Issue #80, opened 2018-09-19
MCD: Log level clean up


ashcrow:
In the logs we are using a combination of default verbosity and 2. However, most, if not all, of our current logging seems like it would be beneficial in normal circumstances. 

## What brings this up

Looking at the log output of https://github.com/openshift/machine-config-operator/issues/79 brought this to mind.

## Proposal

- Use default V for general information useful to admins
- Use `V(2)` for more fine grained items
- Use `V(3)` for debug level logging

## Example

`Daemon.enableUnit` could be updated like:

```golang
// enableUnit enables a systemd unit via symlink
func (dn *Daemon) enableUnit(unit ignv2_2types.Unit) error {
	// The link location
	wantsPath := filepath.Join(wantsPathSystemd, unit.Name)
	// sanity check that we don't return an error when the link already exists
	if _, err := os.Stat(wantsPath); err == nil {
		glog.Infof("%s already exists. Not making a new symlink", wantsPath)
		return nil
	}
	// The originating file to link
	servicePath := filepath.Join(pathSystemd, unit.Name)
	err := os.Symlink(servicePath, wantsPath)
        if err == nil {
    	    glog.Infof("Enabled %s", unit.Name)
            glog.V(3).Infof("Symlinked %s to %s", servicePath, wantsPath)
        } else {
            glog.Warningf("Unable to enable unit %s: %s", unit.Name, err)
        }
        return err
}
```

/cc @sdemos @jlebon @kikisdeliveryservice

jlebon:
Or maybe just two levels? I.e. the default, and `V(2)` for debugging? The more levels you have, the more you have to think about each print statement you add. Whereas "is this only useful for debugging?" is easier to answer.

ashcrow:
I'd be fine with two levels.

ashcrow:
@kikisdeliveryservice is picking this work up. Thanks @kikisdeliveryservice!

kikisdeliveryservice:
So to confirm, I'm updating functions in daemon/update.go using the default and V(2) as appropriate?

ashcrow:
@kikisdeliveryservice in part yes :-). There are a few things to do to close this issue:

1. Ensure that default (no `V`) use is consistent
2. Ensure that `V(2)` is consistent
3. Add logs for both verbosity to help support folks know what's happening.

Think of the default as `info` level and `V(2)` as `debug` level.

kikisdeliveryservice:
[...]

PR #87, opened 2018-09-20
MCD: Log level clean up (daemon/update.go)

kikisdeliveryservice:
Closes Issue #80

- Adds V(2) logging to /daemon/update.go

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -23,6 +23,8 @@ import (
 func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	var err error
 
+	glog.Info("Updating node with new config")
+
 	// make sure we can actually reconcile this state
 	reconcilable, err := dn.reconcilable(oldConfig, newConfig)
 	if err != nil {
@@ -57,6 +59,7 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	if err != nil {
 		return err
 	}
+	glog.V(2).Infof("Node successfully drained")
 
 	// reboot. this function shouldn't actually return.
 	return dn.reboot()
@@ -70,6 +73,8 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 // directories, links, and systemd units sections of the included ignition
 // config currently.
 func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool, error) {
+	glog.Info("Checking if configs are reconcilable")
+
 	oldIgn := oldConfig.Spec.Config
 	newIgn := newConfig.Spec.Config
 
@@ -134,6 +139,7 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool
 	// we can reconcile any state changes in the systemd section.
 
 	// we made it through all the checks. reconcile away!
+	glog.V(2).Infof("Configs are reconcilable")
 	return true, nil
 }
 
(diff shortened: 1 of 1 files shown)


ashcrow:
It's looking good so far!

pkg/daemon/update.go
```diff
@@ -226,17 +229,27 @@ func (dn *Daemon) enableUnit(unit ignv2_2types.Unit) error {
 	}
 	// The originating file to link
 	servicePath := filepath.Join(pathSystemd, unit.Name)
-	return os.Symlink(servicePath, wantsPath)
+	err := os.Symlink(servicePath, wantsPath)
+	if err == nil {
```
nstielau: I've seen `if err := os.Symlink(servicePath, wantsPath); err == nil {...}` style as well.  Which is idiomatic?

pkg/daemon/update.go
```diff
@@ -176,6 +178,7 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 // it encounters.
 func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) {
 	var path string
+	glog.Info("Deleting stale data")
```
nstielau: Is it too verbose to log each file that is being deleted, or would that be useful information when debugging?

ashcrow: I personally think this is nice. It's letting us know that garbage collection is occurring. If this ends up being called once or twice a minute then I think we should up the `V` level. But for now :+1:

pkg/daemon/update.go
kikisdeliveryservice: @nstielau I've only seen the above when there is not an else statement. But since we are adding an else with V2 level logging declaration to cover both cases (err is nil and err != nil) then declaration--if--else seemed more clear and readable.

pkg/daemon/update.go
kikisdeliveryservice: Ok I'll leave this for now then, but of course am happy to change the logging level/ remove logging as people see fit.

pkg/daemon/update.go
rphillips: other places of the code in this file check the negative case first (!=)... I have seen other golang code do the same thing.

kikisdeliveryservice: @rphillips I'll make this consistent. Good catch. Thanks!

pkg/daemon/update.go
```diff
@@ -56,6 +58,8 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	})
 	if err != nil {
 		return err
+	} else {
+		glog.V(2).Infof("Node successfully drained")
```
sdemos: nit: this doesn't have to be in an else block, since if there is an error we return anyway, so if we get past that we successfully drained. not a big deal but golint complains about it. (it complains about the other instances too, my editor suggests there are 14)

pkg/daemon/update.go
```diff
@@ -226,17 +236,27 @@ func (dn *Daemon) enableUnit(unit ignv2_2types.Unit) error {
...
 	servicePath := filepath.Join(pathSystemd, unit.Name)
-	return os.Symlink(servicePath, wantsPath)
+	err := os.Symlink(servicePath, wantsPath)
+	if err != nil {
+		glog.Warningf("Cannot enable unit %s: %s", unit.Name, err)
+	} else {
+		glog.Infof("Enabled %s", unit.Name)
+		glog.V(2).Infof("Symlinked %s to %s", servicePath, wantsPath)
+	}
+    return err
```
sdemos: nit: `gofmt`. this line has spaces instead of tabs, and there are a couple of other whitespace issues in the file as well.

ashcrow:
@sdemos I'll defer to you on the merge.

sdemos:
oh also are there plans to audit the log messages in `pkg/daemon/daemon.go`? a cursory glance suggests that could use some eyes as well.

pkg/daemon/update.go
kikisdeliveryservice: Thanks! Will update.

ashcrow:
@sdemos I'd say yes, but it could be done in a different PR.

pkg/daemon/update.go

pkg/daemon/update.go

pkg/daemon/update.go

PR #89, opened 2018-09-21
MCD: Check what OS MCD is running on

ashcrow:
This set of work adds:
- a function to check what OS the MCD is running on
- usage of said function on start up and exit if the OS is not supported
- plumbs the `operatingSystem` found at start into `Daemon` for use later

/cc @derekwaynecarr @abhinavdahiya @sjenning @crawford @aaronlevy @rphillips

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -41,6 +41,11 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	// To help debugging, immediately log version
 	glog.Infof("Version: %+v", version.Version)
 
+	operatingSystem, err := daemon.GetHostRunningOS(startOpts.rootMount)
+	if err != nil {
+		glog.Fatalf("Error found when checking operating system: %s", err)
+	}
+
 	if startOpts.nodeName == "" {
 		name, ok := os.LookupEnv("NODE_NAME")
 		if !ok || name == "" {
@@ -67,6 +72,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	daemon, err := daemon.New(
 		startOpts.rootMount,
 		startOpts.nodeName,
+		operatingSystem,
 		cb.MachineConfigClientOrDie(componentName),
 		cb.KubeClientOrDie(componentName),
 	)
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -57,6 +57,14 @@
   version = "v1.0.0"
 
 [[projects]]
+  digest = "1:563d5933a25ebbd7e938e04d633bad92ae8f58cb88d70b55c3918857328376c6"
+  name = "github.com/ashcrow/osrelease"
+  packages = ["."]
+  pruneopts = "NUT"
+  revision = "9b292693c55c791c79825cdce5b7e0b3b6da6e88"
+  version = "v1"
+
+[[projects]]
   digest = "1:aba270497eb2d49f5cba6f4162d524b9a1195a24cbce8be20bf56a0051f47deb"
   name = "github.com/blang/semver"
   packages = ["."]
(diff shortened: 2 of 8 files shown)


ashcrow:
Error is on having a new dependency. I'll check the docs to find the right way to vendor for this repo.

cgwalters:
LGTM (superficial review only though)

ashcrow:
@derekwaynecarr @abhinavdahiya @sjenning @crawford @aaronlevy @rphillips It would be pretty easy to add the result of the OS check to a node annotation. Should I do a follow on for that, should we report it a different way, *or* should we not report it at all for now?

pkg/daemon/util.go
```diff
@@ -0,0 +1,26 @@
+package daemon
```
jlebon: I think other maintainers were leaning towards avoiding `util`-style modules. Maybe just `osrelease.go`?

cmd/machine-config-daemon/start.go
```diff
@@ -41,6 +41,11 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	// To help debugging, immediately log version
 	glog.Infof("Version: %+v", version.Version)
 
+	operatingSystem, err := daemon.GetRunningOS()
```
jlebon: Doesn't this need to happen after the `chroot` though? Otherwise, we'll be parsing the `/etc/os-release` of the container. Alternatively, might be beneficial to have a new e.g. `osrelease.NewAtRoot()` in general.

ashcrow: You're 100% right. Updating.

ashcrow:
@jlebon I made the following changes:

- `Daemon.operatingSystem` is now `Daemon.OperatingSystem` to allow wider access
- `GetRunningOS` is now `GetHostRunningOS` and uses the `rootFS` provided at container start to read the proper `os-release` files. I made this choice rather than moving the call after the chroot as I didn't like how the multistage `Daemon` struct creation looked (IE: make a new struct with `New` so it reads the container files and then populate `OperatingSystem after chroot to read the hosts files`).

jlebon:
Looks sane to me! Have you been able to test this?

ashcrow:
@jlebon I tested the code from a container to ensure it did use the proper os-release location (IE: from the mount) but did not do full testing of the MCD.

EG:

## Test Code
```golang
package main

import (
        "fmt"

        "github.com/openshift/machine-config-operator/pkg/daemon"
)

func main() {
        o, err := daemon.GetHostRunningOS("/rootfs")
        if err != nil {
                panic(err)
        }
        fmt.Printf("%+v", o)
}
```
## Exec
- Copied my hosts os-release to /mnt/etc/os-release changing `fedora` to `rhel`
- Copied `main` to a container and ...
```
$ tree /mnt/
/mnt/
└── etc
    └── os-release

1 directory, 1 file
$ podman run -ti -v /mnt/:/rootfs:ro --entrypoint=/main  localhost/test:latest
RHEL$
```

I also did the same with `fedora` in the os-release file and it returned the error as expected.

aaronlevy:
> It would be pretty easy to add the result of the OS check to a node annotation. Should I do a follow on for that, should we report it a different way, or should we not report it at all for now?

Little late to the discussion - but I'm not sure that MCD should be responsible for labeling the node. We would likely need the label *for* MCD to be schedule to the correct nodes. For example, it would be better to never schedule MCD to a host if it would just be crash-looping on an unsupported OS.

The Kubelet providing OS information would at a glance seem reasonable, but I believe we're trying to move away from kubelet's labeling themselves (security issues like "Hey I'm a master - now send me secrets").

Not sure long term direction we go here. Maybe one option here would be that the Kubelet reports OS data as part of the node.status (but not label), and MCO would then update node labels based on that? Or maybe we could just say "MCD goes everywhere, and sleeps if it's not on a supported host", and then we don't need labels at all...

ashcrow:
@aaronlevy great points. We still need this in the MCD so it can react differently to supported OS's but I agree, reporting makes more sense to happen before MCD itself.

PR #92, opened 2018-09-25
cmd/daemon: simplify OSImageURL querying

jlebon:
The booted OSImageURL is clearly not going to change during the same run
of the daemon, so query it just once at the beginning instead. This
simplifies comparisons later on against new machine configs and avoids
running `rpm-ostree status --json` twice.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -27,6 +27,9 @@ type Daemon struct {
 	// OperatingSystem the operating system the MCD is running on
 	OperatingSystem string
 
+	// bootedOSImageURL is the currently booted URL of the operating system
+	bootedOSImageURL string
+
 	// login client talks to the systemd-logind service for rebooting the
 	// machine
 	loginClient *login1.Conn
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -62,17 +65,21 @@ func New(
 		return nil, fmt.Errorf("Error establishing connection to logind dbus: %v", err)
 	}
 
-	if err := loadNodeAnnotations(kubeClient.CoreV1().Nodes(), nodeName); err != nil {
+	if err = loadNodeAnnotations(kubeClient.CoreV1().Nodes(), nodeName); err != nil {
 		return nil, err
 	}
 
+	osImageURL, osVersion, err := getBootedOSImageURL(rootMount)
```
ashcrow: We may want to start denoting functions that are meant to run pre-chroot.

ashcrow: True. I don't see a reason for my comment to hold this PR up though. It was more of a thought as I was reading through changes.

jlebon:
Yup! :+1:

PR #96, opened 2018-09-26
migrate to kubelet config file

rphillips:
Migrates the kubelet to a kubelet.conf file.

The customer typically had to set configuration settings within the kubelet unit file and pass command-line flags to reconfigure the Kubelet. With this patch, the configuration file will allow a subset of options to be set by the customer.

### DIFF
--- a/templates/_base/worker/units/kubelet.yaml
+++ b/templates/_base/worker/units/kubelet.yaml
@@ -7,32 +7,23 @@ contents: |
 
   [Service]
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
-  Environment=KUBELET_RUNTIME_REQUEST_TIMEOUT=10m
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
   EnvironmentFile=-/etc/kubernetes/kubelet-env
 
   ExecStart=/usr/bin/hyperkube \
       kubelet \
+        --config=/etc/kubernetes/kubelet.conf \
         --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig \
         --kubeconfig=/var/lib/kubelet/kubeconfig \
-        --rotate-certificates \
-        --serialize-image-pulls=false \
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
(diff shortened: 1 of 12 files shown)


templates/_base/worker/units/kubelet.yaml
```diff
@@ -7,32 +7,26 @@ contents: |
...
         --allow-privileged \
         --node-labels=node-role.kubernetes.io/worker \
         --minimum-container-ttl-duration=6m0s \
-        --cluster-dns={{.ClusterDNSIP}} \
-        --cluster-domain=cluster.local \
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider={{cloudProvider .}} \
-        {{.CloudProviderConfig}} \
+        {{.CloudProviderConfig -}} \
+         \
```
abhinavdahiya: why this extra line ?

rphillips: fixed in 0ee9e75.

abhinavdahiya:
/approve

@aaronlevy do you mid taking a quick look

aaronlevy:
To my naive eyes this lgtm. I'm assuming flags that are left over are not represented in the config file?

Might want someone else from pod team / someone more familiar with kubelet mechanics to take a pass - but could always be follow up.

Not strictly related to this, but could also drop `--exit-on-lock-contention` flag -- I really really hope no one is still trying to self-host the kubelet (we might want to get that deprecated/removed upstream -- and I'll apologize to everyone for asking for it in the first place...)

rphillips:
Removed the `--exit-on-lock-contention`. The rest of the flags have not been migrated into the kubelet configuration upstream yet.

aaronlevy:
One last question is that it looks like `serialize-image-pulls` changed from false to true. Just confirming that this is intentional / expected?

rphillips:
@aaronlevy good catch.

fixed and squashed.

aaronlevy:
This has been dragging on (apologies) so I can lgtm, but also just noticed that `--lock-file=/var/run/lock/kubelet.lock` also still exists - which is also part of the self-hosting kubelet behavior we no longer use.

rphillips:
@aaronlevy updated to remove the lock

PR #100, opened 2018-09-28
Turn rpm-ostree code into a client

ashcrow:
This change allows for modification of rpm-ostree related code so
unittesting can occur. Instead of the functionality being spread
through private functions within the daemon package, a structure
which implements DeploymentClient is used and is passed in to
the Daemon at creation.

Part of https://github.com/openshift/machine-config-operator/issues/101

/cc @jlebon @sdemos @kikisdeliveryservice

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -28,7 +28,24 @@ type RpmOstreeDeployment struct {
 	CustomOrigin []string `json:"custom-origin"`
 }
 
-func getBootedDeployment(rootMount string) (*RpmOstreeDeployment, error) {
+// NodeUpdaterClient is an interface describing how to interact with the host
+// around content deployment
+type NodeUpdaterClient interface {
+	GetBootedOSImageURL(string) (string, string, error)
+	RunPivot(string) error
+}
+
+// RpmOstreeClient provides all RpmOstree related methods in one structure.
+// This structure implements DeploymentClient
+type RpmOstreeClient struct{}
+
+// NewNodeUpdaterClient returns a new instance of the default DeploymentClient (RpmOstreeClient)
+func NewNodeUpdaterClient() NodeUpdaterClient {
+	return &RpmOstreeClient{}
+}
+
+// getBootedDeployment returns the current deployment found
+func (r *RpmOstreeClient) getBootedDeployment(rootMount string) (*RpmOstreeDeployment, error) {
 	var rosState RpmOstreeState
 	output, err := RunGetOut("chroot", rootMount, "rpm-ostree", "status", "--json")
 	if err != nil {
@@ -48,9 +65,9 @@ func getBootedDeployment(rootMount string) (*RpmOstreeDeployment, error) {
 	return nil, fmt.Errorf("Not currently booted in a deployment")
 }
 
-// getBootedOSImageURL returns the image URL as well as the OSTree version (for logging)
-func getBootedOSImageURL(rootMount string) (string, string, error) {
-	bootedDeployment, err := getBootedDeployment(rootMount)
+// GetBootedOSImageURL returns the image URL as well as the OSTree version (for logging)
+func (r *RpmOstreeClient) GetBootedOSImageURL(rootMount string) (string, string, error) {
+	bootedDeployment, err := r.getBootedDeployment(rootMount)
 	if err != nil {
 		return "", "", err
 	}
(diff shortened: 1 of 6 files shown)


ashcrow:
/hold

I'd like to wait on merging things like this until after the current round of testing.

kikisdeliveryservice:
I'm really excited about these PRs (and future unit testing). Thanks for getting this started!

ashcrow:
Deployment is also an `rpm-ostree` component. I fear there will be more collisions in names like this. I am, however, good with prefixing with something like `NodeDeploymentClient` or `HostDeploymentClient`.

jlebon:
Hmm yeah, I think that makes sense overall for the goal of unit testing. That said for upstream CI, hopefully we can get a real cluster going and test there?

> I am, however, good with prefixing with something like `NodeDeploymentClient` or `HostDeploymentClient`.

Or maybe something like `NodeUpdater`?

ashcrow:
> Or maybe something like NodeUpdater?

I'm :+1: to that. I'll make the change in a bit.

ashcrow:
> Hmm yeah, I think that makes sense overall for the goal of unit testing. That said for upstream CI, hopefully we can get a real cluster going and test there?

Agreed, but that's a different set of work :-)

ashcrow:
Updated.

pkg/daemon/rpm-ostree.go
```diff
@@ -28,7 +28,25 @@ type RpmOstreeDeployment struct {
...
+// This structure implements DeploymentClient
+type RpmOstreeClient struct{}
+
+// NewNodeUpdaterClient returns a new instance of the default DeploymentClient (RpmOstreeClient)
+func NewNodeUpdaterClient() *RpmOstreeClient {
+	return &RpmOstreeClient{}
+}
+
+// GetBootedDeployment returns the current deployment found
+func (r *RpmOstreeClient) GetBootedDeployment(rootMount string) (*RpmOstreeDeployment, error) {
```
sdemos: is this function called by anything external to this package? I only see the one change. if not, I would probably advocate for removing it from the client interface and just having it as a method on `RpmOstreeClient`.

ashcrow: /cc @jlebon

ashcrow: Updated.

jlebon:
Code looks sane to me!

> Should test implementations be in the same file as the main implementation or should they be implemented within a $NAME_test.go with the understanding other $NAME_test.gos will use them as well?

I'll let folks more experienced with golang decide.

ashcrow:
In terms of testing usage. This would look something like this (untested, uncompiled, but close :-)):

```golang
testClient := RpmOstreeClientStub{
	GetBootedOSImageURLReturns: []GetBootedOSImageURLReturn{
            OsImageURL: "theurl",
            Version: "some-version",
            Error: nil,
        }
	RunPivotReturns: []error{nil, fmt.Errorf("broken")}
}

// Create an instance that uses the client
// Execute a method that calls GetBootedOSImageURL off the client. It will get the contents past to GetBootedOSImageURLReturns every time it is executed.
// Execute a method that calls RunPivot. The first time it will return no error. The second run and beyond will return an the "broken" error.

```

pkg/daemon/update_test.go
```diff
@@ -0,0 +1,62 @@
...
+		GetBootedOSImageURLReturns: []GetBootedOSImageURLReturn{},
+		RunPivotReturns: []error{
+			// First run will return no error
+			nil,
+			// Second rrun will return our expected error
+			expectedError},
+	}
+
+	// Create a Daemon instance with mocked clients
+	d := Daemon{
```
ashcrow: I would expect creating a test `Daemon` instance would likely be added as well to keep many of the tests from recreating the instances each time.

ashcrow:
Moved test code into `*_test.go` and added an example test for `update.go`.

ashcrow:
@sdemos PTAL

ashcrow:
FWIW
```
Running tool: /usr/bin/go test -timeout 30s github.com/openshift/machine-config-operator/pkg/daemon -run ^TestUpdateOS$

ok  	github.com/openshift/machine-config-operator/pkg/daemon	0.007s
Success: Tests passed.
```

sdemos:
just kidding, he is out today, I'll just do it. 

/lgtm

Issue #101, opened 2018-09-28
MCD: Move sets of functionality with side effects to clients to allow for unit testing


ashcrow:
In https://github.com/openshift/machine-config-operator/pull/33 we identified that a major blocker for unittesting the MCD are all the side effects that occur from internal functions. The best way we came up with for getting around this limitation is to move things which cause side effects into clients. This issue is a tracker to link to as more of our items move over.

cgwalters:
For rpm-ostree we end up not having a whole lot of unit tests and heavily lean on "integration tests" which inject the built code into a VM.   We're slowly adding some more unit tests since Rust makes it trivial but still...

I haven't hacked on this code really much myself yet, but I imagine we'd like to have a fairly fast iterative flow that tests something more "real" with live Ignition, rpm-ostree etc.

runcom:
now that we have e2e properly setup, let's go e2e for these kinds of testing as the abstraction just adds complexity (and has some bug as well)

PR #102, opened 2018-09-28
daemon/fsclient: Wraps file system operations

ashcrow:
Part of #101 

/cc @jlebon @sdemos @kikisdeliveryservice

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -76,6 +76,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		daemon.NewNodeUpdaterClient(),
 		cb.MachineConfigClientOrDie(componentName),
 		cb.KubeClientOrDie(componentName),
+		daemon.NewFileSystemClient(),
 	)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ashcrow:
/hold

Holding until after the current round of testing.

ashcrow:
Updated:

- Cleaned up some confusing code
- Added mock
- Wired client into the daemon

ashcrow:
@kikisdeliveryservice I think this will be the next need to be able to do testing of your PRs code path. Since `Daemon.update()` (and likely whatever ends up writing SSH keys) attempts to do operating system operations this mock would be able to intercept and return to cover different scenarios.

kikisdeliveryservice:
@ashcrow sounds good, will take a look. thank you!

ashcrow:
```
operator Instantiate the template e2e-aws into the operator namespace and wait for any pods to complete 50m54s
```

Totally unrelated.

ashcrow:
/hold cancel

Originally I was going to write an example test with it but since it's so low level and requires other stubs in combination I'm dropping the hold.

PR #103, opened 2018-10-01
Client for Executing Commands

ashcrow:
Requires #100 

Added a test as an example. The stub probably should live in it's own file.

Note: The 1st commit is #100.

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -73,6 +73,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		startOpts.rootMount,
 		startOpts.nodeName,
 		operatingSystem,
+		daemon.NewNodeUpdaterClient(),
 		cb.MachineConfigClientOrDie(componentName),
 		cb.KubeClientOrDie(componentName),
 	)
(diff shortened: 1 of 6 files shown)


jlebon:
Hmm, given how thin the rpm-ostree wrapper is, I'm not sure if it's worth adding another abstration layer here too. I.e. we're more interested with how the rest of the MCD interacts with the `NodeUpdateClient` (i.e. #100) than how rpm-ostree deserializes JSON, right?

ashcrow:
We're not really interested in how `rpm-ostree` deserializes json so much as we're interested in that the functions/methods react properly when results are produced. For example, with `TestGetBootedDeployment` we would need to continue adding to the test to ensure that the deployment matching (or lack there of) produces the proper results. If a change is made and the test starts failing we know that the internal contract is broken and we must refactor the code to honor the contract, or change the contract by updating the test.

ashcrow:
But it's still a valid note @jlebon. We could alternatively not worry about the `NodeUpdateClient` for testing, but instead test from that level above with by stubbing the `NodeUpdateClient` in a similar way shown with the stub in this PR.

ashcrow:
Let's keep this on hold and revisit it after #100 and others. This PR serves as an example of how to use the clients for testing, but it doesn't have nearly as much value as the other PRs for clients and testing.

PR #104, opened 2018-10-01
templates: change etcd-member to notify type

crawford:
Now that podman is being used, we can actually use systemd's notify
mechanism. This allows systemd to more accurately track etcd's state and
it removes the following warning:

    etcdmain: forgot to set Type=notify in systemd service file?


abhinavdahiya:
@crawford can you run `go test ./pkg/controller/template -u` to update the golden files?

PR #107, opened 2018-10-03
kubelet: fix the clusterDNS property within the config

rphillips:
case matters in yaml and kubelet configuration. clusterDNS was not being propagated into the config.

/cc @abhinavdahiya


wking:
https://github.com/kubernetes/kubernetes/blob/3a8a7114fa22bd687302d31548bb393e2de9f0cd/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L286

PR #108, opened 2018-10-03
sign for the local etcd service name that cluster-network apiservers use

deads2k:
the clusternetwork apiservers like openshift-apiserver use the `etcd.kube-system.svc` to get at etcd.  The cert needs to be signed for that name

/assign @abhinavdahiya


abhinavdahiya:
:P I can push to @deads2k PR branch

PR #109, opened 2018-10-03
daemon: restructure the process function

### LINKED ISSUE #97: MCD doesn't always fully apply new configs
The [top-level logic of the daemon](https://github.com/openshift/machine-config-operator/blob/3a73b1d142eb1efb3fad5b160915e6ddf3d38311/pkg/daemon/daemon.go#L104) currently goes something like this -
* daemon starts up
* check if the machine is in the desired state, as defined by the desiredConfig annotation. this call returns true if currentConfig == desiredConfig [1], and if it doesn't it checks to make sure all the files, systemd units, and and the operating system are all what we want them to be.
  * if it's not, make sure the config is reconcilable, update, and reboot
  * if it is, set the update as done and make the machine schedulable
* wait for the desiredConfig annotation to change
  * when it does, jump back to the "checking machine state" step and continue

this logic is the source of the bug. if the desired machine config contains _only_ irreconcilable differences between it and the current config, the update is never attempted, because the validation logic only validates that all the _reconcilable_ changes are the same. it should be setting the node as degraded. it's also a time-bomb; if someone later makes a valid, reconcilable change, we attempt to apply the changes and only then does everything get rejected and the node set to degraded.

there are three ways I can see to fix this issue -
1. check that a machine is reconcilable every time the desired annotation changes, before validating machine state. this would also happen on startup, because it's entirely possible that while we weren't running the desiredConfig annotation was updated.
2. update the validation logic so that it checks everything defined by ignition regardless of whether we can reconcile it or not. 
3. only validate the node state either when a machine starts up or when we apply an update that we decided doesn't need a reboot (that doesn't actually ever happen right now but the logic is written with that possibility in mind for the future). the node validation logic acts more explicitly as a gate for marking the update as complete. we always attempt to apply config changes when told to, whether or not they seem to represent the same machine state we are already in.

option 1 lets us reject configs we can't reconcile (fixing this problem). it also seems like it may reduce the number of reboots, if, say, we are told to update to a config that we _can_ reconcile, but it turns out that we don't actually need to make any changes to the machine (although, of the top of my head, I can't think of any examples of when this would happen that aren't bugs in the validation logic). the problem with this approach is that it relies much more heavily on our validation logic being totally complete and correct, since if it thinks that a machine is in the correct state, it never goes through the update application logic.

option 2 seems like it would be a lot of work, and it seems like unecessary work, making the assumption that we are the only people who can modify a node under normal operation (plus, even if the node is modified, arguably it's better to just leave that modification be then wipe everything away and start over, since presumably there is a reason for the change).

option 3 is essentially in line with the logic I've always had in mind for the node-agent, now mcd. we rely on the mcc to only tell us to update to a new configuration if there is actually a difference (which it does right now) for minimizing our reboots. if there are subtle side-effects of the update application that we don't account for when validating the machine state, they still happen. this is the option I'm leaning towards.

[1]: I don't think that this should be in the validation logic in general, but I'm going to be opening a separate bug for the issue of removing it since it's not really related to this bug.

/cc @ashcrow @jlebon @kikisdeliveryservice

sdemos:
there was a bug where config changes where the only difference was
unreconcilable the node wouldn't be marked as degraded. it happened
because the update would not be applied if the node verification was
successful, and the verification logic only checks things it
understands, and skips unreconcilable stuff.

this restructuring of the top level logic sets up the process function
to verify the machine every time the mcd starts, and after a successful
update that didn't require a reboot, but only then. essentially this
means verification happens after every update. however, unlike before,
the update is _always_ attempted regardless of whether or not the
machine seems like it is in the correct state. the update logic is
responsible for checking if a config is reconcilable, so that is still
called first.

fixes #97 

/cc: @ashcrow @jlebon

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -93,45 +93,61 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
 	return nil
 }
 
-// process starts the main loop that actually does all the work.
+// process is the main loop that actually does all the work. the flow goes
+// something like this -
+// 1. we restarted for some reason. the happy path reason we restarted is
+//    because of a machine reboot. validate the current machine state is the
+//    desired machine state. if we aren't try updating again. if we are, update
+//    the current state annotation accordingly.
+// 2. watch the desired config annotation, waiting for an update to be
+//    requested by the controller.
+// 3. if an update is requested by the controller, we assume that that means
+//    something changed and apply the desired config no matter what.
+// 4. the update function doesn't return right now, but at some point in the
+//    future if a reboot isn't required for an update it will. if it returns,
+//    validate the machine state and set the update to done.
+//
+// the only reason this function will return is if an error occurs. otherwise it
+// will keep trying to update the machine until it reboots.
 func (dn *Daemon) process() error {
-
-	// do a first pass before entering the main loop
-	if err := dn.syncOnce(); err != nil {
-		return err
-	}
-
 	for {
-		glog.V(2).Infof("Watching for node annotation updates on %q", dn.name)
-		if err := waitUntilUpdate(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
+		// validate machine state
+		isDesired, err := dn.isDesiredMachineState()
+		if err != nil {
 			return err
 		}
-		if err := dn.syncOnce(); err != nil {
-			return err
+
+		if isDesired {
+			// we got the machine state we wanted. set the update complete!
+			if err := dn.completeUpdate(); err != nil {
+				return err
+			}
+
+			// now wait until we need another one.
+			glog.V(2).Infof("Watching for node annotation updates on %q", dn.name)
+			if err := waitUntilUpdate(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
+				return err
+			}
 		}
-	}
-}
 
-// syncOnce only completes once.
-func (dn *Daemon) syncOnce() error {
-	// validate that the machine correctly made it to the target state
-	isDesired, err := dn.isDesiredMachineState()
-	if err != nil {
-		return err
-	}
-
-	if !isDesired {
-		// this currently doesn't return, but may return in the future if the
-		// update could be done without rebooting
+		// either the machine state isn't what we wanted and we should try
+		// again, or the machine state is what we wanted, and now another update
+		// is was triggered.
 		if err := dn.triggerUpdate(); err != nil {
 			return err
 		}
 
+		// we managed to update the machine without rebooting. in this case, we
+		// basically just restart the logic, but working under the assumption
+		// that everything is already initialized for us, so we just go to the
+		// top
 		glog.V(2).Infof("Successfully updated without reboot")
-	} else {
-		glog.V(2).Infof("Node is up to date")
 	}
+}
 
+// completeUpdate does all the stuff required to finish an update. right now, it
+// sets the status annotation to Done and marks the node as schedulable again.
+func (dn *Daemon) completeUpdate() error {
 	if err := setUpdateDone(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


sdemos:
/hold

after playing around with this a bit more I found something I want to dig into before this gets merged in

sdemos:
okay yeah I confirmed - this pr is actually a regression on the current behavior, if there are only irreconcilable changes it successfully recognizes them, but then the daemon stops and restarts, it comes back up again, validates the node (which succeeds for the same reason this change is required), sets the update to done and waits for another change.

it seems like if we figure out what the daemon is supposed to be doing if the node is degraded (it seems like it shouldn't be doing anything? just sleep forever?) that'll fix this issue, which I think is a better way to fix this pr than to restructure the logic here in any other way.

ashcrow:
When degraded it should note to the cluster that it is in that state and continuing it's normal behavior. The node may be destroyed via the cluster and re-provisioned.

pkg/daemon/daemon.go
```diff
@@ -93,45 +93,61 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
-	}
-
 	for {
-		glog.V(2).Infof("Watching for node annotation updates on %q", dn.name)
-		if err := waitUntilUpdate(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
-			return err
-		}
-		if err := dn.syncOnce(); err != nil {
+		// validate machine state
+		isDesired, err := dn.isDesiredMachineState()
```
jlebon: We still need to enhance `isDesiredMachineState()` to check if the configs are reconcilable, right?

sdemos: I wasn't going to initially do that, but it would also fix the issue with this pr, and it's probably a good idea to do that, since it makes the validation function much more complete.

sdemos:
the daemon is already not going to continue its normal behavior if the node is in a degraded state. it can't apply the config for whatever reason, so it fundamentally can't do what it's supposed to. the current behavior without this pr is to crash loop, particularly if the node is degraded because the config is irreconcilable, since it starts up, tries to apply the config, fails, quits, and starts again. I think the best option is either to sleep forever after starting up and finding a degraded status, or to do something with taints/tolerations to prevent the daemon from being rescheduled again once it quits.

ashcrow:
> so it fundamentally can't do what it's supposed to. 

That could be the case. It depends heavily on the pods running on the node. But I follow your logic.

> I think the best option is either to sleep forever after starting up and finding a degraded status, or to do something with taints/tolerations to prevent the daemon from being rescheduled again once it quits.

I'm :+1: for tainting the node when degraded so that no new pods are deployed to the node. If the node is recycled then the pods will reschedule on their own. By adding in taint logic to the MCD we also can take advantage of that for the SSH access taint work.

sdemos:
> I'm :+1: for tainting the node when degraded so that no new pods are deployed to the node

that was actually not what I intended when I wrote that suggestion, but I like that approach. it also is worth noting that at this point we have drained the node and set it unschedulable already and haven't undone that yet, so we are basically doing a more extreme version of that and marking the node tainted so anything that tolerates NoSchedule also won't get scheduled there anymore.

ashcrow:
@sdemos let folks know when it's ready for a re-review!

sdemos:
/hold cancel

@ashcrow should be good now. the behavior right now is to crashloop if the node is degraded - it keeps trying to update, discovering it can't reconcile the node, and exiting. we can add the tainting in a different pr.

jlebon:
Checking for reconcilability in `isDesiredMachineState()` looks good!

> the behavior right now is to crashloop if the node is degraded - it keeps trying to update, discovering it can't reconcile the node, and exiting. we can add the tainting in a different pr.

Agreed on adding taints in a separate PR. Though... that looks like a pretty indirect way of crash looping. Let's just check right at the start if it's degraded and exit? (This also complements the tainting; i.e. if somehow we are scheduled on a tainted node, we immediately jump ship). Open to doing this in a separate PR as well, though.

sdemos:
I agree that we should probably do the check more explicitly, but I think it would be fine to do in another pr. crashlooping is essentially the behavior that already existed, so I would prefer to keep this pr focused on fixing the issue and deal with it later.

jlebon:
OK, this PR looks good to me. Will let @ashcrow have a look and do the merge.

PR #110, opened 2018-10-04
profile: add KUBECONFIG environment variable

rphillips:
Enables kubectl or oc from a worker or master.


rphillips:
If a customer SSHs to a worker or master they can automatically use kubectl or oc out of the box without having to configure anything. Openshift worked like this as well; oc was configured to be used on the machine after logging into it.

abhinavdahiya:
SSH is optional in openshift/installer. So I don't think we want to support this configuration by default. Users will have ability to add any custom configuration by adding the extra machineconfig.

rphillips:
Leaving it up to the customer to configure the machines does not seem very customer friendly. If there is an ELB, port issue, or some other issue that needs diagnosed then a default experience of using oc/kubectl on the node seems like the best use case.

crawford:
Users should absolutely not be SSHing into their machines. In fact, we've discussed cordoning nodes that have had a user log in. There is no need to make things comfy.

PR #112, opened 2018-10-05
Add OpenStack as a valid platform

flaper87:
I'm adding support for OpenStack to the various operators, controllers, and providers that manage an openshift cluster.

I've started by pushing this patch as it seems to be the first step in the flow. I'm not sure if this PR should depend on others so I'd really appreciate some guidance in the process. :smile:

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -353,6 +353,8 @@ func cloudProvider(cfg renderConfig) (interface{}, error) {
 	switch cfg.Platform {
 	case "aws":
 		return "aws", nil
+	case "openstack":
+		return "openstack", nil
 	}
 	return "", nil
 }
--- a/pkg/controller/template/test_data/templates/openstack/master/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/openstack/master/units/kubelet.service
@@ -1,0 +1,34 @@
+contents: |
+  [Unit]
+  Description=Kubernetes Kubelet
+  Wants=rpc-statd.service
+
+  [Service]
+  ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
+  EnvironmentFile=-/etc/kubernetes/kubelet-workaround
+  EnvironmentFile=-/etc/kubernetes/kubelet-env
+
+  ExecStart=/usr/bin/hyperkube \
+      kubelet \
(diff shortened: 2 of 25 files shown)


pkg/controller/template/render.go
```diff
@@ -43,6 +43,7 @@ const (
 //                    /master/units/kubelet.tmpl
 //                           /files/hostname.tmpl
 //       templates/aws/worker/units/kubelet-dropin.conf.tmpl
+//       templates/openstack/worker/units/kubelet-dropin.conf.tmpl
```
abhinavdahiya: Same only example.

pkg/controller/template/test_data/templates/openstack/master/units/kubelet.service
```diff
@@ -0,0 +1,34 @@
...
+        --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig \
+        --rotate-certificates \
+        --kubeconfig=/var/lib/kubelet/kubeconfig \
+        --container-runtime=remote \
+        --container-runtime-endpoint=/var/run/crio/crio.sock \
+        --allow-privileged \
+        --node-labels=node-role.kubernetes.io/master \
+        --minimum-container-ttl-duration=6m0s \
+        --client-ca-file=/etc/kubernetes/ca.crt \
+        --cloud-provider=openstack \
```
abhinavdahiya: @flaper87 openstack is a valid cloud provider, right?

pkg/operator/assets/bindata.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by go-bindata.
+// Code generated by go-bindata. DO NOT EDIT.
```
abhinavdahiya: Can you drop these changes to bindata.go; they are not required.

pkg/controller/template/test_data/templates/openstack/master/units/kubelet.service
flaper87: it is!

pkg/operator/assets/bindata.go
flaper87: yup, will do! Thanks

abhinavdahiya:
/lgtm
I will keep an eye on installer repo to make sure we don't regress. :)

PR #115, opened 2018-10-05
Update SSH keys via MCD

kikisdeliveryservice:
Allow MCD to update SSH Keys.  

We are only allowing one user named "core" and only updating the field SSHAuthorizedKeys for this user.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -14,11 +14,11 @@ import (
 	"strconv"
 	"time"
 
-	errors "github.com/pkg/errors"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	errors "github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -54,7 +54,7 @@ func replaceFileContentsAtomically(fpath string, b []byte) error {
 func (dn *Daemon) writePendingState(desiredConfig *mcfgv1.MachineConfig) error {
 	t := &pendingConfigState{
 		PendingConfig: desiredConfig.GetName(),
-		BootID: dn.bootID,
+		BootID:        dn.bootID,
 	}
 	b, err := json.Marshal(t)
 	if err != nil {
@@ -173,27 +173,38 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 
 	// Passwd section
 
-	// we don't currently configure groups or users in place. we can't fix it if
-	// something changed here.
+	// we don't currently configure Groups in place. we don't configure Users except
+	// for setting/updating SSHAuthorizedKeys for the only allowed user "core".
+	// otherwise we can't fix it if something changed here.
 	if !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd) {
 		if !reflect.DeepEqual(oldIgn.Passwd.Groups, newIgn.Passwd.Groups) {
 			msg := "Ignition Passwd Groups section contains changes"
 			return &msg
 		}
-		// check if the prior config is empty and that this is the first time running.
-		// if so, the SSHKey from the cluster config and user "core" must be added to machine config,.
 		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
+			// check if the prior config is empty and that this is the first time running.
+			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
 			if len(oldIgn.Passwd.Users) == 0 && len(newIgn.Passwd.Users) == 1 {
-				if newIgn.Passwd.Users[0].Name == "core" && len(newIgn.Passwd.Users[0].SSHAuthorizedKeys) > 0 {
-					glog.Info("SSH Keys reconcilable")
-				} else {
-					msg := "Ignition passwd user section contains unsupported changes"
+				glog.Infof("user data sent to verify before ssh init: %v", newIgn.Passwd.Users[0])
+				msg := verifyUserFields(newIgn.Passwd.Users[0])
+				if msg != "" {
+					return &msg
+				}
+			} else if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) >= 1 {
+				// there is an update to Users, we must verify that it is ONLY making an acceptable
+				// change to the SSHAuthorizedKeys for the user "core"
+				for _, user := range newIgn.Passwd.Users {
+					if user.Name != "core" {
+						msg := "Ignition passwd user section contains unsupported changes: non-core user"
+						return &msg
+					}
+				}
+				glog.Infof("user data to be verified before ssh update: %v", newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
+				msg := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
+				if msg != "" {
 					return &msg
 				}
 			}
-		} else {
-			msg := "Ignition passwd section contains unsupported changes"
-			return &msg
 		}
 	}
 
@@ -231,6 +242,29 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 	// we made it through all the checks. reconcile away!
 	glog.V(2).Infof("Configs are reconcilable")
 	return nil
+}
+
+// verifyUserFields returns nil if the user Name = "core", if 1 or more SSHKeys exist for
+// this user and if all other fields in User are empty.
+// Otherwise, an error msg will be returned and the proposed config will not be reconcilable.
+// At this time we do not support non-"core" users or any changes to the "core" user
+// outside of SSHAuthorizedKeys.
+func verifyUserFields(pwdUser ignv2_2types.PasswdUser) string {
+	var msg string
+	emptyUser := ignv2_2types.PasswdUser{}
+	tempUser := pwdUser
+	if tempUser.Name == "core" && len(tempUser.SSHAuthorizedKeys) >= 1 {
+		tempUser.Name = ""
+		tempUser.SSHAuthorizedKeys = nil
+		if !reflect.DeepEqual(emptyUser, tempUser) {
+			msg = "Ignition passwd user section contains unsupported changes: non-sshKey changes"
+		} else {
+			glog.Info("SSH Keys reconcilable")
+		}
+	} else {
+		msg = "Ignition passwd user section contains unsupported changes: user must be core and have 1 or more sshKeys"
+	}
+	return msg
 }
 
 // updateFiles writes files specified by the nodeconfig to disk. it also writes
@@ -536,7 +570,7 @@ func (dn *Daemon) updateSSHKeys(newUsers []ignv2_2types.PasswdUser) error {
 	}
 	sshDirPath := filepath.Join("/home", newUsers[0].Name, ".ssh")
 	// we are only dealing with the "core" User at this time, so only dealing with the first entry in Users[]
-	glog.Infof("Writing SSHKeys at %q:", sshDirPath)
+	glog.Infof("Writing SSHKeys at %q", sshDirPath)
 	if err := dn.fileSystemClient.MkdirAll(filepath.Dir(sshDirPath), os.FileMode(0600)); err != nil {
 		return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(sshDirPath), err)
 	}
@@ -544,7 +578,7 @@ func (dn *Daemon) updateSSHKeys(newUsers []ignv2_2types.PasswdUser) error {
 
 	authkeypath := filepath.Join(sshDirPath, "authorized_keys")
 	var concatSSHKeys string
-	for _, k := range newUsers[0].SSHAuthorizedKeys {
+	for _, k := range newUsers[len(newUsers)-1].SSHAuthorizedKeys {
 		concatSSHKeys = concatSSHKeys + string(k) + "\n"
 	}
 
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -65,22 +65,6 @@ func TestUpdateOS(t *testing.T) {
 // TestReconcilable attempts to verify the conditions in which configs would and would not be
 // reconcilable. Welcome to the longest unittest you've ever read.
 func TestReconcilable(t *testing.T) {
-	// checkReconcilableResults is a shortcut for verifying results that should be reconcilable
-	checkReconcilableResults := func(key string, reconcilableError *string) {
-
-		if reconcilableError != nil {
-			t.Errorf("Expected the same %s values would be reconcilable. Received error: %v", key, *reconcilableError)
-		}
-	}
-
-	// checkIreconcilableResults is a shortcut for verifing results that should be ireconcilable
-	checkIreconcilableResults := func(key string, reconcilableError *string) {
-
-		if reconcilableError == nil {
-			t.Errorf("Expected different %s values would not be reconcilable.", key)
-		}
-	}
-
 	d := Daemon{
 		name:              "nodeName",
 		OperatingSystem:   MachineConfigDaemonOSRHCOS,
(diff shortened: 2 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -100,17 +104,58 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool
...
+						break
+					}
+				} else {
+					// SSH keys match, nothing to do on this
+					// prob unnecessary since defaul bool value is false
+					onlySSHChanged = false
+					break
+				}
+			}
+			// length of Users is different, we don't do those kinds of changes, so nope.
```
ashcrow: I'm stuck between an attempted addition of a new user should noop (as it's doing here) or if we should raise the irreconcilable flag as the user was attempting to do something we are not going to do. @sdemos thoughts?

pkg/daemon/update.go
```diff
@@ -100,17 +104,58 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool
...
 	if !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd) {
-		glog.Warningf("daemon can't reconcile state!")
-		glog.Warningf("Ignition passwd section contains changes")
-		return false, nil
+		// MCD will handle only ignition SSH keys
+		// This section will check (using some tbd magic) to see if the ONLY diff between 2
+		// Passwd configs is an SSHAuthorizedKey change
+		// Any change other than an SSHChanage will return false.
+
+		var onlySSHChanged bool
```
ashcrow: nit: It is true that bools are false by default ingolang, but `onlySSHChanged := false` would be a bit more clear for a casual reader.

pkg/daemon/update.go
```diff
@@ -100,17 +104,58 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool
...
+				} else {
+					// SSH keys match, nothing to do on this
+					// prob unnecessary since defaul bool value is false
+					onlySSHChanged = false
+					break
+				}
+			}
+			// length of Users is different, we don't do those kinds of changes, so nope.
+		} else {
+			// prob unnecessary since defaul bool value is false
```
ashcrow: Agreed. This else can be dropped, though it's not a review blocker.

pkg/daemon/update.go
```diff
@@ -176,6 +221,19 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	return nil
 }
 
+//updatePasswd can be used to update Passwd with changes for Users at indices contained
```
ashcrow: nit: missing space after comment

pkg/daemon/update.go
```diff
@@ -406,6 +464,22 @@ func (dn *Daemon) writeFiles(files []ignv2_2types.File) error {
...
 
+// seperated out writeSSHKeys from updatePasswd even though this is the only update allowed to Passwd right now
+func (dn *Daemon) writeSSHKeys(oldUsers []ignv2_2types.PasswdUser, newUsers []ignv2_2types.PasswdUser, userKeys []int) error {
+	var path string
+	for _, idx := range userKeys {
+		if !reflect.DeepEqual(oldUsers[idx].SSHAuthorizedKeys, newUsers[idx].SSHAuthorizedKeys) {
+			// ssh keys are in PasswdUser.HomeDir/.ssh
+			path = filepath.Join(oldUsers[idx].HomeDir, ".ssh")
+			glog.V(2).Infof("Writing SSHKeys at %q:", path)
+			// write the SSHKeys
```
ashcrow: :+1:

pkg/daemon/update.go
```diff
@@ -406,6 +464,22 @@ func (dn *Daemon) writeFiles(files []ignv2_2types.File) error {
...
+		} else {
+			return fmt.Errorf("Failed to write SSHKeys")
```
ashcrow: Once the ssh key writing is added I recommend returning the error that is bubbled up from that work OR adding a log statement with details before this return

kikisdeliveryservice:
Thanks @ashcrow for your thoughts!  I'll incorporate them into the PR. :smiley_cat:

pkg/daemon/update.go

pkg/daemon/update.go

pkg/daemon/update.go

pkg/daemon/update.go
```diff
@@ -470,15 +470,31 @@ func (dn *Daemon) writeFiles(files []ignv2_2types.File) error {
...
-			// write the SSHKeys
+			if err := dn.fileSystemClient.MkdirAll(filepath.Dir(path), os.FileMode(0600)); err != nil {
+				return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(path), err)
+			}
+			glog.V(2).Infof("Created directory: %s", path)
+			authkeypath := filepath.Join(path, "authorized_keys")
+
+			var concatSSHKeys string
+			for _, k := range newUsers[idx].SSHAuthorizedKeys {
+				concatSSHKeys = concatSSHKeys + string(k) + "/n"
```
ashcrow: Why `/n` and not `\n`?

kikisdeliveryservice: DOH

pkg/daemon/update.go
```diff
@@ -470,15 +470,31 @@ func (dn *Daemon) writeFiles(files []ignv2_2types.File) error {
...
+			}
+
+			err := ioutil.WriteFile(authkeypath, []byte(concatSSHKeys), os.FileMode(0600))
```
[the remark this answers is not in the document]
kikisdeliveryservice: That seems like it would make sense.  The other funcs in update.go are also using ioutils.WriteFile  so it will be helpful. :)

pkg/daemon/update.go
kikisdeliveryservice: @rphillips Thank you for spotting that!

kikisdeliveryservice: @rphillips ah just remembered that I tried strings.Join before, but it won't work bc SSHAuthorizedKeys -> type []SSHAuthorizedKey -> type SSHAuthorizedKey string

ashcrow:
/retest

Flakes.

kikisdeliveryservice:
Temporarily blocked by:
https://github.com/openshift/installer/issues/578

ashcrow:
Woo! All green!

ashcrow:
Rebase needed.

eparis:
Are we going to be able to get back to this soon? I'm in a fight with @crawford here: https://github.com/openshift/installer/issues/862
I'm not happy shipping without ssh keys without this PR...

ashcrow:
Yes. Kirsten is currently at Kubecon but will be coming back to this soon.

pkg/daemon/update.go
```diff
@@ -121,17 +129,54 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (bool
...
-		glog.Warningf("Ignition passwd section contains changes")
-		return false, nil
+		// MCD will handle only ignition SSH keys
+		// This section will check (using some tbd magic) to see if the ONLY diff between 2
+		// Passwd configs is an SSHAuthorizedKey change
+		// Any change other than an SSHChanage will return false.
+
+		sshReconcilable := false
+		// If the size of list Users is different, a User has been added or deleted -> nope
+		if len(oldIgn.Passwd.Users) == len(newIgn.Passwd.Users) {
```
cgwalters: Probably worth splitting this into a separate `sshConfigsReconcilable()` or so.

pkg/daemon/update_test.go
```diff
@@ -63,28 +62,29 @@ func TestUpdateOS(t *testing.T) {
 	}
 }
 
+// checkReconcilableResults is a shortcut for verifying results that should be reconcilable
+func checkReconcilableResults(t *testing.T, key string, err error, isReconcilable bool) {
```
cgwalters: Could split up these test changes and land them now.

pkg/daemon/update.go
kikisdeliveryservice: This is subject to change once my other PR lands.

kikisdeliveryservice:
Reported the tests that are failing due to "cluster-osin-operator". Awaiting new payload getting built.

kikisdeliveryservice:
Trying to get a cluster up to test this.

ashcrow:
Let me know when I should review!

kikisdeliveryservice:
Ran into an Ignition config issue in testing where new configs are appended which hurts the current reconcilable logic. Consulted with @abhinavdahiya and @ajeddeloh and will tweak reconcilable.

kikisdeliveryservice:
I'll pick this back up tomorrow, I'm still seeing some wonky ignition behaviour when it's combining the configs.  I'll update you tomorrow @ashcrow

kikisdeliveryservice:
Manual testing status:
- replace current key for user "core" with a new key via a machineconfig: Success!  Verified that I am able to ssh in with new key, can't ssh in with old key and authorized key file on worker is as predicted.

pkg/daemon/update.go
```diff
@@ -262,6 +239,24 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
...
+	if tempUser.Name == "core" && len(tempUser.SSHAuthorizedKeys) > 1 {
+		tempUser.Name = ""
+		tempUser.SSHAuthorizedKeys = nil
+		if !reflect.DeepEqual(emptyUser, tempUser) {
+			msg = "Ignition passwd user section contains unsupported changes: non-sshKey changes"
+		} else {
+			glog.Info("SSH Keys reconcilable")
+		}
+	} else {
+		msg = "Ignition passwd user section contains unsupported changes: user must be core and have 1 or more sshKeys"
```
cgwalters: Don't need to do in this PR, but it should be supported to entirely remove SSH keys I'd say.

pkg/daemon/update.go
```diff
@@ -184,44 +184,21 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 			// check if the prior config is empty and that this is the first time running.
 			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
 			if len(oldIgn.Passwd.Users) == 0 && len(newIgn.Passwd.Users) == 1 {
```
cgwalters: Offhand, it feels like this "first time running" is actually just for transitioning clusters from *before* this PR to clusters with it (or in the future to support re-installing SSH keys if they've been removed).  Freshly installed clusters *with* this PR and a SSH key supplied should always have `len(oldIgn.Passwd.Users)` right?

pkg/daemon/update.go
```diff
@@ -184,44 +184,21 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
...
-						} else {
-							// there are differences other than just the SSH key
-							msg := "Ignition passwd user section contains unsupported changes"
-							return &msg
-						}
-					} else {
-						msg := "Ignition passwd user section contains unsupported changes"
+			} else if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) >= 1 {
+				for _, user := range newIgn.Passwd.Users {
+					if user.Name != "core" {
```
cgwalters: It'd probably be clearer to explicitly error out if `len(newIgn.Passwd.Users) > 1` since we only support the "core" user, and there can only be one user with a given name so something like:

```
if len(newIgn.Passwd.Users) > 1 {
   msg := "Adding multiple users is not supported"
   return &msg
}
```

If we do that *first*, every following conditional can then be sure the length is either 0 or 1.

cgwalters: Basically all the code after that could actually operate on a `*User` which can be `nil` (in Rust terms `Option<&User>`) and we don't need to think about arrays of users.

IOW after the above code, something like:

```
newUser := nil
oldUser := nil
if len(newIgn.Passwd.Users) > 0 {
    newUser = &newIgn.Passwd.Users[0];
}
if len(newIgn.Passwd.Users) > 0 {
    oldUser = &oldIgn.Passwd.Users[0];
}
```

And then instead of doing `newUsers[0].Name` it'd just be `newUser.Name`.

cgwalters:
(Though to be clear I'm also OK to `lgtm` these changes as is, particularly since they've been tested; can always do follow up work later)

pkg/daemon/update.go
kikisdeliveryservice: Nope the first time running part is used from my first SSH PR where we are passing the SSH Keys from the cluster config thru to the MCD.

pkg/daemon/update.go
```diff
@@ -180,20 +180,30 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 			msg := "Ignition Passwd Groups section contains changes"
 			return &msg
 		}
-		// check if the prior config is empty and that this is the first time running.
-		// if so, the SSHKey from the cluster config and user "core" must be added to machine config,.
 		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
+			// check if the prior config is empty and that this is the first time running.
+			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
+			glog.Infof("oldUsers: %v", oldIgn.Passwd.Users)
```
ashcrow: Thank you for logging.

pkg/daemon/update.go
```diff
@@ -233,6 +243,26 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 	return nil
 }
 
+func verifyUserFields(pwdUser ignv2_2types.PasswdUser) string {
```
ashcrow: If you end up making changes to this PR please add a doc to this function. While it's not required being an internal function, it's still helpful for future developers.

pkg/daemon/update.go
kikisdeliveryservice: I can double check on this req after this PR and follow-on, thanks!

pkg/daemon/update.go
kikisdeliveryservice: You would think this would work and this suggestion is closer to how this PR was initially structured _before_ I discovered Ignition's quirks :sob:  

Short explanation: the way Ignition works the newUsers array is actually the compilation of all the Users from all of the machineconfigs in `oc get machineconfigs`. What ignition does is only _apply_ the last one. So that the newUsers array will likely always be greater than 1, except for the first time this runs when getting the initial SSH Key from the installer.

pkg/daemon/update.go

pkg/daemon/update.go
kikisdeliveryservice: retained the logging that made most sense looking at my cluster. we can add more/less in a follow-on if necessary :)

kikisdeliveryservice:
Passed manual testing, only changes made since last review:
- removed some of my logging that was added for testing
- added additional comments for clarity
- squashed commits

pkg/daemon/update.go
```diff
@@ -173,27 +173,38 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
...
+				// there is an update to Users, we must verify that it is ONLY making an acceptable
+				// change to the SSHAuthorizedKeys for the user "core"
+				for _, user := range newIgn.Passwd.Users {
+					if user.Name != "core" {
+						msg := "Ignition passwd user section contains unsupported changes: non-core user"
+						return &msg
+					}
+				}
+				glog.Infof("user data to be verified before ssh update: %v", newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
+				msg := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
```
cgwalters: Should this be `verifyUserFields(user)`?

kikisdeliveryservice: no, we are taking the last user in users in line with ignition logic.

pkg/daemon/update.go
```diff
@@ -233,6 +244,29 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 	return nil
 }
 
+// verifyUserFields returns nil if the user Name = "core", if 1 or more SSHKeys exist for
```
ashcrow: Thank you!

pkg/daemon/update.go
cgwalters: Wait sorry, I missed this was outside the `for` loop.

ashcrow:
Looks like @cgwalters reviewed at the same time.

/lgtm

ashcrow:
Infra failure

```
could not wait for build: the build machine-config-controller failed with reason PullBuilderImageFailed: Failed pulling builder image
```

sjenning:
We are hitting the same image build failures in https://github.com/openshift/machine-config-operator/pull/249

kikisdeliveryservice:
My tests finally all passed and now they are rerunning. :frowning_face:

ashcrow:
CI MERGED! :tada:

pkg/daemon/update.go
kikisdeliveryservice: I cleaned up this section in #293

PR #118, opened 2018-10-09
templates: add flannel-dns-hack.service for masters

abhinavdahiya:
when pod requesting dns resolution is located on the same node as the dns pod,
the dns query fails with error
```console
nslookup test1-etcd-0.tt.testing
;; reply from unexpected source: 10.2.0.6#53, expected 10.3.0.10#53
```

@rchopra found that the cause is flannel using the bridge to route pod traffic and
the bridge does not participate in iptables if routed locally. The fix is to force
the cni0 bridge to use iptables.

https://github.com/openshift/installer/issues/440
/cc @crawford


crawford:
This is no install section in that service so it doesn't run.

PR #121, opened 2018-10-10
daemon: exit early if node is degraded

jlebon:
As discussed in #109, if the node is marked degraded, then we don't want
the MCD to do anything at all. As a first step, just check for this on
start and exit right away if that's the case. Down the line, we should
investigate tainting the node to make sure we don't get scheduled in the
first place as was discussed in #109.


jlebon:
This one looks ready for review; all tests passing! :heavy_check_mark:

PR #124, opened 2018-10-10
daemon: also check file mode when comparing

jlebon:
Otherwise we'll think we successfully reconciled even if the modes
differ.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"strings"
+	"os"
 
 	"github.com/coreos/go-systemd/login1"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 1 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -19,6 +19,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	DefaultDirectoryPermissions os.FileMode = 0755
```
sdemos: it doesn't look like this is used anywhere?

pkg/daemon/daemon.go
```diff
@@ -322,7 +327,16 @@ func (dn *Daemon) checkFiles(files []ignv2_2types.File) bool {
 // checkFileContents reads the file from the filepath and compares its contents
 // with the  expectedContent. It logs an error in case of an error or mismatch
 // and returns the status of the evaluation.
-func checkFileContents(filePath, expectedContent string) bool {
+func checkFileContentsAndMode(filePath, expectedContent string, mode os.FileMode) bool {
+	fi, err := os.Lstat(filePath)
+	if err != nil {
+		glog.Errorf("could not read file: %q, error: %v", filePath, err)
```
sdemos: might be better if it's `could not stat file` or `could not get file info` or something like that.

pkg/daemon/daemon.go
```diff
@@ -322,7 +327,16 @@ func (dn *Daemon) checkFiles(files []ignv2_2types.File) bool {
 // checkFileContents reads the file from the filepath and compares its contents
```
sdemos: should update doc comment with new name and additional info on file mode handling

pkg/daemon/update.go
jlebon: It does now! :) I added a commit to make sure we match Ignition there.

sdemos:
woah, wrecked.

/retest

PR #126, opened 2018-10-15
daemon: Run once mode

ashcrow:
Non RHCOS nodes will need to apply an MC once and exit.

Requires: https://github.com/openshift/machine-config-operator/pull/139

/cc @aaronlevy @dustymabe @sdodson 

Still todo:
- [x] Rebase on #130 
- [x] Update remote http(s) `onceFrom` to request resources from the cluster
- [x] Update local ignition file `onceFrom` to be able to run without making requests for state from the cluster
- [x] Ensure kubernetes client doesn't attempt connection when there is no cluster

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -24,6 +24,7 @@ var (
 		kubeconfig string
 		nodeName   string
 		rootMount  string
+		onceFrom   string
 	}
 )
 
@@ -32,6 +33,7 @@ func init() {
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
 	startCmd.PersistentFlags().StringVar(&startOpts.nodeName, "node-name", "", "kubernetes node name daemon is managing.")
 	startCmd.PersistentFlags().StringVar(&startOpts.rootMount, "root-mount", "/rootfs", "where the nodes root filesystem is mounted for chroot and file manipulation.")
+	startCmd.PersistentFlags().StringVar(&startOpts.onceFrom, "once-from", "", "Runs the daemon once using a provided file path or URL endpoint as its machine config source")
 }
 
 func runStartCmd(cmd *cobra.Command, args []string) {
@@ -82,6 +84,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		cb.KubeClientOrDie(componentName),
 		daemon.NewFileSystemClient(),
 		ctx.KubeInformerFactory.Core().V1().Nodes(),
+		startOpts.onceFrom,
 	)
 	if err != nil {
 		glog.Fatalf("failed to initialize daemon: %v", err)
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -105,6 +105,9 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		if err != nil {
 			glog.Fatalf("error checking initial state of node: %v", err)
 		}
+		if err = dn.StartInformer(stopCh, startOpts.nodeName, componentName, startOpts.kubeconfig); err != nil {
+			glog.Fatalf("error starting kubernetes informers: %v", err)
+		}
 	}
 
 	glog.Infof(`Calling chroot("%s")`, startOpts.rootMount)
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -6,6 +6,7 @@ import (
 	"syscall"
 
 	"github.com/golang/glog"
+	"github.com/openshift/machine-config-operator/cmd/common"
 	"github.com/openshift/machine-config-operator/pkg/daemon"
 	"github.com/openshift/machine-config-operator/pkg/version"
 	"github.com/spf13/cobra"
@@ -68,6 +69,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	stopCh := make(chan struct{})
 	defer close(stopCh)
 	var dn *daemon.Daemon
+	var ctx *common.ControllerContext
 
 	// If we are asked to run once and it's a valid file system path use
 	// the bare Daemon
@@ -85,6 +87,11 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		}
 		// Else we use the cluster driven daemon
 	} else {
+		cb, err := common.NewClientBuilder(startOpts.kubeconfig)
+		if err != nil {
+			glog.Fatalf("failed to initialize daemon: %v", err)
+		}
+		ctx = common.CreateControllerContext(cb, stopCh, componentName)
 		// create the daemon instance. this also initializes kube client items
 		// which need to come from the container and not the chroot.
 		dn, err = daemon.NewClusterDrivenDaemon(
@@ -92,21 +99,14 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 			startOpts.nodeName,
 			operatingSystem,
 			daemon.NewNodeUpdaterClient(),
-			startOpts.kubeconfig,
+			cb.MachineConfigClientOrDie(componentName),
+			cb.KubeClientOrDie(componentName),
 			daemon.NewFileSystemClient(),
 			startOpts.onceFrom,
-			stopCh,
-			componentName,
+			ctx.KubeInformerFactory.Core().V1().Nodes(),
 		)
 		if err != nil {
 			glog.Fatalf("failed to initialize daemon: %v", err)
-		}
-		err = dn.CheckStateOnBoot(stopCh)
-		if err != nil {
-			glog.Fatalf("error checking initial state of node: %v", err)
-		}
-		if err = dn.StartInformer(stopCh, startOpts.nodeName, componentName, startOpts.kubeconfig); err != nil {
-			glog.Fatalf("error starting kubernetes informers: %v", err)
 		}
 	}
 
@@ -118,6 +118,15 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	glog.V(2).Infof("Moving to / inside the chroot")
 	if err := os.Chdir("/"); err != nil {
 		glog.Fatalf("unable to change directory to /: %s", err)
+	}
+
+	if startOpts.onceFrom == "" {
+		err = dn.CheckStateOnBoot(stopCh)
+		if err != nil {
+			glog.Fatalf("error checking initial state of node: %v", err)
+		}
+		ctx.KubeInformerFactory.Start(ctx.Stop)
+		close(ctx.KubeInformersStarted)
 	}
 
 	glog.Info("Starting MachineConfigDaemon")
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -6,7 +6,6 @@ import (
 	"syscall"
 
 	"github.com/golang/glog"
-	"github.com/openshift/machine-config-operator/cmd/common"
 	"github.com/openshift/machine-config-operator/pkg/daemon"
 	"github.com/openshift/machine-config-operator/pkg/version"
 	"github.com/spf13/cobra"
@@ -40,6 +39,8 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
+	glog.V(2).Infof("Options parsed: %+v", startOpts)
+
 	// To help debugging, immediately log version
 	glog.Infof("Version: %+v", version.Version)
 
@@ -56,11 +57,6 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		startOpts.nodeName = name
 	}
 
-	cb, err := common.NewClientBuilder(startOpts.kubeconfig)
-	if err != nil {
-		glog.Fatalf("error creating clients: %v", err)
-	}
-
 	// Ensure that the rootMount exists
 	if _, err := os.Stat(startOpts.rootMount); err != nil {
 		if os.IsNotExist(err) {
@@ -71,23 +67,44 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 
 	stopCh := make(chan struct{})
 	defer close(stopCh)
+	var dn *daemon.Daemon
 
-	ctx := common.CreateControllerContext(cb, stopCh, componentName)
-	// create the daemon instance. this also initializes kube client items
-	// which need to come from the container and not the chroot.
-	daemon, err := daemon.New(
-		startOpts.rootMount,
-		startOpts.nodeName,
-		operatingSystem,
-		daemon.NewNodeUpdaterClient(),
-		cb.MachineConfigClientOrDie(componentName),
-		cb.KubeClientOrDie(componentName),
-		daemon.NewFileSystemClient(),
-		ctx.KubeInformerFactory.Core().V1().Nodes(),
-		startOpts.onceFrom,
-	)
-	if err != nil {
-		glog.Fatalf("failed to initialize daemon: %v", err)
+	// If we are asked to run once and it's a valid file system path use
+	// the bare Daemon
+	if startOpts.onceFrom != "" && daemon.ValidPath(startOpts.onceFrom) {
+		dn, err = daemon.New(
+			startOpts.rootMount,
+			startOpts.nodeName,
+			operatingSystem,
+			daemon.NewNodeUpdaterClient(),
+			daemon.NewFileSystemClient(),
+			startOpts.onceFrom,
+		)
+		if err != nil {
+			glog.Fatalf("failed to initialize single run daemon: %v", err)
+		}
+		// Else we use the cluster driven daemon
+	} else {
+		// create the daemon instance. this also initializes kube client items
+		// which need to come from the container and not the chroot.
+		dn, err = daemon.NewClusterDrivenDaemon(
+			startOpts.rootMount,
+			startOpts.nodeName,
+			operatingSystem,
+			daemon.NewNodeUpdaterClient(),
+			startOpts.kubeconfig,
+			daemon.NewFileSystemClient(),
+			startOpts.onceFrom,
+			stopCh,
+			componentName,
+		)
+		if err != nil {
+			glog.Fatalf("failed to initialize daemon: %v", err)
+		}
+		err = dn.CheckStateOnBoot(stopCh)
+		if err != nil {
+			glog.Fatalf("error checking initial state of node: %v", err)
+		}
 	}
 
 	glog.Infof(`Calling chroot("%s")`, startOpts.rootMount)
@@ -103,15 +120,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	glog.Info("Starting MachineConfigDaemon")
 	defer glog.Info("Shutting down MachineConfigDaemon")
 
-	err = daemon.CheckStateOnBoot(stopCh)
-	if err != nil {
-		glog.Fatalf("error checking initial state of node: %v", err)
-	}
-
-	ctx.KubeInformerFactory.Start(ctx.Stop)
-	close(ctx.KubeInformersStarted)
-
-	err = daemon.Run(stopCh)
+	err = dn.Run(stopCh)
 	if err != nil {
 		glog.Fatalf("failed to run: %v", err)
 	}
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -12,7 +12,6 @@ import (
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
-	"github.com/openshift/machine-config-operator/cmd/common"
 	"github.com/openshift/machine-config-operator/lib/resourceread"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
@@ -20,6 +19,7 @@ import (
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
 	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -113,11 +113,11 @@ func NewClusterDrivenDaemon(
 	nodeName string,
 	operatingSystem string,
 	nodeUpdaterClient NodeUpdaterClient,
-	kubeconfig string,
+	client mcfgclientset.Interface,
+	kubeClient kubernetes.Interface,
 	fileSystemClient FileSystemClient,
 	onceFrom string,
-	stopCh chan (struct{}),
-	componentName string,
+	nodeInformer coreinformersv1.NodeInformer,
 ) (*Daemon, error) {
 	dn, err := New(
 		rootMount,
@@ -131,41 +131,20 @@ func NewClusterDrivenDaemon(
 		return nil, err
 	}
 
-	cb, err := common.NewClientBuilder(kubeconfig)
-	if err != nil {
-		glog.Fatalf("error creating clients: %v", err)
-	}
-
-	dn.kubeClient = cb.KubeClientOrDie(componentName)
-	dn.client = cb.MachineConfigClientOrDie(componentName)
+	dn.kubeClient = kubeClient
+	dn.client = client
 
 	if err = loadNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), nodeName); err != nil {
 		return nil, err
 	}
-	return dn, nil
-}
 
-// StartInformer initializes and starts the informers.
-func (dn *Daemon) StartInformer(stopCh chan (struct{}), nodeName, componentName, kubeconfig string) error {
-	cb, err := common.NewClientBuilder(kubeconfig)
-	if err != nil {
-		return err
-	}
-
-	ctx := common.CreateControllerContext(cb, stopCh, componentName)
-
-	nodeInformer := ctx.KubeInformerFactory.Core().V1().Nodes()
 	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		UpdateFunc: dn.handleNodeUpdate,
 	})
-
 	dn.nodeLister = nodeInformer.Lister()
 	dn.nodeListerSynced = nodeInformer.Informer().HasSynced
 
-	ctx.KubeInformerFactory.Start(ctx.Stop)
-	close(ctx.KubeInformersStarted)
-
-	return nil
+	return dn, nil
 }
 
 // Run finishes informer setup and then blocks, and the informer will be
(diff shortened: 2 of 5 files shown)


ashcrow:
/cc @crawford 

@aaronlevy @sdodson

If we can pin down the format of the remote service I think we can actually write a card and execute on this. If at all possible I'd like the remote file to be the same as the ignition config ... OR the same as the CRD. The fewer things we have to parse to execute the same/similar operations the better IMHO.

aaronlevy:
I would lean toward the format being an ignition config (not CRD). My reasoning for this:

- We will need to generate raw/complete ignition configs for bootstrap/master/worker to support the eventual "bring your own RHCOS" environments. For example, "I already run a DC and have my own PXE infra - I just need the ignition profiles to use. 

- The bootstrap ignition config (largest/most complex) is already generated only as raw ignition and served from S3 - so we'd be wrapping it in a CRD just to immediately unwrap it anyway.

If we don't already generate ignition configs for master/worker (and instead just their CRD forms) - we likely should add them as an output format of the installer asset generation phase (cc @abhinavdahiya )

ashcrow:
@aaronlevy the MCD ends up consuming MCs (machine config) which wraps an ignition config with more metadata used for defining what version the host should be.

ashcrow:
> So would this seem reasonable to everyone:
>
>    - installer generates bootstrap machineConfig CR (in addition to raw ignition config)

:+1: 

>    - MCD can execute a run-once mode either from api endpoint (machineConfig CR) or locally (machineConfig CR)

That would sound reasonable to me. To reiterate to make sure I'm understanding properly, the api endpoint (full URI with scheme) would be passed to `--run-from` *or* a full file system path (/something/like/this/$FILENAME) would be passed to `--run-from`. In either case the same parsing mechanics used when watching CRD endpoints would be utilized, the only difference is it would parse and execute once, and then exit.

ashcrow:
If my reiteration above is correct I think we have enough to groom a card and do this work.

/cc @aaronlevy @sdodson @crawford

aaronlevy:
That sounds good to me. FWIW - naming of flags and such I have no strong feelings about

ashcrow:
OK cool. I've created a card based on this and we'll start filling it in soon.

ashcrow:
@sdemos When you have a few PTAL at what I have so far before I start wiring up a single process mode.

pkg/daemon/daemon.go
```diff
@@ -102,6 +109,25 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
 	glog.Info("Starting MachineConfigDaemon")
 	defer glog.Info("Shutting down MachineConfigDaemon")
 
+	// Catch quickly if we've been asked to run once.
+	if dn.onceFrom != "" {
+		glog.V(2).Info("Running once per request")
+		// If we sense a local file has been provided ...
+		switch {
+		case strings.HasPrefix(string(os.PathSeparator), dn.onceFrom):
```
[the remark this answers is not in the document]
ashcrow: This does mean absolute. We could support relative ... I can add that in my next update.

cmd/machine-config-daemon/start.go
```diff
@@ -32,6 +33,7 @@ func init() {
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
 	startCmd.PersistentFlags().StringVar(&startOpts.nodeName, "node-name", "", "kubernetes node name daemon is managing.")
 	startCmd.PersistentFlags().StringVar(&startOpts.rootMount, "root-mount", "/rootfs", "where the nodes root filesystem is mounted for chroot and file manipulation.")
+	startCmd.PersistentFlags().StringVar(&startOpts.onceFrom, "once-from", "", "Runs the daemon once using the provided URI as it's configuration source")
```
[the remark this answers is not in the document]
ashcrow: Good catch. I'll clarify that this can be a path to a file or a URL. URI does make it sound like `http://`, `ftp://`, `file://`, etc.. would be supported when really it's `http[s]://` and `/...`.

pkg/daemon/daemon.go
```diff
@@ -370,6 +396,43 @@ func (dn *Daemon) Close() {
 	dn.loginClient.Close()
 }
 
+// getMachineConfigFromFile parses a valid machine config file in yaml format and returns
+// a MachineConfig struct.
+func (dn *Daemon) getMachineConfigFromFile(filePath string) (*mcfgv1.MachineConfig, error) {
+	config := mcfgv1.MachineConfig{}
+	data, err := ioutil.ReadFile(filePath)
```
kikisdeliveryservice: Maybe we should add these ioutils to fsclient.go?:)

ashcrow: Will do.

cmd/machine-config-daemon/start.go
kikisdeliveryservice: minor typo: "its"

pkg/daemon/daemon.go
```diff
@@ -102,6 +109,25 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
+		case strings.HasPrefix(string(os.PathSeparator), dn.onceFrom):
+			// TODO: read file
+		case strings.HasPrefix("http", dn.onceFrom):
+			// TODO: pull config
+		default:
+			return fmt.Errorf("%s is not a path nor url; can not run once", dn.onceFrom)
+		}
+
+		// TODO: parse mcfgv1.MachineConfig{}
+		// TODO: run
```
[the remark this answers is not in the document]
sdemos: yeah, makes sense. plus there is the work in #130 which is going to change that logic up even further. maybe we can just work on unifying it moving forward.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
sdemos: either way is fine with me, but if it's just absolute we should make sure to be explicit about it, it might be a confusing error if you specify a relative file path and get told that you need to provide a file.

pkg/daemon/daemon.go
ashcrow: https://github.com/openshift/machine-config-operator/pull/139

pkg/daemon/daemon.go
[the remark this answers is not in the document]
kikisdeliveryservice: re: Errors, I'm wondering if we need to validate the URI before proceeding or if we think that the errors from the calls read files/pull configs will be enough?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
ashcrow: @sdemos as in encapsulate the logic for `runOnce` on it's own as to make merging easy and follow up later to unify them?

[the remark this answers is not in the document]
sdemos: yeah, since it would take some significant work to fully unify them today

ashcrow:
(not starting a retest yet as this requires another PR which hasn't merged yet due to yesterdays bot outage. It's being re-reviewed now)

ashcrow:
Will rebase with #134 later today.

pkg/daemon/daemon.go
```diff
@@ -129,6 +142,41 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
+			return err
+		}
+	} else {
+		// Otherwise return an error as the input format is unsupported
+		return fmt.Errorf("%s is not a path nor url; can not run once", dn.onceFrom)
+	}
+
+	// At this point we have a populated MachineConfig struct for our desired config
+	// and we run update using an empty machineConfig as there is no provided current state.
+	oldConfig := mcfgv1.MachineConfig{}
```
ashcrow: @sdemos thoughts on this. Should we try to create an initial MC based on something, or does an empty one work?

ashcrow:
Rebased.

Still todo:
- [ ] Update `process` to be able to run without making requests for state from the cluster
- [ ] Add tests

pkg/daemon/daemon.go
sdemos: an empty one will get rejected by `reconcilable` since there are "changes" to the config in places we can't fix.

it would be much better to use something that represents the actual system state. I'm not sure when or where this will be running exactly, will you be able to get it from kube? is it reasonable to expect the user to provide the current config? maybe we can have the current config always available somewhere on disk?

there are a few other possibilities if getting the current config isn't feasible. we could construct the old config from the irreconcilable parts of the new config. we could rip apart the update function so that the part we call always applies configs and we just say "trust us this one works". or we might be able to just pass it the new config for both the old and the new, since we already determined that we need to update, and most of the functionality in update just applies the config directly and doesn't diff. all of those approaches would break anything that does rely on the current configuration, like file deletion (I think that's the only thing right now, but conceivably there could be more in the future).

sdemos: tl;dr is having an accurate current configuration seems like the vastly preferable option if that's at all possible, but you might be able to hack around it.

ashcrow: I don't think we will be able to get the config from kubernetes but let me defer to folks who brought up and will be utilizing this use case. I tend to think we can't being that we can't pull the MC from the API itself, but it doesn't hurt to ask :smile: 

@sdodson / @aaronlevy at the point that MCD runs in the `onceFrom` path is the cluster up and serving machine configs?

aaronlevy: In the case of the bootstrap ignition config, there would be no API/cluster yet (the output of this step would create one). And for bootstrap it's also not a state that we want to "reconcile". The bootstrap assets normally are on a separate throw-away node - but in the BYO case, they are run on one of the control-nodes - but then cleaned up after.

For all other cases of `onceFrom`, it would be retrieving the state from the API/cluster (either bootstrap API or real API).

sdemos: so in the case of bootstrap we just want to blindly apply it, it sounds like. will we be required to delete files for the bootstrap case? at what point are we running on the bootstrap node? why can't we rely on ignition at that point, since if there is no bootstrap api then isn't this just the initial provisioning step that we are intervening in?

sorry if any of these questions are already answered somewhere else, I guess I haven't been keeping up with this feature as well as I thought I was.

aaronlevy: > will we be required to delete files for the bootstrap case?

We do want to clean up these files after bootstrap completes, but I don't initially think it should be the responsibility of MCD (worry that there's fair amount of complexity here we could punt on a simple post-cleanup step).

> at what point are we running on the bootstrap node? 

In the BYO-RHEL host case, you've booted 3 masters and then we execute MCD with `--once-from=/on/disk/boostrap.ign` on one of them (no api is available yet). The bootstrap process then brings up a bootstrap MCO that will serve the master.ign. On all the master nodes then we would need to execute MCD `--once-from=https://control/plane` which would retrieve those roles. This overlap of MCD runs is where the deletion of bootstrap would get a little wonky (we shouldn't actually drop bootstrap assets until the control-plane is fully functional).

> why can't we rely on ignition at that point, since if there is no bootstrap api then isn't this just the initial provisioning step that we are intervening in?

This is to support RHEL nodes so they're already booted / no ignition involved. We're 'using' the ignition definitions to keep the source of truth for node configuration the same -- even if it's a bit of overloaded behavior.

ashcrow: Thanks for the clarification @aaronlevy. We really appreciate it!

ashcrow:
Rebased on informers code by @yuqi-zhang

ashcrow:
Starting to work on:

- Update remote http(s) onceFrom to request resources from the cluster
- Update local ignition file onceFrom to be able to run without making requests for state from the cluster

ashcrow:
Added code and updated flow for:

- Update remote http(s) onceFrom to request resources from the cluster
- Update local ignition file onceFrom to be able to run without making requests for state from the cluster

It feels messy. As expected, we now have 3 cases for attempting updates instead of 1 and each has a slightly different path:

- Fully cluster integrated
- Partially cluster integrated
- Local like only

However, this is what I have so far. PTAL and help me simplify and restructure if possible!

sdemos:
the logic looks good to me. I toyed a bit with the code to see if I could think of a good way to simplify it, and none of my initial approaches panned out.

ashcrow:
@aaronlevy @sdodson ready for review (though holding merge till I add tests)

ashcrow:
```
could not wait for pod to complete: could not wait for pod completion: the pod ci-op-i8c6txwz/e2e-aws failed after 13m46s (failed containers: setup, test):  unknown
```

and AWS errors.

ashcrow:
No code changes, but e2e now passes.

ashcrow:
> This is perhaps off topic for this PR but are there any suggestions for how we should run this? I assume we'll want to run it from a container rather than packaging it up and installing it on the host on BYOR. Should it be a static pod, should we craft a systemd unit, should we just run the container via ansible command module?

Good question. I'm not sure what the best way to kick off the container would be, but I agree that using a container would be the proper way to run MCD. I believe [this image](https://github.com/openshift/machine-config-operator/blob/master/Dockerfile.machine-config-daemon) could be used within or without the cluster.

ashcrow:
Assuming tests in CI all pass this should be ready to be tested. My quick smoke tests worked, but I'd like to have others give it a go.

cmd/machine-config-daemon/start.go
```diff
@@ -69,22 +67,44 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+			startOpts.kubeconfig,
+			daemon.NewFileSystemClient(),
+			startOpts.onceFrom,
+			stopCh,
+			componentName,
+		)
+		if err != nil {
+			glog.Fatalf("failed to initialize daemon: %v", err)
+		}
+		err = dn.CheckStateOnBoot(stopCh)
```
yuqi-zhang: So now that you start the informer in `NewClusterDrivenDaemon` above, we run into the problem that is: let's say a previous run degraded the node while updating, so we never got to the state where `desiredConfig` = `currentConfig`. Upon restarting, we start the informer and it will realize we need an update (dc != cc), and trigger that likely before we reach the CheckStateOnBoot (since in practice the update gets triggered quite often from each node). This way instead of sleeping (which preserves the actual degrade error from the previous run), we crash loop instead.

I recommend splitting out the starting of the informer to after this line instead.

pkg/daemon/daemon.go
```diff
@@ -177,38 +226,116 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
 	// First check if the node that was updated is this daemon's node
-	if (node.Name != dn.name) {
-		// The node that was changed was not ours
-		return
+	if node.Name == dn.name {
+		// Pass to the shared update prep method
+		_, err := dn.prepUpdateFromCluster()
+		if err != nil {
+			glog.V(2).Infof("Unable to prep update: %s", err)
+			return
```
yuqi-zhang: So for this and the following error, note that this is a callback function, so the return actual doesn't go anywhere useful. In that case, if say, the `prepUpdate` fails, we just return to... listening to the informer, which will trigger this again very soon, and it will just silently fail infinitely.

Barring a better way to do this, I recommend that on any update failure, for the time being to set the node to `degraded` and reboot the machine, so we can "catch" the error and sleep. I recognize this may not be the most clean way to handle this, and an alternative I tried was to throw/catch a `panic` and handle it that way if you think that makes sense.

pkg/daemon/daemon.go
```diff
@@ -177,38 +226,116 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
+		}
+		if err = dn.executeUpdateFromCluster(); err != nil {
+			glog.V(2).Infof("Unable to execute update: %s", err)
+			return
```
yuqi-zhang: same as above

pkg/daemon/daemon.go
```diff
@@ -177,38 +226,116 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
-		// The node that was changed was not ours
-		return
+	if node.Name == dn.name {
+		// Pass to the shared update prep method
+		_, err := dn.prepUpdateFromCluster()
+		if err != nil {
+			glog.V(2).Infof("Unable to prep update: %s", err)
+			return
+		}
+		if err = dn.executeUpdateFromCluster(); err != nil {
```
yuqi-zhang: If there is no update from the above `prepUpdateFromCluster`, shouldn't we not run this? Currently we only return if that previous function errors.

Also, we should be passing the `desiredConfig` from the `prepUpdate` above, otherwise we'll hit https://github.com/openshift/machine-config-operator/pull/145 again, right?

pkg/daemon/daemon.go
ashcrow: Hrm, that's a good point. `panic` handling is kind of a last ditch effort IMHO. I'll look at marking degraded and rebooting.

pkg/daemon/daemon.go
ashcrow: :+1:

pkg/daemon/daemon.go
ashcrow: Note that failure in `prepUpdateFromCluster()` will mark the node degraded on it's own. It doesn't, however, reboot. Will add that and some comments.

pkg/daemon/daemon.go
ashcrow: Re-checking the code `executeUpdateFromCluster()` will mark the node degraded and reboot upon error. I'll add some comments though.

pkg/daemon/daemon.go
ashcrow: Agreed on first point.

On the second point are you recommending that we should return the `desiredConfig` as a return value for use later?

yuqi-zhang: Actually maybe I mis-understood the update method the first time around, I think this should be fine.

Originally, the main concern being that let's say 2 updates happen in quick succession, then the first update is not stored, and we might see the second `desiredConfig` when we poll it again at: https://github.com/openshift/machine-config-operator/pull/126/files#diff-06b532a2a5a5d3ad1d5341b3e00945efR404 , but that should be fine since we'd just apply 2 updates and bring it to the most updated "desired state"? Correct me if I'm wrong in this understanding. If that's how it operates, then it's fine to leave as is.

cmd/machine-config-daemon/start.go
ashcrow: Updated with logic we came to via chat.

pkg/daemon/daemon.go
ashcrow: `prepUpdateFromCluster` is:

ashcrow:
/test e2e-aws

Router didn't create.

vrutkovs:
Looks good to me. MCD would probably have to run in a plain container as its ready to use, while kubelet needs to be pre-configured on the node

ashcrow:
Thanks @vrutkovs! With the `--once-from` the kubelet config would either need to be part of the Machine Config file provided during the single run, OR, if on a non-immutable host, configured by an external entity as noted.

ashcrow:
/test e2e-aws

Router didn't start again.

ashcrow:
With the updates from the last review I'm getting an error when running once from a file only:

```
I1107 15:50:24.160249    1292 update.go:81] Checking if configs are reconcilable                                                                  
W1107 15:50:24.160256    1292 update.go:91] daemon can't reconcile state!                                                                         
W1107 15:50:24.160273    1292 update.go:92] Ignition version mismatch between old and new config: old:  new: 2.2.0                                
F1107 15:50:24.160279    1292 start.go:128] failed to run: daemon can't reconcile this config     
```

Looking into this.

ashcrow:
`daemon/update: Allow reconcile skip` assumes there was no previous state when we are in runOnce mode and the previous MC has no Kind.

Tested with:

```
./machine-config-daemon start -v 2 --node-name=test --root-mount=/ --once-from=/tmp/inside.ign
```
Using a valid MC at `/tmp/inside.ign`

@yuqi-zhang :arrow_up:

kikisdeliveryservice:
@ashcrow agree that you will def have to skip using func reconcilable() for this PR. :+1:

cmd/machine-config-daemon/start.go
```diff
@@ -69,22 +67,47 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
```
yuqi-zhang: Another note for here: I remember there were issues doing state checking, etc. before the `Chroot` and the `Chdir` right after, I will double check.

pkg/daemon/daemon.go
```diff
@@ -67,59 +71,112 @@ const (
...
+// StartInformer initializes and starts the informers.
+func (dn *Daemon) StartInformer(stopCh chan (struct{}), nodeName, componentName, kubeconfig string) error {
+	cb, err := common.NewClientBuilder(kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	ctx := common.CreateControllerContext(cb, stopCh, componentName)
+
+	ctx.KubeInformerFactory.Start(ctx.Stop)
```
yuqi-zhang: In the current workflow we do this after setting up the nodeInformer/Lister. I'm not sure if the ordering will cause issues, but I remember that is what we did elsewhere for MCO/MCC, so perhaps move this line and the following line to the end of this function?

pkg/daemon/daemon.go
```diff
@@ -177,38 +234,123 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
+		if err != nil {
+			glog.V(2).Infof("Unable to prep update: %s", err)
+			return err
+		} else if needUpdate == false {
+			glog.V(2).Infof("No update needed")
+			return nil
+		}
+		// At this point we have verified we need to update
+		if err := dn.executeUpdateFromClusterWithMachineConfig(machineConfig); err != nil {
+			glog.Warningf("Unable to update: %s", err)
```
yuqi-zhang: We do already `Errorf` inside the `executeUpdateFromClusterWithMachineConfig`, probably unnecessary

pkg/daemon/daemon.go
```diff
@@ -177,38 +234,123 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
+			glog.V(2).Infof("Unable to prep update: %s", err)
+			dn.reboot()
+			return
+		}
+		// Only executeUpdateFromCluster when we need to update
+		if needUpdate {
+			// Note that if executeUpdateFromCluster errors it will mark the node
+			// degraded and reboot.
+			if err = dn.executeUpdateFromCluster(); err != nil {
+				glog.V(2).Infof("Unable to execute update: %s", err)
```
yuqi-zhang: Can probably remove, errors are logged in the update execution

pkg/daemon/daemon.go
```diff
@@ -177,38 +234,123 @@ func (dn *Daemon) CheckStateOnBoot(stop <-chan struct{}) error {
...
+	if err != nil {
+		return false, err
 	}
 
 	// Detect if there is an update
-	if (node.Annotations[DesiredMachineConfigAnnotationKey] == node.Annotations[CurrentMachineConfigAnnotationKey]) {
+	if node.Annotations[DesiredMachineConfigAnnotationKey] == node.Annotations[CurrentMachineConfigAnnotationKey] {
 		// No actual update to the config
-		return
+		glog.V(2).Info("No updating is required")
```
yuqi-zhang: I suggest removing this since this will likely get logged a lot in cluster operation.

cmd/machine-config-daemon/start.go
yuqi-zhang: Confirmed, looks good!

pkg/daemon/daemon.go
yuqi-zhang: Yes, I mistook the application of config and updating the annotation. This should be fine.

ashcrow:
Same router flake. Waiting till after manual testing before trying an e2e test.

abhinavdahiya:
@ashcrow
```console
Setup success
NAME                           STATUS                     ROLES     AGE       VERSION
ip-10-0-130-104.ec2.internal   Ready,SchedulingDisabled   worker    25m       v1.11.0+d4cacc0
ip-10-0-154-42.ec2.internal    Ready,SchedulingDisabled   worker    25m       v1.11.0+d4cacc0
ip-10-0-166-190.ec2.internal   Ready,SchedulingDisabled   worker    25m       v1.11.0+d4cacc0
ip-10-0-27-206.ec2.internal    Ready,SchedulingDisabled   master    30m       v1.11.0+d4cacc0
ip-10-0-32-194.ec2.internal    Ready,SchedulingDisabled   master    29m       v1.11.0+d4cacc0
ip-10-0-7-73.ec2.internal      Ready,SchedulingDisabled   master    30m       v1.11.0+d4cacc0
```

this is not a router flake.. somebody is making the node SchedulingDisabled,
MCD does cordon nodes... might be something to look into...

ashcrow:
@abhinavdahiya hrm ok. Thanks for the note. You're right, MCD does set the `degraded` label so it's possible that it's flowing back into disabled. @yuqi-zhang is testing it out manually now. If he sees the same issue we'll have data to debug with.

yuqi-zhang:
@ashcrow Locally, it seems that the daemons are running fine (no sign of degradation), but no updates are being applied when I do `oc apply` (i.e. there are no reboots). I suspect this could be because of the ordering of informer workflow as mentioned above? Not sure what is causing the e2e degradation though.

pkg/daemon/daemon.go
ashcrow: Updated and moved start to the end of this function before return.

ashcrow:
Updated the informer code per request.

kikisdeliveryservice:
Seems like another flake? :(

abhinavdahiya:
> Seems like another flake? :(

We are not seeing this specific error anywhere else **yet**
```console
NAME                           STATUS                     ROLES     AGE       VERSION
ip-10-0-12-35.ec2.internal     Ready,SchedulingDisabled   master    1h        v1.11.0+d4cacc0
ip-10-0-137-214.ec2.internal   Ready,SchedulingDisabled   worker    1h        v1.11.0+d4cacc0
ip-10-0-145-104.ec2.internal   Ready,SchedulingDisabled   worker    1h        v1.11.0+d4cacc0
ip-10-0-175-235.ec2.internal   Ready,SchedulingDisabled   worker    1h        v1.11.0+d4cacc0
ip-10-0-23-254.ec2.internal    Ready,SchedulingDisabled   master    1h        v1.11.0+d4cacc0
ip-10-0-44-102.ec2.internal    Ready,SchedulingDisabled   master    1h        v1.11.0+d4cacc0
Waiting for router to be created ...
```

That makes it seem like it is not a flake.

ashcrow:
Agreed, I don't think this is a flake. It does look like a flake we hit previously BUT @yuqi-zhang and I have found cases where nodes are set degraded when they shouldn't be. We're working on this now.

ashcrow:
/hold cancel

- e2e now passes
- onceFrom a local file verified to work

pkg/daemon/daemon.go
```diff
@@ -120,6 +151,12 @@ func New(
 // responsible for triggering callbacks to handle updates. Successful
 // updates shouldn't return, and should just reboot the node.
 func (dn *Daemon) Run(stop <-chan struct{}) error {
+	// Catch quickly if we've been asked to run once.
+	if dn.onceFrom != "" {
+		glog.V(2).Info("Running once per request")
```
kikisdeliveryservice: Is this logging clear enough? Would "daemon running once per request" be clearer? Thoughts?

kikisdeliveryservice:
Also, thanks for adding comments in this PR @ashcrow , it makes it very easy to read through! :+1:

pkg/daemon/daemon.go
ashcrow: I like your wording better. How about I rework some of the log strings in a follow up?

kikisdeliveryservice: Sounds great!

yuqi-zhang:
Some manual testing passes for me both for cluster operation and runOnce with qemu. Will LGTM after logs are cleaned up.

Thanks for the work @ashcrow !

ashcrow:
Thanks @yuqi-zhang and @kikisdeliveryservice! I'll do the log clean up in another PR post merge.

PR #128, opened 2018-10-15
daemon: Remove punctuation from error

ashcrow:
`golint` doesn't like punctuation ending errors


jlebon:
Sure, although that hunk is reworked in #129. Will fix it there as well!

ashcrow:
Ah ok. Will close in favor of the bigger set of work.

Issue #127, opened 2018-10-15
MCD: pods are crashlooping


abhinavdahiya:
@sallyom reported that all the mcd pods are crashlooping.

```console
installer(master)$ oc logs machine-config-daemon-5d862 -n openshift-machine-config-operator
I1015 15:56:56.802774    5350 start.go:42] Version: 3.11.0-152-g0fe65acd-dirty
I1015 15:56:56.828854    5350 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
W1015 15:56:56.884945    5350 rpm-ostree.go:88] Working around "://dummy" OS image URL until installer ➰ pivots
I1015 15:56:56.885256    5350 daemon.go:81] Booted osImageURL: ://dummy (4.0.6844)
I1015 15:56:56.885341    5350 start.go:85] Calling chroot("/rootfs")
I1015 15:56:56.885424    5350 daemon.go:101] Starting MachineConfigDaemon
I1015 15:56:56.893192    5350 daemon.go:110] Shutting down MachineConfigDaemon
F1015 15:56:56.893217    5350 start.go:99] failed to run: Node is degraded; exiting loudly...
```

/cc @sdemos

jlebon:
Hmm, OK just had reports of this internally as well. Odd though, since that logic hasn't changed recently. Will look into it.

ashcrow:
Looks related to https://github.com/openshift/machine-config-operator/pull/121

ashcrow:
I could see this happening if the [node was previously marked degraded](https://github.com/openshift/machine-config-operator/pull/121/files#diff-06b532a2a5a5d3ad1d5341b3e00945efR105)

jlebon:
Ahh OK, sorry for the confusion. This is indeed not the same thing as was reported internally with handling of `://dummy` URLs.

> If so the exit is expected, but since Openshift is rescheduling over and over it's crash looping.

Right, though right now aren't we the only operators that set the degraded label? So it must have encountered an error in a previous run. Will try to reproduce this though `oc logs -p` from someone who's hitting this would help.

sdemos:
> Right, though right now aren't we the only operators that set the degraded label?

We should be, I'm pretty sure that the MCD owns the node status field entirely so if anything else is writing to it it's a problem. However, even if that is the case, (unfortunately) right now crash-looping is the name of the game until we get the work described in #109.

[...]

PR #129, opened 2018-10-15
daemon: just sit quietly if node is degraded

jlebon:
An issue with immediately exiting if the node is degraded is that we
then clobber the logs of the previous MCD containers, which means that
the logs from the MCD instance that actually set the node to degraded is
lost in the ether.

Let's just block on `poll()` so we don't exit and don't waste any CPU
resources. The stop channel here isn't actually being written/closed by
any other goroutine so this just sits there.

See also #127.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -107,7 +107,14 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
 		glog.Errorf("Marking degraded due to: %v", err)
 		return setUpdateDegraded(dn.kubeClient.CoreV1().Nodes(), dn.name)
 	} else if state == MachineConfigDaemonStateDegraded {
-		return fmt.Errorf("Node is degraded; exiting loudly...")
+		// just sleep so that we don't clobber output of previous run which
+		// probably contains the real reason why we marked the node as degraded
+		// in the first place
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -107,7 +107,14 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
 		glog.Errorf("Marking degraded due to: %v", err)
 		return setUpdateDegraded(dn.kubeClient.CoreV1().Nodes(), dn.name)
 	} else if state == MachineConfigDaemonStateDegraded {
-		return fmt.Errorf("Node is degraded; exiting loudly...")
+		// just sleep so that we don't clobber output of previous run which
+		// probably contains the real reason why we marked the node as degraded
+		// in the first place
+		glog.Info("Node is degraded; going to sleep")
+		select {
```
ashcrow: This looks good to me. Cleaner than my `for{}` loops for sure. The path here then would be:

- If the machine was already degraded, hold (unless explicitly told to stop via the channel)
- If we get an error during `getNodeAnnotation`, set degraded, exit, and hit the above line on reschedule and hold
- Otherwise, continue into main loop

jlebon: Yup, that's correct!

PR #130, opened 2018-10-15
daemon: use informers to check for updates

yuqi-zhang:
Change existing watch workflow to an informer instead, as watches
are prone to breaking and restarting the MCD. The Run() function
within MCD is changed from an infinite loop to instead block,
and updates are handled by callbacks via the informer.

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -67,6 +67,10 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		glog.Fatalf("unable to verify rootMount %s exists: %s", startOpts.rootMount, err)
 	}
 
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	ctx := common.CreateControllerContext(cb, stopCh, componentName)
 	// create the daemon instance. this also initializes kube client items
 	// which need to come from the container and not the chroot.
 	daemon, err := daemon.New(
@@ -77,6 +81,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		cb.MachineConfigClientOrDie(componentName),
 		cb.KubeClientOrDie(componentName),
 		daemon.NewFileSystemClient(),
+		ctx.KubeInformerFactory.Core().V1().Nodes(),
 	)
 	if err != nil {
 		glog.Fatalf("failed to initialize daemon: %v", err)
@@ -92,8 +97,17 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		glog.Fatalf("unable to change directory to /: %s", err)
 	}
 
-	stopCh := make(chan struct{})
-	defer close(stopCh)
+	glog.Info("Starting MachineConfigDaemon")
+	defer glog.Info("Shutting down MachineConfigDaemon")
+
+	err = daemon.CheckStateOnBoot(stopCh)
+	if err != nil {
+		glog.Fatalf("error checking initial state of node: %v", err)
+	}
+
+	ctx.KubeInformerFactory.Start(ctx.Stop)
+	close(ctx.KubeInformersStarted)
+
 	err = daemon.Run(stopCh)
 	if err != nil {
 		glog.Fatalf("failed to run: %v", err)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -17,6 +17,10 @@ import (
 	"github.com/vincent-petithory/dataurl"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/api/core/v1"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
 )
 
 // Daemon is the dispatch point for the functions of the agent on the
@@ -47,6 +51,11 @@ type Daemon struct {
 
 	// rootMount is the location for the MCD to chroot in
 	rootMount string
+
+	// nodeLister is used to watch for updates via the informer
+	nodeLister corelisterv1.NodeLister
+
+	nodeListerSynced cache.InformerSynced
 }
 
 const (
@@ -68,6 +77,7 @@ func New(
 	client mcfgclientset.Interface,
 	kubeClient kubernetes.Interface,
 	fileSystemClient FileSystemClient,
+	nodeInformer coreinformersv1.NodeInformer,
 ) (*Daemon, error) {
 	loginClient, err := login1.New()
 	if err != nil {
@@ -84,7 +94,7 @@ func New(
 	}
 	glog.Infof("Booted osImageURL: %s (%s)", osImageURL, osVersion)
 
-	return &Daemon{
+	dn := &Daemon{
 		name:              nodeName,
 		OperatingSystem:   operatingSystem,
 		NodeUpdaterClient: nodeUpdaterClient,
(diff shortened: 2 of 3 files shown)


yuqi-zhang:
Marking WIP as the newest version isn't fully tested pending other issues.

cc/ @jlebon @sdemos @ashcrow

ashcrow:
Makes sense so far. I recommend rebasing though as I think there has been a commit or two in the areas you're updating.

abhinavdahiya:
you are using an informer that reacts to changes from all the nodes. Please only watch changes to the node the daemon was scheduled.

sdemos:
@abhinavdahiya do you have a good example of an informer that only watches one instance of a resource? I wasn't sure if you could do that, right now the update callback checks if it's the node it cares about and if it's not it returns immediately, which I thought would be enough.

yuqi-zhang:
Rebased.
@abhinavdahiya that's a good point, but I'm not sure what the proper method to do so would be. Am I able to just set an informer on a singular node instead of `KubeInformerFactory.Core().V1().Nodes()`?

yuqi-zhang:
Added some fixes and removing WIP:
 - instead of panic, instead reboot the node after setting degraded. This makes it consistent with other failures (the node will sleep post reboot)
 - add a check to continue sleeping upon updates while we're degraded to preserve previous run logs (with actual failures)
 - stop checking for node updates and deletions for the informer (causes race conditions when update has not fully applied, and we handle those cases elsewhere regardless)

It is possible to filter for nodes to watch, albeit roundabout and adds complexity. I feel that the check in handleUpdate is working, and doesn't cause performance issues to warrant a custom informer. Note that there is the option to use UntilWithSync: https://github.com/kubernetes/client-go/blob/master/tools/watch/until.go#L98 sometime in the future (basically a custom informer that mimics our old workflow without causing issues), but that is with v9.0.0 (we use 8.0.0: https://github.com/openshift/machine-config-operator/blob/master/Gopkg.toml#L60), so I personally think the fix here should work for now.

I've tested with updates to configs, as well as leaving it running for 6+ hours, haven't broken so far or hot looped. I have the latest image built at docker.io/jerzhang/origin-machine-config-daemon:v3.11.0-178-g936c032f if you want to try for yourself.

yuqi-zhang:
@ashcrow @sdemos PTAL

pkg/daemon/daemon.go
```diff
@@ -94,14 +105,36 @@ func New(
 		rootMount:         rootMount,
 		fileSystemClient:  fileSystemClient,
 		bootedOSImageURL:  osImageURL,
-	}, nil
+	}
+
+	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
ashcrow: No update needed, but I'm curious: Was there a specific reason to do the informer setup after the `Daemon` creation? Or is it more so it reads better to you this way rather than:

```golang
return  &Daemon{
        ....
        nodeLister:       nodeInformer.Lister(),
        nodeListerSynced: nodeInformer.Informer().HasSynced,
}, nil

```

cmd/machine-config-daemon/start.go
```diff
@@ -77,11 +81,15 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
 		cb.KubeClientOrDie(componentName),
 		daemon.NewFileSystemClient(),
+		ctx.KubeInformerFactory.Core().V1().Nodes(),
 	)
 	if err != nil {
 		glog.Fatalf("failed to initialize daemon: %v", err)
 	}
 
+	ctx.KubeInformerFactory.Start(ctx.Stop)
+	close(ctx.KubeInformersStarted)
```
ashcrow: I think a comment here would be helpful. Seeing a closing of the channel without going through the informer docs seems odd.

ashcrow:
Will let @sdemos and/or @jlebon review and pull the merge lever if they also :+1:.

pkg/daemon/daemon.go
```diff
@@ -121,64 +154,78 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
+		return
+	}
 
-		// either the machine state isn't what we wanted and we should try
-		// again, or the machine state is what we wanted, and now another update
-		// is was triggered.
-		if err := dn.triggerUpdate(); err != nil {
-			return err
+	// The desired machine config has changed, trigger update
+	if err := dn.triggerUpdate(); err != nil {
```
jlebon: Do I understand correctly that there's nothing preventing this to run at the same time as the initial `triggerUpdate()`? Or do callbacks not actually occur until after `cache.WaitForCacheSync()`

pkg/daemon/daemon.go
```diff
@@ -121,64 +154,78 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
-		if isDesired {
-			// we got the machine state we wanted. set the update complete!
-			if err := dn.completeUpdate(); err != nil {
-				return err
-			}
+	// Then check we're not already in a degraded state.
+	if state, err := getNodeAnnotation(dn.kubeClient.CoreV1().Nodes(), dn.name, MachineConfigDaemonStateAnnotationKey); err != nil {
+		// try to set to degraded... because we failed to check if we're degraded
+		glog.Errorf("Marking degraded due to: %v", err)
+		setUpdateDegraded(dn.kubeClient.CoreV1().Nodes(), dn.name)
```
jlebon: I think I'd rather we stop watching and exit on error conditions in this callback function and instead only have the "sleep on Degraded" logic at the beginning of the daemon.

pkg/daemon/daemon.go
yuqi-zhang: I noticed that elsewhere we split off the Lister, etc. (e.g. https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L105). I can change it to your suggested method if you'd like.

pkg/daemon/daemon.go
```diff
@@ -121,64 +154,78 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
-	if err := dn.process(); err != nil {
+	// validate machine state
+	isDesired, err := dn.isDesiredMachineState()
+	if err != nil {
 		glog.Errorf("Marking degraded due to: %v", err)
 		return setUpdateDegraded(dn.kubeClient.CoreV1().Nodes(), dn.name)
 	}
 
+	if isDesired {
+	// we got the machine state we wanted. set the update complete!
```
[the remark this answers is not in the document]
yuqi-zhang: will fix

pkg/daemon/daemon.go
```diff
@@ -121,64 +154,78 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
 		}
 	}
 
-	if err := dn.process(); err != nil {
```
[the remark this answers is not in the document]
yuqi-zhang: My initial thought was that `process()` was more of a loop processing commands, whereas now it just blocks at the `<-stop`, and callbacks happen instead. This is also somewhat consistent with how we have informers elsewhere in this repo. I have no strong feelings either way so I can change that if you feel that makes more sense,

pkg/daemon/daemon.go
yuqi-zhang: Sure, just to make sure, you mean degrade, stop the informer, and then `reboot()` so we can catch it and sleep?

pkg/daemon/daemon.go
yuqi-zhang: Good point, I haven't seen this cause an issue/race condition (even when there are repeated failed updates so it should have triggered), but let me double check.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
ashcrow: IMHO I think it's ok that `Run` and `process` are not split in this PR wince the work in #126 was going to have to change `process` and `Run` significantly anyway.

pkg/daemon/daemon.go
jlebon: Definitely (I thought that was already the case -- when are callbacks actually active? Is it at `nodeInformer.Informer().AddEventHandler()` time? In that case, yeah we should move that to after checking if the node is even degraded.

But also in this function if the node is degraded, let's just stop watching for updates and exit?

pkg/daemon/daemon.go
ashcrow: No this is fine. I was just curious :smile:

yuqi-zhang:
Added fixup to move the state checking before the informer starts, which degrades early (and doesn't start informer anymore) as well as prevent race conditions between multiple updates. Will squash if it makes sense. Tested.

@jlebon PTAL

sdemos:
changes all make sense to me, and the logic looks good. leaving it to @jlebon to confirm the changes he requested.

pkg/daemon/daemon.go
```diff
@@ -117,27 +117,36 @@ func New(
...
-// 4. if an update is requested by the controller, we assume that that means
-//    something changed and apply the desired config no matter what.
+// Run finishes informer setup and then blocks, and the informer will be
+// responsible for triggering callbacks to handle updates. Successful
+// updates shouldn't return, and should just reboot the node.
 func (dn *Daemon) Run(stop <-chan struct{}) error {
 	defer utilruntime.HandleCrash()
 
 	glog.Info("Starting MachineConfigDaemon")
 	defer glog.Info("Shutting down MachineConfigDaemon")
```
jlebon: Minor: at this point, we might as well move these to `runStartCmd` before we call `CheckStateOnBoot`?

cmd/machine-config-daemon/start.go
```diff
@@ -92,8 +97,14 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
 	}
 
-	stopCh := make(chan struct{})
-	defer close(stopCh)
+	err = daemon.CheckStateOnBoot(stopCh)
+	if err != nil {
+		glog.Fatalf("error checking initial state of node: %v", err)
+	}
+
+	ctx.KubeInformerFactory.Start(ctx.Stop)
```
jlebon: IIUC, there's a race condition here if the node annotation is modified *after* `CheckStateOnBoot()`, but *before* cache is synced and we can start receiving callbacks, right?

Here's an idea: we spawn another goroutine that just listens for notifications on a channel before doing `triggerUpdate()`. Then in `handleUpdate` we send a notification. But crucially, we also send a notification after `WaitForCacheSync()` is done if annotations don't match. That way updates are still always serialized, and we close the race condition.

pkg/daemon/daemon.go
yuqi-zhang: Good point, will fix

cmd/machine-config-daemon/start.go
yuqi-zhang: Yes, I thought about this as well, but while testing this it seems that the notifications should still be there for when the informer actually starts listening (Maybe I'm mis-understanding, I used to have a addNode callback as well which triggers despite having happened before the informer starts listening)

Also, during testing, I found that the callback happens quite frequently with the node (once every couple seconds), so updates are generally caught eventually anyways (since we only care about differences in CC vs DC anyways).

yuqi-zhang:
@jlebon pushed a minor fixup

As discussed, it is probably better to handle via. a new goroutine to ensure no updates are lost, but for the time being this is unlikely to cause any race conditions. Instead I've PR'ed a fix to the other race condition at: https://github.com/openshift/machine-config-operator/pull/145

We will likely improve this behaviour as follow-up work. If you think this makes sense, I will squash the commits.

Thanks for the reviews!

ashcrow:
@yuqi-zhang needs a rebase.

yuqi-zhang:
Rebased, squashed, and re-tested locally

PR #131, opened 2018-10-15
daemon: allow no state label to be set on node

jlebon:
Regression from #121: we would fail at retrieving the annotation on the
first run since it wasn't set yet. Except... that annotation gets set by
us normally.

Tweak this by adding an `Ext` version of `getNodeAnnotation` which
doesn't error out if the annotation is not yet set. So now, we exit
early only if the label is set *and* its value is set to "Degraded".

Fixes: #127

### DIFF
--- a/pkg/daemon/node.go
+++ b/pkg/daemon/node.go
@@ -96,7 +96,13 @@ func loadNodeAnnotations(client corev1.NodeInterface, node string) error {
 	return nil
 }
 
+// getNodeAnnotation gets the node annotation, unsurprisingly
 func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (string, error) {
+	return getNodeAnnotationExt(client, node, k, false)
+}
+
+// getNodeAnnotationExt is like getNodeAnnotation, but allows one to customize ENOENT handling
+func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allow_noent bool) (string, error) {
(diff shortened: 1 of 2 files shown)


abhinavdahiya:
Hmm, i would expect the MCD to be not racey and set if not set...

jlebon:
It does set it later on once it validates the system state. But it won't be set on the first run.

pkg/daemon/node.go
```diff
@@ -96,15 +96,25 @@ func loadNodeAnnotations(client corev1.NodeInterface, node string) error {
 	return nil
 }
 
+// getNodeAnnotation gets the node annotation, unsurprisingly
```
ashcrow: :laughing:  :+1:

pkg/daemon/node.go
```diff
@@ -96,15 +96,25 @@ func loadNodeAnnotations(client corev1.NodeInterface, node string) error {
...
 func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (string, error) {
+	return getNodeAnnotationExt(client, node, k, false)
+}
+
+// getNodeAnnotationExt is like getNodeAnnotation, but allows one to customize ENOENT handling
+func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allow_noent bool) (string, error) {
```
ashcrow: I'd like us to start writing tests for functions, but we can do that later.

PR #133, opened 2018-10-16
kubelet: enable readOnlyPort for 10255

rphillips:
When we migrated to the kubelet configuration file, the config disabled
the readOnlyPort by default. The port is enabled for the checkpointer
and other components to read from the kubelet without credentials.


abhinavdahiya:
> The port is enabled for the checkpointer and other components to read from the kubelet without credentials.

This seems not ideal. is there a way to turn on authentication?

rphillips:
This is currently how the checkpointer works... Is there another way @aaronlevy ?

rphillips:
Without this fix the checkpointer currently installed in 4.0 isn't doing anything and reboots across the cluster will not work.

aaronlevy:
This scares me - how can we get tests that ensure checkpointing is working in 4.0? I thought the smoke tests did api server reboots / apiserver pod deletions?

aaronlevy:
We should also be able to have checkpointer use credentials to contact kubelet on secure port. This is probably just leftover cruft that no one got around to fixing.

aaronlevy:
There might be some complexity of behavior there though.

Kubelet authorization does a round-trip to apiserver. But if the checkpointer is trying to decide to restore an apiserver - there's nothing to contact (and request will just be denied).

It looks like the insecureClient is only being used to determine local parents. This might be fine if the logic for local parents is primarily about creating a checkpoint (not whether we start a checkpoint). But I don't remember off top of my head - would need to dig into logic/tests. If it had to do with starting/stopping a checkpoint - might take a bit more work to get rid of the insecurePort

aaronlevy:
Another option might be to move to using the CRI endpoint instead of kubelet api if we still need un-authed access. But it's likely messier to dig this info out of CRI and map to pods.

abhinavdahiya:
/lgtm

For now this is fine. But it would be nice for pod team to look into alternatives...

rphillips:
Agreed. We can look into different alternatives. Using a certificate to connect to the kubelet's secure port is... interesting... because it adds a dependency to the checkpointer to contact the API server for a cert and cert rotation. In a disaster scenario, a valid cert is problematic.

aaronlevy:
Might be reasonable to just assume it uses the same cert as the kubelet.

aaronlevy:
But still would round-trip to apiserver for the actual auth check... so may not be viable regardless.

If only upstream pod checkpointing hadn't died....

derekwaynecarr:
Why did we update master and workers?  Only master would have been needed for checkpointer, right?

Issue #99, opened 2018-09-27
MCD: consider taking action if validation fails multiple times


sdemos:
a node is validated after being updated but before being marked as completed. this validation normally happens after a reboot. if this validation fails, the update is attempted again, ad infinitum. 

there are three reasons I can think of that would cause validation to fail.
1. a bug in the validation logic.
2. a bug in the update application logic.
3. there is something wrong with the machine that is causing it to either revert the changes or never actually be applied.

for situation 3, it seems like the obvious solution is to set the node status to degraded. after all, that's what that status is essentially for (something's wrong with this machine that we can't fix! help!).

however, it's not possible to reliably distinguish between the situations, and setting the node as degraded for situations 1 and 2 [is not ideal](https://github.com/openshift/machine-config-operator/issues/74).

so I guess the question is, how should we handle this situation? do we mark the node as degraded anyway? do we just keep trying to attempt the update forever? should there be some kind of exponential backoff strategy? do we stop trying after a while and bubble up this issue so a cluster administrator will notice the problem, but without setting the node as degraded (which has other semantic meaning)? thoughts?

ashcrow:
> so I guess the question is, how should we handle this situation? do we mark the node as degraded anyway? do we just keep trying to attempt the update forever?

No! :smiley: 

> should there be some kind of exponential backoff strategy? do we stop trying after a while and bubble up this issue so a cluster administrator will notice the problem, but without setting the node as degraded (which has other semantic meaning)? thoughts?

If we can not validate the host then we are not in the state the cluster wants. My initial thought on the path here would be:

1. Attempt to validate the node
2. It fails. Log. Mark the node as degraded and try again.
3. It validates, log, remove degredation, note the node finished and is once again ready. It fails again. log then try once more
4. It validates, log, remove degredation, note the node finished and is once again ready. It fails again stay in degraded mode, log, and wait for cluster or admin intervention.

I wouldn't be opposed to coming up with a mid-term flag such as `reconciling` while we retry a few times to ensure that it's literally an impossible state for the node to enforce.

sdemos:
[...]

PR #138, opened 2018-10-17
pkg/daemon: remove annotation equality test from validation

### LINKED ISSUE #98: MCD: consider removing annotation equality checking from node validation logic
the current [validation logic](https://github.com/openshift/machine-config-operator/blob/3a73b1d142eb1efb3fad5b160915e6ddf3d38311/pkg/daemon/daemon.go#L198) in the MCD first checks if the annotations on the node are equal. if they are, it assumes that the node has a valid configuration. 

this assumption does make some amount of sense. we are generally careful to validate the node state before marking the update as done, so if everything is as expected and nothing ever changes, the node _is_ going to be in a valid state. 

however, I think we should always validate the machine state in full every time we call the validation function. it's not a particularly expensive check, and it allows us to easily validate our assumptions about machine state being constant, and fix it if the state changes for some reason. plus, it insulates us a little bit from transitory issues with the values in the annotations, which obviously shouldn't happen, but we can trivially not rely as much on them and instead anchor ourselves in reality by removing this check. there might be other good reasons I'm not coming up with too.

thoughts?

/cc @jlebon @ashcrow

sdemos:
the node state validation function had a test right at the beginning
that checked if the values of the current annotation and the desired
annotation were equal. if they were, it returned, under the assumption
that because the annotations were equal the node was in the state we
wanted it to be in.

with the way that we try to behave, in general that would be true.
however, we instead opt for more rigorous testing of the node - every
time the validation function is called, it does a full validation of the
node state. if something changed, we attempt to reconcile the state back
to the desired. this also reflects the fact that nobody should be
modifying the node outside of the daemon, and if they do they should
expect to get clobbered.

also, my editor ran gofmt on the file and fixed the import ordering.

fixes #98

/cc @ashcrow @jlebon

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -282,11 +282,6 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 	if err != nil {
 		return false, "", err
 	}
-	// if the current annotation is equal to the desired annotation,
-	// system state is valid.
-	if strings.Compare(dcAnnotation, ccAnnotation) == 0 {
-		return true, dcAnnotation, nil
-	}
 
 	currentConfig, err := getMachineConfig(dn.client.MachineconfigurationV1().MachineConfigs(), ccAnnotation)
 	if err != nil {


pkg/daemon/daemon.go
```diff
@@ -237,11 +237,6 @@ func (dn *Daemon) isDesiredMachineState() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	// if the current annotation is equal to the desired annotation,
-	// system state is valid.
-	if strings.Compare(dcAnnotation, ccAnnotation) == 0 {
-		return true, nil
-	}
```
jlebon: This makes sense to me. Though... now I wonder about the opposite case. If the annotations *do* match, but `isDesiredMachineState` is `false`, then something has gone wrong, right? I guess this gets into #99, but it's slightly different: there, we're talking about repeatedly trying and failing to meet the desired config so that we can set the current config label to the same. Here, we've already in the past validated the node for that config (since the current config label already matches), but somehow the state doesn't match. I feel like we should just set it to degraded in that case?

So basically, it'd mean still doing this comparison but e.g. in `process` after the call to `isDesiredMachineState()`. Anyway, we can do that in a follow-up!

pkg/daemon/daemon.go
```diff
@@ -3,9 +3,9 @@ package daemon
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
-	"os"
```
jlebon: That's probably from me not getting in the habit yet of running `gofmt`. :)

pkg/daemon/daemon.go
sdemos: my thought was that if there was a one-time change of a file under our control, that we can (and I think should) easily just change it back to what we want. if it happens multiple times right now, we don't do anything, but that was already true (maybe there are more cases of it now). I think doing some kind of backoff like described in #99 is the way to go in the long run, which I think would cover this case as well.

ashcrow:
I don't think the e2e test is going to return. It's been "building" for 4 hours and the Jenkins link 404s.

sdemos:
I see a page when I click on the details link - it says it's not finished, started at 7:30ish PDT and doesn't have any build logs. seems busted. will it restart if I tell it to retest?

sdemos:
let's try it

/retest e2e-aws

sdemos:
well that didn't seem to restart it

@smarterclayton any ideas?

ashcrow:
@sdemos the bot is currently down and being fixed.

ashcrow:
/test e2e-aws

Let's see if `Fix how we wait on router becoming available` moves us past this.

ashcrow:
@sdemos Needs a rebase.

jlebon:
Heh, OK this is failing on:

```
--> RUN WHAT=machine-config-daemon ./hack/build-go.sh
Using version from git...
Building github.com/openshift/machine-config-operator/cmd/machine-config-daemon (v3.11.0-193-g038d67de-dirty)
pkg/daemon/daemon.go:285:1: syntax error: unexpected <<, expecting }
error: build error: running 'WHAT=machine-config-daemon ./hack/build-go.sh' failed with exit code 2
```

Which took me a while to understand, but I think it's due to the merge conflicts. Can you rebase this?

ashcrow:
rebase :smile:

ashcrow:
@sdemos 
```
pkg/daemon/daemon.go:285:1: syntax error: unexpected <<, expecting }
error: build error: running 'WHAT=machine-config-controller ./hack/build-go.sh' failed with exit code 2
```

Specifically:

**pkg/daemon/daemon.go:285:1: syntax error: unexpected <<, expecting }**

sdemos:
alright, it's rebased. let's see if the tests pass.

ashcrow:
/lgtm

... while we can!

PR #139, opened 2018-10-18
daemon: Add ReadFile to FsClient

ashcrow:


kikisdeliveryservice:
Yay! 

@ashcrow Q: line 424 also uses ioutils.ReadAll. Do we need to add that too or will we not need to mock it bc we will/ intend to have dummy resp.Body input that we can test with?

ashcrow:
@kikisdeliveryservice both methods will be updated once this merges :-). The only one I think we can't change at the present is `checkFileContentsAndMode` (which is OK for now).

kikisdeliveryservice:
@ashcrow sounds good!
/lgtm

pkg/daemon/fsclient_test.go
```diff
@@ -92,3 +100,12 @@ func (f FsClientMock) Chown(name string, uid, gid int) error {
 func (f FsClientMock) WriteFile(filename string, data []byte, perm os.FileMode) error {
 	return updateErrorReturns(&f.WriteFileReturns)
 }
+
+// ReadFile provides a mocked implemention
+func (f FsClientMock) ReadFile(filename string) ([]byte, error) {
+	returnValues := f.ReadFileReturns[0]
```
jlebon: Shouldn't this be under the if statement to make sure the array is not empty first? E.g.

```go
	if len(f....) > 0 {
		returnValues := f....[0]
		f.... = f.ReadFileReturns[1:]
		return returnValues
	}
	panic("no more values!")
```

?

I guess, as it is it'll panic anyway with index out of range, which I suppose was the intent since the unit test knows exactly how many times it should be called? But then it seems unnecessary to do the `len` check since if we got to there it means there definitely is at least one element. So an alternative is to just simplify it to:

```go
	returnValues := f....[0]
	f.... = f.ReadFileReturns[1:]
	return returnValues
```

? (Though I still prefer the explicitness of the first alternative!)

ashcrow: The idea behind this is that:

- If you provide nothing it panics
- If you provide N returns and use N+X then X+1 will return the same last element (for example if I want everything to return no error I can just give it one nil, not 2, 5, 50, etc..)

ashcrow: I'm happy to do a follow on PR to make the panic explicit :-).

jlebon:
OK, with that last commit, that looks sane to me. :+1: 
/lgtm

Issue #137, opened 2018-10-17
MCD: Failure after reboot when applying new machine config 


yuqi-zhang:
Even with https://github.com/openshift/machine-config-operator/commit/9a53b9bcaafc707327df84fbf779796deb7fd038, I'm still seeing errors upon applying a new machine config requiring a restart:

```
# kubectl logs -p -n openshift-machine-config-operator machine-config-daemon-b2hvq
I1017 19:42:06.534306   15382 start.go:42] Version: 3.11.0-158-g8f6563ae
I1017 19:42:06.565619   15382 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I1017 19:42:06.565788   15382 daemon.go:81] Booted osImageURL:  ()
I1017 19:42:06.565806   15382 start.go:85] Calling chroot("/rootfs")
I1017 19:42:06.565835   15382 daemon.go:101] Starting MachineConfigDaemon
I1017 20:16:58.980305   15382 update.go:26] Updating node with new config
...
I1017 20:16:59.009178   15382 update.go:455] Updating OS to ://dummy
I1017 20:16:59.009225   15382 run.go:13] Running: /bin/pivot ://dummy
pivot version 0.0.1
I1017 20:16:59.022150   30919 run.go:27] Running: rpm-ostree status --json
I1017 20:16:59.144617   30919 run.go:27] Running: skopeo inspect docker://://dummy
time="2018-10-17T20:16:59Z" level=fatal msg="invalid reference format" 
F1017 20:16:59.201422   30919 run.go:33] skopeo: exit status 1
E1017 20:16:59.202069   15382 daemon.go:121] Marking degraded due to: exit status 255
I1017 20:16:59.223021   15382 daemon.go:122] Shutting down MachineConfigDaemon
```

yaml used: the example one in: https://docs.google.com/document/d/1V8JQLCj_IegPouNCYYTCUAccTJBiC1xdtE4yhb-Fgys/edit#heading=h.k3hh962j6y1r

RHCOS image: 

```
rpm-ostree status
State: idle
AutomaticUpdates: disabled
Deployments:
* ostree://rhcos:openshift/3.10/x86_64/os
                   Version: 4.0.6912 (2018-10-17 01:00:41)
                    Commit: 8999e2f9635ed074a7b52c62da87b49a0d6be40b07ac16c07024e5b580819739
```

Installer: latest

Further with:
https://github.com/openshift/machine-config-operator/pull/136

I see:
```
[...]

PR #142, opened 2018-10-19
Dockerfiles: base off of origin-base

jlebon:
Be closer to production by using the same base that all origin images
use. Notably, this has a full rootfs, with utilities like `/sbin/chroot`
which is needed for the MCD.

Resolves: #137


jlebon:
Yup, this fixes it for me!

abhinavdahiya:
The images created by the CI already have that as base layer. Its fine if we want developers create similar images.

/lgtm

PR #144, opened 2018-10-25
update hacking.md with oc commands and check for successful deployment.

kikisdeliveryservice:
Adding some of my notes to the mcd documentation.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -22,12 +22,30 @@ These instructions were tested with installer version https://github.com/openshi
    CVO as it owns the configmap and it will revert your changes:
 
    ```sh
+   # if using kubectl:
    kubectl delete ds cluster-version-operator -n openshift-cluster-version
+
+   # if using oc:
+   oc delete daemonset cluster-version-operator -n openshift-cluster-version --config=auth/kubeconfig
    ```
 
-   then:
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
No idea why this docfix is failing these tests..

ashcrow:
They are total flakes. No worries :smile:

HACKING.md
```diff
@@ -22,12 +22,30 @@ These instructions were tested with installer version https://github.com/openshi
    CVO as it owns the configmap and it will revert your changes:
 
    ```sh
+   # if using kubectl:
    kubectl delete ds cluster-version-operator -n openshift-cluster-version
+
+   # if using oc:
+   oc delete daemonset cluster-version-operator -n openshift-cluster-version --config=auth/kubeconfig
```
jlebon: Sidenote: I usually do `export KUBECONFIG=$(pwd)/auth/kubeconfig`, then you don't have to pass that switch every time.

kikisdeliveryservice: thanks for the tip jlebon!

abhinavdahiya:
Why is this oc and kubectl both, can't we just pick one? `oc` seems to be a super set of all the available features from kubectl, so just use `oc`?

i would rather add `export KUBECONFIG=<path to kubeconfig>` than add that `--config` flag to each command.

@ashcrow @kikisdeliveryservice

ashcrow:
We can pick one. @kikisdeliveryservice's PR was meant to provide openshift commands. I'll follow up and remove the kubernetes commands.

PR #145, opened 2018-10-25
mcd: fix race condition for annotation updates

yuqi-zhang:
Currently when we finish an update, we re-check the desired config
and indiscriminately set the current config to it. This can cause
an update to be lost since the desired config may have changed
inbetween the isDesired check and when the map is updated. Instead
we should be updating the config to the one we checked for in
isDesired.

Tested the update process, seems fine to work.


jlebon:
Awesome, great catch!
/lgtm

PR #146, opened 2018-10-25
templates: switch kubelet systemd unit to type notify

abhinavdahiya:
kubelet supports running as systemd unit and sends READY=1 signal when running.
https://github.com/kubernetes/kubernetes/blob/12cf5451952c8425a47334465f3e03d7ca181956/cmd/kubelet/app/server.go#L761-L762

/cc @crawford @sjenning @rphillips

### DIFF
--- a/pkg/controller/template/test_data/templates/aws/master/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/aws/master/units/kubelet.service
@@ -4,6 +4,7 @@ contents: |
   Wants=rpc-statd.service
 
   [Service]
+  Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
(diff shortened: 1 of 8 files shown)


rphillips:
/hold

We tried this before and hit an issue in crio. Has that issue been fixed?

crawford:
@rphillips Kubelet doesn't run in crio.

rphillips:
Maybe the issue was with the systemd we are using. I'm searching for the issue.

crawford:
@rphillips Are you thinking of https://github.com/opencontainers/runc/pull/1807?

rphillips:
@crawford Ah yes... Has that been resolved? Does it apply to this issue?

rphillips:
/hold cancel

Sorry for the noise... Thanks for finding the issue for me.

crawford:
@rphillips No, it doesn't apply here. That issue affects instances where we use Podman in a notify service unit.

PR #149, opened 2018-10-30
pkg/daemon: gofmt changes

sdemos:
my editor runs gofmt on write and I ran into these because of that.

I wonder if we can hook gofmt into the merge bot somehow?

/cc @ashcrow

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -3,9 +3,9 @@ package daemon
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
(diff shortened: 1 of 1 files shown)


ashcrow:
Thanks for this @sdemos. If possible, do you mind looking at https://github.com/openshift/machine-config-operator/pull/126? If that's close I can pull this in after that merges (or part of it). I'm trying not to rebase for a 4th time :laughing:.

ashcrow:
/close

We'll take care of this in a clean up PR.

Issue #150, opened 2018-10-31
MCD CrashLoopBackOff skopeo inspect docker://://dummy ... invalid reference format


sjenning:
MCD on worker is in CrashLoopBackOff
```
$ oc get machineconfig
NAME                               AGE
00-master                          14m
00-worker                          14m
3c6b6d8063551a7c8a4561755ea5bae5   14m
f0db02dc9fb258ad37bed110741198c1   14m

$ oc get pod -o wide
NAME                                         READY     STATUS             RESTARTS   AGE       IP               NODE                 NOMINATED NODE
machine-config-controller-57c78dc856-xkr6w   1/1       Running            0          15m       10.2.0.20        dev-master-0         <none>
machine-config-daemon-cgmdx                  0/1       CrashLoopBackOff   7          13m       192.168.126.51   dev-worker-0-qplk8   <none>
machine-config-daemon-dm8kj                  1/1       Running            0          14m       192.168.126.11   dev-master-0         <none>
machine-config-operator-76c4d5db69-xxt42     1/1       Running            0          18m       10.2.0.9         dev-master-0         <none>
machine-config-server-m565j                  1/1       Running            0          14m       192.168.126.11   dev-master-0         <none>

$ oc logs machine-config-daemon-dm8kj -n openshift-machine-config-operator
I1031 15:49:02.518478    7375 start.go:42] Version: 3.11.0-195-g214b930d-dirty
I1031 15:49:02.549275    7375 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I1031 15:49:02.632582    7375 daemon.go:95] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:cf407efbac50e1499c39a8fe4af46cb423aa87cb36264a25cff6c294bee342eb (47.36)
I1031 15:49:02.632690    7375 start.go:90] Calling chroot("/rootfs")
I1031 15:49:02.632738    7375 start.go:100] Starting MachineConfigDaemon
E1031 15:49:02.655460    7375 daemon.go:163] Marking degraded due to: machineconfigs.machineconfiguration.openshift.io "f0db02dc9fb258ad37bed110741198c1" not found

$ oc logs machine-config-daemon-cgmdx -n openshift-machine-config-operator
I1031 15:57:03.421947    5277 start.go:42] Version: 3.11.0-195-g214b930d-dirty
[...]

PR #151, opened 2018-10-31
daemon: Adapt to OS being pre-pivoted

jlebon:
Now that RHCOS is pre-pivoted (see
https://github.com/openshift/os/issues/307), we need to tweak the logic
here on how we handle `osImageURL`. Notably, we allow `://dummy` from
the installer to always match the current image, regardless of whether
we're pivoted or not. In fact, that should've been the logic from the
start.

Closes: #150

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -321,6 +321,12 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 
 // checkOS validates the OS image URL and returns true if they match.
 func (dn *Daemon) checkOS(osImageURL string) (bool, error) {
+	// XXX: The installer doesn't pivot yet so for now, just make "://dummy"
+	// match anything. See also: https://github.com/openshift/installer/issues/281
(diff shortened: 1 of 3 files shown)


jlebon:
/hold

Not tested yet.

ashcrow:
```
pkg/daemon/daemon.go:328:3: not enough arguments to return
	have (bool)
	want (bool, error)
pkg/daemon/rpm-ostree.go:8:2: imported and not used: "github.com/openshift/machine-config-operator/vendor/github.com/golang/glog"
pkg/daemon/daemon.go:328:3: not enough arguments to return
	have (bool)
	want (bool, error)
pkg/daemon/rpm-ostree.go:8:2:
```

jlebon:
This is blocked on libvirt installations working again so that I can test this change.

ashcrow:
Will hand off review to @sdemos, but this makes sense to me.

ashcrow:
@jlebon remove the hold when ready

PR #155, opened 2018-11-02
templates: use cluster-name.base-domain for etcd srv search

abhinavdahiya:
Requires: https://github.com/openshift/installer/pull/596

/cc @crawford

PR #156, opened 2018-11-02
templates: fix etcd discovery-srv endpoint

abhinavdahiya:
https://github.com/openshift/machine-config-operator/pull/155 missed changing the
--discovery-srv flag on etcd to cluster-name.basedomain

/cc @crawford


abhinavdahiya:
@wking tests are green. :)

PR #157, opened 2018-11-05
[WIP] Switch to using library-go helpers to read installconfig from file

damemi:
Depends on https://github.com/openshift/library-go/pull/77/

Pulls code that parses `cluster-config.yaml` from file into a helper in `library-go`. This change uses that helper


damemi:
cc @sttts

abhinavdahiya:
/close

Looks like this is obsolete approach

PR #164, opened 2018-11-10
Pass ssh keys from clusterconfig to machineconfig

kikisdeliveryservice:
Closes https://github.com/openshift/installer/issues/578

Summary of issue: the ClusterConfig has the SSH key and while that key is being passed into the MCO, the MCO isn't properly adding it to the MachineConfig. This prevents us from being able add functionality to MCD updating existing SSH keys in a MachineConfig's Spec.Config.Passwd.Users.

Comments & feedback welcome!
cc: @abhinavdahiya @wking

### DIFF
--- a/lib/resourcemerge/machineconfig.go
+++ b/lib/resourcemerge/machineconfig.go
@@ -57,7 +57,7 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
 	setStringIfSet(modified, &existing.ClusterName, required.ClusterName)
 	setStringIfSet(modified, &existing.Platform, required.Platform)
 	setStringIfSet(modified, &existing.BaseDomain, required.BaseDomain)
-	setStringIfSet(modified, &existing.Platform, required.Platform)
+	setStringIfSet(modified, &existing.SSHKey, required.SSHKey)
 
 	setBytesIfSet(modified, &existing.EtcdCAData, required.EtcdCAData)
 	setBytesIfSet(modified, &existing.RootCAData, required.RootCAData)
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -58,6 +58,8 @@ type MCOConfigSpec struct {
 	Platform string `json:"platform"`
 
 	BaseDomain string `json:"baseDomain"`
+
+	SSHKey string `json:"sshKey"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -129,6 +131,9 @@ type ControllerConfigSpec struct {
 	// PullSecret is the default pull secret that needs to be installed
 	// on all machines.
 	PullSecret *corev1.ObjectReference `json:"pullSecret,omitempty"`
+
+	// Public SSH
+	SSHKey string `json:"sshKey"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -61,6 +61,7 @@ func generateMachineConfigs(config *renderConfig, templateDir string) ([]*mcfgv1
 	}
 
 	cfgs := []*mcfgv1.MachineConfig{}
+
 	for _, info := range infos {
 		if !info.IsDir() {
 			glog.Infof("ignoring non-directory path %q", info.Name())
@@ -83,8 +84,17 @@ func generateMachineConfigsForRole(config *renderConfig, role string, path strin
 	if err != nil {
 		return nil, fmt.Errorf("failed to read dir %q: %v", path, err)
 	}
+	// for each role a machine config is created containing the sshauthorized keys to allow for ssh access
+	// ex: role = worker -> machine config "00-worker-ssh" created containing user core and ssh key
+	var tempIgnConfig ignv2_2types.Config
+	tempUser := ignv2_2types.PasswdUser{Name: "core", SSHAuthorizedKeys: []ignv2_2types.SSHAuthorizedKey{ignv2_2types.SSHAuthorizedKey(config.SSHKey)}}
+	tempIgnConfig.Passwd.Users = append(tempIgnConfig.Passwd.Users, tempUser)
+	sshConfigName := "00-" + role + "-ssh"
+	sshMachineConfigForRole := machineConfigFromIgnConfig(role, sshConfigName, &tempIgnConfig)
 
 	cfgs := []*mcfgv1.MachineConfig{}
+	cfgs = append(cfgs, sshMachineConfigForRole)
+
 	for _, info := range infos {
 		if !info.IsDir() {
 			glog.Infof("ignoring non-directory path %q", info.Name())
@@ -119,7 +129,6 @@ func generateMachineConfigForName(config *renderConfig, role, name, path string)
 
 	files := map[string]string{}
 	units := map[string]string{}
-
 	// walk all role dirs, with later ones taking precedence
 	for _, platformDir := range platformDirs {
 		// magic param
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -56,6 +56,10 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 		return err
 	}
 
+	if err = dn.updateSSHKeys(newConfig.Spec.Config.Passwd.Users); err != nil {
+		return err
+	}
+
 	// TODO: Change the logic to be clearer
 	// We need to skip draining of the node when we are running once
 	// and there is no cluster.
@@ -93,7 +97,9 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 // we can only update machine configs that have changes to the files,
 // directories, links, and systemd units sections of the included ignition
 // config currently.
+
 func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *string {
+	glog.Info("Checking if configs are reconcilable")
 	// We skip out of reconcilable if there is no Kind and we are in runOnce mode. The
 	// reason is that there is a good chance a previous state is not available to match against.
 	if oldConfig.Kind == "" && dn.onceFrom != "" {
@@ -131,8 +137,25 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 	// we don't currently configure groups or users in place. we can't fix it if
 	// something changed here.
 	if !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd) {
-		msg := "Ignition passwd section contains changes"
-		return &msg
+		if !reflect.DeepEqual(oldIgn.Passwd.Groups, newIgn.Passwd.Groups) {
+			msg := "Ignition Passwd Groups section contains changes"
+			return &msg
+		}
+		// check if the prior config is empty and that this is the first time running.
+		// if so, the SSHKey from the cluster config and user "core" must be added to machine config,.
+		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
+			if len(oldIgn.Passwd.Users) == 0 && len(newIgn.Passwd.Users) == 1 {
+				if newIgn.Passwd.Users[0].Name == "core" && len(newIgn.Passwd.Users[0].SSHAuthorizedKeys) > 0 {
+					glog.Info("SSH Keys reconcilable")
+				} else {
+					msg := "Ignition passwd user section contains unsupported changes"
+					return &msg
+				}
+			}
+		} else {
+			msg := "Ignition passwd section contains unsupported changes"
+			return &msg
+		}
 	}
 
 	// Storage section
@@ -462,6 +485,37 @@ func getFileOwnership(file ignv2_2types.File) (int, int, error) {
 		}
 	}
 	return uid, gid, nil
+}
+
+// Update a given PasswdUser's SSHKey
+func (dn *Daemon) updateSSHKeys(newUsers []ignv2_2types.PasswdUser) error {
+	// Keys should only be written to "/home/core/.ssh"
+	// Once Users are supported fully this should be writing to PasswdUser.HomeDir
+	if newUsers[0].Name != "core" {
+		// Double checking that we are only writing SSH Keys for user "core"
+		return fmt.Errorf("Expecting user core. Got %s instead", newUsers[0].Name)
+	}
+	sshDirPath := filepath.Join("/home", newUsers[0].Name, ".ssh")
+	// we are only dealing with the "core" User at this time, so only dealing with the first entry in Users[]
+	glog.Infof("Writing SSHKeys at %q:", sshDirPath)
+	if err := dn.fileSystemClient.MkdirAll(filepath.Dir(sshDirPath), os.FileMode(0600)); err != nil {
+		return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(sshDirPath), err)
+	}
+	glog.V(2).Infof("Created directory: %s", sshDirPath)
+
+	authkeypath := filepath.Join(sshDirPath, "authorized_keys")
+	var concatSSHKeys string
+	for _, k := range newUsers[0].SSHAuthorizedKeys {
+		concatSSHKeys = concatSSHKeys + string(k) + "\n"
+	}
+
+	if err := dn.fileSystemClient.WriteFile(authkeypath, []byte(concatSSHKeys), os.FileMode(0600)); err != nil {
+		return fmt.Errorf("Failed to write ssh key: %v", err)
+	}
+
+	glog.V(2).Infof("Wrote SSHKeys at %s", sshDirPath)
+
+	return nil
 }
 
 // updateOS updates the system OS to the one specified in newConfig
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -9,8 +9,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
-	configclientset "github.com/openshift/client-go/config/clientset/versioned"
-	installertypes "github.com/openshift/installer/pkg/types"
+
 	"k8s.io/api/core/v1"
 	apiextclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apiextinformersv1beta1 "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1beta1"
@@ -28,7 +27,9 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
-
+	
+	configclientset "github.com/openshift/client-go/config/clientset/versioned"
+	installertypes "github.com/openshift/installer/pkg/types"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/scheme"
@@ -310,6 +311,7 @@ func getRenderConfig(mc *mcfgv1.MCOConfig, etcdCAData, rootCAData []byte, ps *v1
 		EtcdCAData:          etcdCAData,
 		RootCAData:          rootCAData,
 		PullSecret:          ps,
+		SSHKey:              mc.Spec.SSHKey,
 	}
 	return renderConfig{
 		TargetNamespace:  mc.GetNamespace(),
--- a/pkg/controller/template/render_test.go
+++ b/pkg/controller/template/render_test.go
@@ -312,6 +312,38 @@ func TestGenerateMachineConfigs(t *testing.T) {
 	}
 }
 
+func TestGenerateMachineConfigsSSH(t *testing.T) {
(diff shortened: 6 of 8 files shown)


pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -125,6 +126,9 @@ type ControllerConfigSpec struct {
 	// CAs
 	EtcdCAData []byte `json:"etcdCAData"`
 	RootCAData []byte `json:"rootCAData"`
+
+	// Admin
+	Admin installertypes.Admin `json:"admin"`
```
abhinavdahiya: i would like not import installer's type as there are more fields than we use.

pkg/controller/template/render.go
```diff
@@ -191,7 +191,7 @@ func generateMachineConfigForRole(config *renderConfig, roleName string, roleDir
 		return vs
 	}
 
-	ignCfg, err := transpileToIgn(keySortVals(files), keySortVals(units))
+	ignCfg, err := transpileToIgn(keySortVals(files), keySortVals(units), config.Admin.SSHKey)
```
abhinavdahiya: WDYT
creating a machineconfig object with ssh key setup and labeling it such that both master and worker machineconfigpools pick it up. (ie. add both `...role=worker` and `...role=master` labels to this new object).

pkg/apis/machineconfiguration.openshift.io/v1/types.go
kikisdeliveryservice: Will do!

pkg/controller/template/render.go
ashcrow: Making a different section for ssh key items worries me a bit. Why not use the Ignition section and honor it for both initial setup and future changes?

kikisdeliveryservice: good point @ashcrow. Even if there is a seperate ssh key item, the sshkey still needs to be set in Passwd.Users otherwise the MCD won't be able to update it later.

abhinavdahiya: I'm not sure I follow the @ashcrow  comment... :innocent: do you mind elaborating a little bit

ashcrow: Sure :smile:. `Ignition` has a section that deals with ssh keys. We have work that's waiting to merge that will honor that `Ignition` section for both MCD as a daemon usage as well as `onceFrom` usage in BYO. Why not continue to use this section for all ssh key modification/addition/removal? Or, put another way, why have a second place that ssh key content may live?

abhinavdahiya: That's exactly i would expect the ssh keys to live, even if you create a new machineconfig object with ignition that only has the `user` section set and label it correctly. MCC will combine it to single ignition blob that has all.

ashcrow: @abhinavdahiya Perfect. I think we're all on the same page on where the content would live. Just to verify, it sounds like you recommend adding `worker` and `master` labels to the MC as well, is that correct?

abhinavdahiya: > @abhinavdahiya Perfect. I think we're all on the same page on where the content would live. Just to verify, it sounds like you recommend adding `worker` and `master` labels to the MC as well, is that correct?

Yes to the new standalone MC we create for SSH keys.

kikisdeliveryservice:
Notes: The best that I can tell is that the sshkeys need to move this way to make it to the MachineConfig: installerConfig(Admin.SSHKey)-> MCOConfig --> RenderConfig/ControllerConfig --> MachineConfig (Passwd.Users)

kikisdeliveryservice:
The best that I can tell is that the sshkeys need to move this way to make it to the MachineConfig: installerConfig(Admin.SSHKey)-> MCOConfig --> RenderConfig/ControllerConfig --> MachineConfig (Passwd.Users)

Is this flow correct?  I've traced through quite a few times, but in need of expertise on this. Can someone PTAL and let me know if my logic is correct?
cc: @ashcrow @abhinavdahiya @wking

jlebon:
Looks like this needs a rebase.

kikisdeliveryservice:
Rebased! @jlebon :)

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -129,6 +131,9 @@ type ControllerConfigSpec struct {
 	// PullSecret is the default pull secret that needs to be installed
 	// on all machines.
 	PullSecret *corev1.ObjectReference `json:"pullSecret,omitempty"`
+
+	// Public SSH
+	SSHKey string `json:"sshKey"`
```
jlebon: Only a single key? Would it make sense to support multiple keys here even if the installer right now only lets one specify one only?

ashcrow: If an array of strings doesn't increase the complexity I say go ahead. Otherwise, it's fine to hold off until the feature set is expected/available at the higher level.

kikisdeliveryservice:
Update: 
Confirming that the SSH Key made it all the way from installerConfig(Admin.SSHKey)-> MCOConfig --> RenderConfig/ControllerConfig --> MachineConfig (Passwd.Users).

Now checking the machineconfigs for both master and worker shows (example)
```
$ oc get machineconfig 00-worker  -o yaml
kind: MachineConfig
metadata:
  creationTimestamp: 2018-12-04T01:07:56Z
  generation: 1
  labels:
  name: 00-worker
  ownerReferences:
  - apiVersion: machineconfiguration.openshift.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ControllerConfig
    name: machine-config-controller
    uid: fae0efad-f760-11e8-b3cc-42f92e967bf5
  resourceVersion: "23232"
  selfLink: /apis/machineconfiguration.openshift.io/v1/machineconfigs/00-worker
  uid: 088750c9-f761-11e8-b3cc-42f92e967bf5
spec:
  config:
    ignition:
      config: {}
      security:
        tls: {}
      timeouts: {}
      version: 2.2.0
    networkd: {}
    passwd:
      users:
      - name: core
        sshAuthorizedKeys:
        - |
          ssh-rsa ABC123....
```

kikisdeliveryservice:
However, I am seeing the daemon degrade and I'm not sure if this is this a product of _when_ and _how_ I'm applying my binaries and their timing or if I also need to allow reconcilable to account for this initialization of the configs...  I'm going to try to test this with better timing tomorrow (assuming I can get a cluster up again) as I think the mcd got a headstart on the mco/mcc when I applied my binaries. But any thoughts are welcome.  This is the last thing holding the PR back.
```
$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-289tf
I1204 01:30:42.968291    5400 start.go:51] Version: 3.11.0-301-g13e15273
I1204 01:30:42.968962    5400 start.go:88] starting node writer
I1204 01:30:42.974021    5400 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I1204 01:30:43.054368    5400 daemon.go:120] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:e4b05527d762ba159c821d53bc8b2478c38b717b0fcfb4b76c57787c2f46ee2c (47.177)
I1204 01:30:43.072829    5400 start.go:139] Calling chroot("/rootfs")
I1204 01:30:43.072944    5400 daemon.go:294] CheckStateOnBoot
I1204 01:30:43.085725    5400 update.go:87] Checking if configs are reconcilable!
I1204 01:30:43.085837    5400 update.go:100] Old Config: {[] []}
I1204 01:30:43.085911    5400 update.go:101] New Config: {[] []}
I1204 01:30:43.085947    5400 daemon.go:565] No target osImageURL provided
I1204 01:30:43.095044    5400 start.go:158] Starting MachineConfigDaemon
I1204 01:30:43.095059    5400 daemon.go:195] Enabling Kubelet Healthz Monitor
I1204 01:30:52.359987    5400 daemon.go:371] handleNodeUpdate
I1204 01:30:52.629908    5400 daemon.go:371] handleNodeUpdate
I1204 01:33:09.537006    5400 update.go:33] Updating node with new config
I1204 01:33:09.538536    5400 update.go:87] Checking if configs are reconcilable!
I1204 01:33:09.538767    5400 update.go:100] Old Config: {[] []}
I1204 01:33:09.538859    5400 update.go:101] New Config: {[] [{<nil>  []  core false false false <nil>  [ssh-rsa ABC123]  false <nil>}]}
W1204 01:33:09.538950    5400 update.go:133] daemon can't reconcile state!
W1204 01:33:09.538990    5400 update.go:134] Ignition passwd section contains changes
I1204 01:33:09.539039    5400 daemon.go:385] Unable to apply update: daemon can't reconcile this config
I1204 01:33:09.539076    5400 daemon.go:371] handleNodeUpdate
E1204 01:33:09.539286    5400 writer.go:85] Marking degraded due to: daemon can't reconcile this config
F1204 01:33:09.548695    5400 start.go:163] failed to run: daemon can't reconcile this config
```

kikisdeliveryservice:
cc @ashcrow @abhinavdahiya (some updates and questions above)

ashcrow:
@kikisdeliveryservice it looks like the reconcilable needs updating to allow for a different Password.User section: https://github.com/openshift/machine-config-operator/pull/164/files#diff-06961b075f1753956d802ba954d2cfb5R133

ashcrow:
Apologies, the above paste was an "expected error" at the time. The actual issue being hit is #203.

kikisdeliveryservice:
Getting blocked on this by: https://github.com/openshift/origin/issues/21612
Can't get a running cluster up.

ashcrow:
Rebase required.

pkg/daemon/update.go
```diff
@@ -512,6 +528,32 @@ func (dn *Daemon) logSystem(format string, a ...interface{}) {
...
+	// we are only dealing with the "core" User at this time, so only dealing with the first entry in Users[]
+	glog.Infof("Writing SSHKeys at %q:", sshDirPath)
+	if err := dn.fileSystemClient.MkdirAll(filepath.Dir(sshDirPath), os.FileMode(0600)); err != nil {
+		return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(sshDirPath), err)
+	}
+	glog.V(2).Infof("Created directory: %s", sshDirPath)
+
+	authkeypath := filepath.Join(sshDirPath, "authorized_keys")
+	var concatSSHKeys string
+	for _, k := range newUsers[0].SSHAuthorizedKeys {
```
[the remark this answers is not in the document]
kikisdeliveryservice: That is because right now we only support SSH Keys for user "core"

kikisdeliveryservice:
@cgwalters I can't reply to your last comment directly, but for now the only SSH Keys supported will be for "core".  Those keys will be able to be updated in another pending PR, but first the keys need to be passed from the install config ----> mcd.

kikisdeliveryservice:
Reworking based on #208

kikisdeliveryservice:
Can confirm that mc with sshkeys are added:
$ oc get machineconfigs
NAME                                      AGE
00-master                                 1h
00-master-ssh                             1h
00-worker                                 1h
00-worker-ssh                             1h
01-master-kubelet                         1h
01-worker-kubelet                         1h

Once the MCD picks up my new image (both masters and workers pick it up):
$ oc logs -p -n openshift-machine-config-operator machine-config-daemon-5s9wr 
I1220 00:19:05.562689    7542 daemon.go:381] handleNodeUpdate
I1220 00:19:06.207150    7542 update.go:37] Checking reconcilable for config master-6d45c76e5936a50db670fe9f76db9df1 to master-1df9375c795afb9e5c07dc03157fa182
I1220 00:19:06.207173    7542 update.go:102] Checking if configs are reconcilable!
I1220 00:19:06.207183    7542 update.go:115] Old Config: {[] []}
I1220 00:19:06.207194    7542 update.go:116] New Config: {[] [{<nil>  []  core false false false <nil>  [ssh-rsa AAAAB3...]  false <nil>}]}
I1220 00:19:06.207265    7542 update.go:152] SSH Keys reconcilable
I1220 00:19:06.207309    7542 update.go:214] Updating files
I1220 00:19:06.207333    7542 update.go:402] Writing file "/etc/containers/registries.conf"
I1220 00:19:06.211255    7542 update.go:402] Writing file "/etc/kubernetes/manifests/etcd-member.yaml"
I1220 00:19:06.212990    7542 update.go:402] Writing file "/etc/sysconfig/crio-network"
I1220 00:19:06.218184    7542 update.go:402] Writing file "/etc/kubernetes/static-pod-resources/etcd-member/ca.crt"
I1220 00:19:06.222016    7542 update.go:402] Writing file "/etc/kubernetes/static-pod-resources/etcd-member/root-ca.crt"
I1220 00:19:06.223319    7542 update.go:402] Writing file "/var/lib/kubelet/config.json"
I1220 00:19:06.225140    7542 update.go:402] Writing file "/etc/kubernetes/ca.crt"
I1220 00:19:06.226932    7542 update.go:402] Writing file "/etc/sysctl.d/forward.conf"
I1220 00:19:06.228609    7542 update.go:402] Writing file "/etc/kubernetes/kubelet.conf"
I1220 00:19:06.229826    7542 update.go:336] Writing systemd unit "kubelet.service"
I1220 00:19:06.230091    7542 update.go:374] Enabling systemd unit "kubelet.service"
I1220 00:19:06.230513    7542 update.go:283] /etc/systemd/system/multi-user.target.wants/kubelet.service already exists. Not making a new symlink
I1220 00:19:06.230829    7542 update.go:235] Deleting stale data
I1220 00:19:06.230855    7542 update.go:522] No target osImageURL provided
I1220 00:19:06.230864    7542 update.go:493] Writing SSHKeys at "/home/core/.ssh":
I1220 00:19:06.231608    7542 update.go:497] Created directory: /home/core/.ssh
I1220 00:19:06.232234    7542 update.go:509] Wrote SSHKeys at /home/core/.ssh
...
I1220 00:19:06.232478    7542 update.go:537] machine-config-daemon initiating reboot: Node will reboot into config master-1df9375c795afb9e5c07dc03157fa182
`

pkg/daemon/update.go
```diff
@@ -56,6 +56,10 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 		return err
 	}
 
+	if err = dn.writeSSHKeys(newConfig.Spec.Config.Passwd.Users); err != nil {
```
jlebon: Should this be `updateSSHKeys` for consistency?

pkg/daemon/update.go
```diff
@@ -131,8 +143,18 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
...
 	if !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd) {
-		msg := "Ignition passwd section contains changes"
-		return &msg
+		// check if the prior config is empty and that this is the first time running.
+		// if so, the SSHKey from the cluster config and user "core" must be added to machine config,.
+		// emptyMC := mcfgv1.MachineConfig{}
+		// if reflect.DeepEqual(emptyMC.Spec.Config.Passwd, oldIgn.Passwd) && oldConfig.Kind == "" {
+		// 	glog.Infof("Reconcilable: Initializing ssh keys")
+		if len(newIgn.Passwd.Users[0].SSHAuthorizedKeys) > 0 {
+			glog.Info("SSH Keys reconcilable")
```
jlebon: Hmm, I think we need to be even stronger here, no? There's lots of other changes that could've snuck into there that aren't reconcilable. For the time being, if the only thing we want to support is SSH key reconciliation for a single user, we can do something like this:

1. rule out group changes by doing `DeepEqual` on `Passwd.Groups`
2. check that `len(Passwd.Users)` on both are the same and equal to 1
3. make *copies* of the old and new `Passwd.Users[0]` and zero out the `SSHAuthorizedKeys` elements and then `DeepEqual` those instead. That way, SSH key changes don't interfere with the `DeepEqual` check.

Not sure if there's a cleaner way of doing this short of busting out golang reflect and checking every field.

pkg/daemon/update.go
```diff
@@ -464,6 +486,31 @@ func getFileOwnership(file ignv2_2types.File) (int, int, error) {
 	return uid, gid, nil
 }
 
+// Update a given PasswdUser's SSHKey
+func (dn *Daemon) writeSSHKeys(newUsers []ignv2_2types.PasswdUser) error {
+	sshDirPath := "/home/core/.ssh"
```
jlebon: `core` here should come from the username in the spec, right?

kikisdeliveryservice: Yep, I was going to move some of my logic from my older PR over https://github.com/openshift/machine-config-operator/blob/3eff300a83970775dcc2ede2b17d17752bb4c535/pkg/daemon/update.go#L348

kikisdeliveryservice: In the future it would come from HomeDir but we don't support Users at this time.

So I will use the name for now and make a note.

abhinavdahiya:
Yeah this definitely needs a squash, and it might be a good chance to use https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

lib/resourcemerge/machineconfig.go
```diff
@@ -52,12 +53,14 @@ func ensureMachineConfigSpec(modified *bool, existing *mcfgv1.MachineConfigSpec,
 }
 
 func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfigSpec, required mcfgv1.ControllerConfigSpec) {
+	glog.Info("ensureControllerConfigSpec")
```
abhinavdahiya: drop this

lib/resourcemerge/machineconfig.go
```diff
@@ -52,12 +53,14 @@ func ensureMachineConfigSpec(modified *bool, existing *mcfgv1.MachineConfigSpec,
...
 	setStringIfSet(modified, &existing.ClusterDNSIP, required.ClusterDNSIP)
 	setStringIfSet(modified, &existing.CloudProviderConfig, required.CloudProviderConfig)
 	setStringIfSet(modified, &existing.ClusterName, required.ClusterName)
 	setStringIfSet(modified, &existing.Platform, required.Platform)
 	setStringIfSet(modified, &existing.BaseDomain, required.BaseDomain)
-	setStringIfSet(modified, &existing.Platform, required.Platform)
+
```
abhinavdahiya: why the extra line?

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -58,6 +58,8 @@ type MCOConfigSpec struct {
 	Platform string `json:"platform"`
 
 	BaseDomain string `json:"baseDomain"`
+
+	SSHKey string
```
abhinavdahiya: missing json tag

pkg/controller/template/render.go
```diff
@@ -47,6 +47,13 @@ const (
 //                                    /files/hostname.tmpl
 //
 func generateMachineConfigs(config *renderConfig, templateDir string) ([]*mcfgv1.MachineConfig, error) {
+	var tempIgnConfig ignv2_2types.Config
```
abhinavdahiya: this piece of code assumes only 2 roles worker,master which conflicts with the other part of the code that discovers the roles based on filestructure...

pkg/operator/operator.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/golang/glog"
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	installertypes "github.com/openshift/installer/pkg/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
```
abhinavdahiya: the ideal case for import grouping that i would like is:
```
<stdlib>

<external>

<k8s>

<internal>
```

pkg/operator/sync.go
```diff
@@ -5,15 +5,14 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/openshift/machine-config-operator/lib/resourceapply"
```
abhinavdahiya: same as above.

kikisdeliveryservice:
@ashcrow @abhinavdahiya Was definitely planning on squashing, but wanted to get a review settled before I removed all of my commits. :)

Taking a look at the comments now, thank you!

pkg/operator/operator.go
wking: > the ideal case for import grouping...

Why split out k8s?  Alphabetizing non-stdlib packages will get them sorted together anyway, and `gofmt` will (I think?) enforce that ordering witin a block of contiguous imports.

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: Done.

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: Done.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
kikisdeliveryservice: Done.

pkg/controller/template/render.go
kikisdeliveryservice: In progress.

pkg/operator/sync.go
kikisdeliveryservice: Done.

pkg/operator/operator.go
kikisdeliveryservice: Made the change but lmk if you want to use a diff import format as @wking suggested.

pkg/controller/template/render.go
kikisdeliveryservice: @abhinavdahiya PTAL at new commit with test [here](https://github.com/openshift/machine-config-operator/pull/164/commits/9527f23128fa3ba4016b6f9c7b7f10f602fc6e9d)

kikisdeliveryservice:
FYI: just squashed my last commit in since I got @abhinavdahiya 's final review. :)

kikisdeliveryservice:
@ashcrow I know you already approved this, so could I get a LGTM so this can be merged? Tests aren't flaking anymore.

PR #165, opened 2018-11-11
Add a RHEL7 dockerfile and standarize format

smarterclayton:


abhinavdahiya:
the tests are green, @ashcrow do you have any concerns? otherwise i'll `/lgtm`

Dockerfile.setup-etcd-environment
```diff
@@ -1,10 +1,8 @@
...
+COPY . .
 RUN WHAT=setup-etcd-environment ./hack/build-go.sh
 
-FROM openshift/origin-base:v4.0.0
-COPY --from=build-env /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/setup-etcd-environment /bin/setup-etcd-environment
-
-ENTRYPOINT ["/bin/setup-etcd-environment"]
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/setup-etcd-environment /usr/bin/
+ENTRYPOINT ["/usr/bin/setup-etcd-environment"]
```
cgwalters: LGTM overall, only very minor nit is some of the files don't have ending newlines.

abhinavdahiya:
e2e-failure https://github.com/openshift/cluster-network-operator/issues/32

smarterclayton:
Tests went green, what would be the most unambiguous signal that the MCD was working (besides a generic crashloopnig test that we will add)?

abhinavdahiya:
The OS team has been planning some e2e test for the daemon. that should catch these. cc @ashcrow

smarterclayton:
I've got a "are any pods in the core suites crash looping" e2e test coming up that will be in the guard suites.

PR #169, opened 2018-11-14
Ignition parse

michaelgugino:
WIP: Enable from-ignition option

This commit enables mcd run-once to parse ignition file
directly instead of needing a MachineConfig.

This is useful first-booted clusters in BYO-RHEL

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -21,10 +21,11 @@ var (
 	}
 
 	startOpts struct {
-		kubeconfig string
-		nodeName   string
-		rootMount  string
-		onceFrom   string
+		kubeconfig   string
+		nodeName     string
+		rootMount    string
+		onceFrom     string
+		fromIgnition bool
 	}
 )
 
@@ -34,6 +35,7 @@ func init() {
 	startCmd.PersistentFlags().StringVar(&startOpts.nodeName, "node-name", "", "kubernetes node name daemon is managing.")
 	startCmd.PersistentFlags().StringVar(&startOpts.rootMount, "root-mount", "/rootfs", "where the nodes root filesystem is mounted for chroot and file manipulation.")
 	startCmd.PersistentFlags().StringVar(&startOpts.onceFrom, "once-from", "", "Runs the daemon once using a provided file path or URL endpoint as its machine config source")
+	startCmd.PersistentFlags().BoolVar(&startOpts.fromIgnition, "from-ign", false, "Configures run-once mode to use ignition file directly instead of MachineConfig")
 }
 
 func runStartCmd(cmd *cobra.Command, args []string) {
@@ -65,7 +67,6 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		}
 		glog.Fatalf("unable to verify rootMount %s exists: %s", startOpts.rootMount, err)
 	}
-
 	stopCh := make(chan struct{})
 	defer close(stopCh)
 	var dn *daemon.Daemon
@@ -81,6 +82,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 			daemon.NewNodeUpdaterClient(),
 			daemon.NewFileSystemClient(),
 			startOpts.onceFrom,
+			startOpts.fromIgnition,
 		)
 		if err != nil {
 			glog.Fatalf("failed to initialize single run daemon: %v", err)
@@ -103,6 +105,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 			cb.KubeClientOrDie(componentName),
 			daemon.NewFileSystemClient(),
 			startOpts.onceFrom,
+			startOpts.fromIgnition,
 			ctx.KubeInformerFactory.Core().V1().Nodes(),
 		)
 		if err != nil {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/coreos/go-systemd/login1"
+	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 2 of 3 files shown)


cmd/machine-config-daemon/start.go
```diff
@@ -65,22 +67,22 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
-	if startOpts.onceFrom != "" && daemon.ValidPath(startOpts.onceFrom) {
+	if startOpts.onceFrom != "" {
 		dn, err = daemon.New(
 			startOpts.rootMount,
 			startOpts.nodeName,
 			operatingSystem,
 			daemon.NewNodeUpdaterClient(),
 			daemon.NewFileSystemClient(),
 			startOpts.onceFrom,
+      startOpts.fromIgnition,
```
vrutkovs: Spaces would make `gofmt` unhappy

ashcrow:
If we go this route I'd like to *replace* the MC usage for `onceFrom` and just support `Ignition` files. No change would be needed in this PR, but I'd follow up re-plumbing to that end.

ashcrow:
Note: the previous commits in this PR have been merged.

sdodson:
@ashcrow meeting being scheduled to go through this ASAP

ashcrow:
@sdodson I actually just tried to keep from yet another meeting to occur :frowning:

eparis:
It sounds like the install/byo side of the house likes this path, but we want to know if it makes sense in your MC* architecture or if we need to discuss it further.

thanks @ashcrow !

ashcrow:
@eparis From the MCD POV this change is fine. We'll actually follow up with removing the MC parsing from `onceFrom` so we don't carry a feature we're not using.

I'm :+1: to merge this if this is what will help @aaronlevy, @sdodson, and @michaelgugino.

ashcrow:
(once the hold is removed of course)

imcleod:
FWIW - Giving my :+1: as well.

sdodson:
This is my preferred solution. I'd leave in the ability to run against a MC though. Imagine a scenario where a host has somehow become out of date and unable to attach to the cluster, would `mcd --once-from https://api/mco/worker` be a solution to bringing it back under management?

ashcrow:
@sdodson the MC within the cluster should suffice in that scenario as the MCD would pick up the new MC, apply it, reboot, and rejoin. I don't think there would be a need to run it only once before having it read the MC and apply the needed configuration.

cgwalters:
Makes sense to me.  

>  Imagine a scenario where a host has somehow become out of date and unable to attach to the cluster, would mcd --once-from https://api/mco/worker be a solution to bringing it back under management?

Eh, in an IaaS environment I imagine everyone would just destroy instances like that (which speaking of, can the cluster api tooling detect that?  Does it record historical instance IDs?)

In a bare metal environment I can imagine someone wanting it but I'd still encourage reprovisioning, right?

sdodson:
> In a bare metal environment I can imagine someone wanting it but I'd still encourage reprovisioning, right?

Yes, but we never actually win that battle. I've got a stack of 50 bugs related to uninstall that indicate no reprovisioning is a hurdle too high.

pkg/daemon/daemon.go
```diff
@@ -266,6 +277,34 @@ func (dn *Daemon) runOnce() error {
...
+		absoluteOnceFrom, err := filepath.Abs(filepath.Clean(dn.onceFrom))
+		if err != nil {
+			return err
+		}
+		ignConfig, err := dn.getIgnitionConfigFromFile(absoluteOnceFrom)
+		if err != nil {
+			return err
+		}
+		// Execute update without hitting the cluster
+		if err := dn.writeFiles(ignConfig.Storage.Files); err != nil {
```
kikisdeliveryservice: Usually after updating files/units, stale data is deleted\config is updated. Do we need to do that type of thing here? cc @ashcrow

ashcrow: We should. We can clean this up in a follow on.

ashcrow:
@eparis good questions. I followed up with @michaelgugino about that and they are wanting to use `MC` functionality for upgrades so there is a need for both `Ignition` and `MC` in a `once` use case.

In a follow on we'll rename them to follow a similar pattern *or* sense which one has been passed to us via file extension and use the correct code path.

jlebon:
> I followed up with @michaelgugino about that and they are wanting to use MC functionality for upgrades so there is a need for both Ignition and MC in a once use case.

I'm confused by this. I understand the need for `--from-ign`. But past that bootstrapping phase, shouldn't updates be delivered and applied the same way regardless of whether it's RHCOS or BYOR? (Modulo handing `osImageURL` of course).

aaronlevy:
That was the original plan. However, it was discussed in slack that allowing config updates that are delivered separately from binary updates could cause some pain. For example, if we're on a RHEL host and MCD (running as a daemon) updates kubelet configuration, but does not update the kubelet RPM - we could have config/binary mismatch that causes problems (these could be mitigated by enforcing from/to versions, etc, but still opens us up to a new set of problems).

Couple alternatives discussed:
- Never run MCD as a daemon on RHEL hosts. Instead it will be executed manually during a "host" upgrade that also manages other steps like updating RPMs (likely ansible).
- Run it as a daemon, but also teach it how to manage RPMs as part of RHEL upgrades. (FWIW - I would personally be pretty hesitant about going this route).

jlebon:
Cool, thanks for the insight.

> Never run MCD as a daemon on RHEL hosts. Instead it will be executed manually during a "host" upgrade that also manages other steps like updating RPMs (likely ansible).

So we're in control of upgrades even in the BYOR scenario? In that case, we could potentially teach the MCD to call out to the upgrader the same way it calls out to `pivot` on RHCOS, right? That'd mean much closer parity across the two cases.

> Run it as a daemon, but also teach it how to manage RPMs as part of RHEL upgrades. (FWIW - I would personally be pretty hesitant about going this route).

Yeah, not a big fan of that either.

PR #173, opened 2018-11-14
Sync pull secret from `kube-system/coreos-pull-secret` to disk

abhinavdahiya:
Updates the MCC to read pull-secret from `Secret` (used in cluster) or `Inline data` (used during bootstrapping).

This reads the `kube-system/coreos-pull-secret` Secret from the api.

### DIFF
--- a/manifests/machineconfigcontroller/controllerconfig.yaml
+++ b/manifests/machineconfigcontroller/controllerconfig.yaml
@@ -4,10 +4,4 @@ metadata:
   name: machine-config-controller
   namespace: {{.TargetNamespace}}
 spec:
-  clusterDNSIP: {{.ControllerConfig.ClusterDNSIP}}
-  cloudProviderConfig: {{.ControllerConfig.CloudProviderConfig}}
-  clusterName: {{.ControllerConfig.ClusterName}}
-  platform: {{.ControllerConfig.Platform}}
-  baseDomain: {{.ControllerConfig.BaseDomain}}
-  etcdCAData: {{.ControllerConfig.EtcdCAData | toString | b64enc}}
-  rootCAData: {{.ControllerConfig.RootCAData | toString | b64enc}}
+{{toYAML .ControllerConfig | toString | indent 2}}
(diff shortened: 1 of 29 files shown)


manifests/machineconfigcontroller/controllerconfig.yaml
```diff
@@ -4,10 +4,4 @@ metadata:
...
   namespace: {{.TargetNamespace}}
 spec:
-  clusterDNSIP: {{.ControllerConfig.ClusterDNSIP}}
-  cloudProviderConfig: {{.ControllerConfig.CloudProviderConfig}}
-  clusterName: {{.ControllerConfig.ClusterName}}
-  platform: {{.ControllerConfig.Platform}}
-  baseDomain: {{.ControllerConfig.BaseDomain}}
-  etcdCAData: {{.ControllerConfig.EtcdCAData | toString | b64enc}}
-  rootCAData: {{.ControllerConfig.RootCAData | toString | b64enc}}
+{{toYAML .ControllerConfig | toString | indent 2}}
```
wking: How is base64 encoding for the CAs handled now?

abhinavdahiya: they were `[]byte` and yaml marshalling automatically converts `[]byte` to `base64encoded string`

abhinavdahiya:
@crawford updated the PR to removing `data` from `pullsecret` field.

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -40,6 +45,16 @@ func (b *Bootstrap) Run(destDir string) error {
 		return err
 	}
 
+	psfraw, err := ioutil.ReadFile(b.pullSecretFile)
```
ashcrow: Will there always be a pull secret available, or will there be times pull secrets are defined and this will be an empty file?

abhinavdahiya: Its my plan to tighten to requirement when the source and location finalizes.

PR #177, opened 2018-11-15
Cloud provider conf

flaper87:
In order to start OpenShift with the OpenStack provider enabled, there ought to be a config file in the server. The Openshift deployment on OpenStack is currently failing because such file is not present.

This patch, combined with https://github.com/openshift/installer/pull/611 aims to create the config file in the destination node using the cloud provided by the operator during the dpeloyment.


ashcrow:
Looks like `"github.com/gophercloud/utils/openstack/clientconfig"` hasn't been vendored. The golang tool `dep` is used for this project.

```
vendor/github.com/openshift/installer/pkg/types/installconfig.go:6:2: cannot find package "github.com/gophercloud/utils/openstack/clientconfig" in any of:
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/gophercloud/utils/openstack/clientconfig (vendor tree)
	/usr/local/go/src/github.com/gophercloud/utils/openstack/clientconfig (from $GOROOT)
	/go/src/github.com/gophercloud/utils/openstack/clientconfig (from $GOPATH)
```

pkg/controller/template/render.go
```diff
@@ -336,6 +340,26 @@ func cloudProvider(cfg renderConfig) (interface{}, error) {
 	return "", nil
 }
 
+func cloudProviderConfig(cfg renderConfig) (interface{}, error) {
+    config, err := base64.StdEncoding.DecodeString(cfg.CloudProviderConfig)
+        if err != nil {
+                return "", err
+        }
+
+	switch cfg.Platform {
```
kikisdeliveryservice: Should there be a default in this switch statement?

abhinavdahiya:
@flaper87 the cloud provider config is usually a list of options.
I am open to MCO / MCC understanding a configmap source for getting the cloud provider config contents rather than the marshal/unmarshal in this PR.

sjenning:
i just hit this issue just this morning trying to do an install.  glad to see someone is already on it :+1:

flaper87:
Thakn y'all for the review!

@abhinavdahiya a config map should also work. Should the config map be created by the installer or by MCO directly? One problem we have on the openstack provider side is that this file has to exist in the filesystem (at least until I manage to remove this constraint).

Would it be ok for us to start with this approach and iterate towards using configmaps?

abhinavdahiya:
> Thakn y'all for the review!
> 
> @abhinavdahiya a config map should also work. Should the config map be created by the installer or by MCO directly? One problem we have on the openstack provider side is that this file has to exist in the filesystem (at least until I manage to remove this constraint).
> 
> Would it be ok for us to start with this approach and iterate towards using configmaps?

I am pretty sure all clouds will eventually have need for a cloud config file, like azure...
So I think config map approach is going to be the best way forward. I'll try to work something up today to see how much effort that will be.

flaper87:
@abhinavdahiya awesome, Thanks! 

So, lemme recap and see if I understood correctly. The installer will create a configmap that MCO will use and from which (at least in this case) we'll be able to create the required file on the node FS. correct?

flaper87:
@abhinavdahiya sounds good to me, I'll give it a go!

derekwaynecarr:
The config map should be editable by end users on bring your own infra environments as noted for azure, gce, and vsphere.  We do need this PR to update so kubelet is passed the flag once the final source is available.

sjenning:
need to rebase and run `go test ./pkg/controller/template/... -u` to update the test_data (why the ci/prow/unit job is failing)

sjenning:
If it can be made to pass e2e-aws, then it seems fine to me.  I'll see if I can test in on my Openstack cluster soon.

flaper87:
@sjenning done! Thanks for the feedback :)

templates/openstack/master/files/clouds.yaml
```diff
@@ -0,0 +1,16 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/openstack/clouds.conf"
+contents:
+  inline: |
+    {{with cloudProviderConfig .}}
+    [Global]
+    username={{.Cloud.AuthInfo.Username}}
```
derekwaynecarr: will this make the username and pw generally visible from a machineconfig?

if so, this seems like a problem since the ignition configs are not generally protected.

we need to determine a way to deliver secret data to nodes securely.

rphillips:
Looks like a number of these files need a gofmt.

kikisdeliveryservice:
how does https://github.com/openshift/machine-config-operator/pull/591 affect this PR?

PR #178, opened 2018-11-15
Clean up some network sysctls

squeed:
* revert the flannel bridge changes
* Enable `ip_forward` by default.


squeed:
cc @abhinavdahiya

PR #179, opened 2018-11-15
daemon: Unify onceFrom usage to one switch

ashcrow:
When once-from is supplied a file that ends in .ign it will follow
the Ignition only path. Otherwise it assumes the file is a
MachineConfig.

/cc @michaelgugino @sdodson @aaronlevy @eparis @kikisdeliveryservice


ashcrow:
/hold

I'm going to rework the sense to attempt parsing as @michaelgugino noted that when we add remote API support the endpoints may not provide a file like API.

pkg/daemon/daemon.go
```diff
@@ -196,10 +191,11 @@ func (dn *Daemon) Run(stop <-chan struct{}) error {
...
-		if dn.fromIgnition {
+		switch configType := SenseOnceFromConfigType(dn.onceFrom); configType {
+		case MachineConfigIgnitionFileType:
 			glog.V(2).Info("Daemon running directly from Ignition")
-			return dn.runFromIgnition()
-		} else {
+			return dn.runOnceFromIgnition()
+		default:
 			glog.V(2).Info("Daemon running once per request")
 			return dn.runOnce()
```
kikisdeliveryservice: I'd suggest renaming this function to runOnceFromMachineConfig (if that is indeed the case) just for naming clarity/consistency. Otherwise, I love this restructure!

ashcrow: Updated

kikisdeliveryservice: Yesss looks good!

ashcrow:
Found the culprit. Working on it now :smile:

ashcrow:
Removed code that's no longer used. Previous e2e tests passed. Local testing worked. Asking @kikisdeliveryservice to run local testing as well to double check.

ashcrow:
/hold

Found a bug

pkg/daemon/daemon.go
```diff
@@ -742,3 +666,58 @@ func ValidPath(path string) bool {
...
+
+		content, err = dn.fileSystemClient.ReadFile(absoluteOnceFrom)
+		if err != nil {
+			return nil, "", contentFrom, err
+		}
+	}
+
+	// Try each supported parser
+	// TODO: Add to resource/machineconfig.go as a function?
+	mcfgScheme := runtime.NewScheme()
```
michaelgugino: We can move this logic and below to the calling function to avoid having to return an interface, just return the file contents.  Probably would make testing easier as well.

pkg/daemon/daemon.go
```diff
@@ -742,3 +666,58 @@ func ValidPath(path string) bool {
...
+	mcfgCodecs := serializer.NewCodecFactory(mcfgScheme)
+	requiredObj, err := runtime.Decode(mcfgCodecs.UniversalDecoder(mcfgv1.SchemeGroupVersion), content)
+	if err == nil {
+		glog.V(2).Infof("onceFrom file is of type MachineConfig")
+		mcConfig := requiredObj.(*mcfgv1.MachineConfig)
+		return mcConfig, MachineConfigMCFileType, contentFrom, nil
+	}
+	glog.V(2).Infof("%s is not a MachineConfig: %s. Trying Ignition.", dn.onceFrom, err)
+
+	ignConfig, _, err := ignv2.Parse(content)
```
michaelgugino: Possibly this should be tried first to avoid the scaffolding required for the MC case?

ashcrow: I like that idea. I'll switch it up.

ashcrow:
Manually tested on RHCOS and CentOS with both MC and Ignition `onceFrom` functionality. Deferring to @kikisdeliveryservice to merge.

ashcrow:
Different failures this time. I **think** this is unrelated to adding a reboot to `Ignition`-only functionality as it's not tested in e2e and the previous build passed fine.

/test e2e-aws

ashcrow:
Stress test failure + others. I believe this is related to the large group of new tests that got added. Backing off from retesting for now.

kikisdeliveryservice:
Testing update: tested this with RHCOS using --onceFrom: files successfully written and OS pivoted from 47.110->47.114

PR #180, opened 2018-11-16
accept unencoded configmaps

deads2k:
configmap values are raw strings, not base64 encoded.
This makes the MCO bilingual until the installer pull merges

/assign @abhinavdahiya


pkg/operator/operator.go
```diff
@@ -270,7 +270,9 @@ func (optr *Operator) getCAsFromConfigMap(namespace, name, key string) ([]byte,
 	} else if d, dok := cm.Data[key]; dok {
 		raw, err := base64.StdEncoding.DecodeString(d)
 		if err != nil {
-			return nil, err
+                        // this is actually the result of a bad assumption.  configmap values are not encoded.
```
abhinavdahiya: `configmap values are not encoded`,  hmm nothing stops them from being so 😋. I think that's between the producer and consumer.

PR #182, opened 2018-11-19
Update instructions to disable CVO.

kikisdeliveryservice:
`oc delete daemonset...` returns an error, the correct instruction is now `oc delete deploy...`


kikisdeliveryservice:
@cgwalters I would think so, tho I'm not sure. But I'm ok with waiting until the job failures get resolved too. :)

kikisdeliveryservice:
cc: @abhinavdahiya

jlebon:
#194

ashcrow:
/test e2e-aws

Continued flakes.

PR #195, opened 2018-11-28
resourceread: use lib function for loading machine configs

darkmuggle:
Use lib/resourceread for loading machineconfigs for the daemon. Added
basic testing for load function.

### DIFF
--- a/lib/resourceread/machineconfig.go
+++ b/lib/resourceread/machineconfig.go
@@ -1,6 +1,9 @@
 package resourceread
 
 import (
+	"errors"
+	"fmt"
+
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -15,14 +15,13 @@ import (
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
+	resr "github.com/openshift/machine-config-operator/lib/resourceread"
```
ashcrow: Note: the rest of the code base uses `resourceread` without an alias, but I'm still OK with this.

lib/resourceread/machineconfig.go
```diff
@@ -17,13 +17,22 @@ func init() {
 	}
 }
 
-// ReadMachineConfigV1OrDie reads MachineConfig object from bytes. Panics on error.
+// ReadMachineConfigV1 reads raw MachineConfig object from bytes. Returns MachineConfig and error.
+func ReadMachineConfigV1(objBytes []byte) (*mcfgv1.MachineConfig, error) {
+	mc, err := runtime.Decode(mcfgCodecs.UniversalDecoder(mcfgv1.SchemeGroupVersion), objBytes)
+	if err == nil {
```
abhinavdahiya:
```go
if err !=nil {
    return nil, err
}
mc, ok := obj.(*mcfgv1.MachineConfig)
if !ok {
    return nil, <error describing this>
}
return mc, nil
```

pkg/daemon/daemon.go
darkmuggle: Fixed

lib/resourceread/machineconfig.go
darkmuggle: Thanks for the feedback...I've hardened the logic and repushed.

darkmuggle:
@abhinavdahiya the rationale for `ReadMachineConfigV1` was create a check function that could be used in other parts of the code. Before this commit, individual callers that wanted to parse a MachineConfig had to implement their own logic [1] . This change removes [1] in favor of `ReadMachineConfigV1`, Since `ReadMachineConfigV1OrDie` is used (when a panic is desirable), `ReadMachineConfigV1` handles the parsing of the config for the rest of the code.  

[1]  https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/daemon/daemon.go#L714-L722

lib/resourceread/machineconfig.go
```diff
@@ -17,13 +19,37 @@ func init() {
...
+	if err != nil {
+		return nil, err
+	}
+	if m == nil {
+		return nil, errDecode
+	}
+
+	mc, ok := m.(*mcfgv1.MachineConfig)
+	if !ok {
+		return nil, errDecode
```
abhinavdahiya: `runtime.Decode(mcfgCodecs.UniversalDecoder(mcfgv1.SchemeGroupVersion), objBytes)` will parse raw bytes of a type in https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/apis/machineconfiguration.openshift.io/v1/register.go#L40-L49

So this `m.(*mcfgv1.MachineConfig)` is that the raw byte was actually `MachineConfig` object. So the error should be `fmt.Errorf("expected *mcfvgv1.MachineConfig bute found %T", m)`

darkmuggle: Ah, indeed :+1: I've made the error more descriptive and updated the other others too.

ashcrow:
```
could not wait for build: the build machine-config-controller failed after 5m13s with reason DockerBuildFailed: Docker build strategy has failed.
```

Failure to build

/test e2e-aws

ashcrow:
Looks like build issues related to dangling imports.

```
pkg/daemon/daemon.go:25:2: imported and not used: "github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/runtime"
pkg/daemon/daemon.go:26:2: imported and not used: "github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/runtime/serializer"
pkg/daemon/daemon.go:25:2: imported and not used: "github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/runtime"
pkg/daemon/daemon.go:26:2: imported and not used: "github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/runtime/serializer"
```

ashcrow:
@darkmuggle PTAL at the dangling imports.

darkmuggle:
@ashcrow looks like merge conflicts were the root cause of the dangling imports. I've rebased from master, and pushed. Maybe it'll work now....

darkmuggle:
@abhinavdahiya I believe your requested changes have been addressed? If so, would you mind removing your objection?

abhinavdahiya:
> @abhinavdahiya I believe your requested changes have been addressed? If so, would you mind removing your objection?

Oh, i forgot :innocent: 

/approve

PR #196, opened 2018-11-29
Use "" for "unspecified OS", tweak comments

cgwalters:
The fact that `://dummy` is not a valid URL syntax was bothering me,
let's support (and default to) the empty string to mean unset.

Use a common function for handling this, and also change the comments
to clarify that we're not today "working around" the installer, it's
simply not specified.

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -203,10 +203,6 @@ func generateMachineConfigForRole(config *renderConfig, roleName string, roleDir
 const (
 	machineConfigNameTmpl     = "00-%s"
 	machineConfigRoleLabelKey = "machineconfiguration.openshift.io/role"
-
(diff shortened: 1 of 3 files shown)


jlebon:
/lgtm

> and also change the comments
to clarify that we're not today "working around" the installer, it's
simply not specified.

Aren't we, though? The goal is for the installer to eventually specify this in the initial config (see https://github.com/openshift/installer/issues/281#issuecomment-440418133). At which point, we shouldn't expect the `osImageURL` to ever be `""` or `://dummy` (I guess BYOR makes this less true, but it's definitely true for RHCOS).

cgwalters:
> At which point, we shouldn't expect the osImageURL to ever be "" or ://dummy

Probably but...I think we should still support not specifying it (for debugging at least).

PR #198, opened 2018-11-29
kubelet: set default system reservations

sjenning:
@smarterclayton @rphillips


cgwalters:
I like to see at least a single line rationale in commit messages personally.  Something like
"We are seeing the kubelet be OOM killed, let's ensure it has baseline capacity" or something?

smarterclayton:
Yeah, I'd phrase this as

> OpenShift always runs with system reserved capacity set to an appropriate per node value in order to protect system daemons from pod workloads.  Set the default to a reasonable value for a small node, leaving the door open to dynamic programming in the future.

PR #201, opened 2018-11-29
debug: Send significant events to the apiserver for a node

smarterclayton:
These three events represent rough actions taken by the MCD that an
administrator should know about.

### DIFF
--- a/lib/resourceapply/rbac.go
+++ b/lib/resourceapply/rbac.go
@@ -29,6 +29,27 @@ func ApplyClusterRoleBinding(client rbacclientv1.ClusterRoleBindingsGetter, requ
 	return actual, true, err
 }
 
+// ApplyRoleBinding applies the required rolebinding to the cluster.
+func ApplyRoleBinding(client rbacclientv1.RoleBindingsGetter, required *rbacv1.RoleBinding) (*rbacv1.RoleBinding, bool, error) {
+	existing, err := client.RoleBindings(required.Namespace).Get(required.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		actual, err := client.RoleBindings(required.Namespace).Create(required)
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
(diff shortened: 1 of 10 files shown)


smarterclayton:
@jlebon example as brought up in chat.  I need to double check that you have the right perms to send events though.

abhinavdahiya:
@smarterclayton 
Will need to add handling to MCO for the new clusterroles and clusterrolebindings [here](https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/operator/sync.go#L155-L173) similar to [here](https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/operator/sync.go#L211-L227)

_sorry for explicit :innocent:_

smarterclayton:
Ah, thanks.

On Nov 29, 2018, at 6:50 PM, Abhinav Dahiya <notifications@github.com>
wrote:

@smarterclayton <https://github.com/smarterclayton>
Will need to add handling to MCO for the new clusterroles and
clusterrolebindings here
<https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/operator/sync.go#L155-L173>
similar to here
<https://github.com/openshift/machine-config-operator/blob/f915a1bfb2d98a2cdde2ea264008f7dfb58bf262/pkg/operator/sync.go#L211-L227>

*sorry for explicit 😇*

—
<https://github.com/openshift/machine-config-operator/pull/201#issuecomment-443037494>,
or mute the thread

jlebon:
Ahh neat, wasn't aware of `oc get events`.

> I need to double check that you have the right perms to send events though.

Sounds good.

cgwalters:
Looks like this overlaps with https://github.com/openshift/machine-config-operator/pull/202

cgwalters:
Not an expert in this code but this PR looks good to me.  Seems like it's passing tests, just missing a slash-lgtm?

I'll rebase https://github.com/openshift/machine-config-operator/pull/202 on top of this one once this lands.

smarterclayton:
@abhinavdahiya tag this too?

PR #202, opened 2018-11-30
mcd: Log a rationale at each reboot

cgwalters:
This should be very prominent in the logs.  In the future we will
likely also e.g. log it to the systemd journal at least too.


ashcrow:
:+1:

jlebon:
@cgwalters Looks like this needs a rebase.

ashcrow:
Parent PR has merged.

PR #204, opened 2018-11-30
daemon: tweak log message

jlebon:
We had two "failed to initialize daemon" log messages in the
`runStartCmd()` function. One can still distinguish which one was
emitted since glog does print the line number too, but really that first
one wasn't accurate, so tweak it.


ashcrow:
Failed to set up test infra:

/test e2e-aws

Issue #203, opened 2018-11-30
MCC taking a while to create MachineConfigs


jlebon:
Split out from https://github.com/openshift/machine-config-operator/issues/199#issuecomment-443033182:

It took quite a while for the MCC to create the MachineConfigs:

```
$ kubectl get machineconfigs
No resources found.
```

A few minutes later:

```
$ kubectl get machineconfigs
NAME        AGE
00-master   1m
00-worker   1m
```

I'm sure `7f7366afead827ab5aa47ce67d5903c9` will show up if I wait a bit longer.

The MCC logs show the following:

```
I1129 23:24:23.120333       1 start.go:44] Version: 3.11.0-278-g628ad5d1-dirty
I1129 23:24:23.319527       1 leaderelection.go:185] attempting to acquire leader lease  openshift-machine-config-operator/machine-config-controller...
E1129 23:26:19.915022       1 event.go:259] Could not construct reference to: '&v1.ConfigMap{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"machine-config-controller", GenerateName:"", Namespace:"openshift-machine-config-operator", SelfLink:"/api/v1/namespaces/openshift-machine-config-operator/configmaps/machine-config-controller", UID:"913656b5-f42d-11e8-9834-12202f72768c", ResourceVersion:"7816", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63679130518, loc:(*time.Location)(0x1c60aa0)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string{"control-plane.alpha.kubernetes.io/leader":"{\"holderIdentity\":\"machine-config-controller-5f644d794f-l55l5_e7aa2bc2-f42d-11e8-a035-0a580a800025\",\"leaseDurationSeconds\":90,\"acquireTime\":\"2018-11-29T23:26:19Z\",\"renewTime\":\"2018-11-29T23:26:19Z\",\"leaderTransitions\":1}"}, OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:""}, Data:map[string]string(nil), BinaryData:map[string][]uint8(nil)}'
[...]

Issue #199, opened 2018-11-29
MCD rebooting early after set up


jlebon:
From @crawford:

```
I1129 21:37:29.566704    8105 start.go:51] Version: 3.11.0-278-g628ad5d1-dirty
I1129 21:37:29.567579    8105 start.go:78] starting node writer
I1129 21:37:29.572698    8105 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I1129 21:37:29.971743    8105 daemon.go:115] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:9c7f465ed8a84ed9d47356258ede44b7aec8b11b4e89defaf3ecee3cef33f0a0 (47.153)
I1129 21:37:30.009173    8105 start.go:127] Calling chroot("/rootfs")
E1129 21:37:30.029005    8105 daemon.go:303] Marking degraded due to: machineconfigs.machineconfiguration.openshift.io "0788cf371d56ac505e5b0e78392ccb14" not found
I1129 21:37:30.044992    8105 start.go:146] Starting MachineConfigDaemon
I1129 21:37:30.045012    8105 daemon.go:187] Enabling Kubelet Healthz Monitor
W1129 21:37:30.804003    8105 reflector.go:341] github.com/openshift/machine-config-operator/vendor/k8s.io/client-go/informers/factory.go:130: watch of *v1.Node ended with: very short watch: github.com/openshift/machine-config-operator/vendor/k8s.io/client-go/informers/factory.go:130: Unexpected watch close - watch lasted less than a second and no items received
E1129 21:37:31.805751    8105 reflector.go:205] github.com/openshift/machine-config-operator/vendor/k8s.io/client-go/informers/factory.go:130: Failed to list *v1.Node: Get https://172.30.0.1:443/api/v1/nodes?limit=500&resourceVersion=0: dial tcp 172.30.0.1:443: connect: connection refused
...
I1129 21:38:09.151806    8105 update.go:489] Rebooting
```

```
I1129 21:40:52.851825    5989 start.go:51] Version: 3.11.0-278-g628ad5d1-dirty
I1129 21:40:52.856779    5989 start.go:78] starting node writer
I1129 21:40:52.870226    5989 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
[...]

PR #206, opened 2018-11-30
Various error-handling fixes

jlebon:
See individual commit messages.

### DIFF
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"fmt"
 
+	"github.com/golang/glog"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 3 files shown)


jlebon:
Marked as WIP for now since this does not yet address (3) in https://github.com/openshift/machine-config-operator/issues/199#issuecomment-443207455.

ashcrow:
Changes look good. I'm not crazy about `daemon: retry for 5 minutes to get MachineConfig ` though. Is that more so related to https://github.com/openshift/machine-config-operator/issues/203?

ashcrow:
These commits are currently being manually tested by @kikisdeliveryservice and @jlebon

jlebon:
OK, dropping WIP on this, though I'd appreciate some manual testing from others. I pushed another commit which reworks how we use channels and error-handling in goroutines. This allows us to avoid the nuclear reboot-on-error pattern in favour of the mark-degraded-and-exit-on-error pattern.

pkg/daemon/writer.go
```diff
@@ -78,8 +79,10 @@ func (nw *NodeWriter) SetUpdateWorking(client corev1.NodeInterface, node string)
 	return <-respChan
 }
 
-// SetUpdateDegraded Sets the state to UpdateDegraded.
-func (nw *NodeWriter) SetUpdateDegraded(client corev1.NodeInterface, node string) error {
+// SetUpdateDegraded logs the error and sets the state to UpdateDegraded.
+// Returns an error if it couldn't set the annotation.
+func (nw *NodeWriter) SetUpdateDegraded(err error, client corev1.NodeInterface, node string) error {
+	glog.Error("Marking degraded due to: %v", err)
```
kikisdeliveryservice: Think this should be `glog.Errorf` because I'm seeing `Marking degraded due to: %vdaemon can't reconcile this config` :)

jlebon: Thanks, fixed!

PR #208, opened 2018-12-04
template: allow multiple MachineConfigs per role

sjenning:
@abhinavdahiya @rphillips @derekwaynecarr @ashcrow 

Proposed change to the `templates` structure to allow for multiple `MachineConfigs` per role.

Current:
```
templates/
 platform/
   role/
     files/
     units/
```

Proposed
```
templates/
 role/
   name/
     platform/
       files/
       units/
```

The name of the `MachineConfig` is now `00-<name>` instead of `00-<role>`.


pkg/controller/template/render.go
```diff
@@ -33,20 +33,19 @@ const (
...
 // All files from platform _base are always included, and may be overridden or
 // supplemented by platform-specific templates
 //
 //  ex:
-//       templates/_base/worker/units/kubelet.conf.tmpl
-//                           /files/hostname.tmpl
-//                    /master/units/kubelet.tmpl
-//                           /files/hostname.tmpl
-//       templates/aws/worker/units/kubelet-dropin.conf.tmpl
+//       templates/worker/worker/_base/units/kubelet.conf.tmpl
```
ashcrow: nit: It took me a second to realize the `worker/worker/` part was `<role>/<name/`. It may be worth adding one more example where they don't match. Not a big deal at all though.

sjenning: good idea. done.

pkg/controller/template/render.go
```diff
@@ -197,23 +199,23 @@ func generateMachineConfigForRole(config *renderConfig, roleName string, roleDir
...
+	return machineConfigFromIgnConfig(role, name, ignCfg), nil
 }
 
 const (
 	machineConfigNameTmpl     = "00-%s"
 	machineConfigRoleLabelKey = "machineconfiguration.openshift.io/role"
 )
 
-func machineConfigFromIgnConfig(role string, ignCfg *ignv2_2types.Config) *mcfgv1.MachineConfig {
-	name := fmt.Sprintf(machineConfigNameTmpl, role)
```
abhinavdahiya: This will cause all machineconfigs to have `00-name`. We want to control the default (first in sorted order) machineconfig because, the ordering is important when appending ignition configs.

with current setup this would be possible..
```console
/worker/
    /00-worker/
/master/
    /00-master/
```

abhinavdahiya:
@sjenning only if you have time can you add some docs [here](https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigController.md#templatecontroller) regarding the directory setup and how controller generates the configs....

I can also add an issue to track that, up to you. :innocent:

sjenning:
> I can also add an issue to track that, up to you.

Let's do that, so @rphillips can move along with the kubelet config controller

I'll make the numeric prefix change.

abhinavdahiya:
> > I can also add an issue to track that, up to you.
> 
> Let's do that, so @rphillips can move along with the kubelet config controller

done, https://github.com/openshift/machine-config-operator/issues/211 :innocent:

pkg/controller/template/render.go
sjenning: fixed this up

PR #215, opened 2018-12-05
daemon: Only log to recorder in cluster driven model

cgwalters:
I was looking at this code again and realized this is probably
an issue.


cgwalters:
(Is there an easy way to test the oncefrom stuff?  any docs?)

jlebon:
> (Is there an easy way to test the oncefrom stuff? any docs?)

Yeah, that'd be nice. I haven't tried it myself either yet. Offhand, it seems like one might be able to just directly execute the binary on the host and pass it an Ignition config through `--once-from config.ign`.

PR #216, opened 2018-12-05
daemon: Log reboot reason to systemd journal as well

cgwalters:
I saw my master reboot (pretty sure it was the MCD) after I killed
the MCD pod to test something.  There was no event (must not have
gotten flushed? - wonder if we should wait a little bit for the
event to propagate...or be more conservative about rebooting masters?)

Anyways let's log to the systemd journal too.  Yes, this implementation
is crude but it works.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -2,9 +2,11 @@ package daemon
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/user"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"strconv"
@@ -487,6 +489,32 @@ func (dn *Daemon) updateOS(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	return dn.NodeUpdaterClient.RunPivot(newConfig.Spec.OSImageURL)
 }
 
+// Log a message to the systemd journal as well as our stdout
+func (dn *Daemon) logSystem(format string, a ...interface{}) {
+	message := fmt.Sprintf(format, a...)
+	glog.Info(message)
+	// Since we're chrooted into the host rootfs with /run mounted,
+	// we can just talk to the journald socket.  Doing this as a
+	// subprocess rather than talking to journald in process since
+	// I worry about the golang library having a connection pre-chroot.
+	logger := exec.Command("logger")
+	stdin, err := logger.StdinPipe()
+	if err != nil {
+		glog.Errorf("Failed to get stdin pipe: %v", err)
+		return
+	}
+
+	go func() {
+		defer stdin.Close()
+		io.WriteString(stdin, message)
+	}()
+	err = logger.Run()
+	if err != nil {
+		glog.Errorf("Failed to invoke logger: %v", err)
+		return
+	}
+}
+
 // reboot is the final step. it tells systemd-logind to reboot the machine,
 // cleans up the agent's connections, and then sleeps for 7 days. if it wakes up
 // and manages to return, it returns a scary error message.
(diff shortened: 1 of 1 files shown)


cgwalters:
I am still testing this; am having issues where for some reason trying to deploy my testing MCD is causing reboots (from the vanilla one) and of course that is missing this rationale...

ashcrow:
Test flakes, but needs a rebase.

pkg/daemon/update.go
```diff
@@ -487,12 +489,38 @@ func (dn *Daemon) updateOS(oldConfig, newConfig *mcfgv1.MachineConfig) error {
...
+	// I worry about the golang library having a connection pre-chroot.
+	logger := exec.Command("logger")
+	stdin, err := logger.StdinPipe()
+	if err != nil {
+		glog.Errorf("Failed to get stdin pipe: %v", err)
+		return
+	}
+
+	go func() {
+		defer stdin.Close()
```
ashcrow: This isn't problematic, but why not do the `WriteString` followed by `Close` since `defer` executes on function end?

kikisdeliveryservice:
> I am still testing this; am having issues where for some reason trying to deploy my testing MCD is causing reboots (from the vanilla one) and of course that is missing this rationale...

What behaviour are you seeing exactly @cgwalters ?

cgwalters:
OK rebased :surfing_man: and tested!

```
[root@osiris-worker-0-z2v8p ~]# journalctl |grep machine-config-daemon.*reboot
Dec 06 17:35:28 osiris-worker-0-z2v8p root[25887]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 17:35:57 osiris-worker-0-z2v8p root[6234]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 17:36:25 osiris-worker-0-z2v8p root[6433]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 18:10:24 osiris-worker-0-z2v8p root[17944]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 18:17:02 osiris-worker-0-z2v8p root[9908]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 18:23:12 osiris-worker-0-z2v8p root[9823]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 19:16:40 osiris-worker-0-z2v8p root[23805]: machine-config-daemon initiating reboot: Node will reboot with new config.
Dec 06 19:38:50 osiris-worker-0-z2v8p root[14589]: machine-config-daemon initiating reboot: Node will reboot with new config.
```

(Which um...hm, that worker rebooted quite a bit, a lot more than I changed configs...)

pkg/daemon/update.go
cgwalters: Because I write Go sometimes by cargo culting [bits from the Internet](https://golang.org/pkg/os/exec/#Cmd.StdinPipe) :smile:

cgwalters: I think the general idea is that it's a best practice to always `defer..*close` since that way if e.g. if one later refactors a function to include an early return it still works?

cgwalters:
> What behaviour are you seeing exactly @cgwalters ?

Following the `HACKING.md` file and deploying a new MCD container was (I believe) causing reboots.

pkg/daemon/update.go
```diff
@@ -495,7 +523,7 @@ func (dn *Daemon) reboot(rationale string) error {
 	if (dn.recorder != nil) {
 		dn.recorder.Eventf(&corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: dn.name}}, corev1.EventTypeNormal, "Reboot", rationale)
 	}
-	glog.Infof("Rebooting: %s", rationale)
```
kikisdeliveryservice: Does this remove the "Rebooting" message from the normal Daemon log output?

cgwalters: Nope, we still call `glog.Info(message)` at the top of `logSystem()`

kikisdeliveryservice: thanks!

pkg/daemon/update.go
ashcrow: Works for me. :+1:

cgwalters:
OK since CI is being flaky again I went ahead and added in a commit I was going to do as a followup.

ashcrow:
Looks good @cgwalters. :crossed_fingers: for CI!

jlebon:
Yup, looks sane to me!
/lgtm

PR #218, opened 2018-12-06
controller: Prefix generated configs with pool type

cgwalters:
This way it's easy to see at a glance where a given
config came from.  Now looks like:

```
$ oc get machineconfig
NAME                                      AGE
00-master                                 6h
00-worker                                 6h
...
master-e92af7e0bdc0591dc11405aeabb57ccd   1m
worker-50748faa54f9cc70af13c8686c2c7a28   1m
```

### DIFF
--- a/pkg/controller/render/hash.go
+++ b/pkg/controller/render/hash.go
@@ -20,7 +20,9 @@ var (
 	}
 )
 
-func getMachineConfigHashedName(config *mcfgv1.MachineConfig) (string, error) {
+// Given a config from a pool, generate a name for the config
+// of the form <poolname>-<hash>
+func getMachineConfigHashedName(pool *mcfgv1.MachineConfigPool, config *mcfgv1.MachineConfig) (string, error) {
 	if config == nil {
 		return "", fmt.Errorf("empty machineconfig object")
 	}
@@ -34,7 +36,7 @@ func getMachineConfigHashedName(config *mcfgv1.MachineConfig) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%x", h), nil
+	return fmt.Sprintf("%s-%x", pool.GetName(), h), nil
 }
 
 func hashData(data []byte) ([]byte, error) {
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -470,7 +470,7 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 
 func generateMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig) (*mcfgv1.MachineConfig, error) {
 	merged := mcfgv1.MergeMachineConfigs(configs)
-	hashedName, err := getMachineConfigHashedName(merged)
+	hashedName, err := getMachineConfigHashedName(pool, merged)
 	if err != nil {
 		return nil, err
 	}


cgwalters:
Though I imagine there's some better code to do this type of thing somewhere; e.g. how deployments turn into pods prefixed with a random ID etc.  I'm a kube-code novice so pointers appreciated if someone has them.

(Also, related to that, is there a reason we use a checksum rather than just a short "unique-enough ID"?)

pkg/controller/render/hash.go
```diff
@@ -20,7 +20,7 @@ var (
 	}
 )
 
-func getMachineConfigHashedName(config *mcfgv1.MachineConfig) (string, error) {
+func getMachineConfigHashedName(prefix string, config *mcfgv1.MachineConfig) (string, error) {
```
abhinavdahiya: the name actually becomes `prefix-hash`, add a comment on function and what about empty prefix.

abhinavdahiya:
> Though I imagine there's some better code to do this type of thing somewhere; e.g. how deployments turn into pods prefixed with a random ID etc. I'm a kube-code novice so pointers appreciated if someone has them.
> 
> (Also, related to that, is there a reason we use a checksum rather than just a short "unique-enough ID"?)

using checksums allows MCC to manage generated machine configs without storing that hash somewhere in annotations like deployment does for pods.

cgwalters:
> using checksums allows MCC to manage generated machine configs without storing that hash somewhere in annotations like deployment does for pods.

I don't understand this; what does "manage" here mean?  As best I can tell nothing e.g. re-hashes a config again to compare it versus something else.

It's just about being unique right?

pkg/controller/render/hash.go
cgwalters: OK, I tweaked this to just take the pool.

abhinavdahiya:
> > using checksums allows MCC to manage generated machine configs without storing that hash somewhere in annotations like deployment does for pods.
> 
> I don't understand this; what does "manage" here mean? As best I can tell nothing e.g. re-hashes a config again to compare it versus something else.
> 
> It's just about being unique right?
No

https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/render/render_controller.go#L423-L434

cgwalters:
OK, understood now.  Thanks!

One thing I'd note here is this is a "schema change"; it will cause new configs to be generated in existing clusters.  I think that's OK because we haven't shipped yet, but now would definitely be the time to think of any other improvements to the naming scheme.

kikisdeliveryservice:
This would be really helpful when looking at  lists of machineconfigs, but I'll hold my approval and defer to @abhinavdahiya :)

ashcrow:
Seeing: `machine-config-server-mfplc is not healthy` in this last test. However, this is not consistent.

/test e2e-aws

ashcrow:
Masters didn't start/stay running.

/test e2e-aws

crawford:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/218/pull-ci-openshift-machine-config-operator-master-e2e-aws/679
> aws_instance.bootstrap: Error launching source instance: timeout while waiting for state to become 'success' (timeout: 30s)

Same failure over [here](https://github.com/openshift/installer/pull/868)... Right around the same time. I wonder if AWS briefly fell over.

/retest

crawford:
@ashcrow Hopefully, this will be fixed by https://github.com/openshift/installer/pull/882.

PR #219, opened 2018-12-07
MCO: update installconfig to remove admin

kikisdeliveryservice:
The InstallConfig no longer has an Admin field, so updating the MCO to account for this change.

Installer PR that originates this change: https://github.com/openshift/installer/pull/771
Shortcut to relevant file/changes in above PR: https://github.com/openshift/installer/pull/771/files#diff-74e3d89f2a13e465091f905f65cfc646


kikisdeliveryservice:
Note: I made this a separate PR to not muddy the waters, but this is needed for the SSHKey PRs.

kikisdeliveryservice:
cc: @abhinavdahiya

ashcrow:
This makes sense to me.

abhinavdahiya:
@kikisdeliveryservice why did the `Gopkg.lock / Gopkg.toml` not change with vendor change... ?

ashcrow:
She cherry picked a minimal set of changes to fix the issue. Though maybe we should rebase this vendor import fully.

cgwalters:
I'd vote for doing a full re-vendor.

But LGTM otherwise!

abhinavdahiya:
So any changes to vendor need to done by modifying Gopkg.toml and using dep ensure. No other methods are allowed. 

So no voting ;)

ashcrow:
Fair enough. Let's go ahead and use dep and update this vendored library. After that we can get one of the repo owners to double check and :+1:.

PR #221, opened 2018-12-07
daemon: Log degredation reason to journal and event too

cgwalters:
We really want admins to know *why* a system is degraded, so let's
add the precise error to the event log, and also log it to the
systemd journal just like we do reboots.

This changes the internal core `reconcible()` method to
return its error message rather than logging it itself, so we can
append it when we e.g. send the event.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -526,11 +526,11 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 	// config, the machine is definitely not in its desired state. this function
 	// will return true (meaning they are reconcilable) if there aren't actually
 	// changes.
-	reconcilable, err := dn.reconcilable(currentConfig, desiredConfig)
+	reconcilableError, err := dn.reconcilable(currentConfig, desiredConfig)
 	if err != nil {
 		return false, "", err
 	}
-	if !reconcilable {
+	if reconcilableError != nil {
 		return false, "", nil
 	}
 
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -5,8 +5,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/user"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"reflect"
 	"strconv"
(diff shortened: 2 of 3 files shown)


jlebon:
Looks good to me. Will give others some time to review as well.

ashcrow:
Error showed up during building for unittesting:

`pkg/daemon/update_test.go:75: Errorf call needs 1 arg but has 2 args`

cgwalters:
> Error showed up during building for unittesting:

How do you run the unit tests?

cgwalters:
OK, this passed e2e-aws and should be ready for review.

pkg/daemon/update.go
```diff
@@ -85,20 +88,19 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 }
 
 // reconcilable checks the configs to make sure that the only changes requested
-// are ones we know how to do in-place. if we can't do it in place, the node is
-// marked as degraded.
+// are ones we know how to do in-place.  If we can reconcile, (nil, nil) is returned.
+// Otherwise, if we can't do it in place, the node is  marked as degraded;
```
yuqi-zhang: There's an extra space here between "is" and "marked" and the line above

pkg/daemon/daemon.go
```diff
@@ -526,11 +526,11 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 	// config, the machine is definitely not in its desired state. this function
 	// will return true (meaning they are reconcilable) if there aren't actually
 	// changes.
-	reconcilable, err := dn.reconcilable(currentConfig, desiredConfig)
+	reconcilableError, err := dn.reconcilable(currentConfig, desiredConfig)
 	if err != nil {
 		return false, "", err
 	}
-	if !reconcilable {
+	if reconcilableError != nil {
```
yuqi-zhang: Since we record the event and log to journal down in `update`, we should perhaps do it here as well

cgwalters: Hum...I started to refactor all of this but...what we do if reconciliation fails there in `isDesiredMachineState()` is then later call down to the core `update()` function again, and we'll log there.  Right?

yuqi-zhang: Ah, you are correct, I forgot we removed the degradation for non-reconcilbale states, apologies for the noise

PR #226, opened 2018-12-11
operator: refactor bootstrapping to allow non-bootstrap manifests

abhinavdahiya:
This allows MCO bootstrap to render manifests in `/mainfests` that are pushed to cluster using bootstrap node.

MCO in bootstrap mode has 3 sets of assets.
1. `bootstrap/manifests` -> used by bootstrap MCC.
2. `bootstrap` - > MCC+MCS bootstrap pod.
3. `manifests` - > manifests installed during cluster bootstrap. eg the clusterrolebindings to allow master node boostrapping.

Previously MCO in bootstrap mode only supported 1 and 3. This PR allows all three.

The MCC+MCS bootstrap pod has a `v2` to support all the directory changes.
See: https://github.com/openshift/installer/pull/879 why this was needed in the first place.


cgwalters:
Can you elaborate slightly on what problem this is solving?

abhinavdahiya:
> Can you elaborate slightly on what problem this is solving?

updated the PR description.

PR #227, opened 2018-12-11
Disable cloud provider openstack

flaper87:
Let's not enable cloud provider openstack until there's a good solution
to upload `cloud.conf` files to the nodes. We're currently blocked by
this as the kubelet won't start until we provider proper configs.


hardys:
lgtm

flaper87:
> > We're currently blocked by this as the kubelet won't start until we provider proper configs.
> 
> Sorry, just trying to gain some context from this. So with this patch, the kubelet now starts up fine but isn't OpenStack aware?

Correct! Also, note that this is a temporary workaround until we are able to push `cloud.conf` files to the nodes. Without this patch, the openstack work is just blocked and we'll have to spend time working on a workaround to push the conf file ourselves :(

flaper87:
@hardys @ashcrow alright, now it should be good :smile:

Issue #230, opened 2018-12-11
create a script to both build + upload to cluster


cgwalters:
Moving this discussion from https://github.com/openshift/machine-config-operator/pull/189

As a developer I want to be able to conveniently hack on the stack.

cgwalters:
Some WIP on this:

`cluster-push-prep.sh`
```
#!/usr/bin/env bash

# Scale the CVO down and set up podman with a secret ready to push
# to the machine-config-operator namespace.

# Assumptions: You have set KUBECONFIG to point to your local cluster,
# and you have exposed the registry via e.g.
# https://github.com/openshift/installer/issues/411#issuecomment-445165262

set -xeuo pipefail

oc -n openshift-cluster-version scale --replicas=0 deploy/cluster-version-operator
if ! oc get -n openshift-image-registry route/image-registry &>/dev/null; then
    oc expose -n openshift-image-registry svc/image-registry
fi
oc patch -n openshift-image-registry route/image-registry -p '{"spec": {"tls": {"insecureEdgeTerminationPolicy": "Redirect", "termination": "reencrypt"}}}'
registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq -r ".spec.host")
curl -k --head https://"${registry}" >/dev/null
builder_secretid=$(oc get secret | egrep '^builder-token-'| head -1 | cut -f 1 -d ' ')
echo "podman login ..."
set +x
secret="$(oc get -o json secret/${builder_secretid} | jq -r '.data.token' | base64 -d)"
podman login --tls-verify=false -u unused -p "${secret}" "${registry}"
set -x
```

`cluster-push.sh`
```
#!/usr/bin/env bash

# Build an image and push it to the cluster registry.
# You must have first run `cluster-push-prep.sh` once.
# Assumptions: You have set KUBECONFIG to point to your local cluster,
# and you have exposed the registry via e.g.
# https://github.com/openshift/installer/issues/411#issuecomment-445165262

set -xeuo pipefail

registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq -r ".spec.host")
curl -k --head https://"${registry}" >/dev/null

WHAT=${WHAT:-machine-config-daemon}
LOCAL_IMGNAME=localhost/${WHAT}
REMOTE_IMGNAME=openshift-machine-config-operator/${WHAT}
#podman build -t "${LOCAL_IMGNAME}" -f Dockerfile.${WHAT}
[...]

PR #232, opened 2018-12-12
bootstrap: remove old boostrap assets

abhinavdahiya:
https://github.com/openshift/installer/pull/879 merged that switches to the new bootstrap manifests. So we can drop the old ones now.

/cc @ashcrow


ashcrow:
Failure is infra related:
```
level=error msg="\t* aws_instance.master.0: Error launching source instance: timeout while waiting for state to become 'success' (timeout: 30s)"
```

cgwalters:
My locally-built mcc (while hacking on https://github.com/openshift/machine-config-operator/issues/230 ) is falling over with a spam of:

```
E1212 22:08:13.004611       1 render.go:65] could not find expected template directory /etc/mcc/templates/_base                                                                                                   
I1212 22:08:13.206327       1 template_controller.go:302] Error syncing controllerconfig openshift-machine-config-operator/machine-config-controller: platform libvirt unsupported  
```

Should this fix it?

cgwalters:
OK, I think the problem is my cluster install *just* predates the original PR:

```
  machine-config-controller                     https://github.com/openshift/machine-config-operator                       bf4341308896665b121a101b0fdcd52e4cf68f8e
```

cgwalters:
And it doesn't look like there's a new release payload with it that was accepted.

PR #235, opened 2018-12-14
daemon: Log reason when isDesiredMachineState() is false

cgwalters:
While we do rather quickly drop down into actually trying
to perform an update, which will also e.g. log reconcilable
messages, let's be very clear about exactly which part
didn't match.

A machineconfig change could include one or both of (config, osImageURL)
and we want to know when each is different.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -531,30 +531,33 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 		return false, "", err
 	}
 	if reconcilableError != nil {
+		glog.Infof("Not in desired state: can't reconcile current=%s with desired=%s: %s",
+			currentConfig.GetName(), desiredConfig.GetName(), *reconcilableError)
 		return false, "", nil
 	}
 
-	isDesiredOS := false
 	// We only deal with operating system management on RHCOS
-	if dn.OperatingSystem != MachineConfigDaemonOSRHCOS {
-		// If we are on anything but RHCOS we set to True as there
-		// is nothing to updated.
-		isDesiredOS = true
-	} else {
-		isDesiredOS, err = dn.checkOS(desiredConfig.Spec.OSImageURL)
+	if dn.OperatingSystem == MachineConfigDaemonOSRHCOS {
+		isDesiredOS, err := dn.checkOS(desiredConfig.Spec.OSImageURL)
 		if err != nil {
 			return false, "", err
 		}
+		if !isDesiredOS {
+			glog.Infof("Not in desired state; booted osImageURL differs from target")
+			return false, "", nil
+		}
 	}
 
-	if dn.checkFiles(desiredConfig.Spec.Config.Storage.Files) &&
-		dn.checkUnits(desiredConfig.Spec.Config.Systemd.Units) &&
-		isDesiredOS {
-		return true, dcAnnotation, nil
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
MCD already logs the targetted OS image here: https://github.com/openshift/machine-config-operator/blob/a00ab593a120448a283ddb0343244d02ada112eb/pkg/daemon/update.go#L485

kikisdeliveryservice:
@cgwalters how are you trying to use MCD to update OS image? what does your machine config file look like?

kikisdeliveryservice:
@cgwalters from what I know the MCC orders machine configs lexigraphically and only picks the osImageURL from the first machine config, so try naming it something that puts it into the list before the original machine config. For ex for my setup if my first worker config is 00-worker I might use: 00-osimageupdate so that that config is read first.

YMMV since I'm not sure what the original machine config's name is, but this helped me recently.

cgwalters:
Hmm.  I already destroyed my cluster that had the reboot-looping worker...those logs would have been useful.  Trying to recreate the situation now.

(The reason I destroyed is I only had one worker, and clearly my MC was broken in some way, and that meant I no longer had a registry which broke my development flow)

kikisdeliveryservice:
To be clear I'm not against adding logging there I just think it should be more specifically phrased to not be redundant with the logging from update.go 

Something like "osImageURL %s does not match current booted osImage..." or something

jlebon:
Yeah, agreed the msg should be about the check finding a match or not. This is similar to what's in `checkFileContentsAndMode` (which... we should change those `Errorf` to `Infof` or something), where we print e.g. "expected" vs "received".

cgwalters:
OK, this PR is now different.  I'd actually like to change this whole stack to return a string just like `reconcilable()` but let's be incremental.

cgwalters:
Also...I don't understand why we try to do a full reconcile here, can't we just do:

```
index e47449c..9bf5d36 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -522,14 +522,13 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 		return false, "", err
 	}
 
-	// if we can't reconcile the changes between the old config and the new
-	// config, the machine is definitely not in its desired state. this function
-	// will return true (meaning they are reconcilable) if there aren't actually
-	// changes.
+	if currentConfig != desiredConfig {
+		glog.Infof("Not in desired state: config current=%s desired=%s",
+			currentConfig.GetName(), desiredConfig.GetName())
+	}
+
 	reconcilableError := dn.reconcilable(currentConfig, desiredConfig)
 	if reconcilableError != nil {
-		glog.Infof("Not in desired state: can't reconcile current=%s with desired=%s: %s",
-			currentConfig.GetName(), desiredConfig.GetName(), *reconcilableError)
 		return false, "", nil
 	}
 
```

?

cgwalters:
> Also...I don't understand why we try to do a full reconcile here, can't we just do:

The answer to that is that machineconfigs don't compare pointer-equal of course :smile:  But testing a patch that compares via deepequal.

jlebon:
This patch looks good to me! Will let @kikisdeliveryservice do the honours. :)

cgwalters:
I decided to rewrite this.  New PR incoming.

cgwalters:
Successor PR https://github.com/openshift/machine-config-operator/pull/245

PR #236, opened 2018-12-14
Makefile: Also support specifying just e.g. `make controller`

cgwalters:
Since we're in the machine-config-operator repository, typing
`machine-config-` is redundant.  Now one can just type e.g.
`make daemon` or `make controller`.  But we still support
`make machine-config-daemon` for backwards compat.


ashcrow:
Makes sense to me. @darkmuggle WDYT?

darkmuggle:
This looks fine by me, :+1:

Issue #71, opened 2018-09-17
Update k8s drain lib so we can use stable daemonset API group


jlebon:
See https://github.com/openshift/machine-config-operator/pull/70#discussion_r218148579.

wking:
Should this be an issue in openshift/kubernetes-drain?  And should we get an `OWNERS` there with some of you on it?

I'd [still][1] like to get that library upstreamed into client-go or some such; anything that would help move in that direction is welcome :).

[1]: https://github.com/kubernetes/kubernetes/issues/25625#issuecomment-405056988

runcom:
drain now has its own library in kube upstream, let's move forward to that (https://github.com/kubernetes/kubernetes/pull/72827)

PR #237, opened 2018-12-14
daemon: Reboot via systemd-run, not logind DBus API

cgwalters:
Today, the operator uses:

```
drain.Drain(..., IgnoreDaemonsets:   true, ...)
```

The problem is this leaves e.g. the mcd lying around
from Kube's perspective; it takes a while for it to conclude the mcd
is dead and make a new one on next reboot.  We really want to drain
the node fully for other daemonsets too for similar reasons.

To fix this, we do a two-phase drain; first we drain all of the
non-daemonsets.  Then, we *queue* a reboot by creating a new systemd
unit that waits for us to unlock a file in `/run`.  Once our process
dies (because our pod is drained), we'll unlock the file and the
reboot will commence.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -10,12 +10,14 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
+	"syscall"
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -61,10 +63,12 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	// TODO: Change the logic to be clearer
 	// We need to skip draining of the node when we are running once
 	// and there is no cluster.
-	if dn.onceFrom != "" && !ValidPath(dn.onceFrom) {
+	onceFromMode := dn.onceFrom == "" || ValidPath(dn.onceFrom)
+	var node *corev1.Node
+	if !onceFromMode {
 		glog.Info("Update prepared; draining the node")
 
-		node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
+		node, err = dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
@@ -78,13 +82,36 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 			IgnoreDaemonsets:   true,
 		})
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "Draining non-daemonsets")
 		}
-		glog.V(2).Infof("Node successfully drained")
+		glog.V(2).Infof("Node successfully drained (non-daemonsets)")
 	}
 
-	// reboot. this function shouldn't actually return.
-	return dn.reboot(fmt.Sprintf("Node will reboot into config %v", newConfigName))
+	err = dn.queueReboot(fmt.Sprintf("Node will reboot into config %v", newConfigName))
+	if err != nil {
+		return err
+	}
+
+	// And now this time without IgnoreDaemonSets, which should as a side
+	// effect kill us and release the reboot lock, allowing the queued
+	// reboot to proceed.
+	if !onceFromMode {
+		err = drain.Drain(dn.kubeClient, []*corev1.Node{node}, &drain.DrainOptions{
+			DeleteLocalData:    true,
+			Force:              true,
+			GracePeriodSeconds: 600,
+		})
+		if err != nil {
+			return errors.Wrapf(err, "Final drain")
+		}
+
+		// cross fingers
+		time.Sleep(24 * 7 * time.Hour)
+	}
+
+	// Otherwise, we fall through, which will in onceFrom mode also
+	// exit the process, releasing the file lock.
+	return nil
 }
 
 // reconcilable checks the configs to make sure that the only changes requested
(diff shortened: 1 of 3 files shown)


cgwalters:
Still testing this but I think it'll help with debugging across reboots.

/hold

pkg/daemon/update.go
```diff
@@ -78,13 +82,36 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
...
-			return err
+			return errors.Wrapf(err, "Draining non-daemonsets")
 		}
-		glog.V(2).Infof("Node successfully drained")
+		glog.V(2).Infof("Node successfully drained (non-daemonsets)")
 	}
 
-	// reboot. this function shouldn't actually return.
-	return dn.reboot(fmt.Sprintf("Node will reboot into config %v", newConfigName))
+	err = dn.queueReboot(fmt.Sprintf("Node will reboot into config %v", newConfigName))
```
ashcrow: Love it!

pkg/daemon/update.go
```diff
@@ -512,25 +539,35 @@ func (dn *Daemon) logSystem(format string, a ...interface{}) {
...
-// cleans up the agent's connections, and then sleeps for 7 days. if it wakes up
-// and manages to return, it returns a scary error message.
-func (dn *Daemon) reboot(rationale string) error {
+const rebootLockPath = "/run/machine-config-daemon-reboot"
+
+// The main idea here is that we *queue* a reboot in a separate systemd unit,
+// which will start once this process exits.  In the orchestrated (not-onceFrom)
+// case, the drain should kill our pod, releasing our lock.  In the onceFrom
+// case we just exit.
+func (dn *Daemon) queueReboot(rationale string) error {
```
ashcrow: Love it!

pkg/daemon/update.go
```diff
@@ -78,13 +82,36 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
...
+	err = dn.queueReboot(fmt.Sprintf("Node will reboot into config %v", newConfigName))
+	if err != nil {
+		return err
+	}
+
+	// And now this time without IgnoreDaemonSets, which should as a side
+	// effect kill us and release the reboot lock, allowing the queued
+	// reboot to proceed.
+	if !onceFromMode {
+		err = drain.Drain(dn.kubeClient, []*corev1.Node{node}, &drain.DrainOptions{
```
jlebon: Hmm, but how will Kubernetes know that it's OK to re-schedule us on that node if it doesn't ignore DaemonSets? We'd need some other service to uncordon the node once it's back up, right?

jlebon: (To clarify, the MCD currently is the one that uncordons the node once it's back up and completes the update).

cgwalters: Ah, right...yeah then to complete this story we'd need to push a systemd unit to the host that does that.  Which gets into interesting issues since we'd need to stash the kube creds.

We could move the uncordoning into the operator?  Actually, we could also move the drain there.  Have the MCD signal that it's queued a reboot successfully via node annotation.

Really it feels like there's some duplication in role here between kubelet and MCD.  If kubelet communicates boot ID to master, then we can update node annotations from that.

abhinavdahiya:
> The problem is this leaves e.g. the mcd lying around
from Kube's perspective; it takes a while for it to conclude the mcd
is dead and make a new one on next reboot. We really want to drain
the node fully for other daemonsets too for similar reasons

examples that back this claim will be very helpful to decide how big a problem is this delay.

cgwalters:
> examples that back this claim will be very helpful to decide how big a problem is this delay.

The main motivation for this work was the fact that as I'm hacking on this code, I commonly want to e.g. `oc logs -f pods/machine-config-daemon-xxx` and when a node is rebooting, that would time out.  It also is IMO just broken to try to preserve the pod across reboot, because we rebooted - it's a new pod!  Which should have its own logs.

cgwalters:
Also related https://github.com/openshift/machine-config-operator/issues/71 and chain from https://github.com/kubernetes/kubernetes/issues/25625#issuecomment-405056988

jlebon:
> We could move the uncordoning into the operator? Actually, we could also move the drain there. Have the MCD signal that it's queued a reboot successfully via node annotation.

I think it'd belong better in the MCC? Sync'ing between the MCC and MCD using annotations to coordinate reboots sounds a lot like the architecture from CLUO. Might be worth asking @crawford whether there was a conscious decision to move away from that.

> But clearly for the use case where we're going to reboot, those pods  are going to go away anyways...so it would make sense to teach the  daemonset controller about some annotation of that form?

Hmm, I think we could do this today using a `nodeSelector` and have the MCD set it after it queues the reboot? (And then have the MCC unset it once the node is back online).

Anyway, probably worth splitting this out into its own issue and discussing there?

pkg/daemon/update.go
```diff
@@ -512,25 +539,35 @@ func (dn *Daemon) logSystem(format string, a ...interface{}) {
...
-	// cross fingers
-	time.Sleep(24 * 7 * time.Hour)
+	if err = syscall.Flock(int(fh.Fd()), syscall.LOCK_EX); err != nil {
+		return errors.Wrapf(err, "flock(%s)", rebootLockPath)
+	}
 
-	// if everything went well, this should be unreachable.
-	return fmt.Errorf("Reboot failed; this error should be unreachable, something is seriously wrong")
+	// We invoke via systemd-run so that this pod can be safely drained
+	rebootCmd := exec.Command("systemd-run", "--unit", "machine-config-daemon-reboot",
```
crawford: Why not use the DBUS API to create this job?

pkg/daemon/update.go
```diff
@@ -512,25 +539,35 @@ func (dn *Daemon) logSystem(format string, a ...interface{}) {
...
+		"--description", fmt.Sprintf("machine-config-daemon: %s", rationale),
+		"flock", rebootLockPath, "reboot")
```
crawford: A reboot inhibitor could also work here and you wouldn't have to manage that lock file. It would also be a bit more discoverable to someone debugging. This approach seems fine though.

cgwalters:
Just a note to myself for this PR in the future; it may make sense to pursue a change to upstream Kube (or just OpenShift) to teach the DS controller not to schedule via some "really don't do it" label, another option we have is since we're in control of the kubelet, we can do a "daemonset drain" locally too in a way that makes it refuse to accept new pods.

cgwalters:
I was thinking about this PR a bit more recently; maybe the simplest thing after we drain the node is to `systemd-run '/bin/sh -c "systemctl stop kubelet && systemctl reboot"'` or so.  Need to verify that stopping the kubelet does the right thing, but it should.

sjenning:
What is "the right thing"?

cgwalters:
About half of this ended up landing; not worth keeping this open, we can revisit the issue later.

PR #238, opened 2018-12-15
MCD: Remove unused error return value from reconcilable()

kikisdeliveryservice:
The err return value in dn.reconcilable is never used.  Given that this function just does comparisons of config values, removing it to clean up function.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -526,10 +526,7 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 	// config, the machine is definitely not in its desired state. this function
 	// will return true (meaning they are reconcilable) if there aren't actually
 	// changes.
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -526,7 +526,7 @@ func (dn *Daemon) isDesiredMachineState() (bool, string, error) {
 	// config, the machine is definitely not in its desired state. this function
 	// will return true (meaning they are reconcilable) if there aren't actually
 	// changes.
-	reconcilableError, err := dn.reconcilable(currentConfig, desiredConfig)
+	reconcilableError := dn.reconcilable(currentConfig, desiredConfig)
 	if err != nil {
```
yuqi-zhang: Probably remove this error check as well

kikisdeliveryservice: thanks! will do!

PR #239, opened 2018-12-16
pkg/server/api: Add /healthz for load-balancer health checks

wking:
The server currently 404s the root path.  From a master:

```console
[core@ip-10-0-26-134 ~]$ curl -ik https://wking-api.devcluster.openshift.com:49500/
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Sun, 16 Dec 2018 06:30:14 GMT
Content-Length: 19

404 page not found
```

but we need a reliable response on the range 200-399 to satisfy our network load balancer health checks, which do not support configurable response status codes (see [here][1] and [here][2], which are Terraform links, but they discuss an AWS restriction that is not Terraform-specific).  This pull-request adds a `/healthz` endpoint which always 204s (when the server is alive to handle it).

Required for openshift/installer#924.

[1]: https://github.com/terraform-providers/terraform-provider-aws/issues/2708#issuecomment-352830642
[2]: https://github.com/terraform-providers/terraform-provider-aws/pull/2906/files#diff-375aea487c27a6ada86edfd817ba2401R612

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -22,6 +22,7 @@ type getConfig func(request poolRequest) (*ignv2_2types.Config, error)
 
 func newHandler(getConfig getConfig) http.Handler {
 	mux := http.NewServeMux()
+	mux.Handle("/healthz", &healthHandler{})
 	mux.Handle(apiPathConfig, &configHandler{getConfig: getConfig})
 	mux.Handle("/", &defaultHandler{})
 	return mux
@@ -83,6 +84,26 @@ func (h *configHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		glog.Errorf("failed to write %v response: %v", cr, err)
 	}
+}
+
+type healthHandler struct{}
+
+// ServeHTTP handles /healthz requests.
+func (h *healthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Length", "0")
+
+	if r.URL.Path == "/healthz" {
+		if r.Method == http.MethodGet || r.Method == http.MethodHead {
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	w.WriteHeader(http.StatusNotFound)
+	return
 }
 
 // defaultHandler is the HTTP Handler for backstopping invalid requests.
--- a/pkg/server/api_test.go
+++ b/pkg/server/api_test.go
@@ -80,6 +80,78 @@ func TestHandler(t *testing.T) {
 			},
 		},
 		{
+			name:    "get /healthz",
+			request: httptest.NewRequest(http.MethodGet, "http://testrequest/healthz", nil),
+			getConfig: func(poolRequest) (*ignv2_2types.Config, error) {
+				return new(ignv2_2types.Config), nil
+			},
+			checkResponse: func(t *testing.T, response *http.Response) {
+				checkStatus(t, response, http.StatusNoContent)
+				checkContentLength(t, response, 0)
+				checkBodyLength(t, response, 0)
+			},
+		},
+		{
+			name:    "head /healthz",
+			request: httptest.NewRequest(http.MethodHead, "http://testrequest/healthz", nil),
+			getConfig: func(poolRequest) (*ignv2_2types.Config, error) {
+				return new(ignv2_2types.Config), nil
+			},
+			checkResponse: func(t *testing.T, response *http.Response) {
+				checkStatus(t, response, http.StatusNoContent)
+				checkContentLength(t, response, 0)
+				checkBodyLength(t, response, 0)
+			},
+		},
+		{
+			name:    "get health path that does not exist",
+			request: httptest.NewRequest(http.MethodGet, "http://testrequest/healthz-does-not-exist", nil),
+			getConfig: func(poolRequest) (*ignv2_2types.Config, error) {
+				return nil, nil
+			},
+			checkResponse: func(t *testing.T, response *http.Response) {
+				checkStatus(t, response, http.StatusNotFound)
+				checkContentLength(t, response, 0)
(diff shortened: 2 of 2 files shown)


wking:
/hold

Using a muxer for this server seems overly complicated.  Lets table this change until I have time to drop the muxer, unless other folks can think of a reason to keep it.

abhinavdahiya:
> /hold
> 
> Using a muxer for this server seems overly complicated. Lets table this change until I have time to drop the muxer, unless other folks can think of a reason to keep it.

Do you mind expanding on why to think it's overcomplicated.

wking:
> Do you mind expanding on why to think it's overcomplicated.

Each sub-handler will be stwitching on `Path` anyway, and we don't have so many paths for the muxer's extra level of path switching to help much.  And it's nice to only have to put things like #240's method guard in one spot (vs. repeating in each of the muxer's sub-handlers).

wking:
Rebased onto #242 with 47439f8 -> 9fc7b2b.  This will collide with #240, but rebasing either will be easy, so I don't think it matters which lands first.

cgwalters:
LGTM
/assign abhinavdahiya

wking:
With #242 closed, I've gone back to a separate sub-handler for `/healthz` and rebased onto #240 to get unit tests with 9fc7b2b -> 628e654.

pkg/server/api.go
```diff
@@ -111,3 +112,21 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+type healthHandler struct{}
+
+// ServeHTTP handles /healthz requests.
+func (h *healthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	if r.URL.Path == "" {
```
abhinavdahiya: what is this check capturing.. ?

abhinavdahiya: also the path is going to be set to  `/healthz` based on setup here https://github.com/openshift/machine-config-operator/pull/239/commits/628e654b1e9a42ce30b11e618146dd051cc3bdad#diff-4bbf1ce644e1b41f1ed61ab858def528R46

wking: Sorry, I'm trying to wrap my head around how muxers work :p.  It looks like they pass the full URL through, and aren't stripping off the portion handled by the muxer's switch.  Anyhow, don't bother looking at this PR until #240 has landed ;).

abhinavdahiya: i did some local testing and this sheds some light on Path reaginrding how the muxer works.
https://github.com/openshift/machine-config-operator/pull/242#issuecomment-447943204

wking:
Ok, I've rebased this onto the new #246 with 628e654 -> 6964b05.  #246 gets `api_test.go` running over the muxer instead of just the API/config handler.

abhinavdahiya:
The PR includes commit https://github.com/openshift/machine-config-operator/commit/6cfa51cc380d4370b98fcbb4d9ef871dc682c0dc

/hold

crawford:
@abhinavdahiya @wking What still needs to happen with this PR?

abhinavdahiya:
> @abhinavdahiya @wking What still needs to happen with this PR?

this is what i would have expected it to look like without #246

wking:
> [this][1] is what i would have expected it to look like without #246 

That looks more complicated to me:

```console
$ git show --oneline --stat 50f1439777
50f1439 pkg/server/api: Add /healthz for load-balancer health checks
 pkg/server/api.go      | 21 +++++++++++++
 pkg/server/api_test.go | 84 ++++++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 105 insertions(+)
$ git show --oneline --stat d0a7ae21b0
d0a7ae2 add healthz
 pkg/server/api.go      |  47 ++++++++--
 pkg/server/api_test.go | 243 ++++++++++++++++++++++++++++++++++++++++++++++---
 2 files changed, 269 insertions(+), 21 deletions(-)
```

But, *shrug*, you've put more work into this repo than me.  Want to file a PR?

[1]: https://github.com/abhinavdahiya/machine-config-operator/commit/d0a7ae21b0e4d3174fd5ec41288204d4b0292019

abhinavdahiya:
> > [this](https://github.com/abhinavdahiya/machine-config-operator/commit/d0a7ae21b0e4d3174fd5ec41288204d4b0292019) is what i would have expected it to look like without #246
> 
> That looks more complicated to me:

Hmm, based on what the lines changed? 

> ```
> $ git show --oneline --stat 50f1439777
> 50f1439 pkg/server/api: Add /healthz for load-balancer health checks
>  pkg/server/api.go      | 21 +++++++++++++
>  pkg/server/api_test.go | 84 ++++++++++++++++++++++++++++++++++++++++++++++++++
>  2 files changed, 105 insertions(+)
> $ git show --oneline --stat d0a7ae21b0
> d0a7ae2 add healthz
>  pkg/server/api.go      |  47 ++++++++--

this is because 50f1439777 carries over `defaultHandler` from https://github.com/openshift/machine-config-operator/commit/6cfa51cc380d4370b98fcbb4d9ef871dc682c0dc

>  pkg/server/api_test.go | 243 ++++++++++++++++++++++++++++++++++++++++++++++---

this is because this has 2 types of tests
- test for each handler api, default, healthz
- test for server ie the muxer that @wking asked here https://github.com/openshift/machine-config-operator/pull/239#issuecomment-451598773

>  2 files changed, 269 insertions(+), 21 deletions(-)
> ```
> 
> But, _shrug_, you've put more work into this repo than me. Want to file a PR?

https://github.com/openshift/machine-config-operator/pull/267

wking:
> Hmm, based on what the lines changed?

Yeah, and also the package API.

> this is because 50f1439 carries over defaultHandler from 6cfa51c

But 6cfa51c as a whole is a net reduction of 40 lines.

> this is because this has 2 types of tests

With comprehensive muxer tests, I don't think we need sub-handler tests.

PR #240, opened 2018-12-16
pkg/server/api: Return 405s for non-GET requests

wking:
We don't want folks `POST`ing, etc. to this server.  With this pull-request, we'll return the appropriate response status code if they do.

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -80,8 +80,14 @@ func NewServerAPIHandler(s Server) *APIHandler {
 // ServeHTTP handles the requests for the machine config server
 // API handler.
 func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet && r.Method != http.MethodHead {
+		w.Header().Set("Content-Length", "0")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
 
 	if r.URL.Path == "" {
+		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
@@ -92,18 +98,34 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	conf, err := sh.server.GetConfig(cr)
 	if err != nil {
+		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusInternalServerError)
 		glog.Errorf("couldn't get config for req: %v, error: %v", cr, err)
 		return
 	}
 	if conf == nil && err == nil {
+		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
 
-	encoder := json.NewEncoder(w)
-	if err := encoder.Encode(conf); err != nil {
+	data, err := json.Marshal(conf)
+	if err != nil {
+		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusInternalServerError)
-		glog.Errorf("couldn't encode the config for req: %v, error: %v", cr, err)
+		glog.Errorf("failed to marshal %v config: %v", cr, err)
+		return
+	}
+
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", len(data)))
+	w.Header().Set("Content-Type", "application/json")
+	if r.Method == http.MethodHead {
+		w.WriteHeader(http.StatusOK)
+		return
+	}
+
+	_, err = w.Write(data)
+	if err != nil {
+		glog.Errorf("failed to write %v response: %v", cr, err)
 	}
 }
--- a/pkg/server/api_test.go
+++ b/pkg/server/api_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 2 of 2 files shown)


wking:
I'd have to extent [`scenario`][1].  Narrowly to add a `method` property?  Broadly to add a `request` property (which would also allow testing of #239)?

[1]: https://github.com/openshift/machine-config-operator/blob/7b029f803bd511ba8747494283156f475f7366e6/pkg/server/api_test.go#L20

wking:
Rebased onto #242 and added a test.

wking:
With #242 closed, I've rebased this onto master again and added a unit test based on `Request`s with 98d3fa2 -> 2dc59a3.

pkg/server/api.go
```diff
@@ -80,6 +80,10 @@ func NewServerAPIHandler(s Server) *APIHandler {
 // ServeHTTP handles the requests for the machine config server
 // API handler.
 func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
```
abhinavdahiya: @wking what about `HEAD`. Somebody might send head for headers before sending `GET` ?

but i'm punting on that before to follow up later.

wking:
[e2e-aws][1]:

```
level=error msg="1 error occurred:"
level=error msg="\t* module.vpc.aws_vpc_endpoint.s3: 1 error occurred:"
level=error msg="\t* aws_vpc_endpoint.s3: Error creating VPC Endpoint: VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached."
level=error msg="\tstatus code: 400, request id: e86bbbe2-9fc6-4177-ba9c-ad2a98f8030b"
```

James is working on getting that limit bumped, but until then it's just luck ;).  I'll wait until the queue calms down tonight before retesting, since this is a fairly minor change.

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/240/pull-ci-openshift-machine-config-operator-master-e2e-aws/721/build-log.txt

wking:
Our limit has been bumped.

/retest

pkg/server/api.go
wking: > @wking what about `HEAD`.

I've added support for `HEAD` and touched up our header handling with 2dc59a3 -> 22d0c45.

pkg/server/api.go
```diff
@@ -80,8 +80,14 @@ func NewServerAPIHandler(s Server) *APIHandler {
 // ServeHTTP handles the requests for the machine config server
 // API handler.
 func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet && r.Method != http.MethodHead {
+		w.Header().Set("Content-Length", "0")
```
abhinavdahiya: Can you point to to a reason why we need to set this?

wking: > Can you point to to a reason why we need to set this?

If people start `HEAD`ing this server, it seems like they might be interested in getting valid `Content-Length` back.  I just started setting it everywhere to make it easier to test that we get it right when we actually care about it.

wking:
[images failure][1] was openshift/release#2070.

/retest

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/240/pull-ci-openshift-machine-config-operator-master-images/444/build-log.txt

pkg/server/api_test.go
```diff
@@ -18,49 +19,90 @@ func (ms *mockServer) GetConfig(pr poolRequest) (*ignv2_2types.Config, error) {
...
-	name           string
+	request        *http.Request
 	expectedStatus int
 	serverFunc     func(poolRequest) (*ignv2_2types.Config, error)
 }
 
 func TestAPIHandler(t *testing.T) {
 	scenarios := []scenario{
 		{
-			name:           "not-found",
```
abhinavdahiya: can we keep the `name` these are useful to users looking at the result for a test.
`fmt.Sprintf("%s %s", scenario.request.Method, scenario.request.URL.Path)` leaves more to be desired...

pkg/server/api_test.go
```diff
@@ -18,49 +19,90 @@ func (ms *mockServer) GetConfig(pr poolRequest) (*ignv2_2types.Config, error) {
...
+			handler.ServeHTTP(w, scenario.request)
+
+			resp := w.Result()
+			defer resp.Body.Close()
+
+			if resp.StatusCode != scenario.expectedStatus {
+				t.Errorf("API Handler test failed: expected: %d, received: %d", scenario.expectedStatus, resp.StatusCode)
+			}
+
+			contentLength := int(resp.ContentLength)
```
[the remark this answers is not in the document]
wking: > does this automatically read Content-Length from headers or is this value different?

The tests suggest that it is populated from the `Content-Length` header, although I haven't dug into the Go implementation and the godocs aren't clear.

pkg/server/api_test.go
wking: > can we keep the `name`...

Actually, I think the method/path thing is more readable ;).  But I've taken a stab at restoring names with 22d0c45 -> 8c9980d.

pkg/server/api_test.go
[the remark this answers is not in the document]
abhinavdahiya: hmm, its fine for now then,

pkg/server/api_test.go
```diff
@@ -18,49 +19,90 @@ func (ms *mockServer) GetConfig(pr poolRequest) (*ignv2_2types.Config, error) {
...
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if scenario.request.Method == http.MethodHead {
```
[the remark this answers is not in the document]
wking: > you can add `checkResponse`...

Once we do this, we might as well pull the status check into the helper as well.  I've taken a stab at this with 8c9980d -> 0eebfed, although personally I find the clubbing-them-into-one approach more readable ;).

pkg/server/api.go
abhinavdahiya: So looking at https://github.com/openshift/machine-config-operator/pull/246/commits/5f0a6a7f5d29f9d97608ff325e52f62308165c5b
```
The defaultHandler gives us our usual empty 404s for requests that
don't match our other handlers.  Without it, we'd have returned Go's
default:

  404 page not found

body, and had unit-test failures like:

  --- FAIL: TestHandler (0.00s)
      --- FAIL: TestHandler/GET_/does-not-exist (0.00s)
          api_test.go:93: response body length 19 does not match Content-Length -1
```

I makes me go back to to the original thing that we should not setting/expecting `Content-Length` when **we** only care about the StatusCode.

So maybe can we remove all the `w.Header().Set("Content-Length", "0")` and only set it when we care, ie HEAD/GET success result.

pkg/server/api_test.go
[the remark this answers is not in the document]
abhinavdahiya: @wking thoughts on https://github.com/abhinavdahiya/machine-config-operator/commit/cc12da4b8c75331eff26f96252d5203c3839f170

IMO ^^ makes the somebosy reading the test scenarios pretty clear upfront what we are testing each one against?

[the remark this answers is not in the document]
wking: > thoughts on abhinavdahiya@cc12da4

I like it, thanks :).  Squashed in with 124fd90 -> 01d22e1.  In addition to your commit, I pulled `checkContentType` into a helper as well so we can run it on `HEAD /config/master` as well.  I also tweaked some of the error messages to be more consistent.

pkg/server/api.go
wking: > It makes me go back to to the original thing that we should not setting/expecting `Content-Length` when *we* only care about the StatusCode.

If we want to return `404 page not found\n` (with `Content-Type: text/plain; charset=utf-8`?) for all of our 404s, I'm ok with that (although I don't see a need to do it).  But I would like to be consistent among our 404s, and not have some that return with no body and some that return with the text/plain body.

abhinavdahiya: lets keep it as is for now..

PR #143, opened 2018-10-23
etcd boostrapping using dns

abhinavdahiya:
1. cmd: add setup-etcd-environment

This outputs an environment file with envs that allow etcd to discover its ip address
and the corresponding dns name from the discovery-srv records.

It looks up the `_etcd-server-ssl._tcp.<domain specified by --discovery-srv>` for reverse lookup
of its own dns name.

An example of such file is
```console
$ cat /etc/etcd.env/etcd-environment
ETCD_IPV4_ADDRESS=192.168.126.11
ETCD_DNS_NAME=adahiya-0-etcd-0.tt.testing
```
when the `--discovery-srv` was `tt.testing`
```console
dig +noall +answer _etcd-server-ssl._tcp.tt.testing SRV
_etcd-server-ssl._tcp.tt.testing. 0 IN  SRV     0 10 2380 adahiya-0-etcd-0.tt.testing.
```

The command retries reverse lookup until 5 minutes to allow for dns to be available.

2.  templates: bootstrap etcd using dns discovery

Using the guide https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/clustering.md#dns-discovery

Also adds `setup-etcd-environment.service` that uses the setup-etcd-environment cli to setup the discovery params.

3. cleanup across controller, server and operator to drop etcd index, etcd initial count.

This requires https://github.com/openshift/installer/pull/526

PR #243, opened 2018-12-17
pkg/server: Drop unused etcdTemplateParam

wking:
The last consumer was removed by 4cc79882 (#143).

### DIFF
--- a/pkg/server/bootstrap_server.go
+++ b/pkg/server/bootstrap_server.go
@@ -54,9 +54,8 @@ func NewBootstrapServer(dir, kubeconfig string) (Server, error) {
 // 		"<serverBaseDir>/machine-configs/<currentConfig>.yaml"
 //
 // 3. Load the machine config.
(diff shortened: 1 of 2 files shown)


pkg/server/bootstrap_server.go
```diff
@@ -54,9 +54,8 @@ func NewBootstrapServer(dir, kubeconfig string) (Server, error) {
 // 		"<serverBaseDir>/machine-configs/<currentConfig>.yaml"
 //
 // 3. Load the machine config.
-// 4. Execute the etcd template function based on the etcd_index, if passed in the request .
-// 5. Append the machine annotations file.
-// 6. Append the KubeConfig file.
+// 4. Append the machine annotations file.
+// 5. Append the KubeConfig file.
 func (bsc *bootstrapServer) GetConfig(cr poolRequest) (*ignv2_2types.Config, error) {
```
wking: This will conflict with #246, but rebasing whichever lands second will be easy, so I don't think the conflict should affect review.

PR #244, opened 2018-12-17
pkg/server/server_test: Use filepath.Join for path creation

wking:
Future-proofing for folks who want to run the unit tests on Windows ;).

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"path/filepath"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Not sure if I really want to support people developing directly on Windows long term...I'd like to add more tests that e.g. may use Linux container features locally.

But sure, doesn't do any harm for now.

/lgtm

wking:
> One question, is this all of the locations where pathing was not using stdlib within the repo or just what was noticed at first look?

I'm not sure what the stdlib part of that sentence was asking, but I don't think the old code caused any problems on POSIX systems.  I just noticed that it was using `path.Join` to construct a filesystem path, and `filesystem.Join` is a better semantic match for that use-case.

ashcrow:
@wking Sorry, the stdlib comment was more so noting if you went through this code base and only found the items in this PR or if we possibly have more places that use `+"/"`, etc.. instead of using libraries provided by the golang standard library for putting file system paths together.

wking:
> ... if you went through this code...

I did not do a full audit, just noticed these two calls in passing.

Issue #234, opened 2018-12-14
Add a Degraded annotation


cgwalters:
Right now one of my nodes went degraded, apparently the MCD restarted so I don't have those logs, and more worrying I can't find anything in the journal.

I'm thinking we want a new node annotation:
```
machineconfiguration.openshift.io/state=Degraded
machineconfiguration.openshift.io/degraded="Unable to reconcile ..."
```

jlebon:
Did you check `oc logs -p` too?

kikisdeliveryservice:
I usually see the reason in the `oc logs -p` as well.

cgwalters:
Labels are query-able; and we definitely want to make it easy for admins to find degraded nodes.  https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

cgwalters:
Updating this since labels aren't meant for free-form text; I think we should cover the extended rationale with both an event a systemd journal entry like we do for reboots etc.

cgwalters:
Retitling this to use an annotation instead; in https://github.com/openshift/machine-config-operator/pull/260 we decided not to add a label.

cgwalters:
Also, the MCP really needs to show `(total: 3, updated: 0, unavailable: 1, degraded: 1)` or so.

runcom:
do we want _any_ error during a sync which can result in degraded/unreconcilable to be added in this annotation? I want to hash this out so we can bubble this information up from MCD logs to nodes annotation for an instant debug, wdyt?

kikisdeliveryservice:
@runcom adding the _reason_ an MC is unreconcilable would be extremely helpful for a user rather than them having to bumble thru MCD logs.

runcom:
Yeah, but we also want to provide explaination when for instance, pivot fails but that just marks degraded, not unreconcilable uhm

PR #245, opened 2018-12-17
daemon: Rework state management for updates 

cgwalters:
The high level of this code is greatly reducing duplication
between "check for updates" and "apply updates".  The
"check for updates" is now just comparing config names.

I found the previous code flow hard to understand and also
very "implicit".

When we were going to apply a config, we'd look at *three*
things (currentConfig, on disk state, and desiredConfig), and
treat things as done if "on disk state" == "desiredConfig".

But that conflates the case where *we* are writing state versus
how we're trying to check for any drift from *other* processes
(e.g. admin logging in via ssh) changing things.

First, let's assume that we can simply compare the *names* of
machineconfigs when validating state.  Since we're hashing them,
and also using their names as node annotations, they should really
be immutable.  In other words, it's explicitly not supported to
`oc edit machineconfigs/worker-<hash>`.

The bigger change here is that when we go to apply a config,
we maintain on-disk the tuple `(desiredConfig, bootID)`.  This
tuple captures whether we've booted into the desired state.

When the MCD comes back up now, we read that file and if it
matches what we expect, then the update is complete.

Now, the "check the on disk state for drift from expected config"
is factored out into a new function.  We do this every time on
MCD start, although now we could e.g. change things to do it at
most once a day, writing a "last checked" timestamp in `/run` or as a node annotation or
whatever.


cgwalters:
Previous PR w/a bit more historical context: https://github.com/openshift/machine-config-operator/pull/235

ashcrow:
```
 net/http: TLS handshake timeout
```

flake

/retest

cgwalters:
Although hmm...I guess we're actually relying on the fact that none of the files change as a way to signal completion!  OK, so let's use a stamp file for that.

pkg/daemon/daemon.go
```diff
@@ -498,60 +505,51 @@ func (dn *Daemon) triggerUpdate() error {
...
-	reconcilableError := dn.reconcilable(currentConfig, desiredConfig)
-	if reconcilableError != nil {
-		return false, "", nil
+		return false, nil, err
 	}
+	desiredConfigName := desiredConfig.GetName()
 
-	isDesiredOS := false
-	// We only deal with operating system management on RHCOS
-	if dn.OperatingSystem != MachineConfigDaemonOSRHCOS {
```
abhinavdahiya: is all checking for OS version betting dropped?

kikisdeliveryservice:
Finally caught up on this and have 2 questions  @cgwalters :
Question 1 :  Some of your other PRs were working toward adding more specific & verbose logging to give the user more usable information during failures.  This PR seems to be essentially saying "machineconfig1 isn't machine config2" which seems to be not as informative as the other types of messages you were proposing.

Question 2 (related to @abhinavdahiya comment) : In the current version, if the OS is not RHCOS, we mark as true since we won't be updating and keep going to compare the configs. So that a config that attempts to change a non-RHCOS machine will be fine so long as the rest of the config is legit and that intent to upgrade the OS is ignored.  It seems that this PR changes that by removing the checkOSImage part and would cause a config to now fail - which I don't think is the intended behavior.

pkg/daemon/daemon.go
cgwalters: Nope, we still have:
```
func (dn *Daemon) updateOS(oldConfig, newConfig *mcfgv1.MachineConfig) error {
	if dn.OperatingSystem != MachineConfigDaemonOSRHCOS {
		glog.V(2).Infof("Updating of non RHCOS nodes are not supported")
		return nil
	}
```

In the big picture, this is dropping a lot of duplicate logic between "check for update" and "update".  Now "check for update" is just "compare machine config names".

cgwalters:
> This PR seems to be essentially saying "machineconfig1 isn't machine config2" which seems to be not as informative as the other types of messages you were proposing.

We still do more verbose logging inside `reconcilable()`.

>  It seems that this PR changes that by removing the checkOSImage part and would cause a config to now fail - which I don't think is the intended behavior.

See https://github.com/openshift/machine-config-operator/pull/245#discussion_r242559495

pkg/daemon/daemon.go
```diff
@@ -498,60 +505,51 @@ func (dn *Daemon) triggerUpdate() error {
...
-		// If we are on anything but RHCOS we set to True as there
-		// is nothing to updated.
-		isDesiredOS = true
+	if currentConfigName != desiredConfigName {
```
jlebon: I'm not sure I follow this. Won't this simpler logic result in a reboot loop?

- a new desired config node annotation is set
- the daemon sees it and applies the update and reboots the system
- the daemon comes back online after the reboot and goes through here
- `currentConfigName != desiredConfigName` evaluates to true since the node annotation hasn't been updated
- daemon tries to apply the same update again

We need to only be caring about the desired config, not the current config, because this is how we determine whether we should update the current config annotation itself after an update.

cgwalters: Yep, I have a followup in https://github.com/openshift/machine-config-operator/pull/245#issuecomment-448017388

cgwalters: I pushed my WIP on top of this below :arrow_down: so you all can see what I'm thinking.

cgwalters:
The high level idea of this "transient state file" is to explicitly record on the node whether the step between "drain and reboot" :arrow_right:  "start up again" has been completed. 

This also nicely solves the problem with https://github.com/openshift/machine-config-operator/pull/237 - the MCD is basically serializing the trivial bit of state it needs (desiredConfig) to disk, so it can be safely killed.

cgwalters:
OK, this PR is entirely reworked!

pkg/daemon/daemon.go
```diff
@@ -124,13 +140,21 @@ func New(
 		}
 		glog.Infof("Booted osImageURL: %s (%s)", osImageURL, osVersion)
 	}
+
+	currentBootIDBytes, err := ioutil.ReadFile("/proc/sys/kernel/random/boot_id")
```
ashcrow: In the future I'd like us to try to move things like this into functions. Not needed now though.

pkg/daemon/daemon.go
```diff
@@ -291,6 +315,76 @@ func (dn *Daemon) getHealth() error {
 	return nil
 }
 
+// stateAndConfigs is the "state" node annotation plus parsed machine configs
+// referenced by the currentConfig and desiredConfig annotations.
+type stateAndConfigs struct {
```
ashcrow: This struct and fields are never mean to be accessed outside of the `daemon` package, correct?

ashcrow:
The changes on their own make sense to me. Did you test that `once-from` continued to work as well?

pkg/daemon/daemon.go
```diff
@@ -124,13 +140,21 @@ func New(
...
+	currentBootID := strings.TrimSpace(string(currentBootIDBytes))
+
 	dn := &Daemon{
 		name:                   nodeName,
 		OperatingSystem:        operatingSystem,
 		NodeUpdaterClient:      nodeUpdaterClient,
 		loginClient:            loginClient,
 		rootMount:              rootMount,
 		fileSystemClient:       fileSystemClient,
+		bootID:                 currentBootID,
```
ashcrow: Does the `NewClusterDrivenDaemon` need to know the `bootID` or only when it's not cluster connected?

pkg/daemon/daemon.go
cgwalters: Yep!

pkg/daemon/daemon.go
cgwalters: Factored out into a new function locally.

pkg/daemon/daemon.go
cgwalters: Currently we're assigning it in both cases (this is the common constructor).  However whether that's necessary gets into the onceFrom question.

pkg/daemon/daemon.go
```diff
@@ -291,6 +315,76 @@ func (dn *Daemon) getHealth() error {
...
+	return &stateAndConfigs{
+		currentConfig: currentConfig,
+		desiredConfig: desiredConfig,
+		state: state,
+	}, nil
+}
+
+// isConfigCompletedPending checks whether we've successfully
+// completed a config, and are coming up after a reboot.
+func (dn *Daemon) isConfigCompletedPending(pending string) (bool, error) {
```
cgwalters: This function is probably the most important change here; it's a bit subtle but makes sense I think.

cgwalters:
I tested `once-from` with this PR, seems fine to me!

pkg/daemon/daemon.go
```diff
@@ -300,31 +402,48 @@ func (dn *Daemon) getHealth() error {
...
+	isPending, err := dn.isConfigCompletedPending(stateAndConfigs.desiredConfig.GetName())
 	if err != nil {
-		return dn.nodeWriter.SetUpdateDegradedIgnoreErr(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
+		return err
 	}
 
-	if isDesired {
-		// we got the machine state we wanted. set the update complete!
-		if err := dn.completeUpdate(dcAnnotation); err != nil {
+	if isPending {
```
jlebon: I think here we should still re-validate the on-disk state before actually completing the update. We rebooted, and lots of services/scripts ran since we last applied the update. Also, in the context of automatic rollback, it's not as clear that a different boot ID implies "the rootfs we wanted to reboot into". In fact, it could be the same deployment, right? (E.g. if deployment finalization failed -- side note: we should probably enhance `pivot` to have a `--finalize` option or something so we do this upfront and fail early).

I think we should still also call `dn.checkOS()` in `validateOnDiskState()` for that reason. Checking the on-disk state is not that expensive when put in the context of how heavyweight a whole reboot process is already. So it's a mostly "free" operation in exchange for maintaining that crucial invariant that the `desiredConfig` annotation actually reflects reality.

cgwalters:
Turns out I broke the case of bringing up new nodes; see (untested) change in https://github.com/openshift/machine-config-operator/pull/245/commits/fdce2bb33f62eba2b2e8cac513aa8b74714302d2
and (tested)  change in https://github.com/openshift/machine-config-operator/pull/245/files#diff-06b532a2a5a5d3ad1d5341b3e00945efR420

pkg/daemon/daemon.go
cgwalters: OK, makes sense.  Will do

pkg/daemon/daemon.go
```diff
@@ -300,31 +402,48 @@ func (dn *Daemon) getHealth() error {
...
-	if state, err := getNodeAnnotationExt(dn.kubeClient.CoreV1().Nodes(), dn.name, MachineConfigDaemonStateAnnotationKey, true); err != nil {
-		// try to set to degraded... because we failed to check if we're degraded
-		return dn.nodeWriter.SetUpdateDegradedIgnoreErr(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
-	} else if state == MachineConfigDaemonStateDegraded {
-		// just sleep so that we don't clobber output of previous run which
-		// probably contains the real reason why we marked the node as degraded
-		// in the first place
+	stateAndConfigs, err := dn.getStateAndConfigs()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
cgwalters: OK, though I don't think we want fetching annotations to result in a degraded state personally.

cgwalters:
OK, updated again to address comments, the biggest of which is https://github.com/openshift/machine-config-operator/pull/245#discussion_r242947874

pkg/daemon/daemon.go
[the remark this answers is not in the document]
jlebon: If we can't fetch annotations, then there's no way to monitor for annotation changes and react to updates, right? Though yeah, we should probably add retry loops like we did for the MachineConfig here to handle transient errors.

kikisdeliveryservice:
@cgwalters Question: Could you write up a bit about testing this PR?  I'm not quite sure what I'm looking for if I pull it.  Thanks!

cgwalters:
> @cgwalters Question: Could you write up a bit about testing this PR? I'm not quite sure what I'm looking for if I pull it. Thanks!

Sure!  I have been testing this in a local libvirt cluster with multiple nodes (e.g. do an install, then `oc -n openshift-cluster-api scale --replicas=3 machinesets/osiris-worker-0`).    Then I've been applying a MC somewhat like the one in `HACKING.md`.

Of note is that I haven't yet actively tested some important *failure* conditions; e.g. forcing `validateOnDiskState()` to fail or a reconcilation failure.

What I find most useful about this PR though is that the logging and state flow is *much* clearer IMO.  For example, after applying a MC and having a node update, use `logs -p` on the daemonset and look at what the MCD did before reboot. 

In the updating logs you should now see e.g.:

```
I1219 21:31:53.341323    5123 daemon.go:421] In desired config worker-20de641884b69a214f67bad9693e1c9d
...
I1219 21:34:07.855179    5123 update.go:72] Checking reconcilable for config worker-20de641884b69a214f67bad9693e1c9d to worker-68ed5f576b4dce3a07f69ae2297776b2
...
I1219 21:34:07.949195    5123 update.go:529] machine-config-daemon initiating reboot: Node will reboot into config worker-68ed5f576b4dce3a07f69ae2297776b2
```

Then after reboot:

```
I1219 21:34:33.928850    3420 daemon.go:435] Completing desired config worker-68ed5f576b4dce3a07f69ae2297776b2
I1219 21:34:33.928897    3420 daemon.go:675] No target osImageURL provided
I1219 21:34:33.932612    3420 daemon.go:444] Validated on-disk state
I1219 21:34:34.033216    3420 update.go:529] machine-config-daemon: completed update for config worker-68ed5f576b4dce3a07f69ae2297776b2
```

jlebon:
This looks good to me! Will give it a whirl as well.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
cgwalters: Exactly, I think we should consistently retry-loop for network bits.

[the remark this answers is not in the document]
jlebon: OK, split this out in #252 to make sure we don't lose it.

cgwalters:
OK so I have been thinking about this more, and there's a case I realized this isn't handling correctly that presumably the old code was trying to do, which is:

What happens when desiredConfig changed while we were rebooting?  In that case, we'd go degraded, because we'd try to validate the on-disk state against the *new* desired config before we'd applied it.

Furthermore, we'd skip changing `currentConfig`...so if the daemon got restarted, we'd also go degraded because we'd be confused about what state we were in.

To make this more rigorous, I think we should enforce that we only go between config A to B "atomically"; we always validate the steps there, and if desiredConfig changes to C while we're working on B, we finish B first.

cgwalters:
Another thing I wanted to add here (and will in the commit message) is that a top-level motivation for this is avoiding reboot loops.  I kept hitting those and trying to reason out why they were happening and how to avoid it led to this PR.

Reboot loops are a huge pain in the butt to debug (I will probably add a PR to add e.g. a 30s sleep before we reboot), and also I think will very problematic in production.

cgwalters:
> To make this more rigorous, I think we should enforce that we only go between config A to B "atomically"; we always validate the steps there, and if desiredConfig changes to C while we're working on B, we finish B first.

Pushed a rework of this as a fixup :arrow_up:

pkg/daemon/daemon.go
```diff
@@ -359,54 +380,65 @@ func (dn *Daemon) getStateAndConfigs() (*stateAndConfigs, error) {
...
-		// TODO; don't go into degraded here
-		dn.nodeWriter.SetUpdateDegradedIgnoreErr(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
-		stateAndConfigs.state = MachineConfigDaemonStateDegraded
+		return err
+	}
+	stateAndConfigs, err := dn.getStateAndConfigs(pendingConfigName)
+	if err != nil {
+		return err
 	}
 	if stateAndConfigs.state == MachineConfigDaemonStateDegraded {
```
jlebon: Let's change the `select {}` here to just return an error so we sleep in the new `EnterDegradedState()`?

pkg/daemon/daemon.go
```diff
@@ -417,49 +449,65 @@ func (dn *Daemon) CheckStateOnBoot() error {
...
-		if stateAndConfigs.state == "" {
-			if err := dn.nodeWriter.SetUpdateDone(dn.kubeClient.CoreV1().Nodes(), dn.name, stateAndConfigs.desiredConfig.GetName()); err != nil {
-				return err
-			}
-		}
-		completingPendingUpdate = false
+	// Validate the on-disk state against what we *expect*.
+	//
+	// In the case where we're booting a node for the first time, or the MCD
+	// is restarted, that will be the current config.
```
jlebon: My worry with this is that we expect the MCD to always be restarted when the machine state is clearly defined. E.g. what if the MCD gets OOM-killed while writing out files? Can we relax this a bit in the following way?

- in the pendingConfig case, error out if validation fails
- in the case currentConfig == desiredConfig already, error out if validation fails
- otherwise, just try to apply the desiredConfig

cgwalters: Well...we could introduce a "writingConfig".  

I dunno though...I see your point, but I feel like the best course is to land this and iterate from there.  Having the MCD die half-way through is a problem we could solve better for sure, but...worst case scenario is that validation fails, we go degraded, and the machine eventually gets reprovisioned right?  Wwe also have bigger problems like needing the retry loops around fetching configs that are impacting us a lot more today.

pkg/daemon/daemon.go
cgwalters: I started to make this change then remembered there's a reason I wrote it this way; it's that we want to clearly distinguish between the case of "we're explicitly transitioning the system to degraded because $REASON" from "we started up degraded".

The idea here of course is that `EnterDegradedState` is where we'd implement https://github.com/openshift/machine-config-operator/issues/234

pkg/daemon/daemon.go
jlebon: I'm mentioning it because this is behaviour we're regressing on, right? I.e. the current code does handle this case. Not sure in prod clusters how likely it is, but it's definitely helpful for when iterating on the MCD itself on the bits that actually perform the update. I'm good with doing this as a follow-up though.

jlebon:
Gave this a spin locally and it works well! Want to squash that fixup in?
@kikisdeliveryservice were you able to test it as well?

kikisdeliveryservice:
@jlebon I haven't had a chance since my cluster has been otherwise occupied.  Happy to trust someone else to test as well, if they can do it before I have some free time. :)

cgwalters:
> Want to squash that fixup in?

Done!

jlebon:
I was trying to test this some more this morning but I'm hitting issues with my local cluster now. Anyway, we can fix up anything else that comes up in followups.
/lgtm

cgwalters:
> I would have liked to see more concrete additions to the docs for the daemon as this changing few core code flow.

Yeah, makes sense.  I will do that.

cgwalters:
> Yeah, makes sense. I will do that.

https://github.com/openshift/machine-config-operator/pull/263

PR #246, opened 2018-12-17
pkg/server: Trim public API to return *http.Server

wking:
Builds on #240; review that first.

The previous API required consumers to construct a handler and pass that handler back into a separate function to create a server based on the handler.  With this commit, I've rerolled the API to return the handler directly, and the new public API is just `NewBootstrapServer` and `NewClusterServer`.  To serve both HTTP and HTTPS from the returned server, I'm making shallow copies and altering `Addr` to set the per-protocol ports in those copies.

I've also made some `server` -> `config` renames in the types that have `getConfig` methods.  And `APIHandler` is now `configHandler`.  That helps consolidate the `/config/*` handling under a consistent name, so we have:

* An `*http.Server` that is the server (e.g. as returned by `NewClusterServer`).
* That server's handler is a muxer, and one of the muxer handlers is the `configHandler` for `/config/*`.
* The `configHandler` structure holds a `getConfig` function, and we have `bootstrapConfig` and `clusterConfig` which can supply `getConfig`.  Using a function type for this avoids the need for an interface and mock `getConfig` type.


abhinavdahiya:
IMO i liked the interface and i want to keep it that way.

wking:
> IMO i liked the interface and i want like to keep it that way.

Do you have suggestions for testing the complete muxer with the approach master's currently taking?  Currently it's [tucked in there with the code launching listen-and-server goroutines][1].

And, if we're looking for smaller chunks of 8eef742816f1 where we agree, do you like `bootstrapServer` as the name of a struct whose only method is `GetConfig`?  Or are you ok with me renaming that to `bootstrapConfig`?

[1]: https://github.com/openshift/machine-config-operator/blob/a5d071369728ded41ddeaa9fca6dff376346d7d2/pkg/server/api.go#L45-L60

wking:
[unit][1]:

```
2018/12/18 05:39:17 Executing test unit
2018/12/18 05:45:00 Ran for 7m12s
error: could not run steps: failed to wait for test pod to complete: could not wait for pod completion: the pod ci-op-wyc89v0p/unit failed after 5m41s (failed containers: ):  unknown
```

Dunno what that's about.

/retest

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/246/pull-ci-openshift-machine-config-operator-master-unit/459/build-log.txt

wking:
Rebased onto master with 5f0a6a7 -> 6cfa51c now that #240 has landed.

PR #253, opened 2018-12-19
controller: Log when we generate a MC

cgwalters:
I'd like a bit more visibility into what the controller is doing.
While I could `oc watch machineconfigs` too, I think this one is
worth a log message.

### DIFF
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -3,6 +3,7 @@ package render
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -410,7 +411,6 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		return fmt.Errorf("no MachineConfigs found matching selector %v", selector)
 	}
 
-	glog.V(4).Infof("Syncing generated machineconfig for pool %s using (%d) machineconfigs", pool.GetName(), len(mcs))
 	return ctrl.syncGeneratedMachineConfig(pool, mcs)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -3,6 +3,7 @@ package v1
 import (
 	"sort"
 
+	"github.com/golang/glog"
```
kikisdeliveryservice: accidental ide auto-add?

pkg/controller/render/render_controller.go
```diff
@@ -427,6 +427,7 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 	_, err = ctrl.mcLister.Get(generated.Name)
 	if apierrors.IsNotFound(err) {
 		_, err = ctrl.client.MachineconfigurationV1().MachineConfigs().Create(generated)
+		glog.Infof("Generated machineconfig %s from %d sources", generated.Name, len(configs))
```
jlebon: I like this. How about we also print the names of all the source configs?

abhinavdahiya: already logs here
https://github.com/openshift/machine-config-operator/pull/253/files#diff-9d1c255b5e10f6159f19c5fafa976458R413

abhinavdahiya:
> I'd like a bit more visibility into what the controller is doing.
While I could oc watch machineconfigs too, I think this one is
worth a log message.

- Use verbosity levels.
- Use events for major actions

cgwalters:
> Use events for major actions

Can we do that as a separate PR after https://github.com/openshift/machine-config-operator/pull/256 ?

pkg/controller/render/render_controller.go
cgwalters: That one seems to happen though even if nothing changed; gets verbose.  I deleted it.

Issue #183, opened 2018-11-19
Design for osImageURL updates - integration with CVO/release payload


cgwalters:
I wanted to elaborate here on the current status of this.  We have a PR in https://github.com/openshift/machine-config-operator/pull/363 which will finally close the loop and inject `machine-os-content` from the release payload all the way into the MachineConfig objects, which will result in the MCD updating.

The final architecture will be:

New kernel errata, turns into RPM, converted into ostree then oscontainer.  A bit more information on the [build system side here](https://github.com/openshift/origin/pull/21998#issuecomment-461943604).  The oscontainer makes it into a new release payload published on quay.io.

At some point the release payload pulled down by CVO, which includes a [osimageurl ConfigMap](https://github.com/openshift/machine-config-operator/pull/305) that references that container (same thing as the `machine-os-content` ImageStream).  The CVO updates the ConfigMap, which you can see via

`oc -n openshift-machine-config-operator get configmap/machine-config-osimageurl`

The operator notices the change to the configmap and updates the "controllerconfig" which is an internal CRD that is used as the primary input to the MCC.  See `oc get -o yaml controllerconfig`.

The "template" sub-controller of the MCC then updates `machineconfigs/00-master` and `machineconfigs/00-worker`.

The "render" sub-controller of the MCC generates new "rendered" MCs that look like `machineconfigs/master-<hash>` and `machineconfigs/worker-<hash>` and updates the MachineConfigPools to target them.  For more information on this, see the MCC docs.

On each node the MCD will get the new osimageurl, and if it's different than what's booted, it will pull down the container and rebase to it and reboot. This is also the same as any other config change.

jlebon:
For completeness, one alternative is to make it a `MachineConfig` object instead, which is then used as the base config containing the `osImageURL` in which all other (Ignition) configs are merged into. I think this maps slightly more closely to what's currently done today.

That said, using a separate `ConfigMap` allows us to have richer metadata. E.g. pkg lists & diffs, OSTree commits & versions, etc... The idea being that an "Update available" button can show all that stuff without having to hunt down the osImage.

ashcrow:
Remember, this data will need to flow one way or another down into a `MachineConfig`. If it's easier we could add some more information to MCs themselves.

jlebon:
Strawman:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: rhcos-release-info
data:
[...]

PR #258, opened 2018-12-20
DO NOT MERGE - Example of doing replacement

smarterclayton:
Demonstrating how you can have an RHCoS content image (located in namespace `openshift` and image stream tag `origin-v4.0:machine-image-content`) and consume it from an operator.

@cgwalters


cgwalters:
This is part of a conversation around https://github.com/openshift/machine-config-operator/issues/183

cgwalters:
So in the case of a reference to an image not generate by the CI process, the release operator will simply pull in that image and mirror it into the release payload?

cgwalters:
Superceding by https://github.com/openshift/machine-config-operator/pull/273

/close

PR #260, opened 2018-12-21
Label degraded

cgwalters:
Depends: https://github.com/openshift/machine-config-operator/pull/245


cgwalters:
```
$ oc logs pods/machine-config-daemon-qgfv6
I1221 14:53:26.735013   27475 start.go:52] Version: 3.11.0-379-ga8e88e7e-dirty
I1221 14:53:26.735643   27475 start.go:89] starting node writer
I1221 14:53:26.741413   27475 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I1221 14:53:26.814095   27475 daemon.go:150] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:9b786d1cbcba81e3fc5a03c744a247613dc469046024847316cfdc20977b487a (47.230)                        
I1221 14:53:26.849526   27475 start.go:140] Calling chroot("/rootfs")
I1221 14:53:26.864780   27475 daemon.go:712] No target osImageURL provided
E1221 14:53:26.865252   27475 daemon.go:793] content mismatch for file: "/home/core/test"; expected: hello world                                                                                                  
; received: hello world OOPS
E1221 14:53:26.865302   27475 daemon.go:330] Fatal error checking initial state of node: Checking initial state: Unexpected on-disk state                                                                         
E1221 14:53:26.865331   27475 writer.go:86] Marking degraded due to: Checking initial state: Unexpected on-disk state                                                                                             
I1221 14:53:26.878936   27475 daemon.go:332] Entering degraded state; going to sleep
$ oc get nodes -l machineconfiguration.openshift.io/degraded
NAME                    STATUS    ROLES     AGE       VERSION
osiris-worker-0-wksbf   Ready     worker    18h       v1.11.0+62de992
$
```

pkg/daemon/constants.go
```diff
@@ -14,6 +14,10 @@ const (
 	// MachineConfigDaemonStateDegraded is set by daemon when update cannot be applied.
 	MachineConfigDaemonStateDegraded = "Degraded"
 
+	// This label was introduced in addition to state/Degraded because we
+	// want operators and admins to be able to easily query this state.
+	MachineConfigDaemonLabelDegraded = "machineconfiguration.openshift.io/degraded"
```
jlebon: How about just making this `machineconfiguration.openshift.io/state` and have the label value be the state, just like the annotation? That way we can also easily query for `Working` and `Done` too.

cgwalters:
/hold

Pending 245

pkg/daemon/constants.go
cgwalters: Well at that point, why have both annotations and labels?  Seems like you're saying we should basically switch from annotations to labels?  That could make sense.  I'd want the opinion of @abhinavdahiya for that.

cgwalters: My thought on this though is that an admin (or operator) that wants to know about working state should be querying the `machineconfigpool`, not looking at nodes.

IOW, the working/done annotations are mostly an implementation detail of `machineconfigpool`.  But degraded is a much bigger deal and something that e.g. the machine-api-operator should be involved with.

abhinavdahiya: Don't add labels. Labels have greater meaning in kubernetes. Only annotations should be used.

abhinavdahiya:
I disagree with motivation of the label that we want admins to query nodes with label degraded. The pool CRD should be the one people use. Also degraded is only valid in realm of the machine config daemon.

cgwalters:
> I disagree with motivation of the label that we want admins to query nodes with label degraded. The pool CRD should be the one people use. Also degraded is only valid in realm of the machine config daemon.

The user story I'm thinking of here (and hit myself regularly when testing) is that I want to find, then `oc adm drain`, then `oc delete` degraded nodes.  Labels seem like a good way to do "find".  How do I use the pool to do that?

cgwalters:
OK so based on real time discussion, closing this one.  One thing I missed is that by default we should have a max of one unavailable.  Admins should be able to find the degraded node by looking at the machineconfigpool.

For the case of automated discovery, Abhinav was suggesting that rather than having the machine-api-operator take care of reprovisioning, that's something we'd handle.

Issue #252, opened 2018-12-19
Add retries when trying to fetch or set node annotations


jlebon:
Because we don't want transient network blips to take down a whole node.

PR #262, opened 2018-12-21
daemon: Add retries around fetching annotations

cgwalters:
Just like we do with `getMachineConfig()`.

Closes: https://github.com/openshift/machine-config-operator/issues/252

### DIFF
--- a/pkg/daemon/node.go
+++ b/pkg/daemon/node.go
@@ -3,10 +3,13 @@ package daemon
 import (
 	"encoding/json"
 	"fmt"
+	"time"
 	"io/ioutil"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/daemon/node.go
```diff
@@ -51,7 +54,15 @@ func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (stri
...
 func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allowNoent bool) (string, error) {
-	n, err := client.Get(node, metav1.GetOptions{})
+	var n *core_v1.Node
+	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+		var err error
+		n, err = client.Get(node, metav1.GetOptions{})
+		if err == nil {
+			return true, nil
+		}
+		return false, nil
```
yuqi-zhang: Shouldn't we return the error here? (We also didn't do it in the loop in `getMachineConfig`)

yuqi-zhang: Ah, right, sorry. I think it's fine as is for now then.

PR #265, opened 2019-01-02
allow operator to tie into node upgrade status

abhinavdahiya:
MachineConfigController controls the the machineconfigs that need to be applied to the nodes.

MachineConfigOperator should be able to recognize when the expected machineconfig has been applied to all the existing nodes once.

This would allow MCO to report Done on its ClusterOperator only when pre-determined machinecongfigpools have updated to newest desired configurations.
xref: https://jira.coreos.com/browse/CORS-830

**allow operator to check what version of controller last modified a machineconfig**
the controller creates and owns some of default configs. There is no indication on the machineconfig that would allow us to capture which version of the
controller actually created or last modified the object.

This inforamtion is super helpful for the operator to decide if the machineconfig has been last modified by `the` controller it is expecting

**operator must wait for certain pools to complete as part of the sync**
Currently the operator only  cares about deploying and maintaining the controller, server and daemon.
But during an upgrade the operator must ensure that all the nodes in *certain* machineconfigpools have upgraded to the latest configuration.

for example, it is essential that during upgrade the operator ensure that the all the nodes in master pool have updated to the latest configuration before it can report `done`.

for *certain* pools the operator performs
- verfiy that the configuration is generated by the latest controller.
- if above is true, verify that all the nodes in the pool have updated and none are unavailable.

Only when both the conditions are met, that the operator deems the pool done.

To verfiy that the configuration is generated by the latest controller, the operator
- verfies that the generated machineconfig is generated by the correct controller (ie. controller's version matches operator's version)
- verfies that all the machineconfigs that are owned/generated by the controller, were actually generated by the correct controller (ie. controller's version matches operator's version)

/cc @cgwalters @ashcrow


cgwalters:
A bit more background about the high-level goals here would help a lot.

It looks like this is superceding https://github.com/openshift/machine-config-operator/pull/253 in a better way, which is cool.

But there's some higher level stuff going on here I'm not sure about.  It seems like it might relate to my question around "how does an admin or operator use the pool to monitor state"?  Which comes up in https://github.com/openshift/machine-config-operator/pull/260#issuecomment-449475375 and another PR I'm not finding offhand.

abhinavdahiya:
> A bit more background about the high-level goals here would help a lot.

This is still WIP , using this PR to get release image with changes. Will ping with more details when ready 😋

abhinavdahiya:
ping

/retest

cgwalters:
> There is no indication on the machineconfig that would allow us to capture which version of the
controller actually created or last modified the object.

Sounds like this would help with the problem I've hit in development where when I deploy my new controller with osImageURL handling it doesn't know to trigger regeneration of the MCs.

> for example, it is essential that during upgrade the operator ensure that the all the nodes in master pool have updated to the latest configuration before it can report done.

Relatedly I wonder if the CVO should special case to run the MCO first or something...otherwise we'll schedule a whole lot of new pods but we're simultaneously draining and rebooting nodes.

cgwalters:
```apis: include the machineconfigs being used in pool's status
controller: use the new configuration field in machineconfigpool status
server: use the new configuration field in machineconfigpool status 
```

The first here is an API change, so we won't compile in an intermediate state right?  I tend to squash those types of changes together.

abhinavdahiya:
> > There is no indication on the machineconfig that would allow us to capture which version of the
> > controller actually created or last modified the object.
> 
> Sounds like this would help with the problem I've hit in development where when I deploy my new controller with osImageURL handling it doesn't know to trigger regeneration of the MCs.

So we are on the same page that we need it.. ?

> > for example, it is essential that during upgrade the operator ensure that the all the nodes in master pool have updated to the latest configuration before it can report done.
> 
> Relatedly I wonder if the CVO should special case to run the MCO first or something...otherwise we'll schedule a whole lot of new pods but we're simultaneously draining and rebooting nodes.

CVO applies operators in a sequence, so this PR allows MCO to incorporate node upgrade status for masters in status reported to clusteroperator. Next step will be to tell CVO to wait for MCO to be ready before going forward (not in this PR).

cgwalters:
> To verfiy that the configuration is generated by the latest controller, the operator

I was thinking about a related issue: Currently the MCC gathers all MCs; this seems to create a compatibility/upgrade hazard where if say we rename a MC (e.g. `00-worker` to `00-base` say), then during upgrade we may have both, and the MCC would try to union them. 

Offhand...I think we also want the MCC to only apply configs that have a matching `machineconfiguration.openshift.io/generated-by-controller-version` annotation *or* don't have the annotation (which then means "user configuration")?

cgwalters:
A more concrete scenario I'm thinking of: we implement https://github.com/openshift/machine-config-operator/issues/269 and then do the natural thing and instead of the template controller generating `00-worker`/`00-master`, we just generate `00-base`.  But the render controller is likely to see both `00-worker` and `00-base` and union them, right?  Because we're relying on Kube GC of the old objects tied to the old render config (AIUI).

abhinavdahiya:
> Offhand...I think we also want the MCC to only apply configs that have a matching machineconfiguration.openshift.io/generated-by-controller-version annotation or don't have the annotation (which then means "user configuration")?

Sure

> A more concrete scenario I'm thinking of: we implement #269 and then do the natural thing and instead of the template controller generating 00-worker/00-master, we just generate 00-base. But the render controller is likely to see both 00-worker and 00-base and union them, right? Because we're relying on Kube GC of the old objects tied to the old render config (AIUI).

that is the correct behavior.

And moving to #269 and common base is an api change that requires migration using the operator. With the freeze in effect every change to behavior is any api change.

cgwalters:
> Sure

To clarify, you agree we should change the render controller to implement that "version matching"?

> And moving to #269 and common base is an api change that requires migration using the operator.

OK, #269 is an API *addition*.  Let's say we decide that the split between `00-worker` + `00-worker-sshkey` + `00-worker-osimageurl` is unnecessary and just generate `00-worker` (which I think would be saner) - that's just an implementation concern of the controller today.  If we require special operator code to do those migrations that seems like it'd introduce a lot more fragility.

But, my concern there should be largely addressed by the "MC version matching".

PR #267, opened 2019-01-05
server: Add /healthz

abhinavdahiya:
this is alternate approach to #239 
closes #239
look at #239 for more info.

/cc @wking

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -9,10 +9,6 @@ import (
 	"github.com/golang/glog"
 )
 
-const (
-	apiPathConfig = "/config/"
-)
-
 type poolRequest struct {
 	machinePool string
 }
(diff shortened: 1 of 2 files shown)


pkg/server/api.go
```diff
@@ -31,8 +27,13 @@ type APIServer struct {
 // that runs the Machine Config Server as a
 // handler.
 func NewAPIServer(a *APIHandler, p int, is bool, c, k string) *APIServer {
+	mux := http.NewServeMux()
+	mux.Handle("/config/", a)
```
wking: Why take an `*APIHandler` argument, when the only way to get an `*APIHandler` with `.server` set is to call `NewServerAPIHandler`?  Seems like it would be easier to have `NewAPIServer` take a `Server` argument (although see [here][1] about naming that interface `Server`).

[1]: https://github.com/openshift/machine-config-operator/pull/246#issuecomment-448053607

PR #270, opened 2019-01-07
controller: Split out "cluster derived" config generation

cgwalters:
This is a minor refactor of the bits from
a46fcadaea8bc4af095e5826645f8096ab1ae5f6
for SSH keys.  I plan to add another "cluster derived" config
for osImageURL, and I think it's worth splitting those into
their own clearly separate function.

(The functionality is a bit like a "sub-controller" of the MCC)

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -79,22 +79,17 @@ func generateMachineConfigs(config *renderConfig, templateDir string) ([]*mcfgv1
 	return cfgs, nil
 }
 
+// generateMachineConfigsForRole is part of generateMachineConfigs; it operates
+// on a specific role which has a set of builtin templates.
 func generateMachineConfigsForRole(config *renderConfig, role string, path string) ([]*mcfgv1.MachineConfig, error) {
+	// Start with the cluster-derived configs
+	cfgs := generateDerivedMachineConfigs(config, role)
+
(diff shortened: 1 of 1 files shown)


cgwalters:
(didn't test this beyond compilation yet though)

pkg/controller/template/render.go
```diff
@@ -79,22 +79,17 @@ func generateMachineConfigs(config *renderConfig, templateDir string) ([]*mcfgv1
 	return cfgs, nil
 }
 
+// generateMachineConfigsForRole is part of generateMachineConfigs; it operates
+// on a specific role which has a set of builtin templates.
 func generateMachineConfigsForRole(config *renderConfig, role string, path string) ([]*mcfgv1.MachineConfig, error) {
+	// Start with the cluster-derived configs
+	cfgs := generateDerivedMachineConfigs(config, role)
```
kikisdeliveryservice: Thinking about this, I think the func call (to your new function) should be after the error check on the ReadDir(path) since we're picking up the roles by walking through the dirs. So I think the ReadDir should execute successfully until we add our config to cfgs.

cgwalters:
While I think this is still a potentially good idea, it's not worth the churn.  PR https://github.com/openshift/machine-config-operator/pull/273 no longer depends on this.

/close

Issue #268, opened 2019-01-07
MCD: Not cleaning up files whe removed from MC


ashcrow:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1663376

kikisdeliveryservice:
Should be covered by[ deleteStaleData()](https://github.com/openshift/machine-config-operator/blob/7773bacc6e79829d3325f7747e076e96aebf7a4e/pkg/daemon/update.go#L273) I thought?

kikisdeliveryservice:
So this is what [deleteStaleData()] (https://github.com/openshift/machine-config-operator/blob/7773bacc6e79829d3325f7747e076e96aebf7a4e/pkg/daemon/update.go#L273 ) does after an update is run.  But I don't know that `oc delete` is calling all of our functions like the same as deleting/updating via MCD does?

PR #272, opened 2019-01-08
 daemon: fix deletion of stale config files 

jlebon:
Minor regression from 1d14b2a. We weren't passing the right variable to
`RemoveAll()`.

Closes: #268


cgwalters:
LGTM, though one thing we could also do is warn if a file we expect to `unlink()` is missing?  Since the recent state rework we can rely more on the state we expect to be in.

cgwalters:
That's optional though, if you want to do it please do otherwise I'm good to lgtm this.

jlebon:
Good idea, updated! :arrow_up: 

Though actually, I'm not sure why we don't propagate errors up in this function. We should probably error out on anything that's not `ENOENT`. I can do that in a follow-up.

PR #228, opened 2018-12-11
WIP: Pick last non-empty OSImageURL

cgwalters:
This is part of the "enabling OSImageURL" merges sanely.
Still testing this.

PR #273, opened 2019-01-08
WIP: Add RHCOS oscontainer into payload, render to 00-$role-osimageurl MC

cgwalters:
PR Status:
---

 - [ ] Fix https://github.com/openshift/machine-config-operator/issues/301
 - [ ] Figure out whether or not we need special handling for the *first* master

---

This closes the gap in getting the RHCOS oscontainer into the update
payload, and having the MCO then render that down into the MachineConfigs.
    
Introduce a new ConfigMap `machine-config-osimageurl` that points to
the oscontainer and is applied by the CVO.  Then, the controller
renders that into `machineconfigs/00-$role-osimageurl` which then finally
goes into the rendered config, and should be applied by the daemon.

Closes: #183

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -91,21 +91,16 @@ func generateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1
 	return cfgs, nil
 }
 
+// GenerateMachineConfigsForRole is part of generateMachineConfigs; it operates
+// on a specific role which has a set of builtin templates.
 func GenerateMachineConfigsForRole(config *RenderConfig, role string, path string) ([]*mcfgv1.MachineConfig, error) {
+	cfgs := []*mcfgv1.MachineConfig{}
+
+	// Add our built-in templates
 	infos, err := ioutil.ReadDir(path)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read dir %q: %v", path, err)
 	}
-	// for each role a machine config is created containing the sshauthorized keys to allow for ssh access
-	// ex: role = worker -> machine config "00-worker-ssh" created containing user core and ssh key
-	var tempIgnConfig ignv2_2types.Config
-	tempUser := ignv2_2types.PasswdUser{Name: "core", SSHAuthorizedKeys: []ignv2_2types.SSHAuthorizedKey{ignv2_2types.SSHAuthorizedKey(config.SSHKey)}}
-	tempIgnConfig.Passwd.Users = append(tempIgnConfig.Passwd.Users, tempUser)
-	sshConfigName := "00-" + role + "-ssh"
-	sshMachineConfigForRole := MachineConfigFromIgnConfig(role, sshConfigName, &tempIgnConfig)
-
-	cfgs := []*mcfgv1.MachineConfig{}
-	cfgs = append(cfgs, sshMachineConfigForRole)
 
 	for _, info := range infos {
 		if !info.IsDir() {
@@ -121,9 +116,57 @@ func GenerateMachineConfigsForRole(config *RenderConfig, role string, path strin
 		cfgs = append(cfgs, nameConfig)
 	}
 
+	// And derived configs
+	derivedCfgs, err := generateDerivedMachineConfigs(config, role)
+	if err != nil {
+		return nil, err
+	}
+	cfgs = append(cfgs, derivedCfgs...)
+
 	return cfgs, nil
 }
 
+// machineConfigForOSImageURL generates a MC fragment that just includes the target OSImageURL.
+func machineConfigForOSImageURL(role string, url string) *mcfgv1.MachineConfig {
+	labels := map[string]string{
+		machineConfigRoleLabelKey: role,
+	}
+	return &mcfgv1.MachineConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: labels,
+			Name:   "00-" + role + "-osimageurl",
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			OSImageURL: url,
+		},
+	}
+}
+
+// generateDerivedMachineConfigs is part of generateMachineConfigsForRole. It
+// takes care of generating MachineConfig objects which are derived from other
+// components of the cluster configuration. Currently, that's:
+//
+//  - SSH keys from the install configuration
+//  - OSImageURL from the machine-config-osimageurl configmap (which comes from the CVO)
+func generateDerivedMachineConfigs(config *RenderConfig, role string) ([]*mcfgv1.MachineConfig, error) {
+	cfgs := []*mcfgv1.MachineConfig{}
+
+	// for each role a machine config is created containing the sshauthorized keys to allow for ssh access
+	// ex: role = worker -> machine config "00-worker-ssh" created containing user core and ssh key
+	var tempIgnConfig ignv2_2types.Config
+	tempUser := ignv2_2types.PasswdUser{Name: "core", SSHAuthorizedKeys: []ignv2_2types.SSHAuthorizedKey{ignv2_2types.SSHAuthorizedKey(config.SSHKey)}}
+	tempIgnConfig.Passwd.Users = append(tempIgnConfig.Passwd.Users, tempUser)
+	sshConfigName := "00-" + role + "-ssh"
+	cfgs = append(cfgs, MachineConfigFromIgnConfig(role, sshConfigName, &tempIgnConfig))
+
+	if config.OSImageURL != "" {
+		cfgs = append(cfgs, machineConfigForOSImageURL(role, config.OSImageURL))
+	}
+
+	return cfgs, nil
+}
+
+// generateMachineConfigForName is part of the implementation of generateMachineConfigsForRole
 func generateMachineConfigForName(config *RenderConfig, role, name, path string) (*mcfgv1.MachineConfig, error) {
 	platformDirs := []string{}
 	for _, dir := range []string{"_base", config.Platform} {
--- a/docs/MachineConfigServer.md
+++ b/docs/MachineConfigServer.md
@@ -36,6 +36,10 @@ It performs the following extra actions on the Ignition config defined in the Ma
 
    The new machines that come up, will need a KubeConfig file which will be added as an Ignition file. 
 
+* *Early pivot* - `/etc/rhcos-initial-pivot-target`
+
+    While the Ignition content configures the node, it may actually be booted into an older OS image than is specified by the release payload managed by the [Cluster Version Operator](https://github.com/openshift/cluster-version-operator/).  The MCS writes out the `osImageURL` to this file, and the system will (if necessary) performs an "early pivot" before the node has actually joined the cluster.  This then ensures that when the MachineConfigDaemon starts, it will validate the `currentConfig` (including files written by Ignition and the `osImageURL`).
+
 ### Running MachineConfigServer
 
 It is recommended that the MachineConfigServer is run as a DaemonSet on all `master` machines with the pods running in host network. So machines can access the Ignition endpoint through load balancer setup for control plane.
--- a/pkg/controller/template/test_data/templates/master/00-master/aws/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/master/00-master/aws/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/master/00-master/libvirt/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/master/00-master/libvirt/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/master/00-master/none/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/master/00-master/none/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/master/00-master/openstack/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/master/00-master/openstack/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/worker/00-worker/aws/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/worker/00-worker/aws/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/worker/00-worker/libvirt/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/worker/00-worker/libvirt/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/worker/00-worker/none/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/worker/00-worker/none/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/controller/template/test_data/templates/worker/00-worker/openstack/units/rhcos-initial-pivot.service
+++ b/pkg/controller/template/test_data/templates/worker/00-worker/openstack/units/rhcos-initial-pivot.service
@@ -1,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
+enabled: true
+name: rhcos-initial-pivot.service
--- a/pkg/server/bootstrap_server.go
+++ b/pkg/server/bootstrap_server.go
@@ -96,7 +96,7 @@ func (bsc *bootstrapServer) GetConfig(cr poolRequest) (*ignv2_2types.Config, err
 		return nil, fmt.Errorf("server: could not unmarshal file %s, err: %v", fileName, err)
 	}
 
-	appenders := getAppenders(cr, currConf, bsc.kubeconfigFunc)
+	appenders := getAppenders(cr, currConf, bsc.kubeconfigFunc, "")
 	for _, a := range appenders {
 		if err := a(&mc.Spec.Config); err != nil {
 			return nil, err
--- a/pkg/server/cluster_server.go
+++ b/pkg/server/cluster_server.go
@@ -70,7 +70,7 @@ func (cs *clusterServer) GetConfig(cr poolRequest) (*ignv2_2types.Config, error)
 		return nil, fmt.Errorf("could not fetch config %s, err: %v", currConf, err)
 	}
 
-	appenders := getAppenders(cr, currConf, cs.kubeconfigFunc)
+	appenders := getAppenders(cr, currConf, cs.kubeconfigFunc, mc.Spec.OSImageURL)
 	for _, a := range appenders {
 		if err := a(&mc.Spec.Config); err != nil {
 			return nil, err
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -16,6 +16,9 @@ const (
 	// of the KubeConfig file on the machine.
 	defaultMachineKubeConfPath = "/etc/kubernetes/kubeconfig"
 
+	// rhcosInitialPivotPath is processed by rhcos-initial-pivot.service
+	rhcosInitialPivotPath = "/etc/rhcos-initial-pivot-target"
+
 	// defaultFileSystem defines the default file system to be
 	// used for writing the ignition files created by the
 	// server.
@@ -34,10 +37,11 @@ type Server interface {
 	GetConfig(poolRequest) (*ignv2_2types.Config, error)
 }
 
-func getAppenders(cr poolRequest, currMachineConfig string, f kubeconfigFunc) []appenderFunc {
+func getAppenders(cr poolRequest, currMachineConfig string, f kubeconfigFunc, osimageurl string) []appenderFunc {
 	appenders := []appenderFunc{
 		// append machine annotations file.
 		func(config *ignv2_2types.Config) error { return appendNodeAnnotations(config, currMachineConfig) },
+		func(config *ignv2_2types.Config) error { return appendInitialPivot(config, osimageurl) },
 		// append kubeconfig.
 		func(config *ignv2_2types.Config) error { return appendKubeConfig(config, f) },
 	}
@@ -50,6 +54,15 @@ func appendKubeConfig(conf *ignv2_2types.Config, f kubeconfigFunc) error {
 		return err
 	}
 	appendFileToIgnition(conf, defaultMachineKubeConfPath, string(kcData))
+	return nil
+}
+
+// Ensures that the node is in the OS we expect; for more information see
+// rhcos-initial-pivot.service in the templates
+func appendInitialPivot(conf *ignv2_2types.Config, osimageurl string) error {
+	if osimageurl != "" {
+		appendFileToIgnition(conf, rhcosInitialPivotPath, osimageurl + "\n")
+	}
 	return nil
 }
 
--- a/templates/master/00-master/_base/units/rhcos-initial-pivot.yaml
+++ b/templates/master/00-master/_base/units/rhcos-initial-pivot.yaml
@@ -1,0 +1,19 @@
+name: "rhcos-initial-pivot.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
--- a/templates/worker/00-worker/_base/units/rhcos-initial-pivot.yaml
+++ b/templates/worker/00-worker/_base/units/rhcos-initial-pivot.yaml
@@ -1,0 +1,19 @@
+name: "rhcos-initial-pivot.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=RHCOS initial pivot
+  # Before we join the cluster
+  Before=kubelet.service
+  # But be sure that we've finished Ignition before we reboot
+  After=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/rhcos-initial-pivot-target
+
+  [Service]
+  Type=simple
+  ExecStart=/usr/bin/sh -c 'pivot --reboot $(cat /etc/rhcos-initial-pivot-target) && rm /etc/rhcos-initial-pivot-target'
+  Restart=on-failure
+  RestartSec=30
+
+  [Install]
+  WantedBy=multi-user.target
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -5,6 +5,10 @@ import (
 	"testing"
 )
 
+const (
+	namespace = "openshift-machine-config-operator"
+)
+
 func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
--- a/test/e2e/sanity_test.go
+++ b/test/e2e/sanity_test.go
@@ -10,8 +10,14 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/apimachinery/pkg/labels"
 
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/cmd/common"
 	"github.com/openshift/machine-config-operator/pkg/daemon"
+)
+
+
+var (
+	controllerKind = mcfgv1.SchemeGroupVersion.WithKind("MachineConfigPool")
 )
 
 // Test case for https://github.com/openshift/machine-config-operator/pull/288/commits/44d5c5215b5450fca32806f796b50a3372daddc2
@@ -22,7 +28,7 @@ func TestOperatorLabel(t *testing.T) {
 	}
 	k := cb.KubeClientOrDie("sanity-test")
 
-	d, err := k.AppsV1().DaemonSets("openshift-machine-config-operator").Get("machine-config-daemon", metav1.GetOptions{})
+	d, err := k.AppsV1().DaemonSets(namespace).Get("machine-config-daemon", metav1.GetOptions{})
 	if err != nil {
 		t.Errorf("%#v", err)
 	}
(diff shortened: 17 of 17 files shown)


cgwalters:
The high level idea with this so far (still writing this code) is that we have a ConfigMap that points to the oscontainer and is applied by the CVO.  Then, the controller renders that into `00-$role-osimageurl`.  I also

This rolls in https://github.com/openshift/machine-config-operator/pull/258 and also takes a small bit of https://github.com/openshift/machine-config-operator/pull/228 in that we're taking the first *non-empty* osImageURL.

smarterclayton:
I would recommend pushing to the image stream because it is *exactly* what ART will do, so we can then get the RHCoS image integrated into OCP exactly like origin more quickly.  You can always control how often you push.

install/image-references
```diff
@@ -18,3 +18,7 @@ spec:
     from:
       kind: DockerImage
       name: docker.io/openshift/origin-machine-config-daemon:v4.0.0
+  - name: oscontainer
```
smarterclayton: Naming wise I would probably prefer this to either be `machine-os` or `machine-os-content`, so that it sorts with the machine api.  That means you would publish the tag to `machine-os`.

cgwalters: Everything here is in a Kube namespace though right?  Is there something somewhere which displays/aggregates without namespaces?  (Hm, I guess `oc get X --all-namespaces`?)

But sure, can prefix with `machine-`.  As far as the suffix...I've been using the "oscontainer" term elsewhere but...eh.  This particular string won't be highly visible, I'm OK with `machine-os-content`.

smarterclayton: No, these aren't kube namespaces.  These names are the official "per release payload" names, so think of this as our RPM name.  It will be highly visible in the errata and the oc adm release diff content.  

I don't want to make this RHCoS specific since that means we have to fork origin vs OCP at some point.  If we make it machine-os-content it's clearly owned by the same team that owns machines, and we are unlikely to have 2 or three (or if we do, it would be like channels so `machine-os-content-X`).

smarterclayton: Naming wise this should be named so that someone looking at the release payload can quickly identify what its purpose is (`machine-config-operator`, `etcd`, `cluster-bootstrap`, `cluster-ingress-operator`, etc)

smarterclayton:
Instead of pinning or pushing, you can also just manually tag a new one in, or we can set up a periodic job that grabs the latest maipo, tests it, and then promotes it if the test passes (I could set that up in < 1hr)

cgwalters:
>  or we can set up a periodic job that grabs the latest maipo, tests it, and then promotes it if the test passes (I could set that up in < 1hr)

This one sounds good!  I'd be interested in seeing the code and understanding how that works...there's a lot of magic in the ci-operator/release stuff that I haven't yet fully grasped.  The "test it" part is blocked a bit on this PR landing though right?

cgwalters:
And it turns out we aren't pushing a `:latest` today; we could easily but it's probably better to get the "rhcos oscontainer promotion" going.

Although...maybe we really do need to separate out "add rhcos to payload" from "apply updates"?

Here's a possibility, we could add a config option or so to ignore `osImageURL` changes and flip that on by default in this PR.  That way landing this wouldn't impact development immediately.

smarterclayton:
yes - once you have an image, with a payload, it's really easy to script overlaying that test.  I can go over what would be necessary there.  That would block promoting a new image, but we would still need a job for PRs that overrides the image appropriately (which is not difficult)

cgwalters:
OK this is updated :new: - we now have a `CONFIG_IGNORE_OSIMAGEURL` and use it by default.  This will allow us to land this PR to act as a *mechanism* and then in both local development and CI we can override the `env` setting to test things out.

cgwalters:
Hm:
> error: unable to create a release: operator "machine-config-operator" failed to map images: image file "/tmp/release-image-0.0.1-2019-01-09-161203356873649/machine-config-operator/image-references" referenced image "machine-os-content" that is not part of the input images

Hmm but it worked in https://github.com/openshift/machine-config-operator/pull/258

ashcrow:
Infra failure it looks like. Waiting a bit before retesting.

cgwalters:
> Infra failure it looks like. Waiting a bit before retesting.

Clayton needed to manually create the `machine-os-content` imagestream in api.ci; then the problem is AIUI the [ci operator reusing the namespace](https://github.com/openshift/ci-operator/blob/master/ARCHITECTURE.md).  I logged into api.ci and deleted it.  Let's see...

smarterclayton:
/test e2e-aws

Release payload was built

smarterclayton:
I didn't correctly tag the image in

cgwalters:
> ci/prow/images — Job succeeded. 

Cool.

smarterclayton:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/273/pull-ci-openshift-machine-config-operator-master-e2e-aws/964/artifacts/release-latest/release-payload/0000_50_machine-config-operator_05_osimageurl.yaml

has the correct content

cgwalters:
OK :new: I rewrote this; the main change is the `00-{worker,master}-osimageurl` MCs are gone, the controller always uses the ConfigMap.  See https://github.com/openshift/machine-config-operator/pull/279#issuecomment-452899113

ashcrow:
```
r/pkg/controller/node	0.086s
--- FAIL: TestUpdatesGeneratedMachineConfig (0.07s)
	render_controller_test.go:135: 1 additional expected actions:[{ActionImpl:{Namespace: Verb:update Resource:{Group: Version: Resource:machineconfigs} Subresource:} Object:0xc4203a2a00}]
FAIL
FAIL	github.com/openshift/machine-config-operator/pkg/controller/render	0.095s
```

cgwalters:
Hmm; best as I can tell that unit test was testing changing `osImageURL`, which no longer applies, so I removed it.

cgwalters:
OK, this passes unit tests and e2e-aws, but obviously neither are testing the real functionality here.  I'm going to play around with this some more locally, but it should be ready to review.

jlebon:
Hmm, won't it pick up the new one when the MCC is restarted? (Similar to how the MCO would pick up the new images.json during an update).

cgwalters:
> Hmm, won't it pick up the new one when the MCC is restarted?

Think of the case where an update payload contains just a new oscontainer.  And the other problem is the render controller doesn't understand to regenerate it when it changes.

cgwalters:
One option is to press forward and add an informer/watch for the configmap, and maybe add some sort of "hidden/internal MC" - when the CM changes we update that hidden MC and treat it like any other one.

The other option is to go back to what I had before, creating `00-$role-osimageurl`.  I am leaning towards this for simplicity and symmetry with e.g. the ssh/kubeconfig work.

jlebon:
>  But the downside of this is it brings back the concerns from #279 (comment) around users being able to override it.

Though we can address those concerns with something like in https://github.com/openshift/machine-config-operator/pull/279#issuecomment-452784795.

ashcrow:
>>    The other option is to go back to what I had before, creating 00-$role-osimageurl. I am leaning towards this for simplicity and symmetry with e.g. the ssh/kubeconfig work.

> But we still need to monitor the configmap. I guess the cleanest would be to add a new controller? I am going to proceed down that path unless someone has another idea.

I don't think you'll need another controller so much as making sure that an [informer](https://godoc.org/k8s.io/client-go/informers) is set up to handle reloads when the configmap changes. This could live in the machine-config-controller controller though.

cgwalters:
> This could live in the machine-config-controller controller though.

You mean the render controller?

ashcrow:
>> This could live in the machine-config-controller controller though.

> You mean the render controller?

Ah I follow. It could be within the render controller inside the machine-config-controller, or it could be a new sub controller in the mcc.

cgwalters:
Er, I meant template controller.  Anyways playing with this more, that code is all oriented around deriving from the `controllerconfig` CRD which is updated by the operator.  So I guess (matching the SSH work) the operator should monitor the configmap and we add `osImageURL` to the controllerconfig CRD?

cgwalters:
I am reworking this in https://github.com/cgwalters/machine-config-operator/tree/cvo-integration-2

pkg/daemon/daemon.go
```diff
@@ -142,6 +145,9 @@ func New(
 		return nil, fmt.Errorf("Error establishing connection to logind dbus: %v", err)
 	}
 
+	ignoreOSImageURLVal, ok := os.LookupEnv("CONFIG_IGNORE_OSIMAGEURL")
```
ashcrow: In the future this should also be a switch. Not needed now.

pkg/controller/template/render_test.go
```diff
@@ -330,20 +330,32 @@ func TestGenerateMachineConfigsSSH(t *testing.T) {
...
+			t.Fatal("Failed to find 00-master-ssh")
+		}
+		if workerSsh == nil {
+			t.Fatal("Failed to find 00-worker-ssh")
 		}
 
-		if cfgs[3].Spec.Config.Passwd.Users[0].Name != "core" && cfgs[3].Spec.Config.Passwd.Users[0].SSHAuthorizedKeys[0] != "1234" {
-			t.Fatalf("Failed to create SSH machine config with user core and sshkey 1234, Got: %v", cfgs[3].Spec.Config.Passwd.Users[0])
+		if masterSsh.Spec.Config.Passwd.Users[0].Name != "core" && masterSsh.Spec.Config.Passwd.Users[0].SSHAuthorizedKeys[0] != "1234" {
+			t.Fatalf("Failed to create SSH machine config with user core and sshkey 1234, Got: %v", cfgs[0].Spec.Config.Passwd.Users[0])
```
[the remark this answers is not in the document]
cgwalters: Yeah, and in the interest of DRY I factored it into a helper function :arrow_down:

ashcrow:
```

**Comment by cgwalters** [comment]
OK I decided for now to drop the `CONFIG_IGNORE_OSIMAGEURL` bits for two reasons; first, it was incorrectly placed in the daemon (we'd end up rebooting but *not* updating which is kind of the worst of both worlds), and second - we can test this fully enabled in PR form and see how it goes.

**Comment by cgwalters** [comment]
OK, I apparently logged in to late to the CI cluster; started to gather some information, but looks like it was GC'd.  I got as far as noticing one of the workers went degraded because of an unexpected `osImageURL`.

```
$ oc get machineconfigpool
master    master-f1ef24f88537277975396b4bf8f3f3a1   False     True       False
worker    worker-903aca06ad6e40f505802c891bed42c6   True      False      True
$ oc logs -f pods/machine-config-daemon-f6xlw
I0114 14:06:43.088372    4650 start.go:52] Version: 3.11.0-447-g67d2674b-dirty
I0114 14:06:43.090058    4650 start.go:88] starting node writer
I0114 14:06:43.097838    4650 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0114 14:06:43.138536    4650 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:e59409184a26eb8a8b4cf9d0f419d4986f9dccb863335065b2eae9975f976a5b (47.263)                        
I0114 14:08:04.980087    4650 start.go:139] Calling chroot("/rootfs")
E0114 14:08:14.179060    4650 daemon.go:771] Expected target osImageURL registry.svc.ci.openshift.org/ci-op-jsmj4956/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a               
E0114 14:08:14.179099    4650 daemon.go:346] Fatal error checking initial state of node: Checking initial state: Unexpected on-disk state
E0114 14:08:14.179125    4650 writer.go:90] Marking degraded due to: Checking initial state: Unexpected on-disk state
I0114 14:08:16.433921    4650 daemon.go:348] Entering degraded state; going to sleep
```

**Comment by cgwalters** [comment]
(Maybe it's simpler if rather than generate separate MCs for "base/ssh/osimageurl" we just generate the base one?)

File: pkg/controller/template/render.go [hunk 4]
```diff
@@ -91,21 +91,16 @@ func generateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1
...
+// on a specific role which has a set of builtin templates.
 func GenerateMachineConfigsForRole(config *RenderConfig, role string, path string) ([]*mcfgv1.MachineConfig, error) {
+	cfgs := []*mcfgv1.MachineConfig{}
+
+	// Add our built-in templates
 	infos, err := ioutil.ReadDir(path)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read dir %q: %v", path, err)
 	}
-	// for each role a machine config is created containing the sshauthorized keys to allow for ssh access
```
jlebon: Minor: would be good to show this refactor as a separate initial commit for clarity.

install/image-references
```diff
@@ -18,3 +19,9 @@ spec:
     from:
       kind: DockerImage
       name: docker.io/openshift/origin-machine-config-daemon:v4.0.0
+  # This one is special, it's the OS payload
+  # https://github.com/openshift/machine-config-operator/issues/183
+  - name: machine-os-content
```
jlebon: Do we still actually need this now? We're just using the configmap, right?

install/0000_50_machine-config-operator_05_osimageurl.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: machine-config-osimageurl
+  namespace: openshift-machine-config-operator
+data:
+  # This is a dummy value
```
jlebon: Minor: some more info here would be nice. E.g. "(it will be replaced by the associated osImageUrl for a given release payload)".

jlebon:
So in order to merge this without `CONFIG_IGNORE_OSIMAGEURL`, we'd need to merge just the configmap first as a separate PR, no? (And then teach CI to inject the real value there when assembling the release payload).

install/image-references
cgwalters: The `image-references` file is used by `oc adm release new` (and hence the release operator) to define which images are substituted in the payload.

> (And then teach CI to inject the real value there when assembling the release payload).

This happens today.  The only thing that needed special intervention was Clayton created the `machine-os-content` imagestream.

jlebon: Gotcha. I'm still confused as to how the pullspec goes from imagestream to the configmap? E.g. in this PR, I see no other reference to `machine-os-content` so I'm guessing it's happening elsewhere?

jlebon: To answer my own question, the image references are used in a blanket find & replace throughout all the other manifests. There's nothing else that needs to directly say "replace this configmap value with the one from the imagestream".

pkg/controller/template/render.go
cgwalters: I did that originally but - merge conflict treadmill.

ashcrow:
Monitor and network flakes

/retest

cgwalters:
Rebased with https://github.com/openshift/machine-config-operator/pull/297 so we can more easily debug any failures.

cgwalters:
OK yeah this seems racy on bringup...

```
$ oc logs -f pods/machine-config-daemon-bsrw5                                                                                                                                                                     
I0114 18:17:30.819085    4662 start.go:52] Version: 3.11.0-451-gd399f1c8-dirty
I0114 18:17:30.820430    4662 start.go:88] starting node writer                                                                                                                                                   
I0114 18:17:30.827233    4662 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0114 18:17:30.866924    4662 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:e59409184a26eb8a8b4cf9d0f419d4986f9dccb863335065b2eae9975f976a5b (47.263)                        
I0114 18:17:30.867641    4662 daemon.go:222] Managing node: ip-10-0-131-5.ec2.internal
I0114 18:18:50.907227    4662 start.go:139] Calling chroot("/rootfs")                                                                                                                                             
I0114 18:18:50.934573    4662 daemon.go:392] Current+desired config: worker-ef999f3a84b1506f8d1dd59944d3ae10                                                                                                      
E0114 18:18:50.934596    4662 daemon.go:778] Expected target osImageURL registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a               
E0114 18:18:50.934617    4662 daemon.go:347] Fatal error checking initial state of node: Checking initial state: Unexpected on-disk state                                                                         
E0114 18:18:50.934641    4662 writer.go:90] Marking degraded due to: Checking initial state: Unexpected on-disk state                                                                                             
I0114 18:18:50.954327    4662 daemon.go:349] Entering degraded state; going to sleep
```

Yet a different MCD running on a master has:

```
$ oc logs -f pods/machine-config-daemon-lgm2j                                                                                                                                                                     
I0114 18:16:02.256845   12960 start.go:52] Version: 3.11.0-451-gd399f1c8-dirty
I0114 18:16:02.259074   12960 start.go:88] starting node writer                                                                                                                                                   
I0114 18:16:02.277432   12960 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0114 18:16:02.448678   12960 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:e59409184a26eb8a8b4cf9d0f419d4986f9dccb863335065b2eae9975f976a5b (47.263)                        
I0114 18:16:02.450680   12960 daemon.go:222] Managing node: ip-10-0-26-105.ec2.internal
I0114 18:16:02.530275   12960 start.go:139] Calling chroot("/rootfs")                                                                                                                                             
I0114 18:16:02.702888   12960 daemon.go:904] While getting MachineConfig master-d18bc913fd25af03dd871de78853c28a, got: machineconfigs.machineconfiguration.openshift.io "master-d18bc913fd25af03dd871de78853c28a" n
ot found. Retrying...                
[repeats a lot]
E0114 18:21:02.715867   12960 daemon.go:347] Fatal error checking initial state of node: Checking initial state: timed out waiting for the condition
E0114 18:21:02.715942   12960 writer.go:90] Marking degraded due to: Checking initial state: timed out waiting for the condition                                                                                   
I0114 18:21:02.733198   12960 daemon.go:349] Entering degraded state; going to sleep
```

And that MC is indeed not present.

Generated MCs:

```
$ oc get machineconfigs
NAME                                      GENERATEDBYCONTROLLER        IGNITIONVERSION   OSIMAGEURL                                                                                                                    CREATED
00-master                                 3.11.0-451-gd399f1c8-dirty   2.2.0                                                                                                                                           15m
00-master-osimageurl                      3.11.0-451-gd399f1c8-dirty                     registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
00-master-ssh                             3.11.0-451-gd399f1c8-dirty                                                                                                                                                   15m
00-worker                                 3.11.0-451-gd399f1c8-dirty   2.2.0                                                                                                                                           15m
00-worker-osimageurl                      3.11.0-451-gd399f1c8-dirty                     registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
00-worker-ssh                             3.11.0-451-gd399f1c8-dirty                                                                                                                                                   15m
01-master-kubelet                         3.11.0-451-gd399f1c8-dirty   2.2.0                                                                                                                                           15m
01-worker-kubelet                         3.11.0-451-gd399f1c8-dirty   2.2.0                                                                                                                                           15m
master-140b1092a21c7f1bc27c9569d7021a96   3.11.0-451-gd399f1c8-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
master-7c4c99f286d609e98619e9995c397fe0   3.11.0-451-gd399f1c8-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
worker-b74b86dcf4ee1eb7b9baf328534306fb   3.11.0-451-gd399f1c8-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
worker-ef999f3a84b1506f8d1dd59944d3ae10   3.11.0-451-gd399f1c8-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-f97d2wb8/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   15m
```

I wonder if one thing that's happening here is that adding the new `-osimageurl` MC is exacerbating an extant race condition where we linearly create the `00-worker`, `00-worker-ssh` and now `00-worker-osimageurl`, and in between the render controller reacts and creates an "intermediate" MC.

I am starting to think we really need to only generate e.g. `00-worker` to ensure "atomicity" of our generated configs.

kikisdeliveryservice:
Can you post the `oc logs -p` for the above daemons?

cgwalters:
> Can you post the oc logs -p for the above daemons?

As far as I can see none of the daemons have been restarted.  (I think...`-p` says e.g. `Error from server (BadRequest): previous terminated container "machine-config-daemon" in pod "machine-config-daemon-jnqzq" not found` but I'm not 100% sure nothing was rebooted)

Maybe this really has to do with the bootstrap system?  Maybe the master above that failed to fetch the MC was using the initial one, and since we rapidly churned out new ones it failed.

I think we should be slowing down the render controller a bit in general (e.g. 10-30s pause before rendering to try to avoid races?).

Another idea here; if fetching the `currentConfig` fails, we can skip validation of our existing state.

cgwalters:
If anyone wants to test this PR *before* it lands, you will need to apply some of the changes from it manually.  Specifically you'll need to do at least:

`oc create -f install/0000_50_machine-config-operator_05_osimageurl.yaml`

(This will likely downgrade your cluster if you're running master because that refers to an old oscontainer that will be updated by the release infra)

In addition of course you'll need to `make deploy-controller` and `make deploy-operator`.  (In theory one can do those in parallel, in practice unfortunately at the moment they race).

cgwalters:
> Another idea here; if fetching the currentConfig fails, we can skip validation of our existing state.

But nope can't do that since we need the current config in order to do the "diff" to apply an update.  Conceptually then...no rendered MC should be deleted until all nodes which have it as `currentConfig` are done updating.  Yet today the rendered MCs solely have the `controllerconfig` as an owner.

Two possibilities occur: 

 - Provide the rendered config object locally via Ignition too (ugly, we ~double our ign size?)
 - Introduce an intermediate object...mmm...a `machineconfigrollout` or something that is managed by the operator, and owns the rendered MC, and is only deleted when no nodes reference it?

jlebon:
> I think we should be slowing down the render controller a bit in general (e.g. 10-30s pause before rendering to try to avoid races?).

And maybe something like Jenkins' "quiet period" where it waits a bit after a new MachineConfig is created?

jlebon:
> `I0114 18:16:02.702888   12960 daemon.go:904] While getting MachineConfig master-d18bc913fd25af03dd871de78853c28a, got: machineconfigs.machineconfiguration.openshift.io "master-d18bc913fd25af03dd871de78853c28a" n
ot found. Retrying...`

Hmm, this one looks like something we should fix in the MCD. E.g. do the retry at a higher level, where we actually reread the node annotation for the MC to fetch?

ashcrow:
> Hm so...probably the simplest is for the MCD to treat this as a config difference. On first boot only, if there is a difference between the booted/target osImageURL, do a pivot?

I believe any time there is a different between the current and cluster provided `osImageURL` there should be a pivot. For example, if someone installs and references an old image at install time, but the cluster is set to be a different version, the systems should honor the clusters `osImageURL`.

jlebon:
Or would this work:

```diff
diff --git a/pkg/server/server.go b/pkg/server/server.go
index 4a8ee75..e71d7dc 100644
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -64,7 +64,6 @@ func appendNodeAnnotations(conf *ignv2_2types.Config, currConf string) error {

 func getNodeAnnotation(conf string) (string, error) {
        nodeAnnotations := map[string]string{
-               daemon.CurrentMachineConfigAnnotationKey: conf,
                daemon.DesiredMachineConfigAnnotationKey: conf,
                daemon.MachineConfigDaemonStateAnnotationKey: daemon.MachineConfigDaemonStateDone,
        }
```

Basically, we don't set the current config annotation on startup from the file and let the MCD naturally do it itself if it deems that the system state matches the desired config (and otherwise, e.g. pivot if the osImageURLs don't match).

pkg/daemon/node.go
```diff
@@ -17,6 +17,16 @@ const (
 	// InitialNodeAnnotationsFilePath defines the path at which it will find the node annotations it needs to set on the node once it comes up for the first time.
 	// The Machine Config Server writes the node annotations to this path.
 	InitialNodeAnnotationsFilePath = "/etc/machine-config-daemon/node-annotations.json"
+
+	// InitialConfigRequiredPath is a stamp file that signifies whether or not we have
+	// completed any "firstboot" configuration from our targeted MachienConfig.  Currently
```
jlebon: Minor: MachineConfig (unless you're talking about your dog named Config in French, in which case it's "MachienneConfig" :) )

ashcrow:
```
--- FAIL: TestBootstrapServer (0.00s)
	server_test.go:197: could not find file: root/etc/machine-config-daemon/initial-config-required.stamp
	server_test.go:200: file validation failed for: root/etc/machine-config-daemon/initial-config-required.stamp, exp: {{root <nil> <nil> /etc/machine-config-daemon/initial-config-required.stamp <nil>} {false { data:, {<nil>}} 0xc42003c568}}, got: {{ <nil> <nil>  <nil>} {false {  {<nil>}} <nil>}}
--- FAIL: TestClusterServer (0.00s)
	server_test.go:197: could not find file: root/etc/machine-config-daemon/initial-config-required.stamp
	server_test.go:200: file validation failed for: root/etc/machine-config-daemon/initial-config-required.stamp, exp: {{root <nil> <nil> /etc/machine-config-daemon/initial-config-required.stamp <nil>} {false { data:, {<nil>}} 0xc420422080}}, got: {{ <nil> <nil>  <nil>} {false {  {<nil>}} <nil>}}
```

ashcrow:
Looks like the stamp file needs to be provided for unittests.

ashcrow:
I **think** it's failing due to `getAppenders` in both `_server.go` files being called, and then there being no initial file on the file system.

cgwalters:
OK on this run, seeing the same master machines timing out fetching MCs.

```
$ oc get machineconfigs
NAME                                      GENERATEDBYCONTROLLER        IGNITIONVERSION   OSIMAGEURL                                                                                                                    CREATED
00-master                                 3.11.0-453-gf0070874-dirty   2.2.0                                                                                                                                           3m
00-master-osimageurl                      3.11.0-453-gf0070874-dirty                     registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
00-master-ssh                             3.11.0-453-gf0070874-dirty                                                                                                                                                   3m
00-worker                                 3.11.0-453-gf0070874-dirty   2.2.0                                                                                                                                           3m
00-worker-osimageurl                      3.11.0-453-gf0070874-dirty                     registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
00-worker-ssh                             3.11.0-453-gf0070874-dirty                                                                                                                                                   3m
01-master-kubelet                         3.11.0-453-gf0070874-dirty   2.2.0                                                                                                                                           3m
01-worker-kubelet                         3.11.0-453-gf0070874-dirty   2.2.0                                                                                                                                           3m
master-383ca913310d861fee0be89e6f1d0127   3.11.0-453-gf0070874-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
master-46c05bfb9cb3d4e05608277bb2cb0a5d   3.11.0-453-gf0070874-dirty   2.2.0                                                                                                                                           3m
master-7734c782bad1ead0f8ef5b6affcaf35c   3.11.0-453-gf0070874-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
master-a206c9459a44d859587164a68bb484f2   3.11.0-453-gf0070874-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
worker-f8ffb3b151cd556f0e52db58a59991e8   3.11.0-453-gf0070874-dirty   2.2.0             registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a   3m
```

OK, but the new MCS/MCD code worked fine:

```
$ oc logs -f pods/machine-config-daemon-g6j45 -p
I0114 22:28:05.900642    4685 start.go:52] Version: 3.11.0-453-gf0070874-dirty
I0114 22:28:05.903124    4685 start.go:88] starting node writer
I0114 22:28:05.911505    4685 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0114 22:28:05.951118    4685 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:5e04a144af8106440fb2dd0ac494c562dece239f653134212546da4348adacaf (47.264)                        
I0114 22:28:05.954575    4685 daemon.go:222] Managing node: ip-10-0-152-195.ec2.internal
I0114 22:29:05.968259    4685 node.go:54] Setting initial node config: worker-f8ffb3b151cd556f0e52db58a59991e8                                                                                                    
I0114 22:29:05.980567    4685 start.go:139] Calling chroot("/rootfs")
I0114 22:29:05.999715    4685 daemon.go:392] Current+desired config: worker-f8ffb3b151cd556f0e52db58a59991e8                                                                                                      
I0114 22:29:06.000397    4685 daemon.go:572] Validated on-disk state
I0114 22:29:06.000489    4685 daemon.go:609] Booting into initial configuration; pivot required
I0114 22:29:06.000546    4685 update.go:646] Updating OS to registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a                           
I0114 22:29:06.000597    4685 run.go:13] Running: /bin/pivot registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a                          
pivot version 0.0.2
I0114 22:29:06.163992    5228 run.go:27] Running: rpm-ostree status --json
I0114 22:29:06.200747    5228 root.go:79] Previous pivot: registry.svc.ci.openshift.org/rhcos/maipo@sha256:5e04a144af8106440fb2dd0ac494c562dece239f653134212546da4348adacaf                                       
I0114 22:29:06.200868    5228 run.go:27] Running: skopeo inspect docker://registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a             
I0114 22:29:09.294667    5228 root.go:89] Resolved to: registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a                                
I0114 22:29:09.294780    5228 root.go:106] Pivoting to: 47.231 (b35193d6f2f34c2a38bbbca9e3bbe618cb1266ffa9c38d10050ec8cef2ccba79)                                                         
...
Removed:
  atomic-openshift-clients-4.0.0-0.136.0.git.0.3903800.el7.x86_64
  atomic-openshift-hyperkube-4.0.0-0.136.0.git.0.3903800.el7.x86_64
  atomic-openshift-node-4.0.0-0.136.0.git.0.3903800.el7.x86_64
  pyOpenSSL-0.13.1-4.el7.x86_64
  python-cffi-1.6.0-5.el7.x86_64
  python-enum34-1.0.4-1.el7.noarch
  python-idna-2.4-1.el7.noarch
  python-ply-3.4-11.el7.noarch
  python-pycparser-2.14-1.el7.noarch
  python2-cryptography-1.7.2-2.el7.x86_64
  python2-pyasn1-0.1.9-7.el7.noarch
  python2-pysocks-1.5.7-4.el7.noarch
  python2-urllib3-1.21.1-1.el7.noarch
Added:
  origin-clients-4.0.0-0.alpha.0.802.62de992.x86_64
  origin-hyperkube-4.0.0-0.alpha.0.802.62de992.x86_64
  origin-node-4.0.0-0.alpha.0.802.62de992.x86_64
  python-urllib3-1.10.2-5.el7.noarch
...
I0114 22:30:37.167113    4685 update.go:672] machine-config-daemon initiating reboot: Node will reboot (initial pivot) into config worker-f8ffb3b151cd556f0e52db58a59991e8
$ oc logs -f pods/machine-config-daemon-g6j45
I0114 22:31:33.224130    4570 start.go:52] Version: 3.11.0-453-gf0070874-dirty
I0114 22:31:33.226886    4570 start.go:88] starting node writer
I0114 22:31:33.274734    4570 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0114 22:31:33.415079    4570 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/ci-op-2zl77531/stable@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a (47.231)
I0114 22:31:33.443501    4570 daemon.go:222] Managing node: ip-10-0-152-195.ec2.internal
I0114 22:31:48.244057    4570 start.go:139] Calling chroot("/rootfs")
I0114 22:31:48.262181    4570 daemon.go:392] Current+desired config: worker-f8ffb3b151cd556f0e52db58a59991e8
I0114 22:31:48.264417    4570 daemon.go:572] Validated on-disk state
I0114 22:31:48.278430    4570 daemon.go:599] Completing pending config worker-f8ffb3b151cd556f0e52db58a59991e8
I0114 22:31:48.284692    4570 update.go:672] machine-config-daemon: completed update for config worker-f8ffb3b151cd556f0e52db58a59991e8
I0114 22:31:48.289525    4570 daemon.go:614] Completed initial pivot
I0114 22:31:48.289592    4570 daemon.go:616] Booting into initial configuration; no pivot required
I0114 22:31:48.289674    4570 daemon.go:624] In desired config worker-f8ffb3b151cd556f0e52db58a59991e8
I0114 22:31:48.289753    4570 start.go:158] Starting MachineConfigDaemon
I0114 22:31:48.289804    4570 daemon.go:241] Enabling Kubelet Healthz Monitor
```

Though one confusing thing about this is the `machineconfigpool` is unaware of this state - it flags the workers as being `updated` while the reboots are happening...and in fact, the reboots won't be coordinated.   So we do need to add an annotation or something for this.

cgwalters:
Alternatively...we could do the pivot early on (boot-time systemd unit), before the node joins the cluster.  Positives: less confusion in the MCC, we don't potentially land workloads only to immediately drain them.    Downsides: this bootstrap phase isn't visible in cluster logs directly (e.g. `oc logs -f pods/machine-config-daemon`).  Maybe something else?

jlebon:
Hmm, would it be possible to bring up the node in drained mode and then once we confirm we're in the desired state (possibly after pivoting), we uncordon? That way, the pivot is still in the cluster logs, but we still don't schedule other things that aren't daemonsets.

cgwalters:
Split out prep into https://github.com/openshift/machine-config-operator/pull/305

cgwalters:
> Didn't see any code so that boostrap mco,mcc use the is image url to make sure the configuration is same during special case bootstrapping and long running

Yes, I'm working on that - see https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454190095

cgwalters:
OK still testing this, but here's a new approach.  Includes https://github.com/openshift/machine-config-operator/pull/305 and other patches like https://github.com/openshift/machine-config-operator/pull/303

Implements a new systemd service `rhcos-initial-pivot.service`.

However, one thing I'm realizing is that while this new code should handle "secondary" machines, we may still need further special handling for the initial master (I think this is what [this comment](https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454433465) was getting at).  

Handling that correctly seems like it's going to require extracting the osimageurl during the bootstrap process?  I need to learn more about that code.

cgwalters:
> $ oc create -f install/0000_50_machine-config-operator_05_osimageurl.yaml

Yep, no longer necessary now that https://github.com/openshift/machine-config-operator/pull/305 landed.

kikisdeliveryservice:
@cgwalters Great! I'll pull the new images changes and try to pivot via MC like normal to test.

cgwalters:
Ok I need to `s,Restart=always,Restart=on-failure,` but given that I totally didn't test `rhcos-initial-pivot.service` locally, that appears to have worked just fine in CI!  I was able to see the logs of it working via `rsh` into the MCD and looking at `journalctl -D /rootfs/var/log/journal -u rhcos-initial-pivot`  (which BTW we should do some strategic journal scraping in the MCD I think).

However, one of my masters is still degraded due to MC cycling.  It looks like the MCC just rendered the two new configs, but (I believe Kube GC) forced out the initial one:

```
I0116 20:11:19.962847   13403 node.go:44] Setting initial node config: master-ed135db49be2e34dc2558e6f3f282aa0                                                                                                    
I0116 20:11:20.036232   13403 start.go:139] Calling chroot("/rootfs")                                                                                                                                             
I0116 20:11:20.140343   13403 daemon.go:904] While getting MachineConfig master-ed135db49be2e34dc2558e6f3f282aa0, got: machineconfigs.machineconfiguration.openshift.io "master-ed135db49be2e34dc2558e6f3f282aa0" n
ot found. Retrying... 
```

And there ends my investigation for now as the cluster was just GC'd...

ashcrow:
```
openshift-apiserver OpenShift API is not responding to GET requests
```

cgwalters:
Ah...it looks like you actually did have a successful update, but we fail on the *second* one, note:

```
I0116 20:24:19.402377    5686 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/openshift/origin-v4.0-2019-01-16-135205@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a (47.231)
```

That's the "rewritten" pull spec from the release infra.  The configmap though is saying:
```
I0116 21:21:20.157124    5686 update.go:638] Updating OS to registry.svc.ci.openshift.org/rhcos/maipo@sha256:61dc83d62cfb5054c4c5532bd2478742a0711075ef5151572e63f94babeacc1a
```

Which notice is actually the same sha256, and while the pivot tool passes it through (since the pull specs are different), rpm-ostree barfs on the fact that the ostree commit is identical.

Two things here; I think this scenario is specific to the case where you're trying to manually edit the `osImageURL` to pivot to the exact same commit from the release payload.
Second though, `pivot` should probably canonicalize based on the pull digest.  Or we could teach `rpm-ostree` to make this a no-op (since that's what it is).

kikisdeliveryservice:
Ok trying again with a diff container! Will report back!

ashcrow:
`Terraform` failure flake.

ashcrow:
Restart testing after giving a little bit of time to spin down.

cgwalters:
This PR is held - it's probably OK to just leave it as it won't be merging yet.  We have more issues to work through.

kikisdeliveryservice:
Success! Retested using a diff os container and it worked (using image files up to commit 7999a25 )

`oc logs -p`:
http://pastebin.test.redhat.com/697033

`oc logs -f`:
http://pastebin.test.redhat.com/697034

cgwalters:
OK so https://github.com/openshift/machine-config-operator/issues/301 is definitely happening in some of the CI runs from this PR - but is it actually a *new* issue?  As far as I understand it...`e2e-aws` could pass just fine if a node went degraded (I think).  But let's add a test for that to `e2e-aws-op`.

The other thing ([in the new PR status at the top](https://github.com/openshift/machine-config-operator/pull/273#issue-243082489)) is...actually I think I am wrong, there's no "initial master" versus "secondary masters", the installer creates them all at once.  Which means they should all consistently do the early pivot.

So I think the design here is going to work.

kikisdeliveryservice:
@cgwalters IME with AWS clusters (which is what I'm running everything on), while I see the error syncing messages, my nodes do not degrade bc of them.  Is this perhaps a libvirt issue?

cgwalters:
> while I see the error syncing messages, my nodes do not degrade bc of them.

Yeah; those error messages aren't relevant to the real problem in #301 - I edited the initial message to remove them.  The actual problem in #301 is that if a MC is garbage collected while a node is booting targeting it, it will never finish booting.  And that's what I saw in at least one CI run.

cgwalters:
OK I rebased :surfing_man: this, squashed the two commits into one for clarity, did `s/Restart=always/Restart=on-failure/,` and added a test case.

This is still though waiting for https://github.com/openshift/pivot/pull/25 and the corresponding MCD work, but we can at least run what we have through e2e here.

cgwalters:
```
--- FAIL: TestNoDegraded (0.10s)
	sanity_test.go:79: 3 degraded nodes found
```

Well, that's a kind of progress!  New test successfully failing.

abhinavdahiya:
Why do i see changes to `pkg/server` the server? I'm asking because I see no docs. ;)

Secondly I would live to avoid any more diff that is served by the server from the generated machineconfig contents.

cgwalters:
> Why do i see changes to pkg/server? I'm asking because I see no docs. ;)

It's to handle the difference between the "bootimage" (e.g. AMI) versus the CVO-driven payload.

https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454150059

Will push this in a bit:
```
* *Early pivot* - `/etc/rhcos-initial-pivot-target`

    While the Ignition content configures the node, it may actually be booted into an older OS image than is specified by the release payload managed by the [Cluster Version Operator](https://github.com/openshift/cluster-version-operator/).  The MCS writes out the `osImageURL` to this file, and the system will (if necessary) performs an "early pivot" before the node has actually joined the cluster.  This then ensures that when the MachineConfigDaemon starts, it will validate the `currentConfig` (including files written by Ignition and the `osImageURL`).
```

> Secondly I would live to avoid any more diff that is served by the server from the generated machineconfig contents.

Not sure we'd pass the target data other than writing a file in Ignition?

cgwalters:
> It's to handle the difference between the "bootimage" (e.g. AMI) versus the CVO-driven payload.

And to elaborate, this is now required because of https://github.com/openshift/machine-config-operator/pull/245 - *previously* the MCD would have just said "oh it's a config difference" and updated, but the problem with that approach was we could get into reboot loops and other issues as described in that PR's commit message.

cgwalters:
> Why is getting served as addition by server not through the machineconfig?

You're arguing for having the MCD handle it?  Then we'd need to special case the "initial pivot" there - maybe a good way to do that would be to have "bootstrap node annotations exist" as a flag specifying that.  That has some advantages; it's more consistent.  But one downside as I noted is the [node comes online and then reboots right away](https://github.com/openshift/machine-config-operator/pull/273#issuecomment-455359398).

Does anyone else have a strong opinion on this?

abhinavdahiya:
> > Why is getting served as addition by server not through the machineconfig?
> 
> You're arguing for having the MCD handle it? Then we'd need to special case the "initial pivot" there - maybe a good way to do that would be to have "bootstrap node annotations exist" as a flag specifying that. That has some advantages; it's more consistent. But one downside as I noted is the [node comes online and then reboots right away](https://github.com/openshift/machine-config-operator/pull/273#issuecomment-455359398).
> 
> Does anyone else have a strong opinion on this?

I think i'm confused, any doc detailing the exact flow that this PR is moving towards will help me judge the approach or even provide feedback (if that matters).. currently i am not flowing whats happening  and why :(

cgwalters:
No worries; we've gone through a lot of architecture changes here, and there are now a whole lot of comments on this PR that Github is hiding.  

Did you see https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454150059 ?

To rephrase a bit, if we landed the bit of this PR to add `osImageURL` to the rendered MC without the MCS changes, then if there was any difference between the AMI and the update payload's oscontainer, the MCD would go degraded - because it validates its expected state on boot.

And see https://github.com/openshift/machine-config-operator/pull/273#issuecomment-455359398 again for the semantics of "validate".

Options:

 - "early pivot" as is implemented by this PR right now; MCS provides target osImageURL so we can pivot *before* any cluster workloads land (including the MCD).
 - Change the MCD to special case this by noticing it's the "firstboot case" as defined by "picked up node annotations from `/etc`"

ashcrow:
@abhinavdahiya @cgwalters it may be worth chatting about this in a higher bandwidth system

cgwalters:
This is obsoleted by https://github.com/openshift/machine-config-operator/pull/363

PR #274, opened 2019-01-08
daemon: error out if failing to delete stale data

jlebon:
The presence of an old config file can affect how the system runs.  So
we should treat the inability of deleting such files as an error.

Follow up to #272 (only the last commit is new). Will rebase once that's merged.


jlebon:
Rebased!

PR #275, opened 2019-01-08
manifests: machine config daemon node selector for linux

derekwaynecarr:
Ensure that machine config daemon only runs on linux nodes.

This enables third-parties to explore Windows nodes on OpenShift.


cgwalters:
Aww, I was looking forward to also taking care of updating Windows!

/approve

Issue #276, opened 2019-01-08
retry osimageurl pulls


cgwalters:
So this is a fatal error today:

```
I0108 22:44:07.371178    9935 run.go:27] Running: skopeo inspect docker://registry.svc.ci.openshift.org/rhcos/maipo:latest
time="2019-01-08T22:44:14Z" level=fatal msg="Error reading manifest latest in registry.svc.ci.openshift.org/rhcos/maipo: manifest unknown: manifest unknown"
```

Which makes sense, but AFAICS we'll also go degraded if for some reason we transiently fail to fetch the container.  We should probably teach `pivot` how to distinguish between potentially transient and fatal errors?  (Where here "transient" would be "anything networking" to start)  Or we could just always retry a few times.

ashcrow:
`pivot` does seem like the cleanest place to do this since it's a separate binary. We'd probably want to include k8s's [`util.retry`](https://godoc.org/k8s.io/client-go/util/retry).

The other option would be to use `util.retry` within MCD and look at exit codes. `pivot` would still need a small update to provide a specific exit code for "network failure" or "registry failure" so MCD knew to keep trying up to X times.

ashcrow:
@jlebon my thought was more so that we should retry on specific cases rather than all failures, but I'm OK with always retrying.

PR #278, opened 2019-01-09
daemon/update: retry pivot 5 times on failed rc

jlebon:
In the spirit of retrying all the things that could fail transiently,
just wrap the `/bin/pivot` call in a retry loop.

Closes: #276


jlebon:
Now tested!

cgwalters:
Kind of wonder whether we should move to scheduling a pod on the node instead; that way we go through all the same pull/retry/auth machinery as the rest of the containers.  But this is fine for now.

PR #279, opened 2019-01-09
controller: Take first non-empty osImageURL

cgwalters:
Today each MC will contain both an Ignition fragment and an
`osImageURL`.  Define "merging" as using the first
non-empty `osImageURL` so we don't have to be very picky about
ordering.

This is a smaller version of
https://github.com/openshift/machine-config-operator/pull/228

Prep for: https://github.com/openshift/machine-config-operator/pull/273

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -11,7 +11,7 @@ import (
 // MergeMachineConfigs combines multiple machineconfig objects into one object.
 // It sorts all the configs in increasing order of their name.
 // It uses the Ign config from first object as base and appends all the rest.
-// It only uses the OSImageURL from first object and ignores it from rest.
+// It uses the first non-empty OSImageURL.
 func MergeMachineConfigs(configs []*MachineConfig) *MachineConfig {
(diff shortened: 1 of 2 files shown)


jlebon:
/lgtm

Looking at https://github.com/openshift/machine-config-operator/pull/228#issuecomment-447096711, I'm wondering if it'd make sense to extend it like so: take first non-empty osImageURL, but then also allow overrides from later non-empty osImageURLs and emitting a warning? OTOH, if we go the configmap route in #273, then we'd be able to override there.

crawford:
I said it back in #228, but I'll say it again:

We're going to need to allow the OS to be overridden, but we should make a stink when it happens. Users will inevitably find this and blow their foot off six months down the road when their cluster updates but the OS doesn't. What is the best way to surface that information? Can we propagate it up via the ClusterOperator CR?

cgwalters:
The way I see it is is this is a short term aid just to have slightly less awkward semantics, though in the end it's really optional.  

Following on from #273 we could easily change things so that the rendered MC *always* has the osImageURL from the CVO driven configmap, i.e. ordering of the MC's doesn't matter, and further 
we can then easily implement your requested warning if the rendered MC doesn't match our expected osImageURL in the operator.

PR #282, opened 2019-01-09
daemon: Fix load of booted OS

cgwalters:
We can almost update but go degraded on boot because we had
incorrect variable scoping here :cry:


kikisdeliveryservice:
This seems sane, I'll try to test when i get a working cluster again.

PR #287, opened 2019-01-11
manifests: include pretty column prints for some CRDs

abhinavdahiya:
This makes the lists more informative
```console
$ oc get machineconfigpools
master    master-49e4fc3d90841496575db504d606eb3c   True      False      False
worker    worker-e397e9633d67ced5bcbe3f4067715bff   True      False      False

oc get machineconfigs
NAME                                      GENERATEDBY                  IGNITIONVERSION   OSIMAGEURL
00-master                                 3.11.0-429-g9f6d7d5f-dirty   2.2.0
00-master-ssh                             3.11.0-429-g9f6d7d5f-dirty
00-worker                                 3.11.0-429-g9f6d7d5f-dirty   2.2.0
00-worker-ssh                             3.11.0-429-g9f6d7d5f-dirty
01-master-kubelet                         3.11.0-429-g9f6d7d5f-dirty   2.2.0
01-worker-kubelet                         3.11.0-429-g9f6d7d5f-dirty   2.2.0
master-49e4fc3d90841496575db504d606eb3c   3.11.0-429-g9f6d7d5f-dirty   2.2.0
worker-e397e9633d67ced5bcbe3f4067715bff   3.11.0-429-g9f6d7d5f-dirty   2.2.0
```

/cc @ashcrow @cgwalters

### DIFF
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -4,6 +4,22 @@ metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
   name: machineconfigs.machineconfiguration.openshift.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .metadata.annotations.machineconfiguration\.openshift\.io/generated-by-controller-version
+    description: Version of the controller that generated the machineconfig. This will be empty if the machineconfig is not managed by a controller.
+    name: GeneratedByController
+    type: string
+  - JSONPath: .spec.config.ignition.version
+    description: Version of the Ignition Config defined in the machineconfig.
+    name: IgnitionVersion
+    type: string
+  - JSONPath: .spec.osImageURL
+    description: URL for the RPM OS-tree image. This is optional and can be empty.
+    name: OSImageURL
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Created
+    type: date
   # group name to use for REST API: /apis/<group>/<version>
   group: machineconfiguration.openshift.io
   # list of versions supported by this CustomResourceDefinition
--- a/manifests/machineconfigpool.crd.yaml
+++ b/manifests/machineconfigpool.crd.yaml
@@ -4,6 +4,22 @@ metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
   name: machineconfigpools.machineconfiguration.openshift.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.configuration.name
+    name: Config
+    type: string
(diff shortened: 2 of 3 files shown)


kikisdeliveryservice:
This would have been really  super helpful with my other PRs!!

But please keep the Age column (which seems to have disappeared in your examples?) in the machineconfigpool list because it has been very helpful when creating configs and finding them.

kikisdeliveryservice:
I think NAME - AGE - GENERATEDBY...etc... would be best.

abhinavdahiya:
> But please keep the Age column (which seems to have disappeared in your examples?) in the `oc get machineconfigs` list because it has been very useful when working with all of the configs.

Okay pods have age like column because it matters how long the pod has been running. Do you really think users care about when machineconfig* were created?

IMO their age doesn't provide useful information.
If you need custom info for dev purpose these are better options http://kubernetes.io/docs/user-guide/kubectl-overview/#custom-columns or http://kubernetes.io/docs/user-guide/jsonpath

abhinavdahiya:
> Actually, yes. As a newer user who has applied multiple machine configs for ssh keys, being able to easily see which config was the last applied quickly and easily in the list is extremely helpful. I could definitely see users, especially new users less familiar with applying machine configs and the product/this functionality generally, finding it helpful. The real estate that column takes up is quite small compared to the ones added in this PR.

Sure, done with https://github.com/openshift/machine-config-operator/commit/184884de40792558abfea12af8ea5ed20c282c82

manifests/machineconfig.crd.yaml
```diff
@@ -4,6 +4,22 @@ metadata:
...
+  additionalPrinterColumns:
+  - JSONPath: .metadata.annotations.machineconfiguration\.openshift\.io/generated-by-controller-version
+    description: Version of the controller that generated the machineconfig. This will be empty if the machineconfig is not managed by a controller.
+    name: GeneratedByController
+    type: string
+  - JSONPath: .spec.config.ignition.version
+    description: Version of the Ignition Config defined in the machineconfig.
+    name: IgnitionVersion
+    type: string
+  - JSONPath: .spec.osImageURL
```
jlebon: Minor: I'd put the image URL as the last column since it can get lengthy.

cgwalters:
I also use `oc get machineconfigs` and eyeball `Age` to find "latest" but one could also look at the machineconfigpools...if we're going to make something prettier it feels like that'd be more useful.

But anyways seems reasonable to me, leaving to Kirsten for review though.

/assign kikisdeliveryservice

kikisdeliveryservice:
Let's go with this as is, and we can always revisit the fields in another PR later. :)

PR #289, opened 2019-01-11
daemon: taint node when an ssh access is detected

yuqi-zhang:
Add functionality to taint nodes with `rhcosSsh=accessed:NoSchedule`
when a user SSH's into the machine. This is done via a file watcher
that detects a write to `/var/tmp/ssh-taint`. Also add fsnotify package
and modify nodewriter to facilitate this.

This by itself does not currently affect behaviour on RHCOS nodes,
and is coupled with RHCOS changes to sshd config to write to above
file.

Tested to work with ssh access before the MCD is up, when the MCD
is running, and after removing the taint and re-sshing. The taint only
applies once even if multiple SSH happens.

Note: if the file watcher fails, the behaviour in the goroutine is to log it and
not degrade the node/error out. Maybe that should be changed.

cc/ @ashcrow @cgwalters @jlebon

### DIFF
--- a/pkg/daemon/constants.go
+++ b/pkg/daemon/constants.go
@@ -30,4 +30,9 @@ const (
 	MachineConfigOnceFromRemoteConfig = "REMOTE"
 	// MachineConfigOnceFromLocalConfig denotes that the config was found locally
 	MachineConfigOnceFromLocalConfig = "LOCAL"
+
+	// MachineConfigDaemonSSHTaintKey is the taint key applied when ssh access is detected
+	MachineConfigDaemonSSHTaintKey = "rhcosSsh"
+	// MachineConfigDaemonSSHTaintValue is the taint value applied when ssh access is detecte
+	MachineConfigDaemonSSHTaintValue = "accessed"
 )
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -32,6 +32,7 @@ import (
 	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	fsnotify "gopkg.in/fsnotify.v1"
 )
 
 // Daemon is the dispatch point for the functions of the agent on the
@@ -99,9 +100,10 @@ const (
 	wantsPathSystemd = "/etc/systemd/system/multi-user.target.wants/"
 	// pathDevNull is the systems path to and endless blackhole
 	pathDevNull = "/dev/null"
-
 	// pathStateJSON is where we store temporary state across config changes
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
+	// pathSSHTaint is the runtime file that notifies the MCD to taint the node
+	pathSSHTaint = "/var/tmp/ssh-taint"
 )
 
 const (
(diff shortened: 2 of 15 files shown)


cgwalters:
Overall LGTM!

Per discussion on the other patch though - we could actually watch logind from inside the MCD.  We already have a logind connection:

```
	// login client talks to the systemd-logind service for rebooting the
	// machine
	loginClient *login1.Conn
```

pkg/daemon/daemon.go
```diff
@@ -99,9 +100,10 @@ const (
 	wantsPathSystemd = "/etc/systemd/system/multi-user.target.wants/"
 	// pathDevNull is the systems path to and endless blackhole
 	pathDevNull = "/dev/null"
-
 	// pathStateJSON is where we store temporary state across config changes
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
+	// pathSSHTaint is the runtime file that notifies the MCD to taint the node
+	pathSSHTaint = "/var/tmp/ssh-taint"
```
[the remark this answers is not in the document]
ashcrow: I'm good with moving to `/etc/` .... though let's do it in a follow on PR.

pkg/daemon/constants.go
```diff
@@ -30,4 +30,9 @@ const (
 	MachineConfigOnceFromRemoteConfig = "REMOTE"
 	// MachineConfigOnceFromLocalConfig denotes that the config was found locally
 	MachineConfigOnceFromLocalConfig = "LOCAL"
+
+	// MachineConfigDaemonSSHTaintKey is the taint key applied when ssh access is detected
+	MachineConfigDaemonSSHTaintKey = "rhcosSsh"
+	// MachineConfigDaemonSSHTaintValue is the taint value applied when ssh access is detecte
```
jlebon: detecte**d**

pkg/daemon/writer.go
```diff
@@ -43,7 +44,11 @@ func (nw *NodeWriter) Run(stop <-chan struct{}) {
 		case <-stop:
 			return
 		case msg := <-nw.writer:
-			msg.responseChannel <- setNodeAnnotations(msg.client, msg.node, msg.annos)
+			if msg.annos != nil {
+				msg.responseChannel <- setNodeAnnotations(msg.client, msg.node, msg.annos)
+			} else if msg.taint != nil {
```
jlebon: Minor: I'd just make this another `if` instead of `else if`. Not that the current code currently tries to set both annotations and taints at the same time, but one would expect it'd handle it correctly.

pkg/daemon/daemon.go
```diff
@@ -424,6 +440,75 @@ func (dn *Daemon) getPendingConfig() (string, error) {
...
+		return err
+	}
+
+	return nil
+}
+
+// applySSHTaint calls nodewriter to apply the specified ssh taint
+func (dn *Daemon) applySSHTaint() error {
+	glog.Infof("Detected ssh! The node is being tainted with: %v=%v:%v",
+	MachineConfigDaemonSSHTaintKey, MachineConfigDaemonSSHTaintValue, corev1.TaintEffectNoSchedule)
```
jlebon: Would `%v` of the taint object created just below print the same way?

pkg/daemon/daemon.go
```diff
@@ -424,6 +440,75 @@ func (dn *Daemon) getPendingConfig() (string, error) {
...
+		}
+
+	}
+
+	// Now create (truncate) the file to set up the watcher
+	if _, err := os.Create(pathSSHTaint); err != nil {
+		return err
+	}
+
+	if err := watcher.Add(pathSSHTaint); err != nil {
```
ashcrow: This looks a bit odd to me ... adding the file to watch after the go routine is running. I don't think it's problematic as the go routine will just not have events until this occurs, *but* it may be worth moving as a follow on for readability.

pkg/daemon/daemon.go
```diff
@@ -424,6 +440,75 @@ func (dn *Daemon) getPendingConfig() (string, error) {
...
+func (dn *Daemon) applySSHTaint() error {
+	glog.Infof("Detected ssh! The node is being tainted with: %v=%v:%v",
+	MachineConfigDaemonSSHTaintKey, MachineConfigDaemonSSHTaintValue, corev1.TaintEffectNoSchedule)
+
+	taint := &corev1.Taint {
+		Key: MachineConfigDaemonSSHTaintKey,
+		Value: MachineConfigDaemonSSHTaintValue,
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	return dn.nodeWriter.SetTaint(dn.kubeClient.CoreV1().Nodes(), dn.name, taint)
```
jlebon: Optional: would be nice to also register an event here when tainting to give it higher visibility.

pkg/daemon/daemon.go
yuqi-zhang: Yes, its a string

PR #290, opened 2019-01-11
daemon: fix taint issues

yuqi-zhang:
Based on comment from previous PR:
 - fix typos and add docs
 - de-dupe taint appends
 - change file location for taint

There is still the problem of:
 - How to make the taint key more descriptive (it is currently `rhcosSsh`)
 - that this is breaking tests (will open in further discussions)

cc @abhinavdahiya @ashcrow

### DIFF
--- a/docs/MachineConfigDaemon.md
+++ b/docs/MachineConfigDaemon.md
@@ -14,6 +14,8 @@
 
 MachineConfigDaemon is scheduled on the machines in a cluster as a DaemonSet. This daemon is responsible for performing machine updates in OpenShift 4. The update will include tasks related to the systemd units, files on disk, operating system upgrades etc. The MachineConfigDaemon updates a machine to configuration defined by MachineConfig as instructed by the MachineConfigController.
 
+The MachineConfigDaemon is also responsible for tainting a machine with rhcosSsh=accessed:NoSchedule when it detects an SSH access to the machine.
+
 ## Supported vs Unsupported Ignition config changes
 
 The MachineConfigDaemon receives machine configuration in the form of a "rendered" or merged MachineConfig which is generated from applicable fragments by the controller.
@@ -128,3 +130,8 @@ The draining of pods on the only master node will not evict the control plane as
 ### Node drain etcd static pods on masters
 
 Etcd is co-located on master nodes as static pods. The draining behavior defined above prevents draining of static pods to prevent interference to etcd cluster by the daemon.
+
+## Tainting
+The MachineConfigDaemon is also responsible for tainting the node every time an SSH access is detected. It does this via. watching a file at `/etc/machine-config-daemon/ssh-tainted`, which other scripts can write to to notify the MCD to apply the taint.
+
+The taint can be removed via. `kubectl taint nodes <node> rhcosSsh-`.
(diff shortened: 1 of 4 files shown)


ashcrow:
*Note*: Tests should be transitioned off of ssh. Some are still using ssh, but we'll likely want to work with them to make sure it changes before we enable the tainting in the OS layer.

ashcrow:
*Note*: Tests should be transitioned off of ssh. Some are still using ssh, but we'll likely want to work with them to make sure it changes before we enable the tainting in the OS layer.

ashcrow:
*Note*: Tests should be transitioned off of ssh. Some are still using ssh, but we'll likely want to work with them to make sure it changes before we enable the tainting in the OS layer.

abhinavdahiya:
>     * How to make the taint key more descriptive (it is currently `rhcosSsh`)
WDYT about:

* if this key is supposed to be used to coordinate between components lets use `machineconfiguration.openshift.io/<'-' separated name>`
* if this key is supposed to be used by only one component then lets use `<component>.machineconfiguration.openshift.io/<'-' separated name>`

i think we already kinda follow this..
- example of coordination key https://github.com/openshift/machine-config-operator/blob/3302907bcae4d5fbd438dab80894b410168fe1ac/pkg/daemon/constants.go#L5 and https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/common/constants.go

- example of single component https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L420

docs/MachineConfigDaemon.md
```diff
@@ -128,3 +130,8 @@ The draining of pods on the only master node will not evict the control plane as
 ### Node drain etcd static pods on masters
 
 Etcd is co-located on master nodes as static pods. The draining behavior defined above prevents draining of static pods to prevent interference to etcd cluster by the daemon.
+
+## Tainting
+The MachineConfigDaemon is also responsible for tainting the node every time an SSH access is detected. It does this via. watching a file at `/etc/machine-config-daemon/ssh-tainted`, which other scripts can write to to notify the MCD to apply the taint.
```
kikisdeliveryservice: typo -> "to to"

kikisdeliveryservice: "via." -> "via"

docs/MachineConfigDaemon.md
```diff
@@ -128,3 +130,8 @@ The draining of pods on the only master node will not evict the control plane as
...
+
+The taint can be removed via. `kubectl taint nodes <node> rhcosSsh-`.
```
kikisdeliveryservice: "via." -> "via"

docs/MachineConfigDaemon.md
```diff
@@ -14,6 +14,8 @@
 
 MachineConfigDaemon is scheduled on the machines in a cluster as a DaemonSet. This daemon is responsible for performing machine updates in OpenShift 4. The update will include tasks related to the systemd units, files on disk, operating system upgrades etc. The MachineConfigDaemon updates a machine to configuration defined by MachineConfig as instructed by the MachineConfigController.
 
+The MachineConfigDaemon is also responsible for tainting a machine with rhcosSsh=accessed:NoSchedule when it detects an SSH access to the machine.
```
kikisdeliveryservice: could we add the quotes things like this: `rhcosSsh=accessed:NoSchedule`

yuqi-zhang:
Will fixup PR pending maybe a rebase onto: https://github.com/openshift/machine-config-operator/pull/291

ashcrow:
This probably can be closed.

yuqi-zhang:
/close

Superceded by #359

PR #291, opened 2019-01-11
 daemon: Rework taints to use logind directly

cgwalters:
This way nothing lives in the critical path of ssh to host; which
admins will turn to (ideally only) when something is badly wrong.
It's also just less code and keeps the functionality in one place.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -32,7 +32,6 @@ import (
 	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
-	fsnotify "gopkg.in/fsnotify.v1"
 )
 
 // Daemon is the dispatch point for the functions of the agent on the
@@ -262,20 +261,6 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 
 	go dn.runLoginMonitor(stopCh, dn.exitCh)
 
-	// Set up the file watcher for SSH taint
-	fileWatcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		glog.Errorf("Error creating file watcher: %v", err)
-		return dn.nodeWriter.SetUpdateDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
-	}
-	defer fileWatcher.Close()
-
-	err = dn.setupFileWatcher(fileWatcher)
-	if err != nil {
-		glog.Errorf("Error setting up file watcher: %v", err)
-		return dn.nodeWriter.SetUpdateDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
-	}
-
 	if !cache.WaitForCacheSync(stopCh, dn.nodeListerSynced) {
 		return dn.nodeWriter.SetUpdateDegradedMsgIgnoreErr("failed to sync cache", dn.kubeClient.CoreV1().Nodes(), dn.name)
 	}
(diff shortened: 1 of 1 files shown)


cgwalters:
I tested login sessions but not tainting yet.

/hold

yuqi-zhang:
I agree that this is likely a better path. However we still cannot apply taints at the moment as it would break e2e. I suggest for the time being, have this to write to `motd` to warn the user? (There is the problem of race condition between motd updates and the user seeing it upon login, though)

cgwalters:
(One minor point of trivia: this approach will *also* work if someone logs in on a "local" console; which you'd have to set a password via a MachineConfig to do.  Think bare metal physical console, and most of the cloud providers offer a connection to a virtual console)

cgwalters:
> However we still cannot apply taints at the moment as it would break e2e. 

Ah.  So no matter what this code has to change.

>  (There is the problem of race condition between motd updates and the user seeing it upon login, though)

Yeah.  My vote is we change the default `/etc/motd` to say something like:

```
WARNING: Direct ssh access to machines is not recommended.
In a future version of Red Hat CoreOS, this will cause nodes to become
tainted.
```

cgwalters:
Tested this patch stack and as expected it just results in some logs:

```
I0111 22:06:28.752824   69224 daemon.go:284] Detected a new logged in session: &{:1.2 /org/freedesktop/login1 org.freedesktop.login1.Manager.SessionNew [1 /org/freedesktop/login1/session/_31]}
I0111 22:06:28.752871   69224 daemon.go:285] Login session taints node
I0111 22:06:28.752994   69224 daemon.go:453] NOTE: In a future version of RHCOS this will taint the node
```

cgwalters:
/hold cancel

Although we still aren't decided if this is the way to go.

pkg/daemon/daemon.go
```diff
@@ -282,11 +269,31 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
+		case msg := <-sessionNewCh:
+			glog.Infof("Detected a new logged in session: %v", msg)
+			glog.Infof("Login session taints node")
+			err := replaceFileContentsAtomically(pathSSHTaint, []byte("tainted"))
+			if err != nil {
+				glog.Errorf("Writing taint file: %v", err)
+			}
+			if err := dn.applySSHTaint(); err != nil {
+				glog.Errorf("Error during taint: cannot apply taint due to: %v", err)
+			}
```
jlebon: We'll need to remember to write any errors here to the `exitCh` once we enable this. Might be worth adding a comment about that somewhere in there.

pkg/daemon/daemon.go
```diff
@@ -282,11 +269,31 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
 
+func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
+	sessionNewCh := dn.loginClient.Subscribe("SessionNew")
+	for {
+		select {
+		case <-stopCh:
+			return
+		case msg := <-sessionNewCh:
+			glog.Infof("Detected a new logged in session: %v", msg)
+			glog.Infof("Login session taints node")
```
jlebon: Minor: do we need this line given the logging in `applySSHTaint` already?

yuqi-zhang:
The one behaviour that could be an issue is, if the user ssh's in before the `logind` (or `mcd` for that matter) is ready, he may see an error message in the `motd` but the taint won't actually be applied.

The original case of the file allows for us to check the pre-existence of a write once the `mcd` is ready. That's potentially an edge case for this route.

ashcrow:
/test e2e-aws

Failures look unrelated to this PR.

ashcrow:
`Terraform` failure flakes

ashcrow:
Rebase needed due to taint work being removed.

cgwalters:
/close

Superceded by https://github.com/openshift/machine-config-operator/pull/359

Issue #292, opened 2019-01-12
daemon panic in `updateSSHKeys`


cgwalters:
Was testing OS updates using:

```
$ oc get clusterversion
NAME      VERSION                           AVAILABLE   PROGRESSING   SINCE     STATUS
version   4.0.0-0.alpha-2019-01-12-162217   True        False         46m       Cluster version is 4.0.0-0.alpha-2019-01-12-162217
```

With locally-built operator/controller from https://github.com/openshift/machine-config-operator/pull/273
But daemon from the update payload, i.e.:
```
machine-config-daemon                         https://github.com/openshift/machine-config-operator                       e1d1ea0df81beb1fce3335ea2df3f7b8096e7bd7
```

```
$ oc logs -f pods/machine-config-daemon-ckfh4 -p
...
E0112 18:54:53.223756   16793 runtime.go:66] Observed a panic: "index out of range" (runtime error: index out of range) 
/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72                                         
/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65                
/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51                                                                                       
/usr/local/go/src/runtime/asm_amd64.s:573                                                                              
/usr/local/go/src/runtime/panic.go:502                                                                           
/usr/local/go/src/runtime/panic.go:28                                                                                        
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/update.go:594                                        
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/update.go:96                                           
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:761                                                                             
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:702                                          
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:714                                                       
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:660                                         
/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:227                                          
/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/client-go/tools/cache/controller.go:202                   
[...]

PR #293, opened 2019-01-13
daemon: fix updateSSHKeys for MC with Users unset/empty

kikisdeliveryservice:
Add logic to updateSSH/reconcilable to account for machineconfigs where Passwd:Users is completely empty/unset.

Closes Issue: #292

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -186,13 +186,7 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) *stri
 		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
 			// check if the prior config is empty and that this is the first time running.
 			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
-			if len(oldIgn.Passwd.Users) == 0 && len(newIgn.Passwd.Users) == 1 {
-				glog.Infof("user data sent to verify before ssh init: %v", newIgn.Passwd.Users[0])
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
@cgwalters This should be the fix, testing it on a cluster now.

kikisdeliveryservice:
Tested with machineconfigs with empty/unset passwds where: `passwd: {}` , `passwd: ` and there is no `passwd` line at all in the config. 

This should fix the bug  :smile:

pkg/daemon/update.go
```diff
@@ -589,32 +583,33 @@ func getFileOwnership(file ignv2_2types.File) (int, int, error) {
...
-		return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(sshDirPath), err)
-	}
-	glog.V(2).Infof("Created directory: %s", sshDirPath)
-
-	authkeypath := filepath.Join(sshDirPath, "authorized_keys")
-	var concatSSHKeys string
-	for _, k := range newUsers[len(newUsers)-1].SSHAuthorizedKeys {
-		concatSSHKeys = concatSSHKeys + string(k) + "\n"
-	}
+	if len(newUsers) >= 1 {
```
cgwalters: LGTM, though (optional) I would tend to do e.g.:

```
if len(newUsers) == 0 {
    return nil
}
```

To avoid re-indenting, and also on the general principle of getting the special cases out of the way in a function first, leaving the main body of the function to handle the general case.

kikisdeliveryservice: will do!

kikisdeliveryservice: Done. Thanks for this comment!

PR #295, opened 2019-01-14
manifests/machineconfig: Print osImageURL last

cgwalters:
Since it's long, and the `Age` column is interesting and we don't
want to obscure it.

Ref: https://github.com/openshift/machine-config-operator/pull/287#discussion_r247128766


ashcrow:
This *DOES* look good to me, but I want to double check we are :+1: to merge into master.

PR #297, opened 2019-01-14
Two MCD logging patches

cgwalters:


cgwalters:
Since this bounced I added another logging patch.

pkg/daemon/node.go
```diff
@@ -40,6 +41,7 @@ func loadNodeAnnotations(client corev1.NodeInterface, node string) error {
 		return fmt.Errorf("Failed to unmarshal initial annotations: %v", err)
 	}
 
+	glog.Infof("Setting initial node config: %s", initial[CurrentMachineConfigAnnotationKey])
```
ashcrow: Note: If the annotation isn't set for some reason this will output empty.

ashcrow:
Unrelated failures

/retest

Issue #257, opened 2018-12-20
add more to test suite


cgwalters:
It should be pretty easy to put together a quick test suite in e.g. bash that given a running cluster, tests at least a nondestructive scenario of creating a MC, waiting for it to roll out, running a quick daemonset that verifies the files exist, then e.g. deletes it and verifies the nodes are back as they were.

We may be able to enhance non-destructive testing by creating a separate machineset too and having MCs being more destructive just to those nodes (e.g. making them degraded etc.)

kikisdeliveryservice:
I'm supposed to work on some e2e testing after my other PRs are done, so count me in! :D

abhinavdahiya:
@smarterclayton whats the best location to add operator level tests ? That require a cluster.

smarterclayton:
See the e2e-aws-operator jobs in openshift/release.  Basically have a binary / script in your repo, copy their pattern, then you get passed a kubeconfig and can run.

cgwalters:
One thing related to https://github.com/openshift/machine-config-operator/issues/294 is that I think it would be an extremely useful pattern for us to support spinning up a "testnode" machine type that is marked tainted; we could then use it as a target for e.g. upgrade and other testing.

cgwalters:
I'm just going to repurpose this issue now for "let's add more to the test suite" since it's not covering much at all.

I think we can add a test case for e.g. "add a MC that adds a file to workers" easily now.

It'd also be good to extend that and script a reconciliation failure (`rsh` into the mcd and make a conflicting file?).

Testing SSH key updates should be pretty straightforward I think.

But most interesting (for me) is testing OS updates, to do that nicely a pattern like https://github.com/openshift/machine-config-operator/issues/257#issuecomment-454585735 would help.

cgwalters:
This issue is basically done, we can track anything else via PRs.

PR #298, opened 2019-01-14
hack/test-e2e.sh: New test suite stub

cgwalters:
Ref: https://github.com/openshift/machine-config-operator/issues/257

Add a basic test file, targeting a test case for
https://github.com/openshift/machine-config-operator/pull/288/commits/44d5c5215b5450fca32806f796b50a3372daddc2
since it's easy.

Next step is to try hooking this up to Prow.

(I'm a bit undecided about bash as a framework; looks like other
 projects are using Go, but eh...if we decide to do something else
 we can easily exec it, and anyways this is just to get the ball
 rolling)

### DIFF
--- a/hack/test-e2e.sh
+++ b/hack/test-e2e.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+#
+# This is the "e2e-aws-operator" on github; see github.com/openshift/release
+# which contains the Prow glue to invoke it.  Add in tests here specific
+# to the machine-config-operator.
+#
+# Currently, all tests here are non-destructive.  They assume the environment
+# has a KUBECONFIG with administrative privileges.
+
+set -xeuo pipefail
+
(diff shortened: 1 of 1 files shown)


cgwalters:
release job PR: https://github.com/openshift/release/pull/2577

jlebon:
Looks sane to me. Will let others take a look.
/approve

abhinavdahiya:
I don't think we should move towards any bash based testing..

we already have Makefile to trigger the e2e test (which seems pretty easy looking at [this](https://github.com/openshift/cluster-image-registry-operator/blob/e6c7544903d2826d171be45cdde5a83a58d97383/Makefile#L23))

Go test based testing is what we should aim for and the setup looks easy too [here](https://github.com/openshift/cluster-image-registry-operator/blob/e6c7544903d2826d171be45cdde5a83a58d97383/test/e2e/main_test.go)

Issue #301, opened 2019-01-14
MachineConfigs can be garbage collected while a node is still booting


cgwalters:
Splitting this out of https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454111408
https://github.com/openshift/machine-config-operator/pull/273#issuecomment-454134115

Here are relevant mcc logs from a cluster spun up by that PR:

```
I0114 22:26:10.482441       1 render_controller.go:437] Generated machineconfig master-46c05bfb9cb3d4e05608277bb2cb0a5d from 1 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  }]
I0114 22:26:11.589272       1 render_controller.go:437] Generated machineconfig master-7734c782bad1ead0f8ef5b6affcaf35c from 2 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  } {Machin
eConfig  00-master-osimageurl  machineconfiguration.openshift.io/v1  }]
I0114 22:26:12.484212       1 render_controller.go:437] Generated machineconfig master-383ca913310d861fee0be89e6f1d0127 from 3 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  } {Machin
eConfig  00-master-osimageurl  machineconfiguration.openshift.io/v1  } {MachineConfig  00-master-ssh  machineconfiguration.openshift.io/v1  }]
I0114 22:26:14.483810       1 render_controller.go:437] Generated machineconfig master-a206c9459a44d859587164a68bb484f2 from 4 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  } {Machin
eConfig  00-master-osimageurl  machineconfiguration.openshift.io/v1  } {MachineConfig  00-master-ssh  machineconfiguration.openshift.io/v1  } {MachineConfig  01-master-kubelet  machineconfiguration.openshift.io/
v1  }]
```

You can see that this caused very fast churn in the machineconfigs, and the previous ones were GCd.  But - there were secondary masters that were still booting and expecting to be able to find that MC.
This is a tricky problem - we need to have a way to avoid pruning "in flight" MCs passed from the MCC to Ignition.

cgwalters:
This problem is actually fairly new (as of basically since Friday) because of the near-simultaneous addition of `-ssh` and `-kubelet` MCs which causes MC regeneration churn, and I'm currently going to add another.

kikisdeliveryservice:
[...]

PR #303, opened 2019-01-15
controller: Add a 5s delay before rendering MCs

cgwalters:
To reduce churn if MCs are being created rapidly - both on general
principle, and also to reduce our exposure to the current bug
that a booting node may fail to find a GC'd MachineConfig:
https://github.com/openshift/machine-config-operator/issues/301

### DIFF
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -36,6 +36,10 @@ const (
 	//
 	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
 	maxRetries = 15
+
+	// renderDelay is a pause to avoid churn in MachineConfigs; see
+	// https://github.com/openshift/machine-config-operator/issues/301
+	renderDelay = 5 * time.Second
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ashcrow:
TF failures

/retest

jlebon:
Hmm, if I'm reading this right, even if we translate three MC creation events by 5s, we're still regenerating three times, right? Though I guess they'll hash to the same name now at least, so we won't get a generated MC quickly appearing then getting deleted.

LGTM though will defer to folks more familiar with the workqueue API.

/approve

cgwalters:
> we're still regenerating three times, right?

See https://godoc.org/k8s.io/client-go/util/workqueue
specifically:

>  and if an item is added multiple times before it can be processed, it will only be processed once.

ashcrow:
tf failures again... looks like somethings up. Let's retest shortly.

jlebon:
OK, after reading up some more on the workqueue API, I'm more confident this works now. I've also just tested it!

/lgtm

Re.

> and if an item is added multiple times before it can be processed, it will only be processed once.

Right, I see that at https://github.com/kubernetes/client-go/blob/b831b8de7155117e51afaffeb647007a756ddc92/util/workqueue/queue.go#L114. But this happens at `Add()` time, and `AddAfter()` just delays the `Add()` call: https://github.com/kubernetes/client-go/blob/b831b8de7155117e51afaffeb647007a756ddc92/util/workqueue/delaying_queue.go#L162. So `Add()` is still called e.g. three times. So if the previous work item has already been processed, we'll just sync again. One test I did to verify this was `oc create -f hello1.yaml  && sleep 3.5 && oc create -f hello2.yaml`.

Anyway, this still mitigates the issue fine since the MCs we're concerned about definitely happen within a 5s window. So even on the earliest event fired, we've already got all the MCs. Though of course closing the race window completely will still require some work.

PR #305, opened 2019-01-15
Add osimageurl to release payload and controllerconfig

cgwalters:
[ And also the SSH test patch ]

Split out of #273

This is the first part of closing the gap in getting the RHCOS oscontainer
into the update payload.

Introduce a new ConfigMap machine-config-osimageurl that points to
the oscontainer and is owned by the CVO.  Then, the operator propagates
that into the `controllerconfig` CRD.

Further patches will have the controller react to the update, but having
this first step in will help us validate the model without actually
having clusters update.


cgwalters:
> Mounting the configmap vs syncing?

I did that originally, but how do you suggest monitoring for changes?

> Operator reading the OS image url and passing to controller using config vs passing pointed to config in controller config.

Can you flesh this out a bit?

kikisdeliveryservice:
As a suggestion I think that we should refrain from making unrelated commits (like changing the SSH tests in osimageurl work) in PRs like this going forward bc they are hard to see/review and get lost in the shuffle.

cgwalters:
> I think that we should refrain from making unrelated commits 

I generally agree, however I'm trying to batch at least some PRs to increase the chances they make it through the e2e-aws roulette.

It *is* a clearly separate commit, and this is already a prep PR split from the bigger one.  But if you still prefer we can separate it out.

ashcrow:
I agree with @kikisdeliveryservice's sentiment. For the time being to help move through CI I think it's OK to have PRs with unrelated commits but, when we do that, let's have the PR name reference all the changes so it's searchable.

cgwalters:
OK so we don't have a whole lot of time left to implement OS updates.  We can redesign it...review is good, but at the moment it's just me actively hacking on it, and right now to work on it one needs to redeploy the whole stack (operator, controller, server, daemon).  It'd make it significantly more convenient for others to hack on if we landed these operator pieces now.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -61,6 +61,9 @@ type MCOConfigSpec struct {
 	BaseDomain string `json:"baseDomain"`
 
 	SSHKey string `json:"sshKey"`
+
+	// Sourced from configmap/machine-config-osimageurl
+	OSImageURL string `json:"osImageURL`
```
jlebon: Doesn't seem like we need this, right?

ashcrow: @jlebon as in the MC no longer needs to house an `OSImageURL`?

cgwalters: > Doesn't seem like we need this, right?

Yep, you're right; the SSH key comes from the installconfig, but the `OSImageURL` doesn't.

> @jlebon as in the MC no longer needs to house an OSImageURL?

This is the `mcoconfig` object.

PR #306, opened 2019-01-15
docs: include cvo scaling in directions for updating ssh keys

kikisdeliveryservice:
Just an extra line to make the directions more complete.

### DIFF
--- a/docs/Update-SSHKeys.md
+++ b/docs/Update-SSHKeys.md
@@ -49,6 +49,10 @@ spec:
  - Updating `user: name`: Do not update the `user: name` field. The only user currently supported is `core` as shown in the above example config.
 
  ## Applying the MachineConfig
+First scale the Cluster Version Operator:
+```sh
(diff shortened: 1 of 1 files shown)


docs/Update-SSHKeys.md
```diff
@@ -49,6 +49,10 @@ spec:
  - Updating `user: name`: Do not update the `user: name` field. The only user currently supported is `core` as shown in the above example config.
 
  ## Applying the MachineConfig
+First scale the Cluster Version Operator:
+```sh
+    oc -n openshift-cluster-version scale --replicas=0 deployment/cluster-version-operator
```
yuqi-zhang: Is this for testing purposes only? In actual cluster operation we would need the CVO still for updates, etc. no?

kikisdeliveryservice: Ah good call.  I'll leave it out!

Thanks for the quick reply, @yuqi-zhang !

PR #308, opened 2019-01-15
controller/tests: Avoid hardcoding SSH MC offsets

cgwalters:
In prep for adding more.

### DIFF
--- a/pkg/controller/template/render_test.go
+++ b/pkg/controller/template/render_test.go
@@ -330,21 +330,33 @@ func TestGenerateMachineConfigsSSH(t *testing.T) {
 			t.Fatalf("failed to generate machine configs: %v", err)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/template/render_test.go
```diff
@@ -316,6 +316,12 @@ func TestGenerateMachineConfigs(t *testing.T) {
 	}
 }
 
+func validateSSHConfig(t *testing.T, mc *mcfgv1.MachineConfig) {
```
kikisdeliveryservice: I would move this down with the other helper functions in the bottom of file to stay consistent with file conventions.

kikisdeliveryservice:
Thanks @cgwalters  Is this prep for osImageUrl tests?

ashcrow:
It's now erroring trying to make/reference release images. I'll kick it off again shortly.

PR #311, opened 2019-01-15
Drop path checking from Makefile

darkmuggle:
The use of variable my_p is spurious and breaks in symlinked environments.  Since the use of $my_p is spurious, drop it and use relative paths.


cgwalters:
Specifically, this breaks my "GOPATH hack" of `ln -s github github.com` in my source directory.

PR #312, opened 2019-01-15
daemon: skip querying currentConfig if possible

jlebon:
The same way we pass the specific desiredConfig we "locked on" when
updating, also similarly pass the currentConfig. This is a minor
optimization so we avoid querying the cluster again for information we
already have, but it does also make the code more resilient to humans
meddling with node annotations.

Requires: #310

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -598,7 +598,7 @@ func (dn *Daemon) CheckStateOnBoot() error {
 	} else {
 		// currentConfig != desiredConfig, and we're not booting up into the desiredConfig.
 		// Kick off an update.
-		if err := dn.triggerUpdateWithMachineConfig(state.desiredConfig); err != nil {
+		if err := dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig); err != nil {
 			return err
 		}
 	}
@@ -706,7 +706,7 @@ func (dn *Daemon) prepUpdateFromCluster() (bool, error) {
 // you want both pulled from the cluster please use executeUpdateFromCluster().
 func (dn *Daemon) executeUpdateFromClusterWithMachineConfig(desiredConfig *mcfgv1.MachineConfig) error {
 	// The desired machine config has changed, trigger update
-	if err := dn.triggerUpdateWithMachineConfig(desiredConfig); err != nil {
+	if err := dn.triggerUpdateWithMachineConfig(nil, desiredConfig); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


cgwalters:
I was looking at a similar patch recently, but I was thinking more like

```
$ git diff
index 5309697..c78133a 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -598,7 +598,7 @@ func (dn *Daemon) CheckStateOnBoot() error {
        } else {
                // currentConfig != desiredConfig, and we're not booting up into the desiredConfig.
                // Kick off an update.
-               if err := dn.triggerUpdateWithMachineConfig(stateAndConfigs.desiredConfig); err != nil {
+               if err := dn.update(stateAndConfigs.currentConfig, stateAndConfigs.desiredConfig); err != nil {
                        return err
                }
        }
diff --git a/pkg/daemon/update.go b/pkg/daemon/update.go
index 158f876..4e78920 100644
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -69,6 +69,12 @@ func (dn *Daemon) writePendingState(desiredConfig *mcfgv1.MachineConfig) error {
 func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
        var err error
 
+       if dn.onceFrom != "" && !ValidPath(dn.onceFrom) {
+               if err = dn.nodeWriter.SetUpdateWorking(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
+                       return err
+               }
+       }
+
        oldConfigName := oldConfig.GetName()
        newConfigName := newConfig.GetName()
        glog.Infof("Checking reconcilable for config %v to %v", oldConfigName, newConfigName)
```

cgwalters:
There's all these layers of "update/triggerUpdate*" functions that get into the "onceFrom" stuff and I get a bit confused by them...but the lowest level seems to be `update()` and so let's move the important bit of setting the node annotation there and have the main cluster path call it?

jlebon:
Yeah, I think that'd work. Though it's a bit more involved than I wanted to go given the current freeze. Whereas the patch as written is pretty easy to reason through. Let's enhance it later? Your diff did make me notice something else though. I pushed another minor commit for that! :arrow_up:

ashcrow:
Infra flake

/retest

ashcrow:
Terraform flake

/retest

ashcrow:
Monitor flake:

```
net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

/retest

cgwalters:
/lgtm

This will help a bit for #273

ashcrow:
`Terraform` errors with AWS. Will force a test once the clean up occurs.

PR #313, opened 2019-01-15
 Makefile: Add `test-unit`, only test pkg/ lib/ and cmd/

cgwalters:
This matches what cluster-image-registry-operator is doing.
Prep for #309


cgwalters:
https://github.com/openshift/release/pull/2606

Issue #125, opened 2018-10-11
MCD: Actually handle directories and symlinks


jlebon:
Right now, we're only handling the `files` list of the `storage` section. We should support `directories` and `links` too.

jlebon:
I started on this, but honestly I'm wondering if maybe we should discuss https://github.com/coreos/ignition/issues/599 some more first. There's already some duplication of Ignition code and there's opportunity for more sharing if we want to support directories and symlinks.

The main difference is that Ignition only has an "update" mode whereas the MCD has both a "check" and an "update" mode. Although a "check" mode might be of interest to Ignition as well to be used as a final pass to confirm file states match the config?

ashcrow:
Having overlap is expected. Though I am all for the ability to share some functionality so we don't have to redevelop all required features as time goes on.

Did you find a use case that requires directories and links already? Or is this more so anticipating (correctly) that eventually we will need to support them?

ashcrow:
In that case I agree, implementing (or updating) similar logic in `ignition` in a way that will let us import and reuse is a better use of time as we don't have a need ASAP.

jlebon:
Hmm, should we discuss whether we want this in 4.0? It's basic functionality that one would expect we'd support.

dustymabe:
> Hmm, should we discuss whether we want this in 4.0? It's basic functionality that one would expect we'd support.

I would like to be able to lay down symbolic links as part of the work I'm doing for kmods-via-containers. I agree that directories and symbolic links are items that users would expect to work.

dustymabe:
bump

runcom:
does anybody remember why we didn't add support for directories and symlinks when the MCO was created? I vaguely remember some issues (esp around symlinks) but other than that I'm under the impression that we just didn't add a reconcile for dirs/symlinks just because there wasn't a need?

ashcrow:
I believe there wasn't a need and the previous work which the MCO was built upon didn't have dirs/links implemented either.

larsks:
@jlebon what's the status of this issue? I keep running into situations in which I just want to create a directory, but instead I have to find some weird workaround (e.g., right now I'm installing a bunch files into `/etc/mco` as a staging location, because I couldn't create a more appropriate directory).

jlebon:
@larsks It's still outstanding, but eventually will be fixed by https://github.com/openshift/enhancements/pull/1032.

sjentzsch:
[...]

PR #315, opened 2019-01-15
Provide a writable location for flexvolume plugins

gnufied:
Fixes https://jira.coreos.com/browse/STOR-157

This parameter can not be configured via `KubeletConfiguration` because of  https://github.com/kubernetes/kubernetes/pull/72937

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -24,6 +24,7 @@ contents: |
         --minimum-container-ttl-duration=6m0s \
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider={{cloudProvider .}} \
+        --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
         {{.CloudProviderConfig -}} \
         --anonymous-auth=false \
         --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
--- a/pkg/controller/template/test_data/README.md
+++ b/pkg/controller/template/test_data/README.md
@@ -1,0 +1,7 @@
+## Updating test data after making changes to original
+
+test data templates can be updated via:
+
+```
+go test ./pkg/controller/template/... -u
+```
--- a/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
+++ b/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
@@ -1,0 +1,6 @@
+contents:
+  source: data:,
+  verification: {}
+filesystem: root
+mode: 493
+path: /etc/kubernetes/kubelet-plugins/volume/exec/.dummy
--- a/pkg/controller/template/test_data/templates/master/00-master/libvirt/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
+++ b/pkg/controller/template/test_data/templates/master/00-master/libvirt/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
@@ -1,0 +1,6 @@
+contents:
+  source: data:,
+  verification: {}
+filesystem: root
+mode: 493
+path: /etc/kubernetes/kubelet-plugins/volume/exec/.dummy
--- a/pkg/controller/template/test_data/templates/master/00-master/none/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
+++ b/pkg/controller/template/test_data/templates/master/00-master/none/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
@@ -1,0 +1,6 @@
+contents:
+  source: data:,
+  verification: {}
+filesystem: root
+mode: 493
+path: /etc/kubernetes/kubelet-plugins/volume/exec/.dummy
--- a/pkg/controller/template/test_data/templates/master/00-master/openstack/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
+++ b/pkg/controller/template/test_data/templates/master/00-master/openstack/files/-etc-kubernetes-kubelet-plugins-volume-exec-.dummy
@@ -1,0 +1,6 @@
+contents:
+  source: data:,
+  verification: {}
+filesystem: root
+mode: 493
+path: /etc/kubernetes/kubelet-plugins/volume/exec/.dummy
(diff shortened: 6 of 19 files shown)


gnufied:
corresponding change will also be needed to controller-manager fwiw.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
```diff
@@ -24,6 +24,7 @@ contents: |
         --minimum-container-ttl-duration=6m0s \
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider={{cloudProvider .}} \
+        --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
```
cgwalters: This directory doesn't seem to exist on the masters today; is something else going to create it?  Does the kubelet ignore it if it's `ENOENT`?

gnufied:
@cgwalters yep. I am going to push that stuff. I had that updated yesterday night, I forgot to push.

ashcrow:
Unit tests look like they need updating. The end results no longer look like what was expected.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
gnufied: fixed.

gnufied:
@cgwalters @ashcrow can you PTAL

pkg/daemon/daemon.go
```diff
@@ -22,6 +22,7 @@ import (
 	mcfgclientv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
+	fsnotify "gopkg.in/fsnotify.v1"
```
cgwalters: Generally it'd be nice to split up "my editor is running a gofmt-like tool" type commits to be first.

pkg/daemon/update.go
```diff
@@ -84,6 +84,11 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 		return fmt.Errorf("%s", msg)
 	}
 
+	// Create direcotry path for flexvolume plugin
+	if err = dn.fileSystemClient.MkdirAll(pathFlexVolumePlugin, DefaultDirectoryPermissions); err != nil {
```
cgwalters: I don't think we want this as Go code; it should just be an Ignition config like any others.  Specifically an entry in `directories`; see [the Ignition spec](https://coreos.com/ignition/docs/latest/configuration-v2_2.html). 

IOW you'd want this to be a file in `templates` like how we have the `kubelet.yaml` there.

*However* - looking at `render.go`, we'd need to extend it with e.g. `directories`.

cgwalters:
Before we proceed farther - does this change have a 4.0 freeze exception?

ashcrow:
@cgwalters I did not see it upon quick scan of the list.

@gnufied can you confirm that this does or does not have the exception approval?

gnufied:
@ashcrow @cgwalters this is a bug fix, not a new feature. Does it require 4.0 freeze exception? Adding @childsb @knobunc @tsmetana

childsb:
@gnufied  bugfixes shouldn't require an exception, but you should open a bugzilla to help support that its a bug.

pkg/daemon/update.go
gnufied: okay, do you want me do that as part of this PR? I am kinda new to this code base, I am still trying to grok how Ignition spec works.

pkg/daemon/daemon.go
gnufied: ack. Yeah it did stick out a little. I will drop this hunk from commit.

ashcrow:
OK, if this is indeed a bugfix it doesn't need an exception.

/hold cancel

pkg/daemon/daemon.go
```diff
@@ -505,11 +508,11 @@ func (dn *Daemon) setupFileWatcher(watcher *fsnotify.Watcher) error {
 // applySSHTaint calls nodewriter to apply the specified ssh taint
 func (dn *Daemon) applySSHTaint() error {
 	glog.Infof("Detected ssh! The node is being tainted with: %v=%v:%v",
-	MachineConfigDaemonSSHTaintKey, MachineConfigDaemonSSHTaintValue, corev1.TaintEffectNoSchedule)
+		MachineConfigDaemonSSHTaintKey, MachineConfigDaemonSSHTaintValue, corev1.TaintEffectNoSchedule)
```
ashcrow: Looks like `gofmt`-like tool in this area too. No biggie, but may be worth dropping as well for now.

pkg/daemon/update.go
ashcrow: No worries @gnufied! 

If the Ignition config comes through the install process Ignition will do exactly what is needed (creating the directories, dropping the file, etc..).

If it comes down through MCD then MCD itself has a subset of Ignition implementation to create files/dirs/etc.. I think this would make more sense within [this area](https://github.com/openshift/machine-config-operator/blob/635cb799b02fda6b182334e2f019fb3ae19fa01b/pkg/daemon/update.go#L271) and/or [here](https://github.com/openshift/machine-config-operator/blob/635cb799b02fda6b182334e2f019fb3ae19fa01b/pkg/daemon/update.go#L500).

kikisdeliveryservice: Is the goal here to just create the dir (that is always the same dir location) and never do anything else with it in the MCD?

ashcrow: For clarity, templates is fine. If you need to do some kind of operations then the areas I pointed out would be where to add logic.

gnufied: Goal is just to create the directory. We don't need to do anything else in MCD.

kikisdeliveryservice: I agree with @cgwalters that templates makes more sense.

cgwalters: Based on https://github.com/coreos/ignition/pull/706 you could probably "cheat" and make a `.dummy` file there or so (assuming kubelet won't try to load a `.dummy` or whatever file as a plugin).

gnufied: Yeah I think that would work. Kubelet seems to skip loading files that begin with `.` as plugins.

jlebon: Related: #125

ashcrow:
`Terraform`/AWS flake

/test e2e-aws-op

kikisdeliveryservice:
I'm going to pull this to verify. :)

cgwalters:
/assign kikisdeliveryservice

Will let her do the final lgtm.

PR #317, opened 2019-01-16
WIP: Test breaking e2e

cgwalters:
:warning: Don't merge :warning: 

Let's test the new test e2e-aws-op suite that should appear since https://github.com/openshift/release/pull/2577

### DIFF
--- a/manifests/machineconfigdaemon/daemonset.yaml
+++ b/manifests/machineconfigdaemon/daemonset.yaml
@@ -48,8 +48,6 @@ spec:
         - key: node-role.kubernetes.io/etcd
           operator: Exists
           effect: NoSchedule
-      nodeSelector:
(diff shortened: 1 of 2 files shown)


ashcrow:
```
=== RUN   TestOperatorLabel
--- FAIL: TestOperatorLabel (0.10s)
	sanity_test.go:26: Expected node selector 'linux', not ''
```

cgwalters:
Yep, cool!  So now we should feel confident in extending our test suite as we make further changes.

/close

cgwalters:
(One interesting aspect of this is that our `e2e-aws-op` is basically testing an installer run; if it succeeds but `e2e-aws` fails it's more likely a flake somewhere in the origin e2e suite, not installation/cluster bringup)

PR #318, opened 2019-01-16
 controller/render: fix deleting ownerReference

jlebon:
We weren't actually deleting the ownerReference because CRDs don't yet
support strategic merges[1], which have the fancy `$patch: delete`
syntax. We were also specifying the wrong PatchType argument.

For now, just drop down to the vanilla JSONPatch[2] to do this.

[1] https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch
[2] https://tools.ietf.org/html/rfc6902

Related: #301

### DIFF
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -503,6 +503,8 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 			}
 			// Let's not make it fatal for now
 			glog.Warningf("Failed to delete ownerReference from %s: %v", gmc.Name, err)
+		} else {
+			glog.Infof("Deleted ownerReference from %s", gmc.Name)
 		}
 	}
 
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -470,7 +470,26 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 			continue
 		}
 
-		deleteOwnerRefPatch := fmt.Sprintf(`{"metadata":{"ownerReferences":[{"$patch":"delete","uid":"%s"}],"uid":"%s"}}`, pool.UID, gmc.UID)
+		// we should be the only owners of the MC, but just in case, check the index
+		idx := -1
+		for i, owner := range gmc.OwnerReferences {
+			if owner.UID == pool.UID {
+				idx = i
+				break
+			}
+		}
+
+		if idx == -1 {
+			// We don't have a ref on this object, done!
+			continue
+		}
+
+		// XXX: There is a race here if another ownerReference gets added
+		// (though they usually get appended so idx would still be valid), but
+		// that shouldn't happen since we should be the only owners.  Once CRDs
+		// support strategic merges, we can use the `$patch: delete` syntax.
+		glog.V(2).Infof("Attempting to delete ownerReference from %s", gmc.Name)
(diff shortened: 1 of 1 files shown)


ashcrow:
Is this for https://github.com/openshift/machine-config-operator/issues/301#issuecomment-454910427?

jlebon:
Yup exactly! This is not strictly required in the immediate future. (I thought it'd be since I was going to make use of ownerReferences in order to fix #301, but I don't think that'll actually work). So we don't have to merge this right away.

ashcrow:
`Terraform` failure flakes.

cgwalters:
OK the first patch here looks like a fairly obvious fix; and may actually explain a failure I've seen a few times where adding a MC generates only the worker but not master pool or vice versa.

Which is clearly required for upgrades and also just a general fix for MCs.

Let's split that into a separate PR?

jlebon:
Yeah, that one should be straightforward & low-risk. Split that into #321!

> and may actually explain a failure I've seen a few times where adding a MC generates only the worker but not master pool or vice versa.

Hmm, I'm not sure. The only effect this fix should have is that the render controller should now keep trying to remove the pool's ownerReference from more MCs, instead of stopping at the first error. Since it's in the last thing it does in that function, the `return` wasn't actually skipping over anything else.

cgwalters:
Are you *sure* it was broken?  I'm testing
```
diff --git a/pkg/controller/render/render_controller.go b/pkg/controller/render/render_controller.go
index 719dcf9..69db28f 100644
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -486,6 +486,8 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 			}
 			// Let's not make it fatal for now
 			glog.Warningf("Failed to delete ownerReference from %s: %v", gmc.Name, err)
+		} else {
+			glog.Infof("Queued for GC: %s", gmc.Name)
 		}
 	}

```

right now.  We know *something* is doing a GC, and if it's not this...then I don't know.

jlebon:
> Are you *sure* it was broken?

Pretty sure. You can see this by just omitting the second commit in this pull. You'll see an error like: `Failed to delete ownerReference from 00-master: json: cannot unmarshal object into Go value of type jsonpatch.Patch`. I.e. we were trying to POST a strategic merge patch as a JSONPatch. The second commit here fixes this by actually using a valid JSONPatch since strategic merge patches AFAIU aren't supported for CRDs yet.

cgwalters:
Looking at the MCC logs in [https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/318/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/13](the e2e-aws-op run), I don't see any errors about failing to patch the MC.  So presumably with this patch we are now GC'ing them reliably.

(Did you test this locally?  Does it cause MCs to be reliably GC'd?)

But now the problem is the lifetime will be too short.

cgwalters:
@jlebon can you cherry-pick https://github.com/openshift/machine-config-operator/pull/324/commits/527ed16f5b04b62f9c7587baa3e3aa0630be111b into this PR?

I'd also say that we shouldn't merge this PR until we are confident about a strategy for GC.

jlebon:
Done! Though I tweaked the wording a bit since we're still testing the idea of whether deleting the ownerReference should cause a GC.

cgwalters:
Doing https://github.com/openshift/machine-config-operator/pull/347 first.

ashcrow:
#347 merged.

kikisdeliveryservice:
In an effort to clean up the MCO repo, closing old open PRs with no activity.  

Feel free to reopen.

PR #319, opened 2019-01-16
WIP: test/e2e: Validate that no nodes went degraded

cgwalters:
I think we have this in CI but we're not noticing it.
If it's happening we need to fix it.

Ref: https://github.com/openshift/machine-config-operator/issues/301

### DIFF
--- a/test/e2e/sanity_test.go
+++ b/test/e2e/sanity_test.go
@@ -2,10 +2,14 @@ package e2e_test
 
 import (
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/machine-config-operator/cmd/common"
+	"github.com/openshift/machine-config-operator/pkg/daemon"
 )
 
 // Test case for https://github.com/openshift/machine-config-operator/pull/288/commits/44d5c5215b5450fca32806f796b50a3372daddc2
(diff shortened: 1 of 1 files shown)


cgwalters:
(Tested this locally by forcing a node degrade and the test failed as expected)

cgwalters:
Hm, looks like the cluster in `e2e-aws-op` here lost a node; 
```
$ oc get pods --all-namespaces | grep NodeLost
openshift-cluster-network-operator           cluster-network-operator-qsztm                                  1/1       NodeLost    2          1h
openshift-cluster-node-tuning-operator       tuned-b4pq6                                                     1/1       NodeLost    0          1h
openshift-dns                                dns-default-gn9jz                                               2/2       NodeLost    0          1h
openshift-image-registry                     node-ca-4x9jw                                                   1/1       NodeLost    0          1h
openshift-machine-config-operator            machine-config-daemon-6t8tn                                     1/1       NodeLost    0          1h
openshift-machine-config-operator            machine-config-server-xt7r4                                     1/1       NodeLost    0          1h
openshift-sdn                                ovs-qvcsd                                                       1/1       NodeLost    0          1h
openshift-sdn                                sdn-controller-mprjl                                            1/1       NodeLost    0          1h
openshift-sdn                                sdn-jg9hv                                                       1/1       NodeLost    0          1h
```

And it looks like the previous MCC was there, so we don't have logs from it.  Ah, and the `master` pool is updating:

```
$ oc get machineconfigpool
master    master-73ad249a9f5189df75bec7acb767cc01   False     True       False
worker    worker-ad3f99702752cb7f13d04e3d1b7202df   True      False      False
```

And...the cluster got GC'd, damn.

ashcrow:
`Failed to connect events watcher` ... `connection refused`

Flake.

cgwalters:
(Any reason no one is `lgtm` this?)

ashcrow:
More `Terraform` + aws flakes.

cgwalters:
Cool:

```
=== RUN   TestOperatorLabel
--- PASS: TestOperatorLabel (0.18s)
=== RUN   TestNoDegraded
--- PASS: TestNoDegraded (0.11s)
PASS
ok      github.com/openshift/machine-config-operator/test/e2e   0.327s
```

test/e2e/sanity_test.go
```diff
@@ -26,3 +32,44 @@ func TestOperatorLabel(t *testing.T) {
...
 }
+
+func TestNoDegraded(t *testing.T) {
+	cb, err := common.NewClientBuilder("")
+	if err != nil{
+		t.Errorf("%#v", err)
+	}
+	k := cb.KubeClientOrDie("sanity-test")
+
+	kubeSharedInformer := informers.NewSharedInformerFactory(k, 20 * time.Minute)
```
abhinavdahiya: Hmm.. listers and informers are useful when controllers want events and cached responses. I think using the client to directly list seems more reasonable here.

test/e2e/sanity_test.go
```diff
@@ -26,3 +32,44 @@ func TestOperatorLabel(t *testing.T) {
 		t.Errorf("Expected node selector 'linux', not '%s'", osSelector)
 	}
 }
+
+func TestNoDegraded(t *testing.T) {
```
abhinavdahiya: This test should check that the daemon has done its sync atleast once before checking if it degraded.

cgwalters: Hmm...like wait for the `state=Done` annotation on the node?

ashcrow: That would work if looking at it from the end of the run. You could also watch to see `Working` and know that it sync'd.

cgwalters:
No point in landing this until we fix https://github.com/openshift/machine-config-operator/issues/367

(Though a PR to fix that could roll in this one)

ashcrow:
```
=== RUN   TestNoDegraded
--- FAIL: TestNoDegraded (0.06s)
	sanity_test.go:67: 3 degraded nodes found
FAIL
FAIL	github.com/openshift/machine-config-operator/test/e2e	0.370s
```

runcom:
with https://github.com/openshift/machine-config-operator/pull/386 do we still need a test like this?

cgwalters:
Yeah, I think this one is obsoleted.

PR #321, opened 2019-01-17
controller/render: fix ownerRef deletion error-handling

jlebon:
We were immediately returning from the function when ignoring an error
instead of moving on to the next MachineConfig in the list. Also, we
were completely ignoring errors we *shouldn't* ignore. For now, I've
just made this a warning rather than a hard failure pending discussions
re. garbage collection.


ashcrow:
`Terraform` flakes.

/retest

PR #322, opened 2019-01-17
*: include 127.0.0.1 and localhost in etcd peer SAN

hexfusion:
The etcd peer SAN does not include localhost/127.0.0.1 this is a problem if you would like to use this cert as a client from say inside the container.

As the peer is the only cert with `TLS Web Client Authentication` we really need to include this.

peer
```
 X509v3 Extended Key Usage: 
                TLS Web Client Authentication, TLS Web Server Authentication

X509v3 Subject Alternative Name: 
                DNS:sbatsche-etcd-0.devcluster.openshift.com, DNS:sbatsche.devcluster.openshift.com, IP Address:10.0.15.187

```
server
```
X509v3 Extended Key Usage: 
              TLS Web Server Authentication

X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:etcd.kube-system.svc, DNS:etcd.kube-system.svc.cluster.local, DNS:sbatsche-etcd-0.devcluster.openshift.com, IP Address:10.0.15.187, IP Address:127.0.0.1
```
Separate issue also I am seeing what appears to be a client attempting to use the server cert
```
2019-01-17T13:10:11.150313341+00:00 stderr F 2019-01-17 13:10:11.150277 I | embed: rejected connection from "127.0.0.1:37686" (error "tls: failed to verify client's certificate: x509: certificate specifies an incompatible key usage", ServerName "")
```

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -331,6 +331,7 @@ func etcdPeerCertDNSNames(cfg RenderConfig) (interface{}, error) {
 	}
 
 	var dnsNames = []string{
+		"localhost",
 		"${ETCD_DNS_NAME}",
 		fmt.Sprintf("%s.%s", cfg.ClusterName, cfg.BaseDomain), // https://github.com/etcd-io/etcd/blob/583763261f1c843e07c1bf7fea5fb4cfb684fe87/Documentation/op-guide/clustering.md#dns-discovery
 	}
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -55,7 +55,7 @@ contents:
                 --assetsdir=/etc/ssl/etcd \
                 --dnsnames={{etcdPeerCertDNSNames .}} \
                 --commonname=system:etcd-peer:${ETCD_DNS_NAME} \
-                --ipaddrs=${ETCD_IPV4_ADDRESS} \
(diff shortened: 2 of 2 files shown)


cgwalters:
And yet another PR trips over [the need to replicate every config change into the renderer unit tests](https://github.com/openshift/machine-config-operator/pull/315#issuecomment-454799008).

Anyways as far as the code; I'm not an expert in this - found [this page](https://stackoverflow.com/questions/21197331/x-509-certificate-is-it-a-good-idea-or-bad-to-add-localhost-in-subject-altern) in a quick Google search.

Could use more details on this bit:

> this is a problem if you would like to use this cert as a client from say inside the container.

And specifically I think we need an argument this is a 4.0 exception.

abhinavdahiya:
> And yet another PR trips over [the need to replicate every config change into the renderer unit tests](https://github.com/openshift/machine-config-operator/pull/315#issuecomment-454799008).


`go ./pkg/controller/template/... -u`

hexfusion:
> > And yet another PR trips over [the need to replicate every config change into the renderer unit tests](https://github.com/openshift/machine-config-operator/pull/315#issuecomment-454799008).
> 
> `go ./pkg/controller/template/... -u`

Aha I will keep this in mind from now on. running some additional tests to make sure this is needed.

hexfusion:
I think I jumped the gun on this going to close and review further, thanks for the input.

cgwalters:
https://github.com/openshift/machine-config-operator/pull/322#issuecomment-455237380 was confusing me...I thought there was a program to run, but the command is actually:

`$ go test ./pkg/controller/template/... -u`

PR #323, opened 2019-01-17
kubeletconfig: add blacklisted options

rphillips:
Certain fields within the KubeletConfiguration should not be settable by the user. This code change checks against a blacklist of readonly fields.


pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -57,6 +58,14 @@ var updateBackoff = wait.Backoff{
 	Jitter:   1.0,
 }
 
+var blacklistKubeletConfigurationFields = []string{
```
abhinavdahiya: nit: some documentation around why might come in handy here or in docs.

rphillips: great idea

rphillips: Added

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -57,6 +58,20 @@ var updateBackoff = wait.Backoff{
 	Jitter:   1.0,
 }
 
+// A list of fields a user cannot set within the KubeletConfig CR. If a user
+// were to set these values, then the system may become unrecoverable (ie: not
+// recover after a reboot).
+//
+// If the KubeletConfig CR instance contains a non-zero or non-empty value for
+// the following fields, then the MCC will not apply the CR and log the message.
```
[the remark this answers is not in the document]
rphillips: Another great idea. I'll add that in.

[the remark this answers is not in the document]
abhinavdahiya: Thanks! :100:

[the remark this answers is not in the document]
abhinavdahiya: > Thanks! 100

For adding the status reporting :innocent:

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -404,6 +404,22 @@ type KubeletConfigSpec struct {
 
 // KubeletConfigStatus defines the observed state of a KubeletConfig
 type KubeletConfigStatus struct {
+	// Represents the latest available observations of current state.
```
abhinavdahiya: `ObservedGeneration` might come in handy

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -404,6 +404,22 @@ type KubeletConfigSpec struct {
...
+	Conditions []KubeletConfigCondition `json:"conditions"`
+}
+
+// KubeletConfigCondition defines the state of the KubeletConfig
+type KubeletConfigCondition struct {
```
abhinavdahiya: keeping status object standard will be easy. good example is here https://github.com/openshift/api/blob/4d49b5754a7067c0bf325b67ac6fc5e7d307aaad/config/v1/types_cluster_operator.go#L82-L98

abhinavdahiya: Also functions like these might also come in handy :)

rphillips: I cleaned this up in 750d707... Should be good for another review.

PR #324, opened 2019-01-18
WIP: Render osImageURL, handle "bootstrap" case in MCD 

cgwalters:
This is option 2 for https://github.com/openshift/machine-config-operator/pull/273

Still a WIP, it builds but haven't tested updates with it.

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -91,21 +91,16 @@ func generateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1
 	return cfgs, nil
 }
 
+// GenerateMachineConfigsForRole is part of generateMachineConfigs; it operates
+// on a specific role which has a set of builtin templates.
 func GenerateMachineConfigsForRole(config *RenderConfig, role string, path string) ([]*mcfgv1.MachineConfig, error) {
+	cfgs := []*mcfgv1.MachineConfig{}
+
+	// Add our built-in templates
 	infos, err := ioutil.ReadDir(path)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read dir %q: %v", path, err)
 	}
-	// for each role a machine config is created containing the sshauthorized keys to allow for ssh access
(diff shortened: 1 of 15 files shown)


pkg/controller/template/render.go
```diff
@@ -91,21 +91,16 @@ func generateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1
 	return cfgs, nil
 }
 
+// GenerateMachineConfigsForRole is part of generateMachineConfigs; it operates
```
ashcrow: Thank you for adding the documentation!

ashcrow:
Terraform/AWS flake **and**
```
level=fatal msg="waiting for openshift-console URL: context deadline exceeded"
```

/retest

cgwalters:
OK rolled in https://github.com/openshift/machine-config-operator/pull/329 to this and also added a hacky patch to give us some debugging data when our e2e fails.

ashcrow:
ISE trying to see logs

/retest

ashcrow:
```
http2: server sent GOAWAY and closed the connection; LastStreamID=11, ErrCode=NO_ERROR, debug=""
error: watch closed before Until timeout
```

/retest

cgwalters:
OK so here's the problem I'm struggling with...the "bootstrap MC" and the "inital rendered MC" need to be the same.  If we land [the installer change](https://github.com/openshift/installer/pull/1149) without this PR, or vice versa - things will break because booted nodes won't be able to find the MC they expect.

That was the core problem that brought this PR to a halt.

We need to figure out how to "ratchet" the PRs to both the installer and here.

What I'm vaguely thinking about is that we somehow detect whether or not the installer set up the osimageurl.  Maybe stuff something in the `install-config-v1` configmap?  If it didn't, then we also don't render in the non-bootstrap path?

cgwalters:
Splitting this PR into:

https://github.com/openshift/machine-config-operator/pull/362
https://github.com/openshift/machine-config-operator/pull/363

PR #328, opened 2019-01-18
daemon: be honest about our ability to reconcile dirs & links

jlebon:
Otherwise we'll end up in a situation where someone thinks they're
supported and we won't error out and also not create them.

Related: #125


jlebon:
Note this is technically a bugfix, not a new feature.

PR #330, opened 2019-01-18
Add ContainerRuntime CRD and Controller

umohnani8:
Add a new CRD and Controller that allows users to configure
certain options in crio.conf and storage.conf. A template is
used for the default values that can be changed with a CR. When
the CR is deleted, the values are reverted back to its defaults.
The following options can be configured in /etc/crio/crio.conf
- pidsLimit
- logLevel
- logSizeMax
- infraImage

The following options can be configured in /etc/containers/storage.conf:
- overlaySize


umohnani8:
@mrunalp @sjenning an overview of the design for the crd and controller.

kikisdeliveryservice:
@umohnani8 Thanks for starting some work on docs! 

Could you retitle this PR "docs: Add ContainerRuntime.." so people don't worry there was some other changes in the PR like I did? :)

abhinavdahiya:
/cc @rphillips 
as he added the kubelet config controller and this is going to be pretty identical to that.

mrunalp:
@kikisdeliveryservice We are planning to add all the code to this PR before it merges :) @umohnani8  started out with the docs  initially to make sure design is fine before we start writing code.

rphillips:
@umohnani8 @sjenning and I are considering putting this into the KubeletConfig CR. Do you think this could fit into the KubeletConfig controller within the MCO?

mrunalp:
Did something change since we last discussed this on Thursday? 🙂

rphillips:
I doubt it changed :smiley:. I just didn't hear about it.

mrunalp:
Okay 😃 we decided to make this separate from the kubelet cr so it isn't too tied to Linux.

rphillips:
After talking with Clayton yesterday [[issue]](https://github.com/openshift/machine-config-operator/pull/341#issuecomment-456683991), the crio.conf pause image needs to be customized as well with the quay.io/openshift/origin-pod:v4.0-[cvo versioned pod image]. Seems like the pause image could be rolled into this design/code-change as well.

/cc @smarterclayton @sjenning

mrunalp:
Yeah, I discussed with @umohnani8 yesterday; we will include the pod infra image as part of this work.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -0,0 +1,502 @@
...
+		go wait.Until(ctrl.worker, time.Second, stopCh)
+	}
+
+	<-stopCh
+}
+
+func (ctrl *Controller) updateContainerRuntimeConfig(oldObj interface{}, newObj interface{}) {
+	oldKC := oldObj.(*mcfgv1.ContainerRuntimeConfig)
+	newKC := newObj.(*mcfgv1.ContainerRuntimeConfig)
+	if oldKC != newKC {
```
runcom: all of this seems to still refers to `KC` or kubeconfig right?

umohnani8: Yup, thanks! I missed it

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -0,0 +1,459 @@
...
+	mtmpl "github.com/openshift/machine-config-operator/pkg/controller/template"
+	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
+	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/scheme"
+	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
+	mcfglistersv1 "github.com/openshift/machine-config-operator/pkg/generated/listers/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/version"
+)
+
+const (
+	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
```
jlebon: Minor: s/machineconfig pool/containerruntime config/

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
+# Summary
+
+Users need a way to update the container runtime configuration. Users can do this via MCO currently, but they need to know the correct options to use. We can hopefully help customers by documenting a CRD with the ContainerRuntime tuneables and have the MCO use these tuneables when rendering the crio.conf, storage.conf, and registries.conf files to Ignition/disk.
+
+The CRI-O runtime is used by default and is the recommended solution.
+
+# Motivation
+
+The MCO contains the logic to write the crio.conf, storage.conf, and registries.conf configuration files to Ignition. When Ignition starts on a machine it writes these three files to configure the container runtime. When these resources change within the MachineConfig, the new configs are rewritten and the MachineConfigDaemon is instructed to reboot the machine to pick up the new configs.
```
jlebon: Minor: well, the MCD is not explicitly "instructed" (I think this is what #360 is suggesting). The MCD could in the future become smarter and not reboot for things it knows don't require a reboot.

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+5. configuring list of blocked registries
+
+6. configuring max size of storage driver
+
+## Non-Goals
+
+# Proposal
+
+Extend the Machine Config Operator to include a ContainerRuntimeConfig CRD and ContainerRuntimeConfigController. By using a ContainerRuntimeConfig CRD there is an implicit whitelist of user controlled options for the container runtime. Upon deleting the ContainerRuntimeConfig instance the default config is restored*.
```
jlebon: Looks like a stray `*` here? Or did this refer to a footnote before?

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Validate the user defined ContainerRuntimeConfig
+
+2. Render the current MachineConfigs (storage.files.contents[crio.conf, storage.conf, registries.conf]) into the ContainerRuntimeConfiguration structure
+
+3. Load the ContainerRuntimeConfig from the passed in Spec.ContainerRuntimeConfig
+
+4. Serialize the ContainerRuntimeConfig to yaml
+
+5. Create or Update the ignition /etc/containers/storage.conf, /etc/containers/registries.conf, and /etc/crio/crio.conf files within a 99-[role]-kubelet-managed MachineConfig
```
jlebon: Hmm, why 99 here? This just has to be > 00 to override the base templates, right? The connotation with "99" is that no MC should override these files but we want to have that flexibility, right? (Of course, one could always have a "999", but that doesn't look as nice :)).

umohnani8: Right, it just has to be > 01 to override the base templates. Saw 99 being used, so just followed that. I can change it to 02.

rphillips: I chose 99- for a managed MachineConfig. The default crio.conf should probably be at 01- to match the kubelet conf default machineconfig.

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+2. setting log level
+
+3. configuring list of search registries
+
+4. configuring list of insecure registries
+
+5. configuring list of blocked registries
+
+6. configuring max size of storage driver
```
mrunalp: We need to add infra_image to this list.

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. setting pids limit
+
+2. setting log level
+
+3. configuring list of search registries
+
+4. configuring list of insecure registries
+
+5. configuring list of blocked registries
```
smarterclayton: How does this overlap with the existing global config fields for this?

docs/ContainerRuntimeConfigDesign.md
smarterclayton: This seems like it is more per machine set, is that correct?

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Motivation
+
+The MCO contains the logic to write the crio.conf, storage.conf, and registries.conf configuration files to Ignition. When Ignition starts on a machine it writes these three files to configure the container runtime. When these resources change within the MachineConfig, the new configs are rewritten and the MachineConfigDaemon is instructed to reboot the machine to pick up the new configs.
+
+## Goals
+
+The following goals should not require explicit knowledge of a customer to know the right command-line or configuration options to use within the ContainerRuntime.
+
+1. setting pids limit
```
smarterclayton: What happens if this becomes a kube level setting in the future?

docs/ContainerRuntimeConfigDesign.md
```diff
@@ -0,0 +1,95 @@
...
+
+2. setting log level
```
smarterclayton: When would an end user set the log level (vs when we tell someone to set it during debugging)?

docs/ContainerRuntimeConfigDesign.md
smarterclayton: How does this config object overlap with the Kubelet configuration object?  When a setting might apply to either kubelet or container runtime, which object “wins”?

docs/ContainerRuntimeConfigDesign.md
mrunalp: kubelet settings apply at the pod level while this setting is at the container level. We can advise admins to not set this or keep this unbounded once kubelet setting lands.

docs/ContainerRuntimeConfigDesign.md
mrunalp: This is more for admins enabling it for debugging.

docs/ContainerRuntimeConfigDesign.md
derekwaynecarr: Kubelet setting is pod level.

We backported the flag to set a default per pod pod limit to 3.11+ from 1.14 to address production issues.

Robert is working on enable node to pod pod isolation.

SIG node kep I wrote proposes a granular pod limit feature as a potential policy knob in future but at pod and not Container boundary.  If we do go to Container boundary in kubelet, crio value is default absent kubelet telling it a specific value.

docs/ContainerRuntimeConfigDesign.md
derekwaynecarr: This is to me an issue for debugging , and wanting to canary a pool of nodes.

docs/ContainerRuntimeConfigDesign.md
derekwaynecarr: The fields right now are scoped to build time behavior and not runtime behavior.  It’s possible we should collapse to a common config for both and have MCO render from that global config.  I could see a benefit in letting us configure this at pool level.  In particular, we could have a whitelist registry setting that says only images that run on control planes should come from this source.

docs/ContainerRuntimeConfigDesign.md
derekwaynecarr: I think configuring overlay is distinct from machine api.  We also need this in UPI modes where machines are not necessarily present.

docs/ContainerRuntimeConfigDesign.md
runcom: If we have a canary pool the mco could easily select just those for debugging (but not sure we want this exposed to everyone like this risking to enable debug on every node).

sdodson:
Will the whitelist of supported flags be maintained in a manner that prevents the need to tightly couple the precise version of CRI-O? ie: make sure to only whitelist options supported by all versions of CRI-O 1.12 and not just a specific 1.12.X?

pkg/controller/container-runtime-config/container_runtime_config_controller.go

docs/ContainerRuntimeConfigDesign.md
umohnani8: Was a stray, thanks! Fixed

umohnani8:
@sdodson yup, the configurable options here are supported by all versions of CRI-O 1.12.

umohnani8:
Removed the registries.conf fields as we are moving that to a global config. @rphillips @abhinavdahiya @mrunalp this is ready for review, PTAL.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -0,0 +1,219 @@
...
+	if !ctrcfgValues.IsValid() {
+		return fmt.Errorf("ContainerRuntimeConfig is not valid")
+	}
+
+	ctrcfg := cfg.Spec.ContainerRuntimeConfig
+	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < 20 {
+		return fmt.Errorf("Invalid PidsLimit %q, cannot be less than 20", ctrcfg.PidsLimit)
+	}
+
+	if ctrcfg.LogSizeMax > 0 && ctrcfg.LogSizeMax < 8193 {
```
mrunalp: We should use a const for `8192` and do a <= here.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -0,0 +1,219 @@
...
+	if cfg.Spec.ContainerRuntimeConfig == nil {
+		return nil
+	}
+	ctrcfgValues := reflect.ValueOf(*cfg.Spec.ContainerRuntimeConfig)
+	if !ctrcfgValues.IsValid() {
+		return fmt.Errorf("ContainerRuntimeConfig is not valid")
+	}
+
+	ctrcfg := cfg.Spec.ContainerRuntimeConfig
+	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < 20 {
```
mrunalp: We should use a const for `20` here.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -0,0 +1,219 @@
...
+
+func createNewStorageIgnition(tomlconfig []byte) ignv2_2types.Config {
+	var tempIgnConfig ignv2_2types.Config
+	mode := 0644
+	du := dataurl.New(tomlconfig, "text/plain")
+	du.Encoding = dataurl.EncodingASCII
+	tempFile := ignv2_2types.File{
+		Node: ignv2_2types.Node{
+			Filesystem: "root",
+			Path:       "/etc/containers/storage.conf",
```
mrunalp: We can use constants for `/etc/containers/storage.conf` and `/etc/crio/crio.conf` as well.

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: Fixed. Should we have a maximum allowed for pidsLimit as well?

pkg/controller/container-runtime-config/helpers.go

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -491,3 +491,109 @@ type KubeletConfigList struct {
...
+	ContainerRuntimeConfig    *ContainerRuntimeConfiguration `json:"containerRuntimeConfig,omitempty"`
+}
+
+// ContainerRuntimeConfiguration defines the tuneables of the container runtime
+type ContainerRuntimeConfiguration struct {
+	PidsLimit   int64  `json:"pidsLimit,omitempty"`
+	LogLevel    string `json:"logLevel,omitempty"`
+	LogSizeMax  int64  `json:"logSizeMax,omitempty"`
+	InfraImage  string `json:"infraImage,omitempty"`
+	OverlaySize string `json:"overlaySize,omitempty"`
```
rphillips: OverlaySize, LogSizeMax, and PidsLimit feel like they could be a resource.Quantity to allow users to nicely set the LogSize to say 1G. Thoughts?

templates/master/01-master-container-runtime/_base/files/container-storage.yaml
```diff
@@ -0,0 +1,120 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/containers/storage.conf"
+contents:
+  inline: |
+    # storage.conf is the configuration file for all tools
```
rphillips: I don't think we need the comments in the yaml files. They sorta clutter what is actually being set.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
mrunalp: We can use it for `OverlaySize` and `LogSizeMax`. However, PidLimit is a number and best left as is.

pkg/controller/container-runtime-config/helpers.go
mrunalp: It is the value in /proc/sys/kernel/pid_max - 1. It will be a good idea to add the check.

mrunalp: Actually that is the system limit. The pids cgroup limit must be higher up to the type used to represent it. It looks like 4 million from quick search but I will look deeper and get back.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
umohnani8: So use resource.Quantity and then convert them to strings? I don't see any validation happening in the Quantity to String function, so is changing the type really helpful in this case?

templates/master/01-master-container-runtime/_base/files/container-storage.yaml
umohnani8: The comments have a lot of info on the options in the config files, that is why I have them there. The comments are lost anyway when a CR is created so not completely attached to having them. @mrunalp WDYT?

mrunalp: Yeah we could get rid of them later. right now they are matching the crio.conf before the CR changes are applied. We want to change the default that is laid down is well first.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
umohnani8: Done, using resource.Quantity

Gopkg.toml
```diff
@@ -151,3 +151,15 @@ required = [
 [[override]]
   name = "k8s.io/apiserver"
   version = "kubernetes-1.12.5"
+
+[[constraint]]
+  branch = "release-1.12"
+  name = "github.com/kubernetes-sigs/cri-o"
+
+[[constraint]]
+  branch = "master"
```
runcom: can we have tags or something to avoid `dep` polluting everything when re-syncing?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+	if len(cc) == 0 {
+		return nil, nil, fmt.Errorf("ControllerConfigList is empty")
+	}
+	// Render the default templates
+	tmplPath := filepath.Join(ctrl.templatesDir, role)
+	rc := &mtmpl.RenderConfig{ControllerConfigSpec: &cc[0].Spec}
+	generatedConfigs, err := mtmpl.GenerateMachineConfigsForRole(rc, role, tmplPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("GenerateMachineConfigsforRole failed with error %s", err)
```
runcom: as discussed on slack, lowecase errors

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -0,0 +1,229 @@
...
+	ctrcfg := cfg.Spec.ContainerRuntimeConfig
+	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < minPidsLimit {
+		return fmt.Errorf("invalid PidsLimit %q, cannot be less than 20", ctrcfg.PidsLimit)
+	}
+
+	if ctrcfg.LogSizeMax.Value() > 0 && ctrcfg.LogSizeMax.Value() <= minLogSize {
+		return fmt.Errorf("invalid LogSizeMax %q, cannot be less than 8kB", ctrcfg.LogSizeMax.String())
+	}
+
+	validLogLevels := []string{"error", "fatal", "panic", "warn", "info", "debug"}
```
runcom: can you use  a `map[string]bool` to avoid the loop below?

runcom: and move `validLogLevels` inside the `if`

Gopkg.toml
umohnani8: Changed to a revision

pkg/controller/container-runtime-config/container_runtime_config_controller.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: Changed to a map. I chose string initially so I can easily print the valid values when an error is returned. But map works too :)

umohnani8:
@kikisdeliveryservice sorry about that, added a description to the commit and initial comment in this PR.
@runcom should be ready to merge.

runcom:
@umohnani8 one last comment, and then could you also squash your commits?

kikisdeliveryservice:
Many thanks @umohnani8 the updated description & commit are great. Well done!

kikisdeliveryservice:
@runcom you want her to squash the 2 into 1? We usually try to keep vendor file changes in their own commit.

kikisdeliveryservice:
@umohnani8 I talked to @runcom and don't sqush your commits just update that Generated commit to something like "add autogen templates & vendor bump"

runcom:
Let's keep them 2 commits then but add a meaningful commit message other than Generated (along the line of "add autogen templates + vendor bump" maybe)

umohnani8:
@runcom this should be ready to merge once tests are green.

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -0,0 +1,247 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/crio/crio.conf"
+contents:
+  inline: |
+    # The "crio" table contains all of the server options.
```
cgwalters: It looks like this is overwriting the version from the RPM?  Is it different?

My 2¢ is that it's best to have defaults live in the code, documentation for options in a man page or so.

Not objecting to this, we don't need to remove the `lgtm` but if this isn't different from the RPM version I'm not sure what we're gaining by writing it here again.

umohnani8: I made sure to match this to the one coming in from the RPM. But yeah this overrides the RPM one.

Issue #332, opened 2019-01-21
mcd annontates node as degraded when using kubeletconfig


jeremyeder:
My objective is to add two lines to kubelet.conf by using the kubeletconfig CR.
This has two steps
1) Add a label to the machineconfigpool for my CR to select from.
```oc edit machineconfigpool worker
<snip>
  labels:
    custom-kubelet: cpumanager-enabled
</snip>
```
2) Create a CR that adds my two lines:
```
$ cat enable-cpumanager.yaml 
apiVersion: machineconfiguration.openshift.io/v1
kind: KubeletConfig
metadata: 
  name: cpumanager-enabled
spec:
  machineConfigPoolSelector:
    matchLabels:
      custom-kubelet: cpumanager-enabled
  kubeletConfig:
     cpuManagerPolicy: static
     cpuManagerReconcilePeriod: 5s
```

After completing those two steps, the node was rebooted, but then annotated as Degraded by MCD:

```
jeder@wks: ~  $ oc logs -n openshift-machine-config-operator machine-config-daemon-n2c7d
I0121 19:53:23.876132   46177 start.go:52] Version: 3.11.0-488-g5ea28209-dirty
I0121 19:53:23.876811   46177 start.go:88] starting node writer
I0121 19:53:23.880567   46177 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0121 19:53:23.939922   46177 daemon.go:153] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:21696f797fb0d03a4e47abd2af9fd35104bc09b8cfc0e17bf70bc068e9c515b7 (47.272)
I0121 19:53:23.940116   46177 daemon.go:222] Managing node: ip-10-0-168-9.us-west-2.compute.internal
I0121 19:53:23.970846   46177 start.go:139] Calling chroot("/rootfs")
I0121 19:53:23.995322   46177 daemon.go:399] Current config: worker-ca3911ceaf7ef6d73c9334f62437a392
I0121 19:53:23.995343   46177 daemon.go:400] Desired config: worker-5a8202fb69921f53b7ca029803e1b4eb
I0121 19:53:23.995355   46177 daemon.go:807] No target osImageURL provided
E0121 19:53:23.995718   46177 daemon.go:888] content mismatch for file: "/etc/kubernetes/kubelet.conf"; expected: kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
[...]

PR #333, opened 2019-01-21
kubelet: always delete cpu_manager_state

rphillips:
Talked to @sjenning about cpu_manager_state.

When a node is drained there is not any cpu state to preserve. Preserving
this file also causes the kubelet to fail to startup when the
cpu_manager_state has changed from none to static (or vice versa).

ref: https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy

/cc @jeremyeder 

related to: https://github.com/openshift/machine-config-operator/issues/332


jeremyeder:
Indeed, I had to do this manually.  ready to merge, if you ask me.

jeremyeder:
However does this help with the fact that even after I manually deleted the file, deleted the annotation and deleted the MCO pod, it was still able to detect that I deleted the file? I am just trying to understand this whole setup.

rphillips:
@jeremyeder getting the flow to be correct should help.

PR #337, opened 2019-01-22
controller/node: Also add a 5s delay here responding to pool changes

cgwalters:
This is like
https://github.com/openshift/machine-config-operator/pull/303
but for the node controller.

We really don't need to react *instantly* to start updating and
rebooting machines, and having a small delay will help avoid
races when MCs are created rapidly.


kikisdeliveryservice:
Testing this now.

cgwalters:
Look at the logs here for an example of this:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/324/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/78

There the node controller tried to sync 15-16 times in the span of 3 seconds.

kikisdeliveryservice:
Note this is a partial fix for Issue #338

PR #340, opened 2019-01-23
mco: If the platform is unrecognized, treat the platform as 'none'

smarterclayton:
In order to incrementally roll new platforms out across the project
we must be able to provision platforms before all support is added.
Components should treat an unrecognized platform as "None", performing
the same function as they would in that scenario, and print a warning
to the logs.

Formalizing this requirement as we assess how future providers will
be added in openshift/installer#1112 (found in openshift/installer#1109)

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller_test.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller_test.go
@@ -262,26 +262,31 @@ func (f *fixture) expectUpdateKubeletConfig(config *mcfgv1.KubeletConfig) {
 }
 
 func TestKubeletConfigCreate(t *testing.T) {
-	f := newFixture(t)
+	for _, platform := range []string{"aws", "none", "unrecognized"} {
+		t.Run(platform, func(t *testing.T) {
+			f := newFixture(t)
 
-	cc := newControllerConfig("test-cluster")
-	mcp := newMachineConfigPool("master", map[string]string{"kubeletType": "small-pods"}, metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role", "master"), "v0")
-	mcp2 := newMachineConfigPool("worker", map[string]string{"kubeletType": "large-pods"}, metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role", "worker"), "v0")
-	kc1 := newKubeletConfig("smaller-max-pods", &kubeletconfigv1beta1.KubeletConfiguration{MaxPods: 100}, metav1.AddLabelToSelector(&metav1.LabelSelector{}, "kubeletType", "small-pods"))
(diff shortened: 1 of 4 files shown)


smarterclayton:
/assign @abhinavdahiya 

Spitballed this a bit on slack today, I want to lay some foundations as I go and incrementally get components ready for this.

cgwalters:
I think this needs something like:

```
diff --git a/pkg/controller/template/render.go b/pkg/controller/template/render.go
index 05758bb..71cac20 100644
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -49,12 +49,15 @@ const (
 //                                    /files/hostname.tmpl
 //
 func generateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1.MachineConfig, error) {
-	if config.Platform == "" {
-		return nil, fmt.Errorf("cannot generateMachineConfigs with an empty Platform")
-	}
+	platform := config.Platform
 
-	if config.Platform == "_base" {
+	if platform == "" {
+		return nil, fmt.Errorf("cannot generateMachineConfigs with an empty Platform")
+	} else if platform == "_base" {
 		return nil, fmt.Errorf("platform _base unsupported")
+	} else if platform == "none" {
+		// none is the supported name for our internal "_base" to
+		// be used during bringup of new platforms.
 	}
 
 	infos, err := ioutil.ReadDir(templateDir)
@@ -125,8 +128,12 @@ func GenerateMachineConfigsForRole(config *RenderConfig, role string, path strin
 }
 
 func generateMachineConfigForName(config *RenderConfig, role, name, path string) (*mcfgv1.MachineConfig, error) {
+	expectedPlatformDirs := []string{"_base"}
+	if config.Platform != "base" {
+		expectedPlatformDirs = append(expectedPlatformDirs, config.Platform)
+	}
 	platformDirs := []string{}
-	for _, dir := range []string{"_base", config.Platform} {
+	for _, dir := range expectedPlatformDirs {
 		platformPath := filepath.Join(path, dir)
 		exists, err := existsDir(platformPath)
 		if err != nil {
```

too

smarterclayton:
I kept a similar structure and fixed the test.

This highlighted something that the MCO API needs before GA - we need to be using consistent constants (the constants in https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L41 need to be used everywhere) @abhinavdahiya also in install config

kikisdeliveryservice:
@smarterclayton I opened an issue to make constants consistent in MCO: https://github.com/openshift/machine-config-operator/issues/342

smarterclayton:
I believe I found a bug, if Platform was empty on ControllerConfig before we would have looked at the root template dir incorrectly when trying to find templates (we weren't guarding some of the config generation with an empty string check, which this PR fixes).

smarterclayton:
Any other comments?

smarterclayton:
Gentle ping ;)

abhinavdahiya:
> Gentle ping ;)

/approve

PR #341, opened 2019-01-23
kubelet: use openshift pod image

rphillips:
Use the OpenShift container pod within the kubelet. Images are not passed to the template rendering engine... yet. There are various other templates that hard-code images for now with a TODO to fix. 

/cc @smarterclayton @sjenning

### DIFF
--- a/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
@@ -19,6 +19,7 @@ contents: |
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
         --node-labels=node-role.kubernetes.io/master \
+        --pod-infra-container-image=quay.io/openshift/origin-pod:v4.0 \
         --minimum-container-ttl-duration=6m0s \
         --client-ca-file=/etc/kubernetes/ca.crt \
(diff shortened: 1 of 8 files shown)


smarterclayton:
This is correct

smarterclayton:
@mrunalp does crio still use the pod image

mrunalp:
@smarterclayton CRI doesn't pass the pod image to us so we still have a setting in crio.conf for it. We would want it to be configurable, right?

smarterclayton:
We need it to come from the payload.  So we need crio.conf to have it and
the payload to inject the exact reference into the operator so it can apply
the correct value if we ever need to upgrade it.

On Jan 23, 2019, at 12:19 AM, Mrunal Patel <notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton> CRI doesn't pass the
pod image to us so we still have a setting in crio.conf for it. We would
want it to be configurable, right?

—
<https://github.com/openshift/machine-config-operator/pull/341#issuecomment-456673865>,
or mute the thread

sjenning:
@smarterclayton is this image going to change OKD vs OCP? or is hardcoding it to quay.io fine?

smarterclayton:
The image has to be defined in the payload, so it's not enough to just
hardcode it, you need to make sure it's plumbed and replaced by
image-references and the release replacement

rphillips:
Closing. We will need to pass the images payload into the rendering engine.

Issue #334, opened 2019-01-21
extend bootstrap MC with osimageurl


cgwalters:
This line from PR 305 (which landed) needs to include the osimageurl from the configmap:

https://github.com/openshift/machine-config-operator/pull/305/files#diff-d6f38878757202a810cfa8038e7ae122R45

It looks to me like we'd need an installer change around here https://github.com/openshift/installer/blob/d1272424396f1d42faf239dc1b976db905d0b338/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L120 to pass the configmap?

I am still trying to figure out this code (and also, how to conveniently test changes to this code).

I think it'd be something like:

```
MACHINE_CONFIG_OS_CONTENT=$(podman run --rm ${release} image machine-os-content)
```

And then we do e.g.
`--machine-config-os-content=${MACHINE_CONFIG_OS_CONTENT}`
and then we plumb that through the bootstrap CLI code into that line of code linked above?

cgwalters:
@abhinavdahiya If you have a minute to sanity check this that'd be appreciated!

cgwalters:
If this is correct then we'd need to land a change to our code to support the CLI option, then a change to the installer to use it (and probably to be nice we should support version skew and have the installer detect the presence of the option).

PR #349, opened 2019-01-27
 operator: Add bootstrap arg --osimg-config-file

cgwalters:
This will allow the installer to render the payload's osImageURL
ConfigMap in the initial MachineConfig as well.

Closes: #334


abhinavdahiya:
can you open an installer PR that shows how this will be used, i don't see how you will extract the os-image-url config file from release image.

cgwalters:
Closing in favor of https://github.com/openshift/machine-config-operator/pull/355

PR #350, opened 2019-01-28
Migrate kubeletconfig to kubernetes@1.12.5

abhinavdahiya:
/cc @rphillips @sjenning


rphillips:
I'll take a look at the failure.

rphillips:
Looks like the CVO needs a bump, and the MCO code updated to use the newer CVO client:

```
        master: Could not introduce github.com/openshift/cluster-version-operator@master, as its subpackage github.com/openshift/cluster-version-operator/pkg/generated/clientset/versioned is missing. (Package is required by (root).)
```

abhinavdahiya:
> Looks like the CVO needs a bump, and the MCO code updated to use the newer CVO client:
> 
> ```
>         master: Could not introduce github.com/openshift/cluster-version-operator@master, as its subpackage github.com/openshift/cluster-version-operator/pkg/generated/clientset/versioned is missing. (Package is required by (root).)
> ```

we should drop any code that imports CVO, we are not using anything from cluster version operator. The OperatorStatus moved to `openshift/api/ config`

rphillips:
/lgtm

nice job!

Issue #346, opened 2019-01-24
MCO Cluster Operator statuses frequently changing


kikisdeliveryservice:
Unsure if this is causing any performance issues (it has not from what I can tell but)

The Cluster Operator machine-config-operator seems to cycle through statuses when the cluster is relatively idle (ie I am not performing and updates/applying mcs/etc...).  Not sure why the status is changing but it makes for a confusing user experience.  See the following.

`$ oc get clusteroperator -n machine-config-operator` yields varying statuses within very short periods of time.  For example:

`machine-config-operator                 3.11.0-494-g18b8dc50-dirty   True        False         False     7s`
`machine-config-operator                 3.11.0-494-g18b8dc50-dirty   False       True          False     1s`

In another example, I see changes in the `conditions:` section but the bottom entry also never changes? So the top two alternate between Available, False / Progressing, True but the oldest entry also never seems to change?

`$ oc get clusteroperator machine-config-operator -o yaml -w`
```
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-01-23T18:14:33Z
  generation: 1
  name: machine-config-operator
  resourceVersion: "919947"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/machine-config-operator
  uid: bbc5a98d-1f3a-11e9-a4a4-02204cc1daaa
spec: {}
status:
  conditions:
  - lastTransitionTime: 2019-01-24T18:39:17Z
    status: "False"
    type: Available
  - lastTransitionTime: 2019-01-24T18:39:17Z
    message: Progressing towards 3.11.0-494-g18b8dc50-dirty
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-01-24T02:24:05Z
    status: "False"
    type: Failing
  extension:
    master: all 3 nodes are at latest configuration master-8b907698492f2829b35f6f826511dfa6
    worker: all 3 nodes are at latest configuration worker-58b30499568a7a2588b05ed01bba91a7
  version: 3.11.0-494-g18b8dc50-dirty
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-01-23T18:14:33Z
  generation: 1
  name: machine-config-operator
  resourceVersion: "919948"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/machine-config-operator
  uid: bbc5a98d-1f3a-11e9-a4a4-02204cc1daaa
[...]

PR #351, opened 2019-01-29
 operator: fix status updates on ClusterOperator

abhinavdahiya:
- ClusterOperator now reports list of versions, each for an operand and the operator itself.
- The versions reported for only available conditions, as per defined behavior in [1]
- Failing and Progressing do not change last Available, as per defined behavor in [2]

[1] https://github.com/openshift/api/blob/677153041bd22c1aa01fef67b68a8502ab1333d5/config/v1/types_cluster_operator.go#L59-L61
[2] https://github.com/openshift/cluster-version-operator/blob/4c2d5e0e3662f9643ed4f0992496539ef6dfdaac/docs/dev/clusteroperator.md#conditions

/cc @kikisdeliveryservice @cgwalters


kikisdeliveryservice:
@abhinavdahiya Thanks for this!  This is for #346 right?

I will pull and test tomorrow :smile:

pkg/operator/status.go
```diff
@@ -26,16 +26,19 @@ func (optr *Operator) syncAvailableStatus() error {
...
 		return nil
 	}
-	now := metav1.Now()
+
+	optrVersion, _ := optr.vStore.Get("operator")
 	// set available
-	SetClusterOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue, LastTransitionTime: now})
+	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+		Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue,
+		Message: fmt.Sprintf("Cluster is available at %s", optrVersion),
```
cgwalters: s/Cluster/MCO/?  (or s/Cluster/Operator/, or...)

pkg/operator/status.go
```diff
@@ -50,23 +53,20 @@ func (optr *Operator) syncProgressingStatus() error {
...
-		SetClusterOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorFailing, Status: configv1.ConditionFalse, LastTransitionTime: now})
-	}
 
-	// set progressing
-	if c := FindClusterOperatorStatusCondition(co.Status.Conditions, configv1.OperatorFailing); c != nil && c.Status == configv1.ConditionTrue {
-		SetClusterOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Message: fmt.Sprintf("Unable to apply %s", version.Version.String()), LastTransitionTime: now})
+	optrVersion, _ := optr.vStore.Get("operator")
+	var message string
+	if optr.vStore.Equal(co.Status.Versions) {
+		// syncing the state to exiting version.
```
cgwalters: s/exiting/existing/ ?

pkg/operator/status.go
```diff
@@ -84,21 +84,29 @@ func (optr *Operator) syncFailingStatus(ierr error) error {
...
-	// clear the available condition
-	SetClusterOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse, LastTransitionTime: now})
 
+	optrVersion, _ := optr.vStore.Get("operator")
+	var message string
+	if optr.vStore.Equal(co.Status.Versions) {
+		// syncing the state to exiting version.
+		message = fmt.Sprintf("Failed to resync %s because: %v", optrVersion, ierr.Error())
+	} else {
+		message = fmt.Sprintf("Failed to progressing towards %s because: %v", optrVersion, ierr.Error())
```
cgwalters: "Failed to progress towards"

pkg/operator/status.go
abhinavdahiya: MCO is reporting that its **view** (in terms of ownership) of the cluster is at that version.

abhinavdahiya:
> Why it changes every 15 seconds is an interesting question...I think we'll try to resynchronize on any status change from a lot of objects, including e.g. the MCD daemonset. It feels like we should be doing more "diff detection" here or so, there are presumably good patterns for this but I am not familiar enough with the ecosystem yet.

based on https://github.com/openshift/machine-config-operator/blob/96228eb286a36944b8a729abeea0442ed73af776/pkg/operator/operator.go#L124-L132

it is reacting to changes to these objects.

pkg/operator/status.go
abhinavdahiya: done

pkg/operator/status.go
abhinavdahiya: done

pkg/operator/status.go
kikisdeliveryservice: ` - lastTransitionTime: 2019-01-29T20:53:23Z
    message: Cluster is available at 3.11.0-521-g377bde3b
    status: "True"
    type: Available
  - lastTransitionTime: 2019-01-30T20:56:32Z
    message: Running resync for 3.11.0-521-g377bde3b
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-01-30T06:21:34Z
    status: "False"
    type: Failing
`
@cgwalters it ends up reading this way in the yaml ^^

cgwalters: OK, sounds fine to me!

kikisdeliveryservice:
Since there were some changes, I'm going to repull this and test again.

abhinavdahiya:
> The version column values seem to be missing?:
> 
> ```
> $ oc get clusteroperators machine-config-operator
> NAME                      VERSION   AVAILABLE   PROGRESSING   FAILING   SINCE
> machine-config-operator             True        False          False     12m
> ```
> but in `oc get clusteroperator machine-config-operator -o yaml -w` :
> 
> ```
>   - name: operator
>     version: 3.11.0-521-gcf207959
> ```

We do not control the definition CRD of cluster operator. CVO does. MCO only creates the CR.

abhinavdahiya:
> /lgtm

hmm, looks like effect of the Prow service degradation.

PR #352, opened 2019-01-29
daemon: fix overwritten files

rphillips:
Files within ignition can be overwritten by later entries. This fixes
the validation to check all ignition written files for validity instead
of checking only the first.

This fixes an issue where the MCD marks a node as degraded when a file within ignition has overwritten a default config (ie: KubeletConfig)

/cc @jeremyeder @sjenning

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -22,6 +22,7 @@ import (
 	mcfgclientv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
+	fsnotify "gopkg.in/fsnotify.v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -847,7 +847,8 @@ func (dn *Daemon) checkUnits(units []ignv2_2types.Unit) bool {
 
 // checkFiles validates the contents of  all the files in the
 // target config.
-func (dn *Daemon) checkFiles(files []ignv2_2types.File) bool {
+func checkFiles(files []ignv2_2types.File) bool {
+	overwrittenFiles := make(map[string]bool)
```
jlebon: Though `checkFileContentsAndMode` will still be spouting scary messages when it handles the initial instance of a path. WDYT about just reversing the list/iterating backwards and skipping over paths we've already checked?

pkg/daemon/daemon.go
```diff
@@ -22,6 +22,7 @@ import (
 	mcfgclientv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
+	fsnotify "gopkg.in/fsnotify.v1"
```
jlebon: Minor/optional: this is from `gofmt` I suppose? We usually try to split those out as separate commits to make review easier.

pkg/daemon/daemon.go
rphillips: will do.

rphillips:
@jlebon Pushed the requested fixes... Thank you for the review!

PR #359, opened 2019-01-31
Change SSH taint to annotation instead

yuqi-zhang:
After some discussion we noted that SSH tainting nodes breaks upstream tests, as well as disrupts many existing workflows without adding much security. Thus we will move to an annotation instead to mark a node as tainted, and warn the admin through  e.g. the console when this is applied.

This builds on top of https://github.com/openshift/machine-config-operator/pull/291 (rebased directly with fixes). Alternatively we could rebase/merge that I can rebase.

This also supercedes https://github.com/openshift/machine-config-operator/pull/290.

cc/ @cgwalters @ashcrow


ashcrow:
I think moving to this PR makes the most sense. I'd recommend the other two to close in favor of this one.

docs/MachineConfigDaemon.md
```diff
@@ -14,6 +14,8 @@
 
 MachineConfigDaemon is scheduled on the machines in a cluster as a DaemonSet. This daemon is responsible for performing machine updates in OpenShift 4. The update will include tasks related to the systemd units, files on disk, operating system upgrades etc. The MachineConfigDaemon updates a machine to configuration defined by MachineConfig as instructed by the MachineConfigController.
 
+The MachineConfigDaemon is also responsible for annotating a node with `machineconfiguration.openshift.io/rhcosSSH=accessed` when it detects an SSH access to the machine.
```
cgwalters: I know this is a bikeshed but "rhcosSSH" looks awkward.  How about just `machineconfiguration.openshift.io/ssh=accessed` ?

docs/MachineConfigDaemon.md
```diff
@@ -129,3 +131,7 @@ The draining of pods on the only master node will not evict the control plane as
 ### Node drain etcd static pods on masters
 
 Etcd is co-located on master nodes as static pods. The draining behavior defined above prevents draining of static pods to prevent interference to etcd cluster by the daemon.
+
+## Annotating on SSH access
+
+RHCOS nodes in Openshift are not meant to be manually accessed via SSH. MCD uses logind to watch for login sessions, which, upon detection, warns the user and annotates the node with `machineconfiguration.openshift.io/rhcosSSH=accessed`. This in turn will be used to warn cluster admins.
```
cgwalters: At some point I think we're going to need to link to a more canonical doc on this as we're...somewhat inconsistent in our "is SSH allowed/encouraged" phrasing.  But this is OK for now.

docs/MachineConfigDaemon.md
ashcrow: Either works from my POV.

pkg/daemon/writer.go
```diff
@@ -122,16 +117,19 @@ func (nw *NodeWriter) SetUpdateDegradedMsgIgnoreErr(msg string, client corev1.No
...
+                MachineConfigDaemonSSHAccessAnnotationKey: MachineConfigDaemonSSHAccessValue,
+        }
+        respChan := make(chan error, 1)
+        nw.writer <- message{
+                client:          client,
+                node:            node,
+                annos:           annos,
+                responseChannel: respChan,
+        }
+        return <-respChan
```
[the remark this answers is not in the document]
yuqi-zhang: Sure, I don't think I've seen this hang before, but either case this is only called in a separate goroutine for SSH monitoring, so it should be fine either way.

That said its good to investigate, all the other annotation writing uses the same mechanism, so its probably better to do that in a separate PR.

docs/MachineConfigDaemon.md
yuqi-zhang: Sure, will change

docs/MachineConfigDaemon.md
yuqi-zhang: Agreed

yuqi-zhang:
Note that this doesn't actually pick up e.g. `oc rsh` access, nor is it able to detect SSH access before MCD is running on the node.

cgwalters:
> nor is it able to detect SSH access before MCD is running on the node.

`journalctl MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66` will give us any logins before.

cgwalters:
(We could also just set up a persistent journald query instead of watching dbus, but eh)

ashcrow:
```
test "release-latest" failed: pod release-latest was already deleted
```

Retry in a ~20 minutes

yuqi-zhang:
> journalctl MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66
True, we could use that check for ssh access since last reboot when MCD starts up, although if an admin was monitoring for the annotation and deletes it after handling alerts, it could wrongly flag an ssh access if e.g. the MCD image was updated, causing MCD to restart.

In either case since we don't really have a finalized workflow for how this annotation is going to be used, I think it's probably better to modify as needed later.

pkg/daemon/writer.go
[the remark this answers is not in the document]
runcom: Well, context and cancelation are generally good practice whenever goroutines are involved (especially for rest client and stuff like this). And ideally I'd love all this to have a context. But yeah, let's defer for another PR.

pkg/daemon/daemon.go
```diff
@@ -280,9 +280,10 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 		case <-stopCh:
 			return
 		case msg := <-sessionNewCh:
-			glog.Infof("Detected a new logged in session: %v", msg)
-			if err := dn.applySSHTaint(); err != nil {
-				exitCh <- fmt.Errorf("Error during taint: cannot apply taint due to: %v", err)
+			glog.Infof("Detected a new login session: %v", msg)
+			glog.Infof("Login access is discouraged! Applying annotation: %v", MachineConfigDaemonSSHAccessAnnotationKey)
+			if err := dn.nodeWriter.SetSSHAccessed(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
+				exitCh <- fmt.Errorf("Error: cannot apply annotation for SSH access due to: %v", err)
```
jlebon: Note this will degrade the node if we can't set the annotation. Do we really want that? (Not rhetorical, actually asking. :)). I'm leaning more towards not making this a hard error and just logging it for now.

yuqi-zhang: Good point, we can switch to a warning, but as far as I can tell all other failed nodewriter invocations also degrade the node. If we cannot write an annotation there is likely an issue with the node, and future updates, etc. would fail.

I don't think there is a specific case that an inability to write an annotation would come only from this?

PR #361, opened 2019-02-01
e2e/testing: A few small patches

cgwalters:


jlebon:
No new features, simply a docs change, typo fix, and CI debugging aid.
/lgtm

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -304,8 +304,9 @@ type MachineConfigPoolSpec struct {
 	// Label selector for Machines.
 	MachineSelector *metav1.LabelSelector `json:"machineSelector,omitempty"`
 
-	// If true, changes to this machine pool should be stopped.
-	// This includes generating new desiredMachineConfig and update of machines.
+	// If true, changes to this machine pool should be stopped,
+	// including generating new desiredMachineConfig and update of machines.
+	// This flag is intended for administrators to change.
```
abhinavdahiya: wait, IMO all the fields in spec are for edit. not just this...

cgwalters: Hm...I thought I saw something else in the operator changing the spec, but trying now I can't find anything.

Dropped that patch out and updated :arrow_down:

cgwalters:
> @cgwalters could we put something about testing/e2e/etc in the PR title so it's easier to find in github list later?

ashcrow did that, ok now?

kikisdeliveryservice:
@cgwalters great! :+1:

PR #364, opened 2019-02-01
HACKING.md: add paragraph to turn verbose on

runcom:
Add a paragraph about turning debugging on using `oc patch` for any
given component.


jlebon:
Doesn't this only work if you disable the MCO though? Otherwise it'll reapply e.g. the daemonset definition and overwrite the patch.

runcom:
I should add a note about that, yeah

Issue #338, opened 2019-01-22
fix initial race between template and render sub-controllers


cgwalters:
In progress PR: https://github.com/openshift/machine-config-operator/pull/385

---

Even with the delay from https://github.com/openshift/machine-config-operator/pull/303
we've still been seeing races;
https://github.com/openshift/machine-config-operator/pull/337
is a further band-aid but obviously not a complete solution.

Currently the template controller generates 3 fragments (base, ssh, kubelet) and my osimage work is about to extend that to 4.

The render controller can see any from 1 to 3 (or 4 post-osimageurl) MCs.

One approach is for the template controller to *only* generate 00-{master,worker}.  That gives us atomicity.  We could easily do this by just appending the Ignition internally.

This is primarily an issue on initial cluster bringup; hopefully the work that landed to have the controller only render MCs from its own version addresses upgrade issues.

Another solution would be for the render sub-controller to wait until the template sub-controller has signaled that it's done an initial complete generation.

kikisdeliveryservice:
Example logs from the race seen today:
```
$ oc -n openshift-machine-config-operator logs deploy/machine-config-controller
Found 2 pods, using pod/machine-config-controller-86f898f7cf-2d9g6
I0122 18:40:02.844556       1 start.go:45] Version: 3.11.0-490-g8e01f051-dirty
I0122 18:40:03.044480       1 leaderelection.go:185] attempting to acquire leader lease  openshift-machine-config-operator/machine-config-controller...
[...]

PR #365, opened 2019-02-01
controller: Start sub-controllers in order of dependency

cgwalters:
The render controller needs kubelet data, so start it first.
And add a bit of docs.


cgwalters:
This is not a fix for https://github.com/openshift/machine-config-operator/issues/338 - it will make it a bit less likely though and is prep for a real fix.

ashcrow:
Agreed. I feel like having some more structure around ordering will help ... or at the very least make it easier to debug.

jlebon:
This looks sane to me, though I'd like @abhinavdahiya to take a look. I haven't really played much with this part of the code before.

abhinavdahiya:
No strong opinions on this, will defer to the team on this.

PR #369, opened 2019-02-01
pkg/daemon: idiomatic Go cleanup

runcom:
Went through a bunch of cleanup to be more idiomatic when writing Go
code (no else if previous return, default values, return on err, bool comparison etc etc)
(mainly went to read and learn the code)
There should be no functional change at all.

(so sorry for this PR :angel: )

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -66,9 +66,7 @@ func (dn *Daemon) writePendingState(desiredConfig *mcfgv1.MachineConfig) error {
 // updateOSAndReboot is the last step in an update(), and it can also
 // be called as a special case for the "bootstrap pivot".
 func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
-	var err error
-
-	if err = dn.updateOS(newConfig); err != nil {
+	if err := dn.updateOS(newConfig); err != nil {
 		return err
 	}
 
@@ -92,7 +90,7 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 		if err != nil {
 			return err
 		}
-		glog.V(2).Infof("Node successfully drained")
+		glog.V(2).Info("Node successfully drained")
 	}
 
 	// reboot. this function shouldn't actually return.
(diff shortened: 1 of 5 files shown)


kikisdeliveryservice:
@runcom this is the best thank you!  checking it out!

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	kscheme "k8s.io/client-go/kubernetes/scheme"
 
-	"github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	v1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
```
abhinavdahiya: Why this change?

pkg/daemon/update.go
```diff
@@ -303,16 +299,10 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	if err := dn.writeFiles(newConfig.Spec.Config.Storage.Files); err != nil {
 		return err
 	}
-
 	if err := dn.writeUnits(newConfig.Spec.Config.Systemd.Units); err != nil {
 		return err
 	}
-
-	if err := dn.deleteStaleData(oldConfig, newConfig); err != nil {
```
abhinavdahiya: nit: if in future you add another step after deleteStaleData, you'll have to bring back this if... The current setup keeps it clearer imo.

pkg/daemon/writer.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	"github.com/golang/glog"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
abhinavdahiya: Why this? Maybe `corev1`?

runcom: This is golint doing this, overlooked

pkg/daemon/update.go
runcom: Boils down to preference for me, if you prefer it that way I'll change it back

pkg/controller/bootstrap/bootstrap.go
runcom: Golint doing stupid stuff

jlebon:
Very nice!
/lgtm

pkg/daemon/writer.go
```diff
@@ -103,9 +103,9 @@ func (nw *NodeWriter) SetUpdateDegraded(err error, client corev1.NodeInterface,
 // node to degraded due to that error.
 func (nw *NodeWriter) SetUpdateDegradedIgnoreErr(err error, client corev1.NodeInterface, node string) error {
 	// log error here since the caller won't look at it
-	degraded_err := nw.SetUpdateDegraded(err, client, node)
-	if degraded_err != nil {
-		glog.Errorf("Error while setting degraded: %v", degraded_err)
+	degradedErr := nw.SetUpdateDegraded(err, client, node)
```
yuqi-zhang: To keep it consistent we should probably just make it `if degradedErr := nw.SetUpdateDegraded(err, client, node); degradedErr != nil` here right?

Issue #358, opened 2019-01-31
daemon: unable to rotate token error


yuqi-zhang:
After the daemon has been running for ~5 minutes, it starts to spam:

`E0131 21:27:01.343248   29281 token_source.go:132] Unable to rotate token: failed to read token file "/var/run/secrets/kubernetes.io/serviceaccount/token": open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory`

repeatedly in the daemon logs. It does not seem to otherwise affect daemon behaviour.

abhinavdahiya:
Is this happening only for daemon? And all other pods, operator, controller, server are working?

yuqi-zhang:
Operater/controller/server seem to be fine. Although I specifically rebuilt the daemon from the `master` branch myself.

abhinavdahiya:
It might have to do with chroot after client has been created [here](https://github.com/openshift/machine-config-operator/pull/46)

abhinavdahiya:
daemon needs to be able to rotate its credentials.

yuqi-zhang:
If that's the case I'm surprised we haven't seen this earlier, unless token rotation is relatively new?

I can check whether this happens with a fresh cluster from `master` of the installer to make sure whether changing image and rebooting does something wonky.

kikisdeliveryservice:
@ashcrow how long lived would we be looking for?

ashcrow:
I don't know how often the rotation occurs by default unfortunately.

cgwalters:
Yeah we should probably back off the chroot into the host rootfs...though that is going to require some nontrivial reworking in the code.

kikisdeliveryservice:
Someone else also saw this in a new cluster from today (logs are in the link above).

derekwaynecarr:
i hit this on a cluster < 1hr.  i discovered it when i appeared to have a master show "degraded" state while trying to apply a new update payload, and not sure what to do to debug further.  fwiw, i see the token is in pod.

PR #370, opened 2019-02-02
daemon: Bind /run/secrets/ in so we can see SA tokens

cgwalters:
Because we `chroot()` we need to ensure that we can still see
the service account tokens injected into our pod.

Closes: https://github.com/openshift/machine-config-operator/issues/358

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


cgwalters:
(Not tested locally yet, just tossing this up as I was thinking about a fix)

pkg/daemon/daemon.go
```diff
@@ -273,6 +274,18 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	return dn.nodeWriter.SetUpdateDegradedIgnoreErr(err, dn.kubeClient.CoreV1().Nodes(), dn.name)
 }
 
+// BindPodMounts ensures that the daemon can still see e.g. /run/secrets/kubernetes.io
+// service account tokens after chrooting.  This function must be called before chroot.
+func (dn *Daemon) BindPodMounts() error {
+	targetSecrets := dn.rootMount + "/run/secrets"
```
runcom: nit: `filepath.Join`

runcom:
Just a utopic point here, cause of the complexity maybe, but is there any way we can add a test for this to make sure we won't regress in the future? (my understanding is also that this was working before)

cgwalters:
This appears to not have worked, I still see the error in the mcd logs in this PR.

>  but is there any way we can add a test for this to make sure we won't regress in the future?

Yeah...I think we could scrape the MCD logs in `make test-e2e`.

> (my understanding is also that this was working before)

Not sure...but I doubt it, I suspect the rebase added an error message?  Nothing in how the MCD works here changed recently.

jlebon:
Yeah, I think that makes sense. Actually, given that we mount the whole rootfs and chroot, we're not really making use of the other mountpoints defined in the daemonset either. We should probably document that.

Related: #6

jlebon:
#377

cgwalters:
OK, figured it out I think - we needed `--rbind` since the secrets are also mount points.

cgwalters:
```
$ oc logs -c test pods/e2e-aws-op
clusterversion.config.openshift.io/version condition met
=== RUN   TestMCDToken
--- PASS: TestMCDToken (2.24s)
=== RUN   TestOperatorLabel
--- PASS: TestOperatorLabel (0.04s)
PASS
ok      github.com/openshift/machine-config-operator/test/e2e   2.298s
```

:confetti_ball:

ashcrow:
```
level=fatal msg="failed to initialize the cluster: timed out waiting for the condition"
```

/retest

PR #371, opened 2019-02-02
pkg/daemon: fix files mode checking

runcom:
~~forget~~


runcom:
turns out it's working just fine and my testing is dumb sometimes

PR #373, opened 2019-02-03
pkg/daemon: check channel closing

### LINKED ISSUE #372: ssh: accessed annotation when creating a new MC
creating a mc to e.g. add a new file applies the ssh: accessed annotation

```
10:20:39 [github.com/openshift/installer] ‹master*› oc get nodes -o yaml | grep accessed                                   
      machineconfiguration.openshift.io/ssh: accessed
      machineconfiguration.openshift.io/ssh: accessed
      machineconfiguration.openshift.io/ssh: accessed
```

related to https://github.com/openshift/machine-config-operator/pull/359 and  the use of `logind.SessionNew` signal

```
10:04:23 [github.com/openshift/installer] ‹master*› cat mcadd.yaml                                   1 ↵
kind: MachineConfig
metadata:
  labels:
  name: worker-add-file-000
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,test
        filesystem: root
        mode: 420
        path: /etc/mytestconf
```

Seems to happen on reboot though,  due to the subscribe channel closing:

```
I0203 09:30:13.729708    5632 update.go:667] machine-config-daemon initiating reboot: Node will reboot into config worker-1aec7264842ef893b80b896acaef984b
I0203 09:30:13.788753    5632 daemon.go:283] Detected a new login session: <nil>
I0203 09:30:13.788781    5632 daemon.go:284] Login access is discouraged! Applying annotation: machineconfiguration.openshift.io/ssh
```

It doesn't happen on real ssh login, and we're indeed getting a proper signal instead of `nil`:
```
I0203 09:49:48.698628   13220 daemon.go:283] Detected a new login session: &{:1.1 /org/freedesktop/login1 org.freedesktop.login1.Manager.SessionNew [1 /org/freedesktop/login1/session/_31]}
I0203 09:49:48.698734   13220 daemon.go:284] Login access is discouraged! Applying annotation: machineconfiguration.openshift.io/ssh
```

runcom:
Fix #372, after this patch, no ssh/accessed annotation is applied on a machine config addition.

On reboot, the SessionNew subscribed channel gets closed
causing the node ssh/accessed annotation to be added.
This patch fixes that by checking if the channel is indeed
closing before going ahead and add the annotation.

Haven't added an e2e cause watching for something _not to_ happen is gonna be the best flake ever, I'll defer to add proper testing around ssh/accessed being added on real login in a follow up.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -279,7 +279,11 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 		select {
 		case <-stopCh:
 			return
-		case msg := <-sessionNewCh:
+		case msg, ok := <-sessionNewCh:
+			if !ok {
+				glog.V(4).Info("Not adding the ssh accessed annotation because the logind SessionNew channel is closed")
+				return
+			}
 			glog.Infof("Detected a new login session: %v", msg)
 			glog.Infof("Login access is discouraged! Applying annotation: %v", MachineConfigDaemonSSHAccessAnnotationKey)
 			if err := dn.nodeWriter.SetSSHAccessed(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {


yuqi-zhang:
/lgtm
Good catch!

pkg/daemon/daemon.go
```diff
@@ -279,7 +279,10 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 		select {
 		case <-stopCh:
 			return
-		case msg := <-sessionNewCh:
+		case msg, ok := <-sessionNewCh:
+			if !ok {
+				continue
```
jlebon: Maybe's let's just `glog.Infof` and `return` here? If the channel is closed, then there's no need to keep this goroutine running, right? (Though of course I agree in practice, closed means probably rebooting).

A stronger approach might be to close the stop channel or something before rebooting so that goroutines can exit cleanly rather than having to deal with "is this an error or are we just rebooting?".

runcom: Mmm not sure, closing a channel isn't really asking error or not, I've chosen continue as if this is still not rebooting the stopCh can still kick off and tear down everything if we really want to

jlebon: Yeah, though if all we're doing is waiting for the stop channel, then we might as well not be waiting at all and just exit the goroutine, right?

runcom: Sure I was just chatting about it, I'm going to change it to return, do you really want a log as well? Reboot already logs about it but I guess i can add one at v4

jlebon: The logging part is mostly so that if we're somehow not rebooting, we'll know why new SSH accesses are not resulting in annotations being added.

What I meant re. errors is, if the channel is closed but it's not because we're rebooting, then something else went wrong and we could e.g. try to resubscribe N times. Hence the idea of killing goroutines cleanly before doing the reboot so we don't have to deal with racy errors we don't care about. Not suggesting doing this now though. :)

runcom: added glog and return, please check it out!

PR #375, opened 2019-02-03
 pkg/daemon: run login monitor for Degraded nodes

runcom:
If there would ever be a way to reconcile a Degraded node, I believe it
would be worth to still learn if someone jumped on a node and messed up
with it so we can decide what to do.

Feel free to close this telling I'm dumb as well, I'm probably missing some context which led to this PR while playing around reading the code but it makes sense to me to still have something like this. This is also not fixing anything related to super early startup where someone can jump on a node _before_ the daemon is running and tracking accesses.

This PR builds on #373, I'll rebase it once that merges

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -103,6 +103,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 			startOpts.kubeletHealthzEndpoint,
 			nodeWriter,
 			exitCh,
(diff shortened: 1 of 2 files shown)


ashcrow:
/hold

until pre-req is merged

runcom:
rebased and push forced

/unhold

ashcrow:
pre-req merged, code looks fine to me. Deferring to a second review for merge.

PR #378, opened 2019-02-04
OWNERS: Rework to move everyone to toplevel, add runcom

cgwalters:
We're reorganizing the teams to have part of the RHCOS team own
this operator too.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,9 +2,12 @@
 
 approvers:
   - abhinavdahiya
-  - crawford
-  - smarterclayton
-  - wking
   - ashcrow
   - cgwalters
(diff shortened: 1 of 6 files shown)


runcom:
is this file still needed though in `pkg/daemon`? https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/OWNERS

cgwalters:
> is this file still needed though in pkg/daemon? https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/OWNERS

Good call; thanks.  I removed all of the "sub-OWNERS" ↑

PR #380, opened 2019-02-05
pkg/controller/template: get rid of pointer to map

runcom:
Just do this as basic gotchas in Go even if it was a nice hack. Unit tests cover this code path.


jlebon:
This looks good to me! Will let others have a look.
/approve

cgwalters:
I'm generally OK with this but one thing we need to keep in mind with general refactoring is that today we don't really have many tests for the MCO in general - see <https://github.com/openshift/machine-config-operator/issues/257>.  Code LGTM just saying let's be careful :smile:

runcom:
Yeah, I fear regressions at any time here but this code path has unit tests which I've tested they can catch up failures. I generally agree though

PR #381, opened 2019-02-05
pkg/daemon: detect ssh accesses before daemon started

### LINKED ISSUE #379: Early ssh access detection
I'm surely missing some context but I believe the issue still holds. Anyone can currently ssh into nodes before the MCD runs or the login monitor goroutines is started leaving a window open where we won't detect that.

~~I was thinking (wrongly cause I don't know the installer code probably), can we completely disable the sshd service between the installer ending needing it, and have the MCD itself then start it through the systemd dbus api? Just a random thought (and guess) but I believe we can keep brainstorming here.~~

runcom:
Reads the journal checking for logind message ID and annotates the node
with the ssh accessed label if it finds out there were previous
accesses.

Close #379 

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-systemd/login1"
+	"github.com/coreos/go-systemd/sdjournal"
 	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
@@ -252,6 +253,48 @@ func NewClusterDrivenDaemon(
 	return dn, nil
 }
 
+const (
+	// IDs are taken from https://cgit.freedesktop.org/systemd/systemd/plain/src/systemd/sd-messages.h
+	sdMessageSessionStart = "8d45620c1a4348dbb17410da57c60c66"
+)
+
+func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
+	journal, err := sdjournal.NewJournal()
+	if err != nil {
+		return err
+	}
+	defer journal.Close()
+	if err := journal.AddMatch("MESSAGE_ID=" + sdMessageSessionStart); err != nil {
+		return err
+	}
+	if err := journal.SeekHead(); err != nil {
+		return err
+	}
+	r, err := journal.Next()
+	if err != nil {
+		return err
+	}
+	// journal EOF
+	if r == 0 {
+		return nil
+	}
+	// just one entry is enough to understand if someone jumped on the node
+	// from the very first boot
+	entry, err := journal.GetEntry()
+	if err != nil {
+		return err
+	}
+	// just sanity checking
+	if entry != nil {
+		glog.Info("Detected a login session before the daemon took over on first boot")
+		glog.Infof("Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
+		if err := dn.applySSHAccessedAnnotation(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // Run finishes informer setup and then blocks, and the informer will be
 // responsible for triggering callbacks to handle updates. Successful
 // updates shouldn't return, and should just reboot the node.
@@ -319,11 +362,18 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 			}
 			glog.Infof("Detected a new login session: %v", msg)
 			glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
-			if err := dn.nodeWriter.SetSSHAccessed(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
-				exitCh <- fmt.Errorf("Error: cannot apply annotation for SSH access due to: %v", err)
+			if err := dn.applySSHAccessedAnnotation(); err != nil {
+				exitCh <- err
 			}
 		}
 	}
+}
+
+func (dn *Daemon) applySSHAccessedAnnotation() error {
+	if err := dn.nodeWriter.SetSSHAccessed(dn.kubeClient.CoreV1().Nodes(), dn.name); err != nil {
+		return fmt.Errorf("error: cannot apply annotation for SSH access due to: %v", err)
+	}
+	return nil
 }
 
 func (dn *Daemon) runKubeletHealthzMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
@@ -502,7 +552,8 @@ func (dn *Daemon) getPendingConfig() (string, error) {
 // CheckStateOnBoot is a core entrypoint for our state machine.
 // It determines whether we're in our desired state, or if we're
 // transitioning between states, and whether or not we need to update
-// to a new state.
+// to a new state. It also checks if someone jumped on a node before
+// the daemon took over.
 //
 // Some more background in this PR: https://github.com/openshift/machine-config-operator/pull/245
 func (dn *Daemon) CheckStateOnBoot() error {
@@ -513,6 +564,11 @@ func (dn *Daemon) CheckStateOnBoot() error {
 	state, err := dn.getStateAndConfigs(pendingConfigName)
 	if err != nil {
 		return err
+	}
+	if state.bootstrapping {
+		if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
+			return fmt.Errorf("error detecting early SSH accesses: %v", err)
+		}
 	}
 	if state.state == constants.MachineConfigDaemonStateDegraded {
 		// We're already degraded.  Sleep so that we don't clobber


runcom:
so, unit tests fail because of the CGO dep I've added:
```
2019/02/05 11:32:21 Resolved source https://github.com/openshift/machine-config-operator to master@649451d6, merging: #381 ce09d423 @runcom
2019/02/05 11:32:21 Resolved openshift/release:golang-1.10 to sha256:7e7effdd1615f3a25b54e9823fca1b572ae14b0b7b2b67a842d711ca2b31c22b
2019/02/05 11:32:21 Resolved openshift/origin-v4.0:base to sha256:acc19bd0ccecd3972fb4c3502bee570bddbd27f63dcdc39f17eb9035e8a7efd5
2019/02/05 11:32:21 Using namespace ci-op-bm4gs4v0
2019/02/05 11:32:21 Running [input:root], src, unit
2019/02/05 11:32:21 Creating namespace ci-op-bm4gs4v0
2019/02/05 11:32:21 Creating rolebinding for user runcom in namespace ci-op-bm4gs4v0
2019/02/05 11:32:21 Setting a soft TTL of 1h0m0s for the namespace
2019/02/05 11:32:21 Setting a hard TTL of 12h0m0s for the namespace
2019/02/05 11:32:21 Setting up pipeline imagestream for the test
2019/02/05 11:32:21 Tagging openshift/release:golang-1.10 into pipeline:root
2019/02/05 11:32:21 Building src
2019/02/05 11:32:21 Build src already succeeded in 24s
2019/02/05 11:32:21 Executing test unit
go test -v ./cmd/... ./pkg/... ./lib/...
# github.com/openshift/machine-config-operator/vendor/github.com/coreos/go-systemd/sdjournal
vendor/github.com/coreos/go-systemd/sdjournal/journal.go:27:33: fatal error: systemd/sd-journal.h: No such file or directory
 // #include <systemd/sd-journal.h>
                                 ^
compilation terminated.
```
Does anybody know how to add dependencies in that _build_ container which runs our unit tests?

jlebon:
I think right now it's failing because it's trying to link it for all the container builds but the dep was only added for the daemon Dockerfiles. @abhinavdahiya mentioned in https://github.com/openshift/machine-config-operator/pull/335#issuecomment-459377601 that we should only turn on `CGO_ENABLED` for the MCD. Doing that should fix it, I think?

(Also, can you do the vendoring part as a separate commit; can probably just cherry-pick https://github.com/openshift/machine-config-operator/pull/335/commits/1068940e1e9bc538e99b998861afc9dbb4270c19 :)).

cgwalters:
I believe the build container is derived from https://github.com/openshift/release/blob/a30cd48207d5f7244ed492450e111c6fbde5b757/projects/origin-release/golang-1.10/Dockerfile

I'm not sure about precedent for either adding to it or making a new one (or a derived container).

pkg/daemon/daemon.go
```diff
@@ -240,6 +245,54 @@ func NewClusterDrivenDaemon(
...
+		if r == 0 {
+			break
+		}
+		entry, err := journal.GetEntry()
+		if err != nil {
+			return err
+		}
+		// this is so fragile, but from testing, it's pretty accurate as a way
+		// to detect if there were sessions created before we took over with the daemon
+		if entry != nil && strings.Contains(entry.Fields["MESSAGE"], "New session") {
```
jlebon: We don't even need to check the `MESSAGE` here. The msg id already represents the event we're looking for.

runcom:
> @abhinavdahiya mentioned in [#335 (comment)](https://github.com/openshift/machine-config-operator/pull/335#issuecomment-459377601) that we should only turn on `CGO_ENABLED` for the MCD. Doing that should fix it, I think?

@jlebon I did that but still fails cause we're probably using another Dockerfile and image in tests (???)

> (Also, can you do the vendoring part as a separate commit; can probably just cherry-pick [1068940](https://github.com/openshift/machine-config-operator/commit/1068940e1e9bc538e99b998861afc9dbb4270c19) :)).

@jlebon is `dep` able to pull in vendors w/o a specific `import` in the code? in case it's not, I'd just wait for either PR to land and rebase the other one I guess. Otherwise, I'll create a separate commit.

> I'm not sure about precedent for either adding to it or making a new one (or a derived container).

thanks, I'll check that out

cgwalters:
While it may seem like a hack honestly I'd say we just fork off that `journalctl` command as a subprocess.

runcom:
> While it may seem like a hack honestly I'd say we just fork off that `journalctl` command as a subprocess.

that's fine with me

pkg/daemon/daemon.go
```diff
@@ -240,6 +245,54 @@ func NewClusterDrivenDaemon(
...
+func (dn *Daemon) detectEarlySSHAccesses() error {
+	journal, err := sdjournal.NewJournal()
+	if err != nil {
+		return err
+	}
+	defer journal.Close()
+	if err := journal.AddMatch(logindNewSessionMessageIDMatch); err != nil {
+		return err
+	}
+	if err := journal.SeekHead(); err != nil {
```
jlebon: Right, but then restarting the node or MCD will always reinstall the annotation, even if it was manually cleared in the past. How about only doing this if we're bootstrapping? (See `getStateAndConfigs()`). Then from that point on, the login monitor will catch subsequent accesses. (Though it still leaves a tiny window during the time the nodes are restarted on upgrades).

pkg/daemon/daemon.go
jlebon: Also, there's no need for a for loop here, right? We can just `.Next()` and then `.GetEntry()` sequentially.

jlebon:
> @jlebon is `dep` able to pull in vendors w/o a specific `import` in the code?

I've just been doing `git add vendor/` and committing that separately first. GitHub does a good job of not expanding the vendoring churn in the "Files changed" tab, but in the terminal it's a pain to wade through.

yuqi-zhang:
Running this locally shows: 
```
# WHAT=machine-config-daemon ./hack/build-go.sh
Using version from git...
Building github.com/openshift/machine-config-operator/cmd/machine-config-daemon (v3.11.0-573-g10b54e42)
# github.com/openshift/machine-config-operator/vendor/github.com/coreos/go-systemd/sdjournal
vendor/github.com/coreos/go-systemd/sdjournal/journal.go:27:11: fatal error: systemd/sd-journal.h: No such file or directory
 // #include <systemd/sd-journal.h>
           ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

runcom:
yes, we're waiting for https://github.com/openshift/release/pull/2783 or go raw journalctl

pkg/daemon/daemon.go
runcom: for the second point, just 1 entry is already enough to mark the node ssh/accessed so yeah, no need for a loop for sequentially calling `GetEntry`

pkg/daemon/daemon.go
runcom: Fixed this and added another check on node startup after reboot

runcom: alright, this still doesn't cover the case when someone clears the annotation and you reboot on upgrades right? this will be called _at every boot_, correct?

runcom: Fixed here https://github.com/openshift/machine-config-operator/pull/381/files#diff-06b532a2a5a5d3ad1d5341b3e00945efR555

ashcrow:
https://github.com/openshift/release/pull/2783 merged

cgwalters:
So rhel-images will fail until the upstream change makes it there.  Not sure how long that will take.

hack/build-go.sh
```diff
@@ -31,4 +31,4 @@ fi
 mkdir -p ${BIN_PATH}
 
 echo "Building ${REPO}/cmd/${WHAT} (${VERSION_OVERRIDE})"
-CGO_ENABLED=0 GOOS=${GOOS} GOARCH=${GOARCH} go build ${GOFLAGS} -ldflags "${GLDFLAGS}" -o ${BIN_PATH}/${WHAT} ${REPO}/cmd/${WHAT}
+CGO_ENABLED=1 GOOS=${GOOS} GOARCH=${GOARCH} go build ${GOFLAGS} -ldflags "${GLDFLAGS}" -o ${BIN_PATH}/${WHAT} ${REPO}/cmd/${WHAT}
```
jlebon: I made this conditional on the MCD only in https://github.com/openshift/machine-config-operator/pull/335/commits/c02eb5de6bcdff49931807070bbdd2cec2508087#diff-9283775b0feecc10455ec28bd08983b0.

jlebon: Hmm, actually that might not work. It looks like other components will then fail to build. From https://github.com/openshift/machine-config-operator/pull/335:

```
--> RUN WHAT=machine-config-controller ./hack/build-go.sh
Using version from git...
Building github.com/openshift/machine-config-operator/cmd/machine-config-controller (v3.11.0-589-gf62a5cf6-dirty)
vendor/github.com/coreos/go-systemd/sdjournal/functions.go...machine-config-operator/vendor/github.com/coreos/pkg/dlopen
error: build error: running 'WHAT=machine-config-controller ./hack/build-go.sh' failed with exit code 1
```

I'm not familiar enough with the golang build system, but it seems like it's trying to bundle all the vendored packages for all the targets instead of only the ones that actually need them?

runcom: it's because the daemon package imports other packages (controller|server) and the build system bundle them together and that's why it's required for other components as well.

abhinavdahiya: You can use build tags to include / build files only when those tags are set...
https://golang.org/pkg/go/build/#hdr-Build_Constraints

runcom: yeah, but that's gonna require some refactor as well

runcom: my point is yeah, we're going to refactor for build tags but I guess for this PR we can ship with CGO_ENABLED=1 anyway, I'm creating a new issue to track this.

runcom: https://github.com/openshift/machine-config-operator/issues/396

ashcrow:
```
could not wait for build: the build machine-config-controller failed after 2m16s with reason DockerBuildFailed: Docker build strategy has failed.
```

runcom:
rebased and removed the vendoring since #335 pulled it already

This should be ready

pkg/daemon/daemon.go
```diff
@@ -252,6 +253,47 @@ func NewClusterDrivenDaemon(
...
+		return err
+	}
+	r, err := journal.Next()
+	if err != nil {
+		return err
+	}
+	// if the journal cursor has something
+	if r == 1 {
```
jlebon: Minor: could avoid the indentation here by checking for 0 and returning early.

runcom: thanks, fixed

Issue #383, opened 2019-02-05
install ClusterOperator CRD as part of release payload


cgwalters:
Currently we install it dynamically; this means the CVO is unaware of our status and won't e.g. error out if we start failing.

Doing this will be a way to get gating on degraded status.

cgwalters:
Example: https://github.com/openshift/cluster-kube-controller-manayger-operator/blob/master/manifests/0000_12_kube-controller-manager-operator_07_clusteroperator.yaml

cgwalters:
And then we'd need to start e.g.

```
diff --git a/pkg/operator/status.go b/pkg/operator/status.go
index 066f289..2d365a2 100644
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -114,12 +114,6 @@ func (optr *Operator) syncFailingStatus(ierr error) error {
 
 func (optr *Operator) fetchClusterOperator() (*configv1.ClusterOperator, error) {
 	co, err := optr.configClient.ConfigV1().ClusterOperators().Get(optr.name, metav1.GetOptions{})
-	if meta.IsNoMatchError(err) {
-		return nil, nil
-	}
-	if apierrors.IsNotFound(err) {
-		return optr.initializeClusterOperator()
-	}
 	if err != nil {
 		return nil, err
 	}
```

Looking at the code in the CKCMO repo I think we need this:

```
	v1helpers.EnsureOperatorConfigExists(
		dynamicClient,
		v311_00_assets.MustAsset("v3.11.0/kube-controller-manager/operator-config.yaml"),
		schema.GroupVersionResource{Group: operatorv1.GroupName, Version: operatorv1.GroupVersion.Version, Resource: "kubecontrollermanagers"},
	)

```

runcom:
> Looking at the code in the CKCMO repo I think we need this:

we'll also need something like `status.NewClusterOperatorStatusController(` from `github.com/openshift/library-go/pkg/operator/status` as other operator uses + I believe https://github.com/openshift/machine-config-operator/pull/351 must be reworked now to use `library-go` as well.

abhinavdahiya:
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

CVO will NOT install your clusteroperator

runcom:
> CVO will NOT install your clusteroperator

[...]

PR #384, opened 2019-02-05
templates: Don't hardcode /usr/local/bin for etcd

cgwalters:
The image we're trying to switch to has it in `/usr/bin`, let's
just search `$PATH`.

Ref: https://github.com/openshift/installer/pull/1194


ashcrow:
:boom::robot::boom:
:cloud::fire::cloud:

runcom:
just wondering and brainstorming, we can take this on Slack if you want, but how do we make sure we don't regress on similar drifts (https://github.com/openshift/installer/pull/1194)? is this on the installer or on mco? how's the installer testing changes in the MCO (or does it use the latest published images)? do we need to start having a job which runs the latest available installer with the very latest mco code to make sure we don't drift? and viceversa? (spending my morning figuring this out also)

cgwalters:
(Can people add approve and lgtm here?  Prow missed it earlier)

> do we need to start having a job which runs the latest available installer with the very latest mco code

That's what e2e-aws does; it runs master installer against the code from the PR.

What we need to do is start gating on degraded: https://github.com/openshift/machine-config-operator/issues/383

The original PR here sailed right through because we didn't have that.

PR #387, opened 2019-02-06
test/e2e: add MC smoke test

runcom:
Far away from being the right tests we need to test this whole workload
but it's a start. Just making sure we can see the daemon (1) actually
converge to the new config. (Next step would be to deploy a pod with
an hostPath referencing an MC file, exec into this pod and verify
the file is rightly there)

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -1,23 +1,29 @@
 package e2e_test
 
 import (
-	"testing"
+	"fmt"
 	"strings"
+	"testing"
+	"time"
 
+	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/machine-config-operator/cmd/common"
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
 // Test case for https://github.com/openshift/machine-config-operator/issues/358
 func TestMCDToken(t *testing.T) {
 	cb, err := common.NewClientBuilder("")
-	if err != nil{
+	if err != nil {
 		t.Errorf("%#v", err)
 	}
-	k := cb.KubeClientOrDie("sanity-test")
+	k := cb.KubeClientOrDie("mcd-token-test")
 
 	listOptions := metav1.ListOptions{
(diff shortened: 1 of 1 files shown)


runcom:
so this is green but I can't check if the test actually passed (like in `e2e-aws` where we have the full list)

test/e2e/mcd_test.go
```diff
@@ -40,3 +46,103 @@ func TestMCDToken(t *testing.T) {
...
+			}
+		}
+		return false, nil
+	})
+
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"k8s-app": "machine-config-daemon"}).String(),
+	}
+
+	err = wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
```
jlebon: Minor: hmm, so this is going over every log of every pod? Could we make this smarter by only going over pods on worker nodes?

jlebon:
> so this is green but I can't check if the test actually passed (like in `e2e-aws` where we have the full list)

It's not intuitive, but it's in there. Click on "Details", then "artifacts", then go to `artifacts/e2e-aws-op/container-logs/` and check `test.log.gz`:

```
clusterversion.config.openshift.io/version condition met
=== RUN   TestMCDToken
--- PASS: TestMCDToken (0.43s)
=== RUN   TestMCDeployed
--- PASS: TestMCDeployed (81.21s)
=== RUN   TestOperatorLabel
--- PASS: TestOperatorLabel (0.02s)
PASS
ok  	github.com/openshift/machine-config-operator/test/e2e	81.688s
```

Anyway, this looks good to me overall (woohoo more tests!). Will let others have a look.
/approve

test/e2e/mcd_test.go
runcom: I could do that yeah

cgwalters: It's also only the MCD pods right?

runcom: yeah, should be https://github.com/openshift/machine-config-operator/pull/387#discussion-diff-254722357R122

jlebon: Yeah, true. So every log of every MCD pod, rather than just worker MCD pods. Guess if it's not hard to select for only worker nodes, it'd be a nice addition, otherwise meh!

runcom: I'm checking that, but MCDs aren't referencing worker or master explicitly so it's gonna be a great dance to get worker nodes and then match on names, what do you think?

jlebon:
> I'm checking that, but MCDs aren't referencing worker or master explicitly so it's gonna be a great dance to get worker nodes and then match on names, what do you think?

Yeah, totally fine as is!

/lgtm

runcom:
> It's not intuitive, but it's in there. Click on "Details", then "artifacts", then go to `artifacts/e2e-aws-op/container-logs/` and check `test.log.gz`:

let's try to populate the prow interface with actual tests though so we don't need to do that everytime

PR #390, opened 2019-02-07
test/e2e: add tests around ssh/accessed annotation

runcom:
Testing we correctly add the ssh accessed annotation on any given node accessed with ssh, there are other permutation of the test to be added though (annotation also on degraded etc etc, I'm working on that but I'd love #387  to land first so I can use that code to create MCs and degrade machines)

    - add ssh/accessed annotation test
    - add test no accessed on reboot

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -1,7 +1,11 @@
 package e2e_test
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"os"
+	"os/exec"
 	"strings"
 	"testing"
 	"time"
@@ -10,11 +14,16 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/machine-config-operator/cmd/common"
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/daemon"
+	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 )
 
 // Test case for https://github.com/openshift/machine-config-operator/issues/358
@@ -47,9 +56,13 @@ func TestMCDToken(t *testing.T) {
 	}
 }
 
-func mcLabelForWorkers() map[string]string {
+func mcRoleLabelForWorkers() map[string]string {
+	return mcRoleLabelFor("worker")
+}
+
+func mcRoleLabelFor(role string) map[string]string {
 	mcLabels := make(map[string]string)
-	mcLabels["machineconfiguration.openshift.io/role"] = "worker"
+	mcLabels["machineconfiguration.openshift.io/role"] = role
 	return mcLabels
 }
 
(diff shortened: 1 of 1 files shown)


test/e2e/mcd_test.go
```diff
@@ -40,3 +43,50 @@ func TestMCDToken(t *testing.T) {
...
+	err = wait.Poll(2*time.Second, 5*time.Minute, func() (bool, error) {
+		node, err := k.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		sshAnnotation, ok := node.ObjectMeta.Annotations["machineconfiguration.openshift.io/ssh"]
+		if !ok {
+			return false, nil
+		}
+		if sshAnnotation == "accessed" {
```
jlebon: It'd be even more convincing if it checked *before* doing `ssh` that the annotation was absent.

runcom: let's do that

runcom: done

test/e2e/mcd_test.go
```diff
@@ -40,3 +43,59 @@ func TestMCDToken(t *testing.T) {
...
+				nodeName = node.Name
+			}
+		}
+	}
+
+	node, err := k.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("cannot get node %q: %v", nodeName, err)
+	}
+	sshAnnotation, ok := node.ObjectMeta.Annotations["machineconfiguration.openshift.io/ssh"]
```
ashcrow: nit: It may be worth using the constants: https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/constants.go#L34-L37

runcom:
tests are flaking like crazy now, I'll stop retesting but I need to investigate why ssh fails to connect to the node from within the test container (which I've spotted on an early run)

cgwalters:
>  need to investigate why ssh fails to connect to the node from within the test container

It looks like the e2e suite does this:
`export KUBE_SSH_KEY_PATH=/tmp/cluster/ssh-privatekey`

runcom:
@cgwalters thanks a lot!

runcom:
updated

runcom:
This still fails on the ssh command for some reason, need to understand why:
```
	mcd_test.go:169: running "ssh -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no -i /tmp/cluster/ssh-privatekey core@52.90.36.182 true"
	mcd_test.go:215: error ssh'ing into node "52.90.36.182": exit status 255
```

runcom:
alright, now that I was on it, I've added another test and a todo as well along with some refactor as well.

cgwalters:
I should have remembered; we dropped direct ssh to masters/nodes by default.  See
https://github.com/openshift/release/pull/2469

ashcrow:
```
	mcd_test.go:269: error ssh'ing into node ".....": exit status 255, No user exists for uid 1122740000
```

runcom:
holding cause we're discussing ssh tests and testing in general on Slack/emails

/hold

runcom:
well, the bastion host env isn't really helping either, damn

runcom:
This https://github.com/openshift/release/pull/2817 should unblock us here

test/e2e/mcd_test.go
```diff
@@ -117,32 +115,259 @@ func TestMCDeployed(t *testing.T) {
...
+}
 
+func waitForMCDeployedOnNodes(kubeClient kubernetes.Interface, mcName string, nodeCount int) error {
 	listOptions := metav1.ListOptions{
 		LabelSelector: labels.SelectorFromSet(labels.Set{"k8s-app": "machine-config-daemon"}).String(),
 	}
 
-	err = wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
-		mcdList, err := k.CoreV1().Pods("openshift-machine-config-operator").List(listOptions)
+	var seen int
```
runcom: note to self, this has to be moved within the `wait` call to be right

kikisdeliveryservice:
In an effort to clean up the MCO repo, closing old open PRs with no activity.  

Feel free to reopen.

PR #391, opened 2019-02-07
Makefile: set GOCACHE=off in tests

runcom:
Since Go 1.10, test results get cached. This leads to the tests not being run
sometimes because Go has cached the results and just assumes it will get the
same results because the Go code hasn't changed.
Set GOCACHE=off to get clean results everytime (this is especially good
for us since we can poke at the cluster and rerun the test w/o having to
change go code).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,8 +95,8 @@ images.rhel7: $(imc7)
 
 # This was copied from https://github.com/openshift/cluster-image-registry-operato
 test-e2e:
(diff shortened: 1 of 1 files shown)


cgwalters:
Ah, I'd seen other projects doing this.  

/lgtm

runcom:
removed GOCACHE=off from unit tests as we're not relying on external services there anyway so it's ok to cache

ptal

runcom:
good to go

runcom:
flakes flakes flakes

/retest

PR #392, opened 2019-02-07
deps: openshift API bump to pull in Features

rphillips:
Bumps the Openshift API dependency to pull in the Features object.

Note: A future PR will be created to enable Features within the Kubelet.

/cc @sjenning


runcom:
> Note: A future PR will be created to enable Features within the Kubelet.

if this is only going to be needed in a future PR (no code use this vendoring) can't we just wait for that before merging this anyway?

rphillips:
Yeah, we could do that.

PR #397, opened 2019-02-08
Decouple consts in pkg/daemon

### LINKED ISSUE #396: decouple daemon code wrt platform
After #335 and https://github.com/openshift/machine-config-operator/pull/381 the daemon code requires CGO to build because of C bindings to systemd journal. The way the code is structured right now doesn't allow us to avoid compiling with cgo also the controller and the server.
We need to decouple the code based on platform using Go build tag so we're eventually compiling only the daemon with cgo.

runcom:
Just refactor, no change in behavior nor nothing (but do please check!)

TL;DR; consts are the only thing shared and by having their own package we can enable cgo just in daemon (as well as having better code in general, I pretend) 

Closes #396 and you can build daemon only with cgo


jlebon:
I'm confused how the osrelease package split helps; it's only used in the daemon already, right? Or is this just for re-organizing and not actually necessary for #396?

runcom:
> I'm confused how the osrelease package split helps; it's only used in the daemon already, right? Or is this just for re-organizing and not actually necessary for #396?

it's not used just in daemon package, it's also used in cmd/*-daemon package (those are 2 different package)

runcom:
oh well, it doesn't help right, it's just better to have it there that way

cmd/machine-config-daemon/start.go
```diff
@@ -52,7 +51,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	// To help debugging, immediately log version
 	glog.Infof("Version: %+v", version.Version)
 
-	operatingSystem, err := osrelease.GetHostRunningOS(startOpts.rootMount)
+	operatingSystem, err := daemon.GetHostRunningOS(startOpts.rootMount)
```
jlebon: Looks like this hunk is just nullifying an opposite hunk in the previous commit. Might as well untangle those to make the commits cleaner.

runcom: re pushed cleaning up all this mess and actually moved osrelease back to daemon cause, meh, we don't really need it outside except cmd/ but that's gonna be cgo anyway

Issue #259, opened 2018-12-21
tweak hack/cluster-push.sh to use CVO overrides


cgwalters:
Was looking at the CVO docs and came across https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md

It looks like we should change our dev script to set up overrides rather than disabling the CVO.  This is particularly important for the MCO as we really want to test the flow of having OS updates flow via the CVO, but also test changes to e.g. the MCD.

PR #399, opened 2019-02-11
 hack/cluster-push-prep.sh: override only needed objects

runcom:
Close https://github.com/openshift/machine-config-operator/issues/259

First and third commits are auxiliary, the second one is the one actually adding overrides for the CVO in the `hack/cluster-push-*` scripts.

Could you test it out? It works fine on AWS.

### DIFF
--- a/hack/cluster-push-prep.sh
+++ b/hack/cluster-push-prep.sh
@@ -15,7 +15,7 @@ if ! oc get -n openshift-image-registry route/image-registry &>/dev/null; then
 fi
 oc patch -n openshift-image-registry route/image-registry -p '{"spec": {"tls": {"insecureEdgeTerminationPolicy": "Redirect", "termination": "reencrypt"}}}'
 registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq -r ".spec.host")
-if ! curl -k --head https://"${registry}" >/dev/null; then
+if ! curl -k -s --head https://"${registry}" >/dev/null; then
     if ! grep -q "${registry}" /etc/hosts; then
         set +x
         echo "error: Failed to contact the registry"
--- a/hack/cluster-push-prep.sh
+++ b/hack/cluster-push-prep.sh
@@ -9,7 +9,11 @@
 
 set -xeuo pipefail
 
-oc -n openshift-cluster-version scale --replicas=0 deploy/cluster-version-operator
+# XXX: --type merge completely overrides any previous "overrides" array
+#      find a way to just append? json op: add isn't working at all
+#      if there's not an overrides array already, that's why we use merge
+oc patch clusterversions.config.openshift.io/version --type merge -p '{"spec":{"overrides": [{"kind": "Deployment","name": "machine-config-operator", "namespace": "openshift-machine-config-operator", "unmanaged": true}, {"kind": "ConfigMap","name": "machine-config-operator-images", "namespace": "openshift-machine-config-operator", "unmanaged": true}]}}'
+
 if ! oc get -n openshift-image-registry route/image-registry &>/dev/null; then
     oc expose -n openshift-image-registry svc/image-registry
 fi
--- a/hack/cluster-push.sh
+++ b/hack/cluster-push.sh
@@ -14,7 +14,7 @@ if [ "${1:-}" = "-n" ]; then
 fi
 
 registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq -r ".spec.host")
-curl -k --head https://"${registry}" >/dev/null
+curl -k -s --head https://"${registry}" >/dev/null
 
 WHAT=${WHAT:-machine-config-daemon}
 LOCAL_IMGNAME=localhost/${WHAT}


HACKING.md
```diff
@@ -123,7 +123,15 @@ You can replace `daemonset/machine-config-daemon` with any other MCO component.
...
 If you watch the pods, you'll notice your component restarting as well.
-Note this still requires [disabling the CVO](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#disabling-the-cluster-version-operator).
+
+Note this still requires [disabling the CVO](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#disabling-the-cluster-version-operator)
+or setting the needed objects to unmanaged (`deployment/machine-config-operator`) in the CVO.
+In case you're working on the `server`, `controller` or `daemon`, It also requires disabling the operator
+since the daemonsets and deployments will be overwritten:
+
+```
+oc scale deployment machine-config-operator --replicas=0
```
cgwalters: But...we want to be able to develop on the operator too right?

runcom: setting unmanaged for the operator itself is already enough to change the image for the operator and and the deployment itself w/o the CVO stomping on us. If you read some lines above, this is just for the server,controller and daemon (unless I didn't make that clear)

runcom: (I'm retesting this flow on a new cluster though)

runcom: alright, worked as I described, if you overrides the Deployment for the operator, you can edit it w/o doing anything else (nor scaling) as the CVO ignores any change to it. e.g you can `make deploy-operator` just fine or `oc edit deployment/machine-config-operator` and the CVO doesn't stomp on you

HACKING.md
```diff
@@ -123,7 +123,15 @@ You can replace `daemonset/machine-config-daemon` with any other MCO component.
 If the componenet already has a `--v=` flag set, the command above still adds
 verbose and ignores the previous one already set.
 If you watch the pods, you'll notice your component restarting as well.
-Note this still requires [disabling the CVO](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#disabling-the-cluster-version-operator).
+
+Note this still requires [disabling the CVO](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#disabling-the-cluster-version-operator)
+or setting the needed objects to unmanaged (`deployment/machine-config-operator`) in the CVO.
+In case you're working on the `server`, `controller` or `daemon`, it also requires disabling the operator
+since the daemonsets and deployments will be overwritten:
```
cgwalters: But that's why the current `make deploy-` model bounces (scales down+up) the operator - because we don't wait to wait for it to notice the images change.  (Though the slowness there must really be a bug right?)

I think this text is incorrect - we shouldn't recommend disabling the operator when developing things as some functionality depends on it...my recent osImageURL work depended on patching both the operator and daemon for example.

runcom: I'm fine with that indeed, we need to remove that wording from the section below as well though and since i.e. the daemon isn't CVO managed there's no real way to edit it without the operator replacing it with its own pristine copy

HACKING.md
cgwalters: I'm in favor of using overrides to be clear.  My objection is to recommending scaling the operator.

HACKING.md
cgwalters: > the daemon isn't CVO managed

The daemon is managed by the operator.

> there's no real way to edit it without the operator replacing it with its own pristine copy

That's not true; the current `make deploy-` scripts have been doing this just fine right?

Again I think the problem is some sort of race condition where even changing `images.json` the operator doesn't notice the change right away and tries to change the deployment back based on its old cached copy.

HACKING.md
runcom: yup, it makes sense indeed, I just grabbed that from the section below as it's really the way to actually develop at least the daemon w/o the operator messing with it

HACKING.md
runcom: > > there's no real way to edit it without the operator replacing it with its own pristine copy
> 
> That's not true; the current `make deploy-` scripts have been doing this just fine right?

I need to dig deeper but I think, from my testing, the current `make deploy-*` works just because you also patch the images and the operator re-syncing still result in the new built image but if you try editing something else (with the overrides) the daemon will be replaced again

runcom:
Ok, I've dropped the last commit while I keep testing - rest should be ok to review and test I guess

runcom:
I want to be sure that overrides vs disabling the CVO doesn't result in any development regression for any of you, so please, take this PR :P

runcom:
flake

/test e2e-aws

hack/cluster-push-prep.sh
```diff
@@ -9,13 +9,17 @@
 
 set -xeuo pipefail
 
-oc -n openshift-cluster-version scale --replicas=0 deploy/cluster-version-operator
+# XXX: --type merge completely overrides any previous "overrides" array
+#      find a way to just append? json op: add isn't working at all
+#      if there's not an overrides array already, that's why we use merge
+oc patch clusterversions.config.openshift.io/version --type merge -p '{"spec":{"overrides": [{"kind": "Deployment","name": "machine-config-operator", "namespace": "openshift-machine-config-operator", "unmanaged": true}, {"kind": "ConfigMap","name": "machine-config-operator-images", "namespace": "openshift-machine-config-operator", "unmanaged": true}]}}'
```
runcom: I believe we don't need to add the Deployment for the operator here or we're going to miss the wiring with the CVO (status report and whatnot). @cgwalters is this your understanding as well?

runcom:
Still poking with this

/hold

HACKING.md
kikisdeliveryservice: Sorry to jump in, but I'm confused by what you wrote @runcom I dont currently use the hack/* tools but when I work on MC*, I only scale down the CVO, I've never had to do anything to the MCO. Once I scale, my changes to each image "stick" and I have no problems.

runcom: yeah, I was talking about moving away from scaling the CVO to zero and just set some objects to unmanaged!

kikisdeliveryservice: Gotcha ty!

kikisdeliveryservice:

Feel free to reopen.

PR #401, opened 2019-02-11
pkg/daemon: add atomic files writing and remove file system abstraction

runcom:
pkg/daemon: add atomic files writing and remove file system abstraction
    
    This patch makes sure we write files and ssh (file)
    atomically so, if and only if, we'll be able to reconcile a failing machine,
    we don't end up having old files lying around.
    
    Last thing this patch does is getting rid of the FileSystemClient abstraction.
    I've replaced that with normal 1st class function assignements so we don't
    regress in testing. We would need to move those tests to e2e anyway.

Closes https://github.com/openshift/machine-config-operator/issues/101

@cgwalters @kikisdeliveryservice @ashcrow ptal


ashcrow:
I'm cool with replacing the abstraction with functions that can be mocked for testing.

cgwalters:
"fix files leak" is a bit understated of a commit message title :wink: 

"Remove filesystem mock, fix bugs"?  dunno.

Personally I am not a huge fan of the current unit testing we have...I tend to prefer "real" testing and make that as convenient as possible.  When the mocking starts to get nontrivial (as is the case not only with the filesystem but also our mock kube client stuff) I think the value proposition gets a lot more uncertain.

OTOH it can be easier to test error conditions via mocking.  On the other other hand...a few strategic points to inject errors can go a long way.

I think though I'd say we probably shouldn't do this change right now...the bug isn't worth the code churn?  Let's fix the leak in the existing code and keep this around for later?

cgwalters:
Something that did change of course since the unit testing here was introduced is that we have our own `e2e-aws-op` where we can test functionality "for real".

runcom:
> "fix files leak" is a bit understated of a commit message title

added potential, even if, on any error, we end up leaving files around :)



> Personally I am not a huge fan of the current unit testing we have...I tend to prefer "real" testing and make that as convenient as possible. When the mocking starts to get nontrivial (as is the case not only with the filesystem but also our mock kube client stuff) I think the value proposition gets a lot more uncertain.

I belive the abstraction introduced was because there was no e2e, not that we have it, we can indeed test it for real

> I think though I'd say we probably shouldn't do this change right now...the bug isn't worth the code churn? Let's fix the leak in the existing code and keep this around for later?

I'm not sure about, this mainly adds atomic file writing to prevent files laying around, the defer file.Close() isn't really doing nothing per-se (I spotted that later, was already adding atomic file write).

ashcrow:
> I think though I'd say we probably shouldn't do this change right now...the bug isn't worth the code churn? Let's fix the leak in the existing code and keep this around for later?

That sounds reasonable to me.

runcom:
> Something that did change of course since the unit testing here was introduced is that we have our own `e2e-aws-op` where we can test functionality "for real".

this is indeed why I've added Closes #101 as well in the first comment

runcom:
https://github.com/openshift/machine-config-operator/pull/402

runcom:
e2e rocks :rocket:

pkg/daemon/update.go
```diff
@@ -512,58 +515,73 @@ func (dn *Daemon) writeUnits(units []ignv2_2types.Unit) error {
...
+	}
+
+	// set chown if file information is provided
+	if file.User != nil || file.Group != nil {
+		uid, gid, err := getFileOwnership(file)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to retrieve file ownership for file %q: %v", file.Path, err)
 		}
-		if _, err = file.WriteString(string(contents.Data)); err != nil {
```
lucab: Discussed OOB but leaving a note here so the context isn't lost: this will misbehave as soon as the content is a pure binary blob that can't be represented as UTF-8.

runcom: Along with that, we'll have to move to buffered writes as well, I'm spinning up another PR from this one (even if I do think we need this)

runcom: https://github.com/openshift/machine-config-operator/pull/411

pkg/daemon/daemon.go
```diff
@@ -187,6 +185,7 @@ func New(
 		exitCh:                 exitCh,
 		stopCh:                 stopCh,
 	}
+	dn.atomicSSHKeysWriter = dn.atomicallyWriteSSHKey
```
kikisdeliveryservice: Could you explain what this is for?

runcom: this is a lightweight version of the filesystem client, as in, the client existed mainly to be able to unit test the flow as we shouldn't wrap the file management API of Golang. By having a function which deals with writes only, we're able to mock it out in unit test till we have proper e2e in place for these scenarios.

kikisdeliveryservice: ah thanks! I missed the note on line 87.

runcom:
(rebased :smiling_imp: )

pkg/daemon/update.go
```diff
@@ -618,26 +679,18 @@ func (dn *Daemon) updateSSHKeys(newUsers []ignv2_2types.PasswdUser) error {
...
-	sshDirPath := filepath.Join("/home", newUsers[0].Name, ".ssh")
-	// we are only dealing with the "core" User at this time, so only dealing with the first entry in Users[]
-	glog.Infof("Writing SSHKeys at %q", sshDirPath)
-	if err := dn.fileSystemClient.MkdirAll(filepath.Dir(sshDirPath), os.FileMode(0600)); err != nil {
-		return fmt.Errorf("Failed to create directory %q: %v", filepath.Dir(sshDirPath), err)
-	}
-	glog.V(2).Infof("Created directory: %s", sshDirPath)
 
-	authkeypath := filepath.Join(sshDirPath, "authorized_keys")
+	// we're also appending all keys for any user to core, so for now
```
kikisdeliveryservice: Not quite: "we are appending all new SSHAuthorizedKeys of user core and passing to atomicallyWriteSSHKeys to write."

runcom: mmm https://github.com/openshift/machine-config-operator/pull/401/files#diff-06961b075f1753956d802ba954d2cfb5R686 seems to suggest that we're ranging through every user in Users and grabbind their ssh keys as well

kikisdeliveryservice: no we are ranging over the SSHAuthorizedKeys

runcom: well, no the last one actually

runcom: there's no check around `len(newUsers)` so that range can be wrong right?

kikisdeliveryservice: We are looking at the last user in new Users(as per Ignition) and ranging over its SSHKeys. There is a len check in updateSSHKeys the first or second line.

runcom: > no we are ranging over the SSHAuthorizedKeys

 yeah, I was wrong it's `SSHAuthorizedKeys`. My point being though, we're checking for the very first user to be `core` but if `newUsers` contains more than 1 entry, we're ending up appending keys from `len(newUsers)-1` User, not `core` (is it possible?)

runcom: alrighty, thanks, just saw your previous comment :smile:

kikisdeliveryservice: Yeah I have a cleanup PR taking that erroneous [0] out :)

runcom: no wait, the len is for newUsers to be empty, but what if it has more than 1 user? we'd fall into https://github.com/openshift/machine-config-operator/pull/401#discussion_r256176416

kikisdeliveryservice: I think it's prob better to go over this with you outside of this pr discussion thread :)

tho i'd like this pr to wait until my cleanup pr hits and you can rebase on it.  though it sounds like this one is going to be held anyway?

runcom: well, sure, I can def wait for you PR to land first. As for holding this up, for now yea, but it's still something we need to have imo (nodes can go down, lose connection and stuff like that)

ashcrow:
Rebase needed

runcom:
rebased finally

runcom:
this is now writes atomically for systemd units and dropins as well, I've greatly reduced the code changes as well by reusing and extending functions.

runcom:
Unit flake tracked here: https://github.com/openshift/machine-config-operator/issues/449

/retest

runcom:
I'm moving this to `master-4.1` once I understand how I can rebase that branch -.-

cgwalters:
Can you push the rebased commit here?

runcom:
> Can you push the rebased commit here?

just done

cgwalters:
/approve

This LGTM.

runcom:
> /approve
> 
> This LGTM.

I guess we can let this go in with #548 as that is based on this now.

Issue #367, opened 2019-02-01
master machineconfig pool reports degraded on new cluster


sjenning:
Just installed the cluster and MCO reports failure

```
$ oc get clusteroperators machine-config-operator
NAME                      VERSION   AVAILABLE   PROGRESSING   FAILING   SINCE
machine-config-operator             False       True          True      1h

$ oc get clusteroperators machine-config-operator -o yaml
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-02-01T20:19:35Z
  generation: 1
  name: machine-config-operator
  resourceVersion: "47718"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/machine-config-operator
  uid: b189665d-265e-11e9-9d7f-06538a4d7af6
spec: {}
status:
  conditions:
  - lastTransitionTime: 2019-02-01T20:19:35Z
    status: "False"
    type: Available
  - lastTransitionTime: 2019-02-01T20:19:35Z
    message: Progressing towards 3.11.0-543-g6c3e3e6a-dirty
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-02-01T20:25:18Z
    message: 'Failed when progressing towards 3.11.0-543-g6c3e3e6a-dirty because:
      error syncing: timed out waiting for the condition during syncRequiredMachineConfigPools:
      error pool master is not ready. status: (total: 3, updated: 0, unavailable:
      1)'
    reason: 'error syncing: timed out waiting for the condition during syncRequiredMachineConfigPools:
      error pool master is not ready. status: (total: 3, updated: 0, unavailable:
      1)'
    status: "True"
    type: Failing
  extension:
    master: pool is degraded because of 1 nodes are reporting degraded status on update.
      Cannot proceed.
    worker: all 3 nodes are at latest configuration worker-9f1b04fd7540807d1cad9739722e3ba5
  relatedObjects: null
  versions: null

$ oc get machineconfigpool master -o yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfigPool
metadata:
  creationTimestamp: 2019-02-01T20:19:35Z
  generation: 1
  labels:
    operator.machineconfiguration.openshift.io/required-for-upgrade: ""
  name: master
  resourceVersion: "11941"
  selfLink: /apis/machineconfiguration.openshift.io/v1/machineconfigpools/master
[...]

PR #403, opened 2019-02-11
bootstrap: Require images on command line

cgwalters:
Since having a mismatch here will result in a broken cluster,
change things so that the command line arguments are required
and drop the `docker.io` image references.  We never want
to pull those into a real cluster.

Ref: https://github.com/openshift/machine-config-operator/issues/367

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


pkg/operator/images.go
```diff
@@ -12,12 +12,5 @@ type Images struct {
 
 // DefaultImages returns default set of images for operator.
 func DefaultImages() Images {
```
runcom: This is called from
```
244 func (optr *Operator) sync(key string) error {                                  
```
do we need to store the images somewhere in the operator to return them there?

runcom:
copy pasted `/lgtm` wrongly while reviewing other PRs lol but apart from a comment it looks good

/lgtm cancel

cgwalters:
And now if anyone wants to add a new MCO-managed image, they need to:

 - Add a CLI option here with a default
 - Change the installer to pass that arg with the image from the CVO
 - Make the option required here

Or we change this bootstrap code to just parse the CVO payload directly which IMO would be saner, but this is the current status.

cmd/machine-config-operator/bootstrap.go
```diff
@@ -48,11 +48,17 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.destinationDir, "dest-dir", "", "The destination directory where MCO writes the manifests.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.imagesConfigMapFile, "images-json-configmap", "", "ConfigMap that contains images.json for MCO.")
```
jlebon: This implies we can also drop the `images-json-configmap`, right? (And drop the "this overrides the image from --images-json-configmap" texts).

pkg/operator/images.go
ashcrow: nit: This is not technically true any longer. I'd say we should either nuke this function all together and move to using `Images{}` *or* rename it to something like `NewImages()` or the like.

runcom: +1 for using the raw literal

cgwalters:
OK updated :arrow_up:

runcom:
these are really aws running out of resources, let's hold retesting for a while

ashcrow:
> these are really aws running out of resources, let's hold retesting for a while

:man_facepalming:

PR #404, opened 2019-02-11
image-references: Update maipo oscontainer

cgwalters:
This isn't the long term path, but let's do this until
we have that implemented.

### DIFF
--- a/install/image-references
+++ b/install/image-references
@@ -33,4 +33,4 @@ spec:
   - name: machine-os-content
     from:
       kind: DockerImage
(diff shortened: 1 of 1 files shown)


ashcrow:
```
aws_route_table_association.route_net.1: timeout while waiting for state to become 'success'
```

/test e2e-aws

jlebon:
Can we add the corresponding OS version as a comment there?

cgwalters:
For reference I was confused about how things work; the images referenced here are basically never really used...we should probably replace them with dummy values now.

The source of truth (AIUI) is the imagestreams maintained at api.ci which feed into the release controller, which uses them to update the release payload's `machine-os-content` and `oc adm release new` will use that digested pull spec and replace the value here.

https://github.com/openshift/origin/pull/21998 was trying to do this a different way but for now we're going to do some manual tagging.

PR #406, opened 2019-02-11
pkg/operator: correctly sync status for the CVO

### LINKED ISSUE #346: MCO Cluster Operator statuses frequently changing

runcom:
Based on:

- https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions
- Slack conversation with Clayton

We should correctly report Progressing only when really progressing towards something. Re-syncing, reconciling, drift detections verifying everything is still where you left it is not _changing_ something, thus is not progressing.

closes #346

/cc @smarterclayton @abhinavdahiya ptal

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -28,15 +28,22 @@ func (optr *Operator) syncAvailableStatus() error {
 	}
 
 	optrVersion, _ := optr.vStore.Get("operator")
+	progressing := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing)
+	failing := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorFailing)
+	message := fmt.Sprintf("Cluster has deployed %s", optrVersion)
+
+	available := configv1.ConditionTrue
+
+	if failing && !progressing {
+		available = configv1.ConditionFalse
+		message = fmt.Sprintf("Cluster not available for %s", optrVersion)
+	}
+
 	// set available
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
-		Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue,
-		Message: fmt.Sprintf("Cluster is available at %s", optrVersion),
+		Type: configv1.OperatorAvailable, Status: available,
+		Message: message,
 	})
-	// clear progressing
-	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse})
-	// clear failure
-	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorFailing, Status: configv1.ConditionFalse})
 
 	co.Status.Versions = optr.vStore.GetAll()
 	optr.setMachineConfigPoolStatuses(&co.Status)
@@ -55,15 +62,20 @@ func (optr *Operator) syncProgressingStatus() error {
 	}
 
 	optrVersion, _ := optr.vStore.Get("operator")
-	var message string
+	progressing := configv1.ConditionFalse
+	message := fmt.Sprintf("Cluster version is %s", optrVersion)
+
 	if optr.vStore.Equal(co.Status.Versions) {
-		// syncing the state to existing version.
-		message = fmt.Sprintf("Running resync for %s", optrVersion)
+		if optr.inClusterBringup {
+			progressing = configv1.ConditionTrue
+		}
 	} else {
-		message = fmt.Sprintf("Progressing towards %s", optrVersion)
+		message = fmt.Sprintf("Working towards %s", optrVersion)
+		progressing = configv1.ConditionTrue
 	}
+
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
-		Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue,
+		Type: configv1.OperatorProgressing, Status: progressing,
 		Message: message,
 	})
 
@@ -91,7 +103,7 @@ func (optr *Operator) syncFailingStatus(ierr error) error {
 		// syncing the state to exiting version.
 		message = fmt.Sprintf("Failed to resync %s because: %v", optrVersion, ierr.Error())
 	} else {
-		message = fmt.Sprintf("Failed when progressing towards %s because: %v", optrVersion, ierr.Error())
+		message = fmt.Sprintf("Unable to apply %s: %v", optrVersion, ierr.Error())
 	}
 	// set failing condition
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -47,7 +47,6 @@ func (optr *Operator) syncAll(rconfig renderConfig) error {
 		if optr.inClusterBringup {
 			glog.Infof("[init mode] synced %s in %v", sf.name, time.Since(startTime))
 		}
-		optr.syncProgressingStatus()
 	}
 
 	agg := utilerrors.NewAggregate(errs)
@@ -55,7 +54,8 @@ func (optr *Operator) syncAll(rconfig renderConfig) error {
 		errs = append(errs, optr.syncFailingStatus(agg))
 		agg = utilerrors.NewAggregate(errs)
 		return fmt.Errorf("error syncing: %v", agg.Error())
-	} else if optr.inClusterBringup {
+	}
+	if optr.inClusterBringup {
 		glog.Infof("Initialization complete")
 		optr.inClusterBringup = false
 	}


runcom:
not entirely sure how to test this out on a running cluster, disabling the CVO isn't gonna work, setting the operator Deployment to unmanaged isn't gonna work, any idea?

cgwalters:
If you want to test things like this cleanly it basically requires building a custom payload and passing it to the installer.

runcom:
> If you want to test things like this cleanly it basically requires building a custom payload and passing it to the installer.

it's probably time I start doing this

pkg/operator/status.go
```diff
@@ -31,12 +31,8 @@ func (optr *Operator) syncAvailableStatus() error {
 	// set available
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
 		Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue,
-		Message: fmt.Sprintf("Cluster is available at %s", optrVersion),
+		Message: fmt.Sprintf("Cluster has deployed %s", optrVersion),
 	})
-	// clear progressing
```
abhinavdahiya: Why this drop?

runcom: setting Available doesn't necessarily clear progressing (taken from docs https://github.com/openshift/installer/blob/master/docs/user/overview.md)
```
Failing is true with a detailed message Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.
Available is true with message Cluster has deployed 4.0.0
Progressing is true with message Unable to apply 4.0.1: a required object is missing
```

It also looks like something other operators are doing (setting Avaiable can still be true while Failing and Progressing are also true)

runcom: The other scenarios I've seen in other operators set the Conditions indipendently from the other conditions as well (which makes sense I believe)

abhinavdahiya: > setting Available doesn't necessarily clear progressing (taken from docs https://github.com/openshift/installer/blob/master/docs/user/overview.md)
> 
> ```
> Failing is true with a detailed message Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.
> Available is true with message Cluster has deployed 4.0.0
> Progressing is true with message Unable to apply 4.0.1: a required object is missing
> ```
> 
> It also looks like something other operators are doing (setting Avaiable can still be true while Failing and Progressing are also true)

you are quoting an example when operator is progressing or failing...
when the operator is available
```
```

runcom: I should check the other Conditions I guess right

runcom: But still, why there's an example of all three to True then? Thats really confusing.

runcom: updated

runcom: anyway, the status you quoted can be already converge to without my latest additions (though they're indeed needed). Progressing and Failing are set _before_ available anyway so when we're at Avaiable, we know where we are (I think)

pkg/operator/sync.go
```diff
@@ -47,15 +47,15 @@ func (optr *Operator) syncAll(rconfig renderConfig) error {
...
 	}
 
 	agg := utilerrors.NewAggregate(errs)
 	if agg != nil {
 		errs = append(errs, optr.syncFailingStatus(agg))
 		agg = utilerrors.NewAggregate(errs)
 		return fmt.Errorf("error syncing: %v", agg.Error())
-	} else if optr.inClusterBringup {
+	}
+	if optr.inClusterBringup {
```
cgwalters: Why this change?  The intention was to stay in "initialization" stage until we'd done a successful sync.  

(Not saying it's wrong, I just want to understand why you're making the change)

pkg/operator/status.go
```diff
@@ -28,15 +28,25 @@ func (optr *Operator) syncAvailableStatus() error {
 	}
 
 	optrVersion, _ := optr.vStore.Get("operator")
+
+	var progressing bool
+	if !optr.vStore.Equal(co.Status.Versions) {
+		progressing = true
```
cgwalters: One thing I'm wondering here...we want `progressing=true` when `optr.inClusterBringup` too right?
IOW this code is handling the upgrade case but we also want to be progressing during initial install.

pkg/operator/sync.go
runcom: just golang linting, the previous if branch has a return at the end, the `else` branch is therefore superfluous

pkg/operator/status.go
runcom: good call, I'm gonna work that out

runcom: fixed here https://github.com/openshift/machine-config-operator/pull/406/files#diff-bee1f8f36240cb95db10244fdf335146R68 - we now set progressing = true on bring up as well

pkg/operator/status.go
runcom: fixed here I believe now https://github.com/openshift/machine-config-operator/pull/406/files#diff-bee1f8f36240cb95db10244fdf335146R37

runcom:
This is good to review and test now, I was able to test it out using a custom payload built using https://github.com/openshift/machine-config-operator/issues/421#issuecomment-463392262 and I can't notice the statuses bouncing anymore.

runcom:
@abhinavdahiya could you also take another look at this now?

pkg/operator/status.go
cgwalters: One thing I've noticed related to this though is that we seem to stay in `inClusterBringup` for multiple times.  Though maybe that's only failing PRs?  Glancing at [the logs in a recent merged PR https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/419/pull-ci-openshift-machine-config-operator-master-e2e-aws/1756/artifacts/e2e-aws/pods/] things look fine.

pkg/operator/status.go
```diff
@@ -28,15 +28,21 @@ func (optr *Operator) syncAvailableStatus() error {
 	}
 
 	optrVersion, _ := optr.vStore.Get("operator")
+	progressing := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing)
+	failing := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorFailing)
+	message := fmt.Sprintf("Cluster has deployed %s", optrVersion)
+	status := configv1.ConditionTrue
```
cgwalters: Let's name this variable `available`

pkg/operator/status.go
```diff
@@ -55,15 +61,20 @@ func (optr *Operator) syncProgressingStatus() error {
 	}
 
 	optrVersion, _ := optr.vStore.Get("operator")
-	var message string
+	status := configv1.ConditionFalse
```
cgwalters: Ditto.

pkg/operator/status.go
```diff
@@ -55,15 +61,20 @@ func (optr *Operator) syncProgressingStatus() error {
...
+	message := fmt.Sprintf("Cluster version is %s", optrVersion)
+
 	if optr.vStore.Equal(co.Status.Versions) {
-		// syncing the state to existing version.
-		message = fmt.Sprintf("Running resync for %s", optrVersion)
+		if optr.inClusterBringup {
+			status = configv1.ConditionTrue
+		}
```
cgwalters: Wait why would we be `available = false` if the version store is equal and `!inClusterBringup`?

pkg/operator/status.go
runcom: mmm what do you mean "stay in `inClusterBringup` for multiple times"?

cgwalters: If the sync fails; you'll see that log multiple times.  But I think that's the right thing?

pkg/operator/status.go
runcom: this is setting Progressing to false, not available. If `!inClusterBringup && version-equal`, `available is gonna be true (assuming failing is false as well). Am I missing something?

cgwalters: Sorry, yeah.  I guess it's clearer when viewing the surrounding code as the functions clearly say `syncProgressingStatus` and `syncAvailableStatus`.

So nevermind, LGTM!

pkg/operator/status.go
runcom: fixed

pkg/operator/status.go
runcom: fixed

pkg/operator/status.go
runcom: oh yeah, it's correct we keep re-syncing and logging that

abhinavdahiya:
@runcom @cgwalters 
With this PR if MCO ever fails it will never go available again because of
https://github.com/openshift/machine-config-operator/pull/406/files#diff-bee1f8f36240cb95db10244fdf335146R37

example:
https://gubernator.k8s.io/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.0/4246
MCO is reporting failing...
but if you look at https://storage.cloud.google.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.0/4246/artifacts/release-e2e-aws/pods/openshift-machine-config-operator_machine-config-operator-7888cd6-c8wp4_machine-config-operator.log.gz

mco looses the connection to apiserver interminently when reconciling and then reports the machine-config failing.

once MCO has reported failing, even on a successful reconcile later on calling syncAvailbleStatus [here](https://github.com/openshift/machine-config-operator/pull/406/files#diff-fa45321336db7ad1cedc28bf643a4f97R63) will still report failing because of that `if` condition I mentioned above.

Please look into this.

abhinavdahiya:
I also suspect this was similar error https://gubernator.k8s.io/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.0/4197

runcom:
Thanks @abhinavdahiya I'm fixing that (I was also thinking about making all of this simpler as well but later).

runcom:
@abhinavdahiya please take a look at https://github.com/openshift/machine-config-operator/pull/442

PR #408, opened 2019-02-11
daemon: Clearly log if error is from draining

cgwalters:
Saw this in a log:

```
I0211 21:20:46.924255   61902 daemon.go:660] Unable to apply update: rpc error: code = Unknown desc =
```

It must be from the drain; let's make that clear.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -82,11 +82,13 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 
 		dn.recorder.Eventf(node, corev1.EventTypeNormal, "Drain", "Draining node to update config.")
 
-		err = wait.ExponentialBackoff(wait.Backoff{
+		backoff := wait.Backoff{
 			Steps:    5,
 			Duration: 10 * time.Second,
 			Factor:   2,
-		}, func() (bool, error) {
+		}
+		var lastErr error
+		wait.ExponentialBackoff(backoff, func() (bool, error) {
 			err := drain.Drain(dn.kubeClient, []*corev1.Node{node}, &drain.DrainOptions{
 				DeleteLocalData:    true,
 				Force:              true,
@@ -94,13 +96,14 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 				IgnoreDaemonsets:   true,
 			})
 			if err != nil {
-				glog.Infof("Draining failed with: %v; retrying...", err)
+				lastErr = err
 				return false, nil
 			}
+			lastErr = nil
 			return true, nil
 		})
-		if err != nil {
-			return err
+		if lastErr != nil {
+			return errors.Wrapf(lastErr, "Failed to drain node (%s tries)", backoff.Steps)
 		}
 		glog.V(2).Info("Node successfully drained")
 	}


pkg/daemon/update.go
```diff
@@ -100,7 +101,7 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 			return true, nil
 		})
 		if err != nil {
```
runcom: I think we're really masking the error here from reporting if this times out, we should set err from line 91 to something like lastErr and check if we timed out here and report also that (otherwise `err == wait.ErrWaitTimeout` and that's gonna be it)

cgwalters:
/lgtm cancel

based on https://github.com/openshift/machine-config-operator/pull/408#discussion_r255720389

pkg/daemon/update.go
cgwalters: Yeah...we need to wrap up that pattern in a helper function.

runcom:
pretty sure the error reported on Slack comes from:

```
func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
	if err := dn.updateOS(newConfig); err != nil {
		return err
	}

	// Skip draining of the node when we're not cluster driven
	if dn.onceFrom == "" {
		glog.Info("Update prepared; draining the node")

		node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
		if err != nil {
			return err
		}
```

Otherwise we should have gotten a timeout error right?

pkg/daemon/update.go
ashcrow: Good catch!

jlebon:
Yeah, I agree the error is likely from the `Nodes().Get()`. See #409.

pkg/daemon/update.go
runcom: opened https://github.com/openshift/machine-config-operator/issues/410 to track this

cgwalters: OK, fixed for now here too.

jlebon: I think I'm missing something here. If this times out, then we're propagating it up already, right? Note the `err` in the closure shadows the outer `err`, though I agree we should name them differently to make it more obvious.

pkg/daemon/update.go
```diff
@@ -82,25 +82,28 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
...
+		var lastErr error
+		wait.ExponentialBackoff(backoff, func() (bool, error) {
 			err := drain.Drain(dn.kubeClient, []*corev1.Node{node}, &drain.DrainOptions{
 				DeleteLocalData:    true,
 				Force:              true,
 				GracePeriodSeconds: 600,
 				IgnoreDaemonsets:   true,
 			})
 			if err != nil {
-				glog.Infof("Draining failed with: %v; retrying...", err)
```
jlebon: Why not keep the logging here to show some progress if we're retrying?

pkg/daemon/update.go
```diff
@@ -82,25 +82,28 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
...
 				return false, nil
 			}
+			lastErr = nil
 			return true, nil
 		})
-		if err != nil {
-			return err
+		if lastErr != nil {
+			return errors.Wrapf(lastErr, "Failed to drain node (%s tries)", backoff.Steps)
 		}
```
runcom: we're gonna miss the `wait.*` error this way cause other err may happen other than just timeout
```
// If the condition never returns true, ErrWaitTimeout is returned. All other
// errors terminate immediately.
func ExponentialBackoff(backoff Backoff, condition ConditionFunc) error {
```
I was thinking something like https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L357-L362

runcom:
just a comment which is really a nit since we control the ConditionFunc

/lgtm

pkg/daemon/update.go
runcom: this is not blocking of course, we control the ConditionFunc here

pkg/daemon/update.go
jlebon: Follow up for this in #412!

Issue #395, opened 2019-02-07
Strengthen error-handling when dealing with transient errors


jlebon:
The MCD should not take lightly the decision to degrade a node. All I/O operations subject to a higher possibility of transient failures should be retried (e.g. API requests, file downloads, waiting for events, etc...). We need to (1) audit the whole MCD to make sure this is the case, and/or (2) add some higher-level retry logic to make sure we catch all instances and don't regress on it as new code is added.

runcom:
we could do some nice type checking on go errors for this but I fear we'll need some wrapping anyway

derekwaynecarr:
log from a variation of a similar problem:

```sh
oc logs machine-config-daemon-cgwxw -n openshift-machine-config-operator -p
I0211 19:16:42.450242   66299 start.go:52] Version: 3.11.0-595-g943ed14c-dirty
I0211 19:16:42.451043   66299 start.go:88] starting node writer
I0211 19:16:42.455490   66299 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0211 19:16:42.526574   66299 daemon.go:166] Booted osImageURL: registry.svc.ci.openshift.org/rhcos/maipo@sha256:83a6d461628380f1dcc057ffef4909992f593b72c5aa4db2e01c0b130004afea (47.312)
I0211 19:16:42.526828   66299 daemon.go:238] Managing node: ip-10-0-16-157.ec2.internal
I0211 19:16:42.547911   66299 start.go:146] Calling chroot("/rootfs")
I0211 19:16:42.577111   66299 daemon.go:443] Current+desired config: master-85b614cf21435f01ecfb1e2d3d972e2a
I0211 19:16:42.577305   66299 daemon.go:797] No target osImageURL provided
I0211 19:16:42.578685   66299 daemon.go:558] Validated on-disk state
I0211 19:16:42.578706   66299 daemon.go:590] In desired config master-85b614cf21435f01ecfb1e2d3d972e2a
I0211 19:16:42.578724   66299 start.go:165] Starting MachineConfigDaemon
I0211 19:16:42.578730   66299 daemon.go:259] Enabling Kubelet Healthz Monitor
E0211 19:19:44.611810   66299 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=89, ErrCode=NO_ERROR, debug=""
[...]

PR #409, opened 2019-02-11
daemon: Add GetNode() wrapper which retries on errors

jlebon:
I think this is what's causing the issue in:
https://github.com/openshift/machine-config-operator/issues/395#issuecomment-462482242

### DIFF
--- a/pkg/daemon/node.go
+++ b/pkg/daemon/node.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
+	"github.com/pkg/errors"
 	core_v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -49,19 +50,32 @@ func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (stri
 	return getNodeAnnotationExt(client, node, k, false)
 }
 
-// getNodeAnnotationExt is like getNodeAnnotation, but allows one to customize ENOENT handling
-func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allowNoent bool) (string, error) {
+// GetNode gets the node object.
+func GetNode(client corev1.NodeInterface, node string) (*core_v1.Node, error) {
+	var lastErr error
 	var n *core_v1.Node
 	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
-		var err error
-		n, err = client.Get(node, metav1.GetOptions{})
-		if err == nil {
+		n, lastErr = client.Get(node, metav1.GetOptions{})
+		if lastErr == nil {
 			return true, nil
 		}
+		glog.Warningf("Failed to fetch node %s (%v); retrying...", node, lastErr)
 		return false, nil
 	})
 	if err != nil {
-		return "", err
+		if err == wait.ErrWaitTimeout {
+			return nil, errors.Wrapf(lastErr, "Timed out trying to fetch node %s", node)
+		}
+		return nil, err
+	}
+	return n, nil
+}
+
+// getNodeAnnotationExt is like getNodeAnnotation, but allows one to customize ENOENT handling
+func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allowNoent bool) (string, error) {
+	n, err := GetNode(client, node)
+	if err != nil {
+		return "", fmt.Errorf("Failed fetching node %s: %v", node, err)
 	}
 
 	v, ok := n.Annotations[k]
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -681,7 +681,7 @@ func (dn *Daemon) prepUpdateFromCluster() (bool, error) {
 	}
 
 	// Grab the node instance
-	node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
+	node, err := GetNode(dn.kubeClient.CoreV1().Nodes(), dn.name)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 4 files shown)


jlebon:
(Not tested yet.)

cgwalters:
I feel like we should acquire this once in the daemon around `getStateAndConfigs` or so.  It's not like anything in it is going to change that we care about.

runcom:
> I feel like we should acquire this once in the daemon around `getStateAndConfigs` or so. It's not like anything in it is going to change that we care about.

aren't we going to cache annotation this way? (unless I'm not understanding the comment)

cgwalters:
> aren't we going to cache annotations this way? (unless I'm not understanding the comment)

Hmm.  Right.  We have a node lister we should probably be using.

Also looking over the code in the daemon we have many places that `.Get()` our own node:

```

$ git grep Nodes.*Get
daemon/daemon.go:               if err := dn.nodeWriter.SetUpdateDone(dn.kubeClient.CoreV1().Nodes(), dn.name, state.pendingConfig.GetName()); err != nil {
daemon/daemon.go:       node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
daemon/daemon.go:       node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
daemon/update.go:               node, err := dn.kubeClient.CoreV1().Nodes().Get(dn.name, metav1.GetOptions{})
```

jlebon:
Yup, I force-pushed an update so we do this everywhere.

jlebon:
Pushed an update which follows the pattern in #410. Still trying to test this though. My worker node for some reason is not coming up.

/hold

pkg/daemon/node.go
```diff
@@ -49,19 +49,32 @@ func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (stri
...
 		}
+		glog.Warningf("Failed to fetch node %s (%v); retrying...", node, lastErr)
 		return false, nil
 	})
 	if err != nil {
-		return "", err
+		if err == wait.ErrWaitTimeout {
+			glog.Warningf("Timed out trying to fetch node %s; last error: %v", lastErr)
+		}
+		return nil, lastErr
```
runcom: so, this is fine, but we can still wrap the time out error https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L397-L400

jlebon: Ahh OK, I thought the goal of #410 was keeping the same error so it could be matched higher up in the stack. If we wrap it, we're losing information, right?

jlebon: As an example, the instance in `updateNodeRetry()` is matching against `errors.IsConflict(err)`.

runcom: well, mmm yeah, wrapping is something we still need to do for these kinds of error. Example: say this times out for some reason, callers want to know 1) is it a timeout? 2) what was the underlying error? 
The code right now isn't telling callers that we actually timed out right? we need to bubble that up to make logic around permanent/temporarly failures (?)

The example here https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L397-L400 is telling you both 1) && 2) even we should start using `pkg/errors.Wrap` which has a `.Cause()` method which will tell us more about the wrapped error

runcom: I'm talking about this https://github.com/pkg/errors#adding-context-to-an-error

jlebon: Ahh yes, `Cause()` is exactly what I mean. I didn't realize that's what you meant since we don't currently do that in https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L397-L400. K, will update to use `Wrap`!

runcom: > Ahh yes, `Cause()` is exactly what I mean. I didn't realize that's what you meant since we don't currently do that in https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L397-L400. K

well, by wrapping in that snippet I meant that we were still _wrapping_ the timeout err with lastErr https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L398

jlebon:
My testing this is blocked on https://github.com/openshift/machine-api-operator/issues/205. Trying to pull up an older cluster.

jlebon:
Test failure is:

> F0212 18:30:50.637693       1 api.go:59] Machine Config Server exited with error: listen tcp :49500: bind: address already in use

which should be fixed now that https://github.com/openshift/installer/pull/1180 is merged.

/test e2e-aws-op

jlebon:
`e2e-aws` passed, which seems to indicate that the happy path works at least.

jlebon:
/hold cancel

Tested this now (albeit on an older cluster).

runcom:
:+1: great

cgwalters:
I'm not opposed to this but I still think we should be using Informers at least eventually.

runcom:
> I'm not opposed to this but I still think we should be using Informers at least eventually.

can we track this somewhere in an issue maybe? just to know we have ideas on how to better tackle this issue

ashcrow:
AWS resource problems. Let's wait a bit before we issue a retest.

ashcrow:
Flake

/retest

pkg/daemon/node.go
```diff
@@ -49,19 +50,32 @@ func getNodeAnnotation(client corev1.NodeInterface, node string, k string) (stri
 	return getNodeAnnotationExt(client, node, k, false)
 }
 
-// getNodeAnnotationExt is like getNodeAnnotation, but allows one to customize ENOENT handling
-func getNodeAnnotationExt(client corev1.NodeInterface, node string, k string, allowNoent bool) (string, error) {
+// GetNode gets the node object.
+func GetNode(client corev1.NodeInterface, node string) (*core_v1.Node, error) {
```
runcom: you can lowercase the func name here as it's only used within `pkg/daemon`

jlebon: Hmm, true. Though WDYT about getting this in, and we just work on https://github.com/openshift/machine-config-operator/pull/409#issuecomment-463372473 as a follow up?

runcom: That's ok as well (though I'm missing context on how to use informers for these scenarios) /me goes to learn

runcom:
wtf is going on

PR #411, opened 2019-02-12
pkg/daemon: write bytes, use buffered writer and compare bytes

runcom:
`contents.Data` is a byte slice and can be anything (i.e. a binary after
    deconding). Use a bytes writer to prevent misbehaves in such scenarios.
    Along with that, use a buffered writer around the file in case we're
    dealing with big files. Also, compare byte slices as well instead of
    using strings.
    
### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -806,7 +807,7 @@ func checkUnits(units []ignv2_2types.Unit) bool {
 	for _, u := range units {
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
-			if status := checkFileContentsAndMode(path, u.Dropins[j].Contents, DefaultFilePermissions); !status {
+			if status := checkFileContentsAndMode(path, []byte(u.Dropins[j].Contents), DefaultFilePermissions); !status {
 				return false
 			}
 		}
@@ -827,7 +828,7 @@ func checkUnits(units []ignv2_2types.Unit) bool {
 				return false
 			}
 		}
-		if status := checkFileContentsAndMode(path, u.Contents, DefaultFilePermissions); !status {
+		if status := checkFileContentsAndMode(path, []byte(u.Contents), DefaultFilePermissions); !status {
 			return false
 		}
 
@@ -854,7 +855,7 @@ func checkFiles(files []ignv2_2types.File) bool {
 			glog.Errorf("couldn't parse file: %v", err)
 			return false
 		}
-		if status := checkFileContentsAndMode(f.Path, string(contents.Data), mode); !status {
+		if status := checkFileContentsAndMode(f.Path, contents.Data, mode); !status {
 			return false
 		}
 		checkedFiles[f.Path] = true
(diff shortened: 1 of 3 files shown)


jlebon:
Good catch! /lgtm

I think we need the same thing in `checkFileContentsAndMode()`.

ashcrow:
/hold

@runcom do you want to update ``checkFileContentsAndMode()` in this PR or keep them separate?

runcom:
> @runcom do you want to update ``checkFileContentsAndMode()` in this PR or keep them separate?

I'll push to this PR

runcom:
pushed with changes to the other func, ptal

ashcrow:
All unit tests passed, but reported failure :confused: 

/test unit

ashcrow:
Ahh the failure was only noted in the full log:

```
ok  	github.com/openshift/machine-config-operator/lib/resourceread	0.047s
make: *** [test-unit] Error 1
2019/02/12 14:54:38 Container test in pod unit failed, exit code 2, reason Error
2019/02/12 14:54:39 Ran for 37s
error: could not run steps: test "unit" failed: the pod ci-op-iwn3fmvt/unit failed after 35s (failed containers: test): ContainerFailed one or more containers exited
Container test exited with code 2, reason Error
---
e
--- 
```

runcom:
yeah, there's an error I'm fixing

jlebon:
Looks good to me. Though the commit message title could be tweaked now that it also fixes the reading side as well.

runcom:
> Looks good to me. Though the commit message title could be tweaked now that it also fixes the reading side as well.

yup, I'll change it once I fix the unit tests :+1:

runcom:
updated and changed commit message/title as well

pkg/daemon/daemon.go
```diff
@@ -881,8 +882,8 @@ func checkFileContentsAndMode(filePath, expectedContent string, mode os.FileMode
 		glog.Errorf("could not read file: %q, error: %v", filePath, err)
 		return false
 	}
-	if strings.Compare(string(contents), expectedContent) != 0 {
-		glog.Errorf("content mismatch for file: %q; expected: %v; received: %v", filePath, expectedContent, string(contents))
+	if !bytes.Equal(contents, expectedContent) {
```
runcom: this should now be much faster compared to string methods as it's directly implemented in assembly as well

pkg/daemon/daemon.go
```diff
@@ -881,8 +882,8 @@ func checkFileContentsAndMode(filePath, expectedContent string, mode os.FileMode
...
+		glog.Errorf("content mismatch for file: %q; expected: %v; received: %v", filePath, expectedContent, contents)
```
jlebon: Hmm, this will print the contents as a byte array, right? We could not print the raw contents at all, though it's definitely helpful if it is indeed human-readable text (which it likely is in most cases). WDYT about casting to `string` here before printing and otherwise fallback on printing as a byte array?

I'm good with this as is too though and doing that as a follow up.

runcom: the issue is casting to string a bytes slice always succeeds, not sure how to differentiate if we can't show something, I'm not really sure this info is helpful for big files as well :thinking:

jlebon: How about we just tweak it to `"content mismatch for file: %q"` for now?

runcom: sounds great :+1:

runcom: done

Issue #410, opened 2019-02-11
Wrap wait.* calls to expose the real error on failure


runcom:
Taken from https://github.com/openshift/machine-config-operator/pull/408#discussion_r255720389

There are other spots where we're not correctly reporting the error (i.e. https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/node.go#L55-L65)

On error, we should save the `lastErr` encountered + check for `wait.ErrWaitTimeout` in the `if err != nil` and report lastErr.

jlebon:
> in the `if err != nil` and report lastErr

Ahh yep, that makes sense to me. (With the caveat that it might actually have been a different error on each retry, but even then, just percolating the last error seems like the right behaviour.)

PR #412, opened 2019-02-12
daemon/update: Log each error before retrying

jlebon:
That way, we get more feedback when retrying since it can take a while.
And we also catch the instance when e.g. different errors might have
happened each time.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -96,6 +96,7 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 				IgnoreDaemonsets:   true,
 			})
 			if err != nil {
+				glog.Infof("Draining failed with: %v; retrying...", err)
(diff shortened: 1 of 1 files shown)


cgwalters:
/lgtm

But...I think the real fix here is something like https://github.com/openshift/machine-config-operator/issues/410 where all of the fiddly details of "retry in a loop" are codified and consistent.

runcom:
could you also fix the formatting some lines below:
```
			return errors.Wrapf(lastErr, "Failed to drain node (%s tries)", backoff.Steps)

```

it should be `%d` as `backoff.Steps` is an int, not a string

ashcrow:
/hold

Per @runcom's comment.

runcom:
> Looks good as is, but there was a request for update.

a really nit(y) one lol sorry for that

jlebon:
> it should be `%d` as `backoff.Steps` is an int, not a string

Fixed!

PR #416, opened 2019-02-12
pkg/daemon/update.go: check wait err and wrap it

runcom:
we missed checking it https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L92

also, refactored the condition function a bit to be more future proof (i.e. someone comes and add a `return false, err`)

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -89,23 +89,25 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
 			Factor:   2,
 		}
 		var lastErr error
-		wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
 			err := drain.Drain(dn.kubeClient, []*corev1.Node{node}, &drain.DrainOptions{
 				DeleteLocalData:    true,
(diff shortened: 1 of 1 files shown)


runcom:
aws :heart:

ashcrow:
> aws heart

Yeah. Looks like resources.

pkg/daemon/update.go
```diff
@@ -89,23 +89,25 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
...
 				IgnoreDaemonsets:   true,
 			})
-			if err != nil {
-				glog.Infof("Draining failed with: %v; retrying...", err)
-				lastErr = err
-				return false, nil
+			if err == nil {
+				return true, nil
+			}
+			lastErr = fmt.Errorf("draining failed with: %v", err)
```
jlebon: How about just doing `glog.Infof()` for the error message here and leaving `lastErr = err`? Then if we wrap it later, we can still get back the original error that happened. (Anyway, "failed to drain node (5 tries): draining failed with: ..." is kinda repetitive, no? :))

runcom: :+1:

runcom: fixed, thanks!

PR #418, opened 2019-02-12
[cleanup] pkg/daemon: do not use pointers to strings for errors + unexport consts

runcom:
Pointer to strings in Golang as used rarely and almost only for struct
properties when deserializing json or yaml (or anything) into a structure.
Else, the idiomatic way of informing something isn't right it's through
errors only. Errors in Go do also have an handy `.Error()` which represents
the error as a string for our use cases.

Second commit unexports consts not used outside the daemon package.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -133,12 +133,12 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	reconcilableError := dn.reconcilable(oldConfig, newConfig)
 
 	if reconcilableError != nil {
-		msg := fmt.Sprintf("Can't reconcile config %v with %v: %v", oldConfigName, newConfigName, *reconcilableError)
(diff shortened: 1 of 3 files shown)


pkg/daemon/update.go
```diff
@@ -124,12 +124,12 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) error {
 	reconcilableError := dn.reconcilable(oldConfig, newConfig)
 
 	if reconcilableError != nil {
```
kikisdeliveryservice: Can we then say `if reconcilableError {` now?

runcom: nope, it's still an error (your case it's only valid for `bool` in Go)

kikisdeliveryservice: ahh TIL thank you!!

cgwalters:
/lgtm

But...this is all good cleanup but I still think we need to be careful about code churn at this point.

runcom:
> But...this is all good cleanup but I still think we need to be careful about code churn at this point.

I'm trying to be as concise as I can when cleaning up something to avoid regressions and overlooks

PR #368, opened 2019-02-01
MCS: change machine-config-server ports

kikisdeliveryservice:
Change machine-config-server ports from 49500/49501 -> 22623/22624
to avoid conflict with local port and node port ranges.

Closes: #166
Related-to: https://github.com/openshift/installer/pull/1180

Issue #166, opened 2018-11-12
machine-config-server should not listen in the local port range


squeed:
The machine-config-operator seems to listen on port 49500 (with `hostNetwork: true`). This is in the default `ip_local_port_range`, which means it can collide with active tcp sessions:

```
[root@test1-master-0 core]# sysctl net.ipv4.ip_local_port_range
net.ipv4.ip_local_port_range = 32768    60999
```

It should serve on a port lower than 32768.

For example, I managed to collide with a persistent connection from the apiserver to etcd:

```
[root@test1-master-0 core]# nc -l -t -p 49500
Ncat: bind to 0.0.0.0:49500: Address already in use. QUITTING.
[root@test1-master-0 core]# ss -np | grep 49500
tcp    ESTAB      0      0      192.168.126.11:49500              192.168.126.11:2379                users:(("hypershift",pid=10044,fd=60))
```

crawford:
@squeed Do you have a specific range that we should use? Does OpenShift define a particular range that we can use for internal services? If not, should we define one?

cgwalters:
To clarify, this port is required to serve Ignition configs, and Ignition runs in the initramfs before a node has joined the cluster and can use cluster networking, etc.

That said, is there any reason we couldn't just pick a free port dynamically on startup?

crawford:
> That said, is there any reason we couldn't just pick a free port dynamically on startup?

All of the machines in the cluster would have to know what port number they should connect to. If it were dynamically when the MCS started, how would new machines know where to connect?

squeed:
You just need to change the port. It cannot be in the local port range. Just pick a new number < 32768

cgwalters:
Was glancing at this just for my own edification, it seems like when we change this we need to make a co-ordinated change to the installer:

https://github.com/openshift/installer/blob/ac006ae671a645553d58c8a29c676968dfa3d85f/pkg/asset/ignition/machine/node.go#L24

wking:
For folks blindly searching issues, the current behavior results in [logs][1] like:

```
F0122 18:58:33.952823       1 api.go:59] Machine Config Server exited with error: listen tcp :49500: bind: address already in use```
```

leading to [e2e errors][2] like

```
fail [github.com/openshift/origin/test/extended/operators/cluster.go:109]: Expected
    <[]string | len:2, cap:2>: [
[...]

PR #423, opened 2019-02-13
MCS: remove legacy servers & ports

kikisdeliveryservice:
legacySecureServer and legacyInsecureServer were added to accommodate the
installer port references when we transitioned the MCS ports from 49500/49501
to 22623/22624. The installer was updated via PR #1180 to use 22623/22624, so
we can now remove the legacy server/port references.

Clean-up step for: #166 
Related to: #368 
Related-to: [openshift/installer#1180](https://github.com/openshift/installer/pull/1180)

### DIFF
--- a/cmd/machine-config-server/bootstrap.go
+++ b/cmd/machine-config-server/bootstrap.go
@@ -9,13 +9,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
-	// we are transitioning from legacy ports 49500/49501 -> 22623/22624
-	// this can be removed once fully transitioned in the installer.
-	legacySecurePort   = 49500
-	legacyInsecurePort = 49501
-)
-
 var (
 	bootstrapCmd = &cobra.Command{
 		Use:   "bootstrap",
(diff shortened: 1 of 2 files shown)


cgwalters:
This looks good to me.  One minor thing, would you mind adding the "topic prefix" like `MCS:` as you have in the PR title (I've been using `daemon:`, `controller:` or in this case `server:`) into the commit message title too?

One other thing I'd note is that while it should be safe from a CI perspective to do this right now, we will break anyone who e.g. is using an older installer and trying to do a new update payload.  So when we make changes like this it might be nice to wait...I dunno, a week?  But eh, I'm fine doing it now.

runcom:
/lgtm cancel 

due to Colin's comment above actually

cgwalters:
(On the other hand, merging this will definitely fix known issues)

ashcrow:
> This looks good to me. One minor thing, would you mind adding the "topic prefix" like `MCS:` as you have in the PR title (I've been using `daemon:`, `controller:` or in this case `server:`) into the commit message title too?

:+1: 

> One other thing I'd note is that while it should be safe from a CI perspective to do this right now, we will break anyone who e.g. is using an older installer and trying to do a new update payload. So when we make changes like this it might be nice to wait...I dunno, a week? But eh, I'm fine doing it now.

I saw let's rip off the band aid now :smile_cat:

cgwalters:
> I saw let's rip off the band aid now smile_cat

Fine by me too!

kikisdeliveryservice:
@cgwalters I just wanted to put the pr in now so that we wouldn't forget. I'm fine waiting too. I'll update my commit message and repush and let yall decide when it goes it. :smile:

ashcrow:
Flakes

/retest

PR #424, opened 2019-02-13
kubelet config deletions fix + example update

rphillips:
@umohnani8 found out that KubeletConfigs were hanging on deletes caused by the reference not getting updated. This PR expands the comparison for the update event to look at the entire object instead of just the spec.

Additionally, a second commit fixes the kubeletconfig example within the examples folder.

### DIFF
--- a/examples/kubeletconfig.crd.yaml
+++ b/examples/kubeletconfig.crd.yaml
@@ -3,5 +3,8 @@ kind: KubeletConfig
 metadata:
   name: set-max-pods
 spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      custom-kubelet: small-pods
   kubeletConfig:
     maxPods: 100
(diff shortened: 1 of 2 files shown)


runcom:
> @umohnani8 found out that KubeletConfigs were hanging on deletes caused by the reference not getting updated.

can you provide steps as to how to test this? Looks good to me otherwise

cgwalters:
> can you provide steps as to how to test this?

On that topic we also can now add our own tests in `test/e2e`, but don't need to block this PR on that.

runcom:
> On that topic we also can now add our own tests in `test/e2e`, but don't need to block this PR on that.

let's collect reproducer for stuff like this that are going in but that we eventually want to have a proper e2e for I guess :muscle:

umohnani8:
Steps to reproduce:
- Create the kubeletconfig CR from https://github.com/openshift/machine-config-operator/blob/master/examples/kubeletconfig.crd.yaml (with the matchLabels stuff that is added in this PR)
- Wait for the changes to take place and double check the kubelet.conf file on the master nodes for the changes
- Delete the kubeletconfig and make sure it goes through. Without this fix it was just hanging around forever.
* Make sure to add the matching label from your CR to the machineconfigpool (e.g oc edit machineconfigpool master)

Issue #400, opened 2019-02-11
Kubelet (and CRI-O) CRDs can be fooled by normal MCs


runcom:
If you create a machine config referencing the kubelet config (or the CRI-O once it lands) you're able to go around the existing CRD we have in place for the kubelet (and soon CRI-O).

The following MC causes at least `maxUnavailable` node(s) to go completely off (kube isn't reporting anymore).
```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: worker
  name: worker-add-file-0004
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,test
        filesystem: root
        mode: 0644
        path: /etc/kubernetes/kubelet.conf
```

```
[core@ip-10-0-171-250 ~]$ systemctl status -l kubelet.service
● kubelet.service - Kubernetes Kubelet
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: exit-code) since Mon 2019-02-11 14:36:39 UTC; 3s ago
  Process: 5028 ExecStart=/usr/bin/hyperkube kubelet --config=/etc/kubernetes/kubelet.conf --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig --kubeconfig=/var/lib/kubelet/kubeconfig --container-runtime=remote --container-runtime-endpoint=/var/run/crio/crio.sock --allow-privileged --node-labels=node-role.kubernetes.io/worker --minimum-container-ttl-duration=6m0s --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec --client-ca-file=/etc/kubernetes/ca.crt --cloud-provider=aws --anonymous-auth=false (code=exited, status=255)
  Process: 5026 ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state (code=exited, status=0/SUCCESS)
  Process: 5023 ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests (code=exited, status=0/SUCCESS)
 Main PID: 5028 (code=exited, status=255)

Feb 11 14:36:39 ip-10-0-171-250 systemd[1]: Failed to start Kubernetes Kubelet.
Feb 11 14:36:39 ip-10-0-171-250 systemd[1]: Unit kubelet.service entered failed state.
[...]

PR #430, opened 2019-02-14
kubeletconfig: add support for observedgeneration

rphillips:
ObservedObservations and Generations were a new concept to me...

Adds support for KubeletConfig Status updates and generations comparisons. Fixes KubeletConfig Updates where the status will get updated causing the watch to trigger an unlimited amount of times.


runcom:
is this fixing anything wrong? or is it just adding support for this?

rphillips:
@runcom It's fixing KubeletConfig Updates where the status will get updated causing the watch to trigger an unlimited amount of times.

runcom:
@umohnani8 FYI in case you're doing something similar in #400

PR #363, opened 2019-02-01
 Inject OSImageURL from CVO into templated MachineConfigs

cgwalters:
This injects the `OSImageURL` into the "base"
config (e.g. `00-worker`, `00-master`).  This differs from
previous pull requests which made it a separate MC, but that
adds visual noise and will exacerbate renderer race conditions.

PR #432, opened 2019-02-14
Match image reference in osimageurl configmap

jlebon:
Minor regression from #404. For the substitution to take place, the
pullspec has to match exactly.

---

I'm still confirming it, but I think this is the root of my issues in https://github.com/openshift/machine-config-operator/pull/363#issuecomment-463705911. The masters are getting their `osImageURL` from the bootstrap path (which gets it from the image-references file in the payload), while the controller gets it from the configmap, so the mismatch was causing a checksum diff.

I think the `e2e-aws` tests didn't fail on this because it follows a slightly different path? https://github.com/openshift/origin/pull/21919

### DIFF
--- a/install/0000_30_machine-config-operator_05_osimageurl.yaml
+++ b/install/0000_30_machine-config-operator_05_osimageurl.yaml
@@ -6,4 +6,7 @@ metadata:
 data:
   # The OS payload, managed by the daemon + pivot + rpm-ostree
   # https://github.com/openshift/machine-config-operator/issues/183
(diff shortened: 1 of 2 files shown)


cgwalters:
#363 has https://github.com/openshift/machine-config-operator/pull/363/commits/d58765b4cddca60ce55de903d03e0bff13f9ffcb

Are you sure you were testing with that commit included?

jlebon:
> Are you sure you were testing with that commit included?

Oh wow, completely missed that. I had only cherry-picked the HEAD. :innocent: 
OK, so sanity confirmed.

Thoughts on the added text here?

cgwalters:
I think my vote is after we branch, make it a completely dummy value not resembling the real ones.  But at this point I'd probably stick with trying to get #363 in as is?  Just since it's a known quantity.

Issue #417, opened 2019-02-12
TestKubeletConfigCreate flake?


runcom:
This test fails from time to time, let's track it down
```
--- FAIL: TestKubeletConfigCreate (0.12s)
    --- PASS: TestKubeletConfigCreate/aws (0.10s)
    --- FAIL: TestKubeletConfigCreate/none (0.01s)
    	kubelet_config_controller_test.go:207: Expected
    			testing.CreateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"create", Resource:schema.GroupVersionResource{Group:"", Version:"", Resource:"machineconfigs"}, Subresource:""}, Name:"", Object:(*v1.MachineConfig)(0xc4202f3180)}
    		got
    			testing.UpdateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"update", Resource:schema.GroupVersionResource{Group:"machineconfiguration.openshift.io", Version:"v1", Resource:"kubeletconfigs"}, Subresource:"status"}, Object:(*v1.KubeletConfig)(0xc4202f0780)}
    	kubelet_config_controller_test.go:176: 2 additional expected actions:[{ActionImpl:{Namespace: Verb:patch Resource:{Group:machineconfiguration.openshift.io Version:v1 Resource:kubeletconfigs} Subresource:} Name:smaller-max-pods Patch:[123 34 109 101 116 97 100 97 116 97 34 58 123 34 102 105 110 97 108 105 122 101 114 115 34 58 91 34 57 57 45 109 97 115 116 101 114 45 104 53 53 50 109 45 115 109 97 108 108 101 114 45 109 97 120 45 112 111 100 115 45 107 117 98 101 108 101 116 34 93 125 125]} {ActionImpl:{Namespace: Verb:update Resource:{Group:machineconfiguration.openshift.io Version:v1 Resource:kubeletconfigs} Subresource:status} Object:0xc420133900}]
```

```
--- FAIL: TestKubeletConfigUpdates (0.17s)
    --- PASS: TestKubeletConfigUpdates/aws (0.07s)
    --- PASS: TestKubeletConfigUpdates/none (0.05s)
    --- FAIL: TestKubeletConfigUpdates/unrecognized (0.05s)
    	kubelet_config_controller_test.go:213: Expected
[...]

PR #434, opened 2019-02-15
[cleanup] pkg/controller: don't use index in ranges, use int instead of int32

runcom:

**- What I did**

- use the object given by range directly, no reason to access the array by index, slices in Golang are guaranteed to be ordered as well
- use int instead of int32 to remove unnecessary casts

**- How I did it**

vscode

**- How to verify it**

CI I guess

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -459,13 +459,13 @@ func (ctrl *Controller) setDesiredMachineConfigAnnotation(nodeName, currentConfi
 	})
 }
 
-func makeProgress(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) (int32, error) {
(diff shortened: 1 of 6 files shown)


runcom:
These PRs can wait of course, no reason to bring them in _now_

runcom:
Ha sorry meant retest

/retest

crawford:
I see a lot of retests but no mention of what previously failed. This is a recipe for introducing flakes. Can you please double check what failed, provide a summary inline, and notify the appropriate team of the flake in the future?

runcom:
@crawford we've been doing that (e.g. https://github.com/openshift/machine-config-operator/issues/417). Most of the time I hit retest is because of aws api calls getting rate limited. The other flakes I find are the topmost you can find in the test grid already.

PR #435, opened 2019-02-15
 pkg/controller: add osimageurl to base ssh MC as well

runcom:
**- What I did**

builds on top of #363, cause I don't want to block that from merging right now, if you guys want instead, I'll close this and, _assuming this is needed_ we'll modify #363 directly.

@cgwalters ptal I'm not entirely sure we need this but I saw the ssh MCs were missing the osimageurl so I've added it (check only last commit)

**- How I did it**

vim

**- How to verify it**

test this PR with a custom payload and verify the `-ssh` MCs do have the osimageurl populated as well

### DIFF
--- a/install/image-references
+++ b/install/image-references
@@ -33,4 +33,4 @@ spec:
   - name: machine-os-content
     from:
       kind: DockerImage
-      name: registry.svc.ci.openshift.org/rhcos/maipo@sha256:83a6d461628380f1dcc057ffef4909992f593b72c5aa4db2e01c0b130004afea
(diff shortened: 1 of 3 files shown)


cgwalters:
This isn't necessary and IMO will just add confusion.

See https://github.com/openshift/machine-config-operator/pull/228 for some history around this.

If I could go back in time I think we should have made `machineconfigsrendered` a distinct object namespace.  It's fairly confusing how we mix our own data along with users' data.  And in particular with how osimageurl is handed.

cgwalters:
Specifically see https://github.com/openshift/machine-config-operator/pull/279

runcom:
makes total sense, closing, thanks for looking

PR #443, opened 2019-02-16
Add sync status tests

runcom:
**- What I did**

This PR builds on #442 so let's hold until that merges.
The sync status is one of the most important thing we need NOT to regress on.
What the last commit really does is setting up a way to easily add more unit tests for the sync calls to avoid regression. I've modified the test in #442 to use the new way and added 2 other smoke tests. The aim is to add even more to cover the majority of the cases. (if not all) 

For now, just review the last commit until #442 merges

**- How to verify it**

Run the unit tests


runcom:
holding caues waiting on #442 

/hold

runcom:
please take a look at https://github.com/openshift/machine-config-operator/pull/443/commits/ac3eb7f87c00f8b7122acdbc53e38c12b0a9ba5c and identify other cases if you can as well

runcom:
those units are too important actually, the code added is just in testing anyway. Adding this to #442

PR #445, opened 2019-02-16
image: MCD depends on util-linux (mount) which is being removed from base

smarterclayton:
As part of the general cleanup we are moving to UBI based Docker images
and util-linux is being removed from the base image. Add an explicit
dependency here.

### DIFF
--- a/Dockerfile.machine-config-daemon
+++ b/Dockerfile.machine-config-daemon
@@ -5,4 +5,5 @@ RUN WHAT=machine-config-daemon ./hack/build-go.sh
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-daemon /usr/bin/
+RUN yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*
 ENTRYPOINT ["/usr/bin/machine-config-daemon"]
--- a/Dockerfile.machine-config-daemon.rhel7
+++ b/Dockerfile.machine-config-daemon.rhel7
@@ -5,4 +5,5 @@ RUN WHAT=machine-config-daemon ./hack/build-go.sh
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-daemon /usr/bin/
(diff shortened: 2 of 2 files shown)


runcom:
not sure why `rhel-images` job isn't running (also on other PRs)

smarterclayton:
rhel-images was removed, but something is wrong with the status.  Assuming this goes green I'm going to force merge (I'm still trying to debug why the status came back)

hexfusion:
@smarterclayton if I try `make image-daemon` build fails for Dockerfile.machine-config-daemon is this rhel only?

```
[....]
STEP 11: RUN yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
http://base-4-0.ocp.svc/rhel-fast-datapath/repodata/repomd.xml: [Errno 14] curl#6 - "Could not resolve host: base-4-0.ocp.svc; Unknown error"
Trying other mirror.
```
/cc @abhinavdahiya

hexfusion:
ref: https://github.com/openshift/machine-config-operator/pull/463

smarterclayton:
Yes, for right now you have to be building the dockerfile on the ci
environment.  We’re going to have a workaround for that soon but it’s
complicated by not wanting to share our certificates too broadly.

On Feb 19, 2019, at 6:38 PM, Sam Batschelet <notifications@github.com>
wrote:

@smarterclayton <https://github.com/smarterclayton> if I try make
image-daemon build fails for Dockerfile.machine-config-daemon is this rhel
only?

[....]
STEP 11: RUN yum install -y util-linux && yum clean all && rm -rf
/var/cache/yum/*
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use
subscription-manager to
register.http://base-4-0.ocp.svc/rhel-fast-datapath/repodata/repomd.xml:
[Errno 14] curl#6 - "Could not resolve host: base-4-0.ocp.svc; Unknown
error"
Trying other mirror.

/cc @abhinavdahiya <https://github.com/abhinavdahiya>

—
<https://github.com/openshift/machine-config-operator/pull/445#issuecomment-465357206>,
or mute the thread

runcom:
> Yes, for right now you have to be building the dockerfile on the ci environment. We’re going to have a workaround for that soon but it’s complicated by not wanting to share our certificates too broadly.

@smarterclayton this breaks our own local development since we build images and deploy them on a running cluster to iterate on changes

PR #446, opened 2019-02-17
test/e2e: check MC deployed on every node

runcom:

**- What I did**

TestMCDeployed wasn't checking that the MC was indeed deployed on every
node (worker). Fix that by checking every node. The test is going to
take a little longer but it's more robust as well.

**- How to verify it**

run e2e

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -12,9 +12,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/machine-config-operator/cmd/common"
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
 )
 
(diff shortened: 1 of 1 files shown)


test/e2e/mcd_test.go
```diff
@@ -118,25 +120,39 @@ func TestMCDeployed(t *testing.T) {
...
-				if strings.Contains(line, "completed update for config "+newMCName) {
-					return true, nil
+			for _, pod := range mcdList.Items {
+				res, err := k.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &v1.PodLogOptions{}).DoRaw()
+				if err != nil {
+					// do not error out, we may be rebooting, that's why we list at every iteration
+					continue
+				}
+				for _, line := range strings.Split(string(res), "\n") {
+					if strings.Contains(line, "completed update for config "+newMCName) {
```
cgwalters: Seems odd to parse the log files for this; why not verify the `currentConfig` annotation?

cgwalters: Or for that matter, watch the worker machineconfigpool, which is doing just that.

runcom: changed to watch for status changes for the pool (even if now it's intrinsically racy cause we may somehow be missing some status, for instance, we may very well miss the Updating status, or check the previous Updated status thinking all is good if we don't check Updating and so on).

runcom: ok changed to check `currentConfig` instead.

runcom:
Haproxy flake

/retest

Issue #451, opened 2019-02-17
TestKubeletConfigUpdates flake


runcom:
Similar to #417 and #444, seen in https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/450/pull-ci-openshift-machine-config-operator-master-unit/1222

```
--- FAIL: TestKubeletConfigUpdates (0.17s)
    --- PASS: TestKubeletConfigUpdates/aws (0.09s)
    --- FAIL: TestKubeletConfigUpdates/none (0.02s)
    	kubelet_config_controller_test.go:213: Expected
    			testing.UpdateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"update", Resource:schema.GroupVersionResource{Group:"", Version:"", Resource:"machineconfigs"}, Subresource:""}, Object:(*v1.MachineConfig)(0xc4203b2a00)}
    		got
    			testing.UpdateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"update", Resource:schema.GroupVersionResource{Group:"machineconfiguration.openshift.io", Version:"v1", Resource:"kubeletconfigs"}, Subresource:"status"}, Object:(*v1.KubeletConfig)(0xc420132500)}
    	kubelet_config_controller_test.go:76: 2 additional expected actions:[{ActionImpl:{Namespace: Verb:patch Resource:{Group:machineconfiguration.openshift.io Version:v1 Resource:kubeletconfigs} Subresource:} Name:smaller-max-pods Patch:[123 34 109 101 116 97 100 97 116 97 34 58 123 34 102 105 110 97 108 105 122 101 114 115 34 58 91 34 57 57 45 109 97 115 116 101 114 45 109 119 119 116 103 45 107 117 98 101 108 101 116 34 93 125 125]} {ActionImpl:{Namespace: Verb:update Resource:{Group:machineconfiguration.openshift.io Version:v1 Resource:kubeletconfigs} Subresource:status} Object:0xc4203aef00}]
```

/cc @sjenning @rphillips

runcom:
notice that the test may be also flake somehow https://github.com/openshift/machine-config-operator/issues/417#issuecomment-464538786

runcom:
notice that the test setup and run may be also flake somehow https://github.com/openshift/machine-config-operator/issues/417#issuecomment-464538786

kikisdeliveryservice:
[...]

Issue #449, opened 2019-02-17
TestContainerRuntimeConfigCreate flake


runcom:
Basically the same flake as #417 but for the containerruntime controller:
Notice that #417 is closed but we've spotted it again, so we'll reopen it.

```
--- FAIL: TestContainerRuntimeConfigCreate (0.25s)
    --- FAIL: TestContainerRuntimeConfigCreate/aws (0.03s)
    	container_runtime_config_controller_test.go:207: Expected
    			testing.GetActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"get", Resource:schema.GroupVersionResource{Group:"", Version:"", Resource:"machineconfigs"}, Subresource:""}, Name:"99-master-426vz-containerruntime"}
    		got
    			testing.UpdateActionImpl{ActionImpl:testing.ActionImpl{Namespace:"", Verb:"update", Resource:schema.GroupVersionResource{Group:"machineconfiguration.openshift.io", Version:"v1", Resource:"containerruntimeconfigs"}, Subresource:"status"}, Object:(*v1.ContainerRuntimeConfig)(0xc420135540)}
[...]

PR #454, opened 2019-02-18
daemon: Don't do OS updates/pivoting on !RHCOS

cgwalters:
Without this we end up in a reboot loop.  Happened
for the UPI scaleup tests on CentOS (classic).

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -846,8 +846,18 @@ func (dn *Daemon) isUnspecifiedOS(osImageURL string) bool {
 	return osImageURL == "" || osImageURL == "://dummy"
 }
 
-// checkOS validates the OS image URL and returns true if they match.
+// checkOS determines whether the booted system matches the target
+// osImageURL and if not whether we need to take action.  This function
+// returns `true` if no action is required, which is the case if we're
+// not running RHCOS, or if the target osImageURL is "" (unspecified).
+// Otherwise if `false` is returned, then we need to perform an update.
 func (dn *Daemon) checkOS(osImageURL string) bool {
+	// Nothing to do if we're not on RHCOS
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		glog.Infof(`Not booted into Red Hat CoreOS, ignoring target OSImageURL %s`, osImageURL)
+		return true
+	}
+
 	// XXX: The installer doesn't pivot yet so for now, just make ""
 	// mean "unset, don't pivot". See also: https://github.com/openshift/installer/issues/281
 	if dn.isUnspecifiedOS(osImageURL) {


pkg/daemon/daemon.go
```diff
@@ -848,6 +848,11 @@ func (dn *Daemon) isUnspecifiedOS(osImageURL string) bool {
 
 // checkOS validates the OS image URL and returns true if they match.
 func (dn *Daemon) checkOS(osImageURL string) bool {
```
kikisdeliveryservice: Just to clarify, the func returns true if no update and false if it should be updated?

kikisdeliveryservice: ie false = os doesn't match target and we can update, true = os matches or for some other reason we can't update ?

kikisdeliveryservice: if that's the case can we update the comment above the func to make it clear?

pkg/daemon/daemon.go
```diff
@@ -848,6 +848,11 @@ func (dn *Daemon) isUnspecifiedOS(osImageURL string) bool {
...
+	// Nothing to do if we're not on RHCOS
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return true
```
kikisdeliveryservice: maybe add a glog here too?

pkg/daemon/daemon.go
cgwalters: Yeah, it is a bit odd; we can clean it up but I want to get a quick fix in for the UPI/RHEL folks.

pkg/daemon/daemon.go
cgwalters: Done

pkg/daemon/daemon.go
kikisdeliveryservice: makes sense!

kikisdeliveryservice:
Looks like #451 hit the unit test.

/test unit

kikisdeliveryservice:
HAProxy Router, EBS, TestKubletConfig flakes

/retest

runcom:
This looks good to me, any way anyone can live test it out? Otherwise, I believe this is right

pkg/daemon/daemon.go
runcom: I would enhance the naming of this function to something like  `systemMatchesOSImageURL` and document that we return false if the system is not support at all to help clear out confusion, not a blocker, let's follow up

runcom:
I'm actually going to approve and lgtm as it's the right fix anyway afaict

/lgtm

runcom:
```
level=fatal msg="failed to initialize the cluster: Cluster operator network has not yet reported success"
```
/retest

cgwalters:
> I'm actually going to approve and lgtm as it's the right fix anyway afaict

Remember it won't get in if it doesn't pass CI, so I often lgtm before tests are known to pass.

runcom:
> Remember it won't get in if it doesn't pass CI, so I often lgtm before tests are known to pass.

yeah, I just wanted someone to actually test this fix live but I get it's not that straightforward  and wanted to prevent a merge with my lgtm if it turned out to be wrong, but again, this is correct anyway

runcom:
network runtime bug again

/retest

runcom:
aws limit, will retest in some minutes

runcom:
unit flake #449

PR #441, opened 2019-02-16
containerruntimeconfig: watch Image CR for registries

umohnani8:
Add support to watch the Image CR for registry changes
by the user. Populate /etc/containers/registries.conf
with these changes.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

PR #456, opened 2019-02-19
pkg/controller: remove fmt.Println

runcom:
This slipped in through #441


runcom:
CI failure is https://github.com/openshift/origin/pull/22070

PR #420, opened 2019-02-13
append kube-ca to root-ca for mco config

mrogers950:
For https://github.com/openshift/installer/pull/1179 it's necessary to include the now self-signed kube-ca in the /etc/kubernetes/ca.crt file for kubelets, since the PR makes root-ca no longer a valid trust anchor for the API server certs. I think this is the correct spot to modify to change /etc/kubernetes/ca.crt, and I'm opting to append it rather than replace the root CA for now to make sure nothing else breaks. (There might be a different configmap that it would be better to grab kube-ca from.)
/cc @abhinavdahiya @deads2k @openshift/sig-auth

PR #459, opened 2019-02-19
bootstrap: set --kube-ca default to empty

mrogers950:
This PR is a follow-up regarding https://github.com/openshift/machine-config-operator/pull/420#issuecomment-463721284
The installer now uses the flag with https://github.com/openshift/installer/pull/1266.
/cc @cgwalters


mrogers950:
> Change looks fine, but why default to empty rather than a default? As an example, the next line has a default in `/assets/` as well.

The code makes including the kube-ca optional if it's empty, so I was keeping it in line with that. I was not sure if this command ran under any other context than the installer in order to have a given default be useful.

ashcrow:
> The code makes including the kube-ca optional if it's empty, so I was keeping it in line with that. I was not sure if this command ran under any other context than the installer in order to have a given default be useful.

That's fair. :+1:

kikisdeliveryservice:
LGTM but will defer to @cgwalters to approve

/assign @cgwalters

kikisdeliveryservice:
400 response & HAProxy flakes:

/test e2e-aws

cgwalters:
/lgtm

But FWIW in https://github.com/openshift/machine-config-operator/pull/403 we changed things to error out if the option isn't provided.  There aren't any use cases for this CLI other than the installer.

But eh, not worth respinning now.

ashcrow:
Flake
```
    --- FAIL: TestContainerRuntimeConfigCreate/unrecognized (0.00s)
```

/retest

PR #460, opened 2019-02-19
WIP: Handle transient errors

runcom:
**- What I did**

I may have missed something, I'm still thinking about it and trying it out but basically:

- the first commit reduce the calls we make to the apiserver just to get a node, most of the time, we do already have the object around
- the second commit introduces a transient error type which can be checked in `handleNodeUpdate` and retried. Code is modified accordingly to gracefully handle it as well
- the third commit expands the above to the case we're bringing up the node on boot

Still missing if we really want it is the use of the nodeInformer to grab the node object (and for machineconfigs as well)

related to https://github.com/openshift/machine-config-operator/issues/395

**- How to verify it**

play around with a cluster creating mcs and monitoring what's happening


runcom:
/hold

but I'd love some inputs as well @cgwalters @kikisdeliveryservice @LorbusChris

cmd/machine-config-daemon/start.go
```diff
@@ -154,10 +155,18 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	}
 
 	if startOpts.onceFrom == "" {
-		err = dn.CheckStateOnBoot()
-		if err != nil {
-			dn.EnterDegradedState(errors.Wrapf(err, "Checking initial state"))
+		for {
```
runcom: this and the other are just busy loops now, maybe we need to delay each execution

pkg/daemon/constants/constants.go
```diff
@@ -28,3 +30,8 @@ const (
 	// For more information, see https://github.com/openshift/pivot/pull/25/commits/c77788a35d7ee4058d1410e89e6c7937bca89f6c#diff-04c6e90faac2675aa89e2176d2eec7d8R44
 	EtcPivotFile = "/etc/pivot/image-pullspec"
 )
+
+var (
+	// ErrTransiet is an error which must be retried before giving up
+	ErrTransient = errors.New("transiet error")
```
kikisdeliveryservice: "transient"

Issue #462, opened 2019-02-20
upgrade fails in CI due to different generated release image names with same content


cgwalters:
I tried an upgrade from:
```
    Completion Time:    2019-02-19T22:16:55Z
    Image:              registry.svc.ci.openshift.org/openshift/origin-release@sha256:73c25681fed7403100a685b57c87acf0645df40075afbb5a436c514176d760bf
    Started Time:       2019-02-19T20:37:24Z
    State:              Partial
    Version:            4.0.0-0.alpha-2019-02-18-164603
```
to:
```
    Completion Time:    <nil>
    Image:              registry.svc.ci.openshift.org/openshift/origin-release:4.0.0-0.alpha-2019-02-19-072345
    Started Time:       2019-02-19T22:16:55Z
    State:              Partial
    Version:            4.0.0-0.alpha-2019-02-19-072345
```

Result:

```
$ oc get clusterversion
NAME      VERSION                           AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.0.0-0.alpha-2019-02-19-072345   False       True          3h39m   Unable to apply 4.0.0-0.alpha-2019-02-19-072345: the cluster operator machine-config is failing
```

Because:

```
$ oc logs -f pods/machine-config-daemon-xs9nm -p                                                                                                                                                                  
I0219 22:25:06.829208   22912 start.go:52] Version: 3.11.0-693-g86a5a55f-dirty                                                                                                                                     ...                                                                                                            
I0219 22:25:06.907763   22912 daemon.go:168] Booted osImageURL: registry.svc.ci.openshift.org/openshift/origin-v4.0-2019-02-18-164603@sha256:660061d6eae3ee6d93ca836cd52e6033f1d611c629c1ce47cf272c9e9bda2488 (47.3
18)                                                                                                                                                                                                               
...                                                                                                                         
I0219 22:26:16.123846   22912 update.go:650] Updating OS to registry.svc.ci.openshift.org/openshift/origin-v4.0-2019-02-19-072345@sha256:660061d6eae3ee6d93ca836cd52e6033f1d611c629c1ce47cf272c9e9bda2488  
...
rpm-ostreed.service: Txn UpdateDeployment on /org/projectatomic/rpmostree1/redhat_coreos_maipo failed: Old and new refs are equal: (null):5ee78f9fcf4296b9542ffc07d0630273d824a1498f924c31692dcfebf4ba6998
```

[...]

PR #463, opened 2019-02-20
daemon: Accept osImageURL with matching digests

cgwalters:
The way CI works, a release payload is generated in the CI namespace
but reusing the digested pull of existing components.  This means
the pull specs for *everything* change format - but the actual
content stays the same.

Let's handle this for the OS image at least.

In practice...I feel like we're going to want higher level special
handling for this - maybe the CVO does pull-through to the registry
and rewrites to an internal canonical form, but this should help
us for now.

Closes: https://github.com/openshift/machine-config-operator/issues/462

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -26,6 +26,7 @@ import (
 	mcfgclientv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
+	imgref "github.com/containers/image/docker/reference"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -580,8 +581,13 @@ func (dn *Daemon) CheckStateOnBoot() error {
 	}
 
 	if state.bootstrapping {
-		if !dn.checkOS(state.currentConfig.Spec.OSImageURL) {
-			glog.Info("Bootstrap pivot required")
+		targetOSImageURL := state.currentConfig.Spec.OSImageURL
+		osMatch, err := dn.checkOS(targetOSImageURL)
+		if err != nil {
+			return err
+		}
+		if !osMatch {
+			glog.Infof("Bootstrap pivot required to: %s", targetOSImageURL)
 			// This only returns on error
 			return dn.updateOSAndReboot(state.currentConfig)
 		}
@@ -825,7 +831,12 @@ func (dn *Daemon) triggerUpdateWithMachineConfig(currentConfig *mcfgv1.MachineCo
 // degraded.
 func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool {
 	// Be sure we're booted into the OS we expect
-	if !dn.checkOS(currentConfig.Spec.OSImageURL) {
+	osMatch, err := dn.checkOS(currentConfig.Spec.OSImageURL)
+	if err != nil {
+		glog.Errorf("%s", err);
+		return false
+	}
+	if !osMatch {
 		glog.Errorf("Expected target osImageURL %s", currentConfig.Spec.OSImageURL)
 		return false
 	}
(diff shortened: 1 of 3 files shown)


cgwalters:
Not tested locally though beyond new unit tests; looks like the UBI work broke `make image-daemon` for me with:

```
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
http://base-4-0.ocp.svc/rhel-fast-datapath/repodata/repomd.xml: [Errno 14] curl#6 - "Could not resolve host: base-4-0.ocp.svc; Unknown error"
Trying other mirror.
```

pkg/daemon/daemon.go
```diff
@@ -841,30 +852,67 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
...
 // isUnspecifiedOS says whether an osImageURL is "unspecified",
 // i.e. we should not try to change the current state.
-func (dn *Daemon) isUnspecifiedOS(osImageURL string) bool {
+func isUnspecifiedOS(osImageURL string) bool {
 	// The ://dummy syntax is legacy
 	return osImageURL == "" || osImageURL == "://dummy"
 }
 
+// compareOSImageURL is the backend for checkOS.
+func compareOSImageURL(current, desired string) (bool, error) {
```
kikisdeliveryservice: Could we somehow add the details in the comments about what Canonical is from here(https://github.com/containers/image/blob/5144ced37a1b21b63c6ef605e56811e29a687528/docker/reference/reference.go#L133) so the logic is more clear in the func (ie so I don't have to click around to find it :smile: )

pkg/daemon/daemon.go
```diff
@@ -841,30 +852,67 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
...
+		return false, err
+	}
+	bootedRefCanon, ok := bootedRef.(imgref.Canonical)
+	if !ok {
+		return false, errors.New("Booted osImageURL not canonical form")
+	}
+
+	ref, err := imgref.ParseNamed(desired)
+	if err != nil {
+		glog.Errorf("Failed to parse osImageURL: %v", err)
```
jlebon: Minor/optional: "Failed to parse desired osImageURL"?

pkg/daemon/daemon.go
```diff
@@ -841,30 +852,67 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
...
+		return false, err
+	}
+	refCanon, ok := ref.(imgref.Canonical)
+	if !ok {
+		return false, errors.New("Target osImageURL not canonical form")
+	}
```
jlebon: We could dedupe a bit here by factoring out a function that canonicalizes a ref else returns an error.

cgwalters: Done :arrow_down:

cgwalters:
Just for general interest, here's a workflow I have for testing PRs like this locally.  As part of submitting a PR, a release image will be generated.  You can log into the CI namespace (see `HACKING.md`) and from there run: `oc logs -c release release-latest`

In those logs you'll see a message like:
`Pushed image sha256:3b3ab2bfd9301f11ae0130c47010828e25a5f26f9caf7322f8d9b14a2571a881 to registry.svc.ci.openshift.org/ci-op-6br3phj4/release:latest`

Now, because the CI namespaces only last ~1h, if you want to do testing you'll need to mirror it, e.g.:

`oc adm release mirror --from registry.svc.ci.openshift.org/ci-op-6br3phj4/release:latest --to quay.io/cgwalters/ostest`

ashcrow:
```
level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
2019/02/20 18:37:00 Container setup in pod e2e-aws failed, exit code 1, reason Error
```

cgwalters:
Hmm, that job failed with:

```
0221 02:06:34.537828    5053 update.go:647] Updating OS to registry.svc.ci.openshift.org/ci-op-39zm8d3x/stable@sha256:660061d6eae3ee6d93ca836cd52e6033f1d611c629c1ce47cf272c9e9bda2488
E0221 02:06:34.557315    5053 daemon.go:436] Fatal error checking initial state of node: Checking initial state: Failed to run pivot: error starting job: Transaction is destructive.
E0221 02:06:34.557411    5053 writer.go:90] Marking degraded due to: Checking initial state: Failed to run pivot: error starting job: Transaction is destructive.
```

Hmm.  In [non-ancient systemd looks like that error has a more useful message](https://github.com/systemd/systemd/blob/56ee4d7001a162a7a526a10482c3007a7115539f/src/core/transaction.c#L519) which would have told us what was conflicting.

I feel like this is another case where we should have retried instead of going degraded maybe?

But clearly need to figure out why this happened.

cgwalters:
Not 100% certain but I think what happened here is since we don't have [this pivot commit](https://github.com/openshift/pivot/pull/38/commits/b4f9e9620b161a8df08cbb0555a3ae828b801dd2) we ended up in a race where the system was shutting down for reboot, but the MCD had come up and tried to start pivot again.

/retest

pkg/daemon/daemon.go
```diff
@@ -839,32 +850,69 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
...
+		return true, nil
+	}
+
+	if current == desired {
+		return true, nil
+	}
+
+	bootedDigest, err := getRefDigest(current)
+	if err != nil {
+		return false, errors.Wrap(err, "Parsing booted osImageURL")
```
runcom: lowercase errors

pkg/daemon/daemon.go
```diff
@@ -839,32 +850,69 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
...
+// its digest, or an error if the string fails to parse as
+// a "canonical" image reference with a digest.
+func getRefDigest(ref string) (string, error) {
+	refParsed, err := imgref.ParseNamed(ref)
+	if err != nil {
+		return "", err
+	}
+	canon, ok := refParsed.(imgref.Canonical)
+	if !ok {
+		return "", errors.New(fmt.Sprintf("Not canonical form: %s", ref))
```
runcom: `fmt.Errorf`

cgwalters:
(Any reason not to drop the `lgtm`?)

runcom:
/lgtm

@jlebon ptal

jlebon:
> @jlebon ptal

Yup, this is still
/lgtm
to me. :)

> Not 100% certain but I think what happened here is since we don't have [this pivot commit](https://github.com/openshift/pivot/pull/38/commits/b4f9e9620b161a8df08cbb0555a3ae828b801dd2) we ended up in a race where the system was shutting down for reboot, but the MCD had come up and tried to start pivot again.

Yeah, that sounds plausible to me.

kikisdeliveryservice:
level=error msg="\t* aws_route.to_nat_gw.5: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"

/retest

runcom:
aws limit hit

/retest

kikisdeliveryservice:
I'll retest in a bit, we keep hitting aws rate limits errors :frowning_face:

smarterclayton:
There is a reason we do this - we will almost certainly migrate our content to another registry location in the future.  It’s good to be smart, but don’t assume that if a new url comes in that it will always be at that location in the future.

PR #464, opened 2019-02-20
pkg/daemon: stash the node object

runcom:
**- What I did**

The node object is around most of the time (`handleNodeUpdated` for instance). When it's not a call is sufficient to get it and pass it around This reduces the call frequency to the api server.

This PR has been extracted from #460 and can go in w/o that. It allows us, working on PRs similar to #460 to have less places which can return a transient error from the apiserver and act accordingly.

**- How to verify it**

e2e should cover most of this I guess

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -92,6 +92,10 @@ type Daemon struct {
 
 	// channel used to ensure all spawned goroutines exit when we exit.
 	stopCh <-chan struct{}
+
+	// node is the current instance of the node being processed through handleNodeUpdate
+	// or the very first instance grabbed when the daemon starts
+	node *corev1.Node
 }
 
 // pendingConfigState is stored as JSON at pathStateJSON; it is only
@@ -241,7 +245,11 @@ func NewClusterDrivenDaemon(
 	dn.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "machineconfigdaemon", Host: nodeName})
 
 	glog.Infof("Managing node: %s", nodeName)
-	if err = loadNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), nodeName); err != nil {
+
+	if err := dn.setInitialNode(nodeName); err != nil {
+		return nil, err
+	}
+	if err = dn.loadNodeAnnotations(); err != nil {
 		return nil, err
 	}
 
@@ -469,11 +477,11 @@ func (dn *Daemon) getStateAndConfigs(pendingConfigName string) (*stateAndConfigs
 		glog.Info("In bootstrap mode")
 	}
 
-	currentConfigName, err := getNodeAnnotation(dn.kubeClient.CoreV1().Nodes(), dn.name, constants.CurrentMachineConfigAnnotationKey)
+	currentConfigName, err := getNodeAnnotation(dn.node, constants.CurrentMachineConfigAnnotationKey)
 	if err != nil {
 		return nil, err
 	}
-	desiredConfigName, err := getNodeAnnotation(dn.kubeClient.CoreV1().Nodes(), dn.name, constants.DesiredMachineConfigAnnotationKey)
+	desiredConfigName, err := getNodeAnnotation(dn.node, constants.DesiredMachineConfigAnnotationKey)
 	if err != nil {
 		return nil, err
 	}
@@ -481,7 +489,7 @@ func (dn *Daemon) getStateAndConfigs(pendingConfigName string) (*stateAndConfigs
 	if err != nil {
 		return nil, err
 	}
-	state, err := getNodeAnnotationExt(dn.kubeClient.CoreV1().Nodes(), dn.name, constants.MachineConfigDaemonStateAnnotationKey, true)
+	state, err := getNodeAnnotationExt(dn.node, constants.MachineConfigDaemonStateAnnotationKey, true)
 	if err != nil {
 		return nil, err
 	}
@@ -647,7 +655,7 @@ func (dn *Daemon) CheckStateOnBoot() error {
 			// Great, we've successfully rebooted for the desired config,
 			// let's mark it done!
 			glog.Infof("Completing pending config %s", state.pendingConfig.GetName())
-			if err := dn.completeUpdate(state.pendingConfig.GetName()); err != nil {
+			if err := dn.completeUpdate(dn.node, state.pendingConfig.GetName()); err != nil {
 				return err
 			}
 		}
@@ -712,6 +720,8 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
 
 	// First check if the node that was updated is this daemon's node
 	if node.Name == dn.name {
+		// stash the current node being processed
+		dn.node = node
 		// Pass to the shared update prep method
 		needUpdate, err := dn.prepUpdateFromCluster()
 		if err != nil {
@@ -737,7 +747,7 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
 // update is required, false otherwise.
 func (dn *Daemon) prepUpdateFromCluster() (bool, error) {
 	// Then check we're not already in a degraded state.
-	state, err := getNodeAnnotation(dn.kubeClient.CoreV1().Nodes(), dn.name, constants.MachineConfigDaemonStateAnnotationKey)
+	state, err := getNodeAnnotation(dn.node, constants.MachineConfigDaemonStateAnnotationKey)
 	if err != nil {
 		return false, err
 	}
@@ -745,14 +755,19 @@ func (dn *Daemon) prepUpdateFromCluster() (bool, error) {
 		return false, fmt.Errorf("state is already degraded")
 	}
 
-	// Grab the node instance
-	node, err := GetNode(dn.kubeClient.CoreV1().Nodes(), dn.name)
+	desiredConfigName, err := getNodeAnnotationExt(dn.node, constants.DesiredMachineConfigAnnotationKey, true)
+	if err != nil {
+		return false, err
+	}
+	// currentConfig is always expected to be there as loadNodeAnnotations
+	// is one of the very first calls when the daemon starts.
+	currentConfigName, err := getNodeAnnotation(dn.node, constants.CurrentMachineConfigAnnotationKey)
 	if err != nil {
 		return false, err
 	}
 
 	// Detect if there is an update
-	if node.Annotations[constants.DesiredMachineConfigAnnotationKey] == node.Annotations[constants.CurrentMachineConfigAnnotationKey] {
+	if desiredConfigName == currentConfigName {
 		// No actual update to the config
 		glog.V(2).Info("No updating is required")
 		return false, nil
@@ -783,13 +798,8 @@ func (dn *Daemon) executeUpdateFromCluster() error {
 // completeUpdate marks the node as schedulable again, then deletes the
 // "transient state" file, which signifies that all of those prior steps have
 // been completed.
-func (dn *Daemon) completeUpdate(desiredConfigName string) error {
-	node, err := GetNode(dn.kubeClient.CoreV1().Nodes(), dn.name)
-	if err != nil {
-		return err
-	}
-	err = drain.Uncordon(dn.kubeClient.CoreV1().Nodes(), node, nil)
-	if err != nil {
+func (dn *Daemon) completeUpdate(node *corev1.Node, desiredConfigName string) error {
+	if err := drain.Uncordon(dn.kubeClient.CoreV1().Nodes(), node, nil); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


runcom:
@jlebon ptal

runcom:
I guess a possible next step after this is to start using the daemon's `nodeLister` + `node.DeepCopy()` + `nodeClient.Update(n)` within `GetNode` and `updateNodeRetry` as well, @cgwalters is that what you meant in https://github.com/openshift/machine-config-operator/pull/409#issuecomment-463372473 ?

cgwalters:
> I guess a possible next step after this is to start using the daemon's `nodeLister` + `node.DeepCopy()` + `nodeClient.Update(n)` within `GetNode` and `updateNodeRetry` as well, @cgwalters is that what you meant in [#409 (comment)](https://github.com/openshift/machine-config-operator/pull/409#issuecomment-463372473) ?

Well I think rather than `Update` we can use e.g. `strategicpatch.CreateTwoWayMergePatch` just like the controllers do right?

But yeah, the main thing is to use the informer's state.

In practice...the daemon is mostly "wait for event, then launch" - so really the simplest may be to just get the node once and stash it in the daemon.  Something like:

```
index 6a64507..c590374 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -710,6 +710,9 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
 
 	// First check if the node that was updated is this daemon's node
 	if node.Name == dn.name {
+
+		dn.node = node
+
 		// Pass to the shared update prep method
 		needUpdate, err := dn.prepUpdateFromCluster()
 		if err != nil {
```

And then use `dn.node` everywhere else.

runcom:
> Well I think rather than `Update` we can use e.g. `strategicpatch.CreateTwoWayMergePatch` just like the controllers do right?

yes noticed controllers are doing that as well

> so really the simplest may be to just get the node once and stash it in the daemon

and yes, this PR is a step behind that as a first pass to start having node objects around. Let me know if you want your changes in this PR or I will follow up to reduce code review

runcom:
or I can add a commit on top of this, let me know.

cgwalters:
> this PR is a step behind that as a first pass to start having node objects around.

Feels less invasive to the code though to stash it in the `Daemon` struct though right?  All the code in the daemon is always going to operate on exactly one node.

runcom:
> Feels less invasive to the code though to stash it in the `Daemon` struct though right? All the code in the daemon is always going to operate on exactly one node.

sure, I'll go that way then, though...handleNodeUpdated can be called concurrently? If so, I'm not sure we can have that object stashed

runcom:
I can see:

```
	// AddEventHandler adds an event handler to the shared informer using the shared informer's resync
	// period.  Events to a single handler are delivered sequentially, but there is no coordination
	// between different handlers.
	AddEventHandler(handler ResourceEventHandler)
```

runcom:
> But yeah, the main thing is to use the informer's state.

there's the case in `CheckStateOnBoot` where we don't have the informers at all so code must be refactored more to differentiate both situations.

I guess we can go ahead with the stashing in this PR and follow up to use informers.

cgwalters:
> there's the case in `CheckStateOnBoot` where we don't have the informers at all so code must be refactored more to differentiate both situations.

Mmm...I think we can just flip the ordering there and do `CheckStateOnBoot` after the informer cache is available.

Actually, we can/should probably change this down the line to be a periodic job.  See also [this comment](https://github.com/openshift/machine-config-operator/pull/245#issue-239291556)

> Now, the "check the on disk state for drift from expected config"
is factored out into a new function. We do this every time on
MCD start, although now we could e.g. change things to do it at
most once a day, writing a "last checked" timestamp in `/run` or as a node annotation or
whatever.

runcom:
> Mmm...I think we can just flip the ordering there and do `CheckStateOnBoot` after the informer cache is available.

sounds fine, I'll do it.

> 
> Actually, we can/should probably change this down the line to be a periodic job. See also [this comment](https://github.com/openshift/machine-config-operator/pull/245#issue-239291556)

agree with that (as a follow up)

runcom:
> Mmm...I think we can just flip the ordering there and do `CheckStateOnBoot` after the informer cache is available.

not really, the informer cache is synced only after `dn.Run`

cgwalters:
> or maybe we can move `CheckStateOnBoot` within `Run` as a first thing to do (if that's what you really meant).

Right, conceptually it's part of the daemon run I think.

> if the cache sync fail also, we need to still use the client to update annotations

I think Derek cited that specific bit of code as something we shouldn't be doing; if talking to Kube fails we should retry, not go degraded.

ashcrow:
```
level=info msg="Creating cluster..."
level=info msg="Waiting up to 30m0s for the Kubernetes API..."
level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
```

and 

```
rpc error: code = 2 desc = oci runtime error: exec failed: container "2686d319707...8e25307157c" does not exist
[...]
level=fatal msg="waiting for Kubernetes API: context deadline exceeded"
2019/02/20 15:17:52 Container setup in pod e2e-aws failed, exit code 1, reason Error
```

runcom:
> I think Derek cited that specific bit of code as something we shouldn't be doing; if talking to Kube fails we should retry, not go degraded

Exactly, my point was that there's still some spot like that one where we cannot use informers since the cache isn't even ready and, retry or not, we need to use the bare kube client.

cgwalters:
All the MCDs here are dying with:

```
I0222 15:26:04.768867    2382 start.go:52] Version: 3.11.0-716-g54a39c89-dirty
I0222 15:26:04.769764    2382 start.go:88] starting node writer
I0222 15:26:04.776029    2382 run.go:22] Running captured: chroot /rootfs rpm-ostree status --json
I0222 15:26:04.854307    2382 daemon.go:172] Booted osImageURL: registry.svc.ci.openshift.org/ci-op-xw2l1jn0/stable@sha256:660061d6eae3ee6d93ca836cd52e6033f1d611c629c1ce47cf272c9e9bda2488 (47.318)
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x10a3dbe]

goroutine 1 [running]:
github.com/openshift/machine-config-operator/pkg/daemon.(*Daemon).setInitialNode(0xc4200d2360, 0xc42004200a, 0x1c, 0x0, 0x2)
	/go/src/github.com/openshift/machine-config-operator/pkg/daemon/node.go:75 +0x2e
github.com/openshift/machine-config-operator/pkg/daemon.New(0x1385e63, 0x7, 0xc42004200a, 0x1c, 0x1385339, 0x5, 0x14a22a0, 0x1f83218, 0x14ccf00, 0x1f83218, ...)
	/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:197 +0x370
github.com/openshift/machine-config-operator/pkg/daemon.NewClusterDrivenDaemon(0x1385e63, 0x7, 0xc42004200a, 0x1c, 0x1385339, 0x5, 0x14a22a0, 0x1f83218, 0x14aeae0, 0xc4203af180, ...)
	/go/src/github.com/openshift/machine-config-operator/pkg/daemon/daemon.go:222 +0x125
main.runStartCmd(0x1f136c0, 0x1f83218, 0x0, 0x0)
	/go/src/github.com/openshift/machine-config-operator/cmd/machine-config-daemon/start.go:120 +0xca7
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).execute(0x1f136c0, 0x1f83218, 0x0, 0x0, 0x1f136c0, 0x1f83218)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:766 +0x2c1
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x1f13460, 0x0, 0xc420101f78, 0x406f6c)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:852 +0x30a
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).Execute(0x1f13460, 0xc420101f60, 0x0)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:800 +0x2b
main.main()
	/go/src/github.com/openshift/machine-config-operator/cmd/machine-config-daemon/main.go:27 +0x31
```

runcom:
> All the MCDs here are dying with:

The `onceFrom` is completely broken then. It requires a kubeClient to operate but then it's not even there?
Not sure what we want to do

runcom:
Force pushed to fix the panic, but see https://github.com/openshift/machine-config-operator/pull/464#issuecomment-466452308

runcom:
looks like it's still failing somehow, I'll take a look later today

/hold

runcom:
figured out what was going on...

/hold cancel

runcom:
aws toomanybuckets limit?

/retest

runcom:
back working and green, and avoid many calls, @cgwalters @kikisdeliveryservice @LorbusChris ptal (tomorrow)

pkg/daemon/daemon.go
```diff
@@ -92,6 +92,10 @@ type Daemon struct {
 
 	// channel used to ensure all spawned goroutines exit when we exit.
 	stopCh <-chan struct{}
+
+	// node is the current instance of the node being processed through handleNodeUpdate
+	// or the very first instance grabbed when he daemon starts
```
LorbusChris: nit: the daemon

pkg/daemon/daemon.go
```diff
@@ -737,22 +747,27 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
...
 	}
 
-	// Grab the node instance
-	node, err := GetNode(dn.kubeClient.CoreV1().Nodes(), dn.name)
+	desiredConfigName, err := getNodeAnnotationExt(dn.node, constants.DesiredMachineConfigAnnotationKey, true)
+	if err != nil {
+		return false, err
+	}
+	// currentConfig is always expected to be there as loadNodeAnnotations
+	// is one of the very first call when the daemon starts.
```
LorbusChris: nit: calls

pkg/daemon/writer.go
```diff
@@ -142,32 +142,40 @@ func (nw *NodeWriter) SetSSHAccessed(client corev1.NodeInterface, node string) e
...
 	}
 
-	return nil
+	return node, nil
 }
 
-// setConfig sets the given annotation key, value pair.
-func setNodeAnnotations(client corev1.NodeInterface, node string, m map[string]string) error {
-	return updateNodeRetry(client, node, func(node *v1.Node) {
+func setNodeAnnotationsNodeBack(client corev1.NodeInterface, nodeName string, m map[string]string) (*v1.Node, error) {
```
LorbusChris: the `NodeBack` in the name is due to the func returning the node, right? Wondering whether we could name that differently...although that'd really just be personal preference, not really necessary. Maybe add a tiny comment?

runcom: I suck at naming, a lot, so if you have any ideas I can change it lol

runcom:
Taken care of comments, ptal

pkg/daemon/writer.go
kikisdeliveryservice: revertNodeAnnotations? resetNodeAnnotations? fallBackNodeAnnotations? <-- i dont even know if those are better or worse tho.. :laughing:

kikisdeliveryservice:
`level=error msg="\t* aws_route_table_association.route_net.4: timeout while waiting for state to become 'success' (timeout: 5m0s)"`  <- rate limiting i think.

/test e2e-aws-op

runcom:
Passing now

cgwalters:
Are we targeting this for 4.0 under the "improves upgrades" banner?

runcom:
> Are we targeting this for 4.0 under the "improves upgrades" banner?

This surely has the benefit to have less code paths that can led to an apiserver timeout and error. So I'd say it goes with "improves upgrades" as well where it's more likely the apiserver can stay offline for a while and we need to not error out.

The other side of this PR which I'll work on is substituting the kube client with the informer and afterwards we need to add a retry logic for transient failiures .

How does that sound?

cgwalters:
Today the code in `writer.go` updates node annotations.  If we cache the node object permanently, then the daemon won't see its own annotation updates.  

Today...this probably doesn't matter much but I think we should e.g. have `setNodeAnnotations` also return the modified node object and update the daemon's cache.

runcom:
> Today...this probably doesn't matter much but I think we should e.g. have `setNodeAnnotations` also return the modified node object and update the daemon's cache.

I've already account for that and `setNodeAnnotations` now returns the updated node object https://github.com/openshift/machine-config-operator/pull/464/files#diff-534838ddb77c16bbccf153bf8916b114R169

Going forward, that becomes a daemon method, and automatically re-save the daemon instance on success.

cgwalters:
OK...I can't find a codepath offhand in the daemon which wants to read back any of the annotations right now.

/lgtm

pkg/daemon/writer.go
LorbusChris: setNodeAnnotationsReturnNode?

runcom:
> OK...I can't find a codepath offhand in the daemon which wants to read back any of the annotations right now.

I guess we'll find out :smile: but I've spent quite some time checking that and couldn't find one either (100% possible I missed it if there's some)

PR #467, opened 2019-02-20
MAO: Stop setting statusProgressing=true when resyincing same version

michaelgugino:
Fixes: https://jira.coreos.com/browse/CLOUD-391


cgwalters:
The JIRA you linked is for the [machine API operator](https://github.com/openshift/machine-api-operator) but this repo is the machine *config* operator.

Issue #455, opened 2019-02-18
Pod sandboxes do not use payload specified image


sjenning:
https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-container-runtime/_base/files/crio.yaml#L206

https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-container-runtime/_base/files/crio.yaml#L206

Should use the `pod` image from the release payload.

There is some plumbing to do here as that image pull spec is not provided to the MCO currently.

@smarterclayton @umohnani8 @mrunalp @runcom @rphillips

sdodson:
/cc @vrutkovs @mtnbikenc 
We'll want to track this change so we're ideally setting the correct pause image during BYO.

cgwalters:
:thinking: Why are we still doing the pause image thing with cri-o? It always seemed silly...plus if we do need to do it we can pick another image to use for this right?

sjenning:
It holds open the pod-level namespaces, namely the network namespace.

This is so the namespaces can be initialized before the application starts (assign IPs and whatnot), and maintained if the application container(s) die and need to be restarted.

vrutkovs:
In RHEL scaleup we use default CRIO config and change `pause_image` and `pause_command` - https://github.com/openshift/openshift-ansible/pull/11203/files#diff-8696e419a88c9744f085bc0a531bc511R49

mrunalp:
Also, since k8s added support for sharing pid namespaces, we can't get rid of the infra(pause) container as it is expected to be pid 1 for the pod.

umohnani8:
@sjenning is the `pod` image coming from docker.io or quay.io?

runcom:
> Should use the `pod` image from the release payload.

what's the name of that image that I can query with `releasepayloadimage image {pod}`?

> 
> There is some plumbing to do here as that image pull spec is not provided to the MCO currently.

this has to go through https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L155 right?

cgwalters:
https://github.com/openshift/machine-config-operator/pull/343 is a good prior change for "get an image from the release payload into configs".

PR #471, opened 2019-02-21
Pick image from release payload for crio.conf pause_image

umohnani8:
We want to pick the release payload image for the infra image.
Passing along the image we get from bootkube during bootstrap to
crio.conf.
Removed `infraImage` from the ctrcfg CR as well.

Note: we don't know what the infra image is called in the release payload yet, still working on figuring that out. I have seen it being `docker.io/openshift/origin-pod:v4.0.0" in openshift-ansible so using that for now. Will update that once we get a confirmation.
Also opening a PR in installer to add the infra image to bootkube.sh

Fixes https://github.com/openshift/machine-config-operator/issues/455


umohnani8:
@runcom @mrunalp PTAL

umohnani8:
bootkube PR https://github.com/openshift/installer/pull/1292

abhinavdahiya:
missing https://github.com/openshift/machine-config-operator/blob/5b65a3cc35e29b1800d4ab1cd0ea4e38942d822d/pkg/operator/operator.go#L327

runcom:
Let's chase down what's the pod image, apart from Abhinav comment, code looks already good (I'll defer a full review after we have everything in place)

umohnani8:
@abhinavdahiya what is missing? I have `InfraImageKey` there as well.

abhinavdahiya:
> missing
> 
> [machine-config-operator/pkg/operator/operator.go](https://github.com/openshift/machine-config-operator/blob/5b65a3cc35e29b1800d4ab1cd0ea4e38942d822d/pkg/operator/operator.go#L327)
> 
> Line 327 in [5b65a3c](/openshift/machine-config-operator/commit/5b65a3cc35e29b1800d4ab1cd0ea4e38942d822d)
>  templatectrl.SetupEtcdEnvKey: imgs.SetupEtcdEnv,

sorry missed the change below all the test_data :P

cgwalters:
Could also add a test to verify that `k8s.gcr.io/pause` hasn't been pulled to a node.  But looks sane to me.

umohnani8:
Fixed the infra_image name. Ready for review and merge.

runcom:
This looks good but I'm wondering about orders of PRs now between here and installer since the note here https://github.com/openshift/machine-config-operator/pull/471/files#diff-c15e2a465008fe4b86940c96d54a17eaR3

@cgwalters ^^^

cgwalters:
Yep this needs to land first, then the installer PR.

runcom:
> Yep this needs to land first, then the installer PR.

cause this is what actually generates the pod image in the release right?

cmd/machine-config-operator/bootstrap.go
```diff
@@ -62,6 +63,8 @@ func init() {
 	bootstrapCmd.MarkFlagRequired("etcd-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.setupEtcdEnvImage, "setup-etcd-env-image", "", "Image for Setup Etcd Environment.")
 	bootstrapCmd.MarkFlagRequired("setup-etcd-env-image")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraImage, "infra-image", "", "Image for Infra Containers.")
```
runcom: do we need a default value according to https://github.com/openshift/machine-config-operator/pull/471/files#diff-c15e2a465008fe4b86940c96d54a17eaR3?

runcom:
Just a comment, looks good to me otherwise and we verified with @umohnani8 and @mrunalp that the pod image is indeed in the payload generated from this PR, with #472 @umohnani8 you should be able to test the flow with a dummy value as well.

cmd/machine-config-operator/bootstrap.go
cgwalters: Yes - until the installer change lands.

cgwalters:
> cause this is what actually generates the pod image in the release right?

No, at least I'm pretty sure that's not true.  Is `pod` an image that's built out of an existing repo?  Answer, looks like yes:

```
# podman inspect quay.io/openshift/origin-pod:v4.0|grep source-location | head -1
                "io.openshift.build.source-location": "https://github.com/openshift/images",
```

So...if this works, I think it only works because that image is already part of the build system.

(But I am fuzzy on this part.  I know for "external" images like etcd and machine-os-content, Clayton added those manually)

runcom:
> So...if this works, I think it only works because that image is already part of the build system.

yes, we went through the doc explaining that yesterday and I forgot to cancel my message :+1:

runcom:
/lgtm

I guess the path now is, merge this, merge installer PR, create another PR here to remove the default

runcom:
aws limit hit

/retest

umohnani8:
@runcom a comma was missing from the yaml file, which I think was causing issues with the aws CI. Just added that.

runcom:
Mmm e2e-op is failing though

runcom:
this is constantly hitting #338 

```
I0222 23:45:02.552354       1 node_controller.go:383] Pool master is unconfigured, pausing 5s for renderer to initialize
I0222 23:45:02.552915       1 node_controller.go:383] Pool worker is unconfigured, pausing 5s for renderer to initialize
I0222 23:45:03.150837       1 render_controller.go:456] Generated machineconfig worker-620afe303e7ba1565eced212c44425bb from 2 configs: [{MachineConfig  00-worker  machineconfiguration.openshift.io/v1  } {MachineConfig  00-worker-ssh  machineconfiguration.openshift.io/v1  }]
I0222 23:45:03.152936       1 render_controller.go:456] Generated machineconfig master-b6aaf2eb6ae7a3e11ac41ae97c240317 from 2 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  } {MachineConfig  00-master-ssh  machineconfiguration.openshift.io/v1  }]
I0222 23:45:09.151055       1 node_controller.go:432] Setting node ip-10-0-131-201.ec2.internal to desired config master-b6aaf2eb6ae7a3e11ac41ae97c240317
I0222 23:45:09.650855       1 render_controller.go:456] Generated machineconfig worker-c0b9ccf2441e508bb9469d8afe5ab48b from 4 configs: [{MachineConfig  00-worker  machineconfiguration.openshift.io/v1  } {MachineConfig  00-worker-ssh  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-container-runtime  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-kubelet  machineconfiguration.openshift.io/v1  }]
I0222 23:45:10.050443       1 render_controller.go:369] Error syncing machineconfigpool worker: Operation cannot be fulfilled on machineconfigpools.machineconfiguration.openshift.io "worker": the object has been modified; please apply your changes to the latest version and try again
I0222 23:45:11.150544       1 render_controller.go:456] Generated machineconfig worker-c0b9ccf2441e508bb9469d8afe5ab48b from 4 configs: [{MachineConfig  00-worker  machineconfiguration.openshift.io/v1  } {MachineConfig  00-worker-ssh  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-container-runtime  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-kubelet  machineconfiguration.openshift.io/v1  }]
I0222 23:45:11.150611       1 render_controller.go:369] Error syncing machineconfigpool worker: machineconfigs.machineconfiguration.openshift.io "worker-c0b9ccf2441e508bb9469d8afe5ab48b" already exists
I0222 23:45:11.352114       1 render_controller.go:456] Generated machineconfig master-78e7909e5f846e944d44135e286e4438 from 4 configs: [{MachineConfig  00-master  machineconfiguration.openshift.io/v1  } {MachineConfig  00-master-ssh  machineconfiguration.openshift.io/v1  } {MachineConfig  01-master-container-runtime  machineconfiguration.openshift.io/v1  } {MachineConfig  01-master-kubelet  machineconfiguration.openshift.io/v1  }]
I0222 23:45:11.750438       1 render_controller.go:369] Error syncing machineconfigpool master: Operation cannot be fulfilled on machineconfigpools.machineconfiguration.openshift.io "master": the object has been modified; please apply your changes to the latest version and try again
I0222 23:45:14.750506       1 node_controller.go:432] Setting node ip-10-0-131-201.ec2.internal to desired config master-78e7909e5f846e944d44135e286e4438
I0222 23:48:06.251035       1 node_controller.go:432] Setting node ip-10-0-147-28.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:11.351316       1 node_controller.go:432] Setting node ip-10-0-147-28.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:16.657754       1 node_controller.go:432] Setting node ip-10-0-147-28.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:22.051072       1 node_controller.go:432] Setting node ip-10-0-147-28.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:40.340347       1 node_controller.go:432] Setting node ip-10-0-132-58.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:45.451232       1 node_controller.go:432] Setting node ip-10-0-172-0.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
I0222 23:48:50.551153       1 node_controller.go:432] Setting node ip-10-0-172-0.ec2.internal to desired config worker-c0b9ccf2441e508bb9469d8afe5ab48b
```

runcom:
can it be because the containerruntimeconfig controller is started after the node one? (even if in a goroutine?). @umohnani8 try this out:

```diff
diff --git a/cmd/machine-config-controller/start.go b/cmd/machine-config-controller/start.go
index e8646c8..1323aff 100644
--- a/cmd/machine-config-controller/start.go
+++ b/cmd/machine-config-controller/start.go
@@ -100,6 +100,17 @@ func startControllers(ctx *common.ControllerContext) error {
 		ctx.ClientBuilder.MachineConfigClientOrDie("kubelet-config-controller"),
 	).Run(2, ctx.Stop)
 
+	go containerruntimeconfig.New(
+		rootOpts.templates,
+		ctx.InformerFactory.Machineconfiguration().V1().MachineConfigPools(),
+		ctx.InformerFactory.Machineconfiguration().V1().ControllerConfigs(),
+		ctx.InformerFactory.Machineconfiguration().V1().ContainerRuntimeConfigs(),
+		ctx.ConfigInformerFactory.Config().V1().Images(),
+		ctx.ClientBuilder.KubeClientOrDie("container-runtime-config-controller"),
+		ctx.ClientBuilder.MachineConfigClientOrDie("container-runtime-config-controller"),
+		ctx.ClientBuilder.ConfigClientOrDie("container-runtime-config-controller"),
+	).Run(2, ctx.Stop)
+
 	// The renderer creates "rendered" MCs from the MC fragments generated by
 	// the above sub-controllers, which are then consumed by the node controller
 	go render.New(
@@ -117,16 +128,5 @@ func startControllers(ctx *common.ControllerContext) error {
 		ctx.ClientBuilder.MachineConfigClientOrDie("node-update-controller"),
 	).Run(2, ctx.Stop)
 
-	go containerruntimeconfig.New(
-		rootOpts.templates,
-		ctx.InformerFactory.Machineconfiguration().V1().MachineConfigPools(),
-		ctx.InformerFactory.Machineconfiguration().V1().ControllerConfigs(),
-		ctx.InformerFactory.Machineconfiguration().V1().ContainerRuntimeConfigs(),
-		ctx.ConfigInformerFactory.Config().V1().Images(),
-		ctx.ClientBuilder.KubeClientOrDie("container-runtime-config-controller"),
-		ctx.ClientBuilder.MachineConfigClientOrDie("container-runtime-config-controller"),
-		ctx.ClientBuilder.ConfigClientOrDie("container-runtime-config-controller"),
-	).Run(2, ctx.Stop)
-
 	return nil
 }
```

umohnani8:
@runcom done.

runcom:
I'm thinking that the error in the installer is caused by the fact that the bootstrapped MC in the installer lacks the image and there's a drift with the new generated ones. Could it be? The currentConfig for the node isn't there:

```
annotations: {
machine.openshift.io/machine: "openshift-machine-api/ci-op-cy4tj6qd-57a9f-hqg5h-master-0",
machineconfiguration.openshift.io/currentConfig: "master-fba43dfe0ed4d5727ae10c5b76763ad6",
```

runcom:
Looks like it's https://github.com/openshift/machine-config-operator/issues/367#issuecomment-459880141

cmd/machine-config-operator/bootstrap.go
runcom: looks like, anyway, there's still a chicken and egg issue where the MC generated at bootstrap in CI are different than the ones we now create, causing e2e-* to just fail. I'm again unsure about the flow.

runcom:
The CI failing is defintely https://github.com/openshift/machine-config-operator/issues/367 to the best of my knoledge, see https://github.com/openshift/machine-config-operator/pull/471#issuecomment-466681520

cgwalters:
Debugging this type of thing is a pain.  Last time I did it I logged into the bootstrap node and grab the bootstrap MCs from there, then diff it versus the target one.

There's also https://github.com/openshift/installer/pull/1189

The code seems fine but presumably what's going wrong is that the final rendered `InfraImage` is different (probably unset?) compared to the bootstrap one.

cgwalters:
OK I think I understand the problem.  You need to drop the change to `image-references` temporarily.  The value is being substituted in the target cluster, but not at bootstrap time.

cgwalters:
Although...one thing that changed recently is we started gating installs on `machine-config` operator being successful.  Before that...we could land changes that caused this problem.

I have a feeling we were actually *relying* on that in order to be able to make a change like this...

umohnani8:
@cgwalters by temporarily you mean till we get the installer PR in https://github.com/openshift/installer/pull/1292 ?

cgwalters:
I am thinking the flow needs to be something like this:

```
diff --git a/pkg/operator/images.go b/pkg/operator/images.go
index 825bbcc..85c1139 100644
--- a/pkg/operator/images.go
+++ b/pkg/operator/images.go
@@ -4,8 +4,11 @@ package operator
 // bootkube.sh provides.  If you want to add a new image, you need
 // to "ratchet" the change as follows:
 //
-// Add the image here and also a CLI option with a default value
-// Change the installer to pass that arg with the image from the CVO
+// 1. Add the image here and also a CLI option to bootstrap.go
+//    with a default value (e.g. something quay.io).  But *ignore* the
+//    value passed on the CLI; always use the default.
+// 1. Change the installer to pass that arg with the image from the CVO
+// 1. Change the code in bootstrap to honor the passed image
 // (some time later) Change the option to required and drop the default
 type Images struct {
 	MachineConfigController string `json:"machineConfigController"`
```

cgwalters:
To rephrase https://github.com/openshift/machine-config-operator/pull/471#issuecomment-467154348 - we need to add the CLI option to `bootstrap.go` and land the installer PR to use it, but have it be a no-op.  Only *then* can we enable it in both bootstrap and runtime in this repo.

cgwalters:
(And the need to do this dance would go away if we changed the MCO bootstrap phase to gain access to the CVO directly)

cgwalters:
Concretely, I would make a new PR that has this *single line* change:

`bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraImage, "infra-image", "quay.io/openshift/origin-pod:v4.0", "Image for Infra Containers.")`

umohnani8:
So the tests here are passing now... Did we just have to break it down into 2 steps? So we get this in first then https://github.com/openshift/installer/pull/1292?

umohnani8:
Now the tests are failing :/ Are the failures expected?

cgwalters:
Hmm.  Something failed early in the bootstrap node there.

PR #472, opened 2019-02-21
Upstream mcd dockerfile

cgwalters:
Make `make deploy-daemon` work again.

### DIFF
--- a/Dockerfile.machine-config-daemon.upstream
+++ b/Dockerfile.machine-config-daemon.upstream
@@ -3,7 +3,8 @@ WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
 RUN WHAT=machine-config-daemon ./hack/build-go.sh
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+# Temporarily until the UBI base allows us to install util-linux without entitlements
+FROM registry.fedoraproject.org/fedora:29
 COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-daemon /usr/bin/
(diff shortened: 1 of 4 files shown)


cgwalters:
AIUI this shouldn't cause `fedora:29` to be used in CI because those directly execute the `Dockerfile`, they don't run `make`.

runcom:
works for me as a stopgap for now

/approve
/lgtm

runcom:
the CI upstream still uses the other dockerfile though, right?

Dockerfile.machine-config-daemon.upstream
```diff
@@ -0,0 +1,10 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+WORKDIR /go/src/github.com/openshift/machine-config-operator
+COPY . .
+RUN WHAT=machine-config-daemon ./hack/build-go.sh
+
+# Temporarily until the UBI base allows us to install util-linux without entitlements
```
smarterclayton: This has nothing to do with UBI.  This is because we only install one repo mirror that corresponds to our working set in CI.  You can get this working locally by downloading:

https://github.com/openshift/release/blob/master/cluster/ci/config/prow/openshift/rpm-mirrors/ocp-4.0-default.repo
https://github.com/openshift/shared-secrets/blob/master/mirror/ops-mirror.pem

and putting them in the same directory, then using -v to imagebuilder or podman.

smarterclayton: I sent an email with instructions, you shouldn't need to add this.

runcom: Fixing this, thanks Clayton for pointing this out

PR #473, opened 2019-02-21
Makefile: fix io redirection for image targets

runcom:

**- What I did**

`make images` and the other targets were just creating this file named `1` because the redirection is inverted

**- How to verify it**

run `make images` and watch it not creating a file named `1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ _build-%:
 # Example:
 #    make _image-machine-config-daemon
 _image-%:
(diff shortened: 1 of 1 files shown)


cgwalters:
/lgtm

(Though personally I'd move that code into the `hack/build-image.sh` itself but eh)

runcom:
> (Though personally I'd move that code into the `hack/build-image.sh` itself but eh)

I can do that in this PR as well, just lemme know (it was too late yesterday to think about that lol)

PR #475, opened 2019-02-22
Bug 1691660: pkg/controller: always use the OSImageURL from the CVO

### LINKED ISSUE #465: Prevent OSImageURL from being set by users
I'd love a way to prevent people from creating machineconfigs that set osimageurl (and/or edit the current ones?). Ideally, I can see the CVO driving that w/o user input at all. This will reduce the chance for users to do something bad (in my view, if it's not wrong).
We could detect if a machineconfig has an osimageurl and just override it with what we have in the controllerconfig resulting in a no-op for users.

runcom:

**- What I did**

Always use the osimageurl coming from the CVO. 

close #465 

**- How to verify it**

Added a unit test to cover this change 

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -12,26 +12,21 @@ import (
 // MergeMachineConfigs combines multiple machineconfig objects into one object.
 // It sorts all the configs in increasing order of their name.
 // It uses the Ign config from first object as base and appends all the rest.
-// It uses the first non-empty OSImageURL.
-func MergeMachineConfigs(configs []*MachineConfig) *MachineConfig {
+// It uses only the OSImageURL provided by the CVO and ignores any MC provided OSImageURL.
+func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineConfig {
 	if len(configs) == 0 {
 		return nil
 	}
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
-	outOSImageURL := configs[0].Spec.OSImageURL
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
-		config := configs[idx]
-		if outOSImageURL == "" {
-			outOSImageURL = config.Spec.OSImageURL
-		}
-		outIgn = ignv2_2.Append(outIgn, config.Spec.Config)
+		outIgn = ignv2_2.Append(outIgn, configs[idx].Spec.Config)
 	}
 
 	return &MachineConfig{
 		Spec: MachineConfigSpec{
-			OSImageURL: outOSImageURL,
+			OSImageURL: osImageURL,
 			Config:     outIgn,
 		},
 	}
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -466,7 +466,15 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 		return nil
 	}
 
-	generated, err := generateRenderedMachineConfig(pool, configs)
+	cc, err := ctrl.ccLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("could not enumerate ControllerConfig %s", err)
+	}
+	if len(cc) == 0 {
+		return fmt.Errorf("ControllerConfigList is empty")
+	}
+
+	generated, err := generateRenderedMachineConfig(pool, configs, cc[0])
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


pkg/controller/render/render_controller.go
```diff
@@ -434,13 +440,20 @@ func (ctrl *Controller) garbageCollectRenderedConfigs(pool *mcfgv1.MachineConfig
...
-	generated, err := generateMachineConfig(pool, configs)
+	cc, err := ctrl.ccLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("could not enumerate ControllerConfig %s", err)
+	}
+	if len(cc) == 0 {
+		return fmt.Errorf("ControllerConfigList is empty")
+	}
+
+	generated, err := generateMachineConfig(pool, configs, cc[0])
```
runcom: maybe someone can shed some light on this snipped I found in the kubeconfig/containerruntimeconfig controllers. Is it correct to grab the controllerconfig like this? isn't there always just 1 controller config at any point in time?

runcom: @cgwalters @kikisdeliveryservice ^^^

kikisdeliveryservice:
Taking a look at this, there've been a couple PRs that change the render/controller config functions so refamiliarizing myself.

pkg/controller/render/render_controller.go
cgwalters: Yeah, AFAIK there's always just one.  It's basically a way for the template controller to react to just one aggregation of data AIUI.

But I don't see the problem with having the render controller grab it directly like this either.

cgwalters:
/approve
I like the idea though am uncertain about landing this right now given the testing we've done with the existing implementation.  But it looks generally good to me.

runcom:
I'm uncertain about landing this as well, but w/o this (assuming it works and I can add e2e) customers and users will be able to upgrade/downgrade the baseos through MCs which id love to avoid

kikisdeliveryservice:
q: are we _sure_ we don't want to allow this at all? just wondering before we go to the trouble of landing this and then having it reversed..

that said, i think this should be held until later when the os upgrades are healthy generally.

runcom:
> q: are we _sure_ we don't want to allow this at all?

I don't see any real value in allowing them doing something like that, we want to control upgrades (generally) through CVO, i've always seen osimageurl in the MCs like something for testing for now

runcom:
This has become more relevant in some hours as upgrades should follow the pattern where master are always upgraded before workers...we can play and check what users pass us of course and compare between pools but I feel the cleanest way to do it is to just forbid osimageurl altogether

runcom:
question now is, since #480, do we want to go out with the possibility for users to do this https://github.com/openshift/machine-config-operator/issues/183#issuecomment-463668093 and then break them?

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -12,25 +13,24 @@ import (
...
 
-	outOSImageURL := configs[0].Spec.OSImageURL
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
-		config := configs[idx]
-		if outOSImageURL == "" {
-			outOSImageURL = config.Spec.OSImageURL
+		mc := configs[idx]
+		if mc.Spec.OSImageURL != osImageURL {
+			glog.Infof("Detected OSImageURL %q in MachineConfig %q which is not what provided by the CVO %q", mc.Spec.OSImageURL, mc.Name, osImageURL)
```
LorbusChris: nit: what *is* provided

runcom:
removing WIP here, up for further discussion as I do really feel we shouldn't really expose this

runcom:
The development flow that this might break could be handled through #496 overriding `machine-os-content` directly (even if you can't really control the pool, but hey, we want https://github.com/openshift/machine-config-operator/issues/480 anyway)

cgwalters:
I think my vote on this is to land tests for OS updates first.  The code here looks reasonable but it's also a nontrivial change; we know today that `oc adm upgrade` works.  The problem is that doing testing here nicely depends on https://github.com/openshift/machine-config-operator/pull/496

cgwalters:
> removing WIP here, up for further discussion as I do really feel we shouldn't really expose this

I understand that but on the other hand...an admin really has to jump through hoops today to actually override `osImageURL`.

runcom:
> I think my vote on this is to land tests for OS updates first.

I'm super ok with this of course

cgwalters:
One thing I think we could envision now potentially is adding a new `custommachineconfig` CRD or so that is like `machineconfig` but for users...and wouldn't have `osImageURL`.

Or alternatively, move our rendered ones into a new `renderedmachineconfig` CRD?

runcom:
The second option sounds very reasonable to me :) (having a rendered MC CRD)

runcom:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1691660 now.
Not sure how many users are now using osimageurl directly in machineconfigs, now that upgrades are generally working fine. 
The related Bugzilla just highlights a flaw where if we have an user provided MC with osimageurl, then we don't respect the machineoscontent from the payload, which is wrong. I don't think there's a nice way to still honor the payload while leaving an osimageurl MC around so probably the time to just get rid of it (or pursue the ideas Colin had in previous comments) 

Any ideas?

runcom:
This is rebased now and ready to go!

runcom:
We could also just get rid of OSImageURL all together in machineconfigs to avoid any confusion. Haven't checked but we just get the OSImageURL from the controller config.

Issue #407, opened 2019-02-11
operator should handle MCD sync gracefully


cgwalters:
We need to start doing testing of `oc adm upgrade` including a new MCO codebase.

Rolling out a new MCC may involve e.g. new configs being generated, which could start machine updates.  If the MCD on a node is involved in trying to do that sync, it's going to be disruptive to have it be replaced by the MCO updating the MCD daemonset.

Offhand...I think if we're operating on an existing cluster, the operator should pause until the pools are idle, then roll out the MCD first?

runcom:
> Offhand...I think if we're operating on an existing cluster, the operator should pause until the pools are idle, then roll out the MCD first?

makes total sense

runcom:
what I'm seeing on `oc adm upgrade` is that the _new_ operator is (ofc) the one being rolled out, do you mean we should pause the new operator until all _old_ pools are idle?

cgwalters:
At the moment it's dangerous to kill the MCD unless it's idle.  We can and should harden things against that - there are small things we could do which would make it more reliable, but getting to "100% interruptable"...not easy.

Simplest thing I think is to enforce the property "if a pool is updating" then "ds/machine-config-daemon is complete".

> do you mean we should pause the new operator until all _old_ pools are idle?

So in this case yeah I think the new operator would come on and see the pools are active, and avoid rolling out the new MCD.

Maybe what we want is something like a `Paused` element in the controller config; the operator sets that when it detects it's upgrading, and the render controller would then stop generating new rendered MCs.  That way we ensure progress

Another thing we should do that would be complementary is for the daemon to catch the `SIGTERM` for pod termination and if it's actively working, ignore it.  And we can bump `terminationGracePeriodSeconds`.

runcom:
can this be closed due to #477

runcom:
as discussed during sync, we feel #477 cover pretty much this, closing, feel free to reopen if you feel otherwise

PR #477, opened 2019-02-22
daemon: Catch SIGTERM and ignore it if we're doing an update

cgwalters:
Ref: https://github.com/openshift/machine-config-operator/issues/407

The MCD is not ready to be killed at any arbitrary point.  Bump
the termination timeout to 5 minutes, and ignore `SIGTERM` if we're
in the middle of performing an update.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -82,6 +82,8 @@ type Daemon struct {
 	kubeletHealthzEnabled  bool
 	kubeletHealthzEndpoint string
 
+	installedSigterm bool
+
 	nodeWriter *NodeWriter
 
 	// channel used by callbacks to signal Run() of an error
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -8,11 +8,13 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"os/user"
 	"path"
 	"path/filepath"
 	"reflect"
 	"strconv"
+	"syscall"
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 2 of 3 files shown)


cgwalters:
Tested this lightly locally as far as verifying we caught `SIGTERM`; added a `time.Sleep()` there to verify I got the log message.

cgwalters:
Though a better way to do this is to install the `SIGTERM` handler *only* when we go to do an update; that way the kernel can still kill us directly when idle rather than jumping into the process only to immediately exit.

pkg/daemon/daemon.go
```diff
@@ -82,6 +85,10 @@ type Daemon struct {
 	kubeletHealthzEnabled  bool
 	kubeletHealthzEndpoint string
 
+	updateMu sync.Mutex // guards working (in Rust this would be enforced by the type system)
+	updateCond *sync.Cond // hold updateMu
```
jlebon: This doesn't seem used anywhere?

cgwalters: Yeah my original thought was that in the future where we do live updates the signal handler could watch the cond and exit as soon as we weren't updating but...we're so far from live updates.

cgwalters:
I reworked this entirely per https://github.com/openshift/machine-config-operator/pull/477#issuecomment-466513731

But only compile tested.

pkg/daemon/update.go
```diff
@@ -124,6 +126,26 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) error {
...
 
+// isUpdating returns true if the MCD is actively applying an update
+func (dn *Daemon) catchIgnoreSIGTERM() {
+	if dn.installedSigterm {
+		return
+	}
+
+	termChan := make(chan os.Signal, 1)
+	signal.Notify(termChan, syscall.SIGTERM)
+
```
rphillips: installedSigterm needs to be set... likely here?

cgwalters: Thanks, fixed!

(Though currently it doesn't matter as the MCD doesn't ever call this function more than once in the same process...this was an attempt to optimize for a future where we might)

Issue #479, opened 2019-02-22
MC* should use Ignition's validation logic to validate Ignition types


ajeddeloh:
**Description**

MCO/MCD can currently use MCs that contain Ignition types that are invalid according to Ignition. It should check that the configs are valid using [Ignition's validation logic](https://github.com/coreos/ignition/blob/spec2x/config/validate/validate.go#L108).

**Steps to reproduce the issue:**
1. Create a machine config that should be invalid (e.g. use a relative path for a file)
2. Have that work instead of failing

**Describe the results you received:**
MCD applied the change

**Describe the results you expected:**
MCO/D/something should fail when presented with an invalid config

cc @kikisdeliveryservice since I was talking about this with her
cc @abhinavdahiya since he had some ideas on using admission plugins for this

kikisdeliveryservice:
I can confirm through testing on my cluster that I am able to create and apply a machineconfig with what should be invalid ignition config fields.
For example, I am able to write a file to a relative path `home/whatever` and a mc is generated and applied successfully without error via the MCD.

runcom:
I would add that we don't need to Degrade in such a case, just (sort of) rejecting the MC if possible

ajeddeloh:
Ideally whatever is generating it should fail instead of whatever is consuming it. Invalid MCs should probably be disallowed everywhere if that's possible

kikisdeliveryservice:
as per convo with @runcom We'll have to go Degraded on bad MCs for now and then iterate once the Unreconcilable epic is done switch to that.

kikisdeliveryservice:
Started on this, will pick it up on Monday.

runcom:
just noticed that we're not setting the ignition version to:

- SSH MCs (00-*-ssh)
- registries MCs (99-*-registries)

we should be doing that afaict.

PR #476, opened 2019-02-22
[WIP] pkg/daemon: add onceFrom unit test

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

**- What I did**

Daemon in onceFrom mode can panic if it goes by a `dn.kubeClient.Something` call, this is an example that the daemon in onceFrom mode assume a kubeClient...

**- How to verify it**

run units, it shouldn't panic...

**- Description for the changelog**

PR #481, opened 2019-02-23
MCD: add ign validation check for mc.ignconfig

kikisdeliveryservice:
**- What I did**
 
- [x] Add ignition validation check on mc.ignconfig
- [x] Added unit tests for ignition validation & update existing unit tests to use valid ign configs

Closes: #479 

**- How to verify it**
create a proper machineconfig and `oc create -f` it - SUCCESS!:
```
kind: MachineConfig
metadata:
  labels:
  name: good-test
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,hello%20world%0A
          verification: {}
        filesystem: root
        mode: 420
        path: /home/core/prtest
```
create a bad/invalid machineconfig (here we are using a relative path which is bad) and apply it - Failure!
```
kind: MachineConfig
metadata:
  labels:
  name: bad-test
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,hello%20world%0A
          verification: {}
        filesystem: root
        mode: 420
        path: home/core/prtest
```
for the bad config daemon logs  `oc logs -p ...` you should see something like:
`I0226 02:48:52.743267    6205 update.go:668] Can't reconcile config worker-54afbf0177f8fb7dd414e3f606b7362d with worker-a59fd0f825db6a767251a2b3cd9563e2: Invalid Ignition config found: error: path not absolute`

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
+	"github.com/coreos/ignition/config/validate"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -206,6 +207,11 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	newIgn := newConfig.Spec.Config
 
 	// Ignition section
+	// First check if this is a generally valid Ignition Config
+	rpt := validate.ValidateWithoutSource(reflect.ValueOf(newIgn))
+	if rpt.IsFatal() {
+		return errors.Errorf("Invalid Ignition config found: %v", rpt)
+	}
 
 	// if the config versions are different, all bets are off. this probably
 	// shouldn't happen, but if it does, we can't deal with it.
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -7,6 +7,7 @@ import (
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
+	"github.com/stretchr/testify/assert"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 )
 
@@ -77,7 +78,7 @@ func TestReconcilable(t *testing.T) {
 		Spec: mcfgv1.MachineConfigSpec{
 			Config: ignv2_2types.Config{
 				Ignition: ignv2_2types.Ignition{
-					Version: "0.0",
+					Version: "2.0.0",
 				},
 			},
 		},
@@ -88,7 +89,7 @@ func TestReconcilable(t *testing.T) {
 		Spec: mcfgv1.MachineConfigSpec{
 			Config: ignv2_2types.Config{
 				Ignition: ignv2_2types.Ignition{
-					Version: "1.0",
+					Version: "2.2.0",
 				},
 			},
 		},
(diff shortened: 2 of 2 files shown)


pkg/daemon/update_test.go
```diff
@@ -318,6 +318,57 @@ func TestUpdateSSHKeys(t *testing.T) {
 	}
 }
 
+// This test should fail until Ignition validation enabled.
+// Ignition validation does not permit writing files to relative paths.
+func TestInvalidIgnConfig(t *testing.T) {
```
runcom: I do love this test and the general approach :+1:

pkg/daemon/update_test.go
```diff
@@ -318,6 +318,57 @@ func TestUpdateSSHKeys(t *testing.T) {
...
+	// expectedError is the error we will use when expecting an error to return
+	expectedError := fmt.Errorf("broken")
```
runcom: inline this error since it's not used anywhere else (also, golang prefers errors to be in the form of `expectedErr` but you can drop it altogether here)

pkg/daemon/update_test.go
```diff
@@ -318,6 +318,57 @@ func TestUpdateSSHKeys(t *testing.T) {
...
+				Storage: ignv2_2types.Storage{
+					Files: []ignv2_2types.File{
+						{Node: ignv2_2types.Node{Path: "home/core/test"}, FileEmbedded1: ignv2_2types.FileEmbedded1{Contents: tempFileContents}},
+					},
+				},
+			},
+		},
+	}
+	err := d.reconcilable(oldMcfg, newMcfg)
+	if err == nil {
```
runcom: you can use `assert.Nil(t, err, "unexpected error during reconcilable")`

pkg/daemon/update_test.go
```diff
@@ -318,6 +318,57 @@ func TestUpdateSSHKeys(t *testing.T) {
...
+		GetBootedOSImageURLReturns: []GetBootedOSImageURLReturn{},
+		RunPivotReturns: []error{
+			// First run will return no error
+			nil,
+			// Second rrun will return our expected error
+			expectedError},
+	}
+	mockFS := &FsClientMock{MkdirAllReturns: []error{nil}, WriteFileReturns: []error{nil}}
+	// Create a Daemon instance with mocked clients
+	d := Daemon{
```
runcom: let's use the exported interface `NewClusterDrivenDaemon` directly, we'll be able to avoid making assumption on the object itself and avoid falling into stuff like #476

kikisdeliveryservice: thanks for this tip!

kikisdeliveryservice:
Gah, rebase messed up my test. Fixing now.

pkg/daemon/update_test.go
kikisdeliveryservice: The whole file should use that exported interface so ill update them all in another commit once I get this generally working :smile:

pkg/daemon/update.go
```diff
@@ -180,6 +181,11 @@ func (dn *Daemon) reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	newIgn := newConfig.Spec.Config
 
 	// Ignition section
+	// First check if this is a generally valid Ignition Config
+	rpt := validate.ValidateWithoutSource(reflect.ValueOf(newIgn))
+	if rpt.IsFatal() {
+		return errors.Errorf("Invalid Ignition config found: %v", rpt)
```
runcom: Lowercase error

runcom:
@ajeddeloh ptal

pkg/daemon/update_test.go
```diff
@@ -318,16 +333,81 @@ func TestUpdateSSHKeys(t *testing.T) {
...
+			},
+		},
+	}
+	// create file to write that contains an impermissable relative path
+	tempFileContents := ignv2_2types.FileContents{Source: "data:,hello%20world%0A"}
+	tempMode := 420
+	newMcfg := &mcfgv1.MachineConfig{
+		Spec: mcfgv1.MachineConfigSpec{
+			Config: ignv2_2types.Config{
+				Ignition: ignv2_2types.Ignition{
```
runcom: As a follow up, I would create another failing case where the MC doesn't have the ignition version and it correctly fails. Not for this PR.

kikisdeliveryservice: sure thing!

pkg/daemon/update.go
kikisdeliveryservice: so you are saying `rpt, err := validate`...?

runcom: Nope, just the format of the error message :) (the actual sentence in the error, start it lowercase)

pkg/daemon/update_test.go
```diff
@@ -77,7 +78,7 @@ func TestReconcilable(t *testing.T) {
 		Spec: mcfgv1.MachineConfigSpec{
 			Config: ignv2_2types.Config{
 				Ignition: ignv2_2types.Ignition{
-					Version: "0.0",
+					Version: "2.0.0",
```
ajeddeloh: Is this supposed to be `2.2.0`?

kikisdeliveryservice: this is testing our general reconcilable logic. we don't reconcile ignition versions so this test is first checking that a change in ign versions fails (the old config and new config have diff ign versions), then the next part makes them both the same which is acceptable.

kikisdeliveryservice: those types of checks in practice come after the general ign validation that this pr adds. to make sure any machineconfig updates only make changes to the ign fields that we allow.

cgwalters:
I'm not opposed to doing this in the MCD.  However, there is IMO a strong argument for doing it in the MCC, specifically the render controller.

It's the renderer which merges in the MC fragments and generates the "rendered" i.e. "master-<hash>" version.  If we validate things there it happens once rather than on every node.  (Or if it fails, on both a  master and worker causing them to go degraded)

ajeddeloh:
@cgwalters We should do it on both ends. Compared to actually applying changes validation is cheap. IMO if we can have different parts of the system check each other instead of trusting each other that makes a more robust system and debugging easier.

cgwalters:
> We should do it on both ends.

:+1:

kikisdeliveryservice:
I think it makes more sense for the validation to be in once place bc in general ign validation doesn't say anything about the mc being reconcilable so the same mc that passes that basic validation in MCC can fail and degrade later in the MCD. 

Thinking about UX, I think it's confusing to have a MC fail in multiple places ie MCC/MCD and have to check logs in both to understand what was wrong. After using the PR on a cluster, having the errors together results in seeing ALL of an mc's failures in one place at the same time, meaning I can fix my mc completely after one failed run.

cgwalters:
> I think it makes more sense for the validation to be in once place

The logic for validating an Ignition config is in one place - Ignition itself right?  We're talking about having two callers for the same code.

> in general ign validation doesn't say anything about the mc being reconcilable

Yes - but the scenario here I think we're mainly worried about is the case where a user uploads a bad Ignition config.  "Reconciliation" is about a difference between two configs.

> Thinking about UX, I think it's confusing to have a MC fail in multiple places ie MCC/MCD and have to check logs in both to understand what was wrong.

I think we should expect admins to look at `oc get machineconfigpool` - and the status there rolls up into our high level operator status (`oc get clusteroperator/machine-config`) and from *there* rolls up into `oc get clusterversion`.

The MCC is in control of the pool and is a good place to report status.

ajeddeloh:
> I think it makes more sense for the validation to be in once place bc in general ign validation doesn't say anything about the mc being reconcilable so the same mc that passes that basic validation in MCC can fail and degrade later in the MCD.

I think that's fine to have failures in multiple places especially if it means that users can get feedback more immediately. The sooner into an update we can catch a potential problem the better. 

> Thinking about UX, I think it's confusing to have a MC fail in multiple places ie MCC/MCD and have to check logs in both to understand what was wrong.

I'm a bit out of my wheelhouse here, but it seems like we should make these kinds of errors (especially when a human is writing things) easy to surface. Even better would be to validate things before adding them to cluster (i.e. do a dry run and make sure it validates before actually applying a change). On CL we get a decent number of errors that are just from Ignition validation failure. Ignition in general tries to do as much as it can up front to make sure things are good, then also double check things when running (and do more checks it couldn't do when not running on the host).

cgwalters:
Another argument for doing this in the MCC as well - today if a bad config is generated, then it will be provided to *new* nodes that boot via the MCS and passed to Ignition where the node will fail to boot entirely because it's actually Ignition running in the initramfs parsing it.

The MCD is about incremental updates post-install.

kikisdeliveryservice:
@cgwalters & @ajeddeloh make great points. I will open a MCC PR separate from this one.

cgwalters:
>  Even better would be to validate things before adding them to cluster (i.e. do a dry run and make sure it validates before actually applying a change).

Right...in Kubernetes land that gets into (AIUI) "validating admission webhooks".  Or it looks like [CRDs can have openAPI validation](https://github.com/kubernetes/sample-controller/blob/034514e7983d05385cdfa047b73e47bc4cacfbc7/artifacts/examples/crd-validation.yaml#L12).

What you're talking about is more giving an error at `oc create -f /path/to/mc.yaml`.  What I'm talking about is having the renderer not even try to provide it to a node at update time.  What this PR is doing is validating it on each machine.

kikisdeliveryservice:
currently walking @LorbusChris through testing the PR. :)

kikisdeliveryservice:
gah aws at it again: `level=error msg="\t* aws_internet_gateway.igw: error attaching EC2 Internet Gateway (igw-0b8e5771c75e84ef7): timeout while waiting for state to become 'success' (timeout: 2m0s)"`

/test e2e-aws

kikisdeliveryservice:
Rebased against master for @LorbusChris testing. No changes to my commits.

pkg/daemon/update.go
kikisdeliveryservice: Gotcha! I'll update the whole file this way to be consistent in a separate PR! :smile:

LorbusChris:
Succesfully tested this on an AWS cluster. Thanks for the guidance @kikisdeliveryservice !

/lgtm

PR #483, opened 2019-02-23
Fix oncefrom with MachineConfig

runcom:
**- What I did**

Builds on top of #464 so we can be sure the node calls are at a minimum required. Looks like onceFrom with a MachineConfig still requires a cluster to be around (not the case for raw ignition).
The patch makes sure a kube and mc clients are around when running in onceFrom with a machineconfig avoiding any panic.

This also incorporates https://github.com/openshift/machine-config-operator/pull/476 and it's now correctly passing.

/cc @ashcrow ptal

**- How to verify it**


runcom:
/hold 

holding cause it requires #464

runcom:
The daemons are failing with something related to https://github.com/openshift/machine-config-operator/pull/463:
```
E0223 19:11:07.625913    5505 daemon.go:444] Fatal error checking initial state of node: Checking initial state: parsing booted osImageURL: invalid reference format
E0223 19:11:07.625940    5505 writer.go:90] Marking degraded due to: Checking initial state: parsing booted osImageURL: invalid reference format
I0223 19:11:07.636545    5505 daemon.go:446] Entering degraded state; going to sleep
```

runcom:
:thinking: 

/retest

runcom:
/hold cancel

rebased and ready to be reviewed

runcom:
origin flakes

/retest

runcom:
openshift-sdn flake (?) - reported

/retest

cmd/machine-config-daemon/start.go
```diff
@@ -81,13 +85,39 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+		ctx *common.ControllerContext
+	)
 	glog.Info("starting node writer")
 	nodeWriter := daemon.NewNodeWriter()
 	go nodeWriter.Run(stopCh)
 
+	cb, err := common.NewClientBuilder(startOpts.kubeconfig)
+	if err != nil {
+		if startOpts.onceFrom != "" {
+			glog.Info("Cannot initialize ClientBuilder, likely in onceFrom mode with Ignition")
```
cgwalters: "likely"?  We just checked it :wink: 

Why not just

```
if startopts.onceFrom == "" {
   glog.Fatalf(...)
}
```

cmd/machine-config-daemon/start.go
```diff
@@ -81,13 +85,39 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+
+	var (
+		kubeClient kubernetes.Interface
+		mcClient   mcfgclientset.Interface
+	)
+
+	if cb != nil {
+		kubeClient, err = cb.KubeClient(componentName)
+		if err != nil {
+			glog.Info("Cannot initialize kubeClient, likely in onceFrom mode with Ignition")
```
cgwalters: Confused by this, won't `cb == nil` if we're in oncefrom?

pkg/daemon/daemon.go
```diff
@@ -149,6 +143,14 @@ const (
 	updateDelay = 5 * time.Second
 )
 
+type onceFromOrigin int
+
+const (
+	onceFromUnknownConfig onceFromOrigin = iota
+	onceFromLocalConfig
+	onceFromRemoteConfig
```
cgwalters: I like this!

cgwalters:
I rebased this :surfing_man:

cgwalters:
@vrutkovs One thing that would help us...are there use cases for *both* oncefrom with raw Ignition and from a MachineConfig object pulled from the cluster?  Why do we need both?

vrutkovs:
>are there use cases for both oncefrom with raw Ignition and from a MachineConfig object pulled from the cluster? Why do we need both?

In scaleup we only use Ignition, can't think of a case where MachineConfig + --once-from would be used

sdodson:
/retest
VPC limits

cgwalters:
We can address the nits above in followup.

/lgtm

cmd/machine-config-daemon/start.go
runcom: we haven't checked that we're in onceFrom _with_ ignition, that's why, the log is confusing though, need to reword it

Issue #487, opened 2019-02-24
upgrade: version sync failure


cgwalters:
OK, in my local upgrade test now:

```
  - lastTransitionTime: 2019-02-24T14:04:19Z
    message: 'Failed to resync 3.11.0-723-g617d35e9-dirty because: timed out waiting
      for the condition during syncRequiredMachineConfigPools: pool master has not
      progressed to latest configuration: controller version mismatch for 99-master-8509bfff-379c-11e9-b58a-52fdfc072182-registries
      expected 3.11.0-723-g617d35e9-dirty has 3.11.0-716-g96e07660-dirty'
    reason: 'timed out waiting for the condition during syncRequiredMachineConfigPools:
      pool master has not progressed to latest configuration: controller version mismatch
      for 99-master-8509bfff-379c-11e9-b58a-52fdfc072182-registries expected 3.11.0-723-g617d35e9-dirty
      has 3.11.0-716-g96e07660-dirty'
    status: "True"
    type: Failing
```

The pod/images state looks OK offhand:

```
$ oc logs deploy/machine-config-controller | head -1
I0224 13:52:34.705761       1 start.go:47] Version: 3.11.0-723-g617d35e9-dirty
$ oc logs deploy/machine-config-operator | head -1
I0224 13:52:17.628611       1 start.go:40] Version: 3.11.0-723-g617d35e9-dirty
```

So presumably the operator code isn't correctly picking up the new versions?

runcom:
What's the expected behavior for this to be correct? How do I reproduce? I'm familiar with the error but did it break?

runcom:
It looks like operator and controllers progressed but the pool was still syncing with an older version?

runcom:
The situation seems to be:

- The master pool was correctly generated from the new controller with the new version
- The master pool from above contains an MC fragment (container runtime registries) that was generated by the old controller

runcom:
oooh, ok I found the bug :tada:

PR #461, opened 2019-02-19
Add extra filter for checking if registries have changed

umohnani8:
A resync happens about every 20 minutes, which sends an updated
event to the image informer even if nothing in the image CR has
changed. Adding and extra filter that checks if there has been
any changes to the registries part of the image CR before syncing
the image handler again.

Helps fix https://github.com/openshift/machine-config-operator/issues/453 where changes to the image config will not be applied unless and actual change happened in the CR.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

PR #489, opened 2019-02-24
 pkg/controller: fix container runtime registries generation

runcom:
The CRC (container runtime config) controller recently added a check to
avoid resyncing and recreating the very same registries config if
nothing has changed on the image crd side [1]. While that's correct,
during an upgrade, the controllers need to generate the MC fragments
using the controller version they're at. Since we weren't checking the
versions of the controller that generated the registries config, we
wrongly assumed the configurations were equal and never generated a new
one with the new controller.
This patch fixes that by adding a version check before skipping a
regeneration on equal content in the registries configs.
Fixes: https://github.com/openshift/machine-config-operator/issues/487

[1] https://github.com/openshift/machine-config-operator/pull/461

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -567,8 +567,13 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 			isNotFound := errors.IsNotFound(err)
 			rci := createNewRegistriesConfigIgnition(registriesTOML)
 			if !isNotFound && equality.Semantic.DeepEqual(rci, mc.Spec.Config) {
-				applied = false
-				return nil
+				// if the configuration for the registries is equal, we still need to compare
+				// the generated controller version because during an upgrade we need a new one
+				mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
+				if mcCtrlVersion == version.Version.String() {
+					applied = false
+					return nil
+				}
 			}
 			if isNotFound {
 				mc = mtmpl.MachineConfigFromIgnConfig(role, managedKey, &ignv2_2types.Config{})


pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -567,14 +567,19 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
 			if !isNotFound && equality.Semantic.DeepEqual(rci, mc.Spec.Config) {
-				applied = false
-				return nil
+				// if the configuration for the registries is equal, we still need to compare
+				// the generated controller version because during an upgrade we need a new one
+				mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
+				if mcCtrlVersion == version.Version.String() {
+					applied = false
+					return nil
+				}
```
runcom: I guess we can also just update that annotation to the new version, but let's keep it this way for now

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -567,14 +567,19 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
 			}
 			if isNotFound {
 				mc = mtmpl.MachineConfigFromIgnConfig(role, managedKey, &ignv2_2types.Config{})
 			}
 			mc.Spec.Config = createNewRegistriesConfigIgnition(registriesTOML)
-			mc.ObjectMeta.Annotations = map[string]string{
+			mc.Annotations = map[string]string{
```
cgwalters: What's the difference here?  It looks like we have `mc.ObjectMeta.Annotations` in another place in this code.

runcom: none, but ObjectMeta is an embedded field and Go let's you access embedded fields' fields' directly on the top level struct...nothing changes functionality wise

cgwalters:
This looks sane to me but - today we don't have any tests that cover this.  Did you test this by hand?

I think we could add something to e2e-aws-op that builds another release payload from the same code and at least verify that a no-op upgrade works.  (Don't have to do it in this PR, just brainstorming)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
runcom: I, of course, can change it back...it's just OCD kicking lol

runcom:
> This looks sane to me but - today we don't have any tests that cover this. Did you test this by hand?

I'm testing this right now on a cluster but unlikely to finish the whole testing today but that's the plan



> I think we could add something to e2e-aws-op that builds another release payload from the same code and at least verify that a no-op upgrade works. (Don't have to do it in this PR, just brainstorming)

do we have `oc` around in the CI env? if yes, then yeah, let's do that, I'll follow up with something like that.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
runcom: ok. moved this back for consistency...

runcom:
@umohnani8 ptal

runcom:
Good to see that CI is now cooperating again...

runcom:
> /lgtm

just FYI, I'm spinning up two cluster, one to reproduce using a CI payload (broken) and another one to upgrade to my own payload built from this PR. First should fail with #487, second should hopefully go through (if no other errors like daemon SIGTERM kicks in?)

runcom:
oh gosh, this was fast lol will report back anyway I guess

runcom:
as a side note, the bug in the linked issue isn't 100% reproducible to my testing, I've upgraded to the very same version as Colin's and did not get the error

runcom:
nevermind, I was actually using the very same version, so it was a no-op

runcom:
I was able to upgrade to my custom payload just fine :+1: it works

runcom:
ok, properly tested this one on 2 clusters, the first cluster is correctly failing w/o this patch on upgrade, whether the second one just goes through and successfully upgrade.

First cluster MCs broken (notice the registries configs being on `-dirty`):
```
23:10:40 [github.com/openshift/machine-config-operator] ‹master› oc get machineconfigs                     
NAME                                                        GENERATEDBYCONTROLLER        IGNITIONVERSION   CREATED
00-master                                                   3.11.0-723-g617d35e9         2.2.0             24m
00-master-ssh                                               3.11.0-723-g617d35e9                           24m
00-worker                                                   3.11.0-723-g617d35e9         2.2.0             24m
00-worker-ssh                                               3.11.0-723-g617d35e9                           24m
01-master-container-runtime                                 3.11.0-723-g617d35e9         2.2.0             24m
01-master-kubelet                                           3.11.0-723-g617d35e9         2.2.0             24m
01-worker-container-runtime                                 3.11.0-723-g617d35e9         2.2.0             24m
01-worker-kubelet                                           3.11.0-723-g617d35e9         2.2.0             24m
99-master-8bd634e9-387d-11e9-a758-06a3c1aa817c-registries   3.11.0-723-g617d35e9-dirty                     22m
99-worker-8bd751d1-387d-11e9-a758-06a3c1aa817c-registries   3.11.0-723-g617d35e9-dirty                     22m
master-2cb481673ed0b572d9a8fa9d2612d31c                     3.11.0-723-g617d35e9-dirty   2.2.0             24m
master-31bbbd1cb081eb6182c0f415f887ae99                     3.11.0-723-g617d35e9         2.2.0             5m4s
master-5597f9e68d6a65e7a7ace765139d93d8                     3.11.0-723-g617d35e9-dirty   2.2.0             24m
master-f06730041e1db83b72af95824d4d48b3                     3.11.0-723-g617d35e9         2.2.0             4m56s
worker-05e89433151d256339179b1f5703fd15                     3.11.0-723-g617d35e9         2.2.0             4m56s
worker-db82c7e62e8082acaedf117785f005d6                     3.11.0-723-g617d35e9-dirty   2.2.0             24m
worker-f375d3645f8c66d61c6fed104af20521                     3.11.0-723-g617d35e9         2.2.0             24m
```

runcom:
@cgwalters I do completely agree our CI needs to have a smoke upgrade testing (be it with `oc adm upgrade` in our e2e, or being an official CI job that _openshift_ provides) to be able to catch silly regressions like this one.

PR #490, opened 2019-02-24
consistenly check MC fragments version

runcom:
**- What I did**

- simplify the version check for MC fragments (and generated one), we check version anyway for all cause the logic was flawed (`t.versionCheckRequired` wasn't honored in the actual version check if the annotation was there anyway)
- added some comments to the code to clarify why we do something...

**- How to verify it**


pkg/operator/status.go
```diff
@@ -222,31 +222,24 @@ func isMachineConfigPoolConfigurationValid(pool *mcfgv1.MachineConfigPool, versi
...
 
+		// we check that all the machineconfigs (generated, and those that were used to create generated)
+		// were generated by correct version of the controller.
 		v, ok := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
-		if t.versionCheckRequired && !ok {
-			return fmt.Errorf("%s must be created by controller version %s", t.name, version)
+		if !ok {
+			return fmt.Errorf("%s must be created by controller version %s", mcName, version)
 		}
 		if ok && v != version {
```
runcom: `t.versionCheckRequired` was missing here, effectively doing the version check for everything as long as the annotation was there, so it's just irrelevant to have that field around...

runcom: we set the annotation anyway for _anything_ so there's little to check if the annotation is there

runcom:
build flake

/retest

LorbusChris:
just a note: #471 includes this change in mcc/start.go as well

runcom:
I'll drop that thanks!

kikisdeliveryservice:
Minor nit pr/commit message: "consistenly" -> "consistently"

runcom:
updated and dropped the commit already in #471

runcom:
aws limit :sob:

kikisdeliveryservice:
@runcom we are hitting limits everywhere, gonna elevate the issue in slack.

runcom:
Do we want this for 4.1?

LorbusChris:
I'd say yes.

runcom:
@cgwalters @kikisdeliveryservice @LorbusChris ptal

kikisdeliveryservice:
@runcom just to double check : this is verifying that the controller version (which for example, can now be seen in `oc get machineconfigs`  ->  `4.0.0-alpha.0-9-gfc1a99e6-dirty`) matches the version we expect? and we want to ensure that they are all generated by the same version?

is this to keep track of which MC versions a user is using (in case we change that)? or..?

just trying to understand why we need the check

runcom:
> @runcom just to double check : this is verifying that the controller version (which for example, can now be seen in `oc get machineconfigs` -> `4.0.0-alpha.0-9-gfc1a99e6-dirty`) matches the version we expect? and we want to ensure that they are all generated by the same version?
> 
> is this to keep track of which MC versions a user is using (in case we change that)? or..?
> 
> just trying to understand why we need the check

yeah, the check makes sure every config is generated by the correct version of the operator (operator = (operator, controllers, daemon, server).

Other than the check itself, _this PR isn't changing the behavior_, just simplifying the code w/o using a struct.

cgwalters:
LGTM, giving to

/assign kikisdeliveryservice

for final review and lgtm.

pkg/operator/status.go
```diff
@@ -222,31 +222,30 @@ func isMachineConfigPoolConfigurationValid(pool *mcfgv1.MachineConfigPool, versi
...
 	}
-	for _, t := range tasks {
-		mc, err := machineConfigGetter(t.name)
+	for _, mcName := range mcs {
+		mc, err := machineConfigGetter(mcName)
 		if err != nil {
 			return err
 		}
 
+		// We check that all the machineconfigs (generated, and those that were used to create generated,
```
kikisdeliveryservice: really really like these comments btw!

PR #495, opened 2019-02-25
daemon: Only print status if os == RHCOS

cgwalters:
Don't crash on non-rpm-ostree based systems.


runcom:
/retest

aws limit hit

PR #474, opened 2019-02-22
log rpm-ostree status, and fix hack/

cgwalters:

PR #497, opened 2019-02-26
update hack/push-image.sh to rename images

kikisdeliveryservice:
**- What I did**
PR #474 broke hack/push-image.sh (and my workflow) by renaming the generated
images.  This PR fixes the names used in this script, so the
latest can be pushed to a remote repo like dockerhub.

I'd like to eventually re-add VERSION support, but I don't have time, so it's commented out for now.

**- How to verify it**
try using the push-image.sh note how it doesn't work. update to use this pr and now it successfully pushes like it did pre-474

### DIFF
--- a/hack/push-image.sh
+++ b/hack/push-image.sh
@@ -55,7 +55,6 @@ fi
 for IMAGE_TO_PUSH in $TOPUSH; do
 	NAME="${IMAGE_TO_PUSH#Dockerfile.}"
 	set -x
-	podman push "${NAME}:${VERSION}" "${REPO}/origin-${NAME}:${VERSION}"
-	podman push "${NAME}:${VERSION}" "${REPO}/origin-${NAME}:latest"
+	podman push "localhost/${NAME}:latest" "${REPO}/origin-${NAME}:latest"
 	set +x
 done


kikisdeliveryservice:
I'd like to get this in so I don't have to update this script every time I pull master.

cgwalters:
> I'd like to get this in so I don't have to update this script every time I pull master.

It should work to copy the script outside of the repo, e.g. `cp hack/build-image.sh ~/build-image.sh`.  But is there a reason you're not using `make deploy-$x`?  Going forward it's going to be important that we're all using a similar workflow.

kikisdeliveryservice:
I haven't been able to get the `make` workflow working with AWS nor a ton of time to troubleshoot it and assumed it was more for people who use libvirt.

That said, as it stands, push-image.sh doesn't work so it should either be updated or removed from the repo.

hack/push-image.sh
```diff
@@ -55,7 +55,7 @@ fi
 for IMAGE_TO_PUSH in $TOPUSH; do
 	NAME="${IMAGE_TO_PUSH#Dockerfile.}"
 	set -x
-	podman push "${NAME}:${VERSION}" "${REPO}/origin-${NAME}:${VERSION}"
-	podman push "${NAME}:${VERSION}" "${REPO}/origin-${NAME}:latest"
+#	podman push "/localhost/${NAME}:${VERSION}" "${REPO}/origin-${NAME}:${VERSION}"
```
cgwalters: Don't think we need this line anymore right?

kikisdeliveryservice: you're right - done!

kikisdeliveryservice:
Rate limits again...

/retest

PR #503, opened 2019-02-26
Eliminate use of cpu limits

derekwaynecarr:
**- What I did**
I made updates to eliminate the use of cpu limits for components that run on control plane.  The intention is to depend entirely on CPU requests and CFS sharing for allocating CPU time.  Setting limits just induces unnecessary latency.

**- How to verify it**
Pass this smoke test:
https://github.com/openshift/origin/pull/22095/files

No longer set cpu limits for components in favor of just setting cpu requests.

### DIFF
--- a/examples/machinecontrollerconfig.deploy.yaml
+++ b/examples/machinecontrollerconfig.deploy.yaml
@@ -19,9 +19,6 @@ spec:
         - "start"
         - "--resourcelock-namespace=kube-system"
         resources:
-          limits:
-            cpu: 20m
-            memory: 50Mi
           requests:
             cpu: 20m
             memory: 50Mi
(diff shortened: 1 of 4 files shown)


examples/machinecontrollerconfig.deploy.yaml
```diff
@@ -19,9 +19,6 @@ spec:
         - "start"
         - "--resourcelock-namespace=kube-system"
         resources:
-          limits:
-            cpu: 20m
-            memory: 50Mi
```
runcom: this is also dropping memory, is it ok?

runcom: ok this is just an example

PR #506, opened 2019-02-27
disable swappiness on workers and master

rphillips:
Disable swappiness to hint to the kernel there is not any swap.

**- What I did**

I'm looking into a [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1679585) regarding performance and noticed the swappiness level set to 60 on masters and workers. My test cluster started having high cpu+memory in kswapd when the cluster was overloaded.

**- How to verify it**
`sysctl vm.swappiness` on a node.

set vm.swappiness=0 on masters and workers

### DIFF
--- a/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-sysctl.d-swappiness.conf
+++ b/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-sysctl.d-swappiness.conf
@@ -1,0 +1,6 @@
+contents:
+  source: data:,vm.swappiness%3D0%0A
+  verification: {}
+filesystem: root
+mode: 420
+path: /etc/sysctl.d/swappiness.conf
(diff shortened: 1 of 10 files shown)


cgwalters:
But the kernel *knows* there's not any swap.  See [this doc](https://www.kernel.org/doc/gorman/html/understand/understand013.html) which talks about kswapd.  It's more like "pagereclaimd".  For example, on a system without swap, an executable binary (in a container image, say `/var/lib/containers/.../merged/usr/bin/bash`) may no longer be actively used.  AIUI kswapd is the kernel thread that will end up reclaiming that when there's memory pressure.

I think high CPU usage from kswapd is just evidence of memory pressure.

Did you test this and see that it had an improvement?

rphillips:
Thanks for the explanation. I'll test this on Wednesday and see what else might be contributing to the memory pressure.

PR #508, opened 2019-02-28
install: Fix reference to pod image

cgwalters:
The name needs to *exactly* match that in `image-references`
in order to be substituted.

### DIFF
--- a/install/image-references
+++ b/install/image-references
@@ -26,7 +26,7 @@ spec:
   - name: pod
     from:
       kind: DockerImage
-      name: quay.io/openshift/pod:v4.0
+      name: quay.io/tectonicshift/pod:42
   - name: setup-etcd-environment
     from:
       kind: DockerImage
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -85,7 +85,7 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 		MachineOSContent:        bootstrapOpts.oscontentImage,
 		Etcd:                    bootstrapOpts.etcdImage,
 		SetupEtcdEnv:            bootstrapOpts.setupEtcdEnvImage,
-		InfraImage:              bootstrapOpts.infraImage,
+		InfraImage:              "quay.io/openshift/origin-pod:v4.0",
 	}
 
 	if err := operator.RenderBootstrap(
--- a/install/0000_30_machine-config-operator_02_images.configmap.yaml
+++ b/install/0000_30_machine-config-operator_02_images.configmap.yaml
@@ -11,5 +11,6 @@ data:
       "machineConfigServer": "docker.io/openshift/origin-machine-config-server:v4.0.0",
       "etcd": "quay.io/coreos/etcd:v3.3.10",
       "setupEtcdEnv": "registry.svc.ci.openshift.org/openshift/origin-v4.0:setup-etcd-environment",
+      "dummyImage": "quay.io/tectonicshift/pod:42",
       "infraImage": "quay.io/openshift/origin-pod:v4.0"
     }


umohnani8:
The `origin-` gets dropped? I got the image name from https://quay.io/repository/openshift/origin-pod?tag=v4.0&tab=tags.

cgwalters:
Oh, then yeah it needs to be the other way around.

umohnani8:
Right, I see what I did there. Thanks for the fix!

cgwalters:
So this has drift still, ugh:
`I0228 21:55:30.673793   20978 daemon.go:1041] While getting MachineConfig master-c972857d9ccbede3699c273abb1ebd65, got: machineconfigs.machineconfiguration.openshift.io "master-c972857d9ccbede3699c273abb1ebd65" not found. Retrying..`

cgwalters:
At least though the installer PR got in, so we only need to land a change in this repo.

cgwalters:
OK I think we're actually back to my original theory - we have the image in the payload now :tada: but need to have a no-op CLI option, land this PR, then land the installer one, then enable the CLI option.

wking:
Can I ask some questions about the pod/infra image?  It's from [here][1] (equivalently [here][2]), right?  And it's a sleeper to hold pod namespaces open?  But CRI-O has [its own][3] and Kubernetes has [its own][4].  And the OpenShift Go seems to have forked off Kubernetes way back in openshift/origin#852.  Then Kubernetes ported to C (kubernetes/kubernetes@f3690e2d5e, kubernetes/kubernetes#23009).  Then CRI-O forked off.  Then Kubernetes added `SIGCHLD` reaping (kubernetes/kubernetes@81d27aa2396, kubernetes/kubernetes#36853).  Do we still want our very elderly Go to be the source of our pod-holding process?

On the `SIGCHLD` front, there's some reaping discussion in openshift/origin#1242 with a fix in openshift/origin#1264 (and then some revisiting in openshift/origin#17743 and openshift/origin#18992).  That reaper now seems to only be [launched by `oc observe`][5], which seems like an unlikely candidate for a generic, pod-holding process (but maybe we really do use `oc` for everything?).

Pulling the infra image from the release payload (as in #471) makes sense to me, but we seem to have made an odd choice to provide that image ([Kubernetes' `pause.c`][4] seems to have the most active maintenance and maturity).  Is there more background around that choice somewhere?  I guess one benefit to getting this from the release image is that it will be easy to update later ;).

[1]: https://github.com/openshift/images/tree/2f60da39a9d2e5cc00293b8ec7ad559fcd32446a/pod
[2]: https://github.com/openshift/origin/tree/777ed6f519dbeed001e29cbc857a39bc3bb0e6cc/images/pod
[3]: https://github.com/kubernetes-sigs/cri-o/tree/v1.12.6/pause
[4]: https://github.com/kubernetes/kubernetes/blob/v1.12.6/build/pause/pause.c
[5]: https://github.com/openshift/origin/blob/777ed6f519dbeed001e29cbc857a39bc3bb0e6cc/pkg/oc/cli/observe/observe.go#L413-L414

runcom:
@wking I would open an issue in CRI-O and discuss there and talk to @mrunal I guess. This feature comes from them and we're merely adding core to support that (though I agree on your points).

cgwalters:
> Can I ask some questions about the pod/infra image

That's an *amazing* comment even for you!  I still wonder if you have scripts to do this kind of lookup or maybe a machine learning system :wink: 

Anyways I think we need all images to come from the release payload; can't be pulling random images we don't control particularly for critical infra.

Once we have that fixed, then we should definitely circle back and be sure we've picked up fixes like the `SIGCHLD` reaping.

(Honestly though it would be a lot saner if one could just tell the kernel to do it, but eh)

install/image-references
```diff
@@ -26,7 +26,7 @@ spec:
   - name: pod
     from:
       kind: DockerImage
-      name: quay.io/openshift/pod:v4.0
+      name: quay.io/tectonicshift/pod:42
```
derekwaynecarr: what is this image?

we need to use the pod image from the release payload.

cgwalters: This image doesn't exist and we aren't actually using it - we're really using the `openshift/pod:v4.0` image still.  The reason to add this new dummy image is to help "ratchet" the change here so we can land [the installer PR](https://github.com/openshift/installer/pull/1292).

Yes, it is *very* confusing...

cgwalters: See https://github.com/openshift/machine-config-operator/pull/471#issuecomment-467153179

cgwalters:
OK, this passed `e2e-aws-op`, so can someone `lgtm` it?

What's going on here is that we have `pod` in `image-references` but it's substituting an unused value.  I believe that was the problem before, if there's something in `image-references` but it's not substituting anything then:

```
	if len(repositories) == 0 {
		glog.V(5).Infof("No images are mapped, will not replace any contents")
		return NopManifestMapper, nil
	}
```



 so that the installer PR should be able to land, but we're not actually *using* it.

I think t

PR #509, opened 2019-02-28
docs/Update-SSHKeys: fix SSH key example

miabbott:
The previous version had the SSH keys listed as `- ssh rsa AAAA...`, so I
thought the MCO was doing some sort of magic parsing.  Turns out it
needed the commonly seen representation of `- ssh-rsah AAAAA...`

I added a few examples showing how you can provide different key types
and cleaned up some trailing whitespace.


kikisdeliveryservice:
thanks for this!!

/lgtm

PR #511, opened 2019-02-28
install: use origin-v4.0 etcd image

hexfusion:
We are currently using upstream etcd image which is based on Alpine Linux OS. This PR updates this to the correct origin-v4.0 etcd image build.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1679834

/cc @abhinavdahiya


kikisdeliveryservice:
install/image-references change overlaps with the change being made in #508 
cc: @cgwalters

cgwalters:
>  install/image-references change overlaps with the change being made in #50

They're separate images though the patches probably will conflict "textually".

kikisdeliveryservice:
thanks @cgwalters !

/hold cancel

cgwalters:
/approve

Just want to double check this is a 4.0 target, but I'm pretty sure it should be.

derekwaynecarr:
this must be a 4.0 change.

PR #514, opened 2019-03-01
initial pass at populating the os version label

rphillips:
**- What I did**

Enumerates the OS version from /etc/os-release and creates a `node.openshift.io/os_version={OS_VERSION}` label.

**- How to verify it**
oc get node [node name] -o yaml

Add an initial OS Version to label (`node.openshift.io/os_version={OS_VERSION}`) to nodes.

**- Some background info**
Kubelet --node-labels only get created on kubelet registration, so changes will need to happen with a daemonset. Liggit has a [number](https://github.com/kubernetes/kubernetes/issues/59314) of [comments](https://github.com/kubernetes/kubernetes/issues/59314#issuecomment-362869655) in upstream saying the kubelet should not update labels, so a custom patch would probably never be merged upstream (IMO). There are differing opinions on the ticket.

This PR is phase 1) create the label with the starting version. The second phase will be to create a daemonset to update the label on RHEL7->RHEL8 upgrades. A final third phase, when RHEL7 is not supported anymore, we can retire the daemonset and delete the code/daemonset.

/cc @jeremyeder @sjenning @smarterclayton @derekwaynecarr

### DIFF
--- a/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
@@ -7,6 +7,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  EnvironmentFile=/etc/os-release
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
   EnvironmentFile=-/etc/kubernetes/kubelet-env
 
@@ -19,7 +20,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/master \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${VERSION},node.openshift.io/os_id=${ID} \
         --minimum-container-ttl-duration=6m0s \
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider=aws \
--- a/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
@@ -7,6 +7,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  EnvironmentFile=/etc/os-release
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
   EnvironmentFile=-/etc/kubernetes/kubelet-env
 
@@ -18,7 +19,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_version=${VERSION},node.openshift.io/os_id=${ID} \
         --minimum-container-ttl-duration=6m0s \
         --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
         --client-ca-file=/etc/kubernetes/ca.crt \
--- a/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/libvirt/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/libvirt/units/kubelet.service
@@ -7,6 +7,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  EnvironmentFile=/etc/os-release
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
   EnvironmentFile=-/etc/kubernetes/kubelet-env
 
@@ -18,7 +19,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_version=${VERSION},node.openshift.io/os_id=${ID} \
         --minimum-container-ttl-duration=6m0s \
         --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
         --client-ca-file=/etc/kubernetes/ca.crt \
--- a/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/none/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/worker/01-worker-kubelet/none/units/kubelet.service
@@ -7,6 +7,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  EnvironmentFile=/etc/os-release
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
   EnvironmentFile=-/etc/kubernetes/kubelet-env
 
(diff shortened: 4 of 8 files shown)


kikisdeliveryservice:
Shouldn't this be against master-4.1?

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
```diff
@@ -18,7 +18,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${OS_VERSION} \
```
kikisdeliveryservice: why is the worker label changed to master in the `01-worker-kubelet`?

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/libvirt/units/kubelet.service
```diff
@@ -18,7 +18,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${OS_VERSION} \
```
kikisdeliveryservice: same q: why is the worker label changed to master in the `01-worker-kubelet`?

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/none/units/kubelet.service
```diff
@@ -18,7 +18,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${OS_VERSION} \
```
kikisdeliveryservice: same q: why is the worker label changed to master in the `01-worker-kubelet`?

templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
```diff
@@ -20,7 +20,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${OS_VERSION} \
```
kikisdeliveryservice: same q: why is the worker label changed to master in the `01-worker-kubelet`?

kikisdeliveryservice:
Until there is confirmation that this should go into 4.0

/hold

rphillips:
@kikisdeliveryservice changed to the 4.1 branch

rphillips:
There has been some discussion on aos-devel regarding how networking can figure out if the underlying OS is RHEL7 or RHEL8.

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
rphillips: It isn't changing. The comma is adding a another label called node.openshift.io/os_version with the os version. These node labels only get added on kubelet registration, so we will need a daemonset to 'update' the label on RHEL7->RHEL8 upgrades.

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/libvirt/units/kubelet.service
rphillips: ditto ^

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
kikisdeliveryservice: to clarify it seems to have gone from `/worker` to `/master, node.openshift.io/os....` 

why is that master?

kikisdeliveryservice:
Thanks @rphillips !
/hold cancel

pkg/controller/template/test_data/templates/worker/01-worker-kubelet/aws/units/kubelet.service
rphillips: good catch. I typoed that.

rphillips: @kikisdeliveryservice got that... transposition... fixed. thanks!

kikisdeliveryservice: np!

derekwaynecarr:
this is only needed for 4.1.  In 4.0, we know a node not labeled is rhel7, and in 4.1, pods that require rhel8 will have nodeSelectors that will hold scheduling until a label is present.  In 4.1, rhel7 workers are new nodes, and would have newly registered labels which should be fine with general approach.

templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
```diff
@@ -20,7 +20,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/worker \
+        --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_version=${OS_VERSION} \
```
squeed: Why not just add `/etc/os-release` as an EnvironmentFile and pull it out from there?

rphillips: that does simplify things. fixed!

sdodson:
Should we add similar support so that we can differentiate between RHCOS and RHEL Server or other?

rphillips:
@sdodson which variable would you like to add from os-version?

ashcrow:
This looks good to me. Just to remind folks, you'll end up getting a version like:

```400.7.20190301.0```

Related: 
- Design: https://github.com/openshift/os/issues/259
- Impl: https://github.com/openshift/os/issues/360

**Comment by sdodson** [comment]
The variable ID from /etc/os-release seems like a good source for the value but we probably want to name the label differently. os_id perhaps?

**Comment by rphillips** [comment]
@sdodson My general thought was to have os_version correspond to ${VERSION} within the os-release. If we want to add more variables from os-release, then they might have a 1 to 1 mapping. os_id would map to ${ID} within os-release.

**Comment by sdodson** [comment]
Yes, that's what I meant, an additional label for os_id.

**Comment by rphillips** [comment]
I think this PR is ready for approval unless there are any further comments.

**Comment by rphillips** [comment]
@sdodson added the ID

**Comment by rphillips** [comment]
@kikisdeliveryservice after talking to @sjenning, we need this in 4.0.

I've switched the target branch back to master.

**Comment by kikisdeliveryservice** [comment]
thanks for the update @rphillips 

Could you amend the body of the commit messages to add a little more info?  For ex, os version using: "Enumerates the OS version from /etc/os-release and creates a node.openshift.io/os_version={OS_VERSION} label." and something substantive for the id label commit.

**Comment by kikisdeliveryservice** [comment]
@runcom @ashcrow @cgwalters 

Any comments on this PR?

**Comment by rphillips** [comment]
@kikisdeliveryservice updated the commit message and comment.

**Comment by kikisdeliveryservice** [comment]
thanks @rphillips !

Spinning up a cluster to test.

**Comment by runcom** [comment]
code lgtm, I'd love @sjenning to approve

File: pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service [hunk 6]
```diff
@@ -19,7 +20,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --allow-privileged \
-        --node-labels=node-role.kubernetes.io/master \
+        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_version=${VERSION},node.openshift.io/os_id=${ID} \
```
vrutkovs: This breaks RHEL scaleup:

```
kubelet_node_status.go:92] Unable to register node "ip-10-0-156-42.ec2.internal" with API server: Node "ip-10-0-156-42.ec2.internal" is invalid: metadata.labels: Invalid value: "7 (Core)": a valid label must be an empty string or consist of alphanumeric characters, ''-'', ''_'' or ''.'', and must start and end with an alphanumeric character (e.g. ''MyValue'',  or ''my_value'',  or ''12345'', regex used for validation is ''(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'')'
```

PR #517, opened 2019-03-01
BUG 1670700: *: add support for etcd metrics proxy

hexfusion:
This PR adds support for new TLS assets to support the etcd metrics proxy. This proxy will create a separate chain of trust for metrics consumer which will not give allow access to server data directly. The proxy itself is a feature of etcd called the `grpc-proxy`.  By exposing a separate metrics port on the server and then using the proxy to listen to that port we eliminate direct gRPC transactions. 

The grpc-proxy forwards only  `/metrics` endpoint requests to the server.

/cc @brancz @abhinavdahiya 

depends on: https://github.com/openshift/installer/pull/1291 **merged**


runcom:
@hexfusion could you please also add a target for these PRs? We're after freeze and if this is a feature I'd love someone like Clayton or others to approve

kikisdeliveryservice:
Is this for 4.0 or 4.1?

brancz:
As mentioned in https://github.com/openshift/installer/pull/1291, this needs to land in 4.0.

runcom:
e2es are failing with starting kube though, could you take a look?

/retest

hexfusion:
working through this now

hexfusion:
I am building 

```
/var/opt/openshift/tls/etcd-metrics-ca-bundle.crt
/var/opt/openshift/tls/etcd-metrics-signer-client.crt
/var/opt/openshift/tls/etcd-metrics-signer-client.key
/var/opt/openshift/tls/etcd-metrics-signer-server.crt
/var/opt/openshift/tls/etcd-metrics-signer-server.key
```

But bootstrap wants different so back to the installer.
```
Mar 06 13:32:06 ip-10-0-5-133 bootkube.sh[29781]: F0306 13:32:06.133573       1 bootstrap.go:111] error rendering bootstrap manifests: open /etc/ssl/etcd/metrics-ca.crt: no such file or directory
```

hexfusion:
We have a little bit of a chicken and egg issue here :) as bootstrap depends on itself from the installer.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -118,8 +118,13 @@ type ControllerConfigSpec struct {
...
 
 	// CAs
-	EtcdCAData []byte `json:"etcdCAData"`
-	RootCAData []byte `json:"rootCAData"`
+	EtcdCAData        []byte `json:"etcdCAData"`
+	EtcdMetricsCAData []byte `json:"etcdMetricsCAData"`
+	RootCAData        []byte `json:"rootCAData"`
+
+	// Metrics
+	EtcdMetricsServerCertData []byte `json:"etcdMetricsServerCertData"`
```
abhinavdahiya: I would like to use objectrefs for secret object.

cmd/machine-config-operator/bootstrap.go
```diff
@@ -24,28 +24,34 @@ var (
...
+		destinationDir            string
 	}
 )
 
 func init() {
 	rootCmd.AddCommand(bootstrapCmd)
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdCAFile, "etcd-ca", "/etc/ssl/etcd/ca.crt", "path to etcd CA certificate")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdMetricsCAFile, "etcd-metrics-ca", "", "path to etcd metrics CA certificate")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdMetricsServerCertFile, "etcd-metrics-server-crt", "", "path to etcd metrics server certificate")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdMetricsServerKeyFile, "etcd-metrics-server-key", "", "path to etcd metrics server key")
```
runcom: AIUI, these flags should provide a dummy/default value for this PR to _not_ depend on the installer. After this goes in, you can land the installer PR and then create another PR here to remove the dummy/default value here to always honor what bootkube gives you.

(this is not changing or adding images from the release payload so it should be an easier dance)

/cc @cgwalters

hexfusion: ok sounds good that was my thought as well.

brancz:
I can see the metrics client cert in the e2e test artifacts :tada: now just need to satisfy CI I guess :)

hexfusion:
@brancz we are close just need to align the MCO stars :).

hexfusion:
> Mar 08 13:09:24 ip-10-0-8-16 bootkube.sh[4572]: "00_openshift-kube-controller-manager-ns.yaml": failed to create: request declared a Content-Length of 135 but only wrote 0 bytes

not sure if this is related digging.....

pkg/operator/operator.go
```diff
@@ -279,11 +279,25 @@ func (optr *Operator) sync(key string) error {
 		return err
 	}
 
+	// sync up keys and certs
+	etcdMetricsServerCert, err := optr.getKeyFromSecret("kube-system", "etcd-metrics-server", "tls.crt")
```
runcom: this is now failing here cause the secret isn't in the cluster

From CI:

> level=fatal msg="failed to initialize the cluster: Cluster operator machine-config is reporting a failure: Failed to resync 0.0.1-2019-03-08-124327 because: secrets \"etcd-metrics-server\" not found"

is the installer, today, creating those secrets?

cmd/machine-config-operator/bootstrap.go
```diff
@@ -24,28 +24,34 @@ var (
...
+		etcdMetricsCAFile         string
+		etcdMetricsServerCertFile string
+		etcdMetricsServerKeyFile  string
 	}
 )
 
 func init() {
 	rootCmd.AddCommand(bootstrapCmd)
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdCAFile, "etcd-ca", "/etc/ssl/etcd/ca.crt", "path to etcd CA certificate")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdMetricsCAFile, "etcd-metrics-ca", "/assets/tls/etcd-metrics-ca-bundle.crt", "path to etcd metrics CA certificate")
```
runcom: are these files actually there on bootstrap? meaning, the installer is already laying down these files right?

hexfusion: No but I am injecting this with my own installer and using this release. https://github.com/openshift/installer/pull/1389

hexfusion: It still fails

hexfusion:
```
OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=registry.svc.ci.openshift.org/ci-op-wz3pbz49/release:latest openshift-install-dev --log-level debug --dir aws-cluster/cluster-dev create cluster
```

pkg/operator/operator.go
hexfusion: see below

cmd/machine-config-operator/bootstrap.go
hexfusion: sorry was responding to above yes those files exist ie  `/assets/tls/etcd-metrics-ca-bundle.crt`

pkg/operator/operator.go
runcom: ok, given these secrets aren't in the cluster, that installer PR where you add these secrets must land first (right?) or this cannot pass

hexfusion:
`clusterversion.json` from artifacts where is this on the server? I see /etc/mcs/bootstrap/machine-configs/master-b177c4ce6a5887b4db529acd28a54f25.yaml but what is it comparing to exactly

pkg/operator/operator.go
hexfusion: right but if I add that to code to my custom installer and use this release from CI as below then it should work but I still get 

```

hexfusion:
Also seeing `DEBUG modified etcd-member-ip-10-0-136-122.ec2.internal.158a0289646d9819: Back-off restarting failed container ` digging into that

hexfusion:
ci/prow/e2e-aws

hexfusion:
I understand the tests will fail just trying to get a better look at this from the inside.

brancz:
where are we with this?

hexfusion:
This is being reworked as a feature set to kubecsr. Allowing for metrics certs to be generated in the same fashion as etcd server and peer csr requests during init.

hexfusion:
@abhinavdahiya  PTAL

/cc @brancz

kikisdeliveryservice:
@hexfusion does this have/need an exception granted?

cc: @runcom

hexfusion:
> @hexfusion does this have/need an exception granted?

Yeah @kikisdeliveryservice monitoring will chime in, probably tomorrow.

s-urbaniak:
/approve
... if the invariants expressed in https://github.com/openshift/installer/pull/1483#pullrequestreview-220552841 still hold.

Note that I do not have the authority to grant exceptions though! This must be done higher level.

runcom:
cc @smarterclayton

s-urbaniak:
xref https://github.com/openshift/cluster-monitoring-operator/pull/302

hexfusion:
We have a confirmation on this exception from @derekwaynecarr.

/cc @runcom

runcom:
just saw that, thanks

/lgtm

s-urbaniak:
@runcom @hexfusion awesome, thanks for the hard work ... now let's see if etcd appears in monitoring! :tada:

PR #518, opened 2019-03-01
bootstrap: Final switch to CVO pod image

cgwalters:
When https://github.com/openshift/installer/pull/1292 lands
this will undo the hacks to enable the CVO-driven pod image.


kikisdeliveryservice:
This merges after the installer PR merges.

cgwalters:
/hold cancel

Installer PR went in, this should just need a `lgtm`.

Issue #519, opened 2019-03-01
investigate high CPU usage


miabbott:
@smarterclayton says:

> MCO is the third highest CPU namespace right now - using about a tenth of a core on an unloaded cluster on average.  that’s ok, but should be lower (that’s higher than SDN).  can someone take the todo to do a quick look at where the CPU is coming from and what is the cause (writes, busy loops, checking too much, etc)?

kikisdeliveryservice:
Some more information:
in general most healthy operators at steady state on unloaded clusters should be 10millicore, and operators that have per node agents can budget 5-10millicore per node

kikisdeliveryservice:
```
Element    Value
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-server",namespace="openshift-machine-config-operator",pod_name="machine-config-server-cmgjq"}    0.02591943592962962
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-server",namespace="openshift-machine-config-operator",pod_name="machine-config-server-b8tf7"}    0.024595734537037065
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-server",namespace="openshift-machine-config-operator",pod_name="machine-config-server-gmd8d"}    0.02412587116296297
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-controller",namespace="openshift-machine-config-operator",pod_name="machine-config-controller-7d5744464-zws45"}    0.007748433951851876
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-operator",namespace="openshift-machine-config-operator",pod_name="machine-config-operator-5bf44bf567-rt2hc"}    0.006262270633333337
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-daemon",namespace="openshift-machine-config-operator",pod_name="machine-config-daemon-gmtld"}    0.0023182287481481514
namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{container_name="machine-config-daemon",namespace="openshift-machine-config-operator",pod_name="machine-config-daemon-5gl8s"}    0.0021795583518518552
[...]

PR #521, opened 2019-03-02
pkg/server: encode json straight to response

runcom:

**- What I did**

Avoid marshaling into memory as well and remove not needed
content-length sets.

Should help with #519 

**- How to verify it**

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -79,13 +79,11 @@ func NewServerAPIHandler(s Server) *APIHandler {
 // API handler.
 func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet && r.Method != http.MethodHead {
-		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusMethodNotAllowed)
 		return
 	}
 
 	if r.URL.Path == "" {
-		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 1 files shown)


cgwalters:
Is this for 4.0?  Can you add the label if so?

LGTM though I find myself wondering if the CPU usage is related to the ELB health checks (`/healthz`).  Looks like the interval there is every 10 seconds.

Machines only pull from this endpoint when they're booting.

Going to
/assign LorbusChris
for another review.

pkg/server/api.go
```diff
@@ -79,13 +79,11 @@ func NewServerAPIHandler(s Server) *APIHandler {
 // API handler.
 func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet && r.Method != http.MethodHead {
-		w.Header().Set("Content-Length", "0")
```
LorbusChris: Tests expect content length to be 0 in error cases. If we don't encode content length at all, tests need to be adapted.

pkg/server/api.go
```diff
@@ -107,24 +104,15 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	data, err := json.Marshal(conf)
-	if err != nil {
-		w.Header().Set("Content-Length", "0")
-		w.WriteHeader(http.StatusInternalServerError)
-		glog.Errorf("failed to marshal %v config: %v", cr, err)
-		return
-	}
-
-	w.Header().Set("Content-Length", fmt.Sprintf("%d", len(data)))
```
LorbusChris: why was this done before?

runcom: I don't know, but if you write something to the response writer which has a length, that header is automatically added with the correct length

pkg/server/api.go
runcom: Yup, on it

runcom:
closing this for now as it's not properly working as I've expected

runcom:
> though I find myself wondering if the CPU usage is related to the ELB health checks (`/healthz`).

That's served by the MCD though and MCD isn't showing high cpu usage

PR #523, opened 2019-03-04
kubelet: use cache configMap and secrets change strategy

rphillips:
**Bug fix for 4.0**

Watched based strategy for ConfigMaps and Secrets has a couple bugs, 1) golang http2 max
streams blocking when the stream limit is reached and 2) the kubelet not
cleaning up watches for terminated pods.

This patch configures the cache based strategy. Once golang 1.12 is in
use, and the kubelet patch is merged we can use the watch based
strategy.

ref: https://github.com/kubernetes/kubernetes/issues/74412
ref: https://github.com/kubernetes/kubernetes/issues/74412#issuecomment-468437599

**- How to verify it**

```Use Cache based watched strategy in Kubelet for ConfigMaps and Secrets```

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -63,6 +63,9 @@ var blacklistKubeletConfigurationFields = []string{
 	"CgroupDriver",
 	"ClusterDNS",
 	"ClusterDomain",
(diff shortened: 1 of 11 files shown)


rphillips:
We are going to patch the kubelet with a 1.12 cherry-pick. A separate PR will blacklist the option witin the kubeconfig.

PR #525, opened 2019-03-05
templates/master: define resource request on etcd container not init

hexfusion:
Upon review, the original resource definition was set against the init container not etcd itself. This PR moves that to the etcd container resulting in a QoS a class of ` Guaranteed` vs `BestEffort`. 

/cc @smarterclayton


smarterclayton:
`Burstable`, not `Guaranteed`

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -59,10 +59,6 @@ contents:
           
         securityContext:
           priviledged: true
```
smarterclayton: This looks like a typo....

smarterclayton: This is a typo.  Can you open a second PR to remove this (since it's not needed apparently?)

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -98,6 +94,10 @@ contents:
             --listen-peer-urls=https://0.0.0.0:2380 \
         securityContext:
           priviledged: true
```
smarterclayton: Also a typo - this won't work....

Actually, how the heck is this working.

smarterclayton:
Changes lgtm, want a second pair of eyes.

Also, the privileged flag isn't set correctly which means etcd works without privileged somehow.  Can we chase that down?  @abhinavdahiya

hexfusion:
@smarterclayton I updated the typo after I saw the first review should I revert or leave this.

smarterclayton:
Let's deal with it in another PR in case it has more implications - this stands on its own

smarterclayton:
@sjenning review please

cgwalters:
/approve

LGTM.

kikisdeliveryservice:
ping @sjenning

sjenning:
sorry for the delay
/lgtm

derekwaynecarr:
upstream was still broken.

kubelet should have still looked at init containers in calculating qos.
https://bugzilla.redhat.com/show_bug.cgi?id=1687004
https://github.com/kubernetes/kubernetes/issues/75212

PR #526, opened 2019-03-05
templates/master: fix typo in etcd-member yaml

hexfusion:
must merge after https://github.com/openshift/machine-config-operator/pull/525

`s/priviledged/privileged`


kikisdeliveryservice:
Just adding a hold to wait for other PR:

/hold

kikisdeliveryservice:
Also, we are going to wait on confirmation that fixing this typo doesn't mess anything else up as per @smarterclayton 's comments: https://github.com/openshift/machine-config-operator/pull/525#issuecomment-469793136

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -20,7 +20,7 @@ contents:
         - "--output-file=/run/etcd/environment"
         - "--v=4"
         securityContext:
-          priviledged: true
+          privileged: true
```
smarterclayton: Since we were working fine before this change, and so we didn't need security, I need to know why we don't just remove this section since we don't need privileged to do actions in this directory (or, if we do, how we were able to work).

smarterclayton:
@sjenning why didn't we need privileged to write to a host path?

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
cgwalters: Isn't this because etcd is running as a static pod, it has the privileges to use `hostPath`?

sjenning:
> @sjenning why didn't we need privileged to write to a host path?

good question :confused:

runcom:
ping @sjenning @hexfusion @smarterclayton what do we want to do with this?

smarterclayton:
We want to fix this and we want to find out why we didn't need privileged
to write to a host path.

sjenning:
@smarterclayton we opened this and @joelsmith investigated
https://bugzilla.redhat.com/show_bug.cgi?id=1688516

Joel closed it with reason.  Please reopen if you do not agree.

joelsmith:
Sorry I didn't circle back to close the loop here.

If anything, I think we ought to remove `privileged` from the pod spec entirely. If the pod runs without it, we ought to run it that way since it will have a smaller capabilities bounding set and run with a more restrictive SELinux profile. If we only wanted to run it privileged because we thought that was necessary for `hostPath` volumes (which it isn't), we'd make it a little harder to escape the container.  Of course, if something has owned an etcd pod, they pretty much own the entire cluster already.

kikisdeliveryservice:
Based on the newest comments, once @hexfusion rebases this, we will be merging.

joelsmith:
> Based on the newest comments, once @hexfusion rebases this, we will be merging.

Sorry if I didn't explain well. My recommendation is not to merge this. If it's working today, my recommendation is to remove the `securityContext` entirely. It's more secure without `privileged` and if we don't need it, we shouldn't run the pod in a less secure manner.

kikisdeliveryservice:
thanks for the clarification.

/hold

hexfusion:
I will rebase this now.

hexfusion:
updated per @joelsmith recommendations.

joelsmith:
Thanks! Dunno if this will have any effect coming from me but...
/hold cancel
/lgtm

hexfusion:
We need a lgtm from owner to merge @kikisdeliveryservice ?

PR #527, opened 2019-03-06
pkg/operator: set operator version to release payload version

runcom:
The other components versions can be left as they are to use the image
pullspec. This is not changing the version we report when starting every
component though for our debugging sanity.

This is a 4.0 requirement to pass https://github.com/openshift/origin/pull/22156

### DIFF
--- a/install/0000_30_machine-config-operator_04_deployment.yaml
+++ b/install/0000_30_machine-config-operator_04_deployment.yaml
@@ -25,6 +25,9 @@ spec:
           requests:
             cpu: 20m
             memory: 50Mi
+        env:
+          - name: RELEASE_VERSION
+            value: "0.0.1-snapshot"
         volumeMounts:
         - name: root-ca
           mountPath: /etc/ssl/kubernetes/ca.crt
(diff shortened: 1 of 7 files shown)


runcom:
/hold

I'm running tests right now and want to test an upgrade as well, I'll unhold once I'm done with that

runcom:
Upgrade path works as expected as well:
```
...
machine-config                        0.0.1-2019-03-06-140934                              True        False         False     27m
...
```

install/0000_30_machine-config-operator_04_deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
           requests:
             cpu: 20m
             memory: 50Mi
+        env:
+          - name: RELEASE_VERSION
```
cgwalters: And this value gets substituted by `oc adm release new` or so?  `$ git grep RELEASE_VERSION` doesn't give me any hits in origin.

cgwalters: Ah, it's substituting the `0.0.1` bit: https://github.com/openshift/origin/commit/6e8a42aef2612db1c6acf494e77a6c96bf2b610d

cgwalters:
/hold cancel
(right?)
/approve

PR #496, opened 2019-02-25
WIP: Add hack/cluster-cvo-push.sh

cgwalters:
We need a development workflow that keeps the CVO active so we
can test upgrades.

PR #531, opened 2019-03-07
controller: Report number of degraded machines in pool

cgwalters:
We want to get rid of "degraded" someday; but today is not that day.
In the meantime, let's make clear that there's a degraded node
in a pool status, as distinct from "unavailable".

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -324,6 +324,10 @@ type MachineConfigPoolStatus struct {
 	// A node is marked unavailable if it is in updating state or NodeReady condition is false.
 	UnavailableMachineCount int32 `json:"unavailableMachineCount"`
 
+	// Total number of machines marked degraded.
+	// A node is marked degraded if applying a configuration failed..
+	DegradedMachineCount int32 `json:"degradedMachineCount"`
+
 	// Represents the latest available observations of current state.
 	Conditions []MachineConfigPoolCondition `json:"conditions"`
 }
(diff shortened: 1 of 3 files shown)


cgwalters:
(Only compile tested as of yet)

runcom:
/hold

till testing

added for 4.0 as would be desiderable to have it and it's pretty straightforward

kikisdeliveryservice:
this makes a lot of sense to me!

cgwalters:
Hit on the fact that `make deploy-controller` no longer works:

```
    Message:               Failed to resync 4.0.0-alpha.0-9-gfc1a99e6-dirty because: timed out waiting for the condition during syncRequiredMachineConfigPools: pool master has not progressed to latest configuration: controller version mismatch for master-a530215ab26a87a39158b9624966c8db expected 4.0.0-alpha.0-9-gfc1a99e6-dirty has 4.0.0-alpha.0-12-g0aaf148e                                                              
    Reason:                timed out waiting for the condition during syncRequiredMachineConfigPools: pool master has not progressed to latest configuration: controller version mismatch for master-a530215ab26a87a39158b9624966c8db expected 4.0.0-alpha.0-9-gfc1a99e6-dirty has 4.0.0-alpha.0-12-g0aaf148e
    Status:                True
    Type:                  Failing
```

:cry:

runcom:
> Hit on the fact that `make deploy-controller` no longer works:

is this a regression? 🤔

kikisdeliveryservice:
@runcom Perhaps it's the same issue as #472 ?

cgwalters:
> is this a regression?

Since we started checking versions of sub-components, yes.  There are two ways I see, first we can go with the [CVO push](https://github.com/openshift/machine-config-operator/pull/496) route (which we need anyways).  But I'd also like a "quickly replace this image without going through CVO" path.  Maybe we add an environment variable to tell the MCO to ignore versions of its components?

runcom:
> Since we started checking versions of sub-components, yes.

That has been in the MCO since Abhinav wrote that, something else drifted?

As for temporarily disable checks with an env, yes let's do that, though it requires an invasive plumbing just for that no?

PR #532, opened 2019-03-07
rebase master-4.1

runcom:
I don't have power to rebase and push to origin so we rebase with PRs from time to time now ~~I guess this would be the way to rebase that branch?~~


runcom:
Clayton confirmed

kikisdeliveryservice:
this makes sense, let's give it a whirl

PR #535, opened 2019-03-08
Bug: 1686556: Symlink /root/.docker/config.json to kubelet auth

cgwalters:
This is a workaround for us switching to an authenticated pause
image and crio not knowing how to use the kubelet's auth for this.

https://bugzilla.redhat.com/show_bug.cgi?id=1686556

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/lib/resourcemerge"
 	"github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/version"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


cgwalters:
(And we need to hack this because the template controller doesn't know about symlinks, just files)

cgwalters:
Adding WIP since I only compile tested this.

wking:
> Adding WIP since I only compile tested this.

Is there a way for hard-code a private pause image while CI-testing this?

cgwalters:
/test e2e-aws

(So I can grab the release image this time)

cgwalters:
> Is there a way for hard-code a private pause image while CI-testing this?

Gave that a shot in

https://github.com/openshift/machine-config-operator/pull/537#issue-259576831

cgwalters:
Testing this out live reveals I *almost* got it right:

```
$ oc debug node/ip-10-0-129-163.ec2.internal
Starting pod/ip-10-0-129-163ec2internal-debug ...
To directly access the host PATH, try `chroot /host /bin/bash`
sh-4.2# chroot /host
sh-4.2# ls -al /root/
total 24
drwx------.  3 root root  123 Mar  8 19:17 .
drwxr-xr-x. 23 root root 4096 Mar  8 19:17 ..
-rw-r--r--.  1 root root   18 Mar  8 19:17 .bash_logout
-rw-r--r--.  1 root root  193 Mar  8 19:17 .bash_profile
-rw-r--r--.  1 root root  231 Mar  8 19:17 .bashrc
drwxr-xr-x.  2 root root   25 Mar  8 19:17 .docker
-rw-------.  1 root root 2667 Mar  6 22:23 anaconda-ks.cfg
-rw-------.  1 root root 2334 Mar  6 22:23 original-ks.cfg
sh-4.2# ls -al /root/.docker/
total 0
drwxr-xr-x. 2 root root  25 Mar  8 19:17 .
drwx------. 3 root root 123 Mar  8 19:17 ..
lrwxrwxrwx. 1 root root  26 Mar  8 19:17 config.json -> /var/lib/kubelet/auth.json
sh-4.2# 
```

cgwalters:
And now with `s/auth.json/config.json/`

cgwalters:
https://github.com/openshift/machine-config-operator/pull/537#issuecomment-471089342

/hold cancel

cgwalters:
Let's ship this!

cgwalters:
Although wait a sec...the link seems to be inside the SSH MC.

/hold

cgwalters:
OK, now this should be really ready to go.

Issue #501, opened 2019-02-26
errors strings must start with lowercase


runcom:
As a general rule as the guideline says: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Let's keep this open till most of the code is polished and we'll be more strict when reviewing PRs.
Keep it open as a tracker for PRs that are cleaning the current code.

Ref: https://github.com/openshift/machine-config-operator/pull/481#discussion_r260157691

kikisdeliveryservice:
I will make the changes in update.go!

kikisdeliveryservice:
Opened PR #539 that updates the mcd gofiles: daemon.go, node.go, osrelease.go, rpm-ostree.go, update.go, writer.go

kikisdeliveryservice:
Closing as the rest of the errors were updated in #2868

PR #539, opened 2019-03-11
mcd: fix error string capitalization

kikisdeliveryservice:
Set first char to lowercase for all error strings to conform to golang conventions.

Files changed: daemon.go, node.go, osrelease.go,
rpm-ostree.go, update.go, writer.go

Related-to: #501


kikisdeliveryservice:
Started with update.go (which had a bunch of changes) but realized the other files in `/daemon` only had a few, so figured might as well get those in too.

PR #537, opened 2019-03-08
WIP: Symlink authfile with private pause

cgwalters:
Additional hack to test https://github.com/openshift/machine-config-operator/pull/535

PR #540, opened 2019-03-11
Use CRI-O’s pause_image_auth_file option

mtrmac:
**- What I did**

Replaced the symlink added in https://github.com/openshift/machine-config-operator/pull/535 with using a CRI-O config option introduced by 
https://github.com/kubernetes-sigs/cri-o/pull/2115 .

Note that this requires CRI-O ≥ 1.12.9 / ≥ 1.13.2 .

**- How to verify it**

I hope the CI will do that?  I honestly have no idea what I am doing.

All I verified was that the unit tests work (with updates to the expected results).


kikisdeliveryservice:
cc: @cgwalters - Thoughts?

runcom:
Thanks Miloslav for putting this up so quickly. Context here is that the symlink option is a stopgap but CRI-O added an option to directly speficy the Auth file to look for when pulling the pause image.

cgwalters:
What I did to test the other PR was to [create a separate PR with a test commit](https://github.com/openshift/machine-config-operator/pull/537).   Cherry picking [this commit](https://github.com/openshift/machine-config-operator/pull/537/commits/bbf81398bead000910b35300905bda5caffef720) onto another PR should be enough to get a good test.

cgwalters:
/approve

OK based on the results in #542 this
/lgtm

PR #544, opened 2019-03-12
etcd: add fallback to logs on error

mfojtik:
**- What I did**

Set the `terminationMessagePolicy` so when the etcd container crash, it will record the last N lines of the container log into pod status. That means we can see the reason it crashed instead of just observe the crashloop.

This will help debugging etcd crashes.

**- How to verify it**

Wait for etcd to crash, then oc get pods/etcd-member -o yaml should show the last few lines from the etcd container.


smarterclayton:
/lgtm cancel

You need to update the unit tests.

mfojtik:
@smarterclayton updated...

cgwalters:
(For future reference I personally would prefer that we squash commits like that)

Issue #543, opened 2019-03-12
Fix dependency on chroot


cgwalters:
Today the way the MCD `chroot` into the host creates all sort of problems.  I believe this was done for a few reasons:

 - We can speak DBus to e.g. systemd/rpm-ostree
 - When processing Ignition we don't have to deal with the rootfs prefix

And over time we've had to add hacks like https://github.com/openshift/machine-config-operator/pull/370 to pick up things from the Kube side.

Dealing with DBus is easy, we can just bind mount `/run/dbus/system-bus-socket` into our code.

The filesystem bits are hard mainly because of symlinks.  See e.g. [AT_NO_JUMPS patches](https://marc.info/?l=linux-api&m=155215087916280&w=2) and [rpm-ostree os-release change](https://github.com/projectatomic/rpm-ostree/commit/e94f8c9b5f8e327c792156d6f154aedcba96ef16).

I think for the filesystem stuff...basically we need to split the MCD into two containers (in the same pod); the "kube mcd" and the "host mgmt MCD".  Only the latter "hostmgmt" MCD would be chrooted.  

Or...actually perhaps even easier, have a single MCD that execs the child process chrooted.  The child process gets passed the Ignition configs as temporary files.

In fact it's *almost* like we exec in once-from mode.

cgwalters:
One thing that might actually help us a lot architecturally too is to also have the "once-from" code (basically Ignition handling) also be built into an RPM that goes into the host.  And then when we want to run it rather than using `chroot` we actually `systemd-run` it just like we do for rpm-ostree.

ashcrow:
:+1: to https://github.com/projectatomic/rpm-ostree/commit/e94f8c9b5f8e327c792156d6f154aedcba96ef16 ... though we will need to build that package and get it into composes for it to be used.

> Or...actually perhaps even easier, have a single MCD that execs the child process chrooted. The child process gets passed the Ignition configs as temporary files.
> In fact it's almost like we exec in once-from mode.

I like this path.

cgwalters:
I just did https://github.com/openshift/machine-config-operator/pull/545
which I think simplifies the code and will help avoid the immediate crasher.

Basically, while I think going down the exec path is a good idea it's fairly risky/involved at the moment.
Dropping our dependency on everything using cgo should get us unblocked.

jlebon:
Previous discussions on this subject in #45. Definitely agree overall we should only be chroot'ing for the parts we really have to.

> I just did #545
[...]

PR #545, opened 2019-03-12
daemon: Drop all dependency on cgo and external shlibs

cgwalters:
The way the MCD today chroots into the host is problematic
for major OS version updates and creates a hidden dependency.

Eventually we need to stop chrooting, see:
https://github.com/openshift/machine-config-operator/issues/543

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -15,7 +16,6 @@ import (
 
 	imgref "github.com/containers/image/docker/reference"
 	"github.com/coreos/go-systemd/login1"
-	"github.com/coreos/go-systemd/sdjournal"
 	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
@@ -422,34 +422,13 @@ const (
 	sdMessageSessionStart = "8d45620c1a4348dbb17410da57c60c66"
 )
 
+// detectEarlySSHAccessesFromBoot taints the node if we find a login before the daemon started up.
 func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
-	journal, err := sdjournal.NewJournal()
+	journalOutput, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart).CombinedOutput()
 	if err != nil {
 		return err
 	}
-	defer journal.Close()
-	if err := journal.AddMatch("MESSAGE_ID=" + sdMessageSessionStart); err != nil {
-		return err
-	}
-	if err := journal.SeekHead(); err != nil {
-		return err
-	}
-	r, err := journal.Next()
-	if err != nil {
-		return err
-	}
-	// journal EOF
-	if r == 0 {
-		return nil
-	}
-	// just one entry is enough to understand if someone jumped on the node
-	// from the very first boot
-	entry, err := journal.GetEntry()
-	if err != nil {
-		return err
-	}
-	// just sanity checking
-	if entry != nil {
+	if len(journalOutput) > 0 {
 		glog.Info("Detected a login session before the daemon took over on first boot")
 		glog.Infof("Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
 		if err := dn.applySSHAccessedAnnotation(); err != nil {
@@ -521,22 +500,47 @@ func (dn *Daemon) BindPodMounts() error {
 }
 
 func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
-	sessionNewCh := dn.loginClient.Subscribe("SessionNew")
-	for {
-		select {
-		case <-stopCh:
-			return
-		case msg, ok := <-sessionNewCh:
-			if !ok {
-				glog.V(4).Info("Not adding the ssh accessed annotation because the logind SessionNew channel is closed")
-				return
-			}
-			glog.Infof("Detected a new login session: %v", msg)
-			glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
-			if err := dn.applySSHAccessedAnnotation(); err != nil {
-				exitCh <- err
-			}
-		}
+	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		exitCh <- err
+		return
+	}
+	if err := cmd.Start(); err != nil {
+		exitCh <- err
+		return
+	}
+	worker := make(chan struct{})
+ 	go func() {
+		for {
+			select {
+			case <-worker:
+ 				return
+			default:
+				buf := make([]byte, 1024)
+				l, err := stdout.Read(buf)
+				if err != nil {
+					if err == io.EOF {
+						return
+					}
+					exitCh <- err
+					return
+				}
+				if l > 0 {
+					line := strings.Split(string(buf), "\n")[0]
+					glog.Infof("Detected a new login session: %s", line)
+					glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
+					if err := dn.applySSHAccessedAnnotation(); err != nil {
+						exitCh <- err
+					}
+				}
+ 			}
+ 		}
+ 	}()
+	select {
+	case <-stopCh:
+		close(worker)
+		cmd.Process.Kill()
 	}
 }
 
@@ -736,10 +740,8 @@ func (dn *Daemon) CheckStateOnBoot() error {
 	if err != nil {
 		return err
 	}
-	if state.bootstrapping {
-		if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
-			return fmt.Errorf("error detecting early SSH accesses: %v", err)
-		}
+	if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
+		return fmt.Errorf("error detecting previous SSH accesses: %v", err)
 	}
 
 	if state.bootstrapping {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -15,7 +15,6 @@ import (
 	"time"
 
 	imgref "github.com/containers/image/docker/reference"
-	"github.com/coreos/go-systemd/login1"
 	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
@@ -58,10 +57,6 @@ type Daemon struct {
 
 	// bootedOSImageURL is the currently booted URL of the operating system
 	bootedOSImageURL string
-
-	// login client talks to the systemd-logind service for rebooting the
-	// machine
-	loginClient *login1.Conn
 
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient kubernetes.Interface
@@ -178,15 +173,11 @@ func New(
 	stopCh <-chan struct{},
 ) (*Daemon, error) {
 
-	loginClient, err := login1.New()
-	if err != nil {
-		return nil, fmt.Errorf("error establishing connection to logind dbus: %v", err)
-	}
-
 	osImageURL := ""
 	osVersion := ""
 	// Only pull the osImageURL from OSTree when we are on RHCOS
 	if operatingSystem == machineConfigDaemonOSRHCOS {
+		var err error
 		osImageURL, osVersion, err = nodeUpdaterClient.GetBootedOSImageURL(rootMount)
 		if err != nil {
 			return nil, fmt.Errorf("error reading osImageURL from rpm-ostree: %v", err)
@@ -203,7 +194,6 @@ func New(
 		name:                   nodeName,
 		OperatingSystem:        operatingSystem,
 		NodeUpdaterClient:      nodeUpdaterClient,
-		loginClient:            loginClient,
 		rootMount:              rootMount,
 		bootID:                 bootID,
 		bootedOSImageURL:       osImageURL,
@@ -1121,7 +1111,6 @@ func checkFileContentsAndMode(filePath string, expectedContent []byte, mode os.F
 
 // Close closes all the connections the node agent has open for it's lifetime
 func (dn *Daemon) Close() {
-	dn.loginClient.Close()
 }
 
 // ValidPath attempts to see if the path provided is indeed an acceptable
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -2,17 +2,16 @@ package daemon
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/coreos/go-systemd/dbus"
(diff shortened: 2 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -270,33 +269,11 @@ const (
 )
 
 func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
-	journal, err := sdjournal.NewJournal()
+	o, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID=" + sdMessageSessionStart).CombinedOutput()
```
kikisdeliveryservice: could we rename this var from `o` to something more descriptive?

cgwalters: Sure, though I think it's better with comments like this to suggest a new name; have one in mind?

runcom: `out`, `journalOutput`?

jlebon:
We also use the sdjournal bindings in `rpm-ostree.go`.

cgwalters:
> We also use the sdjournal bindings in rpm-ostree.go.

Yeah, I was looking at unwinding the dependency on logind for rebooting first, then will look at that one next.

cgwalters:
Hm, and also this needs to cover the `runLoginMonitor()` case.

/hold

pkg/daemon/daemon.go
kikisdeliveryservice: ahh didn't want to be too prescriptive @cgwalters but point taken :)

journalOutput  wfm.

cgwalters:
OK this PR is now updated to remove all dependency on external shared libraries.  Giving it more of a test now.

pkg/daemon/daemon.go
```diff
@@ -13,8 +14,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/coreos/go-systemd/login1"
-	"github.com/coreos/go-systemd/sdjournal"
```
runcom: can we `rm -rf vendor/github.com/coreos/go-systemd` now?

runcom: or run `dep ensure` (but not sure it can cleanup unused vendors)

pkg/daemon/daemon.go
```diff
@@ -361,23 +329,32 @@ func (dn *Daemon) BindPodMounts() error {
...
+			glog.Fatal(err)
+		}
+		buf := make([]byte, 1024)
+		l, err := stdout.Read(buf)
+		if err != nil {
+			if err == io.EOF {
 				return
 			}
-			glog.Infof("Detected a new login session: %v", msg)
+			glog.Fatal(err)
```
runcom: all these `err` should be sent to the `exitCh` for correct reporting and stop. These should follow with a `return` to exit the goroutine and cleanup now that we don't listen on `stopCh` anymore.

On top of that, `stopCh` must still be around to signal this goroutine to exit

pkg/daemon/daemon.go
cgwalters: No, it's still needed by Ignition.

cgwalters:
So I just tested ssh tainting by:

```
oc rsh pods/machine-config-daemon-26w8c
chroot /rootfs
passwd root
  <12345>
ssh 127.0.0.1
  <12345>
ctrl-d
```

and it worked :wink:

pkg/daemon/daemon.go
cgwalters: OK, fixed.

pkg/daemon/daemon.go
```diff
@@ -361,23 +330,39 @@ func (dn *Daemon) BindPodMounts() error {
...
+	}
+	if err := cmd.Start(); err != nil {
+		exitCh <- err
+		return
+	}
+	go func() {
+		<- stopCh
+		cmd.Process.Kill()
+	}()
+	go func() {
```
runcom: This goroutine only runs once and exits, should keep reading from the stream right?

The code below should be inside a for loop, with a select receiving from stopCh and exiting in that case (and killing the process). The other case in the select should be a default case with the actual code to read from the stream.

pkg/daemon/daemon.go
```diff
@@ -269,34 +259,13 @@ const (
...
-	}
-	// journal EOF
-	if r == 0 {
-		return nil
-	}
-	// just one entry is enough to understand if someone jumped on the node
-	// from the very first boot
-	entry, err := journal.GetEntry()
+// detectThisBootSSHAccesses taints the node if we find a login before the daemon started up.
+func (dn *Daemon) detectThisBootSSHAccesses() error {
```
kikisdeliveryservice: could we use the old name `detectEarlySSHAccessesFromBoot` I think it reads clearer (and summarizes the func comment pretty well.

pkg/daemon/daemon.go
```diff
@@ -1025,7 +1008,6 @@ func checkFileContentsAndMode(filePath string, expectedContent []byte, mode os.F
 
 // Close closes all the connections the node agent has open for it's lifetime
 func (dn *Daemon) Close() {
-	dn.loginClient.Close()
```
kikisdeliveryservice: question: it's ok that this func is empty?

runcom: yup, just doing nothing/returning nothing - as for keeping it, I don't have a strong opinion

pkg/daemon/daemon.go
runcom: in code:

```diff
diff --git a/pkg/daemon/daemon.go b/pkg/daemon/daemon.go
index 199a480..058f7b8 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	imgref "github.com/containers/image/docker/reference"
 	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
@@ -25,7 +26,6 @@ import (
 	mcfgclientv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
-	imgref "github.com/containers/image/docker/reference"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -261,7 +261,7 @@ const (
 
 // detectThisBootSSHAccesses taints the node if we find a login before the daemon started up.
 func (dn *Daemon) detectThisBootSSHAccesses() error {
-	journalOutput, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID=" + sdMessageSessionStart).CombinedOutput()
+	journalOutput, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart).CombinedOutput()
 	if err != nil {
 		return err
 	}
@@ -330,7 +330,7 @@ func (dn *Daemon) BindPodMounts() error {
 }
 
 func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
-	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "MESSAGE_ID=" + sdMessageSessionStart)
+	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		exitCh <- err
@@ -340,29 +340,38 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 		exitCh <- err
 		return
 	}
+	worker := make(chan struct{})
 	go func() {
-		<- stopCh
-		cmd.Process.Kill()
-	}()
-	go func() {
-		buf := make([]byte, 1024)
-		l, err := stdout.Read(buf)
-		if err != nil {
-			if err == io.EOF {
+		for {
+			select {
+			case <-worker:
 				return
-			}
-			exitCh <- err
-			return
-		}
-		if l > 0 {
-			line := strings.Split(string(buf), "\n")[0]
-			glog.Infof("Detected a new login session: %s", line)
-			glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
-			if err := dn.applySSHAccessedAnnotation(); err != nil {
-				exitCh <- err
+			default:
+				buf := make([]byte, 1024)
+				l, err := stdout.Read(buf)
+				if err != nil {
+					if err == io.EOF {
+						return
+					}
+					exitCh <- err
+					return
+				}
+				if l > 0 {
+					line := strings.Split(string(buf), "\n")[0]
+					glog.Infof("Detected a new login session: %s", line)
+					glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
+					if err := dn.applySSHAccessedAnnotation(); err != nil {
+						exitCh <- err
+					}
+				}
 			}
 		}
 	}()
+	select {
+	case <-stopCh:
+		close(worker)
+		cmd.Process.Kill()
+	}
 }
 
 func (dn *Daemon) applySSHAccessedAnnotation() error {
@@ -384,6 +393,7 @@ func (dn *Daemon) runKubeletHealthzMonitor(stopCh <-chan struct{}, exitCh chan<-
 				failureCount++
 				if failureCount >= kubeletHealthzFailureThreshold {
 					exitCh <- fmt.Errorf("kubelet health failure threshold reached")
+					return
 				}
 			} else {
 				failureCount = 0 // reset failure count on success
```

cgwalters: Oh I had completely missed that this function was already in a goroutine.

runcom: also, sorry, the `return` after `exitCh <- err` in the read buffer error isn't needed, we still want to stay in the goroutine and keep retrying

cgwalters: So I'm getting a conflict trying to apply your patch...can you verify you're rebased on master?  Or since you're rewriting it entirely, just paste the new function?

cgwalters: Actually I have unpushed changes but anyways just toss the new function my way literally instead of a diff and I think that'd be easier.

runcom: mmm I've used this PR to create the patch above :confused:

cgwalters:
OK so I tested this PR ensures the MCD works fine in a rhel8 host scenario when rolling a config change.  Going to test OS updates next.

cgwalters:
I rebased this on top of https://github.com/openshift/machine-config-operator/pull/548

runcom:
This can be now rebased on top of my merged PR. I'll submit you a patch with what we discussed in https://github.com/openshift/machine-config-operator/pull/545#discussion_r265499406

pkg/daemon/rpm-ostree.go
```diff
@@ -123,93 +122,27 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
...
 	}
-	defer reader.Close()
 
-	// We're kinda abusing the API here. The idea is that the stop channel is
-	// used with time.After(), hence the `chan time.Time` type. But really, it
-	// works to just never output anything and just close it as we do here.
-	if err := reader.Follow(stopCh, os.Stdout); err != sdjournal.ErrExpired {
-		glog.Warningf("Failed to follow journal: %v", err)
-	}
+	go func() {
```
runcom: you don't really need this goroutine, this whole function is called from a goroutine itself, you can just wait on the channel and kill the process once it finishes I guess (basically remove the wrapping `go func() ...`)

pkg/daemon/daemon.go
```diff
@@ -521,22 +500,47 @@ func (dn *Daemon) BindPodMounts() error {
...
+	if err != nil {
+		exitCh <- err
+		return
+	}
+	if err := cmd.Start(); err != nil {
+		exitCh <- err
+		return
+	}
+	worker := make(chan struct{})
+ 	go func() {
```
runcom: nevermind, I realized this goroutine and the extra channel isn't needed, just this is ok:

```diff
diff --git a/pkg/daemon/daemon.go b/pkg/daemon/daemon.go
index 279891f..e818bd0 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -414,7 +414,7 @@ const (
 
 // detectEarlySSHAccessesFromBoot taints the node if we find a login before the daemon started up.
 func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
-	journalOutput, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID=" + sdMessageSessionStart).CombinedOutput()
+	journalOutput, err := exec.Command("journalctl", "-b", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart).CombinedOutput()
 	if err != nil {
 		return err
 	}
@@ -490,7 +490,7 @@ func (dn *Daemon) BindPodMounts() error {
 }
 
 func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
-	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "MESSAGE_ID=" + sdMessageSessionStart)
+	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "MESSAGE_ID="+sdMessageSessionStart)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		exitCh <- err
@@ -500,37 +500,30 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 		exitCh <- err
 		return
 	}
-	worker := make(chan struct{})
- 	go func() {
-		for {
-			select {
-			case <-worker:
- 				return
-			default:
-				buf := make([]byte, 1024)
-				l, err := stdout.Read(buf)
-				if err != nil {
-					if err == io.EOF {
-						return
-					}
-					exitCh <- err
+	for {
+		select {
+		case <-stopCh:
+			cmd.Process.Kill()
+			return
+		default:
+			buf := make([]byte, 1024)
+			l, err := stdout.Read(buf)
+			if err != nil {
+				if err == io.EOF {
 					return
 				}
-				if l > 0 {
-					line := strings.Split(string(buf), "\n")[0]
-					glog.Infof("Detected a new login session: %s", line)
-					glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
-					if err := dn.applySSHAccessedAnnotation(); err != nil {
-						exitCh <- err
-					}
+				exitCh <- err
+				return
+			}
+			if l > 0 {
+				line := strings.Split(string(buf), "\n")[0]
+				glog.Infof("Detected a new login session: %s", line)
+				glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
+				if err := dn.applySSHAccessedAnnotation(); err != nil {
+					exitCh <- err
 				}
- 			}
- 		}
- 	}()
-	select {
-	case <-stopCh:
-		close(worker)
-		cmd.Process.Kill()
+			}
+		}
 	}
 }
 ```

cgwalters: OK...I will do this but in the end the ssh tainting stuff IMO can regress a little bit...there's a massive pile of stuff backlogged behind the rhel8 rebase and I don't want to have the journal tainting code being *perfect* blocking that.

runcom: sure thing, I concur with that, I just want to leave traces that we need to fix this. Take my comments as nits when appropriate

cgwalters: But briefly, I don't understand; we'll block the goroutine in the `stdout.Read` invocation and hence never check `stopCh` right?

cgwalters:
Rebased :surfing_man: but didn't change the journal bits pending resolution of discussion.

kikisdeliveryservice:
aws not happy these past 2 days: `aws_route.to_nat_gw.4: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"` 

/test e2e-aws-op

runcom:
/approve
/lgtm

let's follow up on the ssh stuff, this is on the critical path now so let's get this in and follow up

kikisdeliveryservice:
EBS and `level=fatal msg="failed to initialize the cluster: Could not update servicemonitor \"openshift-apiserver-operator/openshift-apiserver-operator\" (299 of 303): the server does not recognize this resource, check extension API servers"`
/retest

PR #551, opened 2019-03-14
test/e2e: roll out new MCs more than just once

runcom:

**- What I did**

Modified the e2e MC test to roll out MCs more than just once to understand if rolling MCs just goes ok

**- How to verify it**

run e2e or wait CI reports

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -55,7 +55,7 @@ func mcLabelForWorkers() map[string]string {
 	return mcLabels
 }
 
-func createMCFile(path, content string, mode int) ignv2_2types.File {
+func createIgnFile(path, content string, mode int) ignv2_2types.File {
 	return ignv2_2types.File{
 		FileEmbedded1: ignv2_2types.FileEmbedded1{
 			Contents: ignv2_2types.FileContents{
@@ -70,19 +70,13 @@ func createMCFile(path, content string, mode int) ignv2_2types.File {
 	}
 }
 
-func TestMCDeployed(t *testing.T) {
-	cb, err := common.NewClientBuilder("")
-	if err != nil {
-		t.Errorf("%#v", err)
-	}
-	mcClient := cb.MachineConfigClientOrDie("mc-file-add")
-	k := cb.KubeClientOrDie("mc-file-add")
-
+func createMCToAddFile(name, filename, data string) *mcv1.MachineConfig {
 	// create a dummy MC
-	mcName := fmt.Sprintf("00-0add-a-file-%s", uuid.NewUUID())
+	mcName := fmt.Sprintf("%s-%s", name, uuid.NewUUID())
 	mcadd := &mcv1.MachineConfig{}
 	mcadd.ObjectMeta = metav1.ObjectMeta{
-		Name:   mcName,
+		Name: mcName,
+		// TODO(runcom): hardcoded to workers for safety
 		Labels: mcLabelForWorkers(),
 	}
 	mcadd.Spec = mcv1.MachineConfigSpec{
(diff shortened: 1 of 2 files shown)


runcom:
@derekwaynecarr ptal

test/e2e/mcd_test.go
```diff
@@ -70,19 +70,13 @@ func createMCFile(path, content string, mode int) ignv2_2types.File {
...
-	if err != nil {
-		t.Errorf("%#v", err)
-	}
-	mcClient := cb.MachineConfigClientOrDie("mc-file-add")
-	k := cb.KubeClientOrDie("mc-file-add")
-
+func createMCToAddFile(filename, data string) *mcv1.MachineConfig {
 	// create a dummy MC
-	mcName := fmt.Sprintf("00-0add-a-file-%s", uuid.NewUUID())
+	mcName := fmt.Sprintf("%s", uuid.NewUUID())
```
cgwalters: That uuid may sort before the built in MCs right?  How about `mctest-addfile-$(date +'%s')` (in shell terms)?

test/e2e/mcd_test.go
```diff
@@ -92,57 +86,70 @@ func TestMCDeployed(t *testing.T) {
...
+		if err := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+			mcp, err := mcClient.MachineconfigurationV1().MachineConfigPools().Get("worker", metav1.GetOptions{})
+			if err != nil {
+				return false, err
 			}
-			if node.Annotations[constants.CurrentMachineConfigAnnotationKey] == newMCName {
-				visited[node.Name] = true
-				if len(visited) == len(nodes) {
+			for _, mc := range mcp.Status.Configuration.Source {
+				if mc.Name == mcadd.Name {
```
cgwalters: I don't understand this loop.  It's going to be racy; the renderer could have just seen the first config and generated a new rendered version and be working on the full one when we drop in here.

I think the check needs to be that `set(sources) = set(mcadd)`.

runcom: we won't create the second config unless the previous one is applied to all nodes, am I missing something?

test/e2e/mcd_test.go
runcom: fixed thanks!

test/e2e/mcd_test.go
runcom: this check makes sure that the created MC is in the rendered, generated one that we're now going to check (and everything is serial as well)

test/e2e/mcd_test.go
```diff
@@ -92,57 +86,70 @@ func TestMCDeployed(t *testing.T) {
...
-				newMCName = mcp.Status.Configuration.Name
-				return true, nil
-			}
-		}
-		return false, nil
-	}); err != nil {
-		t.Errorf("machine config hasn't been picked by the pool: %v", err)
-	}
+	for i := 0; i < 10; i++ {
+		mcadd := createMCToAddFile("00-0add-a-file", fmt.Sprintf("/etc/mytestconf%d", i), "test")
```
kikisdeliveryservice: doesn't the `00-0add-a-file` name need to change here as well? since it's passed to `createMCToAddFile` and runs into the same problem?

runcom: the issue before was that there wasn't anything other than a random UID, with this now it's that name (which makes it to the top of the list) + a random hash just to differentiate from the others

kikisdeliveryservice: like just `add-a-file` so the name genereated in createMCToAddFile will be `add-a-file-XXXXXXXXX`?

runcom: I can do that sure, not sure if it changes something though :thinking:

kikisdeliveryservice: i thought the worry was that the uuid would sort before the build in MCs. but if we use 00-0-add... won't this sort before the built-in MCs now every time?

runcom: oooo gotcha finally, you're completely right, changing

runcom: fixed

test/e2e/mcd_test.go
cgwalters: Oh right it's inside the loop.  OK.

derekwaynecarr:
yes, this is exactly what i had in mind.  this way, we catch flakes here at a higher rate.

PR #552, opened 2019-03-14
pkg/operator: keep calling sync functions for MCP

### LINKED ISSUE #550: increase frequency of pool -> operator status reporting
https://github.com/openshift/origin/pull/22302#issuecomment-472694227

From Derek:

master-e2e-aws-upgrade/14/build-log.txt

looking at https://github.com/openshift/machine-config-operator/blob/f97747550f42821725f2853b211cc48d4c0df42e/pkg/operator/sync.go#L320

i see on every operator sync, we drop into that function and wait 10m for the sync of all pools to complete.

is it possible to 1) report what machine is unavailable to aid debugging, 2) get feedback on cluster operator object at a frequency better than 10m chunks of time? if we dropped the polling altogether in that function, would we just get updated data on the operator next sync loop iteration? its not clear what benefit the 10m poll provides, but maybe its a limitation of current structure?

runcom:
instead of polling for that long...

**- What I did**

handle point number 2 from https://github.com/openshift/machine-config-operator/issues/550

close #550

**- How to verify it**

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -240,8 +240,6 @@ func (optr *Operator) handleErr(err error, key interface{}) {
 		return
 	}
 
-	optr.syncFailingStatus(err)
-
 	if optr.queue.NumRequeues(key) < maxRetries {
 		glog.V(2).Infof("Error syncing operator %v: %v", key, err)
 		optr.queue.AddRateLimited(key)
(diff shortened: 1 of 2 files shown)


runcom:
@cgwalters @kikisdeliveryservice ptal

aws limit

/retest

runcom:
This PR is included in #554 for ease of testing, I'll leave this open for review but this is going with #554, so

/hold

PR #548, opened 2019-03-13
pkg/daemon: rewrite the MCD as a controller

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

**- What I did**

rewrite the daemon as a controller with a sync handler to deal with transient vs permanent error mainly.

effectively close https://github.com/openshift/machine-config-operator/issues/395
also close https://github.com/openshift/machine-config-operator/issues/401
close https://github.com/openshift/machine-config-operator/issues/299

**- How to verify it**

CI should tell us if this is generally good (hopefully)

**- Description for the changelog**

/cc @derekwaynecarr

PR #554, opened 2019-03-15
Fix MCD status reporting, retry loop and max unavailable progress

runcom:

**- What I did**

This PR is a follow up from #548 which accounts for the change in behavior that we introduced with that. Most notably:

- we are never in a permanent Degraded state anymore, we flip Working and Degraded but the logic for determining the updated/unavailable machines rely only on current/desired annotation. I'm going to fix that and remove Degraded
- progressing on applying a machineconfig is currently broken because of the above, so max unavailable isn't respected
- with #548, we can now delete a bad machineconfig which is erroring out in a loop, and just reapply the old one which is going to just stabilize the machine again
- probably something else I'm still not aware of but I'm actively testing

This PR includes https://github.com/openshift/machine-config-operator/pull/552 as well
so, close #552

**- How to verify it**

I'll add e2e(s) + the CI

### DIFF
--- a/manifests/machineconfigpool.crd.yaml
+++ b/manifests/machineconfigpool.crd.yaml
@@ -16,10 +16,6 @@ spec:
     description: When at least one of machine is not either not updated or is in the process of updating to the desired machine config.
     name: Updating
     type: string
-  - JSONPath: .status.conditions[?(@.type=="Degraded")].status
-    description: When the update for one of the machine is not progressig due to an error.
-    name: Degraded
-    type: string
   # group name to use for REST API: /apis/<group>/<version>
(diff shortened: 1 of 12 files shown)


runcom:
This is good to be reviewed now and there's a self explanatory e2e test for rolling back a bad machineconfig

runcom:
not related and known failure

/retest

runcom:
openshift-sdn failing

/retest

runcom:
green :)

runcom:
@cgwalters @kikisdeliveryservice @LorbusChris ptal

cgwalters:
Nice!

/lgtm

PR #556, opened 2019-03-15
ctrruntimeconfig: Resync on failure and accuratey detect object updates

umohnani8:
When a CR was being created and the label didn't exist in any of the pools
the controller would fail and exit it out. It should try resyncing so that
when the label is added later on, it can finally sync properly and make
the necessary changes to the config fies.
Also adds patch to only update when a real change in the containerruntimeconfig
is detected. It also fetches a fresh config so that we limit the number of "object
has been updated" messages.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686282

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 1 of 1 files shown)


umohnani8:
@runcom @mrunalp @rphillips PTAL

kikisdeliveryservice:
Does this need updated tests? Seems like 2 funcs were changed a lot.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -366,12 +379,26 @@ func (ctrl *Controller) generateOriginalContainerRuntimeConfigs(role string) (*i
...
+		return lerr
+	})
+	// If an error occured in updating the status just log it
+	if statusUpdateErr != nil {
+		glog.Warningf("error updating container runtime config status: %v", statusUpdateErr)
 	}
-	_, lerr := ctrl.client.MachineconfigurationV1().ContainerRuntimeConfigs().UpdateStatus(cfg)
-	return lerr
+	// Want to return the actual error received from the sync function
+	return err
```
kikisdeliveryservice: I think the `lerr` `err` is kind of confusing. Perhaps renaming the func argument `err` to something like syncErr and renaming `lerr` -> `err` would be a little more clear?

umohnani8: Renamed it to `updateErr`, hopefully that is more clear.

umohnani8:
@kikisdeliveryservice the tests should be fine. This is just ensuring that the correct error is being returned for the sync functions.

umohnani8:
Tests are green. @cgwalters @runcom PTAL

pkg/controller/container-runtime-config/container_runtime_config_controller.go
kikisdeliveryservice: thanks this is perfect!

PR #557, opened 2019-03-18
pkg/daemon: fix oncefrom panic

runcom:

**- What I did**

we were panicking in onceFrom since the dn.queue isn't initialized in
onceFrom.

**- How to verify it**

rhel scaleup CI

/cc @vrutkovs 


vrutkovs:
This should fix crash in https://github.com/openshift/machine-config-operator/pull/557

runcom:
this is now included in #483 so @vrutkovs can test it out as a whole

Issue #254, opened 2018-12-20
Add an "Unreconcilable" label distinct from degraded


cgwalters:
Building on https://github.com/openshift/machine-config-operator/issues/234
First, we should never "reboot loop".

The two other outcomes today are (config updated successfully | degraded)

Now, what I want to argue is that we shouldn't go degraded if we can't reconcile a new config.  Going degraded is going to be expensive in some cases (particularly for masters).  An admin will want to *know* that e.g. the machineconfig they created is going to reprovision their entire cluster.

Let's introduce a distinct "unreconcilable" label.  If for example the admin messes up a machineconfig, they should totally have the option to just revert it and not degrade their machines.

IOW, degraded is *only* for the case where the node changed out from under us in an unexpected way, or some sort of fatal error (rpm-ostree isn't working, whatever).

Now, it may be that the admin *wants* to eat that reprovisioning; the MAO should also support reprovisioning unreconcilable systems too.  But let's distinguish those two cases.

Related: https://github.com/openshift/machine-config-operator/issues/234

kikisdeliveryservice:
Adding a reminder to myself to add some thoughts to this since I use reconcilable a lot and have degraded so many nodes accidentally.

jlebon:
Hmm, though if the MAO is going to reprovision degraded and unconcilable systems, then won't the admin be racing against the MAO either way?

It's also not clear to me how well we want to support admins creating their own MCs and risk the chance of blowing up their cluster vs a more validated path through the console or something.

kikisdeliveryservice:
@jlebon makes a good point generally: how much of our varying workflows is consistent with the workflow of real admins/users? I think in the New Year it's worth hashing this out more because sometimes it's not clear how much of an issue things are in, say AWS vs libvirt, pinned vs master, or come up due to our unsupported-and-undesirable-for-users tinkering.  Thought of this bc @smarterclayton made a point of saying that we should be using (at least some of the time) AWS, etc.. and having a closer to real life user experience.

cc: @ashcrow @yuqi-zhang

yuqi-zhang:
So is the proposed workflow then if we're in a unreconcilable state, sleep the node (but don't degrade) so the admin can be alerted and remove/apply a new MC? Should we maybe e.g. taint the node as well?

[...]

PR #560, opened 2019-03-19
pkg/daemon: don't flip Degraded/Working and add an Unreconcilable state

runcom:
pkg/daemon: keep degraded state after failure

Now that the Degraded state is just a state with the only scope of being
informative, keep it instead of just flipping between Working->Degraded
when we fail. When an MC degrades the machine, it still keeps retrying
to reconcile but at least administrators know that something is wrong
instantly.

pkg/daemon: add an Unreconcilable state

This new state is, like Degraded, the way we communicate to
administrators that something went specifically wrong with their
MachineConfigs. A natural follow up to this would be going ahead and
expose the real error on the object itself. With this state admins can also
just delete the bad MC and roll back.

### DIFF
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -18,8 +18,11 @@ const (
 	MachineConfigDaemonStateWorking = "Working"
 	// MachineConfigDaemonStateDone is set by daemon when it is done applying an update.
 	MachineConfigDaemonStateDone = "Done"
-	// MachineConfigDaemonStateDegraded is set by daemon when update cannot be applied.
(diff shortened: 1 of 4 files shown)


runcom:
This should help with https://github.com/openshift/machine-config-operator/issues/254 but that issue talks about a label (not an annotation state like this PR does). Also related to https://github.com/openshift/machine-config-operator/issues/234 which could be a follow up to this...

cgwalters:
Don't we also need to remove `Degraded` on success?

runcom:
> Don't we also need to remove `Degraded` on success?

uhm, if the update goes through, we reboot and set `Done` directly

cgwalters:
> uhm, if the update goes through, we reboot and set Done directly

Right, duh!

/lgtm

PR #566, opened 2019-03-20
 render: Add `core` user explicitly to groups

cgwalters:
This is prep for increasing alignment with FCOS where we'd
like to not have the user hardcoded.

See also: openshift/installer#1445

This way the privileged user is only created if a SSH key is
provided.

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -152,7 +152,13 @@ func generateSSHConfig(sshKey string, role string) (*mcfgv1.MachineConfig, error
 	// validated Ignition Config (instead of creating one by hand) that is then added to a
 	// MachineConfig.
 	var tempctCfg cttypes.Config
-	tempUser := cttypes.User{Name: "core", SSHAuthorizedKeys: []string{sshKey}}
+
+	// Keep the set of groups here in sync with `coreos-useradd-core.service`
+	// from https://github.com/coreos/fedora-coreos-config
+	// This list is also referenced in the installer.
+	tempUser := cttypes.User{Name: "core",
+		Groups: []string{"wheel", "sudo", "adm", "systemd-journal"},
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
We explicitly don't support groups -we don't support Pass:Users:SSH, so I don't see how we can present MCs to the user where core has a group?  They will expect to be able to change this and we do not allow it.

cgwalters:
> so I don't see how we can present MCs to the user where core has a group?

The user can't edit it.  But actually you're right in that may  cause a problem for the MCD, even though it will be correctly handled by the Ignition run at provisioning time.

Although actually I think it should work, because we do:

```
		if !reflect.DeepEqual(oldIgn.Passwd.Groups, newIgn.Passwd.Groups) {
			return errors.New("ignition Passwd Groups section contains changes")
		}
```

And unless the user tries to change the groups, the old/new will be the same (previously empty, now with those hardcoded values).

kikisdeliveryservice:
I think it's really confusing to have something in the MC that they cannot edit and will try to edit as opposed to just not showing it to them in the first place.   For better or for worse, we are using MCs as a UI, and I think this is bound to cause confusion.

kikisdeliveryservice:
Would this mean that they user has to copy the user:groups that they dont even know exist into a new MC if they are just trying to update their keys via a new MC?

For ex: i run installer and give it some keys. Later I want to update the keys so I create a new MC with user: core and keys:... if I dont copy over some groups that I dont even know about, my MC will fail?

What if I want to keep my keys and remove the groups? I can't change the groups without getting rid of the user and keys?

kikisdeliveryservice:
Spoke to @cgwalters and he's trying a different approach, for now will mark this as hold.

/hold

kikisdeliveryservice:

Feel free to reopen.

PR #568, opened 2019-03-21
docs: Directories and links aren't supported

cgwalters:
Reported by QE.  Now separately we seemed not to fail trying to
do this for some reason; will investigate that.

### DIFF
--- a/docs/MachineConfigDaemon.md
+++ b/docs/MachineConfigDaemon.md
@@ -26,15 +26,15 @@ Not all Ignition config sections are supported; see the following table:
 
 Ignition sections | Supported
 --- | ---
+Files | YES
+systemd Units | YES
 Networkd | NO
 Users | NO *
-Directories | YES
-Files | YES
+Directories | NO
 FileSystems | NO
-Links | YES
+Links | NO
 Disks | NO
 RAID | NO
-systemd Units | YES
 
 \* At this time only updates to `sshAuthorizedKeys` for user `core` are permitted. Please see [Update-SSHKeys](./Update-SSHKeys.md) for details.
 


cgwalters:
/hold

more to this

schituku:
https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigDaemon.md#supported-vs-unsupported-ignition-config-changes
      Directories, files, links and systemd units are supported ignition config changes as per the doc

jlebon:
Related: #328

cgwalters:
Yep, I updated this to mark links as unsupported too.  IOW just files and systemd units are supported.

kikisdeliveryservice:
good catch @cgwalters 

/lgtm

kikisdeliveryservice:
is this good to go? still has a hold on it.

schituku:
It looks like the doc based on which I reported this issue is now updated to have the directories and links not supported. Does that mean that this PR will be closed and will not make it? @cgwalters @jlebon 
https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigDaemon.md#supported-vs-unsupported-ignition-config-changes

cgwalters:
I don't quite understand, which doc are you talking about?  Do you see something wrong with the code/docs in this repo?

schituku:
I'm referencing the table in [1] which lists the supported vs unsupported. Last week it used to say
Links YES
Directories YES
    Now it says NO.

[1] https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigDaemon.md#supported-vs-unsupported-ignition-config-changes

kikisdeliveryservice:
@schituku we have updated the table to accurately reflect what is supported. The current chart is the canonical status of our ignition support. Please use that table and its contents going forward.

Issue #571, opened 2019-03-22
Error finding pools for node: node belongs to more than one MachineConfigPool


praveenkumar:
**Description**

machine-config-operator is in False state because master is labeled as master and worker which should be the case.

**Steps to reproduce the issue:**
1.  Use https://github.com/openshift/installer/blob/master/docs/user/customization.md#control-plane-with-no-taints to make master node no-taint one.
2. Edit the $CLUSTER_DIR/openshift/99_openshift-cluster-api_master-machines-0.yaml and add worker as label also
3. start the installer on libvirt provider using this manifest.

**Describe the results you received:**

```
$ oc get co | grep machine-config
machine-config                        4.0.0-0.alpha-2019-03-22-124110   False       True          True      18m
$ oc get co machine-config -oyaml
apiVersion: config.openshift.io/v1
kind: ClusterOperator
metadata:
  creationTimestamp: 2019-03-22T15:48:43Z
  generation: 1
  name: machine-config
  resourceVersion: "16803"
  selfLink: /apis/config.openshift.io/v1/clusteroperators/machine-config
  uid: f8c2b310-4cb9-11e9-bbb1-52fdfc072182
spec: {}
status:
  conditions:
  - lastTransitionTime: 2019-03-22T15:48:43Z
    message: Cluster not available for 4.0.0-0.alpha-2019-03-22-124110
    status: "False"
    type: Available
  - lastTransitionTime: 2019-03-22T15:48:43Z
    message: Cluster is bootstrapping 4.0.0-0.alpha-2019-03-22-124110
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-03-22T15:49:35Z
    message: 'Failed to resync 4.0.0-0.alpha-2019-03-22-124110 because: error pool
      master is not ready. status: (total: 1, updated: 0, unavailable: 0)'
    reason: 'error pool master is not ready. status: (total: 1, updated: 0, unavailable:
      0)'
    status: "True"
    type: Failing
  extension:
    master: 0 out of 1 nodes have updated to latest configuration rendered-master-020cc6964b5847dfc037fc2f53a192de
    worker: 0 out of 1 nodes have updated to latest configuration rendered-worker-60eee69a909e9f4604754d616f4dd3db
  relatedObjects:
  - group: ""
    name: openshift-machine-config-operator
    resource: namespaces
  versions:
  - name: operator
    version: 4.0.0-0.alpha-2019-03-22-124110
    
$ oc get csr
[...]

PR #575, opened 2019-03-25
controller/node: Support master/worker combined, and 1 custom role

cgwalters:
For CodeReady Containers and general single-node testing cases,
we need to support having a single node with both master/worker
labels.  In this case, it should fall under the master pool.

We also want to support users defining custom pools that conceptually
"derive" from worker.  If we find both `node-role/worker` and
`node-role/$x`, choose `$x`.  But we don't support more than one
custom role.

Closes: https://github.com/openshift/machine-config-operator/issues/571

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -254,6 +254,9 @@ func nodeChanged(old, cur *corev1.Node) bool {
 	return false
 }
 
+// getPoolForNode chooses the MachineConfigPool that should be used for a given node.
+// It disambiguates in the case where e.g. a node has both master/worker roles applied,
+// and where a custom role may be used.
 func (ctrl *Controller) getPoolForNode(node *corev1.Node) (*mcfgv1.MachineConfigPool, error) {
 	pl, err := ctrl.mcpLister.List(labels.Everything())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/node/node_controller_test.go
```diff
@@ -233,21 +233,30 @@ func TestGetPoolForNode(t *testing.T) {
...
 
-		expected: newMachineConfigPool("test-cluster-master", metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role", "master"), nil, "v0"),
+		expected: newMachineConfigPool("master", metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role", "master"), nil, "v0"),
+		err:      false,
+	}, {
+		pools: []*mcfgv1.MachineConfigPool{
+			newMachineConfigPool("master", metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role/master", ""), nil, "v0"),
+			newMachineConfigPool("worker", metav1.AddLabelToSelector(&metav1.LabelSelector{}, "node-role/worker", ""), nil, "v0"),
+		},
+		nodeLabel: map[string]string{"node-role/master": "", "node-role/worker": ""},
```
runcom: can we just add another test with the logic of the custom worker label? (if it's here already, sorry for the noise)

runcom:
just one comment for an additional test, lgtm otherwise

/approve

pkg/controller/node/node_controller_test.go
cgwalters: Done!

pkg/controller/node/node_controller.go
```diff
@@ -279,10 +282,37 @@ func (ctrl *Controller) getPoolForNode(node *corev1.Node) (*mcfgv1.MachineConfig
...
+	if len(custom) > 1 {
+		return nil, fmt.Errorf("node %s belongs to %d custom roles, cannot proceed with this Node", node.Name, len(custom))
+	} else if len(custom) == 1 {
+		// We don't support making custom pools for masters
+		if master != nil {
+			return nil, fmt.Errorf("node %s has both master role and custom role %s", node.Name, custom[0].Name)
+		}
+		// One custom role, let's use its pool
+		return custom[0], nil
+	} else if master != nil {
```
kikisdeliveryservice: just to check if i am misreading but to get into this third else if.. won't we also get into it if a node is only tagged master and not worker? so shouldnt we check: master != nil and worker == nil??

or am i misreading?

cgwalters: > won't we also get into it if a node is only tagged master and not worker?

If it's tagged master and not worker, it's master right?

> so shouldnt we check: master != nil and worker == nil??

We're explicitly choosing master in that case for the CRC and single-machine-test cases; if master is set we want master, regardless of whether `worker` is nil-or-not.

kikisdeliveryservice: thanks i needed that! 

if 1 custom: custom
if master: master
if only worker: worker

Got it!

kikisdeliveryservice:
this LGTM will defer to @runcom for another pair of :eyes: 

do we need to document this somewhere for the codeready folks?

runcom:
I believe all the new cases are covered into unit testing and CI is fully passing.

/lgtm

Issue #388, opened 2019-02-07
support kernel arguments


cgwalters:
We landed a lot of good work in `rpm-ostree kargs`.  What we need to do is wrap that in MachineConfigs.  One decision point is whether this is an extension to Ignition (if specified then does Ignition reboot?).  Or is it a new item in the MachineConfig spec itself?

And the MCD would need to understand how to reconcile the user-specified kargs versus what's the default.

mrguitar:
@runcom ^^

cgwalters:
A lot of discussion around this; one thing that seems increasingly clear is that we should use `MachineConfig` as a "sugar" for Ignition where it makes sense.

IOW, we start from this:

``
diff --git a/pkg/apis/machineconfiguration.openshift.io/v1/types.go b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
index f7df1d5..318506c 100644
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -226,6 +226,8 @@ type MachineConfig struct {
 
 // MachineConfigSpec is the for MachineConfig
 type MachineConfigSpec struct {
+	// KernelArguments defines additional kernel arguments beyond the host defaults.
+	KernelArguments string `json:"kernelArguments"`
 	// OSImageURL specifies the remote location that will be used to
 	// fetch the OS.
 	OSImageURL string `json:"osImageURL"`
```

runcom:
> One thing that seems increasingly clear is that we should use `MachineConfig` as a "sugar" for Ignition where it makes sense.

cannot agree more with this...

cgwalters:
With https://github.com/projectatomic/rpm-ostree/issues/1802

the MCD in `upgrade.go` when detecting a diff between the current/target `KernelArgs` would first call `--remove` on the old ones, then `--add` for the new ones.

There's an alternative implementation path where the MCD reads the BLS config performs the diff internally, and calls the appropriate `rpm-ostree kargs ` invocation.

runcom:
> With [projectatomic/rpm-ostree#1802](https://github.com/projectatomic/rpm-ostree/issues/1802)
> 
> the MCD in `upgrade.go` when detecting a diff between the current/target `KernelArgs` would first call `--remove` on the old ones, then `--add` for the new ones.

if projectatomic/rpm-ostree#1802 provides a way to achieve the above it would be my preferred way. 
[...]

PR #576, opened 2019-03-25
pkg/daemon: allow reboots from external componenents

runcom:
This is a temporary stopgap to allow other componenents ask for a reboot
since the MCD is the one in charge of draining a node and rebooting it.

The logic is simple with this:

- if an update is already coming, we just go through the update and if
successful, the machine is gonna reboot just fine
- if we're not updating through MachineConfigs, then we check for a file
and drain+reboot
- the "reboot file" is the removed on MCD boot

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -319,9 +319,15 @@ func (dn *Daemon) bootstrapNode() error {
 		return err
 	}
 	dn.node = node
+	if _, err := os.Stat("/etc/defaults/rhcos/reboot-needed"); err == nil {
+		if err := os.Remove("/etc/defaults/rhcos/reboot-needed"); err != nil {
+			return err
+		}
+	}
 	if err := dn.CheckStateOnBoot(); err != nil {
 		return err
 	}
+
 	// finished syncing node for the first time
 	dn.booting = false
 	return nil
@@ -330,6 +336,8 @@ func (dn *Daemon) bootstrapNode() error {
 func (dn *Daemon) handleErr(err error, key interface{}) {
 	if err == nil {
 		dn.queue.Forget(key)
+		// This is a workaround to have the controller keep calling the syncHandler for the MCD reboot signaling
+		dn.queue.AddAfter(key, 5*time.Minute)
 		return
 	}
 
@@ -397,6 +405,18 @@ func (dn *Daemon) syncNode(key string) error {
 		if needUpdate {
 			if err := dn.triggerUpdateWithMachineConfig(nil, nil); err != nil {
 				glog.Infof("Unable to apply update: %s", err)
+				return err
+			}
+		}
+		// FIXME: this is a temporary workaround, remove once an MCD signaling system is in place
+		if _, err := os.Stat("/etc/defaults/rhcos/reboot-needed"); err == nil {
+			dn.catchIgnoreSIGTERM()
+			if err := dn.drainNode(); err != nil {
+				dn.cancelSIGTERM()
+				return err
+			}
+			if err := dn.reboot("rebooting due to another component signaling"); err != nil {
+				dn.catchIgnoreSIGTERM()
 				return err
 			}
 		}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -94,6 +94,45 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
 	}
 }
 
+func (dn *Daemon) drainNode() error {
+	// Skip draining of the node when we're not cluster driven
+	if dn.onceFrom != "" {
+		return nil
+	}
+
+	glog.Info("Update prepared; draining the node")
+	dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Drain", "Draining node to update config.")
+
+	backoff := wait.Backoff{
+		Steps:    5,
+		Duration: 10 * time.Second,
+		Factor:   2,
+	}
+	var lastErr error
+	if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		err := drain.Drain(dn.kubeClient, []*corev1.Node{dn.node}, &drain.DrainOptions{
+			DeleteLocalData:    true,
+			Force:              true,
+			GracePeriodSeconds: 600,
+			IgnoreDaemonsets:   true,
+		})
+		if err == nil {
+			return true, nil
+		}
+		lastErr = err
+		glog.Infof("Draining failed with: %v, retrying", err)
+		return false, nil
+
+	}); err != nil {
+		if err == wait.ErrWaitTimeout {
+			return errors.Wrapf(lastErr, "failed to drain node (%d tries): %v", backoff.Steps, err)
+		}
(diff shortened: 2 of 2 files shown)


runcom:
/hold

cc @ashcrow @crawford

pkg/daemon/daemon.go
```diff
@@ -400,6 +406,15 @@ func (dn *Daemon) syncNode(key string) error {
 				return err
 			}
 		}
+		// FIXME: this is a temporary workaround, remove once an MCD signaling system is in place
```
ashcrow: :+1:

pkg/daemon/daemon.go
```diff
@@ -319,9 +319,15 @@ func (dn *Daemon) bootstrapNode() error {
 		return err
 	}
 	dn.node = node
+	if _, err := os.Stat("/etc/defaults/rhcos/reboot-needed"); err == nil {
```
cgwalters: IMO, `/etc/defaults` is a legacy of SySV init; no reason to put anything new there.

The right place for this is more likely `/run/machine-config-operator/reboot-needed` or so.  (We also have `/etc/machine-config-daemon` used for `/etc/machine-config-daemon/state.json` but I think `/run` makes more sense since it goes away automatically if the machine is rebooted)

pkg/daemon/daemon.go
```diff
@@ -400,6 +406,15 @@ func (dn *Daemon) syncNode(key string) error {
 				return err
 			}
 		}
+		// FIXME: this is a temporary workaround, remove once an MCD signaling system is in place
+		if _, err := os.Stat("/etc/defaults/rhcos/reboot-needed"); err == nil {
```
ashcrow: :+1:

cgwalters:
What other components are going to be asking for a reboot?

pkg/daemon/daemon.go
runcom: cc @ashcrow

runcom:
> What other components are going to be asking for a reboot?

hopefully just the one @ashcrow needs for 4.1

ashcrow:
Yeah, it's just RHCOS and Installer team for 4.1

Context: https://url.corp.redhat.com/bc7c76c

cgwalters:
I don't understand why that work isn't driven by the MCD.  We already have other needs to adjust kernel arguments (per https://github.com/openshift/machine-config-operator/issues/388 ) - and further it's going to be problematic to have some other service running `rpm-ostree kargs` which will adjust deployments at possibly the same time as the MCD is possibly executing pivot.

As is today AFAICS it will race; there's no ordering relationship between that service and `pivot.service` which runs to do the early pivot.

Further it overlaps heavily with https://github.com/ostreedev/ostree/issues/479 which Robert is working on to aid in dropping Anaconda and removing the dependency on grub2 with ostree in general which will solve various other issues.

cgwalters:
Short term hack, we can make this new service `Before=pivot.service` to address the race.

ashcrow:
@cgwalters will update the host level service with

```
Before=pivot.service
```

cgwalters:
Remember that the "early pivot" happens before the node has joined the cluster.  So the MCD isn't involved here - no need to drain anything.

Can we de-scope this from "arbitrary kernel arguments are supported changing at any time" to "If `/etc/openshift-nosmt` is present during `ConditionFirstBoot`, then `nosmt` is added to kernel arguments` at "early pivot" time?

Today, we can assume that `pivot.service` will itself cause a reboot, so kernel args handling doesn't need to additionally do so.

Maybe the simplest is actually to add this to `pivot.service` itself as `/etc/pivot/kernel-args`.  It's already in a place to do all of this transactionally anyways.

(Though that leaves the "BYO host" scenario, but will they really be expecting us to inject kernel args and reboot traditional hosts?  This seems like more in the territory of "you brought it, you manage it"?)

cgwalters:
To elaborate on this a bit...it looks like the current prototype service runs every boot.  That greatly increases the scope; if we only have to care about it during early pivot, thereafter the `nosmt` kernel argument will be correctly carried forwards across upgrades etc.  We don't have a concern about the "source of truth" for kernel arguments.  (The source of truth is the BLS config files as is the original ostree design - while BLS didn't take over the world it's at least a somewhat agreed upon place for kernel arguments implemented by multiple bootloaders.  Invoking `rpm-ostree kargs` is unfortunately still necessary though to regenerate the bootloader config)

ashcrow:
I'm OK with the above proposal for the near term. I'm going to bounce it off other folks who were part of the design before we commit to change.

The POC today runs whenever the `/etc/default/rhcos/kargs` file changes on an RHCOS or RHEL machine. This can be right after boot, or whenever the restricted file is written to.

abhinavdahiya:
> /cc @abhinavdahiya
I'm fine if we support `nosmt` thorugh a special file.. but I would like decide the file location as that would unblock https://github.com/openshift/installer/pull/1392 .

ashcrow:
> Can we de-scope this from "arbitrary kernel arguments are supported changing at any time" to "If /etc/openshift-nosmt is present during ConditionFirstBoot, then nosmt is added to kernel arguments` at "early pivot" time?

I spoke with others and they are cool with this for 4.1. How about `/etc/pivot/kernel-args`. Whatever is put in there will be applied via `rpm-ostree kargs`? The file change will come from one of the following locations at a time:

- openshift-installer
- MCD (via an MC)

If this sounds good I'll shelve my system service API for now and throw something together starting in the morning.

runcom:
> I spoke with others and they are cool with this for 4.1. How about `/etc/pivot/kernel-args`. Whatever is put in there will be applied via `rpm-ostree kargs`? The file change will come from one of the following locations at a time:
> 
> * openshift-installer
> * MCD (via an MC)

isn't this allowing "arbitrary kernel arguments"? for 4.1, we just want `nosmt` to take effect at early pivot.

For the future, then yeah, I agree with what you wrote, that can be a path towards allowing general and arbitrary kargs.

ashcrow:
> Isn't this allowing "arbitrary kernel arguments"? 

It can be expanded to be arbitrary. For now it will only be the subset for tuning.

> for 4.1, we just want nosmt to take effect at early pivot.

Almost. It needs to be in early pivot *or* toggled later.

ashcrow:
@runcom thank you for putting this PR together, but I think it can be closed. I'll hopefully have something up today or tomorrow in it's place.

runcom:
Thanks!!!

cgwalters:
>  It needs to be in early pivot or toggled later.

If we really need to support toggling it later I don't see how we avoid implementing https://github.com/openshift/machine-config-operator/issues/388 right?

ashcrow:
> If we really need to support toggling it later I don't see how we avoid implementing #388 right?

For a later version, yes. But if we scope this to the few tuning parameters that are needed we the current MCO/MC and update to pivot should work. The installer can lay down a file at install time as needed and, if the cluster admin wants to modify the supported tuning options, they can lay down an updated `/etc/pivot/kernel-args` (which will only white list the options we're allowing for now).

ashcrow:
Actually, I see what you mean now @cgwalters. Since `pivot` ONLY runs when MCD gets a new `machine-os-content`.

PR #577, opened 2019-03-26
controller/bootstrap: use files with multiple yaml documents

abhinavdahiya:
**- What I did**
Users can push manifests during bootstrap that of the form:

```yaml
---
```

Especially for the installer: setting authorizes_keys [1] and setting hyperthreading [2] will push a manifest that includes multiple machineconfig objects for
control-plane (master) and compute (worker) roles.

Single file with multiple k8s objects separated by `---` is also a supported structure for `oc create|apply` ie. there is a high chance that users trying to push machineconfigs at
install time might create such files.

This commit allows bootstrap controller to read all k8s objects, even ones described above to find all the `machineconfiguration.openshift.io` Objects.

[1]: https://github.com/openshift/installer/pull/1150
[2]: https://github.com/openshift/installer/pull/1392

**- How to verify it**

Added a unit test.

`Allow bootstrap controller to read multi object yaml files`

### DIFF
--- a/pkg/controller/bootstrap/bootstrap.go
+++ b/pkg/controller/bootstrap/bootstrap.go
@@ -81,9 +81,12 @@ func (b *Bootstrap) Run(destDir string) error {
 		for idx, m := range manifests {
 			obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(v1.SchemeGroupVersion), m.Raw)
 			if err != nil {
-				glog.V(4).Infof("skipping path %q [%d] manifest because of error: %v", file.Name(), idx+1, err)
-				// don't care
-				continue
+				if runtime.IsNotRegisteredError(err) {
+					// don't care
+					glog.V(4).Infof("skipping path %q [%d] manifest because it is not part of expected api group: %v", file.Name(), idx+1, err)
+					continue
+				}
+				return fmt.Errorf("error parsing %q [%d] manifest: %v", file.Name(), idx+1, err)
 			}
 
 			switch obj := obji.(type) {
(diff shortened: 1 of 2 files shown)


pkg/controller/bootstrap/bootstrap.go
```diff
@@ -63,28 +67,35 @@ func (b *Bootstrap) Run(destDir string) error {
...
-			configs = append(configs, obj)
-		case *v1.ControllerConfig:
-			cconfig = obj
-		default:
-			glog.Infof("skipping %q %T", path, obji)
+		for idx, m := range manifests {
+			obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(v1.SchemeGroupVersion), m.Raw)
+			if err != nil {
+				glog.V(4).Infof("skipping path %q [%d] manifest because of error: %v", file.Name(), idx+1, err)
+				// don't care
```
cgwalters: I know this isn't a problem introduced in this patch, but I think silently ignoring errors is going to lead someone to a lot of debugging pain at some point.

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -63,28 +67,35 @@ func (b *Bootstrap) Run(destDir string) error {
...
+
+			switch obj := obji.(type) {
+			case *v1.MachineConfigPool:
+				pools = append(pools, obj)
+			case *v1.MachineConfig:
+				configs = append(configs, obj)
+			case *v1.ControllerConfig:
+				cconfig = obj
+			default:
+				glog.Infof("skipping %q [%d] manifest because %T", file.Name(), idx+1, obji)
```
cgwalters: `s/because %T/of unhandled type %T/` ?

abhinavdahiya: we are skipping because of the type, not `unhandled`

pkg/controller/bootstrap/bootstrap.go
abhinavdahiya: pushed df19bd7 that makes decoding of invalid objects in mco apigroup fatal.

PR #578, opened 2019-03-26
kubeletconfig_features: fix ignition object creation

rphillips:
**- What I did**
FeatureGates creates a brand new ignition object, and the ignition
subsystem validates the Version field is a valid Semver. This fix uses
the common constructor to create a valid ignition object.

Note: This patch also changes the kubelet_config_controller to make it consistent
with how the featuregates controller constructs the ignition object.

**- How to verify it**

None


runcom:
cc @umohnani8 for container runtime stuff

PR #580, opened 2019-03-26
Bug 1677198: pkg/daemon: allow empty new links section

runcom:
This PR is further fixing https://bugzilla.redhat.com/show_bug.cgi?id=1677198 by
allowing an old cluster to contain links (coming from #535) which are
then removed when upgrading to post #540.

What's happening to QE is that:

1) they're starting a cluster with an MCO version which contains #535
2) they're upgrading to a payload which doesn't have #535 cause it has #540 which reverts #535

The above means that point 1) generates MachineConfigs with an unsupported symlink and when
upgrading to 2), the symlink is removed causing drift and an unreconcilable error.

QE can start testing with a newer MCO version to avoid this, but for
ease of testing, let's add this snippet to make sure getting rid of
links, which we don't support anyway, just works.


runcom:
@cgwalters ptal

cgwalters:
Software...fix one thing break another, always the story...

/lgtm

PR #581, opened 2019-03-26
systemd: install systemd.conf to enable CPU Accounting

rphillips:
Install system.conf to enable CPU Accounting.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1692131

**- What I did**
Install a custom system.conf file enabling CPU, BlockIO, and Memory accounting.

ref: https://github.com/openshift/origin/blob/master/contrib/systemd/origin-accounting.conf

Enable CPU Accounting within systemd

/cc @sjenning @derekwaynecarr

### DIFF
--- a/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-systemd-system.conf.d-kubelet-cgroups.conf
+++ b/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-systemd-system.conf.d-kubelet-cgroups.conf
@@ -1,0 +1,6 @@
+contents:
+  source: data:,%23%20Turning%20on%20Accounting%20helps%20track%20down%20performance%20issues.%0A%5BManager%5D%0ADefaultCPUAccounting%3Dyes%0ADefaultMemoryAccounting%3Dyes%0ADefaultBlockIOAccounting%3Dyes%0A
+  verification: {}
+filesystem: root
+mode: 420
+path: /etc/systemd/system.conf.d/kubelet-cgroups.conf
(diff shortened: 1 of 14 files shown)


pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-systemd-system.conf.d-origin-accounting.conf
```diff
@@ -0,0 +1,6 @@
+contents:
```
smarterclayton: Don't use `origin`.  Use `kubelet-cgroups` or something.

templates/worker/00-worker/_base/files/system-conf.yaml
```diff
@@ -0,0 +1,9 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/systemd/system.conf.d/origin-accounting.conf"
```
smarterclayton: `kubelet-cgroup-accounting.conf`

Can you add a description into the system unit so people know why this is there?

rphillips:
@smarterclayton thanks. Fix both comments.

sjenning:
do we want to enable accounting just for {kubelet,crio}.service?  enabling accounting globally will incur kernel overhead and will make the stats summary api response much larger.

sjenning:
meh, nevermind.  there aren't that many service on rhcos anyway and global memory accounting is already on by default.
/lgtm

smarterclayton:
lgtm as well, approving

rphillips:
/test unit

works locally...

rphillips:
Looks like master might be broken?

```
--- FAIL: TestGenerateMachineConfigs (0.17s)
	render_test.go:363: can't find actual file test_data/templates/master/00-master/vsphere/files:
```

rphillips:
Had to rebase and regenerate the templates due to a change on master.

Issue #582, opened 2019-03-26
Add new node.openshift.io/rhelmajor=7|8 label


cgwalters:
https://github.com/openshift/machine-config-operator/pull/514 added the OS version but it isn't usable as a selector for e.g. daemonsets that just care about RHEL major (e.g. iptables vs nftables).

Let's add a new label and implement it for both RHCOS and BYO RHEL.

Proposal is: `node.openshift.io/rhelmajor=7|8`.

ashcrow:
`node.openshift.io/rhelmajor=7|8` makes sense to me.

ashcrow:
The RHCOS `VERSION` or `VERSION_ID` can be parsed to set this like so: `echo $VERSION | cut -d . -f 2`

sjenning:
fyi @rphillips

chambridge:
We wish to use this as part of our metrics information that will be utilized for subscription entitlement reconciliation. The sooner we have this information the better.

sdodson:
> We wish to use this as part of our metrics information that will be utilized for subscription entitlement reconciliation. The sooner we have this information the better.

Do you also have a use for differentiating between RHEL and RHCOS for those same purposes?

chambridge:
We are currently using the label `node.openshift.io/os_id` which shows up on `kube_node_labels` as `label_node_openshift_io_os_id` to differentiate between RHEL and RHCOS.

adelton:
Wouldn't using something like `PLATFORM_ID="platform:el8"` directly from `/etc/os-release` be a reasonable value and distinguisher? So something like `node.openshift.io/platform=platform:el8`?

adelton:
Or is it important for the value to be a single integer value?

ashcrow:
In the latest devel builds I do see a new key in `/etc/os-release` called `RHEL_VERSION` that @cgwalters has kindly added.

cgwalters:
> Wouldn't using something like PLATFORM_ID="platform:el8" directly from /etc/os-release be a reasonable value and distinguisher? 

That doesn't seem to exist in RHEL7.  It's some modularity invention right?  Are you aware of any docs for what it means?  Doesn't seem to be more useful than just `VERSION_ID`  to me?

adelton:
You are right, `PLATFORM_ID` does not seem documented at https://www.freedesktop.org/software/systemd/man/os-release.html.

adelton:
[...]

PR #584, opened 2019-03-26
WIP: templates: Add node.openshift.io/rhelmajor label, drop os_version one

cgwalters:
This removes the functionality added in
#514
which I don't think really worked for what people needed; a node
selector can't parse the value.

All we care about is currently 7 or 8; this signals things like
`iptables` vs `nftables`.

And the presence of this label also signals "RHEL" too;
though we keep the `os_id` label in case e.g. a daemonset should only
run on RHCOS (or traditional RHEL).

Closes: #582

### DIFF
--- a/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet-rhelmajor.service
+++ b/pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet-rhelmajor.service
@@ -1,0 +1,16 @@
+contents: |
+  [Unit]
+  Description=Generate rhelmajor version for kubelet
+  Before=kubelet.service
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  EnvironmentFile=/etc/os-release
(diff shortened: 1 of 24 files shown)


cgwalters:
Haven't done full testing of this yet, submitting WIP for visibility/early review.

kikisdeliveryservice:
to clarify, if:

- rhelmajor = 7|8 then os_id = nil
- os_id != nil then rhelmajor = nil

?

cgwalters:
> to clarify, if:

Mmm, `os_id` is always set as it was before, I haven't changed that (right?).  We will either have `os_id=rhcos` or `os_id=rhel`.

kikisdeliveryservice:
@cgwalters thanks, got it!

cgwalters:
OK so I discovered this: `--node-labels mapStringString     <Warning: Alpha feature> Labels to add when registering the node in the cluster.`

In other words, the labels aren't dynamic which is rather problematic for something that has a version number.  I think the MCD should probably start managing the labels.

cgwalters:
Obsoleting in favor of https://github.com/openshift/machine-config-operator/pull/657

PR #586, opened 2019-03-26
Remove amd64 from dockerfile copy path to support Mulit-Arch Builds.

jeffdyoung:
**- What I did**
Removed hardcoded 'amd64' in build path from the .rhel7 Dockerfiles so we can start building these containers on other architectures.

**- How to verify it**
 Build the containers locally (with doozer or some other method).

Remove amd64 from dockerfile copy path to support Mulit-Arch Builds.

### DIFF
--- a/Dockerfile.machine-config-controller.rhel7
+++ b/Dockerfile.machine-config-controller.rhel7
@@ -1,9 +1,11 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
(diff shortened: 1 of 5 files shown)


jeffdyoung:
@cgwalters Can you help me get this through? We need to start building multi-arch containers in 4.1 so we can test and deliver by 4.2

cgwalters:
We should fix the non-rhel7 dockerfiles too, but doesn't have to be this commit.

/approve
/lgtm

jeffdyoung:
@cgwalters Won't OKD and OCP be built off of ubi7 or ubi8 eventually? I was thinking Dockerfile.rhel would turn into Dockerfile.ubi

PR #591, opened 2019-03-28
 operator: sync the cloud config in bootstrap and cluster mode

abhinavdahiya:
**- What I did**

Use the cloud config defined in https://github.com/openshift/api/pull/245 & https://github.com/openshift/installer/pull/1479 to make sure mco can use it in both bootstrap and cluster.

Also updated the internal templates to write the cloudconfig to `/etc/kubernetes/cloud.conf` so that kubelets on nodes can use it.

**- How to verify it**

```
Sync the cloud config to kubelet on nodes.
```
/cc @staebler @derekwaynecarr


runcom:
/hold

is this fixing a bug and has clearance to go in?

abhinavdahiya:
This PR is required for Azure work to proceed. as kubelets for Azure need to use this cloudconfig.

crawford:
@runcom This can be merged as long as you are confident it won't affect existing functionality.

cmd/machine-config-operator/bootstrap.go
```diff
@@ -24,24 +24,25 @@ var (
 	}
 
 	bootstrapOpts struct {
+		cloudConfigFile      string
```
kikisdeliveryservice: this is just adding cloudConfigFile and alphabetizing bootstrapOpts?

pkg/operator/operator.go
```diff
@@ -421,6 +430,18 @@ func (optr *Operator) getCAsFromConfigMap(namespace, name, key string) ([]byte,
 	}
 }
 
+func (optr *Operator) getCloudConfigFromConfigMap(namespace, name, key string) (string, error) {
+	cm, err := optr.kubeClient.CoreV1().ConfigMaps(namespace).Get(name, metav1.GetOptions{})
```
cgwalters: We have `optr.clusterCmLister.ConfigMaps(namespace)` now.

cgwalters:
Only one minor nit, otherwise looks sane to me.

pkg/controller/template/render.go
```diff
@@ -359,6 +360,13 @@ func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	return "", nil
 }
 
+func cloudConfigFlag(cfg RenderConfig) (interface{}, error) {
```
kikisdeliveryservice: I don't quite understand this func? can you add comments explaining why this is here?

runcom: +1 to comment. I believe this is just a template function, so whenever a template gets processed, if you wrote cloudConfigFlag, it calls this func and renders the result.

pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
```diff
@@ -25,7 +25,7 @@ contents: |
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider=aws \
         --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
-        \
+         \
```
runcom: ?

pkg/controller/template/render.go
```diff
@@ -359,6 +360,13 @@ func cloudProvider(cfg RenderConfig) (interface{}, error) {
...
+	if len(cfg.CloudProviderConfig) > 0 {
+		return "--cloud-config=/etc/kubernetes/cloud.conf", nil
+	}
+	return "", nil
```
kikisdeliveryservice: would this func ever return an error?

runcom:
Haven't fully reviewed this but looks rightly wrapped to not affect other stuff in the MCO. @cgwalters @kikisdeliveryservice wdyt?

pkg/controller/template/test_data/templates/master/01-master-kubelet/aws/units/kubelet.service
abhinavdahiya: because of change [here](https://github.com/openshift/machine-config-operator/pull/591/files#diff-d5047a0acdc71dcfb6fc0b86a556fd4eR29)

cmd/machine-config-operator/bootstrap.go
abhinavdahiya: Yes

pkg/operator/operator.go
abhinavdahiya: done

pkg/controller/template/render.go
abhinavdahiya: added the docs.

pkg/controller/template/render.go
abhinavdahiya: nope, remove the error return.

pkg/controller/template/render.go
kikisdeliveryservice: thanks!

kikisdeliveryservice:
confirming that in CI seeing cloud.conf refs in "01-master-kubelet" and "01-worker-kubelet" that ends up in rendered-master & rendered-worker

abhinavdahiya:
rebased around #648

abhinavdahiya:
[e2e-aws](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/591/pull-ci-openshift-machine-config-operator-master-e2e-aws/3326): failed due to flake :(
```
Failing tests:

[Feature:Builds][Conformance] oc new-app  should succeed with a --name of 58 characters [Suite:openshift/conformance/parallel/minimal]
```
/retest

derekwaynecarr:
this looks fine to me, we will need to handle migration when external cloud controller managers are available when kubelets no longer need this conf file.  that is a few releases away.

Issue #447, opened 2019-02-17
Add static checks to the project in PRs


runcom:
Setup either the CI or wire in Travis to do static check on code changes, `make verify` in this repo isn't checking everything nor it's wired to check automatically. Something like https://github.com/kubernetes-sigs/cri-o/blob/master/.tool/lint + yamllint would be reasonable. Checks must be done only on the diff of a current PR, not the whole project (otherwise, there are a lot of failures and we just want to stop introducing more).

PR #595, opened 2019-04-02
skip test if not linux + hack/Makefile cleanups

runcom:
I'm occasionally developing from a mac, and found out that test is
failing cause it's assuming we're on linux and the kernel wrote the boot
id under /proc. It's not the case on osx so just skip the test if that
file isn't there.

The second commit accounts for an empty rev-parse --show-cdup on osx, and fixes that by testing the path to cd (which is not changing behavior on linux)

The other commits are doing a pass over a cleanup in hack and Makefile in order to start addressing #447 - specifically, we need to avoid stuff like #596 


runcom:
This is also needed for https://github.com/openshift/release/pull/3332

so we can have a verify job which just calls `make verify` from this which is verifying we never have issues like #596 - in the future, we can add gofmt and other checks as well

kikisdeliveryservice:
these failures have nothing to do with this PR. I'll check the forums to see if they are already reported.

ill retest later.

kikisdeliveryservice:
a similar failure has been reported in BZ 1695278.  It seems to be happening on every run, so no point of mashing retest.

kikisdeliveryservice:
openshift-bot is trolling me. :rofl:

PR #597, opened 2019-04-02
Bug 1699134: pkg/daemon: add reason annotation key for degraded/unreconcilable and bubble up errors from render controller

runcom:

**- What I did**

Close https://github.com/openshift/machine-config-operator/issues/234
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1699134

Added a new Reason annotation key set to the specific error we get when failing to reconcile + reintroduced degraded from https://github.com/runcom/machine-config-operator/commit/e753f108d3e0027500fb2c473ceb162c0fed74ae#diff-3bd49282ca65e5525da44a27caaa55c7L80 + bubbling up errors from pool

**- How to verify it**

added a check in one of the e2e

### DIFF
--- a/manifests/machineconfigpool.crd.yaml
+++ b/manifests/machineconfigpool.crd.yaml
@@ -16,6 +16,10 @@ spec:
     description: When at least one of machine is not either not updated or is in the process of updating to the desired machine config.
     name: Updating
     type: string
+  - JSONPath: .status.conditions[?(@.type=="Degraded")].status
+    description: When progress is blocked on updating one or more nodes, or the pool configuration is failing.
+    name: Degraded
+    type: string
   # group name to use for REST API: /apis/<group>/<version>
   group: machineconfiguration.openshift.io
   # list of versions supported by this CustomResourceDefinition
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -46,7 +46,7 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	unavailableMachineCount := int32(len(unavailableMachines))
 
 	degradedMachines := getDegradedMachines(pool.Status.Configuration.Name, nodes)
-	degradedReasons := make([]string, len(nodes))
+	degradedReasons := []string{}
 	for _, n := range degradedMachines {
 		reason, ok := n.Annotations[daemonconsts.MachineConfigDaemonReasonAnnotationKey]
 		if ok && reason != "" {
@@ -86,17 +86,26 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
 	}
 
-	if len(degradedMachines) > 0 {
-		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionTrue, fmt.Sprintf("%d nodes are reporting degraded status on sync", len(degradedMachines)), strings.Join(degradedReasons, ","))
+	var nodeDegraded bool
+	if degradedMachineCount > 0 {
+		nodeDegraded = true
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolNodeDegraded, corev1.ConditionTrue, fmt.Sprintf("%d nodes are reporting degraded status on sync", len(degradedMachines)), strings.Join(degradedReasons, ", "))
 		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
 	} else {
-		// Do not clear out a possible degradation if everything is ok here but render_controller reported
-		// a failure in rendering
-		poolAlreadyDegraded := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolDegraded)
-		if !poolAlreadyDegraded {
-			sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionFalse, "", "")
-			mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
-		}
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolNodeDegraded, corev1.ConditionFalse, "", "")
+		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
+	}
+
+	// here we now set the MCP Degraded field, the node_controller is the one making the call right now
+	// but we might have a dedicated controller or control loop somewhere else that understands how to
+	// set Degraded. For now, the node_controller understand NodeDegraded & RenderDegraded = Degraded.
+	renderDegraded := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolRenderDegraded)
+	if nodeDegraded || renderDegraded {
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionTrue, "", "")
+		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
+	} else {
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionFalse, "", "")
+		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
 	}
 
 	return status
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/golang/glog"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -44,12 +45,23 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	unavailableMachines := getUnavailableMachines(pool.Status.Configuration.Name, nodes)
 	unavailableMachineCount := int32(len(unavailableMachines))
 
+	degradedMachines := getDegradedMachines(pool.Status.Configuration.Name, nodes)
+	degradedReasons := make([]string, len(nodes))
+	for _, n := range degradedMachines {
+		reason, ok := n.Annotations[daemonconsts.MachineConfigDaemonReasonAnnotationKey]
+		if ok && reason != "" {
+			degradedReasons = append(degradedReasons, fmt.Sprintf("Node %s is reporting: %q", n.Name, reason))
+		}
+	}
+	degradedMachineCount := int32(len(degradedMachines))
+
 	status := mcfgv1.MachineConfigPoolStatus{
 		ObservedGeneration:      pool.Generation,
 		MachineCount:            machineCount,
 		UpdatedMachineCount:     updatedMachineCount,
 		ReadyMachineCount:       readyMachineCount,
 		UnavailableMachineCount: unavailableMachineCount,
+		DegradedMachineCount:    degradedMachineCount,
 	}
 
 	status.Configuration = pool.Status.Configuration
@@ -73,6 +85,20 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, fmt.Sprintf("All nodes are updating to %s", pool.Status.Configuration.Name), "")
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
 	}
+
+	if len(degradedMachines) > 0 {
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionTrue, fmt.Sprintf("%d nodes are reporting degraded status on sync", len(degradedMachines)), strings.Join(degradedReasons, ","))
+		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
+	} else {
+		// Do not clear out a possible degradation if everything is ok here but render_controller reported
+		// a failure in rendering
+		poolAlreadyDegraded := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolDegraded)
+		if !poolAlreadyDegraded {
+			sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionFalse, "", "")
+			mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
+		}
+	}
+
 	return status
 }
 
@@ -160,4 +186,27 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
 		}
 	}
 	return unavail
+}
+
+func getDegradedMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
+	var degraded []*corev1.Node
+	for _, node := range nodes {
+		if node.Annotations == nil {
+			continue
+		}
+		dconfig, ok := node.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey]
+		if !ok || dconfig == "" {
+			continue
+		}
+		dstate, ok := node.Annotations[daemonconsts.MachineConfigDaemonStateAnnotationKey]
+		if !ok || dstate == "" {
+			continue
+		}
+
+		if dconfig == currentConfig &&
+			(dstate == daemonconsts.MachineConfigDaemonStateDegraded || dstate == daemonconsts.MachineConfigDaemonStateUnreconcilable) {
+			degraded = append(degraded, node)
+		}
+	}
+	return degraded
 }
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -433,7 +433,20 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		return fmt.Errorf("no MachineConfigs found matching selector %v", selector)
 	}
 
-	return ctrl.syncGeneratedMachineConfig(pool, mcs)
+	if err := ctrl.syncGeneratedMachineConfig(pool, mcs); err != nil {
+		return ctrl.syncFailingStatus(pool, err)
+	}
+
+	return nil
+}
+
+func (ctrl *Controller) syncFailingStatus(pool *mcfgv1.MachineConfigPool, err error) error {
+	sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to render configuration for pool %s: %v", pool.Name, err), "")
+	mcfgv1.SetMachineConfigPoolCondition(&pool.Status, *sdegraded)
+	if _, updateErr := ctrl.client.MachineconfigurationV1().MachineConfigPools().UpdateStatus(pool); updateErr != nil {
+		glog.Errorf("Error updating MachineConfigPool %s: %v", pool.Name, updateErr)
+	}
+	return err
 }
 
 // This function will eventually contain a sane garbage collection policy for rendered MachineConfigs;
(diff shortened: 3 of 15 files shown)


pkg/daemon/writer.go
```diff
@@ -97,6 +97,7 @@ func (nw *NodeWriter) SetUnreconcilable(err error, client corev1.NodeInterface,
 	glog.Errorf("Marking Unreconcilable due to: %v", err)
 	annos := map[string]string{
 		constants.MachineConfigDaemonStateAnnotationKey: constants.MachineConfigDaemonStateUnreconcilable,
+		constants.MachineConfigDaemonReasonAnnotationKey: err.Error(),
```
cgwalters: `setDone` should also clear this annotation right?

runcom: yup good point

runcom: wondering if setting the annotation to an empty string actually clears it out

test/e2e/mcd_test.go
```diff
@@ -309,12 +309,15 @@ func TestReconcileAfterBadMC(t *testing.T) {
 		}
 		for _, node := range nodes {
 			if node.Annotations[constants.DesiredMachineConfigAnnotationKey] == newMCName && node.Annotations[constants.MachineConfigDaemonStateAnnotationKey] != constants.MachineConfigDaemonStateDone {
-				return true, nil
+				// just check that we have the annotation here, w/o strings checking anything that can flip fast causing flakes
+				if node.Annotations[constants.MachineConfigDaemonReasonAnnotationKey] != "" {
```
kikisdeliveryservice: i stared at this and can't figure out what this check is doing here? help.

runcom: makes sure that a Reason for failure is set if this is the node that is in unreconcilable state (just an added check to verify the code in this PR)

runcom:
> At least in the case where there's only 1 unavailable node, we may want to also change the operator to roll this up into its status?

I thought about this and maybe we just need a Failing status in the MCP itself telling the most recent failures if we MCD keeps failing (and that can be then bubbled up to the operator itself)

runcom:
WIP cause I'm reworking this one to take into account https://bugzilla.redhat.com/show_bug.cgi?id=1699134 (which is gonna be simpler to fix in this PR with the changes being implemented here)

/hold

runcom:
@cgwalters @kikisdeliveryservice @LorbusChris ptal at this (last BZ for 4.1). I still have to take care of Colin's comment to clear out the degraded/unreconcilable annotation when flipping back to a good state but this should now fixes the BZ linked as in:

- if the render_controller fails (e.g. missing Ignition version), we now set the pool to Degraded
- if the pool is in Degraded, the operator will report Failing=True as well (given it's the master pool as we behave today anyway)
- the pool reports degraded with reasons from nodes now

pkg/daemon/writer.go
runcom: fixed

manifests/machineconfigpool.crd.yaml
```diff
@@ -16,6 +16,10 @@ spec:
     description: When at least one of machine is not either not updated or is in the process of updating to the desired machine config.
     name: Updating
     type: string
+  - JSONPath: .status.conditions[?(@.type=="Degraded")].status
+    description: When the update for one of the machine is not progressig or the pool can't be configured due to an error.
```
cgwalters: How about "When progress is blocked on updating one or more nodes, or the pool configuration is failing"?

pkg/controller/render/render_controller.go
```diff
@@ -433,7 +433,20 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
...
-	return ctrl.syncGeneratedMachineConfig(pool, mcs)
+	if err := ctrl.syncGeneratedMachineConfig(pool, mcs); err != nil {
+		return ctrl.syncFailingStatus(pool, err)
+	}
+
+	return nil
+}
+
+func (ctrl *Controller) syncFailingStatus(pool *mcfgv1.MachineConfigPool, err error) error {
+	sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to render MCs for MCP %s: %v", pool.Name, err), "")
```
cgwalters: Maybe just "config" and "pool" instead of "MC" and "MCP"?

runcom:
The pool now reports the reason for degrading/unreconciling w/o even bothering anymore to grab logs from MCDs

```
Status:
  Conditions:
    Last Transition Time:  2019-04-23T14:40:20Z
    Message:
    Reason:
    Status:                False
    Type:                  Updated
    Last Transition Time:  2019-04-23T14:40:20Z
    Message:
    Reason:                All nodes are updating to rendered-worker-14b59fdcfeca03d68b1f4ac6e0a3917f
    Status:                True
    Type:                  Updating
    Last Transition Time:  2019-04-23T14:40:25Z
    Message:               ,,,Node ip-10-0-143-208.us-west-1.compute.internal is reporting: "can't reconcile config rendered-worker-42e70505bc4c5e60668060ec30f5d3fc with rendered-worker-14b59fdcfeca03d68b1f4ac6e0a3917f: invalid Ignition config found: warning: File \"fdsalg\" references nonexistent filesystem \"wtft\". (This is ok if it is defined in a referenced config)\nerror: path not absolute: unreconcilable"
    Reason:                1 nodes are reporting degraded status on sync
    Status:                True
```

🎉

pkg/controller/node/status.go
```diff
@@ -44,12 +45,23 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	unavailableMachines := getUnavailableMachines(pool.Status.Configuration.Name, nodes)
 	unavailableMachineCount := int32(len(unavailableMachines))
 
+	degradedMachines := getDegradedMachines(pool.Status.Configuration.Name, nodes)
+	degradedReasons := make([]string, len(nodes))
+	for _, n := range degradedMachines {
+		reason, ok := n.Annotations[daemonconsts.MachineConfigDaemonReasonAnnotationKey]
+		if ok && reason != "" {
+			degradedReasons = append(degradedReasons, fmt.Sprintf("Node %s is reporting: %q", n.Name, reason))
```
cgwalters: This is OK as is though in a larger cluster this could become a lot of text if e.g. 50 worker nodes are hitting the same issue.

Most issues I think though we can catch on the controller side before we try to roll out to each node.

manifests/machineconfigpool.crd.yaml
runcom: 👍 thanks

runcom:
more exciting logs:
```
➜  installer git:(master) ✗ oc get machineconfigpools
NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED
master   rendered-master-12926114a5f51ebf79d5ef2aa0745405   True      False      False
worker   rendered-worker-dabe9018f3df61dc5982625c8957eea9   True      False      True
```

we flip degraded when we fail to render :)

and if we fail to render on the master pool, we also bubble up the error in the clusteroperator:

TODO: I have to fix the message tho 😄 

```
➜  installer git:(master) ✗ oc describe co/machine-config
Name:         machine-config
Namespace:
Labels:       <none>
Annotations:  <none>
API Version:  config.openshift.io/v1
Kind:         ClusterOperator
Metadata:
  Creation Timestamp:  2019-04-23T15:19:29Z
  Generation:          1
  Resource Version:    18890
  Self Link:           /apis/config.openshift.io/v1/clusteroperators/machine-config
  UID:                 304cf8ea-65db-11e9-9955-0626a76c833a
Spec:
Status:
  Conditions:
    Last Transition Time:  2019-04-23T15:39:54Z
    Message:               Cluster not available for 0.0.1-2019-04-23-145721
    Status:                False
    Last Transition Time:  2019-04-23T15:20:18Z
    Message:               Cluster version is 0.0.1-2019-04-23-145721
    Status:                False
    Last Transition Time:  2019-04-23T15:39:54Z
    Message:               Failed to resync 0.0.1-2019-04-23-145721 because: error pool master is not ready, retrying. Status: (total: 3, updated: 3, ready: 3, unavailable: 0)
    Reason:                error pool master is not ready, retrying. Status: (total: 3, updated: 3, ready: 3, unavailable: 0)
    Status:                True
    Type:                  Failing
  Extension:
    Master:  all 3 nodes are at latest configuration rendered-master-12926114a5f51ebf79d5ef2aa0745405
    Worker:  all 3 nodes are at latest configuration rendered-worker-dabe9018f3df61dc5982625c8957eea9
  Related Objects:
    Group:
    Name:      openshift-machine-config-operator
    Resource:  namespaces
  Versions:
    Name:     operator
    Version:  0.0.1-2019-04-23-145721
Events:       <none>
```

manifests/machineconfigpool.crd.yaml
runcom: fixed

pkg/controller/render/render_controller.go
runcom: fixed thanks

runcom:
/hold cancel

please all have another look at this

pkg/controller/node/status.go
kikisdeliveryservice: +1 on Colin's comment above. Couldn we just say X nodes report degraded and let user list them themselves?

pkg/controller/node/status.go
```diff
@@ -73,6 +85,20 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, fmt.Sprintf("All nodes are updating to %s", pool.Status.Configuration.Name), "")
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
 	}
+
+	if len(degradedMachines) > 0 {
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionTrue, fmt.Sprintf("%d nodes are reporting degraded status on sync", len(degradedMachines)), strings.Join(degradedReasons, ","))
```
kikisdeliveryservice: what does sdegraded mean?

kikisdeliveryservice: maybe rename?

runcom: status degraded, same as supdated, supdating above - just being consistent

pkg/controller/node/status.go
```diff
@@ -161,3 +187,26 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
...
 	return unavail
 }
+
+func getDegradedMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
+	var degraded []*corev1.Node
+	for _, node := range nodes {
+		if node.Annotations == nil {
+			continue
+		}
+		dconfig, ok := node.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey]
```
kikisdeliveryservice: daemon config? desiredconfig? can we make more explicit?

pkg/controller/node/status.go
```diff
@@ -161,3 +187,26 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
...
+		if !ok || dconfig == "" {
+			continue
+		}
+		dstate, ok := node.Annotations[daemonconsts.MachineConfigDaemonStateAnnotationKey]
```
kikisdeliveryservice: same

pkg/controller/node/status.go
```diff
@@ -161,3 +187,26 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
...
+		if !ok || dstate == "" {
+			continue
+		}
+
+		if dconfig == currentConfig &&
```
kikisdeliveryservice: same

pkg/controller/render/render_controller.go
```diff
@@ -433,7 +433,20 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
...
-	return ctrl.syncGeneratedMachineConfig(pool, mcs)
+	if err := ctrl.syncGeneratedMachineConfig(pool, mcs); err != nil {
+		return ctrl.syncFailingStatus(pool, err)
+	}
+
+	return nil
+}
+
+func (ctrl *Controller) syncFailingStatus(pool *mcfgv1.MachineConfigPool, err error) error {
+	sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to render configuration for pool %s: %v", pool.Name, err), "")
```
kikisdeliveryservice: same

pkg/controller/node/status.go
runcom: > maybe rename?

it looks sane and consistent with the others in this context, we're switching status here to (updating,updated,degraded). Longer would be `statusDegraded` or something like this. I'm ok with that but let's follow up as other needs changes as well (above) and doesn't belong to this PR

pkg/controller/render/render_controller.go
runcom: see https://github.com/openshift/machine-config-operator/pull/597#discussion_r277763131

pkg/operator/sync.go
```diff
@@ -321,15 +322,25 @@ func (optr *Operator) syncRequiredMachineConfigPools(config renderConfig) error
...
 		}
-		if pool.Generation <= pool.Status.ObservedGeneration && pool.Status.MachineCount == pool.Status.UpdatedMachineCount && pool.Status.UnavailableMachineCount == 0 {
+		degraded := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolDegraded)
+		if pool.Generation <= pool.Status.ObservedGeneration &&
+			isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated) &&
+			!degraded {
 			continue
 		}
-		return fmt.Errorf("error pool %s is not ready, retrying. Status: (total: %d, updated: %d, unavailable: %d)", pool.Name, pool.Status.MachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount)
+		return fmt.Errorf("error pool %s is not ready, retrying. Status: (total: %d, updated: %d, ready: %d, unavailable: %d, pool degraded: %v)", pool.Name, pool.Status.MachineCount, pool.Status.UpdatedMachineCount, pool.Status.ReadyMachineCount, pool.Status.UnavailableMachineCount, degraded)
```
kikisdeliveryservice: i think order should be ready/updated/unavailable bc ready is the final state we are looking for right?

runcom: yup, changing thanks!

kikisdeliveryservice: tho maybe leading with pool degraded would be clearer?

so pool degraded, total, ready updated, unavailable?

runcom: pool degraded, total, ready, updated, unavailable looks good 💪

pkg/controller/node/status.go
kikisdeliveryservice: yeah i had no clue that was supposed to mean status degraded. sounds good

kikisdeliveryservice:
did adding the new test cause the timeout? looking

runcom:
> did adding the new test cause the timeout? looking

yeah, cause I'm not clearing the Degraded=True which is basically node_controller stomping on template_controller for who sets that - I'm trying to figure that out

kikisdeliveryservice:
> > did adding the new test cause the timeout? looking
> 
> yeah, cause I'm not clearing the Degraded=True which is basically node_controller stomping on template_controller for who sets that - I'm trying to figure that out

gotcha! tho looking at the numbers we are starting to get close to the max time anyway (your test only ran for 5 minutes), so we'll likely have to extend timeout in the near future (since we have that one test that takes 36 minutes)

runcom:
> gotcha! tho looking at the numbers we are starting to get close to the max time anyway (your test only ran for 5 minutes), so we'll likely have to extend timeout in the near future (since we have that one test that takes 36 minutes)

yup 👍 

Anyway, so we do have this issue now:

NodeController wants to set/clear Degraded
RenderController wants to set/clear Degraded

There seems to be no good solution to each of the stomping on each other and flipping status is a no-go anyway. One solution I see is letting NodeController manage the MCP anyway but that requires the RenderController a CRD-like-ControllerConfig where it can report that it failed to render. If it failed to render, the NodeController would check that CRD to see if it needs to set/clear out Degraded on MCP.
Looks like a more elegant solution (if not the only one) and something we need to solve anyway as current code doesn't work as expected?

WDYT?

kikisdeliveryservice:
@runcom if the state is unreconcilable it is _also_ degraded?

runcom:
> @runcom if the state is unreconcilable it is _also_ degraded?

yeah, unreconcilable is a subset of degraded cause it's something in the config. Degraded is when things fail in a general way (transient errors as well)

kikisdeliveryservice:
if unreconcilable is a subset of degraded but now degraded has a reason, is there a point to unreconcilable? 

i thought that the idea of having both was to have them be independent error/failure states esp since we had no way of knowing _why_ something had been previously degraded and to avoid something small (like a small MC issue) causing a node to degrade since it was overkill?

runcom:
Lengthy talked with Abhinav who helped me figuring this out, I'm pushing out a change shortly but basically:

- introduce a NodeDegraded condition which flips to true when it does today (as usual w/o changing logic)
- introduce a RenderDegraded condition which flips to true when render fails
- NodeController makes the final call to set Degraded on the MCP (as it understands NodeDegraded & RenderDegraded = Degraded but we can change that in the future to a dedicated controller or control loop)

kikisdeliveryservice:
> Lengthy talked with Abhinav who helped me figuring this out, I'm pushing out a change shortly but basically:
> 
>     * introduce a NodeDegraded condition which flips to true when it does today (as usual w/o changing logic)
> 
>     * introduce a RenderDegraded condition which flips to true when render fails
> 
>     * NodeController makes the final call to set Degraded on the MCP (as it understands NodeDegraded & RenderDegraded = Degraded but we can change that in the future to a dedicated controller or control loop)

I think that this makes sense.

runcom:
> if unreconcilable is a subset of degraded but now degraded has a reason, is there a point to unreconcilable?

clarity between "something in kube fails" vs "something in the MC you provided" fails

> i thought that the idea of having both was to have them be independent error/failure states esp since we had no way of knowing _why_ something had been previously degraded and to avoid something small (like a small MC issue) causing a node to degrade since it was overkill?

Oh yeah, I'm still on board with that, but we're not yet on "MC issue -> no degrade", that's a long term aim that we need to tackle

pkg/controller/node/status.go
```diff
@@ -86,17 +86,26 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
...
+	}
+
+	// here we now set the MCP Degraded field, the node_controller is the one making the call right now
+	// but we might have a dedicated controller or control loop somewhere else that understands how to
+	// set Degraded. For now, the node_controller understand NodeDegraded & RenderDegraded = Degraded.
+	renderDegraded := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolRenderDegraded)
+	if nodeDegraded || renderDegraded {
+		sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, corev1.ConditionTrue, "", "")
+		mcfgv1.SetMachineConfigPoolCondition(&status, *sdegraded)
+	} else if !nodeDegraded && !renderDegraded {
```
cgwalters: This could just be `} else {` though if you end up respinning.

pkg/controller/render/render_controller.go
```diff
@@ -430,18 +430,27 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
...
+	}
 	return nil
 }
 
 func (ctrl *Controller) syncFailingStatus(pool *mcfgv1.MachineConfigPool, err error) error {
-	sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to render configuration for pool %s: %v", pool.Name, err), "")
+	sdegraded := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolRenderDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to render configuration for pool %s: %v", pool.Name, err), "")
 	mcfgv1.SetMachineConfigPoolCondition(&pool.Status, *sdegraded)
 	if _, updateErr := ctrl.client.MachineconfigurationV1().MachineConfigPools().UpdateStatus(pool); updateErr != nil {
 		glog.Errorf("Error updating MachineConfigPool %s: %v", pool.Name, updateErr)
```
kikisdeliveryservice: lower?

runcom: oh it's a log so we want logs to start uppercase (when it's `fmt.Errorf` or `errors.New` we want them lowercase cause we might end up bubbling up and concatenating)

kikisdeliveryservice: ah then transfer this comment over to all of the utilruntime ones like:
`utilruntime.HandleError(fmt.Errorf("Couldn't get object from tombstone %#v", obj))`

kikisdeliveryservice: will address this later :smile:

pkg/controller/node/status.go
runcom: fixed eventually

PR #600, opened 2019-04-03
add golangci-lint

### LINKED ISSUE #447: Add static checks to the project in PRs

runcom:

**- What I did**

Add https://github.com/golangci/golangci-lint to our verify target and fix all the current warnings to start fresh once this PR merges. The linter binary is already `go get`'ed with https://github.com/openshift/release/pull/3332

closes #447 

**- How to verify it**

CI


runcom:
/hold

since there a bunch of stuff to be fixed

runcom:
oh cool, verify it's working already :heart_eyes:

kikisdeliveryservice:
i love this!!

cgwalters:
Oh hum...we probably should have noticed the MCO was using `md5` earlier :cry: 
Not totally sure if we can change it now easily.

runcom:
Yeah, I'm not sure about even landing this for 4.1 due to the many stuff to take care of but still. Now we noticed and we can think about migrating/fixing some of this stuff.

runcom:
made another pass at this, changes are trivial though

runcom:
Removed WIP, some linters are disabled or this PR becomes a never ending PR. After this merges, I'll open an issue for contributors and us to re-enable and fix the remaining disabled linters.

Still on hold cause #603

ashcrow:
```
golangci-lint run
2019/04/04 09:20:21 unknown architecture wasm
```

runcom:
> golangci-lint run
> 2019/04/04 09:20:21 unknown architecture wasm
> pkg/controller/kubelet-config/kubelet_config_controller.go:335:67: `(*Controller).syncStatusOnly` - `err` is unused (unparam)
> func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.KubeletConfig, err error, args ...interface{}) error {

https://github.com/openshift/machine-config-operator/pull/600#issuecomment-479815299

runcom:
This requires #618 in order to be consistent

/hold

runcom:
uhm not sure why it's still using go1.10 despite https://github.com/openshift/release/pull/3413 being merged already 🤔

LorbusChris:
don't we need to wait for https://github.com/openshift/machine-config-operator/pull/618?

runcom:
> /test verify

this is not gonna pass as we aren't and we won't be at 1.11 for now

runcom:
depends on https://github.com/openshift/release/pull/3799

Makefile
```diff
@@ -50,6 +50,8 @@ update:
 #    make verify
 verify:
 	@which go-bindata 2> /dev/null >&1 || { echo "go-bindata must be installed to verify generated code";  exit 1; }
+	@which golangci-lint 2> /dev/null >&1 || { echo "golangci-lint must be installed to lint code";  exit 1; }
```
DanyC97: @runcom qq - is it bound to a particular release ?

DanyC97: i suspect it is by looking https://github.com/openshift/release/pull/3799/files so maybe the echo can be extended to mention that ... my 0.02$ though

runcom:
This is good to go now

Makefile
runcom: closed that PR as it's not needed now that we're all above go1.10 (1.12 to be precise)

ericavonb: How about adding a target to install it for you? e.g.

```
$(GOBIN)/golangci-lint: 
	go get -u github.com/golangci/golangci-lint/cmd/golangci-lint

verify: $(GOBIN)/golangci-lint
```

runcom: I've gona back and forth on the target (we used to have it in CRI-O as well but ended up being dropped cause the CI was taking care of it https://github.com/cri-o/cri-o/pull/2247/files#diff-b67911656ef5d18c4ae36cb6741b7965L308)

The real pain with go1.12 and onwards would be modules also (and calculating $GOBIN). I'm not yet familiar with that. I'm super fine following up to this and add targets to install tools also 👍

ericavonb: 👍 I guess I'm used to having a golang version check to handle compatibility. there were issues with GOBIN? never come across that myself.

runcom: > I guess I'm used to having a golang version check to handle compatibility.

oh cool, that might be something useful to have here in MCO as well :) 

> there were issues with GOBIN? never come across that myself.

afaict with go modules,  go build creates binaries in PWD so that must be conditionalized as well (hence your suggestion above to version check is amazing)

.golangci.yml
```diff
@@ -0,0 +1,109 @@
+---
+run:
+  concurrency: 6
+  deadline: 5m
+  skip-files:
+    - ".*_test\\.go"
```
sgreene570: we definitely want to exclude unit test files from linting? Just double checking.

runcom: yep, but as with the other "ignored" linters in this file, we might want, at some point, remove this and fix everything. Right now we need to get this in and fix code (not tests) to have this run and check on every PR w/o asking ppl to gofmt as you noticed.

sgreene570: ok sounds good

runcom:
rebased again 😄

PR #601, opened 2019-04-03
pkg/controller: resync on unavailable 

kikisdeliveryservice:
**- What I did**

- add check to syncStatusOnly to allow for resync of status when
an unavailable machine or unupdated machines are found.
- update unit tests to account for new error generated in syncStatusOnly 

Closes BZ 1695721

**- How to verify it**
CI runs should now pass without the error.

### DIFF
--- a/pkg/controller/node/node_controller_test.go
+++ b/pkg/controller/node/node_controller_test.go
@@ -784,7 +784,7 @@ func TestShouldMakeProgress(t *testing.T) {
 	expMcp.Status = expStatus
 	f.expectUpdateMachineConfigPoolStatus(expMcp)
 
-	f.run(getKey(mcp, t))
+	f.runExpectError(getKey(mcp, t))
 }
 
 func TestEmptyCurrentMachineConfig(t *testing.T) {
@@ -816,7 +816,7 @@ func TestPaused(t *testing.T) {
 	expMcp.Status = expStatus
 	f.expectUpdateMachineConfigPoolStatus(expMcp)
 
-	f.run(getKey(mcp, t))
+	f.runExpectError(getKey(mcp, t))
 }
 
 func TestShouldUpdateStatusOnlyUpdated(t *testing.T) {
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
a few unit tests are broken, need to update them.

kikisdeliveryservice:
`could not wait for build: the build machine-config-server failed after 1m21s with reason DockerBuildFailed: Docker build strategy has failed.`

not sure why, so retesting..
/retest

kikisdeliveryservice:
Ah so because of #575 I think tests are failing because of the way the unit tests are set up:
```
E0403 12:48:07.570477   17728 node_controller.go:177] error finding pools for node: node node-0 has both master role and custom role infra
E0403 12:48:07.772049   17728 node_controller.go:177] error finding pools for node: node node-0 belongs to 2 custom roles, cannot proceed with this Node
E0403 12:48:07.973102   17728 node_controller.go:177] error finding pools for node: node node-0 belongs to 2 custom roles, cannot proceed with this Node
--- FAIL: TestShouldMakeProgress (0.20s)
	/home/kg/go/src/github.com/openshift/machine-config-operator/pkg/controller/node/node_controller_test.go:116: error syncing machineconfigpool: resync needed: unavailable machines or un-updated machine present
--- FAIL: TestPaused (0.20s)
```

kikisdeliveryservice:
so unit tests were initially failing because it falled f.run() instead of f.runExpectedError() which is what we'd now expect given the existence of the new unavailable/not updated check initially.

pkg/controller/node/node_controller_test.go
```diff
@@ -816,7 +816,7 @@ func TestPaused(t *testing.T) {
 	expMcp.Status = expStatus
 	f.expectUpdateMachineConfigPoolStatus(expMcp)
 
-	f.run(getKey(mcp, t))
+	f.runExpectError(getKey(mcp, t))
```
runcom: should we now have a successful test where `f.run` just runs? (mocking machines len)

runcom:
/hold

after taking a closer look at https://bugzilla.redhat.com/show_bug.cgi?id=1695721, it looks like the MCO isn't causing any issue so I'm holding this until Trevor further clarifies why the MCO (apart from the transient) error is causing the upgrade to fail.

kikisdeliveryservice:
makes sense @runcom

runcom:
/hold cancel

not sure how we would ever get in such scenario where we don't resync anymore (and need this) but looks like we're getting reports that we can fall into this...

runcom:
/hold

I shouldn't have removed hold tho 😄

kikisdeliveryservice:

Feel free to reopen.

PR #603, opened 2019-04-04
kubelet-config: fix variable shadowing

runcom:
Otherwise, we were only adding nil as Conditions even when there was an
actual error...

@rphillips ptal, maybe this function should be split in 2 and unit tests added to catch stuff like this..

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -334,9 +334,9 @@ func (ctrl *Controller) generateOriginalKubeletConfig(role string) (*ignv2_2type
 
 func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.KubeletConfig, err error, args ...interface{}) error {
 	return retry.RetryOnConflict(updateBackoff, func() error {
-		newcfg, err := ctrl.mckLister.Get(cfg.Name)
-		if err != nil {
-			return err
+		newcfg, getErr := ctrl.mckLister.Get(cfg.Name)
+		if getErr != nil {
+			return getErr
 		}
 		newcfg.Status.Conditions = append(newcfg.Status.Conditions, wrapErrorWithCondition(err, args...))
 		_, lerr := ctrl.client.MachineconfigurationV1().KubeletConfigs().UpdateStatus(newcfg)


pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -334,9 +334,9 @@ func (ctrl *Controller) generateOriginalKubeletConfig(role string) (*ignv2_2type
...
 func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.KubeletConfig, err error, args ...interface{}) error {
 	return retry.RetryOnConflict(updateBackoff, func() error {
-		newcfg, err := ctrl.mckLister.Get(cfg.Name)
-		if err != nil {
-			return err
+		newcfg, getErr := ctrl.mckLister.Get(cfg.Name)
+		if getErr != nil {
+			return getErr
 		}
 		newcfg.Status.Conditions = append(newcfg.Status.Conditions, wrapErrorWithCondition(err, args...))
```
runcom: and maybe this needs to be wrapped in a `if err != nil` to avoid reporting nothing?

LorbusChris: wrapErrorWithCondition returns Success if error = nil (https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/helpers.go#L119).
So to me it looks like it'll always append a new condition here and we don't need to avoid this step,

LorbusChris:
looks like apiserver didn't come up in e2e

/test e2e-aws

pkg/controller/kubelet-config/kubelet_config_controller.go
LorbusChris: ah but looks like the return below ends up making the syncHandler return nil on no-error runs. Maybe we want to report an success/no errors reported msg here?

kikisdeliveryservice: does the err here need to be `getErr`?

kikisdeliveryservice:
I agree with @runcom that this function should be split and add that comments should be added to make this easier to understand at a glance.

rphillips:
/lgtm

I'll look into improving that status reporting.

PR #605, opened 2019-04-04
hack: switch diff to show positive diffing

runcom:
We were showing what's missing w/o make update. Makes more sense to
report the lines that make update would add actually.


kikisdeliveryservice:
I like this!
/lgtm

PR #608, opened 2019-04-08
Change the default crio.conf to use the values we want

cgwalters:
Rather than using `/etc/sysconfig/crio-network`, which was removed in
https://github.com/cri-o/cri-o/commit/cce25da96fb9e6c2d53744e49235b06b0a1c4537#diff-8f149e7773eb6fec135aa90b23aac30b


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1696774

dcbw:
/lgtm

Certainly seems cleaner that setting crio-network, if we're already writing a crio.conf somewhere.

kikisdeliveryservice:
sorry but what does the comment:
`Note this default is changed from base.` mean in this context?

cgwalters:
> sorry but what does the comment: Note this default is changed from base. mean in this context?

From the RPM, which is itself a copy of upstream I think.

kikisdeliveryservice:
@cgwalters could we put that in the comment then just so those of us not familiar will know?

cgwalters:
> /lgtm cancel

OK I addressed those two things.

kikisdeliveryservice:
thanks @cgwalters !!!!!!

/lgtm

runcom:
How has this regressed though? Why wasn't there any gating which could have found this

cgwalters:
(And now also changing the worker config)

PR #612, opened 2019-04-09
pkg/daemon: add a current config on disk check

runcom:

**- What I did**

Add an on disk fingerprint about what our current config is on a node (avoiding resync if we changed it or restored from an etcd backup). We check what we have on disk with what we have in annotations before triggering a sync.

**- How to verify it**

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -127,6 +128,9 @@ const (
 	pathDevNull = "/dev/null"
 	// pathStateJSON is where we store temporary state across config changes
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
+	// currentConfigPath is where we store the current config on disk to validate
+	// against annotations changes
+	currentConfigPath = "/var/machine-config-daemon/currentconfig"
 
 	kubeletHealthzEndpoint         = "http://localhost:10248/healthz"
 	kubeletHealthzPollingInterval  = time.Duration(30 * time.Second)
@@ -401,13 +405,13 @@ func (dn *Daemon) syncNode(key string) error {
 		// stash the current node being processed
 		dn.node = node
 		// Pass to the shared update prep method
-		needUpdate, err := dn.prepUpdateFromCluster()
+		current, desired, err := dn.prepUpdateFromCluster()
 		if err != nil {
 			glog.Infof("Unable to prep update: %s", err)
 			return err
 		}
-		if needUpdate {
-			if err := dn.triggerUpdateWithMachineConfig(nil, nil); err != nil {
+		if current != nil || desired != nil {
+			if err := dn.triggerUpdateWithMachineConfig(current, desired); err != nil {
 				glog.Infof("Unable to apply update: %s", err)
 				return err
 			}
@@ -822,6 +826,15 @@ func (dn *Daemon) CheckStateOnBoot() error {
 		state.currentConfig = state.pendingConfig
 	}
 
+	mcJSON, err := json.Marshal(state.currentConfig)
+	if err != nil {
+		return err
+	}
+	// write a file with current fingerprint overriding the previous
+	if err := writeFileAtomicallyWithDefaults(currentConfigPath, mcJSON); err != nil {
+		return err
+	}
+
 	inDesiredConfig := state.currentConfig == state.desiredConfig
 	if inDesiredConfig {
 		if state.pendingConfig != nil {
@@ -852,16 +865,16 @@ func (dn *Daemon) runOnceFromMachineConfig(machineConfig mcfgv1.MachineConfig, c
 			panic("running in onceFrom mode with a remote MachineConfig without a cluster")
 		}
 		// NOTE: This case expects a cluster to exists already.
-		needUpdate, err := dn.prepUpdateFromCluster()
+		current, desired, err := dn.prepUpdateFromCluster()
 		if err != nil {
 			dn.nodeWriter.SetDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
 			return err
 		}
-		if !needUpdate {
+		if current == nil || desired == nil {
 			return nil
 		}
 		// At this point we have verified we need to update
-		if err := dn.triggerUpdateWithMachineConfig(nil, &machineConfig); err != nil {
+		if err := dn.triggerUpdateWithMachineConfig(current, &machineConfig); err != nil {
 			dn.nodeWriter.SetDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -904,6 +911,20 @@ func (dn *Daemon) prepUpdateFromCluster() (bool, error) {
 		return false, err
 	}
 
+	currentFingerprint, err := ioutil.ReadFile(currentConfigFingerprint)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, errors.Wrapf(err, "loading current config fingerprint")
+		}
+		return false, nil
```
runcom: this is wrong (note to self)

pkg/daemon/daemon.go
```diff
@@ -124,6 +125,9 @@ const (
 	pathDevNull = "/dev/null"
 	// pathStateJSON is where we store temporary state across config changes
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
+	// currentConfigFingerprint is where we store the current config on disk to validate
+	// against annotations changes
+	currentConfigFingerprint = "/etc/machine-config-daemon/current"
```
cgwalters: This isn't a fingerprint, it's the whole config right?  I find this idea that we serialize all of the files back into JSON as a file in the filesystem kind of funny.  But I guess it's not a huge amount of data...

```
oc get -o yaml machineconfigs/rendered-worker-74e29c89eb37d7aaa30c33c481fae174 | gzip | wc -c
10197
```

We can certainly assume 10k of data is free.

On the kind-of bikeshed topic, I think `/var` is probably a better place for this.  The ostree model is that `/etc` is data humans *might* edit (e.g. "ssh to node and `vi`").

pkg/daemon/daemon.go
```diff
@@ -888,29 +901,50 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
...
+	if err != nil {
+		return nil, nil, err
 	}
 	state, err := getNodeAnnotation(dn.node, constants.MachineConfigDaemonStateAnnotationKey)
 	if err != nil {
-		return false, err
+		return nil, nil, err
+	}
+
+	mcJSON, err := os.Open(currentConfigFingerprint)
```
cgwalters: Don't we need to do this *before* we load `currentConfig`?

runcom: uhm, we may, I can't see why though, would you explain (I feel dumb)

cgwalters: If the current config isn't available in the apiserver, won't the `Get` above fail before we can find our cached version?

runcom: oh gosh yeah

runcom: No wait, the scenario here is that after a restore we still have something in the apiserver, which is really what we want on the machine itself, if current isn't on the apiserver, there's not much we can do anyway (which is the BZ we got from David yesterday but this PR isn't tackling that).

This is something I wrote to understand this better:
```
- current=configA == desired=configA
upgrade
- current=configA != desired=configB (desired being generated by upgrade and it's being applied as new current)
roll out new configs and MCD syncs
- current=configB == desired=configB

...

trigger a backup restore pre-upgrade which isn't calling informers
- current=configA == desired=configA
but now what's on disk is still configB

we now need something which retriggers the MCD to sync to configA
```

`configA` is what we need anyway, so it must be in the apiserver after a restore

kikisdeliveryservice: can i ask: when is a backup restore triggered? i feel like i understand what this pr is supposed to do, but not fully why?

runcom: there's work still ongoing on others to spec out how/when/where an etcd backup is performed (and a subsequent restore)

runcom: @cgwalters can you check my answer above? I think this is working as intended but I'm waiting on a way to test this out

pkg/daemon/daemon.go
runcom: fixed naming and moved under `/var/machine-config-daemon`

runcom:
This is good to review and get in - as it's treated as a bug. The testing goes later once the other teams/steps are complete

kikisdeliveryservice:
aws route errors:
/test e2e-aws-op

kikisdeliveryservice:
I  know that this is a bug, but I cant find a BZ/something else for it? Could we add Bug to title?

runcom:
> I know that this is a bug, but I cant find a BZ/something else for it? Could we add Bug to title?

there's no BZ for this, came out from a conversation with Derek on Slack :( not sure we need one tho

kikisdeliveryservice:
@runcom cool! just wanted to make sure I didn't miss it somewhere.

kikisdeliveryservice:
weird test didn't re-run

/retest

pkg/daemon/daemon.go
```diff
@@ -900,29 +913,50 @@ func (dn *Daemon) handleNodeUpdate(old, cur interface{}) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	defer mcJSON.Close()
+	fingerprintMC := &mcfgv1.MachineConfig{}
+	if err := json.NewDecoder(bufio.NewReader(mcJSON)).Decode(fingerprintMC); err != nil {
+		return nil, nil, err
+	}
+	if fingerprintMC.GetName() != currentConfig.GetName() {
+		return fingerprintMC, currentConfig, nil
```
cgwalters: So I read your comment a few times...I am finding it really hard to capture the flow/state of things in my head right now.

It feels like this would all be a *lot* simpler to reason about if basically we had `/var/machine-config-daemon/config<hash>.json` and we kept those for every config version that was relevant to us (current, desired) and used it if it was present instead of talking to the api server.

This would also help address the other BZ about having things be deleted from the apiserver.

That said...after thinking about this more I *think* you're right, it will work.  What I was worried about is "if we're saying currentConfig is desiredConfig, how do we then actually honor the real desiredConfig?".  But if they are different then we'll do a config transition and then on the next boot we should transition to the real desiredConfig?

cgwalters: OK here's another way to look at it - we're basically saying we can't trust our annotations to describe our current state, the real state is a file on disk.  So why are we looking at the annotation at all, versus setting it to the right thing if it's different from what's on disk?

runcom: > So why are we looking at the annotation at all, versus setting it to the right thing if it's different from what's on disk?

RIght, I had the same thought indeed, assuming we're talking about the currentConfig (desiredConfig has to come from somewhere/apiserver otherwise we can't really progress right?). So I guess a natural follow up to this would be to avoid talking to the apiserver when querying the currentConfig and just use what it's reflected on disk right?

runcom: > That said...after thinking about this more I _think_ you're right, it will work. What I was worried about is "if we're saying currentConfig is desiredConfig, how do we then actually honor the real desiredConfig?". But if they are different then we'll do a config transition and then on the next boot we should transition to the real desiredConfig?

right, that's my understand of what will happen if there's a real drift between currentOnAPI vs currentOnDisk.

Basically, the first sync would reconcile currentOnAPI with currentOnDisk, after we're done with that, there will be another immediate sync to honor desiredConfig (which, in our disaster recovery story is gonna be almost always the same as currentOnAPI, unless you backup an etcd with current != desired at the time you actually backup, right?)

runcom: > It feels like this would all be a _lot_ simpler to reason about if basically we had `/var/machine-config-daemon/config<hash>.json` and we kept those for every config version that was relevant to us (current, desired) and used it if it was present instead of talking to the api server.

I'm not sure how could we keep desired on disk 🤔 that has always have to come from apiserver right? I feel I'm missing something

cgwalters: I think my TL;DR though is this code is correct, happy to lgtm it if you want to land as-is; the rework above is optional.

cgwalters: > I'm not sure how could we keep desired on disk thinking that has always have to come from apiserver right? I feel I'm missing something

Yeah I edited to ~strikethrough~ that comment.

runcom: > I think my TL;DR though is this code is correct, happy to lgtm it if you want to land as-is; the rework above is optional.

This commit scares me somehow lol, something I'm going to change cause I Iike it, it's actually using a `current-<hash>.json` naming cause I find it best. Then yeah, I guess we can pull the trigger on this to get things flowing

cgwalters: One thing bugging me too is...why don't we

`return fingerprintMC, desiredConfig, nil` ?

kikisdeliveryservice: would make sense since below we return `currentConfig, desiredConfig, nil` in this case, we are using the fingerprintMC as current, we dont need the current I dont think?

kikisdeliveryservice: or are we saying that in _this_ case the currentConfig is the desired from the fingerprint? would it go from fingerprint->current-> desired? or....?

runcom: > fingerprint->current-> desired? or....?

exactly that but in the DR case, fingerprint=configB,current=configA,desired=configA. So it's just one real sync+reboot

> is there any case where if we have fingerprint, current and desired we would not want to ultimately end up in the desiredConfig?

nope, there shouldn't be any such case afaict. Desired is always evaluated at sync+1 if fingerprint and current drift

runcom: > One thing bugging me too is...why don't we
> 
> `return fingerprintMC, desiredConfig, nil` ?

Uhm, yeah, so using my flow from above comments, would that work in this case where I take a snapshot exactly in the middle of a sync where current != desired:

```
current=configA != desired=configB
snapshot

upgrade
- current=configB != desired=configC
roll out new configs and MCD syncs
- current=configC == desired=configC

...

trigger a backup restore pre-upgrade which isn't calling informers
- current=configA != desired=configB
but now what's on disk is still configC
```

should  we first go from configC on disk to configA on disk, and only then configB? You code would go straight from configC on disk to configB (which avoids a sync, but should we anyway?)

cgwalters: >  (which avoids a sync, but should we anyway or do we risk missing something?)

I am not sure what we'd miss.  It sounds like you're trying to be more conservative here just on general principle?  I don't object to that to be clear.  But I don't understand how this would be different from any other config change.

runcom: yup, it's really me being just conservative indeed, I guess it would be fine anyway to avoid a sync, so yeah, changing

runcom:
alrighty, let's see what tests say and then pull the trigger on this for the DR story

PR #620, opened 2019-04-10
make featuregate missing consistent with default

deads2k:
Defaults should be sane and missing feature gates isn't.


deads2k:
machines and nodes came up correctly.  I'm not going to force it yet because

`   "Pod openshift-multus/multus-bjxsf was pending entire time: unknown error",`

Issue #619, opened 2019-04-10
MachineConfigPool machineSelector is confusing


runcom:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1698250

MachineConfigPool matches a set of MachineConfig (machine.spec.machineConfigSelector) with a set of nodes (machine.spec.machineSelector).

machine.spec.machineSelector is confusingly named.

It should be machine.spec.nodeSelector.

We would provide a 2 days compatibility period, where we look for machineSelector and update nodeSelector, afterwards, we're just gonna drop machineSelector. This must be communicated via emails as well.

PR #625, opened 2019-04-10
Bug 1698250: * : pt1 transition from MachineSelector to NodeSelector labels

kikisdeliveryservice:
**- What I did**

MachineSelector will eventually be replaced with NodeSelector.
This PR allows all current MachineSelector values to be copied
into the NodeSelector label if MachineSelector is not nil, otherwise 
reads and uses NodeSelector value as part 1 of this process.

Closes: #619

**- How to verify it**

### DIFF
--- a/lib/resourcemerge/machineconfig.go
+++ b/lib/resourcemerge/machineconfig.go
@@ -41,6 +41,15 @@ func EnsureMachineConfigPool(modified *bool, existing *mcfgv1.MachineConfigPool,
 		*modified = true
 		existing.Spec.MachineSelector = required.Spec.MachineSelector
 	}
+
+	if existing.Spec.NodeSelector == nil {
+		*modified = true
+		existing.Spec.NodeSelector = required.Spec.NodeSelector
+	}
+	if !equality.Semantic.DeepEqual(existing.Spec.NodeSelector, required.Spec.NodeSelector) {
+		*modified = true
+		existing.Spec.NodeSelector = required.Spec.NodeSelector
+	}
 }
 
 func ensureMachineConfigSpec(modified *bool, existing *mcfgv1.MachineConfigSpec, required mcfgv1.MachineConfigSpec) {
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -265,9 +265,17 @@ func (ctrl *Controller) getPoolForNode(node *corev1.Node) (*mcfgv1.MachineConfig
 
 	var pools []*mcfgv1.MachineConfigPool
 	for _, p := range pl {
+		// first get MachineSelector
 		selector, err := metav1.LabelSelectorAsSelector(p.Spec.MachineSelector)
 		if err != nil {
 			return nil, fmt.Errorf("invalid label selector: %v", err)
+		}
+		// if MachineSelector is empty, get and use NodeSelector
+		if p.Spec.MachineSelector == nil {
+			selector, err = metav1.LabelSelectorAsSelector(p.Spec.NodeSelector)
+			if err != nil {
+				return nil, fmt.Errorf("invalid label selector: %v", err)
+			}
 		}
 
 		// If a pool with a nil or empty selector creeps in, it should match nothing, not everything.
@@ -424,8 +432,9 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	// TODO: Deep-copy only when needed.
 	pool := machineconfigpool.DeepCopy()
 	everything := metav1.LabelSelector{}
-	if reflect.DeepEqual(pool.Spec.MachineSelector, &everything) {
-		ctrl.eventRecorder.Eventf(pool, v1.EventTypeWarning, "SelectingAll", "This machineconfigpool is selecting all machines. A non-empty selector is required.")
+
+	if reflect.DeepEqual(pool.Spec.MachineSelector, &everything) && reflect.DeepEqual(pool.Spec.NodeSelector, &everything) {
+		ctrl.eventRecorder.Eventf(pool, v1.EventTypeWarning, "SelectingAll", "This machineconfigpool is selecting all nodes. A non-empty selector is required.")
 		return nil
 	}
 
@@ -439,6 +448,13 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	selector, err := metav1.LabelSelectorAsSelector(pool.Spec.MachineSelector)
 	if err != nil {
 		return err
+	}
+	// if MachineSelector is empty, get and use NodeSelector
+	if pool.Spec.MachineSelector == nil {
+		selector, err = metav1.LabelSelectorAsSelector(pool.Spec.NodeSelector)
+		if err != nil {
+			return fmt.Errorf("invalid label selector: %v", err)
+		}
 	}
 	nodes, err := ctrl.nodeLister.List(selector)
 	if err != nil {
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -16,6 +16,12 @@ func (ctrl *Controller) syncStatusOnly(pool *mcfgv1.MachineConfigPool) error {
 	if err != nil {
 		return err
 	}
+	if pool.Spec.MachineSelector == nil {
+		selector, err = metav1.LabelSelectorAsSelector(pool.Spec.NodeSelector)
+		if err != nil {
+			return err
+		}
+	}
 	nodes, err := ctrl.nodeLister.List(selector)
 	if err != nil {
 		return err
--- a/docs/MachineConfigController.md
+++ b/docs/MachineConfigController.md
@@ -39,7 +39,7 @@ type MachineConfigPoolSpec struct {
     MachineConfigSelector *metav1.LabelSelector `json:"machineConfigSelector,omitempty"`
 
     // Label selector for Machines.
-    MachineSelector *metav1.LabelSelector `json:"machineSelector,omitempty"`
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
 
     // If true, changes to this machine config pool should be stopped.
     // This includes generating new desiredMachineConfig and update of machines.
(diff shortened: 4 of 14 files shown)


kikisdeliveryservice:
@runcom is this what you had in mind and then in another PR MachineSelector will be replace throughout with NodeSelector?

runcom:
> @runcom is this what you had in mind and then in another PR MachineSelector will be replace throughout with NodeSelector?

yeah, I believe we're missing some points where we can get an MCP but we're still reading machineSelector e.g. under `pkg/controller/node/...` you can see we read from machineSelector).

So this pt1 should also be:

we now read from

1) machineSelector
1a) if machineSelector is not nil, we assign to nodeSelector and we keep using it (not machineSelector)
1b) if machineSelector is nil, we do nothing and we just read from nodeSelector (new version)

The above ensures that when this PR merges, both machineSelector & nodeSelector are read (and nodeSelector is used). machineSelector just becomes a mere proxy till we completely get rid of it.

Looks like we need to add code for nodeSelector in  `lib/resourcemerge/machineconfig.go` (right now this PR just assigns to nodeSelector, but we need code to _support_ nodeSelector as well)

kikisdeliveryservice:
Ahhhh. Thanks for clarifying @runcom !

runcom:
e.g. https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L268

That should read also from nodeController cause if we're dealing with a new MCP, machineSelector is gonna be empty

lib/resourcemerge/machineconfig.go
```diff
@@ -41,6 +41,18 @@ func EnsureMachineConfigPool(modified *bool, existing *mcfgv1.MachineConfigPool,
 		*modified = true
 		existing.Spec.MachineSelector = required.Spec.MachineSelector
 	}
+
+	if existing.Spec.MachineSelector != nil {
+		existing.Spec.NodeSelector = existing.Spec.MachineSelector
+	}
+	if existing.Spec.NodeSelector == nil {
```
kikisdeliveryservice: I also added these two `ifs` since the last push for NodeSelector because wouldn't we need it for new MCs?

pkg/controller/node/node_controller.go
```diff
@@ -265,10 +265,21 @@ func (ctrl *Controller) getPoolForNode(node *corev1.Node) (*mcfgv1.MachineConfig
...
 		if err != nil {
 			return nil, fmt.Errorf("invalid label selector: %v", err)
 		}
+		// if MachineSelector is empty, get and use NodeSelector
+		if selector == nil {
+			selector, err = metav1.LabelSelectorAsSelector(p.Spec.NodeSelector)
+			if err != nil {
+				return nil, fmt.Errorf("invalid label selector: %v", err)
+			}
+		} else {
```
kikisdeliveryservice: wait hmm do i need this else?

runcom: nope, as it's gonna be the initial machineSelector you grabbed at line 268/269

runcom: there are other spots that we need this 

https://github.com/openshift/machine-config-operator/blob/76232a00900850733d05b049d82683cbdb1401e0/pkg/controller/node/node_controller.go#L438

https://github.com/openshift/machine-config-operator/blob/76232a00900850733d05b049d82683cbdb1401e0/pkg/controller/node/node_controller.go#L450

lib/resourcemerge/machineconfig.go
runcom: yes this is correct

lib/resourcemerge/machineconfig.go
```diff
@@ -41,6 +41,20 @@ func EnsureMachineConfigPool(modified *bool, existing *mcfgv1.MachineConfigPool,
 		*modified = true
 		existing.Spec.MachineSelector = required.Spec.MachineSelector
 	}
+
+	if existing.Spec.MachineSelector != nil {
+		existing.Spec.NodeSelector = existing.Spec.MachineSelector
+		required.Spec.NodeSelector = required.Spec.MachineSelector
+
```
runcom: empty line

lib/resourcemerge/machineconfig.go
```diff
@@ -41,6 +41,20 @@ func EnsureMachineConfigPool(modified *bool, existing *mcfgv1.MachineConfigPool,
 		*modified = true
 		existing.Spec.MachineSelector = required.Spec.MachineSelector
 	}
+
+	if existing.Spec.MachineSelector != nil {
+		existing.Spec.NodeSelector = existing.Spec.MachineSelector
+		required.Spec.NodeSelector = required.Spec.MachineSelector
```
runcom: this line isn't required, as that's what's coming from `./manifests/...` and we drive that (to the new one already)

runcom: actiually, this whole if statement is not needed as it's only a new payload with nodeSelector when we merge this PR. We need to change the `machineSelector`in stuff like this indeed https://github.com/openshift/machine-config-operator/blob/76232a00900850733d05b049d82683cbdb1401e0/manifests/master.machineconfigpool.yaml#L11

pkg/controller/node/node_controller.go
kikisdeliveryservice: gotcha!

pkg/controller/node/node_controller.go
```diff
@@ -424,7 +432,8 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	// TODO: Deep-copy only when needed.
 	pool := machineconfigpool.DeepCopy()
 	everything := metav1.LabelSelector{}
-	if reflect.DeepEqual(pool.Spec.MachineSelector, &everything) {
+
+	if reflect.DeepEqual(pool.Spec.MachineSelector, &everything) && reflect.DeepEqual(pool.Spec.NodeSelector, &everything) {
```
runcom: we should still do the "if MachineSelector is empty, get and use NodeSelector" and only compare a selector which can be either (like you did in the other places)

runcom:
just one last comment and I think this can go in, also #619 can be closed with this already as it's fixing it so add a `close #619`in your first comment and prefix the PR title (not the commits) with `Bug 1698250`

/approve

pkg/controller/node/node_controller.go
kikisdeliveryservice: for this one i was a little confused, bc it's just checking to see if MachineSelector is empty, but we shouldnt' proceed into this if if Machine Selector is empty and NodeSelector is not.

pkg/controller/node/node_controller.go
```diff
@@ -424,7 +432,8 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
...
 		ctrl.eventRecorder.Eventf(pool, v1.EventTypeWarning, "SelectingAll", "This machineconfigpool is selecting all machines. A non-empty selector is required.")
```
runcom: can we change `all machines` to `all nodes` here as well?

pkg/controller/node/node_controller.go
kikisdeliveryservice: the nil check and use NodeSelector instead is below in starting at line 452

pkg/controller/node/node_controller.go
kikisdeliveryservice: sure!

pkg/controller/node/node_controller.go
runcom: but we can still do

selector
if machineselector empty
   selector = nodeSelector
if reflect.DeepEqual(selector) ...

for new mcp, machineSelector is gonna be empty and nodeSelector is still correctly checked. For old mcp, machineselector is empty and nodeSelector is going to be empty anyway and correctly checked right?

runcom: ok, your code is basically equivalent and better/cleaner/shorter :)

runcom:
Testing an upgrade to this from installer master before lgtm to be sure we're handling the transition correctly

/hold

runcom:
/retest 

(timeout on our tests)

pkg/controller/node/node_controller.go
```diff
@@ -265,10 +265,18 @@ func (ctrl *Controller) getPoolForNode(node *corev1.Node) (*mcfgv1.MachineConfig
 
 	var pools []*mcfgv1.MachineConfigPool
 	for _, p := range pl {
+		// first get MachineSelector
 		selector, err := metav1.LabelSelectorAsSelector(p.Spec.MachineSelector)
```
runcom: ok lol, I just realized that every call to `metav1.LabelSelectorAsSelector(` will return `label.Nothing()` which is obviously not nil in the check below (and elsewhere) and this is making our e2e to fail.

The right check should be:

```go
if p.Spec.MachineSelector == nil {
...
}
```

pkg/controller/node/node_controller.go
```diff
@@ -440,6 +449,13 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	if err != nil {
 		return err
 	}
+	// if MachineSelector is empty, get and use NodeSelector
+	if selector == nil {
```
runcom: this is gonna be always not nil, check https://github.com/openshift/machine-config-operator/pull/625/files#r274715843

pkg/controller/node/status.go
```diff
@@ -16,6 +16,12 @@ func (ctrl *Controller) syncStatusOnly(pool *mcfgv1.MachineConfigPool) error {
 	if err != nil {
 		return err
 	}
+	if selector == nil {
```
runcom: ditto https://github.com/openshift/machine-config-operator/pull/625/files#r274715843

pkg/controller/node/node_controller.go
runcom: if you're wondering the installation goes through because the initial node annotations are loaded from disk (laid down by the bootstrap process so there's no node controller involved which is failing here after cluster is up and running)

kikisdeliveryservice: ahhh ok

kikisdeliveryservice: done

pkg/controller/node/status.go
kikisdeliveryservice: done

pkg/controller/node/node_controller.go
kikisdeliveryservice: done

runcom:
let's schedule the machineSelector removal for uhm, Tuesday?

kikisdeliveryservice:
@runcom sounds good!

kikisdeliveryservice:
i'll get the next PR ready. :smile:

PR #626, opened 2019-04-10
try to make MCD render more self-contained

deads2k:
Trying to fix https://bugzilla.redhat.com/show_bug.cgi?id=1685704, we've bumped into MCD ghost dependencies to dead/broken resources.  In this case, it appears to be a dead configmap with ca-bundle.crt inside that cannot actually verify anything except a deprecated serving-cert.  However, we cannot fix the original here https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/manifests/configmap-initial-kube-apiserver-client-ca.yaml#L1 (which was actually only working by a invalid coincidence of a client-ca validating a serving cert).

To avoid the ghost, the MCD can render out its own new configmap and then rely upon that.  Keep in mind that this still leaves it disconnected from the actual trust bundle because the actual values aren't known until after rotation starts and runs.  But if you depend on the new configmap, you end up getting invalid machine configs that prevent a cluster from ever moving forward.  See https://bugzilla.redhat.com/show_bug.cgi?id=1698007

@sjenning if this doesn't work, hopefully it moves you much closer.

### DIFF
--- a/pkg/operator/render.go
+++ b/pkg/operator/render.go
@@ -22,6 +22,7 @@ type renderConfig struct {
 	ControllerConfig mcfgv1.ControllerConfigSpec
 	APIServerURL     string
 	Images           Images
+	KubeAPIServerServingCA string
 }
 
 func renderAsset(config renderConfig, path string) ([]byte, error) {
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -50,7 +50,7 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdCAFile, "etcd-ca", "/etc/ssl/etcd/ca.crt", "path to etcd CA certificate")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdMetricCAFile, "etcd-metric-ca", "/assets/tls/etcd-metric-ca-bundle.crt", "path to etcd metric CA certificate")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.rootCAFile, "root-ca", "/etc/ssl/kubernetes/ca.crt", "path to root CA certificate")
-	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.kubeCAFile, "kube-ca", "", "path to kube CA certificate")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.kubeCAFile, "kube-ca", "", "path to kube-apiserver serving-ca bundle")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.pullSecretFile, "pull-secret", "/assets/manifests/pull.json", "path to secret manifest that contains pull secret.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.destinationDir, "dest-dir", "", "The destination directory where MCO writes the manifests.")
 	bootstrapCmd.MarkFlagRequired("dest-dir")
--- a/manifests/machineconfigserver/kube-apiserver-serving-ca-configmap.yaml
+++ b/manifests/machineconfigserver/kube-apiserver-serving-ca-configmap.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: initial-kube-apiserver-server-ca
+  namespace: openshift-config
+data:
+  ca-bundle.crt: |
+{{.KubeAPIServerServingCA | indent 4}}
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -24,6 +24,7 @@
 // manifests/machineconfigserver/csr-bootstrap-role-binding.yaml
 // manifests/machineconfigserver/csr-renewal-role-binding.yaml
 // manifests/machineconfigserver/daemonset.yaml
+// manifests/machineconfigserver/kube-apiserver-serving-ca-configmap.yaml
 // manifests/machineconfigserver/node-bootstrapper-sa.yaml
 // manifests/machineconfigserver/node-bootstrapper-token.yaml
 // manifests/machineconfigserver/sa.yaml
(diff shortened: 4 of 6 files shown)


kikisdeliveryservice:
can we make this title/commit a little clearer about what the pr is doing exactly (explain the kubeapiserver cabundle) and add a link to the bugzilla.

kikisdeliveryservice:
Should we not just add this to the spec like the other things (see operator.go line 319 ) then you can pass `kubeAPIServerServingCABytes` via the the spec arg in getRenderConfig like everything else?
thoughts? @runcom @cgwalters

deads2k:
Might be easier to just break head for a couple hours than try to fix the MCD.  We just have no way to coordinate an MCD change.

cgwalters:
I haven't followed this much yet, sorry.  The cert chains are not an area I currently know much about.  Looks like what's rendered from this PR is broken?

```
Apr 10 20:55:21 ip-10-0-0-209 bootkube.sh[2295]: Assert creation failed: failed to load some manifests:
Apr 10 20:55:21 ip-10-0-0-209 bootkube.sh[2295]: "kube-apiserver-serving-ca-configmap.yaml": unable to convert asset "kube-apiserver-serving-ca-configmap.yaml" from YAML to JSON: yaml: line 10: could not find expected ':'
```

> To avoid the ghost, the MCD can render out its own new configmap and then rely upon that. Keep in mind that this still leaves it disconnected from the actual trust bundle because the actual values aren't known until after rotation starts and runs. 

The rotated certs live in a dynamic configmap, we're just using the on-disk version to bootstrap effectively?   Basically we have data that "transitions" from MCO ownership (host disk) to cluster?

deads2k:
The alternative to this pull is merging https://github.com/openshift/installer/pull/1593 and https://github.com/openshift/cluster-kube-apiserver-operator/pull/384 together because the MCD is trying to re-use another component's rendered asset.

deads2k:
> I haven't followed this much yet, sorry. The cert chains are not an area I currently know much about. Looks like what's rendered from this PR is broken?

I suspect an indention failure.  If I get it fixed, can we talk about landing without refactoring the usage in this operator?



> The rotated certs live in a dynamic configmap, we're just using the on-disk version to bootstrap effectively? Basically we have data that "transitions" from MCO ownership (host disk) to cluster?

No.  What's currently pointed to is completely dead (only kept in kube-apiserver-operator to keep the MCD "working").  The new thing that's point to by this pull is current and co-located with this operator to avoid the chicken and egg problem.  The new thing pointed to is never updated, just like the one today.

pkg/operator/render.go
```diff
@@ -32,6 +34,7 @@ func renderAsset(config renderConfig, path string) ([]byte, error) {
 
 	funcs := sprig.TxtFuncMap()
 	funcs["toYAML"] = toYAML
+	funcs["indent"] =    indent
```
runcom: indent is already defined as a built-in function (we use it in other places)

runcom:
> I might be missing something here but: should we not just add this to the ControllerConfigSpec like the other things (see operator.go line 319 ) then you can pass `kubeAPIServerServingCABytes` via the the spec arg in getRenderConfig like everything else? (you could also drop the change in render.go if it it did that)
> thoughts? @runcom @cgwalters

ControllerConfigSpec is for templates in `./templates` whether the whole renderConfig is for `./manifests`as well

runcom:
makes sense to me 👍 

/approve

runcom:
aws hiccups

runcom:
Failing with:

```

**Comment by deads2k** [comment]
I force merged the installer pull and the kas-o pull over failing CI to make progress.  this should still probably be done at some point, but I think I'm past this blocking 4.1 cert fixes.

**Comment by runcom** [comment]
> I force merged the installer pull and the kas-o pull over failing CI to make progress. this should still probably be done at some point, but I think I'm past this blocking 4.1 cert fixes.

I believe https://github.com/openshift/installer/pull/1593 has caused a drift between MCO and bootstrap as bootstrap is now generating the machineconfigs with kube-apiserver-complete-server-ca-bundle.crt but MCO (master) is still using the old kube-ca, right? (I've tested the installer master with 1593 and it fails, I'm reverting and retesting w/o)

**Comment by runcom** [comment]
Actually, looking at the installer failing job https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_installer/1593/pull-ci-openshift-installer-master-e2e-aws/5125

That has indeed caused a drift, so we definitely need this PR at least to generate the same MCs that the bootstrap generates

**Comment by deads2k** [comment]
The pair of pulls that I merged  (reference above) matches.  You can see this in current successful CI runs where the content of the configmap is indeed updated.  https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-config-operator/40/pull-ci-openshift-cluster-config-operator-master-e2e-aws/166/artifacts/e2e-aws/configmaps.json for instance

**Comment by runcom** [comment]
can we close this @deads2k

**Comment by deads2k** [comment]
> can we close this @deads2k

Funny story.  @sjenning is facing the exact same problem a second time.

**Comment by deads2k** [comment]
/retest

to get new images

**Comment by kikisdeliveryservice** [comment]
is this pr still going to go in? 

if so please: https://github.com/openshift/machine-config-operator/pull/626#issuecomment-481853358

**Comment by kikisdeliveryservice** [comment]
to avoid an erroneous lgtm-merge
/approve cancel

**Comment by kikisdeliveryservice** [comment]
since this PR seems dependent on https://github.com/openshift/installer/pull/1646

/hold

**Comment by deads2k** [comment]
w00t green

**Comment by deads2k** [comment]
> can we make this title/commit a little clearer about what the pr is doing exactly explain the kubeapiserver cabundle and that you are adding kubeAPIServerServingCA to the args for getRenderConfig in the MCO. Also add a link to the bugzilla.

I think all those things are already in the description.

This doesn't have to wait on the debugging pull

/hold cancel

**Comment by deads2k** [comment]
terraform
/retest

**Comment by runcom** [comment]
```
--- FAIL: TestMCDToken (0.52s)
	mcd_test.go:38: an error on the server ("unknown") has prevented the request from succeeding (get pods machine-config-daemon-mbhk4)
	mcd_test.go:38: an error on the server ("unknown") has prevented the request from succeeding (get pods machine-config-daemon-mlc87)
	mcd_test.go:38: an error on the server ("unknown") has prevented the request from succeeding (get pods machine-config-daemon-mppvg)
```

Not sure what's happening 🤔 digging

kikisdeliveryservice:
@runcom I saw an error in api server:
`watch chan error: etcdserver: mvcc: required revision has been compacted`

MCS logs seem normal

runcom:
> @runcom I saw an error in api server:
> `watch chan error: etcdserver: mvcc: required revision has been compacted`

awesome, makes total sense then! thanks!

PR #627, opened 2019-04-11
templates: Add a templates/common/, move pull secret there

cgwalters:
A lot of templates are duplicated between master/worker.  There's already
a `_base` for things not specific to a platform, so introduce `_base/_base`
which is templates that are shared between master/worker.

Start by moving the pull secret there.  If this PR is approved I'll
do all of the common files.


cgwalters:
This was bugging me for a while...doesn't quite pass tests but let me know if you guys like the approach.

kikisdeliveryservice:
@cgwalters is the idea to take the set of worker files that also exist in master and move to base?

generally, (without looking at the code that much)  i think i like the idea since it allows us to be explicit & consistent with the files that are exactly the same between the two.

as a note: you do have some file name issues  currently, a few end up being `templates/_base/_base...`

runcom:
> This was bugging me for a while...doesn't quite pass tests but let me know if you guys like the approach.

I like it

cgwalters:
Rebased :surfing_man: - this depends on https://github.com/openshift/machine-config-operator/pull/749 too now.

Keeping as WIP for feedback.

runcom:
needs a rebase to pick up #749

cgwalters:
:surfing_man: rebased!

DanyC97:
@cgwalters if you don't mind adding my thought, i think (maybe is a preference _thing_) would be a bit cleaner if we had `/templates/_base/{common|shared}/` instead of `templates/_base/_base`.

IMHO it much more natural thinking what it is w/o questioning the diff between `_base` and `_base/_base`

cgwalters:
Updated to use `templates/common/`

kikisdeliveryservice:
thanks for updating @cgwalters !!!

/LGTM

PR #628, opened 2019-04-11
docs: fix links to mco docs in README.md

kikisdeliveryservice:
After the README.md was moved to `/docs`, the links to other files in the `docs` folder are broken need to be updated.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,9 +26,9 @@ format.  Operating system updates use [rpm-ostree](http://github.com/projectatom
 This operator is split into 4 components; the above covers
 the operator.  Here are links to design docs for the sub-components:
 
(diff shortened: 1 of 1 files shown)


runcom:
but this is now breaking the README.md in the main page :) if you click on links here https://github.com/openshift/machine-config-operator/tree/f747a68e2b3138d14f88d00d74f34b73ce0c5b60#sub-components-and-design you can see it's broken

kikisdeliveryservice:
Ah, I see @runcom 

I had clicked on the file from the docs repo not from the main page.

runcom:
so, I believe that README in docs/ is just intended to be shown/used from the main repo page, but it's actually an issue if someone is under docs/ and clicks on the links which are not going to work (not sure what we can do though)

PR #632, opened 2019-04-12
Bug 1686117: pkg/controller: re-sync on pull secret change

runcom:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1686117

I've noticed, from code, that we don't trigger a resync when the
pull-secret changes. That leaves us with just a "oc delete pod
controller" to actually trigger a resync with a changed pull-secret.

Added an event handler for that particular secret to trigger a re-sync.

Also added a new doc explaining how to change the pull-secret (taking
care of the referenced BZ).

### DIFF
--- a/docs/PullSecret.md
+++ b/docs/PullSecret.md
@@ -1,0 +1,17 @@
+# Changing the payload pull secret
+
+The pull secret is used to pull the release payload image and it's currently stored as a secret in openshift-config/pull-secret.
+The ControllerConfig has an ObjectRefrence pointing to said secret and the Operator sets this up during its sync.
+The reference is then used by the TemplateController to grab the actual pull secret and generate the templates for MachineConfigs using it.
+That results in the pull secret being laid down on nodes, therefore nodes are now be able to pull the release payload image.
+
+The pull secret has an expiration, and when it expires you need to change it as follows:
+
+```
+oc edit secrets pull-secret -n openshift-config
+```
+
+Changing the `.dockerconfigjson` in the secret and saving it will result in the MCO syncing the new pull secret with the templates which then
+result in the new pull secret being laid down on nodes.
+
+Note that the `.dockerconfigjson` field is a base64 encoded JSON. 
--- a/pkg/controller/template/template_controller.go
+++ b/pkg/controller/template/template_controller.go
@@ -10,16 +10,18 @@ import (
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/lib/resourceapply"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/controller/common"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/scheme"
 	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
 	mcfglistersv1 "github.com/openshift/machine-config-operator/pkg/generated/listers/machineconfiguration.openshift.io/v1"
-	"k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
+	coreinformersv1 "k8s.io/client-go/informers/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	corev1clientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -64,6 +66,7 @@ func New(
 	templatesDir string,
 	ccInformer mcfginformersv1.ControllerConfigInformer,
 	mcInformer mcfginformersv1.MachineConfigInformer,
+	secretsInformer coreinformersv1.SecretInformer,
 	kubeClient clientset.Interface,
 	mcfgClient mcfgclientset.Interface,
 ) *Controller {
@@ -92,6 +95,12 @@ func New(
 		DeleteFunc: ctrl.deleteMachineConfig,
 	})
 
+	secretsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    ctrl.addSecret,
+		UpdateFunc: ctrl.updateSecret,
+		DeleteFunc: ctrl.deleteSecret,
+	})
+
 	ctrl.syncHandler = ctrl.syncControllerConfig
 	ctrl.enqueueControllerConfig = ctrl.enqueue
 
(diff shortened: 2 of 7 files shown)


runcom:
I'm having an hard time spinning up a new cluster to test this out, if someone could help me testing this I'd appreciate, I'm holding this for now.

/hold

@deads2k can you check this one which is related to the BZ you opened please

runcom:
wtf, no hold :smile: ?

runcom:
@abhinavdahiya @cgwalters @kikisdeliveryservice ptal

docs/PullSecret.md
```diff
@@ -0,0 +1,17 @@
+# Changing the payload pull secret
+
+The pull secret is used to pull the release payload image and it's currently stored as a secret in openshift-config/pull-secret.
+The ControllerConfig has an ObjectRefrence pointing to said secret and the Operator sets this up during its sync.
+The reference is then used by the TemplateController to grab the actual pull secret and generate the templates for MachineConfigs using it.
+That results in the pull secret being laid down on nodes, therefore nodes are now be able to pull the release payload image.
+
+The pull secret has an expiration, and when it expires you need to change it as follow:
```
kikisdeliveryservice: "follows"

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+func (ctrl *Controller) addSecret(obj interface{}) {
+	s := obj.(*v1.Secret)
+	if s.DeletionTimestamp != nil {
+		ctrl.deleteSecret(s)
+		return
+	}
+	if s.GetNamespace() == "openshift-config" && s.GetName() == "pull-secret" {
+		cfg, err := ctrl.ccLister.Get(common.ControllerConfigName)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("Couldn't get ControllerConfig on secret callback %#v", err))
```
kikisdeliveryservice: lowercase?

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
 	return ctrl
 }
 
+func (ctrl *Controller) addSecret(obj interface{}) {
+	s := obj.(*v1.Secret)
```
kikisdeliveryservice: could we make this var just called "secret" or something more descriptive instead of `s`

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+			utilruntime.HandleError(fmt.Errorf("Couldn't get ControllerConfig on secret callback %#v", err))
+			return
+		}
+		glog.V(4).Infof("Re-syncing ControllerConfig %s due to secret change", cfg.Name)
+		ctrl.enqueueControllerConfig(cfg)
+	}
+}
+
+func (ctrl *Controller) updateSecret(old, new interface{}) {
+	s := new.(*v1.Secret)
```
kikisdeliveryservice: same as above

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+
+	if s.GetNamespace() == "openshift-config" && s.GetName() == "pull-secret" {
+		cfg, err := ctrl.ccLister.Get(common.ControllerConfigName)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("Couldn't get ControllerConfig on secret callback %#v", err))
+			return
+		}
+		glog.V(4).Infof("Re-syncing ControllerConfig %s due to secret change", cfg.Name)
```
kikisdeliveryservice: do we want all of these logs in the v4 logs? Since this doesn't happen often I'm fine with it being in the main logs

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+		ctrl.enqueueControllerConfig(cfg)
+	}
+}
+
+func (ctrl *Controller) deleteSecret(obj interface{}) {
+	s, ok := obj.(*v1.Secret)
```
kikisdeliveryservice: same as above

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+			return
+		}
+		s, ok = tombstone.Obj.(*v1.Secret)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a MachineConfig %#v", obj))
+			return
+		}
+	}
+
+	if s.GetNamespace() == "openshift-config" && s.GetName() == "pull-secret" {
```
kikisdeliveryservice: since all 3 funcs seem to use the same if-if-block  maybe just move it out to have 1 as a func and call that?

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,65 @@ func New(
...
+
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Couldn't get object from tombstone %#v", obj))
+			return
+		}
+		s, ok = tombstone.Obj.(*v1.Secret)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a MachineConfig %#v", obj))
```
cgwalters: "Not a secret"

pkg/controller/template/template_controller.go
runcom: I'm fine changing this although I'll be more confident doing it in another PR and probably after 4.1 to avoid some weird logs flooding before we get there (we can discuss this in an issue maybe if we really want it for 4.1?)

kikisdeliveryservice: sure. i just figured since this is something user initiated, we'd want to expose some sort of log. all of the MC based updates for ex result in a log line exposed

pkg/controller/template/template_controller.go
runcom: done, though, for the delete action this is not the case as that's not enqueuing anything but I've added a TODO for later.

runcom:
taken care of comments thanks!

pkg/controller/template/template_controller.go
kikisdeliveryservice: nice!

kikisdeliveryservice:
all of the comments seem to be addressed. 

/lgtm

runcom:
flakes mainly
```
Flaky tests:

[Area:Networking] services basic functionality should allow connections to another pod on the same node via a service IP [Suite:openshift/conformance/parallel]
[Conformance][Area:Networking][Feature:Router] The HAProxy router should run even if it has no access to update status [Suite:openshift/conformance/parallel/minimal]
[Conformance][templates] templateinstance impersonation tests should pass impersonation update tests [Suite:openshift/conformance/parallel/minimal]
[Feature:Builds][pruning] prune builds based on settings in the buildconfig  should prune completed builds based on the successfulBuildsHistoryLimit setting [Suite:openshift/conformance/parallel]
[sig-cli] Kubectl client [k8s.io] Update Demo should do a rolling update of a replication controller  [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: emptydir] [Testpattern: Inline-volume (default fs)] subPath should support existing single file [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Pre-provisioned PV (filesystem volmode)] volumeMode should create sc, pod, pv, and pvc, read/write to the pv, and delete all created resources [Suite:openshift/conformance/parallel] [Suite:k8s]

Failing tests:

[sig-storage] In-tree Volumes [Driver: aws] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [Suite:openshift/conformance/parallel] [Suite:k8s]
```

/retest

runcom:
we need https://github.com/openshift/machine-config-operator/pull/640 in order to merge this now

runcom:
> New changes are detected. LGTM label has been removed.

rebased to bring the other PR, I'm now testing it to remove the hold

runcom:
alright, this worked just fine according to my testing, e.g. follow the docs I've added here, edited the secret and I can see 2 new rendered-<role> with the updated pull-secret

/hold cancel

runcom:
/hold

seeing something happening which I don't like but it doesn't look related to this anyway, I'm going to check

runcom:
:/

/retest

runcom:
Finally figured this out finally...

- adding a non-namespaced informer for kube secrets and filtering in the controller is just madness
- ...and it's madness cause all secrets are definitely more than the 50M memory limit we have in the controller deployment...
- ..which is mostly why installing a cluster works, but if you restart the controller, it crashes in loop cause the container is killed by the oom manager (typical exit code 137)

So, the solution would be to just have a namespaced infomer just for the `openshift-config` namespace (which we'll start using here for secrets but can be expanded in the future if we need more from that ns).

Let's see if it works, I'm testing it out locally.

kikisdeliveryservice:
@runcom great!!!!

runcom:
Should be ready to go now, I've tested locally and I can now restart the controller w/o it being oom'ed when it starts up again, let's see if tests complain

/hold cancel

docs/PullSecret.md
```diff
@@ -0,0 +1,17 @@
+# Changing the payload pull secret
```
kikisdeliveryservice: yay docs! :+1:

pkg/controller/template/template_controller.go
```diff
@@ -103,6 +112,59 @@ func New(
...
+func (ctrl *Controller) addSecret(obj interface{}) {
+	secret := obj.(*v1.Secret)
+	if secret.DeletionTimestamp != nil {
+		ctrl.deleteSecret(secret)
+		return
+	}
+	ctrl.filterSecret(secret)
+}
+
+func (ctrl *Controller) updateSecret(old, new interface{}) {
```
kikisdeliveryservice: updateSecret only seems to call filterSecret and only uses 1 of its args (just new not old)?

runcom: yup, we don't really care about `old` and we filter the correct one in `filterSecret` anyway so `old` can be discarded. We can't really use `filterSecret` in the event handler for the informer cause it's not type casted to secret automatically (which is what this method does eventually).

kikisdeliveryservice: gotcha on the filterSecret vs updateSecret part :+1:  but not understanding why old is an arg for updateSecret if we don't need it or have the logic? shouldn't unused args throw some kind of golang linting error? won't it do that in my vscode?

runcom: that's an interface for the event handler for update. The interface for the informer callbacks are as follows:
```
type ResourceEventHandlerFuncs struct {
	AddFunc    func(obj interface{})
	UpdateFunc func(oldObj, newObj interface{})
	DeleteFunc func(obj interface{})
}
```
You can see `UpdateFunc` has old and new (as an update action should have). `updateSecret` is using the `UpdateFunc` signature so we must have old and new.

Anyway, for linting golang doesn't care and vscode either (at least not mine). Golang has the don't care underscore for not used args/variable but that generally sucks, i.e. the updateSecret func becomes

```
func updateSecret(_, new interface{}) {
```

kikisdeliveryservice: thank you for explaining, i didn't quite get it the first two times around :smiley:

kikisdeliveryservice:
LGTM the two tiny fixes are good as well.

/lgtm

PR #613, opened 2019-04-09
Add etcd-quorum-guard manifests and doc

RobertKrawitz:
**- What I did**
Add `etcd-quorum-guard` manifests and documentation describing it.

**- How to verify it**
```oc get pods -n kube-system | grep etcd-quorum-guard```

**- Description for the changelog**
Add etcd-quorum-guard

PR #633, opened 2019-04-12
MCO: add etcd-quorum-guard image

runcom:
Needed to unblock https://github.com/openshift/installer/pull/1597

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -41,6 +41,7 @@ var (
 		setupEtcdEnvImage    string
 		infraImage           string
 		kubeClientAgentImage string
+		etcdQuorumGuardImage string
 		destinationDir       string
 	}
 )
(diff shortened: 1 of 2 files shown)


runcom:
Actually, re-following what we did for the InfraImage, this should go in with #613, merge that, then the installer PR is going to be merged, and then something like https://github.com/openshift/machine-config-operator/pull/518

omg, all of this is so silly

RobertKrawitz:
So should I merge this into #613?

runcom:
I believe so, and on top of this, you should also do something similar to this https://github.com/openshift/machine-config-operator/pull/508/files (I'm resyncing with Colin on this but this is what I would do following what we did with the infra image)

cmd/machine-config-operator/bootstrap.go
```diff
@@ -41,6 +41,7 @@ var (
 		setupEtcdEnvImage    string
 		infraImage           string
 		kubeClientAgentImage string
+		etcdQuorumGuardImage string
```
RobertKrawitz: Already in #613

cmd/machine-config-operator/bootstrap.go
```diff
@@ -70,6 +71,8 @@ func init() {
 	bootstrapCmd.MarkFlagRequired("kube-client-agent-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraImage, "infra-image", "", "Image for Infra Containers.")
 	bootstrapCmd.MarkFlagRequired("infra-image")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdQuorumGuardImage, "etcd-quorum-guard-image", "registry.svc.ci.openshift.org/openshift/origin-v4.0:base", "Image for etcd Quorum Guard.")
```
RobertKrawitz: Already in #613 (including marking it required).

install/image-references
```diff
@@ -27,6 +27,10 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-pod:v4.0
+  - name: etcd-quorum-guard-image
```
RobertKrawitz: Added just now to #613.

RobertKrawitz:
@runcom is this even needed any longer, as all of the changes are in #613?

runcom:
> @runcom is this even needed any longer, as all of the changes are in #613?

nope, this is closed since 4 days now, just tests rekicked for no good reason I guess

PR #640, opened 2019-04-17
render_controller: render on template MCs changes

runcom:
I was working on triggering a resync when the pull-secret change and I
got to this:

- pull-secret changes
- template_controller runs and update 00-<role> with the new pull-secret
- stop, the new pull secret is now just in 00-<role> MCs and not on
nodes since there's no rendered MC which contains it

Now, the above isn't quite right since after an admin updates the
pull-secret, it must be rolled out to nodes to be able to pull images
(for e.g. an upgrade).
What's happening is that the template_controller sets an OwnerReference
of Kind: ControllerConfig, whether the render_controller only reacts to
a Kind: MachineConfigPool (and MCs that aren't controller at all, like
the template_controller tho).

This patch fixes the controller checks and we now enqueue an MCP for the
given MC if we have a pool anyway.

### DIFF
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -15,7 +15,7 @@ import (
 	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
 	mcfglistersv1 "github.com/openshift/machine-config-operator/pkg/generated/listers/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/version"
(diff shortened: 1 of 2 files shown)


runcom:
/hold

I'm syncing with Abhinav on this, related to #632

runcom:
Synced with Abhinav, this is now ready to go and I've added a failing test to avoid regression.

kikisdeliveryservice:
just waiting until it goes green to lgtm

runcom:
> just waiting until it goes green to lgtm

I think it avoids a round-trip if you put it now anyway (assuming it lgt(y)). The PR won't be merged anyway if the tests aren't green.

Issue #641, opened 2019-04-17
Change the ClusterStatusConditionType "Failing" to "Degraded"


runcom:
Requires at least, in order:

- bump of openshift/api to master
- bump of openshift/client-go to master
- change in this repo

(for bumping, last time I've used `dep ensure -update github.com/openshift/api github.com/openshift/client-go` or something like this I guess)

There are still PRs in flight in those repo, so let's hold this till official go-ahead. I've  opened this for tracking. This is critical to release.

runcom:
This is ready to go and work 💯

PR #644, opened 2019-04-18
featuregates: Add default featuregates and comparison check

rphillips:
**- What I did**
Add check to skip machineconfig changes if featuregates are the same

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1699353
Fixes: https://github.com/openshift/cluster-config-operator/pull/39

**- How to verify it**

/cc @sjenning @deads2k

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_features.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features.go
@@ -103,6 +103,10 @@ func (ctrl *Controller) syncFeatureHandler(key string) error {
 		if err != nil {
 			return err
 		}
+		// Check to see if FeatureGates are equal
+		if reflect.DeepEqual(originalKubeConfig.FeatureGates, *featureGates) {
+			continue
+		}
 		// Merge in Feature Gates
 		err = mergo.Merge(&originalKubeConfig.FeatureGates, featureGates, mergo.WithOverride)
 		if err != nil {
--- a/pkg/controller/kubelet-config/kubelet_config_features_test.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features_test.go
@@ -1,13 +1,40 @@
 package kubeletconfig
 
 import (
+	"reflect"
 	"testing"
 
+	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
+	"github.com/vincent-petithory/dataurl"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/openshift/machine-config-operator/pkg/controller/common"
 )
+
+func TestFeatureGateDrift(t *testing.T) {
+	for _, platform := range []string{"aws", "none", "unrecognized"} {
+		t.Run(platform, func(t *testing.T) {
+			f := newFixture(t)
+			cc := newControllerConfig(common.ControllerConfigName, platform)
+			f.ccLister = append(f.ccLister, cc)
+
+			ctrl := f.newController()
+			kubeletConfig, err := ctrl.generateOriginalKubeletConfig("master")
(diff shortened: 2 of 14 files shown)


rphillips:
/hold cancel

Added unit test to confirm openshift/api FeatureGates match MCC templates.

rphillips:
OperatorFailing was removed from openshift/api... Figuring out the patch. https://github.com/openshift/api/commit/a9fb3b1629b23b3198c3d94017fc63430afcf6cc

runcom:
> OperatorFailing was removed from openshift/api... Figuring out the patch. [openshift/api@a9fb3b1](https://github.com/openshift/api/commit/a9fb3b1629b23b3198c3d94017fc63430afcf6cc)

we do have this open for that
https://github.com/openshift/machine-config-operator/issues/641

rphillips:
Thanks for the reference!

This PR will depend on #641 getting resolved to make sure all the featuregates are in sync.

deads2k:
/hold

https://github.com/openshift/cluster-kube-apiserver-operator/pull/423 need to test green .  I'm willing to try https://github.com/openshift/origin/pull/22611 but https://github.com/openshift/api/pull/292 is the backup plan

rphillips:
@runcom @deads2k can you give this PR a final review?

deads2k:
if they are the same, why is there a diff when they are serialized?  Do you just need a stable ordering in the serialization?

rphillips:
@deads2k We need to check if they are the same, which the original code did not do.

deads2k:
I'm ok with this, but without having a stable serialization, I worry that you'll end up chasing this kind of problem around.  I'll leave it with @runcom, but I see this as a risk.

runcom:
> @deads2k We need to check if they are the same, which the original code did not do.

is the risk of not having this PR that we re-compute this at every sync? hearing from @deads2k (since I'm not familiar with this code/feature) I would move this to 4.2 if it's just to skip a sync.

rphillips:
@runcom the main issue is that the default FeatureGate CR cannot be created. The FeatureGate controller within the MCC sees the new CR and attempts to apply the new MachineConfig. The new machineconfig causes a reboot cycle across the nodes. This is problematic on the install (bootstrapping) stage of the cluster.

This PR will see the new FeatureGate CR and notice it should not create a new MachineConfig; skipping the reboot cycle. I'm open to suggestions for another way to skip the MachineConfig creation, but noticing the FeatureGates are equal seems like the most prudent path forward (especially on initial bootstrap).

runcom:
> I'm ok with this, but without having a stable serialization, I worry that you'll end up chasing this kind of problem around. I'll leave it with @runcom, but I see this as a risk.

@deads2k can you clarify on the risk? the risk could be that we're going to reboot once more right?

rphillips:
```
level=error msg="\t* module.vpc.aws_route_table_association.private_routing[3]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.private_routing.3: timeout while waiting for state to become 'success' (timeout: 5m0s)"
```

/retest

sjenning:
/lgtm
@deads2k https://github.com/openshift/cluster-kube-apiserver-operator/pull/423 is merged
/hold cancel
@runcom can you review/approve?

deads2k:
> > I'm ok with this, but without having a stable serialization, I worry that you'll end up chasing this kind of problem around. I'll leave it with @runcom, but I see this as a risk.
> 
> @deads2k can you clarify on the risk? the risk could be that we're going to reboot once more right?

Without stable serialization in the MCO as a whole, any other change you're making and detecting can result in infinite looping or mismatched SHAs.  It's a generic problem, not a specific one.

I not worried about this pull in particular, only the pattern as a whole which this pull doesn't make any worse.

runcom:
Thanks David

/lgtm

PR #645, opened 2019-04-18
pkg/daemon: add MCD skeleton tests and test prepUpdateFromCluster

runcom:
Mimic (and copy from) pkg/controller testing, hopefully, one day I'll spend
a week trying to abstract all the duplicate code in pkg/controller
testings and now pkg/daemon...but not today.

This patch is adding tests for
https://github.com/openshift/machine-config-operator/pull/612 which
would greatly benefit from some testing.

This can be used as a start to add more tests *hint hint*


kikisdeliveryservice:
looks great!

/lgtm

PR #647, opened 2019-04-19
pkg/daemon: sync using current MC on disk if current on etcd not found

runcom:
With #612 we're now able to avoid transient errors or something really weird happening if we fail to grab the current config from the cluster. We can now just rely on what we have on disk and sync to desired.


runcom:
Uhm I don't like this actually even if it makes some sense

PR #649, opened 2019-04-19
start informers and wait for cache sync on informers

runcom:
It's not really syncing otherwise nor reacting...def a bug since syncs aren't even firing at this point.

This PR basically does the following:

- start informer for secrets
- add some v4 debugging which I found useful
- wait for cache syncs


runcom:
I'm dumb, there's a far worst issue actually, we're not starting most of the informers

runcom:
Ok, this should work as intended now, this is a bug

PR #652, opened 2019-04-22
add permissions to the cluster-reader role

deads2k:
Adds permissions to non-sensitive API resources for cluster-reader.

The general rule is that they can read all non-escalating resources, so not things like secrets, oauthtokens

### DIFF
--- a/install/0000_30_machine-config-operator-00_clusterreader_clusterrole.yaml
+++ b/install/0000_30_machine-config-operator-00_clusterreader_clusterrole.yaml
@@ -1,0 +1,19 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:openshift:machine-config-operator:cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+  - apiGroups:
+      - machineconfiguration.openshift.io
+    resources:
+      - containerruntimeconfigs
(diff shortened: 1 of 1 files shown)


install/0000_30_machine-config-operator-00_clusterreader_clusterrole.yaml
```diff
@@ -0,0 +1,19 @@
...
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+  - apiGroups:
+      - machineconfiguration.openshift.io
+    resources:
+      - containerruntimeconfigs
+      - controllerconfigs
+      - kubeletconfigs
+      - machineconfigpools
+      - mcoconfigs
```
cgwalters: The machineconfig objects can end up with secrets embedded though, like the pull secret.

runcom: That's not the machineconfigs tho. I believe at this point it's just a leftover from the early days of this operator and currently contains nothing

kikisdeliveryservice: are you both talking about the `mcoconfigs`?

cgwalters: Yep, this was discussed in Slack too I see now, and I was indeed thinking `mcoconfigs` was `machineconfigs`.

PR #653, opened 2019-04-23
templates: Remove empty platform directories

cgwalters:
So far we have not made any use of per-platform templated MCs.  Going
forward I don't really expect us to do that much, rather the conditional
code ends up in other components.  For the operating system, we
expect components to look at the `ignition.platform.id` kernel
command line argument.  For example, coreos/afterburn
has Azure specific code and it keys off this.

OpenShift components should be looking at the `install-config`, as
e.g. the machine-api-operator does.

The goal here is simply the templates; specifically this
cuts down on the combinatorial explosion of (master,worker) x platform.


runcom:
sounds reasonable to me

/approve

kikisdeliveryservice:
we are totally sure we dont need these? just double checking bc i keep seeing people add them and but have also never seen them used.

kikisdeliveryservice:
(I've always wondered what these were for)

kikisdeliveryservice:
to be held for 4.2 (cc: @runcom)
/hold

abhinavdahiya:
_to be fair didn't see the change in the PR_

it would be shortsightedness to say MCO will never add things that are for a specific platform.
So we should not completely close the path of platform specific MC fragments.

for example
- Azure internal load balancers black hole flows that have the same source and destination... (components on control plane host talking to api through load balancer black hole requests when the same control plane host is chosen as destination)
so we might at some point need to deploy some special configuration (scripts,units) to can prevent that.
- RHHI bare metal platform will require special static pods for infra setup.

cgwalters:
> it would be shortsightedness to say MCO will never add things that are for a specific platform.
> So we should not completely close the path of platform specific MC fragments.

Right, and to be clear this doesn't remove that functionality, just makes the platform directories not required.

cgwalters:
Rebased :surfing_man:
/hold cancel
/joke

cgwalters:
Following up to https://github.com/openshift/machine-config-operator/pull/653#issuecomment-485921342

I reworded the commit message and PR text to explicitly note that we're not removing functionality.  But let's clean things up for the common case and only create platform dirs where we really need it.

PR #655, opened 2019-04-23
etcd-member: exec the etcd binaries

squeed:
Save on the shell, and we want to be sure that termination signals are passed to the etcd process correctly.

Discovered this while looking through etcd connectivity issues. Probably won't fix the problem, but it's good practice regardless.


squeed:
cc @wking

hexfusion:
/lgtm

thanks, @squeed @wking @mfojtik

cgwalters:
Oh yes.  Missing `exec` calls are a pet peeve of mine.
/approve

PR #656, opened 2019-04-23
pkg/daemon: abstract NodeWriter to an interface

runcom:
Current units are broken on master on linux since the NodeWriter is nil
in fixtures (sorry I didn't catch that on osx).
Anyway, this patch finally abstract NodeWriter behing an interface so
it's easy to use it in tests as well. I've created a fakeNodeWriter
which can also be expanted in the future to make assertions if needed
(like counting actions and stuff like that).

cc @LorbusChris since you spotted this one

### DIFF
--- a/pkg/daemon/fake_writer.go
+++ b/pkg/daemon/fake_writer.go
@@ -1,0 +1,46 @@
+package daemon
+
+import (
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
+)
+
+// fakeNodeWriter is a fake single writer to Kubernetes to prevent race conditions
+type fakeNodeWriter struct {
+}
+
+// newFakeNodeWriter Create a new NodeWriter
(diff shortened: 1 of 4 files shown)


pkg/daemon/fake_writer.go
```diff
@@ -0,0 +1,46 @@
...
+// fakeNodeWriter is a fake single writer to Kubernetes to prevent race conditions
+type fakeNodeWriter struct {
+}
+
+// newFakeNodeWriter Create a new NodeWriter
+func newFakeNodeWriter() NodeWriter {
+	return &fakeNodeWriter{}
+}
+
+// Run reads from the writer channel and sets the node annotation. It will
```
LorbusChris: super tiny nit: Run**s** ... and sets .. 

also there is some capitalized **S**ets in the comments below which should be lower case.

runcom: Run is the function name (it's the subject in the sentence)

changed the uppercase->lowercase Sets

LorbusChris:
nice, wfm now!

/lgtm

PR #657, opened 2019-04-23
Drop os_version label on nodes

cgwalters:
It isn't useful today because most containers just want the RHEL major
which they can't get at without parsing it, and you can't do that
in a selector.

Also, it turns out that the label commandline is a one-shot.  We
need to move any label version updates into the MCD.


cgwalters:
Supercedes https://github.com/openshift/machine-config-operator/pull/584

kikisdeliveryservice:
can this go into 4.1 then if it's not required?

kikisdeliveryservice:
to be held for 4.2
/hold

cgwalters:
Rebased :surfing_man: 
/hold cancel
The next step is to move this fully into the MCD, but we might as well do this first.

PR #659, opened 2019-04-24
pkg/operator: fire an event when we change version

runcom:
Going to help with debugging, especially on roll out timings.

@deads2k ptal - notice that we fire this event only after we finish progressing to a new version.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -7,10 +7,12 @@ import (
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
 	cov1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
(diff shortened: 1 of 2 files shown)


pkg/operator/status.go
```diff
@@ -32,6 +34,18 @@ func (optr *Operator) syncVersion() error {
...
+	previousVersions := co.Status.Versions
+	newVersions := optr.vStore.GetAll()
+
+	if !optr.vStore.Equal(previousVersions) {
+		mco := &corev1.ObjectReference{
+			Kind: co.Kind,
+			Name: co.Name,
+			UID:  types.UID(co.GetUID()),
+		}
+		optr.eventRecorder.Eventf(mco, corev1.EventTypeNormal, "OperatorVersionChanged", fmt.Sprintf("clusteroperator/machine-config-operator version changed from %v to %v", previousVersions, newVersions))
```
deads2k: Let's say we have an upgrade from A to B.  I think the timeline looks like this:

1. operator updated to B
2. operator starts rolling out operand to B
3. operator finishes rolling out operand to B
4. operator updates status.version to indicate B

I'm interested in a marker event for 1 and 4 so that we can track how long the set of steps took.  Which step does this fire for?

runcom: Alrighty, can do

pkg/operator/status.go
```diff
@@ -86,13 +97,25 @@ func (optr *Operator) syncProgressingStatus() error {
 			Status:  configv1.ConditionFalse,
 			Message: fmt.Sprintf("Cluster version is %s", optrVersion),
 		}
+		mcoObjectRef = &corev1.ObjectReference{
+			Kind: co.Kind,
+			Name: co.Name,
```
deads2k: you need a namespace or your will emit into `default`

pkg/operator/status.go
```diff
@@ -32,6 +34,15 @@ func (optr *Operator) syncVersion() error {
 		return nil
 	}
 
+	if !optr.vStore.Equal(co.Status.Versions) {
+		mcoObjectRef := &corev1.ObjectReference{
+			Kind: co.Kind,
+			Name: co.Name,
```
deads2k: you need a namespace or your will emit into `default`

deads2k:
minor (but important) comment on namespace.

/lgtm
/hold

holding for namespace update

kikisdeliveryservice:
timeouts and apiserver errors let's try again.

/retest

runcom:
This is updated and tests are passing as well. @deads2k ptal

/hold cancel

runcom:
For reference, this is how and event introduced here looks like:
```
{
apiVersion: "v1",
count: 2,
eventTime: null,
firstTimestamp: "2019-04-24T17:36:18Z",
involvedObject: {
name: "machine-config",
uid: "77df1bc0-66b7-11e9-b9f0-1271eefad662"
},
kind: "Event",
lastTimestamp: "2019-04-24T17:37:29Z",
message: "clusteroperator/machine-config-operator is bootstrapping to [{operator 0.0.1-2019-04-24-171909}]",
metadata: {
creationTimestamp: "2019-04-24T17:36:18Z",
name: "machine-config.1598799b3d97f71a",
namespace: "default",
resourceVersion: "4647",
selfLink: "/api/v1/namespaces/default/events/machine-config.1598799b3d97f71a",
uid: "77f63f25-66b7-11e9-b9f0-1271eefad662"
},
reason: "OperatorVersionChanged",
reportingComponent: "",
reportingInstance: "",
source: {
component: "machineconfigoperator"
},
type: "Normal"
},
```
we don't have upgrade events, but they will pop up after this merges in upgrade jobs

kikisdeliveryservice:
@runcom neat!

kikisdeliveryservice:
EBS messes..

/retest

runcom:
e2e-aws passed but still shows Pending 🤷‍♂️

runcom:
maybe...

/retest

PR #663, opened 2019-04-25
Bug 1702832: pkg/operator: poll during master MCP sync

runcom:
The master MCP we sync in the operator can flip Degraded=True too early
if it's still rolling out. This patch introduces back a 10 minutes
polling to give it some time to correctly roll. We also now bubble up
any error to the CO's Extension field along with a full MCPs status. The
net result of this patch is that we only flip Degraded=True if: 1)
10m are up and we do have a real error preventing roll out 2) 10m are up
and we have transient errors slowing down the roll out. In both case we
report what's happening at every iteration of the sync during the 10
mins poll w/o flipping Degraded=True (as most errors are transient
anyway). If we get a permanent error during sync, we still wait 10 mins
before reporting Degraded but the error can be seen immediately in the
clusteroperator status extension.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/pkg/errors"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -318,10 +319,6 @@ func (optr *Operator) syncRequiredMachineConfigPools(config renderConfig) error 
 	if err != nil {
 		return err
 	}
-	pools, err := optr.mcpLister.List(sel)
-	if err != nil {
-		return err
-	}
 	isPoolStatusConditionTrue := func(pool *mcfgv1.MachineConfigPool, conditionType mcfgv1.MachineConfigPoolConditionType) bool {
 		for _, condition := range pool.Status.Conditions {
 			if condition.Type == conditionType {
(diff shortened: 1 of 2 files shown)


runcom:
let's see what e2e-aws-upgrade has to say about this also

runcom:
There are some gotchas with this PR tho, most notably, 10minutes timeout has been chosen by what was there previously but if today we have a 100 nodes master pool, 10m aren't obviously enough (and we never really tested a performance run on such a huge pool I guess).

cgwalters:
> today we have a 100 nodes master pool

We only support 3 masters today, I don't think the etcd upstream suggests anything higher than 7 https://coreos.com/etcd/docs/latest/v2/admin_guide.html

cgwalters:
OK the commit message keeps talking about "master" as if it's special cased but...we didn't special case it in any way before this PR, and we still aren't with it right?

cgwalters:
(And we know we need to apply OS updates to the master pool first and we're not doing that yet)

runcom:
> OK the commit message keeps talking about "master" as if it's special cased but...we didn't special case it in any way before this PR, and we still aren't with it right?

the syncRequiredConfigPool only takes care of masters (since the master pool has a special annotation) so we only care about master effectively

cgwalters:
So...now we end up blocking in `sync()` for the pools; I had thought the operator was responsible for e.g. regenerating the `controllerconfig` but...oh yes it is, it templates it from the renderconfig looks like.  So if...hm, I guess we don't care about immediately reflecting anything in the render config.

runcom:
> So...now we end up blocking in `sync()` for the pools

this is what we were doing in the past as well but we failed at providing why we were still syncing when we got some transient/permanent error. The change today is introducing that back but bubbling up any error which is what we wanted at first removing the polling.

cgwalters:
> the syncRequiredConfigPool only takes care of masters (since the master pool has a special annotation) so we only care about master effectively

Ahh...yes I see that now.  Wait but hm that makes me wonder about how we're rolling up any failures in the worker pool up to the operator.  Anyways, that doesn't need to block this PR.  LGTM from me, will let Kirsten do any other review and take final merge.

runcom:
> Ahh...yes I see that now. Wait but hm that makes me wonder about how we're rolling up any failures in the worker pool up to the operator.

so, we bubble the error from workers in the ClusterOperator.Status.Extension field but we don't set Degraded but yeah, at some point maybe we'll need something like that for worke

runcom:
repushing to trigger another e2e-aws-upgrade job

pkg/operator/sync.go
```diff
@@ -330,17 +327,51 @@ func (optr *Operator) syncRequiredMachineConfigPools(config renderConfig) error
 		}
 		return false
 	}
-	for _, pool := range pools {
-		if err := isMachineConfigPoolConfigurationValid(pool, version.Version.String(), optr.mcLister.Get); err != nil {
-			return fmt.Errorf("pool %s has not progressed to latest configuration: %v, retrying", pool.Name, err)
+
+	var lastErr error
+	if err := wait.Poll(time.Second, 10*time.Minute, func() (bool, error) {
```
kikisdeliveryservice: is 10 too much/too little?

runcom: looks like it's fine according to Derek so I'd say it's good enough (and what was there before I removed it some months ago)

wking: Cross-ref https://github.com/openshift/machine-config-operator/pull/554/files#diff-fa45321336db7ad1cedc28bf643a4f97L326

kikisdeliveryservice: gotcha and thanks @wking !

kikisdeliveryservice:
is this ready or do you need to test more?

runcom:
> is this ready or do you need to test more?

waiting to hear from upgrade job

runcom:
second upgrade job in a row passing, let me repush again to trigger another and then we can merge

kikisdeliveryservice:
sounds good!

runcom:
green

PR #664, opened 2019-04-25
templates/master/00-master: add 127.0.0.1 to etcd metric proxy SAN.

hexfusion:
etcd-quorum-guard needs the ability to pin to the host etcd instance to check health. Because we are now using the metric chain of trust we need to allow 127.0.0.1 in the SAN the same as etcd server currently does to simplify these TLS connections.

Assists #613


kikisdeliveryservice:
@hexfusion I thought 613 was closed?

hexfusion:
> @hexfusion I thought 613 was closed?

see that don't pay attention for 5 mins :).

hexfusion:
@kikisdeliveryservice actually we still want quorum guard standalone to use metric certs so I would like to keep this open if possible.

RobertKrawitz:
@hexfusion @kikisdeliveryservice this could also be used to allow the standalone etcd-quorum-guard to use the metrics cert.

kikisdeliveryservice:
I think there are some wider spread aws issues right now, so retesting might not help.

RobertKrawitz:
@hexfusion is there any guarantee that the proxy won't cache health status (which would defeat the whole purpose of the quorum guard)?

hexfusion:
@RobertKrawitz the proxy has no caching mechanism it has a handler that makes an HTTP request. On the curl side we could pass `-H 'Cache-Control: no-cache'` as a general safety mechanism for this concern.

hexfusion:
We are not going to use metric certs for this use-case anymore closing

PR #665, opened 2019-04-25
*: add resource sync for etcd-quorum-guard

hexfusion:
This PR adds functionality which will sync ConfigMaps and Secrets from `openshift-config` namespace to `openshift-etcd`. The result will be etcd TLS resources available for etcd-quorum-guard to consume.

Note: although etcd-quorum-guard is solo it still requires the same assets.


RobertKrawitz:
See comment https://github.com/openshift/machine-config-operator/pull/664#issuecomment-486854312

hexfusion:
NLN

PR #672, opened 2019-04-26
controller/template: put the cloudconf for kubelet behind platform gate

abhinavdahiya:
Previously, the assumption was that all platforms that had CloudProviderConfig set can have kubelet configured to
use the config file. But this created a regression for user-provided-infrastructure VSphere case, as everybody like kube-controller-manager etc. except kubelet can use the cloud conf
file from API for vSphere

/cc @staebler

### DIFF
--- a/pkg/controller/template/test_data/templates/master/01-master-kubelet/vsphere/units/kubelet.service
+++ b/pkg/controller/template/test_data/templates/master/01-master-kubelet/vsphere/units/kubelet.service
@@ -25,7 +25,7 @@ contents: |
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider=vsphere \
         --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
-        --cloud-config=/etc/kubernetes/cloud.conf \
+         \
         --anonymous-auth=false \
         --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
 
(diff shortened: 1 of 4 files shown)


runcom:
@kikisdeliveryservice can you also review this and put your lgtm to merge if it's good?

pkg/controller/template/test_data/templates/master/01-master-kubelet/vsphere/units/kubelet.service
```diff
@@ -25,7 +25,7 @@ contents: |
         --client-ca-file=/etc/kubernetes/ca.crt \
         --cloud-provider=vsphere \
         --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
-        --cloud-config=/etc/kubernetes/cloud.conf \
+         \
```
LorbusChris: let's remove the line entirely? same in the worker service below

LorbusChris:
I don't want to be a killjoy and put a `hold` on it at this stage, but I have a nit :smile:

pkg/controller/template/test_data/templates/master/01-master-kubelet/vsphere/units/kubelet.service
abhinavdahiya: currently this is status for kubelet service files in master, and it would require changing the templates to allow skipping this file. Would it be okay to take it as a follow up to update the template to do this for all platforms?

abhinavdahiya: EDIT: also I want this PR's scope confined to changing the render function behavior

kikisdeliveryservice: no worries @abhinavdahiya 

we can get this PR in and make whatever small changes as a followup.

PR #671, opened 2019-04-26
Add etcd-quorum-guard to the machine config operator

RobertKrawitz:
**- What I did**
Add etcd-quorum-guard to the machine config operator

**- How to verify it**
Until the e2e is added, look at the log files in the artifacts.

**- Description for the changelog**
Add guard to prevent more than one etcd member from being taken down at any one time

PR #673, opened 2019-04-26
Add etcd-quota-guard e2e and remove debugging code

RobertKrawitz:
**- What I did**
Add e2e for etcd-quota-guard, and remove debugging startup code from the pod

Followon to #671.

**- How to verify it**
Run e2e

Add e2e for etcd-quota-guard, and remove debugging startup code from the pod

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,4 +98,4 @@ images.rhel7: $(imc7)
 
 # This was copied from https://github.com/openshift/cluster-image-registry-operato
 test-e2e:
-	GOCACHE=off go test -timeout 50m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
+	GOCACHE=off go test -timeout 55m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -63,16 +63,9 @@ spec:
         - mountPath: /mnt/kube
           name: kubecerts
         command:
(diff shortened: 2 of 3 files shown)


kikisdeliveryservice:
will probably have to raise the e2e timeout for this?  let's see, but we're pretty close to the 50 mins already.

RobertKrawitz:
@kikisdeliveryservice this test is fairly quick, but where is that limit so I can make the change if I need to?

kikisdeliveryservice:
@RobertKrawitz i believe it's set in the Makefile

RobertKrawitz:
Yep.  I'll see what happens; if it times out, I'll raise it to 55m.

RobertKrawitz:
/hold
I want to see why the test failed the first time.  It shouldn't have failed in that way.

runcom:
looks like e2e went fine no?

hexfusion:
Was this the concern?

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/673/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/1767/

RobertKrawitz:
@hexfusion yes.  I don't like that it failed once in that way and then passed.

RobertKrawitz:
/hold cancel

Unable to reproduce this after about 5 tries.  I've bumped the timeout in question to 5 minutes in case that was the issue.

RobertKrawitz:
Also bumped the e2e timeout to 55 minutes per @kikisdeliveryservice recommendation.

RobertKrawitz:
/retest
infra failure

RobertKrawitz:
@runcom see comments above; I believe this is ready to merge.

PR #675, opened 2019-04-27
Bug 1703703 *: add etcd configuration file

hexfusion:
Currently, etcd uses a static configuration with defaults. While this is sane for testing the reality is these configs need to be adjusted depending on scale, environment etc. I generated the etcd.conf in this way vs having it managed by MCO to eliminate restarting of masters after the change was made.

Although MCO would make sure all etcd had the same config I think this is still something the human operator should do via ssh-basion. This would reduce the possibility of a bad configuration taking down the cluster on reboot.

IMO we can’t ship without this as it would force people to update etcd-member.yaml which would eventually restart nodes unnecessarily.

The variables defined in the conf in my experience are the most commonly redefined in tuning. The human operator can extend.

resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1703703

/hold


hexfusion:
CI results show this is working.

```
2019-04-27 17:06:27.777995 I | pkg/flags: recognized and used environment variable ETCD_DATA_DIR=/var/lib/etcd
2019-04-27 17:06:27.778108 I | pkg/flags: recognized and used environment variable ETCD_DEBUG=false
2019-04-27 17:06:27.778119 I | pkg/flags: recognized and used environment variable ETCD_ELECTION_TIMEOUT=1000
2019-04-27 17:06:27.778130 I | pkg/flags: recognized and used environment variable ETCD_HEARTBEAT_INTERVAL=100
2019-04-27 17:06:27.778145 I | pkg/flags: recognized and used environment variable ETCD_NAME=etcd-member-ip-10-0-143-28.ec2.internal
2019-04-27 17:06:27.778162 I | pkg/flags: recognized and used environment variable ETCD_QUOTA_BACKEND_BYTES=4294967296
2019-04-27 17:06:27.778171 I | pkg/flags: recognized and used environment variable ETCD_SNAPSHOT_COUNT=100000
```

cgwalters:
> But by doing so because this config is monitored by MCO will result in eventually restarting the node.

Is only kind of true today, we don't actually monitor any files actively.  It would probably make sense to do so.

> Although MCO would make sure all etcd had the same config I think this is still something the human operator should do via ssh-basion.

I always reach for `oc debug node` over ssh first btw.

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -86,6 +87,10 @@ contents:
 
           source /run/etcd/environment
 
+          set -a
+          source /run/etcd/etcd.conf
```
cgwalters: I find the choice of `/run` here to be odd - are you explicitly choosing to have this config file wiped if the node reboots for another reason?  Shouldn't it be in `/etc/etcd/`¹?

In general a pattern I think we should take is to have the MCD generate/own "base config" files, and allow the admin to use "drop-ins" - this pattern is [supported by systemd](https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html) for example.

It's pretty easy to implement in shell script which you already have here, just make it:

```
if -d /etc/etcd.d; then
  for x in /etc/etcd.d/*.conf; do
    source $x
  done
fi
```

¹ I put an etcd in your etc so you can config while you config

hexfusion: > I find the choice of /run here to be odd - are you explicitly choosing to have this config file wiped if the node reboots for another reason?

Nice catch!! No this was an oversight, to be honest.

>> But by doing so because this config is monitored by MCO will result in eventually restarting the node.

>Is only kind of true today, we don't actually monitor any files actively. It would probably make sense to do so.

So if I lay down the conf with MCO and it changes on disk it will no cause restart or it will?

>  I put an etcd in your etc so you can config while you config

:) while I like your train of thought emulating systemd (did I say that?) I would really like a simple config setup that is familiar to 3.x devs. These configs should not change very often except during tuning or disaster? In this case I think a single config that a user does not worry about inheritance or  would be optimal. No matter what we say or do folks will adjust the MCO base config. Then send me a BZ because the base config was overridden by foobar.conf. We could iterate on that in the future?

Single config WYSIWYG is my vote but the floor is open.

hexfusion: @abhinavdahiya I am curious about your feelings on this as well.

kikisdeliveryservice: > So if I lay down the conf with MCO and it changes on disk it will no cause restart or it will?

MCO doesn't watch the files on disk to see if they are being changed by non-config/non-MCO means.

hexfusion: Ok that is not what I was understanding at all. If I adjust `/etc/kubernetes/manifests/etcd-member.yaml` my understanding is MCO will roll that out to the other nodes and eventually restart.

abhinavdahiya: IMO writing the default configuration for etcd using setup-etcd-env is not the best move forawrd, mostly as MC* already writes and manages config files for crio/kubelet both of which technically do not need the reboot to pickup the changes but still we go through the reboot cycle.
on the other hand we would need to restart the etcd pod to pick up new configuration. And we should use MC to roll those changes as that would give us a way to not roll invalid configuration to all the etcd nodes with quorum guard in place.
So I would move towards uniformity and hopefully when / if MCD gets do-not-reboot on certain we can piggyback.

on the single file vs systemd style, i'm leaning towards single file similar to kubelet.conf. dropins are not useful for this usecase.

i'm +1 on stetting up our etcd static pods to consume configurations from a file (I'm okay on any location)

hexfusion:
> release "release-latest" failed: pod release-latest was already deleted

/test e2e-aws

templates/master/00-master/_base/files/etc-etcd-etcd-conf.yaml
```diff
@@ -0,0 +1,29 @@
...
+mode: 0644
+path: "/etc/etcd/etcd.conf"
+contents:
+  inline: |
+    #[member]
+    ETCD_SNAPSHOT_COUNT=100000
+    ETCD_HEARTBEAT_INTERVAL=100
+    ETCD_ELECTION_TIMEOUT=1000
+
+    #[cluster]
```
abhinavdahiya: nit this should not exist here.

templates/master/00-master/_base/files/etc-etcd-etcd-conf.yaml
```diff
@@ -0,0 +1,29 @@
...
+    #[logging]
+    ETCD_DEBUG=false
+
+    #[profiling]
+    ETCD_ENABLE_PPROF=false
+
+    #[validation]
+    ETCD_STRICT_RECONFIG_CHECK=true
+
+    #[restore]
```
abhinavdahiya: nit this should not exist here

templates/master/00-master/_base/files/etc-etcd-etcd-conf.yaml
hexfusion: After talking about this I am fine dropping these. In general, I am only explicitly defining defaults. But to eliminate possible confusion I have removed.

templates/master/00-master/_base/files/etc-etcd-etcd-conf.yaml
hexfusion: same as above

hexfusion:
To be clear all values besides `ETCD_QUOTA_BACKEND_BYTES` are etcd defaults and what we currently use.

`ETCD_QUOTA_BACKEND_BYTES` default is 2GB and is too small for general purposes. 4GB is the default for 3.x and is a good starting point.

hexfusion:
```
fail [github.com/openshift/origin/test/extended/router/scoped.go:58]: Expected error:
    <*util.ExitError | 0xc002925b30>: {
        Cmd: "oc new-app --config=/tmp/admin.kubeconfig --namespace=e2e-test-router-scoped-z6dbq -f /tmp/fixture-testdata-dir599304041/test/extended/testdata/router-common.yaml",
        StdErr: "--> Deploying template \"e2e-test-router-scoped-z6dbq/\" for \"/tmp/fixture-testdata-dir599304041/test/extended/testdata/router-common.yaml\" to project e2e-test-router-scoped-z6dbq\n\n--> Creating resources ...\n    rolebinding.authorization.openshift.io \"system-router\" created\n    route.route.openshift.io \"route-1\" created\n    route.route.openshift.io \"route-2\" created\n    route.route.openshift.io \"route-override-domain-1\" created\n    route.route.openshift.io \"route-override-domain-2\" created\n    service \"endpoints\" created\n    error: pods \"endpoint-1\" is forbidden: no providers available to validate pod request\n--> Failed",
        ExitError: {
            ProcessState: {
                pid: 1588,
                status: 256,
                rusage: {
                    Utime: {Sec: 0, Usec: 301207},
                    Stime: {Sec: 0, Usec: 78313},
                    Maxrss: 88576,
                    Ixrss: 0,
                    Idrss: 0,
                    Isrss: 0,
                    Minflt: 14198,
                    Majflt: 0,
                    Nswap: 0,
                    Inblock: 0,
                    Oublock: 0,
                    Msgsnd: 0,
                    Msgrcv: 0,
                    Nsignals: 0,
                    Nvcsw: 1241,
                    Nivcsw: 56,
                },
            },
            Stderr: nil,
        },
    }
    exit status 1
not to have occurred
```

/test e2e-aws

kikisdeliveryservice:
For posterity, we're hitting a problem in `e2e-aws-op` across the board right now, so this is going to keep failing until we get a fix in.

hexfusion:
test was still 55m timeout  so I guess it did not have what we needed yet

/test e2e-aws-op

kikisdeliveryservice:
@hexfusion a fix is not in yet

hexfusion:
I thought Clayton merged 685

kikisdeliveryservice:
yes but that isn't the fix for the e2e-aws-op bugs, that ci job is still going to fail until we likely get 697 (we hope) merged

runcom:
please rebase to pick up the CI fix

hexfusion:
yeeSSSS!!!!!

PR #677, opened 2019-04-27
Remove reference to etcd-quorum-guard image, and use cli instead

smarterclayton:
The quorum guard image does not exist in OCP and agreement was not
to create a new one. Use the CLI image which is required to have
bash. Simplify the container name to be easier to input when debugging.


smarterclayton:
/assign @runcom 

OCP nightly builds are broken because this took a reference to an image that we didn't want to create (which was the problem in the first place).  This references an image that already exists correctly.

wking:
/lgtm

but I'm not a maintainer.

Issue #679, opened 2019-04-29
MCD unable to "just" enable a systemd unit


JAORMX:
**Description**

When providing a minimal systemd unit configuration that just "enabled" a built-in unit, the MCD is not able to start it.

**Steps to reproduce the issue:**
1. create a minimal MachineConfig object. e.g.
```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: worker
  name: 60-worker-enable-auditd
spec:
  config:
    ignition:
      version: 2.2.0
    systemd:
      units:
      - name: auditd.service
        enabled: true
```
2. apply the configuration

**Describe the results you received:**
The service wasn't started.

**Describe the results you expected:**
`systemctl status auditd` should have shown the service as started.

**Additional information you deem important (e.g. issue happens only occasionally):**
```
  machine-config-controller                     https://github.com/openshift/machine-config-operator                       0ae25d4cd189125c981aee9bdb6d00a48ca3b978
  machine-config-daemon                         https://github.com/openshift/machine-config-operator                       0ae25d4cd189125c981aee9bdb6d00a48ca3b978
  machine-config-operator                       https://github.com/openshift/machine-config-operator                       0ae25d4cd189125c981aee9bdb6d00a48ca3b978
  machine-config-server                         https://github.com/openshift/machine-config-operator                       0ae25d4cd189125c981aee9bdb6d00a48ca3b978
  setup-etcd-environment                        https://github.com/openshift/machine-config-operator                       0ae25d4cd189125c981aee9bdb6d00a48ca3b978
```

**Additional environment details (platform, options, etc.):**
Cluster version is 4.1.0-0.okd-2019-04-29-081740

PR #680, opened 2019-04-29
mcd: Allow enabling units even if no content was provided

JAORMX:
Closes: #679

**- What I did**
when dealing with systemd units, the MCD used to check the "content"
section. If no content was provided, it would skip the unit. This
doesn't work for built-in units, which we could merely enable without
writing anything. This allows that.

**- How to verify it**

Allows enabling built-in systemd units without providing the "content" section.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -541,12 +541,6 @@ func (dn *Daemon) writeUnits(units []ignv2_2types.Unit) error {
 			glog.V(2).Infof("Wrote systemd unit dropin at %s", dpath)
 		}
 
-		if u.Contents == "" {
-			continue
-		}
-
-		glog.Infof("Writing systemd unit %q", u.Name)
-
 		fpath := filepath.Join(pathSystemd, u.Name)
 
 		// check if the unit is masked. if it is, we write a symlink to
@@ -566,12 +560,16 @@ func (dn *Daemon) writeUnits(units []ignv2_2types.Unit) error {
 			continue
 		}
 
-		// write the unit to disk
-		if err := writeFileAtomicallyWithDefaults(fpath, []byte(u.Contents)); err != nil {
-			return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
-		}
+		if u.Contents != "" {
+			glog.Infof("Writing systemd unit %q", u.Name)
 
-		glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+			// write the unit to disk
+			if err := writeFileAtomicallyWithDefaults(fpath, []byte(u.Contents)); err != nil {
+				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
+			}
+
+			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		}
 
 		// if the unit doesn't note if it should be enabled or disabled then
 		// skip all linking.


cgwalters:
/approve

Looks reasonable offhand; some definite risk vs reward balancing here though as far as shipping this for 4.1.

runcom:
def 4.2 :)

runcom:
also, is there a workaround for this or is it just not working?

pkg/daemon/update.go
```diff
@@ -541,17 +541,13 @@ func (dn *Daemon) writeUnits(units []ignv2_2types.Unit) error {
...
-		}
-
-		glog.Infof("Writing systemd unit %q", u.Name)
-
 		fpath := filepath.Join(pathSystemd, u.Name)
 
 		// check if the unit is masked. if it is, we write a symlink to
 		// /dev/null and continue
 		if u.Mask {
+			glog.Infof("Writing systemd unit %q", u.Name)
```
kikisdeliveryservice: If you want a message here, shouldn't this message be something like "Symlinking.."

JAORMX: Not something I wanted really; just left the logic as it was before (It should print exactly what it was printing before).

kikisdeliveryservice: That logging line was moved into the `if u.Mask` block by this PR. Previously it was on line 548 before the `if u.Mask.` It's placement at line 549 is inaccurate and it now appears twice in the function. The same logging on line 566 is correct. You can either change the logging on line 549 as suggested above or remove it entirely.

JAORMX: I'll remove it.

JAORMX: Done

kikisdeliveryservice: Awesome, thanks!

JAORMX:
> Can you add the `Closes: #679` to the commit message Also prepend the commit and title of this PR with `mcd: Allow enabling...`

@kikisdeliveryservice done.

Thanks for the review!

runcom:
> also, is there a workaround for this or is it just not working?

maybe this has been overlooked, but I would be interested into knowing that 💪

JAORMX:
> > also, is there a workaround for this or is it just not working?
> 
> maybe this has been overlooked, but I would be interested into knowing that muscle

I actually don't know :D I just didn't manage to get it working.

runcom:
> @runcom I actually don't know :D I just didn't manage to get it working.

uhm ok, I believe this work isn't impacting 4.1 as there won't be anyone trying to enable auditd right? this is more of a 4.2 stuff but at that point, auditd might be already enabled by default in rhcos (like chronyd) right?

JAORMX:
@runcom actually, IIUC there are no plans to enable auditd by default in RHCOS

runcom:
> @runcom actually, IIUC there are no plans to enable auditd by default in RHCOS

then this is a bug actually.

JAORMX:
@runcom thanks for looking into it

pkg/daemon/update.go
```diff
@@ -566,12 +560,16 @@ func (dn *Daemon) writeUnits(units []ignv2_2types.Unit) error {
...
+
+			// write the unit to disk
+			if err := writeFileAtomicallyWithDefaults(fpath, []byte(u.Contents)); err != nil {
+				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
+			}
 
-		glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		}
```
runcom: I believe the change in behavior here is that, after this lands, if someone tries to enable a non-existant systemd unit, this is now going to crashloop.

JAORMX: I thought it would just issue a warning. Let me try that out and arrange a fix for that issue.

runcom: I believe this is fine tho, it's what systemctl does when the unit doesn't exist actually. Let me think one more minute

runcom:
We chatted about this and we believe this is good to merge

/lgtm

cgwalters:
(Ignore this, using this PR as a random test space for something else)

/test e2e-aws-scaleup

PR #682, opened 2019-04-29
WIP: test/osupdate: New test

cgwalters:
This is a new end-to-end test that:

 - Builds a new oscontainer with test content in /usr/share
 - Builds a new release image that derives from the existing
   one (so in theory we're not testing upgrades of other things)
 - Targets the cluster at that release image
 - Waits for the cluster to update
 - Verifies that the test content made it

Not hooked up into CI yet.  Tentatively thinking of creating a new context (only run for merges by default) for this, but we could also glom it onto our existing e2e-aws-op.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,4 +104,5 @@ images.rhel7: $(imc7)
 
 # This was copied from https://github.com/openshift/cluster-image-registry-operato
 test-e2e:
+	./test/osupdate/test-osupdate.sh
 	go test -timeout 75m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
--- a/test/osupdate/mco-os-bc.yaml
+++ b/test/osupdate/mco-os-bc.yaml
@@ -1,0 +1,32 @@
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
+  labels:
+    app: mco-os
+  name: mco-os
+spec:
+  lookupPolicy:
+    local: true
+  output:
+    to:
+      kind: ImageStreamTag
+      name: mco-os:latest
+  postCommit: {}
+  runPolicy: Serial
+  source:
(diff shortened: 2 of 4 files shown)


cgwalters:
Hm, in this I want to follow a pod's logs while also monitoring its status for completion and propagate the exit code.  I could write some custom code for that (I guess `oc start-build -F` is probably similar) but if anyone knows a convenient  `kubectl/oc` CLI incantation for this that'd be nice!

cgwalters:
Rebased this to see if the bug in `oc adm release new` is fixed now.

cgwalters:
```
error: the release could not be reproduced from its inputs
```

Yeah, this still hits the regression introduced by https://github.com/openshift/origin/commit/15292e028d2ab954ffb23b35e75b573e258f2c77

cgwalters:
> error: the release could not be reproduced from its inputs

So I consulted my "What would ci-operator do?" bracelet and noticed that it was saying `--to-image-base` *and* `--to-image` - and indeed passing both as the same arg seems to fix this.  I think it may be something like without that, the other images still try to pull from the original location or something?

Anyways, let's see how this works.

We still need to figure out whether this test is run as part of `e2e-aws-op` as is, if it's a separate context, or even one we only run manually, etc.

runcom:
> We still need to figure out whether this test is run as part of `e2e-aws-op` as is, if it's a separate context, or even one we only run manually, etc.

I'm leaning towards setting up a new job just for this - it's an extra cluster, I know, but still something we might want to offload from e2e-aws-op. Setting up a new job just requrires a new Makefile target and some files in openshift/release also. What do you think?

runcom:
looks like a network glitch now

kikisdeliveryservice:

Feel free to reopen.

PR #685, opened 2019-04-30
Bug 1704706: remove openshift-sdn and multus CNI configuration files on boot

squeed:
Discovered in BZ1704706 that we weren't actually cleaning these files
up, leading to failed deletions.

Fixes: rhbz1704706

### DIFF
--- a/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-tmpfiles.d-cleanup-cni.conf
+++ b/pkg/controller/template/test_data/templates/master/00-master/aws/files/-etc-tmpfiles.d-cleanup-cni.conf
@@ -1,5 +1,5 @@
 contents:
-  source: data:,r%20%2Fetc%2Forigin%2Fopenvswitch%2Fconf.db%0Ar%20%2Fetc%2Fcni%2Fnet.d%2F80-openshift-network.conf%0Ar%20%2Fetc%2Fcni%2Fnet.d%2F10-ovn-kubernetes.conf%0A
+  source: data:,r%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F80-openshift-network.conf%0Ar%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F10-ovn-kubernetes.conf%0Ar%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F00-multus.conf%0A
   verification: {}
 filesystem: root
 mode: 420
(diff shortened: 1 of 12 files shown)


squeed:
cc @fpan @dcbw - the question is whether or not we should be removing the multus configuration on boot. I argue yes.

runcom:
@squeed can you change the PR title to what agreeed on for BZ? (email thread from Clayton)

squeed:
Ah yeah, good point. Updated.

runcom:
/approve

@dcbw @fpan can you ack on this?

fepan:
LGTM.

squeed:
someone needs to actually do the /lgtm to make it count.

I don't understand what's up with that operator test failure. Surely it's a flake, right?

squeed:
Is there a chance I somehow broke something? This is never passing.

cgwalters:
It's possible that e2e-aws-op is failing for unrelated reasons; looks like it's timing out.  However it also reboots, which I don't think `e2e-aws` does today...so it will actually be testing removing the files on boot.  As of this moment `e2e-aws-upgrade` is also not actually testing OS updates either...I'd guess it's not rebooting either.  But probably what would make a ton of sense is to inject https://github.com/openshift/machine-config-operator/pull/682 into that suite.

cgwalters:
Did you try doing this change live on a running cluster and rebooting nodes?  (Probably simplest way to do that is to create a dummy MC for the workers)

squeed:
> Did you try doing this change live on a running cluster and rebooting nodes? (Probably simplest way to do that is to create a dummy MC for the workers)

Basically, yes. I created the file manually, rather than through the MCO, but yes. What is the test doing that goes above and beyond e2e-aws-upgrade?

runcom:
if something is broken in e2e-aws-op (as opposed as just being a legit timeout), you should see that in MCD logs in artifacts (probably have an error looping)

kikisdeliveryservice:
We are seeing these errors in `e2e-aws-op` across all prs now, so expect this to keep failing until we get a fix.

kikisdeliveryservice:
FTR: this PR has passed the now failing on all PRs e2e-aws-op job at least once on April 30th, at 9:40 PST:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/685/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/1918

smarterclayton:
Force merging at team's request, test failures are unrelated and separate investigation stream is covering them.

dcbw:
@squeed FTR yes, we should be removing Multus's config too.

PR #688, opened 2019-05-01
Remove unreachable toleration

RobertKrawitz:
Unreachable toleration is bad news; removing it.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1706204

### DIFF
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -37,24 +37,6 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
         operator: Exists
-      - key: node.kubernetes.io/memory-pressure
-        effect: NoSchedule
-        operator: Exists
-      - key: node.kubernetes.io/disk-pressure
-        effect: NoSchedule
-        operator: Exists
-      - key: node.kubernetes.io/not-ready
-        effect: NoExecute
-        operator: Exists
-      - key: node.kubernetes.io/unreachable
-        effect: NoExecute
-        operator: Exists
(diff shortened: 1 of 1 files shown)


cgwalters:
This PR should have a bugzilla link.

RobertKrawitz:
/hold

TL;dr: this will probably be closed.

@sjenning @ravisantoshgudimetla and I discussed this; we believe that the quorum guard (eqg) is doing the right thing and that the problem is the MCC requesting multiple masters to go unschedulable concurrently.

The eqg should tolerate everything _except_ for the node being unschedulable.  It should tolerate (both for scheduling and execution) disk pressure, memory pressure, etc.  If resource pressure gets so severe that the etcd member itself can't function, then the etcd will report unhealthy and the eqg will reflect that back upwards, but otherwise the eqg should just keep running.

cgwalters:
>  and that the problem is the MCC requesting multiple masters to go unschedulable concurrently.

Our `maxUnavailable: 1` should prevent this.  But it's possible.  Do you have any data or is this a theory?

RobertKrawitz:
We have evidence that two nodes are being marked SchedulingDisabled (check your slack notifications).  Whether that's the MCC doing so or something else we don't know.

runcom:
> We have evidence that two nodes are being marked SchedulingDisabled (check your slack notifications). Whether that's the MCC doing so or something else we don't know.

the MCC fix is in #697

RobertKrawitz:
/hold cancel

Still need to determine exactly what taints should be tolerated, but unreachable is definite not one of them.

@sjenning @ravisantoshgudimetla @runcom

runcom:
please rebase to pick up the CI fix

RobertKrawitz:
Closing in lieu of #700

PR #689, opened 2019-05-01
install: EQG: tolerate all taints

sjenning:
Rather than listing out all known taints and potentially breaking when a new one pops up, just tolerate all taints.

I think this is causing this gridlock in the MCC/MCD during upgrades
https://search.svc.ci.openshift.org/?search=failed+to+reboot&maxAge=48h&context=2&type=all

@runcom @RobertKrawitz


sjenning:
/hold
still discussing this...
@RobertKrawitz if we tolerate the `node.kubernetes.io/unschedulable` does the EQG even work any more?

sjenning:
move to https://github.com/openshift/machine-config-operator/pull/688

PR #690, opened 2019-05-01
Tolerate every taint

ravisantoshgudimetla:
As of now, we are tolerating only `node-role.kubernetes.io/master` taint during scheduling stage. If we don't have the blanket toleration, there is a very good chance that these pods won't tolerate other taints that got added to the node, for example `disk-pressure` etc. The side-effect is that we will tolerate `NoExecute` taints as well

Please feel free to close this PR, if you think, you just need to tolerate the current taint(s) you have.

/cc @sjenning @derekwaynecarr

### DIFF
--- a/manifests/machineconfigcontroller/deployment.yaml
+++ b/manifests/machineconfigcontroller/deployment.yaml
@@ -29,6 +29,4 @@ spec:
       priorityClassName: "system-cluster-critical"
       restartPolicy: Always
       tolerations:
-      - key: "node-role.kubernetes.io/master"
         operator: "Exists"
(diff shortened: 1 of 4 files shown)


sjenning:
we do not want to do this

wkulhanek:
@sjenning @derekwaynecarr @ravisantoshgudimetla 

I just opened a bugzilla because this breaks our cluster if we set up Infranodes with Taints to protect them from "regular" worker pods.

https://bugzilla.redhat.com/show_bug.cgi?id=1780318

If you have another way to fix this I'd like to know, please.

PR #694, opened 2019-05-02
WIP: DO NOT MERGE disable eqg e2e

runcom:
testing CI...


cgwalters:
Hmm offhand I don't think the problem is e-q-g since the master pool is updating OK, the problem is the worker pool - right?

kikisdeliveryservice:
@cgwalters the first time we saw the timeout was in ci when this test was added. But looks like e2e is stil failing even with this removed.

PR #695, opened 2019-05-02
WIP: kubelet: add disaster recovery for expired certificate

rphillips:
**- What I did**
If a cluster has been suspended for a period of time, then the Kubelet certificate may have expired. We need some way to re-bootstrap the machine if the expiry period has happened. This PR checks the expiry of the certificate, if expired, clears the required assets and reboots the machine.

**- How to verify it**


cgwalters:
https://stackoverflow.com/a/31718838 looks useful here.

But what does this fix?  If our cert is expired we can't join the cluster right?

rphillips:
@cgwalters updated the description

PR #696, opened 2019-05-02
controller: Log when a node has been updated

cgwalters:
The MCC logs show when we tell a node to start, but don't show us
when it's done.  This helps the MCC logs be a centralized
aggregate of events.

Doing this now to aid debugging slowness in updates in our `e2e-aws-op`
test suite.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -200,7 +200,14 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 	if pool == nil {
 		return
 	}
-	glog.V(4).Infof("Node %s updated", curNode.Name)
+
+	// Specifically log when a node has completed an update so the MCC logs are a useful central aggregate of state changes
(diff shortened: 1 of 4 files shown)


kikisdeliveryservice:
lgtm but will we be able to get this merged if ci keeps hitting the bug?

cgwalters:
> lgtm but will we be able to get this merged if ci keeps hitting the bug?

Yeah, we can force override if need be, but it's not urgent at the moment to do so.

kikisdeliveryservice:
aws-op is gonna keep failling so going to try to just retest images bc for some reason that failed:
`could not wait for build: could not create watcher for build machine-config-server: Get https://172.30.0.1:443/apis/build.openshift.io/v1/namespaces/ci-op-r2n14tsb/builds?fieldSelector=metadata.name%3Dmachine-config-server&watch=true: dial tcp 172.30.0.1:443: connect: connection refused`

/test images

cgwalters:
Added to https://github.com/openshift/machine-config-operator/pull/697

PR #697, opened 2019-05-02
adjust sigterm handler and bump test timeouts

runcom:
The NodeController shouldn't rely just on what it's syncing at that
moment to deduce that a node is unavailable. It may happen that the pool
is updating to a rendered-config-A but it's not done, and the code was
still going to apply a new rendered-config-B causing more than 1 node at
the time to go unschedulable. This patch should fix that.

This is also easy to reproduce... create a machineconfig, wait for 1 node to go unschedulable, then instantly create another MC which creates another rendered for a given pool, and start progressing on a second node.

This PR also: 

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1703877
and bump e2e timeouts

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -181,10 +181,6 @@ func isNodeReady(node *corev1.Node) bool {
 	return true
 }
 
-// TODO(runcom): drop currentConfig arg
-// getUnavailableMachines is the opposite of getReadyNodes - it filters the provided
-// set of nodes to ones which are either not ready, or whose current config does
-// not match the target.g
 func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
 	var unavail []*corev1.Node
 	for _, node := range nodes {
@@ -201,7 +197,7 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
 		}
 
 		nodeNotReady := !isNodeReady(node)
-		if dconfig != cconfig || nodeNotReady {
+		if dconfig == currentConfig && (dconfig != cconfig || nodeNotReady) {
 			unavail = append(unavail, node)
 			glog.V(2).Infof("Node %s unavailable: different configs (desired: %s, current %s) or node not ready %v", node.Name, dconfig, cconfig, nodeNotReady)
 		}
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -111,6 +111,8 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	return status
 }
 
+// getUpdatedMachines filters the provided nodes to ones whose current == desired
+// and the "done" flag is set.
 func getUpdatedMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
 	var updated []*corev1.Node
 	for _, node := range nodes {
@@ -138,6 +140,8 @@ func getUpdatedMachines(currentConfig string, nodes []*corev1.Node) []*corev1.No
 	return updated
 }
 
+// getReadyMachines filters the provided nodes to ones which are updated
+// and marked ready.
 func getReadyMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
 	updated := getUpdatedMachines(currentConfig, nodes)
 	var ready []*corev1.Node
@@ -178,6 +182,9 @@ func isNodeReady(node *corev1.Node) bool {
 }
 
 // TODO(runcom): drop currentConfig arg
+// getUnavailableMachines is the opposite of getReadyNodes - it filters the provided
+// set of nodes to ones which are either not ready, or whose current config does
+// not match the target.g
 func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
 	var unavail []*corev1.Node
 	for _, node := range nodes {
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -177,6 +177,7 @@ func isNodeReady(node *corev1.Node) bool {
 	return true
 }
 
+// TODO(runcom): drop currentConfig arg
 func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node {
 	var unavail []*corev1.Node
 	for _, node := range nodes {
@@ -193,7 +194,7 @@ func getUnavailableMachines(currentConfig string, nodes []*corev1.Node) []*corev
 		}
 
 		nodeNotReady := !isNodeReady(node)
-		if dconfig == currentConfig && (dconfig != cconfig || nodeNotReady) {
+		if dconfig != cconfig || nodeNotReady {
 			unavail = append(unavail, node)
 			glog.V(2).Infof("Node %s unavailable: different configs (desired: %s, current %s) or node not ready %v", node.Name, dconfig, cconfig, nodeNotReady)
 		}
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -135,7 +135,7 @@ const (
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
 	// currentConfigPath is where we store the current config on disk to validate
 	// against annotations changes
-	currentConfigPath = "/var/machine-config-daemon/currentconfig"
+	currentConfigPath = "/etc/machine-config-daemon/currentconfig"
 
 	kubeletHealthzEndpoint         = "http://localhost:10248/healthz"
 	kubeletHealthzPollingInterval  = time.Duration(30 * time.Second)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1000,7 +1000,7 @@ func (dn *Daemon) completeUpdate(node *corev1.Node, desiredConfigName string) er
 		return err
 	}
 
-	dn.logSystem("machine-config-daemon: completed update for config %s", desiredConfigName)
+	dn.logSystem("completed update for config %s", desiredConfigName)
 
 	return nil
 }
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -268,7 +268,7 @@ func NewClusterDrivenDaemon(
 	eventBroadcaster.StartRecordingToSink(&clientsetcorev1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 	dn.recorder = eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "machineconfigdaemon", Host: nodeName})
 
-	glog.Infof("Managing node: %s", nodeName)
+	dn.logSystem("Starting to manage node: %s", nodeName)
 
 	go dn.runLoginMonitor(dn.stopCh, dn.exitCh)
 
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -472,7 +472,7 @@ func (dn *Daemon) runOnceFrom() error {
 }
 
 // InstallSignalHandler installs the handler for the signals the daemon should act on
-func (dn *Daemon) InstallSignalHandler(stopCh chan struct{}) {
+func (dn *Daemon) InstallSignalHandler(signaled chan struct{}) {
 	termChan := make(chan os.Signal, 2048)
 	signal.Notify(termChan, syscall.SIGTERM)
 
@@ -486,7 +486,7 @@ func (dn *Daemon) InstallSignalHandler(stopCh chan struct{}) {
 				if dn.updateActive {
 					glog.Info("Got SIGTERM, but actively updating")
 				} else {
-					close(stopCh)
+					close(signaled)
 				}
 			}
 		}
@@ -496,7 +496,7 @@ func (dn *Daemon) InstallSignalHandler(stopCh chan struct{}) {
 // Run finishes informer setup and then blocks, and the informer will be
 // responsible for triggering callbacks to handle updates. Successful
 // updates shouldn't return, and should just reboot the node.
-func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
+func (dn *Daemon) Run(stopCh, signaled <-chan struct{}, exitCh <-chan error) error {
 	if dn.kubeletHealthzEnabled {
 		glog.Info("Enabling Kubelet Healthz Monitor")
 		go dn.runKubeletHealthzMonitor(stopCh, dn.exitCh)
@@ -519,6 +519,8 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	for {
 		select {
 		case <-stopCh:
+			return nil
+		case <-signaled:
 			return nil
 		case err := <-exitCh:
 			// This channel gets errors from auxiliary goroutines like loginmonitor and kubehealth
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -11,8 +11,10 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	imgref "github.com/containers/image/docker/reference"
@@ -88,7 +90,7 @@ type Daemon struct {
 	kubeletHealthzEnabled  bool
 	kubeletHealthzEndpoint string
 
-	installedSigterm bool
+	updateActive bool
 
 	nodeWriter NodeWriter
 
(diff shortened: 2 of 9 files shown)


pkg/controller/node/status.go
```diff
@@ -177,6 +177,7 @@ func isNodeReady(node *corev1.Node) bool {
 	return true
 }
 
+// TODO(runcom): drop currentConfig arg
```
cgwalters: I think actually this argument is acting as "gather nodes for pool", since it effectively filters them to just the pool's config.

Now is `getCandidateMachines()` trying to filter to just nodes owned by a pool?  Um...maybe?  I find all of these undocumented functions confusing...

runcom: at the time we're here, we already have the subset of nodes for our pool though, that's why I don't think that arg is necessary...

runcom: The thing with this is that a pool can only have a rendered config at a time it's syncing. If we allow 2 nodes to have different configs and progress them indipendently, we'll end up with the issue we're seeing. This shouldn't gate everything by saying: alrigth, everything must be idle in general in order to progress.

kikisdeliveryservice:
@runcom just to clarify is this PR ensuring that we don't apply config B unless and until config A is successfully applied/finished and fixing a problem where things get trampled bc another config was applied too soon right?

runcom:
> @runcom just to clarify is this PR ensuring that we don't apply config B unless and until config A is successfully applied/finished and fixing a problem where things get trampled bc another config was applied too soon right?

yup, that's the aim :)

kikisdeliveryservice:
> yup, that's the aim :)
 makes sense to me!!! :+1:

RobertKrawitz:
Sounds like a good idea on general principles

pkg/daemon/daemon.go
```diff
@@ -469,6 +471,28 @@ func (dn *Daemon) runOnceFrom() error {
...
+	// having the process be killed.
+	// https://github.com/openshift/machine-config-operator/issues/407
+	go func() {
+		for sig := range termChan {
+			switch sig {
+			case syscall.SIGTERM:
+				if dn.installedSigterm {
+					glog.Info("Got SIGTERM, but actively updating")
+				} else {
+					close(stopCh)
```
cgwalters: Copying from the BZ your comment:
> Now, in the MCD we have an handler for sigterm only during our sync, the rest of the code doesn't really care about sigterm so we don't catch it and we exit with 143 instead of 0 (if we had an handler).

I have serious difficulty believing that all other pods have `SIGTERM` handlers of this form.  Being killed by `SIGTERM` seems like a *completely normal* event - what makes the MCD special here?  

Maybe it's that most other pods are drained rather than killed by kubelet?  If that's the case...OK I guess, but I would vehemently argue that whatever is treating "container got `SIGTERM`" as an error is the real problem.

runcom: other operators aren't indeed getting sigtermed like us (we're a daemonset and we don't get drained).



> but I would vehemently argue that whatever is treating "container got `SIGTERM`" as an error is the real problem.

I would be with you :)

pkg/daemon/update.go
```diff
@@ -751,9 +727,15 @@ func (dn *Daemon) logSystem(format string, a ...interface{}) {
 	}
 }
 
+func (dn *Daemon) catchIgnoreSIGTERM() {
+	if dn.installedSigterm {
+		return
+	}
+	dn.installedSigterm = true
```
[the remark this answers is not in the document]

kikisdeliveryservice:
Looks good will defer to @cgwalters to give his signoff.

cgwalters:
OK I don't quite understand yet how your fix is addressing the problem.  Looking at this from the top, it certainly seems to me that `getCandidateMachines()` is broken - if `getReadyMachines()` returns nothing, it will happily reset the desired annotation for a node in progress.

Although...our default of `maxUnavailable = 1` means that if there is an unavailable node we'd have already skipped out before going into the candidates.

I guess I can try your reproducer.

OK nevermind, I think I do understand now, `getUnavailableNodes()` was happily reporting nodes as *not unavailable* if they didn't match the pool's current target.  One good way to think of this probably is that `getReadyNodes()` should be symmetric with `getUnavailableNodes` and with this change I think it is.

(Actually, why do we even have `getUnavailableNodes()` instead of just doing the set difference of `nodes - getAvailable(nodes)`?)

runcom:
> (Actually, why do we even have `getUnavailableNodes()` instead of just doing the set difference of `nodes - getAvailable(nodes)`?)

I wanted to actually do that, but I'm too afraid of changing the code.......

kikisdeliveryservice:
> > (Actually, why do we even have `getUnavailableNodes()` instead of just doing the set difference of `nodes - getAvailable(nodes)`?)
> 
> I wanted to actually do that, but I'm too afraid of changing the code.......

Yeah it makes sense bc the math _should_ check out. But... we could do it later too?

cgwalters:
/lgtm

Proof will be if this merges I guess.  Though it's not totally clear to me yet why this suddenly broke.

runcom:
> Though it's not totally clear to me yet why this suddenly broke.

the e2e I'm not sure about either, looks like a general slowdown? the race in node_controller was there for long apparently

cgwalters:
Looking at more PRs I see this consistently:

```
            "spec": {
                "providerID": "aws:///us-east-1b/i-0cc60b89b38b0eb6d",
                "taints": [
                    {
                        "effect": "NoSchedule",
                        "key": "node.kubernetes.io/unschedulable",
                        "timeAdded": "2019-05-02T17:11:05Z"
                    },
                    {
                        "effect": "NoSchedule",
                        "key": "node.kubernetes.io/unreachable",
                        "timeAdded": "2019-05-02T17:11:41Z"
                    },
                    {
                        "effect": "NoExecute",
                        "key": "node.kubernetes.io/unreachable",
                        "timeAdded": "2019-05-02T17:11:47Z"
                    }
                ],
                "unschedulable": true
            },
```

which looks to me like something is actually killing that node.

cgwalters:
> which looks to me like something is actually killing that node.

Or hm maybe that's normal while it's rebooting.

runcom:
looking into https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/692/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/1858 (which was @cgwalters PR which 120m timeout):

```
--- FAIL: TestMCDeployed (4936.88s)
	mcd_test.go:143: machine config didn't result in file being on any worker: timed out waiting for the condition
```

So it's about 82 minutes on just that to timeout the whole test

runcom:
I'm bringing up a cluster and run the e2e there - let's see what's going on

cgwalters:
Some debugging in a cluster runcom spun up shows the time for MCD initiating reboot to systemd actually bouncing the machine to take just about 1m30s consistently which seems fine.

cgwalters:
@runcom I added even more debugging to https://github.com/openshift/machine-config-operator/pull/696 - maybe simpler to roll the commits from that PR into this one?

runcom:
> @runcom I added even more debugging to #696 - maybe simpler to roll the commits from that PR into this one?

yep, feel free to push directly here

cgwalters:
OK though let's give `e2e-aws-op` a chance to complete before we push anything more.  I may need to head out for a bit, so feel free to do the merge if I'm not around.

runcom:
OK, I'm now bumping max unavailable to 3 and it should just pass

cgwalters:
I merged in https://github.com/openshift/machine-config-operator/pull/696
/lgtm

runcom:
now this is failing for no good reason as well:

```
--- FAIL: TestReconcileAfterBadMC (1816.32s)
	mcd_test.go:449: machine config didn't roll back on any worker: timed out waiting for the condition
```

runcom:
it just works (tm) on my laptop, I swear...

runcom:
Alrighty, I've reverted the very first commit about not progressing the pool if it's not idle, that's breaking the roll back on bad mc. I've taken that out, I'm opening another PR so we can get this in to unblock tests.

pkg/daemon/daemon.go
```diff
@@ -135,7 +135,7 @@ const (
 	pathStateJSON = "/etc/machine-config-daemon/state.json"
 	// currentConfigPath is where we store the current config on disk to validate
 	// against annotations changes
-	currentConfigPath = "/var/machine-config-daemon/currentconfig"
```
cgwalters: Wait I'm confused, at this point aren't we chrooted into the host?

kikisdeliveryservice:
@cgwalters added a question so:
/lgtm cancel

pkg/daemon/daemon.go
runcom: yaeh, I changed this for consistency with the other /etc paths and since /etc/machine-config-daemon is mounted in the daemonset. Will wait for tests to report and revert this if you don't want it (which I understand why)

kikisdeliveryservice:
this fixed the e2e-aws-op, first pass in days:
https://openshift-gce-devel.appspot.com/builds/origin-ci-test/pr-logs/directory/pull-ci-openshift-machine-config-operator-master-e2e-aws-op

/lgtm

cgwalters:
I wanted to look at the logs of that passed run to see if my new logging added any information; turns out the logs are now compressed twice just FYI everyone.

I noticed this in the MCD logs:

```
I0503 19:18:27.290073    2440 start.go:214] Shutting down MachineConfigDaemon
panic: close of closed channel

goroutine 1 [running]:
main.runStartCmd(0x1d55a60, 0x1d7e4f8, 0x0, 0x0)
	/go/src/github.com/openshift/machine-config-operator/cmd/machine-config-daemon/start.go:214 +0xaa9
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).execute(0x1d55a60, 0x1d7e4f8, 0x0, 0x0, 0x1d55a60, 0x1d7e4f8)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:766 +0x2c1
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x1d55800, 0x0, 0xc42043df78, 0x403dfc)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:852 +0x30a
github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra.(*Command).Execute(0x1d55800, 0xc42043df60, 0x0)
	/go/src/github.com/openshift/machine-config-operator/vendor/github.com/spf13/cobra/command.go:800 +0x2b
main.main()
	/go/src/github.com/openshift/machine-config-operator/cmd/machine-config-daemon/main.go:27 +0x31
```

Honestly I don't know why we're trying to do this stop channel stuff, let's just exit the process?

kikisdeliveryservice:
> I wanted to look at the logs of that passed run to see if my new logging added any information; turns out the logs are now compressed twice just FYI everyone.
> 
Ohhh. so that's why I had problems yesterday afternoon. Thanks!

runcom:
> Honestly I don't know why we're trying to do this stop channel stuff, let's just exit the process?

fixed that in the latest commit

cgwalters:
This PR feels like a bit of a mess now...it's basically ended up just being adjusting the timeouts, right?  If that's the case maybe it is clearer to do separate PRs for other things.  But eh.

There was a theory about this being network related with crio/kubelet; wonder if we just ended up winning a race.

Anyways I was looking at the controller logs in the passed `e2e-aws-op` run (after decompressing twice) and this looks odd:

```
I0503 19:18:04.818702       1 node_controller.go:476] Setting node ip-10-0-132-181.ec2.internal to desired config rendered-worker-fb998f7806949569640baf62cd48d0ad
I0503 19:18:04.830101       1 node_controller.go:476] Setting node ip-10-0-152-113.ec2.internal to desired config rendered-worker-fb998f7806949569640baf62cd48d0ad
I0503 19:18:04.840559       1 node_controller.go:476] Setting node ip-10-0-140-228.ec2.internal to desired config rendered-worker-fb998f7806949569640baf62cd48d0ad
I0503 19:18:04.851071       1 status.go:202] Node ip-10-0-132-181.ec2.internal unavailable: different configs (desired: rendered-worker-fb998f7806949569640baf62cd48d0ad, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready false
I0503 19:18:04.851190       1 status.go:202] Node ip-10-0-152-113.ec2.internal unavailable: different configs (desired: rendered-worker-fb998f7806949569640baf62cd48d0ad, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready false
... repeats some ...
I0503 19:18:14.841463       1 status.go:202] Node ip-10-0-140-228.ec2.internal unavailable: different configs (desired: rendered-worker-fb998f7806949569640baf62cd48d0ad, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready false
I0503 19:18:19.849176       1 node_controller.go:476] Setting node ip-10-0-132-181.ec2.internal to desired config rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:19.864522       1 node_controller.go:476] Setting node ip-10-0-152-113.ec2.internal to desired config rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:19.864672       1 node_controller.go:207] Pool worker: node ip-10-0-132-181.ec2.internal has completed update to rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:19.876107       1 node_controller.go:476] Setting node ip-10-0-140-228.ec2.internal to desired config rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:19.877922       1 node_controller.go:207] Pool worker: node ip-10-0-152-113.ec2.internal has completed update to rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:19.889322       1 node_controller.go:207] Pool worker: node ip-10-0-140-228.ec2.internal has completed update to rendered-worker-5eedbe776afba6cb62ce01ba6255a653
I0503 19:18:24.864998       1 status.go:178] Node ip-10-0-132-181.ec2.internal is reporting Unschedulable
I0503 19:18:24.865030       1 status.go:202] Node ip-10-0-132-181.ec2.internal unavailable: different configs (desired: rendered-worker-5eedbe776afba6cb62ce01ba6255a653, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready true
I0503 19:18:24.865083       1 status.go:178] Node ip-10-0-140-228.ec2.internal is reporting Unschedulable
I0503 19:18:24.865089       1 status.go:202] Node ip-10-0-140-228.ec2.internal unavailable: different configs (desired: rendered-worker-5eedbe776afba6cb62ce01ba6255a653, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready true
I0503 19:18:24.865095       1 status.go:178] Node ip-10-0-132-181.ec2.internal is reporting Unschedulable
...
I0503 19:21:06.775118       1 status.go:178] Node ip-10-0-140-228.ec2.internal is reporting Unschedulable
I0503 19:21:06.775124       1 status.go:202] Node ip-10-0-140-228.ec2.internal unavailable: different configs (desired: rendered-worker-5eedbe776afba6cb62ce01ba6255a653, current rendered-worker-5eedbe776afba6cb62ce01ba6255a653) or node not ready true
```

You can see how when we revert the config to the previous we get the "completed update" message (which I hadn't considered when adding it but makes sense).

However, for 2 minutes afterwards the nodes are still marked Unschedulable.  Then the logs stop...and I don't see the nodes as Unschedulable in the `nodes.json` so I guess it cleared up?  But why/how were they `Unschedulable` for 2 minutes after they shouldn't have been?

runcom:
I'm not sure what we're doing in those 2 minutes. Are MCD logs telling us something? It may be stuck in checkstateonboot?

cgwalters:
Here's another interesting thing, check out the logs from [this failed run](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/697/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/1919/), specifically note the MCD is in an infinite loop of:

```
I0503 17:46:48.659693    1586 update.go:180] Checking reconcilable for config rendered-worker-9dbe0ae65f449ba6974cd4b769c3f539 to rendered-worker-7abedcbf98d3be5c0b4b6f7f5c575e6f
I0503 17:46:48.659724    1586 update.go:239] Checking if configs are reconcilable
I0503 17:46:48.661298    1586 update.go:708] can't reconcile config rendered-worker-9dbe0ae65f449ba6974cd4b769c3f539 with rendered-worker-7abedcbf98d3be5c0b4b6f7f5c575e6f: invalid Ignition config found: error: no filesystem specified
I0503 17:46:48.664038    1586 daemon.go:411] Unable to apply update: can't reconcile config rendered-worker-9dbe0ae65f449ba6974cd4b769c3f539 with rendered-worker-7abedcbf98d3be5c0b4b6f7f5c575e6f: invalid Ignition config found: error: no filesystem specified: unreconcilable
```

Which...hum indeed, I think we want
```
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -288,7 +288,7 @@ func getNodesByRole(cs *framework.ClientSet, role string) ([]v1.Node, error) {
 func TestPoolDegradedOnFailToRender(t *testing.T) {
        cs := framework.NewClientSet("")
 
-       mcadd := createMCToAddFile("add-a-file", "/etc/mytestconfs", "test", "")
+       mcadd := createMCToAddFile("add-a-file", "/etc/mytestconfs", "test", "root")
        mcadd.Spec.Config.Ignition.Version = "" // invalid, won't render
 
        // create the dummy MC now
```
right?  It looks like this test was intending to have validation fail due to the version and not the filesystem.

So it looks like that MC source was deleted but its contents are still in the *rendered* version...the pool is correctly targeting the new one ```
                        "reason": "All nodes are updating to rendered-worker-9dbe0ae65f449ba6974cd4b769c3f539",
```

but the MCD hasn't noticed this.  Actually the problem is the node controller hasn't changed the desired annotation:

```
                    "machineconfiguration.openshift.io/desiredConfig": "rendered-worker-7abedcbf98d3be5c0b4b6f7f5c575e6f",
```

runcom:
I believe the failed run you're looking at was just related to the fact that we couldn't reconcile back with the patch I've not reverted (in the other PR now). That test is supposed to create a failing MC and that should roll back if you delete it (that's what the test does).

hexfusion:
Shazam!!

cgwalters:
OK, might as well get this in
/lgtm

PR #704, opened 2019-05-05
Small fixes + remove not needed mounts

runcom:
Couple of small fixes helping in debugging + removal of not needed mounts (and I suspect something is going on under `/etc` so just remove that as we chroot into rootfs anyway.


LorbusChris:
LGTM!

/lgtm

PR #707, opened 2019-05-06
pkg/daemon: lock when checking if update is active

runcom:
**- What I did**

Added a lock to read/write `dn.updateActive` as we're reading/writing it from at least 2 goroutines concurrently

**- How to verify it**

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -14,6 +14,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -90,7 +91,8 @@ type Daemon struct {
 	kubeletHealthzEnabled  bool
 	kubeletHealthzEndpoint string
 
-	updateActive bool
+	updateActive     bool
+	updateActiveLock sync.Mutex
```
cgwalters: /me mumbles something about `Mutex<bool>` :wink:

cgwalters:
(If we ever tried running with [the race detector](https://blog.golang.org/race-detector) we'd need this though)

kikisdeliveryservice:
`aws_internet_gateway.igw: error attaching EC2 Internet Gateway (igw-0dfbe89567c17a175): timeout while waiting for state to become 'success' (timeout: 2m0s)"` 

/test e2e-aws

runcom:
> though I am not sure how the current code could cause a real world bug - the conditionals as is are redundant right?

yeah, highly unlikely we'd hit this but as you pointed out, race detector might complain and surely it's good to protect that variable.

PR #708, opened 2019-05-06
daemon: Move chroot to very early on

cgwalters:
Today at least `detectEarlySSHAccessesFromBoot()` was racing
as it was spawning a goroutine that was expecting the target root,
but we were only calling `chroot()` later on.

I think everything will be a lot clearer if the daemon code is
basically *always* chrooted from as early on as possible.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -45,7 +45,7 @@ type RpmOstreeDeployment struct {
 // around content deployment
 type NodeUpdaterClient interface {
 	GetStatus() (string, error)
-	GetBootedOSImageURL(string) (string, string, error)
+	GetBootedOSImageURL() (string, string, error)
 	RunPivot(string) error
 }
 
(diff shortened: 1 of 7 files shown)


pkg/daemon/rpm-ostree.go
```diff
@@ -90,8 +90,8 @@ func (r *RpmOstreeClient) GetStatus() (string, error) {
 }
 
 // GetBootedOSImageURL returns the image URL as well as the OSTree version (for logging)
-func (r *RpmOstreeClient) GetBootedOSImageURL(rootMount string) (string, string, error) {
-	bootedDeployment, err := r.getBootedDeployment(rootMount)
+func (r *RpmOstreeClient) GetBootedOSImageURL() (string, string, error) {
```
kikisdeliveryservice: I like that this change removes these args!

ashcrow:
```
May 06 19:10:03.523 I test="[Disruptive] Cluster upgrade should maintain a functioning cluster [Feature:ClusterUpgrade] [Suite:openshift] [Serial]" failed
```

`connect: connection refused` when accessing the API server.

ashcrow:
Same connection refused error in `e2e-aws-upgrade`. Not retesting (at least not yet).

runcom:
/retest
/approve

is this still wip?

cgwalters:
I have a worry that we may be missing some mounts Kube injects besides the `/run/secrets` ones.  This did pass CI...but it's not fixing a critical bug that we know of, and we have other PRs we need to test.  Hence I'm adding the 4.2 label for now.

cgwalters:
Rebased :surfing_man: and lifting WIP.  I looked over kube mounts and one thing I did notice is that the chroot breaks the ability to write `/dev/termination-log`.  Which would actually be nice to handle...probably best done by holding an `openat()` file descriptor for it.

But it's not broken newly by this patch.  I'd like to get this one in first.

PR #709, opened 2019-05-06
daemon: Only log rpm-ostree status once, add boot list

cgwalters:
Having `rpm-ostree status` in the control loop greatly
adds to noise, it shouldn't change (except by us) before
a reboot.

While we're adding stuff that's logged only once, add
`journalctl --list-boots` which should be useful - it contains
the boot ID too.  I may also add some code later here to scrape
logs from the *previous* boot's MCD run.


kikisdeliveryservice:
This is an A+ idea. Cleaner logs would be great.

Just waiting to double check the e2e logs to confirm.

cgwalters:
This is to improve logs for https://bugzilla.redhat.com/show_bug.cgi?id=1706632

PR #710, opened 2019-05-06
daemon: Greatly slow down rate of retries

cgwalters:
If we hit an error, for the MCD it's very silly to retry in *milliseconds*.
Nothing we care about changes at a rate faster than a second really.

Tweak the queue to start from a second for the exponential backoff
with a cap at 60s (so we do keep retrying at a reasonable rate in
case an admin fixes something).  Drop the token bucket as I don't
think it applies to us.

Ideally what we'd do is slow down the rate of retries for the *same*
node object (that may not be hard, I didn't try), but this should
help avoid the MCD logs turning into spam in cases of error.  In
particular I really often want to see the starting MCD logs, and
that can get GC'd if we output too much.

Also drop `maxRetries` as we're managing the node and it has
to work, we'll keep retrying until it's fixed.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	"golang.org/x/time/rate"
 	imgref "github.com/containers/image/docker/reference"
 	ignv2 "github.com/coreos/ignition/config/v2_2"
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
@@ -144,17 +145,15 @@ const (
 	kubeletHealthzTimeout          = time.Duration(30 * time.Second)
 	kubeletHealthzFailureThreshold = 3
 
-	// TODO(runcom): increase retry and backoff?
-	//
-	// maxRetries is the number of times a node will be retried before it is dropped out of the queue.
-	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
-	// a node is going to be requeued:
-	//
-	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
-	maxRetries = 15
-
-	// updateDelay is a pause to deal with churn in Node
+	// updateDelay is the baseline speed at which we react to changes.  We don't
+	// need to react in milliseconds as any change would involve rebooting the node.
+	// Having this be relatively high limits the number of times we retry before
+	// the MCC/MCO will time out.  We don't want to spam our logs with the same
+	// error.
 	updateDelay = 5 * time.Second
+	// maxUpdateBackoff is the maximum time to react to a change as we back off
+	// in the face of errors.
+	maxUpdateBackoff = 60 * time.Second
 )
 
 type onceFromOrigin int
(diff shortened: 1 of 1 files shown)


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1706632

cgwalters:
Hmm testing this with a MC that writes to `/usr` I just get an infloop of 5s retries...the exponential backoff isn't kicking in.

cgwalters:
Oh whee and when I revert the bad MC we reboot for:

`I0506 20:58:20.804352   18279 update.go:708] Starting update from rendered-worker-700aa70cb983711bd6c4d882f46b6560 to rendered-worker-700aa70cb983711bd6c4d882f46b6560`

kikisdeliveryservice:
> Hmm testing this with a MC that writes to `/usr` I just get an infloop of 5s retries...the exponential backoff isn't kicking 

I double-checked and the format of workqueues seems correct.

cgwalters:
> I double-checked and the format of workqueues seems correct.

I've reread the workqueue docs a few times and looked at other controllers as well as the [sample controller](https://github.com/kubernetes/sample-controller/blob/master/controller.go) and what we're doing seems right...

rphillips:
@cgwalters ItemExponentialFailureRateLimiter has a backoff on each error type. If you are throwing different errors, then there isn't a backoff. You may want to add a BucketRateLimiter to control the speed.

https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L94
https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L43

cgwalters:
> ItemExponentialFailureRateLimiter has a backoff on each error type. If you are throwing different errors, then there isn't a backoff. 

I don't understand this...how does the queue see the error types?  It only sees the item right?

cgwalters:
After changing the default log level I verified I see the message e.g.
`I0506 21:31:47.794508   21763 daemon.go:346] Error syncing node ip-10-0-140-69.ec2.internal (retries 17): open /usr/lib/.mctest-should-not-exist654725038: read-only file system`
And note the `retries 17` which comes from the `ItemExponentialFailureRateLimiter` code..

rphillips:
True, you are right on the items. Doesn't the time.Sleep need to come after the AddRateLimited? There would a be a race to add the key to the queue for ~5 seconds, potentially giving you a batch of errors in that 5 second window.

```
dn.queue.AddRateLimited(key)
time.Sleep(updateDelay)
```

cgwalters:
OK I reworked this and tested it, seems to work pretty well; I see us backing off to 1 minute retries on errors.  The problem turned out to be (AFAICS) that we always need to use the `queue.AddRateLimited()` API - if we call `queue.AddAfter` explicitly that bypasses the rate limiter wrappers.

cgwalters:
Restarting the MCD revealed another bug, quoting commit:

> The handling for the bootstrap case needs to be in the main sync function; we were forgetting to call `.Done()` and so in the case where the MCD was restarted on a bad MC it'd go into a tight retry loop.

runcom:
> Restarting the MCD revealed another bug, quoting commit:

can you link the commit? not sure I understand what's the issue

runcom:
needs an unfortunate rebase...

kikisdeliveryservice:
hitting EBS:
/test e2e-aws

PR #711, opened 2019-05-06
pkg/daemon: log pending config to journal

runcom:
Let's try to avoid losing a file write...

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -11,6 +12,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
@@ -72,18 +74,6 @@ func writeFileAtomically(fpath string, b []byte, dirMode, fileMode os.FileMode, 
 	return t.CloseAtomicallyReplace()
 }
 
-func (dn *Daemon) writePendingState(desiredConfig *mcfgv1.MachineConfig) error {
-	t := &pendingConfigState{
-		PendingConfig: desiredConfig.GetName(),
-		BootID:        dn.bootID,
-	}
-	b, err := json.Marshal(t)
-	if err != nil {
-		return err
-	}
-	return writeFileAtomicallyWithDefaults(pathStateJSON, b)
-}
-
 func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
 	return &corev1.ObjectReference{
 		Kind: "Node",
@@ -99,13 +89,13 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) (retErr err
 		return err
 	}
 
-	if err := dn.writePendingState(newConfig); err != nil {
-		return errors.Wrapf(err, "writing pending state")
+	if err := dn.storePendingState(newConfig, 1); err != nil {
+		return errors.Wrap(err, "failed to log pending config")
 	}
 	defer func() {
 		if retErr != nil {
-			if err := os.Remove(pathStateJSON); err != nil {
-				retErr = errors.Wrapf(retErr, "error removing pending config file %v", err)
+			if err := dn.storePendingState(newConfig, 0); err != nil {
+				retErr = errors.Wrapf(retErr, "error rolling back pending config %v", err)
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -702,6 +692,36 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+
+func (dn *Daemon) logPendingConfig(pending *mcfgv1.MachineConfig) error {
+	logger := exec.Command("logger", "--journald")
+	stdin, err := logger.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("failed to get stdin pipe: %v", err)
+	}
+
+	pendingConfigStr := fmt.Sprintf(`MESSAGE_ID=34c7912c5dd2454286097f8f92a22e9e
+MESSAGE=%s`, pending.GetName())
```
runcom: TODO write bootid

pkg/daemon/update.go
```diff
@@ -702,6 +692,36 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	return nil
 }
 
+func (dn *Daemon) readPendingConfig() (string, error) {
+	// TODO(runcom): msgid has been generated with journalctl --new-id128, move it to a const
+	journalOutput, err := exec.Command("journalctl", "-b", "-1", "-o", "cat", "MESSAGE_ID=34c7912c5dd2454286097f8f92a22e9e").CombinedOutput()
```
runcom: output to json to read BOOT_ID and MESSAGE

runcom: also, we might want to have a customized "machine-config-daemon-pending-config" as msgid - not sure about opinions on this?

pkg/daemon/update.go
runcom: same, message_id can be just what we want...see https://github.com/openshift/machine-config-operator/pull/711/files#r281399898

pkg/daemon/update.go
```diff
@@ -702,6 +691,36 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+}
+
+func (dn *Daemon) logPendingConfig(pending *mcfgv1.MachineConfig) error {
+	logger := exec.Command("logger", "--journald")
+	stdin, err := logger.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("failed to get stdin pipe: %v", err)
+	}
+
+	pendingConfigStr := fmt.Sprintf(`MESSAGE_ID=34c7912c5dd2454286097f8f92a22e9e
```
cgwalters: This could use some abstraction into a wrapper that takes `key=value` strings as varargs or so.

(One other random thought I had is that we `systemd-run` a service on the host and send our logs to it and have it log them; then we can properly `systemctl status openshift-machine-config` or so - it would also basically act as a host-side mutex too to make doubly sure there aren't multiple MCDs)

cgwalters:
I know I suggested this...but I'm wavering a bit.   I'd be a lot happier if we had a stronger idea of what was happening...I really really want to get live access to an affected cluster.

Maybe the best bet is to keep throwing in more logging PRs and see what comes from that.

runcom:
> I know I suggested this...but I'm wavering a bit. I'd be a lot happier if we had a stronger idea of what was happening...I really really want to get live access to an affected cluster.
> 
> Maybe the best bet is to keep throwing in more logging PRs and see what comes from that.

I concur with that - I opened this to validate if it was something which we might pursue..

runcom:
console/authentication failures

/retest

ashcrow:
```
level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: authentication, console: timed out waiting for the condition"
```

pkg/daemon/update.go
```diff
@@ -702,6 +691,61 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+		}
+		if entry.BootID == dn.bootID {
+			return "", fmt.Errorf("pending config %s bootID %s matches current! Failed to reboot?", entry.Message, dn.bootID)
+		}
+		if entry.Pending == 1 {
+			return entry.Message, nil
+		}
+		return "", nil
+	}
+	return "", fmt.Errorf("no pending config found in journal: %v", string(journalOutput))
```
ashcrow: Could this end up being a very large error to return? IE: `journalOutput` containing lots of content?

pkg/daemon/update.go
```diff
@@ -702,6 +691,61 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+}
+
+func (dn *Daemon) logPendingConfig(pending *mcfgv1.MachineConfig, isPending int) error {
+	logger := exec.Command("logger", "--journald")
```
ashcrow: It may be worth noting why writing to `stdin` is used rather than directly executing the command with the message. At first I was going to recommend simplifying via a direct call until I gave `logger` a run and realized that direct calling it is awkward (double enter to end input).

pkg/daemon/update.go
runcom: we're filtering out the journal itself only on our message id and even if we reboot 100 times, we only get 100 lines - this error isn unreacheable though, I need to remove it

runcom:
oh so nice this worked and we're now reading/writing to journal the pending config

kikisdeliveryservice:
@runcom are we going this route for now then?

runcom:
> @runcom are we going this route for now then?

let's hear back from @cgwalters, but I would greatly love to merge this and validate it properly in the upgrade jobs (we need this to land to master in order for the upgrade to pick up this change as the starting point of the job itself).
I share the hesitation with Colin into going this route anyway but I'm still standing towards finding the root cause of the bugzilla and properly handle that. Having said that, as closer as we might be, if this turns out to be stable enough, why not having it.

cgwalters:
I think what's pushing me towards merging this the most is that it will make auditing events easier.

The code design looks good to me.

/lgtm

runcom:
rebased and re-pushed.
/lgtm

PR #712, opened 2019-05-07
daemon: Tweak logging during updates

cgwalters:
We were printing errors 3 times, reduce that to 1.
Also strip out an extra reconciliation log.


cgwalters:
Just to clean up the logs for https://bugzilla.redhat.com/show_bug.cgi?id=1706606

PR #713, opened 2019-05-07
Bug 1707212: only log reboot command error and change it to systemd-run

runcom:
The reboot command itself is being terminated as shown in this log:
```
worker: "pool is degraded because nodes fail with "1 nodes are reporting degraded status on sync": "Node ip-10-0-144-80.ec2.internal is reporting: \"failed to reboot: signal: terminated\"""
```

By the time we issue a reboot it's probably ok to just log reboot errors

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -167,8 +167,12 @@ const (
 
 var (
 	defaultRebootTimeout = 24 * time.Hour
-	defaultRebootCommand = "reboot"
 )
+
+func rebootCommand(rationale string) *exec.Cmd {
+	return exec.Command("systemd-run", "--unit", "machine-config-daemon-reboot",
+		"--description", fmt.Sprintf("machine-config-daemon: %s", rationale), "reboot")
+}
 
 // New sets up the systemd and kubernetes connections needed to update the
 // machine.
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -474,20 +474,20 @@ func (dn *Daemon) runOnceFrom() error {
 // InstallSignalHandler installs the handler for the signals the daemon should act on
 func (dn *Daemon) InstallSignalHandler(signaled chan struct{}) {
 	termChan := make(chan os.Signal, 2048)
-	signal.Notify(termChan, syscall.SIGTERM)
+	signal.Notify(termChan, syscall.SIGTERM, os.Interrupt)
```
cgwalters: I don't think we're getting `SIGINT`; why do you think so?

pkg/daemon/update.go
```diff
@@ -768,9 +768,11 @@ func (dn *Daemon) reboot(rationale string, timeout time.Duration, rebootCmd *exe
 
 	// reboot, executed async via systemd-run so that the reboot command is executed
 	// in the context of the host asynchronously from us
-	err := rebootCmd.Run()
-	if err != nil {
-		return errors.Wrapf(err, "failed to reboot")
+	// We're not returning the error from the reboot command as it can be terminated by
+	// the system itself with signal: terminated. We can't catch the subprocess termination signal
+	// either, we just have one for the MCD itself.
+	if err := rebootCmd.Run(); err != nil {
```
cgwalters: See also https://github.com/openshift/machine-config-operator/pull/237 - at one point when trying to get rpm-ostree upgrades (with reboots) and Ansible to play nicely together I spent a *lot* of time learning about reboots...it turned out to work a lot better to do basically `systemd-run reboot`.

In that PR I was trying to do something more ambitious but it should work here to just change the reboot command as I did in that commit to e.g.:

```
rebootCmd := exec.Command("systemd-run", "--unit", "machine-config-daemon-reboot",
		"--description", fmt.Sprintf("machine-config-daemon: %s", rationale)
```
or so.

runcom: I'll change it to that

pkg/daemon/daemon.go
runcom: dropped indeed

pkg/daemon/update.go
runcom: Fixed

PR #701, opened 2019-05-03
Bug 1707212: Rework controller progress

cgwalters:
Bigger rewrite for #699

This should ensure that the node controller avoids exceeding maxUnavailable,
by changing the notion of "unavailable" to mean nodes which are either
not ready, *or* may become not ready as they're working on an update.

That list is also further filtered by nodes which are degraded/unreconcilable.

PR #718, opened 2019-05-08
 controller: Fix "changed" logic to account for nodeReady changes

cgwalters:
The `nodeChanged` logic was not accounting for the fact that
the controller takes readiness into account too.  However, rather
than change that function, let's just remove it and extend our
logic below that which was also effectively doing change detection
so it could log it.

This also removes the logging from `nodeReady()` which got very
noisy in status; we now consistently log just changes in the node
controller.

However, the change detection logic was also implicitly ignoring
nodes which didn't appear to be managed by the MCD at all - think
Windows nodes.  Let's explicitly skip nodes that don't have a `currentConfig`
annotation.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -188,13 +188,13 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 	oldNode := old.(*corev1.Node)
 	curNode := cur.(*corev1.Node)
 
-	if !nodeChanged(oldNode, curNode) {
+	if !isNodeManaged(curNode) {
 		return
 	}
 
 	pool, err := ctrl.getPoolForNode(curNode)
 	if err != nil {
-		glog.Errorf("error finding pools for node: %v", err)
+		glog.Errorf("error finding pool for node: %v", err)
 		return
 	}
 	if pool == nil {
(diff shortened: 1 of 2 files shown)


cgwalters:
OK this one is rebased :surfing_man: on master which has #719 now.

cgwalters:
Ok this is rebased :surfing_man: on master which as #719 now.

pkg/controller/node/status.go
```diff
@@ -153,32 +171,32 @@ func getReadyMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node
...
 	for i := range node.Status.Conditions {
 		cond := &node.Status.Conditions[i]
 		// We consider the node for scheduling only when its:
 		// - NodeReady condition status is ConditionTrue,
 		// - NodeOutOfDisk condition status is ConditionFalse,
 		// - NodeNetworkUnavailable condition status is ConditionFalse.
 		if cond.Type == corev1.NodeReady && cond.Status != corev1.ConditionTrue {
-			glog.Infof("Node %s is reporting NotReady", node.Name)
-			return false
+			return fmt.Errorf("Node %s is reporting NotReady", node.Name)
```
kikisdeliveryservice: nit: these should be lowercase error messages. we can fix later.

pkg/controller/node/status.go
```diff
@@ -153,32 +171,32 @@ func getReadyMachines(currentConfig string, nodes []*corev1.Node) []*corev1.Node
...
 	if node.Spec.Unschedulable {
-		glog.Infof("Node %s is reporting Unschedulable", node.Name)
-		return false
+		return fmt.Errorf("Node %s is reporting Unschedulable", node.Name)
 	}
-	return true
+	return nil
+}
+
+func isNodeReady(node *corev1.Node) bool {
```
kikisdeliveryservice: couldn't we just return bool, error in checkNodeReady()?

pkg/controller/node/node_controller.go
```diff
@@ -188,31 +188,63 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
...
 	}
 	if pool == nil {
 		return
 	}
 	glog.V(4).Infof("Node %s updated", curNode.Name)
 
-	oldReady := isNodeReady(oldNode)
-	newReady := isNodeReady(curNode)
+	var changed bool
+	oldReadyErr := checkNodeReady(oldNode)
```
kikisdeliveryservice: later: we could make this a small function instead of repeating code

kikisdeliveryservice:
went through this and LGTM

/lgtm

runcom:
/hold

 for beta I guess

pkg/controller/node/status.go
cgwalters: Yeah, though keeping the API the same involved less code churn for the callers.

pkg/controller/node/status.go
```diff
@@ -75,8 +77,12 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		readyMachineCount == machineCount &&
 		unavailableMachineCount == 0 {
 		//TODO: update api to only have one condition regarding status of update.
-		supdated := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdated, corev1.ConditionTrue, fmt.Sprintf("All nodes are updated with %s", pool.Status.Configuration.Name), "")
+		updatedMsg := fmt.Sprintf("All nodes are updated with %s", pool.Status.Configuration.Name)
+		supdated := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdated, corev1.ConditionTrue, updatedMsg, "")
```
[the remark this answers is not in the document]
cgwalters: I think so, I didn't introduce the term though.

cgwalters:
This was merged as part of https://github.com/openshift/machine-config-operator/pull/701

PR #720, opened 2019-05-08
BUG 1703581: templates/master/00-master: bump etcd ETCD_QUOTA_BACKEND_BYTES default to 7GB

hexfusion:
After conversations with various folks, we have decided to increase the default for ETCD_QUOTA_BACKEND_BYTES to 7516192768 (7GB). The value protects users from uncontrolled growth. If a user deploys a cluster and does not perform proper maintenance procedures such as defragmentation they will hit an NOSPACE alarm. This alarm puts the cluster into a maintenance mode which only allows read and delete.

But this process is disruptive, also we would like to reduce day 2 configuration changes required. This new default will cover most users and also add 1GB protection against runaway clusters. To allow time to plan next steps and create a second cluster.

/cc @jupierce


runcom:
can you rebase to pick #719

kikisdeliveryservice:
hopefully this passes e2e-upgrade now with the rebase picking up 719.

/approve

kikisdeliveryservice:
This is for GA so we'll hold until tomorrow.

/hold

blomquisg:
> This is for GA so we'll hold until tomorrow.

GA code freeze is today (May 8th).  Should this say, "This is *not* for GA"?  If it holds until tomorrow, it will not be in GA.

kikisdeliveryservice:
re-adding LGTM since @jupierce already LGTM'ed the substance of the PR.  only change since then was a rebase to pick up 719 to pass e2e.

all tests pass now.

/lgtm

hexfusion:
\moonwalk

hexfusion:
this passed all tests FTR

PR #721, opened 2019-05-08
daemon: Remove dummy OnceFrom Test

cgwalters:
Now that the daemon tries to talk to the systemd journal, this unit
test hangs in my dev environment.  We can't really support this
without mocking a lot more, and it's not really covering much
interesting right now.  Delete it.

### DIFF
--- a/pkg/daemon/daemon_test.go
+++ b/pkg/daemon/daemon_test.go
@@ -128,38 +128,6 @@ func TestCompareOSImageURL(t *testing.T) {
 	}
 }
 
-func TestDaemonOnceFromNoPanic(t *testing.T) {
-	if _, err := os.Stat("/proc/sys/kernel/random/boot_id"); os.IsNotExist(err) {
-		t.Skip("we're not on linux")
-	}
-
-	exitCh := make(chan error)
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
this test is passing in ci though?

kikisdeliveryservice:
is your dev env aws or libvirt?

cgwalters:
> is your dev env aws or libvirt?

I'm talking about just running the unit tests, not a cluster.  My local dev env is a [toolbox container](https://github.com/debarshiray/toolbox/) (on Silverblue 30 host FWIW) but the important part is the toolbox container is somewhat privileged (or at least, not confined); e.g. I can talk to journald from it.  It runs as my main uid, not root though.

The `ci/prow/unit` test build is a container that isn't at all privileged, so talking to journald just fails and doesn't panic.

runcom:
I'm fine dropping this

kikisdeliveryservice:
@cgwalters thanks for the details

/lgtm

kikisdeliveryservice:
ok reconfirmed we're ok with dropping this test.

/hold cancel

PR #719, opened 2019-05-08
pkg/daemon: fix pending config logic

runcom:
we don't need to compare bootid if we have a _not_ pending config...

xref https://github.com/openshift/machine-config-operator/pull/711


imcleod:
LGTM - Merging immediately per request from Antonio

PR #724, opened 2019-05-08
MCD: bump terminationGracePeriodSeconds

runcom:
This job https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_installer/1726/pull-ci-openshift-installer-master-e2e-aws-upgrade/319 is showing us an interesting pattern where:

1) we start a drain
2) we're asked to terminate
3) we exceed the termination grace period (currently 300s)
4) we get terminated
5) we don't reboot and fail the update

I'm still investigating why we're asked to terminate but drain's timeout (not counting the backoff on possible failures) is 600s so this PR is just bumping the DS termination to at least that.

### DIFF
--- a/manifests/machineconfigdaemon/daemonset.yaml
+++ b/manifests/machineconfigdaemon/daemonset.yaml
@@ -35,7 +35,7 @@ spec:
       hostNetwork: true
       hostPID: true
       serviceAccountName: machine-config-daemon
-      terminationGracePeriodSeconds: 300
+      terminationGracePeriodSeconds: 600
       tolerations:
       - key: node-role.kubernetes.io/master
         operator: Exists
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -590,7 +590,7 @@ spec:
       hostNetwork: true
       hostPID: true
       serviceAccountName: machine-config-daemon
-      terminationGracePeriodSeconds: 300
+      terminationGracePeriodSeconds: 600
       tolerations:
       - key: node-role.kubernetes.io/master
         operator: Exists


kikisdeliveryservice:
This makes sense to me.

/lgtm

cgwalters:
We don't think there's a relationship between 1 and 2 right?   In other words, we could also be in the middle of doing an OS update or writing configs and get the termination request.  This is giving us more headroom for that.

Unless...somehow we think that the drain is causing the delete.

runcom:
> We don't think there's a relationship between 1 and 2 right?

I hope there isn't cause we ignore DS in Drain specifically. Turns our the DS controller is asking ask to terminate though:

```
{"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Metadata","auditID":"d91224d0-7781-4188-8be1-845683902ba8","stage":"ResponseComplete","requestURI":"/api/v1/namespaces/openshift-machine-config-operator/pods/machine-config-daemon-g9l25","verb":"delete","user":{"username":"system:serviceaccount:kube-system:daemon-set-controller","uid":"f45fe268-71c3-11e9-b028-12af4ec327be","groups":["system:serviceaccounts","system:serviceaccounts:kube-system","system:authenticated"]},"sourceIPs":["::1"],"userAgent":"hyperkube/v1.13.4+21f1917 (linux/amd64) kubernetes/21f1917/system:serviceaccount:kube-system:daemon-set-controller","objectRef":{"resource":"pods","namespace":"openshift-machine-config-operator","name":"machine-config-daemon-g9l25","apiVersion":"v1"},"responseStatus":{"metadata":{},"code":200},"requestReceivedTimestamp":"2019-05-08T19:50:49.639613Z","stageTimestamp":"2019-05-08T19:50:49.646305Z","annotations":{"authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":"RBAC: allowed by ClusterRoleBinding \"system:controller:daemon-set-controller\" of ClusterRole \"system:controller:daemon-set-controller\" to ServiceAccount \"daemon-set-controller/kube-system\""}}
```

Can't tell why tho. The above matches the timestamp kubelet sends us a terminate.

PR #725, opened 2019-05-09
server: update healthz to return 200 instead of 204

abhinavdahiya:
**- What I did**

Returning 204 `StatusNoContent` for `/healthz` is technically the correct response. And on AWS [1] the health checks treat 200 to 399 as Healthy.
But on Azure, the health checks on treat 200 `StatusOK` [2] as the healthy response. :shake_fist:

This moves the ign server to return 200 for `/heathz`

[1]: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html
[2]: https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-custom-probe-overview#httpprobe

**- How to verify it**

Updates the unit tests

`update healthz to return 200 instead of 204`

/cc @wking @runcom

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -134,7 +134,7 @@ type healthHandler struct{}
 func (h *healthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", "0")
 	if r.Method == http.MethodGet || r.Method == http.MethodHead {
-		w.WriteHeader(http.StatusNoContent)
+		w.WriteHeader(http.StatusOK)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


wking:
Sigh.

/lgtm

Is there anywhere we can complain to Azure about this?

abhinavdahiya:
Looking at the GCP docs, it looks like even google requires `200 (Status OK)` for healthy.
https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http
_ even more :shake_fist: _

serbrech:
@wking we'll make sure to complain about this internally already.. I'd be surprised if we were the only ones.
Problem is, I suppose, that we could break existing systems by changing this behaviour.

wking:
I don't care about 3xx, but 2xx, and especially 200 and 204, are pretty clearly "I'm good"

cgwalters:
/approve
But moving to 4.2 unless I hear otherwise.

kikisdeliveryservice:
No idea why prow is stuck. I reported it and seeing it elsewhere...

PR #727, opened 2019-05-09
Bug 1707162: machine-config-server: Bind to a specific address, not all

ashcrow:
**- What I did**
Use `k8s.io/apimachinery/pkg/util/net` to default to a specific network rather than `0.0.0.0`

- Inclusion of `k8s.io/apimachinery/pkg/util/net` for finding the MCS default bind network
- New MCS root option of address
- NewAPIServer now takes an address

/cc @runcom @imcleod

### DIFF
--- a/cmd/machine-config-server/bootstrap.go
+++ b/cmd/machine-config-server/bootstrap.go
@@ -43,8 +43,8 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 	}
 
 	apiHandler := server.NewServerAPIHandler(bs)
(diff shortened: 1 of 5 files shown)


knobunc:
@danwinship @squeed PTAL

cgwalters:
Just recording thoughts here: One possible fallout is around e.g. baremetal network setups, they may be relying on accessing the MCS from a different interface

cgwalters:
/hold
also from discussion the above issue probably requires fixing if we were going to do this.

ashcrow:
Closing per offline discussion.

PR #728, opened 2019-05-09
Bug 1707928: hack/build-go.sh: use just git hashes for building

runcom:
**- What I did**

Do not rely on tags (as they may contain timestamps which can cause us to drift)

**- How to verify it**

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -20,7 +20,9 @@ if [ -z ${VERSION_OVERRIDE+a} ]; then
 	VERSION_OVERRIDE=$(git describe --abbrev=8 --dirty --always)
 fi
 
-GLDFLAGS+="-X ${REPO}/pkg/version.Raw=${VERSION_OVERRIDE}"
+HASH=$(git rev-parse --verify 'HEAD^{commit}')
+
+GLDFLAGS+="-X ${REPO}/pkg/version.Raw=${VERSION_OVERRIDE} -X ${REPO}/pkg/version.Hash=${HASH}"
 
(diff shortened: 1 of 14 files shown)


hack/build-go.sh
```diff
@@ -17,7 +17,7 @@ cdup="$(git rev-parse --show-cdup)" && test -n "$cdup" && cd "$cdup"
 
 if [ -z ${VERSION_OVERRIDE+a} ]; then
 	echo "Using version from git..."
-	VERSION_OVERRIDE=$(git describe --abbrev=8 --dirty --always)
+	VERSION_OVERRIDE=$(git rev-parse --verify 'HEAD^{commit}')
```
cgwalters: Wouldn't `--verify HEAD` work too?  I had to go look up that syntax.

Are there no other repos which do something similar?  We should try to keep this in common across repos.

cgwalters: It's fine by me as is too.

runcom:
damn this is failing cause we want a valid semver version

runcom:
alright, changed to just have a new field in the version pkg that points to the git hash and use that whenever we need to compare MCC versions. @cgwalters ptal

runcom:
gosh, this is going through another round of testing cause #701 has been merged meanwhile

runcom:
```
annotations: {
machineconfiguration.openshift.io/generated-by-controller-version: "087bb5c36cdba30edeb2f52c69f9b2a953b800bf"
},
```
That's what we get now, and we easily point to a given MCO git commit when asked.

PR #729, opened 2019-05-09
[wip] See what breaks when node-bootstrapper cannot use CSR

enj:
@deads2k @sjenning @openshift/sig-auth


cgwalters:
I believe this is obsoleted by https://github.com/openshift/machine-config-operator/pull/741
/close
Feel free to reopen if I'm wrong.

PR #734, opened 2019-05-10
Bug 1708602:  pkg/daemon: workaround old util-linux logger in rhel7.6

runcom:
**- What I did**

RHEL7.6 has an old util-linux's logger which doesn't support the `--journald` flag to log to journal in a structured way. Work that around by logging a json directly and fall back to that if we're on rhel7.6 then.

**- How to verify it**

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -95,6 +95,9 @@ func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) (retErr err
 	}
 	defer func() {
 		if retErr != nil {
+			if dn.recorder != nil {
+				dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "PendingConfigRollBack", fmt.Sprintf("Rolling back pending config %s: %v", newConfig.GetName(), retErr))
+			}
 			if out, err := dn.storePendingState(newConfig, 0); err != nil {
 				retErr = errors.Wrapf(retErr, "error rolling back pending config %v: %s", err, string(out))
 				return
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -13,6 +13,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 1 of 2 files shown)


runcom:
/hold

Depends on #733

pkg/daemon/update.go
```diff
@@ -701,15 +712,25 @@ func (dn *Daemon) getPendingState() (string, error) {
 		return "", err
 	}
 	if len(journalOutput) == 0 {
-		return "", nil
+		if !loggerWithoutJournalOption() {
+			return "", nil
+		}
+		cmd := "journalctl -o cat | grep OLD_LOGGER"
```
cgwalters: Eeek.  Scraping all journal messages for that?

How about we detect on *startup* whether the host is rhel7, and add that as a flag in the daemon struct.  That'd be useful for a lot of other things.

pkg/daemon/update.go
```diff
@@ -734,6 +755,16 @@ BOOT_ID=%s
...
 
 	logger.Stdin = &pendingState
+	loggerOutput, err := logger.CombinedOutput()
+	if err != nil {
+		// RHEL 7.6 logger (util-linux) doesn't have the --journald flag and it's failng to log the above.
+		if !strings.Contains(string(loggerOutput), "unrecognized option '--journald'") {
+			return loggerOutput, err
+		}
+		// Super hacky workaround for the above
+		oldLogger := exec.Command("logger", fmt.Sprintf(`{"MESSAGE": "%s", "BOOT_ID": "%s", "PENDING": "%d", "OLD_LOGGER": "1"}`, pending.GetName(), dn.bootID, isPending))
```
cgwalters: Then in this code we wouldn't try and fall back each time, we'd just do:

```
if (!dn.hostRhel7) {
   logger --journald
} else {
  logger with literal JSON as message
}
```

pkg/daemon/update.go
runcom: we scrape all journal oly if --journald isn't supported so we need to do this anyway right? I didn't want to add a specific rhel7 flag as I was just testing for functionality which should be generally better. If rhel7.7 upgrades util linux to a logger that does have --journald, we'd be still using this awful ack if we just test RHEL versions

pkg/daemon/update.go
runcom: see https://github.com/openshift/machine-config-operator/pull/734#discussion_r282886553

pkg/daemon/update.go
cgwalters: > If rhel7.7 upgrades util linux to a logger that does have --journald, we'd be still using this awful ack if we just test RHEL versions

Even if it did upgrade which...seems *very* unlikely (right?) we'd still have an upgrade hazard where for previous versions we'd write one way and then later the other way.  Seems better to just always live with one format.

runcom: > Eeek. Scraping all journal messages for that?

to reply specifically on that. With this patch, we're not _always_ scraping all journal, `if !loggerWithoutJournalOption() {` takes care of telling us if we need to do that in case `logger` doesn't have `--journald`.

runcom: > Even if it did upgrade which...seems very unlikely (right?) we'd still have an upgrade hazard where for previous versions we'd write one way and then later the other way. Seems better to just always live with one format.

Yeah, agree it's very unlikely indeed, but since we don't need to play with a rhel7 flag today, isn't testing for the logger functionality in itself enough? (I'm super happy to change it anyway)

pkg/daemon/update.go
```diff
@@ -690,6 +690,17 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	return nil
 }
 
+// RHEL 7.6 logger (util-linux) doesn't have the --journald flag
+func loggerWithoutJournalOption() bool {
+	loggerOutput, err := exec.Command("logger", "--journald").CombinedOutput()
```
cgwalters: I think we should use `logger --help`...as running this command just hangs for me; maybe it works because we're not providing stdin?

Or better, do:

```
if dn.OperatingSystem == machineConfigDaemonOSRHCOS {
    return true
```
?

pkg/daemon/update.go
```diff
@@ -701,15 +712,25 @@ func (dn *Daemon) getPendingState() (string, error) {
 		return "", err
 	}
 	if len(journalOutput) == 0 {
-		return "", nil
+		if !loggerWithoutJournalOption() {
```
cgwalters: note double negative here, probably better as `loggerSupportsJournal`

pkg/daemon/update.go
runcom: uhm, does it hang on RHCOS? works fine on rhel7.6 :thinking:

runcom: oh yeah I get it now thanks!

pkg/daemon/update.go
cgwalters: `OLD_LOGGER` seems a bit too generic...while it's unlikely, that string could appear in log messages from any othe process.

(Ideally...we'd have the MCD logging under `machine-config-daemon.service` that we create on the host.  That's another topic)

How about calling it `OPENSHIFT_MACHINE_CONFIG_DAEMON_LEGACY_LOG_HACK` ? :wink:

pkg/daemon/update.go

pkg/daemon/update.go

pkg/daemon/update.go
runcom: > How about calling it `OPENSHIFT_MACHINE_CONFIG_DAEMON_LEGACY_LOG_HACK` ?

:joy: I'll for it, that's a good one indeed to avoid collisions

pkg/daemon/update.go
```diff
@@ -701,15 +711,25 @@ func (dn *Daemon) getPendingState() (string, error) {
...
 	last := lines[len(lines)-1]
 	type journalMsg struct {
-		Message string `json:"MESSAGE,omitempty"`
-		// TODO(runcom): journal messages have a _BOOT_ID field, might use that
-		BootID  string `json:"BOOT_ID,omitempty"`
-		Pending string `json:"PENDING,omitempty"`
+		Message   string `json:"MESSAGE,omitempty"`
+		BootID    string `json:"BOOT_ID,omitempty"`
+		Pending   string `json:"PENDING,omitempty"`
+		OldLogger string `json:"OPENSHIFT_MACHINE_CONFIG_DAEMON_LEGACY_LOG_HACK,omitempty"` // unused today
```
kikisdeliveryservice: love this. :laughing:

runcom:
@vrutkovs just for awareness, not sure who the owner is, but scaleup is broken somehow

runcom:
aws limits.... will retest later...

pkg/daemon/update.go
```diff
@@ -734,7 +754,17 @@ BOOT_ID=%s
 PENDING=%d`, pendingStateMessageID, pending.GetName(), dn.bootID, isPending)))
 
 	logger.Stdin = &pendingState
-	return logger.CombinedOutput()
+	loggerOutput, err := logger.CombinedOutput()
```
cgwalters: I still think it'd be cleaner to call `loggerSupportsJournal` *once* or so on MCD startup rather than trying and failing each time.

runcom: oh I can definitely do that if it's for `loggerSupportsJournal`

runcom: should be fixed

runcom:
something super funny playing with this on a real cluster with scaleup:

```
I0511 18:59:39.538616   12394 daemon.go:860] Using current config on disk rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5
I0511 18:59:39.538704   12394 update.go:819] Using current config on disk rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5
...
I0511 18:59:39.561255   12394 update.go:171] Checking reconcilable for config rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5 to rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5
I0511 18:59:39.564968   12394 update.go:819] Starting update from rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5 to rendered-worker-d2ddc7c15c9eefdd6f354c9cae089bf5
```

So, for this PR the pending config isn't there somehow (I'm debugging that), besides, the recent config on disk fingerprint PR made it possible to move on anyway I guess. @cgwalters thoughts?

runcom:
ok, this is now stable in my testing against a real scaleup cluster with 3 rhcos masters + 2 centos 7 workers. We should run our e2e-aws-op in scaleup job though or we risk these kinds of regressions everytime

runcom:
/retest

The cluster and machineconfig operator go up in the scale up test but the job itself fails. This is not related to this change as it wasn't passing even before this (reason why it was never enabled by default)

runcom:
Scale-up passed this time..

ashcrow:
Let's strike while e2e is succeeding. @runcom is this ready for final review/merge?

runcom:
> Let's strike while e2e is succeeding. @runcom is this ready for final review/merge?

it is, if @cgwalters @kikisdeliveryservice @LorbusChris can ack on this it would be ready to merge

LorbusChris:
:cry: 

/test e2e-rhel-scaleup
/test e2e-aws

PR #736, opened 2019-05-10
server: Add optional auth token

cgwalters:
Since the bootstrap node serves Ignition to master, and today we
don't support scaling up/down masters, just disable access to
the master Ignition config in the in-cluster MCS by default.

### DIFF
--- a/pkg/server/cluster_server.go
+++ b/pkg/server/cluster_server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"github.com/pkg/errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
@@ -10,8 +11,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	rest "k8s.io/client-go/rest"
+	clientset "k8s.io/client-go/kubernetes"
 	clientcmd "k8s.io/client-go/tools/clientcmd"
 	clientcmdv1 "k8s.io/client-go/tools/clientcmd/api/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 )
@@ -22,6 +25,9 @@ const (
 	inClusterConfig = ""
 
 	bootstrapTokenDir = "/etc/mcs/bootstrap-token"
+
+	componentNamespace = "openshift-machine-config-operator"
+	ignitionAuth = "ignition-auth"
 )
 
 // ensure clusterServer implements the
@@ -29,6 +35,8 @@ const (
 var _ = Server(&clusterServer{})
 
 type clusterServer struct {
+	// kubeClient is used for the Ignition secret currently
+	kubeClient clientset.Interface
 	// machineClient is used to interact with the
 	// machine config, pool objects.
 	machineClient v1.MachineconfigurationV1Interface
@@ -48,8 +56,13 @@ func NewClusterServer(kubeConfig, apiserverURL string) (Server, error) {
 		return nil, fmt.Errorf("Failed to create Kubernetes rest client: %v", err)
 	}
 
+	kc, err := clientset.NewForConfig(restConfig)
+	if err != nil {
+		return nil, err
+	}
 	mc := v1.NewForConfigOrDie(restConfig)
 	return &clusterServer{
+		kubeClient:     kc,
 		machineClient:  mc,
 		kubeconfigFunc: func() ([]byte, []byte, error) { return kubeconfigFromSecret(bootstrapTokenDir, apiserverURL) },
 	}, nil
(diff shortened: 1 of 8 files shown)


runcom:
Code lgtm, guess this can't hurt BYO either now since we just support workers there. Looks sane

pkg/server/cluster_server.go
```diff
@@ -42,23 +52,61 @@ type clusterServer struct {
...
+		enableMaster:   enableMaster,
 	}, nil
 }
 
 // GetConfig fetches the machine config(type - Ignition) from the cluster,
 // based on the pool request.
-func (cs *clusterServer) GetConfig(cr poolRequest) (*ignv2_2types.Config, error) {
-	mp, err := cs.machineClient.MachineConfigPools().Get(cr.machineConfigPool, metav1.GetOptions{})
+func (cs *clusterServer) GetConfig(cr poolRequest, auth string) (*ignv2_2types.Config, error) {	
+	authSecretObj, err := cs.kubeClient.CoreV1().Secrets(componentNamespace).Get(ignitionAuth, metav1.GetOptions{})
```
kikisdeliveryservice: unit test failing here, I think:

```panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0xf8 pc=0xec8d91]

goroutine 85 [running]:
testing.tRunner.func1(0xc4203d9ef0)
	/usr/local/go/src/testing/testing.go:742 +0x29d
panic(0xfcf0e0, 0x18ee930)
	/usr/local/go/src/runtime/panic.go:502 +0x229
github.com/openshift/machine-config-operator/pkg/server.(*clusterServer).GetConfig(0xc420499f38, 0x1101419, 0x9, 0x0, 0x0, 0x0, 0x0, 0xf)
	/go/src/github.com/openshift/machine-config-operator/pkg/server/cluster_server.go:77 +0x41
github.com/openshift/machine-config-operator/pkg/server.TestClusterServer(0xc4203d9ef0)
	/go/src/github.com/openshift/machine-config-operator/pkg/server/server_test.go:175 +0x936
testing.tRunner(0xc4203d9ef0, 0x117d5a0)
	/usr/local/go/src/testing/testing.go:777 +0xd0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:824 +0x2e0
FAIL	github.com/openshift/machine-config-operator/pkg/server	0.039s```

Comment (cgwalters) [reply pkg/server/cluster_server.go]: Yeah, had forgotten to `git commit --amend` before pushing.  Fixed already in the latest :arrow_down:

**Comment by cgwalters** [comment]
Just for general info, I just discovered Ignition will do a retry loop infinitely on http status `>= 500` but if you give it a `403` it gives up, so you need to de-provision the machine (e.g. `oc -n openshift-machine-api delete machine/$x`) if you were testing auth.

`aws ec2 --region=us-east-2 get-console-output --instance-id i-0bc13fd82d10aee7a` is handy to debug things like this.

File: pkg/server/server_test.go [hunk 2]
```diff
@@ -174,7 +176,7 @@ func TestClusterServer(t *testing.T) {
 
 	res, err := csc.GetConfig(poolRequest{
 		machineConfigPool: testPool,
-	})
+	}, "")
```
kikisdeliveryservice: is this "" the problem?

cgwalters:
Just writing up some more steps for testing/using this.  After you do something like this:
`oc -n openshift-machine-config-operator create secret generic --from-literal=worker=$(pwgen 32 1) ignition-auth`

Your next step is to `oc -n openshift-machine-api edit secret/worker-user-data` - base64 decode it to get the JSON, then ensure that the Ignition request has the same auth key you generated in the secret, e.g.:

```
{"ignition":{"config":{"append":[{"source":"https://api-int.mycluster.example.com:22623/config/worker?auth=pxzbz5n2zgfcd4kff8krv78z8wwdndvj","verification":{}}]}, ...
```

pkg/server/server.go
```diff
@@ -31,10 +31,30 @@ type kubeconfigFunc func() (kubeconfigData []byte, rootCAData []byte, err error)
 // appenderFunc appends Config.
 type appenderFunc func(*ignv2_2types.Config) error
 
+// Error is returned by the GetConfig API
+type Error struct {
```
runcom: we might want to call it `GetConfigError` or something like that (doesn't need uppercase to be exported also, does it?)

kikisdeliveryservice: (@runcom I thought upper to be an exported type)

kikisdeliveryservice: like this naming suggestion btw.

kikisdeliveryservice:
aws route53 issues

/retest

cgwalters:
Now a much stronger elaboration of this PR would be for the MCO to support one-time-use tokens.

Rather than having a secret, an admin (or automation) would do e.g.:

```
oc -n openshift-machine-config-operator create secret generic --from-literal=auth=$(pwgen 32 1) ignition-auth-worker-us-east-2a-sks8v` (concatenating `ignition-auth-` with the machineAPI object name).

The MCS would check for this secret's existence, and if it existed, accept the request and then delete the secret.  This would also mean access to the EC2 metadata API from a pod couldn't gain a useful token.

File: pkg/server/server.go (hunk 3 above)
Comment (runcom) [reply pkg/server/server.go]: yeah, I realized only after commenting that I uppercased my suggestion 😄

**Comment by cgwalters** [comment]
OK, all tests passed on this one, confirming my manual testing before submitting the PR that nothing depends on access to the `/config/master` endpoint.   And we have the ability to gate access to `/worker`, but turning it on by default will require an installer PR.

I'd like to consider landing this as we're pretty sure it's not going to break anything, and gives us the mechanism to gate access to `/config/worker`.

**Comment by cgwalters** [comment]
Installer PR: https://github.com/openshift/installer/pull/1740

**Comment by runcom** [comment]
/approve

This lgtm, can this land separately from the installer PR? do we need a go-on from Auth team?

File: pkg/server/server.go (hunk 3 above)
Comment (ashcrow) [reply pkg/server/server.go]: Or possibly `configError` to be explicit.

File: pkg/server/cluster_server.go (hunk 1 above)
Comment (squeed) [inline pkg/server/cluster_server.go]: The ability to configure multiple secrets would make later rotation much easier.

File: pkg/server/cluster_server.go [hunk 4]
```diff
@@ -42,23 +52,61 @@ type clusterServer struct {
...
+		} else {
+			return nil, errors.Wrapf(err, "getting ignition-auth")
+		}
+	}
+
+	name := cr.machineConfigPool
+
+	// If there's a secret, require that it was passed as a query parameter.
+	if authEnabled {
+		authSecret := string(authSecretObj.Data[name])
```
squeed: If an administrator forgets to configure an authsecret for a new machine pool, this will fail open.

cgwalters: That's intentional for backcompat reasons.

crawford:
/hold

This doesn't but us much in terms of security and it makes disaster recovery more difficult.

cgwalters:
> This doesn't but us much in terms of security

If access to the EC2 metadata endpoint is shut off (which it needs to be anyways), or in cases where the bootstrap config isn't accessible at all, I think this is fairly strong security.  It *entirely* shuts down the problem.

How is this *not* buying much?

> and it makes disaster recovery more difficult.

Slightly...and if desired we can switch this to just landing code to enable auth tokens for both master/worker without actually enabling it by default.  In other words, drop the change to disable the master pool.

Further, keep in mind that this code is a step towards further [integration with machineAPI](https://github.com/openshift/machine-config-operator/pull/736#issuecomment-491442647).

cgwalters:
OK dropped the master pool change; this PR now is pure optional `auth` handling if the secret exists, otherwise it does nothing [per this comment](https://github.com/openshift/machine-config-operator/pull/736#discussion_r283448257).

kikisdeliveryservice:
uhoh, is there due to the branching that just happened?
`ould not resolve inputs: could not determine inputs for step [input:base]: could not resolve base image: imagestreamtags.image.openshift.io "4.2" not found`

cgwalters:
Closing in favor of https://github.com/openshift/machine-config-operator/pull/2223

PR #738, opened 2019-05-13
build-go.sh: use OS_GIT_COMMIT if available

sosiouxme:
This will be supplied in downstream builds to relay the commit from the source git repo.

Fixes: bz 1709365

**- How to verify it**
Check that the resulting OCP image built binary reports a commit from this github repo.

build-go.sh: use OS_GIT_COMMIT if available

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -20,7 +20,7 @@ if [ -z ${VERSION_OVERRIDE+a} ]; then
 	VERSION_OVERRIDE=$(git describe --abbrev=8 --dirty --always)
 fi
 
-HASH=$(git rev-parse --verify 'HEAD^{commit}')
+HASH=${OS_GIT_COMMIT:-$(git rev-parse --verify 'HEAD^{commit}')}
 
 GLDFLAGS+="-X ${REPO}/pkg/version.Raw=${VERSION_OVERRIDE} -X ${REPO}/pkg/version.Hash=${HASH}"
 


runcom:
/approve
/lgtm

Not sure the CI jobs here will trigger any testing for this, I assume you've tested it internally right

sosiouxme:
@runcom PTAL

BTW I would pick this to the release-4.1 branch but that looks like it has not been fast-forwarded in quite some time.

runcom:
> BTW I would pick this to the release-4.1 branch but that looks like it has not been fast-forwarded in quite some time.

I'm not sure branching happened yet, have it? :thinking:

sosiouxme:
that's sorta happening today... maybe...

sosiouxme:
This runs during image build, so OKD CI presumably will at least catch gross errors (but I made sure the syntax works...). Matching downstream code to insert the env vars is in testing.

smarterclayton:
/hold

We use GIT_COMMIT in origin, why would we have a different one here?

sosiouxme:
I am blind, release-4.1 is a FF of master. not master-4.1.

sosiouxme:
@smarterclayton 
I originally was going to use `OPENSHIFT_SOURCE_GIT_COMMIT` but that seemed overly specific, while `GIT_COMMIT` seems nonspecific enough to surprise someone who's already using it for something else. These will be applied in everything ART builds. But it is not tremendously important to me what we call it as long as no one is surprised.

cgwalters:
There is a [whole specification and standard](https://reproducible-builds.org/docs/source-date-epoch/) for a related variable `SOURCE_DATE_EPOCH` which we could also use.

cgwalters:
That said I don't care too much; I think the root fix here is to fold the MCO's images into one, *or* fix the dist-git so it can build multiple containers from one dist-git.

Happy to honor `GIT_COMMIT` (if it's already set in the ART system).

smarterclayton:
I was under the impression we were consistent about using GIT_COMMIT elsewhere.  If we were, then let's be consistent here.

runcom:
Opened this https://github.com/openshift/machine-config-operator/issues/739 as a follow up (which I'm not comfortable enough doing now, as trivial as it is)

sosiouxme:
/retest

I don't see that origin uses `GIT_COMMIT`. I see that it uses `OS_GIT_COMMIT`

```
grep -rI GIT_COMMIT hack/
hack/lib/build/version.sh:      if [[ -n ${OS_GIT_COMMIT-} ]] || OS_GIT_COMMIT=$("${git[@]}" rev-parse --short "HEAD^{commit}" 2>/dev/null); then
hack/lib/build/version.sh:              if [[ -n ${OS_GIT_VERSION-} ]] || OS_GIT_VERSION=$("${git[@]}" describe --long --tags --abbrev=7 --match 'v[0-9]*' "${OS_GIT_COMMIT}^{commit}" 2>/dev/null); then
[...]
hack/lib/constants.sh:  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/pkg/version.commitFromGit" "${OS_GIT_COMMIT}"))
hack/lib/constants.sh:  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/k8s.io/kubernetes/pkg/version.gitCommit" "${OS_GIT_COMMIT}"))
hack/lib/constants.sh:  ldflags+=($(os::build::ldflag "${OS_GO_PACKAGE}/vendor/k8s.io/client-go/pkg/version.gitCommit" "${OS_GIT_COMMIT}"))
hack/release.sh:export OS_GIT_COMMIT="${commit}" 
hack/release.sh:  hack/env OS_GIT_COMMIT="${commit}" make build-cross
hack/build-cross.sh:    echo "${OS_GIT_COMMIT}" > "${OS_OUTPUT_RELEASEPATH}/.commit"
hack/build-release.sh:os::build::environment::withsource "${container}" "${OS_GIT_COMMIT:-HEAD}"
hack/build-release.sh:echo "${OS_GIT_COMMIT}" > "${OS_OUTPUT_RELEASEPATH}/.commit"
[...]
```

...and I'm concerned that if we define that, it looks like the logic may expect a bunch of related things to be defined too.

sosiouxme:
Though looking a little more closely there... https://github.com/openshift/origin/blob/7e86a1a1c5dda216da22f950a8e139f294125ca2/hack/lib/build/version.sh#L31 ... it's probably fine.

sosiouxme:
@smarterclayton  ^^ `OS_GIT_COMMIT` is also the short commit, while I think we want the full commit here. So we'd need a new var name anyway?
@cgwalters although the epoch could be useful for showing build parity, it doesn't show up in `git ls-remote` so is harder to verify against the source. Also that's a bit more change than I think we want at this date :)

smarterclayton:
There is no assumption in openshift that requires it to be short.  Use OS_GIT_COMMIT

sosiouxme:
Changed to `OS_GIT_COMMIT`

cgwalters:
But since @smarterclayton added the `/hold` I think he needs to un-hold.  

Some days with all the holds it feels like...
HOLD

smarterclayton:
Note that this will need to be cherry-picked back to 4.1, use the cherry-pick bot!

PR #740, opened 2019-05-13
openstack: Add coredns, mdns-publisher and haproxy static pods

trown:
This is part of the work to remove the "service" VM from the OpenStack architecture. Instead of running coredns on this extra VM, we will run it in a static pod on all masters. We are using the mdns work from metal3.io in order to coordinate the DNS records on the masters.


templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,77 @@
...
+      volumes:
+      - name: resource-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/mdns"
+      - name: conf-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/mdns"
+      initContainers:
+      - name: render-config
+        image: quay.io/openshift/origin-node:latest
```
cgwalters: This will need to be added to `image-references` and come from the payload.

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,77 @@
...
+        resources: {}
+        volumeMounts:
+        - name: resource-dir
+          mountPath: "/etc/kubernetes/static-pod-resources"
+        - name: conf-dir
+          mountPath: "/etc/mdns"
+        imagePullPolicy: IfNotPresent
+      containers:
+      - name: mdns-publisher
+        image: quay.io/openshift-metalkube/mdns-publisher:latest
```
cgwalters: Ditto.

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,77 @@
...
+          SUBNET_CIDR="$(ip addr show | grep -v "scope host" | grep -Po 'inet \K[\d.]+/[\d.]+' | head -n1)"
+          PREFIX="${SUBNET_CIDR#*/}"
+          NON_VIRTUAL_IP="${SUBNET_CIDR%/*}"
+          MASTER_HOSTNAME="$(hostname -s | cut -d '-' -f 4)-$(hostname -s | cut -d '-' -f 5).local."
+          ETCD_HOSTNAME="$(echo "$MASTER_HOSTNAME" | sed 's;master;etcd;')"
+          export MASTER_HOSTNAME
+          export ETCD_HOSTNAME
+          export NON_VIRTUAL_IP
+          export DOMAIN
+          export CLUSTER_NAME
```
cgwalters: Minor but you can just prefix the vars above with `export` too to reduce duplication.

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,77 @@
...
+          /usr/libexec/platform-python -c "from __future__ import print_function
```
cgwalters: I think the more correct approach is for the image to install the Python it wants.

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - "/bin/bash"
+        - "-c"
+        - |
+          #/bin/bash
+          set -ex
+
+          CLUSTER_NAME={{(split "." .EtcdDiscoveryDomain)._0}}
+          DOMAIN={{.EtcdDiscoveryDomain}}
+          API_VIP="$(dig +noall +answer "api.${DOMAIN}" | awk '{print $NF}')"
+          SUBNET_CIDR="$(ip addr show | grep -v "scope host" | grep -Po 'inet \K[\d.]+/[\d.]+' | head -n1)"
```
cgwalters: This feels a bit hacky...

trown:
/hold

need to properly publish the coredns and mdns-publisher images this patch uses

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
trown: Done

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
trown: I agree. I added a TODO to come up with a better method of getting these.

templates/master/00-master/openstack/files/openstack-coredns.yaml
```diff
@@ -0,0 +1,86 @@
...
+  inline: |
+    kind: Pod
+    apiVersion: v1
+    metadata:
+      name: coredns
+      namespace: kube-system
+      creationTimestamp:
+      deletionGracePeriodSeconds: 65
+      labels:
+        app: kni-infra-mdns
```
abhinavdahiya: nit: remove `kni`

templates/master/00-master/openstack/files/openstack-coredns.yaml
```diff
@@ -0,0 +1,86 @@
...
+        imagePullPolicy: IfNotPresent
+      containers:
+      - name: coredns
+        securityContext:
+          privileged: true
+        image: quay.io/openshift-metalkube/coredns-mdns:latest
+        args:
+        - "--conf"
+        - "/etc/coredns/Corefile"
+        resources:
```
abhinavdahiya: These requests are pretty high.

tomassedovic:
/retitle openstack: Add coredns, mdns-publisher and haproxy static pods

tomassedovic:
/retest

This worked for me manually on the master. Looks more like a timeout on the openstack job.

tomassedovic:
@cgwalters @runcom would you mind having a look at this PR?

It's the OpenStack equivalent to: https://github.com/openshift/machine-config-operator/pull/795

We've tested it locally (with and without the corresponding installer PR https://github.com/openshift/installer/pull/1959), it passes the CI and OpenStack has a FFE for this work. Is there anything else you'd like us to do?

pkg/controller/template/render.go
```diff
@@ -180,6 +180,11 @@ func generateMachineConfigForName(config *RenderConfig, role, name, templateDir,
 	platformDirs := []string{}
 	// Loop over templates/common which applies everywhere
 	for _, dir := range []string{platformBase, platform} {
+		// Bypass OpenStack template rendering until
+		// https://github.com/openshift/installer/pull/1959 merges
```
cgwalters: Hum...so if I understand this correctly, this PR is adding all of the code, and then we need to get the installer PR in first, then we'll do a PR to drop these conditionals?

cgwalters:
I find this part encouraging:

> These differences are not fundamental to OpenStack and we will be
looking at aligning more closely with the Baremetal provider in the
future.

Bigger picture, I wonder whether we should consider having the cluster own DNS by default across the board in a uniform way.  That's another topic though.

The commit message in https://github.com/openshift/installer/pull/1959 is great, very helpful!

pkg/controller/template/render.go
tomassedovic: Yes that's the intention. The installer PR depends on the changes there, but this PR depends on the changes in the installer PR so this let's us break that cycle.

The conditionals are isolated in a single commit we can revert afterwards.

PR #741, opened 2019-05-13
operator: remove the cluster role binding that auto approves all node client CSRs

abhinavdahiya:
Because of the existense of this ClusterRoleBinding, kube-controller-manager approves all node client CSRs requests and therefore
allowing wrongful actors to get node level client priviledges. [cluster-machine-approver][1] takes on the role of approving only certain node client CSR requests.

[installer][2] takes the responsibility for approving CSRs during bootstrapping.

[1]: https://github.com/openshift/cluster-machine-approver/pull/26
[2]: https://github.com/openshift/installer/pull/1747


cgwalters:
When investigating this a bit, I noticed that the [openshift-apiserver appears to install roles](https://github.com/openshift/origin/blob/7e86a1a1c5dda216da22f950a8e139f294125ca2/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go#L495) which ultimately calls into [this code](https://github.com/openshift/origin/blob/7e86a1a1c5dda216da22f950a8e139f294125ca2/pkg/cmd/server/bootstrappolicy/policy.go#L809) AFAICS.

cgwalters:
/approve
Since I generally trust you :wink: and I personally don't know much about this code, and I think e2e covers us generally here.

cgwalters:
Without diving deep into the KCM code, it feels like it'd be cleaner if we e.g. added a `--disable-node-bootstrap` CLI flag or something to entirely turn off that bit if we're always relying on cluster-machine-approver, rather than doing it indirectly via RBAC.  But just an offhand comment.

abhinavdahiya:
> When investigating this a bit, I noticed that the [openshift-apiserver appears to install roles](https://github.com/openshift/origin/blob/7e86a1a1c5dda216da22f950a8e139f294125ca2/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go#L495) which ultimately calls into [this code](https://github.com/openshift/origin/blob/7e86a1a1c5dda216da22f950a8e139f294125ca2/pkg/cmd/server/bootstrappolicy/policy.go#L809) AFAICS.

Both the links are code-paths not utilized in Openshift 4 and exist for Openshift 3.

We moved closer to https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/ for Openshift 4.

kikisdeliveryservice:
this makes sense to me, but ill leave it to someone else to lgtm

cc: @runcom

runcom:
Lgtm as well, I believe Alex is going to look at this PR and installer one to also say it's good

cgwalters:
> Both the links are code-paths not utilized in Openshift 4 and exist for Openshift 3.

OK.  Looks like that one would be fun to delete.

knobunc:
/hold
Until we get https://github.com/openshift/installer/pull/1747 merged, we need to hold this, otherwise it's good to go.

enj:
/retest

Curious how far this will get.

runcom:
installer PR merged

/retest

knobunc:
/hold cancel
Since the dependent PR merged.

runcom:
retesting cause I suppose the installer PR hasn't been picked up by CI by the time I re-triggered tests.

cc @abhinavdahiya this should just pass now that installer PR is in right?

PR #744, opened 2019-05-14
build-go.sh: OS_GIT_COMMIT => SOURCE_GIT_COMMIT

sosiouxme:
This will be supplied in downstream builds to relay the commit from the source git repo.

Fixes: bz 1709365

**- How to verify it**
Check that the resulting OCP image built binary reports a commit from this github repo.

build-go.sh: use SOURCE_GIT_COMMIT if available

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -20,7 +20,7 @@ if [ -z ${VERSION_OVERRIDE+a} ]; then
 	VERSION_OVERRIDE=$(git describe --abbrev=8 --dirty --always)
 fi
 
-HASH=${OS_GIT_COMMIT:-$(git rev-parse --verify 'HEAD^{commit}')}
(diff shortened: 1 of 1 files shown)


cgwalters:
I liked the commit message on the other one better, and we should have been using cherrypick right?

Also while we'e reworking this I probably would have preferred adding `SOURCE_DATE_EPOCH` handling *in addition*.  But eh.

/approve
/lgtm

sosiouxme:
I can probably add the `EPOCH` envvar in doozer pretty easily and you can use it at will.

wking:
What happened to `OS_GIT_COMMIT`, after #738 [explicitly pivoted to it][1]?

[1]: https://github.com/openshift/machine-config-operator/pull/738#issuecomment-491963277

sosiouxme:
@wking builds from origin didn't like it, see #aos-art

PR #746, opened 2019-05-14
daemon: Drop unnecessary logSystem

cgwalters:
We show this log in the normal case where the file exists.
I don't think we need to `logSystem` this anyways.


runcom:
/lgtm

yes, this isn't needed, I added it initially cause I wasn't sure we were getting this code path in that nasty bug

PR #749, opened 2019-05-14
controller: Just sanity check templates, don't copy all of them

cgwalters:
We haven't changed how the renderer works in forever, but a lot
of people keep changing the templates.  This causes pointless churn
in the test data and makes changing the templates unnecessarily
painful.

Further, I want to change how the rendering works to inherit
from a common base, and having the unit test code also parsing
the templates is problematic.

Let's just sanity check that we have the pull secret and kubelet
unit.


runcom:
love this

kikisdeliveryservice:
This makes a lot of sense to me!

pkg/controller/template/render_test.go
```diff
@@ -282,6 +275,11 @@ func TestGenerateMachineConfigs(t *testing.T) {
 			t.Fatalf("failed to generate machine configs: %v", err)
 		}
 
+		foundPullSecretMaster := false
+		foundPullSecretWorker := false
+		foundKubeletUnitMaster := false
+		foundKubeletUnitWorker := false
```
runcom: this is the most stupid nits every but declaring a boolean in Golang is the same as setting it to false so `var found......, found .... , ..... bool` is the same here (and this nits can live here for the time being, just wanted to point it out :D

PR #752, opened 2019-05-14
daemon: Log high level diff between MCs

cgwalters:
It's really useful to see in the logs explicitly when the osImageURL
changes and the files change (or didn't).

Eventually we should try to get a real diff and consider only
applying that.


cgwalters:
This looks like
`I0514 22:08:28.665320   22488 update.go:852] Starting update from rendered-worker-c9bd3d73209d4ac6a484562d34c307ef to rendered-worker-6371e8bba4b54530674d0919b62c17fb: &{osUpdate:false passwd:false files:true units:false}`

runcom:
That looks fine for now I guess, we could try to make that more human friendly tho

/approve
/lgtm

PR #753, opened 2019-05-15
pkg/controller: calculate fault tolerance for the master pool

runcom:
**- What I did**

Calculate the maxUnavailable for the master pool based also on the etcd fault tolerance and thus removing the hardcoded 1.

**- How to verify it**

added units

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -471,7 +471,7 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	maxunavail, err := maxUnavailable(pool, nodes)
 	if err != nil {
 		return err
-	}	
+	}
 
 	candidates := getCandidateMachines(pool, nodes, maxunavail)
 	for _, node := range candidates {
@@ -554,22 +554,25 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
 }
 
 func maxUnavailable(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) (int, error) {
-	// Hardcoded for now per discussion in
-	// https://github.com/openshift/machine-config-operator/pull/701
-	if pool.Name == "master" {
-		return 1, nil
-	}
 	intOrPercent := intstrutil.FromInt(1)
 	if pool.Spec.MaxUnavailable != nil {
 		intOrPercent = *pool.Spec.MaxUnavailable
 	}
-
 	maxunavail, err := intstrutil.GetValueFromIntOrPercent(&intOrPercent, len(nodes), false)
 	if err != nil {
 		return 0, err
 	}
 	if maxunavail == 0 {
 		maxunavail = 1
+	}
+	if pool.Name == "master" {
+		// calculate the fault tolerance dynamically for the master pool
+		// to avoid risking losing etcd quorum.
+		tolerance := len(nodes) - ((len(nodes) / 2) + 1)
+		if maxunavail > tolerance {
+			glog.Warningf("Refusing to honor master pool maxUnavailable %d to prevent losing etcd quorum, using %d instead", maxunavail, tolerance)
+			return tolerance, nil
+		}
 	}
 	return maxunavail, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -554,22 +554,25 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
...
-
 	maxunavail, err := intstrutil.GetValueFromIntOrPercent(&intOrPercent, len(nodes), false)
 	if err != nil {
 		return 0, err
 	}
 	if maxunavail == 0 {
 		maxunavail = 1
 	}
+	if pool.Name == "master" {
+		// calculate the fault tolerance dinamycally for the master pool
```
kikisdeliveryservice: "dynamically"

pkg/controller/node/node_controller.go
```diff
@@ -554,22 +554,25 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
...
 	}
 	if maxunavail == 0 {
 		maxunavail = 1
 	}
+	if pool.Name == "master" {
+		// calculate the fault tolerance dynamically for the master pool
+		// to avoid risking losing etcd quorum.
+		tolerance := len(nodes) - ((len(nodes) / 2) + 1)
+		if maxunavail > tolerance {
+			glog.Warningf("Refusing to honor master pool maxUnavailable %d to prevent losing etcd quorum, using %d instead", maxunavail, tolerance)
```
kikisdeliveryservice: just to double check:
wanting a higher threshold is more likely to lead to quorum loss because that's more nodes that have to stay available right? so we are sticking to the threshold that is the min needed to maintain a quorum?

pkg/controller/node/node_controller.go
```diff
@@ -554,22 +554,25 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
 }
 
 func maxUnavailable(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) (int, error) {
-	// Hardcoded for now per discussion in
-	// https://github.com/openshift/machine-config-operator/pull/701
-	if pool.Name == "master" {
-		return 1, nil
-	}
 	intOrPercent := intstrutil.FromInt(1)
```
kikisdeliveryservice: random Q but why are we using IntOrString types?  bc people might say: 1 (node) or 20%?
Like I'm a little confused by it.  Maybe we could add some comments?

kikisdeliveryservice: I see that I was right by looking at the test, but I think a 1 line here would be good for when people skim later. :smile:

pkg/controller/node/node_controller.go
runcom: Yes exactly

pkg/controller/node/node_controller.go
runcom: > Maybe we could add some comments?

Code is pretty explicit, maybe we should add some examples around that in docs/ or examples/

pkg/controller/node/node_controller.go
```diff
@@ -554,22 +554,25 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
...
 	if err != nil {
 		return 0, err
 	}
 	if maxunavail == 0 {
 		maxunavail = 1
 	}
+	if pool.Name == "master" {
+		// calculate the fault tolerance dynamically for the master pool
+		// to avoid risking losing etcd quorum.
+		tolerance := len(nodes) - ((len(nodes) / 2) + 1)
```
cgwalters: While we're here, let's also warn if the number of nodes isn't odd? 
https://github.com/etcd-io/etcd/blob/master/Documentation/faq.md#what-is-failure-tolerance

runcom: :thinking: should we warn here? isn't that something well documented in the etcd docs? having an even number doesn't mean you're doing something wrong per-se, just that you need a wider quorum for less total nodes and no real gain

runcom: just to share more on my take: I believe that if someone is looking for a warning in the MCC logs, they already have a cluster up and running which might be not perfectly configured with an odd number of masters/etcd. But the warning, in general, is a good point, maybe it should be something in the installer so that people trying to use 4 masters are warned that the size isn't _optimal_.

PR #755, opened 2019-05-15
 server: Add a log message when a pool is requested

cgwalters:
This is very useful for debugging, and is prep for adding
further auditing.


cgwalters:
(Also snuck in another small fix here)

PR #759, opened 2019-05-15
WIP: server: Add /pointerconfig API

cgwalters:
This is prep for fixing
https://github.com/openshift/machine-config-operator/issues/683

Today the installer generates this "pointer Ignition" which ends up
as e.g. user-data in AWS, which is just a pair of (MCS URL, root CA).
We need to do this because of size limitations on AWS user data.

It makes a lot of sense for the MCO to be in control of generating
the pointer ignition config too, as it helps centralize knowledge
of Ignition.

Ultimately, the goal here is tighter integration between machineAPI
and the the MCO; e.g. the machineAPI asks to generate the pointer
Ignition config, rather than having it stored as a static secret.
This could then unlock the ability for the MCO to inject e.g.
one time auth tokens.

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -7,6 +7,8 @@ import (
 	"path"
 
 	"github.com/golang/glog"
+
+	ignv2_2types "github.com/coreos/ignition/config/v2_2/types"
 )
 
 type poolRequest struct {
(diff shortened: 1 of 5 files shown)


cgwalters:
Changing this to WIP as I'm not sure this is exactly how we should do this, but I wanted to experiment with what this would look like, and it was simple enough.

kikisdeliveryservice:

Feel free to reopen.

PR #762, opened 2019-05-15
 Add KernelArguments to MachineConfig

cgwalters:
We need to support things like adding `nosmt` to the kernel
commandline, for users that want to choose security over performance;
this is particularly relevant for multi-tenant clusters.
There are a lot of other "tuning" features that are expressed as
kernel cmdline args; down the line we want `node-tuning-operator`
to write something like `machineconfig/05-node-tuning-operator`.

Per discussion in the PR, we down the line probably want to add
an `operatingsystem` type or so that would be a high level wrapper
for this, but today everything else is expressed in `MachineConfig`,
so we need to extend this type.

There have also been some discussions about extending Ignition,
but that's not likely to land soon, and even if it did we still
need to handle "early pivot" pulling `machine-os-content`, and
scoping both OS updates and kargs into Ignition is too far out.

This is the first step which will support "day 2" configuration.
After this lands, we have further work to have the MCS provide
`MachineConfig` and not just Ignition, so that we can replace
the "early pivot" with MCD code.

Closes: https://github.com/openshift/machine-config-operator/issues/388

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -11,7 +11,8 @@ import (
 
 // MergeMachineConfigs combines multiple machineconfig objects into one object.
 // It sorts all the configs in increasing order of their name.
-// It uses the Ign config from first object as base and appends all the rest.
+// It uses the Ignition config from first object as base and appends all the rest.
+// Kernel arguments are concatenated.
 // It uses only the OSImageURL provided by the CVO and ignores any MC provided OSImageURL.
 func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineConfig {
 	if len(configs) == 0 {
@@ -23,10 +24,15 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	for idx := 1; idx < len(configs); idx++ {
 		outIgn = ign.Append(outIgn, configs[idx].Spec.Config)
 	}
+	kargs := []string{}
+	for _, cfg := range configs {
+		kargs = append(kargs, cfg.Spec.KernelArguments...)
+	}
 
 	return &MachineConfig{
 		Spec: MachineConfigSpec{
 			OSImageURL: osImageURL,
+			KernelArguments: kargs,
 			Config:     outIgn,
 		},
 	}
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -229,6 +229,8 @@ type MachineConfigSpec struct {
 	OSImageURL string `json:"osImageURL"`
 	// Config is a Ignition Config object.
 	Config igntypes.Config `json:"config"`
+
+	KernelArguments []string `json:"kernelArguments"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -233,6 +233,11 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 		}
 	}()
 
+	// kargs
+	if err := dn.updateKernelArguments(oldConfig, newConfig); err != nil {
+		return err
+	}
+
 	return dn.updateOSAndReboot(newConfig)
 }
 
@@ -242,6 +247,7 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 // improved logging.
 type MachineConfigDiff struct {
 	osUpdate bool
+	kargs    bool
 	passwd   bool
 	files    bool
 	units    bool
@@ -355,6 +361,7 @@ func Reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*MachineConfigDif
 	glog.V(2).Info("Configs are reconcilable")
 	return &MachineConfigDiff {
 		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
+		kargs: !reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments),
 		passwd: passwdChanged,
 		files: !reflect.DeepEqual(oldIgn.Storage.Files, newIgn.Storage.Files),
 		units: !reflect.DeepEqual(oldIgn.Systemd.Units, newIgn.Systemd.Units),
@@ -380,6 +387,49 @@ func verifyUserFields(pwdUser igntypes.PasswdUser) error {
 		return errors.New("ignition passwd user section contains unsupported changes: user must be core and have 1 or more sshKeys")
 	}
 	return nil
+}
+
+// generateKargsCommand performs a diff between the old/new MC kernelArguments,
+// and generates the command line arguments suitable for `rpm-ostree kargs`.
+// Note what we really should be doing though is also looking at the *current*
+// kernel arguments in case there was drift.  But doing that requires us knowing
+// what the "base" arguments are.  See https://github.com/ostreedev/ostree/issues/479
+func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
+	oldKargs := make(map[string]bool)
+	for _, arg := range oldConfig.Spec.KernelArguments {
+		oldKargs[arg] = true
+	}
+	newKargs := make(map[string]bool)
+	for _, arg := range newConfig.Spec.KernelArguments {
+		newKargs[arg] = true
+	}
+	cmdArgs := []string{}
+	for _, arg := range oldConfig.Spec.KernelArguments {
+		if !newKargs[arg] {
+			cmdArgs = append(cmdArgs, "--delete=" + arg)
+		}
+	}
+	for _, arg := range newConfig.Spec.KernelArguments {
+		if !oldKargs[arg] {
+			cmdArgs = append(cmdArgs, "--append=" + arg)
+		}
+	}
+	return cmdArgs
+}
+
+// updateKernelArguments adjusts the kernel args
+func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	diff := generateKargsCommand(oldConfig, newConfig)
+	if len(diff) == 0 {
+		return nil
+	}
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kargs on non-RHCOS nodes is not supported: %v", diff)
+	}
+
+	args := append([]string{"kargs"}, diff...)
+	dn.logSystem("Running rpm-ostree %v", args)
+	return exec.Command("rpm-ostree", args...).Run()
 }
 
 // updateFiles writes files specified by the nodeconfig to disk. it also writes
--- a/docs/MachineConfiguration.md
+++ b/docs/MachineConfiguration.md
@@ -48,11 +48,9 @@ type MachineConfig struct {
 }
 
 type MachineConfigSpec struct {
-    // OSImageURL specifies the remote location that will be used to
-    // fetch the OS. This must be in the canonical $name@$digest format.
-    OSImageURL string `json:"osImageURL"`
     // Config is a Ignition Config object.
     Config ign.Config `json:"config"`
+    KernelArguments []string `json:"kernelArguments"`
 }
 ```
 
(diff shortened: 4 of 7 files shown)


pkg/daemon/update.go
```diff
@@ -382,6 +389,35 @@ func verifyUserFields(pwdUser ignv2_2types.PasswdUser) error {
...
+			cmdArgs = append(cmdArgs, "--append=" + arg)
+		}
+	}
+	return cmdArgs
+}
+
+// updateKernelArguments adjusts the kernel args
+func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	args := append([]string{"kargs"}, generateKargsCommand(oldConfig, newConfig)...)
+	return exec.Command("rpm-ostree", args...).Run()
```
runcom: need a filter for !rhcos?

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -229,6 +229,8 @@ type MachineConfigSpec struct {
 	OSImageURL string `json:"osImageURL"`
 	// Config is a Ignition Config object.
 	Config ignv2_2types.Config `json:"config"`
+
+	KernelArguments []string `json:"kernelArguments"`
```
runcom: wondering if we need to update DeepCopy as well

cgwalters:
Updated and tested with this MC:

```
kind: MachineConfig
metadata:
  labels:
  name: 05-walters-mctest-worker-nosmt
spec:
  config:
    ignition:
      version: 2.2.0
  kernelArguments:
    - nosmt
```

And I correctly get `nosmt` in the kargs on my worker nodes :tada:

pkg/daemon/update.go
```diff
@@ -382,6 +389,40 @@ func verifyUserFields(pwdUser ignv2_2types.PasswdUser) error {
 	return nil
 }
 
+func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
```
runcom: this func is super cool, it could easily be extended to support fips for instance as a last case in the routine itself

pkg/daemon/update.go
```diff
@@ -382,6 +389,40 @@ func verifyUserFields(pwdUser ignv2_2types.PasswdUser) error {
...
+	for _, arg := range newConfig.Spec.KernelArguments {
+		newKargs[arg] = true
+	}
+	cmdArgs := []string{}
+	for arg := range oldKargs {
+		if !newKargs[arg] {
+			cmdArgs = append(cmdArgs, "--delete=" + arg)
+		}
+	}
+	for arg := range newKargs {
```
runcom: just a note that ranging over a map in golang isn't consistent wrt ordering to that `cmdArgs` can have a different kargs order than what we have in the newConfig.Spec.KernelArguments. If it's important we should account for that

cgwalters: Hmm...I *think* this could only matter if one had multiply specified/conflicting arguments, like `mem=4096 mem=8192` or something...I think like most argument parsers the kernel is "last one wins" so indeed in that scenario it could be unpredictable.

I'd probably say "just don't do that"?  But we can also try to handle it.

runcom: `console=` is also another example where ordering matters right? handling it would be easy also, we're not expecting huge kargrs lists. We can use a slice with more logic instead of a map[string]bool to maintain consisten order (again, the kargs slice isn't huge so we don't really care about O(1) when checking if an arg is there or not, right?)

cgwalters: I reworked this to iterate explicitly over the new args.  Though probably the testing here for order is better done in an e2e fashion because there's a lot of code in the rpm-ostree/ostree stack here processing this too.

ashcrow:
Looks good so far! When this makes it in let's not forget to drop the `pivot` interface.

cgwalters:
> When this makes it in let's not forget to drop the pivot interface.

Yeah, though we'll need to ratchet by landing this, then switch the installer over to use it, then we can drop the pivot component.

And...hum actually this goes straight into https://github.com/coreos/ignition-dracut/issues/81 - we're served Ignition, not MachineConfig, so we won't have `KernelArguments`.  But we really want to handle those kargs during the early pivot so we're not rebooting twice.

Hmm.  Hmm.  I think what we really want to do is have a mechanism to "invert" things and inject the extra stuff (today `osImageURL`, w/this PR also `kernelArguments`) from `MachineConfig` into the Ignition, and then have a host side service that processes the extra bits.

This would really be moving `pivot` to be "machineconfig handling for updates kernel".  I think the simplest thing may be that we actually move the `pivot` code into the MCD, and when we go to build RHCOS we extract the binary from the MCD container or so, rather than building an RPM.  (That entangles the build systems in an interesting way though)

ashcrow:
```
Or: we make this bit common between FCOS/RHCOS and add kargs into Ignition.
```

I feel like that would be a big win and help bring the two closer together if it's feasible. /cc @bgilbert @ajeddeloh @dustymabe

bgilbert:
For Ignition kargs discussion, see https://github.com/coreos/ignition-dracut/issues/81.  Note that it currently presumes a mechanism along the lines of https://github.com/ostreedev/ostree/issues/479#issuecomment-489768307.

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -23,10 +24,15 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	for idx := 1; idx < len(configs); idx++ {
 		outIgn = ign.Append(outIgn, configs[idx].Spec.Config)
 	}
+	kargs := []string{}
+	for _, cfg := range(configs) {
```
runcom: you don't need parenthesis around `config` - also, why this loop here? `configs` and `cfg.Spec.KernelArguments` are slides so order is maintained anyway (and `generateKargsCommand` now loops over slices as well so order is maintained as well and we're still O(1) on checks).

cgwalters:
Oh hey cool, this passed the new e2e test I added! 

```
--- PASS: TestKernelArguments (189.01s)
        mcd_test.go:296: Created kargs-81bd2992-78dc-11e9-b495-0a58ac10166e
        mcd_test.go:115: Pool worker has rendered config kargs-81bd2992-78dc-11e9-b495-0a58ac10166e with rendered-worker-431eab0a737953d758eacf1c99c1b9f1 (waited 2.032938681s)
        mcd_test.go:139: Pool worker has completed rendered-worker-431eab0a737953d758eacf1c99c1b9f1 (waited 2m52.017149132s)
        mcd_test.go:327: Node ip-10-0-131-46.ec2.internal has expected kargs
        mcd_test.go:327: Node ip-10-0-143-244.ec2.internal has expected kargs
        mcd_test.go:327: Node ip-10-0-147-161.ec2.internal has expected kargs
```

But it failed units...looking at that.

Will look at the other review comments here too.

But we need to close the debate pretty soon - are we all OK with extending `MachineConfig` in this fashion, or do we want to try creating a new `operatingsystem.openshift.io` CRD or something?

runcom:
> But we need to close the debate pretty soon - are we all OK with extending `MachineConfig` in this fashion, or do we want to try creating a new `operatingsystem.openshift.io` CRD or something?

This is tricky, we got so may inputs and ideas (at a point, we wanted also something on top of MachineConfig to be more sugar to allow for a nice upgrade to ignv2s3). The CRD to allow us `oc edit kargs` was from the beginning an amazing view and idea (`oc edit fips` anyone?).

My preference would go to a new CRD though (`operatingsystem` sounds perfect and extensible/reusable) and it follows what we did with KubeletConfig and ContainerRuntimeConfig (both of them eventually create MachineConfigs so we could also use the same pattern here somehow?). Also, we could still have a nice CRD on top of the MachineConfig and leave the implementation in this PR to land as is now.

cgwalters:
> My preference would go to a new CRD though (operatingsystem sounds perfect and extensible/reusable) and it follows what we did with KubeletConfig and ContainerRuntimeConfig (both of them eventually create MachineConfigs so we could also use the same pattern here somehow?). 

Yeah...what has me hesitating about that is all the copy-paste involved in the "sub-controllers".  But eh, we can figure that out

> Also, we could still have a nice CRD on top of the MachineConfig and leave the implementation in this PR to land as is now.

That's an interesting point.  I had thought that for this we'd "bypass" `MachineConfig` so we wouldn't have the options twice, but that would involve a bigger rearchitecting.

Maybe there aren't that many special things that *aren't* basically files (i.e. Ignition), so we won't have to duplicate too much...I am actually having a bit of trouble thinking of anything besides kargs and osimageurl that are special.

runcom:
> I am actually having a bit of trouble thinking of anything besides kargs and osimageurl that are special.

sebooleans? we had a doc at some point 🤔

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
cgwalters: Dropped the parens.  The loop is to merge (concatenate) the kargs, just like we're merging together the Ignition specs right?

cgwalters:
> sebooleans? 

Oh yeah that's a good one.

As far as other things like timeservers - this topic is kind of more about things that can't obviously "compile down" to just writing files or changing kernel args.  Timeservers should basically just be writing a file.  I think FIPS is potentially a good example as you need to run this `update-crypto-policy` command.  But on the other hand...we can write a systemd unit that does it based on a file (or perhaps better, ship the unit with RHCOS by default).

runcom:
> I think FIPS is potentially a good example as you need to run this `update-crypto-policy` command. But on the other hand...we can write a systemd unit that does it based on a file (or perhaps better, ship the unit with RHCOS by default).

oh yeah, FIPS is the follow up of this anyway (I'm ok on a pre-shipped unit as well if it works for RHCOS, but exercising it w/o a unit would be nice also).

Ok, let's sum this up, I believe we can _overload_ MachineConfig with Kargs|FIPS|Sebooleans and abstract those out with CRDs in follow ups if we really want to (even before next release for kargs|fips).
The MC object is still the way to let MCD know what it has to do to when reconciling to a given config. Am I missing something? 🤔 even if we want sugar on top (or on the side), the MC is still the last unit in the chain that the MCD understands.

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
runcom: > The loop is to merge (concatenate) the kargs, just like we're merging together the Ignition specs right?

damn I must be blind, sorry for the noise

pkg/apis/machineconfiguration.openshift.io/v1/types.go
cgwalters: Just checked, our current `MachineConfig` deepcopy is implemented manually, but it already recurses on `Spec` so that covers all of this.

pkg/daemon/update.go
cgwalters: There are libostree-level issues to fix here https://github.com/ostreedev/ostree/issues/1859
Eventually we'll have to get this right through the whole stack, but I think what we're doing here in the MCD is OK for now.

ashcrow:
```
The container could not be located when the pod was terminated
May 20 14:21:18.352 I test="[Disruptive] Cluster upgrade should maintain a functioning cluster [Feature:ClusterUpgrade] [Suite:openshift] [Serial]" failed
```

cgwalters:
The upgrade failure seems unrelated, seeing that in a few other jobs in https://prow.svc.ci.openshift.org/?job=*e2e-aws-upgrade*&state=failure
/retest

cgwalters:
OK, updated the commit message, and also updated the `MachineConfig` docs a bit.  (I dropped the `osImageURL` stuff mostly from the user-facing docs since...they shouldn't/can't-really touch that)

I think this is probably good to go :fast_forward: ?

runcom:
Uhm, looks like upgrade is stuck from e2e?

runcom:
Guess this is really good now, @ashcrow @miabbott this is a prereq for FIPS as well so we now have something to test with

/lgtm

pkg/daemon/update.go
```diff
@@ -382,6 +389,49 @@ func verifyUserFields(pwdUser igntypes.PasswdUser) error {
...
+}
+
+// updateKernelArguments adjusts the kernel args
+func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	diff := generateKargsCommand(oldConfig, newConfig)
+	if len(diff) == 0 {
+		return nil
+	}
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kargs on non-RHCOS nodes is not supported: %v", diff)
```
runcom: now that I'm looking at this, should we just return nicely as we do for `updateOS` :thinking:

pkg/daemon/update.go
```diff
@@ -233,6 +233,11 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 		}
 	}()
 
+	// kargs
+	if err := dn.updateKernelArguments(oldConfig, newConfig); err != nil {
+		return err
+	}
```
runcom: and I've now found that we need to roll this back if we fail :smile:

pkg/daemon/update.go
runcom: https://github.com/openshift/machine-config-operator/pull/824

pkg/daemon/update.go
runcom: https://github.com/openshift/machine-config-operator/pull/824

PR #765, opened 2019-05-16
test/e2e: More refactoring

cgwalters:
Rather than poll all of the daemons, add a helper that waits
for a pool to complete a config.

One of our tests walks over the MCDs, change it to just assert
on all of the nodes.

The SSH test can also just wait for a pool and then `rsh` to
each node.

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
+	"github.com/stretchr/testify/assert"
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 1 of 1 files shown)


test/e2e/mcd_test.go
```diff
@@ -136,28 +157,16 @@ func TestMCDeployed(t *testing.T) {
...
-					if len(visited) == len(nodes) {
-						return true, nil
-					}
-					continue
-				}
-			}
-			return false, nil
-		}); err != nil {
-			t.Errorf("machine config didn't result in file being on any worker: %v", err)
+		if err := waitForPoolComplete(cs, "worker", newMCName); err != nil {
```
runcom: isn't thi racy if we jump here before the pool starts rolling? I believe that's why I went node by node confirming current is there and of the new pool

runcom: ok, you added a check below but this may flake then?

cgwalters: Just above that we do `waitForRenderedConfig` right?

cgwalters: For the record though I didn't test this much locally, threw it up to have CI do it.

runcom: oh yeah, I had an old master on my laptop and didn't notice that, cool yeah, no flake no race :tada:

cgwalters:
Hum, so that run failed; added some more debug logs to help me figure it out.  Offhand...I think there *is* a race here because the tests were relying on checking for config=X and status=Updated, but we don't update those atomically.  Really...it feels like we should have `Spec.Configuration` and `Status.Configuration` [just like other objects](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) right?

kikisdeliveryservice:
would checking if state is Done before checking that they match clarify what's happening  (ie switch the 2 asserts?)

cgwalters:
OK this test passes "locally" now - I added a `time.Sleep(13*time.Second)` as a temporary hack.  Pretty sure the race is https://github.com/openshift/machine-config-operator/pull/765#issuecomment-493136113 - going to take a quick look at a PR for that.

cgwalters:
Hooray, e2e-aws-op passed!

https://github.com/openshift/machine-config-operator/pull/773 should help us avoid the race.

kikisdeliveryservice:
ive seen these weird e2e-aws-upgrade failures before, checking them out elsewhere. let's try again:
/test e2e-aws-upgrade

cgwalters:
Rebased :surfing_man: and CI is good, can I get a lgtm?

PR #772, opened 2019-05-16
Bug 1710626: kubelet: add crio dependency to systemd unit file

rphillips:
**- What I did**
Add a Wait= dependency for crio to start within the kubelet unit file. Prevents a race starting kubelet before crio has initialized.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1710626

`Add crio dependency to Kubelet systemd unit file`

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -3,7 +3,8 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service
+  Wants=rpc-statd.service crio.service
+  After=crio.service
 
   [Service]
   Type=notify
--- a/templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
+++ b/templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
@@ -3,7 +3,8 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service
+  Wants=rpc-statd.service crio.service
+  After=crio.service
 
   [Service]
   Type=notify


runcom:
makes sense to me

/approve

templates/master/01-master-kubelet/_base/units/kubelet.yaml
```diff
@@ -3,7 +3,7 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service
+  Wants=rpc-statd.service crio.service
```
cgwalters: That's not going to guarantee what you want though - the crio socket will still be created async right?  I think we want both `Requires=crio.service` and `After=crio.service`.

cgwalters: (lots more detail in `man systemd.unit`)

runcom:
cc @mrunalp

rphillips:
Requires= is a bit more heavy handed. If the required unit fails (crio), then the kubelet would be disabled...

/cc @sjenning

cgwalters:
> Requires= is a bit more heavy handed. If the required unit fails (crio), then the kubelet would be disabled deactivated...

Yeah.  It's debatable I guess.  However what's not debatable is that we need `Wants=crio.service` too right?

rphillips:
> > Requires= is a bit more heavy handed. If the required unit fails (crio), then the kubelet would be disabled deactivated...
> 
> Yeah. It's debatable I guess. However what's not debatable is that we need `Wants=crio.service` too right?

I think you meant `After=`. I agree, I'll change the PR.

cgwalters:
This looks fine to me; on a procedural note, mind squashing the commits and using a "topic prefix"?  See other examples in this repo with `git log --no-merges`.  (And as a special case, the bug reference) For example:

```
Bug 1710626: kubelet.service: Add Wants= and After= crio.service

Prevents a race starting kubelet before crio has initialized.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1710626
```

rphillips:
Squashed. Thanks.

runcom:
/hold

mainly cause I noticed a ping to Seth, feel free to remove the hold if not necessary anymore (I believe the unit is good now) https://github.com/openshift/machine-config-operator/pull/772#issuecomment-493224251

cgwalters:
Going to
/hold cancel
since Seth is pretty busy, I think having 3 engineers review a few lines of systemd unit is good enough, and this is a clear improvement.  If there are any followups we can always do them as new PRs!

PR #773, opened 2019-05-16
Add a Spec.Configuration to MachineConfigPool

cgwalters:
See https://github.com/openshift/machine-config-operator/pull/765#issuecomment-493136113

MachineConfigPool needs a `Spec.Configuration` and `Status.Configuration`
[just like other objects][1] so that we can properly detect state.
Currently there's a race because the render controler may set `Status.Configuration`
while the pool's `Status` still has `Updated`, so one can't check whether the
pool is at a given config.

[1] https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -34,14 +34,12 @@ func (ctrl *Controller) syncStatusOnly(pool *mcfgv1.MachineConfigPool) error {
 }
 
 func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv1.MachineConfigPoolStatus {
-	previousUpdated := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdated)
-
 	machineCount := int32(len(nodes))
 
-	updatedMachines := getUpdatedMachines(pool.Status.Configuration.Name, nodes)
+	updatedMachines := getUpdatedMachines(pool.Spec.Configuration.Name, nodes)
 	updatedMachineCount := int32(len(updatedMachines))
 
-	readyMachines := getReadyMachines(pool.Status.Configuration.Name, nodes)
+	readyMachines := getReadyMachines(pool.Spec.Configuration.Name, nodes)
 	readyMachineCount := int32(len(readyMachines))
 
 	unavailableMachines := getUnavailableMachines(nodes)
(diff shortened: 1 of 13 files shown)


cgwalters:
This doesn't pass units and is definitely broken, just putting up this WIP for early review/tracking.

derekwaynecarr:
be careful you dont break backwards compatibility with 4.1 and 4.2 clients.

pkg/controller/render/render_controller.go
```diff
@@ -500,18 +503,30 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
...
 
-	glog.V(2).Infof("Pool %s is now targeting: %s", pool.Name, generated.Name)
-	pool.Status.Configuration.Name = generated.Name
-	pool.Status.Configuration.Source = source
-	_, err = ctrl.client.MachineconfigurationV1().MachineConfigPools().UpdateStatus(pool)
+	oldPool, err := json.Marshal(pool)
+	if err != nil {
+		return err
+	}
+	pool.Spec.Configuration.Name = generated.Name
```
runcom: you need to:

```
newPool := pool.DeepCopy()
newPool.Spec....
newPool.Spec....
newPoolJSON, err := ...
```

Or we mutate informer cache?

cgwalters:
> be careful you dont break backwards compatibility with 4.1 and 4.2 clients.

Hmm...are there any other clients?  Maybe the console...I should probably try that out at some point.

pkg/controller/render/render_controller.go
cgwalters: Yeah I checked for that, AFAICS we copy the pool earlier in `syncMachineConfigPool`.  Mumble mumble Rust `&` vs `&mut` :wink:

cgwalters:
OK so the console definitely renders MachineConfigs and MachineConfigPools...but it's super primitive.  And hum, looking at the console code it's TypeScript/React, interesting.  I don't think what we're doing here would break it, conceptually we're just *extending* the schema.

pkg/controller/render/render_controller.go
cgwalters: (I ended up deep copying anyways, since why not)

cgwalters:
OK, did some more fixes and am lifting WIP on this one.  Been playing with it interactively now and it is looking good, but let's see what e2e tests say!

cgwalters:
Hooray, e2e passed!  :tada:

pkg/controller/node/status.go
```diff
@@ -34,14 +34,12 @@ func (ctrl *Controller) syncStatusOnly(pool *mcfgv1.MachineConfigPool) error {
 }
 
 func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv1.MachineConfigPoolStatus {
-	previousUpdated := mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdated)
-
 	machineCount := int32(len(nodes))
 
-	updatedMachines := getUpdatedMachines(pool.Status.Configuration.Name, nodes)
+	updatedMachines := getUpdatedMachines(pool.Spec.Configuration.Name, nodes)
```
runcom: if the `Spec` of an object describes the desired state, shouldn't this routine use Status? we're interested in nodes at a given state (not the desired) - I may be missing something and re-reading https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

runcom:
just a question, LGTM otherwise

/approve

pkg/controller/node/status.go
cgwalters: Hmm.  We're trying to calculate the `Updated`/`Updating` conditions here, so we need to know how many nodes are at our *desired* state.

I'll admit to not being steeped in the art here - can you think of anything else similar to this?

But, it would feel very strange if these values followed Status - the `updatedMachineCount` would go *down* as we updated...

runcom: > I'll admit to not being steeped in the art here - can you think of anything else similar to this?

the template controller is the only one I can think of - it uses Status to tell the render ctrl "I'm done"

but I believe you're right and we're just calculating where we are wrt desired state here.

cgwalters: Another way to think of this is; we probably should never have had `Configuration.Name` as part of `Status`, it should have always been in `Spec`. 

Having parts of `Status` should be derived from other `Status` fields seems wrong.

In effect, the `Status.Configuration.Name` is more like what the CVO is doing with its `history` field.

Yet another argument; `Status.Configuration` is almost meaningless given that multiple configurations could have been rolled out.  If any API client wants to know the state of the system, there's really either "fully updated" or "updating".  Any more detail on "updating" really requires looking at each node for its `currentConfig`.  The `Status.Configuration` isn't so useful for that.

runcom:
oh gosh

PR #771, opened 2019-05-16
[release-4.1] templates/master/00-master: add DR tools

runcom:
Signed-off-by: Sam Batschelet <sbatsche@redhat.com>
Signed-off-by: Antonio Murdaca <runcom@linux.com>

**- What I did**

backport of #757 

**- How to verify it**

**- Description for the changelog**

PR #776, opened 2019-05-17
JIRA PROD-1027: templates/master/00-master: optionally pass etcd connection string

vrutkovs:
…

This enables etcd restore script restore the multimember cluster. This is required for CI where a 3 member cluster needs to be restored.

/cc @hexfusion @runcom

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -9,8 +9,8 @@ contents:
     set -o pipefail
     
     # example
-    # etcd-snapshot-restore.sh $path-to-snapshot
     
+    # etcd-snapshot-restore.sh $path-to-snapshot $etcd-connection-string
     if [[ $EUID -ne 0 ]]; then
       echo "This script must be run as root"
       exit 1
@@ -23,6 +23,12 @@ contents:
       SNAPSHOT_FILE="$1"
     fi
     
+
+    ETCD_CONNSTRING=""
+    if [ "$2" != "" ]; then
+      ETCD_CONNSTRING="$2"
+    fi
+
     CONFIG_FILE_DIR=/etc/kubernetes
     MANIFEST_DIR="${CONFIG_FILE_DIR}/manifests"
     MANIFEST_STOPPED_DIR="${CONFIG_FILE_DIR}/manifests-stopped"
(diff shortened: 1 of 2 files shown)


vrutkovs:
/hold

Requires manual verification

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
```diff
@@ -23,6 +23,12 @@ contents:
       SNAPSHOT_FILE="$1"
     fi
     
+
+    ETCD_CONNSTRING=""
+    if [ "$2" != "" ]; then
```
hexfusion: Can you add this to the example [1] otherwise lgtm

https://github.com/openshift/machine-config-operator/blob/28011239345ebeb0ba8de73c6deb142160511c75/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml#L12

vrutkovs: Right, fixed.

These are optional params, not sure if we want to specify that in the example string

hexfusion: yeah, figure folks will read docs but if they see this it will give them hint to functionality, thanks.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -148,7 +151,7 @@ contents:
         
       env ETCDCTL_API=3 ${ETCDCTL} snapshot restore $SNAPSHOT_FILE \
         --name $ETCD_NAME \
-        --initial-cluster ${ETCD_NAME}=https://${ETCD_DNS_NAME}:2380 \
+        --initial-cluster ${ETCD_CONNSTRING} \
         --initial-cluster-token etcd-cluster-1 \
         --skip-hash-check=true \
         --initial-advertise-peer-urls https://${ETCD_IPV4_ADDRESS}:2380 \
```
hexfusion: Since we are here we should really make this a param as well for flexability?

 if [ -z "${ETCD_INIT_PEERS_URLS}" ]; then
        ETCD_INITIAL_PEERS="https://${ETCD_IPV4_ADDRESS}:2380
 fi

vrutkovs: Good catch, I'll address that in the next PR and make sure CI uses this param

runcom:
cherry-picked in https://github.com/openshift/machine-config-operator/pull/771

PR #779, opened 2019-05-20
DR start_static_pods: reread a list of static pods when starting them

vrutkovs:
**- What I did**

Updated DR scripts to start all stopped pods.

Script won't start all stopped pods if it was terminated and restarted.
This may happen when pod-based ssh bastion is used and connection may break 
when other masters are starting.

Instead `start_static_pods` function should re-read a list of pods in MANIFEST_STOPPED_DIR.

**- How to verify it**
* Follow snapshot restore DR procedure and stop the script when master is being restored.
* Stop the script in the middle
* Restart the script
* Verify all pods from `/etc/kubernetes/manifests-stopped` were started

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -299,6 +299,7 @@ contents:
 
     start_static_pods() {
       echo "Starting static pods.."
+      find ${MANIFEST_STOPPED_DIR} -maxdepth 1 -type f -printf "%f\n" > $STOPPED_STATIC_PODS
       while read STATIC_POD; do
         echo "..starting $STATIC_POD"
         mv ${MANIFEST_STOPPED_DIR}/${STATIC_POD} $MANIFEST_DIR
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-member-recover-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-member-recover-sh.yaml
@@ -33,8 +33,8 @@ contents:
     ASSET_DIR_TMP="$ASSET_DIR/tmp"
     CONFIG_FILE_DIR=/etc/kubernetes
     MANIFEST_DIR="${CONFIG_FILE_DIR}/manifests"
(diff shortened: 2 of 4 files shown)


templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -302,6 +302,7 @@ contents:
 
     start_static_pods() {
       echo "Starting static pods.."
+      find ${MANIFEST_STOPPED_DIR} -maxdepth 1 -type f -printf "%f\n" > $STOPPED_STATIC_PODS
       while read STATIC_POD; do
```
hexfusion: My fear of doing this is that we are possibly starting pods which we did not stop. Meaning after this dir $MANIFEST_STOPPED_DIR is created it is possible that additional static pods get placed in this dir. For example the node gets restored then in 6 months restored again. It makes me nervous that we might start something explicitly stopped by admin by placing in this dir.

I think if we should make stop_static_pods smarter. Check if $STOPPED_STATIC_PODS exists and if so don't overwrite/clobber. Then that list should be true and we can reset to that state. What do you think?

vrutkovs: I don't think we expect anyone else to reuse `/etc/kubernetes/manifests-stopped`, this is not a standard dir for stopped pods (the point is to move away from `/etc/kubernetes/manifests`).

A safer alternative would be `mktemp` I guess

hexfusion: > I don't think we expect anyone else to reuse /etc/kubernetes/manifests-stopped

Right unexpected usage is what I was worried about. Another idea is backup the manifests and then also stop them locally in assets/$STOPPED_DIR. This way the stop dir is never created in /etc/kubernetes?

vrutkovs: >Another idea is backup the manifests and then also stop them locally in assets/$STOPPED_DIR

Sounds good to me, implemented this in 8c4e589

vrutkovs:
/retest

`Unable to mount volumes for pod "support-operator-.."`

PR #784, opened 2019-05-21
server: Support denying serving Ignition to active nodes and pods 

cgwalters:
Ignition may contain secret data; pods running on the cluster
shouldn't have access.

This PR closes of access to any IP that responds on port 22, as that
is a port that is:

 - Known to be active by default
 - Not firewalled

A previous attempt at this was to have an [auth token](#736);
but this fix doesn't require changing the installer and people's PXE setups.

In the future we may reserve a port in the 9xxx range and have the
MCD respond on it so that admins who disable/firewall SSH don't
have indirectly reduced security.

### DIFF
--- a/pkg/server/cluster_server.go
+++ b/pkg/server/cluster_server.go
@@ -4,16 +4,25 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
 
 	yaml "github.com/ghodss/yaml"
+	"github.com/golang/glog"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	rest "k8s.io/client-go/rest"
 	clientcmd "k8s.io/client-go/tools/clientcmd"
 	clientcmdv1 "k8s.io/client-go/tools/clientcmd/api/v1"
 
+	oconfigv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	v1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
 )
 
@@ -24,13 +33,27 @@ const (
 
 	//nolint:gosec
 	bootstrapTokenDir = "/etc/mcs/bootstrap-token"
+
+	// machineProvisionedTimeoutSecs is the maximum amount of time we will wait for a reply
+	// to the port.  Note that Ignition by default times out requests after 10 seconds, so
+	// we need to be lower than that.
+	machineProvisionedTimeoutSecs = 5
 )
+
+// machineProvisionedPorts is a set of TCP ports the MCS connects to in order
+// to check if a machine has already been provisioned.  For now this is
+// the SSH port and the kubelet. In the future we might have the MCD itself listen
+// on a port that's inside the reserved 9000-9900 range.
+var machineProvisionedPorts = []string{"22", "10250"}
 
 // ensure clusterServer implements the
 // Server interface.
 var _ = Server(&clusterServer{})
 
 type clusterServer struct {
+	client kubernetes.Clientset
+
+	configClient oconfigv1.ConfigV1Client
 	// machineClient is used to interact with the
 	// machine config, pool objects.
 	machineClient v1.MachineconfigurationV1Interface
@@ -47,14 +70,127 @@ type clusterServer struct {
 func NewClusterServer(kubeConfig, apiserverURL string) (Server, error) {
 	restConfig, err := getClientConfig(kubeConfig)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to create Kubernetes rest client: %v", err)
+		return nil, errors.Wrapf(err, "Failed to create Kubernetes rest client")
 	}
 
-	mc := v1.NewForConfigOrDie(restConfig)
+	client, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "creating core client")
+	}
+
+	mc, err := v1.NewForConfig(restConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "creating mc client")
+	}
+	oc, err := oconfigv1.NewForConfig(restConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "creating config client")
+	}
 	return &clusterServer{
+		client:         *client,
 		machineClient:  mc,
+		configClient:   *oc,
 		kubeconfigFunc: func() ([]byte, []byte, error) { return kubeconfigFromSecret(bootstrapTokenDir, apiserverURL) },
 	}, nil
+}
+
+// remoteAddrIsProvisioned returns an error if the given IP address responds on a known port.
+// If an error is returned, the machine is already up and should not be served Ignition.
+func remoteAddrIsProvisioned(remoteAddr string) error {
+	startTime := time.Now()
+	defer func() {
+		glog.Infof("Checked address %s for provisioning in %v", remoteAddr, time.Since(startTime))
+	}()
+	var wg sync.WaitGroup
+	provisionedPort := make(chan error)
+	wg.Add(len(machineProvisionedPorts))
+	for _, port := range machineProvisionedPorts {
+		go func(port string) {
+			conn, err := net.DialTimeout("tcp", net.JoinHostPort(remoteAddr, port), time.Second*machineProvisionedTimeoutSecs)
+			if err != nil {
+				glog.Infof("Checking provisioned port %s: %v", port, err)
+			} else {
+				provisionedPort <- fmt.Errorf("Address %s responds on port %s; already provisioned", remoteAddr, port)
+				conn.Close()
+			}
+			wg.Done()
+		}(port)
+	}
+	// The above goroutines race to return the first port we find that is provisioned.
+	// But in the case where none are found, this goroutine waits for their completion
+	// and sends the empty string.
+	go func() {
+		wg.Wait()
+		provisionedPort <- nil
+	}()
+	ret := <-provisionedPort
+	return ret
+}
+
+// remoteAddrIsFromPod returns a non-empty string containing a CIDR if the given remote IP
+// is from a pod.  We don't want to allow
+// pods running on a node to potentially access Ignition; it should
+// only be read by the Ignition software itself when the machine
+// boots up, before it joins the cluster.
+// If the remote address is OK, then the empty string "" is returned.
+// Otherwise, error is set.
+func remoteAddrIsFromPod(configClient oconfigv1.ConfigV1Client, remoteAddr string) (string, error) {
+	network, err := configClient.Networks().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	remoteIP := net.ParseIP(remoteAddr)
+	if remoteIP == nil {
+		return "", nil
+	}
+	for _, n := range network.Status.ClusterNetwork {
+		_, ipr, err := net.ParseCIDR(n.CIDR)
+		if err != nil {
+			glog.V(4).Infof("Parsing CIDR %s: %v", n.CIDR, err)
+			continue
+		}
+		if ipr.Contains(remoteIP) {
+			return n.CIDR, nil
+		}
+	}
+	return "", nil
+}
+
+func isLocalhost(addr string) bool {
+	return addr == "127.0.0.1" || addr == "::1"
+}
+
+// Don't serve Ignition to extant nodes; see
+// https://bugzilla.redhat.com/show_bug.cgi?id=1707162
+// https://github.com/openshift/machine-config-operator/issues/731
+func (cs *clusterServer) shouldServeIgnition(remoteAddr string) error {
+	host, _, err := net.SplitHostPort(remoteAddr)
+	if err != nil {
+		return err
+	}
+
+	if !isLocalhost(host) {
+		fromPod, err := remoteAddrIsFromPod(cs.configClient, host)
+		if err != nil {
+			return err
+		}
+		if fromPod != "" {
+			return &configError{
+				msg:       fmt.Sprintf("Address %s is in the pod network %s", host, fromPod),
+				forbidden: true,
+			}
+		}
+
+		err = remoteAddrIsProvisioned(host)
+		if err != nil {
+			return &configError{
+				msg:       err.Error(),
+				forbidden: true,
+			}
+		}
+	}
+
+	return nil
 }
 
 // GetConfig fetches the machine config(type - Ignition) from the cluster,
@@ -63,6 +199,27 @@ func (cs *clusterServer) GetConfig(cr poolRequest) (*runtime.RawExtension, error
 	mp, err := cs.machineClient.MachineConfigPools().Get(context.TODO(), cr.machineConfigPool, metav1.GetOptions{})
 	if err != nil {
 		return nil, fmt.Errorf("could not fetch pool. err: %v", err)
+	}
+
+	if cr.remoteAddr != "" {
+		// By default for now, we run in warn-only mode
+		provisionCheckError := false
+		config, err := cs.client.CoreV1().ConfigMaps("openshift-machine-config-operator").Get(context.TODO(), "machine-config-server", metav1.GetOptions{})
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return nil, errors.Wrapf(err, "Fetching mcs config")
+			}
+		} else {
+			if config.Data["provision-check"] != "" {
+				provisionCheckError = true
+			}
+		}
+		if err := cs.shouldServeIgnition(cr.remoteAddr); err != nil {
+			if provisionCheckError {
+				return nil, err
+			}
+			glog.Warningf("Would deny serving ignition (enable configmap/machine-config-server/provision-check to enforce): %s", err)
+		}
 	}
 
 	currConf := mp.Status.Configuration.Name
--- a/manifests/machineconfigserver/clusterrole.yaml
+++ b/manifests/machineconfigserver/clusterrole.yaml
@@ -4,6 +4,12 @@ metadata:
   name: machine-config-server
   namespace: {{.TargetNamespace}}
 rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: ["machineconfiguration.openshift.io"]
   resources: ["machineconfigs", "machineconfigpools"]
   verbs: ["*"]
+- apiGroups: ["config.openshift.io"]
+  resources: ["networks"]
+  verbs: ["get", "list", "watch"]
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -2536,9 +2536,15 @@ metadata:
   name: machine-config-server
   namespace: {{.TargetNamespace}}
 rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: ["machineconfiguration.openshift.io"]
   resources: ["machineconfigs", "machineconfigpools"]
   verbs: ["*"]
+- apiGroups: ["config.openshift.io"]
+  resources: ["networks"]
+  verbs: ["get", "list", "watch"]
 `)
 
 func manifestsMachineconfigserverClusterroleYamlBytes() ([]byte, error) {
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -11,6 +11,7 @@ import (
 
 type poolRequest struct {
 	machineConfigPool string
+	remoteAddr        string
 }
 
 // APIServer provides the HTTP(s) endpoint
@@ -92,6 +93,7 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	cr := poolRequest{
 		machineConfigPool: path.Base(r.URL.Path),
+		remoteAddr:        r.RemoteAddr,
 	}
 
 	glog.Infof("Pool %s requested by %s", cr.machineConfigPool, r.RemoteAddr)
@@ -99,8 +101,13 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	conf, err := sh.server.GetConfig(cr)
 	if err != nil {
 		w.Header().Set("Content-Length", "0")
-		w.WriteHeader(http.StatusInternalServerError)
-		glog.Errorf("couldn't get config for req: %v, error: %v", cr, err)
+		if IsForbidden(err) {
+			w.WriteHeader(http.StatusForbidden)
+			glog.Infof("Denying unauthorized request: %v", err)
+		} else {
+			w.WriteHeader(http.StatusInternalServerError)
+			glog.Errorf("couldn't get config for req: %v, error: %v", cr, err)
+		}
 		return
 	}
 	if conf == nil {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -40,6 +40,27 @@ type kubeconfigFunc func() (kubeconfigData []byte, rootCAData []byte, err error)
 // appenderFunc appends Config.
 type appenderFunc func(*mcfgv1.MachineConfig) error
 
+// configError is returned by the GetConfig API
+type configError struct {
+	msg       string
+	forbidden bool
+}
+
+// configError returns the string
+func (e *configError) Error() string {
+	return e.msg
+}
+
+// IsForbidden says if err is an configError with forbidden set
+func IsForbidden(err error) bool {
+	switch t := err.(type) {
+	case *configError:
+		return t.forbidden
+	default:
+		return false
+	}
+}
+
 // Server defines the interface that is implemented by different
 // machine config server implementations.
 type Server interface {


cgwalters:
To test this, I did:
`oc debug node/<worker>` and then did `chroot /host iptables -I OPENSHIFT_BLOCK_OUTPUT -p tcp --dport 22623 -j ACCEPT` to undo the SDN filtering, then from that same pod, I see:


```
# curl -k --head https://10.0.131.197:22623/config/master
HTTP/2 403 
content-length: 0
date: Tue, 21 May 2019 15:34:35 GMT
```

And in the MCS logs:
```
oc logs pods/machine-config-server-6snf2
I0521 15:27:25.572401       1 start.go:37] Version: 4.0.0-alpha.0-437-g307e6bea-dirty (307e6bea9ed1025202d431be21184fe9ea4f6066)
I0521 15:27:25.574374       1 api.go:52] launching server
I0521 15:27:25.574434       1 api.go:52] launching server
I0521 15:27:42.202818       1 api.go:106] Denying unauthorized request: Node ip-10-0-134-102.us-east-2.compute.internal with address 10.0.134.102 is already provisioned
```

ashcrow:
```
 func manifestsMachineconfigserverClusterroleYamlBytes() ([]byte, error) {
make: *** [verify] Error 1
hack/../pkg is out of date. Please run make update
```

cgwalters:
From a disaster recovery perspective, I think if you have e.g. master node with a static IP that you want to reprovision, in order to allow access you'd need to edit the `node` object to drop its `status/addresses` data.

Or we could add a config flag to allow this.

In practice I think most people are going to start with trying to recover a master in-place rather than completely re-set it, and that path won't be affected by this.

cgwalters:
OK, passing the main tests now, though upgrade looks stuck but I doubt it's related.

I also verified that scaling up the worker machineset still works, i.e. it doesn't deny the request.

cgwalters:
Maybe a better architecture would be for the MCS to make an HTTP request to the MCC asking if a given IP is OK?  Would increase the availability requirement for the MCC.   Eh, this is probably fine.  If there's some transient apiserver issue it will mean a node's Ignition request will fail, but Ignition will keep retrying.

abhinavdahiya:
> In practice I think most people are going to start with trying to recover a master in-place rather than completely re-set it, and that path won't be affected by this.

Do we have concrete data from Dr team on this. Seems like Amazon taking a VM away might also be common....

How do we know that the interface that acts as source ip is the one node is reporting as it's internal/public IP?

cgwalters:
> How do we know that the interface that acts as source ip is the one node is reporting as it's internal/public IP?

Bear in mind this PR is not claiming to be a complete solution to the problem.  It's adding a layer of defense, much like the other layers we added.  For example, it does nothing about *external* access.  The auth key approach would be much stronger.

I'll check with the network team about this but remember the primary thing we're trying to prevent with this is *in-cluster* pods accessing the endpoint.  I have some confidence that that access will appear to come from the IP address associated with the node that the kubelet reports.  But again let's see what the SDN team says.

> Seems like Amazon taking a VM away might also be common....

Right, [instance retirement](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-retirement.html) definitely happens.  As I said, in that case if a newly provisioned master happens to get the same IP you'd need to explictly drop out the node object.

Or alternatively, we could tweak this PR to only disallow *reachable* nodes which would be pretty easy.

danwinship:
> the primary thing we're trying to prevent with this is in-cluster pods accessing the endpoint. I have some confidence that that access will appear to come from the IP address associated with the node that the kubelet reports.

Assuming the pod isn't using an egress IP or egress router, then traffic from a pod on node A addressed to node B's primary node IP will appear to come from node A's primary node IP.

But we didn't change the MCS to not listen on `0.0.0.0` did we? So a pod on node A could connect to node B's `tun0` IP instead, and that would appear to come from the pod's IP directly. (So you'd want to filter out all connections with source IPs in the pod network as well. Or more simply, filter out connections if the *destination* IP is the `tun0` IP.)

Also, in some environments, if nodes have multiple IPs, then a connection from a pod on node A to a non-primary IP on node B might appear to come from node A's non-primary IP rather than its primary IP.

> if a newly provisioned master happens to get the same IP you'd need to explictly drop out the node object.

VMware at least definitely leaves stale node objects around when dynamically scaling nodes. (I don't know if that would ever effect masters though.)

cgwalters:
> (So you'd want to filter out all connections with source IPs in the pod network as well. Or more simply, filter out connections if the destination IP is the tun0 IP.)

Ah, OK.  Hmm though I'm not sure we can get that from the golang side HTTP request... but thinking about this, why don't we just add iptables rules to the masters that require that the destination IP for the MCS is not `tun0`?

danwinship:
iptables doesn't see the packet because it's delivered by OVS

cgwalters:
/hold
Per discussion above, this needs more work to implement the suggestions in https://github.com/openshift/machine-config-operator/pull/784#issuecomment-494833954

cgwalters:
/hold cancel

OK, updated :new: to also deny requests coming from `tun0`.  I verified both approaches work, doing a request to the main master IP directly, as well as targeting its `tun0`.

```
oc logs pods/machine-config-server-l7wd2
I0523 22:27:09.952731       1 start.go:37] Version: 4.0.0-alpha.0-444-g9eebd1d4-dirty (9eebd1d4a17eb2d26ae74709252cf6ea77330703)
I0523 22:27:09.955394       1 api.go:52] launching server
I0523 22:27:09.955474       1 api.go:52] launching server
I0524 18:00:24.919941       1 api.go:99] Pool master requested by 10.0.137.206:32870
I0524 18:00:24.936254       1 api.go:106] Denying unauthorized request: Node ip-10-0-137-206.us-east-2.compute.internal with address 10.0.137.206 is already provisioned
I0524 18:06:39.321439       1 api.go:99] Pool master requested by 10.131.0.1:59360
I0524 18:06:39.330104       1 api.go:106] Denying unauthorized request: Requesting host 10.131.0.1 is within pod network CIDR 10.128.0.0/14
```

We also now only deny nodes that are `NodeReady=true`.

And I verified with this patch that scaling up a machineset still works.

pkg/server/cluster_server.go
```diff
@@ -48,13 +54,78 @@ func NewClusterServer(kubeConfig, apiserverURL string) (Server, error) {
...
+// remoteAddrAlreadyIsNode returns true if the given remoteAddr is assigned
+// to a provisioned node.
+func remoteAddrAlreadyIsNode(kc clientset.Interface, remoteAddr string) (string, error) {
+	nodes, err := kc.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	for _, node := range nodes.Items {
+		// Ignore NodeReady=false nodes; this allows reprovisioning them.
```
abhinavdahiya: Node not ready doesn't mean node is getting re-provisioned. where did we get that assumption?

abhinavdahiya:
> Also, in some environments, if nodes have multiple IPs, then a connection from a pod on node A to a non-primary IP on node B might appear to come from node A's non-primary IP rather than its primary IP.

This is still not solved.

cgwalters:
> [multi-NIC] is still not solved.

Right, but...again, not claiming a comprehensive solution here.  The issue is important enough to have layers of defenses.

That said...one approach that we could take (derived from an approach @ericavonb mentioned) is run a daemonset service on each node (probably as part of the MCD) that listens on a well-known port and provides a static reply.  Then we could have the MCS "call back" to the requesting IP - if it gets a reply it knows to deny the request.

cgwalters:
Or...maybe way simpler, just try to connect to port 22.

cgwalters:
> Or...maybe way simpler, just try to connect to port 22.

And, if we can assume `sshd.socket` is up - this check basically subsumes the other two and also solves multi-NIC.

cgwalters:
Answering my own question, from looking at the sdn daemonset:

```
        ports:
        - containerPort: 10256
          hostPort: 10256
          name: healthz
          protocol: TCP
```

So that looks like a good thing to check.

cgwalters:
> So that looks like a good thing to check.

Ah but the default AWS firewall doesn't let me go across that.

cgwalters:
OK this PR is now *completely different* - we just check SSH, which [as noted above](https://github.com/openshift/machine-config-operator/pull/784#issuecomment-495761436) solves most prior issues with this approach.

danwinship:
You don't want to try to connect to an openshift-sdn service anyway; then you're only protected when openshift-sdn is being used. Can't you connect to the kubelet itself?

pkg/server/cluster_server.go
```diff
@@ -63,6 +86,27 @@ func (cs *clusterServer) GetConfig(cr poolRequest) (*igntypes.Config, error) {
...
+	// Don't serve Ignition to extant nodes; see
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1707162
+	// https://github.com/openshift/machine-config-operator/issues/731
+	if cr.remoteAddr != "" {
+		host, _, err := net.SplitHostPort(cr.remoteAddr)
+		if err != nil {
+			return nil, err
+		}
+
+		remoteHasSDN, err := remoteAddrIsProvisioned(host)
```
danwinship: Variable name is wrong now.

danwinship:
> Can't you connect to the kubelet itself?

No, because you need to test a privileged port, so that if a non-privileged pod tries to connect via an SDN or multus IP, it can't fake out the check by listening on whatever port the MCS is going to call back on.

cgwalters:
> No, because you need to test a privileged port, so that if a non-privileged pod tries to connect via an SDN or multus IP, it can't fake out the check by listening on whatever port the MCS is going to call back on.

Yeah.  Though it may be worth still carrying the code to explicitly filter out the SDN addresses before this check.

cgwalters:
Hum interesting, [that e2e job](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/784/pull-ci-openshift-machine-config-operator-master-e2e-aws-op/2246/artifacts/e2e-aws-op/must-gather/namespaces/openshift-machine-config-operator/pods/machine-config-server-5r95j/machine-config-server/machine-config-server/logs/current.log) looks like the workers were in Ignition retry loops?

cgwalters:
So I went to look at the Ignition docs and:

> httpResponseHeaders (integer) the time to wait (in seconds) for the server's response headers (but not the body) after making a request. 0 indicates no timeout. Default is 10 seconds.

Pretty awesome I happened to pick 10 seconds as well for the SSH timeout.  Says something about human nature perhaps.  Anyways let's drop the SSH provisioning check down to 5 seconds...

I'm going to also readd the check to deny anything from the pod network.

If we do that, I think we can also safely check whether the kubelet responds as well.

Another thing occurs to me; we could change Ignition to listen on a well-known port and advertise that it's provisioning.  We'd need to depend on a respun RHCOS though.

cgwalters:
Tentatively thinking the problem is that the network CRD isn't available when the MCS starts up the first time, and this crashes `	oc := oconfigv1.NewForConfigOrDie(restConfig)`.  But let's see if the termination log patch helps tell us.

pkg/server/cluster_server.go
```diff
@@ -45,16 +62,70 @@ type clusterServer struct {
...
+func remoteAddrIsProvisioned(remoteAddr string) (string, error) {
+	var wg sync.WaitGroup
+	var provisionedPort string
+	wg.Add(len(machineProvisionedPorts))
+	for _, port := range machineProvisionedPorts {
+		go func(port string) {
+			conn, err := net.DialTimeout("tcp", net.JoinHostPort(remoteAddr, port), time.Second * machineProvisionedTimeoutSecs)
+			defer conn.Close()
+			if err != nil {
+				return
```
ericavonb: nit: debug log the error rather than swallow?

runcom: also, put `defer conn.Close()` after the err check

pkg/server/cluster_server.go
```diff
@@ -45,16 +62,70 @@ type clusterServer struct {
...
 }
 
+// remoteAddrIsProvisioned returns true if the given IP address responds on a known port.
+// If that's true, the machine is already up and should not be served Ignition.
+func remoteAddrIsProvisioned(remoteAddr string) (string, error) {
+	var wg sync.WaitGroup
+	var provisionedPort string
+	wg.Add(len(machineProvisionedPorts))
+	for _, port := range machineProvisionedPorts {
+		go func(port string) {
```
ericavonb: can this variable be named something else? avoid shadowing and the errors that come with it

pkg/server/cluster_server.go
```diff
@@ -45,16 +62,70 @@ type clusterServer struct {
...
+		return "", err
+	}
+	remoteIP := net.ParseIP(remoteAddr)
+	if remoteIP == nil {
+		return "", nil
+	}
+	for _, n := range network.Status.ClusterNetwork {
+		_, ipr, err := net.ParseCIDR(n.CIDR)
+		if err != nil {
+			continue
```
ericavonb: nit: debug log the error rather than swallow

pkg/server/cluster_server.go
cgwalters: Might as well shadow - [it'd be a trap](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables) to use the outer loop variable right?

ericavonb: I don't think it should matter if the function arg has a different variable name so long as the range value is passed to it properly. That linked example tho 😳 . Guess it doesn't matter too much if there have't been complaints about it?
```go
	for _, p := range machineProvisionedPorts {
 		go func(port string) {
			// use `port` here
 		}(p)
	}
```

cgwalters:
Blah, I would have sworn I had tested this locally too...I have a few branches for this and maybe I was confused as to which I was running.  Anyways, we were missing a `wg.Done()`.

cgwalters:
:tada: Fixed the issues, this now passes e2e.  Verified I get the expected denials if I [disable the firewall](https://github.com/openshift/machine-config-operator/pull/784#issuecomment-494443285).

@crawford @abhinavdahiya - you both voiced concerns about *prior* versions of this PR, but it is now significantly different, and I believe both quite risk-free and a strong mitigation.

runcom:
/hold

Just adding the hold cause I was about to merge but it would be nice to have approvals from Abhinav/Alex

crawford:
This looks like it's going to incur a five-second penalty for each machine that gets provisioned. Am I reading this correctly?

cgwalters:
> This looks like it's going to incur a five-second penalty for each machine that gets provisioned.

I haven't tested it to be sure, but I *think* the cost in most cases is going to be trying to establish a TCP connection and immediately getting `ECONNREFUSED` becase there's nothing listening on those ports.  That should be sub-second latency.  But I'll add some logs so we can see!

cgwalters:
Hm indeed, [5 seconds of latency](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/784/pull-ci-openshift-machine-config-operator-master-e2e-aws/3855/artifacts/e2e-aws/must-gather/namespaces/openshift-machine-config-operator/pods/machine-config-server-48bcw/machine-config-server/machine-config-server/logs/current.log).  I may need to redeploy a cluster with this PR and do some investigation if we care about that.

crawford:
I think we do need to worry about that delay. The time it takes to provision a new node is a key metric that is used when comparing K8S offerings.

cgwalters:
> I think we do need to worry about that delay. The time it takes to provision a new node is a key metric that is used when comparing K8S offerings.

It seems to take about 4 minutes to provision a new worker for me - offhand that time looks like it's dominated by the early pivot.  So this adds 2% to the provisioning time but...yeah agree it's the wrong direction.

I don't understand at all why we're getting connection timeouts instead of having the connection refused; may relate to the network state in the initramfs?  Going to dig at that.

michaelgugino:
@cgwalters what about checking for listening on node's server port?  All nodes run that port, we took away ability to change the port (afaik).   This would definitively say 'this is a node'.

cgwalters:
> what about checking for listening on node's server port? All nodes run that port, we took away ability to change the port (afaik). This would definitively say 'this is a node'.

Sorry, which port is that?  You're not talking about the kubelet `10250` one?

michaelgugino:
> > what about checking for listening on node's server port? All nodes run that port, we took away ability to change the port (afaik). This would definitively say 'this is a node'.
> 
> Sorry, which port is that? You're not talking about the kubelet `10250` one?

Yeah, I'm talking about whatever port the kubelet listens on for incoming connections (I know it as the 'server' port, I'm not sure what it's actually called, but it's for getting pod logs and such).

Sorry it took me 10 years to reply, too much spam from github. @cgwalters

cgwalters:
OK I think I finally got this commit correctly past `make verify` and `make test-unit`, my bad for not running those locally.
Now we have some registry flakes in builds.
/retest

cgwalters:
OK yeah, still adds 5s:

```
2019-10-18T18:31:10.248255764Z I1018 18:31:10.248203       1 api.go:99] Pool worker requested by 10.0.132.105:35785
2019-10-18T18:31:15.342547142Z I1018 18:31:15.342482       1 cluster_server.go:100] Checking provisioned port 10250: dial tcp 10.0.132.105:10250: i/o timeout
2019-10-18T18:31:15.342547142Z I1018 18:31:15.342504       1 cluster_server.go:100] Checking provisioned port 22: dial tcp 10.0.132.105:22: i/o timeout
2019-10-18T18:31:15.342547142Z I1018 18:31:15.342528       1 cluster_server.go:91] Checked address 10.0.132.105 for provisioning in 5.000313618s
```

pkg/server/cluster_server.go
```diff
@@ -46,16 +65,84 @@ type clusterServer struct {
...
 	return &clusterServer{
 		machineClient:  mc,
+		configClient:   *oc,
 		kubeconfigFunc: func() ([]byte, []byte, error) { return kubeconfigFromSecret(bootstrapTokenDir, apiserverURL) },
 	}, nil
 }
 
+// remoteAddrIsProvisioned returns true if the given IP address responds on a known port.
+// If that's true, the machine is already up and should not be served Ignition.
+func remoteAddrIsProvisioned(remoteAddr string) string {
```
abhinavdahiya: hmm, this returns `true` as string or bool ??

cgwalters: Neither, it returns the port that responded as a string, or `""` if none.  See what the calling code does.

cgwalters: Maybe it'd be better to have it return `error`?  Dunno.

cgwalters:
I am not able to reproduce the timeout at least locally in libvirt - I immediately get `ECONNREFUSED` when the system is sitting there in the initramfs.

OH!  But here's something interesting...this delay is AWS specific!  Which we can tell now that we're running this operator's test suite in GCP (i.e. it's now `e2e-gcp-op`):

See [this log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/784/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/97/artifacts/e2e-gcp-op/must-gather/registry-svc-ci-openshift-org-ci-op-x1iij6q2-stable-sha256-dae1257b516a5c177237cfef5a6a3e241962b0d20cf54bcb2b66dc1671c5035e/namespaces/openshift-machine-config-operator/pods/machine-config-server-5zm74/machine-config-server/machine-config-server/logs/current.log):

```
2019-10-18T18:25:40.924407037Z I1018 18:25:40.924056       1 start.go:38] Version: machine-config-daemon-4.2.0-201907161330-345-gd0a52f85-dirty (d0a52f8506e63a74612dd4713b0dd24d8cbada5a)
2019-10-18T18:25:40.925178352Z I1018 18:25:40.925148       1 api.go:52] Launching server on :22624
2019-10-18T18:25:40.925250916Z I1018 18:25:40.925209       1 api.go:52] Launching server on :22623
2019-10-18T18:28:56.405696136Z I1018 18:28:56.405619       1 api.go:99] Pool worker requested by 10.0.32.3:37624
2019-10-18T18:28:56.517856181Z I1018 18:28:56.517235       1 cluster_server.go:100] Checking provisioned port 22: dial tcp 10.0.32.3:22: connect: connection refused
2019-10-18T18:28:56.517856181Z I1018 18:28:56.517309       1 cluster_server.go:100] Checking provisioned port 10250: dial tcp 10.0.32.3:10250: connect: connection refused
2019-10-18T18:28:56.517856181Z I1018 18:28:56.517321       1 cluster_server.go:91] Checked address 10.0.32.3 for provisioning in 1.859659ms
2019-10-18T18:29:00.587612577Z I1018 18:29:00.587561       1 api.go:99] Pool worker requested by 10.0.32.2:51946
2019-10-18T18:29:00.597302341Z I1018 18:29:00.595481       1 cluster_server.go:100] Checking provisioned port 22: dial tcp 10.0.32.2:22: connect: connection refused
2019-10-18T18:29:00.597302341Z I1018 18:29:00.595507       1 cluster_server.go:100] Checking provisioned port 10250: dial tcp 10.0.32.2:10250: connect: connection refused
2019-10-18T18:29:00.597302341Z I1018 18:29:00.595547       1 cluster_server.go:91] Checked address 10.0.32.2 for provisioning in 1.518867ms
2019-10-18T18:29:03.833399559Z I1018 18:29:03.833333       1 api.go:99] Pool worker requested by 10.0.32.4:44096
2019-10-18T18:29:03.841810966Z I1018 18:29:03.841737       1 cluster_server.go:100] Checking provisioned port 22: dial tcp 10.0.32.4:22: connect: connection refused
2019-10-18T18:29:03.841866064Z I1018 18:29:03.841822       1 cluster_server.go:100] Checking provisioned port 10250: dial tcp 10.0.32.4:10250: connect: connection refused
2019-10-18T18:29:03.841866064Z I1018 18:29:03.841843       1 cluster_server.go:91] Checked address 10.0.32.4 for provisioning in 1.706479ms
```

Which looks great, exactly what we want.

Is there some sort of ingress firewalling on AWS that's lifted *after* the machines have booted or something?

cgwalters:
So given that the delay is seemingly AWS specific, I think I'd like to propose merging this - the security benefits seem quite worth it.  We can let it cook in master for a while and see if there's any unexpected fallout elsewhere.  We can then revert if anything breaks (or perhaps make this code into a runtime option via configmap?).  And we can investigate/debug the AWS delay in parallel.

pkg/server/cluster_server.go
cgwalters: I realize now what you meant, the documentation string was wrong.  Fixed that and also changed it to return `error` and clarified the docs!

cgwalters:
Now that 4.4 is open, any objections to landing this?

ashcrow:
No objection from me!

abhinavdahiya:
I just have few things.

RHEL 7 nodes are joined to the cluster using the ignition server and those machines are _provisioned_. We need to make sure they keep working.

Would like to keep requests from 127.0.0.1 or something to make sure we have a way to debug the content being served

also since this might affect users.. and enhancement that provides the details on new behavior would be very nice.

cgwalters:
> RHEL 7 nodes are joined to the cluster using the ignition server and those machines are provisioned. We need to make sure they keep working.

See https://github.com/openshift/openshift-ansible/pull/11614 - for the same reasons motivating this PR, openshift-ansible doesn't query the MCS directly anymore.

> Would like to keep requests from 127.0.0.1 or something to make sure we have a way to debug the content being served

OK, done.

> also since this might affect users.. and enhancement that provides the details on new behavior would be very nice.

Understood, though two things:

First, this is building on top of a lot of patches to other components that came before the enhancements process - most of such an enhancement would probably be writing up all of the history behind the *current* state, right?

Second and more importantly: it's *possible* this could break a user, but the reason I'm proposing this approach is I think it's unlikely.  In general, anyone who has a valid reason to get data from the MCS should also have the credentials to pull it from the cluster directly (as openshift-ansible does now).  

Third, part of the idea here is to land this in master and let it soak for a bit to see if e.g. something unusual in bare metal CI shakes out - if there's fallout we can revert, whereas an enhancement feels like more of a commitment.

Finally, I'd like to spend any enhancement-writing time related to this on strengthening the "machine identity" model - ideally supporting TPMs, including linkage between the MCO and machineAPI, probably having the MCO be able to match up `machine` objects when serving Ignition, etc.

crawford:
> Second and more importantly: it's possible this could break a user, but the reason I'm proposing this approach is I think it's unlikely.

This has the potential to significantly complicate disaster recovery. How much thought has been put into how this will affect those procedures?

> Finally, I'd like to spend any enhancement-writing time related to this on strengthening the "machine identity" model - ideally supporting TPMs, including linkage between the MCO and machineAPI, probably having the MCO be able to match up machine objects when serving Ignition, etc.

This has already been solved at the Kubernetes layer, which is where I still believe we should tackle this problem. I'm still unconvinced that Ignition should learn how to help attest the identity of the underlying machine.

cgwalters:
> This has the potential to significantly complicate disaster recovery. How much thought has been put into how this will affect those procedures?

How so?  If you want to re-run Ignition manually or something?  

> This has already been solved at the Kubernetes layer, which is where I still believe we should tackle this problem. I'm still unconvinced that Ignition should learn how to help attest the identity of the underlying machine.

I don't think we've solved machine identity well at the Kubernetes layer - why do you think it is?  I really think the https://github.com/openshift/cluster-machine-approver is a hack personally...
There's some prior work on TPMs for Kube, see https://github.com/coreos/kubernetes-dtc-admission at least.

(This isn't trying to get machine identity into Ignition really, but Ignition would be part of enabling it at a higher level)

michaelgugino:
Commit message should be updated to reflect kubelet port.

cgwalters:
One thing that came up in realtime chat about this is that in e.g. AWS, we might be able to use [instance identity signatures](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) somehow.  Ultimately what we need is a per-machine secret that we can know the requester has.

And actually, we should validate this when fetching Ignition *and* in the approver I think.

cgwalters:
That said anyways...I'd reiterate that we have a problem as long as Ignition contains secrets (and it does) and that's not likely to change in the near future.

Without actually trying to write an exploit I can't say for sure this is *needed* now, but my argument is basically:

- It's a pretty safe check to add and unlikely to break anything (and I'm arguing for having it soak in master a bit to see)
- We know this isn't the end solution, but it doesn't block or harm the ability to add a stronger solution
- It's an easy thing to backport if the need arises

cgwalters:
OK I changed this PR to just warn by default so we can get it in with zero risk, check whether anyone hits this (e.g. some bare metal install paths), and also make it easy to switch to enforcing via a configmap.

cgwalters:
OK this came up in a conversation again; rebased :surfer: - only compile tested now though.

cgwalters:
Since this is now informational only, I think what we need to do now is export a Prometheus metric for denials, and the roll that up into [telemetry](https://github.com/openshift/telemeter) ?

jamescassell:
> This PR closes of access to any IP that responds on port 22, as that
> is a port that is:
> 
>     Known to be active by default
>     Not firewalled
> 

I think the way this works is broken.  All I (as an attacker) need to do to get the ignition is to block SSH on my local machine, then request the ignition.

michaelgugino:
What I think we should do is secure the endpoint with authentication.  A firewall or iptables rule is not a subsitute for authentication.  The machine-api will be available in baremetal clusters soon, as well as vmware, and there's no telling what the network topology of those is going to look like.  It's very well possible that the MCS is behind a routable IP.

cgwalters:
> What I think we should do is secure the endpoint with authentication. 

Yep, that was https://github.com/openshift/machine-config-operator/pull/736 but where it died is *requiring* it would be a huge UX issue for many baremetal flows.

That said...I'm increasingly feeling like a "good enough" mitigation would be something like switching to requiring an auth token *after* the cluster is initialized.

michaelgugino:
> > What I think we should do is secure the endpoint with authentication.
> 
> Yep, that was #736 but where it died is _requiring_ it would be a huge UX issue for many baremetal flows.
> 
> That said...I'm increasingly feeling like a "good enough" mitigation would be something like switching to requiring an auth token _after_ the cluster is initialized.

Consider the reverse of optionally disabling it for baremetal flows.

jomeier:
We can access the ignition files from any PC in our network that doesn't belong to the cluster:

https://github.com/openshift/okd/issues/176

The curl command described there doesn't work out of the box in pods but I'm not sure if this can be enabled somehow.

I assume that also vSphere credentials are inside of the ignition files? This would be a major security leak. As proposed earlier in this PR: is it possible to secure the API endpoint?

If that's not possible in short term what is the proposed workaround to secure the ignition files with a firewall? Could you provide a best practice network layout for that?

crawford:
/hold

This needs an enhancement. Casually skimming the history, it's clear that there are still open questions.

cgwalters:
So in the middle of this epic PR discussion, the change turned from "deny" to "warn and add opt-in mechanism to deny".  I forgot to change the PR title which probably led to a lot of confusion.

I completely agree we need an enhancement if we try to do anything that would deny (and particularly anything that ties together machineAPI and MCO or affects bare metal provisioning flow, disaster recovery etc.)  I'm less in agreement that we need an enhancement to *log* by default.  If it helps I can remove the ability to deny.

(But probably instead of logging to the pod we really want saner observability like an event and prometheus metrics, I need to look at that)

jomeier:
Is there any hint in the docs that a firewall should be set up to prevent anyone from pulling the ignition files with cloud credentials contained in it? Or can this deny Switch be configured during installation of the cluster? By providing a switch in the install-config.yaml for example.

cgwalters:
After https://github.com/openshift/enhancements/pull/368 lands we'll be in a better place to enforce an auth token for MAO managed setups.

I do agree with Alex we want an enhancement for this but it should basically be:

- installer generates `oc -n openshift-config create secret generic provisioning token=<random token>`
- installer injects that into its user data
- MCO also does so for the pointer configs it manages
- MCO denies requests which don't have a header with the token

It's harder to do better than that unless we go to *per machine* user data, but this would suffice to start.  In MAO managed scenarios we should be able to iteratively upgrade later.  But if we start requiring this token e.g. to be specified at the PXE commandline on UPI metal then it will become a bit of an "API".

cgwalters:
Working on an enhancement for this: https://hackmd.io/k7Mfb1lpSIWTzRFvo6o9ig

cgwalters:
Will be obsoleted by https://github.com/openshift/machine-config-operator/pull/2223

PR #785, opened 2019-05-21
Bug 1712507: etcdquorumguard: Respond correctly to TERM when pod shut down

smarterclayton:
sleep, when run as pid 1, does not correctly exit on TERM and so stays
running. Correct that by running a small shim that terminates on TERM,
lower the graceful deletion period (the minimum is 3s in practice).

Correct the image reference to be a real image so that this can be
tested against a working cluster.

@hexfusion

### DIFF
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -33,6 +33,7 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
+      terminationGracePeriodSeconds: 3
       tolerations:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
@@ -49,16 +50,21 @@ spec:
         operator: Exists
         effect: NoSchedule
       containers:
-      - image: registry.svc.ci.openshift.org/openshift:cli
+      - image: quay.io/openshift/origin-cli:latest
         imagePullPolicy: IfNotPresent
         name: guard
         volumeMounts:
         - mountPath: /mnt/kube
           name: kubecerts
         command:
-        - "/bin/sleep"
+        - /bin/bash
         args:
-        - "infinity"
+        - -c
+        - |
+          # properly handle TERM and exit as soon as it is signaled
+          set -euo pipefail
+          trap 'jobs -p | xargs -r kill; exit 0' TERM
+          sleep infinity & wait
         readinessProbe:
           exec:
             command:
--- a/install/image-references
+++ b/install/image-references
@@ -45,4 +45,4 @@ spec:
   - name: cli
     from:
       kind: DockerImage
-      name: registry.svc.ci.openshift.org/openshift:cli
+      name: quay.io/openshift/origin-cli:latest


cgwalters:
Hum,
```
$ rpm -qf /usr/bin/sleep
coreutils-8.31-1.fc30.x86_64
$ ```
appears to die on `SIGTERM`...oh interesting but not as pid 1 as you say, why the heck is that?

It's pretty crazy honestly how prevalent shell script is; we try to write elegant solid "bricks" and end up still gluing them together with messy shell mortar...

On a different topic this is partially reverting https://github.com/openshift/machine-config-operator/pull/750 which I liked a lot...

**Comment by smarterclayton** [comment]
Fixing this exposed that waitForPods is broken in the test `TestEtcdQuorumGuard` (the test only works because it currently takes 30s to shut down).  I'll have to fix the test before this can merge.

**Comment by smarterclayton** [comment]
Shell is fine.  Just one of those things.  I'd rather have shell I can iterate on in a single place than complexity sprawled into multiple places elsewhere.

The point of the CLI images being real was so you could test the yaml directly on the cluster.  Not sure what the value of #750 is, it's inconsistent with other repos.  I'm not directly opposed to people experimenting, but it defeats the point of having the yaml be testable independently.

**Comment by smarterclayton** [comment]
/hold

the e2e still has to be fixed

File: install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml [hunk 1]
```diff
@@ -49,16 +50,23 @@ spec:
...
-        - "/bin/sleep"
+        - /bin/bash
         args:
-        - "infinity"
+        - -c
+        - |
+          # properly handle TERM and exit as soon as it is signaled
+          set -euo pipefail
+          trap 'jobs -p | xargs -r kill; exit 0' TERM
+          while true; do
```
RobertKrawitz: Why this as opposed to

`sleep infinity & wait`

or simply 

`sleep infinity`?

smarterclayton: that's fine

RobertKrawitz:
I'm not able to reproduce that test failure in multiple tries with either the new code or the old.  I tried waiting for >5 minutes with the old code prior to running the test in case there was a problem with the loop, but no failure there either.

I'm going to continue running the test in a loop for a while to see if there are any failures.

RobertKrawitz:
There does appear to be a problem with the test:

```Make all schedulable
Check for all running
  Deployment is ready! 3 3
Make one unschedulable
  Evicting pod openshift-machine-config-operator/etcd-quorum-guardx-78bd998976-mwbm4...
Wait for 2 running
  Deployment is ready! 3 2
Make second unschedulable
  Evicting pod openshift-machine-config-operator/etcd-quorum-guardx-78bd998976-mwbm4...
  Pod should not have been evicted
Make all schedulable
Wait for all running
  Deployment is ready! 3 3
Make one unschedulable
  Evicting pod openshift-machine-config-operator/etcd-quorum-guardx-78bd998976-mnw44...
Wait for one not running
  Deployment is ready! 3 2
Make all schedulable
Wait for all
  Deployment is ready! 3 3
--- FAIL: TestEtcdQuorumGuardx (32.33s)
    etcdquorumguard_test.go:61: Pod should not have been evicted because it violated disruption budget: Unable to evict pod openshift-machine-config-operator/etcd-quorum-guardx-78bd998976-mwbm4: pods "etcd-quorum-guardx-78bd998976-mwbm4" not found
```
The second one it tries to evict should not be the same as the first one.

RobertKrawitz:
Test fix in #789

hexfusion:
@kikisdeliveryservice @cgwalters we need approval please :)

smarterclayton:
Folded into #789

RobertKrawitz:
This is merged into #789 -- @smarterclayton please review that one.

PR #786, opened 2019-05-22
daemon: Explicitly stop kubelet.service first before rebooting

cgwalters:
This is resurrecting the other part of:
https://github.com/openshift/machine-config-operator/pull/237

Basically it's not advantageous to us to try to shut down everything
at once; we want to ensure kubelet can fully shut itself down
gracefully, and only after that initiate the reboot for the rest
of the machine.

This also sets the MCD up to be in full explicit control of
kubelet shutdown; e.g. if we wanted to configure timeout behavior
more, we could easily do that with this "2 phase" shutdown (though
it would involve some custom code and not a shell script one-liner).

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -160,9 +160,16 @@ var (
 	defaultRebootTimeout = 24 * time.Hour
 )
 
+// rebootCommand creates a new transient systemd unit to reboot the system.
+// We explictily try to stop kubelet.service first, before anything else; this
+// way we ensure the rest of system stays running, because kubelet may need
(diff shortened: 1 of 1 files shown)


cgwalters:
The thing we want to add at some point on top of this that https://github.com/openshift/machine-config-operator/pull/237 was trying to do is ensure that kubelet's `SIGTERM` handling evicts daemonset pods too, and doesn't allow new ones to be scheduled.

rphillips:
lgtm

cgwalters:
@rphillips Were you intentionally omtting a `/` there, to defer to a second review?  (I'm fine if that's the case, just curious)

PR #788, opened 2019-05-22
DR: reload systemd services on disk before starting kubelet in etcd restore script

vrutkovs:
**- What I did**
Updated etcd restore script to reload systemd before starting kubelet

**- How to verify it**

1. Change kubelet on disk
2. Run etcd restore script

If the service has changed kubelet won't start without daemon reload:
```
+ echo 'Starting kubelet..'
+ systemctl start kubelet.service
Warning: The unit file, source configuration file or drop-ins of kubelet.service changed on disk. Run 'systemctl daemon-reload' to reload units
```
I'm not quite sure what's changing kubelet on disk during this procedure though

/cc @patrickdillon @hexfusion

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -313,6 +313,7 @@ contents:
 
     start_kubelet() {
       echo "Starting kubelet.."
+      systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


cgwalters:
Can you add `DR:` or so as a "topic prefix" in the commit messages?  I thought this PR was about the main kubelet config at first.

Also...hopefully at some point soon these scripts will gain a new canonical upstream in a container and we can only do backports where necessary here?  4.2 development is open, we can make new containers now...

PR #789, opened 2019-05-22
Bug 1712507: etcdquorumguard: Respond correctly to TERM when pod shut down

RobertKrawitz:
sleep, when run as pid 1, does not correctly exit on TERM and so stays
running. Correct that by running a small shim that terminates on TERM,
lower the graceful deletion period (the minimum is 3s in practice).

Correct the image reference to be a real image so that this can be
tested against a working cluster.

Also, the etcd quorum guard test does not correctly make nodes unschedulable, resulting in occasional failures when the quorum guard exits more quickly.

@hexfusion @smarterclayton

### DIFF
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -33,6 +33,7 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
+      terminationGracePeriodSeconds: 3
       tolerations:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
@@ -49,16 +50,21 @@ spec:
         operator: Exists
         effect: NoSchedule
       containers:
-      - image: registry.svc.ci.openshift.org/openshift:cli
+      - image: quay.io/openshift/origin-cli:latest
         imagePullPolicy: IfNotPresent
         name: guard
         volumeMounts:
         - mountPath: /mnt/kube
           name: kubecerts
         command:
-        - "/bin/sleep"
+        - /bin/bash
         args:
-        - "infinity"
+        - -c
+        - |
+          # properly handle TERM and exit as soon as it is signaled
+          set -euo pipefail
+          trap 'jobs -p | xargs -r kill; exit 0' TERM
+          sleep infinity & wait
         readinessProbe:
           exec:
             command:
--- a/test/e2e/etcdquorumguard_test.go
+++ b/test/e2e/etcdquorumguard_test.go
@@ -37,7 +37,7 @@ type podinfo map[string]podstatus
 func TestEtcdQuorumGuard(t *testing.T) {
 	cs := framework.NewClientSet("")
 	if err := waitForEtcdQuorumGuardDeployment(cs); err != nil {
-		t.Fatalf("etcdQuotaGard deployment not present: %s", err.Error())
+		t.Fatalf("etcdQuorumGuard deployment not present: %s", err.Error())
 	}
 	fmt.Print("Make all schedulable\n")
 	if err := makeAllNodesSchedulable(cs); err != nil {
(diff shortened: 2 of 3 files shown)


RobertKrawitz:
This needs to be backported to 4.1.z prior to merge of #785

test/e2e/etcdquorumguard_test.go
```diff
@@ -98,23 +98,44 @@ func makeNodeUnSchedulableOrSchedulable(cs *framework.ClientSet, node string, un
...
+				continue;
+			}
+			return errors.Wrapf(err, "failed to make node %s %sschedulable", node, prefix)
 		}
-		return nil
+		return wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
+			err1 := getMasterNodes(cs)
+			if err1 != nil {
+				fmt.Printf("Error getting master nodes: %s\n", err1.Error())
+				return true, err
```
hexfusion: Should this return err1?

Just an optional optimization
```golang
if err := getMasterNodes(cs); err != nil {
   fmt.Printf("Error getting master nodes: %s\n", err.Error())
   return true, err
}
```
that would also allow you to use err vs err1 below.

test/e2e/etcdquorumguard_test.go
```diff
@@ -98,23 +98,44 @@ func makeNodeUnSchedulableOrSchedulable(cs *framework.ClientSet, node string, un
...
+			}
+			fmt.Printf("Node %s not yet %sschedulable\n", node, prefix)
+			return false, nil
+		})
 	}
 }
 
 func makeAllNodesSchedulable(cs *framework.ClientSet) error {
+	err := getMasterNodes(cs)
+	if err != nil {
```
hexfusion: same as above as the scope of err is only within this check

test/e2e/etcdquorumguard_test.go
RobertKrawitz: +1

hexfusion:
@cgwalters one more here please sir. :)

hexfusion:
/hold 

will hold for @RobertKrawitz

hexfusion:
https://github.com/openshift/machine-config-operator/blob/8d7a17a38c50a83fc582abc426cbc30bb6f6c208/test/e2e/etcdquorumguard_test.go#L182-L186

Is this what you intended?

hexfusion:
I see that works maybe just `return getMasterNodes(cs)`?

RobertKrawitz:
> I see that works maybe just `return getMasterNodes(cs)`?

It's written as intended.  It's a bit confusing, but the logic is to always update the list of master nodes even if there was an earlier error.  If there was an error from evictEtcdQuotaGuardPodsFromNode I want to return that, otherwise if getMasterNodes took an error I want to return that one.

install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
```diff
@@ -49,16 +50,21 @@ spec:
...
         - mountPath: /mnt/kube
           name: kubecerts
         command:
-        - "/bin/sleep"
+        - /bin/bash
         args:
-        - "infinity"
+        - -c
+        - |
+          # properly handle TERM and exit as soon as it is signaled
```
cgwalters: We don't need to dive deep into this now, but...man this code.  Why isn't `sleep` just dying on `SIGTERM`?  I think this *might* be related to something in `runc` setting `SIGTERM` to `SIG_IGN` and passing that onto its children; not totally sure yet.

RobertKrawitz: The issue that Clayton noted is that pid1 -- and the command run by the container has pid #1 -- is treated specially by Linux; it ignores all signals unless the process specifically registers a handler for the signal.  IIRC this is to prevent `init` (which is `systemd` these days) from being killed and immediately rebooting the system, as older UNIX systems did.

I think we could simply make the command something like

```
/bin/sh -c 'trap exit TERM; sleep infinity'
```

without all the other stuff.

hexfusion:
/lgtm
/hold
@RobertKrawitz unhold when ready

RobertKrawitz:
@smarterclayton do you want to review this, or can I just remove the hold?

runcom:
ok I need to manually cherry-pick

RobertKrawitz:
release-4.1 doesn't appear to have the quorum guard at all:

```
[3(rkrawitz)|||{!51}<rkrawitz>/home/rkrawitz/go/src/github.com/openshift/machine-config-operator]
$ git status
On branch release-4.1
Your branch is up to date with 'upstream/release-4.1'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        install/X/
        install/image-references.orig
        install/image-references.rej
        install/log-bundle.tar.gz
        install/rkrawitz-libvirt/
        mdiff
        mdiffs
        pkg/operator/operator.go.orig
        pkg/operator/operator.go.rej
        pkg/operator/sync.go.orig

nothing added to commit but untracked files present (use "git add" to track)
[3(rkrawitz)|||{!52}<rkrawitz>/home/rkrawitz/go/src/github.com/openshift/machine-config-operator]
$ git ls-files |grep quorum
[3(rkrawitz)|||{!53}<rkrawitz>/home/rkrawitz/go/src/github.com/openshift/machine-config-operator]
```

Issue #782, opened 2019-05-20
after replacing default config file, removing the MC causes the config file to disappear


miabbott:
Using OpenShift 4.1.0-rc.5:

1.  Create a MC that replaces `/etc/containers/policy.json`.  In the example below, this is the default contents of `/etc/containers/policy.json` run through `jq -c` and requoted:

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: worker
  name: 50-policy-json
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,%7B%22default%22%3A%5B%7B%22type%22%3A%22insecureAcceptAnything%22%7D%5D%2C%22transports%22%3A%7B%22docker-daemon%22%3A%7B%22%22%3A%5B%7B%22type%22%3A%22insecureAcceptAnything%22%7D%5D%7D%7D%7D
        filesystem: root
        mode: 0644
        path: /etc/containers/policy.json
```

2.  Wait for nodes to be updated
3.  Remove the MC
4.  Wait for nodes to be updated
5.  Check the nodes and find that `/etc/containers/policy.json` is missing

FWIW, `/etc/containers/policy.json` is owned by the `containers-common` package

cgwalters:
To solve this generally, on an ostree-based system we need to restore the defaults from `/usr/etc` when the new MC stops modifying a file created by the old one.  But we don't have `/usr/etc` on classic yum-based systems :cry: ...I guess we could instead copy the originals to `.mcdorig` when we update a file?

jlebon:
Hmm, is this a liability for future updates? Right now, we need to make sure that any new MC we generate does not stop modifying pre-existing files until this is fixed, it seems? (I.e. re-encoding the default file, rather than removing the file entry entirely). Which... seems like a very easy bit to trip up on :grimacing:

runcom:
I'm adding a test (and a fix for this following Colin's suggestion)

runcom:
PR here https://github.com/openshift/machine-config-operator/pull/797

PR #790, opened 2019-05-22
controller: Add a `testrender` command

cgwalters:
This will be used for unit testing; it accepts a `ct/` directory of
`.yaml` CT files (a bit like our builtin templates but without
the actual templating), as well as an `inline/` subdirectory,
and generates a MachineConfig as JSON.

The value of `inline/` is that for world-readable regular files
you don't need to indirect through YAML.


cgwalters:
```
$ _output/linux/amd64/machine-config-controller testrender --srcdir ./test/daemon/mc/ | jq
{
  "metadata": {
    "name": "testmc",
    "creationTimestamp": null,
    "labels": {
      "machineconfiguration.openshift.io/role": "worker"
    }
  },
  "spec": {
    "osImageURL": "",
    "config": {
      "ignition": {
        "config": {},
        "security": {
          "tls": {}
        },
        "timeouts": {},
        "version": "2.2.0"
      },
      "networkd": {},
      "passwd": {},
      "storage": {
        "files": [
          {
            "filesystem": "root",
            "path": "/etc/tmpfiles.d/cleanup-cni.conf",
            "contents": {
              "source": "data:,r%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F80-openshift-network.conf%0Ar%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F10-ovn-kubernetes.conf%0Ar%20%2Fetc%2Fkubernetes%2Fcni%2Fnet.d%2F00-multus.conf%0A",
              "verification": {}
            },
            "mode": 420
          },
          {
            "filesystem": "root",
            "path": "/etc/sysctl.d/forward.conf",
            "contents": {
              "source": "data:,net.ipv4.ip_forward%20%3D%201%0A",
              "verification": {}
            },
            "mode": 420
          },
          {
            "filesystem": "root",
            "path": "/usr/share/foo.txt",
            "contents": {
              "source": "data:,foo%0A",
              "verification": {}
            },
            "mode": 420
          }
        ]
      },
      "systemd": {}
    }
  }
}

```

cgwalters:
Basically I want to add some shell-script based tests that run the MCD in a once-from like mode and just test `updateFiles`.

Specifically with an eye to fixing https://github.com/openshift/machine-config-operator/issues/782
(which would also require stubbing out a more ostree-like system with `usr/etc` etc.)

cmd/machine-config-controller/testrender.go
```diff
@@ -0,0 +1,153 @@
+package main
```
runcom: it's ok since we're not compiling this anywhere but could we move this out of cmd :thinking:

kikisdeliveryservice:

Feel free to reopen.

PR #791, opened 2019-05-23
DR: backup etcd env file during snapshot restore

vrutkovs:
During snapshot restore this file may be deleted when pods are stopped, 
so restore_snapshot should use backed up copy of this file

/cc @hexfusion @runcom


hexfusion:
/hold

@vrutkovs see if https://github.com/openshift/machine-config-operator/pull/793 resolves your problem. In the case that DNS somehow changes I think I would rather validate against DNS vs use backup if possible. Let me know your thoughts.

vrutkovs:
Closing as #793 is a better solution to the same problem - `/run/etcd/environment` would be generated if it doesn't exist

PR #793, opened 2019-05-23
DR: add validate_environment to openshift-recovery tools

hexfusion:
/run/etcd/environment contains important ENV variables required by most of the DR scripts. This file is populated by the discovery container in the etcd-member init. In some circumstances such as the node rebooting this file can be missing.  `validate_environment` now checks for the existence of the file and if it is missing we populate.

The second part of this PR makes $SNAPSHOT_FILE a required parameter. Currently, we default to the local members DB file backup. This I feel is not a good default and could result in unexpected results.

/cc @vrutkovs

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -4,47 +4,58 @@ path: "/usr/local/bin/etcd-snapshot-restore.sh"
 contents:
   inline: |
     #!/usr/bin/env bash
-    
+
     set -o errexit
     set -o pipefail
-    
+
     # example
-    
     # etcd-snapshot-restore.sh $path-to-snapshot $etcd-connection-string
+
     if [[ $EUID -ne 0 ]]; then
       echo "This script must be run as root"
       exit 1
     fi
-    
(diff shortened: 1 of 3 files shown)


hexfusion:
/hold 

holding for manual validation and @vrutkovs review

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
```diff
@@ -4,47 +4,58 @@ path: "/usr/local/bin/etcd-snapshot-restore.sh"
...
     function run {
       init
       dl_etcdctl
       backup_manifest
+      DISCOVERY_DOMAIN=$(grep -oP '(?<=discovery-srv=).*[^"]' $ASSET_DIR/backup/etcd-member.yaml )
+      if [ -z "$DISCOVERY_DOMAIN" ]; then
+        echo "Discovery domain can not be extracted from $ASSET_DIR/backup/etcd-member.yaml"
+        exit 1
+      fi
+      verify_environment
```
vrutkovs:
```suggestion
      validate_environment
```

vrutkovs:
/hold cancel

Verified that it works correctly (with a fix for https://github.com/openshift/machine-config-operator/pull/793#discussion_r287329087 applied)

Issue #629, opened 2019-04-11
Add `timeservers` section to MachineConfig


cgwalters:
MachineConfigs are low level and force host reboots.  Today we have "internal CRDs" like container/kubelet config that "render down" to MCs but there's no reason for changing the kubelet config to force a reboot.

Related to this, I think we should go down the path of adding some "high level" options for the operating system to MachineConfig.  A bit like the [kernel args](https://github.com/openshift/machine-config-operator/issues/388) proposal except let's add e.g.

```
diff --git a/pkg/apis/machineconfiguration.openshift.io/v1/types.go b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
index 33cca60..236d409 100644
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -229,6 +229,8 @@ type MachineConfigSpec struct {
 	OSImageURL string `json:"osImageURL"`
 	// Config is a Ignition Config object.
 	Config ignv2_2types.Config `json:"config"`
+
+	NTPServers []string `json:"ntpServers"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```

Or it could be a new `MachineTimeServers` CRD though that feels a bit like overkill unless we really need to expose a lot of options.

If we had this it'd be much easier for the MCD to say "oh it's just the timeservers changing, let's not reboot the host".

runcom:
> Or it could be a new MachineTimeServers CRD though that feels a bit like overkill unless we really need to expose a lot of options.

I would go CRDs as the MachineConfigSpec could grow with kargs, this and maybe more and it makes more sense from a UI pov that we have new types so that `oc get` can be helpful as well (might be just me, not a big deal either way for me)

cgwalters:
(Of course though for time...maybe we should only use chrony as a bootstrap and have workers get time from the control plane or something)

JAORMX:
@cgwalters what do you mean by getting the time from the control plane?

JAORMX:
It would be very beneficial for us to be able to configure chronyd and rely on it setting up the correct time for the host, given that it would help us meet some compliance controls.

e.g. For FedRAMP, we need to comply with [AU-8](https://nvd.nist.gov/800-53/Rev4/control/AU-8), which specifies the following:

> The information system:
[...]

PR #795, opened 2019-05-28
baremetal: Add coredns, mdns-publisher, haproxy, keepalived static pods

bcrochet:
Adds pods to master and worker nodes as appropriate for baremetal deployment.

Requires: #943

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -89,9 +89,14 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 			MachineOSContent:      bootstrapOpts.oscontentImage,
 		},
 		ControllerConfigImages: operator.ControllerConfigImages{
-			Etcd:            bootstrapOpts.etcdImage,
-			InfraImage:      bootstrapOpts.infraImage,
-			KubeClientAgent: bootstrapOpts.kubeClientAgentImage,
+			Etcd:                bootstrapOpts.etcdImage,
+			InfraImage:          bootstrapOpts.infraImage,
+			KubeClientAgent:     bootstrapOpts.kubeClientAgentImage,
+			Keepalived:          bootstrapOpts.keepalivedImage,
+			Coredns:             bootstrapOpts.corednsImage,
+			MdnsPublisher:       bootstrapOpts.mdnsPublisherImage,
+			Haproxy:             bootstrapOpts.haproxyImage,
+			BaremetalRuntimeCfg: bootstrapOpts.baremetalRuntimeCfgImage,
 		},
 	}
 
--- a/templates/common/baremetal/files/baremetal-coredns.yaml
+++ b/templates/common/baremetal/files/baremetal-coredns.yaml
@@ -1,0 +1,91 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/kubernetes/manifests/coredns.yaml"
+contents:
+  inline: |
+    kind: Pod
+    apiVersion: v1
+    metadata:
+      name: coredns
+      namespace: openshift-kni-infra
+      creationTimestamp:
+      deletionGracePeriodSeconds: 65
+      labels:
+        app: kni-infra-mdns
+    spec:
+      volumes:
+      - name: resource-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/coredns"
+      - name: kubeconfig
+        hostPath:
+          path: "/etc/kubernetes/kubeconfig"
+      - name: conf-dir
+        hostPath:
+          path: "/etc/coredns"
+      initContainers:
+      - name: render-config
+        image: {{ .Images.baremetalRuntimeCfgImage }}
+        command:
+        - runtimecfg
+        - render
+        - "/etc/kubernetes/kubeconfig"
+        - "--api-vip"
+        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
+        - "--dns-vip"
+        - "{{ .Infra.Status.PlatformStatus.BareMetal.NodeDNSIP }}"
+        - "--ingress-vip"
+        - "{{ .Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
+        - "/config"
+        - "--out-dir"
+        - "/etc/coredns"
+        resources: {}
+        volumeMounts:
+        - name: kubeconfig
+          mountPath: "/etc/kubernetes/kubeconfig"
+        - name: resource-dir
+          mountPath: "/config"
+        - name: conf-dir
+          mountPath: "/etc/coredns"
+        imagePullPolicy: IfNotPresent
+      containers:
+      - name: coredns
+        securityContext:
+          privileged: true
+        image: {{.Images.corednsImage}}
+        args:
+        - "--conf"
+        - "/etc/coredns/Corefile"
+        resources:
+          requests:
+            cpu: 150m
+            memory: 1Gi
+        volumeMounts:
+        - name: conf-dir
+          mountPath: "/etc/coredns"
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          successThreshold: 1
+          failureThreshold: 3
+          timeoutSeconds: 10
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 60
+          timeoutSeconds: 5
+          successThreshold: 1
+          failureThreshold: 5
+        terminationMessagePolicy: FallbackToLogsOnError
+        imagePullPolicy: IfNotPresent
+      hostNetwork: true
+      tolerations:
+      - operator: Exists
+      priorityClassName: system-node-critical
+    status: {}
--- a/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
@@ -1,0 +1,107 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/kubernetes/manifests/haproxy.yaml"
+contents:
+  inline: |
+    kind: Pod
+    apiVersion: v1
+    metadata:
+      name: haproxy
+      namespace: openshift-kni-infra
+      creationTimestamp:
+      deletionGracePeriodSeconds: 65
+      labels:
+        app: kni-infra-api-lb
+    spec:
+      volumes:
+      - name: resource-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/haproxy"
+      - name: kubeconfig
+        hostPath:
+          path: "/etc/kubernetes/kubeconfig"
+      - name: run-dir
+        empty-dir: {}
+      - name: conf-dir
+        hostPath:
+          path: "/etc/haproxy"
+      - name: chroot-host
+        hostPath:
+          path: "/"
+      containers:
+      - name: haproxy
+        image: {{.Images.haproxyImage}}
+        command:
+        - "/bin/bash"
+        - "-c"
+        - |
+          #/bin/bash
+          reload_haproxy()
+          {
+            old_pids=$(pidof haproxy)
+            if [ -n "$old_pids" ]; then
+                /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid -x /var/lib/haproxy/run/haproxy.sock -sf $old_pids &
+            else
+                /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid &
+            fi
+          }
+
+          msg_handler()
+          {
+            while read -r line; do
+              echo "The client send: $line"  >&2
+              # currently only 'reload' msg is supported
+              if [ "$line" = reload ]; then
+                  reload_haproxy
+              fi
+            done
+          }
+          set -ex
+          declare -r haproxy_sock="/var/run/haproxy/haproxy-master.sock"
+          export -f msg_handler
+          export -f reload_haproxy
+          if [ -S "$haproxy_sock" ]; then
+              rm "$haproxy_sock"
+          fi
+          socat UNIX-LISTEN:${haproxy_sock},fork system:'bash -c msg_handler'
+        volumeMounts:
+        - name: conf-dir
+          mountPath: "/etc/haproxy"
+        - name: run-dir
+          mountPath: "/var/run/haproxy"
+        livenessProbe:
+          initialDelaySeconds: 10
+          httpGet:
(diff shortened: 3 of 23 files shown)


templates/master/00-master/baremetal/files/baremetal-chrony.yaml
```diff
@@ -0,0 +1,48 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/chrony.conf"
+contents:
+  inline: |
+    #
+    # Configure an NTP server accessible on the redhat network.
```
cgwalters: This isn't suitable for upstreaming.  See also https://github.com/openshift/machine-config-operator/issues/629

templates/master/00-master/baremetal/files/baremetal-clusterinfo.yaml
```diff
@@ -0,0 +1,18 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/clusterinfo"
```
cgwalters: We should probably ship stuff like this with RHCOS generically across platforms.

templates/master/00-master/baremetal/files/baremetal-fletcher8.yaml
```diff
@@ -0,0 +1,16 @@
...
+    #!/usr/libexec/platform-python
+    import sys
+    
+    data = map(ord, sys.argv[1])
+    ckA = ckB = 0
+    
+    for b in data:
+        ckA = (ckA + b) & 0xf
+        ckB = (ckB + ckA) & 0xf
+    print((ckB << 4) | ckA )
```
cgwalters: This file could use some sort of comment with what it's doing.  I don't find `fletcher8` very self-explanatory :wink:

templates/master/00-master/baremetal/files/baremetal-get-vip-subnet-cidr.yaml
```diff
@@ -0,0 +1,30 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/get_vip_subnet_cidr"
```
cgwalters: Ditto.

If you have a collection of Python helpers, maybe simpler to put them in one file with subcommands?

bcrochet:
> First, this duplicates/overlaps #740 right?
> 
> The `.gitkeep` files shouldn't be necessary anymore.

There is some little overlap, but does not directly duplicate. If you have a suggestion for de-duping between just two platforms, that would be helpful.

I'll remove the .gitkeep files.

templates/master/00-master/baremetal/files/baremetal-chrony.yaml
bcrochet: I'll remove this.

templates/master/00-master/baremetal/files/baremetal-fletcher8.yaml
bcrochet: TBH, neither do I. :)

This commit is just moving a bunch of stuff from openshift-metal3/dev-scripts. It could use some cleanup.

templates/master/00-master/baremetal/files/baremetal-get-vip-subnet-cidr.yaml
bcrochet: That's a good suggestion. Thanks.

templates/master/00-master/baremetal/files/baremetal-clusterinfo.yaml
bcrochet: Any pointers on doing so would be appreciated.

templates/master/00-master/baremetal/files/baremetal-fletcher8.yaml
bcrochet: It generates an 8-bit id for the keepalived routers. I'll add a comment stating such.

templates/master/00-master/baremetal/files/baremetal-clusterinfo.yaml
rgolangh: @cgwalters I have the same needs, I need DNS_VIP, API_VIP, DOMAIN and so on and I'm suggesting to add the `install-config.yaml` values to RenderConfig and by that we would have access to all the data coming from the installer, during render time. All places in the code will become just:
```
{{ .Domain }}
{{ (split "." .Domain -1) | first }}   // for the host
{{ .Platform.SomeSurveryValue }}
```

cgwalters:
Sorry about the delay on this...there's a lot going on and we may need periodic pings if PRs are important/ready.

I feel like this would probably be best as a separate git repository, with the contents built into one or more container images.  The relevant code in the MCO would boil down to injecting the static pod entries, much like is done for etcd.

Among other things, that would avoid the awkward shell-script-in-yaml-in-yaml and more generally allow your team to easily iterate on the content without blocking on us here.

bcrochet:
> Sorry about the delay on this...there's a lot going on and we may need periodic pings if PRs are important/ready.
> 
> I feel like this would probably be best as a separate git repository, with the contents built into one or more container images. The relevant code in the MCO would boil down to injecting the static pod entries, much like is done for etcd.
> 
> Among other things, that would avoid the awkward shell-script-in-yaml-in-yaml and more generally allow your team to easily iterate on the content without blocking on us here.

We are working on the amount of shell-scripty things. Is there an existing component with such a repo that I can look at?

cgwalters:
Since the runway is relatively short here for 4.2, we discussed landing this roughly as is and iterating from there.

One argument to having this be a separate operator/controller was that the MCO's role here should be more around "infrastructure" rather than platform-specific code, but on the other hand in practice actually the MCO codebase has a lot of convenient parts for this (templating MachineConfig files based on data from the cluster config), plus its already extant support for platform conditionals for MachineConfig, etc.

Another aspect is that these templates also need to run at install time for the masters, and introducing a new operator/controller rendering MachineConfigs would require changes to the installer's `bootkube.sh` (not complex changes though).

templates/common/baremetal/files/baremetal-coredns.yaml
```diff
@@ -0,0 +1,93 @@
...
+        - name: resource-dir
+          mountPath: "/etc/kubernetes/static-pod-resources"
+        - name: conf-dir
+          mountPath: "/etc/coredns"
+        imagePullPolicy: IfNotPresent
+      containers:
+      - name: coredns
+        securityContext:
+          privileged: true
+        image: quay.io/openshift-metalkube/coredns-mdns:latest
```
cgwalters: I thought I mentioned this before but am not seeing it...

There is one thing that I think blocks landing here - these images need to be part of the release image and substituted by the MCO's template machinery, just like the `etcd` and `pod` images for example.  The current state means installs on metal aren't reproducible/versioned, controlled by the CVO, etc.

You want to add them to `image-references`, and at the same time I'd probably rename them to a "dummy" value like the others, i.e. make them consistently `registry.svc.ci.openshift.org/openshift:coredns-mdns` etc.

I think it may require intervention from a release admin (Clayton?) to add external images to the payload as well.

celebdor: coredns will be there soon. Hopefully today

runcom: holding this PR till we have images before landing this

runcom:
/hold

per comment https://github.com/openshift/machine-config-operator/pull/795#discussion_r301587696

cmd/machine-config-operator/bootstrap.go
```diff
@@ -72,6 +74,8 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.networkConfigFile, "network-config-file", "/assets/manifests/cluster-network-02-config.yml", "File containing network.config.openshift.io manifest.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.cloudConfigFile, "cloud-config-file", "", "File containing the config map that contains the cloud config for cloudprovider.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.proxyConfigFile, "proxy-config-file", "/assets/manifests/cluster-proxy-01-config.yaml", "File containing proxy.config.openshift.io manifest.")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.keepalivedImage, "keepalived-image", "registry.svc.ci.openshift.org/openshift:keepalived-ipfailover", "Image for Keepalived.")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.corednsMDNSImage, "coredns-mdns-image", "registry.svc.ci.openshift.org/openshift:coredns-mdns", "Image for CoreDNS-mDNS.")
```
celebdor: The image will be coredns, we have a PR to itegrate it there.

templates/master/01-master-kubelet/baremetal/units/kubelet.yaml
```diff
@@ -1,36 +0,0 @@
-name: "kubelet.service"
```
markmc: isn't this undoing the remove-noschedule-taint part of openshift/machine-config-operator#846 ?

bcrochet: I don't think that's supposed to be here. Not really certain how that snuck in. Good catch. Thanks @markmc

templates/common/baremetal/files/baremetal-coredns.yaml
bcrochet: I think I've got all of the image references correct now. /cc @cgwalters @runcom

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -0,0 +1,185 @@
...
+        - "/etc/kubernetes/kubeconfig"
+        - "/etc/kubernetes/static-pod-resources/clusterrc"
+        resources: {}
+        volumeMounts:
+        - name: resource-dir
+          mountPath: "/etc/kubernetes/static-pod-resources"
+        - name: kubeconfig
+          mountPath: "/etc/kubernetes/kubeconfig"
+      - name: render-haproxy-cfg
+        image: quay.io/openshift/origin-node:latest
```
cgwalters: This image (used in many places in this PR) doesn't seem like it's actually part of our `image-references`.  `quay.io/openshift/origin-cli:latest` is however already used by the etcd quorum guard, and if you just want to run some shell script, let's use that.

templates/common/baremetal/files/baremetal-coredns.yaml
```diff
@@ -0,0 +1,111 @@
...
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/coredns"
+      - name: kubeconfig
+        hostPath:
+          path: "/etc/kubernetes/kubeconfig"
+      - name: conf-dir
+        empty-dir: {}
+      initContainers:
+      - name: clusterrc-generation
+        image: quay.io/openshift-metalkube/kubeconfig-extractor:latest
```
cgwalters: This image also needs to be part of the release image.

cgwalters:
Looking a lot better.  I spotted at least two images that still need pinning.  I know this is a bit painful, but it's critical for reproducibility and also for supporting "offline/disconnected" installs which I think has a strong intersection with people who want to run on metal.

(The `verify` failure is a known issue, we're looking at it)

install/image-references
```diff
@@ -48,3 +48,23 @@ spec:
...
+      name: registry.svc.ci.openshift.org/openshift:coredns
+  - name: mdns-publisher
+    from:
+      kind: DockerImage
+      name: registry.svc.ci.openshift.org/openshift:mdns-publisher
+  - name: haproxy-router
+    from:
+      kind: DockerImage
+      name: registry.svc.ci.openshift.org/openshift:haproxy-router
+  - name: baremetal-runtimecfg
```
smarterclayton: I raised this on the PR but this name doesn't seem to make sense.  What is the name that is long term stable for this?  What does it own?  As a regular user, how do I know what is in here?

markmc:
xref: openshift/release#4364

install/image-references
celebdor: Discussed in slack
https://coreos.slack.com/archives/CBN38N3MW/p1563379222233400?thread_ts=1562941823.460800&cid=CBN38N3MW
https://coreos.slack.com/archives/CBN38N3MW/p1563379226233600?thread_ts=1562941823.460800&cid=CBN38N3MW

templates/worker/00-worker/baremetal/files/NetworkManager-dns-vip-prepender-worker.yaml
```diff
@@ -0,0 +1,45 @@
...
+  inline: |
+    #!/bin/bash
+
+    IFACE=$1
+    STATUS=$2
+
+    case "$STATUS" in
+        pre-up)
+        logger -s "NM dns-vip-prepender-worker triggered by pre-upping ${1}."
+        DNS_VIP="127.0.0.1"
```
celebdor: This IP should be from runtime. We used to have some python script to obtain it.

celebdor: You actually have the script included below, but are not calling it :-)

templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml
```diff
@@ -0,0 +1,7 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/dhcp/dhclient.conf"
+contents:
+  inline: |
+    supersede domain-search "{{ .EtcdDiscoveryDomain }}";
+    prepend domain-name-servers {{`{{ .NonVirtualIP }}`}};
```
celebdor: Who will render this? DId you run baremeta-runtimecfg over this file? IIRC the prepender script was in charge for that.

bcrochet: > Who will render this? DId you run baremeta-runtimecfg over this file? IIRC the prepender script was in charge for that.

It's rendered directly. This is why I don't think we need the prepender anymore. It will be there from the first boot.

templates/worker/00-worker/baremetal/files/NetworkManager-dns-vip-prepender-worker.yaml
bcrochet: I think we need can just get rid of the prependers anyway.

templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml
celebdor: resolved with non-virtual-ip-prepender

templates/common/baremetal/files/baremetal-coredns.yaml
celebdor: @runcom I guess we are waiting for the separate approved PR to merge to remove the hold. Right?

celebdor: The image references patch got merged.

bcrochet:
@runcom Review please.

runcom:
an exception has been granted for this so
/lgtm

PR #797, opened 2019-05-28
pkg/daemon: do not delete host files when deleting an MC

runcom:
Deploying an MC changing a file and later deleting it causes the file to
be deleted (it could have been already on disk i.e. shipped by an RPM).
Following Colin's suggestion, this patch adds a backup mechanism which,
when deleting an MC, causes the old file to be restored.
Added an e2e test as well.

closes https://github.com/openshift/machine-config-operator/issues/782

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -485,6 +485,13 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
+			if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
+					return err
+				}
+				glog.V(2).Infof("Restored file %q", f.Path)
+				continue
+			}
 			glog.V(2).Infof("Deleting stale config file: %s", f.Path)
 			if err := os.Remove(f.Path); err != nil {
 				newErr := fmt.Errorf("unable to delete %s: %s", f.Path, err)
@@ -680,9 +687,32 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
 				return fmt.Errorf("failed to retrieve file ownership for file %q: %v", file.Path, err)
 			}
 		}
+		if err := createOrigFile(file.Path); err != nil {
+			return err
+		}
 		if err := writeFileAtomically(file.Path, contents.Data, defaultDirectoryPermissions, mode, uid, gid); err != nil {
 			return err
 		}
+	}
+	return nil
+}
+
+func origFileName(fpath string) string {
+	return fpath + ".mcdorig"
+}
+
+func createOrigFile(fpath string) error {
+	if _, err := os.Stat(fpath); err != nil {
+		// the file isn't there, no need to back it up
+		// we could check ENOENT only maybe?
+		return nil
+	}
+	if _, err := os.Stat(origFileName(fpath)); err == nil {
+		// the orig file is already there and we avoid creating a new one to preserve the real default
+		return nil
+	}
+	if out, err := exec.Command("cp", "-a", "--reflink=auto", fpath, origFileName(fpath)).CombinedOutput(); err != nil {
+		return errors.Wrapf(err, "creating orig file for %q: %s", fpath, string(out))
 	}
 	return nil
 }
@@ -781,12 +811,12 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 
 // RHEL 7.6 logger (util-linux) doesn't have the --journald flag
 func (dn *Daemon) isLoggingToJournalSupported() bool {
+	if dn.OperatingSystem == machineConfigDaemonOSRHEL {
+		return true
+	}
 	loggerOutput, err := exec.Command("logger", "--help").CombinedOutput()
 	if err != nil {
 		dn.logSystem("error running logger --help: %v", err)
-		if dn.OperatingSystem == machineConfigDaemonOSRHCOS {
-			return true
-		}
 		return false
 	}
 	return strings.Contains(string(loggerOutput), "--journald")
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -482,3 +482,68 @@ func TestReconcileAfterBadMC(t *testing.T) {
 		t.Errorf("machine config didn't roll back on any worker: %v", err)
 	}
 }
+
+func TestDontDeleteRPMFiles(t *testing.T) {
+	cs := framework.NewClientSet("")
+	bumpPoolMaxUnavailableTo(t, cs, 3)
+
+	mcHostFile := createMCToAddFile("modify-host-file", "/etc/motd", "mco-test", "root")
+
+	// grab the initial machineconfig used by the worker pool
+	// this MC is gonna be the one which is going to be reapplied once the previous MC is deleted
+	mcp, err := cs.MachineConfigPools().Get("worker", metav1.GetOptions{})
+	if err != nil {
+		t.Error(err)
+	}
+	workerOldMc := mcp.Status.Configuration.Name
+
+	// create the dummy MC now
+	_, err = cs.MachineConfigs().Create(mcHostFile)
+	if err != nil {
+		t.Errorf("failed to create machine config %v", err)
+	}
+
+	renderedConfig, err := waitForRenderedConfig(t, cs, "worker", mcHostFile.Name)
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+
+	// wait for the mcp to go back to previous config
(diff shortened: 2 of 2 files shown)


runcom:
Thinking of any way this could break upgrades/backwards compatibily now lol

pkg/daemon/update.go
```diff
@@ -477,6 +478,13 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
+			if _, err := os.Stat(f.Path + ".mcdorig"); err == nil {
+				if err := os.Rename(f.Path+".mcdorig", f.Path); err != nil {
+					return err
+				}
+				glog.V(2).Infof("Restored file %q", f.Path)
+				continue
+			}
```
runcom: might as well wrap this in a func

pkg/daemon/update.go
```diff
@@ -672,13 +680,37 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
...
 
+func createOrigFile(fpath string) error {
+	f, err := os.Open(fpath)
+	if err != nil {
+		// the file isn't there, no need to back it up
+		// we could check ENOENT only maybe?
+		return nil
+	}
+	if err := os.MkdirAll(filepath.Dir(fpath), defaultDirectoryPermissions); err != nil {
+		return fmt.Errorf("failed to create directory %q: %v", filepath.Dir(fpath), err)
```
cgwalters: I prefer `errors.Wrapf()` personally.

pkg/daemon/update.go
```diff
@@ -672,13 +680,37 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
 				return fmt.Errorf("failed to retrieve file ownership for file %q: %v", file.Path, err)
 			}
 		}
+		if err := createOrigFile(file.Path); err != nil {
```
cgwalters: We should *only* do this if the old config doesn't have the file.  Otherwise, in the case of applying configs A, B, C where A and B both modify a file and C doesn't, we'll end up with the content of config A's file, not the true default right?

(Or maybe simplest to implement, only create the backup file if it doesn't already exist)

pkg/daemon/update.go
```diff
@@ -406,12 +407,12 @@ func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
 	cmdArgs := []string{}
 	for _, arg := range oldConfig.Spec.KernelArguments {
 		if !newKargs[arg] {
-			cmdArgs = append(cmdArgs, "--delete=" + arg)
+			cmdArgs = append(cmdArgs, "--delete="+arg)
```
cgwalters: May be worth splitting out formatting stuff in a prep patch.

ashcrow: If it's not too much work, I agree :+1:

pkg/daemon/update.go
```diff
@@ -672,13 +680,37 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
...
+	}
+	if err := os.MkdirAll(filepath.Dir(fpath), defaultDirectoryPermissions); err != nil {
+		return fmt.Errorf("failed to create directory %q: %v", filepath.Dir(fpath), err)
+	}
+	t, err := renameio.TempFile("", fpath+".mcdorig")
+	if err != nil {
+		return err
+	}
+	defer t.Cleanup()
+	if _, err := io.Copy(t, f); err != nil {
```
ashcrow: Do we need to check permissions for what was originally used at install time? IE: `rpm -q --dump $PACKAGE | grep $FILE_PATH`? Or do we believe that defaults are sufficient?

pkg/daemon/update.go
runcom: My vscode is doing this automatically somehow, need to figure that out first lol

cgwalters:
@runcom Are you going to get back to this PR?  Pretty important bug I think, also I realized in later discussion that this would help us support the Ignition `append` mode.

runcom:
> @runcom Are you going to get back to this PR? Pretty important bug I think, also I realized in later discussion that this would help us support the Ignition `append` mode.

yep, been dragged and buried into something else (mainly bugzillas and next features), I'll get back on this shortly.

pkg/daemon/update.go
runcom: I believe default are sufficient and umask should kick in if the parent directory is there maybe 🤔 I'll check this out

runcom:
Updated and ready for another round of reviews now

pkg/daemon/update.go

pkg/daemon/update.go
cgwalters: umask doesn't have anything to do with parent directories.

Since we do have some secret files (like the pull secret) we definitely need to be matching the permissions.  Also, some of the files we write need to be executable.

cgwalters: Personally I'd just fork off `exec.Command("cp", "-a", "--reflink=auto", fpath, origFileName(fpath))` since among other things that will automatically use reflinks which `io.Copy()` doesn't know about.  (See also [this code](https://gitlab.gnome.org/GNOME/libglnx/blob/8f34033b256da5a8e2cc66faf8e81cfdfb9cdf37/glnx-fdio.c#L712) which I stole from systemd).

runcom: Uhm ok I'll shell out to cp directly

cgwalters: Of course...since we're about to be *replacing* this file just after this, we could simply `rename()` - the problem with that will be a small window where the file doesn't exist, so the more correct thing would be to use [RENAME_EXCHANGE](https://gitlab.gnome.org/GNOME/libglnx/blob/8f34033b256da5a8e2cc66faf8e81cfdfb9cdf37/glnx-fdio.c#L120).

runcom: Fixed by forking to cp. thanks Colin & Steve

knobunc:
Image build failing: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/797/pull-ci-openshift-machine-config-operator-master-images/2698

```
could not wait for build: the build machine-config-daemon failed after 45s with reason DockerBuildFailed: Docker build strategy has failed.

Using version from git...
Building github.com/openshift/machine-config-operator/cmd/...-gcfa394e4-dirty, cfa394e44bc9f4f52330fdedc599eaacdab0102f)
pkg/daemon/update.go:8:2: imported and not used: "io"
error: build error: running 'WHAT=machine-config-daemon ./hack/build-go.sh' failed with exit code 2
```

runcom:
> Image build failing: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/797/pull-ci-openshift-machine-config-operator-master-images/2698

yup, error on my side :P

cgwalters:
A brief search turns up https://github.com/docker/docker-ce/blob/9f7430572607260b058cc69c03da91278beac746/components/engine/daemon/graphdriver/copy/copy.go#L30 which looks reasonable if we ever want to copy some code to do this in-process.

Rust also recently gained a [good implementation of this](https://github.com/rust-lang/rust/blob/96636f3eb8206f7499451ebc5b347d000d7aee08/src/libstd/sys/unix/fs.rs#L873) FWIW.

runcom:
> A brief search turns up https://github.com/docker/docker-ce/blob/9f7430572607260b058cc69c03da91278beac746/components/engine/daemon/graphdriver/copy/copy.go#L30 which looks reasonable if we ever want to copy some code to do this in-process.

I'm familiar with that piece of code, also willing to use that, I'm just worried about the maintenance status of anything within Docker/Moby.

pkg/daemon/update.go
```diff
@@ -680,13 +687,36 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
...
+	if _, err := os.Stat(fpath); err != nil {
+		// the file isn't there, no need to back it up
+		// we could check ENOENT only maybe?
+		return nil
+	}
+	if _, err := os.Stat(origFileName(fpath)); err == nil {
+		// the orig file is already there and we avoid creating a new one to preserve the real default
+		return nil
+	}
+	if out, err := exec.Command("cp", "-a", "--reflink=auto", fpath, origFileName(fpath)).CombinedOutput(); err != nil {
```
cgwalters: Of course if this is interrupted in the middle we'll end up with an incomplete backup but...let's just get this in!

cgwalters: The [libglnx code](https://gitlab.gnome.org/GNOME/libglnx/blob/8f34033b256da5a8e2cc66faf8e81cfdfb9cdf37/glnx-fdio.c#L930) I wrote for this properly clones to an `O_TMPFILE` which also makes it superior to what the Rust libstd is doing.  It's a bit surprising actually `cp` doesn't have a `--tmpfile` flag or something, would probably be easy to implement.

If only calling C from golang wasn't unspeakably awful...

Anyways we can circle back to polishing this bit later.

PR #800, opened 2019-05-29
WIP: Add FIPS to MachineConfig

cgwalters:


pkg/daemon/update.go
```diff
@@ -432,6 +432,26 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+
+	arg := "enable"
+	if !newConfig.Spec.FIPS {
+		arg = "disable"
+	}
+
+	cmd := "/usr/libexec/rhcos-tools/coreos-fips"
+	args := []string{arg}
+	dn.logSystem("Running %s %v", cmd, args)
+	return exec.Command(cmd, args...).Run()
```
ashcrow: This looks like the right command to me. One thing to keep in mind is that, within OCP, the `coreos-fips` will not default to updating the kernel args and, instead, defer to MCO to own those args.

cgwalters: Let's debate that in https://github.com/openshift/rhcos-tools/pull/2 since...it seems to me doing it that way would make everything unnecessarily more complex.

ashcrow: I closed the PR which would offload kernel args to the MCO. This call should be good as is.

pkg/daemon/update.go
```diff
@@ -432,6 +432,26 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
+// updateFIPS handles changes in FIPS
+func (dn *Daemon) updateFIPS(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	if oldConfig.Spec.FIPS != newConfig.Spec.FIPS {
```
runcom: wait, shouldn't this be `==`?

runcom: I know this is wip, just leaving also a comment that we still miss a call to `updateFIPS` 💪

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -13,6 +13,7 @@ import (
 // It sorts all the configs in increasing order of their name.
 // It uses the Ignition config from first object as base and appends all the rest.
 // Kernel arguments are concatenated.
+// FIPS uses the last specified value.
```
ericavonb: I think this will be confusing for users, especially because unset values will decode into `false`.  Like, the following would have FIPS mode OFF:
```yaml
metadata:
  name: 00_config_controller
spec:
  config: ...
---
metadata:
  name: 99_config_fips
spec:
  FIPS: true
---
metadata:
  name: 99_config_master
spec:
  osImageURL: thelatestversion
```

runcom: We might want to go pointer then to differentiate between set/unset as well?

ericavonb: Yeah, probably. Another option would be to fail closed, i.e. if ANY configs have FIPS=true, it will be turned on.

Issue #798, opened 2019-05-28
kargs phase 2


cgwalters:
Expanding this: https://github.com/openshift/machine-config-operator/pull/762#issuecomment-493453115

We need to support "early pivot" kargs.  I think the best implementation path for this is to "invert" the MachineConfig and inject it as extra data into the Ignition.  Then we change `pivot` to also accept a `MachineConfig` object; it would handle `osImageURL` and `kernelArguments`.

What would make further sense then I think is to fold `pivot` into the MCO repo and have it basically become the "host portion" of the MCD.  But we don't need to do that right away.

ashcrow:
> Then we change pivot to also accept a MachineConfig object; it would handle osImageURL and kernelArguments.

WFM

>  What would make further sense then I think is to fold pivot into the MCO repo and have it basically become the "host portion" of the MCD. But we don't need to do that right away.

I worry that the MCO repo has a lot already (MCO, MCC, MCS, MCD). I won't block but just something to keep in mind.

cgwalters:
> I worry that the MCO repo has a lot already (MCO, MCC, MCS, MCD). I won't block but just something to keep in mind.

Yeah, on that topic one thing I personally would like to move *out* of the MCO is the "static" templates (i.e. files not templated) like [this one](https://github.com/openshift/machine-config-operator/blob/3ec65501fa6e39ed1e4be3b3439138b39aa6fbbe/templates/common/_base/files/cleanup-cni-conf.yaml) - we could reuse the [tools repo](https://github.com/openshift/machine-config-operator/blob/3ec65501fa6e39ed1e4be3b3439138b39aa6fbbe/templates/common/_base/files/cleanup-cni-conf.yaml)?   It's weird for the MCO to feed files into Ignition that could just be shipped as part of the read-only base OS.

cgwalters:
> Yeah, on that topic one thing I personally would like to move out of the MCO is the "static" templates (i.e. files not templated) like this one - we could reuse the tools repo?

Except we can't do that as is because it'd break BYO RHEL...they'd need some other way to add those files.   Aarrgh...

cgwalters:
Eh, if we stick them in an RPM that would work.

cgwalters:
PR in https://github.com/openshift/machine-config-operator/pull/859

cgwalters:
Important PR in https://github.com/openshift/machine-config-operator/pull/868

[...]

PR #801, opened 2019-05-29
Add a machine-config-daemon.spec

cgwalters:
As we move towards having more things in MachineConfig, we also
in many cases need the functionality during "early pivot" before
the cluster is up.

Let's just ship the MCD as part of the host as well so we can use
the same code then.

Now of course we have the same code in two places.  I think
what we'd want to do after this is have the "pod-MCD" call out
to the "host-MCD" as a subprocess.  That would actually simplify
other things as well.

### DIFF
--- a/machine-config-daemon.spec
+++ b/machine-config-daemon.spec
@@ -1,0 +1,35 @@
+%define debug_package %{nil}
+%global commit          4e75a8f20e5cf44374fd1bf3b3df997b8689d3ff
+%global shortcommit     %(c=%{commit}; echo ${c:0:7})
+
+Name:           machine-config-daemon
+Version:        4.0.0
+Release:        1.rhaos4.2.git%{shortcommit}%{?dist}
+Summary:        https://github.com/openshift/machine-config-operator
+License:        ASL 2.0
+URL:            https://github.com/openshift/machine-config-operator
+Source0:        https://github.com/openshift/machine-config-operator/archive/%{commit}/machine-config-operator-%{shortcommit}.tar.gz
+
+BuildRequires:  git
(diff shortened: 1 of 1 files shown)


cgwalters:
xref https://github.com/openshift/machine-config-operator/issues/798

ashcrow:
spec looks proper to me. One question is how this would be built and provided where RHCOS could consume the result.

cgwalters:
> One question is how this would be built and provided where RHCOS could consume the result.

The easy answer today is that we build it the same way we build pivot right?  (And then a logical step after that is to fold pivot into this repo)

cgwalters:
I started the paperwork to do the RPM builds.

cgwalters:
OK I rebased and did some further tweaks to the spec, which required the `VERSION_OVERRIDE` and `SOURCE_GIT_COMMIT` sourcery¹ too.

What I'd like to do is try to do some of this work incrementally; if we land the spec file I can get the `machine-config-daemon` binary into rhcos, but it won't *do* anything until we start more work.  If the binary is there though it'll just make it easier to hack on without having to custom deploy both RHCOS and the MCD. 

¹ See what I did there?

runcom:
let's get this in to allow more testing then

/lgtm

runcom:
oh this did go in just fine :/

kikisdeliveryservice:
@runcom @cgwalters wait why is ci running on something that merged??

cgwalters:
I think the parts of Prow here are asynchronous, and for non-required contexts it doesn't kill them after the PR has merged.  (For required contexts, the PR needs them to merge)

kikisdeliveryservice:
> I think the parts of Prow here are asynchronous, and for non-required contexts it doesn't kill them after the PR has merged. (For required contexts, the PR needs them to merge)

but...4 hours async? zoinks

PR #803, opened 2019-05-29
Bug 1713389: containerRuntimeConfigController manages policy.json

umohnani8:
Since containerRuntimeConfigController also watches the
images.config.openshift.io CR, it needs to manage
/etc/containers/policy.json as that is where the list of allowed
and blocked registries are stored for se by the builder pods and
all other pods/containers in the cluster.


umohnani8:
@runcom PTAL

Makefile
```diff
@@ -2,6 +2,7 @@ COMPONENTS = daemon controller server operator
 
 # vim: noexpandtab ts=8
 export GOPATH=$(shell echo $${GOPATH:-$$HOME/go})
+export GOFLAGS="-tags=containers_image_openpgp"
```
runcom: we may want this specifically in ./hack/build-go.sh (at the end, filtering for machine-config-controller)

runcom:
@mtrmac also please take an higher look at this 

@umohnani8 thanks for this!

templates/worker/01-worker-container-runtime/_base/files/policy.yaml
```diff
@@ -0,0 +1,19 @@
...
+  inline: |
+    {
+        "default": [
+            {
+                "type": "insecureAcceptAnything"
+            }
+        ],
+        "transports":
+            {
+                "docker-daemon":
```
mtrmac: `updatePolicyJSON` does not create this entry. Does it need to? (I don’t think so, unless there are users of Docker left around.) If not, it seems a bit unexpected to me that the template is non-trivially different from configs generated by `updatePolicyJSON`.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -264,6 +297,49 @@ func updateRegistriesConfig(data []byte, internalInsecure, internalBlocked []str
...
 
+// updatePolicyJSON decodes the data rendered from the template, merges the changes in and encodes it
+// back into a JSON format. It returns the bytes of the encoded data
+// It also returns an error if both allowed and blocked registries are set
+func updatePolicyJSON(data []byte, internalBlocked, internalAllowed []string) ([]byte, error) {
+	if len(internalAllowed) != 0 && len(internalBlocked) != 0 {
+		return nil, fmt.Errorf("invalid images config: only one of AllowedRegistries or BlockedRegistries may be specified")
+	}
+
+	policyObj := &signature.Policy{}
```
mtrmac: This is more of a question than a review comment:

`updateRegistriesConfig` decodes the original config, and only updates it (well… partially; it just overwrites some values regardless of whether the previous values were MCO-written or got there some other way). Does this need to do the same thing? Can `updateRegistriesConfig` be simplified to generate a fresh configuration instead of reading the old one? Or, maybe, are the two files somehow different to warrant different treatment?

(I rather hope that overwriting from scratch is OK; it’s not quite obvious how to update a pre-existing policy to implement `internalBlocked`/`internalAllowed`.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -156,6 +178,15 @@ func findRegistriesConfig(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
 	return nil, fmt.Errorf("could not find Registries Config")
 }
 
+func findPolicyJSON(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
```
mtrmac: I can’t see any caller; was this intended to be added to `generateOriginalContainerRuntimeConfigs`?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -613,13 +616,17 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 				if err != nil {
 					return fmt.Errorf("could not update registries config with new changes: %v", err)
 				}
+				policyJSON, err = updatePolicyJSON(dataURL.Data, blockedRegs, allowedRegs)
```
mtrmac: This is passing the `originalRegistriesIgn` (for `registries.conf`) data contents to `updatePolicyJSON` (for `policy.json`); luckily that does not break because `updatePolicyJSON` completely ignores the input — but that also means that the original template is completely ignored by `updatePolicyJSON` (and if I understand the render controller correctly, that means that the original template is completely ignored _if_ `insecureRegs != nil || blockedRegs != nil || allowedRegs != nil` and used unmodified otherwise).

pkg/controller/container-runtime-config/helpers.go
mtrmac: … reading more of the code, if I understand it correctly,
> `updateRegistriesConfig` decodes the original config
It does not decode the “current on-disk status” (with possible custom edits); it decodes the result of processing the built-in `templates/`, which would otherwise be used. `updateRegistriesConfig` then creates a new `MachineConfig` object, and the render controller ensures that the last version (by name) is applied, effectively overwriting the template output with the result of `updateRegistriesConfig` (template).

This means that `updateRegistriesConfig`, strictly speaking, needs to only handle updates to the template file, not to arbitrary files.

OK, that’s easier.

---

Per the other review comments, the `policy.json` generation, in that case, does not quite follow the existing pattern AFAICS.

mtrmac: Either way, the generation of `getManagedKeyReg`-named `MachineConfig` which includes `policy.json` will make it **rather difficult** for users to use custom `policy.json` configurations (e.g. to actually specify trusted keys and signature enforcement).  I guess there is always `999-*` and the like for custom configurations, at the cost of sacrificing the whitelist/blacklist functionality.  Is that good enough?

templates/worker/01-worker-container-runtime/_base/files/policy.yaml
umohnani8: This is the default template that s laid down by the MCO at start up. I looked at what a fresh cluster has for policy.json and just used that for the template her.
When we want to add something to policy.json, the MCO uses this template and adds to it.

pkg/controller/container-runtime-config/helpers.go
umohnani8: Yeah, `updateRegistriesConfig` just updates the registries.conf files based on the values sent in the CRD. That is what the MCO does for crio.conf and storage.conf as well.
So `updateRegistriesConfig` is just for reading in the registries.conf file and updating it. `updatePolicyJSON` is just for policy.json.
Do we want to create a new policy.json from scratch every time a change happens?

pkg/controller/container-runtime-config/helpers.go
umohnani8: woops, yeah fixed.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Fixed as well.

pkg/controller/container-runtime-config/helpers.go
mtrmac: > Do we want to create a new policy.json from scratch every time a change happens?
 
(I am new to this code.) AFAICS nodes always use the configuration in `MachineConfigPoolStatus.Configuration`, which is only generated by the render controller, so the files in `templates/*` are never used directly/unmodified; `container-runtime-config.Controller.syncImageConfig` is always involved. Therefore, it does not really matter much whether there is a template that is updated by `updatePolicyJSON`, or whether there is no template and this function generates a configuration from scratch.

Still, the code should behave consistently whether or not `blockedRegs != nil || allowedRegs != nil`; currently if this is false, the template is used with the `docker-daemon` entry, and if this is false, the template is ignored and there is no `docker-daemon` entry. That would be a clearly unintended side effect of the `blockedRegs`/`allowedRegs` configuration.

---

It seems simplest to me to drop the template and to always generate the policy using this function; that way all the decisions are clearly located in one place.

Using the template-created `data` object in this function would also work (whether or not the template removes the `docker-daemon` entry… but seems a bit more risky: the `policyObj.Default` `PolicyRequirements` value has fundamental impact to whether other, more specific, scopes, are configured correctly or not (e.g. whether there needs to exist a `reject` entry for a scope or whether the default would catch it) - so giving future maintainers a `policy.json` template that they can edit without necessarily noticing that the default policy can change is a bit more risky. (OTOH, editing the template JSON is a bit more readable than editing the Go code in this function; on balance, I don’t see this as a strong enough argument, but opinions can differ.)

umohnani8: @runcom WDYT? Is it okay if we just generate the policy.json without having a template from before every time there is an allowed or blocked registry entry in the images CR?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -414,11 +414,19 @@ func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.Cont
...
-		return nil, nil, nil, fmt.Errorf("could not generate old container runtime configs: %v, %v, %v", errStorage, errCRIO, errRegistries)
+	// Find Policy JSON
+	for _, gmc := range generatedConfigs {
+		config, errPolicy = findPolicyJSON(gmc)
+		if errPolicy == nil {
+			gmcPolicyJSON = config
+			break
+		}
+	}
+	if errStorage != nil || errCRIO != nil || errRegistries != nil || errPolicy != nil {
```
mtrmac: Non-blocking: This repetitive code doing 4 times the same two nested loops with different loops, with 4 sets of variables, 4 sets of errors, and 4 sets of return values (which are never used at the same time), is about at the point of being too ugly.

A single piece of code building a generic `Map<filename string, ignore *igntypes.File>`, and the callers just indexing into that with the necessary path, and handling failures separately, would be quite a bit shorter and somewhat more efficient.

Perhaps even make the function extract `contents []byte` for all of those files; that would do a fair amount of unnecessary data conversions to simplify the consumers further.

(And, going _way_ beyond the scope of this PR, it intuitively seems to me that that inefficiency could be best avoided by passing the required file names down into `GenerateMachineConfigsForRole` so that the unnecessary data is not generated in the first place, instead of only not converting it.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -58,17 +61,21 @@ type tomlConfigCRIO struct {
 
 type updateConfigFunc func(data []byte, internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error)
 
-func createNewCtrRuntimeConfigIgnition(storageTOMLConfig, crioTOMLConfig []byte) igntypes.Config {
+// createNewIgnition returns the data from the ignition files
+func createNewIgnition(configs map[string][]byte) igntypes.Config {
```
mtrmac: (Absolutely non-blocking, for a separate PR: This could be moved into a shared helper, maybe somewhere next to `NewIgnConfig`, and called also from `pkg/controller/kubelet-config`.)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -541,7 +549,11 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 				tempIgnCfg := ctrlcommon.NewIgnConfig()
 				mc = mtmpl.MachineConfigFromIgnConfig(role, managedKey, &tempIgnCfg)
 			}
-			mc.Spec.Config = createNewCtrRuntimeConfigIgnition(storageTOML, crioTOML)
+			tomlConfig := map[string][]byte{
```
mtrmac: (Absolutely non-blocking: inlining the map constructor into the call would avoid a single-use short-term variable. This works perfectly fine as well, feel free to keep it this way if it is easier to read for you. [Both callers of `createNewIgnition` should be consistent in this respect. See the comments in `createNewIgnition` first, though.])

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -431,29 +452,31 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 // getValidRegistries gets the insecure and blocked registries in the image spec and validates that the user is not adding
 // the registry being used by the payload to the list of blocked registries.
 // If the user is, we drop that registry and continue with syncing the registries.conf with the other registry options
-func getValidRegistries(clusterVersionStatus *apicfgv1.ClusterVersionStatus, imgSpec *apicfgv1.ImageSpec) ([]string, []string, error) {
+func getValidRegistries(clusterVersionStatus *apicfgv1.ClusterVersionStatus, imgSpec *apicfgv1.ImageSpec) ([]string, []string, []string, error) {
```
mtrmac: Absolutely non-blocking: I don’t really think it’s worth the extra code to return unmodified `insecureRegs` / `allowedRegs` from this function; the caller can just as well read them directly from `imgSpec` and avoid the return values / variables. OTOH then we would have to think about naming this function something else, and `blockedRegs` would be treated differently from the others…

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -724,7 +742,21 @@ func registriesConfigIgnition(templateDir string, controllerConfig *mcfgv1.Contr
...
+	if blockedRegs != nil || allowedRegs != nil {
+		dataURL, err := dataurl.DecodeString(originalPolicyIgn.Contents.Source)
+		if err != nil {
+			return nil, fmt.Errorf("could not decode original policy json: %v", err)
+		}
+		policyJSON, err = updatePolicyJSON(dataURL.Data, blockedRegs, allowedRegs)
+		if err != nil {
+			return nil, fmt.Errorf("could not update policy json with new changes: %v", err)
+		}
+	}
```
mtrmac: (It feels a bit off to have `blockedRegs` enforced in two configurations… OTOH it does not hurt anything.)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -707,14 +719,20 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 }
 
 func registriesConfigIgnition(templateDir string, controllerConfig *mcfgv1.ControllerConfig, role string,
```
mtrmac: (Non-blocking: The return value from this function no longer contains only the `registries.conf` data; maybe it should be renamed. (OTOH, arguably, the name refers to “registries-related configuration”.))

---

(It’s also not quite ideal that `RunImageBootstrap` invokes this with always-empty `*Regs` parameters, but splitting this into two functions and then somehow merging the ignition configuration in `syncImageConfig` would probably not be cleaner.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -392,6 +367,52 @@ func updateRegistriesConfig(data []byte, internalInsecure, internalBlocked []str
 	return newData.Bytes(), nil
 }
 
+// updatePolicyJSON decodes the data rendered from the template, merges the changes in and encodes it
+// back into a JSON format. It returns the bytes of the encoded data
+// It also returns an error if both allowed and blocked registries are set
+func updatePolicyJSON(data []byte, internalBlocked, internalAllowed []string) ([]byte, error) {
```
mtrmac: Please add something like
> WARNING: This can not safely edit policy files with arbitrary complexity, especially files which include `signedBy` requirements. It expects the input policy to be generated by templates in this project.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -331,31 +332,49 @@ func (f *fixture) expectUpdateContainerRuntimeConfig(config *mcfgv1.ContainerRun
...
 		icsps = append(icsps, icsp)
 	}
 	updatedMC, err := f.client.MachineconfigurationV1().MachineConfigs().Get(mcName, metav1.GetOptions{})
 	require.NoError(t, err)
-	verifyRegistriesConfigContents(t, updatedMC, mcName, imgcfg, icsps)
+	verifyRegistriesConfigAndPolicyJSONContents(t, updatedMC, mcName, imgcfg, icsps, numFiles)
 }
 
-func verifyRegistriesConfigContents(t *testing.T, mc *mcfgv1.MachineConfig, mcName string, imgcfg *apicfgv1.Image, icsps []*apioperatorsv1alpha1.ImageContentSourcePolicy) {
+func verifyRegistriesConfigAndPolicyJSONContents(t *testing.T, mc *mcfgv1.MachineConfig, mcName string, imgcfg *apicfgv1.Image, icsps []*apioperatorsv1alpha1.ImageContentSourcePolicy, numFiles int) {
```
mtrmac: Non-blocking I guess: The `numFiles` variable with magic 1/2 values is non-obvious; a `verifyPolicyJSON` bool would be cleaner. Or maybe split this into two/three functions, one to verify `registries.conf`, one to verify `policy.json`, one to verify the layout of `mc`, or something like that. OTOH, it’s ”only tests” in the end…

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -331,31 +332,49 @@ func (f *fixture) expectUpdateContainerRuntimeConfig(config *mcfgv1.ContainerRun
...
+		regfile = mc.Spec.Config.Storage.Files[1]
+	}
+	assert.Equal(t, registriesConfigPath, regfile.Node.Path)
+	registriesConf, err := dataurl.DecodeString(regfile.Contents.Source)
 	require.NoError(t, err)
 	assert.Equal(t, string(expectedRegistriesConf), string(registriesConf.Data))
+
+	// Validate the policy.json contents if a change is expected from the tests
+	if numFiles == 2 {
+		expectedPolicyJSON, err := updatePolicyJSON(templatePolicyJSON,
```
mtrmac: Please add actual tests for `updatePolicyJSON`, then — at least a smoke test for “no blocked/no allowed”, “something blocked”, “something allowed”.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -754,7 +781,7 @@ func TestRegistriesValidation(t *testing.T) {
 	for _, test := range successTests {
 		imgcfg := newImageConfig(test.name, test.config)
 		cvcfg := newClusterVersionConfig("version", "blah.io/myuser/myimage:test")
-		insecure, blocked, err := getValidRegistries(&cvcfg.Status, &imgcfg.Spec)
+		insecure, blocked, _, err := getValidRegistries(&cvcfg.Status, &imgcfg.Spec)
```
mtrmac: Non-blocking: Add actual tests for `allowedRegs` here? Or simplify by not returning that value, maybe, see elsewhere.

hack/build-go.sh
```diff
@@ -32,7 +32,13 @@ fi
 
 mkdir -p ${BIN_PATH}
 
+# Need to enable CGO and use the containers_image_opengpg flag because we are modifying the
+# /etc/containers/policy.json file whenever allowed/blocked registries are added
 CGO_ENABLED=0
```
mtrmac: But this is _disabling_ CGO :) Also `s/gpg/pgp/`.

Maybe something like
> Use the containers_image_openpgp` flag to avoid the default CGO implementation of signatures dragged in by containers/image/signature, which we use only to edit the /etc/containers/policy.json file without doing any cryptography

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -58,17 +61,21 @@ type tomlConfigCRIO struct {
 
 type updateConfigFunc func(data []byte, internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error)
 
-func createNewCtrRuntimeConfigIgnition(storageTOMLConfig, crioTOMLConfig []byte) igntypes.Config {
+// createNewIgnition returns an ignition config containing the new data configure by the CRs
```
mtrmac: This function does not see any CRs. It would be nice to describe what are the keys and values in the `configs` map (especially if the `nil` semantics will remain).

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -58,17 +61,21 @@ type tomlConfigCRIO struct {
...
+func createNewIgnition(configs map[string][]byte) igntypes.Config {
 	tempIgnConfig := ctrlcommon.NewIgnConfig()
 	mode := 0644
-	// Create storage.conf ignition
-	if storageTOMLConfig != nil {
-		storagedu := dataurl.New(storageTOMLConfig, "text/plain")
+	// Create ignitions
+	for filePath, data := range configs {
+		if data == nil {
```
mtrmac: (Original question: Why is `data` treated specially here? It seems to me that it should in principle be possible to use this function to create an empty file, and `nil` is (sadly) a mostly valid empty slice in Go.

If the caller does not want a file created, the caller can just not include the key.)

---

Ah, this is actually used to avoid updating some of the files with variables like `registriesTOML []byte` initialized to `nil` by default.

That works, but it’s pretty non-obvious; if we are introducing a new helper, I’d prefer making this explicit in the callers. Let the callers build the `configs` map, using either a literal that contains the always-used files, or conditionally adding individual files to the map if they should be set up.

Non-blocking: If you want to keep the current interface, and explicitly document the “data == `nil` means the file is not included” semantics, that would be fine as well.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -58,17 +61,21 @@ type tomlConfigCRIO struct {
...
+			continue
+		}
+		storagedu := dataurl.New(data, "text/plain")
```
mtrmac: This function is a nice generalization — make the name of this variable generic as well, please.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Will fix this up in a following PR

pkg/controller/container-runtime-config/container_runtime_config_controller.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: Fixed up to only validate for blocked registries

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: It is needed in registries.conf and policy.json, so going to leave it for now since it is not causing any problems.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Yeah, I don't want to mess around with this now since we need this in 4.2 and feature freeze is already over. Will try to do something for it in 4.3.

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: changed to `verifyPolicyJSON` bool instead

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: Added tests to helpers_test.go

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: Changed this function to only validate blocked registries, got rid of the insecure and allowed. Fixed

hack/build-go.sh
umohnani8: woops, I overlooked the 0. Fixed now

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: Documented why this is being done. Fixed

pkg/controller/container-runtime-config/helpers.go

umohnani8:
@mtrmac addressed all the comments above, PTAL
cc @runcom

runcom:
> @mtrmac addressed all the comments above, PTAL

/approve

will wait for a final ack from Miloslav, then lgtm

runcom:
> /test e2e-aws-scaleup-rhel7

we can /skip that test for now if it turns red again, there have been issues already reported.

umohnani8:
hold this, let's get https://github.com/openshift/machine-config-operator/pull/1087 in first. I will rebase on top of that then.

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -354,3 +356,90 @@ func TestUpdateRegistriesConfig(t *testing.T) {
...
+				t.Errorf("updatePolicyJSON() error = %v", err)
+				return
+			}
+			gotConf := signature.Policy{}
+			if err := json.Unmarshal(got, &gotConf); err != nil {
+				t.Errorf("error unmarshalling result: %v", err)
+				return
+			}
+			// This assumes a specific order of Registries entries, which does not actually matter; ideally, this would
+			// sort the two arrays before comparing, but right now hard-coding the order works well enough.
```
mtrmac: This comment is not applicable to `signature.Policy`.

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -354,3 +356,90 @@ func TestUpdateRegistriesConfig(t *testing.T) {
...
+			if !reflect.DeepEqual(gotConf, tt.want) {
+				t.Errorf("updatePolicyJSON() Diff:\n %s", diff.ObjectGoPrintDiff(tt.want, gotConf))
+			}
+			// Ensure that the generated configuration is actually valid.
```
mtrmac: This did not do that AFAICS?! `c/image/signature.NewPolicyFromBytes` can be used without the temporary file.

pkg/controller/container-runtime-config/helpers.go
mtrmac: ACK. Could you move the `nil` documentation to the function doc string? It’s a property of the interface, not an internal implementation detail.

runcom:
> hold this, let's get #1087 in first. I will rebase on top of that then.

/hold

pkg/controller/container-runtime-config/helpers_test.go
umohnani8: Fixed

pkg/controller/container-runtime-config/helpers_test.go
umohnani8: Fixed

umohnani8:
Rebased, we can cancel the hold now and this is ready once tests are green.
cc @runcom @mtrmac @kikisdeliveryservice

kikisdeliveryservice:
I'm aware of some issues with the upgrade tests, cant' tell if the AWS failure and this specific upgrade test is related. Will retest tomorrow.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -378,17 +378,17 @@ func (ctrl *Controller) handleImgErr(err error, key interface{}) {
...
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("generateMachineConfigsforRole failed with error %s", err)
+		return nil, nil, nil, nil, fmt.Errorf("generateMachineConfigsforRole failed with error %s", err)
 	}
 	// Find generated storage.config, and crio.config
 	var (
-		config, gmcStorageConfig, gmcCRIOConfig, gmcRegistriesConfig *igntypes.File
-		errStorage, errCRIO, errRegistries                           error
+		config, gmcStorageConfig, gmcCRIOConfig, gmcRegistriesConfig, gmcPolicyJSON *igntypes.File
+		errStorage, errCRIO, errRegistries, errPolicy                               error
```
runcom: this is growing out of control :smile: probably needs a struct to wrap all these configs

runcom:
non-blocking comment, lgtm overall :)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Will fix in another PR

umohnani8:
Tests are finally green! :)

PR #804, opened 2019-05-30
DR: use param to populate etcd name

hexfusion:
The PR resolves a few issues.

* `ETCD_NAME` is currently populated from `hostname` but this is not always valid. For example, `hostname` in AWS returns the hostname of the node. But in baremetal hostname returns FQDN. Because we can no longer assume how etcd name is formatted we will ask for it as a mandatory param for `etcd-member-recover.sh`. With the addition of `validate_etcd_name function,` we can now extract the ETCD_NAME from `ETCD_INITIAL_CLUSTER` for `etcd-snapshot-restore.sh`.

* ETCD_CONNSTRING is now renamed ETCD_INTIAL_CLUSTER to more accurately reflect value.

* Fail early if SNAPSHOT_FILE does not actually exist.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1714457

/hold

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -9,7 +9,7 @@ contents:
     set -o pipefail
 
     # example
-    # etcd-snapshot-restore.sh $path-to-snapshot $etcd-connection-string
+    # ./etcd-snapshot-restore.sh $path-to-snapshot $inital_cluster
 
     if [[ $EUID -ne 0 ]]; then
       echo "This script must be run as root"
@@ -17,20 +17,16 @@ contents:
     fi
 
     usage () {
-        echo 'Path to snapshot is required: ./etcd-member-recover.sh $path-to-snapshot'
-        exit
+        echo 'Path to snapshot and initial cluster are required: ./etcd-snapshot-restore.sh $path-to-snapshot $initial_cluster'
+        exit 1
     }
 
-    if [ "$1" == "" ]; then
+    if [ "$1" == "" ] || [ "$2" == "" ]; then
         usage
     fi
 
     SNAPSHOT_FILE="$1"
-    ETCD_CONNSTRING=""
-    if [ "$2" != "" ]; then
-      ETCD_CONNSTRING="$2"
-    fi
-
+    ETCD_INITIAL_CLUSTER="$2"
     ASSET_DIR=./assets
     CONFIG_FILE_DIR=/etc/kubernetes
     MANIFEST_DIR="${CONFIG_FILE_DIR}/manifests"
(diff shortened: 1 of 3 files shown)


hexfusion:
@vrutkovs Looks like we fail to get bastion installed correctly?
...Still waiting for DNS...
nslookup: '' is not a legal name (unexpected end of input)
[....]
nslookup: '' is not a legal name (unexpected end of input)

vrutkovs:
>@vrutkovs Looks like we fail to get bastion installed correctly?

Yep, that's Route53 not replying back after 30 attempts :/

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
```diff
@@ -9,26 +9,27 @@ contents:
     set -o pipefail
 
     # example
-    # etcd-snapshot-restore.sh $path-to-snapshot $etcd-connection-string
+    # ./etcd-snapshot-restore.sh $path-to-snapshot $inital_cluster
```
sgreene570: spelling?

sgreene570: might explain CI failures

hexfusion: The param var is misspelled but how would it cause CI failure?

sgreene570: oh whoops didn't realize this was commented out my mistake

hexfusion:
@vrutkovs adding exit 1 to usage() gave us what we were expecting.

```
Path to snapshot and initial cluster are required: ./etcd-snapshot-restore.sh $path-to-snapshot $initial_cluster
exit code: 1
```

vrutkovs:
This is fixed in https://github.com/openshift/release/pull/3919

vrutkovs:
I think I started `etcd-quorum-loss` too soon - it still fails with this error.

/test e2e-etcd-quorum-loss

hexfusion:
> + echo 'Snapshot file not found, restore failed: /root/assets/backup/etcd/member/snap/db.'
+ exit 1

/cc @vrutkovs 

/test e2e-etcd-quorum-loss

vrutkovs:
Fix: https://github.com/openshift/release/pull/3967

hexfusion:
I think the issue was actually on me I wanted to check that the snapshot existed early. But if we want to use the backup the check was too soon as it did not exist yet.

hexfusion:
all tests passed!

vrutkovs:
```
Failing tests:

```
Known flake (although I assumed the fix has been merged by that time).
/skip

hexfusion:
>   "Pod openshift-sdn/sdn-kd54v is not healthy: container sdn has restarted more than 5 times",

The tests are still flakey due to other issues than this code directly. The changes here are needed for DR to work properly for bare-metal so we need to merge this and then solve other problems.

hexfusion:
@kikisdeliveryservice @cgwalters could I get an approval, please.

PR #805, opened 2019-05-30
Mirrors support

mtrmac:
**- What I did**
Implemented support for `operator/v1alpha1.ImageContentSourcePolicy`, added by https://github.com/openshift/api/pull/354 , to support multiple mirrors for pulling OpenShift images.

NOTE: The CRI interface does not support passing more than one set of credentials for pulling an image, i.e. the pull secret, if any, only applies to the primary image location (the one explicitly included in the pull spec). If any of the mirrors need credentials, the credentials must be available on the node, probably via `ControllerConfig.spec.pullSecret`. That implies that configuring per-tenant credentials for accessing the mirrors is not currently possible.

To make this work at least for the OpenShift images, this PR also hard-codes CRI-O’s `global_auth_file` option to point to the pull secret file also used by kubelet, which provides access to OpenShift mirrors.

**- How to verify it**
WIP.  There are fairly extensive unit tests, but I didn’t try this end-to end yet at all.

(Depending on which variant is chosen):
- Added support for `operator/v1alpha1.ImageContentSourcePolicy`, allowing to configure mirrors for image registries


cgwalters:
This is totally fine but it'd also be nice to update the PR description with at least a sentence or two for what the end goal is - think of it like the "seed" upon which an eventual proper commit message could grow :smile: - offhand I'm guessing something like:

> We're working on supporting fetching container images from mirrors.

mtrmac:
> This is totally fine but it'd also be nice to update the PR description with at least a sentence or two for what the end goal is - think of it like the "seed" upon which an eventual proper commit message could grow 😄

Updated now.  Still WIP, notably completely untested.

I _really_ don’t know much about what I’m doing, any pointers would be appreciated.

Specific questions, for now:
- What should the eventual API look like: this adds **two separate** commits adding members to `vendor/github.com/openshift/api/config/v1/types_image.go`, we need one of them only.
- I suppose those members will have to be added to the `openshift/api` repository. Or should this instead introduce a new object / some other kind of change in the `machineconfiguration.openshift.io` namespace?
- The mirror support requires using (a current version of) the v2 `registries.conf` format, which implies requiring a new enough version of CRI-O (and Podman?) (Neither of which exist at this moment.). Can we just assume that the version will be new enough (and enforce that in some repository by updating MCO and CRI-O/… in lockstep), or does this code need to detect and transparently support older versions of the two tools as well (degrading to no-mirror operation)?

cgwalters:
This work is targeted at 4.2 right?

Can you link to any more information on the dependent crio work?  

It looks like this depends on https://github.com/openshift/api/pull/354 - may be good to do a separate PR to update our vendored openshift/api as a prep step.

mtrmac:
> This work is targeted at 4.2 right?

Yes.

> Can you link to any more information on the dependent crio work?
https://github.com/cri-o/cri-o/pull/2494 + https://github.com/cri-o/cri-o/pull/2510 for 1.14,
https://github.com/cri-o/cri-o/pull/2439 + https://github.com/cri-o/cri-o/pull/2511 on master.

> It looks like this depends on [openshift/api#354](https://github.com/openshift/api/pull/354) - may be good to do a separate PR to update our vendored openshift/api as a prep step.

Will do, at least to get the other effects of `dep ensure -update` out of this PR — but note that the API repo changes are pure additions, and only really happen when this PR adds the uses of those subpackages.

mtrmac:
> It looks like this depends on [openshift/api#354](https://github.com/openshift/api/pull/354) - may be good to do a separate PR to update our vendored openshift/api as a prep step.

https://github.com/openshift/machine-config-operator/pull/936 . This PR now includes code from that branch, so don’t merge it as it is; after https://github.com/openshift/machine-config-operator/pull/936 is merged, I’ll rebase on top of master again.

umohnani8:
@mtrmac you will need to add something like https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfigcontroller/clusterrole.yaml#L16 for `operator.openshift.io` to get permissions to read the ImageContentSourcePolicy CRs.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -586,6 +613,12 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		return err
 	}
 
+	// Find all ImageContentSourcePolicy objects
+	icspRules, err := ctrl.icspLister.List(labels.Everything())
+	if err != nil {
```
runcom: if this has been added in 4.2.master and 4.1 doesn't have it, we need to tolerate a notfound error or 4.1->4.2 cluster upgrade fails

runcom:
@mtrmac any reason this is still WIP? looks generally good, would love to find/have a way to test this out on a cluster also, but probably QE can verify that after landing as well (same as what happened with Proxies)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
mtrmac: AFAICS `icspLister.List` can never fail just because an object is missing (though maybe something creating the `icspLister` might fail, I haven’t read all of the code); only `icspLister.Get` can return a `StatusReasonNotFound` error.

Anyway, added a check to make the behavior explicit.

mtrmac:
> @mtrmac any reason this is still WIP? looks generally good, would love to find/have a way to test this out on a cluster also

It’s only now being tested on a live cluster (and reportedly failing), and it is missing tests for the OpenShift API integration. If you want to merge it anyway, I won’t complain too loud ;)

runcom:
> > 
> 
> It’s only now being tested on a live cluster (and reportedly failing), and it is missing tests for the OpenShift API integration. If you want to merge it anyway, I won’t complain too loud ;)

ha, that's why I asked - if t can be of any help with testing this ou, let me know

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -244,22 +241,176 @@ func updateCRIOConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfiguratio
...
-	if _, err := toml.Decode(string(data), tomlConf); err != nil {
+// prefixMatchesRegistry returns true if a prefix value (as in sysregistriesv2.Registry.Prefix / sysregistriesv2.Endpoint.Location)
+// matches a host[:port] value in reg.
+func prefixMatchesRegistry(prefix, reg string) bool {
+	// NOTE: "prefix" is the sysregistries.Registry name of the object, and it can be a host[:port] or a namespace within;
+	// i.e. we want to return true if "reg" is a prefix of "prefix"!
+	if reg == prefix {
+		return true
+	}
+	if len(prefix) > len(reg) {
```
kikisdeliveryservice: sorry but could you explain what this comparision is doing? ive read it a few times and dont fully get it?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -244,22 +241,176 @@ func updateCRIOConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfiguratio
...
+// disjointOrderedMirrorSets processes icspRules and returns the resulting disjoint sets of mutual mirrors, each ordered in consistently with
+// the icspRules preference order (if possible)
+// This exists to merge different mirror sets, e.g. given mirror sets (B, C) and (A, B), it will combine them into a single (A, B, C) set.
+func disjointOrderedMirrorSets(icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy) ([][]string, error) {
+	// Collect all mirror sets into a flat array to abstract the rest of the code from the ImageContentSourcePolicy format
+	mirrorSets := []*apioperatorsv1alpha1.RepositoryDigestMirrors{}
+	for _, icsp := range icspRules {
+		for i := range icsp.Spec.RepositoryDigestMirrors {
+			set := &icsp.Spec.RepositoryDigestMirrors[i]
+			if len(set.Sources) < 2 {
```
kikisdeliveryservice: why not if len(set.Sources) >= 2 { mirrorSets....}

pkg/controller/container-runtime-config/union.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return res
+}
+
+// find is Find(), except it returns a *internalUnionNode
+func (g *unionGraph) find(query unionNode) *internalUnionNode {
+	queryNode := g.getNode(query)
+	// Find root
+	x := queryNode
+	for x.parent != nil {
```
kikisdeliveryservice: ahh took me a minute to realize this is golang's while loop. :) nice

pkg/controller/container-runtime-config/helpers.go
mtrmac: `prefix` is a scope for the configuration, which can be a host[:port] or host[:post]/namespace[/namespace…], or a host[:port]/namespaces…/image (or, ultimately, include a tag and/or digest, but we don’t generate such configs in this controller).

`reg` is a host[:port] value, and the purpose of this function is to match all scopes (i.e. `prefix` values) that use the host[:port] value of `reg`. 
e.g.

`reg` | `prefix` | result
--- | ------ | ------
`example.com` | `quay.io` | false
`quay.io` | `quay.io` | true
`quay.io:443` | `quay.io` | false
`quay.io` | `quay.io/namespace` | true
`quay.io` | `quay.io/rhel8/rhel` | true

Yes, the naming is confusing, as the comment says, I’d appreciate any suggestion.

pkg/controller/container-runtime-config/helpers.go
mtrmac: This way the comment about the `len < 2` case can be placed in the matching code path; using `len >= 2` would force the comment to be above the `if`, and redundantly describe the condition.

I don’t care at all strongly about this.

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: if we check the reg==prefix and return true, is there anyway to just fold that into the strings.HasPrefix logic? Because HasPrefix returns true if reg==prefix as well

https://play.golang.org/p/ZJFrQhNErbt

Just wondering if we need that '/' check? And trying to wrap my brain around this PR ;)

kikisdeliveryservice: actually disregard as im seeing the ':' is a possibility. thanks for the chart.

kikisdeliveryservice: @mtrmac ok im a little confused about line 4 of your chart bc that seems like it would yield false? in your example reg (quay.io) will return False for HasPrefix quay.io/namespace that and since it's an && it will return false there...
https://play.golang.org/p/zZWdgopC4WI

kikisdeliveryservice: wait i see it. HasPrefix(prefix, reg) not (reg, prefix)... Going to try to think of a new name for prefix here bc i think it's confsing me...

kikisdeliveryservice: if prefix = `quay.io` and reg = `quay.io/somethingelse` this will return false but that's a textbook prefix. 

we are really checking if reg is in (as a prefix lol) prefix so maybe rename prefix to sysregistriesRegistryName or sysregistriesName and eliminate the prefix wording altogether??

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: i think it would be better to open a github issue to cover this FIXME since they tend to get lost in the shuffle.

pkg/controller/container-runtime-config/helpers.go
mtrmac: > actually disregard as im seeing the ':' is a possibility. thanks for the chart.

Not just that, it’s also making sure that `quay.io.example.com` and `quay.ioprefix` don’t match `quay.io`.

mtrmac: > we are really checking if reg is in (as a prefix lol) prefix so maybe rename prefix to `sysregistriesRegistryName` or `sysregistriesName` (taken from leading comments before func) or `configScope` and eliminate the prefix wording altogether??

Thanks, `scope` does express this much better. (I’ll also add unit tests to record some of the aspects of the behavior.)

pkg/controller/container-runtime-config/helpers.go
mtrmac: Yeah, filed https://github.com/openshift/machine-config-operator/issues/949 .

mtrmac:
Dropping the WIP marker, reportedly this works, and the new behavior now has tests at least as good as any other aspect of the controller (in particular, the tests now actually verify expected contents of the generated `registries.conf`).

Still, I don’t yet fully understand the test scaffolding, so, please, make sure to review that the “Fix TestContainerRuntimeConfigUpdate and TestImageConfigUpdate” commit makes sense.

umohnani8:
Tested this out on a cluster and everything works fine. Tried out adding multiple ImageContentSourcePolicy objects with overlaps in the mirrors defined. Tested deleting the CRs one by one and registries.conf was populates as expected. And it plays well with any insecure and blocked registries added to the cluster wide Image CR.
The only thing I think that is potentially missing is checking whether there was an actual change to all the ImageContentSourcePolicy CRs when a resync happens. This way we can prevent rebooting the nodes when a resync happens every few minutes if nothing with the registries.conf file has changed. We do a check for the Image CR, which was added here https://github.com/openshift/machine-config-operator/pull/461. cc @mtrmac

mtrmac:
> The only thing I think that is potentially missing is checking whether there was an actual change to all the ImageContentSourcePolicy CRs when a resync happens. This way we can prevent rebooting the nodes when a resync happens every few minutes if nothing with the registries.conf file has changed. We do a check for the Image CR, which was added here #461. cc @mtrmac

I thought exactly that path, in particular the https://github.com/openshift/machine-config-operator/blob/aed3d2a170dbad0bc9ebb723a51d989380d74e1c/pkg/controller/container-runtime-config/container_runtime_config_controller.go#L627 check, should already work prevent no-op changes. Is that not enough?

OTOH, I can’t get `TestICSPUpdate` to stop the MCS object updates just by not modifying ICSP, but so far that seems to be more likely to be a test issue (the difference is actually in `OSImageURL` being `""` vs. `"dummy:///"`).

umohnani8:
@mtrmac ah yes, that should do it. I missed it, my bad.
Everything LGTM then, this should be ready to merge once tests pass @runcom @kikisdeliveryservice :)

umohnani8:
@runcom @cgwalters tests are green!

kikisdeliveryservice:
this LGTM

will let walters/runcom give the final approval

cgwalters:
> Because any users without credentials were already able to use the
kubelet's config.json secrets

Hmm...not as of https://github.com/openshift/machine-config-operator/pull/827 right?

mtrmac:
> > Because any users without credentials were already able to use the
> > kubelet's config.json secrets
> 
> Hmm...not as of #827 right?

The above is not talking about non-container users with local accounts on the node, but about users of the cluster with the ability to create pods (who should, typically, have no access outside of their containers at all). For such users, the kubelet automatically uses the  kubelet’s secret file if the user does not explicitly provide a pull secret (and the kubelet is, before this PR, the only process on the node that actually needs access to the cluster pull secret file), and sends the credentials to CRI-O to cause the image to be pulled.

cgwalters:
So...this is a lot to take in.  I'd like to see a higher level design doc if one exists.  But the feature is critical, and I think we can learn more after it lands in master.
/lgtm

kikisdeliveryservice:
> So...this is a lot to take in. I'd like to see a higher level design doc if one exists. But the feature is critical, and I think we can learn more after it lands in master.

+1

PR #806, opened 2019-05-30
pkg/daemon: backup the initial node annotations file

runcom:

Trying to understand something about https://bugzilla.redhat.com/show_bug.cgi?id=1702626

Let's check if that file is at all around. This isn't really changing any logic but the rename vs delete of the initial node annotation file.


runcom:
@hexfusion @vrutkovs is the etcd-quorum-loss e2e expected to pass? I can't see it in other PRs

vrutkovs:
Prometheus alerts test may fail - its not clear what's causing it

runcom:
> Prometheus alerts test may fail - its not clear what's causing it

why is it running only on some PR though, can we skip it here, let me repush

runcom:
closing as it's in #807

vrutkovs:
>why is it running only on some PR though

we've enabled it to run on all PRs yesterday, so it won't start on older PRs.

Use `/skip` to make it go away if it still flakes

Issue #660, opened 2019-04-24
oc edit node -> remove currentConfig annotation never reconciles


runcom:
As title says, if we drop the currentConfig annotation from a node, we're not going back. One thing we might want to do is (non trivial I guess) lookup what we've stored on disk in the MCD and re-load that.

runcom:
now that ondisk state is there and respected/relied on for DR, I believe it's gonna be something we might want to fix instead of having the node crashlooping or manual intervention. Basically, if we, for whatever reason fail to grab the currentConfig in MCD, we're going to rely on what we have on disk and proceed w/o bailing.

kikisdeliveryservice:
Addressed

PR #807, opened 2019-05-30
Bug 1702626: pkg/daemon: reconcile if we lost annotations

runcom:
~Builds on #806 for no good reason actually.~ brought it in here

Should take care of this one as well https://github.com/openshift/machine-config-operator/issues/660 but requires an MCD restart (we might want to drop `dn.booting` to have it automatically reconciles)

This is tricky, but the aim is to workaround missing annotations on node (best effort) - I haven't fully tested this just yet but reviews are highly appreciated.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -839,11 +839,11 @@ func (dn *Daemon) CheckStateOnBoot() error {
 		}
 		glog.Info("No bootstrap pivot required; unlinking bootstrap node annotations")
 
-		// Delete the bootstrap node annotations; the
+		// Rename the bootstrap node annotations; the
 		// currentConfig's osImageURL should now be *truth*.
 		// In other words if it drifts somehow, we go degraded.
-		if err := os.Remove(constants.InitialNodeAnnotationsFilePath); err != nil {
-			return errors.Wrapf(err, "removing initial node annotations file")
+		if err := os.Rename(constants.InitialNodeAnnotationsFilePath, constants.InitialNodeAnnotationsBakPath); err != nil {
+			return errors.Wrap(err, "renaming initial node annotation file")
 		}
 	}
 
--- a/pkg/daemon/node.go
+++ b/pkg/daemon/node.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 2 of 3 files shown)


runcom:
Looking at the node controller, this PR makes some sense. In the node controller, we filter out nodes w/o currentConfig annotation and we'll thus never add/change desired. With this change, we're able to _publish_ our currentConfig from on-disk if something went wrong (like in the BZ I'm investigating). Setting currentConfig from on-disk also lets the node controller run again with the node in question and add a desired config as well (reconciling etc etc).

runcom:
ok so the issue in the BZ turned out not to be us, so I don't really want this PR anymore but let's check with everyone :)

runcom:
going to properly test this out actually as we might still be in charge of reconciling such scenario

pkg/daemon/node.go
```diff
@@ -23,9 +24,18 @@ func (dn *Daemon) loadNodeAnnotations(node *core_v1.Node) (*core_v1.Node, error)
...
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
 		return nil, fmt.Errorf("failed to read initial annotations from %q: %v", constants.InitialNodeAnnotationsFilePath, err)
 	}
+	if os.IsNotExist(err) {
+		// try currentConfig if, for whatever reason we lost annotations? this is super best effort.
+		currentOnDisk, err := dn.getCurrentConfigOnDisk()
+		if err == nil {
+			glog.Infof("Setting initial node config based on current configuration on disk: %s", currentOnDisk.GetName())
+			return setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, node.Name, map[string]string{constants.CurrentMachineConfigAnnotationKey: currentOnDisk.GetName()})
```
runcom: we only set current cause desired is coming after the node controller kicks in

runcom:
ok, manually tested, we might want QE to provide a test for this as well. @cgwalters ptal

pkg/daemon/daemon.go
```diff
@@ -660,7 +660,7 @@ func (dn *Daemon) getStateAndConfigs(pendingConfigName string) (*stateAndConfigs
 		return nil, err
 	}
 	desiredConfigName, err := getNodeAnnotation(dn.node, constants.DesiredMachineConfigAnnotationKey)
-	if err != nil {
+	if err != nil && pendingConfigName == "" {
```
cgwalters: I don't follow the logic here; we get an error finding `desiredConfigName` but then we're conditionalizing on the `pendingConfigName`?

pkg/daemon/daemon.go
```diff
@@ -839,11 +839,11 @@ func (dn *Daemon) CheckStateOnBoot() error {
 		}
 		glog.Info("No bootstrap pivot required; unlinking bootstrap node annotations")
 
-		// Delete the bootstrap node annotations; the
+		// Rename the bootstrap node annotations; the
 		// currentConfig's osImageURL should now be *truth*.
 		// In other words if it drifts somehow, we go degraded.
-		if err := os.Remove(constants.InitialNodeAnnotationsFilePath); err != nil {
-			return errors.Wrapf(err, "removing initial node annotations file")
+		if err := os.Rename(constants.InitialNodeAnnotationsFilePath, constants.InitialNodeAnnotationsBakPath); err != nil {
```
cgwalters: AFAICS no code *reads* the backup file - right?  So how is this helping?

runcom: This was just for debugging purposes to hash out doubts like "did we ever had an initial node annotations file?" - should be harmless

pkg/daemon/daemon.go
runcom: I need to drop this. Not sure why I've added it at this point.

runcom:
The thing with this PR is only the last commit also. The first one was just for debugging. The second commit is the one adding reconciliation logic when current isn't there coming from a restored node.

runcom:
To test this out on a cluster:

- oc delete NODE
- ssh NODE && systemctl restart kubelet
- NODE registered again but no annotations

With this patch we're able to reconcile otherwise we get into the error in the Bugzilla.

cgwalters:
LGTM with the first bit dropped.

On the first patch, I found it confusing to make a constant for the `.bak` file - made me think something else would read it.  But it's fine as is.

runcom:
> LGTM with the first bit dropped.

alrighty, pushed again now

/hold cancel

runcom:
wat

/retest

runcom:
@vrutkovs known etcd loss failure?

vrutkovs:
```
Network timed out while trying to connect to https://registry.svc.ci.openshift.org/v1/repositories/openshift/release/images. You may want to check your internet connection or if you are behind a proxy.
```
Registry is having a bad day :/
/retest

PR #810, opened 2019-06-03
Update the e2e tests to use .yaml

skitt:
This matches the change in
https://github.com/openshift/installer/pull/1748 and would allow us to
switch all YAML extensions to .yaml.

Related to https://github.com/openshift/installer/issues/1501

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -73,8 +73,8 @@ func init() {
 	bootstrapCmd.MarkFlagRequired("infra-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.configFile, "config-file", "", "ClusterConfig ConfigMap file.")
 	bootstrapCmd.MarkFlagRequired("config-file")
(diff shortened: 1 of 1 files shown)


cgwalters:
This code isn't user facing, not sure why this matters?  But not opposed either.

ashcrow:
I agree with @cgwalters and @runcom.

/retest

cgwalters:
OK I see what you're trying to do now.  This can't work as is because the PRs depend on each other.  I think what you need to do is:

 - Change the installer to *explicitly pass* those arguments

And now you could stop there, but if you want to make it slightly nicer:

 - Update this PR to make the arguments required rather than having defaults

skitt:
> OK I see what you're trying to do now. This can't work as is because the PRs depend on each other. I think what you need to do is:
> 
>     * Change the installer to _explicitly pass_ those arguments
> 
> 
> And now you could stop there, but if you want to make it slightly nicer:
> 
>     * Update this PR to make the arguments required rather than having defaults

Fantastic, thanks; I was trying to figure out how to deal with the _n_-step dance involved, and hadn’t thought of parameterising the file names...

kikisdeliveryservice:

Feel free to reopen.

PR #813, opened 2019-06-04
Bug 1703234: pkg/operator: ignore CM changes that we don't care about

runcom:
Triggering a sync on the following CMs chnages causes a wasteful sync (every 14s):
```
NAME                             DATA   AGE
machine-config                   0      27m
machine-config-controller        0      27m
```
Those cms seem to be bound to operator and controller leases. Since we
don't need to listen on them for our sync, just ignore them.

I'm following up with someone who understands more than me on this but it shouldn't be an issue to just ignore those.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -28,6 +28,7 @@ import (
 	appslisterv1 "k8s.io/client-go/listers/apps/v1"
 	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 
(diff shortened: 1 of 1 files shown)


runcom:
/hold

this is quite spammy actually - holding till I at least figure stuff in my test cluster - we can drop this to glog v2 maybe.

runcom:
ok, somehow we have a ConfigMap resyncinc every 14s for no good reason - trying to understand why now.

runcom:
```

runcom:
/hold

We need to wholly ignore Config maps based on the leader annotation (got word from Clayton). I'll change this later today.

runcom:
/hold cancel

fully good to go now!

runcom:
leaning towards skipping that test..........

kikisdeliveryservice:
> leaning towards skipping that test..........

the etcd quorum loss one?  looking at the runs it hasn't had a successful one on this pr at all and only 2 successful runs since Monday... yikes

runcom:
> the etcd quorum loss one? looking at the runs it hasn't had a successful one on this pr at all and only 2 successful runs since Monday... yikes

yeah, it's known to be highly flake :(

PR #814, opened 2019-06-04
pkg/operator: Remove infra platform string constants

sgreene570:
**- What I did**

The openshift v1 infrastructure API types have been updated (https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go).  As is, `render.go` does not take into account bare metal or GCP cluster type strings. The currently used struct fields are also marked as deprecated in the API source code.  

**- How to verify it**

Should not be semantically different.

Update platform type strings in controller config spec

### DIFF
--- a/pkg/operator/render.go
+++ b/pkg/operator/render.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"net"
+	"strings"
 	"text/template"
 
 	"github.com/Masterminds/sprig"
@@ -67,23 +68,19 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 		return nil, err
 	}
 
+	infraPlatformString := ""
+	// The PlatformStatus field is set in cluster versions >= 4.2
+	// Otherwise use the Platform field
+	if infra.Status.PlatformStatus != nil {
+		infraPlatformString = string(infra.Status.PlatformStatus.Type)
+	} else {
+		//nolint:staticcheck
+		infraPlatformString = string(infra.Status.Platform)
+	}
+
 	platform := "none"
-	//nolint:staticcheck
-	switch infra.Status.Platform {
-	case configv1.AWSPlatformType:
-		platform = "aws"
-	case configv1.AzurePlatformType:
-		platform = "azure"
-	case configv1.BareMetalPlatformType:
-		platform = "baremetal"
-	case configv1.GCPPlatformType:
-		platform = "gcp"
-	case configv1.OpenStackPlatformType:
-		platform = "openstack"
-	case configv1.LibvirtPlatformType:
-		platform = "libvirt"
-	case configv1.VSpherePlatformType:
-		platform = "vsphere"
+	if infraPlatformString != "" {
+		platform = strings.ToLower(infraPlatformString)
 	}
 
 	ccSpec := &mcfgv1.ControllerConfigSpec{
--- a/pkg/operator/render_test.go
+++ b/pkg/operator/render_test.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"fmt"
+	configv1 "github.com/openshift/api/config/v1"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


pkg/operator/render.go
```diff
@@ -67,19 +69,7 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
...
-	case configv1.AzurePlatformType:
-		platform = "azure"
-	case configv1.OpenStackPlatformType:
-		platform = "openstack"
-	case configv1.LibvirtPlatformType:
-		platform = "libvirt"
-	case configv1.VSpherePlatformType:
-		platform = "vsphere"
-	}
+	platform := strings.ToLower(string(infra.Status.PlatformStatus.Type))
```
runcom: the only semantic difference is that we don't fallback to a `none` platform right? cc @abhinavdahiya for confirmation

runcom:
Looks great as is, just a friendly ping to installer :+1:

pkg/operator/render.go
sgreene570: actually looks like the none platform is defined? @runcom not sure if that's used by installer though.
https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go

runcom: oh great, nevermind this comment then

pkg/operator/render.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net"
 	"text/template"
+	"strings"
+
```
abhinavdahiya: extra line

sgreene570: nice catch

runcom:
gosh, fat fingers (wanted to comment)

/approve
/lgtm

runcom:
uhm, something isn't properly working here 

/hold

pkg/operator/render.go
kikisdeliveryservice: question: is there any way the platform won't be set to anything (not None, but empty)?  maybe we should check to make sure it has a value?

runcom:
> Is the intention of this PR to add support for additional platforms? For example, the previous state put baremetal, GCP, libvirt to None as opposed to passing them as actual platform values. I'm pretty sure this is why the switch statement was originally used and that we should make sure we actually want to expand the potential platform values that we are passing before we merge this.
> 
> See: https://github.com/openshift/api/blob/43c075c38fd931d1c85b41039e79caac63fc1925/config/v1/types_infrastructure.go#L70
> 
> @runcom

4.2 supports that set of platforms so we should be good landing this to master (cross checked with Ian as well)

kikisdeliveryservice:
> 4.2 supports that set of platforms so we should be good landing this to master (cross checked with Ian as well)

is there any issue merging here before they are fully supported?

runcom:
> s there any issue merging here before they are fully supported?

nope afaict, the installer sets the platform and if someone is using master is not something _released_ so free to break anytime w/o support.

kikisdeliveryservice:
> nope afaict, the installer sets the platform and if someone is using master is not something _released_ so free to break anytime w/o support.

Fair enough! :smile: 

Just 1 comment from me above, as I'd like a check that platform has a value otherwise set to None: https://github.com/openshift/machine-config-operator/pull/814#discussion_r290830598

runcom:
/lgtm cancel

keep holding this PR because the PlatformStatus is still empty because we need https://github.com/openshift/installer/pull/1725#issuecomment-492934400. We can go ahead with this PR once the installer PR lands

sgreene570:
Added empty string check so this will be ready once the installer PR is merged.

pkg/operator/render.go
```diff
@@ -68,17 +69,9 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
...
-		platform = "azure"
-	case configv1.OpenStackPlatformType:
-		platform = "openstack"
-	case configv1.LibvirtPlatformType:
-		platform = "libvirt"
-	case configv1.VSpherePlatformType:
-		platform = "vsphere"
+	infraPlatformString := string(infra.Status.PlatformStatus.Type)
+	if infraPlatformString != "" {
+		platform = strings.ToLower(infraPlatformString)
```
kikisdeliveryservice: Nice update :smile:

sgreene570:
Added a unit test for the function I modified :smiley:

pkg/operator/render.go
```diff
@@ -68,30 +69,16 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
...
 	}
 
 	ccSpec := &mcfgv1.ControllerConfigSpec{
 		ClusterDNSIP:        dnsIP,
 		CloudProviderConfig: "",
 		EtcdDiscoveryDomain: infra.Status.EtcdDiscoveryDomain,
 		Platform:            platform,
-		Proxy:               &proxy.Status,
```
runcom: You might want to rebase to get rid of this but there won't be conflicts anyway I guess

pkg/operator/render.go
```diff
@@ -68,30 +69,16 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
...
-		platform = "baremetal"
-	case configv1.GCPPlatformType:
-		platform = "gcp"
-	case configv1.OpenStackPlatformType:
-		platform = "openstack"
-	case configv1.LibvirtPlatformType:
-		platform = "libvirt"
-	case configv1.VSpherePlatformType:
-		platform = "vsphere"
+	infraPlatformString := string(infra.Status.PlatformStatus.Type)
```
abhinavdahiya: This will most likely fail 4.1 to 4.2 upgrades

abhinavdahiya: You should be using the .platformStatus only when set and fall back to old .platformType when not

pkg/operator/render.go
sgreene570: done

pkg/operator/render.go
sgreene570: Taken care of. Thanks!

kikisdeliveryservice:
verify isn't allowing the new check bc the type is deprecated here: https://github.com/openshift/api/blob/49f610c1a3a8461a9d0623ad35fa08d943b12c68/config/v1/types_infrastructure.go#L42

runcom:
Let's just add a `//nolint:linter_name` on top of it to silence the l'Inter - standard practice

kikisdeliveryservice:
> Let's just add a `//nolint:linter_name` on top of it to silence the l'Inter - standard practice

nice!! i thought that was the way but wasn't sure. just saw similar here: https://github.com/openshift/machine-config-operator/pull/918/files#diff-fa45321336db7ad1cedc28bf643a4f97

kikisdeliveryservice:
i think we're just probably waiting for that e2e test to stop flaking

Issue #815, opened 2019-06-04
filter out objects changes in MCO


runcom:
```
I0604 22:42:25.706357       1 operator.go:258] resyncing due to "consoles.config.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706388       1 operator.go:258] resyncing due to "servicecas.operator.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706400       1 operator.go:258] resyncing due to "alertmanagers.monitoring.coreos.com" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706410       1 operator.go:258] resyncing due to "prometheusrules.monitoring.coreos.com" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706422       1 operator.go:258] resyncing due to "schedulers.config.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706432       1 operator.go:258] resyncing due to "dnses.operator.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706445       1 operator.go:258] resyncing due to "ingresscontrollers.operator.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706455       1 operator.go:258] resyncing due to "kubeapiservers.operator.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706468       1 operator.go:258] resyncing due to "networks.operator.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706479       1 operator.go:258] resyncing due to "projects.config.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706491       1 operator.go:258] resyncing due to "securitycontextconstraints.security.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706501       1 operator.go:258] resyncing due to "clusterversions.config.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706511       1 operator.go:258] resyncing due to "tuneds.tuned.openshift.io" *v1beta1.CustomResourceDefinition object change
I0604 22:42:25.706521       1 operator.go:258] resyncing due to "containerruntimeconfigs.machineconfiguration.openshift.io" *v1beta1.CustomResourceDefinition object change
[...]

PR #817, opened 2019-06-05
Filter ClusterRole and ClusterRoleBindings on our labels

runcom:
This is a subset of #815 and #813 for ClusterRole and ClusterRoleBindings

manifests: remove namespace from ClusterRole/ClusterRoleBinding  …
ClusterRole(s) and ClusterRoleBinding(s) aren't namespaced anyway, see
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

pkg/operator: use filters for ClusterRole/Bindings informers  …
Since ClusterRole and ClusterRoleBindings aren't namespaced, we're
triggering a resync for every ClusterRole/Binding that don't pertain us.
Use a custom filtered informers for objects with specific MCO labels.


pkg/controller/common/controller_context.go
```diff
@@ -54,6 +58,16 @@ func CreateControllerContext(cb *clients.Builder, stop <-chan struct{}, targetNa
...
+	// This informer is specifically filtering for mco-built-in labels for cluster-role and cluster-role-bindings
+	filterClusterRoleAndClusterRoleBindings := func(opts *metav1.ListOptions) {
+		ls, err := labels.ConvertSelectorToLabelsMap(opts.LabelSelector)
+		if err != nil {
+			glog.Warningf("unable to convert selector %q to map: %v", opts.LabelSelector, err)
+			return
+		}
+		opts.LabelSelector = labels.Merge(ls, map[string]string{"mco-built-in/cluster-role": "", "mco-built-in/cluster-role-binding": ""}).String()
+	}
+	clusterRoleAndBindingsInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod()(), targetNamespace, filterClusterRoleAndClusterRoleBindings)
```
runcom: I might just name this `mcoBuiltInInformer` and we keep adding labels to the map above when we need specific objects

runcom:
/hold

cause testing and fixing and all :tada:

manifests/machineconfigcontroller/clusterrole.yaml
```diff
@@ -2,7 +2,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: machine-config-controller
-  namespace: {{.TargetNamespace}}
+  labels:
+    "mco-built-in/cluster-role": ""
```
runcom: this must be in the form `<component>.openshift.io/<functionality>` (`mco.openshift.io/something-something`)

pkg/controller/common/controller_context.go
runcom: ` NewSharedInformerFactoryWithOptions` instead of `NewFilteredSharedInformerFactory`

manifests/machineconfigcontroller/clusterrole.yaml
cgwalters: A lot of other operators are going to be installing roles too right?  I wonder if we should have a standard annotation for this like `openshift.io/operator-managed: ""`?

runcom: we could yeah, Maceij suggested the above as a common standard pattern to use for labels but I'm open to a more generalized way as well

cgwalters:
LGTM pending resolution of the label.

manifests/machineconfigcontroller/clusterrole.yaml
runcom: I think we can agree on `mco.openshift.io/managed`?

sgreene570: > I think we can agree on `mco.openshift.io/managed`?

I like this label. Much clearer than the previous label. 
Anyone have any opposing thoughts?

kikisdeliveryservice: what was the status of the generalized: `openshift.io/operator-managed: ""` bc i think that would make the most sense here if it's possible.

ericavonb: I think most operators have the CVO create the resources for their operators (CRDs, RBAC, namespaces, etc) directly and doesn't need any sort of label for filtering or ownership designations. Is there a reason the MCO is different and has to do it itself rather than using the CVO?

cgwalters: I did a quick `git log install/` which turned up:

```
commit 3792c944f2ced1be4fe8c2758f390d097a0b7927
Author:     Abhinav Dahiya <abhinav.dahiya@redhat.com>
AuthorDate: Tue Aug 7 12:42:10 2018 -0700
Commit:     Abhinav Dahiya <abhinav.dahiya@redhat.com>
CommitDate: Mon Aug 13 11:32:07 2018 -0700

    *: add operator sync and  manifests
    
    manifests/ -> are all the object that operator continuously manages.
    install/ -> are all the objects that need to be installed but not
    maned by the operator.
```

Which...hmm, is some sort of hint?  I think part of the idea is that we may need to specially handle upgrades for some of our objects?

I am pretty sure there's a real reason for this but it's escaping me at the moment...

abhinavdahiya: >  I think part of the idea is that we may need to specially handle upgrades for some of our objects?

Yes, CVO is designed to create and manage objects that are **required** for the operator to run correctly and not any of it's operands. So it made most sense to control objects so that MCO can at late stage perform migrations for these objects during upgrade...

cgwalters: Based on this I'd personally lean towards a generic `openshift.io/operator-managed: ""` and try to propagate that to other codebases as well?

runcom: Alrighty, I like it and it makes total sense, let's go for it, I'll change this and Steve can change the other PR accordingly

kikisdeliveryservice:

Feel free to reopen.

PR #820, opened 2019-06-05
templates: update default etcd conf on Azure

abhinavdahiya:
on Azure, it seems like 40%+ of peer round trips are over 25ms. Therefore etcd team has made a recommendation for

```
current: ETCD_ELECTION_TIMEOUT=1000

We would want to bump that to 2500

current: ETCD_HEARTBEAT_INTERVAL=100

500
```

/cc @hexfusion @crawford

### DIFF
--- a/templates/master/00-master/azure/files/etc-etcd-etcd-conf.yaml
+++ b/templates/master/00-master/azure/files/etc-etcd-etcd-conf.yaml
@@ -1,0 +1,18 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/etcd/etcd.conf"
+contents:
+  inline: |
+    #[member]
+    ETCD_SNAPSHOT_COUNT=100000
+    ETCD_HEARTBEAT_INTERVAL=500
+    ETCD_ELECTION_TIMEOUT=2500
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
Just for posterity, this template replaces the defaults from here for Azure:
https://github.com/openshift/machine-config-operator/blob/master/templates/master/00-master/_base/files/etc-etcd-etcd-conf.yaml

/approve

deads2k:
/hold

This cluster is getting 425 writes per minute.  That may be correct, but I'd like some confirmation before modifying tunables.

smarterclayton:
why in the world is the round trip time 25ms?

cgwalters:
In https://github.com/openshift/machine-config-operator/pull/837 we're dropping a per-platform override because people forgot to update the override copies.  This PR is introducing a new one, we don't want to e.g. override `ETCD_QUOTA_BACKEND_BYTES` too.

(Does etcd not have a `/etc/etcd.d` drop-in like dir?  Not only would it make sense, the naming would surely would win some kind of Internet award :medal_sports: )

PR #824, opened 2019-06-05
pkg/daemon: roll back kargs on error and ignore kargs update on non-RHCOS

runcom:

**- What I did**

Fixed some late review comments that I've just noticed over at https://github.com/openshift/machine-config-operator/pull/762/files#diff-13f28de2a32e0dab8f73b67aa5af9803R24

- roll back kargs on error
- do not bail when trying to update kargs on non-RHOS, just log it as we're doing for OS upgrades

**- How to verify it**

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -237,6 +237,14 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 	if err := dn.updateKernelArguments(oldConfig, newConfig); err != nil {
 		return err
 	}
+	defer func() {
+		if retErr != nil {
+			if err := dn.updateKernelArguments(newConfig, oldConfig); err != nil {
+				retErr = errors.Wrapf(retErr, "error rolling back kernel arguments %v", err)
+				return
+			}
+		}
+	}()
 
 	return dn.updateOSAndReboot(newConfig)
 }


kikisdeliveryservice:
This makes sense to me :+1: , will let @cgwalters review

pkg/daemon/update.go
```diff
@@ -432,7 +432,8 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 		return nil
 	}
 	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
-		return fmt.Errorf("Updating kargs on non-RHCOS nodes is not supported: %v", diff)
+		glog.V(2).Info("Updating kargs of non RHCOS nodes is not supported")
```
cgwalters: Hm.  It feels more to me like erroring out is the correct thing.  Probably?
I dunno.  This MCD-on-not-RHCOS is a super messy topic...

Is there a specific reason you're making this change?  I just want to understand this more.

The reason we silently ignore `osImageURL` on not-RHCOS is because it's always there.  But at least today kargs are not always there.

When the installer changes to use `KernelArguments` for `nosmt`, is it really right for us to silently ignore that request for BYO RHEL?

runcom: oh well, the only reasoning was to emulate osimageurl but your point that Kargs aren't always there makes total sense, dropping that commit! Thanks!

runcom: done

cgwalters: I keep vacillating on this; following the installer-nosmt scenario above though, since the MCD today doesn't handle kargs for not-RHCOS, the user would need to use Ansible or whatever for that, but the MCD would go into a degraded state and not handle the parts it can handle (files).

So...maybe your change is right?  I dunno.

I guess perhaps the more ideal thing would be for us to *inspect* the host and not make a change if we didn't need to - then the MCD would be degraded until Ansible ran.

runcom: > I guess perhaps the more ideal thing would be for us to _inspect_ the host and not make a change if we didn't need to - then the MCD would be degraded until Ansible ran.

I'm leaning towards this indeed. This way an admin can see something is wrong in the pool and take action instead of silently wtf'ing because kargs isn't applied (same for fips at this point)

cgwalters: Yeah, it seems clearly better, but would require reworking our kargs code.  Anyways this PR as is looks clearly right.

kikisdeliveryservice: `I guess perhaps the more ideal thing would be for us to inspect the host and not make a change if we didn't need to `

just to understand, inspect the host to see if we need to make the kargs changes? or...?

PR #825, opened 2019-06-06
pkg/operator: expose more resources to must-gather

runcom:

**- What I did**

added collection of master,worker MCP and CC

**- How to verify it**

run must-gather

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -207,6 +207,9 @@ func (optr *Operator) initializeClusterOperator() (*configv1.ClusterOperator, er
 	// RelatedObjects are consumed by https://github.com/openshift/must-gather
 	co.Status.RelatedObjects = []configv1.ObjectReference{
 		{Resource: "namespaces", Name: "openshift-machine-config-operator"},
+		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigpools", Name: "master"},
+		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigpools", Name: "worker"},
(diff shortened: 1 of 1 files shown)


pkg/operator/status.go
```diff
@@ -207,6 +207,9 @@ func (optr *Operator) initializeClusterOperator() (*configv1.ClusterOperator, er
 	// RelatedObjects are consumed by https://github.com/openshift/must-gather
 	co.Status.RelatedObjects = []configv1.ObjectReference{
 		{Resource: "namespaces", Name: "openshift-machine-config-operator"},
+		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigpools", Name: "master"},
+		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigpools", Name: "worker"},
```
cgwalters: We also need to support custom pools right?

runcom: I've already a follow up to dynamically collect and populate this - I'm running through some testing and will open it once ready

cgwalters: OK, so land as is and then improve later?  Cool by me!

runcom: yup, that's the plan, at the beginning I actually just thought about having `master` only here. The information we gather from this is questionable also, logs from controller is what we'll look at anyway (and we gather that)

runcom:
I'll add the 4.1.z BZ to the resulting cherry-pick PR, not this (bug is https://bugzilla.redhat.com/show_bug.cgi?id=1717738)

PR #826, opened 2019-06-06
pkg/daemon: support FIPS

runcom:

**- What I did**

Added a new FIPS field to MachineConfig to enable/disable FIPS on rhcos (disabled by default). This is, at this point, only intended as a day-2 operation.

The RHCOS I'm testing with hasn't yet rhe rhcos-tools to flip FIPS so I'm putting this on hold for now.

FIPS would also greatly benefit a dedicated CRD (`oc edit fips`?) - I'm experimenting with that but it will ofc require a new controller.

/hold

**- How to verify it**

added an e2e


cgwalters:
Wait dup of https://github.com/openshift/machine-config-operator/pull/800 right?

runcom:
> Wait dup of #800 right?

oh crap

Issue #294, opened 2019-01-14
How to I add machineconfigpool named infra to the cluster


jatanmalde:
Hello.

I currently only see two default `machineconfigpool` with the installation
```
[root@localhost ~]# oc get machineconfigpool
NAME      AGE
master    6d
worker    6d
[root@localhost ~]# 

```
How do I tell installer about additional `machineconfigpool` for the cluster so that I can have some infra nodes on the cluster.

Please advice on it.

wking:
The essiest way is to add additional machine(set)s and machineconfigpools as day-2 operations.  There is very little difference between you pushing those into the cluster after the installer succeeds and getting the installer to push them for you.

jatanmalde:
@wking I agree to your point, How could I add the `machineconfigpool` prior installation.

Lets say if I want to add another `machineconfigpool`. How can I achive that?

Thanks.

wking:
> I agree to your point, How could I add the machineconfigpool prior installation.

If you agree about it being easiest as a day-2 change, why do you care about adding these prior to installation?  Just push the objects you want into the cluster post-install.

jatanmalde:
@wking I did not find the Day2 -operation sections. If you can post the link then that would help.

I can take look at it once to see how can I add machineconfigpool?

Thanks

wking:
I dunno if there are docs for this yet, but if you want a new machineconfig pool, you can probably yse the existing worker pool as a template.

wking:
Looks like there may eventually be docs [here][1].

[1]: https://github.com/openshift/training/blob/master/docs/compute/09-multiple-machine-pools.md

runcom:
@wking just to iterate on this one, is it possible today, as a day-2 operation, to bring up instances on aws and mark them in a new MCpool? I've tried scaling the worker set, creating a machine pool and a machine config but everything blew up cause previous machine configs tied to "worker"s role are taking over anyway. Just trying to find a pointer to follow so I can test this out further.

PR #828, opened 2019-06-06
pkg/controller/node: fix controller to handle multiple roles

### LINKED ISSUE #429: nodes limited to just one pool
I've been following https://github.com/openshift/training/blob/master/docs/05-infrastructure-nodes.md in an attempt to create "workers" with different roles for distruptive e2e tests. So, no matter anything, creating a new machineset always apply the "worker" role to a node, resulting in nodes having multi roles:

```
ip-6.us-west-2.compute.internal   Ready                         infra,worker   9m44s   v1.12.4+a532756e37
ip-.us-west-2.compute.internal    Ready                         worker         120m    v1.12.4+a532756e37
ip-.us-west-2.compute.internal    NotReady,SchedulingDisabled   infra,worker   45m     v1.12.4+a532756e37
ip-.us-west-2.compute.internal   Ready                         worker         54m     v1.12.4+a532756e37
ip-.us-west-2.compute.internal   Ready                         infra,worker   5m27s   v1.12.4+a532756e37
ip-.us-west-2.compute.internal   Ready                         worker         120m    v1.12.4+a532756e37
ip-.us-west-2.compute.internal    Ready                         worker         120m    v1.12.4+a532756e37
ip-.us-west-2.compute.internal    Ready                         master         135m    v1.12.4+a532756e37
ip-.us-west-2.compute.internal     Ready                         master         135m    v1.12.4+a532756e37
ip-.us-west-2.compute.internal    Ready                         master         135m    v1.12.4+a532756e37
```
if you add a pool matching `infra` now (and you already have one matching `worker`) and give it an MC, it results in errors in the controller like:
```
E0214 13:55:36.896535       1 node_controller.go:195] error finding pools for node: node ip-10-0-143-182.us-west-2.compute.internal belongs to more than one MachineConfigPool, cannot proceed with this Node
```
Which is expected from the code but actually prevents multi labeled nodes from working correctly with MCPs and MCs.

This is probably not a bug (or maybe it is since multi labeled nodes are a thing). My point here is we still have to find a way to create nodes to degrade for testing in a cluster.

runcom:
If the node belongs to more than 1 pool, run it through the same logic we use
to add it to the queue for a specific pool. This avoids including nodes in
multiple pools for status reporting and multiple desiredConfig swaps.

see https://github.com/openshift/machine-config-operator/issues/429#issuecomment-499458138 for testing. W/o this patch, the worker pool will show Updating with 6 nodes cause 3 nodes are both worker and infra. If you then add an MC for worker only, nodes go through reconciliation twice (which is super bad cause of reboot).

Close #429 
Close https://github.com/openshift/machine-config-operator/issues/294

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -452,11 +452,7 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		return ctrl.syncStatusOnly(pool)
 	}
 
-	selector, err := metav1.LabelSelectorAsSelector(pool.Spec.NodeSelector)
-	if err != nil {
-		return fmt.Errorf("invalid label selector: %v", err)
-	}
-	nodes, err := ctrl.nodeLister.List(selector)
+	nodes, err := ctrl.getNodesForPool(pool)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


runcom:
This is working so nice, I encourage everyone to test this out using https://github.com/openshift/machine-config-operator/issues/429#issuecomment-499458138 (so we can keep testing it as well). Worked fine for me to create an MC just for `infra` and see it rolling on `infra` nodes only. Then create an MC for `worker` and see it rolling on `infra,worker` nodes as well as `worker`. Pretty cool.

cgwalters:
Sorry about the delay,
/lgtm

runcom:
Thanks Colin, gonna hold it for a moment once I run an upgrade with the infra pool to check if everything is ok also

/hold

runcom:
/hold cancel

upgrade went through just fine with an additional infra pool

Issue #819, opened 2019-06-05
get rid of ./lib/* fork


runcom:
This somehow depends on #588 but can be done w/o that finished as well. Code under `./lib` is basically a fork of an old project that now fully lives in openshift/library-go.

We need to:

- Fix #588 (but again, this issue can be done 99% w/o that)
- Move existing code to use `openshift/library-go/pkg/operator/resource/*` for `core` types
- drop `./lib`

As of today, since #588 isn't done yet, we can work on the second point above as follows:

- Move existing code to use  `openshift/library-go/pkg/operator/resource/*` for `core` types (instead of our `./lib/resource*` packages)
- leave under `./lib` just `./lib/*/machineconfig*` files, types and funcs since we still need those till #588 is done
- while removing funcs from the above, contribute any core func to library-go as well (if any, haven't checked).

PR #829, opened 2019-06-06
WIP: Use resource{apply,merge,read} functions provided by library-go where possible

LorbusChris:
This is a first PR to partly tackle #819

**- What I did**
Added changes to use resource{apply,merge,read} functions provided by library-go where possible instead of the forked ones from ./lib/resource{apply,merge,read}.
Removed now unused files in ./lib/resource{apply,merge,read}.
Regenerated

**- How to verify it**
CI


pkg/controller/render/render_controller.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/coreos/ignition/config/validate"
 	"github.com/golang/glog"
-	"github.com/openshift/machine-config-operator/lib/resourceapply"
+	mcoresourceapply "github.com/openshift/machine-config-operator/lib/resourceapply"
```
kikisdeliveryservice: i find this name super hard to read, would leave it as `resourceapply`

pkg/daemon/daemon.go
```diff
@@ -23,7 +23,7 @@ import (
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
-	"github.com/openshift/machine-config-operator/lib/resourceread"
+	mcoresourceread "github.com/openshift/machine-config-operator/lib/resourceread"
```
kikisdeliveryservice: same comment

pkg/operator/sync.go
```diff
@@ -15,8 +15,10 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 
-	"github.com/openshift/machine-config-operator/lib/resourceapply"
-	"github.com/openshift/machine-config-operator/lib/resourceread"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
```
kikisdeliveryservice: i think this should have a clearer name and indicate that it's from libgo

runcom: the default `resourceapply` should be from libgo going forward, so maybe have a better name for the MCO's resourceapply is what we need

kikisdeliveryservice: ah fair enough, my eyeballs just literally can't read mcoresourceapply clearly.

runcom: let's change that to be more readable (I'm afraid I'm so bad at naming tho)

kikisdeliveryservice: mcoResourceApply?

LorbusChris: > the default `resourceapply` should be from libgo going forward, so maybe have a better name for the MCO's resourceapply is what we need (related to [#829 (comment)](https://github.com/openshift/machine-config-operator/pull/829#discussion_r291268678))

that was my thinking here. I'll change it to `mcoResourceApply`.

kikisdeliveryservice: awesome!! thanks @LorbusChris !

LorbusChris:
The added override in Gopkg.toml brings in a huge amount of changes in the second commit, i.e. when running `dep ensure -update` :/
cc @runcom

pkg/operator/sync.go
```diff
@@ -138,7 +140,7 @@ func (optr *Operator) syncMachineConfigController(config renderConfig) error {
 		return err
 	}
 	sa := resourceread.ReadServiceAccountV1OrDie(saBytes)
-	_, _, err = resourceapply.ApplyServiceAccount(optr.kubeClient.CoreV1(), sa)
+	_, _, err = resourceapply.ApplyServiceAccount(optr.kubeClient.CoreV1(), nil, sa)
```
runcom: we need the recorder here and wherever we're now using library-go. Look at https://github.com/openshift/cluster-kube-controller-manager-operator/blob/a01770cde88d646d3dfaa82a16bc8950afc7bb04/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L52-L77

we need basically the very same in the operator object/instance and just pass it here `optr.recorder`

runcom: This is gonna panic otherwise

cgwalters:
Verify going OOM may need us to request more memory in the pod in openshift/release?  Probably worth trying to figure out how much memory it's using.

pkg/operator/render.go
```diff
@@ -68,6 +68,7 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 	}
 
 	platform := "none"
+	//nolint:staticcheck
 	switch infra.Status.Platform {
 	case configv1.AWSPlatformType:
```
LorbusChris: strangely, this errors out on this branch, but not on master.

runcom:
This is now fully passing - I'd go with this PR and tackle the other library-go replacements in a follow up. I want this to fully get rid of our, now, super old fork and then we can try to understand a move to sigs.k8s.io/controller-runtime as @ericavonb suggested on Slack.

LorbusChris:
ack. will tidy the commit messages and bit before the merge.
/hold

ericavonb:
To be clear, I don't think we need controller-runtime or librarygo for what we're doing. Here's what I think we should do:
1- have the CVO create the RBAC and CRDS. This is consistent with other operators and makes sense in terms of what the purpose of each operator is (the CVO is an operator to handle deploying other controllers/operators, the MCO is an operator for handling machine configuration).
2- keep the helpers in `apply/machineconfig.go`, those don’t/shouldn’t have something in library-go anyways
3- if there are any other resources that the operator needs to apply after modifications, write the struct out in the code directly rather than rely on data in some file (removing any resourceread functions) and use the client for that type (removing any resourceapply)

runcom:
> To be clear, I don't think we need controller-runtime or librarygo for what we're doing. Here's what I think we should do:

isn't that going against https://github.com/openshift/machine-config-operator/pull/817#discussion_r294566520? (apart from moving to library-go or not).

I think here in MCO isn't about consistency, but rather flexibility of managing our own resources (as Abhinav highlighted in his comment above)

ericavonb:
> isn't that going against #817 (comment)? (apart from moving to library-go or not).

Per #817 (comment):
> Yes, CVO is designed to create and manage objects that are required for the operator to run correctly and not any of it's operands.

The operand for the MCO are machine configurations and the server/daemons to facilitate that. As the operator reconciles its CR types, those CRD definitions would be requirements (along with the RBAC required to handle them) for the operator.

> I think here in MCO isn't about consistency, but rather flexibility of managing our own resources (as Abhinav highlighted in his comment above)
CRDs in particular can't be handled only by the operator without consideration to the rest of the clsuter because they are global objects. If anything needed to interact with one of them before the MCO ran, it would need that definition to exist and having more than one place attempt to create the definitions would be a much more difficult migration situation. This is why the openshift/api config CRDs are created via the cluster-config-operator* before other operators run and other operators include their CRDs in their payloads at earlier runlevels.

* created by the CVO but included in a release via the cluster-config-operator payload. with the "create-only" annotation, you can utilize the CVO to create resources that you may want to modify later.

runcom:
> 2- keep the helpers in `apply/machineconfig.go`, those don’t/shouldn’t have something in library-go anyways

I think a plus one for library-go (at least today) is about not having those helpers everywhere. See https://github.com/openshift/library-go/pull/472 also, we're currently "broken", and having a single place for the helpers is at least beneficial to catch and fix stuff like that (I just noticed about that)

LorbusChris:
will resurrect this PR on Monday!

kikisdeliveryservice:

Feel free to reopen.

PR #831, opened 2019-06-06
Standardize import paths and run gofmt

eparis:
Some files imported the same thing twice with 2 different names. Many files
called the same import something different. This just creates a bit of
continuity using the most common upstream practice:

corev1 "k8s.io/api/core/v1"
corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
corev1lister "k8s.io/client-go/listers/core/v1"
appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"

I also run the 1.12 formatter across the repo.


pkg/operator/assets/bindata.go
```diff
@@ -41,6 +41,7 @@ import (
 	"strings"
 	"time"
 )
+
```
cgwalters: This file is generated, I think you need to skip it or the next person running `hack/update-generated-bindata.sh` will revert.

cgwalters: (or, we update that script to run gofmt)

eparis: I'm just not going to touch it.

runcom:
Thanks for this @eparis (we also need to pick https://github.com/openshift/machine-config-operator/pull/600 at some point)

/approve
as well

eparis:
I split it into 2 commits. If e2e-etcd-quorum-loss known bad? If so, I think it's ready to go...

kikisdeliveryservice:
thanks @eparis !

/lgtm

PR #835, opened 2019-06-07
Bug 1718421: templates: Add missing fields to crio.conf 

umohnani8:
Certain options such as default_runtime were missing
from the crio.conf template in the MCO. This was causing
issues when upgrading cri-o from 1.13 to 1.14.

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -28,6 +28,13 @@ contents:
     #storage_option = [
     #]
 
+    # file_locking is whether file-based locking will be used instead of
+    # in-memory locking
+    file_locking = false
+
+    # Path to the lock file.
+    file_locking_path = "/run/crio.lock"
+
     # The "crio.api" table contains settings for the kubelet/gRPC interface.
     [crio.api]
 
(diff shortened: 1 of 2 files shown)


umohnani8:
@mrunalp @haircommander @runcom PTAL
Testing is currently in progress, will comment back in a bit when done

kikisdeliveryservice:
@umohnani8 is this a bug or just a normal pr for master?

umohnani8:
@kikisdeliveryservice it is a blocker for upgrades. We will need to cherry pick it into 4.1 as well.

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -28,6 +28,13 @@ contents:
     #storage_option = [
     #]
 
+    # file_locking is whether file-based locking will be used instead of
+    # in-memory locking
+    file_locking = false
+
+    # Path to the lock file.
+    file_locking_path = ""
```
haircommander: crio by default sets: `file_locking_path = "/run/crio.lock"` do we want that here too?

kikisdeliveryservice:
@umohnani8 gotcha. we'll need a BZ then (if there isnt one) and need to add Bug:XXXX to the title of the pr i think.

thanks for this!

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -196,6 +196,17 @@ contents:
     # ranges are separed by comma.
     gid_mappings = ""
 
+    # The minimal amount of time in seconds to wait before issuing a timeout
+    # regarding the proper termination of the container.
+    ctr_stop_timeout = 10
```
haircommander: rpms on 1.13.9 will generate 0 here

umohnani8: So we need to change the rpm to 10 then I believe.

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: Done!

kikisdeliveryservice:
:heart: the comments in the templates!

also fixed the title format so it would get the correct label.

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: we don't really need it, it will be set to 10 if the value is less than 10

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -56,41 +63,24 @@ contents:
...
+    #default_ulimits = []
+
+    # Path to the OCI compatible runtime used for trusted container workloads. This
+    # is a mandatory setting as this runtime will be the default and will also be
+    # used for untrusted container workloads if runtime_untrusted_workload is not
+    # set.
+
+    # default_runtime is the _name_ of the OCI runtime to be used as the default.
+    # The name is matched against the runtimes map below.
+    default_runtime = "runc"
```
cgwalters: Personally...I think it's a much cleaner pattern to have the defaults live in the *code*, and have the config files just be commented out, or mostly empty with just a few key config options and their defaults listed.  See e.g. `/etc/systemd/journald.conf`.

So...if we made that change to crio would this cease to be an upgrade hazard?

templates/master/01-master-container-runtime/_base/files/crio.yaml
mrunalp: Let's just skip adding this as discussed. It will default to 10 automatically.

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: yeah if we commented out these fields they'd be correctly populated

cgwalters: So...let's do that?

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -140,6 +130,16 @@ contents:
...
+    default_sysctls = []
+
+    # List of additional devices. specified as
+    # "<device-on-host>:<device-on-container>:<permissions>", for example: "--device=/dev/sdc:/dev/xvdc:rwm".
+    #If it is empty or commented out, only the devices
+    # defined in the container json file by the user/kube will be added.
+    additional_devices = []
+
     # hooks_dir_path is the oci hooks directory for automatically executed hooks
     hooks_dir_path = "/usr/share/containers/oci/hooks.d"
```
haircommander: it seems hooks_dir_path has been replaced with hooks_dir. hooks_dir is a slice so we will want:
`hooks_dir = ["/usr/share/containers/oci/hooks.d", ]`

cgwalters: Again if we can just comment this out and not have to change the MCO when crio changes...

templates/master/01-master-container-runtime/_base/files/crio.yaml
mrunalp: We can probably make it work. @umohnani8 If we find no handlers and a default set, then we can default to runc. Let's see if we can make 1.14.x work without having to touch the configuration. In future iterations, we can reduce the config by using more defaults :+1:

umohnani8:
Decided to take the "set defaults if nothing is set" route and opened https://github.com/cri-o/cri-o/pull/2461 for cri-o 1.14. Closing this in favour of that.

PR #836, opened 2019-06-08
Sync OpenStack kubelet config and everyone else's kubelet config

eparis:
It appears OpenStack added their own kubelet config without cloudconfig set. The pod team updated the _base kubelet config but never updated the OpenStack version. This re-sync's the files and it creates an (extremely hacky) verify script to hopefully keep them in sync.

I think the existence of this PR means we're doing it wrong. And probably shouldn't allow these per platform file overwrites. We should find a better way to templatize such differences.


eparis:
@flaper87 @sjenning 
Tagging you 2 since Flavio owns the file and since Seth's team is who forgot to update it.

eparis:
closing, i think i can do this better already

PR #837, opened 2019-06-08
Disable OpenStack cloud provider in the templating not by replacing the file

eparis:
The kubelet config for _base had been updated numerous times since the OpenStack
version of the file was forked to remove the cloud provider. Instead of forking
the template file, instead just don't set the OpenStack cloud provider config
in the template config data.


eparis:
@flaper87 @sjenning 
This came in via: 6a6fa80cc
But the file got out of date as the pod team updated _base and not the openstack version
So let's just kill the OpenStack version, then they can't get out of date

cgwalters:
I think what we really wanted here originally was to use "drop in" files rather than overrides, see https://github.com/openshift/machine-config-operator/issues/763

eparis:
I also thought about continuing to use the overrides, in https://github.com/openshift/machine-config-operator/pull/836 but decided that was even worse....

kikisdeliveryservice:
this test is flaky and has been updated to not always run so:

/skip e2e-etcd-quorum-loss

PR #840, opened 2019-06-11
Add shortnames for MachineConfig and MachineConfigPool CRDs

LorbusChris:
**- What I did**
Added mc and mcd shortnames for MachineConfig and MachineConfigPool CRDs respectively.

**- How to verify it**
`oc describe mc` and `oc describe mcp`

Added mc and mcd shortnames for MachineConfig and MachineConfigPool CRDs respectively.

### DIFF
--- a/manifests/machineconfigpool.crd.yaml
+++ b/manifests/machineconfigpool.crd.yaml
@@ -40,3 +40,6 @@ spec:
     singular: machineconfigpool
     # kind is normally the CamelCased singular type. Your resource manifests use this.
     kind: MachineConfigPool
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - mcp
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -34,3 +34,6 @@ spec:
     singular: machineconfig
     # kind is normally the CamelCased singular type. Your resource manifests use this.
     kind: MachineConfig
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - mc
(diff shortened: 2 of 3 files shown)


manifests/machineconfigpool.crd.yaml
```diff
@@ -40,3 +40,7 @@ spec:
     singular: machineconfigpool
     # kind is normally the CamelCased singular type. Your resource manifests use this.
     kind: MachineConfigPool
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - mcp
+
```
runcom: drop empty lines

runcom:
needs a `make update` as well (look at the verify job)

kikisdeliveryservice:
Did we confirm that these are unique?  A little worried about `mc` since it's so short it might have name collision elsewhere?

runcom:
> Did we confirm that these are unique? A little worried about `mc` since it's so short it might have name collision elsewhere?

Kubernetes should be taking care of conflicts when _registering_ names and short names for a CRD *(trying to find the code that does that which I've forgotten now)

runcom:
> Kubernetes should be taking care of conflicts when _registering_ names and short names for a CRD *(trying to find the code that does that which I've forgotten now)

can't find function but looking at the kubernetes repo there are various examples of how shortnames are expanded and checked for conflicts, CI would also shed some light on this

kikisdeliveryservice:
> can't find function but looking at the kubernetes repo there are various examples of how shortnames are expanded and checked for conflicts, CI would also shed some light on this

Gotcha. Just wanted to make sure there wasn't something obvious we were missing esp since mc is so short. :+1:

runcom:
sign-off from @derekwaynecarr that this isn't disrupting anything because of _adding_ it

PR #845, opened 2019-06-12
[WIP] Set labels and taints on nodes via the MCO

eparis:
This provides a declarative API to drive the setting and removing of labels and taints. It does not work.


eparis:
@derekwaynecarr mwahahahaha

cgwalters:
>  It does not work.

:laughing:

cgwalters:
What other components would consume this API?
AIUI we need labels like the `node-role` to be set via the CLI for the initial setup, but doesn't it work just fine for other components to use the API to update the `node` object via a merge?

I remember a while ago we had an issue where cluster/machine API was entirely *overwriting* taints but that was just a bug, right?

eparis:
ok, weird, maybe it does work? unexpected on the first try....

pkg/controller/node/node_controller.go
```diff
@@ -496,6 +506,88 @@ func (ctrl *Controller) getNodesForPool(pool *mcfgv1.MachineConfigPool) ([]*core
...
+		return err
+	}
+
+	return clientretry.RetryOnConflict(nodeUpdateBackoff, func() error {
+		oldData, err := json.Marshal(oldNode)
+		if err != nil {
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
```
runcom: you need a fresh copy of the node object here at every iteration if there's a conflict?

eparis:
My/The problem is that adding and removing labels and taints is an imperative action on top of what is supposed to be a declarative system. This means that it's not possible for us to create a working 3-node cluster declaratively (since the masters are created with the master taint). Yes imperative code can run and then remove the taint, but that is not kube like.

This would allow us to create an MC which explicitly removed the master taint. Such a thing is a requirement for 3-node clusters and for the RHHI/foundation initiative.

https://jira.coreos.com/browse/PROD-1207

eparis:
> ok, weird, maybe it does work? unexpected on the first try....

well, i'm not plumbing the MC into the RenderConfig yet, so I guess it doesn't work yet.


What would be the team's thought on putting the 'role' into the renderConfig? It would allow me to cut down the number of kubelet templates at least...

runcom:
> What would be the team's thought on putting the 'role' into the renderConfig?

I can't see anything against it to drive this (as it requires differentiating between roles anyway) - I'll keep reading and understanding this though

eparis:
> > What would be the team's thought on putting the 'role' into the renderConfig?
> 
> I can't see anything against it to drive this (as it requires differentiating between roles anyway) - I'll keep reading and understanding this though

It might make the code a little simpler.  Not sure. I'll probably keep slowly poking. Thank you!

cgwalters:
> This means that it's not possible for us to create a working 3-node cluster declaratively (since the masters are created with the master taint).

OK so you're trying to fix https://github.com/openshift/machine-config-operator/issues/763 - a `Closes:` here would be appropriate.  Also I think some design discussion in that issue before a big code change would have been a good idea.

> Such a thing is a requirement for 3-node clusters and for the RHHI/foundation initiative.

(Also a requirement for Code Ready Containers)

cgwalters:
With respect to implementation, did you have any opinions on my other suggestions in that issue?

eparis:
I like to show up with PoC code to show something is possible. I don't care if this code is completely rejected. Probably my years in the kernel where it is normal. I understand and can have a design discussion a heck of a lot better having coded (most of) it.

I think it's more than 763, but agree, that is the current thing forcing me to look at the space. The declarative management of labels and taints, instead of imperative, is a meaningful win to me.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
```diff
@@ -22,12 +22,12 @@ contents: |
         --kubeconfig=/var/lib/kubelet/kubeconfig \
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
-        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
+        --node-labels={{masterLabels .}} \
```
cgwalters: One thing I discovered is that `--node-labels` docs clearly says:
> Labels to add when registering the node in the cluster

So changing it post-registration won't do anything.  Hence I think we shouldn't template this; the systemd unit should register with the labels/taints.  And then the MCO comes by later and updates them dynamically.

We also want to avoid using `MachineConfig` for this because e.g. changing a MC today involves rebooting and we clearly don't need to reboot for removing the master taint.

cgwalters:
How about having the taints be part of `MachineConfigPool`?

cgwalters:
I haven't fully traced through your logic here but I suspect we have "ownership" problems.  This is quite analogous to kernel arguments which are an array of key=value strings too.  Today the way the MCD handles this is to [diff two MCs](https://github.com/openshift/machine-config-operator/blob/64af482e531f250931c442c28c5e63359e829fc4/pkg/daemon/update.go#L400) but I don't see your code doing a similar diff.

If we don't diff, then we can't easily know which args *we* wrote versus which args were written by a user or other code.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
eparis: This part was intended to make the NEXT node join correctly. Instead of getting 1 set of labels and then later having that set get updated. For existing nodes I update those:
https://github.com/openshift/machine-config-operator/pull/845/files#diff-a44cd14a0381ca49bcbaaf967b3cc9b0R556

The fact that an MC change requires a reboot is kinda stinky....

cgwalters: > This part was intended to make the NEXT node join correctly.

Doesn't hurt to have a master start !schedulable briefly.  For workers...we may indeed need support for adding user-specified custom labels on bringup for special hardware and the like, but I think most cases of that end up adding taints instead post-registration.

eparis:
> I haven't fully traced through your logic here but I suspect we have "ownership" problems. This is quite analogous to kernel arguments which are an array of key=value strings too.

> If we don't diff, then we can't easily know which args _we_ wrote versus which args were written by a user or other code.

I tried to create an API that didn't suffer from these difficulties. The API I'm proposing has an explicit 'these labels must exist' and 'these labels must not exist' concept. The hope being that I don't need to track previous state or user changes, and i'm only going to make sure the specified labels are either present or not present.

cgwalters:
> The API I'm proposing has an explicit 'these labels must exist' and 'these labels must not exist' concept. 

Mmm.  OK, I think that's workable.

So where that leaves us then is whether or not this should go through `MachineConfig` (which also goes into whether we template as above).  I still feel like it shouldn't, because of the MC-reboot issue but also because I think it's cleaner to just have *one* thing reacting to state rather than two (template controller and node controller).

Having the node controller do this is by far the easiest code wise but it makes the architecture uglier - this label stuff is a "sidecar" to what it is currently doing and isn't correctly tracked via status e.g.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
eparis: Adding labels and removing taints post-join is reasonable.  But adding a taint post join isn't really the best in my mind, since an inappropriate workload may have landed before the taint was added. So I sort of feel like the right taints should be added by the kubelet at join.

We could add every node with some megataint in the template. But then removing that taint would be required for every new pool. Which of course we can get right for our defaults, but is a likely bad stumbling block for users who aren't super savy.

I guess that means I feel like I can drop this part for labels, but the complete story seems weak if I drop it for taints....

eparis:
> How about having the taints be part of `MachineConfigPool`? (We can't have the `node-role` as part of the pool as that'd be circular, but we could have _additional_ labels be part of the pool).

We can probably do something here, yes (and that would avoid a reboot). But I admittedly really like the MC merging concept.

templates/master/01-master-kubelet/_base/units/kubelet.yaml
cgwalters: >  But adding a taint post join isn't really the best in my mind, since an inappropriate workload may have landed before the taint was added. 

True.

pkg/controller/node/node_controller.go
cgwalters: Happened to notice the mcd has `updateNodeRetry()`.

cgwalters:
I did an alternative (subset) in https://github.com/openshift/machine-config-operator/pull/849

(I don't want to discourage a new contributor but this is also a new nontrivial API surface)

Let me know what you think about it!

If we move forward with this PR, one thing to consider is doing what I do in 849 and have the MCD own updating the node.  That makes everything much clearer I think; the `spec/status` of the pool will automatically include updates to labels/taints, node controller's role will be unchanged, etc.

eparis:
I'm going to close this generic label/taint management PR. After discussion with the other openshift architects they thought this work was too far ahead of itself and we should wait for more users to want generic taint/label reconciler before we do more here.

Instead they want to do something more directed (sorta like in #849) but they thought that the scheduler operator should be what implements that.  Not the MC* codebase.

PR #849, opened 2019-06-12
daemon: Automatically sync master taint

cgwalters:
If a user or code adds the `worker` role to the masters, automatically
remove our `NoSchedule` taint.

Alternative to https://github.com/openshift/machine-config-operator/pull/845

Closes: https://github.com/openshift/machine-config-operator/issues/763

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -47,6 +47,11 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
+const (
+	masterRole = "node-role.kubernetes.io/master"
+	workerRole = "node-role.kubernetes.io/worker"
+)
+
 // Daemon is the dispatch point for the functions of the agent on the
 // machine. it keeps track of connections and the current state of the update
 // process.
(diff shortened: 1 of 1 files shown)


cgwalters:
Not tested yet

cgwalters:
This PR is silent on how the `worker` label ends up on the masters; so it isn't a complete solution (I should remove the `Closes` probably).  However, since we need to make both changes (router selects for worker explicitly at least), it simplifies things to reduce the problem down to "label worker or not", instead of both labels and taints.

This was inspired by [this bit of code](https://github.com/openshift/machine-config-operator/issues/763#issuecomment-501383523) in the GCP provider.

(It's a bit surprising this isn't an upstream kube default actually but I guess it gets into policy vs mechanism)

cgwalters:
Cleaned up a bit, added some comments.  Lifting WIP, I tested this and it works as advertised.  I can add an e2e if people want.

cgwalters:
I still feel this is true:

>  However, since we need to make both changes (router selects for worker explicitly at least), it simplifies things to reduce the problem down to "label worker or not", instead of both labels and taints.

So I'd like to propose landing this, even if we don't have a consensus on how the worker label should be added.

kikisdeliveryservice:
hey our new linter works?!

cgwalters:
...and then, Colin hesitated briefly over the `git push` button to fix some whitespace, knowing that 3 clusters would be launched as a result of this action.  But, YOLO (and so do the clusters from a PR).

cgwalters:
Closing per https://github.com/openshift/machine-config-operator/issues/763#issuecomment-502205542

PR #851, opened 2019-06-13
pkg/controller/node: add tests for multiple pools

runcom:
**- What I did**

Add tests for `getNodesForPool` - the e2e version of this is going to QE directly for now but I'm hoping we can add it to our suite eventually.

**- How to verify it**

### DIFF
--- a/pkg/controller/node/node_controller_test.go
+++ b/pkg/controller/node/node_controller_test.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	kubeinformers "k8s.io/client-go/informers"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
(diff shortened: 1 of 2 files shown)


pkg/controller/node/node_controller_test.go
```diff
@@ -325,6 +380,23 @@ func TestGetPoolForNode(t *testing.T) {
 
 func intStrPtr(obj intstr.IntOrString) *intstr.IntOrString { return &obj }
 
+func newMixedNodeSet(setlen int, roles ...map[string]string) []*corev1.Node {
```
kikisdeliveryservice: could we add a 1 liner comment about what these new helpers are?

pkg/controller/node/node_controller_test.go
```diff
@@ -213,6 +214,60 @@ func (f *fixture) expectPatchNodeAction(node *corev1.Node, patch []byte) {
 	f.kubeactions = append(f.kubeactions, core.NewPatchAction(schema.GroupVersionResource{Resource: "nodes"}, node.Namespace, node.Name, types.MergePatchType, patch))
 }
 
+func TestGetNodesForPool(t *testing.T) {
```
kikisdeliveryservice: 1 liner comment for this?

pkg/controller/node/node_controller_test.go
runcom: yup I'll add it

pkg/controller/node/node_controller_test.go
runcom: This is testing the `getNodesForPool` function - following how `Test...` in golang are written

kikisdeliveryservice: ah sorry, still drinking coffee

pkg/controller/node/node_controller_test.go
kikisdeliveryservice: thanks!

runcom: fixed!

PR #853, opened 2019-06-13
[wip] pkg/controller: Remove hard coded platform consts in controller template

sgreene570:
**- What I did**
Remove hard coded platform descriptors from `pkg/contrller/template/render.go`.
This PR depends on MCO #814 (which depends on https://github.com/openshift/installer/pull/1725).
This _may_ break upgrade testing. TBD.

**- How to verify it**
CI / run unit tests locally. 
ex. `go test -v github.com/openshift/machine-config-operator/pkg/controller/template`

Remove hard coded platform strings from controller template in pkg/controller

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -17,6 +17,7 @@ import (
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
+	configv1 "github.com/openshift/api/config/v1"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/version"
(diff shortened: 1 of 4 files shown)


LorbusChris:
Is this the right time to change [our Platform type](https://github.com/openshift/machine-config-operator/blob/master/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L116) to natively use configv1.PlatformType, instead of converting to string?

sgreene570:
> Is this the right time to change [our Platform type field](https://github.com/openshift/machine-config-operator/blob/master/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L116) to natively use configv1.PlatformType, instead of converting to string?

Maybe it is. Would switching to the API type be much of a concern for upgrades? (My guess is no, but I'm not super sure 😂)

LorbusChris:
neither am I :smile:

/cc @runcom

pkg/controller/template/render.go
```diff
@@ -33,15 +34,15 @@ const (
...
-	platformNone      = "none"
-	platformVSphere   = "vsphere"
-	platformBase      = "_base"
+	platformAWS       = string(configv1.AWSPlatformType)
+	platformAzure     = string(configv1.AzurePlatformType)
+	platformOpenstack = string(configv1.OpenStackPlatformType)
+	platformLibvirt   = string(configv1.LibvirtPlatformType)
+	platformNone      = string(configv1.NonePlatformType)
+	platformVSphere   = string(configv1.VSpherePlatformType)
+	platformBareMetal = string(configv1.BareMetalPlatformType)
```
runcom: we can get rid of these assignements and just use `string(...)`  in the switch cases below right?

runcom:
> Is this the right time to change [our Platform type field](https://github.com/openshift/machine-config-operator/blob/master/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L116) to natively use configv1.PlatformType, instead of converting to string?

I'd wait for the dependent installer PR to land first and then we can definitely think about moving the ControllerConfig to use the right values

pkg/controller/template/render.go
sgreene570: sure, just did this for readability because the vars get used a few times

sgreene570:
> > Is this the right time to change [our Platform type field](https://github.com/openshift/machine-config-operator/blob/master/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L116) to natively use configv1.PlatformType, instead of converting to string?
> 
> I'd wait for the dependent installer PR to land first and then we can definitely think about moving the ControllerConfig to use the right values

Just realized this PR uses capitalized constants (from type defs in API) while the rest of MCO repo uses all lowercase constants.  Will push up a lower case fix for the time being.

While switching to natively use configv1.PlatformType may cause some issues, it would certainly remove any capitalization / formatting questions.

kikisdeliveryservice:

Feel free to reopen.

PR #855, opened 2019-06-13
operator: fix logic setting version for progressing status

kikisdeliveryservice:
Background: in CI we are seeing a cluster that is in the middle of upgrading having a CVO status of Available in version target version as opposed to the expected status of progressing until the upgrade is completed.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -29,7 +29,7 @@ func (optr *Operator) syncVersion() error {
 	}
 
 	// keep the old version and progressing if we fail progressing
-	if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) && cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded) {
+	if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) || cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded) {
 		return nil
 	}
 


cgwalters:
On the face of it, this seems like an obvious change.  *But* the logic here is... twisty.  `syncProgressingStatus()` which is responsible for setting progressing also looks at the version store, so this feels circular.

kikisdeliveryservice:
@cgwalters Yeah im not 100% on this but what i think is happening is that syncProgressing is correctly called but then syncVersion which is called right after (without this change) is updating the version of Progressing operator. Which fastforward, ends up as the new version is available while the cluster is updating when it _should_ be the old version is available and the new version is only available once the update is complete.

wking:
Cross-linking https://bugzilla.redhat.com/show_bug.cgi?id=1708454

runcom:
We surely have a faulty check in syncRequitedMachineConfitPools as well. I can't explain how it passes the check to report Available otherwise. I believe the issue lays down in a race which I'm still trying to find.

kikisdeliveryservice:
Test timed out but poking around in the logs all of the nodes were succesfully updated. going to retest again:

/test e2e-aws-upgrade

kikisdeliveryservice:
so poking around in the logs progressing is now correc since the upgrade did not finish and stopped at 90%t:
```
{
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-06-13T22:48:43Z",
                "generation": 1,
                "name": "machine-config",
                "resourceVersion": "35095",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/machine-config",
                "uid": "657b3a25-8e2d-11e9-9c98-12d94da3a8e8"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-06-13T22:49:22Z",
                        "message": "Cluster has deployed 0.0.1-2019-06-13-223247",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2019-06-13T23:19:51Z",
                        "message": "Working towards 0.0.1-2019-06-13-223247",
                        "status": "True",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-06-13T22:48:43Z",
                        "status": "False",
                        "type": "Degraded"
                    }
                ],
                "extension": {
                    "master": "all 3 nodes are at latest configuration rendered-master-b24a8176c20faebab011f3ed6e554d34",
                    "worker": "all 3 nodes are at latest configuration rendered-worker-18b7b8832dcd836e7d385662b1a4d7cf"
                },```

But I really am not sure about the Available message bc it's misleading?  Shouldn't the available version be the operator version not the cluster version? The progressing message tells us what we are progressing to, but the Available message should tell us where we are - no?? https://github.com/openshift/machine-config-operator/blob/0d843aa29c6a5ce5aedcd1cb1638d2383ea17e41/pkg/operator/status.go#L65

**Comment by kikisdeliveryservice** [comment]
but wouldnt also it make sense if the mco statuses in the clusteroperator.json were ordered from newest to oldest?

**Comment by kikisdeliveryservice** [comment]
so this PR seems to solve the initial problem of the version being set before the upgrade is finished (see operator logs, it's no longer immediately updating version), but I'm hitting another problem where the test keeps timing out?  I rekicked another PR on the e2e-aws-upgrade here: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/851/pull-ci-openshift-machine-config-operator-master-e2e-aws-upgrade/699  that test eventually passed but took 3 hours(???)  So I really can't tell if this pr is failing on the upgrade e2e bc of my change (even tho looking at logs it seems ok) or bc the test is just taking forever and timing out due to other things. 

cc: @runcom @wking

**Comment by runcom** [comment]
```
I0614 04:03:02.618360       1 status.go:82] Pool master: All nodes are updated with rendered-master-624af8792aa560921625cd72e57628cf


I0614 04:03:03.005071       1 event.go:209] Event(v1.ObjectReference{Kind:"", Namespace:"", Name:"machine-config", UID:"9437fbd5-8e53-11e9-82ab-12f47fb64940", APIVersion:"", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'OperatorVersionChanged' clusteroperator/machine-config-operator version changed from [{operator 0.0.1-2019-06-14-030242}] to [{operator 0.0.1-2019-06-14-030413}]
```

MCC and MCO logs look sane

**Comment by runcom** [comment]
Ok, now I'm understanding more, from the first job failure (see also @kikisdeliveryservice comments at https://github.com/openshift/machine-config-operator/pull/855#issuecomment-501934752).

- Progressing=True at the new version
- Available=True at the new version
- Degraded=False
- **ClusterOperator status still at old version** (check here https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/855/pull-ci-openshift-machine-config-operator-master-e2e-aws-upgrade/694/artifacts/e2e-aws-upgrade/clusteroperators.json)

The above happens because in setting Available we only check if we're not degraded. We should also do NOT report the new version if we're still progressing I guess.

**Comment by runcom** [comment]
```diff
diff --git a/pkg/operator/status.go b/pkg/operator/status.go
index ec89f98e..05a2b71b 100644
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -61,15 +61,14 @@ func (optr *Operator) syncAvailableStatus() error {
                return nil
        }
 
-       optrVersion, _ := optr.vStore.Get("operator")
        degraded := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded)
-       message := fmt.Sprintf("Cluster has deployed %s", optrVersion)
+       message := fmt.Sprintf("Cluster has deployed %s", co.Status.Versions)
 
        available := configv1.ConditionTrue
 
        if degraded {
                available = configv1.ConditionFalse
-               message = fmt.Sprintf("Cluster not available for %s", optrVersion)
+               message = fmt.Sprintf("Cluster not available for %s", co.Status.Versions)
        }
 
        coStatus := configv1.ClusterOperatorStatusCondition{
```

The above takes care of my last comment. We were wrongly picking up the new version and set that w/o waiting completely for Progressing=False which sets the new version in the clusteroperator object.

pkg/operator/status.go
```diff
@@ -29,7 +29,7 @@ func (optr *Operator) syncVersion() error {
 	}
 
 	// keep the old version and progressing if we fail progressing
-	if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) && cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded) {
+	if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) || cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded) {
```
runcom: debugging with Alex. This `||` makes it impossing to flip from progressing=True to False because we now never report the new version to the cluster object.

runcom: in other words, there's no way right now to flip from progressing=true & old version to progressing=false & new version reported here

runcom:
Ok, looking again at https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/23161/pull-ci-openshift-origin-release-4.1-e2e-aws-upgrade/51/artifacts/e2e-aws-upgrade/pods/

The progressing logic is flawed and we're thinking about relying that status on the master pool status as well

runcom:
Ok, the race is:

- 18:07:56: new Operator goes up
- start syncing
- `syncRequiredMachineConfigPools` passes cause pool master is idle but didn't start updating to new rendered-mc
- 18:09:28: new version reported
- 18:10:06: the master pool starts rolling the new templates

`syncRequiredMachineConfigPools` needs a stronger check

cgwalters:
We really need the full chain to be complete to report progressing=false: 

 - operator is updated
 - controller has rolled out and is idle (most importantly the template, render sub-controllers)
 - the master pool's spec config must = status

Note that we don't need to wait for the node controller to be idle because that's covered by the last condition.

runcom:
What's happening is that the Generation of a pool isn't updated (always at 1) because we never update its spec. We only update its status.configuration in 4.1. Colin's patch shoudl fix that

runcom:
> We really need the full chain to be complete to report progressing=false:

the problem isn't there though. That code runs fine, the failure is in checking the correct generation of the MCP we're comparing against in syncRequiredMachineConfigPool and I'm gonna check that we don't have such issue in master cause of https://github.com/openshift/machine-config-operator/pull/773

kikisdeliveryservice:

Feel free to reopen.

PR #610, opened 2019-04-09
*: pre-transition ign refactor

kikisdeliveryservice:
**- What I did**
Now that we have more time, I am breaking up PR #583 . This PR is removing the changes from #583 that can go into the code-base now to make the future transition easier. 

This PR makes the following changes:
- update helpers.go NewIgnConfig() to use MaxVersion and sub igntypes for ignv2_2types as package alias
- for all non-test files, substitute igntypes for ignv2_2types as package alias (these changes are trivial but take up a lot of space in the PR, let's do it now!)
- in test files, make the above substitution, and also refactor to use NewIgnConfig() instead of generating empty config and hardcoding version.
- minor doc update to reflect above changes

After this lands, the future subtantive ignition change PR should be clearer and easier to test and evaluate.

**- How to verify it**
This PR makes no functional changes and everything should work exactly the same as before.

PR #856, opened 2019-06-14
[release-4.1] Bug 1706082: Add Spec.Configuration to MachineConfigPool, render controller writes it

cgwalters:
Cherry-picked from 37e8466a1a5d809b31a1e7782aeecb8abf62752b

See https://github.com/openshift/machine-config-operator/pull/765#issuecomment-493136113

MachineConfigPool needs a `Spec.Configuration` and `Status.Configuration`
[just like other objects][1] so that we can properly detect state.
Currently there's a race because the render controller may set `Status.Configuration`
while the pool's `Status` still has `Updated`, so one can't reliably check whether the
pool is at a given config.

With this, ownership is clear: the render controller sets the spec, and the node controller
updates the status.

[1] https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -299,6 +299,9 @@ type MachineConfigPoolSpec struct {
 	// MaxUnavailable specifies the percentage or constant number of machines that can be updating at any given time.
 	// default is 1.
 	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable"`
+
+	// The targeted MachineConfig object for the machine config pool.
+	Configuration MachineConfigPoolStatusConfiguration `json:"configuration"`
 }
 
 // MachineConfigPoolStatus is the status for MachineConfigPool resource.
--- a/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
@@ -633,6 +633,7 @@ func (in *MachineConfigPoolSpec) DeepCopyInto(out *MachineConfigPoolSpec) {
 		*out = new(intstr.IntOrString)
 		**out = **in
 	}
+	in.Configuration.DeepCopyInto(&out.Configuration)
 	return
 }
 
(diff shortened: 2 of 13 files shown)


runcom:
This patch makes sure the Generation field for an MCP is correctly increased (it's not right now in 4.1 because we never update the Spec field of the MCP).
With this, `syncRequiredMachineConfigPools` correctly checks that the MCP is actually rolling the new upgrade changes.

The race we've been seeing is:

the race is:

- 18:07:56: new Operator goes up
- start syncing
- syncRequiredMachineConfigPools passes cause pool master is idle but didn't start updating to new rendered-mc
- 18:09:28: new version reported
- 18:10:06: the master pool starts rolling the new templates

cgwalters:
Going to wait for a CI run before I force push again to retrigger Prow's check of the Bugzilla target (which is now `4.1.z` again)

runcom:
Looks like there is a compilation issue in e2e

cgwalters:
Oh duh, looking at the diff the test suite changes are entangled with some prep PRs like https://github.com/openshift/machine-config-operator/pull/765/commits/9970f247f93713e15b79bedb40ae5f7b8fe63b58
which itself is built on top of https://github.com/openshift/machine-config-operator/pull/610 which touches the whole codebase.  Going to try redoing the e2e test changes without that last one.

cgwalters:
> ci/prow/e2e-aws-op — Job succeeded. 

:tada:

kikisdeliveryservice:
yay!!

/lgtm

cgwalters:
Anyone know who takes care of "Pending — Not mergeable. Needs cherry-pick-approved label. " ?

kikisdeliveryservice:
> Anyone know who takes care of "Pending — Not mergeable. Needs cherry-pick-approved label. " ?

yeah i talked to @runcom and i'll get that on next week.

eparis:
(Parent BZ)[https://bugzilla.redhat.com/show_bug.cgi?id=1721583] is still not verified by QA. Reaching out to QA might be helpful when I re-review for 4.1.4 inclusion tomorrow.

I see you have some CI testing around this. But I'm not entirely sure of the source of these changes. Is this a combination of https://github.com/openshift/machine-config-operator/pull/765 and https://github.com/openshift/machine-config-operator/pull/773 ?

What's the impact of not merging this?

runcom:
> What's the impact of not merging this?

we won't tell exactly that an upgrade completed because we'll report "OK" doing the comparison with the old version (this is what this fixes)

cgwalters:
> But I'm not entirely sure of the source of these changes. Is this a combination of #765 and #773 ?

Yep, exactly.

> What's the impact of not merging this?

The bug symptoms would persist; the MCO could claim it was done upgrading when it hadn't (transiently).

And/or we'd have to come up with a different fix which...seems at best equally risky.  We know this code works in 4.2.

eparis:
we've been in master since May 16. Even though QA has not VERIFIED I'm going to let this through. We likely have had a reasonable about of CI soak time in master.

PR #858, opened 2019-06-14
Update templates for CRI-O 1.14

haircommander:
**- What I did**
Updated the crio.yaml templates to the current configuration
**- How to verify it**
./bin/crio --config $OLD_TEMPLATE > $NEW_TEMPLATE
then replace file_locking{,_path} defaults, as file locking isn't desired (these values were in the wrong place before, so they weren't effecting the config despite being set)
Update crio.yaml templates to match CRI-O 1.14.x config format

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -118,17 +118,6 @@ contents:
     # only the capabilities defined in the containers json file by the user/kube
     # will be added.
     default_capabilities = [
-        "CHOWN",
-        "DAC_OVERRIDE",
-        "FSETID",
-        "FOWNER",
-        "NET_RAW",
-        "SETGID",
-        "SETUID",
-        "SETPCAP",
-        "NET_BIND_SERVICE",
-        "SYS_CHROOT",
-        "KILL",
     ]
 
     # List of default sysctls. If it is empty or commented out, only the sysctls
@@ -145,6 +134,7 @@ contents:
 
     # Path to OCI hooks directories for automatically executed hooks.
     hooks_dir = [
+        "/usr/share/containers/oci/hooks.d"
     ]
 
     # List of default mounts for each container. **Deprecated:** this option will
@@ -274,9 +264,10 @@ contents:
     [crio.network]
 
     # Path to the directory where CNI configuration files are located.
+    # Note this default is changed from the RPM.
     network_dir = "/etc/kubernetes/cni/net.d/"
 
-    # Paths to directories where CNI plugin binaries are located.
-    plugin_dirs = [
-        "/var/lib/cni/bin",
-    ]
+    # Path to directories where CNI plugin binaries are located.
+    # This option is deprecated and will be replaced with plugin_dirs
+    # Note this default is changed from the RPM.
+    plugin_dir = "/var/lib/cni/bin"
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -3,152 +3,182 @@ mode: 0644
 path: "/etc/crio/crio.conf"
 contents:
   inline: |
-    # The "crio" table contains all of the server options.
+    # The CRI-O configuration file specifies all of the available configuration
+    # options and command-line flags for the crio(8) OCI Kubernetes Container Runtime
+    # daemon, but in a TOML format that can be more easily modified and versioned.
+    #
+    # Please refer to crio.conf(5) for details of all configuration options.
+
+    # CRI-O reads its storage defaults from the containers-storage.conf(5) file
+    # located at /etc/containers/storage.conf. Modify this storage configuration if
+    # you want to change the system's defaults. If you want to modify storage just
+    # for CRI-O, you can change the storage configuration options here.
     [crio]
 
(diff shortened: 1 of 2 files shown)


haircommander:
I think we also want this in release-4.2, I can open a PR if needed, but I figured we could review/get it ready here before doing the backport.
cc: @umohnani8 @mrunalp

umohnani8:
LGTM, thanks @haircommander!

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
```diff
@@ -3,250 +3,276 @@ mode: 0644
...
-    # registries is used to specify a comma separated list of registries to be used
-    # when pulling an unqualified image (e.g. fedora:rawhide).
+    # List of registries to be used when pulling an unqualified image (e.g.,
+    # "alpine:latest"). By default, registries is set to "docker.io" for
+    # compatibility reasons. Depending on your workload and usecase you may add more
+    # registries (e.g., "quay.io", "registry.fedoraproject.org",
+    # "registry.opensuse.org", etc.).
     #registries = [
-    # ]
+    #     "docker.io",
```
ashcrow: nit: uneven indentation

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -3,250 +3,276 @@ mode: 0644
...
-    # registries is used to specify a comma separated list of registries to be used
-    # when pulling an unqualified image (e.g. fedora:rawhide).
+    # List of registries to be used when pulling an unqualified image (e.g.,
+    # "alpine:latest"). By default, registries is set to "docker.io" for
+    # compatibility reasons. Depending on your workload and usecase you may add more
+    # registries (e.g., "quay.io", "registry.fedoraproject.org",
+    # "registry.opensuse.org", etc.).
     #registries = [
-    # ]
+    #     "docker.io",
```
ashcrow: nit: uneven indentation

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
haircommander: ah good catch, fixed

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: also fixed here

kikisdeliveryservice:
> I think we also want this in release-4.2, I can open a PR if needed, but I figured we could review/get it ready here before doing the backport.
No backports needed. :smile:

runcom:
@umohnani8 @haircommander do we need to wait for that particular crio version to first land in rhcos before merging this?

/approve
/retest

haircommander:
@runcom yeah the new format isn't backwards compatible. CRI-O should be rolled out in the next week and I'll revisit this once it is.

runcom:
> @runcom yeah the new format isn't backwards compatible. CRI-O should be rolled out in the next week and I'll revisit this once it is.

thanks, I guess that it's fine as long as we don't break upgrades. Holding till we have the new CRI-O

/hold

haircommander:
1.14.5 is in the puddle, let's try again
/retest
/hold cancel

kikisdeliveryservice:
@runcom this hit the same bug as my PR as well! (BZ 1724346)

haircommander:
Happy tests @runcom PTAL

haircommander:
ah, shoot, @runcom you were right, it messes with upgrades https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.1/100

I may update with something that will work for 4.3, but in the meantime

/close

haircommander:
/reopen
I lied, I updated it so it is backwards compatible but still updates a couple of things. We can go for the plugin_dirs change in 4.3. I'll test the upgrade once I have the ci image

cgwalters:
To handle upgrades like this I think we'd need to have the renderer verify that the target `osImageURL` is new enough..though honestly a mess this gets into is that we're rewriting the configuration for the *current* deployment so it's not really transactional...

Probably indeed the simplest is to land incompatible config changes only once we know that OS content we're upgrading *from* is compatible, so drop 4.1 cruft in 4.3 as you say, relying there being no direct Cincinnati edges from 4.1 :arrow_right: 4.3.

haircommander:
@cgwalters my original hope was the config changes would roll over after crio had been upgraded, and opened the PR with that assumption. Given that is not necessarily the case, I see no problem updating the config a release later

runcom:
> I lied, I updated it so it is backwards compatible but still updates a couple of things. We can go for the plugin_dirs change in 4.3. I'll test the upgrade once I have the ci image

isn't there no way to make these crio config changes backward compatible?

haircommander:
@runcom The one change that breaks backwards compatibility was reverted (plugin_dirs ->plugin_dir). everything else is mostly a slight rearrangement/update to the config version for 1.13.

I tried this update to test an upgrade from 4.1.4, but something is wrong with the way I'm testing it (cluster bot), as it fails along with another PR from this repo that seemed inoffensive (https://github.com/openshift/machine-config-operator/pull/947) I will keep trying the upgrade, but this PR should not offend crio 1.13

/retest

haircommander:
I'm going to close this for now, but prepare for a 1.16 version coming asap

PR #859, opened 2019-06-14
Import pivot code

cgwalters:
This is prep for having "early MCD" process kernel arguments
from MachineConfig objects at pivot time, and more generally
building up the model that the host processes `MachineConfig`,
not just Ignition.

Right now what I've tested is just manually running
`machine-config-daemon pivot registry.svc.ci.openshift.org/rhcos/machine-os-content:latest`

But, next step is to test having this take over the early pivot role,
then we can drop `/usr/bin/pivot` out of RHCOS.

See: https://github.com/openshift/machine-config-operator/issues/798

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -1,0 +1,376 @@
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	// Enable sha256 in container image references
+	_ "crypto/sha256"
+
+	"github.com/golang/glog"
+	daemon "github.com/openshift/machine-config-operator/pkg/daemon"
+	"github.com/openshift/machine-config-operator/pkg/daemon/pivot/types"
+	"github.com/openshift/machine-config-operator/pkg/daemon/pivot/utils"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// flag storage
+var keep bool
+var reboot bool
+var exit77 bool
+
+const (
+	// the number of times to retry commands that pull data from the network
+	numRetriesNetCommands = 5
+	etcPivotFile          = "/etc/pivot/image-pullspec"
+	runPivotRebootFile    = "/run/pivot/reboot-needed"
+	// Pull secret.  Written by the machine-config-operator
+	kubeletAuthFile = "/var/lib/kubelet/config.json"
+	// File containing kernel arg changes for tuning
+	kernelTuningFile = "/etc/pivot/kernel-args"
+	cmdLineFile      = "/proc/cmdline"
+)
+
+// TODO: fill out the whitelist
+// tuneableArgsWhitelist contains allowed keys for tunable arguments
+var tuneableArgsWhitelist = map[string]bool{
+	"nosmt": true,
+}
+
+var pivotCmd = &cobra.Command{
+	Use:                   "pivot",
+	DisableFlagsInUseLine: true,
+	Short:                 "Allows moving from one OSTree deployment to another",
+	Args:                  cobra.MaximumNArgs(1),
+	Run:                   Execute,
+}
+
+// init executes upon import
+func init() {
+	rootCmd.AddCommand(pivotCmd)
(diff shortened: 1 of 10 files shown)


runcom:
This is amazing.

cgwalters:
> This is amazing.

https://www.youtube.com/watch?v=yhmrIGw5Voc

cgwalters:
This introduces into the MCO the same CI issue that kubelet has: https://github.com/openshift/release/pull/2747

But once we fix that there will be huge benefits because this repository will become an easy place to add functionality to RHCOS that's on Github and CI gated in a useful way.

cgwalters:
Lifting WIP on this, it doesn't actually *do* anything right now by default, but should be a useful baseline for further iteration.  (For example there's various cleanups/dedup that we can do to the imported pivot code, but I want to do that once we have CI testing set up, etc.)

pkg/daemon/daemon.go
```diff
@@ -1103,10 +1103,10 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) bool
 	return true
 }
 
-// getRefDigest parses a Docker/OCI image reference and returns
+// GetRefDigest parses a Docker/OCI image reference and returns
 // its digest, or an error if the string fails to parse as
 // a "canonical" image reference with a digest.
-func getRefDigest(ref string) (string, error) {
+func GetRefDigest(ref string) (string, error) {
```
runcom: for later, we might want to get these funcs out of pkg/daemon into an helper/rhcos utilities package

pkg/daemon/pivotutils/run.go
```diff
@@ -0,0 +1,87 @@
...
+
+// RunIgnoreErr is like Run(..), but doesn't exit on errors
+func RunIgnoreErr(command string, args ...string) {
+	if _, err := runImpl(false, command, args...); err != nil {
+		glog.Warningf("(ignored) %s: %s", command, err)
+	}
+}
+
+// RunGetOut is like Run(..), but get the output as a string
+func RunGetOut(command string, args ...string) string {
```
runcom: this is dup'ed with some other code we have in pkg/daemon (let's clean it up afterwards though).

runcom:
Some really trivial nits which we can postpone but I'd love to get the pkg/daemon/pivot for this. Otherwise, it looks great.

/approve

pkg/daemon/pivotutils/run.go
ashcrow: Agreed :+1:

pkg/daemon/daemon.go
cgwalters: I think where we should likely go is having a high level API exposed by the daemon that's just called by `cmd/machine-config-daemon/main.go`.

runcom: uhm, code under `cmd` isn't something we might want to have as API. What's under `pkg/daemon` might be a best fit for something like that for sure

pkg/daemon/pivottypes/tuneargument.go
```diff
@@ -0,0 +1,8 @@
+package types
```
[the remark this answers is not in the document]
cgwalters: Done

pkg/daemon/daemon.go
cgwalters: Yeah, that's what I meant.

runcom: I'm blind, you're right sorry for the extra noise

cmd/machine-config-daemon/pivot.go
```diff
@@ -0,0 +1,376 @@
...
+		args = append(args, authArgs...)
+		args = append(args, container)
+		utils.RunExt(false, numRetriesNetCommands, "podman", args...)
+	}
+
+	inspectArgs := []string{"inspect", "--type=image"}
+	inspectArgs = append(inspectArgs, fmt.Sprintf("%s", container))
+	output := utils.RunExt(true, 1, "podman", inspectArgs...)
+	var imagedataArray []types.ImageInspection
+	json.Unmarshal([]byte(output), &imagedataArray)
```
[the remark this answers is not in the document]
sgreene570: oops i commented too late :joy:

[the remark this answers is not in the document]
runcom: This probably got imported as is from openshift/pivot and yeah, we should have an error check here - let's follow up all the things here :)

PR #860, opened 2019-06-15
Bug 1720771 : pkg/operator: wait for all pools

runcom:
**- What I did**

It seems clear that it's expected that we wait for all pools before flipping Progressing=False at the end an upgrade so this patch just does that. I've increased the pool timeout to 15 mins also but we give back feedback anyway.

I still suspect the CVO is gonna just timeout and report a failed upgrade but we'll see how that goes during the BZ testing with more than 3 workers.

**- How to verify it**

Install a cluster and run an upgrade

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -5,21 +5,19 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/pkg/errors"
-
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/util/wait"
-
 	"github.com/openshift/machine-config-operator/lib/resourceapply"
 	"github.com/openshift/machine-config-operator/lib/resourceread"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/operator/assets"
 	"github.com/openshift/machine-config-operator/pkg/version"
+	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


cgwalters:
Is this special enough that the CVO should gain some custom logic for the worker pool, and it gets reported separately?  I think I'd tentatively vote that we keep the current behavior instead, and just teach people to do:

`oc describe machineconfigpool/worker`

And also the console should render that by default.

runcom:
> Is this special enough that the CVO should gain some custom logic for the worker pool, and it gets reported separately? I think I'd tentatively vote that we keep the current behavior instead, and just teach people to do:
> 
> `oc describe machineconfigpool/worker`
> 
> And also the console should render that by default.

I'm also leaning towards such approach (I've commented in the Bugzilla), we can move the conversation there as well.

runcom:
My idea was to still report "upgraded" but Degraded=True for the MCO cause it's really "degraded" till the worker pool or any other pool is still upgrading.

cgwalters:
To be clear I'm not opposed to this patch either!  It deserves some architectural discussion for sure though.

cgwalters:
To be clear I'm not opposed to this patch either!  It deserves some architectural discussion for sure though.

runcom:
> To be clear I'm not opposed to this patch either! It deserves some architectural discussion for sure though.

yep, I'll try to gather people and discuss this in the next few days also :+1:

cgwalters:
OK for now then let's
/hold

pkg/operator/sync.go
```diff
@@ -326,7 +320,7 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 	}
 
 	var lastErr error
-	if err := wait.Poll(time.Second, 10*time.Minute, func() (bool, error) {
+	if err := wait.Poll(time.Second, 15*time.Minute, func() (bool, error) {
```
ericavonb: Use a const like the other timeouts? e.g. https://github.com/openshift/machine-config-operator/blob/eab95d598b9b32361d0ea971906e2ebe93217201/pkg/operator/sync.go#L375

runcom:
@crawford could you chime in here?

jupierce:
For clarification on https://bugzilla.redhat.com/show_bug.cgi?id=1720771 , `oc describe machineconfigpool/worker` provides useful feedback for humans but the continuous release pipeline programmatically monitors cluster upgrades and ideally we would have frozen apis (like high level cluster operator status) to rely on instead of trying to inspect describe output.

kikisdeliveryservice:

Feel free to reopen.

sinnykumari:
I see this question keeps coming on that why MCO considers upgrade as done while few nodes are still being upgraded. Seems a valid concern to me. Shall we resurrect this  PR?

PR #861, opened 2019-06-17
Bug 1720872: Add coreos-growpart systemd dependency to kubelet

rphillips:
**- What I did**
Two part series to fix. 1. Adds a dependency to the kubelet unit file to
start after coreos-growpart.service. 2. RHCOS needs a tweak in the unit
file for coreos-growpart to add `Type=oneshot` to wait for the
coreos-growpart service to complete.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1720872

**- How to verify it**

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -3,8 +3,9 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service crio.service
-  After=crio.service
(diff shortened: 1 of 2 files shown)


cgwalters:
Since the [RHCOS change here](https://url.corp.redhat.com/e3fe57a) needs to land and includes the inverse (`Before` instead of `After`), is it still worth doing this?

I'm not so sure about the `Wants=` either...it feels like if the service wasn't enabled for some reason, that's probably the admin's decision and we should respect it?  But I'm not opposed either.

rphillips:
Nice. that fix will work great. closing this one.

PR #862, opened 2019-06-17
Mcd startup cleanup

cgwalters:
More prep for https://github.com/openshift/machine-config-operator/issues/798

I think the second commit is right but need a BYOH team member to verify.


runcom:
For the second commit, as long as it doesn't break the Ingnition path, we're good as they only use that for now.

vrutkovs:
/test e2e-rhel-scaleup

This would probably fail on CRI-O anyway

vrutkovs:
That doesn't seem to break MCD once-from, however we can't fully validate the nodes join until https://github.com/openshift/openshift-ansible/pull/11698 lands

cgwalters:
Pushed a few more changes to this one.

ashcrow:
This is looking good so far!! :heart:

cgwalters:
Ohh now I see...I'm basically trying to undo some of the changes in https://github.com/openshift/machine-config-operator/commit/165f8d47071af70d076f870cbd52c37e54b4c713 that factored out the boot id into start so the unit tests could pass a dummy value.

I think testing most of the MCD is just hard to mock fundamentally.  @runcom (and others) how about for the MCD we just unit test "easy" parts and leave more heavy lifting testing to the cluster e2e tests?

kikisdeliveryservice:
> I think testing most of the MCD is just hard to mock fundamentally. @runcom (and others) how about for the MCD we just unit test "easy" parts and leave more heavy lifting testing to the cluster e2e tests?

Yeah I agree.. I think the unit tests work best when we are testing some of our logic heavy parts (for ex reconcilable) & "give it this/get that" type functions but don't map that well onto other things imo - sometimes it's either too hard or I end up thinking "I wrote this long unit test but is it really adding any value???" .  So I'd say if you feel like the unit test adds value sure, but if you're just doing it to have a unit test but feel like it's diminishing returns it's prob better suited as e2e.

cgwalters:
And my bootID changed passed CI because it runs on Linux, but I think Antonio at least was running the unit tests on a Mac, so `make test-unit` would fail in that case.

cgwalters:
Another variant of this is https://github.com/openshift/machine-config-operator/issues/760

cgwalters:
I think I suggested this earlier but maybe what we can do is call `os.Getuid()` inside the daemon and if it's non-zero assume we're in a dummy mode and don't try to do anything that changes the OS or even assumes we're on Linux.

(To emphasize the reason I'm trying to clean up the MCD entrypoint is I need to add a new variant and there was already a lot of duplication with just 2, adding 3 needs cleanup)

runcom:
> I think testing most of the MCD is just hard to mock fundamentally. @runcom (and others) how about for the MCD we just unit test "easy" parts and leave more heavy lifting testing to the cluster e2e tests?

yup, I concur with that. I also think our code should be simple enough and well factored that unit tests are still a thing. Unit and e2e testing is anyway really different and there's value in both (otherwise we would have one or the other).

The value in writing units is about writing simple code cause you're forced to factor out functions into simple ones to unit test (and validate them).

The MCD is huge thought so again, I agree with that above in general.

This is ready to go now right? I can see all tests passing.

/approve

cgwalters:
>  I also think our code should be simple enough and well factored that unit tests are still a thing. 

Oh of course.  For example I think the tests I added for [kargs](https://github.com/openshift/machine-config-operator/blob/9fc686030853b8a03e7a102287e71aa0d59377da/pkg/daemon/update_test.go#L229) worked out well; it's just changing data structures in memory, totally safe and appropriate for a unit.

Looking at say [the ssh key unit tests](https://github.com/openshift/machine-config-operator/blob/9fc686030853b8a03e7a102287e71aa0d59377da/pkg/daemon/update_test.go#L402) we mock out the actual writing.  But if during a code refactor we stopped using the mocking accidentally the unit tests would try to rewrite the user's ssh keys potentially...

> This is ready to go now right? I can see all tests passing.

Yeah, but it will break running the unit tests on MacOS as is.  Do you care?  Any opinion on which approach to take from above?  I think I'd lean towards the `os.Getuid()` approach.

I can do that as a followup?  Or do we want to do it in this PR?

runcom:
> Yeah, but it will break running the unit tests on MacOS as is. Do you care? Any opinion on which approach to take from above? I think I'd lean towards the `os.Getuid()` approach.
> 
> I can do that as a followup? Or do we want to do it in this PR?

I'm the only one with MacOS so I'll just go ahead and live with that test till we follow up to fix this with, uhm, yeah `os.Getuid()` sounds ok. Let's do a follow up though as this is ready to go anyway now 💪 

/lgtm

PR #865, opened 2019-06-18
pkg/server: Add cluster server unit test

sgreene570:
**- What I did**
Added a unit test for creating kube configs from file secrets to `pkg/server`
**- How to verify it**
`go test -v github.com/openshift/machine-config-operator/pkg/server`

Add kubeconfig unit test to pkg/server


pkg/server/server_test.go
```diff
@@ -6,13 +6,16 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
+	b64 "encoding/base64"
```
runcom: Just use base64 directly

PR #866, opened 2019-06-19
Mcd startup cleanup further

cgwalters:
More patches like https://github.com/openshift/machine-config-operator/pull/862


pkg/daemon/update.go
```diff
@@ -810,19 +810,6 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	return nil
 }
 
-// RHEL 7.6 logger (util-linux) doesn't have the --journald flag
-func (dn *Daemon) isLoggingToJournalSupported() bool {
-	if dn.OperatingSystem == machineConfigDaemonOSRHEL {
```
cgwalters: One thing I wanted to call out...this conditional looked wrong to me, so I inverted it above.

runcom: This was actually completely wrong.

runcom: Can you create a separate PR for this?

runcom: it was meant to compare against RHCOS, not OSRHEL

pkg/daemon/daemon.go
```diff
@@ -208,62 +214,55 @@ func New(
...
+		bootID, err = getBootID()
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to read boot ID")
+		}
+	}
+
+	// RHEL 7.6 logger (util-linux) doesn't have the --journald flag
+	loggerSupportsJournal := true
+	if !mock {
+		if operatingSystem == machineConfigDaemonOSRHEL {
```
runcom: can we just have like before:

```
if operatingsystem == RHCOS { return true }
else
check --journald
```

runcom: (where the bug was `s/OSRHEL/RHCOS`)

cgwalters: The code is now in the constructor so we can't `return true`.

pkg/daemon/update.go
cgwalters: Done in https://github.com/openshift/machine-config-operator/pull/867

runcom: Thanks a lot Colin!

cgwalters:
Rebased :surfing_man:

PR #868, opened 2019-06-19
 server: Add /etc/ignition-machine-config-encapsulated.json

cgwalters:
This effectively "inverts" a `MachineConfig` object so that
when it boots, Ignition processes the subset of MachineConfig
that is Ignition, and then we have the other stuff (`osImageURL`,
`kernelArguments` and in the future more) as a file on disk
that the MCD can read.

The way to think about this is we're aiming to have RHCOS boot
into `MachineConfig`, not just Ignition.

### DIFF
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -34,4 +34,10 @@ const (
 	// EtcPivotFile is used by the `pivot` command
 	// For more information, see https://github.com/openshift/pivot/pull/25/commits/c77788a35d7ee4058d1410e89e6c7937bca89f6c#diff-04c6e90faac2675aa89e2176d2eec7d8R44
 	EtcPivotFile = "/etc/pivot/image-pullspec"
+
+	// MachineConfigEncapsulatedPath contains all of the data from a MachineConfig object
+	// except the Spec/Config object; this supports inverting+encapsulating a MachineConfig
+	// object so that Ignition can process it on first boot, and then the MCD can act on
+	// non-Ignition fields such as the osImageURL and kernelArguments.
+	MachineConfigEncapsulatedPath = "/etc/ignition-machine-config-encapsulated.json"
 )
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -7,6 +7,7 @@ import (
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	"github.com/vincent-petithory/dataurl"
 )
@@ -48,6 +49,13 @@ func getAppenders(currMachineConfig string, f kubeconfigFunc, osimageurl string)
 
 // machineConfigToIgnition converts a MachineConfig object into Ignition.
 func machineConfigToIgnition(mccfg *mcfgv1.MachineConfig) *igntypes.Config {
+	tmpcfg := mccfg.DeepCopy()
+	tmpcfg.Spec.Config = ctrlcommon.NewIgnConfig()
+	serialized, err := json.Marshal(tmpcfg)
+	if err != nil {
+		panic(err.Error())
+	}
+	appendFileToIgnition(&mccfg.Spec.Config, daemonconsts.MachineConfigEncapsulatedPath, string(serialized))
 	return &mccfg.Spec.Config
 }
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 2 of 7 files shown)


cgwalters:
Part of https://github.com/openshift/machine-config-operator/issues/798

pkg/daemon/constants/constants.go
```diff
@@ -34,4 +34,11 @@ const (
 	// EtcPivotFile is used by the `pivot` command
 	// For more information, see https://github.com/openshift/pivot/pull/25/commits/c77788a35d7ee4058d1410e89e6c7937bca89f6c#diff-04c6e90faac2675aa89e2176d2eec7d8R44
 	EtcPivotFile = "/etc/pivot/image-pullspec"
+
+	// MachineConfigEncapsulatedPath contains all of the data from a MachineConfig object
+	// except the Spec/Config object; this supports inverting+encapsulating a MachineConfig
+	// object so that Ignition can process it on first boot, and then the MCD can act on
+	// non-Ignition fields such as the osImageURL and kernelArguments.  We don't want to
+	// write this file out to disk permanently; so it lives in /run.
```
cgwalters: Hm the `/run` thing is a lie, just like cake; will fix.

cgwalters:
This does nothing at the moment until the MCD learns to read it, which is going to happen in another PR after more prep work lands from https://github.com/openshift/machine-config-operator/pull/866 etc - but we have unit coverage and I think this is looking good.

kikisdeliveryservice:
I'm a little confused about the mention of "inverting" it?? Would this eventually change our mcs?

From what I can tell here you are taking a machine config, extracting the IGN config part from spec and then adding the non-spec extras like osimageurl into the ignition config? <- is this correct?

cgwalters:
Currently MachineConfig contains Ignition.  With this, the generated Ignition contains the rest of the MachineConfig.  "invert" seems right for that, no?

> From what I can tell here you are taking a machine config, extracting the IGN config part from spec and then adding the non-spec extras like osimageurl into the ignition config? <- is this correct?

Yep!

> I dont fully understand what the final flow will look like or how it will differ from what we have?

The main rationale is in https://github.com/openshift/machine-config-operator/issues/798
Today we don't have a good way for `pivot` which runs on the first boot to get access to the `kernelArguments` in the MachineConfig.  This fixes that by serving the whole MachineConfig, and then later PRs will replace `pivot` with `machine-config-daemon` which will soon be built into RHCOS (*in addition* to being run as a container).

cgwalters:
>  With this, the generated Ignition contains the rest of the MachineConfig. "invert" seems right for that, no?

Though to be very clear this is just a transient implementation detail for the first boot, invisible to users and most of the code.  This inversion is a lot like what we are doing with the node annotations today, just an extension to provide the full MC and not just its hash.

kikisdeliveryservice:
> > With this, the generated Ignition contains the rest of the MachineConfig. "invert" seems right for that, no?
> 
> Though to be very clear this is just a transient implementation detail for the first boot, invisible to users and most of the code. This inversion is a lot like what we are doing with the node annotations today, just an extension to provide the full MC and not just its hash.

AHA!!!!!!!! Ok gotcha!

cgwalters:
Another way to think of this is that in theory we could do something like have `machine-config-daemon` actually *wrap* Ignition on the firstboot; the MCS would literally serve MachineConfig YAML or whatever and the MCD would unwrap the Ignition part and pass it to Ignition etc.

That would be...a lot more invasive of a change though.

runcom:
Awesome, I'll approve and leave it for lgtm because of the comment above /run

/approve

runcom:
Awesome, I'll approve and leave it for lgtm because of the comment above /run

/approve

runcom:
Awesome, I'll approve and leave it for lgtm because of the comment above /run

/approve
/retest

runcom:
Waiting for https://github.com/openshift/machine-config-operator/pull/868#discussion_r295509063, otherwise it looks great

/approve
/retest

runcom:
Waiting for https://github.com/openshift/machine-config-operator/pull/868#discussion_r295509063 but looks great anyway

cgwalters:
Rebased :surfing_man: and dropped the comment about `/run`.

cgwalters:
Interesting, failure to bootstrap in both cases.  I haven't tested that case locally yet, just did a `make deploy-server`.  Looking at the logs gathered by the installer, the bootstrap MCS seems fine...but I'll see if I can reproduce this locally.

kikisdeliveryservice:
Not sure if relevant, but in the kublet.log I see:
```Jun 20 13:09:29 ip-10-0-11-17 hyperkube[1214]: E0620 13:09:29.110280    1214 container_manager_linux.go:476] failed to find cgroups of kubelet - cpu and memory cgroup hierarchy not unified.  cpu: /, memory: /system.slice/kubelet.service```

File: pkg/server/server.go [hunk 2]
```diff
@@ -45,6 +46,19 @@ func getAppenders(currMachineConfig string, f kubeconfigFunc, osimageurl string)
 	return appenders
 }
 
+// machineConfigToIgnition converts a MachineConfig object into Ignition.
+func machineConfigToIgnition(mccfg *mcfgv1.MachineConfig) *igntypes.Config {
+	tmpcfg := mccfg.DeepCopy()
+	spec := &tmpcfg.Spec.Config
+	tmpcfg.Spec.Config = igntypes.Config{}
```
kikisdeliveryservice: Could we use NewIgnConfig() instead (from controller/common/helpers.go) for consistency and gen a valid ign confg base?

cgwalters: Sure, done

runcom:
This is green now :)

cgwalters:
> This is green now :)

Yeah, I broke the main case when refactoring to make the unit tests happy.  Both are good now!

PR #870, opened 2019-06-20
Pivot cleanups

cgwalters:


cmd/machine-config-daemon/pivot.go
```diff
@@ -193,25 +194,29 @@ func podmanRemove(cid string) {
 }
 
 // getDefaultDeployment uses rpm-ostree status --json to get the current deployment
-func getDefaultDeployment() types.RpmOstreeDeployment {
+func getDefaultDeployment() (types.RpmOstreeDeployment, error) {
 	// use --status for now, we can switch to D-Bus if we need more info
 	var rosState types.RpmOstreeState
+	var errRet types.RpmOstreeDeployment
```
runcom: oh, usually you return `types.RpmOstreeDeployment{}` in the return statements in go. Actually, this might as well be a pointer so we can return `nil, err`

cmd/machine-config-daemon/pivot.go
```diff
@@ -193,25 +194,29 @@ func podmanRemove(cid string) {
...
+	return rosState.Deployments[0], nil
 }
 
 // pullAndRebase potentially rebases system if not already rebased.
-func pullAndRebase(container string) (imgid string, changed bool) {
-	defaultDeployment := getDefaultDeployment()
+func pullAndRebase(container string) (imgid string, changed bool, err error) {
+	defaultDeployment, err := getDefaultDeployment()
+	if err != nil {
+		return
```
runcom: do we want to warn since the callee isn't logging?

cmd/machine-config-daemon/pivot.go
```diff
@@ -228,17 +233,18 @@ func pullAndRebase(container string) (imgid string, changed bool) {
 
 	// If we're passed a non-canonical image, resolve it to its sha256 now
 	isCanonicalForm := true
-	if _, err := daemon.GetRefDigest(container); err != nil {
+	if _, err = daemon.GetRefDigest(container); err != nil {
```
runcom: leave `:`as it's scoped to this if (w/o changing below as well)

cmd/machine-config-daemon/pivot.go
```diff
@@ -291,10 +297,12 @@ func pullAndRebase(container string) (imgid string, changed bool) {
 			glog.Infof("Using ref %s", refs[0])
 			ostreeCsum = utils.RunGetOut("ostree", "rev-parse", "--repo", repo, refs[0])
 		} else if len(refs) > 1 {
-			glog.Fatalf("Multiple refs found in repo!")
+			err = errors.New("multiple refs found in repo")
+			return
 		} else {
 			// XXX: in the future, possibly scan the repo to find a unique .commit object
-			glog.Fatalf("No refs found in repo!")
+			err = errors.New("No refs found in repo")
```
runcom: why are we setting err here and above?

cmd/machine-config-daemon/pivot.go
```diff
@@ -326,19 +333,22 @@ func Execute(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		return err
+	}
 
 	// Delete the file now that we successfully rebased
 	if fromFile {
 		if err := os.Remove(etcPivotFile); err != nil {
 			if !os.IsNotExist(err) {
-				glog.Fatalf("Failed to delete %s: %v", etcPivotFile, err)
+				return errors.Wrapf(err, "Failed to delete %s", etcPivotFile)
```
runcom: lowercase Failed (not blocking you didn't make this wrong here anyway)

cmd/machine-config-daemon/pivot.go
cgwalters: The whole point of this is that the toplevel callee is catching the error right?  `pullAndRebase` :arrow_right: `run` :arrow_right: `Execute` which logs.

cmd/machine-config-daemon/pivot.go
cgwalters: Because we're using the named return value.

cmd/machine-config-daemon/pivot.go
cgwalters: I don't think that works because we need to set the named return value, right?

cgwalters:
Fixed a few things, added another commit too.

cgwalters:
/retest
e2e flakes

cgwalters:
Tide?  Helloooo tide?

cgwalters:
Here's a cookie :cookie: ?

kikisdeliveryservice:
> Tide? Helloooo tide?

:laughing:

cgwalters:
OK, I think I understand; https://github.com/openshift/machine-config-operator/pull/872 was landing and since they didn't get batched, this one needs to be retested on top of it.

cgwalters:
Hooray, merged, so now on to https://github.com/openshift/machine-config-operator/pull/882

PR #872, opened 2019-06-20
daemon: Use pivot/machine-config-daemon-host.service as needed

cgwalters:
RHCOS will soon drop `pivot` entirely and I only tested the
bootimage pivot, but as soon as that lands in `machine-os-content`
we need this fix too.

Also this reveals the `-initial` nomenclature in the service
is wrong; I was originally thinking this was just for the first
boot.  But I'd forgotten about the whole saga around why we always
execute on the host because of SELinux.  I fleshed out some of
the comments there.

Rename the systemd unit to make that clearer too.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -106,7 +106,11 @@ func (r *RpmOstreeClient) GetBootedOSImageURL() (string, string, error) {
 }
 
 // RunPivot executes a pivot from one deployment to another as found in the referenced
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
@cgwalters to be clear - this can't merge until the RHCOS changes goes in right?

cgwalters:
> @cgwalters to be clear - this can't merge until the RHCOS changes goes in right?

Should be fine to merge now - the code is trying to do "use pivot if it exists", so we work in both cases.

PR #873, opened 2019-06-20
Disable Go Unit Test Cache in Go 1.12

sgreene570:
**- What I did**
In https://github.com/openshift/machine-config-operator/pull/818, the flag to disable `GOCACHE` was removed for go 1.12 compatibility.  Turns out, running unit tests with the cache enabled is frustrating.  For example, if a test asset (and not a test file itself) is updated, re-running `make test-unit` will pull test results from the test cache. Also, changing some parts of a unit test may not actually clear the entire test cache for that particular test (even though it should).

The _new_ way to disable the go testing cache is to use `-count=1` in the `go test` command.
(see https://github.com/golang/go/issues/24573)

**- How to verify it**
`make test-unit` and verify that no test were pulled from the cache (ie. look for `(cached)`)

Disable Go Cache for unit tests in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ test: test-unit test-e2e
 
 # Unit tests only (no active cluster required)
 test-unit:
-	go test -v ./cmd/... ./pkg/... ./lib/...
+	go test -count=1 -v ./cmd/... ./pkg/... ./lib/...
 
 # Run the code generation tasks.
 # Example:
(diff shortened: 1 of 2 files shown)


cgwalters:
/approve
Thanks for this.  As [this says](https://twitter.com/codinghorror/status/506010907021828096), 
> There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.
And this is incorrect caching, the go test framework's caching would need to know external inputs.

Will give this a bit for someone else to lgtm.

kikisdeliveryservice:
@sgreene570 and I discovered this yesterday while debugging a unit test, so A+ on this PR :smile: 

/lgtm

runcom:
Great catch Steve!

PR #875, opened 2019-06-21
Mcd sync cleanups

cgwalters:


pkg/daemon/daemon.go
```diff
@@ -363,18 +341,16 @@ func (dn *Daemon) syncNode(key string) error {
...
-	}
-
 	_, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		return err
 	}
+	// If this isn't our node, nothing to do.  The node controller
+	// handles other nodes.
+	if name != dn.name {
+		return nil
```
ericavonb: we're still going to get a bunch of "Started syncing node <other-node>.. \nFinished syncing node <other-node>..." in the logs. Unless we're filtering earlier and this would be unexpected in which case we should log something about it.

pkg/daemon/daemon.go
```diff
@@ -383,32 +359,44 @@ func (dn *Daemon) syncNode(key string) error {
...
+	// Check for Deleted Node
+	if node.DeletionTimestamp != nil {
+		glog.Infof("Node %s was deleted!", node.Name)
+		return nil
+	}
 
 	// Deep-copy otherwise we are mutating our cache.
 	node = node.DeepCopy()
+	// Update our cached copy of the node
+	dn.node = node
```
ericavonb: Do we need a cached copy? Passing in the node object to sync functions directly would make it much easier to follow and test state transitions. From what I can tell, the uses are:
* `completeUpdate` uses the node object; it doesn't need anything but `node.Name` and `node.Spec.Unschedulable` (which if stale is fine: https://github.com/openshift/kubernetes-drain/blob/master/drain.go#L567).
* event recorder noderef
* getting the current and desired config name via annotations. would be much clearer to get the values for `currentConfigName` and `desiredConfigName` from the annotations once and pass them through to the helper sync functions.

cgwalters: Yeah, I think you're right that would be a good cleanup.  I can do that on top of this?

pkg/daemon/daemon.go
cgwalters: These aren't printed at our default log level.  I don't think I've ever turned it up beyond the non-default myself honestly, I think there's...a *lot* of spam.

pkg/daemon/daemon.go
ericavonb: always 😄 . please don't take my comments as blockers (yet), I'm still learning a lot myself!

cgwalters: I started on this but it's not a trivial patch and would also conflict with some other local branches in my MCD-on-host work.  OK to circle back to a separate PR?

PR #876, opened 2019-06-21
Bug 1722894: pkg/operator: Add FailedToSync Reason

LorbusChris:
Setting reason to error message can cause Prometheus to report too many series and can potentially bring it down completely.
This patch adds a reason constant to fix the above.

This will be followed up with the addition of more reason constants.


runcom:
4.1.z BZ is here https://bugzilla.redhat.com/show_bug.cgi?id=1722887 - Christian's going to reference it in the backport PR as well.

PR #881, opened 2019-06-21
[WIP] [templates/openstack] Dnshacks

iamemilio:
**- What I did** Watches for machines in kubernetes api and creates an A record for them in coredns

**- How to verify it**


cgwalters:
Similar comment as here https://github.com/openshift/machine-config-operator/pull/795#issuecomment-504567632

templates/master/00-master/openstack/files/openstack-mdns-publisher.yaml
```diff
@@ -0,0 +1,90 @@
...
+          path: "/etc/kubernetes/static-pod-resources/mdns"
+      - name: kubeconfig
+        hostPath:
+          path: "/etc/kubernetes/kubeconfig"
+      - name: conf-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/mdns"
+      initContainers:
+      - name: clusterrc-generation
+        image: quay.io/openshift-metalkube/kubeconfig-extractor:latest
```
cgwalters: OK so I was wrong I see we are pulling in other containers here, so that's good.  However they will also need to be added to the release image.  See [the image-references file](https://github.com/openshift/machine-config-operator/blob/master/install/image-references).

PR #882, opened 2019-06-21
More pivot/MCD cleanups/prep work

cgwalters:


cgwalters:
> 2019/06/21 20:28:40 Container test in pod unit failed, exit code 2, reason OOMKilled

Whee, transient OOM.

/retest

cmd/machine-config-daemon/pivot.go
```diff
@@ -366,10 +364,7 @@ func run(_ *cobra.Command, args []string) error {
 	}
 
 	if !changed {
-		glog.Info("Already at target pivot; exiting...")
-		if exit77 {
-			os.Exit(77)
-		}
+		glog.Info("Already at target oscontainer")
 	} else if reboot || utils.FileExists(runPivotRebootFile) {
```
runcom: debugging the very latest weirdness I've witnessed with pivot, it would be nice to log if we rebooted cause of the flag or the file

pkg/daemon/pivot/utils/run_test.go
```diff
@@ -6,21 +6,22 @@ import (
 	"time"
 	"io/ioutil"
 
+    "github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/util/wait"
```
runcom: nit, off indent?

runcom:
/approve

looked at commits one by one 👍 will leave to another review to lgtm!

cgwalters:
Rebased :surfing_man: and addressed both those issues!

cmd/machine-config-daemon/pivot.go
```diff
@@ -365,14 +226,20 @@ func run(_ *cobra.Command, args []string) error {
 		changed = true
 	}
 
+	runPivotRebootExists := utils.FileExists(runPivotRebootFile)
```
runcom: let's not use FileExists here either. Also, the impl for FileExists isn't correct, it returns `true` if for some reason, other errors occur causing pivot to reboot in the middle of an MCD sync (which is what I believe it's happening in an issue I'm looking at)

cgwalters: small bikeshed: do we have a `FileExists` that ignores errors (and returns `false`) or "throws" (in golang terms)?  I tend to lean towards the latter, but was actually surprised that the "usually so correct it's painful" [Rust API for this](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists) just swallows errors.

cgwalters:
Updated to drop the `pivotutils.FileExists` bit.

runcom:
timeouts (?)

/retest

cgwalters:
```
Jun 25 18:22:41 ip-10-0-135-155 machine-config-daemon[1133]: F0625 18:22:41.240630    1133 main.go:28] Error executing mcd: unknown flag: --from-etc-pullspec
```

Right.  Need to get an updated build through the sausage factory.

cgwalters:
Hmm, can't easily do https://github.com/openshift/machine-config-operator/pull/882/commits/53c6d6855d5e5d8d0af26e4a7e9d8ba8fc28af56 without ratcheting I think...

cgwalters:
> Hmm, can't easily do 53c6d68 without ratcheting I think...

OK, made that one optional for now and will circle back and require it later.

cgwalters:
And we're green again!

runcom:
approved already, let's have another round of review to get this in!

cmd/machine-config-daemon/pivot.go
```diff
@@ -365,14 +227,24 @@ func run(_ *cobra.Command, args []string) error {
 		changed = true
 	}
 
+	_, err = os.Stat(runPivotRebootFile)
```
ericavonb: Is there a reason to check this if `!changed`?

ericavonb: Or if `reboot` is set?

cmd/machine-config-daemon/pivot.go
```diff
@@ -365,14 +227,24 @@ func run(_ *cobra.Command, args []string) error {
...
+	if err != nil && !os.IsNotExist(err) {
+		return errors.Wrapf(err, "Checking %s", runPivotRebootFile)
+	}
+	runPivotRebootExists := err == nil
 	if !changed {
-		glog.Info("Already at target pivot; exiting...")
-		if exit77 {
-			os.Exit(77)
+		glog.Info("Already at target oscontainer")
+	} else if reboot || runPivotRebootExists {
```
ericavonb: nit: return in if statement above and get rid of the `else`'s

cmd/machine-config-daemon/pivot.go
```diff
@@ -381,6 +253,7 @@ func run(_ *cobra.Command, args []string) error {
 func Execute(cmd *cobra.Command, args []string) {
 	err := run(cmd, args)
 	if err != nil {
-		glog.Fatalf("%v", err)
+		fmt.Printf("error: %v\n", err)
+		os.Exit(1)
```
ericavonb: Any reason for this (other than that we should move from glog to klog 🙂 )?

cmd/machine-config-daemon/pivot.go
cgwalters: Hm you're right, I regressed this compared to https://github.com/openshift/pivot/blob/63d3093d32f25649241cc09ae5fdc35181ead73e/cmd/root.go#L409

cmd/machine-config-daemon/pivot.go
cgwalters: `Fatalf` dumped a stack trace which wasn't necessary here.

cgwalters: This was noted in the commit message from 0bc206007612159988721f39f47dd8f12d3ffb25
>  (Also change the toplevel to not spew a stack trace on error)

cgwalters:
Added a :new: commit.

cmd/machine-config-daemon/pivot.go
ericavonb: How about when `reboot` is set by cmdline flag?

kikisdeliveryservice:
this looks good will let @ericavonb review and LGTM.

cmd/machine-config-daemon/pivot.go
```diff
@@ -366,13 +228,26 @@ func run(_ *cobra.Command, args []string) error {
...
-		glog.Info("Already at target pivot; exiting...")
-		if exit77 {
-			os.Exit(77)
-		}
-	} else if reboot || utils.FileExists(runPivotRebootFile) {
+		glog.Info("No changes; already at target oscontainer, no kernel args provided")
+		return nil
+	}
+
+	_, err = os.Stat(runPivotRebootFile)
```
ericavonb: Same question about if this should be checked when `reboot` is set via cmd line flag.

cgwalters: OK, right, I thought I'd changed that but I hadn't.  Reworked it and updated that last commit again.

At a higher procedural level though this isn't net new code, we imported it from elsewhere and actually I am working on replacing it and have another PR ready to go after this one lands.  This code will mostly be dead after that in favor of reading the new encapsulated config.

Also I personally hesitate a bit around nits like this since the previous version had all passed, every push needs 3 new clusters spun up and an hour of validation...i.e. there's a velocity vs code review tradeoff.  On the other hand it is good to come to a shared team understanding of code style for future changes!

So anyways, updated :new: again!

ericavonb: I tried to restrict my review to new code; I'm sorry if I missed something and re-reviewed lines. My concern here was introducing a new failure point at a place where you could be trying to force through a reboot due to some issue. That seemed like a potentially problematic state to get into.

Understood on the CI tradeoff, please keep pushing back as I come up to speed 👍

cgwalters:
Hmm, looks like a flake...maybe some e.g. EC2 networking blip.
/retest

runcom:
> /lgtm

oooh why @ericavonb lgtm didn't work? 🤔 

@ericavonb can you try it again cause sometimes it just doesn't go through?

kikisdeliveryservice:
sometimes yelling works:

/LGTM

PR #883, opened 2019-06-24
WIP: pkg/daemon: inject proxies from cluster

runcom:
The MCD runs an external service (pivot) which has to understand if the
cluster has proxies configured and use them accordingly.
This patch grabs the proxy config from the cluster and inject it if
available. This patch also just sets the env for the service, but skopeo/pivot/containers/image automatically read those when set and use it (uses the `http.ProxyFromEnv` func from Golang).

related to https://jira.coreos.com/browse/MCO-55


runcom:
/hold

Mainly to keep discussing in the related card

runcom:
I guess I can see now what @cgwalters commented in the card - we might just go and support proxies through MachineConfig(s) for the whole OS (though, that requires a KubeletConfig-like controller to keep it in sync with cluster changes right?)

pkg/daemon/update.go
```diff
@@ -810,6 +815,22 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		// no proxy configured
+		return os.Remove(proxyPath)
+	}
+ 	proxyEnvs := fmt.Sprintf(`HTTP_PROXY=%s
+HTTPS_PROXY=%s
+NO_PROXY=%s
+`, proxyCfg.Status.HTTPProxy, proxyCfg.Status.HTTPSProxy, proxyCfg.Status.NoProxy)
```
ericavonb: How do we distinguish between user has set the proxy resource spec but whatever operator has not reconciled and set the status yet, and empty values for each? At the least, there could be an unnecessary flip-flop in this file

runcom: yeah, this isn't hashed out though, just a preliminary test - we might want to go https://github.com/openshift/machine-config-operator/pull/883#issuecomment-504977641 anyway

runcom: > How do we distinguish between user has set the proxy resource spec but whatever operator has not reconciled and set the status yet

This could be a reason for sycing proxies specifically (like we do for ImageConfig and ContainerRuntimeConfig). On the other end, since this rerun on failure (assuming your connections have to go through a proxy) we can fail this and once Status is again populated, this will go through. But I'm starting to prefer the first option. Emtpy values go through systemd env file and process as always (unset?)

templates/common/_base/units/machine-config-daemon-host.service
```diff
@@ -3,6 +3,7 @@ enabled: true
 contents: |
   [Unit]
   Description=Machine Config Daemon Initial
+  EnvironmentFile=/etc/machine-config-daemon/proxy
```
cgwalters: What if we just inject them everywhere?
https://github.com/coreos/bugs/issues/1507#issuecomment-240233869

This way the env vars would be picked up automatically by e.g. a user logging in via `ssh` as well as e.g. `cri-o.service`.  But I don't think they'd be picked up by `oc debug node/`...whole other can of worms there.

cgwalters:
A few novel things going on here.  First, each MCD here is directly watching a cluster resource and applying changes "live" without...wait, no we do reboot because this is part of the main update sync func?

Or actually is this missing the bit that adds the node into the MCD change queue?

I had thought this would be a new "sub-controller" like the kubelet/containerruntime ones that generates a small MachineConfig fragment.

But this approach is actually a new way for us to approach "rebootless" updates - bypass MachineConfig entirely and have the MCD apply live.

runcom:
> I had thought this would be a new "sub-controller" like the kubelet/containerruntime ones that generates a small MachineConfig fragment.
> 
> But this approach is actually a new way for us to approach "rebootless" updates - bypass MachineConfig entirely and have the MCD apply live.

so I created this and then thought about the sub-controller as well so I'm now torn between the two since this looks much simpler.

On the other hand, if this is cluster-wide and we want to go the "MCO syncs proxies on OS as well" - then yeah we definitely need a subcontroller. This is not clear enough to me, we might want to run our questions through others to understand more.

> A few novel things going on here. First, each MCD here is directly watching a cluster resource and applying changes "live" without...wait, no we do reboot because this is part of the main update sync func?

My first impression to the proxy enablement was that we only cared about this "on-the-fly", meaning, we use tools that talks to external services and we need this enablement. This is/was not about configuring proxies cluster-wise on the OS itself, that can come later afaict.

templates/common/_base/units/machine-config-daemon-host.service
runcom: Coming from Slack conversations, this enablement is ad-hoc for our services that use external services needing a proxy

pkg/daemon/update.go
runcom: note to self, we need to check `status != nil` as per slack conversations

templates/common/_base/units/machine-config-daemon-host.service
cgwalters: The scope of the MCO is everything inclusive of kernel to kubelet right?  The way you say "our services" seems to argue for just handling one of our core functions, but I would say we do want the same thing to work if a user runs `podman` directly at least.

Actually it's not clear to me...the MCO is also going to take care of this for cri-o/kubelet right?

runcom:
Closing this as we're really going through the Proxy controller way

PR #885, opened 2019-06-24
Take over `/usr/bin/pivot` too

cgwalters:
Since it exists in the docs, and there's already a nontrivial amount
of "muscle memory" around it.

### DIFF
--- a/machine-config-daemon.spec
+++ b/machine-config-daemon.spec
@@ -28,8 +28,10 @@ env VERSION_OVERRIDE=%{version} SOURCE_GIT_COMMIT=%{commit} make daemon
 
 %install
 install -D -m 0755 _output/linux/*/%{name} $RPM_BUILD_ROOT/usr/libexec/%{name}
+install -D -m 0755 cmd/machine-config-daemon/pivot.sh $RPM_BUILD_ROOT/%{_bindir}/pivot
 
 %files
 %license LICENSE
 %doc README.md
 %{_libexecdir}/%{name}
+%{_bindir}/pivot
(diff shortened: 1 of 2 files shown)


machine-config-daemon.spec
```diff
@@ -28,8 +28,10 @@ env VERSION_OVERRIDE=%{version} SOURCE_GIT_COMMIT=%{commit} make daemon
 
 %install
 install -D -m 0755 _output/linux/*/%{name} $RPM_BUILD_ROOT/usr/libexec/%{name}
+install -D -m 0755 cmd/machine-config-daemon/pivot.sh $RPM_BUILD_ROOT/usr/bin/pivot
```
ashcrow: nit: `$RPM_BUILD_ROOT/%{_bindir}/pivot`

machine-config-daemon.spec
```diff
@@ -28,8 +28,10 @@ env VERSION_OVERRIDE=%{version} SOURCE_GIT_COMMIT=%{commit} make daemon
 
 %install
 install -D -m 0755 _output/linux/*/%{name} $RPM_BUILD_ROOT/usr/libexec/%{name}
+install -D -m 0755 cmd/machine-config-daemon/pivot.sh $RPM_BUILD_ROOT/%{_bindir}/pivot
 
 %files
 %license LICENSE
 %doc README.md
 %{_libexecdir}/%{name}
+%{_bindir}/pivot
```
runcom: isn't this conflicting with https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/rpm-ostree.go#L132

runcom: Uhm, this slipped through, @cgwalters can you take a super fast look?

cgwalters: You're totally right but because [we aren't auto-building the RPM yet](https://jira.coreos.com/browse/ART-640) we have time to do a PR to drop the compat code which I think we can do now that [a RHCOS bootimage update](https://github.com/openshift/installer/pull/1895) landed.

Issue #878, opened 2019-06-21
Simplify controller architecture, merge "operator" and "controller"


ericavonb:
The operator controller for deploying the core controller is an extraneous layer. The "controller" _is_ the operator, it handles the operational logic specific to this operator (the machine configuration). Other operators exist to handle deploying the operators themselves (CVO in this case,.

cgwalters:
I agree with this - although with this type of thing I have a temptation to ping Abhinav to understand more of his reasoning, but OTOH he's pretty busy with things like reviewing a gigantic PR to add baremetal to the installer, etc.

Probably makes the most sense to approach this in incremental phases; the fact that we have just one image now is a big one!  Next would be to run the operator and controller in the same pod probably.

After that we can start to take advantage of that and drop e.g. the `controllerconfig` CRD which I think is only used as a way to sync from the operator :arrow_right: controller...we could just use a yaml file in the pod shared memory or whatever.

And then perhaps after that we could link the controller's code into the operator process.

abhinavdahiya:
I disagree with operator is the controller.

The controller + server + daemon is the ecosystem that realizes the machineconfiguration API, provides a way to manage configuration of machines in the cluster.

The operator is the one that knows information about how to configure them and update them through versions.

For example, 
configpool is transparent to controller, it updates the nodes with new configuration with max unavail set. The operator knows a configpool is for control plane so it should be set to something safe.

Or that when upgrading it's required that control plane upgrades while the compute ones can trail..

I think the management and control-loop layer separation is useful.

PR #886, opened 2019-06-24
[release-4.1] Bug 1723327: pkg/operator: sync deployment before controllerconfig

runcom:
and also rename the CM so the old version can't use the new osimageurl coming from an upgrade.

This should take care of clusters at 4.1.0 upgrading to 4.1.x+?

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -122,6 +122,9 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 	defer close(journalStopCh)
 	go followPivotJournalLogs(journalStopCh)
 
+	if err := os.Remove("/etc/pivot/reboot-needed"); err != nil {
+		return errors.Wrap(err, "deleting pivot reboot-needed file")
+	}
 	err := exec.Command("systemctl", "start", "pivot.service").Run()
 	if err != nil {
 		return errors.Wrapf(err, "failed to start pivot.service")
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -395,9 +395,9 @@ func (optr *Operator) sync(key string) error {
 	// any error marks sync as failure but continues to next syncFunc
 	var syncFuncs = []syncFunc{
 		{"pools", optr.syncMachineConfigPools},
+		{"mcd", optr.syncMachineConfigDaemon},
 		{"mcc", optr.syncMachineConfigController},
 		{"mcs", optr.syncMachineConfigServer},
-		{"mcd", optr.syncMachineConfigDaemon},
 		{"required-pools", optr.syncRequiredMachineConfigPools},
 	}
 	return optr.syncAll(rc, syncFuncs)
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -143,16 +143,6 @@ func (optr *Operator) syncMachineConfigController(config renderConfig) error {
 		return err
 	}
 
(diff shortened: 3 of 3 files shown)


cgwalters:
This would be another argument to do https://github.com/openshift/machine-config-operator/issues/878

cgwalters:
Ah nevermind, renaming the URL fixes the controller sync problem *this* time but doesn't solve the general case.

runcom:
> Actually, I don't think this fixes the race, it just narrows it as we discussed because the operator's rollout doesn't block on it rolling out the controller, does it?

uhm, the new MCO is guaranteed to be up and running for the new MCC to roll out (MCO starts the MCC)

runcom:
The new payload would be rolled out as follows (aiui):

- Old MCO running
- Old MCC running
- Old CM osimageurl
- New MCO code kicks in
- Old MCC running
- Old CM osimageurl
- New MCC kicked in through new MCO and wait for roll out
- New CM osimageurl applied and picked by now-new MCC above

cgwalters:
> With this patch, when the new MCO runs, it'll wait for the new MCC to roll out before going ahead and creating the new ControllerConfig for the MCC (template controller).

Right.  We discussed this even and I forgot that argument.  I am convinced again :smile: 
/approve

runcom:
> Right. We discussed this even and I forgot that argument. I am convinced again 😄
> /approve

hold on, we still have to fix the reboot-needed file 😄 the MCD is rolled out last so we should still have the new code running right?

runcom:
/hold

actually, the old MCD can still see the new rendered MC from the new MCC and attempt an upgrade afaict

cgwalters:
Wait did I miss something or this is going straight to 4.1?  Why no PR for master first and a cherry pick?

runcom:
> Wait did I miss something or this is going straight to 4.1? Why no PR for master first and a cherry pick?

cause I had the branch prepared for 4.1 will do a master PR very shortly

also, regarding my last comment, can't why just roll out the MCD before the MCC? 🤔

runcom:
> What about if more than one MCO is running? Is there leader election to prevent that?

yes, MCO and MCC use leader election

cgwalters:
> actually, the old MCD can still see the new rendered MC from the new MCC and attempt an upgrade afaict

I don't *think* so...the osimageurl comes from the controllerconfig which is from the operator.
Not unless the CVO applies the resources in parallel...hm, the docs say:

> In install and reconciliation, the CVO runs components in random, parallel fashion and retrying as necessary. 

What's "reconciliation"?  Forcing a failed upgrade?

runcom:
> What's "reconciliation"? Forcing a failed upgrade?

sounds like a generic term for "upgrade" as well I guess

pkg/daemon/rpm-ostree.go
```diff
@@ -122,6 +122,9 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 	defer close(journalStopCh)
 	go followPivotJournalLogs(journalStopCh)
 
+	if err := os.Remove("/etc/pivot/reboot-needed"); err != nil {
+		return errors.Wrap(err, "deleting pivot reboot-needed file")
```
cgwalters: This needs to ignore `ENOENT`.

runcom: 👍 thanks!

cgwalters:
> cause I had the branch prepared for 4.1 will do a master PR very shortly

I'd lean towards doing master and the cherry pick dance so we follow full procedure here.

PR #887, opened 2019-06-24
MCO/MCD: sync deployment before controllerconfig, roll out MCDs before MCC and remove pivot's reboot-needed file

runcom:
This patch set takes care of mainly three issues we've spotted:

- an osimageurl change from an upgrade can race with the old MCC
- we can still be running the old MCDs on nodes once the new MCC creates the new rendered
- if the node has never rebooted and ConditionFirstBoot is thus True, pivot reboots the node causing an upgrade failure

The above have been fixed by:

- rolling out the new MCC before applying the new osimageurl ConfigMap that is then templated by the new MCC
- roll out the MCDs before the MCC
- always remove `/etc/pivot/reboot-needed` when running pivot from the MCD

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -921,6 +921,13 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 				return err
 			}
 		}
+		// if we're degraded here, it means we got an error likely on startup and we retried
+		// if it's the case, clear it out
+		if state.state == constants.MachineConfigDaemonStateDegraded {
+			if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.currentConfig.GetName()); err != nil {
+				return errors.Wrap(err, "error setting node's state to Done")
+			}
+		}
 
 		glog.Infof("In desired config %s", state.currentConfig.GetName())
 
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -124,6 +124,11 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 	defer close(journalStopCh)
 	go followPivotJournalLogs(journalStopCh)
 
+	// This is written by code injected by the MCS, but we always want the MCD to be in control of reboots
+	if err := os.Remove("/run/pivot/reboot-needed"); err != nil && !os.IsNotExist(err) {
+		return errors.Wrap(err, "deleting pivot reboot-needed file")
+	}
+
 	service := "machine-config-daemon-host.service"
 	// We need to use pivot if it's there, because machine-config-daemon-host.service
 	// currently has a ConditionPathExists=!/usr/bin/pivot.  This code can be dropped
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -410,9 +410,9 @@ func (optr *Operator) sync(key string) error {
 	// any error marks sync as failure but continues to next syncFunc
 	var syncFuncs = []syncFunc{
 		{"pools", optr.syncMachineConfigPools},
+		{"mcd", optr.syncMachineConfigDaemon},
 		{"mcc", optr.syncMachineConfigController},
 		{"mcs", optr.syncMachineConfigServer},
-		{"mcd", optr.syncMachineConfigDaemon},
 		{"required-pools", optr.syncRequiredMachineConfigPools},
 	}
 	return optr.syncAll(rc, syncFuncs)
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -143,16 +143,6 @@ func (optr *Operator) syncMachineConfigController(config *renderConfig) error {
 		return err
 	}
 
-	ccBytes, err := renderAsset(config, "manifests/machineconfigcontroller/controllerconfig.yaml")
-	if err != nil {
-		return err
-	}
-	cc := resourceread.ReadControllerConfigV1OrDie(ccBytes)
-	_, _, err = resourceapply.ApplyControllerConfig(optr.client.MachineconfigurationV1(), cc)
-	if err != nil {
-		return err
-	}
-
 	mccBytes, err := renderAsset(config, "manifests/machineconfigcontroller/deployment.yaml")
 	if err != nil {
 		return err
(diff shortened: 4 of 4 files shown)


runcom:
I'm not sure about the second commit `pkg/operator: roll out the MCDs first` but wanted to test what's happening in the CI with that at least

runcom:
This actually went through and we can see in logs that the MCD waited for the MCC as well to comes up and publish the rendered MC on cluster setup. Now, not related to this PR clearly since I've spotted them elsewhere but I need to write them down before forgetting:

```
I0624 21:36:38.224136    5027 update.go:723] logger doesn't support --jounald, grepping the journal
```
The above shouldn't be the case on RHCOS in master, why does it happen, and:
```
I0624 21:36:38.309457    5027 daemon.go:944] In desired config rendered-master-e539c6f166c366576fef29f64a2d8941
I0624 21:36:43.625102    5027 update.go:174] Checking reconcilable for config rendered-master-e539c6f166c366576fef29f64a2d8941 to rendered-master-e539c6f166c366576fef29f64a2d8941
I0624 21:36:43.628216    5027 update.go:836] Starting update from rendered-master-e539c6f166c366576fef29f64a2d8941 to rendered-master-e539c6f166c366576fef29f64a2d8941
```
The above shouldn't really happen, why is it

I'll move the above to separate issues tomorrow

runcom:
this looks sane, though I'll hold cause I need to investigate an additional unexpected sync+reboot which shouldn't happen

kikisdeliveryservice:
> This actually went through and we can see in logs that the MCD waited for the MCC as well to come up and publish the rendered MC on cluster setup.
> 
> Now, not related to this PR clearly since I've spotted it elsewhere but I need to write them down before forgetting:
> 
> ```
> I0624 21:36:38.224136    5027 update.go:723] logger doesn't support --jounald, grepping the journal
> ```
> 
> The above shouldn't be the case on RHCOS in master, why does it happen, and:
> 
> I'll move the above to separate issues tomorrow

From just poking around maybe bc that logging comes from checkStateOnFirstRun() https://github.com/openshift/machine-config-operator/blob/05bdb217416517801f2c62fcc58b763e9f01e1d5/pkg/daemon/daemon.go#L800 the default value (false) of var loggerSupportsJournal is being used instead of the set value of true that is done here: https://github.com/openshift/machine-config-operator/blob/05bdb217416517801f2c62fcc58b763e9f01e1d5/pkg/daemon/daemon.go#L223

From looking through the files that's the only way outside of being a RHEL7 that it would retain the value of false.

runcom:
working as expected now and MCDs roll out before MCC

cgwalters:
All the code looks good to me, and I know you wanted to test this in CI but - I am going to be a pain about the commit message bodies being empty.  Mind doing e.g. `git rebase -i origin/master`, then do `r` on each commit to reword, and add the text you've written in the PR body into some of the commits?  Even just a single line can help later.

I use `git blame` a *lot* to find historical context and having the text in the commit message directly is a lot faster than trying to go back to Github to find the PR text.

Another important thing is...5 years down the line, maybe the GH business model isn't quite working out and they let go one of their database people and then later some of the historical PR discussions get lost...  but with git, it's forever.  (I still wish sometimes we backed up even the PR review discussions reliably but that's another issue)

runcom:
> All the code looks good to me, and I know you wanted to test this in CI but - I am going to be a pain about the commit message bodies being empty. Mind doing e.g. `git rebase -i origin/master`, then do `r` on each commit to reword, and add the text you've written in the PR body into some of the commits? Even just a single line can help later.

no pain, I usually do that anyway but forget from time to time and I agree it's freaking important to have it (I look it up from vscode as well!)

pkg/daemon/rpm-ostree.go
```diff
@@ -124,6 +124,10 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 	defer close(journalStopCh)
 	go followPivotJournalLogs(journalStopCh)
 
+	if err := os.Remove("/run/pivot/reboot-needed"); err != nil && !os.IsNotExist(err) {
```
cgwalters: Also this one could really use a comment like:

`// This is written by code injected by the MCS, but we always want the MCD to be in control of reboots`

runcom:
Took care of @cgwalters comments!

pkg/daemon/daemon.go
```diff
@@ -921,6 +921,13 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 				return err
 			}
 		}
+		// if we're degraded here, it means we got an error likely on startup and we retried
+		// if it's the case, clear it out
+		if state.state == constants.MachineConfigDaemonStateDegraded {
+			if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.currentConfig.GetName()); err != nil {
```
cgwalters: Would it be better to just call `completeUpdate` as the code right above this is doing in case there's somehow still a cordon?  I wonder if we should always call `SetDone` here like this:

```
index 0c8630d2..ff33afa0 100644
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -895,9 +895,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		if dn.recorder != nil {
 			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "NodeDone", fmt.Sprintf("Setting node %s, currentConfig %s to Done", dn.node.Name, state.pendingConfig.GetName()))
 		}
-		if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.pendingConfig.GetName()); err != nil {
-			return errors.Wrap(err, "error setting node's state to Done")
-		}
 		if out, err := dn.storePendingState(state.pendingConfig, 0); err != nil {
 			return errors.Wrapf(err, "failed to reset pending config: %s", string(out))
 		}
@@ -922,6 +919,11 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 			}
 		}
 
+		if state.state != constants.MachineConfigDaemonStateDone {
+			if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.pendingConfig.GetName()); err != nil {
+				return errors.Wrap(err, "error setting node's state to Done")
+			}
+		}
 		glog.Infof("In desired config %s", state.currentConfig.GetName())
 
 		// All good!
```

?

cgwalters: Eh...it's OK we can do this type of detail in a later followup.

runcom:
I knew it -.-

runcom:
working on a patch for 4.1

PR #889, opened 2019-06-25
pkg/daemon: support FIPS

runcom:

**- What I did**

resurrected from https://github.com/openshift/machine-config-operator/pull/826
close https://github.com/openshift/machine-config-operator/pull/800

Added a new FIPS field to MachineConfig to enable/disable FIPS on rhcos (disabled by default). This is, at this point, only intended as a day-2 operation.

FIPS would also greatly benefit a dedicated CRD (oc edit fips?) - I'm experimenting with that but it will ofc require a new controller.

**- How to verify it**

added an e2e and manually verified

Added FIPS support

### DIFF
--- a/docs/MachineConfiguration.md
+++ b/docs/MachineConfiguration.md
@@ -51,6 +51,7 @@ type MachineConfigSpec struct {
     // Config is a Ignition Config object.
     Config ign.Config `json:"config"`
     KernelArguments []string `json:"kernelArguments"`
+    Fips bool `json:"fips"`
 }
 ```
 
@@ -93,6 +94,10 @@ Ignition config keys as well.
 ### KernelArguments
 
 This extends the host's kernel arguments.  Use this for e.g. [nosmt](https://access.redhat.com/solutions/rhel-smt).
+
+### FIPS
+
+This allows to enable/disable [FIPS mode](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations). If any of the configuration has FIPS enabled, it'll be set.
 
 ### OSImageURL
 
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -20,8 +20,13 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	}
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
+	var fips bool
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
+		// if any of the config has FIPS enabled, it'll be set
+		if configs[idx].Spec.Fips {
+			fips = true
+		}
 		outIgn = ign.Append(outIgn, configs[idx].Spec.Config)
 	}
 	kargs := []string{}
@@ -34,6 +39,7 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 			OSImageURL:      osImageURL,
 			KernelArguments: kargs,
 			Config:          outIgn,
+			Fips:            fips,
 		},
 	}
 }
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -37,6 +37,8 @@ const (
 	coreUserName = "core"
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
+	// fipsCommand is the command to use when enabling or disabling FIPS
+	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
 )
 
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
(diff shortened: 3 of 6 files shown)


pkg/daemon/update.go
```diff
@@ -369,6 +400,7 @@ func Reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*MachineConfigDif
 	return &MachineConfigDiff{
 		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
 		kargs:    !reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments),
+		fips:     oldConfig.Spec.Fips != newConfig.Spec.Fips,
```
ashcrow: This will force the reboot so that the change takes effect, correct?

runcom: nope, the sync itself does the reboot the node (and it can't be avoided today as we know). This is just to log what we're changing during a sync.

ashcrow: WFM. Just wanted to ensure a reboot does happen when FIPS changes :+1:

runcom: 👍 the e2e added here also takes care of validating fips has been enabled on nodes so it will fail if we skip a reboot after enabling fips (iiuic)

pkg/daemon/update.go
```diff
@@ -37,6 +37,8 @@ const (
 	coreUserName = "core"
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
+	// fipsCommand is the command to use when enabling or disabling FIPS
+	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
```
cgwalters: Sort of a side comment but I recently was thinking that we probably should have taken the path of *replacing* `/usr/bin/fips-mode-setup`; we can totally do that now via [override layers](https://github.com/projectatomic/rpm-ostree/pull/1830).

docs/MachineConfiguration.md
```diff
@@ -51,6 +51,7 @@ type MachineConfigSpec struct {
     // Config is a Ignition Config object.
     Config ign.Config `json:"config"`
     KernelArguments []string `json:"kernelArguments"`
+    Fips bool `json:"fips"`
```
cgwalters: What about [Erica's comment](https://github.com/openshift/machine-config-operator/pull/800#discussion_r296311266) here?

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -20,8 +20,13 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	}
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
+	var fips bool
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
+		// if any of the config has FIPS enabled, it'll be set
```
cgwalters: Ah OK I see we went with the last approach.  I think that's sane for this indeed.

cgwalters:
LGTM, giving to
/assign ericavonb
since she reviewed an earlier version.

docs/MachineConfiguration.md
```diff
@@ -94,6 +95,10 @@ Ignition config keys as well.
 
 This extends the host's kernel arguments.  Use this for e.g. [nosmt](https://access.redhat.com/solutions/rhel-smt).
 
+### FIPS
+
+This allows to enable/disable [FIPS mode](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations).
```
ericavonb: Can you add a comment here about `if any of the config has FIPS enabled, it'll be set` as well?

pkg/daemon/update.go
```diff
@@ -245,16 +247,45 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+		return nil
+	}
+	arg := "enable"
+	if !desired.Spec.Fips {
+		arg = "disable"
+	}
+	cmd := exec.Command(fipsCommand, arg)
+	dn.logSystem("Running %s %s", fipsCommand, arg)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return errors.Wrapf(err, "enabling FIPS: %s", string(out))
```
ericavonb: Can you include `arg` in the error to distinguish between errors from enabling and disabling FIPS?

test/e2e/mcd_test.go
```diff
@@ -546,3 +546,91 @@ func TestDontDeleteRPMFiles(t *testing.T) {
 		}
 	}
 }
+
+func TestFIPS(t *testing.T) {
```
ericavonb: non-blocking suggestion: it might be nice to have this test structured as a table test with a couple variations of machineconfig sets.

pkg/daemon/update.go
```diff
@@ -245,16 +247,45 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+				return
+			}
+		}
+	}()
+
 	return dn.updateOSAndReboot(newConfig)
 }
 
+func (dn *Daemon) updateFIPS(current, desired *mcfgv1.MachineConfig) error {
+	if current.Spec.Fips == desired.Spec.Fips {
```
ericavonb: Does the current get written to Status anywhere?

runcom: nope

test/e2e/mcd_test.go
runcom: yeah, that would be amazing but test timeout is 75m today and going through various reboots for applying the MC + other tests would exceed that

pkg/daemon/update.go
runcom: oh good catch this is completely wrong indeed

runcom:
Took care of @ericavonb comments

kikisdeliveryservice:
Gah. This is hitting AWS/ECS errors now: `Error: error attaching EC2 Internet Gateway...`

PR #891, opened 2019-06-25
daemon: Look for pivot.service unit file instead of binary

cgwalters:
Since we merged a change to add `/usr/bin/pivot` redirecting
to the MCD, we need to fix this compat code.  Checking for the unit
file gives us the semantics we need here.


cgwalters:
Decided not to drop the compat code yet...while we merged the bootimage update I'm not sure if the `machine-os-content` promotion went through yet, and it was just as easy to update the compat code as delete it.

PR #892, opened 2019-06-25
[release-4.1] Bug 1723327: sync deployment before controllerconfig, roll out MCDs before MCC and remove pivot's reboot-needed file

runcom:
Backport of https://github.com/openshift/machine-config-operator/pull/887

There's a simple reproducer for this but requires "bootimage == target oscontainer" to avoid an early pivot + reboot. Basically, at least on UPI, install 4.1.0 and upgrade to whatever this PR lands to (4.1.4?).

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -940,6 +940,13 @@ func (dn *Daemon) CheckStateOnBoot() error {
 				return err
 			}
 		}
+		// if we're degraded here, it means we got an error likely on startup and we retried
+		// if it's the case, clear it out
+		if state.state == constants.MachineConfigDaemonStateDegraded {
+			if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.currentConfig.GetName()); err != nil {
+				return errors.Wrap(err, "error setting node's state to Done")
+			}
+		}
 
 		glog.Infof("In desired config %s", state.currentConfig.GetName())
 
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -122,6 +122,11 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 	defer close(journalStopCh)
 	go followPivotJournalLogs(journalStopCh)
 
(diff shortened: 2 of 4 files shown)


runcom:
```
{"component":"entrypoint","file":"prow/entrypoint/run.go:163","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","time":"2019-06-25T15:46:08Z"}
```

/retest

eparis:
I don't see a 4.2 bug associated with the master PR nor do I see any unit/integration/e2e tests proving the master (or this) PR is safe. Hopefully either we can get a 4.2 BZ VERIFIED by review time next week or we can get some CI tests to do that job.

runcom:
> I don't see a 4.2 bug associated with the master PR nor do I see any unit/integration/e2e tests proving the master (or this) PR is safe. Hopefully either we can get a 4.2 BZ VERIFIED by review time next week or we can get some CI tests to do that job.

we might have a way to test this (bootimage == oscontainer) - I'll test with that somehow so we have a reproducer and a way to verify this as well.

runcom:
I guess one way to reproduce is to bring up a cluster (even aws), then touch `/run/pivot/reboot-needed` on every nodes, then start an upgrade to a payload that upgrades the os as well and watch it failing (this is the failure case). To verify this, same process but the payload to upgrade to also needs to contain this PR. I'm gonna try this out but I'm unsure on how to grab a payload which contains an os update as well

runcom:
created a 4.2 BZ https://bugzilla.redhat.com/show_bug.cgi?id=1724155 - we can hopefully get that verified soonish and backport this as per Eric's comment

runcom:
ok, was able to verify that we rm -rf that file before starting a pivot with latest master (collecting steps for QE as well) - I also need a valid oscontainer to upgrade/downgrade to to fully verify this.
Still looking for a failing case to triple check this one.

runcom:
ok, provided steps to verify in the 4.1 and 4.2 BZ - also provided steps to workaround this issue in 4.1 clusters

runcom:
@eparis the 4.2 BZ has been VERIFIED  and I've provided steps to reproduce for 4.1 as well when we get this in.

PR #893, opened 2019-06-25
pkg/controller: Shorten error Reasons by moving them to Messages

### LINKED ISSUE #890: Reason should be one word not full message
There are several occurrences where we're setting the Reason field to a long message rather than a one-word simple message. This happens today with the ControllerConfig and MachineConfigPool. It hasn't caused issues like https://github.com/openshift/machine-config-operator/pull/879 but we need to basically move what we're now setting as Reason to Message (and either leave Reason empty cause we don't really need it, we have the Type field).

Examples:

Must be flipped and Reason set to `""`

https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/status.go#L76

Flip as well and Reason empty https://github.com/openshift/machine-config-operator/blob/0d843aa29c6a5ce5aedcd1cb1638d2383ea17e41/pkg/controller/template/status.go#L21

There are other occurrences of course, the above are just examples from code

sgreene570:
Closes #890.

**- What I did**
Found all instances of `MachineConfigPoolCondition` and `NewControllerConfigStatusCondition` in the repo and swapped the (left) Reason field with the (right) Message field, as described in #890.
**- How to verify it**
Unit testing / CI success

pkg/controller: Shorten error Reasons by moving them to Messages

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -73,7 +73,7 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	if allUpdated {
 		//TODO: update api to only have one condition regarding status of update.
 		updatedMsg := fmt.Sprintf("All nodes are updated with %s", pool.Spec.Configuration.Name)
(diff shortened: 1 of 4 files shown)


kikisdeliveryservice:
nice!

@sgreene570 : could you add your what i did message body to the commit body along with a final line containing The Closes #890?

kikisdeliveryservice:
will let @runcom do the lgtm

/approve

Issue #888, opened 2019-06-25
Ignition version validation with Kargs|FIPS


runcom:
Playing around with kargs and fips, it's not possible to just create an MC like this:
```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: worker
  name: 50-kargs
spec:
  kargs:
    - nosmt
```

We should allow such a MC w/o incurring into ignition versions errors. It's not trivial though cause the ignition config is not a pointer so we can't really differentiate.

```
    Reason:                Failed to render configuration for pool worker: machine config: 50-kargs contains invalid ignition config: error: invalid config version (couldn't parse)
```

cgwalters:
We can detect if an Ignition spec is empty, and make that silently OK.

That said this is also potentially an argument to backtrack on extending `MachineConfig` as is and instead introducing some other CRD `operatingsystemconfig` or whatever.

runcom:
> We can detect if an Ignition spec is empty, and make that silently OK.

yes, but that requires looking at fields for the Ingnition Config struct cause it's not a pointer though

cgwalters:
> yes, but that requires looking at fields for the Ingnition Config struct 

We already walk the struct to e.g. verify that it doesn't have a non-empty `storage` section etc.

runcom:
> We already walk the struct to e.g. verify that it doesn't have a non-empty `storage` section etc.

yeah, I'm not against doing that 😄 I just wished we had a pointer to straight compare to `nil` and get done with it 😃

ericavonb:
I ran into this issue my first time playing around with machineconfigs 😄 . As annoying as the check for `nil` everywhere is, I'm in favor of the pointer for this one.

Should this we discussed in the openshift/api PR somewhere?

runcom:
> Should this we discussed in the openshift/api PR somewhere?

as a follow up to that I guess yeah. If that PR makes 4.2, we need a transition plan for 4.3 to account for this change but I won't risk changing this for 4.2 (as annoying as it is I concur 😄 )

kikisdeliveryservice:
The ignition verification specifically requires a version for a MC to be valid. That's kind of baked in and not specific to Kargs/FIPs in the MCO.

kikisdeliveryservice:
[...]

PR #894, opened 2019-06-26
daemon/controller: handle empty ignition configs

kikisdeliveryservice:
Wrap ignition validation in helper function that adds exception
for empty ign configs.  Use helper validation function in update.go
and render_controller.go

Update and add unit tests

Closes: #888

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -2,9 +2,12 @@ package common
 
 import (
 	"io/ioutil"
+	"reflect"
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
+	validate "github.com/coreos/ignition/config/validate"
 	"github.com/golang/glog"
+	errors "github.com/pkg/errors"
 )
 
 // NewIgnConfig returns an empty ignition config with version set as latest version
@@ -21,4 +24,20 @@ func WriteTerminationError(err error) {
 	msg := err.Error()
 	ioutil.WriteFile("/dev/termination-log", []byte(msg), 0644)
 	glog.Fatal(msg)
+}
+
+// ValidateIgnition wraps the underlying Ignition validation, but explicitly supports
+// a completely empty Ignition config as valid.  This is because we
+// want to allow MachineConfig objects which just have e.g. KernelArguments
+// set, but no Ignition config.
+// Returns nil if the config is valid (per above) or an error containing a Report otherwise.
+func ValidateIgnition(cfg igntypes.Config) error {
+	// only validate if Ignition Config is not empty
+	if reflect.DeepEqual(igntypes.Config{}, cfg) {
+		return nil
+	}
+	if report := validate.ValidateWithoutSource(reflect.ValueOf(cfg)); report.IsFatal() {
+		return errors.Errorf("invalid Ignition config found: %v", report)
+	}
+	return nil
 }
--- a/pkg/controller/common/helpers_test.go
+++ b/pkg/controller/common/helpers_test.go
@@ -1,0 +1,26 @@
+package common
+
+import (
+	"testing"
+
+	igntypes "github.com/coreos/ignition/config/v2_2/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateIgnition(t *testing.T) {
+	// Test that an empty ignition config returns nil
+	testIgnConfig := igntypes.Config{}
+	isValid := ValidateIgnition(testIgnConfig)
+	require.Nil(t, isValid)
+
+	// Test that an invalid ignition config returns and error
+	tempUser1 := igntypes.PasswdUser{Name: "core", SSHAuthorizedKeys: []igntypes.SSHAuthorizedKey{"5678", "abc"}}
+	testIgnConfig.Passwd.Users = []igntypes.PasswdUser{tempUser1}
+	isValid = ValidateIgnition(testIgnConfig)
+	require.NotNil(t, isValid)
+
+	// Test that a valid ignition config returns nil
+	testIgnConfig.Ignition.Version = "2.0.0"
+	isValid = ValidateIgnition(testIgnConfig)
+	require.Nil(t, isValid)
+}
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -5,11 +5,11 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/coreos/ignition/config/validate"
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/lib/resourceapply"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/controller/common"
+	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/scheme"
 	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
@@ -526,9 +526,8 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 func generateRenderedMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig, cconfig *mcfgv1.ControllerConfig) (*mcfgv1.MachineConfig, error) {
 	// Before merging all MCs for a specific pool, let's make sure each contains a valid Ignition Config
 	for _, config := range configs {
-		rpt := validate.ValidateWithoutSource(reflect.ValueOf(config.Spec.Config.Ignition))
-		if rpt.IsFatal() {
-			return nil, fmt.Errorf("machine config: %v contains invalid ignition config: %v", config.ObjectMeta.Name, rpt)
+		if err := ctrlcommon.ValidateIgnition(config.Spec.Config); err != nil {
+			return nil, err
 		}
 	}
 	merged := mcfgv1.MergeMachineConfigs(configs, cconfig.Spec.OSImageURL)
--- a/pkg/controller/render/render_controller_test.go
+++ b/pkg/controller/render/render_controller_test.go
@@ -265,11 +265,13 @@ func TestIgnValidationGenerateRenderedMachineConfig(t *testing.T) {
 	mcp := helpers.NewMachineConfigPool("test-cluster-master", helpers.MasterSelector, nil, "")
 	files := []igntypes.File{{
 		Node: igntypes.Node{
-			Path: "/dummy/0",
+			Filesystem: "root",
+			Path:       "/dummy/0",
 		},
 	}, {
 		Node: igntypes.Node{
-			Path: "/dummy/1",
+			Filesystem: "root",
+			Path:       "/dummy/1",
 		},
 	}}
 	mcs := []*mcfgv1.MachineConfig{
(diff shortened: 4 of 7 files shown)


runcom:
> Instead of letting an empty Ignition version trigger an error, let's bypass that for all MachineConfigs since it seems unnecessarily confusing for most users.

uhm, the version check *when* a ignition config is provided is actually valuable and Andrew has been telling us forever and that was probably why we added it.

Why don't we do like Colin said in https://github.com/openshift/machine-config-operator/issues/888#issuecomment-505474586 (since it's not a pointer)? that way we still leave the version check (w/o changing behavior between openshift versions) and allow for non-Ignition MC to go through as well w/o version

kikisdeliveryservice:
@runcom this PR _only_ covers the case when an ignition version is empty - if an ignition version is provided, everything functions (including the checks) as it did before.

I think that errorring out for general empty ignition values across the board makes little sense as a UX (I get questions about this a lot) and is super confusing whether or not the user is submitting kargs. if empty why not just supply a default value (which this PR does) and let the checks process whether or not the ignition config is valid otherwise and meets our reconcilable criteria (including our ignition checks)?

runcom:
> I think that errorring out for general empty ignition values across the board makes little sense as a UX (I get questions about this a lot) and is super confusing whether or not the user is submitting kargs. if empty why not just supply a default value (which this PR does) and let the checks process whether or not the ignition config is valid otherwise and meets our reconcilable criteria (including our ignition checks)?

The version check is to make sure the ignition embedded has the correct version to work with us and be understandable by the MCD - I'm not sure why people complain about this but it's something we enforce and it must be enforced to have _valid_ ignition embedded (till we change it to something else?). It's a check for correctness, as annoying as it may be (when we move away from ignition hopefully that goes away as well and we can rely on kube types versions?)
Example, if someones provide Ignition version v3 w/o providing the version, it's completely wrong to assume v2 and inject v2 (the MCD is also gonna fail anyway).

The issue here is that we shouldn't validate the inner `config` for the MC when it's _empty_ (Colin provided a suggestion in the issue to just walk the struct and derive emptiness/changes) nor I believe we should add default w/o knowledge of what it's embedded in the MC itself, that's just skipping the check so we might as well get rid of the check as well cause it's not fully honored anymore.

kikisdeliveryservice:
> Example, if someones provide Ignition version v3 w/o providing the version, it's completely wrong to assume v2 and inject v2 (the MCD is also gonna fail anyway).

The question, I think, is failing on an empty version field or failing on the rest of the MC as a UX. Given past discussions of the place of ignition in MCs, I wanted to bring up the possibility of not failing on empty and using a default instead.

I can switch this only to kargs, but also wanted to bring up this possibility.

cgwalters:
Right, my initial suggestion was to make it not an error to have an "empty" Ignition config, in other words it's OK to not have a version specified if the config is otherwise empty.

In other words replace all of our calls to `validate.ValidateWithoutSource` with a wrapper function that does something like:


```
func ValidateIgnition(cfg *Ignition) error {
  emptyIgnition := &Ignition{}
  if reflect.DeepEqual(&emptyIgnition, cfg) {
    return nil
  }
  return validate(cfg)
}
```

kikisdeliveryservice:
> Right, my initial suggestion was to make it not an error to have an "empty" Ignition config, in other words it's OK to not have a version specified if the config is otherwise empty.
> 

Ok, gotcha.

runcom:
>  if reflect.DeepEqual(&emptyIgnition, cfg) {

I wonder if that takes care of the comparison for emptiness but I think it will just fine since we're not moving to newer ignition config _now_ so we don't risk a new added field faling the DeepEqual check 👍

kikisdeliveryservice:
I think I was thinking too much about our future of MCs discussions, which, is not happening in this PR or soon.. :smile: 

> > if reflect.DeepEqual(&emptyIgnition, cfg) {
> 
> I wonder if that takes care of the comparison for emptiness but I think it will just fine since we're not moving to newer ignition config _now_ so we don't risk a new added field faling the DeepEqual check +1

It should be fine, I'll let you know if I run into any trouble. I'll add a test for that as well to cover it.

pkg/controller/common/helpers.go
```diff
@@ -22,3 +25,19 @@ func WriteTerminationError(err error) {
...
+	emptyIgnition := igntypes.Config{}
+
+	// only validate if Ignition Config is not empty
+	if !reflect.DeepEqual(emptyIgnition, cfg) {
+		report := validate.ValidateWithoutSource(reflect.ValueOf(cfg))
+		if report.IsFatal() {
+			return errors.Errorf("invalid Ignition config found: %v", report)
+		}
+	}
+	return nil
```
runcom: logic is perfect, I'd just invert the `if` to only have an indent level and less vars (readability mostly), but not blocking:

```go
        if reflect.DeepEqual(igntypes.Config{}, cfg) {
            return nil
        }
	if report := validate.ValidateWithoutSource(reflect.ValueOf(cfg)); report.IsFatal() {
		return errors.Errorf("invalid Ignition config found: %v", report)
	}
	return nil
```

kikisdeliveryservice: ahh i had it this way at first but switched it.  will make this change thanks!

kikisdeliveryservice: Done.

kikisdeliveryservice:
While working on this PR I noticed an error so this bug will close BZ 1726866 in master and once approved and merged will be backported.

kikisdeliveryservice:
I will have to update TestReconcileAfterBadMC() since applying a bad MC will not result in a render config that is generated but then fails.  With this PR a bad MC will result in no new render config being generated...

runcom:
> Looking at the e2e I think that there will end up being some redundancy between TestReconcileAfterBadMC() and the existing TestPoolDegradedOnFailToRender() if I fix it to account for the fixed validation since they are both now testing the same thing so I'll refactor them into one test.

I think we just have to tweak an MC to still test `NodeDegraded` in `TestReconcileAfterBadMC`- this can easily be done by tweaking `TestReconcileAfterBadMC` to create an MC that fails the `Reconcilable` check but still passes Ignition validation, like this:

```yaml
kind: MachineConfig
metadata:
  labels:
  name: 51-test
spec:
  config:
    ignition:
      version: 2.2.0
    networkd:
      units:
        - name: 00-eth0.network
          contents: |
            [Match]
            Name=eth0

            [Network]
            DNS=1.2.3.4
            Address=10.0.0.101/24
            Gateway=10.0.0.1
    storage:
      files:
      - contents:
          source: data:,test
        filesystem: root
        mode: 0644
        path: /test
```

pkg/controller/render/render_controller.go
```diff
@@ -526,9 +526,9 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 func generateRenderedMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig, cconfig *mcfgv1.ControllerConfig) (*mcfgv1.MachineConfig, error) {
 	// Before merging all MCs for a specific pool, let's make sure each contains a valid Ignition Config
 	for _, config := range configs {
-		rpt := validate.ValidateWithoutSource(reflect.ValueOf(config.Spec.Config.Ignition))
-		if rpt.IsFatal() {
-			return nil, fmt.Errorf("machine config: %v contains invalid ignition config: %v", config.ObjectMeta.Name, rpt)
+		ignErrMsg := ctrlcommon.ValidateIgnition(config.Spec.Config)
+		if ignErrMsg != nil {
```
runcom: this can just all be:

```go
if err := ctrlcommon.ValidateIgnition(config.Spec.Config); err != nil {
    return err
}
```

pkg/daemon/update.go
```diff
@@ -305,9 +305,10 @@ func Reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*MachineConfigDif
...
 	// Ignition section
 	// First check if this is a generally valid Ignition Config
-	rpt := validate.ValidateWithoutSource(reflect.ValueOf(newIgn))
-	if rpt.IsFatal() {
-		return nil, errors.Errorf("invalid Ignition config found: %v", rpt)
+
+	ignErrMsg := ctrlcommon.ValidateIgnition(newIgn)
+	if ignErrMsg != nil {
+		return nil, ignErrMsg
 	}
```
runcom: same as https://github.com/openshift/machine-config-operator/pull/894/files#r300422961 - no need for the extra err variable

LorbusChris:
Should we use reflect to check for the empty field or compare the typed object directly here for validation?
see for a change away from reflection to that: https://github.com/openshift/machine-config-operator/pull/923/files

runcom:
@LorbusChris this is a bit different cause we would have to compare the whole Ignition struct which is rather big to go through and reflection gives us a one off comparison with an empty (just initialized) struct

kikisdeliveryservice:
> I think we just have to tweak an MC to still test `NodeDegraded` in `TestReconcileAfterBadMC`- this can easily be done by tweaking `TestReconcileAfterBadMC` to create an MC that fails the `Reconcilable` check but still passes Ignition validation, like this:
> 

Sounds good, I'll try it out! Thanks!!

pkg/controller/common/helpers_test.go
```diff
@@ -0,0 +1,33 @@
...
+	testIgnConfig.Passwd.Users = []igntypes.PasswdUser{tempUser1}
+	isValid = ValidateIgnition(testIgnConfig)
+	if isValid == nil {
+		t.Errorf("Expected an error from invalid ignition config")
+	}
+
+	// Test that a valid ignition config returns nil
+	testIgnConfig.Ignition.Version = "2.0.0"
+	isValid = ValidateIgnition(testIgnConfig)
+	if isValid != nil {
```
cgwalters: optional: This could just be `require.Nil(isValid)`

pkg/controller/render/render_controller_test.go
```diff
@@ -283,23 +285,36 @@ func TestIgnValidationGenerateRenderedMachineConfig(t *testing.T) {
...
 	_, err = generateRenderedMachineConfig(mcp, mcs, cc)
 	if err == nil {
 		t.Fatalf("expected error. mcs contains a machine config with invalid ignconfig version")
 	}
+
+	// verify that a machine config with no ignition content will not fail validation
+	mcs[1].Spec.Config = igntypes.Config{}
+	mcs[1].Spec.KernelArguments = append(mcs[1].Spec.KernelArguments, "test1")
+	_, err = generateRenderedMachineConfig(mcp, mcs, cc)
+	if err != nil {
```
cgwalters: Ditto

pkg/controller/common/helpers.go
```diff
@@ -22,3 +25,17 @@ func WriteTerminationError(err error) {
 	ioutil.WriteFile("/dev/termination-log", []byte(msg), 0644)
 	glog.Fatal(msg)
 }
+
+// ValidateIgnition evaluates whether a given non-empty Ignition Config is valid
```
cgwalters: How about:

```
// ValidateIgnition wraps the underlying Ignition validation, but explicitly supports
// a completely "empty" Ignition config as valid in addition.  This is because we
// want to allow MachineConfig objects which just have e.g. KernelArguments
// set, but no Ignition config.
// Returns nil if the config is valid (per above) or an error otherwise.
```

kikisdeliveryservice: this is much better. thanks!

pkg/daemon/update.go

pkg/controller/common/helpers_test.go
kikisdeliveryservice: done

pkg/controller/render/render_controller.go
kikisdeliveryservice: done

pkg/controller/render/render_controller_test.go
kikisdeliveryservice: done

pkg/controller/common/helpers.go
kikisdeliveryservice: done

test/e2e/mcd_test.go
```diff
@@ -355,8 +355,20 @@ func TestReconcileAfterBadMC(t *testing.T) {
 	cs := framework.NewClientSet("")
 	bumpPoolMaxUnavailableTo(t, cs, 3)
 
-	// create a bad MC w/o a filesystem field which is going to fail reconciling
-	mcadd := createMCToAddFile("add-a-file", "/etc/mytestconfs", "test", "")
+	// create a bad MC that contains a valid ignition config that is not reconcilable
+	mcadd := &mcfgv1.MachineConfig{
```
runcom: you're missing metadata.name at least  and test is failing - why don't we just re-use `createMCToAddFile` which creates a valid ignition config and then we append the networkd bits as you're doing below?

kikisdeliveryservice: gah. thank you! done.

kikisdeliveryservice:
this really shouldn't be failing seeing weird things in e2e logs...

test/e2e/mcd_test.go
```diff
@@ -355,8 +355,16 @@ func TestReconcileAfterBadMC(t *testing.T) {
 	cs := framework.NewClientSet("")
 	bumpPoolMaxUnavailableTo(t, cs, 3)
 
-	// create a bad MC w/o a filesystem field which is going to fail reconciling
+	// create a MC that contains a valid ignition config but is not reconcilable
 	mcadd := createMCToAddFile("add-a-file", "/etc/mytestconfs", "test", "")
```
runcom: you need to add `root` as the last arg to this func, right now it's creating an invalid ignition config anyway but we need a valid one with unreconcilable code (as you did below)

kikisdeliveryservice: thanks will. still unsure why i hit those weird e2e errors tho.

kikisdeliveryservice:
`"Error: Error creating S3 bucket: TooManyBuckets: You have attempted to create more buckets than allowed"`

/retest

PR #898, opened 2019-06-26
API Updates

LorbusChris:
Various API Updates as suggested in and in preparation for https://github.com/openshift/api/pull/301

Added KubeletConfiguration to RawExtension conversion to make API independent from kubelet type/api versioning.


pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -223,9 +223,9 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
 		return fmt.Errorf("status for ControllerConfig %s is being reported for %d, expecting it for %d", ccName, cur.Status.ObservedGeneration, cur.Generation)
 	}
 
-	completed := IsControllerConfigStatusConditionTrue(cur.Status.Conditions, TemplateContollerCompleted)
-	running := IsControllerConfigStatusConditionTrue(cur.Status.Conditions, TemplateContollerRunning)
-	failing := IsControllerConfigStatusConditionTrue(cur.Status.Conditions, TemplateContollerFailing)
+	completed := IsControllerConfigStatusConditionTrue(cur.Status.Conditions, TemplateControllerCompleted)
```
kikisdeliveryservice: is this a true change from the original? i can't see the difference?

kikisdeliveryservice: it's a typo fix!!  Can you pull this out as a seperate commit?

kikisdeliveryservice:
uhohs. your e2e-aws-op is hitting the same  hyperkube error: `1347 runtime.go:69] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)` as my other pr... :frowning: 

https://github.com/openshift/machine-config-operator/pull/899#issuecomment-506470095

LorbusChris:
@rphillips could you review the second commit especially. Want to make sure it's ok to encode to JSON internally here instead of YAML.

@sttts please have a look at the KubeletConfiguration<->RawExtension conversion in the third commit. Note that I am re-using and not adding the actual conversion funcs. 

cc @runcom @cgwalters

pkg/controller/kubelet-config/helpers.go
```diff
@@ -67,10 +67,14 @@ func getManagedKubeletConfigKey(pool *mcfgv1.MachineConfigPool) string {
 
 // validates a KubeletConfig and returns an error if invalid
 func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
-	if cfg.Spec.KubeletConfig == nil {
+	if cfg.Spec.KubeletConfig.Raw == nil {
 		return nil
 	}
-	kcValues := reflect.ValueOf(*cfg.Spec.KubeletConfig)
+	kcDecoded, err := decodeKubeletConfig(cfg.Spec.KubeletConfig.Raw)
```
sttts: where is decodeKubeletConfig ?

pkg/controller/kubelet-config/helpers.go
```diff
@@ -161,12 +165,12 @@ func encodeKubeletConfig(internal *kubeletconfigv1beta1.KubeletConfiguration, ta
...
 }
 
-func newKubeletconfigYAMLEncoder(targetVersion schema.GroupVersion) (runtime.Encoder, error) {
+func newKubeletconfigJSONEncoder(targetVersion schema.GroupVersion) (runtime.Encoder, error) {
 	_, codecs, err := kubeletconfigscheme.NewSchemeAndCodecs()
 	if err != nil {
 		return nil, err
 	}
-	mediaType := "application/yaml"
+	mediaType := "application/json"
```
sttts: why not yaml?

pkg/controller/kubelet-config/helpers.go
LorbusChris: it's a func in this file that I am reusing for this

pkg/controller/kubelet-config/kubelet_config_controller_test.go
```diff
@@ -109,11 +109,18 @@ func newControllerConfig(name, platform string) *mcfgv1.ControllerConfig {
 }
 
 func newKubeletConfig(name string, kubeconf *kubeletconfigv1beta1.KubeletConfiguration, selector *metav1.LabelSelector) *mcfgv1.KubeletConfig {
+	kcRaw, err := encodeKubeletConfig(kubeconf, kubeletconfigv1beta1.SchemeGroupVersion)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
sttts: oh, this is in a test. nvmd. Then it is fine.

pkg/controller/kubelet-config/helpers.go
```diff
@@ -67,10 +67,14 @@ func getManagedKubeletConfigKey(pool *mcfgv1.MachineConfigPool) string {
...
-	if cfg.Spec.KubeletConfig == nil {
+	if cfg.Spec.KubeletConfig.Raw == nil {
 		return nil
 	}
-	kcValues := reflect.ValueOf(*cfg.Spec.KubeletConfig)
+	kcDecoded, err := decodeKubeletConfig(cfg.Spec.KubeletConfig.Raw)
+	if err != nil {
+		return fmt.Errorf("KubeletConfig could not be unmarshalled, err: %v", err)
+	}
+	kcValues := reflect.ValueOf(*kcDecoded)
```
sttts: the banned field logic below works, but it is totally untyped. Would prefer to have real types here, especially because field names might change from config version to config version.

pkg/controller/kubelet-config/helpers.go
LorbusChris: RawExtension.Raw expects raw JSON

rphillips:
Does this upgrade a config that is already in yaml to json?

rphillips:
/hold

for @sttts to confirm his comments were addressed.

runcom:
uhm, too late, @LorbusChris can you make sure you follow up Stephan comments asap?

pkg/controller/kubelet-config/helpers.go
LorbusChris: I don't think I changed the logic here. (kcDecoded is still typed as KubeletConfiguration as before). Do you mean to put a type around [blacklistKubeletConfigurationFields](https://github.com/LorbusChris/machine-config-operator/blob/master/pkg/controller/kubelet-config/kubelet_config_controller.go#L62)? Do you have an example for guidance at hand?

sttts: no, you did not change that. But with RawExtension it's getting harder to reason about.

A good pattern is to type-cast to a real type after decoding. And then compare the real fields, not via reflection.

runcom: > A good pattern is to type-cast to a real type after decoding.

decode routine already type-cast to `KubeletConfiguration`. Are you suggesting instead of playing with reflections below, do something like:

```go
// basically remove the loop and go one by one for banned fields
if kcDecoded.CgroupDriver != "" {
    return errors.New("CgroupDriver is not allowed to be set")
}
...
```

LorbusChris: resolved in https://github.com/openshift/machine-config-operator/pull/923

PR #850, opened 2019-06-12
*: create a single MCO image containing all MC* binaries

kikisdeliveryservice:
This is the 3rd pr in a series to fold all MC* images into one.
PR does the following as step 3:
- update dockerfile.machine-config-operator to generates all component
images (mco/mcc/mcd/mcs/setup-etcd-env)
- update dockerfile.machine-config-operator.rhel7 the same as sbove
- update image-references, configmap and manifests to use MCO image
- update MCO image flag in bootstrap.go to be required

TO-DO in follow-on prs:
- removes other dockerfiles to have 1 dockerfile generating image
- make another PR in release to only reference the above dockerfile
- make another PR in installer later to remove individual refs in bootkube
- make another PR here to remove the component image refs
- remove component image refs in ART(??)

Related-to: #847 (merged)
Requires: openshift/installer#1847 (merged)
Related-to: #739 (issue will only close when final installer pr merges)

Issue #739, opened 2019-05-13
Fold MCO images into one


runcom:
Since MCO and MCC are strictly related, it would be appropriate to fold all our images into just one to avoid falling into bugs and error whenever the build system isn't properly setup and could lead to different images being built from different commits. 

This requires changes at least to:

- installer's bootkube
- mco itself

To expand on the above, this should be done in 2 steps to avoid any breakage (and we need to think about backwards compatibility as well):

- tweak bootkube to accept the MCO image itself (requires a PR in MCO to add a new flag for the MCO image, then a PR in the installer for bootkube to pass it)
- change Dockerfiles to build all binaries into a single image here in MCO
- remove the entrypoint(s) from the image above
- create a compatibility entrypoint for https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L187 and set it to the above image
- tweak the `machine-config-operator` cmd to no-op on these flags https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L196-L200 (except kube-client-agent) image, and instead use the MCO image itself directly (which we plugged at the beginning)
- now the CI should be good so merge the above PR
- create a PR in the installer to tweak bootkube to only use 1 image
- merge the above
- remove compatibility code from point 1 and the fake entrypoint and all other compatibility code
- merge the above
- done

runcom:
I've expanded the points above to something clearer now

runcom:
@cgwalters ptal

cgwalters:
A preparatory step for this is probably to go to a single binary (`hypermco`) too right?  Otherwise we're wasting a lot of space.  But I guess it's not required.

I was thinking of this more like:

 - Change e.g. `Dockerfile.machine-config-operator` to put all the binaries inside there
 - Drop the default `ENTRYPOINT`, change our manifests to launch that image

Now we have a single image running in the cluster by default; no other PRs to installer required right?
But we are building 3 unnecessary images, so the remaining work would be similar to your outline.  I suspect we'd also need patches to the release controller and ART's configuration to delete builds?

runcom:
> I suspect we'd also need patches to the release controller and ART's configuration to delete builds?

yep, that's gonna be a required thing anyway.

> I was thinking of this more like:
> 
> * Change e.g. `Dockerfile.machine-config-operator` to put all the binaries inside there
[...]

PR #895, opened 2019-06-26
* : remove refs to component images & dockerfiles

kikisdeliveryservice:
Remove all references to mcc/mcd/mcs/setupetdenv images so that we rely on the new MCO mega image only.

Note: Dockerfile.rhel7 images will be removed in a followup PR pending ART merge

Requires: https://github.com/openshift/installer/pull/1875
Related-to: openshift/machine-config-operator#739

PR #899, opened 2019-06-26
templates/mco: update template to replace setupetcdenv image with MCO image

kikisdeliveryservice:
Operator.go & bootstrap.go need to reference the MCO image instead of the
old setupetcdenv image.

Update template to add setupetcdenv entrypoint for MCO image.

Required-for: openshift/installer#1875
Related-to: #850
Related-to: #739 (issue will only close when final installer pr merges)

cc: @runcom

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -112,7 +112,7 @@ func RenderBootstrap(
 	spec.OSImageURL = imgs.MachineOSContent
 	spec.Images = map[string]string{
 		templatectrl.EtcdImageKey:            imgs.Etcd,
-		templatectrl.SetupEtcdEnvKey:         imgs.SetupEtcdEnv,
+		templatectrl.SetupEtcdEnvKey:         imgs.MachineConfigOperator,
 		templatectrl.InfraImageKey:           imgs.InfraImage,
 		templatectrl.KubeClientAgentImageKey: imgs.KubeClientAgent,
 	}
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -14,6 +14,7 @@ contents:
       initContainers:
       - name: discovery
         image: "{{.Images.setupEtcdEnv}}"
+        command: ["/usr/bin/setup-etcd-environment"]
         args:
         - "run"
         - "--discovery-srv={{.EtcdDiscoveryDomain}}"
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -406,7 +406,7 @@ func (optr *Operator) sync(key string) error {
 	spec.OSImageURL = imgs.MachineOSContent
 	spec.Images = map[string]string{
 		templatectrl.EtcdImageKey:            imgs.Etcd,
-		templatectrl.SetupEtcdEnvKey:         imgs.SetupEtcdEnv,
+		templatectrl.SetupEtcdEnvKey:         imgs.MachineConfigOperator,
 		templatectrl.InfraImageKey:           imgs.InfraImage,
 		templatectrl.KubeClientAgentImageKey: imgs.KubeClientAgent,
 	}


templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -13,7 +13,8 @@ contents:
     spec:
       initContainers:
       - name: discovery
-        image: "{{.Images.setupEtcdEnv}}"
+        image: "{{.Images.MachineConfigOperator}}"
+        command: ["/usr/bin/setup-etcd-environment"]
```
runcom: going through some other changes, and noticed we need to take care of https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/bootstrap.go#L104 as well?

CI seems to be unhappy about this change also (at least the last run)

kikisdeliveryservice: that change was going to be covered in https://github.com/openshift/machine-config-operator/pull/895 where i was removing all of the non-template refs to images. was on the fence about doing it here.  but you are probably right that i should. ill give it a try and see if that fixes ci

kikisdeliveryservice: after i first look thru the logs of course :)

kikisdeliveryservice:
@runcom I think this is not my PR that is failing but BZ: 1724346 other people are seeing the same: ```hyperkube[1339]: E0626 21:47:25.979131    1339 runtime.go:69] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)``` as this pr starting early today.  Will investigate more.

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
kikisdeliveryservice: on second thought making the change in boostrap.go would be inconsistent with how I'm treating mcc/mcs/mcd up til now. #850 added the mco image and uses it in templates (like the change i have here) but doesn't remove any other references to the images throughout.

runcom: don't we need https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/bootstrap.go#L104 to actually _render_ the template for the etcd pod?

runcom: guess not actually, so #895 it is

kikisdeliveryservice:
(Pushed a commit updating the author date to stop the bot's /retest insanity for now.)

kikisdeliveryservice:
Hitting BZ 1724346

See in tests logs e2e-aws: artifacts/e2e-aws/installer/logbundle..../bootstrap/journals/kubelet.log :
`Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer`

kikisdeliveryservice:
namespace deleted, trying again

/retest

runcom:
This 100% fails with that panic 🤔 I'm wondering if something else is going on here (but upgrade passed!)

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
runcom: I'd say let's move to https://github.com/openshift/machine-config-operator/pull/895 where this diff is there as well and work from there

runcom: I'm also thinking this can be the issue actually, the MCO image isn't plumbed here at all today so this PR is running w/o the MCO image for setupEtcdEnv which relates to the whole convo in https://github.com/openshift/machine-config-operator/pull/895/files#r298159766

kikisdeliveryservice: the thing is that the other substitutions for the other images were in `manifests/` where this change is in templates so i was thinking that maybe that is the problem?

kikisdeliveryservice: i can't move this to 895 bc this change is needed for the installer pr. but can i pull that change from 895 and move it here?

kikisdeliveryservice: ok moved the 895 change to this pr so we'll see soon!

pkg/operator/bootstrap.go
```diff
@@ -112,7 +112,7 @@ func RenderBootstrap(
 	spec.OSImageURL = imgs.MachineOSContent
 	spec.Images = map[string]string{
 		templatectrl.EtcdImageKey:            imgs.Etcd,
-		templatectrl.SetupEtcdEnvKey:         imgs.SetupEtcdEnv,
+		templatectrl.SetupEtcdEnvKey:         imgs.MachineConfigOperator,
```
runcom: missed https://github.com/openshift/machine-config-operator/pull/895/files#r298159078 ? (but I saw you commented in there)

kikisdeliveryservice: yep got it! trying again :)

runcom:
aaarg

kikisdeliveryservice:
Ok I can confirm that the problem is this branch. The mco image isn't being passed in, in the template- templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml-  (tho it is passed elsewhere successfully from the bootstrap.go/operator.go):
```Jul 02 22:28:13 ip-10-0-134-0 hyperkube[1002]: I0702 22:28:13.014075    1002 kuberuntime_manager.go:617] computePodActions got {KillPod:false CreateSandbox:false SandboxID:08f5fe55cac2d4092e184e2c95436da69e0e4f14e557c5196535127ae9992919 Attempt:0 NextInitContainerToStart:&Container{Name:discovery,**Image:<no value>,Command:[/usr/bin/setup-etcd-environment]**,Args:[run --discovery-srv=ci-op-2y2yd539-c4a31.origin-ci-int-aws.dev.rhcloud.com --output-file=/run/etcd/environment --v=4],WorkingDir:,Ports:[],Env:[],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{},},VolumeMounts:[{discovery false /run/etcd/  <nil> }],LivenessProbe:nil,ReadinessProbe:nil,Lifecycle:nil,TerminationMessagePath:/dev/termination-log,ImagePullPolicy:IfNotPresent,SecurityContext:nil,Stdin:false,StdinOnce:false,TTY:false,EnvFrom:[],TerminationMessagePolicy:File,VolumeDevices:[],} ContainersToStart:[] ContainersToKill:map[]} for pod "etcd-member-ip-10-0-134-0.ec2.internal_openshift-etcd(d9c9ef576d3104b9b56d58e8e6490a12)"```

File: templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml [hunk 3]
```diff
@@ -13,7 +13,8 @@ contents:
     spec:
       initContainers:
       - name: discovery
-        image: "{{.Images.setupEtcdEnv}}"
+        image: "{{.Images.machineConfigOperator}}"
```
kikisdeliveryservice: so there's an issue here that is still failing.  I thought I found a fix by subbing MachineConfigOperator -> machineConfigOperator I did this bc I noticed the image referenced was setupEtcdEnv and kubeClientAgentImage which is from https://github.com/openshift/machine-config-operator/blob/master/install/0000_80_machine-config-operator_02_images.configmap.yaml

kikisdeliveryservice: is this image somehow referenced _before_  it has a value? (since this is in a different place than the other mc* images)

runcom: Ok, I see it 😄 we still have to use `Images. setupEtcdEnv`cause that's the key being substituted

runcom: right, I've created a smoke PR (now closed) https://github.com/openshift/machine-config-operator/pull/915 and it shows tests passing

kikisdeliveryservice: i stared at this for so long!!!! can't believe i didnt see this - thank you!

PR #900, opened 2019-06-26
docs/README: provide base64 encoded example

miabbott:
**- What I did**
Updated the docs with a `base64` encoded example

**- How to verify it**
Apply the example MachineConfig and verify the file is landed on nodes

docs/README: Provided an example of using `base64` encoded file contents


miabbott:
Maybe it is just me, but using `base64` is easiest when trying to encode files.

runcom:
👍 thanks @miabbott ! (this is also great cause in theory, you could ship a binary fine by base64 encoding it)

/lgtm

PR #901, opened 2019-06-26
MCO: add Proxy support

runcom:
**- What I did**

Add Proxy support as per https://jira.coreos.com/browse/MCO-55

This PR basically wires Proxy enablement through proxy.config.openshift.io.
This is actually different from the sub-controller as I think having the MCO resync the MC for proxy is equivalent anyway w/o having a specialized subcontroller.
This approach has the downside of being applied to master and worker - the other approach could have been more granular but proxy is cluster-wise anyway so I can't see why you want a proxy on workers only when pivot at least could require a master&worker proxy anyway.

EDIT: just went and checked 3.x docs:

> The proxy configuration must be the same on each host in the cluster. Therefore, when setting up the proxy or modifying it, you must update the files on each OpenShift Container Platform host to the same values. Then, you must restart OpenShift Container Platform services on each host in the cluster.

From https://docs.openshift.com/container-platform/3.5/install_config/http_proxies.html#overview so this looks like the correct approach

cc @mrunalp 

**- How to verify it**


runcom:
looks like compile is getting killed, might need a memory bump in CI job, I'll take a look

runcom:
good good

cgwalters:
Going to
/lgtm
based on the above and the fact we have green CI but there is one detail I want to follow up on after this which is we need to also patch `/etc/systemd/user.conf.d/proxy.conf` so that e.g. `systemctl --user run env` picks it up too.  While systemd user sessions may seem like a bit of an odd case they're *very* useful in general particularly with rootless podman.

PR #902, opened 2019-06-27
Fix ControllerConfigStatusConditionType typos

LorbusChris:
**- What I did**
Fixed ControllerConfigStatusConditionType typos by adding the missing r in TemplateContoller status conditions

**- How to verify it**
CI


LorbusChris:
let's get this in first.
/cc @kikisdeliveryservice

runcom:
This shouldn't disrupt upgrades afaict (and CI will tell) - we may only risk to have the old status "types" laying around forever but we might be able to clear those out somehow

/approve

Issue #662, opened 2019-04-25
Expose API for blocking and initiating reboots


cgwalters:
We have multiple requests for a way for external components to initiate the drain+reboot.  Other software *can* do this but if it's not coordinated with the MCO problems will arise.

We face a tension for whether some of this functionality should be part of the MCO itself, but anyways here's a strawman:

`/run/machine-config-operator/rebooting` is a flag file the MCO writes when it starts a drain+reboot.

`/run/machine-config-operator/reboot-lock.d/`
is a directory, and a component can do e.g.:
`echo preparing new kernel module > /run/machine-config-operator/reboot-lock.d/nvidia`

As long as at least 1 entry exists, the MCO will not initiate a drain + reboot.  However in order to avoid a race condition, external components must check if `/run/machine-config-operator/rebooting` exists after they wrote a lock file - if it does, they lost a race and should not attempt any host modifications.

`/run/machine-config-operator/reboot-start.d/`

When the app has completed its preparation, it does:
`mv /run/machine-config-operator/reboot-{lock,start}.d/nvidia`

And the MCD will start a drain+reboot when there is at least 1 entry in `.start.d` and no locks.

pliurh:
I have a use case for this API. Now we are working on an operator which can setup the SRIOV NIC on baremetal hosts. We plan to support NICs from Intel and Mellanox. 

However, for Mellanox NICs, to enable the SRIOV and set the max number of VFs, it requires to use 'mstconfig' tool to configure the firmware, then to reboot the node to take effect[1]. So we would like to have this API so that rebooting can be better coordinated among the cluster. So that we don't need to implement the drain+reboot logic in our own daemon.
 
[1]https://access.redhat.com/articles/3082811

jeremyeder:
The tuned use-case for needing to trigger reboots is when tuned adds kernel command line options that are part of a tuned profile.  Several discussions occurred with the perf team and MCO team over the last 6 months about ensuring tunings delivered via tuned are clearly & cleanly coordinated with the MCO and MCD.

/cc @runcom @jmencak

cgwalters:
> The tuned use-case for needing to trigger reboots is when tuned adds kernel command line options that are part of a tuned profile.

I really rather strongly think only the MCO should be writing kernel arguments.  See https://github.com/openshift/machine-config-operator/issues/388

pliurh:
@cgwalters #388 is for RHCOS only, right? What is the plan for BYOH case?

[...]

PR #904, opened 2019-06-27
 daemon: Add a m-c-d-firstboot.service that handles encapsulated MC

cgwalters:
The saga of adding "firstboot MachineConfig" in
#798
is getting closer.  This is a small amount of code that builds on
a lot of prep work that landed to add a systemd service + entrypoint
in the MCD that reads the
`/etc/ignition-machine-config-encapsulated.json` file.  You
could think of this a lot like a variant of "once-from".

However, it doesn't run by default right now because the MCS still
serves `/etc/pivot/image-pullspec`, and this gives us a way to
"ratchet" the change as we need this new MCD code to land in RHCOS
before we can use it.

### DIFF
--- a/cmd/machine-config-daemon/firstboot_complete_machineconfig.go
+++ b/cmd/machine-config-daemon/firstboot_complete_machineconfig.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	daemon "github.com/openshift/machine-config-operator/pkg/daemon"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+var firstbootCompleteMachineconfig = &cobra.Command{
+	Use:                   "firstboot-complete-machineconfig",
+	DisableFlagsInUseLine: true,
+	Short:                 "Complete the host's initial boot into a MachineConfig",
+	Args:                  cobra.MaximumNArgs(0),
+	Run:                   executeFirstbootCompleteMachineConfig,
+}
+
+// init executes upon import
+func init() {
+	rootCmd.AddCommand(firstbootCompleteMachineconfig)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+}
+
+func runFirstBootCompleteMachineConfig(_ *cobra.Command, _ []string) error {
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+
+	exitCh := make(chan error)
+	defer close(exitCh)
+
+	dn, err := daemon.New(daemon.NewNodeUpdaterClient(), exitCh)
+	if err != nil {
+		return err
+	}
+
+	err = dn.RunFirstbootCompleteMachineconfig()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func executeFirstbootCompleteMachineConfig(cmd *cobra.Command, args []string) {
+	err := runFirstBootCompleteMachineConfig(cmd, args)
+	if err != nil {
+		fmt.Printf("error: %v\n", err)
+		os.Exit(1)
+	}
+}
--- a/cmd/machine-config-daemon/firstboot_complete_machineconfig.go
+++ b/cmd/machine-config-daemon/firstboot_complete_machineconfig.go
@@ -36,12 +36,7 @@ func runFirstBootCompleteMachineConfig(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	err = dn.RunFirstbootCompleteMachineconfig()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return dn.RunFirstbootCompleteMachineconfig()
 }
 
 func executeFirstbootCompleteMachineConfig(cmd *cobra.Command, args []string) {
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -163,7 +163,7 @@ func (dn *Daemon) drainAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error)
 
 var errUnreconcilable = errors.New("unreconcilable")
 
-func canonicalizeEmptyConfig(config *mcfgv1.MachineConfig) *mcfgv1.MachineConfig {
+func canonicalizeEmptyMC(config *mcfgv1.MachineConfig) *mcfgv1.MachineConfig {
 	if config != nil {
 		return config
 	}
@@ -175,9 +175,9 @@ func canonicalizeEmptyConfig(config *mcfgv1.MachineConfig) *mcfgv1.MachineConfig
 	}
 }
 
-func (dn *Daemon) checkUpdateDiff(oldConfig, newConfig *mcfgv1.MachineConfig) bool {
+func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfig) bool {
 	var diff *MachineConfigDiff
-	oldConfig = canonicalizeEmptyConfig(oldConfig)
+	oldConfig = canonicalizeEmptyMC(oldConfig)
 	oldConfigName := oldConfig.GetName()
 	newConfigName := newConfig.GetName()
 	diff = NewMachineConfigDiff(oldConfig, newConfig)
@@ -190,7 +190,7 @@ func (dn *Daemon) checkUpdateDiff(oldConfig, newConfig *mcfgv1.MachineConfig) bo
 
 // update the node to the provided node configuration.
 func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr error) {
-	oldConfig = canonicalizeEmptyConfig(oldConfig)
+	oldConfig = canonicalizeEmptyMC(oldConfig)
 
 	if dn.nodeWriter != nil {
 		state, err := getNodeAnnotationExt(dn.node, constants.MachineConfigDaemonStateAnnotationKey, true)
@@ -333,21 +333,18 @@ type MachineConfigDiff struct {
 	units    bool
 }
 
-// See if this was Rust we wouldn't have nullable pointers by default...
-func canonicalizeStringArray(a []string) []string {
-	if a == nil {
-		return []string{}
-	}
-	return a
-}
-
 // NewMachineConfigDiff compares two MachineConfig objects.
 func NewMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) *MachineConfigDiff {
 	oldIgn := oldConfig.Spec.Config
 	newIgn := newConfig.Spec.Config
+
+	// Both nil and empty slices are of zero length,
+	// consider them as equal while comparing KernelArguments in both MachineConfigs
+	kargsEmpty := len(oldConfig.Spec.KernelArguments) == 0 && len(newConfig.Spec.KernelArguments) == 0
+
 	return &MachineConfigDiff{
 		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
-		kargs:    !reflect.DeepEqual(canonicalizeStringArray(oldConfig.Spec.KernelArguments), canonicalizeStringArray(newConfig.Spec.KernelArguments)),
+		kargs:    !(kargsEmpty || reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments)),
 		fips:     oldConfig.Spec.FIPS != newConfig.Spec.FIPS,
 		passwd:   !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd),
 		files:    !reflect.DeepEqual(oldIgn.Storage.Files, newIgn.Storage.Files),
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -25,6 +25,7 @@ import (
 	errors "github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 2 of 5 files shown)


cgwalters:
Vertical pod autoscaler flake looks like.
/test e2e-aws

runcom:
This looks right to me and loving the diff methods as well

/approve

cgwalters:
Further progress here blocks on https://github.com/openshift/machine-config-operator/issues/662#issuecomment-507785189 I think since we need to handle the "4.1.0 bootimage" case no matter what.

But, I also think this PR is pretty safe to land as is; there's just some prep refactoring, the end commit doesn't do anything since as noted above the `ConditionPathExists=` will always be false.

cgwalters:
Would it help to split off a separate PR with the prep commits here?

runcom:
> Would it help to split off a separate PR with the prep commits here?

it would :pray:

cgwalters:
Done in https://github.com/openshift/machine-config-operator/pull/935

sinnykumari:
> However, it doesn't run by default right now because the MCS still
> serves `/etc/pivot/image-pullspec`, and this gives us a way to
> "ratchet" the change as we need this new MCD code to land in RHCOS
> before we can use it.

This PR has nice work done in MCD to read encapsulated MC available on nodes.  It looks to me that we will need these changes in machine-config-daemon package shipped with RHCOS. It seems that this PR is going to make into 4.3+. Since we don't have a way to update bootimages, are we going to need some way to get latest machine-config-daemon binary for 4.2 cluster upgrade? Something similar to what we have been discussing for 4.1 cluster update for Day1 kargs in https://github.com/openshift/machine-config-operator/issues/798#issuecomment-512819035 and further comment?

cgwalters:
> Something similar to what we have been discussing for 4.1 cluster update for Day1 kargs in #798 (comment) and further comment?

Yeah, encoding the mcd binary into Ignition is the best thing I can think of.

sinnykumari:
@cgwalters Can we get this PR rebased to get it tested with latest changes?

cgwalters:
So https://github.com/openshift/machine-config-operator/pull/935 had some prep cleanup but a bit surprisingly this PR seems to rebase cleanly and build at least without it.  It might be there was some semantic stuff necessary there, or maybe I just jumped the gun and churned the code unnecessarily.  Unfortunately now I've forgotten the full context.

Regardless...I think before we can really use this PR we need to land support for "MCD injection" i.e. injecting the MCD binary via Ignition.

templates/common/_base/units/machine-config-daemon-firstboot.service
```diff
@@ -0,0 +1,19 @@
+name: "machine-config-daemon-firstboot.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Machine Config Daemon Firstboot
+  # This effectively disables this unit unit we patch the MCS, so
+  # we can land this code and then get it built into the host.
+  ConditionPathExists=!/etc/pivot/image-pullspec
+  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
+  After=ignition-firstboot-complete.service
```
sinnykumari: While I was testing this locally, noticed that `machine-config-daemon-firstboot.service` runs when a newly created cluster reboots after applying ignition configs and updates, which leads to another reboot. It's because /etc/pivot/image-pullspec gets deleted once cluster is updated to latest machine-os-content during first boot.  Should we also add something like `BindsTo=ignition-firstboot-complete.service` here to avoid running it in next reboot or it's an acceptable behavior?

sinnykumari:
> So #935 had some prep cleanup but a bit surprisingly this PR seems to rebase cleanly and build at least without it. It might be there was some semantic stuff necessary there, or maybe I just jumped the gun and churned the code unnecessarily. Unfortunately now I've forgotten the full context.

Thanks for rebasing the PR!

> Regardless...I think before we can really use this PR we need to land support for "MCD injection" i.e. injecting the MCD binary via Ignition.

Are we talking this in context of testing this PR by injecting machine-config-daemon binary  (containing this change) through ignition config during cluster install or this is something else ?

I tried to test this PR locally by generating manifests and editing existing [99_openshift-machineconfig_worker.yaml](https://sinnykumari.fedorapeople.org/99_openshift-machineconfig_master.yaml) and  [99_openshift-machineconfig_master.yaml](https://sinnykumari.fedorapeople.org/99_openshift-machineconfig_worker.yaml) files injecting machine-config-daemon into /usr/local/bin/ path and machine-config-daemon-firstboot.service file which gets successfully added to nodes, even runs and applies kargs during initial boot if I edit service file accordingly. Is it right way to test it?

One issue which I am seeing with my test approach is mcd fails to apply rendered config with error `daemon.go:1293] couldn't parse file: missing data prefix` (error message from one of running mcd pod). It's because I am injecting mcd binary into ignition as `https` path instead of `data` in source and MCD doesn't seems to support it while [validating source](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/daemon.go#L1248)

templates/common/_base/units/machine-config-daemon-firstboot.service
cgwalters: Hmm...we should be deleting `/etc/pivot/image-pullspec` during firstboot.

cgwalters:
> Are we talking this in context of testing this PR by injecting machine-config-daemon binary (containing this change) through ignition config during cluster install or this is something else ?

Yep!  I think that's a hard prerequisite/requirement for this.

> I tried to test this PR locally by generating manifests 

Sounds reasonable to me!

> One issue which I am seeing with my test approach is mcd fails to apply rendered config with error daemon.go:1293] couldn't parse file: missing data prefix (error message from one of running mcd pod).

Ah right; in general in the MCO I don't think we can easily support remote data sources in Ignition, becuase it breaks the "immutability" of the checksum in `rendered-worker-<sha1>`.  We don't know when remote sources change so we wouldn't know when to update.   So I think our code should encode the MCD binary directly via base64 encoding (as ugly as that is).

That said, I think I'd test this out via changing the MCS to serve the binary as that should be the final plan - we only need the MCD binary during *firstboot* - after that we've already updated to the target `machine-os-content` (and for that matter the MCD runs also as a pod) - anyways we know it's new enough.  To say this another way, we're encoding the MCD as a special "extra" in Ignition but *not* MachineConfig.

templates/common/_base/units/machine-config-daemon-firstboot.service
sinnykumari: yes, we delete `/etc/pivot/image-pullspec` during firstboot which happens after OS is upgraded to latest machine-os-content but `machine-config-daemon-firstboot.service` runs early enough and at that time image-pullspec file still exist. Once system reboots after updating OS, `/etc/pivot/image-pullspec` doesn't exist and `/etc/ignition-machine-config-encapsulated.json` exist which leads `machine-config-daemon-firstboot.service`run. That's probably because  systemd `After=ignition-firstboot-complete.service` is not enough to not run `machine-config-daemon-firstboot.service` service if `ignition-firstboot-complete.service` service failed

sinnykumari:
> Ah right; in general in the MCO I don't think we can easily support remote data sources in Ignition, becuase it breaks the "immutability" of the checksum in `rendered-worker-<sha1>`. We don't know when remote sources change so we wouldn't know when to update. So I think our code should encode the MCD binary directly via base64 encoding (as ugly as that is).

got it!

> That said, I think I'd test this out via changing the MCS to serve the binary as that should be the final plan - we only need the MCD binary during _firstboot_ - after that we've already updated to the target `machine-os-content` (and for that matter the MCD runs also as a pod) - anyways we know it's new enough.

Hmm, to start with we were thinking of having kargs day1 feature available to 4.3 and onward clusters where we can have latest machine-config-daemon binary available in installer bootimage(updating machine-config-daemon package which ships with RHCOS) .  Injecting mcd binary through ignition which is served by MCS  was next step (which I believe will also help us to support karg day1 in 4.1/4.2 based cluster)

> To say this another way, we're encoding the MCD as a special "extra" in Ignition but _not_ MachineConfig.

Right! 
I was editing MachineConfig to inject mcd binary just for testing. Once this code gets merged, we should only be updating machine-config-daemon package which should provide required functionality to apply kargs during first boot

cmd/machine-config-daemon/firstboot_complete_machineconfig.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = dn.RunFirstbootCompleteMachineconfig()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
ericavonb: nit: unnecessary error check (can do simply `return dn.RunFirstbootCompleteMachineconfig()`)

pkg/daemon/update.go
```diff
@@ -312,6 +333,38 @@ type MachineConfigDiff struct {
 	units    bool
 }
 
+// See if this was Rust we wouldn't have nullable pointers by default...
+func canonicalizeStringArray(a []string) []string {
```
kikisdeliveryservice: btw this isn't an array it's a slice.  isn't nil slice []somename vs empty canonical?

pkg/daemon/update.go
```diff
@@ -162,8 +163,35 @@ func (dn *Daemon) drainAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error)
 
 var errUnreconcilable = errors.New("unreconcilable")
 
+func canonicalizeEmptyConfig(config *mcfgv1.MachineConfig) *mcfgv1.MachineConfig {
```
kikisdeliveryservice: this could be simpler by just having a canonicalizeMC() and instead of only returning the empty config also doing a check to see if the kargs slice is empty and if so, making it a nil slice. if you did this here, you wouldn't need the other helper below - you'd just use this on both the oldConfig and newConfig immediately in update()

pkg/daemon/update.go
```diff
@@ -162,8 +163,35 @@ func (dn *Daemon) drainAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error)
...
+	if diff.IsEmpty() {
+		glog.Infof("No changes from %s to %s", oldConfigName, newConfigName)
+		return false
+	}
+	return true
+}
+
 // update the node to the provided node configuration.
 func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr error) {
+	oldConfig = canonicalizeEmptyConfig(oldConfig)
```
kikisdeliveryservice: see comment above you can just do both configs here with the func i suggested

pkg/daemon/update.go
kikisdeliveryservice: you dont really need this anyway see above.

pkg/daemon/update.go
```diff
@@ -312,6 +333,38 @@ type MachineConfigDiff struct {
...
+	return a
+}
+
+// NewMachineConfigDiff compares two MachineConfig objects.
+func NewMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) *MachineConfigDiff {
+	oldIgn := oldConfig.Spec.Config
+	newIgn := newConfig.Spec.Config
+	return &MachineConfigDiff{
+		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
+		kargs:    !reflect.DeepEqual(canonicalizeStringArray(oldConfig.Spec.KernelArguments), canonicalizeStringArray(newConfig.Spec.KernelArguments)),
```
kikisdeliveryservice: you were hitting the issue of nil vs empty slices bc you are using deepequal to compare. you can do my suggestion above. or just check the len of both slices if !=0 then deepequal else true.

cmd/machine-config-daemon/firstboot_complete_machineconfig.go
sinnykumari: Nice catch, fixed it!

pkg/daemon/update.go
sinnykumari: Removed this function in latest push

pkg/daemon/update.go
sinnykumari: I would prefer not to change anything with an existing config. Empty slice and nil being different becomes a problem only in NewMachineConfigDiff (because it uses DeepEqual), so ideally the fix should be added there itself.

pkg/daemon/update.go
sinnykumari: yeah! Added checks to consider both empty and nil slices of kargs as same and not perform deepcopy() in this case

pkg/daemon/update.go
```diff
@@ -162,8 +163,35 @@ func (dn *Daemon) drainAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error)
...
+	}
+	return &mcfgv1.MachineConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: "mco-empty-mc"},
+		Spec: mcfgv1.MachineConfigSpec{
+			Config: ctrlcommon.NewIgnConfig(),
+		},
+	}
+}
+
+func (dn *Daemon) checkUpdateDiff(oldConfig, newConfig *mcfgv1.MachineConfig) bool {
```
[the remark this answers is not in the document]
sinnykumari: Renamed to compareMachineConfig()

pkg/daemon/update.go
kikisdeliveryservice: Can we change this to canonicalizeEmptyMC? just to indicate it is a machineconfig?

pkg/daemon/update.go
kikisdeliveryservice: yay!

pkg/daemon/update.go
```diff
@@ -312,6 +333,35 @@ type MachineConfigDiff struct {
...
+	oldIgn := oldConfig.Spec.Config
+	newIgn := newConfig.Spec.Config
+
+	// Both nil and empty slices are of zero length,
+	// consider them as equal while comparing KernelArguments in both MachineConfigs
+	kargsEmpty := len(oldConfig.Spec.KernelArguments) == 0 && len(newConfig.Spec.KernelArguments) == 0
+
+	return &MachineConfigDiff{
+		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
+		kargs:    !(kargsEmpty || reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments)),
```
kikisdeliveryservice: yes perf! simple and clean.

pkg/daemon/update.go
sinnykumari: SGTM, renamed!

sinnykumari:
yay! all tests are passing.Thanks everyone for the review!

sinnykumari:
Have done another round of cluster run with mcd binary and custom payload with latest changes, looks good to me.

@cgwalters Please take a look at it, want to make sure we didn't miss anything.

runcom:
awesome
/approve

I'll leave to others on the team to lgtm but looks good otherwise!

templates/common/_base/units/machine-config-daemon-firstboot.service
cgwalters: OK, I see, though I think we should probably be deleting the `/etc/ignition-machine-config-encapsulated.json` file too.
But this `BindsTo=` approach is also fine.

PR #905, opened 2019-06-27
templates: Add proxy conf for systemd user sessions as well

cgwalters:
Followup from: https://github.com/openshift/machine-config-operator/pull/901
We want the proxies to be set up for systemd user sessions as well, as it's
quite valid to do e.g. `systemd-run --user podman ...` even if not a primary
OpenShift use case today.


runcom:
/lgtm
/approve 

Thanks for the follow up Colin

PR #906, opened 2019-06-28
Fix pkg/daemon/pivot/utils tests on macOS

mtrmac:
**- What I did**

Fixed pkg/dameon/pivot/utils Linux assumptions:

- `echo -n` is not recognized in the `/bin/sh` mode; use universal `printf`.
- `stat -c` does not exist; use `wc`.

**- How to verify it**

`make test-unit` now works for me on macOS; Linux will presumably be verified by CI.

N/A

### DIFF
--- a/pkg/daemon/pivot/utils/run_test.go
+++ b/pkg/daemon/pivot/utils/run_test.go
@@ -1,10 +1,10 @@
 package utils
 
 import (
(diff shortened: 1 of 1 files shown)


runcom:
Thanks Miloslav!

/approve
/lgtm

PR #908, opened 2019-06-30
Fix upgrades/pivoting

cgwalters:
daemon: Fix pivoting by trimming trailing newlines

Subtle API difference in the `runGetOut` that was duplicated
between the pivot and MCD codebases.  The trailing newline
wasn't expected by podman in further commands obviously.
My bad for not testing this, and doubly bad for not having CI coverage.


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1725478

PR #910, opened 2019-07-02
Bug 1726296: pkg/operator: handle missing proxy object

runcom:
**- What I did**

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-rollback-4.1-to-4.2/10

The job above failed a 4.1 -> 4.2 upgrade, logs in MCO shows:

```
E0702 05:58:10.557039       1 operator.go:311] proxy.config.openshift.io "cluster" not found
E0702 06:02:00.258227       1 operator.go:311] proxy.config.openshift.io "cluster" not found
E0702 06:05:49.965328       1 operator.go:311] proxy.config.openshift.io "cluster" not found
E0702 06:09:39.666065       1 operator.go:311] proxy.config.openshift.io "cluster" not found
E0702 06:13:29.368922       1 operator.go:311] proxy.config.openshift.io "cluster" not found
```

It means the proxy object isn't in the apiserver and we fail at getting it, disrupting the whole upgrade. Proxy comes in 4.2 but it's still valid to have no proxy during an upgrade (4.1 isn't creating a proxy object nor the upgrade is) and we shouldn't fail like that but just avoid configuring the proxy.

**- How to verify it**

4.1 -> 4.2 CI && manual 4.1 -> 4.2 upgrade

### DIFF
--- a/pkg/operator/render.go
+++ b/pkg/operator/render.go
@@ -86,13 +86,19 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 		platform = "vsphere"
 	}
 
-	return &mcfgv1.ControllerConfigSpec{
+	ccSpec := &mcfgv1.ControllerConfigSpec{
 		ClusterDNSIP:        dnsIP,
 		CloudProviderConfig: "",
 		EtcdDiscoveryDomain: infra.Status.EtcdDiscoveryDomain,
 		Platform:            platform,
 		Proxy:               &proxy.Status,
-	}, nil
+	}
+
+	if proxy != nil {
+		ccSpec.Proxy = &proxy.Status
+	}
+
+	return ccSpec, nil
 }
 
 func clusterDNSIP(iprange string) (string, error) {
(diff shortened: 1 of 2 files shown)


cgwalters:
Yeah...we are going to need to think about 4.1 :arrow_right: 4.2 for all PRs.
/approve
/lgtm

runcom:
> Yeah...we are going to need to think about 4.1 ➡️ 4.2 for all PRs.

ideally, we should have them running in our CI here in MCO as well

runcom:
I'm following up with a more robust way to bubble up this kind of errors during an operator sync, right now the co object isn't aware of what's happening..

pkg/operator/render.go
```diff
@@ -86,13 +86,19 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 		platform = "vsphere"
 	}
 
-	return &mcfgv1.ControllerConfigSpec{
+	ccSpec := &mcfgv1.ControllerConfigSpec{
 		ClusterDNSIP:        dnsIP,
 		CloudProviderConfig: "",
 		EtcdDiscoveryDomain: infra.Status.EtcdDiscoveryDomain,
 		Platform:            platform,
 		Proxy:               &proxy.Status,
```
runcom: I feel so dumb, I left this here.... I'm following up

PR #911, opened 2019-07-02
pkg/operator: bubble up errors during MCO initial sync

runcom:
Ref: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-rollback-4.1-to-4.2/10

From the above jobs the MCO isn't reporting the internal error
preventing the upgrade (or sync) to go through. This patch moves the
renderConfig setup to its own sync function in order to be accounted in
the CO object as any other sync functions.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -2,10 +2,7 @@ package operator
 
 import (
 	"encoding/base64"
-	"encoding/json"
-	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"time"
 
@@ -37,7 +34,6 @@ import (
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-	templatectrl "github.com/openshift/machine-config-operator/pkg/controller/template"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/scheme"
 	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
@@ -106,6 +102,8 @@ type Operator struct {
 	queue workqueue.RateLimitingInterface
 
 	stopCh <-chan struct{}
+
+	renderConfig *renderConfig
 }
 
 // New returns a new machine config operator.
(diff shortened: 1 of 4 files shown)


runcom:
/hold

manually testing 4.1 -> 4.2 to check if the ClusterObject isn't just stuck at old version w/o info about what's going wrong

runcom:
> manually testing 4.1 -> 4.2 to check if the ClusterObject isn't just stuck at old version w/o info about what's going wrong

it's gonna be tough to test tho...that job has never passed basically so there might be other errors, I'll follow up.

runcom:
/hold

testing testing testing

runcom:
but I can 100% reproduce on 4.1.3 -> 4.2.master

runcom:
Backing off from retesting in CI because of `VpcLimitExceeded` - reported. Will keep testing manually.

runcom:
Tested on 4.2 by hacking stuff to make the MCO fails:
```
Status:
  Conditions:
    Last Transition Time:  2019-07-02T17:58:15Z
    Message:               Cluster not available for 0.0.1-2019-07-02-162416
    Status:                False
    Last Transition Time:  2019-07-02T16:56:13Z
    Message:               Error while reconciling 0.0.1-2019-07-02-162416
    Status:                False
    Last Transition Time:  2019-07-02T17:58:15Z
    Message:               Failed to resync 0.0.1-2019-07-02-162416 because: configmap "root-ca" not found
    Reason:                FailedToSync
    Status:                True
```
machine-config                             0.0.1-2019-07-02-162416   False       False         True       2m8s
```

I'm validating 4.1 -> 4.2 w/o proxy now but this is good to go as is!

/hold cancel

```

runcom:
/hold

ehm, testing this on a cluster and it's not fully right cause it flips available instantly after the initial step, will fix

pkg/operator/sync.go
```diff
@@ -62,6 +72,97 @@ func (optr *Operator) syncAll(rconfig *renderConfig, syncFuncs []syncFunc) error
 	return agg
 }
 
+func (optr *Operator) syncRenderConfig(config *renderConfig) error {
```
cgwalters: It's confusing that this function takes a `config` as a parameter but really it will be either the previous version or `nil`.  Maybe cleaner to separate it out of the other sync funcs array and have it take `()`?

pkg/operator/operator.go
```diff
@@ -321,108 +321,20 @@ func (optr *Operator) sync(key string) error {
...
+		// for the sync funcs below. This is the only function that prevents the others from even running also.
+		{"render-config", optr.syncRenderConfig},
 		{"pools", optr.syncMachineConfigPools},
 		{"mcd", optr.syncMachineConfigDaemon},
 		{"mcc", optr.syncMachineConfigController},
 		{"mcs", optr.syncMachineConfigServer},
 		{"required-pools", optr.syncRequiredMachineConfigPools},
 	}
-	return optr.syncAll(rc, syncFuncs)
+	return optr.syncAll(optr.renderConfig, syncFuncs)
```
cgwalters: Feels like this should really be:

```
if err := optr.SyncRenderConfig(); err != nil {
    return err
}
return optr.syncAll(syncFuncs)
```

And we use `optr.renderConfig` in the sync funcs or so?

runcom: the thing is syncRenderConfig must follow operator versions settings as well, that's why it's in the array.

also, to reply to https://github.com/openshift/machine-config-operator/pull/911#discussion_r299654948 - it looks weird I know but we do already have functions taking renderConfig but not using it e.g.: `syncRequiredMachineConfigPools` and that's why we have the function signagture typed I guess

```
func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
```

runcom: also, wiring syncRenderConfig _outside_ of syncAll to respect all the Status flipping is really non-trivial (that's why I went this way)

runcom: also, the other sync functions are gated on this first func to run anyway

pkg/operator/sync.go
runcom: see your other comment

cgwalters:
Right, when I looked closer I see why you did it this way.
/approve

runcom:
I'm removing the lift once the cluster I'm testing 4.1->4.2 finishes correct and I run some other smoke tests

pkg/operator/sync.go
```diff
@@ -62,6 +72,97 @@ func (optr *Operator) syncAll(rconfig *renderConfig, syncFuncs []syncFunc) error
...
+	}
+	etcdMetricCA, err := optr.getCAsFromConfigMap("openshift-config", "etcd-metric-serving-ca", "ca-bundle.crt")
+	if err != nil {
+		return err
+	}
+	rootCA, err := optr.getCAsFromConfigMap("kube-system", "root-ca", "ca.crt")
+	if err != nil {
+		return err
+	}
+	// as described by the name this is essentially static, but it no worse than what was here before.  Since changes disrupt workloads
```
kikisdeliveryservice: i know that this comment predates this PR but I dont quite understand what it is trying to say?

runcom: worth asking @deads2k maybe

runcom:
/hold cancel

works as expected now and fixed the ordering gating actually

cgwalters:
> pkg/operator: fail fast if a sync func fails 

I'm guessing here that it's to ensure we make progress if there are any dependencies.  Offhand...I don't think there are here.

runcom:
> I'm guessing here that it's to ensure we make progress if there are any dependencies. Offhand...I don't think there are here.

syncRenderConfig is a dep to the others and MCD is a dep for MCC for instance, but I didn't fully get the comment 🤔 

/retest

pkg/operator/render.go
```diff
@@ -91,7 +91,6 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 		CloudProviderConfig: "",
 		EtcdDiscoveryDomain: infra.Status.EtcdDiscoveryDomain,
 		Platform:            platform,
-		Proxy:               &proxy.Status,
```
cgwalters: Stray change from the other PR here?

runcom: yup, but didn't want to rebase since this won't be a conflict once it merges I guess 🤔

runcom:
W** is happening here

PR #912, opened 2019-07-02
templates: only add proxy to crio, kubelet and machine-config-daemon-host

runcom:
**- What I did**

Following discussion with @mrunalp we want to enable proxy only for kubelet, crio and machine-config-daemon-host (which manages pivot)

**- How to verify it**


runcom:
/hold

actually, pivot needs to be using this as well and we have a service we can inject envs to

runcom:
> actually, pivot needs to be using this as well and we have a service we can inject envs to. I think all the services we ship in MCD need this override?

added envs for `machine-config-daemon-host` as well

/hold cancel

mrunalp:
:+1:

cgwalters:
I guess it's easier to be conservative to start, though I wonder if at some point we should gain a knob that's just "our whitelisted service" vs "everything".
/lgtm

PR #913, opened 2019-07-02
pkg/operator: actually do not set proxy if empty...

runcom:
**- What I did**

I've been silly here https://github.com/openshift/machine-config-operator/pull/910/files#r299605265, shame on me, the previous PR works but makes the MCO panic and fail the upgrade anyway

Fix that

**- How to verify it**


sgreene570:
Thanks for quick fix :pray:

PR #916, opened 2019-07-03
pkg/controller: Handle errors for cascadeDelete

JAORMX:
cascadeDelete for the container-runtime-config-controller and the
kubelet-config-controller can return errors, so lets handle those
(or at least issue the appropriate error message for visibility).

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -232,8 +232,11 @@ func (ctrl *Controller) deleteContainerRuntimeConfig(obj interface{}) {
 			return
 		}
 	}
-	ctrl.cascadeDelete(cfg)
-	glog.V(4).Infof("Deleted ContainerRuntimeConfig %s and restored default config", cfg.Name)
+	if err := ctrl.cascadeDelete(cfg); err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't delete object %#v: %v", cfg, err))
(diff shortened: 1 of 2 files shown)


runcom:
cc @rphilips

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -232,8 +232,11 @@ func (ctrl *Controller) deleteContainerRuntimeConfig(obj interface{}) {
 			return
 		}
 	}
-	ctrl.cascadeDelete(cfg)
-	glog.V(4).Infof("Deleted ContainerRuntimeConfig %s and restored default config", cfg.Name)
+	if err := ctrl.cascadeDelete(cfg); err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't delete object %#v: %v", cfg, err))
```
cgwalters: In new code let's please use `errors.Wrapf` like `errors.Wrapf(err, "couldn't delete object %s", cfg.Name)` because in the past I and others have forgotten to do the trailing `%v, err` part.

runcom:
/approve

This code is @umohnani8 and @rphillips to lgtm

umohnani8:
LGTM

pkg/controller/container-runtime-config/container_runtime_config_controller.go
cgwalters: @JAORMX did you see this comment?

PR #919, opened 2019-07-03
CI: Add gosec to make verify

JAORMX:
While golangci-lint runs gosec as well; the way they did it was by
forking the gosec code and importing it. The issue is that they
haven't updated the fork in almost a year.

So to get the latest gosec, lets call it explicitly.

This depends on  openshift/release#4273

If golangci/golangci-lint#597 gets fixed, we could get rid of this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,8 @@ verify:
 	@which golangci-lint 2> /dev/null >&1 || { echo "golangci-lint must be installed to lint code";  exit 1; }
 	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
 	golangci-lint run
+	# Remove once https://github.com/golangci/golangci-lint/issues/597 is
+	# addressed
 	gosec -severity high --confidence medium -exclude G204 -quiet ./...
 	hack/verify-codegen.sh
 	hack/verify-generated-bindata.sh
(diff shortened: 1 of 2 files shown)


JAORMX:
@runcom could you check this out?

runcom:
see https://github.com/openshift/release/pull/4273#issuecomment-508102952

Makefile
```diff
@@ -51,7 +51,9 @@ update:
 verify:
 	@which go-bindata 2> /dev/null >&1 || { echo "go-bindata must be installed to verify generated code";  exit 1; }
 	@which golangci-lint 2> /dev/null >&1 || { echo "golangci-lint must be installed to lint code";  exit 1; }
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
 	golangci-lint run
+	gosec -severity high --confidence medium -exclude G204 -quiet ./...
```
runcom: let's add a comment pointing to https://github.com/golangci/golangci-lint/issues/597 to make sure we get rid of this once golangci-lint updates gosec

JAORMX: Done.

runcom: uhm, did you push? 😄

JAORMX: lol, yep. NOW I pushed :smiley:

PR #920, opened 2019-07-03
pkg/operator: clarify MCO images and use better names

runcom:
Honestly finding images in the struct confusing as it's really the
collection of two different sets of images (renderConfig and
ControllerConfig).
This also led to some hassles around changing images in #899 for
instance.
This patch adds some comments around the scope of the images, split the
set in two different ones (still tied to the original struct) and change
the template keys used to substite.

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -85,15 +85,14 @@ type ControllerConfigSpec struct {
 	// on all machines.
 	PullSecret *corev1.ObjectReference `json:"pullSecret,omitempty"`
 
-	// images is the map of images that are used by the controller,
-	// i.e. images for etcd Environment Setup, etcd, Kube Client Agent, and Infra containers.
+	// images is map of images that are used by the controller to render templates under ./templates/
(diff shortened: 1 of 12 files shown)


runcom:
/hold

we'll get back to this asap

kikisdeliveryservice:
I love this and it would have save be so much time and staring at open tabs.. :heart:

sgreene570:
🙏🙏🙏

runcom:
/hold cancel

rebased and hopefully fixed

runcom:
ready and green now

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -85,8 +85,7 @@ type ControllerConfigSpec struct {
 	// on all machines.
 	PullSecret *corev1.ObjectReference `json:"pullSecret,omitempty"`
 
-	// images is the map of images that are used by the controller,
-	// i.e. images for etcd Environment Setup, etcd, Kube Client Agent, and Infra containers.
+	// Images is map of images that are used by the controller to render templates under ./templates/
```
LorbusChris: images at the start of the comment must correspond to the json name, so it has be lowercase here.

PR #921, opened 2019-07-03
Two cleanup patches to tests/node

cgwalters:


pkg/controller/node/node_controller_test.go
```diff
@@ -157,29 +155,17 @@ func checkAction(expected, actual core.Action, t *testing.T) {
 		e, _ := expected.(core.CreateAction)
 		expObject := filterLastTransitionTime(e.GetObject())
 		object := filterLastTransitionTime(a.GetObject())
-
-		if !equality.Semantic.DeepEqual(expObject, object) {
-			t.Errorf("Action %s %s has wrong object\nDiff:\n %s",
-				a.GetVerb(), a.GetResource().Resource, diff.ObjectGoPrintDiff(expObject, object))
-		}
+		assert.Equal(t, expObject, object)
```
kikisdeliveryservice: Migth be worth including the diff as an args to asset.Equal since that's a helpful error message.

cgwalters: Note this in the commit message:

> ... does diffs automatically, etc.

Try it!  The diffs output are pretty nice.

This one you quoted indeed tries to make a more useful error message but...it's IMO not worth the extra verbosity, and trying to do that consistently requires a *lot* more typing when adding new unit tests.

kikisdeliveryservice: > Note this in the commit message:
> 
> > ... does diffs automatically, etc.
> 
> Try it! The diffs output are pretty nice.
> 
I'll have to check this out!  Missed the commit message. :)

cgwalters:
Any further comments on this?

PR #922, opened 2019-07-03
two small MCD patches

cgwalters:


runcom:
`daemon: Don't check our own state flag for update check` I thought I changed that somewhere else so it makes total sense 👍 

/approve

runcom:
I'm starting to think the op error is a valid error? 🤔

cgwalters:
Yeah, we may need to always do "if current == desired, set state to done".

cgwalters:
> Yeah, we may need to always do "if current == desired, set state to done".

Dropped that patch for now, this one is just two trivial patches now.

kikisdeliveryservice:
i think it's gonna be one of those wacky ci afternoons...

/retest

PR #924, opened 2019-07-03
test/e2e: More use of assert/require

cgwalters:
I actually just discovered the `require` package is what I really
wanted, make use of it in more places so we consistently panic
rather than stumbling on in some cases, and we don't need to duplicate
`err`, it's 1 line instead of 3 etc.

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -36,15 +36,11 @@ func TestMCDToken(t *testing.T) {
 	}
 
 	mcdList, err := cs.Pods("openshift-machine-config-operator").List(listOptions)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/mcd_test.go
```diff
@@ -36,15 +36,11 @@ func TestMCDToken(t *testing.T) {
 	}
 
 	mcdList, err := cs.Pods("openshift-machine-config-operator").List(listOptions)
-	if err != nil {
-		t.Fatalf("%#v", err)
-	}
+	require.Nil(t, err)
```
ericavonb: you can also use `require.NoError(t, err)`, but simpler tests 👍

PR #928, opened 2019-07-04
manifests/machineconfig.crd: set a minimal openAPIV3Schema

### LINKED ISSUE #777: MC spec must be required for oc create
Haven't dug enough into this but today it's possible to successfully create the following (bad) MC:

```yaml
kind: MachineConfig
metadata:
 labels:
 name: 99-rollback-test
spec:
config:
 ignition:
   version: 2.2.0
 storage:
   files:
   - contents:
       source: data:,A
     filesystem: root
     mode: 420
     path: /etc/rollback-test
```

It's wrongly formatted for the spec, which results in the MCC refusing to generate a new rendered one (which is good!). Still, we should have a way to prevent the creation (rolling back the above just requires oc delete badmc also).

For instance, if you create a pod w/o the spec.containers field, you'll get:
```
$ oc create -f test.yaml 
The Pod "nginx" is invalid: spec.containers: Required value
```

vrutkovs:
**- What I did**
Added a minimal openAPI spec definition to MachineConfig CRD to ensure it gets (minimally) validated when it is being created or changed

**- How to verify it**
Attempt to create an MC with an empty spec -> error message
Create MC with a valid spec -> pass

Validate MachineConfigs have `spec:` field and add description to the ignition subset we support

Fixes #777

### DIFF
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -39,3 +39,186 @@ spec:
     # shortNames allow shorter string to match your resource on the CLI
     shortNames:
     - mc
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        apiVersion:
+          description: "APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources"
+          type: string
+        kind:
+          description: "Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds"
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: "spec hold the intent of how this operator should behave"
+          type: object
+          properties:
+            config:
+              description: "config contains options related to the configuration"
+              type: object
+              properties:
+                ignition:
+                  description: "ignition section contains metadata about the configuration itself"
+                  type: object
+                  properties:
+                    version:
+                      description: "version string is the semantic version number of the spec"
+                      type: string
+                storage:
+                  description: "storage describes the desired state of the system's storage devices"
+                  type: object
+                  properties:
+                    directories:
+                      description: "directories is the list of directories to be created"
+                      type: array
+                      items:
+                        description: "items is list of directories to be written"
+                        type: object
+                        properties:
+                          filesystem:
+                            description: "filesystem is the internal identifier of the filesystem in which to write the file. This matches the last filesystem with the given identifier"
+                            type: string
+                          mode:
+                            description: "mode is the file's permission mode. Note that the mode must be properly specified as a decimal value (i.e. 0644 -> 420)"
+                            type: integer
+                          path:
+                            description: "path is the absolute path to the file"
+                            type: string
+                          user:
+                            description: "user object specifies the file's owner"
+                            type: object
+                            properties:
+                              id:
+                                description: "id is the user ID of the owner"
+                                type: integer
+                              name:
+                                description: "name is the user name of the owner"
+                                type: string
+                          group:
+                            description: "group object specifies group of the owner"
+                            type: object
+                            properties:
+                              id:
+                                description: "id specifies group ID of the owner"
+                                type: integer
+                              name:
+                                description: "name is the group name of the owner"
+                                type: string
+                          overwrite:
+                            description: "overwrite specifies whether to delete preexisting nodes at the path"
+                            type: boolean
+                    files:
+                      description: "files is the list of files to be created"
+                      type: array
+                      items:
+                        description: "items is list of files to be written"
+                        type: object
+                        properties:
+                          contents:
+                            description: "contents specifies options related to the contents of the file"
+                            type: object
+                            properties:
+                              compression:
+                                description: "the type of compression used on the contents (null or gzip). Compression cannot be used with S3."
+                                type: string
+                              source:
+                                description: "source is the URL of the file contents. Supported schemes are http, https, tftp, s3, and data. When using http, it is advisable to use the verification option to ensure the contents haven't been modified."
+                                type: string
+                              verification:
+                                description: "verification specifies options related to the verification of the file contents"
+                                type: object
+                                properties:
+                                  hash:
+                                    description: "hash is the hash of the config, in the form <type>-<value> where type is sha512"
+                                    type: string
+                          filesystem:
+                            description: "filesystem is the internal identifier of the filesystem in which to write the file. This matches the last filesystem with the given identifier"
+                            type: string
+                          mode:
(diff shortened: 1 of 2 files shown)


runcom:
> Validate MachineConfigs have `spec:` field

is this also validating config.Ignition.Spec within a spec?

Thanks . for this Vadim, been looking for this forever 😄 🙏

runcom:
ref: https://github.com/openshift/machine-config-operator/issues/930

manifests/machineconfig.crd.yaml
```diff
@@ -39,3 +39,208 @@ spec:
...
+                            items:
+                              description: "items describes unit dropin"
+                              type: object
+                              properties:
+                                contents:
+                                  description: "contents is the contents of the drop-in"
+                                  type: string
+                                name:
+                                  description: "name is the name of the drop-in. This must be suffixed with '.conf'."
+                                  type: string
```
runcom: do we need:
```
required:
      - spec
```
to make it error out on `oc create`? (following https://github.com/openshift/cluster-version-operator/pull/168/files#diff-9b1e80d552a6692388b5e226ebf4c02cR137)

vrutkovs: Right, I'll add more `requires` there

runcom: I think spec is enough - `config`, `kargs`, `fips` can be `anyOf`

vrutkovs: Lets leave it like this for now - IIUC `anyOf` and detailed type validation is 1.15 feature and not yet enabled in 4.2

manifests/machineconfig.crd.yaml
```diff
@@ -39,3 +39,210 @@ spec:
...
+                              name:
+                                description: "name is the group name of the owner"
+                                type: string
+                          overwrite:
+                            description: "overwrite specifies whether to delete preexisting nodes at the path"
+                            type: boolean
+                          append:
+                            description: "append specifies whether to append to the specified file. Creates a new file if nothing exists at the path. Cannot be set if overwrite is set to true."
+                            type: boolean
+                    links:
```
runcom: we don't support links, this whole section can go away

runcom: we need also:

```diff
+            kargs:
+              description: ""
+              type: array
+              items:
+                description: ""
+                type: string
+            fips:
+              description: ""
+              type: boolean
```

LorbusChris:
Note: If we move to converting ign.Config to RawExtension for our external API, I don't think we could validate the RawExtension this way. However, we might end up not doing that move. Let's discuss this further next week?

manifests/machineconfig.crd.yaml
```diff
@@ -39,3 +39,183 @@ spec:
...
+                              description: "items describes unit dropin"
+                              type: object
+                              properties:
+                                contents:
+                                  description: "contents is the contents of the drop-in"
+                                  type: string
+                                name:
+                                  description: "name is the name of the drop-in. This must be suffixed with '.conf'."
+                                  type: string
+                kargs:
```
runcom: kargs and fips are part of the MachineConfigSpec, not the ignition config (unless I'm messing up with indentation)

runcom:
manually testing this out

runcom:
uhm, something is off:
```
15:03:46 [github.com/openshift/installer] ‹master*› oc explain machineconfigs.machineconfiguration.openshift.io                                                                                                 
error: field "machineconfiguration" does not exist
```

/hold

vrutkovs:
Weird, it shows 
```
$ oc explain machineconfigs.machineconfiguration.openshift.io
KIND:     MachineConfig
VERSION:  machineconfiguration.openshift.io/v1

DESCRIPTION:
     <empty>
```
here

/retest

runcom:
> Weird, it shows

that is shown to me _before_ this PR - after I install a cluster with this PR's payload I get that weird error

vrutkovs:
Oh, right, can reproduce now

manifests/machineconfig.crd.yaml
```diff
@@ -39,3 +39,186 @@ spec:
...
+              description: "kernel argument"
+              type: string
+          fips:
+            description: "fips controls FIPS mode"
+            type: boolean
+          osImageURL:
+            description: "osImageURL specifies the remote location that will be used to fetch the OS"
+            type: string
+      required:
+      - spec
```
sttts: you require spec. So an empty object will not work.

runcom: we don't want an empty object to be created

sttts: misread Slack. Nvmd.

vrutkovs: we only require spec really, other items in spec are optional (@runcom 	correct me if I'm wrong)

runcom: yep - and as you pointed out Vadim, we're gonna use something like `anyOf` once available

vrutkovs:
```
$ oc explain --api-version=machineconfiguration.openshift.io/v1 machineconfigs                                                                     
KIND:     MachineConfig
VERSION:  machineconfiguration.openshift.io/v1

DESCRIPTION:
     <empty>

FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#resources

   kind	<string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds

   metadata	<Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata

   spec	<Object> -required-
     spec hold the intent of how this operator should behave
```
Not sure why `oc explain machineconfigs.machineconfiguration.openshift.io` fails, sounds like a bug / bad error message to me

sttts:
> Not sure why `oc explain machineconfigs.machineconfiguration.openshift.io` fails, sounds like a bug / bad error message to me

No. The syntax is `oc explain --api-version=machineconfiguration.openshift.io/v1 machineconfigs.spec` to dive into the type.

runcom:
> No. The syntax is `oc explain --api-version=machineconfiguration.openshift.io/v1 machineconfigs.spec` to dive into the type.

so this works as expected? cool! (still not sure about the error above, looks wrong/weird)

/approve
/assign LorbusChris

kikisdeliveryservice:
@vrutkovs can you squash some of these commits?

vrutkovs:
Rebased and squashed

kikisdeliveryservice:
thanks @vrutkovs !

re-adding:
/lgtm

damemi:
I've opened https://github.com/openshift/machine-config-operator/pull/955 which should help with generating the manifest validation

kikisdeliveryservice:
@vrutkovs once #957 merges, you'll probably have to pull that in to get `test verify` passing.

PR #935, opened 2019-07-08
Prep commits for having the MCD process encapsulated MachineConfig

cgwalters:
Prep commits for https://github.com/openshift/machine-config-operator/pull/904

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -634,7 +634,7 @@ type stateAndConfigs struct {
 }
 
 // getStateAndConfigs captures initial state of the MCD/node
-func (dn *Daemon) getStateAndConfigs() (*stateAndConfigs, error) {
+func (dn *Daemon) getStateAndConfigs(bootstrapping bool) (*stateAndConfigs, error) {
 	pendingConfigName, err := dn.getPendingState()
 	if err != nil {
 		return nil, err
@@ -646,18 +646,6 @@ func (dn *Daemon) getStateAndConfigs() (*stateAndConfigs, error) {
 		if err != nil {
 			return nil, err
 		}
-	}
-	_, err = os.Lstat(constants.InitialNodeAnnotationsFilePath)
-	var bootstrapping bool
-	if err != nil {
-		if !os.IsNotExist(err) {
-			return nil, err
-		}
-		// The node annotation file (laid down by the MCS)
-		// doesn't exist, we must not be bootstrapping
-	} else {
-		bootstrapping = true
-		glog.Info("In bootstrap mode")
 	}
 
 	currentConfigName, err := getNodeAnnotation(dn.node, constants.CurrentMachineConfigAnnotationKey)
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -803,6 +791,53 @@ func (dn *Daemon) storeCurrentConfigOnDisk(current *mcfgv1.MachineConfig) error
...
+	if err == nil && ccAnnotation != "" {
+		return false, nil
+	}
+
+	glog.Infof("No %s annotation on node %s: %v, in cluster bootstrap, loading initial node annotation from %s", constants.CurrentMachineConfigAnnotationKey, dn.node.Name, dn.node.Annotations, constants.InitialNodeAnnotationsFilePath)
+
+	d, err := ioutil.ReadFile(constants.InitialNodeAnnotationsFilePath)
+	if err != nil && !os.IsNotExist(err) {
+		return false, fmt.Errorf("failed to read initial annotations from %q: %v", constants.InitialNodeAnnotationsFilePath, err)
+	}
```
runcom: super silly nit: shouldn't we return `true` from now on here? I'm just worried in the future we're gonna change the outer logic and might miss the bootstrap bool

pkg/daemon/update_test.go
```diff
@@ -453,21 +451,3 @@ func checkIrreconcilableResults(t *testing.T, key string, reconcilableError erro
...
-	require.Nil(t, d.reboot("", 0, nil))
-
-	// skipReboot in normal cluster run is just a no-op and we reboot anyway
-	d = &Daemon{
-		mock:       true,
-		onceFrom:   "",
-		skipReboot: true,
-	}
-	require.NotNil(t, d.reboot("", 0, exec.Command("true")))
-}
```
runcom: I'm still reluctant to remove a test however easy and straightforward it is...if someone messes up with the skipReboot `if` this was going to catch it

cgwalters: We [talked about these unit tests before](https://github.com/openshift/machine-config-operator/pull/862#issuecomment-503264620) right?

runcom: yup, no blocking, was just pointing out my initial reaction when a test gets deleted :D

pkg/daemon/update.go
```diff
@@ -169,14 +169,18 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
-	// reason is that there is a good chance a previous state is not available to match against.
-	if oldConfig.Kind == "" && dn.onceFrom != "" {
-		glog.Info("Missing kind in old config. Assuming no prior state.")
+	oldConfigUnset := oldConfig == nil
+	if oldConfigUnset {
+		// Rather than change the rest of the code to deal
+		// with a nil oldConfig, just pass an empty one in
+		// most places.
+		emptyMC := mcfgv1.MachineConfig{}
+		oldConfig = &emptyMC
```
ericavonb: note for next time (don't have to fix in this pr): can remove `emptyMC` declaration and just do `oldConfig = &mcfgv1.MachineConfig{}` directly

pkg/daemon/daemon.go
ericavonb: Would that would apply for the `json.Unmarshal`  and `setNodeAnnotations(..., initial)` error cases as well?

pkg/daemon/daemon.go
```diff
@@ -790,6 +793,53 @@ func (dn *Daemon) storeCurrentConfigOnDisk(current *mcfgv1.MachineConfig) error
...
+	if os.IsNotExist(err) {
+		// try currentConfig if, for whatever reason we lost annotations? this is super best effort.
+		currentOnDisk, err := dn.getCurrentConfigOnDisk()
+		if err == nil {
+			glog.Infof("Setting initial node config based on current configuration on disk: %s", currentOnDisk.GetName())
+			_, err := setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.node.Name, map[string]string{constants.CurrentMachineConfigAnnotationKey: currentOnDisk.GetName()})
```
ericavonb: why don't we update the cached copy `dn.node = n` like in the other case?

pkg/daemon/daemon.go
runcom: yes, from this point it should be safe to return true

pkg/daemon/daemon.go
```diff
@@ -1278,15 +1312,16 @@ func ValidPath(path string) bool {
 // senseAndLoadOnceFrom gets a hold of the content for supported onceFrom configurations,
 // parses to verify the type, and returns back the genericInterface, the type description,
 // if it was local or remote, and error.
-func (dn *Daemon) senseAndLoadOnceFrom() (interface{}, onceFromOrigin, error) {
+func (dn *Daemon) senseAndLoadOnceFrom(onceFrom string) (interface{}, onceFromOrigin, error) {
```
sinnykumari: Just wondering, since this function is not utilizing Daemon struct anymore, is there a reason to keep it part of Daemon ?

runcom: not really, we can remove the object reference to Daemon as a cleanup

sinnykumari:
@cgwalters there are some conflict with latest master, can we get this PR rebased?

cgwalters:
Eh, the reboot parts of this were merged, and there's been a lot of code churn since then that makes rebasing this tricky/painful.  I'm just going to try putting everything in https://github.com/openshift/machine-config-operator/pull/904 instead.

PR #937, opened 2019-07-08
Watch scheduler CR and make masters schedulable accordingly

ravisantoshgudimetla:
**- What I did**
Added  logic in node controller to watch for scheduler CR changes and update the master nodes to make them schedulable and unschedulable

**- How to verify it**

~~I did not test it. I want to check if the approach seems reasonable before I proceed further. Please review the https://github.com/openshift/machine-config-operator/commit/c82786d5061928080031eb83274d94601451cf4d~~

The e2e test verifies if the worker label has been added and master taint has been removed for all the masters when scheduler CR's `mastersSchedulable` field has been set or unset. That field was added to Scheduler CRD here https://github.com/openshift/api/pull/366


pkg/controller/node/node_controller.go
```diff
@@ -97,7 +107,9 @@ func New(
 		UpdateFunc: ctrl.updateNode,
 		DeleteFunc: ctrl.deleteNode,
 	})
-
+	schedulerInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
ravisantoshgudimetla: Need to add event handlers for Delete and Create

pkg/controller/node/node_controller.go
```diff
@@ -97,7 +107,9 @@ func New(
...
+		UpdateFunc: ctrl.updateMasterNodes,
```
runcom: I've a doubt and currently thinking Colin meant to wire all this into the current `sync` handler to couple this with the other upgrade decisions (but this might as well work out correctly)

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
 	<-stopCh
 }
 
+func (crtl *Controller) updateMasterNodes(old, cur interface{}) {
+	oldScheduler := old.(*configv1.Scheduler)
+	curScheduler := cur.(*configv1.Scheduler)
+
+	if oldScheduler.Name != "cluster" && curScheduler.Name != "cluster" {
```
cgwalters: I think we want `||` here.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+				if t.Key == "node-role.kubernetes.io/master" && t.Effect == corev1.TaintEffectNoSchedule {
+					continue
+				}
+				newTaints = append(newTaints, t)
+			}
+			//
+			node.Spec.Taints = newTaints
+		})
+		if err != nil {
+			glog.Errorf("error making node %v schedulable with error %v", err)
```
cgwalters: As a general rule, rather than logging and continuing, please return the error and let the higher level logic retry.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+			}
+			node.Labels = newLabels
+			// Add master taint
+			newTaints := node.Spec.Taints
+			MasterUnSchedulableTaint := corev1.Taint{Key: "node-role.kubernetes.io/master", Effect: corev1.TaintEffectNoSchedule}
+			newTaints = append(newTaints, MasterUnSchedulableTaint)
+			node.Spec.Taints = newTaints
+		})
+		if err != nil {
+			glog.Errorf("error making node %v unschedulable with error %v", err)
```
cgwalters: Ditto.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+		f(nodeClone)
+
+		newNode, err := json.Marshal(nodeClone)
+		if err != nil {
+			return err
+		}
+
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldNode, newNode, corev1.Node{})
+		if err != nil {
+			return fmt.Errorf("failed to create patch for node %q: %v", nodeName, err)
```
cgwalters: See https://github.com/openshift/machine-config-operator/pull/916#discussion_r300063695

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+		// may be conflict if max retries were hit
+		return nil, fmt.Errorf("unable to update node %q: %v", node, err)
+	}
+	return node, nil
+}
+
+func (crtl *Controller) getCurrentMasters() []*corev1.Node {
+	nodeList, err := crtl.nodeLister.List(labels.SelectorFromSet(labels.Set{"node-role.kubernetes.io/master": ""}))
+	if err != nil {
+		glog.Errorf("error while listing master nodes %v", err)
```
cgwalters: This seems like a particularly egregious "log and continue" case.  While it's *probably* safe to have the rest of the code operate on the empty array in an error situation, why even worry about it?  Let's error out early and avoid thinking about it.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+	}
+	desiredSchedulableMasters := curScheduler.Spec.PercentageOfMastersSchedulable
+	currentMasters := crtl.getCurrentMasters()
+	currentMasterCount := len(currentMasters)
+	validateDesiredSchedulableMasterField(desiredSchedulableMasters, currentMasterCount)
+
+	currentSchedulableMasters, currentUnSchedulableMasters := getSchedulableAndUnSchedulabeMasterNodes(currentMasters)
+	desiredNumberOfMastersSchedulable, err := intstrutil.GetValueFromIntOrPercent(intstrutil.ValueOrDefault(&curScheduler.Spec.PercentageOfMastersSchedulable, intstrutil.FromInt(0)), currentMasterCount, false)
+	if err != nil {
+		glog.Errorf("error converting integer")
```
cgwalters: This hides the underlying error message; exactly why I prefer `errors.Wrapf`.

pkg/controller/node/node_controller.go
cgwalters: Yeah...indeed that was my original thought; if they're in separate handlers then it's possible that at least transiently we are violating the max schedulable.  But OTOH Kube is all about reconciliation/eventual consistency and the window for the labels being out of sync would likely be very small.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+// The acceptable values currently are 100% or 0%. In future, we will relax this restriction.
+// TODO: @ravig remove this validation when we relax this restriction and this code could be in api validation when
+// it doesn't depend on number of master nodes
+func validateDesiredSchedulableMasterField(desiredSchedulableMasters intstrutil.IntOrString, currentMasterNodesCount int) {
+	switch desiredSchedulableMasters.Type {
+	case intstrutil.String:
+		maxMastersToBeSchedulable := "100%"
+		minMastersToBeSchedulable := "0%"
+		if desiredSchedulableMasters != intstrutil.FromString(maxMastersToBeSchedulable) || desiredSchedulableMasters != intstrutil.FromString(minMastersToBeSchedulable) {
+			glog.Errorf("Acceptable values for percentangeOfMastersSchedulable are 100 and 0 but found %v", desiredSchedulableMasters)
```
cgwalters: Should return `error`.

pkg/controller/node/node_controller.go
```diff
@@ -128,6 +143,199 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+			return
+		}
+	case intstrutil.Int:
+		if desiredSchedulableMasters.IntValue() == 0 {
+			desiredSchedulableMasters = intstrutil.FromInt(0)
+		}
+		// If this value is greater than no of master nodes, so set it to all the master available.
```
cgwalters: Comment is a lie...this is basically always assigning it to `currentMasterNodesCount` right?

pkg/controller/node/node_controller.go
ravisantoshgudimetla: I thought about this but in that case, we need to have scheduler lister list CRs continuously. My past experience with this CR listing continuously was that it causes a performance impact(a hot read loop). Also, out of sync can happen in even in that case where the update happened and sync loop just completed, we'd have to wait for next sync loop for changes to come into effect but the eventual consistency model as @cgwalters pointed out would still hold good :).

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Yup. Thanks for catching. I will update it.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Sure. Let me look into `errors.Wrapf`. BTW, we can also verify checks in CI which capture this.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: In this case, it might be different - we cannot return error when it fails for a single node, perhaps we need to capture slice of errors for individual nodes and return. I don't have a strong preference but I see your point.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: As I told earlier, I have seen both the patterns. Either log and continue or return early with an error. I agree that in this case, perhaps error-ing early might help. As I told this is rough shape of what you can see, not the exact fn.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Sure.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: This is copied code and function will be removed to a neutral place soon.

manifests/machineconfigcontroller/clusterrole.yaml
```diff
@@ -14,5 +14,5 @@ rules:
   resources: ["configmaps", "secrets"]
   verbs: ["*"]
 - apiGroups: ["config.openshift.io"]
-  resources: ["images", "clusterversions", "featuregates"]
+  resources: ["images", "clusterversions", "featuregates", "schedulers"]
```
ravisantoshgudimetla: This can be changed futher to just list, instead of giving `*`

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +153,235 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+	return errors
+}
+
+// makeMastersUnSchedulable makes all the masters in the cluster unschedulable
+func (ctrl *Controller) makeMastersUnSchedulable(currentMasters []*corev1.Node) []error {
+	errors := []error{}
+	for _, node := range currentMasters {
+		if !checkMasterIsAlreadySchedulable(node) {
+			continue
+		} else {
```
runcom: you can drop the else?

runcom:
@ravisantoshgudimetla this patch looks good - is there some testing I can perform manually?

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +153,235 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+	return v1helpers.NewMultiLineAggregate(errors)
+}
+
+// makeMastersSchedulable makes all the masters in the cluster schedulable
+func (ctrl *Controller) makeMastersSchedulable(currentMasters []*corev1.Node) []error {
+	errors := []error{}
+	for _, node := range currentMasters {
+		if CheckMasterIsAlreadySchedulable(node) {
+			continue
+		} else {
```
runcom: this else is superflous since you're `continue`'ing above right?

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +153,235 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+	_, err := updateNodeRetry(ctrl.kubeClient.CoreV1().Nodes(), ctrl.nodeLister, node.Name, func(node *corev1.Node) {
+		// Remove worker label
+		newLabels := node.Labels
+		if _, hasWorkerLabel := newLabels[workerLabel]; hasWorkerLabel {
+			delete(newLabels, workerLabel)
+		}
+		node.Labels = newLabels
+		// Add master taint
+		newTaints := node.Spec.Taints
+		MasterUnSchedulableTaint := corev1.Taint{Key: masterLabel, Effect: corev1.TaintEffectNoSchedule}
```
runcom: lowercase var

runcom:
```
--- FAIL: TestMastersSchedulable (1.15s)
    mco_test.go:52: Error while updating scheduler CR
```

test/e2e/mco_test.go
```diff
@@ -27,3 +31,59 @@ func TestClusterOperatorRelatedObjects(t *testing.T) {
...
+
+func TestMastersSchedulable(t *testing.T) {
+	cs := framework.NewClientSet("")
+	schedulerCR, err := cs.ConfigV1Interface.Schedulers().Get("cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal("Error while listing scheduler CR")
+	}
+	schedulerCR.Spec.MastersSchedulable = true
+	if _, err = cs.ConfigV1Interface.Schedulers().Update(schedulerCR); err != nil {
+		t.Fatal("Error while updating scheduler CR")
```
runcom: let's `Fatalf` here and elsewhere to log the error we got as well - e2e is failing right now and this is not providing further info about what's happening

cgwalters: See also other changes to use `require.NoError(t, error)` which enforces that.

pkg/controller/node/node_controller.go
cgwalters: > we cannot return error when it fails for a single node, 

Why not?  

> perhaps we need to capture slice of errors for individual nodes and return.

Yes, that's a better pattern indeed, but I think "fast fail for single item" is still better than "log and continue".

test/e2e/mco_test.go
ravisantoshgudimetla: Ya, I saw it too. Will do that but it's taking 1:30 hrs to fail. Let me try it locally first.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: yes. Thnx for catching it

pkg/controller/node/node_controller.go
ravisantoshgudimetla: I will change it.

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +153,235 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// This is duplicated code and is copied from package daemon. Perhaps a more neutral place for this exists. If not can we
+// create one?
+func updateNodeRetry(client corev1client.NodeInterface, lister corev1lister.NodeLister, nodeName string, f func(*corev1.Node)) (*corev1.Node, error) {
```
runcom: ok, let's move this to `./internal/kubeutils`(?)

ravisantoshgudimetla: Ohh, I just created a top level utils package with `node_utils.go`. I don't have a strong preference. I can move it `./internal/kubeutils`. Do you want me to do that?

ravisantoshgudimetla:
/retest

`fail [k8s.io/kubernetes/test/e2e/upgrades/apps/deployments.go:167]: Unexpected error:`

`fail [k8s.io/kubernetes/test/e2e/framework/framework.go:338]: Jul 11 01:26:53.041: Couldn't delete ns: "svcaccounts-4488": namespace svcaccounts-4488 was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed (&errors.errorString{s:"namespace svcaccounts-4488 was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed"})`

ravisantoshgudimetla:
`Jul 11 01:55:14.869 E clusterversion/version changed Failing to True: UpdatePayloadFailed: Could not update deployment "openshift-machine-config-operator/etcd-quorum-guard" (359 of 401)`

runcom:
/test e2e-aws

verify needs some care

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Yeah, I have done slice collection and then returned because in the next iteration it could cause problems with other node. Let me know, if you prefer it other way

test/e2e/mco_test.go
```diff
@@ -27,3 +31,63 @@ func TestClusterOperatorRelatedObjects(t *testing.T) {
...
+		t.Fatalf("Error while updating scheduler CR with error %v", err)
+	}
+	err = waitForAllMastersUpdate(cs, t, true)
+	if err != nil {
+		t.Fatalf("Expected all master nodes to be schedulable but it's not the case with %v", err)
+	}
+	// Reset scheduler CR
+	schedulerCR, err = cs.ConfigV1Interface.Schedulers().Get("cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal("Error while listing scheduler CR")
```
kikisdeliveryservice: minor: do we need the error added to the output here as well?

test/e2e/mco_test.go
```diff
@@ -27,3 +31,63 @@ func TestClusterOperatorRelatedObjects(t *testing.T) {
...
+	if _, err = cs.ConfigV1Interface.Schedulers().Update(schedulerCR); err != nil {
+		t.Fatalf("Error while updating scheduler CR with error %v", err)
+	}
+	err = waitForAllMastersUpdate(cs, t, false)
+	if err != nil {
+		t.Fatalf("Expected all master nodes to be unschedulable but it's not the case with %v", err)
+	}
+}
+
+func toggleMasterNodesSchedulability(cs *framework.ClientSet, masterSchedulable bool, t *testing.T) bool {
```
kikisdeliveryservice: question: this isn't really toggling the actual schedulableness of master it's just checking if whether or not it's correctly set no?

test/e2e/mco_test.go
ravisantoshgudimetla: Yeah, we should. It'll be useful later

test/e2e/mco_test.go
ravisantoshgudimetla: Yeah, you're right. I initially wanted the toggling to happen here instead of https://github.com/openshift/machine-config-operator/pull/937/files#diff-737b90e6d061aba6fa31500eba42868eR45 but I kind of moved away. I will change it `checkAllMastersSchedulable` or something like that.

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +154,197 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+}
+
+// makeMasterNodeSchedulable makes master node schedulable by removing NoSchedule master taint and
+// adding worker label
+func (ctrl *Controller) makeMasterNodeSchedulable(node *corev1.Node) error {
+	_, err := internal.UpdateNodeRetry(ctrl.kubeClient.CoreV1().Nodes(), ctrl.nodeLister, node.Name, func(node *corev1.Node) {
+		// Add worker label
+		newLabels := node.Labels
+		if _, hasWorkerLabel := newLabels[workerLabel]; !hasWorkerLabel {
+			newLabels[workerLabel] = ""
```
cgwalters: Doesn't this need to be:
`delete node.Labels[workerLabel]` ?

cgwalters: Er sorry, had this backwards!

pkg/controller/node/node_controller.go
```diff
@@ -128,11 +154,197 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
...
+// makeMasterNodeUnSchedulable makes master node unschedulable by removing worker label and adding `NoSchedule`
+// master taint to the master node
+func (ctrl *Controller) makeMasterNodeUnSchedulable(node *corev1.Node) error {
+	_, err := internal.UpdateNodeRetry(ctrl.kubeClient.CoreV1().Nodes(), ctrl.nodeLister, node.Name, func(node *corev1.Node) {
+		// Remove worker label
+		newLabels := node.Labels
+		if _, hasWorkerLabel := newLabels[workerLabel]; hasWorkerLabel {
+			delete(newLabels, workerLabel)
+		}
+		node.Labels = newLabels
```
cgwalters: I  think all of the code above can just be:
`delete(node.Labels, workerLabel)` since it's a no-op if the key doesn't exist.

cgwalters:
Overall this looks good, thanks for working on it!  Can you please squash it into one commit - since this is really one logical change.  And also now the last commit has intermixed fixups from review.  And please write a commit message, something like:

```
Watch scheduler CR and make masters schedulable accordingly

By default, the MCO's kubelet configuration injects a taint to disable scheduling.  This adds support in the MCO for watching the API recently added to allow the masters to be schedulable.  https://github.com/openshift/api/pull/366

Use cases here are for 3 node bare metal clusters (which have significant resources on the masters), as well as CodeReady Containers which wants to make a single node OpenShift as a VM.

Closes: https://github.com/openshift/machine-config-operator/issues/763
```

ravisantoshgudimetla:
Thanks for the review @runcom @cgwalters @kikisdeliveryservice. I made the changes suggested. PTAL

kikisdeliveryservice:
this lgtm 

will let @cgwalters / @runcom give the final approval

runcom:
needs a rebase :(

cgwalters:
> needs a rebase :(

The bright side is it's an opportunity to use the :surfing_man: emoji

ravisantoshgudimetla:
Rebased :see_no_evil:

kikisdeliveryservice:
ci didn't like that last commit :(

kikisdeliveryservice:
uhoh hitting limits now:

`./tmp/openshift-install-050178053/vpc/master-elb.tf line 1, in resource \"aws_lb\" \"api_internal\":"`

kikisdeliveryservice:
reported but ci has been hitting this for the last few hours.

kikisdeliveryservice:
Still seeing ci problems:
``` level=error msg="Error: timeout while waiting for state to become 'active' (last state: 'provisioning', timeout: 10m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-001008030/vpc/master-elb.tf line 19, in resource \"aws_lb\" \"api_external\":"
level=error msg="  19: resource \"aws_lb\" \"api_external\" {" ```

/retest

PR #938, opened 2019-07-08
daemon: restrict journal queries

cgwalters:


cgwalters:
Only compile tested locally.

cgwalters:
Maybe it'd be better to make the uid/unit bindings come first since they're low risk and important, then do the journal API rework on top.  Will probably do that after this CI run finishes.

cgwalters:
/hold
will rework soon per above

PR #941, opened 2019-07-09
fold mco images/ ratchet: add new master Dockerfiles with finalized names

kikisdeliveryservice:
The final mco master dockerfiles will be named:
Dockerfile and Dockerfile.rhel7 We need to ratchet in the name
change by first adding these files and copying contents of the
current master dockerfiles Dockerfile.machine-config-operator/.rhel7

Once this is is merged will be able to update the names in release
and ART in followup prs and remove all other Dockerfiles.

Related-to: #739
Related-to: #894
Required-by: openshift/release#4314

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/machine-config-operator
+COPY . .
+RUN WHAT=machine-config-operator ./hack/build-go.sh
+RUN WHAT=machine-config-daemon ./hack/build-go.sh
+RUN WHAT=machine-config-controller ./hack/build-go.sh
+RUN WHAT=machine-config-server ./hack/build-go.sh
+RUN WHAT=setup-etcd-environment ./hack/build-go.sh
+
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-operator /usr/bin/
+COPY install /manifests
+COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-daemon /usr/bin/
+RUN if ! rpm -q util-linux; then yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*; fi
+COPY --from=builder /go/src/github.com/openshift/machine-config-operator/_output/linux/amd64/machine-config-controller /usr/bin/
+COPY templates /etc/mcc/templates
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/machine-config-operator
+COPY . .
+RUN WHAT=machine-config-operator ./hack/build-go.sh
```
cgwalters: Not required to do in this PR but I think as a followup we should just make `./hack/build-go.sh` build everything by default and do that in `RUN` to reduce duplication.

Dockerfile.rhel7
```diff
@@ -0,0 +1,30 @@
...
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/machine-config-operator/_output/linux/$(go env GOARCH)/machine-config-controller /tmp/build/machine-config-controller
+RUN WHAT=machine-config-server ./hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/machine-config-operator/_output/linux/$(go env GOARCH)/machine-config-server /tmp/build/machine-config-server
+RUN WHAT=setup-etcd-environment ./hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/machine-config-operator/_output/linux/$(go env GOARCH)/setup-etcd-environment /tmp/build/setup-etcd-environment
+
+FROM registry.svc.ci.openshift.org/ocp/4.0:base
```
cgwalters: Another unrelated question...anyone know what's the status of switching to the UBI image, specifically UBI8?  If we did that I think we wouldn't need to duplicate the two Dockerfiles at all even.

But still not a blocker for this just noting the question.

kikisdeliveryservice: not sure. fwiw: i did double check that this change worked for ART, but that was all.

Dockerfile
kikisdeliveryservice: happy to refactor as a followup PR :+1:

kikisdeliveryservice:
`I0709 19:40:29.638259     178 request.go:530] Throttling request took 896.885297ms, request: GET:https://api.ci-op-wz9x2t4k-49e8c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/apps.openshift.io/v1?timeout=32s` 

/retest

Dockerfile.rhel7
runcom: This is a good one, probably worth following up with openshift/release as I imagine it would be projects wide

kikisdeliveryservice: I can followup on this if you'd  like @runcom

runcom: sure thing, as Colin said, not blocking

PR #942, opened 2019-07-09
pkg/controller: Fix up node status.go docs

sgreene570:
**- What I did**
Cleaned up some inaccurate / hard to follow function doc strings.

**- How to verify it**

Fix doc strings in pkg/controller node status

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -148,7 +148,7 @@ func isNodeDoneAt(node *corev1.Node, targetConfig string) bool {
 	return isNodeDone(node) && node.Annotations[daemonconsts.CurrentMachineConfigAnnotationKey] == targetConfig
 }
 
-// isNodeMCDDone checks the MCD state
+// isNodeMCDState checks the MCD state against the state parameter
 func isNodeMCDState(node *corev1.Node, state string) bool {
 	dstate, ok := node.Annotations[daemonconsts.MachineConfigDaemonStateAnnotationKey]
 	if !ok || dstate == "" {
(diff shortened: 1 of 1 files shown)


runcom:
amazing
/approve
/lgtm

pkg/controller/node/status.go
```diff
@@ -167,9 +167,8 @@ func isNodeMCDFailing(node *corev1.Node) bool {
 		isNodeMCDState(node, daemonconsts.MachineConfigDaemonStateUnreconcilable)
 }
 
-// getUpdatedMachines filters the provided nodes to ones whose current config
-// matches the desired config, which also matches the target config.
-// and the "done" flag is set.
```
runcom: this is actully true if you look at `isNodeDoneAt`

pkg/controller/node/status.go
```diff
@@ -221,8 +220,7 @@ func isNodeReady(node *corev1.Node) bool {
 	return checkNodeReady(node) == nil
 }
 
-// isNodeUnavailable is the backend for getUnavailableMachines;
-// see the docs for that for more information.
```
runcom: any reason to drop this?

sgreene570: I guess not. Can put back.

pkg/controller/node/status.go
sgreene570: I see. I thought having "desired config" and "target config" at the same time was confusing.

kikisdeliveryservice:
Nice changes @sgreene570 !!!

Issue #812, opened 2019-06-03
bm/ovirt: pass API_VIP/DNS_VIP from cluster config/installer and render templates with it


rgolangh:
I'm developing the platform provider for ovirt #766, which is a platform which doesn't provide a DNS service, and as such it utilzes the work done on metal3. To bootstrap a working DNS solution, the installer asks the user for an IP to be the virtual IP of the DNS - DNS_VIP.
This DNS_VIP is a value which must end up in the masters ignition files for the ovirt platform, and possibly others. 

Today I see that there are number of files which bootstrap MCO, one of them is `config-file`:
https://github.com/openshift/machine-config-operator/blob/acf5b49de3692801ce9cd3d799f64961f9283b00/cmd/machine-config-operator/bootstrap.go#L74

Which is the serialized `InstallConfig` of the installer, which has all I need:
https://github.com/openshift/installer/blob/479ca0fc0fd46c489c4b97d783bf1dd58cc373a2/pkg/types/installconfig.go#L42

What I want is to add this ClusterConfigMap to the RenderConfig and by that I could render my templates.
Can I rely on the installer InstallConfig type and to serialize the content to it and put it in render config? It creates dependency on the installer types, and I don't know yet how does it fit the overall design of those components.

kikisdeliveryservice:
Briefly looking through, it looks like the following prs would need to land before we'd tackle this:
https://github.com/openshift/api/pull/314
https://github.com/openshift/machine-config-operator/pull/767

What is the status of the first PR in openshift/api ??

rgolangh:
I have nothing to change there, just waiting for approval

kikisdeliveryservice:
specifically which value are you needing from the cluster config?

kikisdeliveryservice:
Also if you can add more details about the platform, that would be great as I don't think we have that much familiarity with it.

rgolangh:
You'll find it here, under the new 'ovirt' platform (not merged yet obviously) 
https://github.com/rgolangh/installer/blob/fd85cbdb57e9d3e5ba20bba757468c9b8c57642d/pkg/types/ovirt/platform.go#L27

rgolangh:
I can think of more data which I need to share, like the the platform authentication details, which will be used by the cluster-api or actuator to provision machines

kikisdeliveryservice:
can you outline what PRs need to merge / the overall status of this ovirt platform? before this PR could even merge?

runcom:
[...]

PR #943, opened 2019-07-10
Make the infra object available for template rendering

rgolangh:
oVirt and possibly other platforms needs access the InfrastructureStatus
to render properly config files with stuff like API VIP, DNS VIP etc.

By embedding the infra object its available for rendering in
tremplates - for example a coredns config file like this:

```
{{ ControllerConfig.Infra.Status.PlatformStatus.Ovirt.DnsVIP }} api-int.{{.EtcdDiscoveryDomain}} api.{{.EtcdDiscoveryDomain}} ns1.{{.EtcdDiscoveryDomain}}

```

**How to verify it**
TODO will create a test

Fixes: #812 
Required-by: #795

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -94,6 +94,10 @@ type ControllerConfigSpec struct {
 
 	// proxy holds the current proxy configuration for the nodes
 	Proxy *configv1.ProxyStatus `json:"proxy"`
+
+	// infra holds the infrastructure details
+	// TODO this makes platform redundant as everything is contained inside Infra.Status
+	Infra *configv1.Infrastructure `json:"infra"`
 }
 
 // ControllerConfigStatus is the status for ControllerConfig
--- a/pkg/operator/render.go
+++ b/pkg/operator/render.go
@@ -88,6 +88,7 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 		CloudProviderConfig: "",
 		EtcdDiscoveryDomain: infra.Status.EtcdDiscoveryDomain,
 		Platform:            platform,
+		Infra:               infra,
 	}
 
 	if proxy != nil {
--- a/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
@@ -253,6 +253,11 @@ func (in *ControllerConfigSpec) DeepCopyInto(out *ControllerConfigSpec) {
 		*out = new(configv1.ProxyStatus)
 		**out = **in
 	}
+	if in.Infra != nil {
+		in, out := &in.Infra, &out.Infra
+		*out = new(configv1.Infrastructure)
+		(*in).DeepCopyInto(*out)
+	}
 	return
 }
 


pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -94,6 +94,10 @@ type ControllerConfigSpec struct {
 
 	// proxy holds the current proxy configuration for the nodes
 	Proxy *configv1.ProxyStatus `json:"proxy"`
+
+	// infra holds the infrastructure details
+	// TODO this makes platform redundant as everything is contained inside Infra.Status
+	Infra configv1.Infrastructure
```
runcom: Pointer?

rgolangh: Why a pointer? are the values changing in a differnet place?

runcom: maintaining the pattern with other infra object and useful to differentiate between set/unset (not that it's needed now, but still)

pkg/operator/render.go
```diff
@@ -23,6 +23,7 @@ type renderConfig struct {
 	APIServerURL           string
 	Images                 *RenderConfigImages
 	KubeAPIServerServingCA string
+	Infra				   configv1.Infrastructure
```
sgreene570: nit: struct spacing.

This might actually get flagged in our linter.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -94,6 +94,10 @@ type ControllerConfigSpec struct {
 
 	// proxy holds the current proxy configuration for the nodes
 	Proxy *configv1.ProxyStatus `json:"proxy"`
+
+	// infra holds the infrastructure details
+	// TODO this makes platform redundant as everything is contained inside Infra.Status
```
kikisdeliveryservice: why is this PR making already existing platform redundant? what do you plan to do to with this TODO?

rgolangh: If this will get merged then we should remove platform and just get it from infra. Can you suggest different wording or you object the idea?

pkg/apis/machineconfiguration.openshift.io/v1/types.go
rgolangh: So are we going to use that for setting value to  it? Is that the generated code maybe? (help me to get that cause I'd like to understand it)

tomassedovic:
I'd just like to mention that the OpenStack and baremetal folks could really use this change (for basically the same reasons as @rgolangh is adding it for oVirt).

pkg/apis/machineconfiguration.openshift.io/v1/types.go
kikisdeliveryservice: I'm asking why you are adding infra when you could use the already existing platform

kikisdeliveryservice:
please correct me but it sounds like this PR requires: https://github.com/openshift/installer/pull/1948 to merge first. 

as a safety mechanism:
/hold

kikisdeliveryservice:
Are we ok with removing platform as this PR implies?  Would like to just have this idea explicitly :+1: so we are clear going forward.

cc: @runcom @cgwalters 

see: https://github.com/openshift/machine-config-operator/pull/943#discussion_r302476941

pkg/apis/machineconfiguration.openshift.io/v1/types.go
runcom: they need the whole infra object to read the ovirt platform status - this is a prep PR to inject infra

rgolangh:
@kikisdeliveryservice @runcom how do we continue?

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -94,6 +94,10 @@ type ControllerConfigSpec struct {
...
+	Infra *configv1.Infrastructure
```
LorbusChris: please add the missing `json:"infra"` tag

LorbusChris:
looks like the only thing left here is to run `make update` and commit :)

rgolangh:
which version of go-bindata is needed? look like `make update` created unwanted changes into

LorbusChris:
I hit the same last night. You need this:
https://github.com/openshift/machine-config-operator/pull/961/files#diff-df701135b6203a3b17b044643056007a
(last file in PR)

pkg/operator/render.go
```diff
@@ -24,6 +24,7 @@ type renderConfig struct {
 	APIServerURL           string
 	Images                 *RenderConfigImages
 	KubeAPIServerServingCA string
+	Infra                  configv1.Infrastructure
```
kikisdeliveryservice: if Infra was added to the `ControllerConfigSpec` above why add it separately into the `renderConfig` as a field? Wouldn't it just be `ControllerConfig.Infra`?

rgolangh: Good catch, it is not needed.

kikisdeliveryservice: thanks!

kikisdeliveryservice:
adding a hold so this doesn't merge without clearance via exception.

/hold

bcrochet:
> adding a hold so this doesn't merge without clearance via exception.

I believe we have a FFE for #795 , and it depends on this patch.

PR #946, opened 2019-07-10
WIP: Clean up Dockerfile

cgwalters:
Fix `make binaries` to also build `setup-etcd-environment`, then
use that in the `Dockerfile` to reduce duplication.

Similarly, just use globs to avoid hardcoding the architecture,
and also grab all binaries.

Add comments in various places.


Dockerfile.machine-config-operator
```diff
@@ -1,20 +1,18 @@
+# This is the "all in one" image for the MCO; we previously
```
runcom: I know it's a pain but `Dockerfile` must be updated as well till we remove this

kikisdeliveryservice:
Thanks @cgwalters !! 
Can you copy these changes into `Dockerfile` since that will be the master mco dockerfile going forward?  (the others will be removed in my other pr). For now `Dockerfile` and `Dockerfile.machine-config-operator` are exact copies.

Dockerfile.machine-config-operator
cgwalters: Oh right I changed the wrong one, argh.

cgwalters:
Can't we just symlink them?

runcom:
> Can't we just symlink them?

we're gonna remove it soon enough - we didn't expect a change to it :smile:

cgwalters:
https://github.com/openshift/machine-config-operator/pull/947

kikisdeliveryservice:
> > Can't we just symlink them?
> 
> we're gonna remove it soon enough - we didn't expect a change to it smile

Basically this. :)

cgwalters:
(The real reason I'm doing this is `make deploy-$x` is broken at the moment and I use it a lot)

kikisdeliveryservice:
this pr will have to be rebased onto #947 and these changes will go into `Dockerfile`

Makefile
```diff
@@ -105,3 +105,7 @@ images.rhel7: $(imc7)
 # This was copied from https://github.com/openshift/cluster-image-registry-operato
 test-e2e:
 	go test -timeout 75m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
+
+Dockerfile.rhel7: Dockerfile Makefile
+	(echo 'THIS FILE IS GENERATED FROM '$<' DO NOT EDIT' && \
+	 sed -e s,org/openshift/release,org/ocp/builder, -e s,/openshift/origin-v4.0:base,/ocp/4.0:base, < $<) > $@.tmp && mv $@.tmp $@
```
cgwalters: Stand back, I'm pulling out the `sed`!

cgwalters:
Wait ugh...does the version of docker in 3.11 not support globs or something?

runcom:
> Wait ugh...does the version of docker in 3.11 not support globs or something?

no version support glob afaict, in the Dockerfile you're not running a shell with expansions and whatnot

cgwalters:
> no version support glob afaict, in the Dockerfile you're not running a shell with expansions and whatnot

See
https://docs.docker.com/engine/reference/builder/#copy

Specifically:

> Each <src> may contain wildcards and matching will be done using Go’s filepath.Match rules. For example:

And this works with `podman-1.3.1-1.git7210727.fc30.x86_64` locally.

runcom:
> See
> https://docs.docker.com/engine/reference/builder/#copy

then I completely missed the feature and was left really behind, TIL

cgwalters:
Obsoleted by https://github.com/openshift/machine-config-operator/pull/1210

PR #947, opened 2019-07-10
build: Symlink rather than copy Dockerfiles for operator

cgwalters:
This way we avoid duplication.


kikisdeliveryservice:
These files will be deleted in a couple of days so I dont think it's necessary to do this.

runcom:
> These files will be deleted in a couple of days so I dont think it's necessary to do this.

Colin actually mentioned he's broken on the make deploy scripts so unless we can fix this _today_ I think this might be a good shortcut

kikisdeliveryservice:
> > These files will be deleted in a couple of days so I dont think it's necessary to do this.
> 
> Colin actually mentioned he's broken on the make deploy scripts so unless we can fix this _today_ I think this might be a good stopgap

Can we just make this change in the other PR then?  Or conversely this needs to merge before his other PR bc of the linking.

cgwalters:
> Or conversely this needs to merge before his other PR bc of the linking.

I wanted to test them independently in CI; the other one depends on this, but this can go in quite safely I think.

kikisdeliveryservice:
This is now hitting BZ1728888 :(

PR #948, opened 2019-07-10
dev: It's just `make image` now

cgwalters:
We only have one image, so massively simplify the code.

(Note that `make deploy` i.e. `hack/cluster-push` is still
 not aware of this but fixing that is more involved and
 this is correct in the interim)


cgwalters:
(Now you might ask, what's the value of even having a wrapper script for something so obvious?  I think I'd like to continue to have `make deploy` work...and `Makefile` is an OK way to wrap these chains of tasks together)

kikisdeliveryservice:
Nice. You simplified this even more than I did.

I'm fine with keeping it in the makefile - I think it keeps it simple by having things in a central place and consistent

Just testing this locally... one sec

PR #952, opened 2019-07-11
pkg/daemon: reconcile being killed by kube on drain+reboot

runcom:

**- What I did**

WIP mainly because I want to test it out on a live cluster by forcing a grace period of something really low to kick the kubelet killing the MCD

This patch removes the perma-failure we have in case:

- we were draining
- we exceeded 600s terminationGracePeriod
- the kubelet killed (9) us

The above can be deducted by:

- pending config is on disk
- bootID equality

In such case, we can go ahead and re-kick and drain+reboot routine till it eventually succedes (modulo: as Colin said in https://bugzilla.redhat.com/show_bug.cgi?id=1728873 we're in a permanent situation where we can't really drain _forever_)

**- How to verify it**

testing it manually for now

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -750,23 +750,20 @@ func (dn *Daemon) writePendingConfig(desiredConfig *mcfgv1.MachineConfig) error 
 
 // XXX: drop this
 // we need this compatibility layer for now
-func (dn *Daemon) getPendingConfig() (string, error) {
+func (dn *Daemon) getPendingConfig() (*pendingConfigState, error) {
 	s, err := ioutil.ReadFile("/etc/machine-config-daemon/state.json")
 	if err != nil {
 		if !os.IsNotExist(err) {
-			return "", errors.Wrapf(err, "loading transient state")
+			return nil, errors.Wrapf(err, "loading transient state")
 		}
-		return "", nil
+		return nil, nil
 	}
 	var p pendingConfigState
 	if err := json.Unmarshal(s, &p); err != nil {
-		return "", errors.Wrapf(err, "parsing transient state")
+		return nil, errors.Wrapf(err, "parsing transient state")
 	}
 
-	if p.BootID == dn.bootID {
-		return "", fmt.Errorf("pending config %s bootID %s matches current! Failed to reboot?", p.PendingConfig, dn.bootID)
-	}
-	return p.PendingConfig, nil
+	return &p, nil
 }
 
 func (dn *Daemon) getCurrentConfigOnDisk() (*mcfgv1.MachineConfig, error) {
@@ -797,6 +794,7 @@ func (dn *Daemon) storeCurrentConfigOnDisk(current *mcfgv1.MachineConfig) error 
 // the daemon took over.
 //
 // Some more background in this PR: https://github.com/openshift/machine-config-operator/pull/245
+//nolint:gocyclo
 func (dn *Daemon) checkStateOnFirstRun() error {
 	node, err := dn.loadNodeAnnotations(dn.node)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -804,22 +801,39 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
+		bootID = legacyPendingState.BootID
 	}
+
 	state, err := dn.getStateAndConfigs(pendingConfigName)
 	if err != nil {
 		return err
 	}
+
+	// if we have a pendingConfig but we're into the same bootid, we failed to drain or reboot
+	// and if we still have a pendingConfig it means we've been killed by kube after 600s
```
cgwalters: These seem like distinct cases.  If we failed to drain, we should easily be able to detect that.

Failure to *reboot* is where we get killed by kube.

runcom: uhm according to logs, we _were_ draining, it just took too long so we're in the middle of a drain and can't even get past the Drain function to reboot. That's where we _also_ get sigkill'ed by kubelet

runcom: a normal failure to drain doesn't incur in the BZ, if we get a Drain error, we just rollback everything and retry, this is about being stuck at drain and getting killed by exceeding the 600s grace term period

cgwalters:
Sorry can you walk me through this one a bit more?  You say

> we were draining
> we exceeded 600s terminationGracePeriod

But our drain shouldn't kill the MCD as it's a daemonset, which we ignore right?  Or does that "ignore" mean "try and drain but don't wait for it"?

runcom:
> Sorry can you walk me through this one a bit more? You say
> 
> > we were draining
> > we exceeded 600s terminationGracePeriod
> 
> But our drain shouldn't kill the MCD as it's a daemonset, which we ignore right? Or does that "ignore" mean "try and drain but don't wait for it"?

The issue here is specific to "doing something in MCD after being requested to shutdown (MCD)" and I'm not sure who asked that (can't really say from logs but will keep investigating). The logs show that the MCD has been requested to shut down but it was still draining (again not sure who asked for that) and after 600s we've been killed by kubelet and we now have the same bootid cause we didn't reboot.

cgwalters:
> The logs show that the MCD has been requested to shut down but it was still draining (again not sure who asked for that) and after 600s we've been killed by kubelet and we now have the same bootid cause we didn't reboot.

Hmmm.  OK, it's making more sense.  Maybe it's something like someone manually doing `systemctl restart kubelet`?  It's not the admin doing `reboot` directly because then we'd reboot right?

runcom:
These are the relevant logs (you can see the pid changed as well for the MCD):
```
Jul 02 00:05:46 ip-192-168-179-98 root[106368]: machine-config-daemon[98533]: controller syncing started
Jul 02 00:05:46 ip-192-168-179-98 root[106369]: machine-config-daemon[98533]: Starting update from rendered-worker-7f899c2c109ce7eea3cde169e3f51092 to rendered-worker-71095e466f5b906cdbcc9ff8cb04414c
Jul 02 00:05:46 ip-192-168-179-98 root[106371]: machine-config-daemon[98533]: Update prepared; beginning drain
Jul 02 00:22:14 ip-192-168-179-98 root[127189]: machine-config-daemon[127147]: Starting to manage node: ip-192-168-179-98.us-west-1.compute.internal
```

cgwalters:
Side note, will conflict some with https://github.com/openshift/machine-config-operator/pull/935 but oh well.

runcom:
> Hmmm. OK, it's making more sense. Maybe it's something like someone manually doing `systemctl restart kubelet`? It's not the admin doing `reboot` directly because then we'd reboot right?

could indeed be, or the kubelet shut down alone? is there anyway to check journal for `systemctl restart kubelet`?

runcom:
so there's no sign of admin rebooting - rather, something interesting popped up, we've been asked to shut down after ~300s from starting drain

runcom:
> so there's no sign of admin rebooting - rather, something interesting popped up, we've been asked to shut down after ~300s from starting drain

so meanwhile speaking to Ravi and others to understand how to grab useful logs as to why the kubelet wants us to shut down.

Besides....I think this patch does make sense as a stopgap/best effort reconciliation - @cgwalters wdyt?

pkg/daemon/daemon.go
```diff
@@ -797,29 +794,49 @@ func (dn *Daemon) storeCurrentConfigOnDisk(current *mcfgv1.MachineConfig) error
...
+	// take a stab at that and re-run the drain+reboot routine
+	if state.pendingConfig != nil && bootID == dn.bootID {
+		dn.logSystem("killed by kube, retrying...")
+		return dn.drainAndReboot(state.pendingConfig)
+	}
```
runcom: oh, sorry, I actually regressed, here we need the bootid check anyway

runcom: uhm, not really

PR #955, opened 2019-07-11
Generate CRD schema

damemi:
This adds the markers to the `machineconfiguration.openshift.io` doc.go to label the group appropriately eg, for use with CRD generators


damemi:
In https://github.com/openshift/machine-config-operator/pull/955/commits/6591658eb7b8f55a6a29351a68867c8cd0126d0d, these vendored types were causing failures in the generator due to lack of JSON tags. I don't know what they are, but adding these tags in allowed the generator to run.

vendor/github.com/coreos/ignition/config/v2_2/types/schema.go
```diff
@@ -30,8 +30,8 @@ type CreateOption string
 type Device string
 
 type Directory struct {
-	Node
-	DirectoryEmbedded1
+	Node               `json:"node,omitempty"`
+	DirectoryEmbedded1 `json:"directoryEmbedded1,omitempty"`
 }
```
LorbusChris: we might want to add these json tags to the upstream types as well.
/cc @runcom @ajeddeloh

ajeddeloh: No, this should be dropped here. It's using struct embedding, the Node/DirectoryEmbedded1 should be invisible to anyone looking at the serialized json.

LorbusChris: @damemi @vrutkovs this concerns the machinconfig crd, why was this not necessary for the precursory PR? (https://github.com/openshift/machine-config-operator/pull/928)

LorbusChris: Note: if we're goping to expose ign.Config as RawExtension externally, we won't need this. Will be discussed in Tuesday's RHCOS cabal.

damemi: @LorbusChris that PR had the CRD validation manually written because the [CRD generator](https://github.com/openshift/crd-schema-gen) was not working against this repo. This PR is just showing the steps to get that generator working. One of the errors was due to these types, which I'm assuming are some kind of dependency, not having JSON tags. 

The fact that I've added these tags here is really more of a proof of concept (showing the CRD generator properly running). Due to my lack of knowledge on these or the upstream types I was hoping people more familiar with them could give more review, and determine if it's a change worth having.

ajeddeloh: The Igntition types vendored here are generated from a json schema. We use struct embedding (which is what our json schema->go types generator outputs) to cut down a bit on code duplication. 

It's also worth noting that Ignition types have their own validation code which can be used.

PR #960, opened 2019-07-12
cmd/setup-etcd-environment: add support for exporting env

hexfusion:
This PR adds support for exporting variables consumed as params by the etcd binary during the setup-etcd-environment init container. As we begin to move to cluster-etcd-operator managing bootstrap. We will need to gracefully transition the etcd-member.yaml spec to something more agnostic.

As an example use-case, I removed the runtime flag `--discovery-srv`  and replaced it with an environment variable. As some of these variables can be member-specific such as name they do not fit well with the MachineConfig model.

It would be easier in general to export all variables but as we prefix them with `ETCD_`  they will get picked up by the binary and print a bit of cruft during startup. For example.

```
2019-07-11 20:36:38.756064 W | pkg/flags: unrecognized environment variable ETCD_IPV4_ADDRESS=192.168.0.12
```
While this is mostly cosmetic I fear it could cause confusion which is the reason for adding the bool to export.

As a bonus all etcd containers now set ETCDCTL_API=3 as default.

/cc @abhinavdahiya @runcom

### DIFF
--- a/cmd/setup-etcd-environment/run.go
+++ b/cmd/setup-etcd-environment/run.go
@@ -86,11 +86,17 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 		out = f
 	}
 
+	if err := writeEnvironmentFile(map[string]string{
+		"DISCOVERY_SRV": runOpts.discoverySRV,
+	}, out, true); err != nil {
+		return err
+	}
+
 	return writeEnvironmentFile(map[string]string{
 		"IPV4_ADDRESS":      ip,
 		"DNS_NAME":          dns,
 		"WILDCARD_DNS_NAME": fmt.Sprintf("*.%s", runOpts.discoverySRV),
-	}, out)
+	}, out, false)
 }
 
 func ipAddrs() ([]string, error) {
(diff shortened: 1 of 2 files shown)


cmd/setup-etcd-environment/run.go
```diff
@@ -150,11 +156,16 @@ func reverseLookupSelf(service, proto, name, self string) (string, error) {
...
-func writeEnvironmentFile(m map[string]string, w io.Writer) error {
+func writeEnvironmentFile(m map[string]string, w io.Writer, export bool) error {
 	var buffer bytes.Buffer
 	for k, v := range m {
 		buffer.WriteString(fmt.Sprintf("ETCD_%s=%s\n", k, v))
 	}
+	if export == true {
+		s := buffer.String()
+		buffer.Reset()
+		buffer.WriteString(fmt.Sprintf("set -a\n%sset +a\n", s))
```
abhinavdahiya: the reason why we want to export certain is to differentiate the access between the env variables available to the script calls `exec etcd` and the `etcd` process itself right?

would it be simpler, if all envs from setup-etcd-environment should be available to both ie set all as export.

hexfusion: > would it be simpler, if all envs from setup-etcd-environment should be available to both ie set all as export.

it would be easier but as we prefix `ETCD_` on all env they get picked up by the binary and print a bit of cruft during startup. Stuff like

2019-07-11 20:36:38.756064 W | pkg/flags: unrecognized environment variable ETCD_IPV4_ADDRESS=192.168.0.12

While this is mostly cosmetic I fear it could cause confusion.

abhinavdahiya:
```
export <>
```
vs
```
set -a
<>
set +a
```
not sure why we picked one over the other.

abhinavdahiya: +1

abhinavdahiya:
LGTM

cmd/setup-etcd-environment/run.go
abhinavdahiya: +1

i think this reasoning would be useful in the commit message or the PR description for future reference.

hexfusion: I like to enable/disable -a as a way to isolate a group of variables or an entire file of variables marked for export. If we were to add multiple vars with in the set it would not require any additional action. But I agree export is more common.

hexfusion: will update which should simplify the code as well.

hexfusion: good idea, updated

Issue #951, opened 2019-07-11
Pin a version of go bin-data or atleast document which one to use


ravisantoshgudimetla:
**Description**
As of now, there are many variants of go bin-data. As a contributor, it becomes difficult to identify which version of bindata I should use. Can this be explicitly documented or have that library vendored and pinned. The other advantage of having the version pinned is, we don't have to explicitly mention which version of bin-data library to use

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `oc adm release info --commits | grep machine-config-operator`:**

```
(paste your output here)
```

**Additional environment details (platform, options, etc.):**

kikisdeliveryservice:
This issue actually brought down our ci today. In summary:

`test verify` started failing across PRs. I noticed it happening #953 which is a simple doc fix. When looking at verify i saw:

```
 --- hack/../_tmp/pkg/operator/assets/bindata.go    2019-07-11 20:25:22.160515877 +0000
+++ hack/../pkg/operator/assets/bindata.go    2019-07-11 20:25:22.176516911 +0000
@@ -66,7 +66,7 @@ func (fi bindataFileInfo) Mode() os.File
     return fi.mode
 }
 
-// Mode return file modify time
+// ModTime return file modify time
 func (fi bindataFileInfo) ModTime() time.Time {
     return fi.modTime
 } 
```

Which matches the first change in a month to go into go-bindata: https://github.com/go-bindata/go-bindata/commit/a87177b60aad08824a6f8b96d9c3dfa1117ebe6c#diff-15036d5e0876436b1832c7721c9c7afd

Note that this changes is a mere comment change but gave us non-matching bindata and brought our CI to a halt.

As a fix we updated bindata.go here in #957 but we should probably be pinning all of these gopkgs going forward and document this as ravi suggested above.

Further, we will also have to update both our repo and release (see https://github.com/openshift/release/pull/4282/files) which also doesn't pin.  walters has suggested that we could change the release job to run a script from our repo.

kikisdeliveryservice:
Also in the course of troubleshooting this I realized that I (and other people) were using an old go bindata version (a change went in recently that we didnt update locally) so it would be good to update documentation as well for dev systems.

kikisdeliveryservice:
[...]

PR #961, opened 2019-07-12
Makefile: pin tools to specific versions

runcom:
This also inverts the dependency we have with openshift/release.

This has a downside tho..if someone adds something that brings down the whole infra it can be our fault to have allowed arbitrary binaries executions. On the other hand.. we're running in pods

Close https://github.com/openshift/machine-config-operator/issues/951

### DIFF
--- a/hack/install-go-bindata.sh
+++ b/hack/install-go-bindata.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+tmp=$(mktemp -d)
+version=ee3c2418e3682cc4a4e6c5dd1b32d0b98f7e2c55
+
+cd $tmp &&\
+git clone https://github.com/go-bindata/go-bindata.git &&\
+cd $tmp/go-bindata &&\
+git checkout -q $version &&\
+go build -o build-go-bindata ./go-bindata &&\
+cp build-go-bindata $GOPATH/bin/go-bindata
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ COMPONENTS = daemon controller server operator
 # vim: noexpandtab ts=8
 export GOPATH=$(shell echo $${GOPATH:-$$HOME/go})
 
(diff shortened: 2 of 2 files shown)


runcom:
infra resources exhaustion...

hack/install-go-bindata.sh
```diff
@@ -0,0 +1,11 @@
...
+
+tmp=$(mktemp -d)
+version=ee3c2418e3682cc4a4e6c5dd1b32d0b98f7e2c55
+
+cd $tmp &&\
+git clone https://github.com/go-bindata/go-bindata.git &&\
+cd $tmp/go-bindata &&\
+git checkout -q $version &&\
+go build -o build-go-bindata ./go-bindata &&\
+cp build-go-bindata $GOPATH/bin/go-bindata
```
sgreene570: nit:
No new line at the end of this file.

abhinavdahiya:
You can also vendor these if you like :p and use them from vendor as go run vendor/....

dep has required packages
go mod, requires a package that imports them

Example
https://github.com/openshift/cluster-ingress-operator/blob/master/tools/tools.go

runcom:
> You can also vendor these if you like :p and use them from vendor as go run vendor/....

this is so nice, thanks for this

runcom:
> this is so nice, thanks for this

however..... it's really pulling so many new vendors :(

runcom:
this is ready to go

hack/install-go-bindata.sh
runcom: fixed

runcom:
addressed Steve's comment, this is now good to go

kikisdeliveryservice:
@sgreene570 do you want to test this?

sgreene570:
Tested these changes locally and everything looks good!
/lgtm

runcom:
fat fingers...

Issue #954, opened 2019-07-11
Image config controller tries to sync MachineConfig's for a custom defined pool


kincl:
**Description**

I am using OCP 4.1.4 on bare-metal with UPI and want to define a MachineConfigPool for infrastructure hosts. I am defining a pool with:

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfigPool
metadata:
  name: infra
spec:
  machineConfigSelector:
    matchLabels:
      machineconfiguration.openshift.io/role: infra
      machineconfiguration.openshift.io/role: worker
  nodeSelector:
    matchLabels:
      node-role.kubernetes.io/infra: ""
  paused: false
```

This creates the `rendered-infra-aaabbb` MachineConfig but the machine-config-controller complains that it cannot access `/etc/mcc/templates/infra` to sync the image configs. Since this is a custom MachineConfigPool there will not be any templates to sync so this should not be an error.

**Steps to reproduce the issue:**
1. oc create -f machineconfigpool.yaml
2. oc logs -n openshift-machine-config-operator $(oc get pod -l k8s-app=machine-config-controller -n openshift-machine-config-operator -o name)
3.

**Describe the results you received:**

```
I0711 15:44:53.570164       1 container_runtime_config_controller.go:336] Error syncing image config openshift-config: could not Create/Update MachineConfig: could not generate origin ContainerRuntime Configs: generateMachineConfigsforRole failed with error failed to read dir "/etc/mcc/templates/infra": open /etc/mcc/templates/infra: no such file or directory
E0711 15:46:15.516315       1 container_runtime_config_controller.go:341] could not Create/Update MachineConfig: could not generate origin ContainerRuntime Configs: generateMachineConfigsforRole failed with error failed to read dir "/etc/mcc/templates/infra": open /etc/mcc/templates/infra: no such file or directory
I0711 15:46:15.516337       1 container_runtime_config_controller.go:342] Dropping image config "openshift-config" out of the queue: could not Create/Update MachineConfig: could not generate origin ContainerRuntime Configs: generateMachineConfigsforRole failed with error failed to read dir "/etc/mcc/templates/infra": open /etc/mcc/templates/infra: no such file or directory
```

**Describe the results you expected:**

No error logs

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `oc adm release info --commits | grep machine-config-operator`:**

```
(paste your output here)
```

**Additional environment details (platform, options, etc.):**

PR #963, opened 2019-07-12
pkg/controller: skip image config MCs sync for custom defined pools

runcom:
**- What I did**

Fix https://github.com/openshift/machine-config-operator/issues/954

For custom defined pools (inheriting mainly from worker) do not sync the image config MCs. That's because we already sync for the default pools and that causes a resync for custom defined pool as well. Avoid also a spammy error log.

**- How to verify it**

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -50,6 +50,8 @@ const (
 	//
 	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
 	maxRetries = 15
+
+	builtInLabelKey = "machineconfiguration.openshift.io/mco-built-in"
 )
 
 var (
@@ -625,8 +627,12 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		return err
 	}
 
-	// Find all the MachineConfig pools
-	mcpPools, err := ctrl.mcpLister.List(labels.Everything())
+	sel, err := metav1.LabelSelectorAsSelector(metav1.AddLabelToSelector(&metav1.LabelSelector{}, builtInLabelKey, ""))
+	if err != nil {
+		return err
+	}
+	// Find all the MCO built in MachineConfigPools
+	mcpPools, err := ctrl.mcpLister.List(sel)
 	if err != nil {
 		return err
 	}
--- a/manifests/master.machineconfigpool.yaml
+++ b/manifests/master.machineconfigpool.yaml
@@ -4,6 +4,7 @@ metadata:
   name: master
   labels:
     "operator.machineconfiguration.openshift.io/required-for-upgrade": ""
+    "machineconfiguration.openshift.io/mco-built-in": ""
 spec:
   machineConfigSelector:
     matchLabels:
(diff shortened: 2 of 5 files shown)


pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -55,6 +55,11 @@ const (
 var (
 	// controllerKind contains the schema.GroupVersionKind for this controller type.
 	controllerKind = mcfgv1.SchemeGroupVersion.WithKind("ContainerRuntimeConfig")
+
+	defaultPools = map[string]bool{
+		"master": true,
+		"worker": true,
```
cgwalters: Hm.  I'm OK with this for now but...feels like we should have a `machineconfiguration.openshift.io/mco-built-in: ""` annotation or something on the pools.

runcom: that would be definitely better indeed

cgwalters:
/approve
and it's ok by me to land as is, looks obviously correct and will fix a spammy/confusing log message.
But holding lgtm - if you want to do https://github.com/openshift/machine-config-operator/pull/963#discussion_r303007365 - if not, let's just lgtm this for now and do it later.

runcom:
> and it's ok by me to land as is, looks obviously correct and will fix a spammy/confusing log message.
> But holding lgtm - if you want to do [#963 (comment)](https://github.com/openshift/machine-config-operator/pull/963#discussion_r303007365) - if not, let's just lgtm this for now and do it later.

I fixed that by adding the label to the built-in pools

cgwalters:
One thing I want to understand too at a higher level - hum, do we have docs for making custom pools?  I don't see it in the official docs nor ours, but clearly people are figuring it out.

This is also related to hum...that other PR/issue I can't find where we had a model for the "worker" role being "required" by custom pools.  

There's also code in the kubelet controller that's similar.  I bet what's happening now is we generate duplicate MCs?  Should the kubelet controller have a similar fix?

Anyways we can still land this as is I think...

runcom:
> There's also code in the kubelet controller that's similar. I bet what's happening now is we generate duplicate MCs? Should the kubelet controller have a similar fix?

both kubelet controller and syncContainerRuntimeConfig have a similar issue since we do not have templates for custom pools (but we should rely on the worker pool only). I can follow up with all this . as well.
The reason this is special is that syncImageConfig applies to _every_ pool, whether kubelet/cr applies only to the _selected_ pool with the selector specified in the CR. Basically, the image config was spammy because it always tried to apply to everything.

> Anyways we can still land this as is I think...

yep, I'll follow up with the others and need to think about how to do that as well



> One thing I want to understand too at a higher level - hum, do we have docs for making custom pools? I don't see it in the official docs nor ours, but clearly people are figuring it out.

Great question, I'll check if we have something official I guess - I used most of https://github.com/openshift/training/blob/master/docs/05-infrastructure-nodes.md to actually creates the additional infra nodes (but then it requires an MCP and MC as well)

kikisdeliveryservice:
> One thing I want to understand too at a higher level - hum, do we have docs for making custom pools? I don't see it in the official docs nor ours, but clearly people are figuring it out.

I looked a bit and couldn't find anything yesterday. There might be something but really not sure.

kikisdeliveryservice:
Do we need a test for this? Otherwise looks good.

runcom:
> Do we need a test for this? Otherwise looks good.

the symptom in this one is a spammy log so I'd say we can merge this

PR #964, opened 2019-07-12
OpenStack: set correct hostnames for masters and workers

Fedosin:
This commit sets hostnames equal to OpenStack node names.

### DIFF
--- a/templates/common/openstack/units/afterburn-hostname.yaml
+++ b/templates/common/openstack/units/afterburn-hostname.yaml
@@ -1,0 +1,13 @@
+name: "afterburn-hostname.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Afterburn Hostname
+  Before=kubelet.service
+
+  [Service]
+  ExecStart=/usr/bin/afterburn --provider openstack-metadata --hostname=/etc/hostname
+  Type=oneshot
+
+  [Install]
+  WantedBy=multi-user.target


templates/master/00-master/openstack/files/set-hostname.yaml
```diff
@@ -0,0 +1,22 @@
...
+    # Read more at https://docs.openstack.org/nova/latest/user/metadata.html
+    response = urllib.request.urlopen(
+        'http://169.254.169.254/openstack/2017-02-22/meta_data.json'
+    )
+
+    data = response.read()
+
+    decoded_data = json.loads(data)
+
+    with open('/etc/hostname', 'w') as f:
```
cgwalters: Please add this to https://github.com/coreos/afterburn/ instead.

cgwalters: See e.g. https://github.com/coreos/afterburn/pull/199 which landed exactly this for Azure.  By contributing to afterburn, you're also contributing to Fedora CoreOS, and the functionality works early from the OS start.

Oh also, doing this from a static pod will run straight into https://github.com/openshift/origin/pull/22826

cgwalters:
In fact afterburn [has OpenStack support today](https://github.com/coreos/afterburn/blob/master/src/providers/openstack/network.rs) except...what's up with [this change](https://github.com/coreos/afterburn/commit/2ee7e10fb2b50cce52220e13be10db865bc0a308#diff-b4aea3e418ccdb71239b96952d9cddb6R99) which calls the provider `openstack-metadata`?  Is it not the same thing as the `ignition.platform.id`?

lucab:
@cgwalters indeed. As Openstack is not an homogeneous platform and the metadata component is only optional, the static Ignition platform ID is not useful here (i.e. it can't signal whether the metadata endpoint exists or not).
Passing a `--provider openstack-metadata` is the correct way to hardcode the fact that you are now expecting/requiring the metadata endpoint.

cgwalters:
> Passing a --provider openstack-metadata is the correct way to hardcode the fact that you are now expecting/requiring the metadata endpoint.

To do that we'd have to have the MCO ship a drop-in for the systemd unit file that overrides `ExecStart=`?  A little ugly.  How about an [overdrop file](https://docs.rs/liboverdrop/0.0.2/liboverdrop/) `afterburn/provider` so we'd just `echo openstack-metadata > /usr/lib/afterburn/provider` in the RHCOS build? 

(And for FCOS + OpenStack...people would use Ignition to write to `/etc/afterburn/provider`)

cgwalters:
An implicit thing in this discussion that we should probably make explicit is this PR is trying to unconditionally set the metadata for OpenStack, but you are implicitly arguing to not to do it an require users (of FCOS) to manually enable it.

I am curious about more background behind your latter argument; it seems...easier...if we just *try* to reach the metadata service always, and require users of OpenStack-without-metadata (the unusual case, surely) to disable it.

cgwalters:
In this PR though, if metadata isn't available, then we'll have a failed systemd unit which...to my knowledge, nothing will react to or care about - something probably for the MCO to roll up in its status?  Not sure.

tomassedovic:
For what it's worth, this fixes the issue we're having with OpenStack.

A worker's hostname and therefore node & machine name is now e.g. `tsedovic-qt2cq-worker-6p9xp` as opposed to `host-172-24-0-14` (the behaviour without this patch).

We need the Nova names and node names to match for the kubernetes cloud provider to work:

https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#node-name-4

Fedosin:
@cgwalters Maybe I'm wrong, but as I see the metadata service must be enabled to deploy ocp anyway. We pass Ignition configs to MCO using metadata service: 

- We generate the config here: https://github.com/openshift/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/machine/actuator.go#L155

- And then inject the string when we boot an instance https://github.com/openshift/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/machine/actuator.go#L218

So, I wouldn't worry about the absence of the metadata service.

lucab:
I have three high-level concerns with the PR in its current form:
 1. [at least in some cases](https://github.com/openshift/installer/blob/2df090bc2e3b8296bd1cf0f8b2f17368e3e60b89/data/data/openstack/masters/main.tf#L10-L22), the installer seems to already own `/etc/hostname`. This PR now blurs the definition of  "who owns the static /etc/hostname"
 1. if the hostname without this change was `host-172-24-0-14`, then this is introducing/amplifying a skew within infrastructure inventory, which will have side-effects on other components trying to match infrastructure services to nodes (e.g. infrastructure metrics/logging analyzers)
 1. ultimately, in a cluster of networked cattle nodes, the DHCP is the source of truth for the hostnames. This guarantees that the whole infrastructure is aligned on resources naming/inventory. A static hostname (either via Ignition or via Afterburn) is the last resort option when DHCP does not provide an hostname, which does not seem to be the case here.

tomassedovic:
@lucab 

Regarding your first point: presumably the change here would apply to the initial master nodes created by the installer as well. If that is the case, we will remove the installer code that sets `/etc/hostname` because it is no longer necessary.

If not, it would have to stay because it fixes the exact same issue on those nodes.

To your points 2 and 3: getting the hostname from DHCP is simply not how OpenStack is expected to operate. Yes the `host-172-24-0-14` value does seem to come from DHCP, but it is not the correct value.

A cluster of networked *OpenStack* cattle nodes is expected to read the cloud metadata and use the hostname values specified there. Which is what this pull request does. This is what every single OpenStack cloud image I'm aware of does.

I have just launched two extra Nova servers (one CirrOS and one CentOS) in my deployed OCP cluster and their hostnames ended up being identical to their names in Nova. The hostname is initially set by the value from DHCP and it is then immediately overridden by the value from the metadata.

It seems to me that we are using Afterburn exactly what it was designed for: to update the system's state by querying the cloud provider. And we do it to get the behaviour that the OpenStack users know and expect.

I suppose the ultimate fix would be to run Afterburn (or something that does the same thing) in the OpenStack CoreOS images directly. I expect that will take longer and our cloud provider work will be blocked in the meantime.

bgilbert:
> To do that we'd have to have the MCO ship a drop-in for the systemd unit file that overrides ExecStart=? A little ugly.

The unit files are arranged so that the dropin can be:

```
[Service]
Environment=AFTERBURN_OPT_PROVIDER="--provider openstack-metadata"
```

That's how CLCT handles it, when invoked with `-platform openstack-metadata`.

> I am curious about more background behind your latter argument; it seems...easier...if we just try to reach the metadata service always, and require users of OpenStack-without-metadata (the unusual case, surely) to disable it.

I think the idea is that it's insecure to request metadata from a random link-local IP address unless we explicitly know that it's safe.

cc @ajeddeloh

cgwalters:
Although, now that I look more closely this could just go in the `templates/common/` directory right?

PR #965, opened 2019-07-12
Bug 1721586: mco: add kubelet CA rotation support

rphillips:
The kubelet client CA expires after one year. This change will follow the client-ca rotation and update cert.

**- What I did**

* changes the kubelet to use /etc/kubernetes/kubelet-ca.crt for the
  client CA file
* Adds an additional informer to the MCO to watch
  openshift-kube-apiserver-operator/kube-apiserver-to-kubelet-client-ca

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1721586

**- How to verify it**

Deleting the annotations on the following object will cause a rotation:

```
oc edit secrets -n openshift-kube-apiserver-operator kube-apiserver-to-kubelet-signer
```

```
Add Kubelet client CA cert rotation
```

/cc @sjenning @deads2k


abhinavdahiya:
will this cause reboot of machines during/immediately after the install...?

rphillips:
/hold

Currently, the ca.crt is a bundle of 5-6 certs that gets laid down. On Monday, I will split out the one cert that we need on the initial deploy. After the tweak to this PR, the `kube-apiserver-to-kubelet-client-ca` cert config map is the same as the deploy and will not reboot.

rphillips:
/hold cancel

Ready for review.

@abhinavdahiya the bundle contains the entire certificate chain, so I needed to inject the rotated cert into the bundle.

/cc @deads2k

pkg/operator/sync.go
```diff
@@ -112,10 +115,20 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
 	// as described by the name this is essentially static, but it no worse than what was here before.  Since changes disrupt workloads
 	// and since must perfectly match what the installer creates, this is effectively frozen in time.
-	kubeAPIServerServingCABytes, err := optr.getCAsFromConfigMap("openshift-config", "initial-kube-apiserver-server-ca", "ca-bundle.crt")
+	initialKubeAPIServerServingCABytes, err := optr.getCAsFromConfigMap("openshift-config", "initial-kube-apiserver-server-ca", "ca-bundle.crt")
 	if err != nil {
 		return err
 	}
+
+	// Fetch the following configmap and merge into the the initial CA. The CA is the same for the first year, and will rotate
+	// automatically afterwards.
```
kikisdeliveryservice: simple question: we are just auto-rotating whenever the cert expires?

rphillips: Yes... The openshift-kube-apiserver-operator will rotate the cert and update the configmap at roughly the 80% expiration of the cert (there is some jitter). This certificate expires after 1 year, so the update will happen somewhere around 9.6ish months.

kikisdeliveryservice: thanks for the explanation!

pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -70,6 +70,9 @@ type ControllerConfigSpec struct {
 
 	// TODO: Use string for CA data
 
+	// managed Kubelet to API Server Cert... Rotated automatically
```
LorbusChris: This comment line has to start with `kubeAPIServerServingCAData ...` (ie. the json name)

kikisdeliveryservice: @LorbusChris do you know if the TODO in line 71 above tracked anywhere?

LorbusChris: Not one that I'm aware of. There are a few TODOs in types.go (mostly for a future v2 of the api). I'd say we'll decide what to do with these/open issues once we move the apis package over to openshift/api (either here or there)

kikisdeliveryservice: sounds good, just dont want to lose track of them

sjenning:
@deads2k do you have time to review this?

pkg/apis/machineconfiguration.openshift.io/v1/types.go
rphillips: @LorbusChris fixed the comment!

rphillips:
AWS issues this morning...

/retest

templates/master/01-master-kubelet/_base/files/kubelet.yaml
```diff
@@ -7,7 +7,7 @@ contents:
     apiVersion: kubelet.config.k8s.io/v1beta1
     authentication:
       x509:
-        clientCAFile: /etc/kubernetes/ca.crt
```
cgwalters: There are a lot of other references to this path, a quick ripgrep shows e.g. a bind in
`install/0000_80_machine-config-operator_04_deployment.yaml`.

Hm, but I guess previously...we were mixing the kubelet ca with the root ca?

sjenning:
While the description is phrased like a feature, this is fixing a bug.  As as, all clusters will stop working and need manual intervention (partial DR) to get new `kube-apiserver-to-kubelet-client-ca` on the nodes after one year.  This will likely need to be backported to 4.1.z as well.

kikisdeliveryservice:
> While the description is phrased like a feature, this is fixing a bug. As is, all clusters will stop working and need manual intervention (partial DR) to get new `kube-apiserver-to-kubelet-client-ca` on the nodes when the CA expires in one year. This will likely need to be backported to 4.1.z as well.

Thanks @sjenning I updated the title to add the BZ so it can get picked up correctly. :+1: 

After this merges in master @rphillips can do a backport PR for z stream.

rphillips:
@kikisdeliveryservice thanks!

cgwalters:
> As is, all clusters will stop working and need manual intervention (partial DR) to get new kube-apiserver-to-kubelet-client-ca on the nodes when the CA expires in one year

Um...woah.  OK, we shipped a time bomb and...the fallout is just `logs`?

Is there any documentation about what these different certs are doing?

PR #970, opened 2019-07-15
hack/cluster-push: Fix for single image

cgwalters:
There are other things I want to do with this script, but at
least it works again.

### DIFF
--- a/hack/cluster-push.sh
+++ b/hack/cluster-push.sh
@@ -19,8 +19,9 @@ registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq 
 curl -k --head https://"${registry}" >/dev/null
 
 WHAT=${WHAT:-machine-config-daemon}
-LOCAL_IMGNAME=localhost/${WHAT}:latest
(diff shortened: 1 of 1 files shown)


hack/cluster-push.sh
```diff
@@ -19,8 +19,9 @@ registry=$(oc get -n openshift-image-registry -o json route/image-registry | jq
 curl -k --head https://"${registry}" >/dev/null
 
 WHAT=${WHAT:-machine-config-daemon}
-LOCAL_IMGNAME=localhost/${WHAT}:latest
-REMOTE_IMGNAME=openshift-machine-config-operator/${WHAT}
+imgname=machine-config-operator
```
kikisdeliveryservice: is there a reason why we are keeping the machine-config-daemon/WHAT above?  RHCOS?

cgwalters: So the script works the same as before basically.  As I said in the commit...I'd like to do more but this fixes the dev workflow to be the same as it was before single image.

PR #971, opened 2019-07-15
Validate nil values in scheduler spec

ravisantoshgudimetla:
**- What I did**

**- How to verify it**

Default masters schedulable to false when scheduler spec doesn't exist.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -170,7 +170,13 @@ func (ctrl *Controller) checkMasterNodesOnAdd(obj interface{}) {
 		glog.V(4).Infof("We don't care about CRs other than cluster created for scheduler config")
 		return
 	}
-	areMastersSchedulable := scheduler.Spec.MastersSchedulable
+	var areMastersSchedulable bool
+	if scheduler.Spec == (configv1.SchedulerSpec{}) {
+		glog.V(4).Infof("Scheduler spec is nil, so set master as unschedulable")
+		areMastersSchedulable = false
+	} else {
+		areMastersSchedulable = scheduler.Spec.MastersSchedulable
+	}
 	currentMasters, err := ctrl.getCurrentMasters()
 	if err != nil {
 		goerrs.Wrap(err, "Reconciling to make master nodes schedulable/unschedulable failed")


pkg/controller/node/node_controller.go
```diff
@@ -170,7 +170,13 @@ func (ctrl *Controller) checkMasterNodesOnAdd(obj interface{}) {
 		glog.V(4).Infof("We don't care about CRs other than cluster created for scheduler config")
 		return
 	}
-	areMastersSchedulable := scheduler.Spec.MastersSchedulable
+	var areMastersSchedulable bool
+	if scheduler.Spec == (configv1.SchedulerSpec{}) {
+		glog.V(4).Infof("Scheduler spec is nil, so set master as unschedulable")
+		areMastersSchedulable = false
```
kikisdeliveryservice: if a bool default value is false, isn't the nil value covered in the original line?  is this change more for the logging or am i missing something?

ravisantoshgudimetla: You mean why explicit setting to false if it's already false(the default value for bool is false)? If yes, it's just for my sanity. As the zen of python says `explicit better than implicit`.

kikisdeliveryservice: bool default value in go is false so it seems redundant to do this check unless the logging you are adding has a high value.

kikisdeliveryservice: also the default value of areMastersSchedulable is false so we don't have to reset it there.

pkg/controller/node/node_controller.go
```diff
@@ -170,7 +170,13 @@ func (ctrl *Controller) checkMasterNodesOnAdd(obj interface{}) {
 		glog.V(4).Infof("We don't care about CRs other than cluster created for scheduler config")
 		return
 	}
-	areMastersSchedulable := scheduler.Spec.MastersSchedulable
+	var areMastersSchedulable bool
+	if scheduler.Spec == (configv1.SchedulerSpec{}) {
+		glog.V(4).Infof("Scheduler spec is nil, so set master as unschedulable")
```
runcom: spec isn't really nil - why don't we change the `MastersSchedulable` within spec to be a `*bool` while we're still in time to do that?

ravisantoshgudimetla: Ya, technically, it's empty struct, I am checking with. I see your point but I think having `bool` is fine too, isn't it?

runcom: well, yeah, it's ok but I think having a pointer to differentiate set/unset (as this case) would be cleaner, no?

ravisantoshgudimetla: >  but I think having a pointer to differentiate set/unset (as this case) would be cleaner, no?

True but we're comparing Spec here which is a struct, not a pointer to struct and the `*bool` is for `MastersSchedulable` field which we added recently. For primitive types like `bool`, I think it's ok not to have pointers and considering we have 3 more days left before feature freeze

pkg/controller/node/node_controller.go
cgwalters: I generally agree that the extra logging feels...unnecessary but I don't care a lot either.
Take this as a "lgtm" from me, will defer to @kikisdeliveryservice for final lgtm.

kikisdeliveryservice: the more i think about it the more i feel that this should be verified via a test in https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller_test.go or elsewhere.  the check in the code doesn't change any functionality (and imho introduces a bit of confusion at first glance) if you are worried about the treatment of nil values, ensuring they are taken care of via a test makes more sense to me.

kikisdeliveryservice: ie add a test that the existing code fails and if so use this change. if the code does not fail, use as is and keep the test to ensure nothing breaks nil values in the future.

kikisdeliveryservice:
going to approve this and @ravisantoshgudimetla is going to check to see whether or not it's necessary to add a test around any of this behavior and if so will add one via a follow-on pr.

/lgtm

kikisdeliveryservice:
probably wait on the retest since hitting:

`"Error: Error creating S3 bucket: TooManyBuckets: You have attempted to create more buckets than allowed"`

PR #972, opened 2019-07-15
pkg/controller: clarify platform none constant in template/render.go

### LINKED ISSUE #857: controller: empty platform vs platform none
So reading through render.go, I'm wondering why we treat an empty platform as an error and not a platformNone? Mostly I'm looking at:
https://github.com/openshift/machine-config-operator/blob/0d843aa29c6a5ce5aedcd1cb1638d2383ea17e41/pkg/controller/template/render.go#L123

```
func platformFromControllerConfigSpec(ic *mcfgv1.ControllerConfigSpec) (string, error) {
	switch ic.Platform {
	case "":
		return "", fmt.Errorf("cannot generate MachineConfigs with an empty platform field")
	case platformBase:
		return "", fmt.Errorf("platform _base unsupported")
	case platformAWS, platformAzure, platformBaremetal, platformOpenstack, platformLibvirt, platformNone:
		return ic.Platform, nil
	default:
		glog.Warningf("Warning: the controller config referenced an unsupported platform: %s", ic.Platform)
		return platformNone, nil
	}
}
```

if platform is ""/ empty it's not construed as None. For it to be none it literally has to be "None", which ok, but should maybe the Error say something about that at least? `cannot generate MachineConfigs with an empty platform field, to indicate no platform use "platform: None"` or something?

Just wondering how clear this is and what we're trying to cover with the "" check.

sgreene570:
Closes #857

**- What I did**
Clarify platform none vs "" in template platform code.
https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L122
The installer explicitly sets the `Platform` field to be `""` when a platform type is not specified.
Also, unsupported Platform fields should not be considered as `platformNone` since technically
`platformNone` is a [hidden-but-supported](https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L36) platform.

**- How to verify it**

pkg/controller: clarify platform none constant in template/render.go

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -124,12 +124,16 @@ func GenerateMachineConfigsForRole(config *RenderConfig, role, templateDir strin
 func platformFromControllerConfigSpec(ic *mcfgv1.ControllerConfigSpec) (string, error) {
 	switch ic.Platform {
 	case "":
-		return "", fmt.Errorf("cannot generate MachineConfigs with an empty platform field")
+		// if Platform is nil, return nil platform and an error message
+		return "", fmt.Errorf("cannot generate MachineConfigs when no platform is set")
 	case platformBase:
 		return "", fmt.Errorf("platform _base unsupported")
 	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenstack, platformLibvirt, platformNone:
 		return ic.Platform, nil
 	default:
+		// platformNone is used for a non-empty, but currently unsupported platform.
+		// This allows us to incrementally roll out new platforms across the project
+		// by provisioning platforms before all support is added.
 		glog.Warningf("Warning: the controller config referenced an unsupported platform: %s", ic.Platform)
 		return platformNone, nil
 	}


sgreene570:
assign @kikisdeliveryservice

sgreene570:
> @sgreene570
> 
> I am reading from here: [da21a3a](https://github.com/openshift/machine-config-operator/commit/da21a3a4ef874f386a641c522d2c2b145f71f544)
> 
> that None is a non-nil platform value that is unsupported currently in OpenShift? Whereas empty is just an empty/nil platform value correct? ie, I think your change from none to nil is not correct.
> 
> But the naming is confusing (should really be platform.Unsupported or something), so better/clarifying comments there (now that we have figured this out) would be great!

Thanks for input...updated the comments to make things easier to read :D

pkg/controller/template/render.go
```diff
@@ -124,12 +124,14 @@ func GenerateMachineConfigsForRole(config *RenderConfig, role, templateDir strin
...
 	case "":
-		return "", fmt.Errorf("cannot generate MachineConfigs with an empty platform field")
+		// if Platform is nil, return nil platform and an error message
+		return "", fmt.Errorf("cannot generate MachineConfigs when no platform is set")
 	case platformBase:
 		return "", fmt.Errorf("platform _base unsupported")
 	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenstack, platformLibvirt, platformNone:
 		return ic.Platform, nil
 	default:
+		// platformNone is used for a non-empty, but currently unsupported platform.
```
kikisdeliveryservice: how about: "..unsupported platform. This allows us to incrementally roll new platforms out across the project
by provisioning platforms before all support is added."

sgreene570: Fixed! Nicely put.

kikisdeliveryservice:
I think the added comments on the None platform will def help people not be confused going forward.

/lgtm

PR #973, opened 2019-07-16
OpenStack: Set owners for OpenStack templates

Fedosin:


OWNERS_ALIACES
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
```
sgreene570: nit: file name is not correct?

kikisdeliveryservice: not a nit, it would be a blocker.

kikisdeliveryservice: :)

Fedosin: Yeah, my bad :( Thanks for your comment

Fedosin:
@sgreene570 @kikisdeliveryservice folks, what do you think now?

PR #974, opened 2019-07-16
pkg/controller: add a node hold API

runcom:

**- What I did**

The storage team has an additional requirement when calculating if a node is ready. The node is ready and the node controller can proceed to target the next node in the pool iif the Ceph cluster is ready. Before this patch, we calculate availability to proceed only based on node readiness. To account for their additional requirement (and given the short notice/period before the freeze), this patch adds a simple mechanism to _hold_ a node from being targeted for an upgrade/config change. The mechanism is driven by an annotation containing a counter. The reason for the counter is the possibility for more than one component to interact with the functionality. The API is pretty simple, a component set the annotation and set the counter (either initializing the annotation to 1 or just by increasing the counter). External components are also responsible to decrease the counter when ready to proceed.

This is far from being ideal and also, we can't use the Paused field on an MCP cause we can't differentiate more than user interacting with the field.

I'll put this on hold but I've tested it out and works fine as is - leaving hold for code review (especially naming and further checks).

The whole discussion started here https://github.com/openshift/machine-config-operator/issues/662#issuecomment-510057616

/cc @rohantmp

**- How to verify it**

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
@@ -221,11 +222,24 @@ func isNodeReady(node *corev1.Node) bool {
 	return checkNodeReady(node) == nil
 }
 
+func isNodeOnHold(node *corev1.Node) bool {
+	owners, ok := node.Annotations[daemonconsts.NodeOnHoldAnnotationKey]
+	if !ok {
+		return false
+	}
+	ownersInt, _ := strconv.Atoi(owners)
+	// if ownersInt < 0, someone is messing up, reconciles to 0 ourselves
+	return ownersInt != 0
+}
+
 // isNodeUnavailable is a helper function for getUnavailableMachines
 // See the docs of getUnavailableMachines for more info
 func isNodeUnavailable(node *corev1.Node) bool {
 	// Unready nodes are unavailable
 	if !isNodeReady(node) {
+		return true
+	}
+	if isNodeOnHold(node) {
 		return true
 	}
 	// Ready nodes are not unavailable
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -14,6 +14,8 @@ const (
 	DesiredMachineConfigAnnotationKey = "machineconfiguration.openshift.io/desiredConfig"
 	// MachineConfigDaemonStateAnnotationKey is used to fetch the state of the daemon on the machine.
 	MachineConfigDaemonStateAnnotationKey = "machineconfiguration.openshift.io/state"
+	// NodeOnHoldAnnotationKey is used by external components to put a node on hold from being accounted for in the node controller
+	NodeOnHoldAnnotationKey = "machineconfiguration.openshift.io/hold"
 	// MachineConfigDaemonStateWorking is set by daemon when it is applying an update.
 	MachineConfigDaemonStateWorking = "Working"
 	// MachineConfigDaemonStateDone is set by daemon when it is done applying an update.
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -432,7 +432,7 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 		glog.Infof("Pool %s: node %s has completed update to %s", pool.Name, curNode.Name, curNode.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey])
 		changed = true
 	} else {
-		annos := []string{daemonconsts.CurrentMachineConfigAnnotationKey, daemonconsts.DesiredMachineConfigAnnotationKey, daemonconsts.MachineConfigDaemonStateAnnotationKey}
+		annos := []string{daemonconsts.CurrentMachineConfigAnnotationKey, daemonconsts.DesiredMachineConfigAnnotationKey, daemonconsts.MachineConfigDaemonStateAnnotationKey, daemonconsts.NodeOnHoldAnnotationKey}
 		for _, anno := range annos {
 			if oldNode.Annotations[anno] != curNode.Annotations[anno] {
 				glog.Infof("Pool %s: node %s changed %s = %s", pool.Name, curNode.Name, anno, curNode.Annotations[anno])


runcom:
These are the controller logs from a config change, after holding a node after the first one rolled out the config change:

```
I0716 14:36:30.881743       1 node_controller.go:432] Pool worker: node ip-10-0-167-91.ec2.internal has completed update to rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:36:39.803523       1 node_controller.go:425] Pool worker: node ip-10-0-167-91.ec2.internal is now reporting ready
I0716 14:37:30.858455       1 node_controller.go:438] Pool worker: node ip-10-0-155-156.ec2.internal changed machineconfiguration.openshift.io/hold = 2
I0716 14:37:46.408678       1 node_controller.go:438] Pool worker: node ip-10-0-155-156.ec2.internal changed machineconfiguration.openshift.io/hold = 1
I0716 14:37:56.327397       1 node_controller.go:438] Pool worker: node ip-10-0-155-156.ec2.internal changed machineconfiguration.openshift.io/hold = 
I0716 14:38:01.327778       1 node_controller.go:709] Setting node ip-10-0-155-156.ec2.internal to desired config rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:38:01.346343       1 node_controller.go:438] Pool worker: node ip-10-0-155-156.ec2.internal changed machineconfiguration.openshift.io/desiredConfig = rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:38:02.360587       1 node_controller.go:438] Pool worker: node ip-10-0-155-156.ec2.internal changed machineconfiguration.openshift.io/state = Working
I0716 14:38:02.490078       1 node_controller.go:423] Pool worker: node ip-10-0-155-156.ec2.internal is now reporting unready: node ip-10-0-155-156.ec2.internal is reporting Unschedulable
I0716 14:38:44.125754       1 node_controller.go:423] Pool worker: node ip-10-0-155-156.ec2.internal is now reporting unready: node ip-10-0-155-156.ec2.internal is reporting NotReady
I0716 14:40:31.343978       1 node_controller.go:432] Pool worker: node ip-10-0-155-156.ec2.internal has completed update to rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:40:40.321042       1 node_controller.go:425] Pool worker: node ip-10-0-155-156.ec2.internal is now reporting ready
I0716 14:40:41.352649       1 node_controller.go:709] Setting node ip-10-0-131-80.ec2.internal to desired config rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:40:41.373051       1 node_controller.go:438] Pool worker: node ip-10-0-131-80.ec2.internal changed machineconfiguration.openshift.io/desiredConfig = rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:40:42.386989       1 node_controller.go:438] Pool worker: node ip-10-0-131-80.ec2.internal changed machineconfiguration.openshift.io/state = Working
I0716 14:40:42.498581       1 node_controller.go:423] Pool worker: node ip-10-0-131-80.ec2.internal is now reporting unready: node ip-10-0-131-80.ec2.internal is reporting Unschedulable
I0716 14:41:19.433843       1 node_controller.go:423] Pool worker: node ip-10-0-131-80.ec2.internal is now reporting unready: node ip-10-0-131-80.ec2.internal is reporting NotReady
I0716 14:43:10.353786       1 node_controller.go:432] Pool worker: node ip-10-0-131-80.ec2.internal has completed update to rendered-worker-f0aea22749bfa8e16a616d27c96849c2
I0716 14:43:19.337095       1 node_controller.go:425] Pool worker: node ip-10-0-131-80.ec2.internal is now reporting ready
I0716 14:43:20.363591       1 status.go:83] Pool worker: All nodes are updated with rendered-worker-f0aea22749bfa8e16a616d27c96849c2
```

cgwalters:
A PR with a hold that implements holds!  We're back in full Braveheart territory :smile:

runcom:
> A PR with a hold that implements holds! We're back in full Braveheart territory

it's even funnier if you think at github labels as annotations

pkg/controller/node/status.go
```diff
@@ -221,13 +222,26 @@ func isNodeReady(node *corev1.Node) bool {
 	return checkNodeReady(node) == nil
 }
 
+func isNodeOnHold(node *corev1.Node) bool {
```
kikisdeliveryservice: Could we add comment here explaining the what/why?  like from your pr description?

pkg/daemon/constants/constants.go
```diff
@@ -14,6 +14,8 @@ const (
 	DesiredMachineConfigAnnotationKey = "machineconfiguration.openshift.io/desiredConfig"
 	// MachineConfigDaemonStateAnnotationKey is used to fetch the state of the daemon on the machine.
 	MachineConfigDaemonStateAnnotationKey = "machineconfiguration.openshift.io/state"
+	// NodeOnHoldAnnotationKey is used by external components to put a node on hold from being accounted for in the node controller
```
kikisdeliveryservice: can we also mention something about "hold a node from being targeted for an upgrade/config change" from your pr desc here too?

pkg/daemon/constants/constants.go
```diff
@@ -14,6 +14,8 @@ const (
...
+	NodeOnHoldAnnotationKey = "machineconfiguration.openshift.io/hold"
```
cgwalters: I was thinking about this in the context of reboot requests...how about "prefix annotations", the API is to e.g. add
`machineconfiguration.openshift.io/hold-ceph = "Hold placed by ceph operator"` and we scan for all annotations with that prefix?

This makes them more self-describing and auditable.

runcom: oh yeah - I went that road but I've been buried into kubernetes specs around annotation and labels that I didn't want to _break_ the way labels/annotations are specified.  I was thinking about `hold.machineconfiguration.openshift.io/<component>` prefixto have an easy split at the slash also. I don't know, I don't feel strongly about this and since, at least for the next release, this is just used by one component we can revisit for 4.3 (we can also mark this beta for 4.2)

cgwalters:
One thing that seems tricky here is for the "holder code" to reconcile its state; basically how does it know whether or not one of those increments are "owned" by it?  This would also be solved with a prefix approach or anything really which acted as a "named slot".

As is today it seems like the operator would need to maintain a configmap or other state with the set of nodes it has holds on...but what happens if e.g. it successfully patches the node to drop the hold, then the operator gets OOM killed before it can patch its configmap; now the next time it starts it's going to drop a hold it doesn't own...

runcom:
> As is today it seems like the operator would need to maintain a configmap or other state with the set of nodes it has holds on...but what happens if e.g. it successfully patches the node to drop the hold, then the operator gets OOM killed before it can patch its configmap; now the next time it starts it's going to drop a hold it doesn't own...

uhm, if we go the named annotation approach, are we concerned about this also? Since this acts like an api it's the component responsibility to track all this and we (MCO) adhere to what we're being told

> One thing that seems tricky here is for the "holder code" to reconcile its state; basically how does it know whether or not one of those increments are "owned" by it? This would also be solved with a prefix approach or anything really which acted as a "named slot".

This sounds right in case something goes wrong and we need to point to someone holding an upgrade  or config change for instance

cgwalters:
> uhm, if we go the named annotation approach, are we concerned about this also? Since this acts like an api it's the component responsibility to track all this and we (MCO) adhere to what we're being told

What I'm saying is that the API seems inherently subject to race conditions.  I gave one around deletions, but there's the same for additions.

If we're only targeting this for exactly one external user for 4.2, we could just make it an explicit `bool`.  Then there's no races, we clearly only support at most one external component adding a hold.

runcom:
> If we're only targeting this for exactly one external user for 4.2, we could just make it an explicit `bool`. Then there's no races, we clearly only support at most one external component adding a hold.

so, I've been thinking about all of this for a while today. I can see two options going forward. This is a readiness check mainly, not an _hold_ requierment (this is also, not strictly speaking, a cluster requirement, not a node one but we'll go with that). One way to solve this is to provide an api that components must implement and the MCC calls into to check if a given node is ready, we can spec this out and have an object drive all this or we can go the in-process way using grpc plugins (as highlighted elsewhere for a similar problem). 
The other option would be to invert how all this work. The MCC holds an object where components write to to signal a node is/isn't ready for them. That should solve the race issue as the external component now doesn't need to track 2 objects (right?).

runcom:
> If we're only targeting this for exactly one external user for 4.2, we could just make it an explicit `bool`. Then there's no races, we clearly only support at most one external component adding a hold.

having said the above in my other comment - I think this needs a deeper design discussion, I'll create a card to track it for 4.3 and meanwhile we can go with the bool (even if I hate it)

cgwalters:
> having said the above in my other comment - I think this needs a deeper design discussion, I'll create a card to track it for 4.3 and meanwhile we can go with the bool (even if I hate it)

To be clear...I am not saying I'm against this PR as is; the races here may be unlikely in practice[1].  So if you want to push back and say "let's ship as is" I'm not standing in the way necessarily.

[1] Although if they do happen it won't be good, there will either be nodes stuck or multiple updating at once.

runcom:
> To be clear...I am not saying I'm against this PR as is; the races here may be unlikely in practice[1]. So if you want to push back and say "let's ship as is" I'm not standing in the way necessarily.

I think the risk of getting stuck or disrupting Ceph is higher than shipping a safe stopgap as a knob as you suggested.

(also, I wrote "I hate the book" meaning I hate to carry any stopgap, not the one you suggested :smile:)

smarterclayton:
/hold

At minimum needs api review - I’m deeply skeptical of allowing MCC to infer readiness on components of the node.  Will review in more detail but this at first glance doesn’t seem in keeping with our system principles and I want to be convinced workloads should be allowed to be treated specially.

smarterclayton:
At least on first glance the storage component needs to delay drain arbitrarily, which is not the responsibility of MCC or something that the node should impact.

If ceph needs this ceph needs to control the PDB in more detail

runcom:
> If ceph needs this ceph needs to control the PDB in more detail

Right, that's what we came up at the beginning but it seems PDB (or MDB) isn't enough from conversation and tests that we had with the Ceph team.

rohantmp:
@smarterclayton
Our constraints are:
- We cannot have OSDs down on more than one node at once.
- Once a node has been brought down, we cannot tolerate further disruption until the whole cluster is rebalanced and healthy.

PDB will not work for ceph in an environment with more than one OSD pod per node.
There are two cases:

- PDB maxUnavailable <= 1: No node can drain
- PDB maxUnavailable > 1: There is a chance that OSDs on more than one node are drained at a time which can cause data availability loss and data loss.

rohantmp:
> If ceph needs this ceph needs to control the PDB in more detail

IIRC the budget part of the PDB does not support update. Will update this comment after verifying.

eparis:
This should not merge in 4.2.
/hold
again.

runcom:
gonna just close this one, thanks

PR #975, opened 2019-07-16
OpenStack: enable cloud provider

Fedosin:
One of the requirements for starting the cloud provider is the ability to read data from the secret.
Despite the fact that the cloud provider supports this feature, there are a number of limitations that do not allow to run it at the initial stage.
The peculiarity of the cloud provider is that it requires a list of nodes from Nova at the earliest stage of kubelet loading, when the certificates required for reading the secret have not been received yet and the kubeconfig file has not been generated yet.
The absence of this file leads to a complete failure of kubelet and the impossibility to continue the installation.

As a solution, changes were proposed to the kubelet itself, both in the upstream and downstream versions. But since these are very serious changes, and the cloud provider was deprecated, we will
probably not be able to merge them.
https://github.com/kubernetes/kubernetes/pull/79221
https://github.com/openshift/origin/pull/23290

The following commit was offered as an alternative, more convenient solution. The idea is to run kubelet without cloud provider support for a certain period of time so that it can generate the necessary kubeconfig file (generate-kubeconfig.service) and then run the real kubelet with cloud provider support (kubelet.service).

### DIFF
--- a/templates/master/00-master/openstack/units/afterburn-hostname.yaml
+++ b/templates/master/00-master/openstack/units/afterburn-hostname.yaml
@@ -1,0 +1,12 @@
+name: "afterburn-hostname.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Afterburn Hostname
+
+  [Service]
+  ExecStart=/usr/bin/afterburn --provider openstack-metadata --hostname=/etc/hostname
+  Type=oneshot
+
+  [Install]
+  WantedBy=multi-user.target
--- a/templates/master/01-master-kubelet/openstack/files/generate-kubeconfig.yaml
+++ b/templates/master/01-master-kubelet/openstack/files/generate-kubeconfig.yaml
@@ -1,0 +1,24 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/generate-kubeconfig.sh"
+contents:
+  inline: |
+    #!/bin/bash
+
+    /usr/bin/timeout 900 /bin/bash -c -- \
+    'while true; do \
+      /usr/bin/hyperkube \
+      kubelet \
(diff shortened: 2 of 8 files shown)


derekwaynecarr:
what is the objection with https://github.com/openshift/origin/pull/23290 ?

/cc @sjenning i thought you were reviewing that.  its a carry for the life of in-tree cloud provider.

Fedosin:
@derekwaynecarr it's just an solution that works... I suggested it as an alternative in case we don't have time to test https://github.com/openshift/origin/pull/23290

The problems with https://github.com/openshift/origin/pull/23290 are:
1. it seems like upstream patch won't be merged.
2. I can't test it properly because I need a modified RHCOS image with a custom hyperkube binary. Since I can't build it myself, I am afraid that something could go wrong there and break the deployment completely.
I tested the binary on a deployed ocp cluster and it worked for me, but it's not enough, obviously.
If there was a way for me to build the image manually and test https://github.com/openshift/origin/pull/23290 before merging it, that would be great, but looks like it's not possible.

In order not to bother you and @sjenning with image building and testing of my patches, I have prepared this alternative that works fine. We can use it as a temporary solution for 4.2 with in-tree cloud provider and switch to the external one in 4.3.

cgwalters:
> I can't test it properly because I need a modified RHCOS image with a custom hyperkube binary. Since I can't build it myself

https://github.com/coreos/coreos-assembler
https://url.corp.redhat.com/redhat-coreos-config

sjenning:
I'm a nack on this in favor of https://github.com/openshift/origin/pull/23290

cgwalters:
<voice type="Mel Gibson">
/hold
</voice>

Issue #976, opened 2019-07-16
ci failures: bootkube.sh: Invalid character(s) found in major number "machine-config-daemon-4"


kikisdeliveryservice:
**Description**
Seeing a pattern of bootkube.sh CI failures beginning around 4 hours ago.

**Steps to reproduce the issue:**
1. looking at e2e ci runs that fail around ~45-5x minutes with exit code 1
2. lookign at log-bundle-xxx.tar -> /bootstrap/journals/bootkube.log

**Describe the results you received:**
```
Jul 16 21:23:18 ip-10-0-11-239 systemd[1]: Started Bootstrap a Kubernetes cluster.
Jul 16 21:23:19 ip-10-0-11-239 podman[7844]: 2019-07-16 21:23:19.784564535 +0000 UTC m=+0.315776402 container create aec1e9a6654e8c8f562654b97d95329152b57635a5853e86fef6c1833dbd7cba (image=registry.svc.ci.openshift.org/ci-op-9n865rsi/release@sha256:0e56f818093ef4777a5e187b91375fb98d2b9b8952e118e59c5e0b093ae4eb95, name=nostalgic_neumann)
Jul 16 21:23:20 ip-10-0-11-239 podman[7844]: 2019-07-16 21:23:20.461016745 +0000 UTC m=+0.992228687 container init aec1e9a6654e8c8f562654b97d95329152b57635a5853e86fef6c1833dbd7cba (image=registry.svc.ci.openshift.org/ci-op-9n865rsi/release@sha256:0e56f818093ef4777a5e187b91375fb98d2b9b8952e118e59c5e0b093ae4eb95, name=nostalgic_neumann)
Jul 16 21:23:20 ip-10-0-11-239 podman[7844]: 2019-07-16 21:23:20.47647134 +0000 UTC m=+1.007683402 container start aec1e9a6654e8c8f562654b97d95329152b57635a5853e86fef6c1833dbd7cba (image=registry.svc.ci.openshift.org/ci-op-9n865rsi/release@sha256:0e56f818093ef4777a5e187b91375fb98d2b9b8952e118e59c5e0b093ae4eb95, name=nostalgic_neumann)
...
[...]

PR #977, opened 2019-07-16
Stop parsing our own version as semver

cgwalters:
Since we now have a single image, we don't need to worry about
drift between them.

Previously we were trying to parse the output
of `git describe` (trimmed) as a semantic version, which blew
up when ART pushed a git tag that didn't conform to the previous pattern.

Just refer to the literal raw tag name instead of trying to parse it.

Closes: https://github.com/openshift/machine-config-operator/issues/976


kikisdeliveryservice:
@cgwalters updated the unit tests for the new version. they should pass now.

kikisdeliveryservice:
This is great and passes e2e.

/lgtm

PR #984, opened 2019-07-17
Update for openshift/api changes to ImageContentSourcePolicy

mtrmac:
<s>DO NOT MERGE as is: Includes an UNMERGED PR to openshift/api.</s>

**- What I did**

https://github.com/openshift/api/pull/384 modifies `ImageContentSourcePolicy` to be explicit about sources == pull specs, and mirrors, instead of defining sets of repositories that can be mirrors of each other.

This updates the vendored API and the relevant code.

**- How to verify it**

Updated unit tests, I suppose manual testing that it all works.

Updated for ImageContentSourcePolicy distinguishing mirrors from primary repositories


bcrochet:
Also needed for #795

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -210,9 +225,10 @@ func TestUpdateRegistriesConfig(t *testing.T) {
 			icspRules: []*apioperatorsv1alpha1.ImageContentSourcePolicy{
 				{
 					Spec: apioperatorsv1alpha1.ImageContentSourcePolicySpec{
-						RepositoryDigestMirrors: []apioperatorsv1alpha1.RepositoryDigestMirrors{
-							{Sources: []string{"insecure.com/ns-i1", "blocked.com/ns-b1", "unrelated.com/ns-u1"}},
-							{Sources: []string{"blocked.com/ns-b/ns2-b", "unrelated.com/ns-u2", "insecure.com/ns-i2"}},
+						RepositoryDigestMirrors: []apioperatorsv1alpha1.RepositoryDigestMirrors{ // other.com is neither iecure nor blocked
```
sgreene570: nit: spelling, secure?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -253,87 +253,67 @@ func scopeMatchesRegistry(scope, reg string) bool {
...
 }
 
-// disjointOrderedMirrorSets processes icspRules and returns the resulting disjoint sets of mutual mirrors, each ordered in consistently with
-// the icspRules preference order (if possible)
-// This exists to merge different mirror sets, e.g. given mirror sets (B, C) and (A, B), it will combine them into a single (A, B, C) set.
-func disjointOrderedMirrorSets(icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy) ([][]string, error) {
-	// Collect all mirror sets into a flat array to abstract the rest of the code from the ImageContentSourcePolicy format
-	mirrorSets := []*apioperatorsv1alpha1.RepositoryDigestMirrors{}
+// mergedMirrorSets processes icspRules and returns a set of RepositoryDigestMirrors, one for each Source value,
+// ordered in consistently with the preference order of the individual entries (if possible)
```
sgreene570: nit: "ordered consistently" instead of "ordered in consistently" ?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -253,87 +253,67 @@ func scopeMatchesRegistry(scope, reg string) bool {
...
+			// Every node in topoGraph, including source, is implicitly added by topoGraph.AddEdge (there are no unconnected nodes that we would
+			// have to add separately from the edges).
 		}
 		sortedRepos, err := topoGraph.Sorted()
 		if err != nil {
 			return nil, err
 		}
-		res = append(res, sortedRepos)
+		if sortedRepos[len(sortedRepos)-1] == source {
+			// We don't need to explicitly include source in the list, it will be automatically tried last per the semantics of sysregistriesv2. Mirrors.
```
sgreene570: Is it harmful to leave source in the list? Would it be possible to prevent source from being added to the list rather than removing it later?

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -39,98 +39,113 @@ func TestScopeMatchesRegistry(t *testing.T) {
...
 				{ // Vendor2-provided default configuration
-					{"registry1.vendor2.com", "registry2.vendor2.com"},
+					{Source: "source.vendor2.com", Mirrors: []string{"registry1.vendor2.com", "registry2.vendor2.com"}},
 				},
 				{ // Admin-configured local mirrors:
-					{"local-mirror.example.com", "registry1.vendor.com"},
-					{"local-mirror2.example.com", "registry2.vendor2.com", "registry1.vendor2.com"}, // Opposite order of the vendor’s mirrors
+					{Source: "source.vendor.com", Mirrors: []string{"local-mirror.example.com"}},
+					// Opposite order of the vendor’s mirrors.
+					// WARNING: The order in this case is unspecified by the ICSP specification, and may change at any time;
```
sgreene570: This comment is identical to the previous test comment. Are these 2 tests testing the same thing? Trying to follow along.

pkg/controller/container-runtime-config/helpers_test.go
mtrmac: “insecure”. Thanks, fixing.

pkg/controller/container-runtime-config/helpers.go
mtrmac: Right now not removing `source` from `sortedRepos` would result in an unnecessary extra attempt to contact that registry (if the image were not available); it’s not catastrophic but it is unnecessary. Plus, removing `source` from the list results in `registries.conf` more similar to how a human would write it, especially for the trivial case of a single `RepositoryDigestMirrors`.

Yes, not adding the source to the list instead of removing it here would be problematic: Consider
```go
[]RepositoryDigestMirrors{
  {Source: "S", Mirrors: {"A"}} // Expected order of attempts: AS
  {Source: "S", Mirrors: {"BSC"}} // Expected order of attempts: BSC(S) , the second S being redundant/pointless/not applicable
}
```
We need to add the AS edge to `topoGraph` so that the merged order in `sortedRepos` is one of ABSC/BASC; without it, it could end up e.g. BSCA (or maybe BSCA(S)), unnecessarily inconsistent with the input entries.

(Sure, this is all corner cases. Everything about including `Source` as a member of `Mirrors` is a corner case. Originally I had an assert “`Source` is last” in here, then I considered the “`Source` is included in `Mirrors`” counterexample, and it turned out simpler/clearer/more general to just implement that fully than to detect the case and fail.)

pkg/controller/container-runtime-config/helpers_test.go
mtrmac: They test two different things:
- `Source` is included in `Mirrors`
- Two `RepositoryDigestMirrors` items contain items in `Mirrors` in conflicting order.

ICSP does not _document_ how such cases are resolved, it’s not part of the API guarantee.

OTOH, MCC does have an interest of keeping the output stable over time, so that the controller does not issue unnecessary updates to the configuration.  So, the implementation sorts the items to behave consistently over time, and these tests hard-code that implementation detail.

The warnings are added to decrease the chance that someone will read the code and the tests, and conclude that the current order is an API promise. Sure, it’s pretty unlikely that anyone will read that, if you think the warnings are too confusing to be worth that, I’d be happy to remove them.

pkg/controller/container-runtime-config/helpers.go
mtrmac: Thanks, fixed.

pkg/controller/container-runtime-config/helpers.go
sgreene570: OK I understand. Thank you for elaborating.

pkg/controller/container-runtime-config/helpers_test.go
sgreene570: I think the warnings are a good idea, since like you said there is room for confusion. Just wanted to make sure that the warning did need to be present in two places.

runcom:
needs rebase

mtrmac:
Rebased.

LorbusChris:
@mtrmac could you try updating just the api dep?
`dep ensure -update github.com/openshift/api`

mtrmac:
Sure, limited the vendor update.

LorbusChris:
we're unfortunately still hitting resource limits on AWS :(
/lgtm
nonetheless, let's hope this gets in quickly

PR #986, opened 2019-07-18
*: remove Dockerfile.machine-config-daemon.upstream

runcom:

**- What I did**

not really used anymore now that we only have 1 image

**- How to verify it**


kikisdeliveryservice:
Wasn't sure if anyone else was using this for some other purpose?

@cgwalters can you verify it's ok to remove?

runcom:
> Wasn't sure if anyone else was using this for some other purpose?
> 
> @cgwalters can you verify it's ok to remove?

yeah, it's not referenced by anything in the current code and the reason why we added it (no entitlements on UBI) seems to be solved since it's working today

kikisdeliveryservice:
> yeah, it's not referenced by anything in the current code and the reason why we added it (no entitlements on UBI) seems to be solved since it's working today

Ahh I think I was confused. In that case:

/lgtm

kikisdeliveryservice:
almost there!

/retest

kikisdeliveryservice:
gah timing out..

/retest

PR #991, opened 2019-07-18
Tighten scheduler permissions in mco clusterrole

ravisantoshgudimetla:
**- What I did**
Reduce the permissions of mco to `watch`,`list` and `get` scheduler config

**- How to verify it**


runcom:
Looks sane since we're not updating the object ourselves

/approve
/lgtm

ravisantoshgudimetla:
`"openshift-machine-api/cluster-autoscaler-operator\" (144 of 431): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-machine-api/machine-api-operator\"`

PR #996, opened 2019-07-22
 Ignition types.Config (v0s2) to runtime.RawExtension conversion

LorbusChris:
**- What I did**
Convert Ignition types.Config to runtime.RawExtension
    
In order to decouple the MCO's API from the Ignition types,
this PR introduces conversion to and from RawExtension.

When reading or mutating a config, the RawExtension is now converted to
a typed Ignition config internally upon receival, and then converted
back to RawExtension for passing it around externally.

Using RawExtension does not change the wire-format of the
transmitted data. The contents of MachineConfig.Spec.Config
will still be in JSON format and readable as such.

Move MergeMachineConfigs func into controller/common pkg,
in preparation of decoupling the API from Ignition Config types.

This PR should fully unblock the migration of the `api` package to the `openshift/api` repo.

I'd advise to review each commit separately.

**- How to verify it**
CI

Added Ignition types.Config (version 0.x spec 2.x) to runtime.RawExtension conversion


pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -21,13 +92,28 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
 	var fips bool
-	outIgn := configs[0].Spec.Config
+
```
kikisdeliveryservice: would like a comment explaining why this encoding/decoding is happening

kikisdeliveryservice:
I'm still going through this, but can you break this up into 2 commits and break out the test changes to make it easier to view?

kikisdeliveryservice:
@runcom @LorbusChris  Still going through this, but basic question to check my understanding: Currently: Machineconfig.Spec.Config is an Ignition Config.
After this PR: Spec.Config is going to be a []byte? 

Won't this make troubleshooting/looking at MachineConfigs generally impossible from a user perspective? Am I missing something????

LorbusChris:
@kikisdeliveryservice please hold off a thorough review for now, this still needs polishing.

With this PR, Spec.Config.Raw is going to be []byte. It's not pretty, but it'll allow us to use Ignition Config spec2 as well as spec3 (by encoding/decoding Spec.Config.Raw from/to either of those).

As we discourage writing Ignition by hand anyway, I don't see a huge problem here. What specific troubleshooting steps do you see becoming much more difficult to do with this PR?

runcom:
RawExtension basically allows an arbitrary struct to be put there and we decide the destination struct to unmarshal to. Iirc this won't change how we write MC or how they're printed right?

kikisdeliveryservice:
> As we discourage writing Ignition by hand anyway, I don't see a huge problem here. What specific troubleshooting steps do you see becoming much more difficult to do with this PR?

My main question is if a MC has an Ign Config (which people do indeed write) that is encoded as []byte, if someone is inspecting MCs will this PR affect what they see in Spec.Config of the MC? Will it be readable? Are we changing the UI?

pkg/server/server_test.go
```diff
@@ -56,13 +57,29 @@ func TestMachineConfigToIgnition(t *testing.T) {
...
+	}
+	assert.Equal(t, 1, len(mcIgn.Storage.Files))
+	assert.Equal(t, mcIgn.Storage.Files[0].Path, "/etc/coreos/update.conf")
 
 	origMc := mc.DeepCopy()
-	ign := machineConfigToIgnition(mc)
-	assert.Equal(t, 1, len(origMc.Spec.Config.Storage.Files))
-	assert.Equal(t, 2, len(mc.Spec.Config.Storage.Files))
+	origIgnCfg, err := v1.DecodeIgnitionConfigSpecV2(origMc.Spec.Config.Raw)
+	if err != nil {
```
LorbusChris: Note to self: This is tests, use DecodeIgnitionConfigSpecV2OrDie func here

test/helpers/helpers.go
```diff
@@ -24,6 +25,17 @@ func NewMachineConfig(name string, labels map[string]string, osurl string, files
 	if labels == nil {
 		labels = map[string]string{}
 	}
+	rawIgnition, _ := mcfgv1.EncodeIgnitionConfigSpecV2(
```
LorbusChris: nts: Use OrDie func

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -1,14 +1,64 @@
 package v1
```
LorbusChris: move all helper funcs out of API pkg, or just those that need the ign config types?

Reasoning: We cannot migrate anything that depends on ignition config over to api repo

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -21,13 +71,28 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
 	var fips bool
-	outIgn := configs[0].Spec.Config
+
+	internalIgnPtr, err := DecodeIgnitionConfigSpecV2(configs[0].Spec.Config.Raw)
+	if err != nil {
+		panic(err)
```
LorbusChris: NTS: this is not a test func, so we shouldn't panic here.

pkg/server/server.go
```diff
@@ -123,7 +140,11 @@ func getNodeAnnotation(conf string) (string, error) {
 	return string(contents), nil
 }
 
-func appendFileToIgnition(conf *igntypes.Config, outPath, contents string) {
+func appendFileToIgnition(raw *runtime.RawExtension, outPath, contents string) {
+	conf, err := mcfgv1.DecodeIgnitionConfigSpecV2(raw.Raw)
+	if err != nil {
+		panic(err.Error())
```
LorbusChris: NTS: Don't panic

LorbusChris:
Rebase and fixup is in progress

LorbusChris:
@openshift/openshift-team-mco ptal

I believe it makes sense to merge this early on in the 4.5 window, and use this code as starting point for v2/v3 conversion and dual support.

/cc @yuqi-zhang @runcom

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -668,12 +673,16 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
+			if err != nil {
+				return fmt.Errorf("could not encode registries Ignition config: %v", err)
+			}
 			mc, err := ctrl.client.MachineconfigurationV1().MachineConfigs().Get(managedKey, metav1.GetOptions{})
 			if err != nil && !errors.IsNotFound(err) {
 				return fmt.Errorf("could not find MachineConfig: %v", err)
 			}
 			isNotFound := errors.IsNotFound(err)
-			if !isNotFound && equality.Semantic.DeepEqual(registriesIgn, mc.Spec.Config) {
+			if !isNotFound && equality.Semantic.DeepEqual(rawRegistriesIgn, mc.Spec.Config) {
```
yuqi-zhang: Are references to `mc.Spec.Config` the exact same as `mc.Spec.Config.Raw`? I'm a little confused since now we use `.Raw` for the object, but here we are comparing directly.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -350,15 +350,17 @@ func verifyRegistriesConfigAndPolicyJSONContents(t *testing.T, mc *mcfgv1.Machin
 		imgcfg.Spec.RegistrySources.BlockedRegistries, icsps)
 	require.NoError(t, err)
 	assert.Equal(t, mcName, mc.ObjectMeta.Name)
+
+	ignCfg := ctrlcommon.DecodeIgnitionConfigSpecV2OrDie(mc.Spec.Config.Raw)
```
yuqi-zhang: Maybe we should be testing MC objects here still and do some extra conversion instead of testing ignCfg directly

pkg/daemon/update.go
```diff
@@ -782,14 +808,14 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 // all the files, units that are present in the old config but not in the new one.
 // this function will error out if it fails to delete a file (with the exception
 // of simply warning if the error is ENOENT since that's the desired state).
-func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+func (dn *Daemon) deleteStaleData(oldConfig, newConfig *igntypes.Config) error {
```
yuqi-zhang: Minor nit, to keep it consistent I think we should refer "Config" as MachineConfig and "IgnConfig" as ignition config. For example here you changed oldConfig from a MC to a ign config, but didn't change the naming, which is slightly confusing

pkg/controller/common/helpers.go
```diff
@@ -1,25 +1,159 @@
...
 
-// NewIgnConfig returns an empty ignition config with version set as latest version
-func NewIgnConfig() igntypes.Config {
+const (
+	kernelTypeDefault  = "default"
+	kernelTypeRealtime = "realtime"
+)
+
+// DecodeIgnitionConfigSpecV2 decodes byte slices to Ignition Config Spec v2 or errors out
+func DecodeIgnitionConfigSpecV2(data []byte) (*igntypes.Config, error) {
```
yuqi-zhang: Actually, now I think about it, wouldn't we need some sort of function to first detect ignition version in a `data []byte` object? If we just have a `MachineConfig object` saved as `mc.Spec.Config.Raw` we wouldn't know what to decode it to.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
LorbusChris: this should indeed be comparing to `mc.Spec.Config.Raw`. I guess isn't caught because the RawExtension is something of an artificial type that doesn't add any extra bytes (i.e. wire-format stays exactly the same).

pkg/controller/common/helpers.go
LorbusChris: Yes we will need such a function! But since we only support V2 at this time, contents are always assumed to be V2 (or error out otherwise).

I thought it'd be overkill to add it already, but we'll definitely need it for dual-support. It could look something like this:
```
// GetSupportedIgnitionVersionFromRawConfig decodes byte slices to get the Ignition version contained or errors out
func GetSupportedIgnitionVersionFromRawConfig(data []byte) (string, error) {
	var cfgMap map[string]*json.RawMessage
	err := json.Unmarshal(data, &cfgMap)
	if err != nil {
		return "", err
	}
	type Ignition struct {
		Version string
	}
	var ign Ignition
	err = json.Unmarshal(*cfgMap["ignition"], &ign)
	if err != nil {
		return "", err
	}

	if ign.Version >= "2.0.0" && ign.Version <= igntypes.MaxVersion.String() {
		return "v2", nil
	}

	return "", fmt.Errorf("Unsupported Ignition Config Spec Version detected")
}
```

LorbusChris:
Thanks @yuqi-zhang! I'll address your review tomorrow :)

pkg/controller/common/helpers.go
runcom: > Yes we will need such a function! But since we only support V2 at this time, contents are always assumed to be V2 (or error out otherwise).

I think that's Kirsten WIP PR

kikisdeliveryservice: Agree with runcom. Scope of this PR should stay V2, we can merge then I can make whatever v3 changes are needed in my PR (also giving this PR some soak time).

pkg/daemon/update.go
```diff
@@ -381,8 +394,8 @@ func canonicalizeKernelType(kernelType string) string {
 
 // NewMachineConfigDiff compares two MachineConfig objects.
 func NewMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) *MachineConfigDiff {
-	oldIgn := oldConfig.Spec.Config
-	newIgn := newConfig.Spec.Config
+	oldIgn, _ := ctrlcommon.DecodeIgnitionConfigSpecV2(oldConfig.Spec.Config.Raw)
```
kikisdeliveryservice: is there a reason we aren't checking for the error here? (everywhere else there is a check when this func is called)

pkg/controller/common/helpers.go
```diff
@@ -1,25 +1,159 @@
...
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	errors "github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
-// NewIgnConfig returns an empty ignition config with version set as latest version
-func NewIgnConfig() igntypes.Config {
+const (
+	kernelTypeDefault  = "default"
```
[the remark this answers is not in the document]
LorbusChris: We have a non-singular number of constants files this could possibly go in -- push that into a follow up?

pkg/daemon/update.go
LorbusChris: I've changed up all functions that use this, including this one to also return an error now :)

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
LorbusChris: Hm looking closer at this, I'm not sure I follow here. The func still takes an MC as arg, and we need to look at the decoded config to be able to tell what's inside. Happy to change it up if you could elaborate a little more tho :)

LorbusChris:
@kikisdeliveryservice thank you for your review, too!
I've addressed the review in code and comments :)
cc @yuqi-zhang

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
yuqi-zhang: Yes, follow up is good, just wanted to make a note

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
yuqi-zhang: I think I confused something. This is good as is

LorbusChris:
@kikisdeliveryservice @yuqi-zhang this ready for another round of reviews :)

LorbusChris:
/hold
for some more improvements
Ignition has a Parse funct, which does validation and type casting...I will change up this PR to use that.

yuqi-zhang:
Also to note like you mentioned in https://github.com/openshift/machine-config-operator/pull/1486#discussion_r383372583 , this may have consequences for api validation.

LorbusChris:
@openshift/openshift-team-mco this is ready for another round of reviews (also my apologies I keep bugging you with this)

LorbusChris:
I can't really tell why MCO's e2e tests don't pass now. It seems the rendered config isn't picked up.
```
Error:      	Expected nil, but got: pool worker didn't report rendered-worker-ee4645cf261480abeac8ac0e54345e8f to updated (waited 20m0.009959737s): timed out waiting for the condition
```
Retrying once more...
/retest

yuqi-zhang:
I wonder if the merging of https://github.com/openshift/machine-config-operator/pull/1485 had anything to do with the failure. Maybe the 
```
              required:
              - ignition
``` part?
That said, the entire config section should be unvalidated.

**Comment by LorbusChris** [comment]
yeh I was wondering about that, too, as that PR seemed to have similar failures during development.

If an API call is rejected for not adhering to the OpenAPI validation, how would that show?

**Comment by yuqi-zhang** [comment]
So as far as I know there are multiple ways it can fail:

1. the field you provide is not in the defined fields. Since we have pruning turned on, it will simply not exist in the generated or rendered config. To check you can see whether the MC in a failed run has what you expect via CI artifacts

2. something you provide is in the defined fields but is a wrong field. This will manifest in either
a) the machine-config-controller logs
b) the machine-config-daemon logs (if it got that far)
c) gets bubbled up to machineconfigpools or the mco operator status

If none of those are reporting anything I am not sure where else to look.

**Comment by LorbusChris** [comment]
Ahh yea, the daemon logs show something
```
Marking Degraded due to: failed to retrieve file ownership for file "/etc/mytestconf0": failed to retrieve UserID for username: foo
```
I've added the `foo` user to the test files, because either username or ID have to be given, otherwise the config isn't valid. I'll change that to a user that actually exists..

**Comment by LorbusChris** [comment]
@yuqi-zhang thanks again for the pointer! Resolved by using an existing user for the test ign files.

All relevant tests seem to pass now -- ptal @openshift/openshift-team-mco

**Comment by LorbusChris** [comment]
A difference to earlier code in this PR is that now the ign.Parse function is used, which has built-in validation.

That is also the resaon why the explicit version mismatch check in the Reconcilable func is not needed anymore (see https://github.com/openshift/machine-config-operator/pull/996/commits/500cdea950f4211ac41ef7e123e623bd15ae404b#diff-06961b075f1753956d802ba954d2cfb5R444) -- the Parser will always try to translate configs passed to it to the latest known version, or fail if that isn't possible. A version mismatch will therefore never occur in the way that is checked by the Reconcilable func.

File: pkg/server/testdata/machine-configs/test-config.yaml [hunk 13]
```diff
@@ -9,6 +9,7 @@ spec:
     - nosmt
   config:
     ignition:
+      version: "2.2.0"
```
yuqi-zhang: Question: did this work before without an explicit ignition version?

LorbusChris: yes, this wasn't caught by tests before

LorbusChris:
I agree, having a commit that large is sub-optimal. I'm happy to take suggestion for a sensible split. The changes are pretty intertwined and the commit grew sort of organically during debugigng, so I didn't immediately see a way for splitting it up without adding a significant amount of hrs to this.

LorbusChris:
Not sure it makes a huge difference, but would you prefer if I split out the KernelType consts change, and the change from `test/e2e/utils_test.go` into seperate commits?

LorbusChris:
I'll split them out in case this PR needs another rebase :smile:

yuqi-zhang:
Sounds good. Will let others take a look and then lgtm

LorbusChris:
/cc @mtrmac 
I've included some of the suggestions from your review over at #1543 here, ptal :)

ashcrow:
@runcom PTAL

yuqi-zhang:
One more thing to note that is we may need to somewhat modify https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfig.crd.yaml#L67,
Although that said since we are not changing UX, I suppose its fine to still list is as an ignition (as opposed to rawExtension) object

pkg/server/server_test.go
```diff
@@ -49,23 +50,42 @@ func TestStringEncode(t *testing.T) {
...
 	assert.Nil(t, err)
 	mc := new(mcfgv1.MachineConfig)
 	err = yaml.Unmarshal([]byte(mcData), mc)
 	assert.Nil(t, err)
-	assert.Equal(t, 1, len(mc.Spec.Config.Storage.Files))
-	assert.Equal(t, mc.Spec.Config.Storage.Files[0].Path, "/etc/coreos/update.conf")
+	mcIgnCfg, _, err := ign.Parse(mc.Spec.Config.Raw)
+	if err != nil {
+		t.Errorf("decoding Ignition Config failed: %s", err)
+	}
```
mtrmac: (These `if err != nil` checks can be `{assert,require}.NoError(t, err)` . OTOH, existing style in parts of this file…)

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -2,67 +2,11 @@ package v1
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// MergeMachineConfigs combines multiple machineconfig objects into one object.
-// It sorts all the configs in increasing order of their name.
-// It uses the Ignition config from first object as base and appends all the rest.
-// Kernel arguments are concatenated.
-// It uses only the OSImageURL provided by the CVO and ignores any MC provided OSImageURL.
-func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineConfig {
```
mtrmac: Does the `pkg/apis` name mean this subpackage is meant to be consumed by other repositories? If so, is it safe to remove this?

(I can’t see any user inside GitHub/openshift , at least).

LorbusChris: The apis package will be moved into `openshift/api` eventually (hence moving out this func which needs ignition types). It's used by `installer` afaik

pkg/server/server_test.go
LorbusChris: yeh, I was trying to adhere to existing style :)

pkg/controller/render/render_controller_test.go
```diff
@@ -285,12 +287,21 @@ func TestIgnValidationGenerateRenderedMachineConfig(t *testing.T) {
 
 	// verify that an invalid igntion config (here a config with content and an empty version,
 	// will fail validation
-	mcs[1].Spec.Config.Ignition.Version = ""
+	ignCfg, _, err := ign.Parse(mcs[1].Spec.Config.Raw)
+	require.Nil(t, err)
```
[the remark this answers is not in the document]
LorbusChris: > One more thing to note that is we may need to somewhat modify https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfig.crd.yaml#L67,
> Although that said since we are not changing UX, I suppose its fine to still list is as an ignition (as opposed to rawExtension) object

AFAICT the only change for Spec3 here would be to remove the `filesystem` field inside dirs and files, but that would make it incompatible with Spec2. I therefore believe it should be safe to leave as-is.

LorbusChris:
> One more thing to note that is we may need to somewhat modify https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfig.crd.yaml#L67,
> Although that said since we are not changing UX, I suppose its fine to still list is as an ignition (as opposed to rawExtension) object

@yuqi-zhang AFAICT the only change for Spec3 here would be to remove the `filesystem` field inside dirs and files, but that would make it incompatible with Spec2. I therefore believe it should be safe to leave as-is.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -546,12 +546,21 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 			}
 			if isNotFound {
 				tempIgnCfg := ctrlcommon.NewIgnConfig()
-				mc = mtmpl.MachineConfigFromIgnConfig(role, managedKey, &tempIgnCfg)
+				mc, err = mtmpl.MachineConfigFromIgnConfig(role, managedKey, tempIgnCfg)
+				if err != nil {
+					glog.V(2).Infoln(tempIgnCfg, err, "could not create MachineConfig from new Ignition config: %v", err)
```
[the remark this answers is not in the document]
LorbusChris: /cc @runcom for final review

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -546,12 +546,21 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
 			}
-			mc.Spec.Config = createNewIgnition(map[string][]byte{
+
+			internalCtrRuntimeConfigIgn := createNewIgnition(map[string][]byte{
 				storageConfigPath: storageTOML,
 				crioConfigPath:    crioTOML,
 			})
+			rawCtrRuntimeConfigIgn, err := json.Marshal(internalCtrRuntimeConfigIgn)
+			if err != nil {
+				glog.V(2).Infoln(internalCtrRuntimeConfigIgn, err, "error marshalling container runtime config Ignition: %v", err)
```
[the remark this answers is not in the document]
LorbusChris: /cc @runcom for final review

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -683,9 +696,12 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 			}
 			if isNotFound {
 				tempIgnCfg := ctrlcommon.NewIgnConfig()
-				mc = mtmpl.MachineConfigFromIgnConfig(role, managedKey, &tempIgnCfg)
+				mc, err = mtmpl.MachineConfigFromIgnConfig(role, managedKey, tempIgnCfg)
+				if err != nil {
+					glog.V(2).Infoln(tempIgnCfg, err, "could not create MachineConfig from new Ignition config: %v", err)
```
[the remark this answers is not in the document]
LorbusChris: /cc @runcom for final review

pkg/controller/render/render_controller_test.go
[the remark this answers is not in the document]
LorbusChris: keeping the current style here

LorbusChris:
@runcom, @mtrmac has raised some concerns, please give this a final review, and lift the hold if this is good to go as-is

ashcrow:
@runcom PTAL

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
runcom: This is going to create an MC which could potentially be nil which isn't that perfect - we should return an error here as well (and syncStatusOnly as above)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
runcom: yes, this has to return an error (how unlikely that is doesn't matter, we can't just crash and this has a path to crashing besides how unlikely that is). This and the other are errors where we want to instantly communicate to users and just bail here.
```
return ctrl.syncStatusOnly(cfg, err, "could not create MachineConfig from new Ignition config: %v", err)
```

LorbusChris:
ptal @runcom

LorbusChris:
rebased

vrutkovs:
Not sure about e2e-gcp-op, but the rest looks like flakes

/retest

LorbusChris:
tests from `test/e2e` seem to time out regularly, not just on this PR

yuqi-zhang:
I don't think the kernelTypes test is a flake actually. I did a manual run of the test and one node got stuck in `NotReady, SchedulingDisabled` . Azure console shows the node as "Running" but I have no access to it. Worker logs are not shown on Azure so maybe you can take a look running this manually on AWS and seeing what happened on reboot.

I also did a test upgrading 4.5 nightly to this and it passed, so that's nice to know

yuqi-zhang:
Actually scratch that the test is failing on other PRs too, so its not due to this PR. But CI is currently blocked on that

yuqi-zhang:
@LorbusChris https://github.com/openshift/machine-config-operator/pull/1612 should fix the test, could you rebase on that?

LorbusChris:
:tada:. rebased.

yuqi-zhang:
/lgtm

This passes my local verifications so once we pass the flaking test I am relatively confident in it.

LorbusChris:
I'm taking the hold off that I applied for more review earlier.
(feel free to apply again)
/hold cancel

kikisdeliveryservice:
@runcom PTAL

kikisdeliveryservice:
would like another review on this as we only have jerry's and @runcom was reviewing earlier.

/hold

Issue #997, opened 2019-07-22
MCC reporting unsupported platform when installing on vSphere


miabbott:
**Description**

During a UPI install of Openshift 4.1 on vSphere, the MCC is reporting vSphere as an unsupported platform.

**Steps to reproduce the issue:**
1.  Perform UPI install of OpenShift 4.1 on vSphere
2.  Inspect MCC logs on bootstrap node
3.

**Describe the results you received:**

```
[root@bootstrap-0 ~]# crictl logs -f 2ad00effa3946
I0721 17:30:31.823340       1 bootstrap.go:40] Version: 4.1.6-201907101224-dirty (365c1cfd14de5b0e3b85e0fc815b0060f36ab955)
W0721 17:30:31.831657       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
W0721 17:30:31.843369       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
W0721 17:30:31.845925       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
W0721 17:30:31.847011       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
W0721 17:30:31.849637       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
W0721 17:30:31.852473       1 render.go:127] Warning: the controller config referenced an unsupported platform: vsphere
```

**Describe the results you expected:**

No warnings about vSphere as an unsupported platform

(filed on behalf of an email sent to aos-devel@redhat.com)

ericavonb:
I'm unsure what the definition of "supported" is for the MCO, but assuming we _do_ support vsphere I put out a PR fixing this issue: https://github.com/openshift/machine-config-operator/pull/998

kikisdeliveryservice:
Is vsphere fully supported? I've asked the person who added it to mco in the PR.. Also can this be made into a BZ since it will not get merged if correct otherwise @miabbott

kikisdeliveryservice:
Some info here: https://github.com/openshift/machine-config-operator/pull/998#discussion_r305925408

This might be desired behavior, waiting for more info...

miabbott:
@kikisdeliveryservice the original email points to our UPI docs for vSphere - https://docs.openshift.com/container-platform/4.1/installing/installing_vsphere/installing-vsphere.html#installation-vsphere-config-yaml_installing-vsphere

[...]

PR #998, opened 2019-07-22
Bug 1736254: template/render: support vsphere platform

ericavonb:
Do not warn when platform is VSphere as this is supported for UPI
installs.

Closes: #997

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -133,7 +133,7 @@ func platformFromControllerConfigSpec(ic *mcfgv1.ControllerConfigSpec) (string, 
 		return "", fmt.Errorf("cannot generate MachineConfigs when no platform is set")
 	case platformBase:
 		return "", fmt.Errorf("platform _base unsupported")
-	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenStack, platformLibvirt, platformOvirt, platformNone:
+	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenStack, platformLibvirt, platformOvirt, platformVSphere, platformNone:
 		return ic.Platform, nil
 	default:
 		// platformNone is used for a non-empty, but currently unsupported platform.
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
Would like @abhinavdahiya to weigh in on this since vsphere wasn't explicitly added to that list. PTAL when you get a minute?

pkg/controller/template/render.go
```diff
@@ -128,7 +128,7 @@ func platformFromControllerConfigSpec(ic *mcfgv1.ControllerConfigSpec) (string,
 		return "", fmt.Errorf("cannot generate MachineConfigs when no platform is set")
 	case platformBase:
 		return "", fmt.Errorf("platform _base unsupported")
-	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenstack, platformLibvirt, platformNone:
+	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenstack, platformLibvirt, platformNone, platformVSphere:
```
abhinavdahiya: cc @eparis @staebler is the cloud provider ready to be turned on the kubelet.
in previous conversation, setting vsphere for kubelet caused failures.

staebler: The last that I saw the kubelet would crash loop with the cloud provider set. I believe that the contention from @eparis was that the crash loop was at least partially due to inappropriate permissions when accessing the creds secret. I do not know of any work to resolve that.

ericavonb: Is there an issue that I can link to in a comment and in the issue?

kikisdeliveryservice:
Just going to add a hold until this gets sorted out

/hold

pkg/controller/template/render.go
runcom: @eparis @abhinavdahiya @staebler would be nice to understand if this is an issue with 4.2 or it's _still_ the expected behavior. The BZ has been closed based on @staebler's comment above but happy to bring it back and merge this if this is 4.2

kikisdeliveryservice:
@rphillips PTAL

kikisdeliveryservice:
Could we get this rebased @ericavonb

ericavonb:
> Could we get this rebased @ericavonb

Done! I had pushed the rebase to my fork but for some reason this pr was against an upstream branch.

rphillips:
/retitle Bug 1736254: template/render: support vsphere platform

PR #999, opened 2019-07-22
Bug 1727022: set upgradeable operator condition

ericavonb:
Set the condition "upgradeable" on the "machine-config"  cluster operator.
https://github.com/openshift/api/blob/9525304a0adb725ab4a4a54539a1a6bf6cc343d3/config/v1/types_cluster_operator.go#L140

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1727022

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -177,6 +177,27 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
 	return optr.updateStatus(co, coStatus)
 }
 
+// syncUpgradeableStatus applies the new condition to the mco's ClusterOperator object.
+func (optr *Operator) syncUpgradeableStatus() error {
+	co, err := optr.fetchClusterOperator()
+	if err != nil {
+		return err
+	}
+	if co == nil {
+		return nil
+	}
+	// Report default "Upgradeable=True" status. When known hazardous states for upgrades are
+	// determined, specific "Upgradeable=False" status can be added with messages for how admins
+	// can resolve it.
+	// [ref] https://github.com/openshift/cluster-version-operator/blob/8402d219f36fc79e03edf45918785376113f2cc1/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource
+	coStatus := configv1.ClusterOperatorStatusCondition{
+		Type:   configv1.OperatorUpgradeable,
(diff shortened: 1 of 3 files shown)


pkg/operator/status.go
```diff
@@ -207,6 +207,7 @@ func (optr *Operator) initializeClusterOperator() (*configv1.ClusterOperator, er
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse})
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse})
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorDegraded, Status: configv1.ConditionFalse})
+	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorUpgradeable, Status: configv1.ConditionTrue})
```
ericavonb: WIP - this is NOT where I plan on setting this condition in the final version

pkg/operator/status.go
```diff
@@ -125,10 +125,6 @@ func (optr *Operator) syncProgressingStatus() error {
 }
 
 func (optr *Operator) updateStatus(co *configv1.ClusterOperator, status configv1.ClusterOperatorStatusCondition) error {
-	existingCondition := cov1helpers.FindStatusCondition(co.Status.Conditions, status.Type)
-	if existingCondition.Status != status.Status {
-		status.LastTransitionTime = metav1.Now()
-	}
```
runcom: any reason to have dropped this?

ericavonb: It was causing a nil-pointer dereference panic when the existing condition wasn't found for the new type, but the logic is already handled in the call to `cov1helpers.SetStatusCondition` anyways so I just removed it.

https://github.com/openshift/library-go/blob/3679c350479175dceb60cbdafe68f0d92280952f/pkg/config/clusteroperator/v1helpers/status.go#L31

runcom: 👍👍

PR #1002, opened 2019-07-23
Render baremetal bootstrap configs

cybertron:
**- What I did**
This moves the baremetal infra bootstrap static pod configurations to MCO, which is necessary so the installer can look up the images from the release and pass them in. Previously the pod definitions were in the installer itself, which meant they were written before bootkube.sh was run and could not use the standard method to retrieve the images. This also moves all of the baremetal infra templates into the same repo and converts them to use the MCO templating just like the master versions.

**- How to verify it**
Remove the existing baremetal systemd services from the installer and deploy with this series of commits. Coredns and keepalived should be deployed on the bootstrap node as static pods.

Baremetal infra static pod definitions are now deployed during the machine-config-operator bootstrap phase.


pkg/operator/bootstrap.go
```diff
@@ -145,6 +150,18 @@ func RenderBootstrap(
 	}, {
 		name:     "manifests/machineconfigserver/kube-apiserver-serving-ca-configmap.yaml",
 		filename: "manifests/kube-apiserver-serving-ca-configmap.yaml",
+	}, {
+		name:     "manifests/baremetal/coredns.yaml",
```
iamemilio: Is there some way for this to take in platform as a variable:
`/manifests/platform/*`

iamemilio: maybe:
`/manifests/infra.Status.APIServerInternalURL/*`

iamemilio: `/manifests/spec.Platform/*`

templates/master/00-master/baremetal/files/NetworkManager-dns-vip-prepender.yaml
```diff
@@ -0,0 +1,41 @@
+filesystem: "root"
```
celebdor: SHouldn't all these be out of this PR and just in #795 Is this just to make it work and we'll rebase as soon as 795 is merged?

cybertron:
This is rebased now that #795 has merged. It also includes a couple of fixes for things we identified in testing.

Also note that this is a dependency for https://github.com/openshift/installer/issues/2067

pkg/operator/bootstrap.go
```diff
@@ -169,6 +169,26 @@ func RenderBootstrap(
...
+			filename: "baremetal/manifests/coredns.yaml",
+		}, {
+			name:     "manifests/baremetal/coredns-corefile.tmpl",
+			filename: "baremetal/static-pod-resources/coredns/Corefile.tmpl",
+		}, {
+			name:     "manifests/baremetal/keepalived.yaml",
+			filename: "baremetal/manifests/keepalived.yaml",
+		}, {
+			name:     "manifests/baremetal/keepalived.conf.tmpl",
+			filename: "baremetal/static-pod-resources/keepalived/keepalived.conf.tmpl",
```
runcom: shouldn't all these end up under `bootstrap/` similar to https://github.com/openshift/machine-config-operator/pull/1002/files#diff-d6f38878757202a810cfa8038e7ae122R161

Have you tested this out manually to verify this is applied at bootstrap?

runcom:
cc @abhinavdahiya - can you take a look at the overall mechanism here? I feel like this is sane to do but maybe you have more insight from the installer and this can be achieved from the installer as well (key thing here is that we want the images from the payload to be available to render these manifests and previously we only had systemd units in the installer bootstrap data).

pkg/operator/bootstrap.go
cybertron: I believe @celebdor made this change because having them in bootstrap caused them to be applied twice - once by kubelet and again during cluster-bootstrap. We will be explicitly copying them into place in the installer[0], so the specific location here doesn't matter much.

0: https://github.com/celebdor/installer/commit/134d74cdab091303721b00d2111330c862420b23

cybertron:
The other reason we wanted to move these here was so we had a consistent templating mechanism between bootstrap and master/worker nodes. When these resources lived in the installer we had to retrieve the inputs in more hacky ways. Here we can use the exact same platform status data as the master templates.

cgwalters:
/approve
I didn't review this in detail, but it looks good to me offhand. I think probably the best strategy is to get this and the installer PR merged and let people do testing from master builds.

runcom:
/lgtm

deferring my question to @abhinavdahiya from https://github.com/openshift/machine-config-operator/pull/1002#issuecomment-515514982

PR #1004, opened 2019-07-24
Mcd encap prep cleanups 1

cgwalters:
Extracted from https://github.com/openshift/machine-config-operator/pull/935

With one other small commit on top. 

Would like to land this so I'm less in rebase hell later...

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -996,7 +996,7 @@ func (dn *Daemon) runOnceFromIgnition(ignConfig igntypes.Config) error {
 	if err := dn.writeUnits(ignConfig.Systemd.Units); err != nil {
 		return err
 	}
-	return dn.reboot("runOnceFromIgnition complete", defaultRebootTimeout, rebootCommand("runOnceFromIgnition complete"))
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -833,7 +829,7 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 	// and if we still have a pendingConfig it means we've been killed by kube after 600s
 	// take a stab at that and re-run the drain+reboot routine
 	if state.pendingConfig != nil && bootID == dn.bootID {
-		dn.logSystem("killed by kube, retrying...")
+		dn.logSystem("drain interrupted, retrying")
```
runcom: thanks for this, I completely forgot to clean this up

runcom:
/approve

I'm not sure we can land this even if it's just a cleanup and it's far better than before - I'll check that out and/or create a `master-4.3` branch otherwise where we can land these kind of PRs from now on

cgwalters:
It's just feature freeze, not code freeze right?

runcom:
> It's just feature freeze, not code freeze right?

right :thinking: this isn't changing anything then or adding anything

/lgtm

PR #1007, opened 2019-07-24
test/e2e: add custom pool test

runcom:

**- What I did**

There has been a lot of interest around custom pools lately and it does make sense to cover the flow with an e2e to avoid regressions.

**- How to verify it**

CI

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -49,10 +49,14 @@ func TestMCDToken(t *testing.T) {
 	}
 }
 
-func mcLabelForWorkers() map[string]string {
+func mcLabelForRole(role string) map[string]string {
 	mcLabels := make(map[string]string)
-	mcLabels["machineconfiguration.openshift.io/role"] = "worker"
+	mcLabels["machineconfiguration.openshift.io/role"] = role
 	return mcLabels
+}
+
+func mcLabelForWorkers() map[string]string {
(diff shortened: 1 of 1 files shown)


runcom:
aws api throttling....

test/e2e/mcd_test.go
```diff
@@ -592,3 +599,59 @@ func TestFIPS(t *testing.T) {
...
+		metav1.LabelSelectorRequirement{
+			Key: "machineconfiguration.openshift.io/role",
+			Operator: metav1.LabelSelectorOpIn,
+			Values: []string{"worker", "infra"},
+		},
+	}
+	_, err = cs.MachineConfigPools().Create(infraMCP)
+	require.Nil(t, err)
+
+	infraMC := createMCToAddFileForRole("modify-host-file", "infra", "/etc/motd", "mco-test", "root")
```
cgwalters: Since a bunch of other tests also write this, I think we could get false positives.  How about:
```
 createMCToAddFileForRole("modify-host-file", "infra", "/etc/mco-custom-pool", "custom-pool", "root")
```
or so?

(And we should in general have each test create a uniquely named file)

runcom: oh yeah, makes total sense

runcom: fixed, thanks!

runcom:
ok, fixed the wrong `cat` command

PR #1011, opened 2019-07-25
pkg/controller: allow custom pool rollback

runcom:
In case a user needs to rollback the custom pool he can do so by
removing the custom pool label from the node. However, w/o this patch
the MCC wasn't reacting to label changes. It should be safe to only
react to label changes also since we only care about the role label and
the labels aren't updated that often.

`oc label node ip-10-0-143-53.ec2.internal node-role.kubernetes.io/infra-`

The command above is going to drop the infra label from a worker,infra node causing it to rollback to a worker rendered MC.

This also expands the e2e added in #1007 

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -654,4 +654,22 @@ func TestCustomPool(t *testing.T) {
 		}
 		t.Logf("Node %s has expected infra MC content", node.Name)
 	}
+
+	out, err = exec.Command("oc", "label", "node", infraNode.Name, "node-role.kubernetes.io/infra-").CombinedOutput()
+	require.Nil(t, err, "unable to remove infra label from node %s: %s", infraNode.Name, string(out))
+
+	workerMCP, err := cs.MachineConfigPools().Get("worker", metav1.GetOptions{})
+	require.Nil(t, err)
+	if err := wait.Poll(2*time.Second, 5*time.Minute, func() (bool, error) {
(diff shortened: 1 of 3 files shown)


runcom:
this is rebased and good to go now

/hold cancel

test/e2e/mcd_test.go
```diff
@@ -654,4 +654,22 @@ func TestCustomPool(t *testing.T) {
...
+
+	out, err = exec.Command("oc", "label", "node", infraNode.Name, "node-role.kubernetes.io/infra-").CombinedOutput()
+	require.Nil(t, err, "unable to remove infra label from node %s: %s", infraNode.Name, string(out))
+
+	workerMCP, err := cs.MachineConfigPools().Get("worker", metav1.GetOptions{})
+	require.Nil(t, err)
+	if err := wait.Poll(2*time.Second, 5*time.Minute, func() (bool, error) {
+		node, err := cs.Nodes().Get(infraNode.Name, metav1.GetOptions{})
+		require.Nil(t, err)
+		if node.Annotations[constants.DesiredMachineConfigAnnotationKey] == workerMCP.Spec.Configuration.Name {
```
runcom: this should be `!=` - we need to wait till the infra,worker node becomes worker-only targeted so this was passing just because we were lucky to be still at a rendered-infra-<hash> desiredConfig.
Fixed

runcom:
This went through, bumped the e2e timeout to account for the new pool test

cgwalters:
:swimming_man: :hourglass_flowing_sand: 
/lgtm

PR #1012, opened 2019-07-25
hack: allow using Docker for development

runcom:
This patch mainly modifies scripts under ./hack to allow people (me) to
develop using docker instead of podman (various reasons behind that, osx
mainly but still, many people are using docker). What this allows is:

podman=docker ./hack/cluster-push.sh

The normal flow with podman shouldn't have changed (please test it out!).

Also dropped a wrong section from HACKING.md and a minor change there as well.

### DIFF
--- a/hack/cluster-push.sh
+++ b/hack/cluster-push.sh
@@ -12,8 +12,6 @@
 
 set -xeuo pipefail
 
-podman=${podman:-podman}
-
 do_build=1
 if [ "${1:-}" = "-n" ]; then
     do_build=0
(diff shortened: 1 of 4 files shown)


runcom:
cc @cgwalters

hack/cluster-push.sh
```diff
@@ -28,9 +26,18 @@ REMOTE_IMGNAME=openshift-machine-config-operator/${imgname}
...
     ./hack/build-image
 fi
-$podman push --tls-verify=false "${LOCAL_IMGNAME}" ${registry}/${REMOTE_IMGNAME}
+builder_secretid=$(oc get -n openshift-machine-config-operator secret | egrep '^builder-token-'| head -1 | cut -f 1 -d ' ')
+secret="$(oc get -n openshift-machine-config-operator -o json secret/${builder_secretid} | jq -r '.data.token' | base64 -d)"
+
+set +e
+skopeo copy --dest-tls-verify=false --dest-creds unused:${secret} "containers-storage:${LOCAL_IMGNAME}" "docker://${registry}/${REMOTE_IMGNAME}"
+if [ "$?" -ne 0 ]; then
+  # awful hack to allow people to use docker as well
```
cgwalters: Hm, maybe simpler to do something like:
```
if [[ "${podman}" =~ "docker" ]]; then
  imgstorage="docker-daemon:"
else
  imgstorage="containers-storage:"
fi
skopeo copy --dest-tls-verify=false --dest-creds unused:${secret} "${imgstorage}:${LOCAL_IMGNAME}" "docker://${registry}/${REMOTE_IMGNAME}"
```

cgwalters:
Side note, I actually live inside a `podman` container using [coretoolbox](https://github.com/cgwalters/coretoolbox) but the problem there is recursive container builds...don't work so I end up with a `podman-host-sudo` alias that runs podman on the host which mostly works, except that doesn't intercept the `skopeo` calls which looks at a different credential chain.  So far I've just been copying them creds manually but at some point I'd like to figure out how to make the podman-in-podman nicer.

This PR doesn't change things here AFAICS, just noting it.

hack/cluster-push.sh
runcom: fixed!

PR #1013, opened 2019-07-25
[WIP] pkg/daemon: recover config from sys state

ericavonb:
TODO write PR description

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -39,7 +39,20 @@ const (
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsCommand is the command to use when enabling or disabling FIPS
 	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
+	// fipsEnabledFile is the filepath for checking if FIPS mode is enabled on the system
+	fipsEnabledFile = "/proc/sys/crypto/fips_enabled"
 )
+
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -39,8 +39,21 @@ const (
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsCommand is the command to use when enabling or disabling FIPS
 	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
+	// fipsEnabledFile is the filepath for checking if FIPS mode is enabled on the system
+	fipsEnabledFile = "/proc/sys/crypto/fips_enabled"
```
cgwalters: This works, but I think it'd be better if we instead executed `/usr/libexec/rhcos-tools/coreos-fips check` or so.  (Which...doesn't exist but should!)

The reason here is it'd be good to centralize logic around FIPS (for RHCOS) in that script, which definitely includes things beyond just the kernel FIPS flag.

Maybe something as simple as:
`/usr/libexec/rhcos-tools/coreos-fips status` outputs JSON or something that the MCD parses.

Or taking this out one level...bigger picture I think it's useful to have "Unix CLI friendly" tools in the OS, and the MCD for the most part just *orchestrates* those tools.

kikisdeliveryservice:

Feel free to reopen.

PR #1014, opened 2019-07-25
Generate registries.conf using ICSP in MCC bootstrap

mtrmac:
**- What I did**

Taught the MCC bootstrap to consume ICSP and use it to generate correct `registries.conf`. This is apparently necessary during OpenShift installation so that masters generated by the bootstrap node are configured correctly and can pull other images necessary to proceed.

This requires the installer to provide ICSP data in the manifest directory consumed by the MCC bootstrap (but it can cleanly work with earlier installer versions that don’t provide such data, it will simply assume that there are no ICSP objects).

See individual commit messages for details.

**- How to verify it**

<s>WIP; this does not yet have any tests, I didn’t even try running it</s>
There are unit tests, and @sjenning has verified that the code can correctly boot a cluster which requires the mirror configuration to work.

MCC bootstrap now consumes ImageContentSourcePolicy and can generate `registries.conf` mirror configuration appropriately.


pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -710,6 +706,63 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
+		tempIgnCfg := ctrlcommon.NewIgnConfig()
+		mc := mtmpl.MachineConfigFromIgnConfig(role, managedKey, &tempIgnCfg)
+		mc.Spec.Config = *registriesIgn
+		// Explicitly do NOT set GeneratedByControllerVersionAnnotationKey so that the first run of the non-bootstrap controller
+		// always rebuilds registries.conf (with the insecureRegs/blockedRegs values actually available).
+		mc.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+			{
+				APIVersion: apicfgv1.SchemeGroupVersion.String(),
+				Kind:       "Image",
+				// Name and UID is not set, the first run of syncImageConfig will overwrite these values.
```
mtrmac: I have no idea whether this is safe.

mtrmac:
> This requires the installer to provide ICSP data in the manifest directory consumed by the MCC bootstrap

https://github.com/openshift/installer/pull/2097 .

mrunalp:
@runcom ptal.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
runcom: what I'm worried about here is that we'll end up with a rendered machineconfig which is different than the one produced by the in-cluster sync-loop. This only means 1 thing, this is not going to work cause there will be a drift if we can't generate exactly the same thing we're generating at bootstrap

runcom: we may want the ImageConfig to be available at bootstrap as well so we can make sure this is going to generate the same stuff - will think more about this.

mtrmac: This code intentionally does not set `GeneratedByControllerVersionAnnotationKey` to anything. That should guarantee that when the MCC starts, `syncImageConfig` will _always_ update the object on first start.

Or are you saying that such an update would break things?

(I’m also assuming that future upgrades of the cluster don’t use the bootstrap process, so the bootstrap code does not need to worry about any pre-existing `getManagedKeyReg(pool)`-named objects.)

runcom: > Or are you saying that such an update would break things?

the state today is:

- the bootstrap process generates a given `rendered-<pool>-<hash>` MC
- the sync loop generates a given `rendered-<pool>-<hash>` MC
- both have to match or it's not going to work

I'm not fully sure the above is the case (too tired now lol). But I suspect that if this code doesn't produce the same MachineConfig (cause it's missing ImageConfig) than what syncImageConfig would produce, then we'll have a drift and it's not gonna work. Will check again better later today.

mtrmac: WRT making `ImageConfig` available, I agree that it would be _clean_ to generate exactly the same configuration on the bootstrap path, but it’s not just `ImageConfig`: `getValidRegistries` requires `ClusterVersion`.**`Status.Desired.Image`**, and looking at
https://github.com/openshift/installer/blob/580b58fa7ef8c8904f65392830aaf18ede2b3ddd/data/data/manifests/bootkube/cvo-overrides.yaml.template and https://github.com/openshift/cluster-version-operator/blob/398afc117e3d418664572cf2b9554ce9fcd2bcc8/bootstrap/bootstrap-pod.yaml , this is not set by the installer, but only by a bootstrap pod — that runs at about the time of the MCC bootstrap, but it’s not available at the MCO bootstrap time, and presumably (?) the MCC bootstrap can’t use the Kubernetes API to read (and block on) the `ClusterVersion.Status` to be available.

mtrmac: > * the bootstrap process generates a given `rendered-<pool>-<hash>` MC
> * the sync loop generates a given `rendered-<pool>-<hash>` MC
> * both have to match or it's not going to work

That happens in the render controller, and AFAICS that should not be affected; this only generates one more MC for the render controller to merge.

Yes, it could be possible that the `registries.conf` input to the render controller is the same during bootstrap and after bootstrap — in that case `syncImageConfig` would create an updated MC, but that would not trigger the render controller to create a new `rendered-*` MC. That seems correct to me in principle.

(Currently, the first post-bootstrap `syncImageConfig` always triggers the render controller to re-render because `syncImageConfig` creates a `registries.conf` MC, even if that MC has no `Spec.Config.Storage.Files` (the `registriesTOMLConfig == nil` case); and because `MergeMachineConfigs` does not exclude no-files empty configs, this causes the hash to change. So, if anything in the render code differed between the bootstrap path and the controller, and silently expected the first post-bootstrap run to trigger a re-render, yes, that could be a problem. Comparing `syncGeneratedMachineConfig` and `RunBootstrap`, it does seem that the bootstrap code does not set `pool.Spec.Configuration.Source`, and does set `pool.Status.Configuration.Name`, so there might be that kind of a difference. That could be problematic for `operator.isMachineConfigPoolConfigurationValid` and `syncRequiredMachineConfigPools`, but that’s too deep a rabbit hole for me right now.)

mtrmac: … I mean, if you are worried only about the hash, we can always add a `# Dummy comment to ensure the registries.conf is different on bootstrap` line to the bootstrap one. But that is a badly targeted hack.

Another way to argue this is that the render controller exists longer than `syncImageConfig`, so _if_ the render controller worked correctly before `syncImageConfig` was added, then the extra `registries.conf` MC that currently triggers a re-render was never required.  OTOH the render controller is not _that_ much older than `syncImageConfig`, so I can’t quickly tell for sure.

mtrmac: > This code intentionally does not set `GeneratedByControllerVersionAnnotationKey` to anything. That should guarantee that when the MCC starts, `syncImageConfig` will _always_ update the object on first start.

Oops, I didn’t notice that the MCC bootstrap only saves the results of rendering, not the intermediate MCs; those exist only in memory. Well, one more reason for `syncImageConfig` to always be used to produce the copy of record.

(This does not affect the discussion about the rendered MC being correct/consistent AFAICS.)

umohnani8:
Changes LGTM. @runcom PTAL

pkg/controller/container-runtime-config/container_runtime_config_controller.go
runcom: has this been tested somehow on a cluster? I want to avoid merging something which has not been manually tested at least if there's a way to test this out

runcom:
> Changes LGTM. @runcom PTAL

https://github.com/openshift/machine-config-operator/pull/1014#discussion_r307811750

pkg/controller/container-runtime-config/container_runtime_config_controller.go
sjenning: @runcom i'm going to try to test it this afternoon if i can figure out how to build the images, get them into a release image, and mirror it

sjenning:
This does not seem to be having the desired effect.  MCs served by the bootstrap MCS are still not including the mirror cfg in `registries.conf`

I can see the installer change on the bootstrap node
```
# ls /etc/mcc/bootstrap/image-content-source-policy-*
/etc/mcc/bootstrap/image-content-source-policy-0.yaml  /etc/mcc/bootstrap/image-content-source-policy-1.yaml

# cat /etc/mcc/bootstrap/image-content-source-policy-*
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  creationTimestamp: null
  name: image-policy-0
spec:
  repositoryDigestMirrors:
  - mirrors:
    - registry.lab.variantweb.net/ocp
    source: registry.svc.ci.openshift.org/ocp/4.2-2019-07-26-152829
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  creationTimestamp: null
  name: image-policy-1
spec:
  repositoryDigestMirrors:
  - mirrors:
    - registry.lab.variantweb.net/ocp
    source: registry.svc.ci.openshift.org/ocp/release
```
But the rendered MCs don't have the mirror cfg
```
# cat rendered-master-0efaedd7047cb6bf1ef99d5bbe325ad9.yaml | grep -B5 "registries.conf"
      - contents:
          source: data:,unqualified-search-registries%20%3D%20%5B'registry.access.redhat.com'%2C%20'docker.io'%5D%0A
          verification: {}
        filesystem: root
        mode: 420
        path: /etc/containers/registries.conf
```

sjenning:
The MCS init container for the MCS static pod on the bootstrap node is very terse in its logging :-/
```
# crictl ps -a
CONTAINER ID        IMAGE                                                              CREATED             STATE               NAME                        ATTEMPT             POD ID
45b630c7cb953       ba6ab05fa648515160acd961d65db18866f4031cec9cd50995af10297782c891   11 minutes ago      Running             machine-config-server       0                   ff589d18b78de
cc63ee3337914       ba6ab05fa648515160acd961d65db18866f4031cec9cd50995af10297782c891   11 minutes ago      Exited              machine-config-controller   0                   ff589d18b78de
# crictl logs cc63ee3337914
I0726 20:35:46.358045       1 bootstrap.go:40] Version: machine-config-daemon-4.2.0-201907161330-74-gfcd80d34-dirty (fcd80d34dcd9dbef012d85e58972bde510cb343e)
```

mtrmac:
Note to self: it seems that files in `/etc/mcc/bootstrap` are not read, only in the subdirectories.

mtrmac:
Hopefully fixed with https://github.com/openshift/machine-config-operator/pull/1014/commits/712d1a223bde3439071dc4956b1003a23a369c6d .  (Is there a better way to decode the object from two different schemas?).  I still need to write tests for this.

Note that the generated `MachineConfig` contains two entries for `registries.conf`, the latter one is correct. AFAICT that’s how the render controller works, there is no code to merge configuration by file name.

sjenning:
@mtrmac testing...

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -79,7 +83,10 @@ func (b *Bootstrap) Run(destDir string) error {
 		}
 
 		for idx, m := range manifests {
-			obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(v1.SchemeGroupVersion), m.Raw)
+			obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(v1.GroupVersion), m.Raw)
```
abhinavdahiya: This is how installer uses a decoder for multiple schemes https://github.com/openshift/installer/blob/6e2977c740853e842247b55c7cd08d9b350f3e93/pkg/asset/machines/master.go#L287-L301

I think we should do the same rather than handling `runtime.IsNotRegisteredError(err)`

sjenning:
My mirrored cluster install succeeded with the current changes in this PR
```
$ oc get imagecontentsourcepolicy -oyaml
apiVersion: v1
items:
- apiVersion: operator.openshift.io/v1alpha1
  kind: ImageContentSourcePolicy
  metadata:
    creationTimestamp: "2019-07-29T16:19:09Z"
    generation: 1
    name: image-policy-0
    resourceVersion: "276"
    selfLink: /apis/operator.openshift.io/v1alpha1/imagecontentsourcepolicies/image-policy-0
    uid: 981879d1-b21c-11e9-be72-fa163e86f289
  spec:
    repositoryDigestMirrors:
    - mirrors:
      - registry.lab.variantweb.net/ocp
      source: registry.svc.ci.openshift.org/ocp/4.2-2019-07-29-133200
- apiVersion: operator.openshift.io/v1alpha1
  kind: ImageContentSourcePolicy
  metadata:
    creationTimestamp: "2019-07-29T16:19:09Z"
    generation: 1
    name: image-policy-1
    resourceVersion: "277"
    selfLink: /apis/operator.openshift.io/v1alpha1/imagecontentsourcepolicies/image-policy-1
    uid: 98359565-b21c-11e9-be72-fa163e86f289
  spec:
    repositoryDigestMirrors:
    - mirrors:
      - registry.lab.variantweb.net/ocp
      source: registry.svc.ci.openshift.org/ocp/release
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""

$ oc get clusteroperators 
NAME                                       VERSION                        AVAILABLE   PROGRESSING   DEGRADED   SINCE
authentication                             4.2.0-0.ci-2019-07-29-133200   True        False         False      61s
cloud-credential                           4.2.0-0.ci-2019-07-29-133200   True        False         False      13m
cluster-autoscaler                         4.2.0-0.ci-2019-07-29-133200   True        False         False      9m14s
console                                    4.2.0-0.ci-2019-07-29-133200   True        False         False      5m34s
dns                                        4.2.0-0.ci-2019-07-29-133200   True        False         False      13m
image-registry                             4.2.0-0.ci-2019-07-29-133200   True        False         False      3m30s
ingress                                    4.2.0-0.ci-2019-07-29-133200   True        False         False      7m53s
kube-apiserver                             4.2.0-0.ci-2019-07-29-133200   True        True          False      12m
kube-controller-manager                    4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
kube-scheduler                             4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
machine-api                                4.2.0-0.ci-2019-07-29-133200   True        False         False      13m
machine-config                             4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
marketplace                                4.2.0-0.ci-2019-07-29-133200   True        False         False      9m2s
monitoring                                 4.2.0-0.ci-2019-07-29-133200   True        False         False      7m3s
network                                    4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
node-tuning                                4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
openshift-apiserver                        4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
openshift-controller-manager               4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
openshift-samples                          4.2.0-0.ci-2019-07-29-133200   True        False         False      48s
operator-lifecycle-manager                 4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
operator-lifecycle-manager-catalog         4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
operator-lifecycle-manager-packageserver   4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
service-ca                                 4.2.0-0.ci-2019-07-29-133200   True        False         False      12m
service-catalog-apiserver                  4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
service-catalog-controller-manager         4.2.0-0.ci-2019-07-29-133200   True        False         False      10m
storage                                    4.2.0-0.ci-2019-07-29-133200   True        False         False      9m39s
support                                    4.2.0-0.ci-2019-07-29-133200   True        False         False      13m
```
particularly of note is that the `machine-config` cluster operator is not `Degraded` indicating that the MachineConfig at boostrap time is matching the MachineConfig generated/adopted by the self-hosted MCO. :+1:

mtrmac:
Thanks, updated, now with the fix fully integrated, and unit tests for both the image config controller and MCC bootstrap.

Do note these commits:
- Make the MCC bootstrap output directories owner-accessible
- Use a versioned serializer for the MCC bootstrap-generated objects

which are not strictly related to the subject of the PR, but ~necessary for tests (at least the first one).

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -111,37 +122,48 @@ func (b *Bootstrap) Run(destDir string) error {
...
+	}
+	configs = append(configs, rconfigs...)
+
 	fpools, gconfigs, err := render.RunBootstrap(pools, configs, cconfig)
 	if err != nil {
 		return err
 	}
 
+	serializer := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme, scheme)
+	encoder := codecFactory.EncoderForVersion(serializer, v1.GroupVersion)
```
mtrmac: I have no idea whether this is idiomatic; it just seems to work.

runcom:
Thanks Miloslav, both the addiornal commits lgtm as well and Seth ran a successful test as well!

/approve 

@cgwalters can you take another fast look to get this in for the disconnected work to go on?

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -111,37 +122,48 @@ func (b *Bootstrap) Run(destDir string) error {
...
+
 	poolsdir := filepath.Join(destDir, "machine-pools")
-	if err := os.MkdirAll(poolsdir, 0664); err != nil {
+	if err := os.MkdirAll(poolsdir, 0764); err != nil {
```
cgwalters: Hah, before we were just relying on `CAP_DAC_OVERRIDE` I guess?

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -59,9 +61,16 @@ func (b *Bootstrap) Run(destDir string) error {
 		return err
 	}
 
+	scheme := runtime.NewScheme()
+	v1.Install(scheme)
```
cgwalters: One thing I found is [the docs here](https://godoc.org/k8s.io/apimachinery/pkg/runtime) say:
> It's recommended that you register your types with runtime in your package's init function.

https://medium.com/@arschles/go-experience-report-generics-in-kubernetes-25da87430301
is also useful on this topic.

runcom:
```
error: unable to read image registry.svc.ci.openshift.org/ci-op-wsyzbi3d/stable-initial@sha256:dd5f011703a04c54dd7d4bb0df39aef7dbc0c91f5773e5327ee7fc39ae879302: received unexpected HTTP status: 504 Gateway Time-out
```

/retest

PR #1015, opened 2019-07-25
WIP: cmd/setup-etcd-environment: add etcd pivot logic

hexfusion:
This PR adds logic required to scale up during install via openshift-etcd-operator. More details soon..

required by: https://github.com/openshift/cluster-etcd-operator/pull/19

### DIFF
--- a/cmd/setup-etcd-environment/run.go
+++ b/cmd/setup-etcd-environment/run.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -14,7 +15,11 @@ import (
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/pkg/version"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
 )
 
 var (
@@ -29,8 +34,22 @@ var (
 		discoverySRV string
 		ifName       string
 		outputFile   string
+		pivot        bool
 	}
+	EtcdScalingAnnotationKey = "etcd.operator.openshift.io/scale"
 )
+
+type EtcdScaling struct {
+	Metadata *metav1.ObjectMeta `json:"metadata,omitempty"`
+	Members  []Member           `json:"members,omitempty"`
+}
+
+type Member struct {
+	ID         uint64   `json:"ID,omitempty"`
+	Name       string   `json:"name,omitempty"`
+	PeerURLS   []string `json:"peerURLs,omitempty"`
+	ClientURLS []string `json:"clientURLs,omitempty"`
+}
 
 func init() {
 	rootCmd.AddCommand(runCmd)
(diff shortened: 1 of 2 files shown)


cgwalters:
Would be nice to avoid the term "pivot" for this since it's highly associated here with the `pivot` command used in RHCOS (4.1).

hexfusion:
> Would be nice to avoid the term "pivot" for this since it's highly associated here with the `pivot` command used in RHCOS (4.1).

not an issue

cmd/setup-etcd-environment/run.go
```diff
@@ -86,9 +105,54 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+			panic(err.Error())
+		}
+		client, err := kubernetes.NewForConfig(clientConfig)
+		if err != nil {
+			return fmt.Errorf("error creating client: %v", err)
+		}
+		duration := 10 * time.Second
+		// wait forever for success and retry every duration interval
+		wait.PollInfinite(duration, func() (bool, error) {
+			result, err := client.CoreV1().ConfigMaps("openshift-etcd").Get("scaling-lock", metav1.GetOptions{})
```
beekhof: At least in this usage, "scaling-lock" seems misnamed.
Perhaps "known-members"?

hexfusion: yeah, this configmap originally was a distributed lock similar to the leader election process. But I did not find that to be necessary will consider renaming.

cmd/setup-etcd-environment/run.go
```diff
@@ -86,9 +105,54 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
-		"DISCOVERY_SRV": runOpts.discoverySRV,
-	}, out, true); err != nil {
+	export := make(map[string]string)
+	etcdName := os.Getenv("ETCD_NAME")
+	if etcdName == "" {
+		return fmt.Errorf("environment variable ETCD_NAME has no value")
+	}
+	var e EtcdScaling
+	if runOpts.pivot {
+		clientConfig, err := rest.InClusterConfig()
```
abhinavdahiya: Does this mean that it requires k8s networking to be up to talk to apiservers... That would be difficult in recovery scenarios

cmd/setup-etcd-environment/run.go
```diff
@@ -86,9 +105,54 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+		client, err := kubernetes.NewForConfig(clientConfig)
+		if err != nil {
+			return fmt.Errorf("error creating client: %v", err)
+		}
+		duration := 10 * time.Second
+		// wait forever for success and retry every duration interval
+		wait.PollInfinite(duration, func() (bool, error) {
+			result, err := client.CoreV1().ConfigMaps("openshift-etcd").Get("scaling-lock", metav1.GetOptions{})
+			if err != nil {
+				klog.Errorf("error creating client %v", err)
```
kikisdeliveryservice: wondering why you used klog here?

cmd/setup-etcd-environment/run.go
```diff
@@ -86,9 +105,54 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+			}
+			if err := json.Unmarshal([]byte(result.Annotations[EtcdScalingAnnotationKey]), &e); err != nil {
+				klog.Errorf("error decoding result %v", err)
+				return false, nil
+			}
+			if e.Metadata.Name != etcdName {
+				klog.Errorf("could not find self in scaling-lock")
+				return false, nil
+			}
+			members := e.Members
```
kikisdeliveryservice: do we need this placeholder var? would using e.Members directly cause an issue?

cmd/setup-etcd-environment/run.go
```diff
@@ -86,9 +105,54 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
 
-	if err := writeEnvironmentFile(map[string]string{
-		"DISCOVERY_SRV": runOpts.discoverySRV,
-	}, out, true); err != nil {
+	export := make(map[string]string)
+	etcdName := os.Getenv("ETCD_NAME")
+	if etcdName == "" {
+		return fmt.Errorf("environment variable ETCD_NAME has no value")
+	}
+	var e EtcdScaling
```
kikisdeliveryservice: could you use a more descriptive name than `e`, it's hard to remember what this stands for as i read thru below.

cmd/setup-etcd-environment/run.go
hexfusion: This functionality is part of cluster bootstrap and relies on the operator so networking would be assumed. DR would require standing up a single node etcd cluster then scaling additional members via operator.

cmd/setup-etcd-environment/run.go
hexfusion: I see we have glog will drop.

cmd/setup-etcd-environment/run.go
hexfusion: good point will adjust

cmd/setup-etcd-environment/run.go
hexfusion: sounds good.

cmd/setup-etcd-environment/run.go
hexfusion: But I will consider how to handle this for that single node. thanks.

hexfusion:
/close

Work is continued in #1221

PR #1016, opened 2019-07-25
pkg/operator: support additional CA bundle

runcom:
This patch implements the last mile to the Proxy support specifically
(and later, CAs for mirrors).
This just reads from the installer objects and set up the bundle on disk
to be later updated via the new update-ca service shipped in RHCOS.
The service should always run early enough to always have the correct CA
on system to services to be consumed.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -19,6 +19,7 @@ import (
 
 // RenderBootstrap writes to destinationDir static Pods.
 func RenderBootstrap(
+	additionalTrustBundleFile,
 	proxyFile,
 	clusterConfigConfigMapFile,
 	infraFile, networkFile,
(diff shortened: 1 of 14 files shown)


runcom:
cc @abhinavdahiya

pkg/operator/bootstrap.go
```diff
@@ -29,6 +30,7 @@ func RenderBootstrap(
 ) error {
 	filesData := map[string][]byte{}
 	files := []string{
+		additionalTrustBundleFile,
```
abhinavdahiya: the trust bundle file will only be present if it was required by the user. this is optional..

pkg/operator/sync.go
```diff
@@ -151,6 +151,12 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		return err
 	}
 
+	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
```
abhinavdahiya: again this config map is only present when requested by the user.

pkg/operator/bootstrap.go
runcom: ack makes sense

runcom: fixed

pkg/operator/sync.go
runcom: fixed

abhinavdahiya:
LGTM

templates/common/_base/files/additional-trust-bundle.yaml
```diff
@@ -0,0 +1,8 @@
+filesystem: "root"
+mode: 0600
+path: "/etc/pki/ca-trust/source/anchors/ca.crt"
```
cgwalters: Super nit but...it'd be a lot more obvious if this was named `openshift-config-user-ca-bundle.crt` or so.

runcom: ack, I thought naming was `ca.crt` by default, changing!

runcom: fixed!

cgwalters:
/lgtm
∕certified
∕shipit

Issue #1017, opened 2019-07-26
Error rendering bootstrap manifests: no kind "ConfigMap"


jcpowermac:
**Description**
When testing an installation with `addditionalTrustBundle` configured in `install-config.yaml` installation will fail.  Reviewing the bootkube.service logs shows:

```
Jul 26 18:14:19 ip-10-0-10-216 bootkube.sh[1596]: Rendering MCO manifests...
Jul 26 18:14:27 ip-10-0-10-216 podman[4769]: 2019-07-26 18:14:27.034940399 +0000 UTC m=+7.837004467 image pull
Jul 26 18:14:27 ip-10-0-10-216 podman[4769]: 2019-07-26 18:14:27.063369124 +0000 UTC m=+7.865432612 container create 0451c1a14021de10f2d0400f68fbcbf8feede03f67e66e98e3fd182afa3e3f79 (image=registry.svc.ci.openshift.org/origin/4.2-2019-07-26-171621@sha256:0b9956c2bad93bc30dccdfe9039e2756eba1fe68500334f7873fb752063f265e, name=stupefied_ellis)
Jul 26 18:14:27 ip-10-0-10-216 podman[4769]: 2019-07-26 18:14:27.388647646 +0000 UTC m=+8.190711046 container init 0451c1a14021de10f2d0400f68fbcbf8feede03f67e66e98e3fd182afa3e3f79 (image=registry.svc.ci.openshift.org/origin/4.2-2019-07-26-171621@sha256:0b9956c2bad93bc30dccdfe9
039e2756eba1fe68500334f7873fb752063f265e, name=stupefied_ellis)
Jul 26 18:14:27 ip-10-0-10-216 podman[4769]: 2019-07-26 18:14:27.401546969 +0000 UTC m=+8.203610579 container start 0451c1a14021de10f2d0400f68fbcbf8feede03f67e66e98e3fd182afa3e3f79 (image=registry.svc.ci.openshift.org/origin/4.2-2019-07-26-171621@sha256:0b9956c2bad93bc30dccdfe
9039e2756eba1fe68500334f7873fb752063f265e, name=stupefied_ellis)
[...]

PR #1018, opened 2019-07-26
pkg/operator: use corev1 codecs to decode configmaps

runcom:

**- What I did**

Fix https://github.com/openshift/machine-config-operator/issues/1017

We were wrongly using the config scheme codec. This fixes that by using the corev1 and fixes the issue linked.

@jcpowermac ptal

**- How to verify it**

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -88,7 +88,7 @@ func RenderBootstrap(
 		return err
 	}
 	if additionalTrustBundleData != nil {
-		obji, err = runtime.Decode(configscheme.Codecs.UniversalDecoder(configv1.SchemeGroupVersion), additionalTrustBundleData)
+		obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(corev1.SchemeGroupVersion), additionalTrustBundleData)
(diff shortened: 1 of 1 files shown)


jcpowermac:
@runcom thx!

pkg/operator/bootstrap.go
```diff
@@ -88,15 +88,15 @@ func RenderBootstrap(
 		return err
 	}
 	if additionalTrustBundleData != nil {
-		obji, err = runtime.Decode(configscheme.Codecs.UniversalDecoder(configv1.SchemeGroupVersion), additionalTrustBundleData)
+		obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(corev1.SchemeGroupVersion), additionalTrustBundleData)
```
ericavonb: Curious, why is this file a configmap rather than just containing the ca-bundle data directly?

runcom: The installer writes that CM to be installed and managed by some other operator (don't know the full rationale behind that but it's because it must be easier for the installer to just create the cm and reuse it here for the MCO and elsewhere)

PR #1021, opened 2019-07-29
OpenStack: create a symlink for cloud provider config

Fedosin:
To attach a cinder volume Kubernetes requires that cloud provider config be placed in /etc/kubernetes/cloud-config file, but OCP puts it in /etc/kubernetes/cloud.conf, and Kubelet refuses to work.
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/cinder/cinder.go#L44

To fix that this patch creates a symlink '/etc/kubernetes/cloud.conf' -> '/etc/kubernetes/cloud-config'

Closes: https://github.com/openshift/installer/issues/2102

### DIFF
--- a/templates/common/openstack/units/openstack-cloud-config-symlink.yaml
+++ b/templates/common/openstack/units/openstack-cloud-config-symlink.yaml
@@ -1,0 +1,14 @@
+name: "openstack-cloud-config-symlink.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Create a symlink '/etc/kubernetes/cloud.conf' -> '/etc/kubernetes/cloud-config'
+  Before=kubelet.service
+  ConditionPathExists=!/etc/kubernetes/cloud-config
+
+  [Service]
+  ExecStart=/usr/bin/ln -s /etc/kubernetes/cloud.conf /etc/kubernetes/cloud-config
+  Type=oneshot
+
+  [Install]
+  WantedBy=multi-user.target


templates/common/openstack/units/provider-config-symlink.yaml
```diff
@@ -0,0 +1,12 @@
+name: "provider-config-symlink.service"
```
cgwalters: `openstack-cloud-config-symlink.service` maybe?

templates/common/openstack/units/provider-config-symlink.yaml
```diff
@@ -0,0 +1,12 @@
...
+enabled: true
+contents: |
+  [Unit]
+  Description=Create a symlink '/etc/kubernetes/cloud.conf' -> '/etc/kubernetes/cloud-config'
```
cgwalters: Surely this would need to be `Before=kubelet.service`?

Also this service will fail if the symlink already exists (on reboot)...so you'd want `ConditionPathExists=!/etc/kubernetes/cloud-config` or so.

Hmm.  It's too bad our internal Ignition templating doesn't support symlinks, then this would be a lot more obvious.

templates/common/openstack/units/provider-config-symlink.yaml
kikisdeliveryservice: agree

Fedosin: done

templates/common/openstack/units/provider-config-symlink.yaml
Fedosin: yeah, I agree. Before=kubelet.service is not really necessary, because CP starts well and fails only when we try to attach a volume from cinder to a container. But I think it wouldn't be bad to have this dependency as well.

templates/common/openstack/units/openstack-cloud-config-symlink.yaml
```diff
@@ -0,0 +1,14 @@
...
+enabled: true
+contents: |
+  [Unit]
+  Description=Create a symlink '/etc/kubernetes/cloud.conf' -> '/etc/kubernetes/cloud-config'
+  Before=kubelet.service
+  ConditionPathExists=!/etc/kubernetes/cloud-config
+
+  [Service]
+  ExecStart=/usr/bin/ln -s /etc/kubernetes/cloud.conf /etc/kubernetes/cloud-config
+  Type=oneshot
```
cgwalters: With `Type=oneshot` you also always want `RemainAfterExit=true`, see https://github.com/ostreedev/ostree/pull/1697

cgwalters:
So much engineering to create a symlink!  Sorry about that...

I also do wonder why we aren't fixing (our?) kubelet to just look in the right place.

runcom:
> I also do wonder why we aren't fixing (our?) kubelet to just look in the right place.

this is actually a great point :thinking:

cgwalters:
I mean it's this code in the MCO doing it right?

```
func cloudConfigFlag(cfg RenderConfig) interface{} {
	if cfg.CloudProviderConfig == "" {
		return ""
	}
	flag := "--cloud-config=/etc/kubernetes/cloud.conf"
	switch cfg.Platform {
	case platformAzure, platformOpenStack:
		return flag
	default:
		return ""
	}
}
```

So...is it likely Azure is busted in the same way?  Why don't we just change that bit of code?

kikisdeliveryservice:
I am +1 on just updating our bit of code..

kikisdeliveryservice:
@Fedosin when you say in the top PR message: "Kubelet refuses to work" what were you seeing in the logs?

Fedosin:
@kikisdeliveryservice 
It says:
Failed to provision volume with StorageClass "standard": OpenStack cloud provider was not initialized properly : stat /etc/kubernetes/cloud-config: no such file or directory

@cgwalters it seems like it happens only with Cinder volume provisioning, because they hardcoded the path. Other cloud providers and even other parts of OpenStack cloud provider work well.

PR #1022, opened 2019-07-30
pkg/controller/kubelet-config: fix owener reference

runcom:

**- What I did**

This fixes a regression introduced in https://github.com/openshift/machine-config-operator/commit/680f79c9c7002bdc56b7d5f7481a2c7712fda164#diff-bc0cd3d2a7a92d7ffb804b2d7b2a5636R54 - I think this can cause the resulting kubeletConfig MC to be deleted when the kube GC kicks in but I haven't successfully verified that. Besides, the controller for kubeConfig is `KubeletConfig` so fix that.

**- How to verify it**

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -51,7 +51,7 @@ const (
 
 var (
 	// controllerKind contains the schema.GroupVersionKind for this controller type.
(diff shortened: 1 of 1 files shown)


cgwalters:
Mind including the info from the PRs in commit messages in the future?

runcom:
> Mind including the info from the PRs in commit messages in the future?

c*ap, I missed that sorry

rphillips:
/retitle Bug 1743160: pkg/controller/kubelet-config: fix owener reference

PR #1026, opened 2019-08-01
baremetal - haproxy static pod, run haproxy process if cfg file exist

yboaron:
In current code, the haproxy process in haproxy container being started/restarted
only upon request (configuration change) from the monitor container.
So in case haproxy container fail due to liveness probe, it won't run the haproxy
process until configuration will be changed.
To fix that,  the haproxy process will be started if the haproxy cfg file exists.

**- What I did**

**- How to verify it**


cgwalters:
/approve
Should probably get LGTM from the maintainer of that pod though.

runcom:
toomanybuckets errors

PR #1027, opened 2019-08-01
demonstrate simple daemonset to fix master kubelet bootstrap creds

deads2k:
related to https://bugzilla.redhat.com/show_bug.cgi?id=1735180

simplest solution I could come up with.  this fixes the "my cluster doesn't sleep issue" and fixes the disaster recovery updates.

/assign @sjenning @rphillips

### DIFF
--- a/install/0000_80_machine-config-operator_04_bootstrapdaemonset.yaml
+++ b/install/0000_80_machine-config-operator_04_bootstrapdaemonset.yaml
@@ -1,0 +1,108 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kubelet-bootstrap-cred-manager
+  namespace: openshift-machine-config-operator
+  labels:
+    k8s-app: kubelet-bootrap-cred-manager
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      k8s-app: kubelet-bootstrap-cred-manager
+  template:
+    metadata:
+      labels:
+        k8s-app: kubelet-bootstrap-cred-manager
+    spec:
+      containers:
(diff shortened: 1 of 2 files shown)


rphillips:
As discusses in slack, the kubeconfig is ignition managed and overwriting the kubeconfig in this manner will cause the MCD to validate the contents of the file and tainting the node since they differ.

deads2k:
> As discusses in slack, the kubeconfig is ignition managed and overwriting the kubeconfig in this manner will cause the MCD to validate the contents of the file and tainting the node since they differ.

I think we just spoke again in slack and this wasn't the case.  The file doesn't appear to be managed and already has "bad" content.

install/0000_80_machine-config-operator_04_bootstrapdaemonset.yaml
```diff
@@ -0,0 +1,108 @@
...
+              kubectl config set-credentials "kubelet" --token="$token" >/dev/null
+              ca_crt="$(mktemp)"; echo "$ca_crt_data" > $ca_crt
+              kubectl config set-cluster $cluster --server="$intapi" --certificate-authority="$ca_crt" --embed-certs >/dev/null
+              kubectl config set-context kubelet --cluster="$cluster" --user="kubelet" >/dev/null
+              kubectl config use-context kubelet >/dev/null
+
+              echo "----------------------------------------------------------------------"
+              echo "Print kubeconfig"
+              echo "----------------------------------------------------------------------"
+              cat "$KUBECONFIG"
```
sjenning: probably want to remove this before merging right? @deads2k was this a kinda DO NOT MERGE experimental PR or is it suitable for merging consideration?

Issue #1024, opened 2019-07-31
mastersSchedulable: true not working when set at install time


russellb:
**Description**

This is related to issue #763 and PR #937 

On a running cluster, I was able to change `mastersSchedulable` to `true` and observe the right things happening.

I just deployed a new 3-node cluster with https://github.com/openshift/installer/pull/2004 included, which sets `mastersSchedulable: true` at install time.  In this case, the MCO did not properly reconcile the master Nodes based on this setting.  After the install timed out because ingress never came up, I manually changed the `mastersSchedulable` setting to `false` and then back to `true`.  This triggered reconciliation and I observed that the Nodes correcting had the worker role applied to them.

**Steps to reproduce the issue:**
1. Deploy a 3-node cluster
2. Observe that `mastersSchedulable: true` is set at install time in the scheduler config
3. Observe that the Nodes are not properly reconciled by MCO

ravisantoshgudimetla:
@russellb - Can you provide mco logs after increasing the loglevel to 5 or above? I think, the scheduler informer is somehow missing the scheduler object creation event. https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L167. I need to look more, if this the case.

You can disable CVO and update the mco log-level here https://github.com/openshift/machine-config-operator/blob/88147c9a50342ee9177ff513970757fa0cf34120/manifests/machineconfigcontroller/deployment.yaml#L22

russellb:
2 updates:

1) This is not consistent.  It seems to happen in roughly half of my deployments.

2) I adjusted the `-v` level on the `machine-config-controller` deployment, and it reconciled the Nodes on restart

cgwalters:
This should be fixed by https://github.com/openshift/machine-config-operator/pull/1029 - please reopen if I'm mistaken.  And thanks for the fixes!

PR #1029, opened 2019-08-01
node_controller: Check scheduler config on Node changes.

russellb:
When the scheduler config option, "mastersSchedulable: true" was set
at install time, I observed inconsistent behavior.  Roughly half of
the time this controller would successfully apply this configuration.

The code was only checking the scheduler configuration in the
scheduler config object informer.  When the Add function was called,
the cache was sometimes not yet populated with Nodes.

The fix in this patch is to reconcile the scheduler config any time a
Node is added or updated, as well.

This will also ensure that this configuration option is re-applied
after unexpected changes like the worker label being manually removed
from a Node.

Closes issue #1024

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -165,30 +165,7 @@ func (ctrl *Controller) getCurrentMasters() ([]*corev1.Node, error) {
 // checkMasterNodesOnAdd makes the master nodes schedulable/unschedulable whenever scheduler config CR with name
 // cluster is created
 func (ctrl *Controller) checkMasterNodesOnAdd(obj interface{}) {
-	scheduler := obj.(*configv1.Scheduler)
-	if scheduler.Name != schedulerCRName {
-		glog.V(4).Infof("We don't care about CRs other than cluster created for scheduler config")
-		return
-	}
-	var areMastersSchedulable bool
-	if scheduler.Spec == (configv1.SchedulerSpec{}) {
-		glog.V(4).Infof("Scheduler spec is nil, so set master as unschedulable")
-		areMastersSchedulable = false
-	} else {
-		areMastersSchedulable = scheduler.Spec.MastersSchedulable
-	}
-	currentMasters, err := ctrl.getCurrentMasters()
-	if err != nil {
-		goerrs.Wrap(err, "Reconciling to make master nodes schedulable/unschedulable failed")
-		return
-	}
-
-	// reconcile master labels and taints to make the master node schedulable, unschedulable
-	err = ctrl.reconcileSchedulableMasters(areMastersSchedulable, currentMasters)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -386,6 +395,17 @@ func (ctrl *Controller) addNode(obj interface{}) {
 		return
 	}
 
+	mastersSchedulable, err := ctrl.getMastersSchedulable()
+	if err != nil {
+		goerrs.Wrap(err, "Getting scheduler config failed")
+		return
+	}
+	err = ctrl.reconcileSchedulableMasters(mastersSchedulable)
```
ravisantoshgudimetla: Why to make all masters schedulable when add a node? we can check if the current node is master and then make it schedulable.

russellb: Makes sense - I guess I was taking the lazy brute force approach and re-using the existing code.  I'll update this

pkg/controller/node/node_controller.go
```diff
@@ -379,13 +379,60 @@ func (ctrl *Controller) deleteMachineConfigPool(obj interface{}) {
...
+}
+
+// Determine if a given Node is a master
+func (ctrl *Controller) isMaster(node *corev1.Node) bool {
+	_, master := node.ObjectMeta.Labels[masterLabel]
+	return master
+}
+
+// Given a master Node, ensure it reflects the current mastersSchedulable setting
+func (ctrl *Controller) reconcileMaster(node *corev1.Node) {
```
ravisantoshgudimetla: Can you call this function from https://github.com/openshift/machine-config-operator/blob/76e5737c1cfc183974c361d9593179535d97176a/pkg/controller/node/node_controller.go#L250, so that we avoid code duplication?

russellb: done

cgwalters:
/approve

This looks *much* better, thank you!

cgwalters:
If no one has any further comments in the next hour or two, I'll drop a lgtm here.

ravisantoshgudimetla:
Yeah, just to give everyone some background, Bryan and I debugged this yesterday and found that node cache is not getting populated by the time scheduler informer handler functions are fired causing the problem. We are not technically waiting till are caches are re-synced. This is kind of producer-consumer problem where we need to wait till node informer event handler fns fill the caches(so the node lister returns non-empty list) and then scheduler informer's event handlers can consume the node list

This might not be a problem if the nodes are filled in properly by the time scheduler informer fires events.(Which was happening when master node scheduling was a day-2 operation). When master scheduling became a day-1 operation(during install time), the approach outlined in the PR is what we could come up with. 

I am fine with lgtm'ing this PR but it'd be lgtming my own PR :(. So, @cgwalters or @runcom, @kikisdeliveryservice please do the honours as you guys are already familiar with the original PR

PR #1030, opened 2019-08-01
Telemetry: add prometheus service monitor and MCD metrics listener

sgreene570:
**- What I did**
Add required components for the Prometheus operator to scrape MCD metrics via TLS. 

We can't serve Prometheus metrics via open `http` since the metrics we will be implementing will consist of data that should be restricted to cluster admins. To serve metrics via authenticated `https`, this PR adds an `oauth-proxy` container to the MCD daemonset. The MCD container then serves a metrics listener via `http` on `localhost`. The `oauth-proxy` forwards incoming secure (and authenticated) connections on port `9001` of the pod to the local metrics listener. This is how the [elastic search operator](https://github.com/openshift/elasticsearch-operator) handles secure metrics.

**- How to verify it**

- Build MCO image and deploy.
- Verify that the MCD is visible in the cluster prometheus UI under the "targets" listing:
 From the OCP page see Monitoring -> Metrics -> Prometheus UI -> Status (Top menu) -> Targets (dropdown)

ex:
proof

To see the Prometheus data manually, run the following on an MCD pod.

`kubectl port-forward -n openshift-machine-config-operator machine-config-daemon-xxxx 9001:9001`

then in a new terminal with `KUBECONFIG` set, 

`curl -H "Authorization: Bearer $(oc serviceaccounts -n openshift-monitoring get-token prometheus-k8s)" -k https://127.0.0.1:9001/metrics`

This should list the contents being scraped by the Prometheus operator.

To try out a sample metric through the golang client, follow this basic example in the [cluster auth operator](https://github.com/openshift/cluster-authentication-operator/blob/226f263d77b90ece9f36d229ed0f6ecff3405cce/pkg/version/version.go#L37-L46)

(The Go prom client dependency is already included in this PR)

[4.3] Add prometheus service monitor and MCD metrics listener

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -46,7 +46,7 @@ func init() {
 	startCmd.PersistentFlags().BoolVar(&startOpts.skipReboot, "skip-reboot", false, "Skips reboot after a sync, applies only in once-from")
 	startCmd.PersistentFlags().BoolVar(&startOpts.kubeletHealthzEnabled, "kubelet-healthz-enabled", true, "kubelet healthz endpoint monitoring")
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeletHealthzEndpoint, "kubelet-healthz-endpoint", "http://localhost:10248/healthz", "healthz endpoint to check health")
-	startCmd.PersistentFlags().StringVar(&startOpts.promMetricsURL, "metrics-url", "127.0.0.1:8080", "URL for prometheus metrics listener")
+	startCmd.PersistentFlags().StringVar(&startOpts.promMetricsURL, "metrics-url", "127.0.0.1:8797", "URL for prometheus metrics listener")
 }
 
 // bindPodMounts ensures that the daemon can still see e.g. /run/secrets/kubernetes.io
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -33,6 +33,7 @@ var (
 		fromIgnition           bool
 		kubeletHealthzEnabled  bool
 		kubeletHealthzEndpoint string
+		promMetricsURL         string
 	}
 )
 
@@ -45,6 +46,7 @@ func init() {
 	startCmd.PersistentFlags().BoolVar(&startOpts.skipReboot, "skip-reboot", false, "Skips reboot after a sync, applies only in once-from")
 	startCmd.PersistentFlags().BoolVar(&startOpts.kubeletHealthzEnabled, "kubelet-healthz-enabled", true, "kubelet healthz endpoint monitoring")
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeletHealthzEndpoint, "kubelet-healthz-endpoint", "http://localhost:10248/healthz", "healthz endpoint to check health")
+	startCmd.PersistentFlags().StringVar(&startOpts.promMetricsURL, "metrics-url", "127.0.0.1:8080", "URL for prometheus metrics listener")
 }
 
 // bindPodMounts ensures that the daemon can still see e.g. /run/secrets/kubernetes.io
@@ -134,6 +136,9 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	// This channel is used to ensure all spawned goroutines exit when we exit.
 	stopCh := make(chan struct{})
 	defer close(stopCh)
+
+	// Start local metrics listener
+	go daemon.StartMetricsListener(startOpts.promMetricsURL, stopCh)
 
 	ctx := controllercommon.CreateControllerContext(cb, stopCh, componentName)
 	// create the daemon instance. this also initializes kube client items
--- a/manifests/machineconfigdaemon/cookie-secret.yaml
+++ b/manifests/machineconfigdaemon/cookie-secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cookie-secret
+  namespace: {{.TargetNamespace}}
+type: Opaque
+data:
+  cookie-secret: {{.GenerateProxyCookieSecret}}
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -8,8 +8,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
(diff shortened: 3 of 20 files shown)


cmd/machine-config-daemon/start.go
```diff
@@ -131,6 +133,9 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	stopCh := make(chan struct{})
 	defer close(stopCh)
 
+	// Start local metrics listener
+	go daemon.StartMetricsListener(startOpts.metricsURL)
```
lucab: This should be taking the stop-channel (or even better, a context) for graceful termination.

pkg/operator/render.go
```diff
@@ -114,3 +116,7 @@ func clusterDNSIP(iprange string) (string, error) {
 	}
 	return ip.String(), nil
 }
+
+func (rc renderConfig) GenerateProxyCookieSecret() string {
```
kikisdeliveryservice: add a comment/description this func.

pkg/operator/sync.go
```diff
@@ -337,6 +337,22 @@ func (optr *Operator) syncMachineConfigDaemon(config *renderConfig) error {
 		return err
 	}
 
+	// Only generate a new proxy cookie secret if the secret does not exist or if it has been deleted.
+	_, err = optr.kubeClient.CoreV1().Secrets(config.TargetNamespace).Get("cookie-secret", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		csBytes, err := renderAsset(config, "manifests/machineconfigdaemon/cookie-secret.yaml")
```
kikisdeliveryservice: let's use cookiesecrets instead of cs here since cs is used elsewhere as client set

manifests/machineconfigdaemon/cookie-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cookie-secret
+  namespace: {{.TargetNamespace}}
+type: Opaque
+data:
+  cookie-secret-seed: {{.GenerateProxyCookieSecret}}
```
kikisdeliveryservice: q: what does cookie-secret-seed mean?

cmd/machine-config-daemon/start.go
```diff
@@ -32,6 +32,7 @@ var (
 		fromIgnition           bool
 		kubeletHealthzEnabled  bool
 		kubeletHealthzEndpoint string
+		metricsURL             string
```
kikisdeliveryservice: Q: metricsURL or promMetricsURL? i dont know, you tell me, is this generic or specific only for prom?

manifests/machineconfigdaemon/cookie-secret.yaml
sgreene570: I'm not really sure, will change to just `cookie-secret`

cmd/machine-config-daemon/start.go
sgreene570: This is technically specific to prom since the metricsURL will have a prom HTTP handler.  Will update.

pkg/daemon/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+
+func StartMetricsListener(addr string, stopCh chan struct{}) {
+	if addr == "" {
+		addr = DefaultBindAddress
+	}
+
+	glog.Infof("Starting metrics listener on %s", addr)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	func() {
```
chancez: What's up with the random closure?

pkg/daemon/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+		if err := http.ListenAndServe(addr, mux); err != nil {
+			glog.Exitf("Unable to start metrics listener: %v", err)
+		}
+	}()
+	<-stopCh
```
chancez: You could `select` on stopCh in a go routine and stop the http server when the channel is closed.

runcom:
@kikisdeliveryservice we need to carry this PR and land it for the telemetry work

kikisdeliveryservice:
Update:

- rebased (normal) and then rebased/updated again (pick up go mods)
- walked through instructions and verified
- updated instructions for clarity

Just want to check this out tomorrow a little more then we should be good to go. :+1:

kikisdeliveryservice:
these tests have consistently passed on this pr, only pushed a rebase on master for this today.  ci/aws issues galore tho so..

/retest

kikisdeliveryservice:
this scaleup-rhel job has failed all day...
/skip

kikisdeliveryservice:
hmm this pr has consistently passed these tests and even passed (and had successful clusters spun up to use) after i updated to use go modules yesterday.  today i just picked up the new commits from master but they were all docs/dr commits so shouldnt have changed anything.  not sure if something else was merged into ocp (reported some things to 4-dev earlier today - the failed to sync stuff wasn't unique to this pr). ill go investigate more tomorrow...

/retest

pkg/daemon/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+		addr = DefaultBindAddress
+	}
+
+	glog.Infof("Starting metrics listener on %s", addr)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	s := http.Server{Addr: addr, Handler: mux}
+	go func() {
+		if err := s.ListenAndServe(); err != nil {
+			glog.Exitf("Unable to start metrics listener: %v", err)
```
lucab: This will abruptly `os.Exit()` the whole process, killing any pending `defer` in any other part of the application. I'm not sure if this this the expected behavior (versus more traditional error bubbling).

kikisdeliveryservice: thanks will double check and update.

kikisdeliveryservice: so i think maybe i should make an error chan send it to this func then send this message up as an error (not an exitf()) and check for it in start.go? not totally sure but seems to be the correct way to pass errors up from go routines?

kikisdeliveryservice:
Ahh that worked! Will clean up.

kikisdeliveryservice:
these 2 tests (distruptive & scaleup ) are constantly failing across prs

/skip

PR #1031, opened 2019-08-02
 cmd: add gcp-routes-controllers to manage the routes service to prevent blackholes

abhinavdahiya:
turn a service on/off based on the status of the health check provided.

** Background

On GCP, network load balancing requires modification of the local machine's routes to include the forwarded_ips of the LB rules.
Therefore, for a VM in the backend pool, the traffic to the LB never leaves the box (regardless of the health checks...)
for more info: https://cloud.google.com/load-balancing/docs/internal/setting-up-internal#test-from-backend-vms

This esp. affects the OpenShift control-plane as the clients like kubelet, openshift-networking (SDN) communicate with the apiserver using the LB.. and therefore can only communicate if the
local apiservers is running. This is very fragile during roll-out of the apiserver pods as any interminent faliure to start the local apiserver pod means that kubelet on the machine can no longer
reach any api to receive new pods for recovery to new version of the apiserver. A lot of the early users and CI is seeing these kinds of failures.

** What

With https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/481, the RHCOS's gcp-routes.service now cleans up any routes that were added for forwarded_ips when stopped, so to recover the blackhole, a user can logon to the machine
and stop the service long enough for progress to happen and then start the service again to get traffic again..

gcp-routes-controller performs health checks on the local apiserver so that,
- When the apiserver is known to be failing, the momentum to mark failure is large enough to not interfer with LB's halthchecks and trigger only in require's intervention cases, stop the gcp-routes.service to allow clients to communicate to other apiservers over the LB
- When the apiserver is running (succeeding), the momentum to mark sucess to small enough that the Lb's health checks will adde the instance to it's backend, it starts the service.

** What it solves

It solves 2 cases for GCP:

1) Improves the reliability of the apiserver roll-outs on GCP, making sure the kubelets have a chance to receive new versions that can unstick roll-outs

2) Improve the installation of GCP.
   Currently, the installer during bootstrapping only adds the bootstrap-host to the LB backend, as adding the control-plane makes the kubelets on those nodes incapaable to reaching out the bootstrap-control-plane on the bootstrap-host
   and switches the backends to control-plane during bootstrap destroy, making destroy bootstrap a required step to complete the bootstrapping and also creating a period of time where there are no active backends.
   On other platforms, installer keeps all nodes in the LB backend and after bootstrap-complete the bootstap-host is remove from the backends by the LB itself and technically the bootstrap destroy step is not required.

   This helps bring the GCP flow at par with others as the gcp-route-controller on the control-plane hosts will run the gcp-routes when the apiserver is running locally and the installer can keep all machines in the backend.

Other suggestions:

1) The same can be achieved by removing and adding nodes from the backend.
For this to work, all forwarding rules must be updated the remove the machine, which requires more nuanced agent that lists the forwarding rule to find out which backends have itself included and tracking membership... which can be api limits heavy.

### DIFF
--- a/cmd/gcp-routes-controller/run.go
+++ b/cmd/gcp-routes-controller/run.go
@@ -73,6 +73,8 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 	httpCheck, err := checkers.NewHTTP(&checkers.HTTPConfig{
 		URL: uri,
 		Client: &http.Client{Transport: &http.Transport{
+			// #nosec G402
+			// health checks to https endpoints can use InsecureSkipVerify.
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		}},
 	})
--- a/cmd/gcp-routes-controller/run.go
+++ b/cmd/gcp-routes-controller/run.go
@@ -1,0 +1,181 @@
+package main
+
+import (
+	"crypto/tls"
+	"flag"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"os/exec"
+	"sync"
+	"syscall"
+	"time"
+
+	health "github.com/InVisionApp/go-health"
+	"github.com/InVisionApp/go-health/checkers"
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/machine-config-operator/pkg/version"
+)
+
+var (
+	runCmd = &cobra.Command{
+		Use:   "run",
+		Short: "Runs the gcp-routes-controller",
+		Long:  "",
+		RunE:  runRunCmd,
+	}
+
+	runOpts struct {
+		gcpRoutesService string
+		rootMount        string
+
+		healthCheckURL string
+	}
+)
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+	runCmd.PersistentFlags().StringVar(&runOpts.gcpRoutesService, "gcp-routes-service", "gcp-routes.service", "The name for the service controlling gcp routes on host")
+	runCmd.PersistentFlags().StringVar(&runOpts.rootMount, "root-mount", "/rootfs", "where the nodes root filesystem is mounted for chroot and file manipulation.")
+	runCmd.PersistentFlags().StringVar(&runOpts.healthCheckURL, "health-check-url", "", "HTTP(s) URL for the health check")
+}
+
+func runRunCmd(cmd *cobra.Command, args []string) error {
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+
+	// To help debugging, immediately log version
+	glog.Infof("Version: %+v (%s)", version.Raw, version.Hash)
+
+	if runOpts.rootMount != "" {
+		glog.Infof(`Calling chroot("%s")`, runOpts.rootMount)
+		if err := syscall.Chroot(runOpts.rootMount); err != nil {
+			return fmt.Errorf("Unable to chroot to %s: %s", runOpts.rootMount, err)
+		}
+
+		glog.V(2).Infof("Moving to / inside the chroot")
+		if err := os.Chdir("/"); err != nil {
+			return fmt.Errorf("Unable to change directory to /: %s", err)
+		}
+	}
+
+	uri, err := url.Parse(runOpts.healthCheckURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse health-check-url: %v", err)
+	}
+	if !uri.IsAbs() {
+		return fmt.Errorf("invalid URI %q (no scheme)", uri)
+	}
+
+	httpCheck, err := checkers.NewHTTP(&checkers.HTTPConfig{
+		URL: uri,
+		Client: &http.Client{Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create httpCheck: %v", err)
+	}
+
+	errCh := make(chan error)
+	tracker := &healthTracker{
+		ErrCh:            errCh,
+		SuccessThreshold: 2,
+		FailureThreshold: 10,
+		OnFailure:        func() error { return exec.Command("systemctl", "stop", runOpts.gcpRoutesService).Run() },
+		OnSuccess:        func() error { return exec.Command("systemctl", "start", runOpts.gcpRoutesService).Run() },
+	}
+
+	h := health.New()
+	h.AddChecks([]*health.Config{{
+		Name:       "dependency-check",
+		Checker:    httpCheck,
+		Interval:   time.Duration(5) * time.Second,
+		Fatal:      true,
+		OnComplete: tracker.OnComplete,
+	}})
+
+	if err := h.Start(); err != nil {
+		return fmt.Errorf("failed to start heath checker: %v", err)
+	}
+
+	for {
+		select {
+		case err := <-errCh:
+			if err != nil {
+				return fmt.Errorf("error running health checker: %v", err)
+			}
+		}
+	}
+}
+
+type trackerState int
+
+const (
+	unknownTrackerState trackerState = iota
+	failedTrackerState
+	succeededTrackerState
+)
+
+type healthTracker struct {
+	sync.Mutex
+
+	state trackerState
+
+	succeeded int
+	failed    int
+
+	// ErrCh is used to collect errors
+	ErrCh chan<- error
+
+	// SuccessThreshold is the number of consecutive success
+	SuccessThreshold int
+
+	// FailureThreshold is the number of consecutive failure that trigger OnFailure func
+	FailureThreshold int
+
+	// OnFailure is the function that is triggered when the health check is in FAILED state
+	// Non nil error are sent over the ErrCh
+	// Only one OnFailure function will be active at a time.
+	OnFailure func() error
+
+	// OnSuccess is the function that is triggered when the health check is in SUCCEEDED state
+	// Non nil error are sent over the ErrCh
+	// Only one OnFailure function will be active at a time.
+	OnSuccess func() error
+}
+
+func (sl *healthTracker) OnComplete(state *health.State) {
+	sl.Lock()
+	defer sl.Unlock()
+
+	switch state.Status {
+	case "ok":
+		sl.failed = 0
+		sl.succeeded++
+		if sl.succeeded >= sl.SuccessThreshold {
+			if sl.state != succeededTrackerState {
+				glog.Info("Running OnSuccess trigger")
+				if err := sl.OnSuccess(); err != nil {
+					sl.ErrCh <- err
+				}
+			}
+			sl.state = succeededTrackerState
+		}
+	case "failed":
+		sl.succeeded = 0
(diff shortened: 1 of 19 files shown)


templates/master/00-master/gcp/files/etc-kubernetes-manifests-gcp-routes-controller.yaml
```diff
@@ -0,0 +1,36 @@
...
+path: "/etc/kubernetes/manifests/gcp-routes-controller.yaml"
+contents:
+  inline: |
+    apiVersion: v1
+    kind: Pod
+    metadata:
+      name: gcp-routes-controller
+      namespace: kube-system
+      labels:
+        k8s-app: etcd
```
cgwalters: Copy-pasta

cgwalters:
If we have this controller here, it may argue for moving the shell script that's currently part of the OS into the MCO too.

One random idea I've been considering lately is allowing different github repos, most notably the MCO and installer, to build code that goes into RHCOS.  The MCO builds the MCD as an RPM, but we could support skipping the RPM bit too.  You could imagine we have an `rhcos-overlay/` directory here or so that works just like the overlay dir cosa supports. Anyways we don't have it right now, just an aside.

abhinavdahiya:
> If we have this controller here, it may argue for moving the shell script that's currently part of the OS into the MCO too.

We the routes service is necessary for general deployment to work with LB.. the gcp-routes-controller (_i am not sure if that is the right name_) is esp for the control-plane which has some restrictions.

cgwalters:
/approve
The description in the PR is great.  The overall idea and code makes sense to me, although I have to say I'm not a GCP or networking expert.

The new code feels like it could be ~5 lines of shell script executed as a systemd unit, although an interesting twist today is I'm not sure we have a standard way to detect there that a node is a master.
(Anywhere we are doing "privileged daemonset" in particular hostNetwork + chroot(/host) we should be considering whether it's better on the host, but again it's OK as is)

cgwalters:
It looks like you have one unused variable blocking the "verify" job.  The "e2e-aws-op" timeouts are known...I think we're as of recently just Doing Too Much in that job, but the team will look at this today.

runcom:
> It looks like you have one unused variable blocking the "verify" job. The "e2e-aws-op" timeouts are known...I think we're as of recently just Doing Too Much in that job, but the team will look at this today.

90m should have been plenty to run everything, maybe we're doing something wrong, somewhere, I'm looking.

abhinavdahiya:
/hold 

for approval from @crawford  / @smarterclayton

cmd/gcp-routes-controller/run.go
```diff
@@ -0,0 +1,181 @@
...
+				return fmt.Errorf("error running health checker: %v", err)
+			}
+		}
+	}
+}
+
+type trackerState int
+
+const (
+	unknownTrackerState trackerState = iota
```
kikisdeliveryservice: this var is unused so it's going to fail verify unless removed.

abhinavdahiya: https://github.com/openshift/machine-config-operator/pull/1031/commits/d83c4e499a093a1cf85fca2f0c314feba515a619

kikisdeliveryservice: :+1:

pkg/controller/template/constants.go
```diff
@@ -7,6 +7,9 @@ const (
 	// SetupEtcdEnvKey is the key that references the setup-etcd-environment image in the controller
 	SetupEtcdEnvKey string = "setupEtcdEnvKey"
 
+	// GCPRoutesController is the key that references the gcp-routes-controller image in the controller
+	GCPRoutesController string = "gcpRoutesController"
```
kikisdeliveryservice: change to "GCPRoutesControllerKey" to maintain consistent naming here.

kikisdeliveryservice: change to "GCPRoutesControllerKey" to maintain consistent naming here.

pkg/operator/bootstrap.go
```diff
@@ -132,6 +132,7 @@ func RenderBootstrap(
 	spec.Images = map[string]string{
 		templatectrl.EtcdImageKey:            imgs.Etcd,
 		templatectrl.SetupEtcdEnvKey:         imgs.MachineConfigOperator,
+		templatectrl.GCPRoutesController:     imgs.MachineConfigOperator,
```
kikisdeliveryservice: name update

pkg/operator/sync.go
```diff
@@ -179,6 +179,7 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 	spec.Images = map[string]string{
 		templatectrl.EtcdImageKey:            imgs.Etcd,
 		templatectrl.SetupEtcdEnvKey:         imgs.MachineConfigOperator,
+		templatectrl.GCPRoutesController:     imgs.MachineConfigOperator,
```
kikisdeliveryservice: name update

templates/master/00-master/gcp/files/etc-kubernetes-manifests-gcp-routes-controller.yaml
```diff
@@ -0,0 +1,35 @@
...
+    apiVersion: v1
+    kind: Pod
+    metadata:
+      name: gcp-routes-controller
+      namespace: kube-system
+    spec:
+      containers:
+      - name: gcp-routes-controller
+        image: "{{.Images.gcpRoutesController}}"
+        command: ["gcp-routes-controller"]
```
kikisdeliveryservice: won't this have to be ["usr-bin-gcp-routes-controller"] like here:

https://github.com/openshift/machine-config-operator/blob/8367a767919c112142f1559d56b0d08aa7bf6613/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml#L17

runcom: `/usr/bin` is in `$PATH`, I guess `gcp-routes-controller` is there as well

kikisdeliveryservice: ah fair enough then. ty @runcom

pkg/controller/template/constants.go
abhinavdahiya: done with https://github.com/openshift/machine-config-operator/pull/1031/commits/a60e68569d9cdfda0ff0d00ca770bb740396437f

pkg/operator/sync.go
abhinavdahiya: done with https://github.com/openshift/machine-config-operator/pull/1031/commits/a60e68569d9cdfda0ff0d00ca770bb740396437f

pkg/operator/bootstrap.go
abhinavdahiya: done with https://github.com/openshift/machine-config-operator/pull/1031/commits/a60e68569d9cdfda0ff0d00ca770bb740396437f

pkg/controller/template/constants.go
kikisdeliveryservice: thanks!

squeed:
Two thoughts:

1. It's important that the failure thresholds are equal or looser to the load balancer's configuration. This is because removing the IP from the local table will *immediately* cause all incoming connections to be dropped, whereas we want to allow connections to drain from a dying apiserver. Because you now have a value duplicated across codebases, could you document in both places (this code and the installer's load balancer definitions)?
2. This assumes that there will only ever be one service pointing to masters that are load-balanced (and possibly host-network... not sure about that). While that assumption is currently true, will it always be?
3. We might be able to solve this by adding a virtual IP address and some static NAT rules via `ip rule`. This would be slightly hacky but could solve the problem neatly.

abhinavdahiya:
> Some thoughts:
> 
> 1. It's important that the failure thresholds are equal or looser to the load balancer's configuration. This is because removing the IP from the local table will _immediately_ cause all incoming connections to be dropped, whereas we want to allow connections to drain from a dying apiserver. Because you now have a value duplicated across codebases, could you document in both places (this code and the installer's load balancer definitions)?

> gcp-routes-controller performs health checks on the local apiserver so that,
When the apiserver is known to be failing, the momentum to mark failure is large enough to not interfer with LB's halthchecks and trigger only in require's intervention cases, stop the gcp-routes.service to allow clients to communicate to other apiservers over the LB

Agree and that's why I mentioned it in the commit and PR description.

> 2. This assumes that there will only ever be one service pointing to masters that are load-balanced (and possibly host-network... not sure about that). While that assumption is currently true, will it always be? In other words, will masters always be special?

Yes, currently it is very naive and treats the apiserver running on the control-plane as the only service that matters.. which I think is fine. That assumption is a fair one to make as I don't think we are going away from self hosted control plane anytime soon.

abhinavdahiya:
/hold cancel

/retest

ping @openshift/openshift-team-mco

kikisdeliveryservice:
like @cgwalters I'm fine with these changes, but would defer to someone who has a better understanding to LGTM.

templates/master/00-master/gcp/files/etc-kubernetes-manifests-gcp-routes-controller.yaml
```diff
@@ -0,0 +1,35 @@
...
+      name: gcp-routes-controller
+      namespace: kube-system
+    spec:
+      containers:
+      - name: gcp-routes-controller
+        image: "{{.Images.gcpRoutesControllerKey}}"
+        command: ["gcp-routes-controller"]
+        args:
+        - "run"
+        - "--health-check-url=https://127.0.0.1:6443/healthz"
```
sttts: shouldn't this be /readyz? We want an early signal in order to not drop traffic or route it wrongly.

runcom: @abhinavdahiya ptal

abhinavdahiya: > shouldn't this be /readyz? We want an early signal in order to not drop traffic or route it wrongly.

https://github.com/openshift/installer/blob/e8b75d0bd2a5c254fa3be8d1fadee902301ca4e3/data/data/gcp/network/lb-private.tf#L12
https://github.com/openshift/installer/blob/e8b75d0bd2a5c254fa3be8d1fadee902301ca4e3/data/data/aws/vpc/master-elb.tf#L66

We use the readyz for the LB... for graceful shutdown. the routes-controller is meant to open the egress to other apiservers when we know the one on this is no longer running, not just accepting new connections..

but i'm fine with what the apiserver team recommends.

sttts: I don't see a difference to the LB. With readyz we report "don't send me traffic". This is the same for local clients. With healthz you will always have a race because when it fails, it is too late already, possibly for seconds.

runcom: created https://github.com/openshift/machine-config-operator/pull/1417 - please let me know about BZs and I'll go ahead and create them

PR #1032, opened 2019-08-02
WIP Use gopkg.in to be able to import versioned ignition

vrutkovs:
**- What I did**
Update gopkg files to avoid vendoring github.com/coreos/ignition so that both spec2 and spec3 could supported in MCO (and eventually in installer)


vrutkovs:
That's not gonna work :/

PR #1040, opened 2019-08-06
Revert "openstack: Allow to run MCO changes independently of the installer patches" 

Fedosin:
This reverts commit 759702e7bd490099065405cb90742a2a5ce6d2fc.

 openshift/installer#1959 is merged, so we can enable template rendering again.

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -180,11 +180,6 @@ func generateMachineConfigForName(config *RenderConfig, role, name, templateDir,
 	platformDirs := []string{}
 	// Loop over templates/common which applies everywhere
 	for _, dir := range []string{platformBase, platform} {
-		// Bypass OpenStack template rendering until
-		// https://github.com/openshift/installer/pull/1959 merges
-		if dir == platformOpenStack && config.ControllerConfigSpec.Infra.Status.PlatformStatus.OpenStack == nil {
-			continue
-		}
 		basePath := filepath.Join(templateDir, "common", dir)
 		exists, err := existsDir(basePath)
 		if err != nil {
@@ -197,11 +192,6 @@ func generateMachineConfigForName(config *RenderConfig, role, name, templateDir,
 	}
 	// And now over the target e.g. templates/master
 	for _, dir := range []string{platformBase, platform} {
-		// Bypass OpenStack template rendering until
-		// https://github.com/openshift/installer/pull/1959 merges
-		if dir == platformOpenStack && config.ControllerConfigSpec.Infra.Status.PlatformStatus.OpenStack == nil {
-			continue
-		}
 		platformPath := filepath.Join(path, dir)
 		exists, err := existsDir(platformPath)
 		if err != nil {


kikisdeliveryservice:
@Fedosin that commit says it should be reverted once openshift/installer#1959 merges. Was that message a mistake?

Fedosin:
@kikisdeliveryservice  as I understand there is a cyclic dependence between #740 and openshift/installer#1959. So we can't merge neither without breaking the ci. As a fix @mandre suggested to temporary disable template rendering for OpenStack as a part of #740 (759702e), then merge openshift/installer#1959 and finally revert 759702e to enable rendering again.

mandre:
Correct, https://github.com/openshift/machine-config-operator/commit/759702e7bd490099065405cb90742a2a5ce6d2fc is only useful to keep e2e-openstack green until https://github.com/openshift/installer/pull/1959 and can be safely reverted after that.

kikisdeliveryservice:
ahhh gotcha!  thanks for the explanations @Fedosin & @mandre

pkg/controller/template/render.go
```diff
@@ -148,6 +148,12 @@ func filterTemplates(toFilter map[string]string, path string, config *RenderConf
 			return nil
 		}
 
+		// Temporarily disable template rendering for all openstack files except "afterburn-hostname.yaml"
+		// This code must be removed when https://github.com/openshift/installer/pull/1959 is merged.
+		if strings.Contains(info.Name(), "openstack") && !strings.Contains(info.Name(), "afterburn-hostname.yaml") {
```
kikisdeliveryservice: should this be more like `config.ControllerConfigSpec.Platform == platformOpenStack && !strings.Contains(info.Name(), "afterburn-hostname.yaml" ` ?

Fedosin: @kikisdeliveryservice even better! thanks :)

kikisdeliveryservice: did some debugging on the now failing unit test and my suggestion was way over-inclusive and incorrect. oops? better revert.

mandre: The problem is that if it merges, can't have green CI on https://github.com/openshift/installer/pull/1959 as this won't render the templates.

In the patch you reverted, I checked for `config.ControllerConfigSpec.Infra.Status.PlatformStatus.OpenStack` that is set in openshift/installer#1959.

mandre: Something like `if config.ControllerConfigSpec.Infra.Status.PlatformStatus.OpenStack == nil && strings.Contains(info.Name(), "openstack") && !strings.Contains(info.Name(), "afterburn-hostname.yaml") {` should do it.

kikisdeliveryservice:
@Fedosin so @mandre said that this PR might be unneeded now?

See here: https://github.com/openshift/machine-config-operator/pull/1050#issuecomment-519497596

kikisdeliveryservice:
AHH ok I see, this is now just doing the reverted mentioned in 1050.

Issue #1037, opened 2019-08-06
baremetal: Bootstrap keepalived should pass --cluster-config


bcrochet:
A recent change to the installer (openshift/installer@82d81d9) copies kubeconfig-loopback instead of kubeconfig to /etc/kubernetes/kubeconfig.

This causes a problem for baremetal bootstrap. It points to localhost.

Proposed fix is to read from cluster-config.yaml first, and fall back to kubeconfig if that's not available.

The coredns bootstrap pod already passes --cluster-config to baremetal-runtimecfg. Keepalived needs to as well.

Related to openshift/baremetal-runtimecfg#15
Depends on openshift/baremetal-runtimecfg#16

PR #1043, opened 2019-08-07
baremetal: keepalived to pass cluster-config.yml to runtimecfg

bcrochet:
A recent installer change started using a loopback address in kubeconfig
as the default on the bootstrap node. baremetal-runtimecfg will begin
using cluster-config.yml if it exists first, in order to get the domain
and name of the cluster.

This patch adds a flag for this to the keepalived bootstrap pod in order
to utilize it. This should only affect the bootstrap pods.

Fixes: #1037
Depends on openshift/baremetal-runtimecfg#16


runcom:
cc @celebdor

PR #1044, opened 2019-08-07
Remove unused internal MCOConfig types and CRD

LorbusChris:
**- What I did**
Remove unused MCOConfig types and CRD

**- How to verify it**
CI

Removed unused MCOConfig types and CRD


kikisdeliveryservice:
is this 4.3?

/hold

runcom:
/hold

pending question to Architects/GL about this being a _breaking API change_

runcom:
asked @derekwaynecarr

LorbusChris:
I aimed this at 4.3, but with the branch not being on par with master, this PR is against master for now (can change that later when 4.3 branch has caught up).

LorbusChris:
/hold
I just realized that MCOConfig is probably the right place to configure which version of Ignition spec (and possibly kubelet config, slightly different situation with that) the MCO chooses to use/serve.
@cgwalters IIRC you suggested to make MCOConfig an internal API (and not move it over to api reo evtly). Does this idea divert that train of thought?

LorbusChris:
@runcom @cgwalters @yuqi-zhang do we want to use MCOConfig for anything in the future, like e.g config'ing the Ignition spec version to run or not? (I guess we don't need to make Ign Spec Verion explicitly configurable, but does any other use-case come to mind? Otherwise I'd revive this PR and remove the stuff entirely, instead of letting it hang around unused for longer)

LorbusChris:
@crawford just to triple-check before I rebase this,
MCOConfig is not going to be used in the future and can still be removed?
Does this need a Note somewhere?

crawford:
Errmmm, I didn't realize that this even existed. If we've never used it for anything (as appears to be the case), we should remove it. In the future, if we want the operator to change behavior based on configuration, that configuration should be expressed in one of the global config objects or potentially a new one (but the config should not be specific to that operator implementation).

LorbusChris:
This aligns with my train of thought, thanks! Will rebase shortly.

LorbusChris:
Rebased, should be good to go now

PR #1045, opened 2019-08-07
pkg/operator/sync.go: udpate the namespace for trust configmap

abhinavdahiya:
followup to https://github.com/openshift/installer/pull/2173

/cc @jcpowermac @runcom @danehans

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -155,11 +155,34 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		return err
 	}
 
+	var trustBundle []byte
+	certPool := x509.NewCertPool()
+	// this is the generic trusted bundle for things like self-signed registries.
 	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config", "user-ca-bundle", "ca-bundle.crt")
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
-	spec.AdditionalTrustBundle = additionalTrustBundle
+	if len(additionalTrustBundle) > 0 {
+		if !certPool.AppendCertsFromPEM(additionalTrustBundle) {
+			return fmt.Errorf("configmap %s/%s doesn't have a valid PEM bundle", "openshift-config", "user-ca-bundle")
+		}
+		trustBundle = append(trustBundle, additionalTrustBundle...)
+	}
+
+	// this is the trusted bundle specific for proxy things and can differ from the generic one above.
+	if proxy != nil && proxy.Spec.TrustedCA.Name != "" && proxy.Spec.TrustedCA.Name != "user-ca-bundle" {
+		proxyTrustBundle, err := optr.getCAsFromConfigMap("openshift-config", proxy.Spec.TrustedCA.Name, "ca-bundle.crt")
+		if err != nil {
+			return err
+		}
+		if len(proxyTrustBundle) > 0 {
+			if !certPool.AppendCertsFromPEM(proxyTrustBundle) {
+				return fmt.Errorf("configmap %s/%s doesn't have a valid PEM bundle", "openshift-config", proxy.Spec.TrustedCA.Name)
+			}
+			trustBundle = append(trustBundle, proxyTrustBundle...)
+		}
+	}
+	spec.AdditionalTrustBundle = trustBundle
 
 	// if the cloudConfig is set in infra read the cloud config reference
 	if infra.Spec.CloudConfig.Name != "" {
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -155,7 +155,7 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		return err
 	}
 
-	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
+	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config", "user-ca-bundle", "ca-bundle.crt")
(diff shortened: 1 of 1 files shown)


pkg/operator/sync.go
```diff
@@ -155,7 +155,7 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		return err
 	}
 
-	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
+	additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config", "user-ca-bundle", "ca-bundle.crt")
```
bparees: when is this flow being executed?  is this a one time "during install" flow, in which case it is fine?  or is it a steady state flow, in which case it should be looking at the proxy config resource to determine the name of the configmap.

runcom: this is during an MCO sync so it's post-install

bparees: yeah, so if you're going to be periodically syncing this information, you need to be looking at the proxy config resource to determine the location(name) of the CA configmap.

The MCO is combining this content w/ the local system trusted from the node?

runcom: > The MCO is combining this content w/ the local system trusted from the node?

yes exactly, so you're saying this should be (using the `proxy` status from above):

```
proxy.ProxySpec.TrustedCA.Name
```
?

runcom:
```
diff --git a/pkg/operator/sync.go b/pkg/operator/sync.go
index 14b31ad2..608c2514 100644
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -155,9 +155,13 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
                return err
        }
 
-       additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
-       if err != nil && !apierrors.IsNotFound(err) {
-               return err
+       var additionalTrustBundle []byte
+       if proxy != nil && proxy.Spec.TrustedCA.Name != "" {
+               var err error
+               additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
+               if err != nil && !apierrors.IsNotFound(err) {
+                       return err
+               }
        }
        spec.AdditionalTrustBundle = additionalTrustBundle
 
```

runcom:
```diff

diff --git a/pkg/operator/sync.go b/pkg/operator/sync.go
index 14b31ad2..b8148e9c 100644
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -155,9 +155,13 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
                return err
        }
 
-       additionalTrustBundle, err := optr.getCAsFromConfigMap("openshift-config-managed", "user-ca-bundle", "ca-bundle.crt")
-       if err != nil && !apierrors.IsNotFound(err) {
-               return err
+       var additionalTrustBundle []byte
+       if proxy != nil && proxy.Spec.TrustedCA.Name != "" {
+               var err error
+               additionalTrustBundle, err = optr.getCAsFromConfigMap("openshift-config-managed", proxy.Spec.TrustedCA.Name, "ca-bundle.crt")
+               if err != nil && !apierrors.IsNotFound(err) {
+                       return err
+               }
        }
        spec.AdditionalTrustBundle = additionalTrustBundle
```

runcom: I thought it was always `user-ca-bundle` as defined by the installer

bparees: The installer, if it creates it, will create it as that.  The user can create it post install(in the case that they didn't initially need CAs but now they do), or change the name post-install.  In which case they can pick any name they want.

runcom: gotcha, thanks for the explaination @bparees - I'll create a PR with the diff above

runcom: I guess my confusion from using the proxy.trusteCA cm reference was that additionaltrustBundle could have meant a general way to plug other CAs, not only for proxies

bparees: there is almost certainly going to be another turn of the crank on this (post 4.2) where we extend our CA support to be that general way to plug in CAs.  Which may entail another api+configmap name, or another api but sharing a configmap, etc.  But for now the only way we have to identify the CAs the user wants included, is by them specifying it in `proxy.spec.trustbundle`

pkg/operator/sync.go
```diff
@@ -155,11 +155,34 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
-	spec.AdditionalTrustBundle = additionalTrustBundle
+	if len(additionalTrustBundle) > 0 {
+		if !certPool.AppendCertsFromPEM(additionalTrustBundle) {
+			return fmt.Errorf("configmap %s/%s doesn't have a valid PEM bundle", "openshift-config", "user-ca-bundle")
+		}
+		trustBundle = append(trustBundle, additionalTrustBundle...)
+	}
+
+	// this is the trusted bundle specific for proxy things and can differ from the generic one above.
+	if proxy.Spec.TrustedCA.Name != "" {
```
runcom: should check for nil

pkg/operator/sync.go
runcom: ack, I think Abhinav modified this patch to account for all of this also

runcom:
/approve

@abhinavdahiya do we need some other ack from someone or testing?

abhinavdahiya:
> /approve
> 
> @abhinavdahiya do we need some other ack from someone or testing?

Not really.. should be fine to merge.

Issue #1046, opened 2019-08-07
Fix byo machine-config-daemon-host.service


cgwalters:
See https://github.com/openshift/openshift-ansible/pull/11801
We probably want `machine-config-daemon-host.service` to be `ConditionPathExists=/run/ostree-booted` but I haven't thought about it a lot yet.

cgwalters:
PR in https://github.com/openshift/machine-config-operator/pull/1047

PR #1047, opened 2019-08-07
templates: Don't run machine-config-daemon-host on BYO RHEL

cgwalters:
Implicitly this was happening before because the `pivot` package
contained the systemd unit, and so there was nothing to react
to the MCS injecting `/etc/pivot/image-pullspec`.  Now that the MCD
injects this systemd unit, we need to ensure it's only run
on ostree-managed systems, not BYO RHEL.

Closes: https://github.com/openshift/machine-config-operator/issues/1046

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-host.service
+++ b/templates/common/_base/units/machine-config-daemon-host.service
@@ -3,6 +3,9 @@ enabled: true
 contents: |
   [Unit]
   Description=Machine Config Daemon Initial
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
> Now that the MCD
> injects this systemd unit, we need to ensure it's only run
> on ostree-managed systems, not BYO RHEL.

Ooh, this makes those errors make sense to me now..

/lgtm

PR #1049, opened 2019-08-08
Bug 1738831: pkg/controller/node: resync all pools for a node

runcom:
When a node belongs to more than one pool, we need to trigger a resync
for all the pools it belongs to or we leave the "parent" MCP at an old
state. The easy example would be to create an infra pool of just 1
node and describe the worker mcp and notice that it still has all nodes
(including the now-infra node).


runcom:
/hold

works but there's a gotcha exposed from unit tests

PR #1050, opened 2019-08-08
openstack: keepalived to pass cluster-config.yml to runtimecfg

mandre:
This pulls in fix
https://github.com/openshift/machine-config-operator/pull/1043 for
openstack.

A recent installer change started using a loopback address in
kubeconfig as the default on the bootstrap node. baremetal-runtimecfg
will begin using cluster-config.yml if it exists first, in order to get
the domain and name of the cluster.

This patch adds a flag for this to the keepalived bootstrap pod in
order to utilize it. This should only affect the bootstrap pods.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/openstack/keepalived.yaml
+++ b/manifests/openstack/keepalived.yaml
@@ -18,6 +18,9 @@ spec:
       path: "/etc/kubernetes/kubeconfig"
   - name: conf-dir
     empty-dir: {}
+  - name: manifests
+    hostPath:
+      path: "/opt/openshift/manifests"
   initContainers:
   - name: render-config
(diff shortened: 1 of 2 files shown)


tomassedovic:
/assign @kikisdeliveryservice

Please take a look, this mirrors a change already merged for baremetal and it fixes a deployment & CI issue we're seeing with OpenStack.

mandre:
Hi @kikisdeliveryservice, this PR unblocks openshift/installer#1959, meaning we likely won't need https://github.com/openshift/machine-config-operator/pull/1040 and can simply revert https://github.com/openshift/machine-config-operator/commit/759702e7bd490099065405cb90742a2a5ce6d2fc as was the original plan.

PR #1051, opened 2019-08-08
FF master-4.3 branch

runcom:
As per email from Clayton, bringing changes via a PR is the current flow.

cc @sgreene570


cgwalters:
Will this really fast forward though?  Won't it generate a merge commit from the PR, or does prow have special support for this somehow?

That's not bad really, we already have a ton of merge commits - just noting that in the end we're going to have merges both ways when we merge back to master.

cgwalters:
/assign sgreene570
Anyways, I'll believe you this is syncing master :smile: , giving to sgreene for lgtm.

runcom:
> Will this really fast forward though? Won't it generate a merge commit from the PR, or does prow have special support for this somehow?
> 
> That's not bad really, we already have a ton of merge commits - just noting that in the end we're going to have merges both ways when we merge back to master.

yeah, thought about that as well - maybe I'll follow up with Clayton/Steve to understand that

runcom:
> Anyways, I'll believe you this is syncing master

I believe Clayton :smile:

PR #1055, opened 2019-08-12
[release-4.1] Bug 1740147: DR: use observed directories vs assumptions for backup_etcd_client_certs

hexfusion:
This is a cherry-pick of #1052


derekwaynecarr:
#1052 is not yet VERIFIED in master.

derekwaynecarr:
The referenced bug has not had its dependent bug verified yet in master.  Updated the bz in order to not trip up our tooling.

/bugzilla refresh

PR #832, opened 2019-06-06
pkg/operator Filter out non-MCO CRDs when re-syncing modified MCO CRDs

sgreene570:
Fixes #815 

**- What I did**
Following #817, added a label for `mco-built-in/crd` to MCO CRDs in the `manifests/` directory.  
Then modified `apiExtSharedInformer` to filter based on this label.
Tested on an AWS cluster to be working.

**- How to verify it**
Add logging statements to the `enqueue` function in `pkg/operator/operator.go` and watch `machine-config-operator` pod logs in kube and verify that non-MCO CRDs are not being re-synced (when they don't need to be, as seen in the issue, #815 ).

**- Description for the changelog**

Filter out object changes not pertaining to MCO when re-syncing MCO CRDs.

PR #1058, opened 2019-08-13
[4.3] pkg/daemon: MCD constants fixup

sgreene570:
**- What I did**
Add the `"openshift.io/operator-managed"` label from #832 to the MCD global constants file.
This way the operator managed label can be easily swapped out if the need arises (once #817 is merged the label will be used more than once, and it is an important label). 

I also cleaned up some small typos in the constants file.

**- How to verify it**
CI


kikisdeliveryservice:
since it's so small i think you can squash commits 1 & 3

kikisdeliveryservice:
just adding a hold for now since this isn't a bugfix.  thanks for this!

/hold

kikisdeliveryservice:
these retests are flooding my inbox.
until ci straightens itself out
/hold

PR #1060, opened 2019-08-14
Bug 1712826: DR: Add pre-run checks for backup and restore scripts

alaypatel07:
Currently, the DR scripts do not check if parameters are valid. 

**- What I did**
Added checks for valid params to backup and restore scripts.

Fixes: [BZ 1712826](https://bugzilla.redhat.com/show_bug.cgi?id=1712826)

/cc @hexfusion  @retroflexer

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
@@ -16,11 +16,17 @@ contents:
       exit 1
     fi
 
+    usage () {
+        echo 'Path to backup file required: ./etcd-snapshot-backup.sh ./backup.db'
+        exit 1
+    }
+
     ASSET_DIR=./assets
-    SNAPSHOT_FILE="${ASSET_DIR}/backup/etcd/member/snap/db"
 
     if [ "$1" != "" ]; then
       SNAPSHOT_FILE="$1"
+    else
+      usage
     fi
 
     CONFIG_FILE_DIR=/etc/kubernetes
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -40,6 +40,11 @@ contents:
     ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
     STOPPED_STATIC_PODS="${ASSET_DIR}/tmp/stopped-static-pods"
 
+    if [ ! -f "$SNAPSHOT_FILE" ]; then
+      echo "etcd snapshot $SNAPSHOT_FILE does not exist."
+      exit 1
+    fi
+
     source "/usr/local/bin/openshift-recovery-tools"
 
     function run {


templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
```diff
@@ -25,6 +25,11 @@ contents:
         usage
     fi
 
+    if [ ! -f "$1" ]; then
+      echo "Backup file $1 file does not existBackup file. Please try again "
```
hexfusion: can we clarify this message a little?

`etcd snapshot $SNAPSHOT_FILE does not exist.`

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
```diff
@@ -16,11 +16,18 @@ contents:
       exit 1
     fi
 
+    usage () {
+        echo 'Path to backup file needed: ./etcd-snapshot-backup.sh ./backup.db'
```
hexfusion: s/needed/required?

kikisdeliveryservice:
Not sure why we got this error:
```go: mvdan.cc/unparam@v0.0.0-20190124213536-fbb59629db34: unrecognized import path "mvdan.cc/unparam" (https fetch: Get https://mvdan.cc/unparam?go-get=1: dial tcp [2a03:b0c0:1:d0::64d:d001]:443: connect: network is unreachable)```

/test verify

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
kikisdeliveryservice: +1

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
kikisdeliveryservice: +1

alaypatel07: +1

retroflexer: I would like the message to include the string "Usage:" somewhere. For example.,
```echo 'Path to backup file is needed.\n Usage: ./etcd-snapshot-backup.sh <path-to-backup-file>'```

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
hexfusion: since we capture $1 as SNAPSHOT_FILE I think it more readable to check that in you if block. WDYT?

hexfusion:
If we are going to require a path can we remove the default?

https://github.com/openshift/machine-config-operator/blob/06df20fc6062899faf8bc21d7002a449e24a0ecf/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml#L25

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
alaypatel07: +1 I think it will be more readable. I will move the check after the variables are defined.

PR #1063, opened 2019-08-14
Bug 1735661: add template with default inotify.max_users settings

kikisdeliveryservice:
add template to create /etc/sysctl.d/inotify.conf with inotify.max_users defaults taken from: 
https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/tuned/templates/openshift-node/tuned.conf

Closes: BZ 1735661

### DIFF
--- a/templates/common/_base/files/sysctl-inotify.conf.yaml
+++ b/templates/common/_base/files/sysctl-inotify.conf.yaml
@@ -1,0 +1,8 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/sysctl.d/inotify.conf"
+contents:
+  inline: |
+ 
+    fs.inotify.max_user_watches = 65536
+    fs.inotify.max_user_instances = 8192


abhinavdahiya:
If we want to update the sysctls.. should we put it in the same directory as https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/sysctl-forward-conf.yaml ??

templates/common/_base/files/tuned-default-conf.yaml
```diff
@@ -0,0 +1,13 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/tuned.conf"
+contents:
+  inline: |
+    [main]
+    summary=Optimize systems running OpenShift nodes
+    include=openshift
+
+    [sysctl]
```
LorbusChris: do we want to use the sysctl plugin of tuned here, or configure sysctl directly?

templates/common/_base/files/tuned-default-conf.yaml
```diff
@@ -0,0 +1,13 @@
...
+    net.ipv4.tcp_fastopen=3
+    fs.inotify.max_user_watches= 65536
+    fs.inotify.max_user_instances= 8192
```
LorbusChris: are the spaces after = wanted/needed here?

kikisdeliveryservice:
Updated.  Talked to Clayton and am going to check to see if the `net.ipv4.tcp_fastopen=3` that was also set in 3.11 needs to be set as default in MCO or if this already exists elsewhere.

kikisdeliveryservice:
ok i think i got the BZ sorted out:

/bugzilla refresh

kikisdeliveryservice:
@abhinavdahiya & @LorbusChris PTAL

templates/common/_base/files/sysctl-inotify.conf.yaml
```diff
@@ -0,0 +1,8 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/sysctl.d/inotify.conf"
```
cgwalters: Nothing here is templated; which means we could ship it by default in RHCOS (in `/usr/lib/sysctl.d`).

This is an interesting topic - one thing to note is that changes made via the MCO only apply when booted as part of a cluster.  So if I just want to e.g. `cosa run` to test the base OS, I'm testing without these types of tweaks.

In this case, maybe that is right - we only want to bump limits once a node has joined the cluster and it's no longer "standalone"?

Anyways for now I'd at least call this `/etc/sysctl.d/rhcos-inotify.conf` to make it clearer where it came from.

cgwalters: Although, the other `/etc/sysctl.d/forward.conf` is also generically-named.  So we could also just try to make a cleanup pass on this stuff later and add things like comments to the files describing why they're doing the changes, etc.

kikisdeliveryservice: @cgwalters i tried to be consistent with the other conf, so a full cleanup later with comments, etc.. makes sense to me so things stay consistent.

PR #1064, opened 2019-08-15
Bug 1744764: Report degraded reason in cluster operator status

LorbusChris:
Needs cherrypicking for 4.1.z, in order to fix BZ 1741065: 

**- What I did**
Add reason for degradation to ClusterOperatorStatusCondition

**- How to verify it**
CI

pkg/operator: Report degraded reason in cluster operator status

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -132,11 +132,11 @@ func (optr *Operator) updateStatus(co *configv1.ClusterOperator, status configv1
 }
 
 const (
-	failedToSyncReason = "FailedToSync"
+	asExpectedReason = "AsExpected"
 )
 
 // syncDegradedStatus applies the new condition to the mco's ClusterOperator object.
-func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
+func (optr *Operator) syncDegradedStatus(ierr syncError) (err error) {
 	co, err := optr.fetchClusterOperator()
 	if err != nil {
 		return err
@@ -148,22 +148,30 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
 	optrVersion, _ := optr.vStore.Get("operator")
 	degraded := configv1.ConditionTrue
 	var message, reason string
-	if ierr == nil {
+	if ierr.err == nil {
 		degraded = configv1.ConditionFalse
 	} else {
 		if optr.vStore.Equal(co.Status.Versions) {
 			// syncing the state to exiting version.
-			message = fmt.Sprintf("Failed to resync %s because: %v", optrVersion, ierr.Error())
+			message = fmt.Sprintf("Failed to resync %s because: %v", optrVersion, ierr.err.Error())
 		} else {
-			message = fmt.Sprintf("Unable to apply %s: %v", optrVersion, ierr.Error())
+			message = fmt.Sprintf("Unable to apply %s: %v", optrVersion, ierr.err.Error())
 		}
-		reason = failedToSyncReason
+		reason = ierr.task + "Failed"
 
 		// set progressing
 		if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) {
-			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Message: fmt.Sprintf("Unable to apply %s", optrVersion)})
+			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorProgressing,
+				Status:  configv1.ConditionTrue,
+				Message: fmt.Sprintf("Unable to apply %s", optrVersion),
+			})
 		} else {
-			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse, Message: fmt.Sprintf("Error while reconciling %s", optrVersion)})
+			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorProgressing,
+				Status:  configv1.ConditionFalse,
+				Message: fmt.Sprintf("Error while reconciling %s", optrVersion),
+			})
 		}
 	}
 
@@ -193,7 +201,7 @@ func (optr *Operator) syncUpgradeableStatus() error {
 	coStatus := configv1.ClusterOperatorStatusCondition{
 		Type:   configv1.OperatorUpgradeable,
 		Status: configv1.ConditionTrue,
-		Reason: "AsExpected",
+		Reason: asExpectedReason,
 	}
 	return optr.updateStatus(co, coStatus)
 }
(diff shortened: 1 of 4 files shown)


LorbusChris:
it's ok @openshift-ci-robot, I'll fix.

/cherrypick release-4.1

kikisdeliveryservice:
Oh nooo.. Bc whoever created the MCO BZ cloned it it's now dependent on all of the other cloned BZs that person opened??? Gaaah.

kikisdeliveryservice:
Updated the BZ to request that the incorrectly set Depends on/Blocks fields get cleared out so we can get the proper labels on this PR to merge.

pkg/operator/status.go
```diff
@@ -157,13 +160,23 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
 		} else {
 			message = fmt.Sprintf("Unable to apply %s: %v", optrVersion, ierr.Error())
 		}
-		reason = failedToSyncReason
+		reason = FailedToSyncReason
 
 		// set progressing
 		if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) {
-			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Message: fmt.Sprintf("Unable to apply %s", optrVersion)})
+			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
```
sgreene570: nice formatting fix :pray:

pkg/operator/status.go
```diff
@@ -132,7 +132,10 @@ func (optr *Operator) updateStatus(co *configv1.ClusterOperator, status configv1
 }
 
 const (
-	failedToSyncReason = "FailedToSync"
+	// FailedToSyncReason is the the reason string for the degraded status condition
+	FailedToSyncReason = "FailedToSync"
+	// AsExpectedReason is the reason string for the upgradable status condition
+	AsExpectedReason = "AsExpected"
```
runcom: any reason to export them if they're not used elsewhere?

LorbusChris:
@runcom I've created  BZ1741914 for 4.2, can you set the target version so the bot picks it up correctly?

pkg/operator/status.go
```diff
@@ -161,9 +162,19 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
 
 		// set progressing
 		if cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorProgressing) {
-			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Message: fmt.Sprintf("Unable to apply %s", optrVersion)})
+			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorProgressing,
+				Status:  configv1.ConditionTrue,
+				Message: fmt.Sprintf("Unable to apply %s", optrVersion),
+				Reason:  reason,
```
runcom: This does look good but I think we need to do a better job into at least saying which of the sync function failed

runcom: I'll open a follow up

LorbusChris: addressing this here directly

pkg/operator/sync.go
```diff
@@ -39,24 +39,35 @@ type syncFunc struct {
 	fn   func(config *renderConfig) error
 }
 
+type syncErr struct {
+	task string
+	error
```
runcom: let's call this `err error` - I'd avoid types embedding as much as we can

pkg/operator/sync.go
```diff
@@ -39,24 +39,35 @@ type syncFunc struct {
...
+}
+
 func (optr *Operator) syncAll(syncFuncs []syncFunc) error {
 	if err := optr.syncProgressingStatus(); err != nil {
-		return fmt.Errorf("error syncing progressing status: %v", err)
+		return syncErr{
+			task:  "ProgressingStatus",
+			error: fmt.Errorf("error syncing progressing status: %v", err),
+		}
```
runcom: is this bubbled up in the degraded? I think this just just return the error it was returning before

pkg/operator/status.go
```diff
@@ -148,22 +148,33 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
...
+		reason = ""
 	} else {
 		if optr.vStore.Equal(co.Status.Versions) {
 			// syncing the state to exiting version.
 			message = fmt.Sprintf("Failed to resync %s because: %v", optrVersion, ierr.Error())
 		} else {
 			message = fmt.Sprintf("Unable to apply %s: %v", optrVersion, ierr.Error())
 		}
-		reason = failedToSyncReason
+		reason = ierr.task + "SyncError"
```
runcom: ok sorry I said `SyncError` but I think `Failed` would be better and more aligned with other telemtry coming from other opererators (sorry for the back and forth)

pkg/operator/status.go
```diff
@@ -148,22 +148,33 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
 	optrVersion, _ := optr.vStore.Get("operator")
 	degraded := configv1.ConditionTrue
 	var message, reason string
-	if ierr == nil {
+	if ierr.error == nil {
 		degraded = configv1.ConditionFalse
+		reason = ""
```
runcom: not needed, `reason` remains `""` if it goes into this `if`

pkg/operator/sync.go
LorbusChris: people in Milan own style, don't they :D

pkg/operator/sync.go
LorbusChris: oh right. forgot to rm this one again

pkg/operator/status.go
```diff
@@ -148,22 +148,33 @@ func (optr *Operator) syncDegradedStatus(ierr error) (err error) {
...
+				Message: message,
+				Reason:  reason,
+			})
 		} else {
-			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse, Message: fmt.Sprintf("Error while reconciling %s", optrVersion)})
+			cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorProgressing,
+				Status:  configv1.ConditionFalse,
+				Message: message,
+				Reason:  reason,
```
runcom: wondering if it's right to report the Degraded message and reason here for progressing as well

runcom: I'd avoid this change if not strictly required for the linked BZ, we can come back to this later on

pkg/operator/sync.go
runcom: fantasy around naming above all lmao

pkg/operator/status.go
LorbusChris: we set the message here in the past. I replaced `fmt.Sprintf("Error while reconciling %s", optrVersion)}` with the message set above. Want me to remove message and reason here entirely?

runcom: nope, just rollback to what we had, so that this PR has a clear scope on the BZ

PR #1066, opened 2019-08-19
Bug 1743361: fix updating version output

kikisdeliveryservice:
Problem: output of  `oc get co machine-config -o yaml` is pointing to current config name: 
```worker: 0 (ready 0) out of 2 nodes are updating to latest configuration rendered-worker-2fe2c50ca1b300646e59b6fad007d537``` 
instead of the name of the desired/target config that is being applied.

Through some testing/adding additional logs I verified the target config in Spec was set properly at the correct time.  The issue was status.go was improperly referencing `pool.Status.Configuration.Name`  (current config) instead of `pool.Spec.Configuration.Name` in its output.

Closes: BZ 1743361

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -335,7 +335,7 @@ func machineConfigPoolStatus(pool *mcfgv1.MachineConfigPool) string {
 	case mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdated):
 		return fmt.Sprintf("all %d nodes are at latest configuration %s", pool.Status.MachineCount, pool.Status.Configuration.Name)
 	case mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating):
-		return fmt.Sprintf("%d (ready %d) out of %d nodes are updating to latest configuration %s", pool.Status.UpdatedMachineCount, pool.Status.ReadyMachineCount, pool.Status.MachineCount, pool.Status.Configuration.Name)
+		return fmt.Sprintf("%d (ready %d) out of %d nodes are updating to latest configuration %s", pool.Status.UpdatedMachineCount, pool.Status.ReadyMachineCount, pool.Status.MachineCount, pool.Spec.Configuration.Name)
 	default:
 		return "<unknown>"
 	}
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -513,7 +513,7 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 	if err != nil {
 		return err
 	}
-	glog.V(2).Infof("Pool %s: now targeting: %s", pool.Name, generated.Name)
+	glog.V(2).Infof("Pool %s: now targeting: %s", pool.Name, pool.Spec.Configuration.Name)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by runcom** [comment]
> I believe this is bc the code is currently using `pool.Spec.Configuration.Name` which doesn't actually change the poolwide config nameot the new config until the update is completed. I think `pool.Status.Configuration.Name` should have the new config name.

it's actually the opposite, the spec maintains the desired state and status doesn't change till spec becomes current state. Maybe we're not setting the spec to desired or no sync runs in MCC and the mcp doesn't yet know about that.

**Comment by kikisdeliveryservice** [comment]
> it's actually the opposite, the spec maintains the desired state and status doesn't change till spec becomes current state. Maybe we're not setting the spec to desired or no sync runs in MCC and the mcp doesn't yet know about that.

ahh you're right.. 	https://github.com/openshift/machine-config-operator/blob/1759c3527d64c80b4792384df43f86b87ab5d9d8/pkg/controller/node/node_controller.go#L788

so yeah then this isn't being updated until the end? will look more.. thanks!

**Comment by kikisdeliveryservice** [comment]
update, I got the version fixed:
```
99-worker-test1                                                                                        2.2.0             4m26s
rendered-master-962c91db7a1ddeb685b5a632883df72c            26e62ecb46ab4c23d69c91324a3ea89bb44d69fb   2.2.0             19m
rendered-worker-27622c5cd8d8863e78630d3198a2de43            26e62ecb46ab4c23d69c91324a3ea89bb44d69fb   2.2.0             19m
rendered-worker-ba38166438eec49d0b1ad762e3cecd09            26e62ecb46ab4c23d69c91324a3ea89bb44d69fb   2.2.0             4m20s
```
and

```
extension:
    master: all 3 nodes are at latest configuration rendered-master-962c91db7a1ddeb685b5a632883df72c
    worker: 0 (ready 0) out of 3 nodes are updating to latest configuration rendered-worker-ba38166438eec49d0b1ad762e3cecd09
```

File: pkg/operator/status.go [hunk 1]
```diff
@@ -335,7 +335,7 @@ func machineConfigPoolStatus(pool *mcfgv1.MachineConfigPool) string {
 	case mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdated):
 		return fmt.Sprintf("all %d nodes are at latest configuration %s", pool.Status.MachineCount, pool.Status.Configuration.Name)
 	case mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating):
-		return fmt.Sprintf("%d (ready %d) out of %d nodes are updating to latest configuration %s", pool.Status.UpdatedMachineCount, pool.Status.ReadyMachineCount, pool.Status.MachineCount, pool.Status.Configuration.Name)
+		return fmt.Sprintf("%d (ready %d) out of %d nodes are updating to latest configuration %s", pool.Status.UpdatedMachineCount, pool.Status.ReadyMachineCount, pool.Status.MachineCount, pool.Spec.Configuration.Name)
```
runcom: oh nice so this was it right? :+1:

kikisdeliveryservice: yep

cgwalters:
Oops, yeah I messed that up!
/lgtm

kikisdeliveryservice:
e2e-aws-upgrade is going to fail until this merges: https://github.com/openshift/insights-operator/pull/7

PR #1069, opened 2019-08-20
Bug 1743487: Do not requeue a failed kubeletconfig decode

rphillips:
**- What I did**
Add a forgettable error type that is caught by the error handler to not requeue the key.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1743487

**- How to verify it**

```NONE```

### DIFF
--- a/pkg/controller/kubelet-config/errors.go
+++ b/pkg/controller/kubelet-config/errors.go
@@ -1,0 +1,13 @@
+package kubeletconfig
+
+type forgetError struct {
+	Err error
+}
+
+func newForgetError(err error) *forgetError {
+	return &forgetError{Err: err}
+}
+
+func (e *forgetError) Error() string {
+	return e.Err.Error()
+}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controller/kubelet-config/errors.go [hunk 1]
```diff
@@ -0,0 +1,13 @@
+package kubeletconfig
+
+type forgetError struct {
+	Err error
+}
+
+func NewForgetError(err error) *forgetError {
```
runcom: can you unexport this?

runcom:
@umohnani8 can you check if container runtime config has the same issue?

pkg/controller/kubelet-config/errors.go
rphillips: fixed! Thanks!

runcom:
I'm torn on this - if we forget the key, are we sure the error sticks for people to take action? otherwise this is good

rphillips:
@runcom yes... the error is pushed into the status field for a user action.

runcom:
> @runcom yes... the error is pushed into the status field for a user action. syncStatusOnly is run prior to the handleErr function.

alrighty, and that's not cleared by successive syncs (right?) :smile: 

/approve

rphillips:
@runcom that is correct

umohnani8:
@runcom will check and fix up if that is the case.

PR #705, opened 2019-05-06
Work around bz1571183

RobertKrawitz:
`curl` results in heavy dentry cache pollution that gets charged to the eqg's cgroup.  The underlying bug is https://bugzilla.redhat.com/show_bug.cgi?id=1571183.

Fixes 1706625.

PR #1070, opened 2019-08-20
Bug 1743768: avoid excessive dentries due to keepalived static pod health check

yboaron:
The 'curl' commands on the keepalived static pod cause a huge amount of memory dentries.
The underlying bug is https://bugzilla.redhat.com/show_bug.cgi?id=1571183.
This commit work-around this problem. 
For more details see:
https://github.com/openshift/machine-config-operator/pull/705
https://github.com/openshift/openshift-ansible/pull/11829

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/baremetal/keepalived.yaml
+++ b/manifests/baremetal/keepalived.yaml
@@ -55,6 +55,9 @@ spec:
     securityContext:
       privileged: true
     image: {{ .Images.KeepalivedBootstrap }}
+    env:
+      - name: NSS_SDB_USE_CACHE
+        value: "no"
     command:
     - /usr/sbin/keepalived
     args:
--- a/templates/common/openstack/files/openstack-keepalived.yaml
+++ b/templates/common/openstack/files/openstack-keepalived.yaml
@@ -53,6 +53,9 @@ contents:
         securityContext:
           privileged: true
         image: {{.Images.keepalivedImage}}
+        env:
+          - name: NSS_SDB_USE_CACHE
+            value: "no"
         command:
         - /usr/sbin/keepalived
         args:


runcom:
oh ok, this is not fixing 1571183, we need a valid bz for 4.2 to land this, can you create one @yboaron and link it here?

runcom:
@yboaron you need to run `make update` also
/approve

would love someone from Baremetal/OpenStack to lgtm this also after verification

bcrochet:
@yboaron You need to re-generate the bindata. 'make update'

manifests/baremetal/keepalived.yaml
```diff
@@ -56,14 +56,11 @@ spec:
...
-    - "-f"
-    - "/etc/keepalived/keepalived.conf"
-    - "--dont-fork"
-    - "--vrrp"
-    - "--log-detail"
-    - "--log-console"
+    - /bin/sh
+    - -c
+    - |
+       export NSS_SDB_USE_CACHE=no
```
cgwalters: Would it be simpler to
```
    env:
    - name: NSS_SDB_USE_CACHE
      value: "no"
```
?

celebdor: Indeed. Please try that, @yboaron

yboaron: @cgwalters , Yep, looks like a nicer way to set this variable. I'll verify it on my env and will update the PR. 10x

runcom:
can we just close this PR as it's effectively a duplicate of https://github.com/openshift/machine-config-operator/pull/1083 now?

celebdor:
Yes. We can indeed.

PR #1072, opened 2019-08-21
Bug 1744133: Openstack - copy Haproxy container log fix from Baremetal

yboaron:
After haproxy container is recreated (for example due to liveness probe failure),
the haproxy events don't log anymore in container's logs, and we hit [1]

This commit copy the fix from baremetal platform.

[1]- + socat UNIX-RECV:/var/run/haproxy/haproxy-log.sock STDOUT
2019/08/14 07:21:03 socat[8] E "/var/run/haproxy/haproxy-log.sock" exists

**- What I did**

**- How to verify it**


runcom:
:+1:

PR #1073, opened 2019-08-21
Bug 1734554: provide etcd-member-add.sh for adding back a member with valid certs

retroflexer:
Fixes: #1734554
The current etcd DR scripts are intended to be used when quorum is lost. However, in simpler cases, an etcd member may fail to restart, but the certificates are valid. In that case, The solution is to remove the member from the cluster, and re-add using etcd-member-add.sh. This script can be used in conjunction with etcd-member-remove.sh.

**- What I did**
Created a new DR script etcd-member-add.sh which allows users to easily add back an etcd member when the certs are valid.

**- How to verify it**
1.) Create a new cluster
2.) remove a member using the etcd-member-remove.sh script.
3.) add back the member using this script
3.) verify with etcdctl member list

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
@@ -1,0 +1,62 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/etcd-member-add.sh"
+contents:
+  inline: |
+    #!/usr/bin/env bash
+    
+    # This script should be invoked only if the member already has valid certs.
+    # example
+    # sudo -E ./etcd-member-add.sh 192.168.1.100 $etcd_name
+    
+    if [[ $EUID -ne 0 ]]; then
+      echo "This script must be run as root"
+      exit 1
+    fi
+    
+    usage () {
+        echo 'Active server IP address and etcd name required: ./etcd-member-add.sh 192.168.1.100 $etcd_name'
+        exit 1
+    }
+    
+    if [ "$1" == "" ] || [ "$2" == "" ]; then
+        usage
+    fi
+    
+    RECOVERY_SERVER_IP=$1
+    ETCD_NAME=$2
+    
+    ASSET_DIR=./assets
+    CONFIG_FILE_DIR=/etc/kubernetes
+    MANIFEST_DIR="${CONFIG_FILE_DIR}/manifests"
+    RUN_ENV=/run/etcd/environment
+    
+    MANIFEST_STOPPED_DIR="$ASSET_DIR/manifests-stopped"
+    ETCD_MANIFEST="${MANIFEST_DIR}/etcd-member.yaml"
+    ETCD_CONFIG=/etc/etcd/etcd.conf
+    ETCDCTL=$ASSET_DIR/bin/etcdctl
+    ETCD_VERSION=v3.3.10
+    ETCD_DATA_DIR=/var/lib/etcd
+    ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
+    
+    source "/usr/local/bin/openshift-recovery-tools"
+    
+    function run {
+      init
+      dl_etcdctl
+      backup_manifest
+      DISCOVERY_DOMAIN=$(grep -oP '(?<=discovery-srv=).*[^"]' $ASSET_DIR/backup/etcd-member.yaml )
+      if [ -z "$DISCOVERY_DOMAIN" ]; then
+        echo "Discovery domain can not be extracted from $ASSET_DIR/backup/etcd-member.yaml"
+        exit 1
+      fi
+      validate_environment
+      source  /run/etcd/environment
+      backup_etcd_conf
+      stop_etcd
+      backup_data_dir
+      etcd_member_add
+      start_etcd
+    }
+    
+    run


templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
```diff
@@ -0,0 +1,72 @@
...
+      backup_manifest
+      DISCOVERY_DOMAIN=$(grep -oP '(?<=discovery-srv=).*[^"]' $ASSET_DIR/backup/etcd-member.yaml )
+      if [ -z "$DISCOVERY_DOMAIN" ]; then
+        echo "Discovery domain can not be extracted from $ASSET_DIR/backup/etcd-member.yaml"
+        exit 1
+      fi
+      validate_environment
+      source  /run/etcd/environment
+      backup_etcd_conf
+      backup_etcd_client_certs
```
retroflexer: We probably don't need to back certs at all, as we are not touching. Perhaps a test to validate the certs would, instead, be a good idea.

alaypatel07: I was thinking of a test to validate the certs at the very beginning, and if the certs are not valid, exit with an appropriate error message. WDYT?

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
```diff
@@ -0,0 +1,72 @@
...
+    usage () {
+        echo 'Active server IP address and etcd name required: ./etcd-member-add.sh 192.168.1.100 $etcd_name'
+        exit 1
+    }
+    
+    if [ "$1" == "" ] || [ "$2" == "" ]; then
+        usage
+    fi
+    
+    RECOVERY_SERVER_IP=$1
```
hexfusion: `RECOVERY_SERVER_IP` what is it used for?

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
```diff
@@ -0,0 +1,72 @@
...
+    
+    MANIFEST_STOPPED_DIR="$ASSET_DIR/manifests-stopped"
+    ETCD_MANIFEST="${MANIFEST_DIR}/etcd-member.yaml"
+    ETCD_CONFIG=/etc/etcd/etcd.conf
+    ETCDCTL=$ASSET_DIR/bin/etcdctl
+    ETCD_VERSION=v3.3.10
+    ETCD_DATA_DIR=/var/lib/etcd
+    ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
+    
+    SHARED=/usr/local/share/openshift-recovery
```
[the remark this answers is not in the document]
retroflexer: :-). It was used in recover.sh for defining the TEMPLATE. I removed TEMPLATE as it is used in generating the certs, but missed to remove the SHARED.

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
hexfusion: The assumption for the script is the certs are valid. If we can't assume that should they just use the current restore that rolls new certs?

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
retroflexer: We need a good server ip  (ip of the one of the active members) for member add.

retroflexer:
> @retroflexer thanks for the PR! Few nits and in general some of these variables can maybe be removed altogether from other scripts as well. Perhaps we can do that in another PR or feel free to do that here.

Thanks. I will try to get rid of as many of the unwanted variables as possible now. Let me do another check in (at the end I will rebase and squish).

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
hexfusion: +1

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
```diff
@@ -0,0 +1,72 @@
...
+        exit 1
+      fi
+      validate_environment
+      source  /run/etcd/environment
+      backup_etcd_conf
+      backup_etcd_client_certs
+      stop_etcd
+      backup_data_dir
+      CLUSTER_NAME=$(echo ${DISCOVERY_DOMAIN} | grep -oP '^.*?(?=\.)')
+      patch_manifest
```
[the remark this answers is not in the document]
retroflexer: IDK. 'll look into it.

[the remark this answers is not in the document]
retroflexer: \s --discovery_service isn't found in the manifest files of 4.2. So, patch_manifest is not needed.

templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
```diff
@@ -0,0 +1,62 @@
...
+mode: 0755
+path: "/usr/local/bin/etcd-member-add.sh"
+contents:
+  inline: |
+    #!/usr/bin/env bash
+    
+    # This script should be invoked only if the member already has valid certs.
+    # example
+    # sudo -E ./etcd-member-add.sh 192.168.1.100 $etcd_name
+
```
hexfusion: optional nit but as we are no longer passing ENV to script we can drop `-E`

hexfusion:
Overall this looks good I am concerned that folks will use this without truly knowing if certs are valid. But as long as we explicitly document the usage and dependencies I think it's fine. Most admin/support staff has been doing these operations by hand for a long time. So why require many more unneeded steps if not required.

/lgtm

kikisdeliveryservice:
thanks @retroflexer & @hexfusion 

please follow-up with documentations as @hexfusion suggested.

/approve

Issue #1074, opened 2019-08-22
MC updates could cause unexpected results for static pod manifests.


hexfusion:
I am seeing circumstances[1] where when MC updates are made to static pod manifests it appears the original file supersede the new file.  

kubelet logs
```
Aug 14 12:08:11 ip-10-0-144-177 hyperkube[1491]: I0814 12:08:11.923936    1491 file.go:200] Reading config file "/etc/kubernetes/manifests/etcd-member.yaml"
Aug 14 12:08:11 ip-10-0-144-177 hyperkube[1491]: I0814 12:08:11.925976    1491 file.go:200] Reading config file "/etc/kubernetes/manifests/etcd-member.yaml.mcdorig"
```

The change was made here[2], while I understand its usefulness I am surprised that kubelet will read any file in manifest dir apparently regardless of extension. Still digging but wanted to get this in front of maintainers eyes early for input.

[1] https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1056/pull-ci-openshift-machine-config-operator-master-e2e-restore-cluster-state/16
[2] https://github.com/openshift/machine-config-operator/pull/797

/cc @runcom @kikisdeliveryservice

runcom:
> The change was made here[2], while I understand its usefulness I am surprised that kubelet will read any file in manifest dir apparently regardless of extension. Still digging but wanted to get this in front of maintainers eyes early for input.

ouch, guess that's a real bug then, maybe we shouldn't use $PWD for the `.mcdorig` backup file, I'll fix that.

runcom:
@hexfusion do we have a BZ for this?

hexfusion:
> @hexfusion do we have a BZ for this?

@runcom I will verify my claim and file bug if that is the case. I am just going from logs at the moment but will test shortly.

runcom:
> @runcom I will verify my claim and file bug if that is the case. I am just going from logs at the moment but will test shortly.

thanks a lot Sam

hexfusion:
@runcom https://bugzilla.redhat.com/show_bug.cgi?id=1744580

PR #1075, opened 2019-08-22
Bug 1744580: pkg/daemon: create orig file outside of pwd for the file

runcom:
Backing up the original file to its $PWD has a side effect on certain
directories like the Kube static pod dir. If a file is backed up there,
it will be still processed by kube which is completely wrong and
unwanted.
This patch is instead backing up the file inside
/etc/machine-config-daemon/$FPATH avoiding the pitfall above.

Fix https://github.com/openshift/machine-config-operator/issues/1074

Let's debate on the `/etc/machine-config-daemon` location, I don't really like it but I'm accepting other suggestions.

cc @hexfusion 

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -747,8 +747,12 @@ func (dn *Daemon) writeFiles(files []igntypes.File) error {
 	return nil
 }
 
+func origParentDir() string {
+	return filepath.Join("/etc", "machine-config-daemon", "orig")
+}
+
 func origFileName(fpath string) string {
-	return fpath + ".mcdorig"
+	return filepath.Join(origParentDir(), fpath+".mcdorig")
 }
 
 func createOrigFile(fpath string) error {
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh wow...

/approve

Longer term, I think it'd be a lot better to use ostree for this by default, since that way it's all transactional by default...it would mean we'd be less transactional on "classic RHEL" but hey, everyone running that just accepts that your computer is toast if it crashes in the middle of `yum update` anyways, so...

hexfusion:
LGTM!

/cc @abhinavdahiya @vrutkovs

sgreene570:
@runcom can you fix the typo in your commit message? (Or maybe reword it, the title is kinda confusing).

runcom:
> @runcom can you fix the typo in your commit message? (Or maybe reword it, the title is kinda confusing).

oh yeah, just noticed

PR #1076, opened 2019-08-23
Bug 1745240: pkg/operator: fix related objects for must-gather

runcom:
The ControllerConfig name is "machine-config-controller", not "cluster".
That was causing the cc to not being collected with must-gather.


kikisdeliveryservice:
this looks good, just waiting on a BZ so we can merge.

PR #1079, opened 2019-08-23
[4.3] baremetal: run coredns healthcheck on 8083

karmab:
**- What I did**
change the port used for coredns health plugin in baremetal platform

**- How to verify it**
deploy with baremetal and check port 8083

switch to port 8083 for coredns healthcheck on baremetal

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -1,6 +1,6 @@
 . {
     errors
-    health
+    health :8083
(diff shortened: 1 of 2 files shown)


e-minguez:
Related: https://github.com/openshift/ocs-operator/issues/53

runcom:
4.2 or next? if 4.2 it needs a BZ

karmab:
> 4.2 or next? if 4.2 it needs a BZ

next

cybertron:
This change will also need to be made in https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/baremetal-coredns-corefile.yaml and https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/baremetal-coredns.yaml in order to be effective on anything but the bootstrap node.

PR #1080, opened 2019-08-23
Bug 1745016: [release-4.1] add base templates with default inotify.max_user settings

kikisdeliveryservice:
Cherrypicks: #1063
Closes: BZ 1745016

cc: @runcom


kikisdeliveryservice:
i think i need to update the tests with the new templates...

/hold

runcom:
/approve
/lgtm

waiting on the 4.2 BZ to be verified, refresh here and this should be good to go

mfojtik:
Approving as this met all formal criteria for backport process, has valid master/4.2 BZ that is in valid state, this is not a feature and already soaked in master for ~9 days.

/cc @jwforres

PR #1083, opened 2019-08-26
Bug 1743768: avoid excessive dentries due to keepalived static pod health check

bcrochet:
curl commands on the keepalived static pod cause a huge amount of memory dentries.
The underlying bug is https://bugzilla.redhat.com/show_bug.cgi?id=1571183.

For more details see:
https://github.com/openshift/machine-config-operator/pull/705
https://github.com/openshift/openshift-ansible/pull/11829

Supersedes PR #1070

**- What I did**

**- How to verify it**


runcom:
@celebdor ptal

/approve

PR #1084, opened 2019-08-26
ctrcfg: setup OWNERS for code and templates

runcom:
Code and templates is owned by the Node team (was Container Runtime team). Setup some OWNERS file
to make sure they're correctly assigned when something changes there.


abhinavdahiya:
on installer repo we use OWNER_ALIASES for better handling https://github.com/openshift/installer/blob/master/OWNERS_ALIASES might be useful here.

PR #1085, opened 2019-08-26
[release 4.1] Bug 1741065: Report degraded reason in cluster operator status

LorbusChris:
Closes: BZ 1741065

**- What I did**
4.1.z backport of BZ 1741064 (GH https://github.com/openshift/machine-config-operator/pull/1064)
Makes syncAll fail fast like upstream
Add reason for degradation to ClusterOperatorStatusCondition
Updated unit tests

**- How to verify it**
CI

pkg/operator: Report degraded reason in cluster operator status

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -15,7 +15,6 @@ import (
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/machine-config-operator/lib/resourceapply"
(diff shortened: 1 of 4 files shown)


runcom:
> * expected dependent [Bugzilla bug 1744764](https://bugzilla.redhat.com/show_bug.cgi?id=1744764) to be in one of the following states: VERIFIED, RELEASE_PENDING, CLOSED (ERRATA), but it is ON_QA instead

yep, perfect

pkg/operator/sync.go
```diff
@@ -30,22 +29,32 @@ type syncFunc struct {
...
 
-	var errs []error
+	var syncErr syncError
 	for _, sf := range syncFuncs {
 		startTime := time.Now()
-		errs = append(errs, sf.fn(rconfig))
+		syncErr = syncError{
+			task: sf.name,
+			err:  sf.fn(rconfig),
+		}
```
runcom: in 4.1 this is going to shadow errors coming from the other sync functions as we're not breaking at the first error like in 4.2 - this needs another angle to tackle um

LorbusChris: hm, I added the break, see below. Is this what you mean here?

runcom: I think this is good actually.

PR #1086, opened 2019-08-26
Bug 1741817: mcd: Add /run/machine-config-daemon-force stamp file

cgwalters:
This causes the MCD to skip validating against `currentConfig` (or `pendingConfig`).

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1741817

A long time ago, this PR introduced the current model:
https://github.com/openshift/machine-config-operator/pull/245
One aspect of this is we need to avoid reboot loops; that was
a real-world problem early in OpenShift development, although
it is probably unlikely to re-occur today.

Another problem is that we can't simply reconcile by default because
we don't have a mechanism to coordinate reboots:
 https://github.com/openshift/machine-config-operator/issues/662#issuecomment-506472687

However, this PR should aid disaster recovery scenarios and others
where administrators want the MCD to "just do it".

### DIFF
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -40,4 +40,9 @@ const (
 	// object so that Ignition can process it on first boot, and then the MCD can act on
 	// non-Ignition fields such as the osImageURL and kernelArguments.
 	MachineConfigEncapsulatedPath = "/etc/ignition-machine-config-encapsulated.json"
+
+	// MachineConfigDaemonForceFile if present causes the MCD to skip checking the validity of the
+	// "currentConfig" state.  Create this file (empty contents is fine) if you wish the MCD
+	// to proceed and attempt to "reconcile" to the new "desiredConfig" state regardless.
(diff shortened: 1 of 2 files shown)


runcom:
vpc errors on e2e-aws-op

/approve

looks really good

runcom:
> We really need to roll up into status something like "draining node ip-10-0-163-146.us-east-2.compute.internal".

like a "Working reason"

runcom:
```
I0826 21:03:09.123262   26695 update.go:89] pod "downloads-5fdcfc8686-tjs5l" removed (evicted)
```

This looks like the pod that made it slow - and I think it's not listening to drain (same bug as the registry one perhaps, I"ll check)

cgwalters:
After some discussion on this I feel like this PR was wrong: what we really want is something more like "automatically reconcile any changes we detect rather than go degraded".

runcom:
The only thing we need to make sure is that whatever we’re force applying, it must be reflected into a rendered mc (like it is usually). I think one of the DR scenario advices to modify the files on host and use the force file which is ineffective if those changes aren’t in a the rendered machine config as well.

PR #1087, opened 2019-08-26
Bug 1745192: Move registries.conf editing into a subpackage

mtrmac:
<s>(WIP because this includes unmerged https://github.com/openshift/machine-config-operator/pull/1082 .)</s>

**- What I did**
Turned `updateRegistriesConfig` from `pkg/controller/container-runtime-config` into 
`pkg/registries.EditRegistriesConfig` in a separate subpackage, with minimal dependencies
to be possible to call it from other registries.

This is in preparation for possibly moving it out of this repository into library-go entirely, see https://github.com/openshift/openshift-controller-manager/pull/19 for the whole discussion.

Moving it only into a subpackage in this repo is clearly non-disruptive, and it allows reviewing the proposed API before moving it.

The API deals with Go structures only, forcing the callers to do their own TOML encoding/decoding. This is to make it easier for the openshift-controller-manager code to start with a Go-only data structure without having to encode it just to submit it to the edit function.

**- How to verify it**
Existing unit tests.

N/A


mtrmac:
@runcom @umohnani8 @gabemontero PTAL.

mtrmac:
@kikisdeliveryservice No, https://github.com/openshift/openshift-controller-manager/pull/19 needs to happen in 4.2 ; this is to avoid duplicating all of that code in two places.

We are still looking for the best place for the newly created subpackage to exist, MCO is not ideal.

pkg/registries/topo.go
```diff
@@ -1,4 +1,4 @@
-package containerruntimeconfig
+package registries
```
runcom: can this pkg be under `pkg/controller/container-runtime-config`?

mtrmac: Sure; I moved it up here to make it easier to find / read for out-of-repo consumers (who don’t care about the M-C-controller much).

Ideally we would want to have it somewhere out of the 2 (3? 4?) known users into some shared infrastructure repo.

If it is going to stay in MCO, I really don’t care where.

mtrmac:
(Oh, and I forgot, just in case: See the individual commits to see what changed.)

pkg/registries/topo.go
mtrmac: Moved.

mtrmac:
Filed https://github.com/openshift/library-go/pull/513 for the library-go inclusion; that would replace this PR.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -241,164 +240,21 @@ func updateCRIOConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfiguratio
...
-		}
-		tomlConf.Registries = append(tomlConf.Registries, sysregistriesv2.Registry{
-			Endpoint: sysregistriesv2.Endpoint{Location: scope},
-		})
-		return &tomlConf.Registries[len(tomlConf.Registries)-1]
-	}
-
-	mirrorSets, err := mergedMirrorSets(icspRules)
-	if err != nil {
+	if err := registries.EditRegistriesConfig(&tomlConf, internalInsecure, internalBlocked, icspRules); err != nil {
```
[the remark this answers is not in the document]
mtrmac: I would be very surprised if that helps; a library-go subpackage with that modified API would pull in _exactly the same_ containers/image, and other, dependencies, as the code calling it does right now, there would just be one layer of indirection.

The effect would basically be only that library-go would _always_ pull in the TOML encoder/decoder as well (which makes no difference for the two currently contemplated consumers, but removes a bit of flexibility for possible future uses).

---

One thing that could possibly actually help is fixing https://github.com/containers/image/issues/692 ; my problems with Glide in the library-go PR manifested in one of the compression implementations (e.g. via `github.com/klauspost/compress/snappy/testdata/alice29.txt` ).

But that would help with either of the APIs, and might not be essential — I was able to make Glide work with that one package using `git add --force`, at least.

gabemontero:
I've sent out an email to the interested parties on getting to a decision on merging this PR and then rebasing https://github.com/openshift/openshift-controller-manager/pull/19 off of this

umohnani8:
LGTM
@runcom PTAL

runcom:
> LGTM
> @runcom PTAL

we need a BZ explaining why we need such PR

umohnani8:
The BZ for this is https://bugzilla.redhat.com/show_bug.cgi?id=1745192. Well it requires https://github.com/openshift/openshift-controller-manager/pull/19 as well to be fully fixed.

runcom:
/approve

will leave to @kikisdeliveryservice one last review

PR #1098, opened 2019-09-03
Bug 1748844: vendor: use cluster-api's drain lib

runcom:
openshift/kubernetes-drain is about to be officially deprecated and
already discontinued. We need to use the cluster-api's one (which is
basically a fork +
https://github.com/openshift/kubernetes-drain/pull/4).
We also need https://github.com/openshift/kubernetes-drain/pull/4 to
avoid flooding the apiserver as the MAO did (https://bugzilla.redhat.com/show_bug.cgi?id=1733708).
Also, MCO already updated in 4.1 and 4.2 with https://github.com/openshift/kubernetes-drain/pull/3 which is the actual fix to limit the rate of requests for eviction - this bump is to make sure to bring the deadlock fix.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -486,12 +486,12 @@
   revision = "c44a8b61b9f46cd9e802384dfeda0bc9942db68a"
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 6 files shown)


runcom:
/hold

this needs verification and BZs for 4.1/4.2 - doing that

runcom:
good to go now, will take care of https://github.com/openshift/machine-config-operator/pull/1099 after this is in

PR #1107, opened 2019-09-10
WIP Bug 1748798: setup-etcd-environment: add flag bootstrap-srv allowing toggle of SRV discovery

hexfusion:
Currently, we offer no direct mechanism for disabling SRV discovery. Today SRV discovery needs to be disabled for various DR related actions. In the future cluster-etcd-operator will handle cluster membership and will also require SRV discovery to be disabled.

If both `ETCD_DISCOVERY_SRV` and `ETCD_INTIAL_CLUSTER` are set at the same time it will result in etcd failing to start.

```
etcdmain: error verifying flags, multiple discovery or bootstrap flags are set. Choose one of "initial-cluster", "discovery" or "discovery-srv". See 'etcd --help'.
```

This PR also changes the way that DR runtimes are persisted. Currently, we append them to a MCO managed config file. Because this information can be dynamic per-peer member it should not be stored here. Also because the envs are only consumed during the first time etcd starts we do not need to continue to persist them. This works well with `/run/etcd/environment` which gets cleared on each reboot.


cmd/setup-etcd-environment/run.go
```diff
@@ -30,13 +30,16 @@ var (
 		discoverySRV string
 		ifName       string
 		outputFile   string
+		bootstrapSRV bool
```
retroflexer: Where does the new flag --bootstrapSRV get used?

hexfusion: it doesn't by default

retroflexer: But during the recovery this should be set to false, right?

hexfusion: I think it should be a general option to disable but the correct way to change the spec would be via MachineConfig. The downside of that update would be a rolling restart of nodes, not great during discovery. For DR we will handle this with sed, after the cluster is restored and stable the ENV could be permanently disabled with MachineConfig update.

retroflexer: Yeah. On reboot the environment gets reset, so we should be fine.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -336,6 +336,9 @@ contents:
     # sometimes $RUN_ENV is not available if the node is rebooted so we recreate here.
     validate_environment() {
       if [ -f "$RUN_ENV" ] && [ -s "$RUN_ENV" ];then
+        # SRV envs need to be removed until etcd is started during recovery.
+        # After etcd has joined the cluster SRV envs are ignored.
+        sed -i '/ETCD_DISCOVERY_SRV/d' $RUN_ENV
```
retroflexer: /lgtm

hexfusion:
@retroflexer is currently testing this, removing WIP if testing is success please remove hold.

/hold

PR #1113, opened 2019-09-13
Revert "Bug 1750433: 1/3 etcd-member pods crashloopback"

hexfusion:
Reverts openshift/machine-config-operator#1112

I am seeing conflicting information on this between leads.

@mrunalp and @cgwalters differ. I am not comfortable with letting this just slide into 4.2 without a consensus. This was originally removed here #526.

/cc @smarterclayton @sjenning 

ref:
- https://github.com/openshift/machine-config-operator/pull/1112#issuecomment-531004708
- https://github.com/openshift/machine-config-operator/pull/1112#issuecomment-531081024

/hold

### DIFF
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -20,8 +20,6 @@ contents:
         - "--discovery-srv={{.EtcdDiscoveryDomain}}"
         - "--output-file=/run/etcd/environment"
         - "--v=4"
-        securityContext:
-          privileged: true
         volumeMounts:
         - name: discovery
           mountPath: /run/etcd/
@@ -69,8 +67,6 @@ contents:
                 --commonname=system:etcd-metric:${ETCD_DNS_NAME} \
                 --ipaddrs=${ETCD_IPV4_ADDRESS} \
 
-        securityContext:
-          privileged: true
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
         - name: discovery
@@ -109,8 +105,6 @@ contents:
             --listen-client-urls=https://0.0.0.0:2379 \
             --listen-peer-urls=https://0.0.0.0:2380 \
             --listen-metrics-urls=https://0.0.0.0:9978 \
-        securityContext:
-          privileged: true
         resources:
           requests:
             memory: 600Mi


smarterclayton:
I am definitely not comfortable with making etcd privileged.  I concur with revert and a deeper discussion.

hexfusion:
@smarterclayton this change was added in response to BZ 1750433. The crux of the issue was that etcd container was found crash looping after it received "permission denied"  trying to source a host file. The same source action happened successfully in separate containers in the pod. Upon review we found that although the kubelet appeared to still have an active connection to the socket crictl could not communicate with the server and received (Connection Refused).

The proposed solution to insolate etcd-member containers from this was to set as privileged.

Based on these two comments on the original PR and https://github.com/openshift/machine-config-operator/pull/526 I wanted clarification on the change.

- https://github.com/openshift/machine-config-operator/pull/1112#issuecomment-531004708
- https://github.com/openshift/machine-config-operator/pull/1112#issuecomment-531081024

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1750433

mrunalp:
@smarterclayton Based on my debugging yesterday I see that static pods for etcd-member are running as privileged even though privileged isn't specified in the pod yaml and that allows those pods to access /run/etcd from the host. However, from the bug it appears that sometimes they do get restricted with a container selinux label and can't access that directory. The node team has to figure out what is the right behavior for static pods.

mrunalp:
What @cgwalters proposed is a different thing that would have to pursued upstream if we want to have the host path volumes have the right selinux label for container access set by k8s. Today we don't do that and it would make sense for the container process to be privileged if we want to write under that path. Another option would be some setup to label /run/etcd by some other privileged process so etcd containers could be not privileged while still being able to write there.

smarterclayton:
It sounds like the previous interpretation (privileged not necessary to
access host path) that justified 526 are wrong.

How long until we have an answer for node team about why an unprivileged
pod was able to access host path.  Is it because the label changed between
two executions of the container (not stable)?  Host path is supposed to
create directories and label them according to the pod, was this just
coincidentally working because the label changed on the container rarely?

smarterclayton:
Isn't there a bug where the right labels aren't being set anyway?

rphillips:
I have test case reproducer. I'll create a slack discussion

zvonkok:
Running a privileged pod means that the container runs with *spc_t *context
label *. *This means that the container can access any label from the host
(etc_t, var_t, ...) or container (container_file_t, ...).
That is why a privileged container can read any host path with any SELinux
context label.

Looking at my cluster, the etcd directory (no the host)
has container_var_run_t this means it is an container label and a container
that mounts this directory
can read the context.

If for some reason the label on the host is changed, or not applied the
container would not been able to read the contents.

That is one reason why we're running all device-plugins privileged, because
the directory /var/lib/kubelet/device-plugins has a host label (var_lib_t)
and the container policy prevents container_t reading
var_lib_t.

If we make sure that host paths that are mounted into the containers have
the correct labels, a lot of infra container could be run non-privileged.

hexfusion:
/close

This was worked out in slack via #waroom-etcd-crashloop

cgwalters:
> This was worked out in slack via #waroom-etcd-crashloop

Can someone summarize the results of that conversation into e.g. a PR adding a comment to the static pod?

PR #1114, opened 2019-09-16
pkg/daemon: remove force validation file if it exists

runcom:
Even if we document that the file needs to be removed it sounds like the
MCD should do it itself if the force validation file exists. The case
would be someone forgetting to remove the file causing validation to
always be skipped, even on upgrades.

cc @cgwalters @rphillips


runcom:
/hold

let's get this in

runcom:
good to go as well

cgwalters:
To rephrase this commit:

We added a "force overwrite" file to allow people to fix things when they know they wanted to synchronize to the current state.  However, persisting it on disk by default thereafter wasn't explicitly intended - it will silently mask errors.

The flag is much better as a "one time" intervention.

PR #1119, opened 2019-09-17
Bug 1741608: logs are flooded with "failed to get pod stats: failed to get imageFs info: non-existent label "crio-images"" messages

RobertKrawitz:
Fixes: 1741608

Cherrypick of #772


runcom:
ok figured
you need to use go-bindata from the old(pre-fork) repo
cause that's what we're sticking with in 4.1
```
$ go get -u github.com/jteeuwen/go-bindata/...  
$ make update
```
the diff should be empty actually, so you might as well just remove the make update step from commit

jwforres:
this is missing the approved and lgtm labels, please try to get this PR fixed up so it can go into the next merge window

/test verify

PR #1122, opened 2019-09-18
Bug 1753067: Create a namespace for OpenStack infra static pods

Fedosin:
Currently infra pods are created in "openshift-kni-infra" namespace, which doesn't exist. It leads to various OOM errors and CI issues.

This patch creates the missing namespace "openshift-openstack-infra" and puts all the pods there, since "openshift-kni-infra" belongs to KNI team.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1753067

PR #1123, opened 2019-09-18
Bug 1753411: manifests: Create namespace for static pods IPI-BM

yboaron:
Currently infra pods are created in "openshift-kni-infra" namespace, which doesn't exist. It leads to various OOM errors.

This patch creates the missing namespace "openshift-kni-infra".

**- What I did**

**- How to verify it**

### DIFF
--- a/install/0000_80_machine-config-operator_00_namespace.yaml
+++ b/install/0000_80_machine-config-operator_00_namespace.yaml
@@ -17,4 +17,14 @@ metadata:
   labels:
     name: openshift-openstack-infra
     openshift.io/run-level: "1"
+---
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
@yboaron you will need to run `make update` since you modified manifests and also update the BZ so that this is a valid BZ.

kikisdeliveryservice:
Note: this PR was split out of https://github.com/openshift/machine-config-operator/pull/1122

Fedosin:
hi! I realized late last night that MCO doesn't include manifests in the image, so looks like you need to add your namespace here after all: https://github.com/openshift/machine-config-operator/blob/master/install/0000_80_machine-config-operator_00_namespace.yaml

PR #1124, opened 2019-09-18
Bug 1751978: templates/baremetal: Fix keepalived dysfunction on vrrp iface change

celebdor:
When using CNV or other operators that modify how the node is connected
to the network, we may end up in the case where the configured VRRP
interface no longer has an address in the network that it is configured
to hold virtual IPs in.

This patch takes a page from what we do for HAProxy and adds a monitor
side car container that checks keepalived and reloads it when necessary.

Fixes: #1751978
Signed-off-by: Antoni Segura Puimedon <antoni@redhat.com>

### DIFF
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -23,31 +23,8 @@ contents:
       - name: conf-dir
         hostPath:
           path: "/etc/keepalived"
-      initContainers:
-      - name: render-config
-        image: {{ .Images.baremetalRuntimeCfgImage }}
-        command:
-        - runtimecfg
-        - render
-        - "/etc/kubernetes/kubeconfig"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
-        - "--dns-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.NodeDNSIP }}"
-        - "--ingress-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
-        - "/config"
-        - "--out-dir"
-        - "/etc/keepalived"
-        resources: {}
-        volumeMounts:
-        - name: resource-dir
-          mountPath: "/config"
-        - name: kubeconfig
-          mountPath: "/etc/kubernetes/kubeconfig"
-        - name: conf-dir
-          mountPath: "/etc/keepalived"
-        imagePullPolicy: IfNotPresent
+      - name: run-dir
+        empty-dir: {}
       containers:
       - name: keepalived
         securityContext:
@@ -57,14 +34,40 @@ contents:
           - name: NSS_SDB_USE_CACHE
             value: "no"
         command:
-        - /usr/sbin/keepalived
-        args:
-        - "-f"
-        - "/etc/keepalived/keepalived.conf"
-        - "--dont-fork"
-        - "--vrrp"
-        - "--log-detail"
-        - "--log-console"
+        - /bin/bash
+        - -c
+        - |
+          #/bin/bash
+          reload_keepalived()
+          {
+            if pid=$(pgrep -o keepalived); then
+                kill -s SIGHUP "$pid"
+            else
+                /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+            fi
+          }
+
+          msg_handler()
+          {
+            while read -r line; do
+              echo "The client sent: $line" >&2
+              # currently only 'reload' msg is supported
+              if [ "$line" = reload ]; then
+                  reload_keepalived
+              fi
+            done
+          }
+
+          set -ex
+          declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
+          export -f msg_handler
+          export -f reload_keepalived
+          if [ -s "/etc/keepalived/keepalived.conf" ]; then
+              /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+          fi
+
+          rm -f "$keepalived_sock"
+          socat UNIX-LISTEN:${keepalived_sock},fork system:'bash -c msg_handler'
         resources:
           requests:
             cpu: 150m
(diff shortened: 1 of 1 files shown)


celebdor:
Depends on https://github.com/openshift/baremetal-runtimecfg/pull/20 being merged

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -23,31 +23,8 @@ contents:
       - name: conf-dir
         hostPath:
           path: "/etc/keepalived"
-      initContainers:
-      - name: render-config
-        image: {{ .Images.baremetalRuntimeCfgImage }}
-        command:
-        - runtimecfg
```
phoracek: Is runtimecfg still used by anyone or it can be dropped from https://github.com/openshift/baremetal-runtimecfg/?

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,79 @@ contents:
...
+          reload_keepalived()
+          {
+            declare -r pid="$(pgrep -o keepalived)"
+            kill -s SIGHUP "$pid"
+          }
+
+          msg_handler()
+          {
+            while read -r line; do
+              echo "The client sent: $line"  >&2
```
phoracek: there is a possibility so safe one (!) character by dropping the extra space

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,79 @@ contents:
...
+
+          set -ex
+          declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
+          export -f msg_handler
+          export -f reload_keepalived
+          if [ -s "/etc/keepalived/keepalived.conf" ]; then
+              /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+          fi
+
+          rm -f "$keepalived_sock"
```
phoracek: You remove it and start listening to it on the next line? Why?

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,79 @@ contents:
...
+                  reload_keepalived
+              fi
+            done
+          }
+
+          set -ex
+          declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
+          export -f msg_handler
+          export -f reload_keepalived
+          if [ -s "/etc/keepalived/keepalived.conf" ]; then
```
phoracek: There is a regression. Previously it would fail if the config does not exist.

kikisdeliveryservice:
since this depends on openshift/baremetal-runtimecfg#20 

/hold

kikisdeliveryservice:
Also updated the PR title, please use this format in the future :)

templates/common/baremetal/files/baremetal-keepalived.yaml
celebdor: Done

templates/common/baremetal/files/baremetal-keepalived.yaml
celebdor: In case the container is restarted the socket file might be there and socat won't be able to start unless it is gone.

templates/common/baremetal/files/baremetal-keepalived.yaml
celebdor: Now instead of failing it will just wait in socat for a reload when the config is written. What's the issue with that?

templates/common/baremetal/files/baremetal-keepalived.yaml
celebdor: It is used by mdns-publisher

templates/common/baremetal/files/baremetal-keepalived.yaml
phoracek: I see

templates/common/baremetal/files/baremetal-keepalived.yaml
phoracek: Ack

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,79 @@ contents:
...
+        - name: run-dir
+          mountPath: "/var/run/keepalived"
         livenessProbe:
           exec:
             command:
-            - pgrep
-            - keepalived
+            - /bin/sh
+            - -c
+            - kill -s SIGUSR1 "$(pgrep -o keepalived)" && grep -q "State = FAULT" /tmp/keepalived.data
```
yboaron: @celebdor  could you please elaborate about the logic of the LivenessProbe?

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,79 @@ contents:
...
-        - "--log-detail"
-        - "--log-console"
+        - /bin/bash
+        - -c
+        - |
+          #/bin/bash
+          reload_keepalived()
+          {
+            declare -r pid="$(pgrep -o keepalived)"
+            kill -s SIGHUP "$pid"
```
yboaron: @celebdor I think that "$pid" could be empty (the keepalived container starts before monitor container and .conf is missing), is it correct?

templates/common/baremetal/files/baremetal-keepalived.yaml
celebdor: Absolutely :-)

If keepalived is not running, the pgrep will be empty and the kill command will fail with exit code 1. This means the liveness check will consider it failed.

If keepalived is running, sending SIGUSR1 to the parent keepalived process will make keepalived to write /tmp/keepalived.data containing information about each vrrp group including the state. The state can be MASTER, BACKUP or FAULT. If any of the states are FAULT grep will exit with exit code 1 and the liveness test will fail.

Oops. I just realized I forgot to put put the negative. Will fix it now

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -57,29 +34,82 @@ contents:
...
+        - name: run-dir
+          mountPath: "/var/run/keepalived"
         livenessProbe:
           exec:
             command:
-            - pgrep
-            - keepalived
+            - /bin/sh
+            - -c
+            - kill -s SIGUSR1 "$(pgrep -o keepalived)" && ! grep -q "State = FAULT" /tmp/keepalived.data
```
yboaron: @celebdor, not sure that I (still :-) ) fully understand the liveness probe logic.
So, if some component (e.g: CNV ) changed network interfaces configuration, that may lead to keepalived failure (State=FAULT), and this container will be restarted by Kubelet. IIUC, the problem will be fixed only after monitor container updates .conf file, right?  
Could u please explain what added value do we get from this Liveness check?

celebdor: The added value is that if any of the VIP management (ingress, api or dns) gets faulty for any cause, we'll restart keepalived.

celebdor:
Verified with success that the IP was moved to brext and the VIPs did as well. The downtime for reconfiguration of the interfaces was of about 90seconds.

kikisdeliveryservice:
looks good.

/lgtm

PR #1126, opened 2019-09-20
OpenStack: CI failure: ensure infra pods have tolerations

Fedosin:
One of the OpenShift requirements is that all pods must have "not-ready" and "unreachable" tolerations. This commit adds them to the infra pods.


Fedosin:
/hold

looks like the issue should be solved by https://github.com/openshift/origin/pull/23833

celebdor:
@yboaron ptal

PR #1127, opened 2019-09-20
Bug 1753953: Temporarily put OpenStack infra pods into non-existing namespace

Fedosin:
#1122 breaks OpenStack CI, so we need to prepare the infrastructure first.
This patch temporarily puts all OpenStack infra pods to a non-existing namespace to make CI green.

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
@@ -7,7 +7,7 @@ contents:
     apiVersion: v1
     metadata:
       name: haproxy
-      namespace: openshift-openstack-infra
+      namespace: non-existent
       creationTimestamp:
       deletionGracePeriodSeconds: 65
       labels:
--- a/manifests/openstack/coredns.yaml
+++ b/manifests/openstack/coredns.yaml
@@ -3,7 +3,7 @@ kind: Pod
 apiVersion: v1
 metadata:
   name: coredns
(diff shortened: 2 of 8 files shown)


tomassedovic:
/lgtm
/hold

Let's make sure the (non-blocking) `e2e-openstack` job passes with this.

kikisdeliveryservice:
why is this merging with a BZ attached?

tomassedovic:
/lgtm
/hold

Again, let's make sure the `e2e-openstack` job passes.

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -7,7 +7,7 @@ contents:
     apiVersion: v1
     metadata:
       name: haproxy
-      namespace: openshift-openstack-infra
+      namespace: non-existent
```
kikisdeliveryservice: Q:  was this namespace supposed to be kni and missed? or is the openstack correct here (i see it's in bm dir)

Fedosin: now kni namespace exists and it is used by "metal" platform. before #1123 it didn't exist and used by both openstack and metal platforms.
Therefore, in order to get things back to normal, we have temporarily place the pods in a non-existent namespace.

kikisdeliveryservice:
Blocked by: https://bugzilla.redhat.com/show_bug.cgi?id=1754154

PR #1133, opened 2019-09-25
Bug 1755560: Ensure ETCD_INITIAL_CLUSTER is preserved during restore

smarterclayton:
`/run/etcd/environment` can contain ETCD_INITIAL_CLUSTER which would
override the value the user provided as an argument. This flow is
used during disaster recovery when only a single master remains and
the INITIAL_CLUSTER value needs to come from the user (the other
members are dead / gone).

This prevents performing a second recovery without editing the
`/run/etcd/environment`.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -26,7 +26,7 @@ contents:
     fi
 
     SNAPSHOT_FILE="$1"
-    ETCD_INITIAL_CLUSTER="$2"
+    INITIAL_CLUSTER="$2"
     ASSET_DIR=./assets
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
/approve

@hexfusion @retroflexer PTAL

hexfusion:
I thought this was 4.1 my bad

/hold cancel

smarterclayton:
Do not put prefixes on titles - that breaks the changelog

runcom:
So, somehow it looks like the router pod can't be evicted anymore:

```
I0926 08:16:44.517793    2694 update.go:89] error when evicting pod "router-default-7644fff96d-qsn8g" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

I'll retest to see if that is consistent

kikisdeliveryservice:
reported the boskos errors in this last run.

PR #1134, opened 2019-09-25
Bug 1755557: Ensure ETCD_INITIAL_CLUSTER is preserved during restore

smarterclayton:
`/run/etcd/environment` can contain ETCD_INITIAL_CLUSTER which would override the value the user provided as an argument. This flow is used during disaster recovery when only a single master remains and the INITIAL_CLUSTER value needs to come from the user (the other members are dead / gone).  In the specific case, the recovery flow failed because we couldn't reestablish the master.

Backport #1133


hexfusion:
avoiding collision makes sense to me

defer to @retroflexer

kikisdeliveryservice:
thanks for reviewing @retroflexer 
/approve

PR #1100, opened 2019-09-03
[4.3] Update containers/image to v3.x

mtrmac:
**- What I did**

… to make it easier for consumers of `pkg/controller/container-runtime-config/registries`
to use the newer major API version.

**- How to verify it**

Existing tests; the bump changes nothing directly used from this repository.

**- Description for the changelog**
Updated containers/image to 3.0.2.

PR #1136, opened 2019-09-26
Migrate to Go Modules

LorbusChris:
**- What I did**
Add tools.go for build tooling
Run `go mod init` and add replace directives in go.mod
Update github.com/containers/image to v3.0.2. Resolves #1100
Remove Gopkg files
Add `go-deps` make target

**- How to verify it**
CI

Migrated to Go Modules


Makefile
```diff
@@ -37,13 +37,21 @@ test-unit:
...
+update: install-go-bindata install-go-deps
 	hack/update-codegen.sh
 	hack/update-generated-bindata.sh
 
 install-go-bindata:
 	hack/install-go-bindata.sh
 
+install-go-deps:
+	go mod tidy
+	rm -rf ./vendor
```
runcom: as discussed in chat, we shouldn't be doing this but all this should be idempotent (like `dep ensure` today)

Makefile
```diff
@@ -37,13 +37,21 @@ test-unit:
...
+	go mod vendor
+	# go mod does not vendor in scripts so we need to get them manually
```
runcom: is there any way to whilelist these files in go mod?

go.mod
```diff
@@ -0,0 +1,107 @@
...
+	golang.org/x/net v0.0.0-20190812203447-cdfb69ac37fc // indirect
+	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45 // indirect
+	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
+	golang.org/x/tools v0.0.0-20190711191110-9a621aea19f8 // indirect
+	google.golang.org/appengine v1.6.1 // indirect
+	google.golang.org/genproto v0.0.0-20190708153700-3bdd9d9f5532 // indirect
+	google.golang.org/grpc v1.22.0 // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	gotest.tools v2.2.0+incompatible // indirect
+	k8s.io/api v0.0.0-00010101000000-000000000000
```
runcom: why don't we have clean tags like go dep?

Makefile
LorbusChris: not that I know of

LorbusChris: I think if the directory itself was recognized as go module, it would probably respect the files. (some scripts in other modules are also vendored)

Makefile
runcom: https://github.com/openshift/cluster-monitoring-operator/blob/master/Makefile#L76

go.mod
LorbusChris: version for this is in the replace directive

Makefile
runcom: I'm also getting this diff when just running `go mod tidy` or `vendor`

LorbusChris: if it is indeed idempotent, this should be safe to do though, right?
Remove anyway?

LorbusChris: which diff are we talking about?
Should this just be:
```
go mod tidy
go mod vendor
go mod verify
```

LorbusChris:
btw I noticed that in order to run `make verify` on the host now, gpgme-devel and libassuan-devel have to be installed (they are on the builder image that CI uses). maybe I missed something in the meantime, but that wasn't required for me locally before I believe.

Makefile
runcom: > if it is indeed idempotent, this should be safe to do though, right?
> Remove anyway?

it will re-download everything if you remove - if you don't it'll probably just do an hash check and prevents downloading if they match

> which diff are we talking about?

when I run `go mod tidy` or `go mod verify` (sorry haven't pasted the diff lmao), I get this from this PR in my local development workspace:

```diff
diff --git a/go.mod b/go.mod
index 2205020f..7f2bf48f 100644
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
        github.com/Microsoft/hcsshim v0.8.6 // indirect
        github.com/ajeddeloh/go-json v0.0.0-20170920214419-6a2fe990e083 // indirect
        github.com/ajeddeloh/yaml v0.0.0-20170912190910-6b94386aeefd // indirect
-       github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf // indirect
        github.com/apparentlymart/go-cidr v1.0.0
        github.com/ashcrow/osrelease v0.0.0-20180626175927-9b292693c55c
        github.com/containerd/cgroups v0.0.0-20190710130057-d596c78861b1 // indirect
@@ -49,7 +48,6 @@ require (
        github.com/joho/godotenv v1.3.0
        github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
        github.com/kubernetes-sigs/cri-o v1.12.10
-       github.com/modern-go/reflect2 v1.0.1 // indirect
        github.com/mtrmac/gpgme v0.0.0-20170102180018-b2432428689c // indirect
        github.com/opencontainers/go-digest v1.0.0-rc1
        github.com/opencontainers/image-spec v1.0.1 // indirect
```

LorbusChris: talking about idempotency, it doesnt seem to do that on my system :smile:

ashcrow:
In general I'm a big fan of moving to `go mod`. Thanks for taking this on @LorbusChris!

Makefile
```diff
@@ -37,13 +37,22 @@ test-unit:
...
-update: install-go-bindata
+update: install-go-bindata install-go-deps
 	hack/update-codegen.sh
 	hack/update-generated-bindata.sh
 
 install-go-bindata:
 	hack/install-go-bindata.sh
 
+install-go-deps:
+	go mod tidy
```
sgreene570: Should this directive set the `GO111MODULE` environment variable before running `go mod ..`?

This is what I get when I run `make install-go-deps` on my machine:
```
go mod tidy
go: modules disabled inside GOPATH/src by GO111MODULE=auto; see 'go help modules'
make: *** [Makefile:48: install-go-deps] Error 1
```

runcom: likely, I think we can bring much from https://github.com/openshift/cluster-monitoring-operator/blob/master/Makefile#L76 :) as they already successfully moved to go modules

LorbusChris: yes, good catch! I forgot to set this explicitly; had it disabled globally.

go.mod
sgreene570: I think any of the weird `v0.0.0-00010101000000-000000000000` version tags can be shortened to just `v0.0.0` for readability since we are replacing the dependency down below, although that's definitely not required (see https://github.com/kubernetes/minikube/blob/master/go.mod#L79 for an example).

LorbusChris: I'll do that if I need to push this branch again for whatever reason

tools.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	// Code generators built at runtime.
+	_ "k8s.io/code-generator" // TODO: Investigate why scripts in this directory are removed and not vendored by go mod.
+	_ "k8s.io/code-generator/cmd/client-gen"
+	_ "k8s.io/code-generator/cmd/conversion-gen"
+	_ "k8s.io/code-generator/cmd/deepcopy-gen"
+	_ "k8s.io/code-generator/cmd/defaulter-gen"
+	_ "k8s.io/code-generator/cmd/informer-gen"
+	_ "k8s.io/code-generator/cmd/lister-gen"
```
runcom: we can move go-bindata and hopefully golangci-lint here as suggested in https://github.com/openshift/machine-config-operator/pull/961#issuecomment-511134245 (but this is a comment for a follow up I'm experimenting with)

Makefile
```diff
@@ -37,13 +39,22 @@ test-unit:
...
+	go mod vendor
+	go mod verify
+	# go mod does not vendor in scripts so we need to get them manually
+	# NOTE: Make sure to keep this up-to-date with the version specified in go.mod
```
runcom: I really hate this - I know it's go mod fault (there are issues in Go's github describing why they're not letting ppl do this). This is highly sub optimal and can lead to silly skews - let's see if we can find a better solution as well (or check if someone else fixes this issue somehow)

LorbusChris: maybe split this and the golangci-lint mod vendoring via tools.go out into new new follow up issues and get this merged for now (also especially now that i's all green lit on CI)? May not get a quick fix here otherwise because I'm travelling atm. I havent found a good explanation of when scripts aren't vendored, maybe we can file this upstream so we get a mod pkg we can vendor that actually contains the scripts as well?

LorbusChris: also feel free to push follow up fix commits directly to this branch :)

runcom:
/hold

on my mac:

```
➜  machine-config-operator git:(pr/origin/1136) ✗ make install-go-deps 
go mod tidy
go: finding gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.3
go: finding gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.3
github.com/openshift/machine-config-operator/cmd/gcp-routes-controller imports
        github.com/InVisionApp/go-health/checkers tested by
        github.com/InVisionApp/go-health/checkers.test imports
        gopkg.in/DATA-DOG/go-sqlmock.v1: cannot find module providing package gopkg.in/DATA-DOG/go-sqlmock.v1
github.com/openshift/machine-config-operator/pkg/daemon imports
        github.com/openshift/cluster-api/pkg/drain tested by
        github.com/openshift/cluster-api/pkg/drain.test imports
        github.com/go-log/log/capture: module github.com/go-log/log@latest (v0.1.0) found, but does not contain package github.com/go-log/log/capture
make: *** [Makefile:50: install-go-deps] Error 1
```

runcom:
> on my mac:

go1.13 might be the issue https://github.com/golang/go/issues/34085

sgreene570:
> ```
>         github.com/go-log/log/capture: module github.com/go-log/log@latest (v0.1.0) found, but does not contain package github.com/go-log/log/capture
> make: *** [Makefile:50: install-go-deps] Error 1
> ```
Hm, `v.0.1.0` definitely won't work for `github.com/go-log/log` since the `capture` code wasn't introduced until after that release. Very strange that go 1.12 doesn't throw an error but 1.13 does? Anyways, maybe we can put that dependency under the replace directive to force a more up to date commit (since there aren't any other tags). For reference, [here](https://github.com/openshift/cluster-api/blob/master/Gopkg.lock#L117) is the version of `go-log` used by the openshift `cluster-api`.

runcom:
> Hm, `v.0.1.0` definitely won't work for `github.com/go-log/log` since the `capture` code wasn't introduced until after that release. Very strange that go 1.12 doesn't throw an error but 1.13 does? Anyways, we can import that dependency at a newer commit (since there aren't any other tags). For reference, [here](https://github.com/openshift/cluster-api/blob/master/Gopkg.lock#L117) is the commit of `go-log` used by the openshift `cluster-api`.

yeah, it's a 1.13 issue/regression

PR #1137, opened 2019-09-26
[DO NOT MERGE] test e2e-aws-op on 4.2

runcom:
`master` seems broken because we can't evict the router pod - a 4.2 PR passed the tests tho so this is making sure 4.2 is unaffected.


runcom:
now, why this hasn't run e2e-aws-op remains a mystery

PR #1139, opened 2019-09-27
test/e2e: do not test nosmt, move maxUnavailable to 2

runcom:
nosmt is known to slow everything down and in our case it's making e2e
fail. It's still not clear why the 4.2 branch works with nosmt but
master doesn't. For the sake of kargs testing tho, we can just test a
dummy arg.

Also move maxUnavailable to 2 cause the router pod can't reschedule anymore with 3 - used to in 4.2 - we'll keep investigating but at least we could unblock master


runcom:
/hold

still broken reboots which kubelet not coming up on 43 rhcos and 1.16 kubelet

PR #1144, opened 2019-10-01
bump kubelet dependency for TopologyManagerPolicy config option

rphillips:
**- What I did**
Bumps kubelet dependency to 1.16 to bring in `TopologyManagerPolicy` kubelet configuration option.

**- How to verify it**
topologyManagerPolicy is settable via KubeletConfig.

```
NONE
```

Note: I attempted to bump all the dependencies to 1.16, but ran into issues with runc/kubernetes and dep not wanting to converge.


rphillips:
Had a great conversation with @kikisdeliveryservice on slack... Found out we are migrating to `go mod`  (https://github.com/openshift/machine-config-operator/pull/1136). To configure the TopologyManager we just need the openshift/kubelet bump. I'll keep an eye out for it.

PR #1152, opened 2019-10-03
baremetal: templates/manifests: configure coredns hosts plugin to point at /dev/null

cybertron:
We aren't actually reading any records from the configured file, but
we need to have it anyway. Currently it points at a non-existent
file, which causes unnecessary warnings at startup. Setting it to
/dev/null instead silences these warnings and better indicates the
intent to have the file be a noop.

**- What I did**

**- How to verify it**

Configure baremetal coredns hosts plugin to point at /dev/null to silence missing file warnings.

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -5,7 +5,7 @@
     forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}}
     cache 30
     reload
-    hosts /etc/coredns/api-int.hosts {{ .ControllerConfig.EtcdDiscoveryDomain }} {
+    hosts /dev/null {{ .ControllerConfig.EtcdDiscoveryDomain }} {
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
I think you will have to: https://github.com/openshift/machine-config-operator/blob/master/docs/HACKING.md#updating-the-manifests

cybertron:
Oops, I had it locally but forgot to commit. Should be fixed now.

cybertron:
Apparently go-bindata on my dev system is too old. :-/

cybertron:
/retest

None of these changes should affect aws, so I assume those failures are flakes.

cybertron:
I guess technically the file is optional, but if you want to specify the zones for which the plugin is authoritative you need to include the file because the plugin parameters are positional. It might work to drop the zone specification too though. Since fallthrough is enabled it may not matter if we limit it to certain zones. I can try that.

cybertron:
Okay, I pushed an alternate implementation in https://github.com/openshift/machine-config-operator/pull/1169 that just removes the parameters to the hosts file.

cybertron:
#1169 fixed the problem this was addressing.

PR #1153, opened 2019-10-04
Bug 1758343: pkg/daemon: default drain grace period to -1

runcom:
Also bump cluster-api's drain library to further fix this by disallowing
higher drain's grace period than what a pod defines in its spec.
Changing the timeout to -1 ensures the library always honor the pod's
grace period and fixes a subtle issue with the "downloads" pod taking
600s to be evicted.


runcom:
wat, clusters in CI jobs are 3 masters only and workers aren't coming up either

eparis:
We merged this in 4.2.0. if we thought it was safe it should be in 4.3, even if CI if red.

Issue #981, opened 2019-07-17
Don't error out for non-oscontainer images


cgwalters:
Today if one does a manual `cosa build` of RHCOS, and doesn't know to run the (internal RHCOS) script to set the [custom origin](https://github.com/projectatomic/rpm-ostree/pull/1406), the MCO will error out.

Also along with the general thread of supporting pivoting from FCOS for OKD, we should make this not an error - if the origin is ostree (or rojig, or whatever), just continue and replace it with content from the oscontainer.  We don't need to care what the booted origin is.

sgreene570:
@cgwalters do you know what the MCO errors will look like in this situation?

cgwalters:
You can reproduce this by using an origin that's not an oscontainer.  ("origin" here being ostree's term for "where you get updates" - it's tricky to explain since there's not a close equivalent in either yum or docker worlds)

Here's an example way to do it:

```
walters@toolbox ~/s/g/c/coreos-assembler> oc debug node/ip-10-0-130-6.us-east-2.compute.internal
Starting pod/ip-10-0-130-6us-east-2computeinternal-debug ...
To use host binaries, run `chroot /host`
chroot /host
rpm-ostree status
Pod IP: 10.0.130.6
If you don't see a command prompt, try pressing enter.
sh-4.2# chroot /host
sh-4.4# rpm-ostree status -v
State: idle
AutomaticUpdates: disabled
Deployments:
* pivot://registry.svc.ci.openshift.org/origin/4.2-2019-07-17-143546@sha256:80a59320ee5543406348f1f47bfa0e3c4135acd43297370a46c7b2581ad5e150
              CustomOrigin: Managed by machine-config-operator
                   Version: 420.8.20190717.1 (2019-07-17T08:59:58Z)
                    Commit: 16392af432d1bf6d20a2809bcd7ec6e10ff4ecdfa235a0cbb36dcefc9e0ac87c
                            |- art-rhcos-4.1 ((invalid timestamp))
                            |- art-rhcos-4.2 ((invalid timestamp))
                            |- ootpa-baseos ((invalid timestamp))
                            `- ootpa-appstream ((invalid timestamp))
                    Staged: no
                 StateRoot: rhcos

  pivot://quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:fdd84b6bdd8da02bac167bd2002261fd4c4f4f626b1978cf6070d659b5c46498
              CustomOrigin: Provisioned from oscontainer
[...]

PR #1155, opened 2019-10-04
daemon: Support unpivoted images

cgwalters:
Today the RHCOS bootimages come "prepivoted"; they have a custom
origin set to the `machine-os-content` that matches.  The idea
here is to avoid an extra reboot.

In practice we're always going to be updating and rebooting today,
because we don't have a mechanism to update bootimages:
https://github.com/openshift/os/issues/381

I plan to change the RHCOS build system to stop "prepivoting"; the
oscontainer will be pushed *after* the bootimages.  This breaks
a complicated dependency and ensures that the RHCOS "source of truth"
is the cosa storage.

Also, this allows one to use the MCD when e.g. booting Fedora CoreOS.

Closes: https://github.com/openshift/machine-config-operator/issues/981

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -114,14 +114,16 @@ func (r *RpmOstreeClient) GetStatus() (string, error) {
 }
 
 // GetBootedOSImageURL returns the image URL as well as the OSTree version (for logging)
+// Returns the empty string if the host doesn't have a custom origin that matches pivot://
+// (This could be the case for e.g. FCOS, or a future RHCOS which comes not-pivoted by default)
 func (r *RpmOstreeClient) GetBootedOSImageURL() (string, string, error) {
 	bootedDeployment, err := r.getBootedDeployment()
 	if err != nil {
 		return "", "", err
 	}
 
-	// the canonical image URL is stored in the custom origin field by the pivot tool
-	osImageURL := "<not pivoted>"
+	// the canonical image URL is stored in the custom origin field.
+	osImageURL := ""
 	if len(bootedDeployment.CustomOrigin) > 0 {
 		if strings.HasPrefix(bootedDeployment.CustomOrigin[0], "pivot://") {
 			osImageURL = bootedDeployment.CustomOrigin[0][len("pivot://"):]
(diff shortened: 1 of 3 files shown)


cgwalters:
Tested locally in a libvirt build with custom local RHCOS via
```
$ xokdinst launch -D -p libvirt -O file:///srv/walters/rhcos-4.3/builds/latest/x86_64/rhcos-43.80.20191002.0-qemu.x86_64.qcow2 -I registry.svc.ci.openshift.org/ocp/release:4.3.0-0.ci-2019-10-04-083724 osiris
```
where the bootimage isn't pivoted.  Will let CI cover the current path.

pkg/daemon/rpm-ostree.go
```diff
@@ -168,14 +174,16 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
...
-			return
-		}
-		if targetMatched {
-			changed = false
-			return
+		if previousPivot != "" {
+			var targetMatched bool
+			targetMatched, err = compareOSImageURL(previousPivot, container)
+			if err != nil {
+				return
```
runcom: log this err?

cgwalters: Not new in this PR; had to look, but this function uses named return values, so we're correctly propagating errors back up to the caller, who will handle it (usually log).

pkg/daemon/rpm-ostree.go
```diff
@@ -150,7 +152,11 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
 		if strings.HasPrefix(defaultDeployment.CustomOrigin[0], "pivot://") {
 			previousPivot = defaultDeployment.CustomOrigin[0][len("pivot://"):]
 			glog.Infof("Previous pivot: %s", previousPivot)
+		} else {
+			glog.Infof("Previous custom origin: %s", defaultDeployment.CustomOrigin[0])
 		}
+	} else {
+		glog.Infof("Current origin is not custom")
```
kikisdeliveryservice: minor: this should glog.Info() no?

pkg/daemon/rpm-ostree.go
runcom: awesome, missed that, sorry for the noise

PR #1164, opened 2019-10-09
OpenStack: change coredns healthcheck port 8080->18080

Fedosin:
Port 8080 is used by cluster-network-operator in 4.3, so we can't deploy a cluster because of the conflict with coredns:

error listening on :8080: listen tcp :8080: bind: address already in use

This patch changes the coredns healthcheck port from 8080 to 18080 to resolve the conflict.

### DIFF
--- a/manifests/openstack/coredns-corefile.tmpl
+++ b/manifests/openstack/coredns-corefile.tmpl
@@ -1,6 +1,6 @@
 . {
     errors
-    health
+    health :18080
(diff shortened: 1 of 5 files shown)


manifests/baremetal/coredns-corefile.tmpl
```diff
@@ -1,6 +1,6 @@
 . {
     errors
-    health
+    health :18080
```
iamemilio: I don't think we should change baremetal's code, lets just alert them of the issue and let them do it on their own.

kikisdeliveryservice: agreed

Fedosin: yes, sorry :(

PR #1165, opened 2019-10-10
doc: Add FAQ.md

cgwalters:
Let's try to record the answers to questions in Git, so that:

- They're easily searchable
- They're public
- They can be fixed over time
- They might make their way into the official documentation more easily


cgwalters:
This question came up on a private mailing list, so I can't link to it.  But my hope is that in the future, when people ask more questions like this, we answer them by submitting a PR to this file if it makes sense - less replying in private email/chat.

runcom:
👍

PR #1169, opened 2019-10-11
baremetal: templates/manifests: Remove parameters from hosts plugin c…

cybertron:
…onfiguration

Currently we have the hosts plugin configured to look at a dummy
file since we serve all of our addresses from the inline content.
However, this causes a warning when coredns starts up because the
file doesn't exist.

Since all of the parameters to the hosts plugin are optional, we can
just drop the file entirely. This also requires dropping the zone,
but since we use fallthrough it doesn't really matter if we specify
a zone. Any requests the hosts plugin can't handle will just be
passed along to the next plugin.

Not specifying a file causes it to default to /etc/hosts, which is
fine because the coredns container /etc/hosts file only contains
the entries for localhost.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -5,7 +5,7 @@
     forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}}
     cache 30
     reload
-    hosts /etc/coredns/api-int.hosts {{ .ControllerConfig.EtcdDiscoveryDomain }} {
+    hosts {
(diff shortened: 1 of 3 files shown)


yboaron:
@cybertron , overall LGTM, but I have one Q: Do you think that setting coredns to point on host's /etc/hosts is safe? I mean coredns may wrongly resolve names that added by the admin/user?

cybertron:
The container has its own copy of /etc/hosts, it's not a bind mount from the host. So even if someone adds entries to /etc/hosts they won't show up in the file coredns looks at.

ericavonb:
I don’t know enough to comment on much except outside like template formatting on this one. From that angle though,
/lgtm

PR #1172, opened 2019-10-14
WIP: implement basic metrics

kikisdeliveryservice:


pkg/daemon/metrics.go
```diff
@@ -5,11 +5,55 @@ import (
...
+
+	// RebootCounter is a metric incremented when MCD reboots
+	RebootCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mcd_reboots",
+			Help: "A metric incremented when MCD reboots",
+		}, []string{"node"})
+
+	// DrainCounter is a metric incremented when MCD drains
+	DrainCounter = prometheus.NewGaugeVec(
```
runcom: this is amazing and opens up to better debugging :pray: another cool one would be an os upgrade one to check on upgrades

pkg/daemon/metrics.go
```diff
@@ -5,11 +5,55 @@ import (
...
+var (
+	// DefaultBindAddress is the port for the metrics listener
+	DefaultBindAddress = ":8797"
+
+	// RebootCounter is a metric incremented when MCD reboots
+	RebootCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mcd_reboots",
+			Help: "A metric incremented when MCD reboots",
+		}, []string{"node"})
```
ericavonb: Should this be a counter rather than a gauge? since it only increment upwards, it seems to fit better with a counter if I'm understanding correctly https://prometheus.io/docs/concepts/metric_types/#counter . 

Although, a gauge metric for active reboots could be useful too 🤔

pkg/daemon/metrics.go
```diff
@@ -5,11 +5,55 @@ import (
...
+
+	// DrainCounter is a metric incremented when MCD drains
+	DrainCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mcd_drains",
+			Help: "A metric incremented when MCD drains",
```
ericavonb: If this is going to be a gauge, it could increment the counter at drain start and decrement at drain finish. Otherwise, should this be a counter like above?

Or also, from my experience looking at that bug yesterday - it would be great to have a metric on how long the drain attempt took. Like, `mcd_drain{node: <node>, state: <"success"|"failure">, err: <error>, duration: <(sec)>}`

runcom: >  it would be great to have a metric on how long the drain attempt took

this would indeed be great to have

pkg/daemon/update.go
```diff
@@ -161,7 +163,12 @@ func (dn *Daemon) drain() error {
 		}
 		return errors.Wrap(err, "failed to drain node")
 	}
+
 	dn.logSystem("drain complete")
+	t := time.Since(startTime).Seconds()
+	glog.Infof("Successful drain took %v sec", t)
```
kikisdeliveryservice: per mcd logs, this logging/block is being hit:
```
I1024 17:58:51.277698    3826 update.go:1044] drain complete
I1024 17:58:51.280250    3826 update.go:169] Successful drain took 32.941232105 sec
```
but the metric never gets updated. wondering if it's bc of timing? it would update <.2 sec before the node reboots?

kikisdeliveryservice: so what i think is happening is that the scrape is taking longer to happen than the actual reboot. so if the scrape is 1 sec and the metric update to reboot is .2 sec then it never picks up the metric.  I _will_ pick it up if something post drain delays the reboot, which could still be helpful to have when trouble shooting, ie "well we know drain did complete at X seconds so what happened after that?"

runcom:
/skip
/approve

leaving to @yuqi-zhang @ericavonb for a final lgtm

ericavonb:
/lgtm
@yuqi-zhang any last thoughts on this one?

yuqi-zhang:
Nope, LGTM!

PR #1176, opened 2019-10-15
split stopping kubelet and reboot into two commands for debugging

rphillips:
**- What I did**
Split the kubelet stop and reboot into two steps to aid in debugging.

**- How to verify it**


kikisdeliveryservice:
cc: @runcom PTAL this would be helpful in sorting through some of those looong logs.

rphillips:
Force pushed to add a `--wait` for the stop kubelet command.

runcom:
this looks good (@cgwalters do you see something off just to triple check?)

/approve

PR #1184, opened 2019-10-17
Revert "split stopping kubelet and reboot into two commands for debugging"

rphillips:
Reverts openshift/machine-config-operator#1176

The shutting down of the kubelet needs to be atomic with the reboot. Shutting down the kubelet will cause the MCD to be killed.


cgwalters:
Ah, right.  yes.
/approve

ashcrow:
Flake getting an image:
```
received unexpected HTTP status: 500 Internal Server Error
```

ashcrow:
More of
```
received unexpected HTTP status: 500 Internal Server Error
```

holding off to retest

rphillips:
fixed yet?
/test images

ashcrow:
> fixed yet?

My understanding is it should be. We will see :crossed_fingers:.

kikisdeliveryservice:
> > fixed yet?
> 
> My understanding is it should be. We will see crossed_fingers.

i think you jinxed it

rphillips:
I mentioned in slack... After this PR we need to migrate from `reboot` to `systemctl reboot` due to:

```
sh-4.4# reboot ; echo $?
Failed to connect to bus: No data available
0
```

kikisdeliveryservice:
even tho this might trigger ci flooding my inbox :laughing: 
/lgtm

cgwalters:
> I mentioned in slack... After this PR we need to migrate from reboot to systemctl reboot due to:

That doesn't make sense to me since `reboot` *is* `systemctl`:

```
$ ls -al /usr/sbin/reboot
lrwxrwxrwx. 1 root root 16 Sep  3 13:01 /usr/sbin/reboot -> ../bin/systemctl*
```

"Failed to connect to bus" sounds like something that might happen if the system was in the middle of rebooting and `dbus.service` had already gone down.

rphillips:
@cgwalters ```podman run --privileged -ti -v /:/rootfs fedora /bin/bash -c 'chroot /rootfs'```
and try `reboot`

rphillips:
@ashcrow mentioned in slack that systemctl.c does have a number of `multiple paths for reboot depending on environment and call`.

Likely a bug with the way $0 is handled in systemd.

Issue #1185, opened 2019-10-17
fails to build in Brew


cgwalters:
```
Mock Version: 1.3.4
Mock Version: 1.3.4
ENTER ['do'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target s390x --nodeps /builddir/build/SPECS/machine-config-daemon.spec'], nspawn_args=[]shell=Falseuid=994env={'LANG': 'en_US.UTF-8', 'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOSTNAME': 'mock', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'HOME': '/builddir', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PS1': '<mock-chroot> \\s-\\v\\$ '}chrootPath='/mnt/build/mock/rhaos-4.3-rhel-8-build-5254082-3544966/root'gid=135user='mockbuild'timeout=86400logger=<mockbuild.trace_decorator.getLog object at 0x3fff6f4d5c0>printOutput=False)
Executing command: ['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target s390x --nodeps /builddir/build/SPECS/machine-config-daemon.spec'] with env {'LANG': 'en_US.UTF-8', 'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOSTNAME': 'mock', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'HOME': '/builddir', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PS1': '<mock-chroot> \\s-\\v\\$ '} and shell False
Building target platforms: s390x
Building for target s390x
Wrote: /builddir/build/SRPMS/machine-config-daemon-4.3.0-201910171625.git.2.3bef96a.el8.src.rpm
Child return code was: 0
[...]

PR #1186, opened 2019-10-17
hack/build-go: Print `go env` if it doesn't give us what we want

cgwalters:
This is failing in Brew (rpm build) and reproducing that is annoying,
so let's have this print the env if it fails.


cgwalters:
xref https://github.com/openshift/machine-config-operator/issues/1185

PR #1187, opened 2019-10-18
Bug 1762019: use systemctl reboot instead of reboot

rphillips:
**- What I did**

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1762019#c10

There are a number of master and worker logs that show `drain complete` from the MCD, but do not reboot.

MCD performs the following operations when starting:

  0. runs in `privileged: true` security context
  1. Mounts hostPath `/` to `/rootfs`
  2. Inside the container: chroot /rootfs

Running a similar test by using podman:

```sh
# podman run --privileged -ti -v /:/rootfs fedora /bin/bash -c 'chroot /rootfs'
# reboot ; echo $?
Failed to connect to bus: No data available
0
```

The `reboot` command it symlinked to systemctl; however, there appears to be a bug with the $0 handling with systemctl versus just running `systemctl reboot`.

Running the same test with `systemctl reboot` will reboot the machine every single time without errors.

**- How to verify it**


runcom:
Talked to Mrunal as well, all cool

/approve
/lgtm

ashcrow:
```
```
Unrelated

PR #1188, opened 2019-10-18
spec: Remove GOPATH hacks

cgwalters:
Now that I look at the build log more closely, the error is obvious:
`$GOPATH/go.mod exists but should not`

Now that in 4.3 we use go modules, we don't need to mess around with
`GOPATH`. 🎊

Closes: https://github.com/openshift/machine-config-operator/issues/1185


ashcrow:
```
```

Unrelated to the change.

cgwalters:
Can i get a lgtm?

PR #1189, opened 2019-10-18
Bug 1762868: revert #1177 and fix common templates in MCs

runcom:
What #1177 had inadvertently regressed is that our MCs aren't rolled back anymore. This was sort of fine till the BZ came in, which exposed another bug we had. Basically, the code was adding duplicate and triplicates of common files all over. If you pair that with #1177, you'll get the bug in the title.

### DIFF
--- a/pkg/controller/template/template_controller.go
+++ b/pkg/controller/template/template_controller.go
@@ -403,9 +403,6 @@ func (ctrl *Controller) syncControllerConfig(key string) error {
 		if err := ctrl.syncRunningStatus(cfg); err != nil {
 			return err
 		}
-	} else {
-		glog.V(2).Info("ControllerConfig didn't change, skipping templates sync")
-		return nil
 	}
 
 	var pullSecretRaw []byte
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -104,6 +104,10 @@ func GenerateMachineConfigsForRole(config *RenderConfig, role, templateDir strin
 	}
 
 	cfgs := []*mcfgv1.MachineConfig{}
+	// This func doesn't process "common"
+	// common templates are only added to 00-<role>
+	// templates/<role>/{00-<role>,01-<role>-container-runtime,01-<role>-kubelet}
+	var commonAdded bool
 	for _, info := range infos {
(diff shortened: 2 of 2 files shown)


pkg/controller/template/template_controller.go
```diff
@@ -403,9 +404,10 @@ func (ctrl *Controller) syncControllerConfig(key string) error {
 		if err := ctrl.syncRunningStatus(cfg); err != nil {
 			return err
 		}
-	} else {
-		glog.V(2).Info("ControllerConfig didn't change, skipping templates sync")
-		return nil
+	}
+
+	if cfg.Spec.Version != version.Hash {
+		return ctrl.syncFailingStatus(cfg, fmt.Errorf("refusing to retemplate with different ControllerConfig and MCC versions: %s != %s", cfg.Spec.Version, version.Hash))
```
cgwalters: OK, so if the MCO has updated but the MCC hasn't, the MCC will block until it's updated.  Makes sense.

cgwalters:
And to help me understand - in order to fix the bug, we need to backport this all the way to 4.1, right?

runcom:
> And to help me understand - in order to fix the bug, we need to backport this all the way to 4.1, right?

I think 4.1 can live w/o the controllerconfig version fix (as it will compare against the empty string and still be correct)

As for the other common template fix, yeah, need proper backport till 4.1 but not as urgent maybe

pkg/controller/template/template_controller.go
runcom: this should guard also the case where, somehow, the new MCC is running with new templates but the controllerconfig is the old one and viceversa

runcom:
Good to go now with the revert to #1177 which is also needed for 4.2 (#1182) @mrunalp PTAL

/cherrypick release-4.2

runcom:
> Good to go now with the revert to #1177 which is also needed for 4.2 (#1182) @mrunalp PTAL
> 
> /cherrypick release-4.2

4.2 BZ is already here https://bugzilla.redhat.com/show_bug.cgi?id=1763205 for the cherry-pick

runcom:
/hold cancel
/skip

Good to go

PR #1192, opened 2019-10-18
pkg/controller: bring back old keys for image names

runcom:
To try and avoid failures in upgrades from 4.1 (???)


runcom:
/hold

this shipped in 4.2... so likely need a compatibility layer on top, will address once we confirm this is _fixing_ the issue somehow.

Issue #1190, opened 2019-10-18
consider making config changes truly transactional on RHCOS


cgwalters:
With ostree, each deployment (bootable target) has its own copy of `/etc`.

Today, the MCD writes into the *current* `/etc`.  This has weird side effects; for example, it means that we may be affecting running static pods for kubelet.  We may rewrite the pull secret.

It also means we don't have rollback at the ostree level.

It'd be fairly easy to change the MCD to (on RHCOS) create a *new* deployment for pure config changes, and write the new changes to `/etc` there - leaving the booted system untouched.

This would mean that config changes would be fully transactional and offline, the same way OS updates are.

cgwalters:
Another way to look at this is, it'd enforce that config changes require a reboot.  

We could drop the hacks around using journald for which config we're in, because `/etc` becomes transactional.  (And much closer to being immutable - see also https://github.com/projectatomic/rpm-ostree/issues/702 )

cgwalters:
Also a prerequisite is moving the pull secret from `/var` into `/etc` - and in general, all MachineConfig changes.

mrunalp:
:+1: IIUC, this would prevent new static pod manifests to take effect until we reboot.

sinnykumari:
This proposal looks interesting!
 One question though - will this apply on firstboot of node as well? This question is because today mcd available on host reads MachineConfig and writes `/etc/ignition-machine-config-encapsulated.json` during firstboot which is later used by mcd to apply kargs in firstboot.

cgwalters:
> One question though - will this apply on firstboot of node as well?

I don't see a blocker to doing this on firstboot too.

sinnykumari:
> > One question though - will this apply on firstboot of node as well?
> 
> I don't see a blocker to doing this on firstboot too.

Looks like I might have misunderstood little bit making /etc/ changes  transactional.

What will happen with this proposal in place during firstboot? Most likely during firstboot we will have config changes (like new file to be added in /etc/ignition-machine-config-encapsulated.json ) and also new machine-os-content. Today, we create new deployment for mchine-os-content changes which gets applied after reboot. What will happen to files like /etc/ignition-machine-config-encapsulated.json, will it be written to current deployment or in new deployment?

cgwalters:
The motivation for filing this issue was the [etcd upgrade bug](https://bugzilla.redhat.com/show_bug.cgi?id=1761557).

[...]

PR #1194, opened 2019-10-20
Bug 1763695: [release-4.2] pkg/daemon: drain before applying changes

runcom:
backport of #1200 

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -830,7 +830,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 	// take a stab at that and re-run the drain+reboot routine
 	if state.pendingConfig != nil && bootID == dn.bootID {
 		dn.logSystem("drain interrupted, retrying")
-		return dn.drainAndReboot(state.pendingConfig)
+		if err := dn.drain(); err != nil {
+			return err
+		}
+		return dn.finalizeAndReboot(state.pendingConfig)
 	}
 
 	if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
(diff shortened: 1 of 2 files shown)


cgwalters:
/approve
I am not sure why we didn't do this from the very start.  It seems a lot safer.
That said, I do think a longer term fix is https://github.com/openshift/machine-config-operator/issues/1190 - but that's harder (and actually would only apply to RHCOS, but we'd still want to drain classic RHEL first).

cgwalters:
Hm, is there no master equivalent of this PR?

cgwalters:
Also worth a least a summary of why we need this in the commit message - something like:

> Since the MCO owns writing kubelet static pod configurations, we need to drain *before* attempting to change them so that e.g. the etcd-quorum-guard's PDB has a chance to operate.  This would have also mitigated a bug in 4.1 to 4.2 upgrades - see https://github.com/openshift/machine-config-operator/pull/1177

runcom:
needs further discussion and approval + master PR etc etc (now BZ is wrong)

runcom:
/cherrypick release-4.1

4.1 BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=1763696

cgwalters:
Does this one need to have been explicitly created as a cherry-pick of the master PR?  Maybe it doesn't matter and we just need an architect to do the label dance.

runcom:
Patch manager will add the label to pick this up, we're waiting on this to let it soak a bit in master

eparis:
This will not be able to merge until BZ 1762536 has been VERIFIED by QE. And after that someone runs `/bugzilla refresh`

kikisdeliveryservice:
@runcom your cherry pick to 4.1 failed..

PR #1198, opened 2019-10-20
Bug 1763635: pkg/operator: fix race between images CM and MCO

runcom:
Full context here https://docs.google.com/document/d/1mBDXujKsQjf0SSP9abLE_Mczvc8KE9D8qmeyoGdrD5Y/edit# (will make it public asap)

holding as it needs a proper bugzilla and whole process as well if this is accepted.

Signed-off-by: Antonio Murdaca runcom@linux.com

### DIFF
--- a/install/0000_80_machine-config-operator_02_images.configmap.yaml
+++ b/install/0000_80_machine-config-operator_02_images.configmap.yaml
@@ -6,6 +6,7 @@ metadata:
 data:
   images.json: >
     {
+      "releaseVersion": "0.0.1-snapshot",
       "machineConfigOperator": "registry.svc.ci.openshift.org/openshift:machine-config-operator",
       "etcd": "registry.svc.ci.openshift.org/openshift:etcd",
       "infraImage": "registry.svc.ci.openshift.org/openshift:pod",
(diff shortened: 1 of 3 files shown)


runcom:
lifting the hold as there is enough evidence that this is the right approach and fix

/hold cancel

install/0000_80_machine-config-operator_02_images.configmap.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 data:
   images.json: >
     {
+      "releaseVersion": "0.0.1-snapshot",
```
cgwalters: Is there something that substitutes this value?  In `oc adm release new`?  I am not seeing it in `$ rg releaseVersion` in oc git master.

runcom: `oc adm release new` is substituting it according to my manual testing, if you can double check that , it would be amazing

runcom: also, grep for `0.0.1-snapshot` in openshift/origin - that's where the substitution is

mrunalp: https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/image_mapper.go#L317

mrunalp: Also, https://github.com/openshift/origin/pull/23626

PR #1203, opened 2019-10-22
Bug 1764116: templates: rename our dropins to include the mco string

runcom:
Mainly to avoid ppl to ship something which could override the MCO files.

The situation can still happen, in which case, the bad MC must be deleted and in order to rollback the maxUnavailable for the pool must be increased by at least 1

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -622,6 +622,13 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
 			if _, ok := newDropinSet[path]; !ok {
+				if _, err := os.Stat(origFileName(path)); err == nil {
+					if err := os.Rename(origFileName(path), path); err != nil {
+						return err
+					}
+					glog.V(2).Infof("Restored dropoin %q", path)
+					continue
+				}
 				glog.V(2).Infof("Deleting stale systemd dropin file: %s", path)
 				if err := os.Remove(path); err != nil {
 					newErr := fmt.Errorf("unable to delete %s: %s", path, err)
@@ -696,6 +703,9 @@ func (dn *Daemon) writeUnits(units []igntypes.Unit) error {
 		for i := range u.Dropins {
 			glog.Infof("Writing systemd unit dropin %q", u.Dropins[i].Name)
 			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
+			if err := createOrigFile(dpath); err != nil {
+				return err
+			}
 			if err := writeFileAtomicallyWithDefaults(dpath, []byte(u.Dropins[i].Contents)); err != nil {
 				return fmt.Errorf("failed to write systemd unit dropin %q: %v", u.Dropins[i].Name, err)
 			}
--- a/templates/common/_base/files/etc-systemd-system-crio.service.d-10-default-env.conf.yaml
+++ b/templates/common/_base/files/etc-systemd-system-crio.service.d-10-default-env.conf.yaml
@@ -1,6 +1,6 @@
 filesystem: "root"
 mode: 0644
-path: "/etc/systemd/system/crio.service.d/10-default-env.conf"
+path: "/etc/systemd/system/crio.service.d/10-mco-default-env.conf"
 contents:
   inline: |
     {{if .Proxy -}}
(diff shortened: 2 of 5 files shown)


runcom:
@cgwalters wondering why this wasn't the case till the beginning :/

runcom:
> @cgwalters wondering why this wasn't the case till the beginning :/

the actual issue here is that the rendered machineconfigs can contain duplicate entries for e.g. a service or unit. If that's the case, the validate routine can validate and fail only the first entry but what we have written on disk is the second one - should we change the validate routine to always check the last entry if there's a duplicate?  This PR just makes sure we can rollback but maybe the fix to validation is needed as well.

runcom:
(This PR isn't working as intended also, so keep holding)

runcom:
maybe we should completely avoid having users override what we ship?

cgwalters:
> @ajeddeloh since a rendered mc can contain duplicates entries, should we validate only against the last one? right now, we're validating and bailing at the first entry, but the MCD applies the last when writing to disk.

I believe it's a hard error to have duplicate files in Ignition spec 3.  We should also indeed disallow this in the MCO I'd say.

But the clear shorter term fix is to change our validation to check the last one for consistency with what we actually write.

runcom:
> I believe it's a hard error to have duplicate files in Ignition spec 3. We should also indeed disallow this in the MCO I'd say.

uhm, so then the MCC has to learn to generate rendered MCs by always using the last entry in alphabetical order? 🤔

cgwalters:
I think we should error out.

ajeddeloh:
> > @ajeddeloh since a rendered mc can contain duplicates entries, should we validate only against the last one? right now, we're validating and bailing at the first entry, but the MCD applies the last when writing to disk.

> I believe it's a hard error to have duplicate files in Ignition spec 3. We should also indeed disallow this in the MCO I'd say.

Correct that duplicates are disallowed in spec 3.0.0+. Spec 2.x is more complicated. The last one listed "wins" but only if overwrite is true. Also they're created in the order: directories, then files, then links. This means a file can never "win" against a link. This is also fixed with spec 3.

runcom:
> The last one listed "wins" but only if overwrite is true

here the scenario is "we write a file at a dropin location" then "a dropin writes on that location again" ouch - does spec v3 disallow this?

kikisdeliveryservice:
> I think we should error out.

Erroring out seems to be the cleanest choice, imo - it's predictable and clear to a user what they need to do going forward.

ajeddeloh:
> here the scenario is "we write a file at a dropin location" then "a dropin writes on that location again" ouch - does spec v3 disallow this?

Can you clarify? Do you mean a systemd drop in or an appended config?

runcom:
> Can you clarify? Do you mean a systemd drop in or an appended config?

yep, a systemd dropin.

ajeddeloh:
Hmm yeah I think we ought to check for that in Ignition as well actually. Failing seems like the right thing to do since the user is asking for the impossible.

runcom:
> But the clear shorter term fix is to change our validation to check the last one for consistency with what we actually write.

I don't believe this is the right thing to do now as a short term hack - the reason here writing a file at a location and writing a dropin config which later writes there. Allowing the validation to pass means that any configuration already shipped (for things like crio and kubelet) can be firstly overridden and secondly skipped from validation.
I believe this a broader issue also, how does someone ship a crio dropin? should we allow that since crio is controlled by the CRC crd? hence it makes sense to just error out when someone provides a configuration that would overwrite anything specified _before_.

What do you all think?

cgwalters:
> I believe this a broader issue also, how does someone ship a crio dropin? 

I think the problem here is that the user chose *the same name* for the dropin `10-default-env.conf` as we did.  Perhaps it'd be clearer if we named our drop-ins like `10-mco-default-env.conf` and hopefully users would not include `-mco` in their names.

runcom:
> and hopefully users would not include `-mco` in their names.

so if they do instead, I think we need to avoid rendering and communicate that, how does that sound?

runcom:
I'm updating this PR to move to `-mco-` for our dropins meanwhile - still doesn't solve the issue tho :(

runcom:
About this PR and Bug tho, the issue was mainly the inability to roll back and I think that's the case because the pools are hitting their maxUnavailable so bumping that to 2 will reconcile the cluster, I'm verifying that, we can then get the rename in and postpone any later discussion about validation when spec 3 will be in MCO maybe (?)

runcom:
Closed by https://github.com/openshift/machine-config-operator/pull/1715

PR #1205, opened 2019-10-22
Include candidate IP in log message.

russellb:
I hit this log message when doing some IPv6 testing.  I found that it
would have been helpful to know the IP address it was checking.
There's a verbose message a few lines later that would print it, but
it won't be hit in this error scenario.


cgwalters:
Limits looks like
/test e2e-gcp-upgrade

PR #1206, opened 2019-10-22
Bug 1763700: kubelet: add dependency on network-online.target

rphillips:
**- What I did**
This fixes a race when the kubelet enumerates the hostname at startup. Wait for the network-online.target to be ready before starting the Kubelet.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1763700

**- How to verify it**

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -3,8 +3,8 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service crio.service
-  After=crio.service
+  Wants=rpc-statd.service network-online.target crio.service
+  After=network-online.target crio.service
 
   [Service]
   Type=notify
--- a/templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
+++ b/templates/worker/01-worker-kubelet/_base/units/kubelet.yaml
@@ -3,8 +3,8 @@ enabled: true
 contents: |
   [Unit]
   Description=Kubernetes Kubelet
-  Wants=rpc-statd.service crio.service
-  After=crio.service
+  Wants=rpc-statd.service network-online.target crio.service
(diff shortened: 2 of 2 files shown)


cgwalters:
Also remember the installer has a copy of `kubelet.service` that needs updating too.  See e.g. https://github.com/openshift/installer/pull/2459

cgwalters:
Also xref:

 - https://github.com/cri-o/cri-o/issues/2478
 - https://github.com/cri-o/cri-o/pull/2662/
 - https://bugzilla.redhat.com/show_bug.cgi?id=1713478

cgwalters:
Given Kube dependency on the hostname (something that'd be good to fix eventually but)...this makes sense.  Did this really never come up before though?  Do other people's `kubelet.service` units do this too?

rphillips:
The cloud provider in the kubelet will gather the hostname from the metadata service. Namely, this fixes the issue with metal or other non-cloud based platforms.

@cgwalters typhoon waits for the resolve.conf to get set [1], but I don't love the implementation. 

1. https://github.com/poseidon/typhoon/blob/24fc440d837afceb189986c5375b68d8ce84ce50/bare-metal/container-linux/kubernetes/cl/worker.yaml.tmpl#L17

cgwalters:
> The cloud provider in the kubelet will gather the hostname from the metadata service (ie, aws, etc). 

:scream: There's another codebase doing that?

rphillips:
@cgwalters same code base as the kubelet, but each cloud provider does the hostname detection slightly different [[source]](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/).

rphillips:
```
level=error msg="Error: Error creating S3 bucket: timeout while waiting for state to become 'success' (timeout: 5m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-518815423/bootstrap/main.tf line 5, in resource \"aws_s3_bucket\" \"ignition\":"
level=error msg="   5: resource \"aws_s3_bucket\" \"ignition\" {"
```

PR #1210, opened 2019-10-23
build-sys: Generate Dockerfile.rhel7 from Dockerfile

cgwalters:
This finally merges all of the cleanups that went into our
main `Dockerfile` into the ART build, and should ensure the
two stay in sync from now on.


kikisdeliveryservice:
Really like this btw :)

PR #1211, opened 2019-10-23
etcd: Add initial support for an IPv6 control plane

russellb:
I'm working on bringing OpenShift up with an IPv6 control plane.  With these changes, the etcd cluster forms and the bootstrap process continues.  There are still some suspicious messages in the etcd-member log, but this seems like a good start.

The changes should have no impact to other installs unless the etcd DNS records are created with IPv6 addresses instead of IPv4.  Otherwise, there should be no changes in the resulting behavior.

This is not intended to be complete IPv6 support, just a start.  There's an `IPV4_ADDRESS` variable I left unchanged to cut down on the size of the patch.  I also noticed similar issues still need to be fixed in the recovery tools.

### DIFF
--- a/cmd/setup-etcd-environment/run.go
+++ b/cmd/setup-etcd-environment/run.go
@@ -338,10 +338,6 @@ func ipAddrs(preferredIP string) ([]string, error) {
 		if ip == nil {
 			continue
 		}
-		ip = ip.To4()
-		if ip == nil {
-			continue // not an ipv4 address
-		}
 		if !ip.IsGlobalUnicast() {
 			continue // we only want global unicast address
 		}
--- a/cmd/setup-etcd-environment/run.go
+++ b/cmd/setup-etcd-environment/run.go
@@ -179,7 +179,7 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 	unexportedEnv := map[string]string{
 		// TODO This can actually be IPv6, so we should rename this ...
 		"IPV4_ADDRESS":         setupEnv.etcdIP,
-		"ESCAPED_IP_ADDR":      escapedIP,
+		"ESCAPED_IP_ADDRESS":   escapedIP,
 		"ESCAPED_ALL_IPS":      escapedAllIPs,
 		"LOCALHOST_IP":         localhostIP,
 		"ESCAPED_LOCALHOST_IP": escapedLocalhostIP,
--- a/cmd/setup-etcd-environment/run.go
+++ b/cmd/setup-etcd-environment/run.go
@@ -156,9 +156,34 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	parsedIP := net.ParseIP(setupEnv.etcdIP)
+	if parsedIP == nil {
+		return fmt.Errorf("Failed to parse IP '%s'", setupEnv.etcdIP)
+	}
+
+	escapedIP := setupEnv.etcdIP
+	escapedAllIPs := "0.0.0.0"
+	localhostIP := "127.0.0.1"
+	escapedLocalhostIP := "127.0.0.1"
+	if parsedIP.To4() == nil {
+		// This is an IPv6 address, not IPv4.
+
+		// When using an IPv6 address in a URL, we must wrap the address portion in
+		// [::] so that a ":port" suffix can still be added and parsed correctly.
+		escapedIP = fmt.Sprintf("[%s]", setupEnv.etcdIP)
+		escapedAllIPs = "[::]"
+		localhostIP = "::1"
+		escapedLocalhostIP = "[::1]"
+	}
+
 	unexportedEnv := map[string]string{
-		"IPV4_ADDRESS":      setupEnv.etcdIP,
-		"WILDCARD_DNS_NAME": fmt.Sprintf("*.%s", setupEnv.opts.discoverySRV),
+		// TODO This can actually be IPv6, so we should rename this ...
+		"IPV4_ADDRESS":         setupEnv.etcdIP,
+		"ESCAPED_IP_ADDR":      escapedIP,
+		"ESCAPED_ALL_IPS":      escapedAllIPs,
+		"LOCALHOST_IP":         localhostIP,
+		"ESCAPED_LOCALHOST_IP": escapedLocalhostIP,
+		"WILDCARD_DNS_NAME":    fmt.Sprintf("*.%s", setupEnv.opts.discoverySRV),
 	}
 	if setupEnv.etcdDNS != "" {
 		unexportedEnv["DNS_NAME"] = setupEnv.etcdDNS
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -404,7 +404,7 @@ func etcdServerCertCommand(cfg RenderConfig) (interface{}, error) {
 			"  --assetsdir=/etc/ssl/etcd \\",
 			fmt.Sprintf("  --dnsnames=%s \\", serverCertDNS),
 			"  --commonname=system:etcd-server:${ETCD_DNS_NAME} \\",
-			"  --ipaddrs=${ETCD_IPV4_ADDRESS},127.0.0.1 \\",
+			"  --ipaddrs=${ETCD_IPV4_ADDRESS},${ETCD_LOCALHOST_IP} \\",
 		}...)
 	} else {
 		commands = append(commands, []string{
--- a/templates/common/_base/files/sysctl-forward-conf.yaml
+++ b/templates/common/_base/files/sysctl-forward-conf.yaml
@@ -4,4 +4,5 @@ path: "/etc/sysctl.d/forward.conf"
 contents:
   inline: |
     net.ipv4.ip_forward = 1
+    net.ipv6.conf.all.forwarding = 1
 
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -135,7 +135,7 @@ contents:
           set +a
 
           exec etcd \
-            --initial-advertise-peer-urls=https://${ETCD_ESCAPED_IP_ADDR}:2380 \
+            --initial-advertise-peer-urls=https://${ETCD_ESCAPED_IP_ADDRESS}:2380 \
             --cert-file=/etc/ssl/etcd/system:etcd-server:${ETCD_DNS_NAME}.crt \
             --key-file=/etc/ssl/etcd/system:etcd-server:${ETCD_DNS_NAME}.key \
             --trusted-ca-file=/etc/ssl/etcd/ca.crt \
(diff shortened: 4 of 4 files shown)


russellb:
This has been rebased.

@runcom any thoughts on this?  I think it should be safe.  There shouldn't be any changes in behavior in the normal case.

hexfusion:
I think we need to be careful about IPv6 until we get kube 1.16.3 there is a known bug[1] that we have resolved in etcd via 3.3.17[2] but won't get into the client (apiserver) until 1.16.3. I believe this is a hole in CI as AWS and GCP use ipv4 to my knowledge.

[1] https://github.com/kubernetes/kubernetes/issues/83550
[2] https://github.com/etcd-io/etcd/pull/11211/files#diff-e51eb0fcc3e32b5460ba6ed83b7399dbL235

russellb:
> I think we need to be careful about IPv6 until we get kube 1.16.3 there is a known bug[1] that we have resolved in etcd via 3.3.17[2] but won't get into the client (apiserver) until 1.16.3. I believe this is a hole in CI as AWS and GCP use ipv4 to my knowledge.
> 
> [1] [kubernetes/kubernetes#83550](https://github.com/kubernetes/kubernetes/issues/83550)
> [2] https://github.com/etcd-io/etcd/pull/11211/files#diff-e51eb0fcc3e32b5460ba6ed83b7399dbL235

Thanks for the pointer to the bug.  I haven't hit that one (yet?).

re: CI, indeed, there's no IPv6 CI yet.  That's on the todo list as soon as I can get enough of the cluster up and running.  I've an IPv6 cluster coming up partially on AWS with a bunch of changes (including the ones in this PR).  The etcd cluster forms using IPv6.  Right now I'm working through cluster SDN issues.  We'll have CI soon to help exercise this further.

hexfusion:
FTR 1.16.3 should land next week[1] so this won't be a prolonged issue for supporting IPv6.

[1]: https://github.com/kubernetes/kubernetes/issues/83550#issuecomment-550132544

cgwalters:
This...seems safer to do in 4.4 to me.  But I know that's annoying as it makes development now harder.

runcom:
> This...seems safer to do in 4.4 to me. But I know that's annoying as it makes development now harder.

I'm agreeing with this actually, we cannot take this PR anyway at this point tho. It looks pretty safeto me tho so when 4.4 opens it can go in

kikisdeliveryservice:
@hexfusion @alaypatel07 PTAL

hexfusion:
I picked 3.3.17 client into apiserver so I believe we are good here as 1.16.3 won't actually land in 4.3.

hexfusion:
/hold 

etcd-team would like a little time to manually test this and review. We are moving on this now.

hexfusion:
cc @alaypatel07 @retroflexer

templates/master/01-master-kubelet/_base/units/kubelet.yaml
```diff
@@ -22,6 +22,10 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
+{{- if .KubeletIPv6}}
+        --node-ip :: \
+        --address :: \
```
abhinavdahiya: --address 0.0.0.0
--
  | The  IP address for the Kubelet to serve on (set to 0.0.0.0 for all IPv4  interfaces and `::` for all IPv6 interfaces) (default 0.0.0.0)  (DEPRECATED: This parameter should be set via the config file specified  by the Kubelet's --config flag. See  https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/  for more information.)

templates/master/01-master-kubelet/_base/units/kubelet.yaml
```diff
@@ -22,6 +22,10 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
+{{- if .KubeletIPv6}}
+        --node-ip :: \
```
abhinavdahiya: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
```
--node-ip string
IP address of the node. If set, kubelet will use this IP address for the node
```

What does it mean to set the node-ip to `::` ?? will this break the apiserver to node communication?

russellb: @danwinship

danwinship: It requires a not-yet-merged upstream patch to kubelet. https://github.com/kubernetes/kubernetes/pull/85850/files

hexfusion:
@russellb general question, how can setup-etcd-env know `isSingleStackIPv6`? We need to be able to make that distinction before the control plane is up. Any ideas?

cmd/setup-etcd-environment/run.go
```diff
@@ -146,10 +146,35 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	parsedIP := net.ParseIP(ip)
+	if parsedIP == nil {
+		return fmt.Errorf("Failed to parse IP '%s'", ip)
+	}
```
abhinavdahiya: i think we should move this to https://github.com/openshift/machine-config-operator/pull/1211/files#diff-f26a1d5dc0cdd6c1f8863a9a3d7a3630R158 so that we are erroring out only when we need to... ?

hexfusion:
> @russellb general question, how can setup-etcd-env know `isSingleStackIPv6`? We need to be able to make that distinction before the control plane is up. Any ideas?

If all of the ips on the bootstrap node are ipv6 can we say the cluster is `isSingleStackIPv6` ?

cmd/setup-etcd-environment/run.go
russellb: Is there a case where we'd want to ignore a failure to parse the IP?  I know it wasn't checked before, but if this is a bad IP, the cluster is going to break badly anyway, right?

templates/master/01-master-kubelet/_base/units/kubelet.yaml
russellb: Thanks for pointing out that this was deprecated.  I missed that.

russellb:
> @russellb general question, how can setup-etcd-env know `isSingleStackIPv6`? We need to be able to make that distinction before the control plane is up. Any ideas?

The best thing to check at the moment is the cluster's SDN CIDR to see if it's IPv4 or IPv6.  That's supplied at install-config time.

russellb:
> > @russellb general question, how can setup-etcd-env know `isSingleStackIPv6`? We need to be able to make that distinction before the control plane is up. Any ideas?
> 
> If all of the ips on the bootstrap node are ipv6 can we say the cluster `isSingleStackIPv6` ?

nope ... in our AWS dev/test env, all nodes will have both IPv4 and IPv6 addresses, though we want to pretend it's single stack Ipv6 as much as possible.

runcom:
@russellb this needs a rebase. Also where do we stand with this? Especially on the etcd side /cc @hexfusion

russellb:
> @russellb this needs a rebase. Also where do we stand with this? Especially on the etcd side /cc @hexfusion

I'm looking at doing a rebase.  It conflicts with a change from @hexfusion that just says "cmd/setup-etcd-environment: refactor and add tests" but changes things that aren't obvious to me, so it'll definitely need his review...

russellb:
Rebase done, not yet tested though ... let's see what CI thinks!

russellb:
> might need a `make update`?

had a typo in a patch that broke the build ... `hack/build-go` didn't actually build all the go so I missed it in a test build.  :-D

cmd/setup-etcd-environment/run.go
```diff
@@ -156,9 +156,34 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+		localhostIP = "::1"
+		escapedLocalhostIP = "[::1]"
+	}
+
 	unexportedEnv := map[string]string{
-		"IPV4_ADDRESS":      setupEnv.etcdIP,
-		"WILDCARD_DNS_NAME": fmt.Sprintf("*.%s", setupEnv.opts.discoverySRV),
+		// TODO This can actually be IPv6, so we should rename this ...
+		"IPV4_ADDRESS":         setupEnv.etcdIP,
+		"ESCAPED_IP_ADDR":      escapedIP,
```
alaypatel07: s/ESCAPED_IP_ADDR/IP_ADDRESS

cmd/setup-etcd-environment/run.go
```diff
@@ -156,9 +156,34 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+		"ESCAPED_ALL_IPS":      escapedAllIPs,
```
alaypatel07: It would really help readability if instead of `ESCAPED_ALL_IPS` we have `LISTEN_CLIENT_URLS` and `LISTEN_PEER_URLS`

cmd/setup-etcd-environment/run.go
```diff
@@ -156,9 +156,34 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+		"LOCALHOST_IP":         localhostIP,
+		"ESCAPED_LOCALHOST_IP": escapedLocalhostIP,
```
alaypatel07: confused why we need both, is it possible to conditionally set LOCALHOST_IP and conditionally escape it if ipv6?

russellb: It's kind of annoying, because we only want to use it in its escaped form in some cases (in a URL), but not when it's provided as a raw IP, that's why it's there twice

cmd/setup-etcd-environment/run.go
russellb: made it ESCAPED_IP_ADDRESS

cmd/setup-etcd-environment/run.go
russellb: we'd need LISTEN_CLIENT_URLS, LISETN_PEER_URLS, LISTEN_METRIC_URLS, and METRICS_ADDR.  sure you want 4 variables instead of this 1?

russellb:
> @russellb Added comments inline. For context, we added changes to cluster-etcd-operator related to ipv6 recently, if possible can you change the naming convention of the env variables to something like this? [openshift/cluster-etcd-operator@57825f6](https://github.com/openshift/cluster-etcd-operator/commit/57825f6eaa0f77f5b6a9475c5a9acefb456c7f0a) this will help the etcd team in reading the code. Thanks

took a look at the PR and left a comment on one issue I spotted.

If you don't like the ESCAPED versions of variables in the PR, I could move that logic into shell code instead, I suppose.

russellb:
I've updated this PR to drop the kubelet config changes, which depended on a kubelet change to land first.  We can address that with a follow-up PR.  @danwinship 

I think this should be fine to merge now ...

kikisdeliveryservice:
@alaypatel07 @hexfusion PTAL

cmd/setup-etcd-environment/run.go
alaypatel07: @russellb thanks for pointing out, I think it would give clarity but 4 instead of 1 would be an overdo. Feel free to mark it resolved.

retroflexer:
I reviewed the changes and looks good to me.

/lgtm

kikisdeliveryservice:
@alaypatel07 @retroflexer please feel free to remove @hexfusion 's hold. this is ready to go.

kikisdeliveryservice:
To expedite, I'll just remove the hold myself. :)

/hold cancel

kikisdeliveryservice:
not sure why tide says it still needs and approved label when it clearly has one?  it's kind of wonky lately.

/approve

russellb:
well, not a clean cherry-pick, but I already have a 4.3 version of these changes in this branch: https://github.com/openshift-kni/machine-config-operator/commits/4.3-ipv6

PR #1214, opened 2019-10-24
Revert "build-sys: Add `make install`, use it in Dockerfile"

kikisdeliveryservice:
Reverts openshift/machine-config-operator#1204

After spending the better part of the afternoon unable to make images, I finally figured out that this PR broke it for me.  When I dropped the commit, I was able to run scripts correctly.

I hit many errors with the proxy.golang.org for ex: 
`go: github.com/onsi/ginkgo@v1.10.2: Get https://proxy.golang.org/github.com/onsi/ginkgo/@v/v1.10.2.info: dial tcp: lookup proxy.golang.org on 127.0.0.1:53: read udp 127.0.0.1:47112->127.0.0.1:53: read: connection refused`

As we are very close to freeze and this was a complete blocker to my epic work, I'd like to revert.  I don't have time to figure out why this isn't working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,16 @@ FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
 RUN make binaries
-# FIXME once we can depend on a new enough host that supports globs for COPY,
-# just use that.  For now we work around this by copying a tarball.
-RUN make install DESTDIR=./instroot && tar -C instroot -cf instroot.tar .
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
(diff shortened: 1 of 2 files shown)


sinnykumari:
I have recently seen this error while packaging latest MCD in brew where you can't download required go packages over network which is expected behavior in Brew buildsystem. To resolve that we explicitly passed `GOFLAGS=-mod=vendor` to `go build` in [spec file](https://github.com/openshift/machine-config-operator/blob/master/machine-config-daemon.spec#L25) so that it uses required packages from vendor/ directory. But this shouldn't be the case here unless `make image` was using dependency package from vendor/ instead of downloading latest one specified in go.mod . If that was the case `make image` should be failing since the the time we switched to go modules

kikisdeliveryservice:
Closing. Resolved.

cgwalters:
Sorry about missing this earlier,
/approve
on general principle.  What was the problem though?  Something unrelated networking?

PR #1215, opened 2019-10-24
templates: Let m-c-d binary on host process encapsulated MC on firstboot

sinnykumari:
During firstboot machine-config-daemon running on host will read
/etc/ignition-machine-config-encapsulated.json file available on
node, process it and applies MachineConfig provided like
kernelArguments

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-firstboot.service
+++ b/templates/common/_base/units/machine-config-daemon-firstboot.service
@@ -5,11 +5,8 @@ contents: |
   Description=Machine Config Daemon Firstboot
   # Make sure it runs only on OSTree booted system
   ConditionPathExists=/run/ostree-booted
-  # This effectively disables this unit unitl we get latest
-  # machine-config-daemon package into bootimage
-  ConditionPathExists=!/etc/pivot/image-pullspec
-  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
   BindsTo=ignition-firstboot-complete.service
+  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
   After=ignition-firstboot-complete.service
   Before=kubelet.service
 


sinnykumari:
/hold
This will go in once we have latest RHCOS bootimage available -  https://github.com/openshift/installer/pull/2547

runcom:
/retest

can't wait to test this out 👍

runcom:
I'm about to drop the hold once I run a 4.1->4.2 upgrade

sinnykumari:
Yay! Thank you all for testing and getting it merged.

cgwalters:
I think this may have broken upgrades from all "born 4.2" clusters (i.e. 4.2 bootimages) to 4.4, because 4.2 doesn't have `/usr/libexec/machine-config-daemon firstboot-complete-machineconfig`:
```
Apr 29 23:36:22 mffied-8v28p-w-c-z9vll.c.openshift-qe.internal machine-config-daemon[1306]: Error: unknown command "firstboot-complete-machineconfig" for "machine-config-daemon"
```

It won't affect 4.1 bootimages because that has `pivot.service` baked into the OS, which ends up running.  It won't affect 4.3 bootimages because that has `firstboot-complete-machineconfig` (which is likely what we tested with in this PR).

cgwalters:
OK sketching out an outline for a fix:

- Add infrastructure to in the MCO to look at `clusterversion` and determine if a cluster was "born" at 4.2
- Change the MCS to look at the cluster "born version" and if it's 4.2, change the Ignition it serves so that we basically `s/firstboot-complete-machineconfig/pivot` in [this template](https://github.com/openshift/machine-config-operator/blob/5396c45464fc1658c65bad980de0a386958fbf9a/templates/common/_base/units/machine-config-daemon-firstboot.service#L16)

cgwalters:
We *might* also be able to inspect the `User-Agent` version...I'm digging into that now.

runcom:
> We _might_ also be able to inspect the `User-Agent` version...I'm digging into that now.

with the user-agent, it would become just a matter of serving the modified template w/o extra plumbing 🤔

cgwalters:
Oooh, or I bet it would work if we generated two systemd units with conditionals!  Going to do a PR to sketch that out.

cgwalters:
https://github.com/openshift/machine-config-operator/pull/1706

PR #1216, opened 2019-10-24
Update and reduce templates for CRI-O 1.16

haircommander:
We have historically run into issues by the MCO setting more fields than it really needs to. Instead, we only really need to have the select fields MCO cares about, and comment out the rest.

**- What I did**
Override the default values with the ones MCO set in the past, then comment out the rest

**- How to verify it**
crio --config "" config > default.conf
crio --config $OLD_TEMPLATE config > mco.conf
sdiff default.conf mco.conf | grep '|'

and the resulting values are the only values uncommented here

Update crio.yaml templates to match CRI-O 1.16 config format as well as removed default values

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -3,254 +3,345 @@ mode: 0644
 path: "/etc/crio/crio.conf"
 contents:
   inline: |
-    # The "crio" table contains all of the server options.
+    # The CRI-O configuration file specifies all of the available configuration
+    # options and command-line flags for the crio(8) OCI Kubernetes Container Runtime
+    # daemon, but in a TOML format that can be more easily modified and versioned.
+    #
+    # Please refer to crio.conf(5) for details of all configuration options.
+
+    # CRI-O supports partial configuration reload during runtime, which can be
+    # done by sending SIGHUP to the running process. Currently supported options
+    # are explicitly mentioned with: 'This option supports live configuration
+    # reload'.
+
+    # CRI-O reads its storage defaults from the containers-storage.conf(5) file
+    # located at /etc/containers/storage.conf. Modify this storage configuration if
+    # you want to change the system's defaults. If you want to modify storage just
+    # for CRI-O, you can change the storage configuration options here.
     [crio]
 
-    # CRI-O reads its storage defaults from the containers/storage configuration
-    # file, /etc/containers/storage.conf. Modify storage.conf if you want to
-    # change default storage for all tools that use containers/storage.  If you
-    # want to modify just crio, you can change the storage configuration in this
-    # file.
-
-    # root is a path to the "root directory". CRIO stores all of its data,
-    # including container images, in this directory.
+    # Path to the "root directory". CRI-O stores all of its data, including
+    # containers images, in this directory.
     #root = "/var/lib/containers/storage"
 
-    # run is a path to the "run directory". CRIO stores all of its state
-    # in this directory.
-    #runroot = "/var/run/containers/storage"
+    # Path to the "run directory". CRI-O stores all of its state in this directory.
+    #runroot = "/run/user/1000"
 
-    # storage_driver select which storage driver is used to manage storage
-    # of images and containers.
-    #storage_driver = ""
+    # Storage driver used to manage the storage of images and containers. Please
+    # refer to containers-storage.conf(5) to see all available storage drivers.
+    #storage_driver = "overlay"
 
-    # storage_option is used to pass an option to the storage driver.
+    # List to pass options to the storage driver. Please refer to
+    # containers-storage.conf(5) to see all available storage options.
     #storage_option = [
     #]
 
-    # The "crio.api" table contains settings for the kubelet/gRPC interface.
+    # The default log directory where all logs will go unless directly specified by
+    # the kubelet. The log directory specified must be an absolute directory.
+    # log_dir = "/var/log/crio/pods"
+
+    # Location for CRI-O to lay down the version file
+    # version_file = "/var/lib/crio/version"
+
+    # The crio.api table contains settings for the kubelet/gRPC interface.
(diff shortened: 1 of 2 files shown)


templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
-      "SETGID", 
-      "SETUID", 
-      "SETPCAP", 
-      "NET_BIND_SERVICE", 
-      "SYS_CHROOT", 
-      "KILL", 
-    ]
+    # Path to the conmon binary, used for monitoring the OCI runtime.
+    # Will be searched for using $PATH if empty.
+    conmon = "/usr/libexec/crio/conmon"
```
cgwalters: Does this one still need explicit specification?

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
-    container_exits_dir = "/var/run/crio/exits"
+    # If true, SELinux will be used for pod separation on the host.
+    # selinux = true
 
-    # Path to directory for container attach sockets.
-    container_attach_socket_dir = "/var/run/crio"
+    # Path to the seccomp.json profile which is used as the default seccomp profile
+    # for the runtime. If not specified, then the internal default seccomp profile
+    # will be used.
+    seccomp_profile = "/etc/crio/seccomp.json"
```
cgwalters: And this?

haircommander: ah actually I don't think this one is needed. good catch

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: for the time being yes. This is a situation in which the CRI-O rpm needs to move in lock step with MCO. I guess we can have the rpm put it in both spots (it may already I have to check). Once that happens, we can drop this, and eventually drop it from /usr/libexec all together

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: fixed

haircommander:
@umohnani8 @mtrmac PTAL

cgwalters:
I think we should do this now to make future upgrades easier.  Code overall
/lgtm

(Though do we need separate crio.conf on masters vs workers?  Could we move it into `common/`?  But not blocking this PR)

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
     [crio]
 
-    # CRI-O reads its storage defaults from the containers/storage configuration
-    # file, /etc/containers/storage.conf. Modify storage.conf if you want to
-    # change default storage for all tools that use containers/storage.  If you
-    # want to modify just crio, you can change the storage configuration in this
-    # file.
+    # Path to the "root directory". CRI-O stores all of its data, including
+    # containers images, in this directory.
+    #root = "/home/pehunt/.local/share/containers/storage"
```
umohnani8: This doesn't look right. Should be `/var/lib/containers/storage`

umohnani8: oh nvm, it is commented out. But we should probably still fix it.

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
-      "KILL", 
-    ]
+    # Path to the conmon binary, used for monitoring the OCI runtime.
+    # Will be searched for using $PATH if empty.
+    conmon = "/usr/libexec/crio/conmon"
 
-    # hooks_dir_path is the oci hooks directory for automatically executed hooks
-    hooks_dir_path = "/usr/share/containers/oci/hooks.d"
+    # Cgroup setting for conmon
+    # conmon_cgroup = "pod"
```
umohnani8: since this is not `system.slice`, we should it update it here.

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
     [crio]
 
-    # CRI-O reads its storage defaults from the containers/storage configuration
-    # file, /etc/containers/storage.conf. Modify storage.conf if you want to
-    # change default storage for all tools that use containers/storage.  If you
-    # want to modify just crio, you can change the storage configuration in this
-    # file.
+    # Path to the "root directory". CRI-O stores all of its data, including
+    # containers images, in this directory.
+    #root = "/home/pehunt/.local/share/containers/storage"
```
umohnani8: Same as above here.

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
```diff
@@ -3,254 +3,346 @@ mode: 0644
...
-      "KILL", 
-    ]
+    # Path to the conmon binary, used for monitoring the OCI runtime.
+    # Will be searched for using $PATH if empty.
+    conmon = "/usr/libexec/crio/conmon"
 
-    # hooks_dir_path is the oci hooks directory for automatically executed hooks
-    hooks_dir_path = "/usr/share/containers/oci/hooks.d"
+    # Cgroup setting for conmon
+    # conmon_cgroup = "pod"
```
umohnani8: move to `system.slice` here as well

cgwalters:
/hold
per review comments
(But feel free to lift and do followup commits; I think most of that is cosmetic)

haircommander:
comments addressed :)

umohnani8:
LGTM

haircommander:
:upside_down_face:

PR #1222, opened 2019-10-29
give etcd-metrics container privilege

haircommander:
**- What I did**
Explicitally give etcd-metrics container privilege, as it is trying to run /run/etc/environment on the host, but wasn't given permissions to do so

### DIFF
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -176,6 +176,8 @@ contents:
         - name: metric
           containerPort: 9979
           protocol: TCP
+        securityContext:
+          privileged: true
       hostNetwork: true
       priorityClassName: system-node-critical
       tolerations:


mrunalp:
@cgwalters @runcom ptal.

runcom:
@rphillips ptal (Sam's out also, not sure who to ping from etcd, but at some point this priviledged bit was dropped because it had a typo and now this is getting back)

runcom:
cross linking the original change that dropped the privileged flag https://github.com/openshift/machine-config-operator/pull/526

rphillips:
lgtm from me. This is related to the CRIO 1.16 bump in progress and the Selinux changes found within that bump.

runcom:
so talked with @mrunalp - here's why we need this PR now (@haircommander would be nice to update the commit message with this):

- CRI-O 1.14 would turn privileged on for every container in a static pod as long as just one had privileged on. That's why etcd was still running in privileged in 1.14 even w/o the flag, see #526 
- CRI-O 1.16 (kube?) has disabled that now and we need to set that for every container in a static pod

runcom:
> @runcom updated. I am pretty sure it's kube that is doing this. CRI-O has no understanding of static pods, and only gives containers privilege when it's told to by kubelet

yeah, s/CRIO/kube/g sounds fine to me 👍 

/lgtm

PR #1224, opened 2019-10-29
cmd/machine-config-operator/bootstrap: add cluster-etcd-operator flag

hexfusion:
This PR adds a flag to bootstrap allowing the `cluster-etcd-operator` image to be be used in the etcd-member spec. We are also using the value of the flag as conditional logic for the spec. Because we need to balance that with what is put in etcd toggling CEO will involve unsetting the flag and the value of the configmap.

We also add a `readinessProbe` to etcd-member container. etcd should not show ready in service until it can serve traffic.

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -34,6 +34,7 @@ var (
 		keepalivedImage           string
 		kubeCAFile                string
 		kubeClientAgentImage      string
+		clusterEtcdOperatorImage  string
 		mcoImage                  string
 		mdnsPublisherImage        string
 		oauthProxyImage           string
@@ -63,6 +64,7 @@ func init() {
 	bootstrapCmd.MarkFlagRequired("etcd-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.kubeClientAgentImage, "kube-client-agent-image", "", "Image for Kube Client Agent.")
 	bootstrapCmd.MarkFlagRequired("kube-client-agent-image")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.clusterEtcdOperatorImage, "cluster-etcd-operator-image", "", "Image for Cluster Etcd Operator. An empty string here means the cluster boots without CEO.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraImage, "infra-image", "", "Image for Infra Containers.")
 	bootstrapCmd.MarkFlagRequired("infra-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.configFile, "config-file", "", "ClusterConfig ConfigMap file.")
@@ -101,6 +103,7 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 			Etcd:                bootstrapOpts.etcdImage,
 			InfraImage:          bootstrapOpts.infraImage,
 			KubeClientAgent:     bootstrapOpts.kubeClientAgentImage,
+			ClusterEtcdOperator: bootstrapOpts.clusterEtcdOperatorImage,
 			Keepalived:          bootstrapOpts.keepalivedImage,
 			Coredns:             bootstrapOpts.corednsImage,
 			MdnsPublisher:       bootstrapOpts.mdnsPublisherImage,
--- a/install/0000_80_machine-config-operator_02_images.configmap.yaml
+++ b/install/0000_80_machine-config-operator_02_images.configmap.yaml
@@ -11,6 +11,7 @@ data:
       "etcd": "registry.svc.ci.openshift.org/openshift:etcd",
       "infraImage": "registry.svc.ci.openshift.org/openshift:pod",
       "kubeClientAgentImage": "registry.svc.ci.openshift.org/openshift:kube-client-agent",
+      "clusterEtcdOperatorImage": "",
       "keepalivedImage": "registry.svc.ci.openshift.org/openshift:keepalived-ipfailover",
       "corednsImage": "registry.svc.ci.openshift.org/openshift:coredns",
       "mdnsPublisherImage": "registry.svc.ci.openshift.org/openshift:mdns-publisher",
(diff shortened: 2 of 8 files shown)


cmd/machine-config-operator/bootstrap.go
```diff
@@ -63,6 +64,7 @@ func init() {
 	bootstrapCmd.MarkFlagRequired("etcd-image")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.kubeClientAgentImage, "kube-client-agent-image", "", "Image for Kube Client Agent.")
 	bootstrapCmd.MarkFlagRequired("kube-client-agent-image")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.clusterEtcdOperatorImage, "cluster-etcd-operator-image", "", "Image for Cluster Etcd Operator. An empty string here means the cluster boots without CEO.")
```
kikisdeliveryservice: i think you will need to update installer to pass this in?

cmd/setup-etcd-environment/run.go
```diff
@@ -12,11 +12,12 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
```
kikisdeliveryservice: drop

pkg/operator/render_test.go
```diff
@@ -2,9 +2,10 @@ package operator
 
 import (
 	"fmt"
-	configv1 "github.com/openshift/api/config/v1"
 	"strings"
 	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
```
kikisdeliveryservice: drop

pkg/server/server_test.go
```diff
@@ -11,10 +11,10 @@ import (
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 	yaml "github.com/ghodss/yaml"
-	"github.com/stretchr/testify/assert"
 	v1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
+	"github.com/stretchr/testify/assert"
```
kikisdeliveryservice: drop

cmd/machine-config-operator/bootstrap.go
kikisdeliveryservice: see https://github.com/openshift/installer/pull/1847

hexfusion: yep we are just working through the maze here. My hopes are that because it is not a required flag we can sneak in it into MCO before the installer.

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -109,8 +168,6 @@ contents:
             --listen-client-urls=https://0.0.0.0:2379 \
             --listen-peer-urls=https://0.0.0.0:2380 \
             --listen-metrics-urls=https://0.0.0.0:9978 \
-        securityContext:
-          privileged: true
```
hexfusion: going to fix these

cmd/machine-config-operator/bootstrap.go
kikisdeliveryservice: looking at my past images stuff: without required should be ok!

hexfusion:
using for test release for debug 

/test e2e-aws

hexfusion:
FTR the failure has been isolated and will be resolved shortly

kikisdeliveryservice:
fyi: that aws-scaleup-rhel7 is basically broken. i filed a bz last night.  so you can just:

/skip

hexfusion:
We have found a bug where using if statements in spec where on disk the machine config renders correctly but in etcd the If statement was rendered in reverse. I will provide an example shortly.

hexfusion:
```
level=error msg="Error: Error creating DNS RecordSet: googleapi: Error 503: Backend Error, backendError"
level=error
level=error msg="  on ../tmp/openshift-install-025962353/dns/base.tf line 39, in resource \"google_dns_record_set\" \"etcd_a_nodes\":"
level=error msg="  39: resource \"google_dns_record_set\" \"etcd_a_nodes\" {"
```
/test e2e-gcp-upgrade

install/0000_80_machine-config-operator_02_images.configmap.yaml
```diff
@@ -11,6 +11,7 @@ data:
       "etcd": "registry.svc.ci.openshift.org/openshift:etcd",
       "infraImage": "registry.svc.ci.openshift.org/openshift:pod",
       "kubeClientAgentImage": "registry.svc.ci.openshift.org/openshift:kube-client-agent",
+      "clusterEtcdOperatorImage": "",
```
hexfusion: to be clear this is intentional and will be added in a separate PR when we enable the operator.

hexfusion:
> mcd_test.go:555: pool worker didn't report updated to rendered-worker-a4935ca22d7431a704c741eba9e4d563: timed out waiting for the condition

Doesn't feel related

/retest

hexfusion:
> level=info msg="Cluster operator authentication Progressing is Unknown with NoData: "

/test e2e-gcp-op

hexfusion:
https://github.com/openshift/machine-config-operator/pull/1232#issuecomment-548880739

FIPS is broken I guess?

hexfusion:
cc @cgwalters

kikisdeliveryservice:
hitting same error as here: https://github.com/openshift/machine-config-operator/pull/1232#issuecomment-548774819

we are trying to get a fix

kikisdeliveryservice:
@cgwalters do you think we can do an override for this as it's blocking other PRs in the bunch?

cgwalters:
Yeah,
/approve
And I'm OK overriding, you want to pull the trigger?

kikisdeliveryservice:
thanks bot for re-running all the tests again, including the one we told you not to run...

/override ci/prow/e2e-gcp-op

kikisdeliveryservice:
i think it has to fail before i can override, ill check in later

PR #1228, opened 2019-10-30
[fcos] mcd-pull: fix command to extract mcd on the hosts

vrutkovs:
This fixed mcd-pull service command, previous failed with `/usr/bin/cat: 
cannot execute binary file`

/cc @LorbusChris

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-pull.service
+++ b/templates/common/_base/units/machine-config-daemon-pull.service
@@ -13,7 +13,7 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/bin/sh -c "/usr/bin/podman run --quiet --net=host --entrypoint=sh '{{ .Images.setupEtcdEnvKey }}' cat /usr/bin/machine-config-daemon > /usr/local/bin/machine-config-daemon"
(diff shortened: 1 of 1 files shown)


templates/common/_base/units/machine-config-daemon-pull.service
```diff
@@ -13,7 +13,7 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/bin/sh -c "/usr/bin/podman run --quiet --net=host --entrypoint=sh '{{ .Images.setupEtcdEnvKey }}' cat /usr/bin/machine-config-daemon > /usr/local/bin/machine-config-daemon"
+  ExecStart=/usr/bin/podman run --quiet --net=host --entrypoint=cat '{{ .Images.setupEtcdEnvKey }}' /usr/bin/machine-config-daemon > /usr/local/bin/machine-config-daemon
```
cgwalters: :smile_cat: :heavy_check_mark:

smarterclayton:
If you have machine-os-content being built on PRs like Origin you could iterate here now.  @LorbusChris I think we should do that while iterating here, otherwise we're wasting time having to merge before we can test.

PR #1231, opened 2019-10-31
Bug 1766356: templates: tell NetworkManager to ignore interfaces

squeed:
The recent os bump includes the OVS plugin for NetworkManager, causing it to manage interfaces that were previously left unmanaged.

This functionality will be used, so we can't tell it to ignore all interfaces of type openvswitch. So, add the interfaces created by openshift-sdn and ovn-kubernetes to the Unmanaged list.

### DIFF
--- a/templates/common/_base/files/nm-ignore-sdn.yaml
+++ b/templates/common/_base/files/nm-ignore-sdn.yaml
@@ -1,0 +1,9 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/NetworkManager/conf.d/sdn.conf"
+contents:
+  inline: |
+    # ignore known SDN-managed devices
+    [device]
+    match-device=interface-name:br-int;interface-name:br-local;interface-name:br-nexthop,interface-name:k8s-*;interface-name:tun0;interface-name:br0;driver:veth
+    managed=0


squeed:
FYI @dcbw

cgwalters:
I think longer term rather than that inline matching which is somehow reminiscent of [modem commands](https://en.wikipedia.org/wiki/Hayes_command_set) we should do something like have a way to attach metadata about an "owner" to network interfaces, so that not-NetworkManager processes could signal to NM directly about this.

A somewhat related situation for example is the default podman/docker bridge devices, which at one point NM saw and said "oh hey, an ethernet device, let me DHCP on it".  

Maybe in udev somehow?  Or maybe patch the kernel to add xattrs on sysfs?

squeed:
> FWIW, in theory this could (and possibly should) go directly in RHCOS too, but...we don't have a nice convenient ci-gated place for that.

I don't have a good intuition for where that line should fall. Since third-party network providers could theoretically have to mirror this (using the correct CRs), keeping it open seems like being a good citizen.

> Maybe in udev somehow?

That might work. Honestly, any networking device created by `ip dev add` should probably be unmanaged by default. I wonder if NetworkManager lets you override another selector.

For something as complicated as an ovs bridge, I really can't think of a specific rule. We, as an SDN provider, want full control of our interface. The CNV people, who *also* create ovs bridges, want those interfaces to be managed normally. At this point, branching on the name is the only thing I can think of.

In my days as CNI maintainer, I've definitely wanted the ability to set arbitrary metadata on an interface...

squeed:
any hints on how to debug this?

cgwalters:
It's not this PR, same symptom here https://github.com/openshift/machine-config-operator/pull/1232#issuecomment-548774819
Something broke here recently elsewhere, will ping the network-edge folks.

squeed:
Gotcha. If there's anything I can do to speed this along, let me know.

squeed:
Rebased on master, added `driver:veth` to the list of ignores.
@dcbw please review

kikisdeliveryservice:
level=error msg="Error: Error waiting for instance (i-02e9b659694cad6cf) to become ready: Failed to reach target state. Reason: Server.InternalError: Internal error on launch"
level=error
level=error msg="  on ../tmp/openshift-install-395096181/master/main.tf line 93, in resource \"aws_instance\" \"master\":"
level=error msg="  93: resource \"aws_instance\" \"master\" {" 

/test e2e-aws

PR #1232, opened 2019-10-31
remove make binaries from Dockerfiles and use -mod=vendor when building

runcom:
Fixes an issue with builds in ART as well, please look at individual commit messages.

cc @cgwalters

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
-RUN make binaries
 # FIXME once we can depend on a new enough host that supports globs for COPY,
 # just use that.  For now we work around this by copying a tarball.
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
gcp iam failure?  reported

/test e2e-gcp-upgrade

cgwalters:
Hm, it looks like that e2e-gcp-op test failed because it couldn't evict the router:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1232/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/235/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-8z6x2_machine-config-daemon.log

I can't imagine that this PR broke it so
/test e2e-gcp-op

kikisdeliveryservice:
this failed test is killing us...

/test e2e-gcp-op

cgwalters:
I think we need to get to the bottom of why the router pod isn't evicting.
That said...I think we should just get this one in with an override.  Any opposition?

PR #1233, opened 2019-10-31
daemon: Refuse to disable FIPS mode

cgwalters:
Our new thought around this is that really FIPS should be a "day 1"
operation, and we don't want to make it really easy to undo.
See also https://github.com/openshift/installer/pull/2594
 Anyone who wants to force this can change the MC flag, then
`oc debug node` and run the disable command by hand, then reboot.

Our MachineConfig merge semantics should make it hard for this
to happen unless the admin explicitly deletes the installer-generated MC,
but still.

Since we don't support it and don't want customers
to do it by accident, let's disable it and also stop wasting compute
hours testing it.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -328,6 +328,23 @@ func (dn *Daemon) updateFIPS(current, desired *mcfgv1.MachineConfig) error {
 	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
 		return errors.New("Updating FIPS on non-RHCOS nodes is not supported")
 	}
+	// Our new thought around this is that really FIPS should be a "day 1"
+	// operation, and we don't want to make it really easy to undo.
+	// See also https://github.com/openshift/installer/pull/2594
+	// Anyone who wants to force this can change the MC flag, then
+	// `oc debug node` and run the disable command by hand, then reboot.
+	if current.Spec.FIPS && !desired.Spec.FIPS {
+		return errors.New("Refusing to undo FIPS mode")
+	}
+	// At this point, we must be trying to enable FIPS, since
+	// current != desired && desired per conditionals above
+	if _, err := os.Stat(fipsCommand); err != nil {
+		if os.IsNotExist(err) {
+			return errors.New("Cannot enable FIPS after firstboot")
+		}
+		return errors.Wrapf(err, "Checking FIPS")
+	}
+
 	arg := "enable"
 	if !desired.Spec.FIPS {
 		arg = "disable"


ashcrow:
This makes sense to me. Should we refuse to let them change FIPS at all since Day 2 wouldn't be valid?

runcom:
> This makes sense to me. Should we refuse to let them change FIPS at all since Day 2 wouldn't be valid?

yeah, I was actually thinking if day2 isn't valid at all we might want to go the route to completely drop this FIPS from MC and either you install with fips or not. That would be the safest option to me w/o providing any way to disable it through the MCO

cgwalters:
> Should we refuse to let them change FIPS at all since Day 2 wouldn't be valid?

Possibly but we *currently* rely on this code since we haven't landed an initramfs replacement yet.

ashcrow:
/override ci/prow/e2e-gcp-op

I've been told there are currently some issues being looked at for the above job. Skipping.

cgwalters:
I just noticed this in the logs:
```
I1031 20:38:14.869340       1 node_controller.go:433] Pool worker: node ci-op--bgjrr-w-d-lpplg.c.openshift-gce-devel-ci.internal is now reporting unready: node ci-op--bgjrr-w-d-lpplg.c.openshift-gce-devel-ci.internal is reporting OutOfDisk
```

kikisdeliveryservice:
in the failed run it took almost and hour to evict arouter pod:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1233/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/231/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-xsg8p_machine-config-daemon.log

kikisdeliveryservice:
@cgwalters i noticed in the fips e2e they were hitting OOD on 2 nodes in the failed runs as well.  sent you the link to chat convo

crawford:
Are we concerned about transitions in the opposite direction? Going from non-FIPS compliant to (technically still not) FIPS compliant seems more dangerous.

cgwalters:
@crawford  Yes but see
https://github.com/openshift/machine-config-operator/pull/1233#issuecomment-548526302
(We need this right now to test FIPS at all until we land the updated RHCOS)

cgwalters:
@ashcrow You added a hold - can you explain why?

ashcrow:
@cgwalters I added it as @kikisdeliveryservice wanted to review why the GCE job failed to ensure it was a flake.

kikisdeliveryservice:
> @ashcrow You added a hold - can you explain why?

held bc we saw the failed FIPs test and wanted to investigate it (ie the stuff we've been looking at)


@cgwalters  If you are fine with merging it without it passing that job at all, I'm fine with removing hold. I dont know enough about FIPs to say so myself... just lmk

cgwalters:
Hmm.  I think this PR is quite safe, and will actually help our issues with the e2e-gcp-op test suite since it'll shorten it a bit.

But, it can also probably wait until we've figured out that suite.

kikisdeliveryservice:
It's late for me, but I've figured out that our test is failing precisely bc FIPS day2 is broken.  I have a PR here removing the test to unblock the PRs that have been stuck since Friday:
See #1244 

Can someone in the morning just double check and then merge this (and then approve my PR)?  We have features blocked so it would be great to get the test gone and things back to working.

ashcrow:
Needs a quick rebase

cgwalters:
OK rebased :surfing_man: and I also added some code to more cleanly handle the [pending RHCOS change](https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/680/diffs) to drop `rhcos-tools`.

ashcrow:
/override ci/prow/e2e-aws-scaleup-rhel7

My understanding is the above failing test is unrelated

kikisdeliveryservice:
GCP errors

/test e2e-gcp-upgrade

PR #1234, opened 2019-10-31
machineconfig.crd.yaml: add passwd schema

ericavonb:
Better CRD validation via OpenAPIV3 schema. Taken from https://github.com/coreos/ignition/blob/spec2x/schema/ignition.json

### DIFF
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -70,6 +70,22 @@ spec:
                     version:
                       description: "version string is the semantic version number of the spec"
                       type: string
+                passwd:
+                  type: object
+                  properties:
+                    users:
+                      type: array
+                      items:
+                        type: object
+                        properties:
+                          name:
+                            description: "name of user. Must be \"core\" user."
+                            type: string
+                          sshAuthorizedKeys:
+                            description: "public keys to be assigned to user core"
+                            type: array
+                            items:
+                              type: string
                 storage:
                   description: "storage describes the desired state of the system's storage devices"
                   type: object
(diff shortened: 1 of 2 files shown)


ericavonb:
/assign @kikisdeliveryservice 
@kikisdeliveryservice these look like the right set of supported fields?

kikisdeliveryservice:
@ericavonb Correct.  The only part of passwd we support is users and only the user core and only property is SSH keys.

ericavonb:
@kikisdeliveryservice I have a feeling this is going to get caught up in whatever 🔥 CI tests are going through today. Can you look over and give a lgtm if it looks give minus CI tests passing?

kikisdeliveryservice:
Will test with a cluster this afternoon to double check.

kikisdeliveryservice:
@ericavonb do you have a sample mc that i can start with (so i can go faster) + expected output

ericavonb:
Bad- `oc apply...` should fail (from https://access.redhat.com/solutions/4073041):
```yaml
kind: MachineConfig
metadata:
  labels:
  name: 00-master-ssh-02
spec:
  config:
    passwd:
      users:
      - name: core
        sshAuthorizedKeys: |
          <<ssh-public rsa key goes here>>
```

Good- `oc apply...` should succeed: 
```yaml
kind: MachineConfig
metadata:
  labels:
  name: 50-master-ssh-02
spec:
  config:
    passwd:
      users:
      - name: core
        sshAuthorizedKeys:
          - |
            <<ssh-public rsa key goes here>>
```

kikisdeliveryservice:
Looks good! Output from my bogus config:

```$ oc create -f evb-config.yaml 
The MachineConfig "sssh-test" is invalid: spec.config.passwd.users.sshAuthorizedKeys: Invalid value: "string": spec.config.passwd.users.sshAuthorizedKeys in body must be of type array: "string"``

/lgtm

kikisdeliveryservice:
gah aws not again..

/retest

PR #1236, opened 2019-11-01
[fcos] templates: place /etc/pivot/kernel-args on masters/workers

vrutkovs:
This would disable cgroupsv2 and Spectre mitigations. These files are served by MCD to ensure correct initial hash on masters is being calculated

### DIFF
--- a/templates/common/_base/files/etc-pivot-kernel-args.yaml
+++ b/templates/common/_base/files/etc-pivot-kernel-args.yaml
@@ -1,0 +1,7 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/pivot/kernel-args"
+contents:
+  inline: |
+    DELETE mitigations=auto,nosmt
+    ADD systemd.unified_cgroup_hierarchy=0


vrutkovs:
/hold

Testing it locally

vrutkovs:
/hold cancel

Works fine here

cgwalters:
This should use the [encapsulated MachineConfig](https://github.com/openshift/machine-config-operator/pull/868) with `kernelArguments`.

vrutkovs:
>This should use the encapsulated MachineConfig with kernelArguments.

I was trying to add a new MC via installer, but that made node annotations to have a wrong rendered machine config hash. Eventually all this would be cleaned up

ericavonb:
> I was trying to add a new MC via installer, but that made node annotations to have a wrong rendered machine config hash.
That shouldn't happen. How were you adding the new machineconfig?

vrutkovs:
>That shouldn't happen. How were you adding the new machineconfig?

Via the installer - https://github.com/openshift/installer/pull/2548/commits/1f837e796ed5fc9d4a7199b055b8b1b13bcb8211

LorbusChris:
ok, let's do this for now. We can do any clean ups/polishing of this a bit further down the road.

/lgtm

vrutkovs:
I agree this is certainly a bad hack and would be reverted, but without e2e its hard to come up with a better fix

PR #1243, opened 2019-11-04
daemon: leave backup copy of encapsulated machineconfig

ericavonb:
Rename rather than remove the encapsulated MachineConfig file after processing
on firstboot. This allows users to check what data was processed on the node
after the machineconfigdaemon-firstboot.service has run successfully.


ericavonb:
/assign @sinnykumari 
cc @cgwalters

PR #1245, opened 2019-11-04
controller: Log the exact status in conditions

cgwalters:
We're seeing `OutOfDisk` in some recent runs and I'm wondering
whether it's `ConditionUnknown` appearing recently.  This should
help us find out.


cgwalters:
Yep, see this
`I1105 05:16:43.120872       1 node_controller.go:433] Pool worker: node ci-op--bzz4h-w-d-42qjt.c.openshift-gce-devel-ci.internal is now reporting unready: node ci-op--bzz4h-w-d-42qjt.c.openshift-gce-devel-ci.internal is reporting OutOfDisk=Unknown`

Uncertain about us marking a node as unready if `OutOfDisk` is just `Unknown`.  But...eh.  I think we know now that it was a red herring at least.

PR #1247, opened 2019-11-05
*: enable cluster-etcd-operator

hexfusion:
Currently, we use the presence/existence of the clusterEtcdOperatorImageKey flag value and the value of the clusterEtcdOperatorImage [1] [2] as a condition for using the old style SRV bootstrap or enable cluster-etcd-operator to scale etcd. This PR would enable 4.3 scaling for MCO.

- https://github.com/openshift/machine-config-operator/blob/release-4.3/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml#L22

- https://github.com/openshift/machine-config-operator/compare/master...hexfusion:enable_ceo?expand=1#diff-cbbafda65814e4ab5960f98e6b8c109fR14


alaypatel07:
/lgtm 

apart from the hold, this lgtm

kikisdeliveryservice:
Required installer pr: https://github.com/openshift/installer/pull/2608

hexfusion:
We will try this again in 4.4

Issue #1250, opened 2019-11-07
Booting in FIPS mode day-1 causes errors on firstboot


jlebon:
```
Nov 07 17:12:13 fips-test-k7v7d-master-0.fips-test.jlebon.testing systemd[1]: Starting Machine Config Daemon Firstboot...
Nov 07 17:12:13 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:13.508185    2031 rpm-ostree.go:365] Running captured: rpm-ostree status --json
Nov 07 17:12:14 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:14.133618    2031 daemon.go:209] Booted osImageURL:  (43.81.201911071657.0)
Nov 07 17:12:14 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:14.134321    2031 update.go:233] Checking Reconcilable for config mco-empty-mc to rendered-master-db1384e29eb0129eab71e6600e7f291f
Nov 07 17:12:14 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:14.134370    2031 update.go:1065] Starting update from mco-empty-mc to rendered-master-db1384e29eb0129eab71e6600e7f291f: &{osUpdate:true kargs:false fips:true passwd:>
Nov 07 17:12:14 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:14.149312    2031 update.go:592] Updating files
Nov 07 17:12:14 fips-test-k7v7d-master-0.fips-test.jlebon.testing machine-config-daemon[2031]: I1107 17:12:14.149331    2031 update.go:611] Deleting stale data
Nov 07 17:12:14 fips-test-k7v7d-master-0 machine-config-daemon[2031]: I1107 17:12:14.296912    2031 update.go:592] Updating files
Nov 07 17:12:14 fips-test-k7v7d-master-0 machine-config-daemon[2031]: I1107 17:12:14.296935    2031 update.go:611] Deleting stale data
Nov 07 17:12:14 fips-test-k7v7d-master-0 machine-config-daemon[2031]: error: Cannot enable FIPS after firstboot
[...]

PR #1252, opened 2019-11-07
daemon/update.go: fix FIPS check and remove day 2 support

yuqi-zhang:
We no longer carry coreos-fips script in RHCOS, so that check fill
fail regardless. As a temporary fix, check state of node via
cat /proc/sys/crypto/fips_enabled instead.

Fixes: https://github.com/openshift/machine-config-operator/issues/1250

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/user"
@@ -38,8 +39,8 @@ const (
 	coreUserName = "core"
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
-	// fipsCommand is the command to use when enabling or disabling FIPS
-	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
+	// fipsFile is the file to check if FIPS is enabled
+	fipsFile = "/proc/sys/crypto/fips_enabled"
 )
 
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -325,6 +325,22 @@ func (dn *Daemon) updateFIPS(current, desired *mcfgv1.MachineConfig) error {
 	if current.Spec.FIPS == desired.Spec.FIPS {
 		return nil
 	}
+
+	cmd := exec.Command(fipsCommand)
```
cgwalters: Seems odd to shell out to `cat` as opposed to reading the file directly...

yuqi-zhang: Yes sorry let me fix that, I just wanted to check if this flow was what people wanted

pkg/daemon/update.go
```diff
@@ -306,53 +307,36 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+	if err != nil {
+		return errors.Wrapf(err, "Error reading FIPS file at %s: %s", fipsFile, string(content))
 	}
-	cmd := exec.Command(fipsCommand, arg)
-	dn.logSystem("Running %s %s", fipsCommand, arg)
-	if out, err := cmd.CombinedOutput(); err != nil {
-		return errors.Wrapf(err, "%s FIPS: %s", arg, string(out))
+	nodeFIPS, err := strconv.ParseBool(string(content))
+	if err != nil {
+		return errors.Wrapf(err, "Error parsing FIPS file at %s", fipsFile,)
```
jlebon: Extra comma here?

ashcrow:
verify failure:
```
pkg/daemon/update.go:331: File is not `goimports`-ed (goimports)
		return errors.Wrapf(err, "Error parsing FIPS file at %s", fipsFile,)
pkg/daemon/update.go:339:21: error strings should not be capitalized or end with punctuation or a newline (golint)
		return errors.New("Refusing to update FIPS mode. This is not a supported operation.")
		                  ^
```

pkg/daemon/update.go
```diff
@@ -306,53 +307,36 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+	}
+	if desired.Spec.FIPS == nodeFIPS {
+		// Check if FIPS on the system is at the desired setting
+		return nil
+	}
+
+	if current.Spec.FIPS != desired.Spec.FIPS {
```
jlebon: Hmm, should we move this check higher up?

pkg/daemon/update.go
yuqi-zhang: Whoops, fixing

pkg/daemon/update.go
yuqi-zhang: Hm so we ran into this error initially because we detected that the current `current.Spec.FIPS` != `desired.Spec.FIPS` right? Because it was nil?

pkg/daemon/update.go
```diff
@@ -38,8 +39,8 @@ const (
 	coreUserName = "core"
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
-	// fipsCommand is the command to use when enabling or disabling FIPS
-	fipsCommand = "/usr/libexec/rhcos-tools/coreos-fips"
+	// fipsFile is the file to check if FIPS is enabled
+	fipsFile = "/proc/sys/crypto/fips_enabled"
```
ericavonb: @cgwalters didn't approve of this method (should use the `coreos-fips --status` or whatever command): https://github.com/openshift/machine-config-operator/pull/1013/files#r307428010

pkg/daemon/update.go
yuqi-zhang: As in, maybe I should change this just to return an error if `!desired.Spec.FIPS == nodeFIPS`

yuqi-zhang: And flat out ignore what `current.Spec.FIPS` is

yuqi-zhang: but then the Spec would not be reflective of the state of the system, perhaps I should update that as well just above?

pkg/daemon/update.go
yuqi-zhang: Yeah we removed that binary unfortunately, which is what lead to this

pkg/daemon/update.go
yuqi-zhang: I changed it to update current.Spec.FIPS so we don't get stuck in a reboot loop. This should definitely be improved of course but I think this can get us past the errors (not tested yet)

pkg/daemon/update.go
ashcrow: `fipscheck`

jlebon: See https://github.com/openshift/machine-config-operator/issues/1250#issuecomment-551189317. That said, checking for `/proc/sys/crypto/fips_enabled` is a good start.

We don't want to use `coreos-fips` anymore. RHCOS has embedded support for turning on FIPS now.

ashcrow: I think using the proc path is fine above is fine. `fipscheck` is an alternate possibility. `coreos-fips` has been removed.

yuqi-zhang:
/retest

> Post http://boskos.ci/acquire?dest=leased&owner=ci-op-3smxij8t-c4a31&request_id=4447156788061523237&state=free&type=aws-quota-slice: dial tcp 172.30.131.17:80: connect: no route to host

jlebon:
/approve

This looks sane to me!

yuqi-zhang:
Minor update: Instead of just returning nil, if FIPS changes were detected, do:
`return errors.New("Detected change to FIPS flag. Refusing to modify FIPS on a running cluster.")`
instead

yuqi-zhang:
/retest

latest failure still boskos

yuqi-zhang:
/hold

Found an error while testing, investigating

yuqi-zhang:
/hold cancel

Can confirm this should work in existing clusters, as well as properly catch and reject day 2 fips applications as we expect. Have not tested with day 1 pending RHCOS updates

ashcrow:
> Have not tested with day 1 pending RHCOS updates

Understood. I don't _think_ an issue will occur but if one does the fix can be done in a different PR.

/lgtm

cgwalters:
@ashcrow FWIW that context is non-blocking, you don't need to override it.

PR #1253, opened 2019-11-07
pkg/controller/kubelet-config: Fix wrapErrorWithCondition string formatting

wking:
Two changes:

* Don't format the additional arguments on success.  Callers set this up with strings like:

        could not generate the original Kubelet config: %v

    that are only appropriate to the error condition.  Exposing additional details about non-error conditions would be nice, but seems awkward without a broader refactoring.

* Fix the `args[:1]` -> `args[1:]` when picking the arguments to pass in to get formatted.  The previous implementation would use `args[0]` as both the format string and the first argument to that format string.

### DIFF
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -146,18 +146,18 @@ func wrapErrorWithCondition(err error, args ...interface{}) mcfgv1.ContainerRunt
 			corev1.ConditionFalse,
 			fmt.Sprintf("Error: %v", err),
 		)
+		if len(args) > 0 {
+			format, ok := args[0].(string)
+			if ok {
+				condition.Message = fmt.Sprintf(format, args[1:]...)
+			}
+		}
 	} else {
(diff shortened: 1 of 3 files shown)


wking:
Pushed e93c4e02 -> 9c79a463 to bump the copy in `pkg/controller/container-runtime-config/helpers.go` to match.

wking:
[Failures][1] were from the CI-cluster Boskos routing issue, since resolved:

```
failed to acquire a resource: Post http://boskos.ci/acquire?dest=leased&owner=ci-op-gxfbx65b-c4a31&request_id=3629321694672897547&state=free&type=aws-quota-slice: dial tcp 172.30.131.17:80: connect: no route to host
```

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1253/pull-ci-openshift-machine-config-operator-master-e2e-aws/5660

runcom:
@rphillips ptal

cgwalters (commented):
Looks sane to me offhand, but I only gave this a superficial look.  Adding 4.7 label.

PR #1254, opened 2019-11-07
[WIP] Bug 1736254: controller: add platformVSphere to ControllerConfig check

kikisdeliveryservice:
Vsphere is supported but is missing from the check here. Not sure why? But seeing in BZ logs they were seeing a warning from `default`.

Related-to: BZ 1736254


ericavonb:
@kikisdeliveryservice any reason we can't unhold this pr? https://github.com/openshift/machine-config-operator/pull/998

I pushed a rebase to it just now.

kikisdeliveryservice:
> @kikisdeliveryservice any reason we can't unhold this pr? #998
> 
> I pushed a rebase to it just now.

@ericavonb Sorry, didn't remember that PR. Looking, it seemed held for reasons, so you should follow-up and find out the current state. We should have a definitive answer before merging. If it's crashlooping it's not supported and the message makes sense.

Btw, I don't think your rebase went through still tagged as needing one.

Closing this.

PR #1258, opened 2019-11-11
daemon/FIPS: move check to reconcilable

yuqi-zhang:
Fix up the previous FIPS patch by moving the FIPS check for the
system to reconcilable. This way we can detect a change to the
FIPS flag much earlier and fail.

The check for FIPS is still comparing FIPS on the system to the
desiredConfig. In a perfect world currentConfig would be the
correct check, but currently this seems to be failing for
firstboot scenarios.

Please provide the following information:
-->

**- What I did**
Move the FIPS check to earlier in the update process

**- How to verify it**
Run a normal (FIPS=false) cluster, apply a machineconfig to turn it on, watch it fail trying to reconcile

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -307,36 +307,7 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 		}
 	}()
 
-	if err := dn.checkFIPS(oldConfig, newConfig); err != nil {
-		return err
-	}
-
 	return dn.updateOSAndReboot(newConfig)
-}
-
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
I will also investigate why the firstboot errors

cgwalters:
> In a perfect world currentConfig would be the correct check, but currently this seems to be failing for firstboot scenarios.
...
> I will also investigate why the firstboot errors

We talked about this before; it's because the initramfs handles FIPS on firstboot, and the MCD is unaware of this.   Probably what we should do is have the `currentConfig` state be populated from the *active* state on firstboot or so.

PR #561, opened 2019-03-19
 docs/OSUpgrades.md: New file

cgwalters:
Describe the pivot magic with reference to the installer; we need
developers/testers and users to understand this.

PR #1259, opened 2019-11-11
docs/README: Fix "docs/whatever" -> "whatever" internal links

wking:
The README is already in the docs/ directory, so we don't need this in
relative links to other docs/ content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,41 @@
-docs/README.md
+# machine-config-operator
+
+OpenShift 4 is an [operator-focused platform](https://blog.openshift.com/openshift-4-a-noops-platform/),
+and the Machine Config operator extends that to the operating system itself,
+managing updates and configuration changes to essentially everything between the kernel and kubelet.
+
+To repeat for emphasis, this operator manages updates to systemd, cri-o/kubelet, kernel, NetworkManager,
+etc.  It also offers a new `MachineConfig` CRD that can write configuration files onto the host.
+
+The approach here is a "fusion" of code from the original CoreOS
+Tectonic as well as some components of Red Hat Enterprise Linux Atomic Host,
+as well as some fundamentally new design.
+
+The MCO (for short) interacts closely with
+both [the installer](https://github.com/openshift/installer/) as well as Red Hat
+CoreOS. See also [the machine-api-operator](https://github.com/openshift/machine-api-operator)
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
We made the choice to use docs/whatever format so that the links work from the README page on the repo landing page.  This change break them: https://github.com/openshift/machine-config-operator/tree/8a26c28a646ba414b1569ae854bd65a3d563a28

wking:
> We made the choice to use docs/whatever format so that the links work from the README page on the repo landing page.

Ah, I hadn't realized the root README was a symlink.  I see the motivation is CI triggering (d906b807, #561).  I've pushed 8a26c28a -> d33aa670, splitting the READMEs (dropping the symlink) with high-level, hopefully stable stuff in the root and more detailed, probably churnier stuff in docs.  How does that look?

Another possibility, if you want to stick with the symlink, is to use [path-absolute references][1].  I don't see them discussed in [GitHub's docs][2], but I'm pretty sure they work.  If that sounds appealing, I can reroll and we can check to see if I'm remembering correctly ;).

[1]: https://tools.ietf.org/html/rfc3986#section-4.2
[2]: https://help.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax#links

dustymabe:
> Another possibility, if you want to stick with the symlink, is to use [path-absolute references](https://tools.ietf.org/html/rfc3986#section-4.2). I don't see them discussed in [GitHub's docs](https://help.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax#links), but I'm pretty sure they work. If that sounds appealing, I can reroll and we can check to see if I'm remembering correctly ;).

I think they do.. At least it works in my testing: https://github.com/openshift/machine-config-operator/pull/1277

sinnykumari:
It's hard to choose between this PR and the one from PR #1277 , both fixes the broken links neatly. I personally like keeping things unified wherever possible and which makes me lean towards PR #1277 . 

/me looks at other MCO folks to see what they think

wking:
> It's hard to choose between this PR and the one from PR #1277...

I'm fine with whatever, as long as the links work ;).

kikisdeliveryservice:
#1277 seems to work, so let's just go with that since it's straightforward.

going to close this PR. ty!

PR #1262, opened 2019-11-12
Bug 1770273: Verify all containers are dead in stop_all_containers() function

retroflexer:
Verify all containers are dead in stop_all_containers() function.
Fixes: #1770273 
**- What I did**
Added a while loop to validate that all containers have stopped.

**- How to verify it**
While running the DR function of restore.sh, verify that stop_all_containers function does indeed stop all containers.

Verify all containers are stopped in stop_all_containers() function.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -330,8 +330,20 @@ contents:
     }
 
     stop_all_containers() {
-      echo "Stopping all containers.."
-      crictl ps -q | xargs -r crictl stop
+      conids=$(crictl ps -q)
+      iterations=0
+      while [ ! -z "$conids" ]; do
+          let iterations=$iterations+1
+          if [ $iterations -ge 60 ]; then
+              echo "Failed to stop all containers after 60 iterations. Exiting!"
+              exit 1
+          fi
+          crictl stop $conids || true
+          echo "Waiting for all containers to stop... ($iterations/60)"
+          sleep 5
+          conids=$(crictl ps -q)
+      done
+      echo "All containers are stopped."
     }
 
     # validate_environment performs the same actions as the discovery container in etcd-member init


retroflexer:
cc @hexfusion

sinnykumari:
To merge this PR in master we need a valid BZ against 4.3, right now RHBZ#1770273  is targeted to 4.2.z

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -331,7 +331,19 @@ contents:
...
+      conids=$(crictl ps -q)
+      iterations=0
+      while [ ! -z "$conids" ]; do
+          let iterations=$iterations+1
+          if [ $iterations -ge 60 ]; then
+              echo "Failed to stop all containers after 60 iterations. Exiting!"
+              exit 1
+          fi
+          crictl stop $conids || true
+          echo "Waiting for all containers to stop... ($iterations/60)"
```
hexfusion: Optional nit/question would this be more informative to folks at the start of the loop?

hexfusion: Could then also drop https://github.com/openshift/machine-config-operator/pull/1262/files#diff-df2f7b2c367c15c3aa83f8b8e002b3d8R333

retroflexer: OK. I will get rid of lineno. 333.

retroflexer: I removed 333, and added 346 (just so that if we don't even get into the while loop, we will have a message).

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -330,8 +330,20 @@ contents:
...
+          sleep 5
+          conids=$(crictl ps -q)
+      done
+      echo "All containers are stopped."
```
retroflexer: Added a message at the end, so that we provide information to the user even if we don't get into the while loop.

PR #1266, opened 2019-11-13
Add manifest annotations for hosted deployment exclusions

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See https://github.com/openshift/cluster-version-operator/pull/252

### DIFF
--- a/install/0000_80_machine-config-operator_04_deployment.yaml
+++ b/install/0000_80_machine-config-operator_04_deployment.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: openshift-machine-config-operator
   labels:
     k8s-app: machine-config-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   replicas: 1
   selector:
--- a/install/0000_80_machine-config-operator_06_clusteroperator.yaml
+++ b/install/0000_80_machine-config-operator_06_clusteroperator.yaml
@@ -2,6 +2,8 @@ apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: machine-config
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec: {}
 status:
   versions:
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: etcd-quorum-guard
   namespace: openshift-machine-config-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   replicas: 3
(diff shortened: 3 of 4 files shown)


csrwng:
/assign @lorbuschris
This should be ready for merge

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: openshift-machine-config-operator
   labels:
     k8s-app: machine-config-daemon
+  annotations:
```
cgwalters: I think we want these Prometheus rules even in external control plane scenarios, right?

install/0001_00_machine-config-operator_00_service.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     k8s-app: machine-config-daemon
   annotations:
     service.beta.openshift.io/serving-cert-secret-name: proxy-tls
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
cgwalters: And similar for this.

csrwng: @cgwalters do you mean that the mcd itself should be available?

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
csrwng: ack, will remove this change

install/0001_00_machine-config-operator_00_service.yaml
cgwalters: Right, the MCD isn't control plane. so we don't want to exclude anything related to it.

csrwng: Thanks, updated

csrwng:
@LorbusChris @cgwalters this should be ready.

sinnykumari:
/hold
since freeze unless there's a BZ

csrwng:
@sinnykumari we have exception approval to get this change in to 4.3 per @derekwaynecarr

cgwalters:
OK, I trust you to know whether or not a BZ is required.
/approve

cgwalters:
(Did other PRs like this need a BZ or not?)
/lgtm

Will let @sinnykumari lift the hold if she agrees.

csrwng:
> (Did other PRs like this need a BZ or not?)

They did not, since it's not a bug, it's a post-feature freeze exception. You can see the PRs linked to https://github.com/openshift/cluster-version-operator/pull/252

csrwng:
@derekwaynecarr can you please ack that this is ok to merge?

sinnykumari:
/hold cancel

Sorry, didn't know about the explicit approval!

runcom:
> /hold cancel

keeping the hold till @derekwaynecarr comes back since we're past freeze

derekwaynecarr:
/hold cancel

This should merge.  Thanks!

PR #1267, opened 2019-11-13
Bug 1766918: Exit the verify_certs loop after 60 iterations

retroflexer:
Closes: #1766918
**- What I did**
Added iteration counter to exit after 60 iterations.
**- How to verify it**
Verify that there is timeout for exit to prevent the script going forever in dead loop

timeout verify_certs() to prevent script going forever in dead loop.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -273,8 +273,14 @@ contents:
     }
 
     verify_certs() {
+      iterations=0
       while [ "$(ls $ETCD_STATIC_RESOURCES | wc -l)" -lt 9  ]; do
-        echo "Waiting for certs to generate.."
+        let iterations=$iterations+1
+        if [ $iterations -gt 60 ]; then
+          echo "Failed to verify cert generation after 60 iterations. Exiting!"
+          exit 1
+        fi
+        echo "Waiting for certs to generate... ($iterations/60)"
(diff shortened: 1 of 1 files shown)


templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -273,8 +273,14 @@ contents:
     }
 
     verify_certs() {
+      iterations=0
       while [ "$(ls $ETCD_STATIC_RESOURCES | wc -l)" -lt 9  ]; do
-        echo "Waiting for certs to generate.."
+        let iterations=$iterations+1
+        if [ $iterations -gt 60 ]; then
```
hexfusion: question:
- How much time does it take for 60 iterations?

- Should the user know about the timeout so they don't kill the script manually? Maybe something like "Waiting N minutes for certs to generate.."

hexfusion: Actually I guess you addressed this

> Waiting for certs to generate... ($iterations/60)

retroflexer: Yeah. Thanks. If there is a different way you want me to tackle it, I can convert it into seconds and minutes too.

hexfusion: 10 mins which I believe is the timeout for the client

PR #1268, opened 2019-11-14
Bug 1772440: OpenStack: decrease infra pods resource consumption

Fedosin:
All infra pods have "requests" section with 1Gb of RAM and 150m of CPU, but they do not consume more than 200mb and 100m respectively.
This leads to irrational use of resources and prevents the launch of workloads on workers.

Fixes: bz1772440


Fedosin:
@kikisdeliveryservice done :)

PR #1269, opened 2019-11-15
add new node role labels

rphillips:
**- What I did**
1.15 [removed](https://github.com/kubernetes/kubernetes/pull/68267) support for node-role.kubernetes.io/{master,worker} labels. We include a carry patch in origin to get around the issue.

To remove the carry patch we will need to upgrade the system in the following manner:

1. Add new labels to Openshift.... node.kubernetes.io/master and node.kubernetes.io/worker
2. Patch MCO to list nodes, and add the new labels to the current nodes
  note: There is not a component in the system that updates a nodes' labels.
3. Over the course of the 4.3.z stream
  a. Update the operators to support the new label role
4. In the 4.4 release use the new label and remove the node-role label from node's registration (systemd config file for the kubelet).

**- How to verify it**

self imposed hold for discussion
/hold

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -21,7 +21,7 @@ contents: |
         --kubeconfig=/var/lib/kubelet/kubeconfig \
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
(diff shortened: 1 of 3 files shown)


ravisantoshgudimetla:
@rphillips - Won't the label `node.kubernetes.io/role=master` still be a problem as per regex `[*.]node.kubernetes.io/*`? I agree with the rest of the approach outlined and this is what we're using for Windows nodes.

rphillips:
@ravisantoshgudimetla thanks... updated.

PR #1271, opened 2019-11-16
update fips

sallyom:
**- What I did**
upgrade from fips-enabled failed in testing, I'm fixing it
fn to updateFIPS changed to always honor old config, since upgrades from fips->non-fips 
or non-fips->fips is unsupported AFAIK
**- How to verify it**
install w/ fips: true
upgrade, verify upgrade succeeds


sallyom:
oops was looking at old branch

PR #1272, opened 2019-11-19
Bug 1770269: DR: etcdctl version should match with what is shipped with product

retroflexer:
Closes: #1770269
**- What I did**
Added etcdctl version in usr-local-bin-openshift-recovery-tools-sh.yaml
**- How to verify it**
Verify that the DR scripts of release 4.3 use etcdctl of version v3.3.17

The DR scripts download the etcdctl on-demand, but the version was set in various scripts to an older version. This change sets the ETCD_VERSION in one place, and removes all the older versions from the other places.


runcom:
@hexfusion ptal

PR #1274, opened 2019-11-19
Bug 1777204: [Baremetal] verify that 'old' HAProxy processes being deleted after HAProxy reload.

yboaron:
Since HAProxy timeout values could be sensitive to some applications (e.g: Kuryr) we use long (24 hours) timeout values for the API LB. 
Because of these timeout values, sometimes old HAProxy processes being terminated only after 24Hours when HAProxy reload operation activated.
So, if HAProxy reload operation triggered many times in a short period we may end-up with many HAProxy processes hanging around (as it's described in Bug 1777204)
 
This change force sending SIGTERM after timeout (default is 120 seconds) to old HAProxy processes which haven't terminated.
Fixes Bug 1777204: Many haproxy processes hanging around due to monitor not closing connections

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
@@ -31,16 +31,37 @@ contents:
       containers:
       - name: haproxy
         image: {{.Images.haproxyImage}}
+        env:
+          - name: OLD_HAPROXY_PS_FORCE_DEL_TIMEOUT
+            value: "120"
         command:
         - "/bin/bash"
         - "-c"
         - |
           #/bin/bash
+          verify_old_haproxy_ps_being_deleted()
+          {
+            local prev_pids
+
+            prev_pids="$1"
+            sleep $OLD_HAPROXY_PS_FORCE_DEL_TIMEOUT
+            cur_pids=$(pidof haproxy)
+
+            for val in $prev_pids; do
+                if [[ $cur_pids =~ (^|[[:space:]])"$val"($|[[:space:]]) ]] ; then
+                   kill $val
+                fi
+            done
+          }
+
           reload_haproxy()
           {
             old_pids=$(pidof haproxy)
             if [ -n "$old_pids" ]; then
                 /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid -x /var/lib/haproxy/run/haproxy.sock -sf $old_pids &
+                #There seems to be some cases where HAProxy doesn't drain properly.
+                #To handle that case, SIGTERM signal being sent to old HAProxy processes which haven't terminated.
+                verify_old_haproxy_ps_being_deleted "$old_pids"  &
             else
                 /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid &
             fi
@@ -61,6 +82,7 @@ contents:
           declare -r haproxy_log_sock="/var/run/haproxy/haproxy-log.sock"
           export -f msg_handler
           export -f reload_haproxy
+          export -f verify_old_haproxy_ps_being_deleted
           rm -f "$haproxy_sock" "$haproxy_log_sock"
           socat UNIX-RECV:${haproxy_log_sock} STDOUT &
           if [ -s "/etc/haproxy/haproxy.cfg" ]; then


yboaron:
This PR fixes [1] for the Baremetal case, after approving this PR we should duplicate it for Openstack

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1771566

kikisdeliveryservice:
Just to make sure this doesn't merged until commit fixed.

/hold

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -41,6 +41,9 @@ contents:
             old_pids=$(pidof haproxy)
             if [ -n "$old_pids" ]; then
                 /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid -x /var/lib/haproxy/run/haproxy.sock -sf $old_pids &
+                #There seems to be some cases where HAProxy doesn't drain properly. 
+                #To handle that case, SIGTERM signal being sent to old HAProxy processes which haven't terminated after few seconds.
+                sleep 5 && bash -c "echo $old_pids|xargs kill " &
```
celebdor: is `$old_pids` multiline? Do we really need xargs?

yboaron: @celebdor , Well, that's an interesting point, my original code didn't include the xargs.
I tested my code by forcing haproxy-monitor to send 'reload' request every few seconds, and sometimes $old_pids was multiline. so to be on the safe side I decided to use xargs.

cybertron: That's odd. pidof defaults to delimiting multiple pids using whitespace, not a newline. When old_pids was multiline was it multiple pids or some other unexpected output?

cybertron: I see there is a -st option to haproxy that is the corresponding SIGTERM command to the -sf one used above. Maybe we could just use that so we have consistency in how we're signalling?

yboaron: The 'old' Haproxy not being deleted since there are still open connections through these processes. 

Using '-st' means that processes will be terminated immediately, and that may break ongoing sessions, while '-sf' terminated the processes only after all the ongoing sessions completed. So, I don't think that '-st' is a good option here.
 
I continued investigating this issue, and it seems that 'old' processes not being deleted (when '-sf' is used) because of the long timeouts we set in [1], I think that decreasing these timeout values is the right way to solve this bug.
I'll update this PR 

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/master/00-master/baremetal/files/baremetal-haproxy-haproxy.yaml#L15-#L17

cybertron: Right, I wasn't suggesting that we replace -sf with -st, but use -st to send the sigterm here.

We may need to be careful about decreasing timeouts. I've seen issues in the past where applications that open persistent connections can be broken by haproxy timeouts. I'm guessing that's why they are so long now.

yboaron: A. Yes, I know that we should be sensitive with HAProxy timeout values, but I'm not sure that current timeouts values are correct. 
I suggested using the default values used by OCP HAProxy routers [1]. 

B. Not sure that I understood your point with the '-st', could you elaborate about that?

[1] https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L114-#L123

yboaron:
I updated haproxy-monitor image to force intensively Haproxy reload (every 30 seconds) to verify this fix.
Seems that the 'old' HAProxy processes terminated properly.

yboaron:
@cc cyberton

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -41,6 +41,9 @@ contents:
             old_pids=$(pidof haproxy)
             if [ -n "$old_pids" ]; then
                 /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid -x /var/lib/haproxy/run/haproxy.sock -sf $old_pids &
+                #There seems to be some cases where HAProxy doesn't drain properly.
+                #To handle that case, SIGTERM signal being sent to old HAProxy processes which haven't terminated.
+                sleep 60  && kill $old_pids &
```
cybertron: What happens if during that 60 seconds another process gets assigned the same pid? Should we re-run pidof before doing this?

yboaron: 1. From the tests  I did in dev-scripts environment, No process was deleted after calling 'haproxy -sf' option, I guess it because of the timeout values we use in Haproxy cfg.

2. If we'll re-run pidof, we'll get also the new haproxy process (the one we don't want to delete), 
so we should run intersect logic between the pidof lists. 

3. Based on the tests on dev-scripts, I don't think we should add this extra logic, let me know if you still think this logic is required.

cybertron: If -sf isn't doing anything, should we just remove it and kill the process immediately? That would eliminate any possibility of a race without any extra logic. It would also remove an unnecessary 60 second delay in stopping the old processes. Or does -sf allow it to to stop some connections?

yboaron: I think that -sf does something ( stops some connections), so to be on the safe side, I suggest to keep the long timeout values and add the kill after timeout (default 120 seconds).
Also to verify that we kill a HAProxy process and not another process that gets assigned the same pid.

kikisdeliveryservice:
all bc of a space in the title... gah.

/bugzilla refresh

PR #1277, opened 2019-11-20
docs/README: fix broken links

dustymabe:
Use absolute paths from the toplevel of the git repo instead of
relative paths because the README exists in the docs directory but
is symlinked from the toplevel of the repo.


kikisdeliveryservice:
Closing since we already have a PR where we're hashing this out. See: https://github.com/openshift/machine-config-operator/pull/1259

kikisdeliveryservice:
Double checked this, and looks good to go! Tyvm

/lgtm

PR #1278, opened 2019-11-20
Bug 1774369: skip setting the kubeletconfig finalizer if it is set

rphillips:
**- What I did**
If the finalizer is already set, then we skip setting another finalizer.

Fixes an issue where Finalizers may be duplicated:

```
  finalizers:
  - 99-worker-935d0eac-d152-4f2a-a754-7f8086c591a1-kubelet
  - 99-worker-935d0eac-d152-4f2a-a754-7f8086c591a1-kubelet
```

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1774369
**- How to verify it**

```
NONE
```

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -547,6 +547,13 @@ func (ctrl *Controller) addFinalizerToKubeletConfig(kc *mcfgv1.KubeletConfig, mc
 			return err
 		}
 
+		// if the finalizer is already set then skip
+		for _, finalizerName := range newcfg.Finalizers {
+			if finalizerName == mc.Name {
+				return nil
+			}
+		}
+
 		kcTmp := newcfg.DeepCopy()
 		kcTmp.Finalizers = append(kcTmp.Finalizers, mc.Name)
 


pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -547,6 +547,13 @@ func (ctrl *Controller) addFinalizerToKubeletConfig(kc *mcfgv1.KubeletConfig, mc
 			return err
 		}
 
+		// if the finalizer is already set then skip
```
kikisdeliveryservice: Isn't this not just checking if set  but checking if matches mc.name?

kikisdeliveryservice: Is there ever a way that fubalizer set but doesn't match mc.name? If so should that error? (I have no idea just wondering)

ericavonb: I don't know that there are any specific cases, but another controller or user could be adding their own finalizer to objects for whatever reason. This check only carrying about our own finalizer seems correct.

(although I'm not sure why we need to use finalizers in this case? what do they provide over using ownerrefs?)

rphillips: I reviewed the code again (it's been awhile). Looks like we can remove the finalizers. Should we do remove the finalizers in this PR? or a different one?

rphillips: Thought some more on this... I think we should fix this bug with this patch, and support a jira story to remove the finalizers.

kikisdeliveryservice: that's fine with me :)

PR #1280, opened 2019-11-21
Bug 1775009: pkg/controller: do not enqueue a nil MCP

runcom:
if a node, for any reason, belongs to only a custom pool, do not enqueue
a nil worker mcp as it would panic when processed.


sinnykumari:
/lgtm

looks good, haven't tested it locally though

PR #1284, opened 2019-11-21
WIP: install: add a cronjob for etcd defragmentation

alaypatel07:
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1772489
**- What I did**
Add a CronJob resource for defragmenting etcd weekly

 add a cronjob for etcd defragmentation

### DIFF
--- a/install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
+++ b/install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
@@ -1,0 +1,55 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: etcd-defrag
+  namespace: openshift-etcd
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          hostNetwork: true
+          containers:
+            - command: ["/usr/bin/sh"]
+              terminationMessagePolicy: FallbackToLogsOnError
+              args:
+                - "-c"
+                - |
+                  set -euo pipefail
+
+                  export ETCDCTL_API=3
+                  export ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)
+                  export ETCDCTL_ENDPOINTS=https://127.0.0.1:2379;
+                  for m in $(etcdctl member list); do
+                    [[ $m =~ "2379" ]] && echo "Defragmenting endpoint $m" && etcdctl --endpoints=$m defrag
+                    sleep 10
+                  done
+              image: registry.svc.ci.openshift.org/openshift:etcd
+              name: etcd-defrag
+              volumeMounts:
+                - mountPath: /etc/ssl/etcd/
+                  name: certs
+          nodeSelector:
+            node-role.kubernetes.io/master: ""
+          restartPolicy: OnFailure
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/master
+              operator: Exists
+            - effect: NoExecute
+              key: node.kubernetes.io/not-ready
+              operator: Exists
+              tolerationSeconds: 120
+            - effect: NoExecute
+              key: node.kubernetes.io/unreachable
+              operator: Exists
+              tolerationSeconds: 120
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/etcd
+              operator: Exists
+          volumes:
+            - hostPath:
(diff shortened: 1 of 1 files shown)


install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
```diff
@@ -0,0 +1,55 @@
...
+              args:
+                - "-c"
+                - |
+                  set -euo pipefail
+
+                  export ETCDCTL_API=3
+                  export ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)
+                  export ETCDCTL_ENDPOINTS=https://127.0.0.1:2379;
+                  for m in $(etcdctl member list); do
+                    [[ $m =~ "2379" ]] && echo "Defragmenting endpoint $m" && etcdctl --endpoints=$m defrag
```
hexfusion: To simplify this why not create the array with only valid client records? `etcdctl member list | awk '{ print $5}' | uniq`

hexfusion: In a larger db this will timeout how are you going to handle that to cover all circumstances?

hexfusion: How will we handle retry if we can not reach a member in the list? If the first defrag fails we exit now right?

hexfusion: IS there a way we can alert admin if cronjob fails I know with traditional cron we could do something like.

```
MAILTO=admin@cluster.com
```

If this fails we should alert somehow. Not email:)

hexfusion:
Functionality should include the following.

- easy to review logging.
- retry on failure: we don't want to have a cron failure that means that member was not defragmented. we should retry reasonably.
- configurable IE: `ETCDCTL_COMMAND_TIMEOUT`, `MAX_RETRY`
- alert on failure

install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
hexfusion: reminder `restartPolicy: Never`

install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
```diff
@@ -0,0 +1,55 @@
...
+                    sleep 10
```
hexfusion: sleep 30

install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
alaypatel07: 1. I don't believe we get to see the alerts, good feature to suggest upstream. 
2. I am not sure how to handle timeouts on large db, maybe set the client timeout to 10 minutes?
3. IMO we shouldn't handle retry if one fails for some reason. We skip the defrag of that member and try again next week
4. IMO we should run the defrag command on each member independent of one another and space it out by 30 seconds as discussed on slack.

install/0000_80_machine-config-operator_09_etcddefragmentation_cronjob.yaml
```diff
@@ -0,0 +1,55 @@
...
+              terminationMessagePolicy: FallbackToLogsOnError
+              args:
+                - "-c"
+                - |
+                  set -euo pipefail
+
+                  export ETCDCTL_API=3
+                  export ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)
+                  export ETCDCTL_ENDPOINTS=https://127.0.0.1:2379;
+                  for m in $(etcdctl member list); do
```
hexfusion: let's hardcode the endpoint vs setting ENV. in 3.4 you can't have both ENV and flag so we should do this here nd be explicit.

hexfusion: I guess we should probably do localhost vs 127.0.0.1 in case of IPv6

cgwalters:
Let's please not encourage the pattern of inline shell script in YAML, particularly for critical services.  Can you instead commit this to the MCO image and make it an entrypoint?

(Bigger picture, do we really need to add this to the MCO now versus getting it in the etcd operator?  Can't we ship the etcd operator in even a "dummy" mode that e.g. just includes this service?)

PR #1285, opened 2019-11-22
Bug 1775647: pkg/daemon: randomize pivot container name

runcom:
If podman has a bug and cannot cleanup container names, we'll be stuck
in the update process as seen in
https://bugzilla.redhat.com/show_bug.cgi?id=1775647.
Randomize container names with UUID to avoid that and I think it's
generally better to do this w/o cleaning up at the beginning every time.

The linked bug won't be closed by this, just using it to attach to it so we can merge as part of that. The podman fix, if any, will close it.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -11,12 +11,12 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/pkg/errors"
-
 	"github.com/opencontainers/go-digest"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	pivottypes "github.com/openshift/machine-config-operator/pkg/daemon/pivot/types"
 	pivotutils "github.com/openshift/machine-config-operator/pkg/daemon/pivot/utils"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/uuid"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/daemon/rpm-ostree.go
```diff
@@ -214,12 +214,11 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
 		imgid = container
 	}
 
-	// Clean up a previous container
-	podmanRemove(pivottypes.PivotName)
+	containerName := pivottypes.PivotNamePrefix+string(uuid.NewUUID())
```
cgwalters: This will have the potential side effect of "leaking" these containers if we fail and retry though.

Possibly a stronger fix here is to use `podman create --annotation=org.openshift.machineconfigoperator.pivot=true`, and then we can filter for containers with that annotation, remove them?

That said, this seems like a raw podman incompatibility across upgrades.  We may need to use a gigantic hammer there like `crio-wipe.service` is doing.

You don't have to solve all of that in this PR, but how about adding the annotation at least while we're here, so a future patch could clean things up more precisely?

runcom: > Possibly a stronger fix here is to use `podman create --annotation=org.openshift.machineconfigoperator.pivot=true`, and then we can filter for containers with that annotation, remove them?

I can do that, prefixing has the same effect when filtering in my mind (+ grep) but annotation looks cooler

> You don't have to solve all of that in this PR, but how about adding the annotation at least while we're here, so a future patch could clean things up more precisely?

Yes and Yes, annotation sounds good to me

runcom: adding a defer as well, so if it's erroring out, we'll purge it, let's see

runcom: should be fixed now

ashcrow:
Will defer to @kikisdeliveryservice for LGTM

kikisdeliveryservice:
this LGTM but letting colin do final review/ check to see if any objections

/assign @cgwalters

cgwalters:
/lgtm
(In the future I want to move this code natively into rpm-ostree FWIW and when we do so we should likely entirely avoid the default podman storage and use e.g. `/run/rpm-ostree/containers-storage` or something like that so that we can't hit incompatibilities like this)

kikisdeliveryservice:
unrelated failures bc of course..

/test e2e-gcp-op

kikisdeliveryservice:
vpc/subnet issues

/test e2e-aws

jlebon:
@kikisdeliveryservice I think we need this backported to 4.1 and 4.3 also, right?

/cherry-pick release-4.1
/cherry-pick release-4.3

jlebon:
> @jlebon: #1285 failed to apply on top of branch "release-4.1"

OK, will look into backporting this one manually. Let's try 4.3 again:

/cherry-pick release-4.3

jlebon:
> OK, will look into backporting this one manually.

OK, so for 4.1 we were still using the separate [pivot](https://github.com/openshift/pivot) binary.... Which, ahh yup, has a `release-4.1` branch. Anyway, I backported it there: https://github.com/openshift/pivot/pull/58. But I'm concerned by the fact that that repo is not hooked into the CI system at all. I guess we can rely on the RHCOS promotion CI coverage. (The alternative I think is backporting https://github.com/openshift/machine-config-operator/pull/882 to 4.1 which although has been used for a while now, is much more new code than strictly necessary to fix this in 4.1).

PR #1291, opened 2019-11-26
Bug 1777082: set vsphere as provider

ericavonb:
fix(pkg/controller/template/render): set vsphere as provider

Set cloudProvider to VSphere in the kubelet configs.
Fixes Bug 1777082: volume cannot mount to node after upgrade

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -467,12 +467,10 @@ func etcdMetricCertCommand(cfg RenderConfig) (interface{}, error) {
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	switch cfg.Platform {
-	case platformAWS, platformAzure, platformOpenStack:
+	case platformAWS, platformAzure, platformOpenStack, platformVSphere:
 		return cfg.Platform, nil
 	case platformGCP:
 		return "gce", nil
-	case platformVSphere:
-		return "", nil // kubelet crash errors when provider set to VSphere right now
 	default:
 		return "", nil
 	}
--- a/pkg/controller/template/render_test.go
+++ b/pkg/controller/template/render_test.go
@@ -43,6 +43,9 @@ func TestCloudProvider(t *testing.T) {
 	}, {
 		platform: "none",
 		res:      "",
+	}, {
+		platform: "vsphere",
+		res:      "vsphere",
 	}}
 	for idx, c := range cases {
 		name := fmt.Sprintf("case #%d", idx)


abhinavdahiya:
would be nice to see if vsphere is also setting the cloud-provider-config ?
https://github.com/openshift/machine-config-operator/pull/1291/files#diff-46149d064bd665c5078bfe4f1c38ae86R73-R96

kikisdeliveryservice:
ok, made a 4.4 bz so we can backport..

/bugzilla refresh

jcpowermac:
We really a e2e-vsphere job for mco.  I was starting to make changes unless someone else wants to.

ericavonb:
/rename Bug 1777082: set vsphere as provider

kikisdeliveryservice:
> We really need a e2e-vsphere job for mco. I was starting to make changes unless someone else wants to.

we dont have any provider specific e2e on mco. you should probably have a cloud-provider-vsphere test somewhere but i dont know that it makes sense to run it in mco constantly and only.  openstack i think has their own tests you can see what they do/where they test?

ericavonb:
> would be nice to see/test if vsphere is also setting the cloud-provider-config ?
> https://github.com/openshift/machine-config-operator/pull/1291/files#diff-46149d064bd665c5078bfe4f1c38ae86R73-R96

I tried adding a vsphere test here but it doesn't seem to be a relevant factor, (it always gets an actual value of `""` acorss releases). Looking at the test and function, it doesn't seem like it would make a difference: https://github.com/openshift/machine-config-operator/blob/a6a863218ba135404217da637751df8a55336c7c/pkg/controller/template/render.go#L495

jcpowermac:
> > We really need a e2e-vsphere job for mco. I was starting to make changes unless someone else wants to.
> 
> we dont have any provider specific e2e on mco. you should probably have a cloud-provider-vsphere test somewhere but i dont know that it makes sense to run it in mco constantly and only. openstack i think has their own tests you can see what they do/where they test?

I could be confused or wasn't clear.  I would like to add a job for vSphere here:
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-master-presubmits.yaml

That way changes like these will spin up a OpenShift cluster on our vSphere CI infrastructure.

kikisdeliveryservice:
leaving this to @gnufied @jcpowermac to LGTM as they are able to test on vsphere

kikisdeliveryservice:
spoke to @jcpowermac and he's going to add a e2e-vsphere test like we have for openstack: https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-master-presubmits.yaml#L362

once that gets done we'll test that way before merging.

/hold

kikisdeliveryservice:
FYI: RHEL7 scaleup test failing across PRs due to BZ1777034 (unrelated to this PR and MCO), seeing same err:
```"Error: Package: cri-o-1.16.0-0.2.dev.rhaos4.3.gita1639b5.el8.x86_64 (openshift-rpms)```

kikisdeliveryservice:
Might need the inflight https://github.com/openshift/release/pull/6131 to merge to have enough clusters for e2e-vsphere to run.

kikisdeliveryservice:
Going to remove the hold so that this can be merged/LGTM'ed by @jcpowermac / @gnufied as they see fit based on the results of e2e-vsphere runs.

/hold cancel

gnufied:
it mostly looks good to me. Although - I do not understand MCO code well enough. :-)

jcpowermac:
@ericavonb the statefulset and PVC tests that were failing succeeded.  I think this PR is good to go.

/lgtm

PR #1299, opened 2019-12-02
CRI-O: add old default hook directories

haircommander:
In CRI-O we moved to not using implicit hooks dirs (https://github.com/cri-o/cri-o/pull/1943, https://github.com/cri-o/cri-o/pull/2731).
Unfortunately, we never set a default, until: https://github.com/cri-o/cri-o/pull/3011. The default isn't enough, though.
We also need /etc/containers/oci/hook.d because /usr is immutable, so a user can add hooks.

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -157,8 +157,12 @@ contents:
     # ]
 
     # Path to OCI hooks directories for automatically executed hooks.
-    # hooks_dir = [
-    # ]
+    # Note: the default is just /usr/share/containers/oci/hooks.d, but /usr is immutable in RHCOS
+    # so we add /etc/containers/oci/hooks.d as well
+    hooks_dir = [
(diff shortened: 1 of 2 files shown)


haircommander:
@mrunalp @umohnani8 PTAL

haircommander:
This will be needed in 1.16 to close https://bugzilla.redhat.com/show_bug.cgi?id=1777838 as well

kikisdeliveryservice:
`level=info msg="Cluster operator support Disabled is True with NotAuthorized: Reporting was not allowed: your Red Hat account is not enabled for remote support or your token has expired"`

This was reported already in 4-dev =/

/test e2e-gcp-upgrade

haircommander:
@kikisdeliveryservice is that failure expected?

kikisdeliveryservice:
@haircommander yes that rhel7 scaleup test always fails.. it's not required for merge so ignore it.

zvonkok:
I tested this change (manually) on my cluster and the hooks are executing again.

PR #1302, opened 2019-12-03
WIP: [baremetal] Add a VIP for provisioning network

bcrochet:
A provisioning VIP to reach Ignition may be necessary in some networks. If
the provisioning-api hostname is present, set up keepalived to use it.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/baremetal/keepalived.conf.tmpl
+++ b/manifests/baremetal/keepalived.conf.tmpl
@@ -33,3 +33,20 @@ vrrp_instance {{`{{.Cluster.Name}}`}}_DNS {
         {{`{{ .Cluster.DNSVIP }}`}}/{{`{{ .Cluster.VIPNetmask }}`}}
     }
 }
+
+{{`{{if .Cluster.APIProvisioningVIP}}`}}
+vrrp_instance {{`{{.Cluster.Name}}`}}_PROVISION {
+    state BACKUP
+    interface {{`{{ .VRRPProvInterface }}`}}
+    virtual_router_id {{`{{ .Cluster.APIProvisioningVirtualRouterID }}`}}
+    priority 140
+    advert_int 1
+    authentication {
+        auth_type PASS
+        auth_pass {{`{{ .Cluster.Name }}`}}_api_provisioning_vip
+    }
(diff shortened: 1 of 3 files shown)


manifests/baremetal/keepalived.conf.tmpl
```diff
@@ -33,3 +33,20 @@ vrrp_instance {{`{{.Cluster.Name}}`}}_DNS {
         {{`{{ .Cluster.DNSVIP }}`}}/{{`{{ .Cluster.VIPNetmask }}`}}
     }
 }
+
+{{`{{if .Cluster.APIProvisioningVIP}}`}}
+vrrp_instance {{`{{.Cluster.Name}}`}}_PROVISION {
+    state BACKUP
+    interface {{`{{ .VRRPProvInterface }}`}}
+    virtual_router_id {{`{{ .Cluster.APIProvisioningVirtualRouterID }}`}}
+    priority 40
```
yboaron: If we prefer that the bootstrap node (as long as it exists) will hold this VIP, shouldn't we set a higher priority?

templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
```diff
@@ -23,6 +23,12 @@ contents:
         weight 50
     }
 
+    {{`{{if .Cluster.APIProvisioningVIP}}`}}
+    vrrp_script chk_api_prov {
+        script "/usr/bin/curl -o /dev/null -kLs https://{{`{{ .Cluster.APIProvisioningVIP }}`}}:2379"
```
yboaron: Shouldn't you check here the local ('localhost:2379') service status?

bcrochet: Done

manifests/baremetal/keepalived.conf.tmpl
bcrochet: Yes. Thanks for catching that.

kikisdeliveryservice:
This hasn't had active development since last year.  Closing, please reopen if in error.

PR #1305, opened 2019-12-04
Bug 1777838: [release-4.3] CRI-O: add old default hook directories

haircommander:
In CRI-O we moved to not using implicit hooks dirs (https://github.com/cri-o/cri-o/pull/1943, https://github.com/cri-o/cri-o/pull/2731).
Unfortunately, we never set a default, until: https://github.com/cri-o/cri-o/pull/3011. The default isn't enough, though.
We also need /etc/containers/oci/hook.d because /usr is immutable, so a user can't add hooks.


kikisdeliveryservice:
This is a backport of #1299

PR #1306, opened 2019-12-05
ovirt: ensure NetworkManager restart after changes

rgolangh:
Workers get a dhclient.conf to prepend the nameserver first
but it doesn't take effect on the /etc/resolv.conf because
NetworkManager is already running.
This makes sure changes take effect.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/ovirt/units/afterburn-hostname.yaml
+++ b/templates/common/ovirt/units/afterburn-hostname.yaml
@@ -1,0 +1,13 @@
+name: "post-network-changes.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Refresh Network Manager after ignition changed settings
+  Before=kubelet.service
+
+  [Service]
+  ExecStart=systemctl restart NetworkManager
+  Type=oneshot
+
+  [Install]
+  WantedBy=multi-user.target


cgwalters:
Er...how could this be special to oVirt?  NM's DHCP implementation should be updating `/etc/resolv.conf`.

Is something in your oVirt install path setting an explicit `/etc/resolv.conf` and overwriting NM's default management of it?

cgwalters:
/hold
since I really don't like this and I think there's some other bug.
This isn't the kind of thing we should be patching in the MCO - if we carried something like this I think it'd be better in RHCOS by default.

yuqi-zhang:
> Refresh Network Manager after ignition changed settings

I'm not very familiar with ovirt, are you writing to /etc/resolv.conf via ignition? If so this should happen before NetworkManager takes over, and /etc/resolv.conf should be already updated by NM.

Today we do a soft refresh of ignition networking by tearing down the dhcp network in the initramfs:
https://github.com/coreos/ignition-dracut/blob/spec2x/dracut/30ignition/coreos-teardown-initramfs-network.service . NetworkManager effectively sees no existing networks anyways

rgolangh:
The thing is I have `dhclient.conf` written [here](https://github.com/openshift/machine-config-operator/blob/master/templates/worker/00-worker/ovirt/files/dhcp-dhclient-conf.yaml)
Its main job is to make the machine resolve the api.$CLUSTER name.

On my workers I see this doesn't take effect, because we need to restart NetworkManager - because I have an ordering problem: NM is up, igntion writes the file. 

Could it be that my master don't restart after pivot? 
If that's not a bug I'm very happy to get the right solution here.

cgwalters:
> On my workers I see this doesn't take effect, because we need to restart NetworkManager - because I have an ordering problem: NM is up, igntion writes the file.

I think what you're most likely seeing actually is that the initramfs networking isn't honoring these DHCP configs.  

For the main and further boots, Ignition has run from the initramfs and wrote the config long before NM runs.

cgwalters:
Did you test this PR and it actually fixed the problem?

yuqi-zhang:
If you can `rd.break` right before the switchroot, you can check the contents of the file and see if its as expected. There have also been scenarios where NM overwrites default confs (although probably not the case here).

rgolangh:
> > On my workers I see this doesn't take effect, because we need to restart NetworkManager - because I have an ordering problem: NM is up, igntion writes the file.
> 
> I think what you're most likely seeing actually is that the initramfs networking isn't honoring these DHCP configs.
> 
> For the main and further boots, Ignition has run from the initramfs and wrote the config long before NM runs.

In that case I need debug NM to see why it doesn't honor the config,

rgolangh:
There is [/etc/NetworkManager/dispatcher/pre-up.d/non-virtual-ip-prepender](https://github.com/openshift/machine-config-operator/blob/master/templates/worker/00-worker/ovirt/files/NetworkManager-non-virtual-ip-prepender.yaml) script that creates the /etc/dhcp/dhclient.conf during pre-up and dhcp comes after it. If I restart NM its ok, so it very much looks like a race between both. I'm closing this, and will open an issue so @cybertron and @celebdor will have a look.

PR #1314, opened 2019-12-09
templates: remove /usr/share/containers/oci/hooks.d from crio config

vrutkovs:
**- What I did**
Removed /usr/share/containers/oci/hooks.d from CRI-O config. This folder is not available on RHEL7 hosts when CRI-O is being installed.

This dir is provided by `oci-systemd-hook` (if I understood correctly), however it is planned to be dropped during 4.3 lifecycle. This PR would ensure CRI-O is configured correctly in preparation to that

**- How to verify it**

Run RHEL7 scaleup - it should pass

/usr/share/containers/oci/hooks.d is no longer being used as a source of CRI-O hooks.

/cc @mtnbikenc

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -161,7 +161,6 @@ contents:
     # so we add /etc/containers/oci/hooks.d as well
     hooks_dir = [
         "/etc/containers/oci/hooks.d",
-        "/usr/share/containers/oci/hooks.d",
     ]
 
     # List of default mounts for each container. **Deprecated:** this option will
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -161,7 +161,6 @@ contents:
     # so we add /etc/containers/oci/hooks.d as well
     hooks_dir = [
         "/etc/containers/oci/hooks.d",
-        "/usr/share/containers/oci/hooks.d",
     ]
 
     # List of default mounts for each container. **Deprecated:** this option will


cgwalters:
It sounds like people were using this hook directory on traditional RHEL systems (since it's writable there), and given `/etc/oci/hooks.d` is new I'm worried we'll put people in a situation where they need to detect the OS and use the correct dir.

Maybe the simplest thing is to merge this PR, and then *also* do `ln -sr /{etc,usr/share}/containers/oci/hooks.d` in the crio RPM?

vrutkovs:
>Maybe the simplest thing is to merge this PR, and then also do ln -sr /{etc,usr/share}/containers/oci/hooks.d in the crio RPM?

That sounds sane to me, not sure if anyone has been installing custom cri-o hooks (especially in /usr/share). @lsm5 wdyt?

vrutkovs:
[Networking issue](https://bugzilla.redhat.com/show_bug.cgi?id=1781242)

/retest

kikisdeliveryservice:
this PR is making a change to something that was recently added in https://github.com/openshift/machine-config-operator/pull/1299/files

@haircommander PTAL

/hold

haircommander:
I am fine with it if it works with @zvonkok

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -159,9 +159,9 @@ contents:
     # Path to OCI hooks directories for automatically executed hooks.
     # Note: the default is just /usr/share/containers/oci/hooks.d, but /usr is immutable in RHCOS
     # so we add /etc/containers/oci/hooks.d as well
+    # /usr/share/containers/oci/hooks.d is not available in recent CRI-O
     hooks_dir = [
         "/etc/containers/oci/hooks.d",
```
haircommander: Since we already have `/etc/containers/oci/hooks.d` as a default in cri-o (as of https://github.com/cri-o/cri-o/commit/16beb92233cf04ef23757d709800586466454f3e), I'd just comment out all of hooks_dir so we can allow cri-o to set the default as it sees fit

vrutkovs: I don't think its a good idea. I'd rather have MCD be the single source of CRI-O configs rather then relying on CRI-O defaults. Other components might use this config to set their own hooks. Also CRI-O might be tested with `oci-systemd-hook` installed (which is not the case for FCOS and currently happening on RHCOS), so the test results from CRI-O suite might be inconvincing

haircommander: per conversations with @cgwalters we decided to keep the templated configs as slim as possible. If other components use this value, they can import the value from `cri-o/cri-o/config`, which we keep public for that very use case. Removing /usr/share/containers/oci/hooks.d will remove the oci-systemd-hook anyway (as it's only shipped there, and also being removed on rhcos), so after this PR the tests will be the same, regardless of whether this field is commented out or not.

my comment isn't necessarily blocking, but we've set the idiom in the other fields to use cri-o defaults when possible, and in this situation it is possible, so I'm still pushing to follow it

kikisdeliveryservice:
@haircommander @vrutkovs relevant bug below:
https://bugzilla.redhat.com/show_bug.cgi?id=1781019

haircommander:
@vrutkovs would you mind changing the name to not reference the bug? I'll backport the fix to release-4.3 and link the bug there

@mrunalp PTAL

vrutkovs:
Done

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
```diff
@@ -159,9 +159,9 @@ contents:
     # Path to OCI hooks directories for automatically executed hooks.
     # Note: the default is just /usr/share/containers/oci/hooks.d, but /usr is immutable in RHCOS
     # so we add /etc/containers/oci/hooks.d as well
+    # /usr/share/containers/oci/hooks.d is not available in recent CRI-O
```
mrunalp: Can we drop this comment?

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -159,9 +159,9 @@ contents:
     # Path to OCI hooks directories for automatically executed hooks.
     # Note: the default is just /usr/share/containers/oci/hooks.d, but /usr is immutable in RHCOS
     # so we add /etc/containers/oci/hooks.d as well
+    # /usr/share/containers/oci/hooks.d is not available in recent CRI-O
```
mrunalp: Can we drop this comment?

vrutkovs: Done. Updated commit message too

cgwalters:
Also adding a
/hold
here since
> This dir is provided by oci-systemd-hook (if I understood correctly), however it is planned to be dropped during 4.3 lifecycle. 

Reference?  It's listed explicitly in the RHCOS manifest I believe because we want to support people doing a "lift and shift" of VMs into containers (where systemd runs in the container) and the like.

Its existence also makes things messier here...I think we probably need both values in the configuration?

haircommander:
/retest

https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/764 actually dropped `oci-systemd-hook`  from rhcos (as it's no longer needed in our container engines)

thus: 
/hold cancel

templates/master/01-master-container-runtime/_base/files/crio.yaml
vrutkovs: Okay, in that case it should be done in a different PR (and moved to shared masters/workers templates)

carbonin:
I saw mention in here that `oci-systemd-hooks` is not needed anymore on cri-o nodes. (specifically from @haircommander in https://github.com/openshift/machine-config-operator/pull/1314#issuecomment-564745602) What does that mean exactly? How does running a container with init work without privileged?

I opened https://bugzilla.redhat.com/show_bug.cgi?id=1807245 a while back and am trying to get some clarity on what's going on.

PR #1317, opened 2019-12-09
Bug 1769847: cmd/gcp-routes-controller: shutdown gcp routes on signal

darkmuggle:
Closes: BZ 1769847

**- What I did**

In BZ 1769847, it was observed that shutting down the GCP controller results in node failures. This change stops the host's routes service on signal.

**- How to verify it**
<in progress>

Shutdown the GCP host route on OS Signal.

### DIFF
--- a/cmd/gcp-routes-controller/run.go
+++ b/cmd/gcp-routes-controller/run.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"sync"
 	"syscall"
 	"time"
@@ -104,6 +105,19 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 	if err := h.Start(); err != nil {
 		return fmt.Errorf("failed to start heath checker: %v", err)
 	}
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for sig := range c {
+			glog.Infof("Signal %s received: shutting down gcp routes service", sig)
+			if err := exec.Command("systemctl", "stop", runOpts.gcpRoutesService).Run(); err != nil {
+				glog.Infof("Failed to terminate gcp routes service on signal: %s", err)
+			} else {
+				break
+			}
+		}
+	}()
 
 	for {
 		select {


darkmuggle:
/cc @ashcrow  I'm not sure how to fully test this. Per [1], I'm not entirely sure that this fixes the root-cause. However, this won't hurt. 

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1769847

kikisdeliveryservice:
this will have to be into 4.4 (current master) and then backported to 4.3 via cherrypick bot

kikisdeliveryservice:
cc: @cgwalters @abhinavdahiya is this what you had in mind per the BZ?

(this is outside of my wheelhouse so deferring to the above)

abhinavdahiya:
I don't see the value add to the BZ 1769847 or existing code with this change??

darkmuggle:
> I don't see the value add to the BZ 1769847 or existing code with this change??

The reason for it was: 
>Clayton Coleman 1:33 PM
>so on GCP thing :slightly_smiling_face:
> the GCP-routes service - does it clear itself on TERM (remove the mapping)?

This ensures that on-term, then route mapping is cleared. I don't _see_ the value, but I was asked to look at it.  This at least ensures that on TERM the node will start to fail the LB health checks a little faster...

kikisdeliveryservice:
@abhin

> I don't see the value add to the BZ 1769847 or existing code with this change??

sounds like you are thinking gcp-routes.service should be changed directly?

darkmuggle:
The crux of the problem is that the node's routes to the LB should not exist until the node is ready/capable of servicing the route. The `gcp-routes.service` setups the kernel's routing so that the requests from the LB can be handled by listening services. The MCO GCP Controller, AFAIK, starts or stops the host's service when its needed. Making the `gcp-routes.service` "smarter" doesn't solve the problem of knowing when/if the node is ready since it not aware of "why" the LB exists or what "what" service/port or for that matter, which health checks it should be checking.

kikisdeliveryservice:
> The crux of the problem is that the node's routes to the LB should not exist until the node is ready/capable of servicing the route. The `gcp-routes.service` setups the kernel's routing so that the requests from the LB can be handled by listening services. The MCO GCP Controller, AFAIK, starts or stops the host's service when its needed. Making the `gcp-routes.service` "smarter" doesn't solve the problem of knowing when/if the node is ready since it not aware of "why" the LB exists or what "what" service/port or for that matter, which health checks it should be checking.

basic q (bc i truly dont know) but wouldn't a service running on the host be better positioned to know if the host is ready/not ready than an operator?

> since it not aware of "why" the LB exists

I dont quite understand what this means?

kikisdeliveryservice:
Also linking over to : https://github.com/openshift/machine-config-operator/pull/1031 that has a good description and discussion on this portion of the codebase (cmd/gcp-route-controller_: https://github.com/openshift/machine-config-operator/pull/1031

darkmuggle:
> basic q (bc i truly dont know) but wouldn't a service running on the host be better positioned to know if the host is ready/not ready?
> 
> > since it not aware of "why" the LB exists
> 
> I dont quite understand what this means?

The node can have many routes (or IP address). Each route corresponds to a specific port (or service). The meta-data service does not provide a way to know which route corresponds to which port. We assume in the case of RHCOS that the route exists for the LB in-front of the API service. But it could be for anything service on the host. 

Each LB has a health check that is setup. Without knowing what the port is, or what health check to test for, we can only make assumptions. 

That's why the `gcp-routes.service` is so dumb. It setups up the routes for the LB's. But if memory serves, the reason why MCO has this code was the route was setup before the API server was ready and that lead to a black-hole situation.

cmd/gcp-routes-controller/run.go
```diff
@@ -105,6 +106,16 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to start heath checker: %v", err)
 	}
 
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for sig := range c {
+			glog.Infof("Signal %s received: shutting down gcp routes service", sig)
+			exec.Command("systemctl", "stop", runOpts.gcpRoutesService).Run()
```
cgwalters: Let's at least log an error if this fails?

cgwalters:
I don't have the expertise here to say this change is correct, but it looks reasonable to me.    One thing I can answer though:

> I'm not sure how to fully test this.

It should be sufficient to deploy a 4.3 (or 4.2 for that matter) cluster in GCP, then cause the MCO to reboot the control plane by deploying a dummy MC that writes `/etc/darkmuggle-was-here` or so.  During the drain/reboot process, the API server should remain fully available.  

So probably first reproduce the failure by doing API calls in a loop `oc get pods` in a shell or whatever, then deploy this patch and verify you don't see some API calls fail.

smarterclayton:
You can run “openshift-tests run-monitor” and it will tell you if a master drops out

darkmuggle:
Aside from trying to figure out how to get this to pass the CI (not my wheel-house), this is ready for review. I'll work on that part.

abhinavdahiya:
just to capture the actual issue for bz1769847

**the mcd daemon issues a reboot of the machine**, the apiserver container is configured with graceful termination such that no new connections are allowed and all current work is completed with it's health check marked as failing.
the gives time for LB to react and create gracefull rolling of apiservers.

now when the reboot is issued, systemd start shutting down services, and it shuts down the gcp-routes.service.. and since gcp-routes.service is designed to cleanup when it receives stop, it removes the ip route immediately dropping/closing connections to the apiserver... and hence all the work done to gracefully close connections etc from above wrt apiserver is not being used here.

this abrupt is what causes non-graceful rollout of apiserver in bz1769847.

> since gcp-routes.service is designed to cleanup when it receives stop

Why is gcp-routes.service setup in this way? that detail should be in https://github.com/openshift/machine-config-operator/pull/1031

now what should be a solution to get the graceful connection dropping back for apiservers?

A) switch the interlink between gcp-routes.service and gcp-routes-controller from STOP to something else, such that stopping gcp-routes doesn't affect connectivity but gcp-routes-controller still has the capability to request _cleanup_ when required.
- we will have to keep in mind that upgrades change machine-os-content on reboot but the files on disk are changed immediately

B) move the gcp-routes-controller health checking to gcp-routes.service it self. and allow mco to **configure** and **turn-on** the behavior using certain configuration.

hopefully that provides more context around https://bugzilla.redhat.com/show_bug.cgi?id=1769847#c7

cgwalters:
> we will have to keep in mind that upgrades change machine-os-content on reboot but the files on disk are changed immediately

Aside: [we can fix that](https://github.com/openshift/machine-config-operator/issues/1190)

cgwalters:
Hm; one preparatory thing that may help here is moving the route script out of RHCOS and into the MCD.

darkmuggle:
> Hm; one preparatory thing that may help here is moving the route script out of RHCOS and into the MCD.

If the MCD knows what the route is, then the problem domain can be a whole lot simpler:
- burn the `gcp-rotues.sh` and `gcp-routes.service` in RHCOS
- the MCD setup only the route it needs

IMHO, that would be the better fix and solves my concerns about the correct route being set up for the service being served by the LB.

ashcrow:
`ci/prow/e2e-aws-scaleup-rhel7` can be ignored

abhinavdahiya:
> > Hm; one preparatory thing that may help here is moving the route script out of RHCOS and into the MCD.
> 
> If the MCD knows what the route is, then the problem domain can be a whole lot simpler:
> 
>     * burn the `gcp-rotues.sh` and `gcp-routes.service` in RHCOS
> 
>     * the MCD setup only the route it needs
> 
> 
> IMHO, that would be the better fix and solves my concerns about the correct route being set up for the service being served by the LB.

you would want RHCOS to be usable without machine-config-daemon running on it. like on bootstrap-host or new control-plane node.

Issue #1319, opened 2019-12-10
Recent containers storage.conf breaking RHCOS promotion


cgwalters:
I'm not *sure* this is a MCO (containerruntimeconfig) bug but it seems likely.  Recent [RHCOS 4.3 promotions](https://prow.svc.ci.openshift.org/?job=promote-release-openshift-machine-os-content-e2e-aws-4.3) are dying with `E1210 16:15:51.105286   11181 daemon.go:1350] content mismatch for file /etc/containers/storage.conf:`:

https://storage.googleapis.com/origin-ci-test/logs/promote-release-openshift-machine-os-content-e2e-aws-4.3/1012/artifacts/e2e-aws/pods/openshift-machine-config-operator_machine-config-daemon-29fqh_machine-config-daemon.log

This seems to have been provoked with a change to the default in the `containers-common` package.  I don't understand how this could break anything though.

Hum...*unless* - if somehow the cconfigc¹ isn't running at bootstrap time, and we were just happening to rely on the contents being the same as the RPM package.  Then, our generated Ignition wouldn't include the changes, but the MCD would expect to find them.  But hmm...we md5sum the generated Ignition, so that seems unlikely.

¹ for short, also it's somehow hypnotic spelled that way

vrutkovs:
MCO uses [static template](https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-container-runtime/_base/files/container-storage.yaml) when containerruntimeconfig is applied, including it in the Ignition file applied to the machine before machine-os-content is applied. This would break if storage.conf is being changed in a newer RPM in machine-os-content - MCD won't be able to sync configs.

Could MCD peek into the staged deployment before the reboot, change particular settings and update Ignition?

cgwalters:
But in the current code, the templated config should override whatever the RPM ships.

cgwalters:
Now, it would obviously be a lot nicer if in fact e.g. we just *merged* any changes we wanted to make with the actual defaults from the RPM, but that would be a rewrite of the code.

(Really what we want is a drop-in `containers.conf.d` style thing)

cgwalters:
Filed https://github.com/openshift/machine-config-operator/pull/1320 as a cleanup while looking at this but also as a CI test canary.

cgwalters:
Ahhh wait, I think I know the problem!  If the templated config is identical to the version shipped in the package, then ostree assumes that the admin hasn't modified it, and will replace it with the new default version.

[...]

PR #1320, opened 2019-12-10
templates/container-storage: Add a "this is generated" note

cgwalters:
It's identical between master/worker, so move it into the
common pool.

Just a cleanup while looking at
https://github.com/openshift/machine-config-operator/issues/1319

---

templates/container-storage: Add a "this is generated" note

This is a best practice, and particularly in this case
when we're intending to modify a config file, it can't be identical
to the shipped version.

This avoids having libostree replace it with a newer version.

Closes: #1319


cgwalters:
If [my theory](https://github.com/openshift/machine-config-operator/issues/1319#issuecomment-564206602) is correct, then in this PR what I'll do is add a second commit which adds a dummy comment like `# Generated from the Machine Config Operator's containerruntimeconfig` - that will mark the file as modified and libostree won't replace it with a new default.

cgwalters:
Now with a new :new:  commit which should fix the bug.
And hopefully CI is fixed now (some errors were unrelated).

cgwalters:
Just logged into the cluster spun up by that e2e-aws and things are looking good there so far!

```
walters@toolbox ~/s/g/o/ostree> oc get machineconfigpool
NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT   AGE
master   rendered-master-5497b82f2b60d6a6e17bd16345e22dc1   True      False      False      3              3                   3                     0                      9m53s
worker   rendered-worker-4fe1fc25bc80ac51a692798946497e92   True      False      False      3              3                   3                     0                      9m53s
```

cgwalters:
I just want to say it's pretty awesome that we're fixing a critical bug by adding a comment.

Funny enough, this bug is effectively the inverse of https://src.fedoraproject.org/rpms/selinux-policy/pull-request/2

vrutkovs:
Clean install works, but gcp-upgrade still fails - not sure why though, I expected this to resolve upgrades as well

cgwalters:
I think if the current release payload is broken, then upgrades can't pass.  Let's retry once, but if it still fails I think we should pull out the `/override` :hammer: .
/test e2e-gcp-upgrade

cgwalters:
One mystery here is how we got into a situation where the promoted payload cause this issue.

kikisdeliveryservice:
Looking here it seems as tho it's failing bc of the storage.conf thing:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1320/pull-ci-openshift-machine-config-operator-master-e2e-gcp-upgrade/605/artifacts/e2e-gcp-upgrade/pods/openshift-machine-config-operator_machine-config-daemon-4wfbw_machine-config-daemon.log

If that's that case we can either hold this until that is resolved (since this isn't emergency) or merge it bc it helps solve that (which doesn't seemlike that case?)

vrutkovs:
IIUC the broken machine-os-content never got promoted to nightlies (https://openshift-release.svc.ci.openshift.org/ 4.3.0.0 broken for ~5 days), but CI tests compose latest release from latest tags, so it attempts to update from a broken payload to a "fixed" payload. I assumed this would fix it too, not quite sure why it doesn't.

Perhaps its MCD on the host not being updated (this requires a new MCD RPM with this PR)?

cgwalters:
> If that's that case we can either hold this until that is resolved (since this isn't emergency)

The storage.conf conflict is exactly what this PR is addressing.  Note this one passed e2e-aws, but other PRs on master aren't.

>  but CI tests compose latest release from latest tags, so it attempts to update from a broken payload to a "fixed" payload. 

Hmm.  Oh, wait, so CI runs are also pulling in the latest `machine-os-content` that's tagged, but before promotion?  That seems like a bug if so.

Anyways, I want to more fully outline how this bug occurred.  First, [libostree](https://github.com/ostreedev/ostree/) has a semantic for config files shared with dpkg/rpm and other OS update systems in general, which is "keep modified config files, replace with latest version otherwise".  

We previously tried to replace the file with its default version, and that worked OK because the *new* version was the same thing.

But the MCD will validate that it's in its expected state, and as soon as we do the upgrade, we fail that validation.

Here we're just adding a comment to flag the file as modified and ensure that our version always wins, regardless of what changes in the package.

(Now we should also stop overwriting the package version and merge our changes, but that's a separate bug)

We need to backport this probably as far back as 4.2 or maybe even 4.1 - otherwise it's a time bomb waiting to bite whenever podman is updated in those streams.

cgwalters:
I think the upgrade is failing because we need this PR backported to older branches too.

So, someone mind validating what I'm saying here and then dropping the `/override e2e-gcp-upgrade` ?

kikisdeliveryservice:
So I'm pretty sure this is failing in bootstrap (as always) bc the bootstrap version wouldn't have the changes the PR has it would be upgrading _to_ this version. (ie a cluster comes up without images from this PR and then upgrades to this PR image). It will always fail until the fix is merged and future prs would be upgrading from the version with the fix. Right?

kikisdeliveryservice:
Ok double checked my e2e understanding and yes: cluster comes up -> upgrades to image with PR's changes -> runs e2e.  This PR is still getting stuck in the initial bootstrap (exactly how/where all the other e2e runs were failing before this PR) bc it doesn't have the fix.  So talked to Colin and will override.

/lgtm

cgwalters:
Oh nice thanks vrutkovs!

vrutkovs:
/retitle Bug 1781708: templates/container-storage: Add a "this is generated" note

PR #1325, opened 2019-12-11
templates: Unify rest of container-runtime into common/

cgwalters:
A while ago we added the `templates/common` for things shared
between master/worker, and my recollection at that time is I only
looked at `templates/master/00-master` and not the other subdirectories.

All of these files for `container-runtime` are identical (as you'd
expect, no reason for container configuration to differ), so move
them into the `templates/common` like we just did for the
`storage.conf`.

Recent PRs have touched both versions in identical ways, e.g.
01108fc0c4d2581802b10d528669891b060ab2c3
and with this we'll only need to change one file.

These directories now go away entirely, including a stale `openstack/OWNERS`
file.


cgwalters:
(I guess a minor downside of this is that the container configs are merged in with the "base" stuff but eh, not sure the separation is really valuable)

haircommander:
oooOOhh I didn't even know this was an option

big
/approve
from me

kikisdeliveryservice:
since this seems to be removing the container runtime configs as a UI/UX, would like @mrunalp / @umohnani8 / @mtrmac to approve & lgtm as well. it's not insignificant change:

https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1325/pull-ci-openshift-machine-config-operator-master-e2e-aws/5966/artifacts/e2e-aws/must-gather/registry-svc-ci-openshift-org-ci-op-by87vn0s-stable-sha256-c31e3068a603b8d8add473dbaaa5b933323a23dc862cb266855248e7cba5ac99/cluster-scoped-resources/machineconfiguration.openshift.io/machineconfigs/

Wondering if maybe there should be a container-runtime-config common instead of throwing everything into base?  Could make it harder to trouble shoot later...

/hold

kikisdeliveryservice:
For ex: in a recent troubleshooting thing, I knew to look for ctrcfg -> new container-runtime-config-> rendered master.  Which was relatively smooth and easy to look at.

cgwalters:
> since this seems to be removing the container runtime configs as a UI/UX,

Creating a `containerruntimeconfig` CRD isn't changed by this.   Users shouldn't have been editing the Ignition generated by the MCO before.

(Though hmm, do we correctly delete old MC fragments that we generated before on upgrade that are unowned now?  Need to test that...)

cgwalters:
/hold
based on https://github.com/openshift/machine-config-operator/pull/1325#issuecomment-565252389

runcom:
also, is the CR controller aware of this change? I can see in the code that they're currently using the `role` to grab templates from the template dir but now this made them to `common` - not sure, but something we might want to check

umohnani8:
Is this something we still want? I can work on getting this to completion if needed :)

cgwalters:
> Is this something we still want? I can work on getting this to completion if needed :)

I think so yes!

PR #1326, opened 2019-12-11
cri-o: remove hooks dir /usr/share/containers/oci/hooks.d

haircommander:
By default in IPI, the only hooks that we add are oci-systemd-hook, which is deprecated and should be removed.
Since /etc... is actually mutable, and the only shipped hook in /usr... is deprecated, we should remove /usr...

This also will fix https://bugzilla.redhat.com/show_bug.cgi?id=1781019 (once backported to 4.3)


haircommander:
@mrunalp @kikisdeliveryservice PTAL

haircommander:
/close 
this is a dup of https://github.com/openshift/machine-config-operator/pull/1314 
which I got confused for the fcos one
sorry for the noise

PR #1330, opened 2019-12-12
Add kernelType to MachineConfig

sinnykumari:
By default we use traditional kernel on RHCOS system. With this,
one can create a MachineConfig specifying `kernelType: realtime` .
Once this MachineConfig is applied to OpenShift cluster,
targetted RHCOS node should switch to using realtime kernel.

Currently, realtime kernel packages are shipped as rpm files in
RHCOS OSContainer. MCO fetches those rpms from OSContainer and
uses rpm-ostree to layer realtime kernel on host.

### DIFF
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -233,6 +233,9 @@ spec:
           fips:
             description: "fips controls FIPS mode"
             type: boolean
+          kernelType:
+            description: "contains which kernel we want to be running like default (traditional), realtime"
+            type: string
           osImageURL:
             description: "osImageURL specifies the remote location that will be used to fetch the OS"
             type: string
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -21,6 +21,7 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
 	var fips bool
+	var kernelType string
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
 		// if any of the config has FIPS enabled, it'll be set
@@ -29,6 +30,23 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 		}
 		outIgn = ign.Append(outIgn, configs[idx].Spec.Config)
 	}
+
+	// sets the KernelType if specified in any of the MachineConfig
+	for _, cfg := range configs {
+		if cfg.Spec.KernelType == "realtime" {
+			kernelType = cfg.Spec.KernelType
+			break
+		}
+		if kernelType == "default" && cfg.Spec.KernelType != "realtime" {
+			kernelType = cfg.Spec.KernelType
+		}
+	}
+
+	// If no MC sets kerneType, then set it to 'default' since that's what it is using
+	if kernelType == "" {
+		kernelType = "default"
+	}
+
 	kargs := []string{}
 	for _, cfg := range configs {
 		kargs = append(kargs, cfg.Spec.KernelArguments...)
@@ -40,6 +58,7 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 			KernelArguments: kargs,
 			Config:          outIgn,
 			FIPS:            fips,
+			KernelType:      kernelType,
 		},
 	}
 }
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -32,12 +32,12 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
 	}
 
 	// sets the KernelType if specified in any of the MachineConfig
+	// Setting kerneType to realtime in any of MachineConfig takes priority
 	for _, cfg := range configs {
 		if cfg.Spec.KernelType == "realtime" {
 			kernelType = cfg.Spec.KernelType
 			break
-		}
-		if kernelType == "default" && cfg.Spec.KernelType != "realtime" {
+		} else if kernelType == "default" {
 			kernelType = cfg.Spec.KernelType
 		}
 	}
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -524,9 +524,9 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 
 // generateRenderedMachineConfig takes all MCs for a given pool and returns a single rendered MC. For ex master-XXXX or worker-XXXX
 func generateRenderedMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig, cconfig *mcfgv1.ControllerConfig) (*mcfgv1.MachineConfig, error) {
-	// Before merging all MCs for a specific pool, let's make sure each contains a valid Ignition Config
+	// Before merging all MCs for a specific pool, let's make sure MachineConfigs are valid
 	for _, config := range configs {
-		if err := ctrlcommon.ValidateIgnition(config.Spec.Config); err != nil {
+		if err := ctrlcommon.ValidateMachineConfig(config.Spec); err != nil {
 			return nil, err
 		}
 	}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -11,6 +11,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"syscall"
@@ -23,10 +24,12 @@ import (
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
+	pivotutils "github.com/openshift/machine-config-operator/pkg/daemon/pivot/utils"
 	errors "github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
@@ -41,6 +44,10 @@ const (
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsFile is the file to check if FIPS is enabled
 	fipsFile = "/proc/sys/crypto/fips_enabled"
+	// Traditional kernel
+	defaultKernelType = "default"
+	// Realtime kernel
+	realtimeKernelType = "realtime"
 )
 
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
@@ -307,6 +314,20 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 		}
 	}()
 
+	// Switch to real time kernel
+	if err := dn.switchKernel(oldConfig, newConfig); err != nil {
+		return err
+	}
+
+	defer func() {
+		if retErr != nil {
+			if err := dn.switchKernel(newConfig, oldConfig); err != nil {
+				retErr = errors.Wrapf(retErr, "error rolling back Real time Kernel %v", err)
+				return
+			}
+		}
+	}()
+
 	return dn.updateOSAndReboot(newConfig)
 }
 
@@ -315,12 +336,21 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 // and the MCO would just operate on that.  For now we're just doing this to get
 // improved logging.
 type MachineConfigDiff struct {
-	osUpdate bool
-	kargs    bool
-	fips     bool
-	passwd   bool
-	files    bool
-	units    bool
+	osUpdate   bool
+	kargs      bool
+	fips       bool
+	passwd     bool
+	files      bool
+	units      bool
+	kernelType bool
+}
+
+// ValidKernelType returns a valid kernelType, otherwise fallback to default kernel
+func ValidKernelType(kernelType string) string {
+	if kernelType == realtimeKernelType {
+		return realtimeKernelType
+	}
+	return defaultKernelType
 }
 
 // NewMachineConfigDiff compares two MachineConfig objects.
@@ -333,12 +363,13 @@ func NewMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) *MachineCo
 	kargsEmpty := len(oldConfig.Spec.KernelArguments) == 0 && len(newConfig.Spec.KernelArguments) == 0
 
 	return &MachineConfigDiff{
-		osUpdate: oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
-		kargs:    !(kargsEmpty || reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments)),
-		fips:     oldConfig.Spec.FIPS != newConfig.Spec.FIPS,
-		passwd:   !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd),
-		files:    !reflect.DeepEqual(oldIgn.Storage.Files, newIgn.Storage.Files),
-		units:    !reflect.DeepEqual(oldIgn.Systemd.Units, newIgn.Systemd.Units),
+		osUpdate:   oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
+		kargs:      !(kargsEmpty || reflect.DeepEqual(oldConfig.Spec.KernelArguments, newConfig.Spec.KernelArguments)),
+		fips:       oldConfig.Spec.FIPS != newConfig.Spec.FIPS,
+		passwd:     !reflect.DeepEqual(oldIgn.Passwd, newIgn.Passwd),
+		files:      !reflect.DeepEqual(oldIgn.Storage.Files, newIgn.Storage.Files),
+		units:      !reflect.DeepEqual(oldIgn.Systemd.Units, newIgn.Systemd.Units),
+		kernelType: ValidKernelType(oldConfig.Spec.KernelType) != ValidKernelType(newConfig.Spec.KernelType),
 	}
 }
 
@@ -554,6 +585,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	args := append([]string{"kargs"}, diff...)
 	dn.logSystem("Running rpm-ostree %v", args)
 	return exec.Command("rpm-ostree", args...).Run()
+}
+
+// MountOSContainer mounts the container and returns the mountpoint
+func (dn *Daemon) MountOSContainer(container string) (mnt, containerName string, err error) {
+
+	var authArgs []string
+	if _, err = os.Stat(kubeletAuthFile); err == nil {
+		authArgs = append(authArgs, "--authfile", kubeletAuthFile)
+	}
+	// Pull the image
+	args := []string{"pull", "-q"}
+	args = append(args, authArgs...)
+	args = append(args, container)
+	pivotutils.RunExt(false, numRetriesNetCommands, "podman", args...)
+
+	containerName = "mcd-" + string(uuid.NewUUID())
+	// `podman mount` wants a container, so let's create a dummy one, but not run it
+	var cidBuf []byte
+	cidBuf, err = runGetOut("podman", "create", "--net=none", "--annotation=org.openshift.machineconfigoperator.pivot=true", "--name", containerName, container)
+	if err != nil {
+		return
+	}
+
+	cid := strings.TrimSpace(string(cidBuf))
+	// Use the container ID to find its mount point
+	mntBuf, err := runGetOut("podman", "mount", cid)
+	if err != nil {
+		return
+	}
+	mnt = strings.TrimSpace(string(mntBuf))
+	return
+}
+
+// switchKernel updates kernel on host with the kernelType specified in MachineConfig.
+// Right now it supports default (traditional) and realtime kernel
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support Kernel update only on RHCOS nodes
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kernel on non-RHCOS nodes is not supported")
+	}
+
+	dn.logSystem("ImageURl  %v - %v ", oldConfig.Spec.OSImageURL, newConfig.Spec.OSImageURL)
+	defaultKernel := []string{"kernel", "kernel-core", "kernel-modules", "kernel-modules-extra"}
+	dn.logSystem("Initiating switch from kernel %v to %v", ValidKernelType(oldConfig.Spec.KernelType), ValidKernelType(newConfig.Spec.KernelType))
+	if ValidKernelType(oldConfig.Spec.KernelType) == realtimeKernelType && ValidKernelType(newConfig.Spec.KernelType) == defaultKernelType {
+		dn.logSystem("Switching to Default kernel now")
+		args := []string{"override", "reset"}
+		args = append(args, defaultKernel...)
+		rtKernelUninstall := []string{"--uninstall", "kernel-rt-core", "--uninstall", "kernel-rt-modules", "--uninstall", "kernel-rt-modules-extra"}
+		args = append(args, rtKernelUninstall...)
+		dn.logSystem("Running rpm-ostree %+q", args)
+		if err := exec.Command("rpm-ostree", args...).Run(); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	var mnt, containerName string
+	var err error
+	if mnt, containerName, err = dn.MountOSContainer(newConfig.Spec.OSImageURL); err != nil {
+		return err
+	}
+
+	defer func() {
+		// Delete container and remove image once we are done with using rpms available in OSContainer
+		podmanRemove(containerName)
+		exec.Command("podman", "rmi", newConfig.Spec.OSImageURL).Run()
+		dn.logSystem("Deleted container and removed OSContainer image")
+	}()
+
+	rtRegex := regexp.MustCompile("kernel-rt(.*).rpm")
+	if ValidKernelType(oldConfig.Spec.KernelType) == defaultKernelType || ValidKernelType(newConfig.Spec.KernelType) == realtimeKernelType {
+		// Switch to RT kernel
+		dn.logSystem("Switching to RT kernel now")
+		args := []string{"override", "remove"}
+		args = append(args, defaultKernel...)
+		rpms, err := ioutil.ReadDir(mnt)
+		if err != nil {
+			return err
+		}
+
+		for _, rpm := range rpms {
+			if rtRegex.MatchString(rpm.Name()) {
+				args = append(args, "--install", fmt.Sprintf("%s/%s", mnt, rpm.Name()))
+			}
+		}
+		cmd := exec.Command("rpm-ostree", args...)
+		dn.logSystem("Command to be executed by rpm-ostree %v", cmd)
+		if err := cmd.Run(); err != nil {
+			return err
+		}
+	}
+
+	if ValidKernelType(oldConfig.Spec.KernelType) == realtimeKernelType && ValidKernelType(newConfig.Spec.KernelType) == realtimeKernelType {
+		if oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL {
+			// Update kernel-rt pacakges from latest OSContainer
+			dn.logSystem("Initiating realtime kernel update on host")
+			args := []string{"uninstall", "kernel-rt-core", "kernel-rt-modules", "kernel-rt-modules-extra"}
+			rpms, err := ioutil.ReadDir(mnt)
+			if err != nil {
+				return err
+			}
+			for _, rpm := range rpms {
+				if rtRegex.MatchString(rpm.Name()) {
+					args = append(args, "--install", fmt.Sprintf("%s/%s", mnt, rpm.Name()))
+				}
+			}
+			cmd := exec.Command("rpm-ostree", args...)
+			dn.logSystem("Updating rt-kernel packages on host: %v", cmd)
+			if err := cmd.Run(); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
 }
 
 // updateFiles writes files specified by the nodeconfig to disk. it also writes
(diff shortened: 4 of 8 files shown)


pkg/daemon/update.go
```diff
@@ -556,6 +572,15 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
+// updateToRealTimeKernel updates to RT Kernel if it is set to true.
+// In case realtime is set to false, switch to traditional Kernel if already
+// on RT kernel, otherwise do nothing
+func (dn *Daemon) updateToRealTimeKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// TODO
```
cgwalters: See https://github.com/williamcaban/ocp-rt/blob/master/rhcos-rt.sh#L18

This will also need to reconcile and handle the case where we're upgrading from an older RT kernel; see https://github.com/williamcaban/ocp-rt/blob/master/README.md#swap-in-an-updated-rt-kernel for that.

sinnykumari: right, thanks for the link

manifests/machineconfig.crd.yaml
```diff
@@ -233,6 +233,9 @@ spec:
           fips:
             description: "fips controls FIPS mode"
             type: boolean
+          realtime:
+            description: "Controls whether we use real time or traditional kernel"
+            type: boolean
```
ericavonb: I was influenced by [abhinavdahiya's comment here](https://github.com/openshift/enhancements/pull/124/files#r349364482) about not using booleans. What about something like `kernelControl: "realtime"`, `kernelControl: "core"`?

pkg/daemon/update.go
ericavonb: In that case, do we want to be more specific and have the RT kernel version specified in the MC?

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -20,13 +20,18 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
...
-	var fips bool
+	var fips, realtime bool
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
 		// if any of the config has FIPS enabled, it'll be set
 		if configs[idx].Spec.FIPS {
 			fips = true
 		}
+
+		// if any of the MachineCOnfig has realtime enabled, it'll be set
```
ericavonb: nit: capitalization `s/MachineCOnfig/MachineConfig`

pkg/daemon/update.go
```diff
@@ -556,6 +572,15 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
+// updateToRealTimeKernel updates to RT Kernel if it is set to true.
+// In case realtime is set to false, switch to traditional Kernel if already
+// on RT kernel, otherwise do nothing
+func (dn *Daemon) updateToRealTimeKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
```
ericavonb: Can we change the function signature to something clearer\*? It gets confusing when you have `updateToRealTimeKernel(oldConfig, newConfig)` then `updateToRealTimeKernel(newConfig, oldConfig)` in the revert where the new is the old and the old is the new 😟 .

Best would be to take only the minimal information from the configs necessary for the function.

\* https://dave.cheney.net/2019/09/24/be-wary-of-functions-which-take-several-parameters-of-the-same-type

pkg/daemon/update.go
cgwalters: I don't think we need to support that; haven't heard of any asks related to having multiple RT kernel choices.  We're only embedding one version of the RT kernel in the oscontainer today (same as we only have one version of the mainline kernel).

Anyone who wants to e.g. test something out of band is probably best using direct `rpm-ostree override` commands via e.g. `oc debug node/`.

manifests/machineconfig.crd.yaml
cgwalters: Yeah, valid thing to consider.  There is also `kernel-alt` at least in RHEL contexts and it's not entirely impossible that that'd be something we want to support.

So perhaps:

`kernel: "default"`
`kernel: "realtime"`

sinnykumari: yeah, sounds reasonable to me to use string instead of boolean here

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -21,14 +21,27 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
...
 	outIgn := configs[0].Spec.Config
 	for idx := 1; idx < len(configs); idx++ {
 		// if any of the config has FIPS enabled, it'll be set
 		if configs[idx].Spec.FIPS {
 			fips = true
 		}
+
+		// sets the Kernel if specified in any of the MachineConfig
+		if configs[idx].Spec.KernelType != "" {
+			kernelType = configs[idx].Spec.KernelType
```
miabbott: Probably should check that the value of `kernelType` is one that is supported

sinnykumari:
There were lot of changes since the original implementation were made. To avoid confusion and get a fresh review, have done force push. This should be good for review now.

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
+// MountOSContainer mounts the container and returns the mountpoint
+func (dn *Daemon) MountOSContainer(container string) (mnt, containerName string, err error) {
+
```
runcom: empty line can be removed

runcom:
just a super minor nit but the rest looks super sane 
/approve

will let others to review again before getting this in

templates/common/_base/units/machine-config-daemon-host.service
```diff
@@ -1,5 +1,5 @@
 name: "machine-config-daemon-host.service"
-enabled: true
+enabled: false
```
ashcrow: On purpose or did this sneak in? :slightly_smiling_face:

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+		if err := cmd.Run(); err != nil {
+			return err
+		}
+	}
+
+	if ValidKernelType(oldConfig.Spec.KernelType) == realtimeKernelType && ValidKernelType(newConfig.Spec.KernelType) == realtimeKernelType {
+		if oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL {
+			// Update kernel-rt pacakges from latest OSContainer
+			dn.logSystem("Initiating realtime kernel update on host")
+			args := []string{"uninstall", "kernel-rt-core", "kernel-rt-modules", "kernel-rt-modules-extra"}
```
cgwalters: Let's at least add

```
// TODO: Compare NEVRAs and avoid doing this if kernel is unchanged
```

(Or if you want, make that optimization now, but not a blocker from my PoV)

cgwalters: Also, one thing related to this is that if the set of RPMs ever changes, this will start failing.  And in fact in recent RHEL kernel packages there's some new subpackages.

Hmm.  Probably the most robust thing here is to parse the `requested-base-removals` bit in `rpm-ostree status --json` and match on the same `kernel-rt*` regexp you have above.

That still doesn't handle the case of "kernel-rt grows a dependency on linux-firmware-rt" e.g. but that's harder.

cgwalters:
Looking at this, I can definitely see the cost to the MCO of shipping kernel-rt the way we are; it would be simplest with two `osImageURL`, a bit less simple with two refs in the same container.  But there'd be a lot of tech debt to pay down to implement that buildsystem side.

templates/common/_base/units/machine-config-daemon-host.service
sinnykumari: It's from PR https://github.com/openshift/machine-config-operator/pull/1366 , picked it from PR in  advance so that this doesn't cause problem while testing.

pkg/daemon/update.go
sinnykumari: Ah yes, somehow missed this. Will fix it during review comment fixes

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -29,6 +30,23 @@ func MergeMachineConfigs(configs []*MachineConfig, osImageURL string) *MachineCo
...
 		outIgn = ign.Append(outIgn, configs[idx].Spec.Config)
 	}
+
+	// sets the KernelType if specified in any of the MachineConfig
+	for _, cfg := range configs {
+		if cfg.Spec.KernelType == "realtime" {
+			kernelType = cfg.Spec.KernelType
+			break
+		}
+		if kernelType == "default" && cfg.Spec.KernelType != "realtime" {
```
ashcrow: nit: Could be a shorter `else if` but this should be fine as is.

pkg/daemon/update.go
```diff
@@ -315,12 +336,21 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+	units      bool
+	kernelType bool
+}
+
+// ValidKernelType returns a valid kernelType, otherwise fallback to default kernel
+func ValidKernelType(kernelType string) string {
+	if kernelType == realtimeKernelType {
+		return realtimeKernelType
+	}
+	return defaultKernelType
```
kikisdeliveryservice: I dont think default should be returned expectedly without a messages saying so - something like: "kernelType %s is not a supported kernel, using defaultKernelType instead," %kernelType

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+
+	var authArgs []string
+	if _, err = os.Stat(kubeletAuthFile); err == nil {
+		authArgs = append(authArgs, "--authfile", kubeletAuthFile)
+	}
+	// Pull the image
+	args := []string{"pull", "-q"}
+	args = append(args, authArgs...)
+	args = append(args, container)
+	pivotutils.RunExt(false, numRetriesNetCommands, "podman", args...)
```
kikisdeliveryservice: can we check this is run successfully ?

kikisdeliveryservice: according to: https://github.com/openshift/machine-config-operator/blob/1e15e3f76c5a050123f38896655c5ca23dafe2ed/pkg/daemon/pivot/utils/run.go#L55  it does return an error if fatal, etc... so we can check for it no?

pkg/daemon/update.go
kikisdeliveryservice: to be clear im suggesting this function returns `string, error`

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+	}
+
+	defer func() {
+		// Delete container and remove image once we are done with using rpms available in OSContainer
+		podmanRemove(containerName)
+		exec.Command("podman", "rmi", newConfig.Spec.OSImageURL).Run()
+		dn.logSystem("Deleted container and removed OSContainer image")
+	}()
+
+	rtRegex := regexp.MustCompile("kernel-rt(.*).rpm")
```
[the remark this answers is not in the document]
sinnykumari: yeah, shipping rpms into kernel-rt/ directory would have been cleaner. 
In future if we have more stuff, we may start supporting multiple OSImageURL

pkg/daemon/update.go
sinnykumari: RunExt() calls runExtBackoff() which calls [glog.Fatalf()](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/pivot/utils/run.go#L48) in case of any error occurred. [glog.Fatalf()](https://godoc.org/github.com/golang/glog#Fatalf) documentation says that it calls os.Exit() after printing error and stacktrace. So, program will exit at that point.

pkg/daemon/update.go
sinnykumari: Yeah, that's a good concern and things to keep an eye on.
Not sure if parsing requested-base-removals will help here as well. I don't see a direct mapping between packages required for default kernel verses rt-kernel. For example: default kernel has kernel sub-packages installed while we don't install kernel-rt sub-package for realtime.

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support Kernel update only on RHCOS nodes
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kernel on non-RHCOS nodes is not supported")
+	}
+
+	dn.logSystem("ImageURl  %v - %v ", oldConfig.Spec.OSImageURL, newConfig.Spec.OSImageURL)
+	defaultKernel := []string{"kernel", "kernel-core", "kernel-modules", "kernel-modules-extra"}
+	dn.logSystem("Initiating switch from kernel %v to %v", ValidKernelType(oldConfig.Spec.KernelType), ValidKernelType(newConfig.Spec.KernelType))
+	if ValidKernelType(oldConfig.Spec.KernelType) == realtimeKernelType && ValidKernelType(newConfig.Spec.KernelType) == defaultKernelType {
```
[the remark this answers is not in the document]
sinnykumari: Yeah, they are valid KernelType at this point. Since KernelType is string, for unset kernelType it contains empty (""). For additional safety. I am using this function to make `default `and empty ("") considered as `default` before comparison

pkg/daemon/update.go
```diff
@@ -556,6 +587,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+
+// switchKernel updates kernel on host with the kernelType specified in MachineConfig.
+// Right now it supports default (traditional) and realtime kernel
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support Kernel update only on RHCOS nodes
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kernel on non-RHCOS nodes is not supported")
+	}
+
+	dn.logSystem("ImageURl  %v - %v ", oldConfig.Spec.OSImageURL, newConfig.Spec.OSImageURL)
```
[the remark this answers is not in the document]
sinnykumari: Ah this shouldn't be needed, I missed to remove this. Was using for local debugging purpose.

pkg/daemon/update.go
```diff
@@ -315,12 +336,21 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+	kargs      bool
+	fips       bool
+	passwd     bool
+	files      bool
+	units      bool
+	kernelType bool
+}
+
+// ValidKernelType returns a valid kernelType, otherwise fallback to default kernel
+func ValidKernelType(kernelType string) string {
```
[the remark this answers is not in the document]
sinnykumari: My initial thought was kind of silently ignoring the invalid kernelType. But yes, it is much better to tell users then being silent.

sinnykumari:
Thank you all for reviewing it. I have tried to incorporate changes from feedback in latest commit

sinnykumari:
seems flakes in the CI

/retest

pkg/controller/render/render_controller.go
```diff
@@ -524,9 +524,9 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 
 // generateRenderedMachineConfig takes all MCs for a given pool and returns a single rendered MC. For ex master-XXXX or worker-XXXX
 func generateRenderedMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig, cconfig *mcfgv1.ControllerConfig) (*mcfgv1.MachineConfig, error) {
-	// Before merging all MCs for a specific pool, let's make sure each contains a valid Ignition Config
+	// Before merging all MCs for a specific pool, let's make sure MachineConfigs are valid
 	for _, config := range configs {
-		if err := ctrlcommon.ValidateIgnition(config.Spec.Config); err != nil {
+		if err := ctrlcommon.ValidateMachineConfig(config.Spec); err != nil {
```
kikisdeliveryservice: Q: if I have a config with an unsupported kernel type will this fail rendering and I need to fix it?  i thought it was not supposed to fail rendering but instead just use the default and let the user know that we are using default kernel instead of the one they have?

sinnykumari:
> Can we have some tests added for this?

Are we talking about e2e test or some other tests?
Adding e2e test is not feasible for this feature because it might fail on some platform like aws (related https://github.com/openshift/installer/issues/2919).

I suggest let''s not block this PR on tests. If some tests can be added, we can get it in another PR.

pkg/controller/render/render_controller.go
cgwalters: Failing rendering was what I suggested - at least until we get to the point of having API validation and rejecting at submission time.

Basically we only support the three values "", "default", and "realtime" - we shouldn't silently ignore a typo for "realtiime" e.g.

This is all the same as how we're validating basic Ignition before trying to roll it out, right/

kikisdeliveryservice: sorry this pr is confusing to me, so it _will_ fail if we have an unsupported kernel type not just use the default? (that canonicalizeKernelType function confused me)

kikisdeliveryservice: so canonicalizeKernelType is really covering the "" case and ensuring that default is used for that?

kikisdeliveryservice:
> > Can we have some tests added for this?
> 
> Are we talking about e2e test or some other tests?
> Adding e2e test is not feasible for this feature because it might fail on some platform like aws (related [openshift/installer#2919](https://github.com/openshift/installer/issues/2919)).
> 
> I suggest let''s not block this PR on tests. If some tests can be added, we can get it in another PR.

is this known to fail on GCP which is where our e2e is run?

sinnykumari:
> > > Can we have some tests added for this?
> > 
> > 
> > Are we talking about e2e test or some other tests?
> > Adding e2e test is not feasible for this feature because it might fail on some platform like aws (related [openshift/installer#2919](https://github.com/openshift/installer/issues/2919)).
> > I suggest let''s not block this PR on tests. If some tests can be added, we can get it in another PR.
> 
> is this known to fail on GCP which is where our e2e is run?

Maybe I am missing something related to different CI test run. RT kernel test shouldn't ideally fail on gcp, so far I did my testing on GCP. But they fail on aws due to limited kernel modules in rt-kernel. o, won't e2e-aws  test run will start failing?

pkg/controller/render/render_controller.go
sinnykumari: yes, canonicalizeKernelType is for code simplification and just for converting "" to default so that we don't un-necessary keep doing comparison for both "" and default.

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
sinnykumari: done

ashcrow:
@sinnykumari 

e2e-gcp-upgrade: infra failure per the `gcloud` output.

e2e-gcp-op: from what I've been told, is "mco testing", but it's been hit with the same/similar e2e infra issues. However, it still will need to pass to verify changes.

sinnykumari:
> @sinnykumari
> 
> e2e-gcp-upgrade: infra failure per the `gcloud` output.
> 
> e2e-gcp-op: from what I've been told, is "mco testing", but it's been hit with the same/similar e2e infra issues. However, it still will need to pass to verify changes.

yeah, I hope infra gets fixed soon.

kikisdeliveryservice:
> > > > Can we have some tests added for this?
> > > 
> > > 
> > > Are we talking about e2e test or some other tests?
> > > Adding e2e test is not feasible for this feature because it might fail on some platform like aws (related [openshift/installer#2919](https://github.com/openshift/installer/issues/2919)).
> > > I suggest let''s not block this PR on tests. If some tests can be added, we can get it in another PR.
> > 
> > 
> > is this known to fail on GCP which is where our e2e is run?
> 
> Maybe I am missing something related to different CI test run. RT kernel test shouldn't ideally fail on gcp, so far I did my testing on GCP. But they fail on aws due to limited kernel modules in rt-kernel. o, won't e2e-aws test run will start failing?

The MCO e2e tests are run on GCP in the suite`e2e-gcp-op`. Since this does work fine on GCP, you should add an e2e test for it.  Adding an MCO e2e test there has nothing to do with the other tests run on aws via e2e-aws, etc...

It should be easy/straighforward to add an e2e test here since this feature does work on GCP, so I'd really like one added.

runcom:
/approve
/lgtm

the CI is back to normal - let's figure a test out after this merges to make it easier for Sinny to spin up a cluster directly with this and test it out.

@sinnykumari did we bump the mcd already in rhcos for this?

sinnykumari:
> @sinnykumari did we bump the mcd already in rhcos for this?

Not yet, we will bump it once this PR gets merged.

sinnykumari:
> > > > > Can we have some tests added for this?
> > > > 
> > > > 
> > > > Are we talking about e2e test or some other tests?
> > > > Adding e2e test is not feasible for this feature because it might fail on some platform like aws (related [openshift/installer#2919](https://github.com/openshift/installer/issues/2919)).
> > > > I suggest let''s not block this PR on tests. If some tests can be added, we can get it in another PR.
> > > 
> > > 
> > > is this known to fail on GCP which is where our e2e is run?
> > 
> > 
> > Maybe I am missing something related to different CI test run. RT kernel test shouldn't ideally fail on gcp, so far I did my testing on GCP. But they fail on aws due to limited kernel modules in rt-kernel. o, won't e2e-aws test run will start failing?
> 
> The MCO e2e tests are run on GCP in the suite`e2e-gcp-op`. Since this does work fine on GCP, you should add an e2e test for it. Adding an MCO e2e test there has nothing to do with the other tests run on aws via e2e-aws, etc...really like one added.

Ah yes, found https://github.com/openshift/release/blob/master/ci-operator/config/openshift/machine-config-operator/openshift-machine-config-operator-release-4.4.yaml#L44 that MCO e2etest runs only in gcp-op. I am working on adding e2e test and will create a separate PR for it

sinnykumari:
/me prays to OpenShift CI god to pass this gcp-op test

runcom:
> /me prays to OpenShift CI god to pass this gcp-op test

it's failing on other PR so it shouldn't be this PR fault

pkg/daemon/update.go
```diff
@@ -556,6 +612,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+		return
+	}
+	mnt = strings.TrimSpace(string(mntBuf))
+	return
+}
+
+// switchKernel updates kernel on host with the kernelType specified in MachineConfig.
+// Right now it supports default (traditional) and realtime kernel
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support Kernel update only on RHCOS nodes
```
runcom: should this function check if we need to do all this?

pkg/daemon/update.go
```diff
@@ -192,6 +199,31 @@ func canonicalizeEmptyMC(config *mcfgv1.MachineConfig) *mcfgv1.MachineConfig {
...
 
+// Returns true if updated packages are available
+func rtKernelUpdateAvailable(rpms []os.FileInfo, rtKernelPkg []string) (bool, error) {
+	for _, pkg := range rtKernelPkg {
+		var out []byte
+		var err error
+		found := false
+
+		if out, err = exec.Command("rpm", "-q", pkg).Output(); err != nil {
+			return false, err
```
runcom: wrap err with something like `failed to run rpm -q blablabla`

runcom: same for the other occurrences of this pattern

pkg/daemon/update.go
runcom:
```
I0122 13:16:51.153341    2609 update.go:1222] Initiating switch from kernel default to default
I0122 13:16:51.159415    2609 run.go:16] Running: podman pull -q --authfile /var/lib/kubelet/config.json registry.svc.ci.openshift.org/ci-op-6mmddzs0/stable@sha256:b2e468365bbc6808c581ee7a7d452bb35c63e1fac05c1501621a76ccb4f8b276
151450e66d9aab561ddc0e94804d9f7b46f1f987f3c9e439f5b746f6a03c2152
I0122 13:17:28.373015    2609 rpm-ostree.go:366] Running captured: podman create --net=none --annotation=org.openshift.machineconfigoperator.pivot=true --name mcd-75d0c6bc-fed2-43f4-8ee5-42c09322fcab registry.svc.ci.openshift.org/ci-op-6mmddzs0/stable@sha256:b2e468365bbc6808c581ee7a7d452bb35c63e1fac05c1501621a76ccb4f8b276
I0122 13:17:28.645906    2609 rpm-ostree.go:366] Running captured: podman mount c6dde396d617bd3e6f74a916b8c753530f7b17c768ac7e7436eb18781e417107
I0122 13:17:28.884382    2609 update.go:1222] Switching to kernelType=default,invoking rpm-ostree ["override" "remove" "kernel" "kernel-core" "kernel-modules" "kernel-modules-extra" "--install" "/var/lib/containers/storage/overlay/b4d98bd61ab6ca1e038351946499de5d802735f014712c4f379875c1856822a0/merged/kernel-rt-core-4.18.0-147.3.1.rt24.96.el8_1.x86_64.rpm" "--install" "/var/lib/containers/storage/overlay/b4d98bd61ab6ca1e038351946499de5d802735f014712c4f379875c1856822a0/merged/kernel-rt-modules-4.18.0-147.3.1.rt24.96.el8_1.x86_64.rpm" "--install" "/var/lib/containers/storage/overlay/b4d98bd61ab6ca1e038351946499de5d802735f014712c4f379875c1856822a0/merged/kernel-rt-modules-extra-4.18.0-147.3.1.rt24.96.el8_1.x86_64.rpm"]
```

this is what the MCD says, basically we don't need to do this _every time_ we need to check if old->new config is asking for a switch, otherwise do as it was nothing

pkg/daemon/update.go
```diff
@@ -556,6 +612,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support Kernel update only on RHCOS nodes
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Updating kernel on non-RHCOS nodes is not supported")
+	}
+
+	defaultKernel := []string{"kernel", "kernel-core", "kernel-modules", "kernel-modules-extra"}
+	rtKernel := []string{"kernel-rt-core", "kernel-rt-modules", "kernel-rt-modules-extra"}
+	args := []string{}
+	dn.logSystem("Initiating switch from kernel %s to %s", canonicalizeKernelType(oldConfig.Spec.KernelType), canonicalizeKernelType(newConfig.Spec.KernelType))
```
runcom: basically, we don't need to initiate this switch unless we have to based on the diff between old and new config right?

pkg/daemon/update.go
```diff
@@ -556,6 +612,122 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+		dn.logSystem("Switching to kernelType=%s, invoking rpm-ostree %+q", newConfig.Spec.KernelType, args)
+		if err := exec.Command("rpm-ostree", args...).Run(); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	var mnt, containerName string
+	var err error
+	if mnt, containerName, err = dn.MountOSContainer(newConfig.Spec.OSImageURL); err != nil {
```
miabbott: Should the mount be checked that the RPMs are present before trying to switch kernels?

It's probably super unlikely that this will happen, but maybe someone tries to do this with a custom release payload or `machine-os-content`?

pkg/daemon/update.go
sinnykumari: yeah right, it should do nothing when kernelType is default and it hasn't been changed.

pkg/daemon/update.go
sinnykumari: yeah, makes sense. I added checks and should throw error when someone tries to switch to realtime but kernel-rt package is not not available in the machine-os-content

pkg/daemon/update.go
sinnykumari: Thanks for catching this! 
Should have looked earlier into the journal logs of this PR. It wasn't obvious because of the similar error message we were seeing in the gcp-op CI in other PRs.

sinnykumari:
Retrying as it says `level=error msg="Error: ResourceLimitExceeded: Task limit exceeded"`

/test e2e-aws
/test e2e-aws-scaleup-rhel7

pkg/daemon/update.go
```diff
@@ -556,6 +613,134 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+	mnt = strings.TrimSpace(string(mntBuf))
+	return
+}
+
+// switchKernel updates kernel on host with the kernelType specified in MachineConfig.
+// Right now it supports default (traditional) and realtime kernel
+func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+
+	// Do nothing if both old and new KernelType are of type default
+	if canonicalizeKernelType(oldConfig.Spec.KernelType) == defaultKernelType && canonicalizeKernelType(newConfig.Spec.KernelType) == defaultKernelType {
```
runcom: Should we do nothing in general if they don’t differ?

sinnykumari: For kernelType=default (traditional kernel ) we don't need to do anything if kernelType in both old and new MC are same. I am using canonicalizeKernelType(), just to make sure that kernelType "" and default are treated same.

This is not the case when both new and old MC has kerneType=realtime, because in this case it could be possible that we have a new kernel-rt package update in latest OSContainer which we fetch from newconfig.spec.OSImageURL. If we don't do that kernel-rt package will never be updated on required nodes. For default kernel, since it is part of OSTree commit, it gets updated the usual way.

Does that make sense? or perhaps I didn't understand your question

sinnykumari:
I moved following rpm-ostree command execution back in the respective if condition so that we never accidentally run it.
```
if err := exec.Command("rpm-ostree", args...).Run(); err != nil {
    return fmt.Errorf("Failed to execute rpm-ostree %+q : %v", args, err)
}
```

pkg/daemon/update.go
runcom: it does perfect cc @cgwalters

pkg/daemon/update.go
```diff
@@ -556,6 +613,136 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
+// MountOSContainer mounts the container and returns the mountpoint
+func (dn *Daemon) MountOSContainer(container string) (mnt, containerName string, err error) {
```
runcom: you can lowercase this method as it's not used outside `pkg/daemon` - `mountOSContainer`

sinnykumari: done

runcom:
op e2e failing for the usual flakes
```
```

PR #1288, opened 2019-11-24
*: conditionally populate the images configmap with cluster-etcd-operator image 

alaypatel07:
[This](https://github.com/openshift/machine-config-operator/pull/1224) adds a flag to MCO for cluster-etcd-operator image and hardcodes
 the value to `""`. 

This PR conditionally populates the cluster-etcd-operator image only
if the etcd custom resource exists and if it has a spec field set as
`managementState: Managed`. It is the same mechanism used 
in the [installer](https://github.com/openshift/installer/pull/2730). This will make sure that the MC rendered during the initial.
boot with the value coming in from the installer flag is always same
 as one generate in the sync loop.

With this, we can use a single [knob](https://github.com/openshift/cluster-etcd-operator/blob/master/manifests/0000_12_etcd-operator_01_operator.cr.yaml#L8) in cluster-etcd-operator 
project to test the bootstrap process in CI with cluster-etcd-operator 
enabled.

The existence of cluster-etcd-operator image is used to render static 
etcd pod yaml. This PR will be followed by another PR which introduces
a boolean flag to be used for conditional rendering of the static pod image. 
It will be a more readable way of achieving the same means.

PR #1334, opened 2019-12-12
[WIP]: *: adds a flag to mco for reading etcd CR yaml

alaypatel07:
**- What I did**
- added a flag to MCO for reading the etcd CR yaml.
- add a field in ControllerConfigSpec to save if cluster-etcd-operator is enabled
- modify the logic to use this added field in ControllerConfigSpec to render etcd static pod yaml

**-  Depends on** 
-  #1288

**- How to verify it**
The current value for Spec is defaulting to False. It will change once we enable cluster-etcd-operator
so should be able to see ControllerConfig as
```
spec:
  ...
  clusterEtcdOperatorEnabled: false
  ...
```

add a flag to mco to read etcd config file

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -26,6 +26,7 @@ var (
 		destinationDir            string
 		etcdCAFile                string
 		etcdImage                 string
+		etcdConfigFile            string
 		etcdMetricCAFile          string
 		haproxyImage              string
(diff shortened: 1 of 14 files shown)


alaypatel07:
FTR, identified the reason for failed runs, pushing up the changes.

kikisdeliveryservice:
This hasn't seen active development since last year.

Closing, please reopen if this is in error.

PR #1336, opened 2019-12-12
DR: save and restore kube apiserver's static pod resources

retroflexer:
Closes: #1783445

**- What I did**
Backup should save static pod resources of kube-apiserver along with the snapshot of etcd database. Similarly, when a restore is invoked, it should restore the static pod resources along with the etcd database.
**- How to verify it**
Follow DR documentation to back up and restore. The cluster should functional after the same backup is restored on all masters.

Save and restore kube apiserver's static pod resources along with etcd database.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
@@ -7,6 +7,7 @@ contents:
 
     set -o errexit
     set -o pipefail
+    set -o errtrace
 
     # example
     # etcd-snapshot-backup.sh $path-to-snapshot
@@ -17,17 +18,26 @@ contents:
     fi
 
     usage () {
-        echo 'Path to backup file required: ./etcd-snapshot-backup.sh ./backup.db'
+        echo 'Path to backup dir required: ./etcd-snapshot-backup.sh <path-to-backup-dir>'
         exit 1
     }
 
     ASSET_DIR=./assets
 
-    if [ "$1" != "" ]; then
-      SNAPSHOT_FILE="$1"
-    else
+    if [ -z "$1" ] || [ -f "$1" ]; then
       usage
     fi
+
+    if [ ! -d "$1" ]; then
+      mkdir -p $1
+    fi
+
+    BACKUP_DIR="$1"
+    DATESTRING=$(date "+%F_%H%M%S")
+    BACKUP_TAR_FILE=${BACKUP_DIR}/snapshot_db_kuberesources_$DATESTRING.tar
+    SNAPSHOT_FILE="${ASSET_DIR}/tmp/snapshot.db"
+
+    trap "rm -f ${BACKUP_TAR_FILE} ${SNAPSHOT_FILE}" ERR
 
     CONFIG_FILE_DIR=/etc/kubernetes
     MANIFEST_DIR="${CONFIG_FILE_DIR}/manifests"
@@ -45,7 +55,12 @@ contents:
       dl_etcdctl
       backup_etcd_client_certs
       backup_manifest
+      backup_latest_kube_static_resources
       snapshot_data_dir
+      tar rf ${BACKUP_TAR_FILE} -C ${ASSET_DIR}/tmp snapshot.db
+      gzip ${BACKUP_TAR_FILE}
+      rm -f ${SNAPSHOT_FILE}
+      echo "snapshot db and kube resources are successfully saved to ${BACKUP_TAR_FILE}.gz!"
     }
 
     run
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
@@ -9,7 +9,7 @@ contents:
     set -o pipefail
 
     # example
-    # ./etcd-snapshot-restore.sh $path-to-snapshot $inital_cluster
+    # ./etcd-snapshot-restore.sh $path-to-backup $inital_cluster
 
     if [[ $EUID -ne 0 ]]; then
       echo "This script must be run as root"
@@ -17,7 +17,7 @@ contents:
     fi
 
     usage () {
-        echo 'Path to snapshot and initial cluster are required: ./etcd-snapshot-restore.sh $path-to-snapshot $initial_cluster'
+        echo 'Path to backup file and initial cluster are required: ./etcd-snapshot-restore.sh $path-to-backup $initial_cluster'
         exit 1
     }
 
@@ -25,7 +25,7 @@ contents:
         usage
     fi
 
-    SNAPSHOT_FILE="$1"
+    BACKUP_FILE="$1"
     INITIAL_CLUSTER="$2"
     ASSET_DIR=./assets
     CONFIG_FILE_DIR=/etc/kubernetes
@@ -39,8 +39,8 @@ contents:
     ETCD_STATIC_RESOURCES="${CONFIG_FILE_DIR}/static-pod-resources/etcd-member"
     STOPPED_STATIC_PODS="${ASSET_DIR}/tmp/stopped-static-pods"
 
-    if [ ! -f "$SNAPSHOT_FILE" ]; then
-      echo "etcd snapshot $SNAPSHOT_FILE does not exist."
+    if [ ! -f "${BACKUP_FILE}" ]; then
+      echo "etcd snapshot ${BACKUP_FILE} does not exist."
       exit 1
     fi
 
@@ -49,6 +49,21 @@ contents:
     function run {
       ETCD_INITIAL_CLUSTER="${INITIAL_CLUSTER}"
       init
+      if [ ${BACKUP_FILE#*.} = "tar.gz" ]; then
+        RESTORE_STATIC_RESOURCES="true"
+        tar xzf ${BACKUP_FILE} -C ${ASSET_DIR}/tmp/ snapshot.db
+        SNAPSHOT_FILE="${ASSET_DIR}/tmp/snapshot.db"
+      else
+        # For backward-compatibility, we support restoring from single snapshot.db file
+        RESTORE_STATIC_RESOURCES="false" 
+        SNAPSHOT_FILE="${BACKUP_FILE}"
+      fi
+
+      if [ ! -f "${SNAPSHOT_FILE}" ]; then
+        echo "etcd snapshot ${SNAPSHOT_FILE} does not exist."
+        exit 1
+      fi
+
       dl_etcdctl
       backup_manifest
       DISCOVERY_DOMAIN=$(grep -oP '(?<=discovery-srv=).*[^"]' $ASSET_DIR/backup/etcd-member.yaml )
@@ -66,7 +81,9 @@ contents:
       stop_all_containers
       backup_data_dir
       remove_data_dir
+      [ "${RESTORE_STATIC_RESOURCES}" = "true" ] && remove_kube_static_resources
       restore_snapshot
+      [ "${RESTORE_STATIC_RESOURCES}" = "true" ] && restore_kube_static_resources
       start_static_pods
       start_kubelet
     }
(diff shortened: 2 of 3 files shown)


templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -148,6 +159,16 @@ contents:
       fi
     }
 
+    remove_kube_static_resources() {
+      COUNT=$(ls -d "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]* 2> /dev/null | wc -l) || true
+      if [ "$COUNT" -gt 1 ]; then
+         echo "Removing etcd certs.."
```
alaypatel07: Typo? "Removing kube static resources.."?

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
```diff
@@ -17,18 +17,21 @@ contents:
     fi
 
     usage () {
-        echo 'Path to backup file required: ./etcd-snapshot-backup.sh ./backup.db'
+        echo 'Path to backup dir required: ./etcd-snapshot-backup.sh <path-to-backup-dir>'
```
alaypatel07: this will need a doc change, can we have the bug to track please?

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -148,6 +159,16 @@ contents:
...
+         rm -rf "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*
```
alaypatel07: why do we need to delete all the revision directories? I know we only need the revision that was backed up, but to be extra safe would be a good idea to only delete the ones after the revision number that was backed up?

kikisdeliveryservice:
does this need to merge in master first then backport using the BZ for 4.3?

retroflexer:
> does this need to merge in master first then backport using the BZ for 4.3?

Yes. That would be the right path, I think.

kikisdeliveryservice:
@retroflexer so let's remove the BZ from this one, after @hexfusion and @alaypatel07 review, we can merge, then do a cherrypick into 4.3 branch using this BZ.

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
```diff
@@ -9,23 +9,34 @@ contents:
...
         exit 1
     }
 
     if [ "$1" == "" ] || [ "$2" == "" ]; then
         usage
     fi
 
-    SNAPSHOT_FILE="$1"
+    if [ ! -d $1 ]; then
+      echo "$1 is not a directory. Please specify backup-dir"
```
hexfusion: if this is getting backported we need to consider any backups that existed which did not have these additional assets

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
```diff
@@ -17,18 +17,21 @@ contents:
...
-    else
+    if [ -z "$1" ] || [ -f "$1" ]; then
       usage
     fi
 
+    if [ ! -d "$1" ]; then
+      mkdir -p $1
+    fi
+
+    SNAPSHOT_FILE="$1/snapshot.db"
```
hexfusion: so this backing up the snapshot to the dir you passed as a param. but then the static pod assets are going into $ASSETS_DIR[1],[2]?

[1] https://github.com/openshift/machine-config-operator/pull/1336/files#diff-df2f7b2c367c15c3aa83f8b8e002b3d8R72
[2] https://github.com/openshift/machine-config-operator/pull/1336/files#diff-f553d41a05548c453f918afebe1b65e5L24

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -61,6 +61,17 @@ contents:
...
 
+    #backup latest static pod resources for kube-apiserver
+    backup_latest_kube_static_resources() {
+      echo "Trying to backup latest static pod resources.."
+      LATEST_STATIC_POD_DIR=$(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]* | head -1) || true
+      if [ -z "$LATEST_STATIC_POD_DIR" ]; then
+          echo "error finding static-pod-resources"
+          exit 1
+      fi
+      tar -czf $ASSET_DIR/backup/kube_static_resources.tar.gz $LATEST_STATIC_POD_DIR
```
hexfusion: I think we should consider adding timestamp to directories so backups do not overwrite themselves by default

hexfusion: Do we need to consider permissions/ownership here at all with `-p`

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
hexfusion: I think this is confusing as I would expect all assets to be included in the directory path I gave. why not a single tar with all assets?

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -61,6 +61,17 @@ contents:
       fi
     }
 
+    #backup latest static pod resources for kube-apiserver
+    backup_latest_kube_static_resources() {
+      echo "Trying to backup latest static pod resources.."
+      LATEST_STATIC_POD_DIR=$(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]* | head -1) || true
```
hexfusion: This isn't doing what you think it is.
```
$ mkdir  kube-apiserver-pod-{99,100,101}
$ ls -td kube-apiserver-pod-[0-9]* | head -1
kube-apiserver-pod-100
```

hexfusion:
/hold
for further testing

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
retroflexer: I agree. I can accommodate both file and directory and only restore the snapshot and not delete the static pod directories. Thanks.

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
retroflexer: No. Both snapshot and static pod resources go into the backup directory supplied. Oh, looks like I missed that change.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
retroflexer: Yeah. -p also retains the date-timestamps.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
retroflexer: -t sorts by time. So, if 101 is the last modified directory, it should get it.

hexfusion: gotcha ok `ls` can handle versions internally `ls -v | tail -1` but if time unique per revision we should be good. Do we have any other corner cases to consider?

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-restore-sh.yaml
hexfusion: we at least need to consider what would happen and possibly document workaround.

hexfusion:
LGTM defer to @alaypatel07

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
```diff
@@ -17,18 +18,27 @@ contents:
...
     fi
 
+    if [ ! -d "$1" ]; then
+      mkdir -p $1
+    fi
+
+    BACKUP_DIR="$1"
+    DATESTRING=$(date "+%F_%H%M%S")
+    BACKUP_TAR_FILE=${BACKUP_DIR}/snapshot_db_kuberesources_$DATESTRING.tar
+    SNAPSHOT_FILE="${BACKUP_DIR}/snapshot.db"
```
hexfusion: should both files have timestamp if we are going this route?

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
```diff
@@ -45,7 +55,12 @@ contents:
       dl_etcdctl
       backup_etcd_client_certs
       backup_manifest
+      backup_latest_kube_static_resources
       snapshot_data_dir
+      tar rf ${BACKUP_TAR_FILE} -C ${BACKUP_DIR} ${SNAPSHOT_FILE#$BACKUP_DIR}
+      gzip ${BACKUP_TAR_FILE}
+      rm -f ${SNAPSHOT_FILLE}
```
hexfusion: `SNAPSHOT_FILLE` typo?

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
retroflexer: Only the combined tar file bundle will have the timestamp. The snapshot.db needs to have this exact name so that we can extract it separately when restoring.

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
retroflexer: Thanks. `rm -f` is silent, and I didn't see the error message.

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
alaypatel07: how about creating a time-stamped directory inside the backup directory and putting the files in there?

something equivalent to
```
DATESTRING=$(date "+%F_%H%M%S")
BACKUP_DIR="$1/$DATESTRING"
mkdir -p BACKUP_DIR
```
This will make sure we don't overwrite the backups every time we run the script

retroflexer: If the user wants he can give a backup directory that is timestamped. Since there is going to be a combined tar file, there is no question of overwriting the file. Let me modify to create the snapshot file in a temp directory and then append to the tar bundle.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
retroflexer: I am working on deleting only later revisions. Look for another push.

templates/master/00-master/_base/files/usr-local-bin-etcd-snapshot-backup-sh.yaml
alaypatel07: We ask the users to copy the snapshot file to other hosts. If we tar the snapshot along with the KAS assets are we going to ask users to copy KAS assets as well? I ask because I am not sure if KAS assets are node dependent or not, @retroflexer can you please confirm that?
 
IIRC the static pod yaml is in the assets directory which might differ from host to host in which case untaring the snapshot would just be an extra step a user has to do before scping for no added advantage.

retroflexer: Yeah. I am not sure either.

retroflexer: Checking with sttts.

retroflexer: On my current installation, they appear to be identical.

retroflexer:
I just did a dummy push to get a newer CI image

hexfusion:
/hold cancel 

Manual tests confirmed

alaypatel07:
@retroflexer since this is going into master, do we wanna remove the `Bug 1783445` from title?

retroflexer:
@hexfusion I created a new PR for 4.4 with the changes you had suggested here. Thanks.
 https://github.com/openshift/machine-config-operator/pull/1346

PR #1346, opened 2019-12-18
DR: Cleanup and comment backup and restore scripts

retroflexer:
**- What I did**
Cleaned up and commented recent changes for readability and easier troubleshooting.
**- How to verify it**
Backup etcd and restore it as described in the OCP documentation.

Clean up of backup and restore DR scripts.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -38,8 +38,8 @@ contents:
       if [ -f "$ASSET_DIR/backup/etcd-ca-bundle.crt" ] && [ -f "$ASSET_DIR/backup/etcd-client.crt" ] && [ -f "$ASSET_DIR/backup/etcd-client.key" ]; then
          echo "etcd client certs already backed up and available $ASSET_DIR/backup/"
       else
-        STATIC_DIRS=($(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*))
-        if [ "$?" -ne 0 ]; then
+        STATIC_DIRS=($(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*)) || true
+        if [ -z "${STATIC_DIRS}" ]; then
           echo "error finding static-pod-resources"
           exit 1
         fi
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
Adding a hold, @alaypatel07 / @hexfusion  please remove once you are satisfied with the e2e results.

/hold

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -38,7 +38,7 @@ contents:
       if [ -f "$ASSET_DIR/backup/etcd-ca-bundle.crt" ] && [ -f "$ASSET_DIR/backup/etcd-client.crt" ] && [ -f "$ASSET_DIR/backup/etcd-client.key" ]; then
          echo "etcd client certs already backed up and available $ASSET_DIR/backup/"
       else
-        STATIC_DIRS=($(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*))
+        STATIC_DIRS=($(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*)) || true
         if [ "$?" -ne 0 ]; then
```
hexfusion: This won't work as expected now in a failure case.

```
#!/bin/bash
  
  set -o errexit
  set -o pipefail
  TEST=($(ls -td ./does/not/exist )) || true
  echo "return is $?"
  return is 0
```
Maybe you want to just wrap this together, one idea would be something like.

```
STATIC_DIRS=($(ls -td "${CONFIG_FILE_DIR}"/static-pod-resources/kube-apiserver-pod-[0-9]*)) || {
  echo "error finding static-pod-resources"
  exit 1
}}
```

retroflexer: Thanks. Fixed it.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -69,9 +69,17 @@ contents:
           echo "error finding static-pod-resources"
           exit 1
       fi
+
+      # tar up the static kube resources, with the path relative to CONFIG_FILE_DIR
       tar -cpf $BACKUP_TAR_FILE -C ${CONFIG_FILE_DIR} ${LATEST_STATIC_POD_DIR#$CONFIG_FILE_DIR/}
     }
 
+    append_snapshot_to_tar_and_gzip() {
```
hexfusion: question: if this function fails do you feel we will provide enough information to debug?

retroflexer: Well. If this function fails, tar or gzip should report error, which should be available to the end user. Then, the trap removes whatever files we backed up. (Also the set +x debug tells us we came this far and failed but end user doesn't run with +x). I am not sure adding more error checking here is any useful.

hexfusion: OK that makes sense, thanks

retroflexer:
> /test e2e-aws-disruptive

Thanks.

PR #1351, opened 2019-12-19
Bug 1785281: Discard audit messages from journald

fbac:
Fixes bz#1785219 with the simplest approach.

Since this is a corner case [1] which probably few people will hit, this is the simplest approach to fix the issue, also it will be short-lived as RHEL7.7 comes with logger supporting --journald flag.

[1] RHEL7.6 + audit enabled + logger not supporting --journald flag


LorbusChris:
No need to use the bot for cherrypicking -- 4.2 and 4.3 PRs already exist

PR #1357, opened 2020-01-08
Bug 1789581: Fix osImageURL upgrade race

runcom:
This patches does two things:

- ~makes sure we create the osimageurl configmap _*before*_ the new MCO deployment manifest is applied~
- version the osimageurl configmap

*NOTE*

- This patch requires a 4.2.z patch to version the osimageurl config map there as well (done)
- ~it might be needed to version the ControllerConfig as well (???)~

related bug https://bugzilla.redhat.com/show_bug.cgi?id=1786993


cgwalters:
Sad we're not :fox_face: ing it...

pkg/operator/sync.go
```diff
@@ -686,6 +686,11 @@ func (optr *Operator) getOsImageURL(namespace string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	releaseVersion := cm.Data["releaseVersion"]
+	optrVersion, _ := optr.vStore.Get("operator")
+	if releaseVersion != optrVersion {
```
ashcrow: Is there any valid cases where the configmap from a previous setup would either be unversioned or not match while transitioning to versioned?

cgwalters: Previous configmaps won't be versioned; but that's fine because Golang doesn't have `Option<>` so the `releaseVersion` will be the empty string, which won't match.  Which is what we want - we'll ignore the previous unversioned configmap.

runcom: notice there would still be a possible race where the 4.2 MCO reads the new 4.3 osimageurl configmap, that's why I belive we need to patch 4.2 as well to take versioning into account

cgwalters:
I haven't traced through the code and the logs from the race, but I completely believe you have and this looks highly likely to fix the problem.

I looked at previous variants of this like https://github.com/openshift/machine-config-operator/pull/1198 and the original was https://github.com/openshift/machine-config-operator/pull/527/commits/835278f6afa784e7705110fe12d4a2eec5c5ca3f

Looking over the manifests we install, I think this is the last thing that is input to the MachineConfig that isn't versioned.  The other CVO-installed manifests are pure kube level stuff like RBAC rules etc.

/approve
/lgtm

ashcrow:
Prometheus was not happy.

/test ci/prow/e2e-aws

pkg/operator/sync.go
ericavonb: > notice there would still be a possible race where the 4.2 MCO reads the new 4.3 osimageurl configmap and could generate a new rendered mc with a newer osimageurl, that's why I believe we need to patch 4.2 as well to take versioning into account

I would rather we address that race condition directly rather than trying to patch 4.2 which requires us to enforce that 4.2.x (unpatched) upgrades through 4.2.x+1(patched) before going to 4.3. That's possible in theory but seems to be cumbersome in practice.

Some thoughts:
1. Version the key in the `osimageurl` configmap (e.g. `osImageURL-0.0.1-snapshot: ...`) so that each MCO release has its own image reference. This would address this upgrade issue by changing the key name in 4.3 and protects us from any future upgrade race-conditions and failures.  It would also allow you to do some weird cross-version testing, like if you wanted to have multiple MCOs referencing different osimages for some reason. 
2. Set the osimageurl value on the MCO deployment directly - it can be a flag passed in to the operator, the release infra will replace the image reference in the deployment manifest the same as it does in the configmap. If there are other reasons for using a configmap, the MCO deployment can be changed to reference a configmap mounted in just like all the other images. This would reduce any version issues to mismatches in the CC/MCC which we've dealt with before.
3. Make MCO updates atomic by putting everything that always and only updates together together. The osimageurl, MCO, MCC controllers, and MC templates. https://github.com/openshift/machine-config-operator/issues/878

sinnykumari:
Haven't really tested this out. Looks sane, also we did something similar in https://github.com/openshift/machine-config-operator/pull/1198 .
/lgtm

kikisdeliveryservice:
aws failure has nothing to do with this pr:
"Error: Error waiting for AMI (ami-0597f0ba49a38f182) to be ready: timeout while waiting for state to become 'available' (last state: 'pending', timeout: 40m0s)"

PR #1358, opened 2020-01-08
Bug 1789565: [release-4.2] Fix osImageURL upgrade race

runcom:
same as #1357 but for 4.2 to test it in CI (4.2->4.2)

### DIFF
--- a/install/0000_80_machine-config-operator_05_osimageurl.yaml
+++ b/install/0000_80_machine-config-operator_05_osimageurl.yaml
@@ -4,6 +4,7 @@ metadata:
   name: machine-config-osimageurl
   namespace: openshift-machine-config-operator
 data:
+  releaseVersion: 0.0.1-snapshot
   # The OS payload, managed by the daemon + pivot + rpm-ostree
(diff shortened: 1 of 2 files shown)


crawford:
Why isn't https://github.com/openshift/machine-config-operator/pull/1359/commits/031fee633ed7a92cd5b4dc9093e809334d3b6ec9 needed as well?

runcom:
> Why isn't [031fee6](https://github.com/openshift/machine-config-operator/commit/031fee633ed7a92cd5b4dc9093e809334d3b6ec9) needed as well?

That has been removed as it can cause a race with the old MCO if we create the osimageurl CM before starting the new MCO. That case is similar to what happened during the 4.1->4.2 etcd bug.
It should be the MCO understanding and waiting on the correct osimageurl to show up, not the other way around.

runcom:
@kikisdeliveryservice @yuqi-zhang @sinnykumari @ericavonb ptal

wking:
[VERIFIED in 4.3][1]:

/bugzilla refresh

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1786993#c41

PR #1368, opened 2020-01-10
build: do not export GOFLAGS

runcom:
GOFLAGS env shouldn't be exported as it's meant for something else but we're using
it for "tags" - that's why the build fails on rhel (not sure why it started recently).

Tested in the golang-1.12 container and works fine, should further fix the ocp builds.

Reference to GOFLAGS env: https://tip.golang.org/cmd/go/#hdr-Environment_variables

```
GOFLAGS
	A space-separated list of -flag=value settings to apply
	to go commands by default, when the given flag is known by
	the current command. Each entry must be a standalone flag.
	Because the entries are space-separated, flag values must
	not contain spaces. Flags listed on the command line
	are applied after this list and therefore override it.
```

ref #1364 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ export GOPATH=$(shell echo $${GOPATH:-$$HOME/go})
 export GO111MODULE
 export GOPROXY=https://proxy.golang.org
 
-GOFLAGS = "containers_image_openpgp exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_ostree_stub"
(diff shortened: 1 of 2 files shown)


cgwalters:
Failures seem unrelated
/retest

runcom:
@kikisdeliveryservice @yuqi-zhang @sinnykumari @ericavonb ptal

cgwalters:
Glanced at some other Go projects (podman, openshift/installer) and it does look like what we were doing was just wrong.  So many copy+pasted and tweaked `Makefiles` that drift over time for go projects...
/lgtm

PR #1370, opened 2020-01-14
[FCOS] Update vendor and fix small go mod issue

sgreene570:
Updated the `vendor` folder for the FCOS branch and ran into a snag with `go mod vendor` and permissions. 

Fixed the `update-codegen.sh` script to explicitly call bash rather than `./` since the vendor'ed script won't have the execute bit set.

See https://github.com/golang/go/issues/34965


sgreene570:
Should we look into some way of verifying that `go mod vendor` stays in sync with `go.mod`?

LorbusChris:
@sgreene570 thank you! I'll rebase the fcps branch again for the Beta release and will include a new commit to update deps there anyway. Can you make this PR just include the first commit?

LorbusChris:
Actually, it doesn't matter, I'll cherry-pick later.
/override ci/prow/e2e-aws
/lgtm

LorbusChris:
Oh well, I'm still not able to override due to a prob bug with upper case letters in GH username.

PR #1375, opened 2020-01-16
WIP baremetal ipv6 fixes

hardys:
WIP fixes for ipv6 deployments with IPI baremetal, this is currently a test branch based on release-4.3 as we're testing with a 4.3 based build, but when we've figured out the fixes needed I'll re-submit the changes to master.

### DIFF
--- a/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
+++ b/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
@@ -5,3 +5,5 @@ contents:
   inline: |
     [main]
     dhcp=dhclient
+    [connection]
+    ipv6.dhcp-duid=ll


cgwalters:
My main question is - is this something that's just convenient for testing?  Or is it something we really see as required for real (non-libvirt) deployments?

hardys:
> My main question is - is this something that's just convenient for testing? Or is it something we really see as required for real (non-libvirt) deployments?

Good question, I think it's required for all deployments, because we expect the hostname to be provided via DHCP, and we also expect to have static reservations which implies a predictable client ID (at least for dnsmasq, I've not yet investigated other DHCP servers, but I do know there are customer requirements to enable this working with dnsmasq).

@celebdor would you agree?  I think at this point we expect ipv6 to be a stateful replacement for ipv4, and as such we need to have predictable IP reservations and thus a duid?

It seems that typical ipv6 server environments are dual-stack, and in that case you can imagine the ipv4 DHCP would be used to provide this information (with a potentially more stateless setup for ipv6 with a non-predictable server IP and router advertisements?), but in this case we know we have to make this work in a single-stack environment, so that isn't possible.

kikisdeliveryservice:
@hardys since this was a test PR and the ipv6 stuff is done - do you still need this PR open?

PR #1381, opened 2020-01-17
Update kubelet.config for single-stack IPv6

danwinship:
**- What I did**
In a single-stack IPv6 environment, update the `kubelet.config` to pass `--node-ip ::` to tell it to pick an IPv6 address rather than an IPv4 address for the node address. (Depends on https://github.com/openshift/origin/pull/24401 which is why this was not merged with the other IPv6 changes before (#1211).)

Specifically, this fixes the problem of trying to bring up a single-stack IPv6 cluster on a node that happens to have IPv4 addresses that you aren't using. (eg, this is necessary for testing single-stack IPv6 on AWS or Azure, since there is no way to bring up a node there there doesn't have an IPv4 address.) By default, if you don't specify an explicit `--node-ip`, and there are IPv4 node IPs addresses present, then kubelet will pick the node's default IPv4 address to be its primary IP, even if the cluster and service CIDRs are IPv6-only (which will then result in a cluster that doesn't work). The newly added `--node-ip ::` option tells it to pick the node's default IPv6 address instead.

**- How to verify it**
... You can't easily. But this has been part of the IPv6 testing environment. And it has unit tests!

In a single-stack IPv6 environment, kubelet will be configured to pick an IPv6 address rather than an IPv4 one for the default node IP (even if there are IPv4 addresses on the node).

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -105,6 +105,9 @@ type ControllerConfigSpec struct {
 	// infra holds the infrastructure details
 	// TODO this makes platform redundant as everything is contained inside Infra.Status
 	Infra *configv1.Infrastructure `json:"infra"`
+
+	// kubeletIPv6 is true to force a single-stack IPv6 kubelet config
+	KubeletIPv6 bool `json:"kubeletIPv6,omitempty"`
 }
 
 // ControllerConfigStatus is the status for ControllerConfig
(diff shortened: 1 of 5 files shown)


cgwalters:
/approve
lgtm

kikisdeliveryservice:
As you can see our testing i somehow broken at the moment =/

LGTM

cc: @rphillips for viz

runcom:
yeah, tests broken but we're working on it

/lgtm

PR #1382, opened 2020-01-17
Bug 1782153: templates: Add a "this is generated" note

vrutkovs:
This is a best practice, and particularly in this case
when we're intending to modify a config file, it can't be identical
to the shipped version.

This avoids having libostree replace it with a newer version.

Cherry-pick of #1320 on release-4.1.


cgwalters:
You'll need to run a command to update the test copies of the templates, `git log` should find it.

templates/worker/01-worker-container-runtime/_base/files/container-storage.yaml
```diff
@@ -117,4 +120,3 @@ contents:
     # attempt to complete IO when ENOSPC (no space) error is returned by
     # underlying storage device.
     # xfs_nospace_max_retries = "0"
-
```
kikisdeliveryservice: can we put this newline back?

templates/master/01-master-container-runtime/_base/files/container-storage.yaml
```diff
@@ -117,4 +120,3 @@ contents:
     # attempt to complete IO when ENOSPC (no space) error is returned by
     # underlying storage device.
     # xfs_nospace_max_retries = "0"
-
```
kikisdeliveryservice: can we put this newline back?

vrutkovs: Done, thanks!

kikisdeliveryservice:
thanks @vrutkovs !

/appprove

will let @cgwalters lgtm...

kikisdeliveryservice:
err... 

/approve

cgwalters:
/approve
Thanks for doing this!

PR #1384, opened 2020-01-18
Fix update codegen script to be compatible with go mod vendor

LorbusChris:
Manual cherry-pick from the fix @sgreene570 added on the `fcos` branch.

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -6,7 +6,7 @@ set -o pipefail
 
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 
-${SCRIPT_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh all \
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
The failure seems to be pretty consistently the same, for example, on upgrade:
` ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []`
Not sure if that's related though... I see tests passing on some other PRs so not sure why this is having trouble

LorbusChris:
I don't see how that could be related :/

sinnykumari:
> I don't see how that could be related :/

It's not this PR specific, gcp tests are failing for other PRs in master too.

sgreene570:
https://github.com/openshift/machine-config-operator/pull/1370 for context, this change won't affect master until the `vendor` folder is updated.

PR #1388, opened 2020-01-20
Bug 1790823: [IPI][BAREMETAL] Verify that mdns-publisher starts after hostname is set

yboaron:
This change verifies that mdns-publisher pod advertises node's
name only after hostname being set.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
+++ b/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
@@ -24,6 +24,21 @@ contents:
         hostPath:
           path: "/etc/mdns"
       initContainers:
+      - name: verify-hostname
+        image: {{ .Images.baremetalRuntimeCfgImage }}
+        env:
+          - name: DEFAULT_LOCAL_HOSTNAME
+            value: "localhost"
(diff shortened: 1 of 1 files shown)


templates/common/baremetal/files/baremetal-mdns-publisher.yaml
```diff
@@ -24,6 +24,21 @@ contents:
...
+        image: {{ .Images.baremetalRuntimeCfgImage }}
+        env:
+          - name: DEFAULT_LOCAL_HOSTNAME
+            value: "localhost"
+        command:
+        - "/bin/bash"
+        - "-c"
+        - |
+          #/bin/bash
+          while [ $(hostname) == "$DEFAULT_LOCAL_HOSTNAME" ]
```
cybertron: Maybe want to put quotes around $(hostname)? I don't know if there would be anything in a hostname that would mess up the check, but this being bash I generally err on the side of caution.

yboaron: @cybertron - Done. 10x!

PR #1390, opened 2020-01-20
Bug 1793144: kubelet: add log level environment variable

rphillips:
**- What I did**
Create an environment variable for the kubelet log level, so it can be overridden in /etc/kubernetes/kubelet-env across reboots.

**- How to verify it**

```NONE```

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.yaml
@@ -10,6 +10,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  Environment="KUBELET_LOG_LEVEL=3"
   EnvironmentFile=/etc/os-release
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: templates/master/01-master-kubelet/_base/units/kubelet.yaml [hunk 1]
```diff
@@ -10,6 +10,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+  Environment="KUBELET_LOG_LEVEL=3"
```
mrunalp: Did we test that we can override this in one of the environment files?

rphillips: yes. it works fine.

umohnani8:
LGTM

umohnani8:
Maybe add an example in the doc somewhere on how to use it?

cgwalters:
xref https://github.com/openshift/machine-config-operator/pull/1439

I think what we *really* want here is `/etc/kubelet.conf.d`.

PR #1392, opened 2020-01-20
OpenStack: Add ca cert to disk on all nodes so kubelet can start up

iamemilio:
**- What I did** Add support for Ca certs being added to the cloud-provider-config configmap, and writing those certs to where kubelet will know to look for them. This is associated with https://github.com/openshift/installer/pull/2932/files, and is part of a greater scheme to add support for self signed platform (openstack) CA certificates outlined here: https://docs.google.com/document/d/1bKXArO_dtIOQMVqV35GTmUdzoRQAgUqIVMbO9kpfknI/edit?folder=0AFMASrm3oxfNUk9PVA#heading=h.9jxn8ewfjnas

**- How to verify it** For all platforms excluding openstack, this should have no affect on the way the MCO runs. This should be verifiable by the MCO being able to pass its normal unit and e2e test suite. For openstack, this is being tested in a specialized cluster in combination with a number of other pending code changes.

### DIFF
--- a/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
@@ -241,6 +241,11 @@ func (in *ControllerConfigSpec) DeepCopyInto(out *ControllerConfigSpec) {
 		*out = make([]byte, len(*in))
 		copy(*out, *in)
 	}
+	if in.CloudProviderCAData != nil {
+		in, out := &in.CloudProviderCAData, &out.CloudProviderCAData
+		*out = make([]byte, len(*in))
+		copy(*out, *in)
+	}
 	if in.AdditionalTrustBundle != nil {
 		in, out := &in.AdditionalTrustBundle, &out.AdditionalTrustBundle
 		*out = make([]byte, len(*in))
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -29,7 +29,7 @@ func RenderBootstrap(
 	proxyFile,
 	clusterConfigConfigMapFile,
 	infraFile, networkFile,
-	cloudConfigFile,
+	cloudConfigFile, cloudProviderCAFile,
 	etcdCAFile, etcdMetricCAFile, rootCAFile, kubeAPIServerServingCA, pullSecretFile string,
 	imgs *Images,
 	destinationDir string,
@@ -47,6 +47,9 @@ func RenderBootstrap(
 	}
 	if kubeAPIServerServingCA != "" {
 		files = append(files, kubeAPIServerServingCA)
+	}
+	if cloudProviderCAFile != "" {
+		files = append(files, cloudProviderCAFile)
 	}
 	for _, file := range files {
 		data, err := ioutil.ReadFile(file)
@@ -128,6 +131,10 @@ func RenderBootstrap(
 	if _, ok := filesData[kubeAPIServerServingCA]; ok {
 		bundle = append(bundle, filesData[kubeAPIServerServingCA]...)
 		spec.KubeAPIServerServingCAData = filesData[kubeAPIServerServingCA]
+	}
+	// Set the cloud-provider CA if given.
+	if data, ok := filesData[cloudProviderCAFile]; ok {
+		spec.CloudProviderCAData = data
 	}
 
 	spec.EtcdCAData = filesData[etcdCAFile]
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -205,6 +205,11 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 			return err
 		}
 		spec.CloudProviderConfig = cc
+
+		caCert, err := optr.getCAsFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, "ca-bundle.pem")
+		if err == nil {
+			spec.CloudProviderCAData = caCert
+		}
 	}
 
 	//TODO: alaypatel07 remove after cluster-etcd-operator deployed via CVO as Managed
--- a/templates/common/_base/files/cloud-provider-ca.yaml
+++ b/templates/common/_base/files/cloud-provider-ca.yaml
@@ -1,0 +1,8 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
+contents:
+  inline: |
+{{if .CloudProviderCAData -}}
+{{.CloudProviderCAData | toString | indent 4}}
+{{end -}}
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -22,6 +22,7 @@ var (
 		baremetalRuntimeCfgImage  string
 		cloudConfigFile           string
 		configFile                string
+		cloudProviderCAFile       string
 		corednsImage              string
 		destinationDir            string
 		etcdCAFile                string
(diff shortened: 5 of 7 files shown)


templates/common/openstack/files/openstack-cacert-file.yaml
```diff
@@ -0,0 +1,6 @@
+filesystem: "root"
```
mandre: Is there a way to only copy the cert to `/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem` when it is set (non empty)?
The clouds.conf shouldn't have the `ca-file` option in this case so we're not breaking anything.

pkg/operator/sync.go
```diff
@@ -206,6 +206,16 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		spec.CloudProviderConfig = cc
 	}
 
+	if spec.Platform == "openstack" {
+		if infra.Spec.CloudConfig.Name != "" {
+			openstackCACert, err := optr.getCAsFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, "ca-bundle.pem")
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+			spec.OpenStackCACertificate = openstackCACert
```
mandre: What happens when the `ca-bundle.pem` key is not present in the config map, or when the config map isn't there?

Should this be enclosed in an else clause?

pkg/operator/sync.go
```diff
@@ -206,6 +206,16 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		spec.CloudProviderConfig = cc
 	}
 
+	if spec.Platform == "openstack" {
+		if infra.Spec.CloudConfig.Name != "" {
+			openstackCACert, err := optr.getCAsFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, "ca-bundle.pem")
```
mandre: Shouldn't it be hardcoded to the `cloud-provider-config` configmap?

kikisdeliveryservice:
from the \prow verify log:
``` hack/../pkg is out of date. Please run make update
make: *** [verify] Error 1
--- 
```

pkg/operator/sync.go
```diff
@@ -204,6 +204,12 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 			return err
 		}
 		spec.CloudProviderConfig = cc
+
+		caCert, err := optr.getCAsFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, "ca-bundle.pem")
+		if err != nil && !apierrors.IsNotFound(err) {
```
mandre: `apierrors.IsNotFound(err)` catches cases where the configmap doesn't exist, but doesn't cover the case where "ca-bundle.pem" is missing from the configmap.

cgwalters:
Hmm, what is consuming this CA set?  Just kubelet?  And more specifically just the cloud provider code in the kubelet?

Won't other components that talk to the IaaS like the router, machine API etc need this too?

cgwalters:
As discussed on Slack you'll also need to change the bootstrap code path.

iamemilio:
This patch alone is for kubelet. We needed to update the cloud-config to look for a ca-certificate, which kubelet consumes. We are patching a bunch of services, which we are tracking/planning here: https://docs.google.com/document/d/1bKXArO_dtIOQMVqV35GTmUdzoRQAgUqIVMbO9kpfknI/edit?folder=0AFMASrm3oxfNUk9PVA#

cgwalters:
>  We are patching a bunch of services, which we are tracking/planning here: 

This seems like a good thing to file as an https://github.com/openshift/enhancements/ - IMO, private Google Docs are an anti-pattern for FOSS project development.  If you want a collaborative editor before sumitting the enhancement, I recommend [hackmd.io](hackmd.io/).

iamemilio:
We had drafted 2 previous enhancements, but we had to abandon each one. The google doc was born from necessity, since we had run out of time, and needed to get this fix to our customers. I can file another enhancement retroactively now if you think its beneficial @cgwalters

pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
```diff
@@ -241,6 +241,11 @@ func (in *ControllerConfigSpec) DeepCopyInto(out *ControllerConfigSpec) {
 		*out = make([]byte, len(*in))
 		copy(*out, *in)
 	}
+	if in.CloudProviderCAData != nil {
+		in, out := &in.CloudProviderCAData, &out.CloudProviderCAData
+		*out = make([]byte, len(*in))
+		copy(*out, *in)
```
wking: The `in` and `out` indirection makes this harder for me to read, although that's obviously subjective.  What about:

```go
if in.CloudProviderCAData != nil {
  out.CloudProviderCAData = make([]byte, len(in.CloudProviderCAData))
  copy(out.CloudProviderCAData, in.CloudProviderCAData)
}
```

?

wking: Ah, I guess the `in` and `out` indirection is part of an established code pattern around here; never mind ;).

pkg/operator/bootstrap.go
```diff
@@ -129,6 +132,10 @@ func RenderBootstrap(
 		bundle = append(bundle, filesData[kubeAPIServerServingCA]...)
 		spec.KubeAPIServerServingCAData = filesData[kubeAPIServerServingCA]
 	}
+	// Set the cloud-provider CA if given.
+	if _, ok := filesData[cloudProviderCAFile]; ok {
+		spec.CloudProviderCAData = filesData[cloudProviderCAFile]
```
wking: you're also matching an existing pattern here, but I would have expected:

```go
if data, ok := filesData[cloudProviderCAFile]; ok {
  spec.CloudProviderCAData = data
}
```

I'm fine with any of "we're just pattern matching, this is fine", "we'll diverge from the pattern to avoid a doubled lookup (which maybe gets optimized out in compilation and won't take long anyway", or "nice.  And we'll fix `kubeAPIServerServingCA` above as well".  And even if I'm not fine with something, I'm not a maintainer here, so don't worry about it ;)

pkg/apis/machineconfiguration.openshift.io/v1/zz_generated.deepcopy.go
iamemilio: I prefer your code, however 
```
// +build !ignore_autogenerated

// Code generated by deepcopy-gen. DO NOT EDIT.
```

templates/common/_base/files/cloud-provider-ca.yaml
```diff
@@ -0,0 +1,8 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
+contents:
+  inline: |
+{{if .CloudProviderCAData -}}
```
wking: is the kubelet fine if you write empty content here?

iamemilio: Yeah, it doesn't seem to care. We tested this in clusters with and without the cert and didnt have trouble with kubelet after this fix.

pkg/operator/bootstrap.go
iamemilio: I like that

kikisdeliveryservice:
I'd like @mandre @Fedosin @deads2k to review before MCO signs off.

Also cc: @runcom are there any objections to this addition?

kikisdeliveryservice:
Is there any e2e that will cover this that can be run via test `e2e-openstack` ? (I'm not sure just asking)

crawford:
Looks good to me. This appears to be safe to merge even if we miss the accompanying installer PR.

iamemilio:
We are testing this manually, and were able to deploy an e2e succesfully. The code should not change how the installer or mco runs, unless they run it with the new functionality that we added in the linked installer patch.

kikisdeliveryservice:
`level=debug msg="Still waiting for the cluster to initialize: Cluster operator machine-config is reporting a failure: Failed to resync 0.0.1-2020-01-22-101826 because: ca-bundle.pem not found in openshift-config/cloud-provider-config`

```
I0122 10:51:37.212833       1 operator.go:246] Starting MachineConfigOperator
E0122 10:52:50.062613       1 operator.go:312] ca-bundle.pem not found in openshift-config/cloud-provider-config
```

Just to confirm this is expected to fail until the installer pr lands?

templates/common/_base/files/cloud-provider-ca.yaml
mandre: Note that the clouds.conf doesn't reference the CA file when it's not provided so kubelet is not even looking at that empty file.

mandre:
> Just to confirm this is expected to fail until the installer pr lands?

Definitely not, thanks for noticing. This PR shouldn't depend on any way on the installer one (otherwise they would be inter-dependent). I'll fix this in the morning.

mandre:
@kikisdeliveryservice you were looking at an old job result and I think I've already fixed the error you mentioned. I've re-triggered the e2e-openstack job, let's see.

mandre:
@kikisdeliveryservice All good, the e2e-openstack job passes.

PR #1395, opened 2020-01-21
baremetal: IPv6 add dhcp-duid to NetworkManager config

hardys:
Adding ipv6.dhcp-duid=ll means we get a predictable
client ID so that reservations can be made on the
DHCP server - this is useful if you want to provide
the hostname via DHCP (although a reverse DNS lookup
should also work)

Reservations are also required to ensure the controlplane
IP of the masters won't ever change post-deployment
which will be necessary if the DNS lookup approach
is taken to set the hostnames.

**- What I did**

Updated the NetworkManager configuration, only for the baremetal (IPI) platform, so we specify the ipv6.dhcp-duid=ll option.  This enables a predictable (mac derived) client ID/DUID when deploying in an IPv6 environment.

**- How to verify it**

If testing with dev-scripts you can check via `virsh net-dhcp-leases baremetal` - on baremetal you can check the client ID in the lease for the nic connected to the controlplane network (`/var/lib/NetworkManager/dhclient6*.lease`)

In both cases the controlplane network must be configured for single-stack ipv6, and there are some other PRs open to enable that e2e with the default OpenShift builds. https://github.com/openshift-metal3/dev-scripts/pull/871 has some more details.

Updated the NetworkManager configuration, for the baremetal (IPI) platform to specify the ipv6.dhcp-duid=ll option.  This enables a predictable (mac derived) client ID/DUID.

### DIFF
--- a/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
+++ b/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
@@ -6,3 +6,5 @@ contents:
     [main]
     dhcp=dhclient
     rc-manager=unmanaged
+    [connection]
+    ipv6.dhcp-duid=ll


hardys:
https://github.com/openshift/machine-config-operator/pull/1375 has some related discussion

kikisdeliveryservice:
PTAL

/assign @celebdor 
/assign @bcrochet

rgolangh:
Its time for a well established common configuration for all 3 baremetal, openstack, ovirt. Possibly more platforms will benefit from it, vsphere, libvirt, and so on. Question is how to properly have platforms opt-in to such common configuration. One way might be symbolic links - but then we need to solve templating:
```
cat template/shared/foo.conf
key="{{ .ControllerConfig.Infra.Status.PlatformStatus.PLATFORM.SomeValue}}"

ls -la template/ovirt/
lrwxrwxrw ../share/foo.conf
```
so `PLATFORM` can be resolved by the internal `Platform` 

@runcom is there an issue on this topic? if not I'll open one

hardys:
@rgolangh agreed figuring out how to de-duplicate this would be good

> One way might be symbolic links - but then we need to solve templating:

If we can't solve the templating another approach would be to rework the dispatcher scripts to they are common to all platforms, but then source a per-platform file with the variables that are templated (and do nothing if such a file doesn't exist?)

kikisdeliveryservice:
Any reviews on this?

@celebdor @hardys @bcrochet

hardys:
> Any reviews on this?
> 
> @celebdor @hardys @bcrochet

This has been tested by several folks now via the https://github.com/openshift-kni/machine-config-operator/commits/4.3-ipv6 branch cc @russellb @derekhiggins - I think it's good to go provided @celebdor and @bcrochet agree

kikisdeliveryservice:
also ping: @celebdor :)

kikisdeliveryservice:
@hardys looks like you need to rebase this PR as another changed the same yaml.

derekhiggins:
/lgtm
works for me

hardys:
/retitle Bug 1798637: baremetal: IPv6 add dhcp-duid to NetworkManager config

PR #1396, opened 2020-01-21
baremetal: ipv6, switch to NM dispatcher for DNS VIP prepending

hardys:
The prepend via dhclient doesn't work via ipv6, so switch to a
NetworkManager dispatcher that runs after dhclient instead as a
workaround.

**- What I did**

Reworked the prepender implementation (with help from @celebdor) to avoid using the `prepend domain-name-servers` option in dhclient.conf - it seems this option only works for ipv4[1] and the suggested alternative of `prepend dhcp6.name-servers` also doesn't seem to work when the resolv.conf is managed by NetworkManager.

Instead we configure NetworkManager to no longer manage the resolv.conf directly, and rely on a dispatcher script which injects the necessary IP on master/worker nodes to correctly reference coredns for the baremetal platform.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=643890

**- How to verify it**

Deploy with ipv4 and ipv6, confirm that on the masters the DNS VIP is prepended to the resolv.conf, and on the workers the local nic IP for the controlplane network (not the DNS VIP) is configured.

Also check `sudo journalctl -b | grep prepender` to see the log output from the dispatcher scripts.

For the baremetal platform management of the resolv.conf is now handled via a NetworkManager dispatcher script, so that the necessary DNS server can be prepended for both ipv4 and ipv6 environments.


celebdor:
/assign @rgolangh 
/assign  @Fedosin 

You may want to copy the approach (specially looking to IPv6 support).

templates/worker/00-worker/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -0,0 +1,27 @@
...
+    case "$STATUS" in
+        up|down|dhcp4-change|dhcp6-change)
+        logger -s "NM resolv-prepender triggered by ${1} ${2}."
+        NAMESERVER_IP=$(/usr/local/bin/non_virtual_ip \
+            "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
+            "{{.Infra.Status.PlatformStatus.BareMetal.NodeDNSIP}}" \
+            "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
+        set +e
+        if [[ -n "$NAMESERVER_IP" ]]; then
+            logger -s "NM resolv-prepender: Prepending 'nameserver $NAMESERVER_IP' to other nameservers in /var/run/NetworkManager/resolv.conf"
```
rgolangh: The log says "to other nameservers in /var/run/..." but this actually updates /etc/resolv.conf

hardys: Ah good point, thanks - will update!

hardys: I think it's just worded incorrectly, should say prepending nameserver ... to other nameservers *from* /var/run/...

I'll reword it to clarify we're reading one file and writing the other

hardys: Ok I pushed an additional commit clarifying the log message

kikisdeliveryservice:
This is passing tests can we get a review @bcrochet / @celebdor ?

hardys:
> This is passing tests can we get a review @bcrochet / @celebdor ?

Also cc @russellb @derekhiggins who have tested this and found it to work OK - I've also re-tested it in a regular ipv4 environment.

hardys:
@celebdor @bcrochet  @russellb I think we may have to make this also apply to the bootstrap VM, particularly because the NM conf where we set rc-manager=unmanaged in this PR is common?

Also we set the prepend dhclient.conf option in the installer here:

https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/etc/dhcp/dhclient.conf#L7

So I think we need to remove that and ensure the prepender gets added for all nodes including bootstrap?

There are only common, master and worker templates directories, how does the precedence work, e.g can we have a common template (which will be for the bootstrap, or perhaps bootstrap and masters if we switch to using the DNS VIP on the bootstrap VM), then override it in the worker directory with a file of the same name?

russellb:
> @celebdor @bcrochet @russellb I think we may have to make this also apply to the bootstrap VM, particularly because the NM conf where we set rc-manager=unmanaged in this PR is common?
> 
> Also we set the prepend dhclient.conf option in the installer here:
> 
> https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/etc/dhcp/dhclient.conf#L7
> 
> So I think we need to remove that and ensure the prepender gets added for all nodes including bootstrap?
> 
> There are only common, master and worker templates directories, how does the precedence work, e.g can we have a common template (which will be for the bootstrap, or perhaps bootstrap and masters if we switch to using the DNS VIP on the bootstrap VM), then override it in the worker directory with a file of the same name?

It looks like the MCO templates don't apply to the bootstrap VM.  I think we need to add what's needed to the installer.

kikisdeliveryservice:
@hardys @russellb 
Just double check I can approve this for merge though right??

russellb:
> @hardys @russellb
> Just double check I can approve this for merge though right??

Yes.  There's a related change still needed, but for the bootstrap VM, and that'll be in the installer.  This PR is still ready to go.

PR #1397, opened 2020-01-21
Bug 1780442: DR: allow use of local etcdctl plus version check

hexfusion:
In airgap and proxy scenario, we need a simple way for allowing an alternative method of obtaining etcd client binary. In the future, this method could be considered the default. But for now we are allowing user to provide the binary in advance of running the script. This process would be part of the documentation as phase 1.

We also do not want old versions of the binary to be used so we added a version check.

This PR needs manual test/verification

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -21,15 +21,25 @@ contents:
 
     # download and test etcdctl from upstream release assets
     dl_etcdctl() {
-      GOOGLE_URL=https://storage.googleapis.com/etcd
-      DOWNLOAD_URL=${GOOGLE_URL}
+      ETCDCTL_BIN="${ASSET_DIR}/bin/etcdctl"
+      if [ -f "$ETCDCTL_BIN" ]; then
+        echo "etcdctl binary found.."
+        LOCAL_ETCD_VERSION=$($ETCDCTL_BIN version | grep 'etcdctl' | awk '{print "v" $3}') || true
+        if [ "$LOCAL_ETCD_VERSION" != "$ETCD_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


hexfusion:
cc @retroflexer

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -21,15 +21,25 @@ contents:
...
-        && tar -xzf $ASSET_DIR/tmp/etcd-${ETCD_VERSION}-linux-amd64.tar.gz -C $ASSET_DIR/shared --strip-components=1 \
-        && mv $ASSET_DIR/shared/etcdctl $ASSET_DIR/bin \
-        && rm $ASSET_DIR/shared/etcd \
-        && $ASSET_DIR/bin/etcdctl version
+      ETCDCTL_BIN="${ASSET_DIR}/bin/etcdctl"
+      if [ -f "$ETCDCTL_BIN" ]; then
+        echo "etcdctl binary found.."
+        LOCAL_ETCD_VERSION=$($ETCDCTL_BIN version | grep 'etcdctl' | awk '{print "v" $3}') || true
+        if [ "$LOCAL_ETCD_VERSION" != "$ETCD_VERSION" ]; then
+           echo "$ETCDCTL_BIN (${LOCAL_ETCD_VERSION}) is not expected (${ETCD_VERSION}) please remove or replace."
```
retroflexer: Nit though. This message is a bit confusing:
```./assets/bin/etcdctl (v3.3.16) is not expected (v3.3.17) please remove or replace.```
Could it be worded as:
```./assets/bin/etcdctl: Version mismatch. Expected v3.3.17, but found v3.3.16. Please remove or replace.```

hexfusion: done

Issue #1389, opened 2020-01-20
baremetal: haproxy doesn't listen on IPv6


russellb:
When trying a bare metal IPv6 cluster, my haproxy pods were crash looping because the health check was failing.  This was due to haproxy not listening on IPv6.  I made this config change to make it listen on Ipv4 and Ipv6, which seems like a reasonable default to switch to in all cases:

```
    # haproxy not configured to listen on IPv6
    fssh core@${MASTER} sudo sed -i \"s/bind :7443/bind :::7443 v4v6/\" /etc/haproxy/haproxy.cfg                                                                                             
    fssh core@${MASTER} sudo sed -i \"s/bind :50936/bind :::50936 v4v6/\" /etc/haproxy/haproxy.cfg                                                                                           
```

PR #1399, opened 2020-01-22
Bug 1793253: [release-4.3] HAProxy infra pod: Make v4/v6 compatible

celebdor:
Have the bind to all interfaces statements listen to both IPv4 and IPv6 so it
works both on single and dual stack deployments.

Relates: #1389 
Backport of: https://github.com/openshift/machine-config-operator/pull/1394


celebdor:
@cuppett I saw you retargetted for 4.3.z which makes the bug be marked as invalid by the bot. Is that correct?

kikisdeliveryservice:
@hardys @bcrochet This needs a LGTM

PR #1404, opened 2020-01-23
Bug 1792914: prevent hitting annotation max size limit on nodes

ericavonb:
**- What I did**
Fixes issue where an "annotation not found" error compounds via wrapping the node in the error message, growing the `machineconfiguration.openshift.io/reason` annotation until it hits the annotation max size limit, thereby preventing updates to that node's annotations by controllers.

- Remove node object from  "annotation not found" error and replace with just node name
- Truncate errors used in `machineconfiguration.openshift.io/reason` node annotation to prevent any that may contain large amounts of data from maxing out annotation size.

**- How to verify it**
- Stop MCO and MCC controllers
- Delete the `machineconfiguration.openshift.io/desiredConfig` annotation on a node and wait for the MCD to sync the update
- Check the node object for the annotation`"machineconfiguration.openshift.io/reason": "machineconfiguration.openshift.io/desiredConfig annotation not found in node '<name here>'"`

- Truncate errors used in `machineconfiguration.openshift.io/reason` node annotation to prevent hitting the max annotation size

Fixes: [Bug 1792914](https://bugzilla.redhat.com/show_bug.cgi?id=1792914)

### DIFF
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -1,6 +1,8 @@
 package daemon
 
 import (
+	"fmt"
+
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/internal"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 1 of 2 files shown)


runcom:
This works perfectly as a stopgap but we might want to revisit this and a better way to expose the reason 👍 

/approve
/lgtm

pkg/daemon/writer.go
```diff
@@ -128,11 +133,14 @@ func (nw *clusterNodeWriter) SetUnreconcilable(err error, client corev1client.No
 // Returns an error if it couldn't set the annotation.
 func (nw *clusterNodeWriter) SetDegraded(err error, client corev1client.NodeInterface, lister corev1lister.NodeLister, node string) error {
 	glog.Errorf("Marking Degraded due to: %v", err)
+	// truncatedErr caps error message at a reasonable length to limit the risk of hitting the total
```
kikisdeliveryservice: Thanks for this comment!!

PR #1407, opened 2020-01-24
test/e2e - Add e2e test for Realtime Kernel

sinnykumari:
Attempt to include e2e test for kernelType

PR #1412, opened 2020-01-28
test/e2e: Add rollback test for kernelType

sinnykumari:
Initially in e2e test for [kernelType](https://github.com/openshift/machine-config-operator/pull/1407),  we created a MachineConfig with `kerneltype:realtime`  to switch to Real Time kernel.

For better test coverage, we are now also making sure that switching back to `default` kernel also works fine. We are testing this by deleting the Real Time MachineConfig we applied earlier.

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -285,6 +285,11 @@ func TestKernelArguments(t *testing.T) {
 
 func TestKernelType(t *testing.T) {
 	cs := framework.NewClientSet("")
+	// Get initial MachineConfig used by the worker pool so that we can rollback to it later on
+	mcp, err := cs.MachineConfigPools().Get("worker", metav1.GetOptions{})
+	require.Nil(t, err)
+	workerOldMC := mcp.Status.Configuration.Name
+
 	kernelType := &mcfgv1.MachineConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   fmt.Sprintf("kerneltype-%s", uuid.NewUUID()),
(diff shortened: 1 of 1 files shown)


runcom:
/approve
/lgtm

@sinnykumari can you just add a one liner to the first post to make it clear what this does (title is ok anyway to me!)

sinnykumari:
> /approve
> /lgtm

Thanks for review!

> @sinnykumari can you just add a one liner to the first post to make it clear what this does (title is ok anyway to me!)

done

kikisdeliveryservice:
i :heart: this!

/retest

runcom:
the e2e-gcp-op failure seems related https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1412/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1113 somehow it did not roll back to what we expected?

test/e2e/mcd_test.go
```diff
@@ -322,6 +327,35 @@ func TestKernelType(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	t.Logf("Deleted MachineConfig %s", kernelType.Name)
+
+	// Wait for the mcp to rollback to previous config
+	if err := waitForPoolComplete(t, cs, "worker", workerOldMC); err != nil {
+		t.Fatal(err)
+	}
+	for _, node := range nodes {
```
runcom: oh yeah, this is what's happening - you're using the `nodes` slice from a previous fetch, you need to add:

```go
nodes, err := getNodesByRole(cs, "worker")
require.Nil(t, err)
```

So it grabs a fresh slice with the correct annotations

sinnykumari: ah, I didn't fetch it again because I thought we should be getting same worker nodes slice during re-fetch. But yeah,  annotations would have changed.

PR #1405, opened 2020-01-23
Bug 1794493: Update crio.yaml templates with correct values

umohnani8:
Update the templates with the corret defaults from cri-o.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

PR #1414, opened 2020-01-28
Bug 1794493: add ctrcfg e2e test

haircommander:
**- What I did**
Added a ctrcfg e2e test, to hopefully begin to prevent regressions 
To get this test to pass, some changes were also needed:
- refactor some of mcd_test.go (and move to utils_test.go) to not duplicate code between it and ctrcfg_test.go 
- pull in https://github.com/openshift/machine-config-operator/pull/1405, because without it the test fails
- revendor cri-o to have an updated config structure
- change a check on LogSizeMax which prevents an empty value ending up in the crio.conf

**- How to verify it**
`go test -v -run="TestContainerRuntimeConfigPidsLimit" ./test/e2e/`

added simple e2e test for ContainerRuntimeConfigs
revendor cri-o to have an updated config structure

~Note: this is WIP because I haven't seen the test pass yet (as I was working without the required patch). Instead of waiting for a cluster to provision manually, I'll now start letting prow do it. I also want to refactor the changes a bit~


runcom:
awesome 💪

runcom:
> Added a ctrcfg e2e test, to hopefully begin to prevent regressions with it. also pull in #1405, because without it the test fails

I'd just close #1405 and retitle this PR to account for the Bug as well

runcom:
uhm, not sure yet if failures in op tests are fully related

/retest

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+		ContainerRuntimeConfig: config,
+	}
+	spec.MachineConfigPoolSelector.MatchLabels[key] = value
+	ctrcfg.Spec = spec
+
+	_, err := cs.ContainerRuntimeConfigs().Create(ctrcfg)
+	require.Nil(t, err)
+	return func() {
+		cs.ContainerRuntimeConfigs().Delete(name, &metav1.DeleteOptions{})
```
runcom: why I love how this all works it's kinda suboptimal as we're ignoring the error - maybe handle it somehow?

haircommander: Do you think the test should fail if it fails in cleaning up? Or is logging it sufficient

runcom: I think this should be a continuation of the actual test where you make sure that rolling back (deleting) the ctrcfg brings the cluster back to normal (see #1412 for instance). TLDR: it should fail

haircommander: Sgtm

runcom: also still not sure why tests failed 🤔looks super weird but this is OT from this convo

runcom:
uhm I think e2e-gcp-op failures are now related

haircommander:
updated to have the verification the rollback worked, as well as a bit more debugging logs to test the gcp-op problems

runcom:
e2e-gcp-op still related to this now that we're back with that

haircommander:
yeah I've observed similar things when testing on my own. somehow the LogSizeMax value is being set to 0, and cri-o complains. Still working on it...

haircommander:
```

level=error
```


/retest

haircommander:
>     * Neither this nor #1405 provides any rationale. Is there a consensus on this being the right thing?
> 

agreed, I am still working on figuring out all of the failures. Upon this being the final version (still wip), I'll update it for explainations, but as I can see it:
Right now, `pkg/controller/container-runtime-config/*` defines a custom struct built off of the crio config. This struct is needs to have all of the options populated (for now, because we base off of a single config file). However, if the template doesn't define a default value for a value, that value will be evaluated as empty ("', 0, [], etc) after we render the ctrcfg. The problem is CRI-O can't discern whether that empty value is a mistake or purposeful (for instance, we set stream_address as empty on purpose). That's the first problem that is being fixed.

The second problem is because we are using an outdated version of the CRI-O config, and there are now some values that, when populated, behave better (version file, for instance). So the vendor bump is to fix that.

Finally, there's some weirdness with LogSizeMax. Somehow, it ends up as 0 after an update is applied. I still haven't gotten to the bottom of that, and once I do I'll add a test for each of the options, update the PR description, and remove the WIP

>     * If it is necessary to explicitly set all of those options, is there a process in place to keep updating the templates as new options are added, or defaults change?
For now, it is necessary and there's no process. However, there's light at the end of the tunnel. What we really should do is utilize [drop in config files](https://github.com/cri-o/cri-o/pull/3152) and define a structure that is a subset of the options that openshift needs (only the ones we care about configuring). When a ctrcfg comes in, we update that file with its values (keeping most out of the config, and thus most defaulting to the system crio.conf)

mtrmac:
> > ```
> > * If it is necessary to explicitly set all of those options, is there a process in place to keep updating the templates as new options are added, or defaults change?
> > ```
> 
> 
> For now, it is necessary and there's no process. However, there's light at the end of the tunnel. What we really should do is utilize [drop in config files](https://github.com/cri-o/cri-o/pull/3172) and define a structure that is a subset of the options that openshift needs (only the ones we care about configuring).

That would be great, and is a good reason not to invest _too much_ in the current hardcode-all-default design — if the drop in solution can happen soon enough.

If there’s a risk that this will persist over a CRI-O major release update (1.y to 1.y+1), it seems desirable to harden the MCO implementation, maybe to _at build time_ call https://github.com/cri-o/cri-o/blob/eb14e0feb01971bf88b2591b744b64dd60272fe5/pkg/config/config.go#L485  to generate the templates.

Or maybe do that in `updateCRIOConfig`: start with the return value of `DefaultConfig()` (and apply any MCO-specific hard-coded edits in the code) instead of reading and decoding a template file.

haircommander:
> > > ```
> > > * If it is necessary to explicitly set all of those options, is there a process in place to keep updating the templates as new options are added, or defaults change?
> > > ```
> > 
> > 
> > For now, it is necessary and there's no process. However, there's light at the end of the tunnel. What we really should do is utilize [drop in config files](https://github.com/cri-o/cri-o/pull/3172) and define a structure that is a subset of the options that openshift needs (only the ones we care about configuring).
> 
> That would be great, and is a good reason not to invest _too much_ in the current hardcode-all-default design — if the drop in solution can happen soon enough.
> 
> If there’s a risk that this will persist over a CRI-O major release update (1.y to 1.y+1), it seems desirable to harden the MCO implementation, maybe to _at build time_ call https://github.com/cri-o/cri-o/blob/eb14e0feb01971bf88b2591b744b64dd60272fe5/pkg/config/config.go#L485 to generate the templates.
> 
> Or maybe do that in `updateCRIOConfig`: start with the return value of `DefaultConfig()` (and apply any MCO-specific hard-coded edits in the code) instead of reading and decoding a template file.

my hope is to get this in for 4.4, for now just fixing the tests and adding some to make sure when we make the change we don't regress. Agreed that's a good method, but right now with the byte buffer and direct toml encoding, it *could* get messy starting with a default. I don't know the code that well, though, so I'm sure it's possible, but not easily nor quickly by me

test/e2e/utils_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"os/exec"
+	"testing"
+	"time"
+
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/test/e2e/framework"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
```
kikisdeliveryservice: you need to import corev1 or this will keep failing:

corev1 "k8s.io/api/core/v1"

kikisdeliveryservice: also do you need "k8s.io/apimachinery/pkg/labels"?

test/e2e/utils_test.go
```diff
@@ -0,0 +1,151 @@
+package e2e_test
+
+import (
```
kikisdeliveryservice: This should all be moved into the e2e/helpers dir

test/e2e/utils_test.go
haircommander: that's what you get when you don't run the tests first :sweat_smile:

test/e2e/utils_test.go
haircommander: framework? I'll look into adding them when I get more consistent passing tests

kikisdeliveryservice: no i mean that the file you created utils_test.go should be in dir e2e/helpers/

kikisdeliveryservice: also moving the functions should be in the commit message as well (or seperate commit)

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,87 @@
...
+			MatchLabels: make(map[string]string),
+		},
+		ContainerRuntimeConfig: config,
+	}
+	spec.MachineConfigPoolSelector.MatchLabels[key] = ""
+	ctrcfg.Spec = spec
+
+	_, err := cs.ContainerRuntimeConfigs().Create(ctrcfg)
+	require.Nil(t, err)
+	return func() error {
```
kikisdeliveryservice: just to double check im reading correctly if ctrcfg isn't created properly we return test error here? so we dont' have to check it before we wait for rendered config right?

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// label one node from the pool to specify which worker to update
+	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", mcpNameToRole(matchValue))
+	defer unlabelFunc()
+
+	// create an MCP to match the node we tagged
+	mcpCleanupFunc := createMCP(t, cs, matchValue)
+
+	// create our ctrcfg and attach it to our created node pool
+	cleanupCtrcfg := createCtrcfgWithConfig(t, cs, ctrcfgName, matchValue, cfg)
```
kikisdeliveryservice: why is this called cleanupCtrcfg?  maybe just generatedCtrcfg?

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// wait for the ctrcfg to complete its update
+	err = waitForPoolComplete(t, cs, matchValue, mcName)
+	require.Nil(t, err)
+
+	// cleanup ctrcfg and make sure it doesn't error
+	err = cleanupCtrcfg()
+	require.Nil(t, err)
+
+	t.Logf("Deleted ContainerRuntimeConfig %s", ctrcfgName)
```
kikisdeliveryservice: A proper rollback test would also be good here.

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+		ContainerRuntimeConfig: config,
+	}
+	spec.MachineConfigPoolSelector.MatchLabels[key] = ""
+	ctrcfg.Spec = spec
+
+	_, err := cs.ContainerRuntimeConfigs().Create(ctrcfg)
+	require.Nil(t, err)
+	return func() error {
+		return cs.ContainerRuntimeConfigs().Delete(name, &metav1.DeleteOptions{})
```
kikisdeliveryservice: Can't we move this up as part of the rollback test to ensure deleted CtrCfg will properly update cluster?

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// cache the old machine config value to compare against
+	oldMC := getMcName(t, cs, "worker")
+
+	// label one node from the pool to specify which worker to update
+	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", mcpNameToRole(matchValue))
+	defer unlabelFunc()
+
+	// create an MCP to match the node we tagged
+	mcpCleanupFunc := createMCP(t, cs, matchValue)
```
kikisdeliveryservice: why cleanup here? i dont understand?

test/e2e/containerruntimeconfig_test.go
kikisdeliveryservice: (after deletion chec that it is no longer applied etc... see an example here: https://github.com/openshift/machine-config-operator/pull/1412/files)

test/e2e/containerruntimeconfig_test.go
runcom: It’s because the function is returning a function which cleanup the resource, pretty standard pattern in golang

test/e2e/containerruntimeconfig_test.go
runcom: Just the return is a function itself which cleans things up

test/e2e/containerruntimeconfig_test.go
haircommander: should be called cleanupCtrcfgFunc to be more clear of its type

test/e2e/containerruntimeconfig_test.go
haircommander: I love this idiom because it declares that the function that provisions a resource is responsible for saying how that resource must be cleaned up. Then it's on the caller to decide when cleaning up is necessary

test/e2e/containerruntimeconfig_test.go
haircommander: proper rollback test added!

test/e2e/containerruntimeconfig_test.go
haircommander: yeah this is returning a function that will do the cleanup and return the error. it's called as cleanupCtrcfgFunc()

test/e2e/utils_test.go
haircommander: I will mention in the commit :+1:

adding them in a separate package means we'd have to create public functions for the test/e2e/ package to use the helpers. seems a bit weird to only have shared code move to helpers, and have some left in mcd_test, yet defining an entirely separate package and moving all the helper code seems out of scope of this PR.

If I update the commit, would you be alright with me leaving them in utils_test for this PR?

test/e2e/containerruntimeconfig_test.go
kikisdeliveryservice: gotcha

haircommander:
```
 --- FAIL: TestContainerRuntimeConfigLogSizeMax (3.30s)
    containerruntimeconfig_test.go:132: 
        	Error Trace:	containerruntimeconfig_test.go:132
        	            				containerruntimeconfig_test.go:47
        	            				containerruntimeconfig_test.go:19
        	Error:      	Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc00043e000), Stderr:[]uint8(nil)}
        	Test:       	TestContainerRuntimeConfigLogSizeMax
        	Messages:   	failed to cat /host/etc/crio/crio.conf on node ci-op-kgsmv-w-b-l8xtc.c.openshift-gce-devel-ci.internal 
```
why am I able to use oc debug from my host but not from the test :thinking:

runcom:
I don’t believe that ever worked , you can use oc rsh to execute commands within a given node

test/e2e/utils_test.go
kikisdeliveryservice: yeah leave it as is with updated commit, tests can be refactored later overall, i think. seems more important to get this all working right now. :)

test/e2e/containerruntimeconfig_test.go
kikisdeliveryservice: ok i understand esp with the rename! ty!

kikisdeliveryservice:
> I don’t believe that ever worked , you can use oc rsh to execute commands within a given node
@haircommander for rsh example see: https://github.com/openshift/machine-config-operator/blob/cef8e6b74eb601d240e73917893aa703f508087e/test/e2e/mcd_test.go#L232

test/e2e/containerruntimeconfig_test.go
```diff
@@ -0,0 +1,142 @@
...
+	mcName, err := waitForRenderedConfigFromCtrcfg(t, cs, matchValue, ctrcfgName)
+	require.Nil(t, err, "failed to render machine config from container runtime config")
+
+	// wait for the ctrcfg to complete its update
+	err = waitForPoolComplete(t, cs, matchValue, mcName)
+	require.Nil(t, err)
+
+	// verify value was changed
+	newConfValue := getValueFromCrioConfig(t, infraNode.Name, regexKey)
+	assert.Equal(t, newConfValue, expectedConfValue)
```
kikisdeliveryservice: maybe a message as an arg here?

haircommander: good idea, I should do that on all of them probably

haircommander:
> > I don’t believe that ever worked , you can use oc rsh to execute commands within a given node
> > @haircommander for rsh example see: https://github.com/openshift/machine-config-operator/blob/cef8e6b74eb601d240e73917893aa703f508087e/test/e2e/mcd_test.go#L232

does the mcd have /etc/crio/crio.conf of the host bind mounted? I need direct access to that file

kikisdeliveryservice:
> > > I don’t believe that ever worked , you can use oc rsh to execute commands within a given node
> > > @haircommander for rsh example see: https://github.com/openshift/machine-config-operator/blob/cef8e6b74eb601d240e73917893aa703f508087e/test/e2e/mcd_test.go#L232
> 
> does the mcd have /etc/crio/crio.conf of the host bind mounted? I need direct access to that file

seems like we have rootfs mounted looking at other tests.

haircommander:
we are both very eager to figure this out :D

haircommander:
alright, I **think** the test should pass, though there are two things I need to do still:

make sure the rollback is working properly (I'm not setting the mc we should roll back to correctly)
update my commit to reflect all I've done.

back again monday :sunglasses:

haircommander:
```
 === RUN   TestContainerRuntimeConfigPidsLimit
--- PASS: TestContainerRuntimeConfigPidsLimit (173.16s) 
```
Huzzah!

test/e2e/ctrcfg_test.go
```diff
@@ -0,0 +1,163 @@
...
+func runTestWithCtrcfg(t *testing.T, testName, regexKey, expectedConfValue string, cfg *mcfgv1.ContainerRuntimeConfiguration) {
+	cs := framework.NewClientSet("")
+	matchValue := fmt.Sprintf("%s-%s", testName, uuid.NewUUID())
+	ctrcfgName := fmt.Sprintf("ctrcfg-%s", matchValue)
+	poolName := fmt.Sprintf("node-%s", matchValue)
+	mcName := fmt.Sprintf("mc-%s", matchValue)
+
+	// label one node from the pool to specify which worker to update
+	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", mcpNameToRole(poolName))
+	defer unlabelFunc()
```
runcom: I think I understand what's going on here, follow https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md#removing-a-custom-pool

Basically, first thing unlabel the node, in this function you should wait for the pool to go idle as it's going back to worker - then only after go ahead and cleanup 1) MC 2) MCP

haircommander: rtfm @haircommander :roll_eyes: 

thanks!

haircommander:
```
 === RUN   TestContainerRuntimeConfigPidsLimit
--- PASS: TestContainerRuntimeConfigPidsLimit (215.71s) 
```
:smile: 
PTAL @kikisdeliveryservice @runcom @umohnani8 @mtrmac

runcom:
there seems to still be some kind of issue/race right? 🤔

haircommander:
> there seems to still be some kind of issue/race right? 

I don't know if it's an "issue". one could call the sleeps a bit hacky, but I don't know of a better way to make sure the operation has made it to the openshift api-server without them.

runcom:
> I don't know if it's an "issue". one could call the sleeps a bit hacky, but I don't know of a better way to make sure the operation has made it to the openshift api-server without them.

oh no sorry, I meant to say the gpg-op test is still failing with something that I think it's related to this change uhm

haircommander:
`Throttling: Rate exceeded\n`

/retest

haircommander:
:tada: 

ok now it's ready (assuming the aws failure is unrelated, which I think it is?)

@runcom @kikisdeliveryservice @umohnani8 @mtrmac PTAL

haircommander:
/retest 

:crossed_fingers:

runcom:
> ok now it's ready (assuming the aws failure is unrelated, which I think it is?)

🎉 yep, it looks like it's ok to merge - we'll fix any flake afterwards :trollface:

runcom:
/skip
/approve
/lgtm

we need a backport for this for 4.3 as well right? @umohnani8 @haircommander

haircommander:
> /skip
> /approve
> /lgtm
> 
> we need a backport for this for 4.3 as well right? @umohnani8 @haircommander

yep, backport here https://github.com/openshift/machine-config-operator/pull/1447

PR #1416, opened 2020-01-28
daemon: Move "Validated on-disk state" log to only when we have

cgwalters:
When the force file is present, we didn't actually validate.
Obviously, this just changes logging; I was just reading the code
as part of looking at
https://bugzilla.redhat.com/show_bug.cgi?id=1766515

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -962,13 +962,13 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		if !dn.validateOnDiskState(expectedConfig) {
 			return fmt.Errorf("unexpected on-disk state validating against %s", expectedConfig.GetName())
 		}
+		glog.Info("Validated on-disk state")
 	} else {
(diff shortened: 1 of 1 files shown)


runcom:
/approve

This should be safe to merge now but let me get back with a more concrete action on what to do here (branch isn't frozen, but we'd need a bug fix for such a change as well perhaps)

kikisdeliveryservice:
> /approve
> 
> This should be safe to merge now but let me get back with a more concrete action on what to do here (branch isn't frozen, but we'd need a bug fix for such a change as well perhaps)

gonna add a hold bc this prob needs a bz

/hold

kikisdeliveryservice:
unless you wanted to make it part of 1766515?

runcom:
(after some chats...) PR is super tiny so it can go in as it's not really requiring a BZ

/hold cancel
/lgtm

runcom:
Uhm probably need a BZ now that master has been enforced , let’s see

/refresh

kikisdeliveryservice:
ok this can finally merge, im doing it.

/hold cancel

PR #1419, opened 2020-01-29
Bug 1796148:  baremetal: Remove .template from path in dhcp-dhclient-conf.yaml

hardys:
This appears to get written out as /etc/dhcp/dhclient.conf.template
with no subsequent render-config templating to create the actual
/etc/dhcp/dhclient.conf - modifying the path results in the file
being correctly written.

### DIFF
--- a/templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml
+++ b/templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml
@@ -1,6 +1,6 @@
 filesystem: "root"
 mode: 0644
-path: "/etc/dhcp/dhclient.conf.template"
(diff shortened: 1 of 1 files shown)


runcom:
/approve
/lgtm

this is ipv6 work right?

hardys:
> this is ipv6 work right?

I found this while testing ipv6 but it looks like it never worked even on ipv4, AFAICT this should work fine in both cases though (and aligns with what's done and known-working on the masters)

runcom:
> but it looks like it never worked even on ipv4

I think we would love a BZ if this bug is present in 4.3 and below

/hold

hardys:
/retitle Bug 1796148:  baremetal: Remove .template from path in dhcp-dhclient-conf.yaml

PR #1421, opened 2020-01-29
Bug 1796147: pkg/server: serve config only to master in bootstrap server

runcom:
The new cluster etcd operator flow is:

1) start bootstrap mcs
2) start etcd on bootstrap
3) wait for bootstrapping to finish i.e. atleast one control-plane is ready and there is MCS running on cluster
4) turn down bootstrap mcs

What the above does is giving a chance to workers to grab
the ignition config from the bootstap server which now stays up longer.
However, by the time they attempt to create a CSR the kube-apiserver has
rotated that bootstrap chain of trust out which causes the workers to error out with:

Jan 29 19:55:20 ip-10-0-130-205 hyperkube[2623]: E0129 19:55:20.869251    2623 certificate_manager.go:421] Failed while requesting a signed certificate from the master: cannot create certificate signing request: Unauthorized

The above results in workers not being able to join the cluster eventually.

What this patch does is denying serving the configuration to all pools but master
within the bootstrap server, effectively delaying workers to grab the wrong config
from the wrong server. Workers will keep polling for configuration and they'll
eventually grab the correct one from the server running within the new cluster.

### DIFF
--- a/pkg/server/bootstrap_server.go
+++ b/pkg/server/bootstrap_server.go
@@ -56,7 +56,9 @@ func NewBootstrapServer(dir, kubeconfig string) (Server, error) {
 // 4. Append the machine annotations file.
 // 5. Append the KubeConfig file.
 func (bsc *bootstrapServer) GetConfig(cr poolRequest) (*igntypes.Config, error) {
-
+	if cr.machineConfigPool != "master" {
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
cc: @crawford

crawford:
It's not terribly surprising that e2e-gcp-upgrade failed since it's using the bootstrap process prior to this change.

PR #1422, opened 2020-01-29
Bug 1780396: Get etcdctl binary from the etcd image and not from the upstream releases

Prashanth684:
For non-x86 architectures there is no etcdctl binary available for the upstream releases of etcd. The suggestion from @cgwalters was to get the binary from the image(https://bugzilla.redhat.com/show_bug.cgi?id=1780396#c4)
which makes sense.

Verified by making the changes on an s390x system and executing the snapshot backup script.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -30,15 +30,13 @@ contents:
            exit 1
         fi
       else
-        GOOGLE_URL=https://storage.googleapis.com/etcd
-        DOWNLOAD_URL=${GOOGLE_URL}
-
-        echo "Downloading etcdctl binary.."
-        curl -s -L ${DOWNLOAD_URL}/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-amd64.tar.gz -o $ASSET_DIR/tmp/etcd-${ETCD_VERSION}-linux-amd64.tar.gz \
-          && tar -xzf $ASSET_DIR/tmp/etcd-${ETCD_VERSION}-linux-amd64.tar.gz -C $ASSET_DIR/shared --strip-components=1 \
-          && mv $ASSET_DIR/shared/etcdctl $ASSET_DIR/bin \
-          && rm $ASSET_DIR/shared/etcd \
-          && $ETCDCTL_BIN version
+        local etcdimg="{{.Images.etcdKey}}"
+        local etcdctr=$(podman create ${etcdimg})
+        local etcdmnt=$(podman mount "${etcdctr}")
+        cp ${etcdmnt}/bin/etcdctl $ASSET_DIR/bin
+        umount "${etcdmnt}"
+        podman rm "${etcdctr}"
+        $ETCDCTL_BIN version
       fi
     }
 


kikisdeliveryservice:
PTAL as you updated this recently

/assign @hexfusion

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -30,15 +30,13 @@ contents:
...
-        DOWNLOAD_URL=${GOOGLE_URL}
-
-        echo "Downloading etcdctl binary.."
-        curl -s -L ${DOWNLOAD_URL}/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-amd64.tar.gz -o $ASSET_DIR/tmp/etcd-${ETCD_VERSION}-linux-amd64.tar.gz \
-          && tar -xzf $ASSET_DIR/tmp/etcd-${ETCD_VERSION}-linux-amd64.tar.gz -C $ASSET_DIR/shared --strip-components=1 \
-          && mv $ASSET_DIR/shared/etcdctl $ASSET_DIR/bin \
-          && rm $ASSET_DIR/shared/etcd \
-          && $ETCDCTL_BIN version
+        local etcdimg="{{.Images.etcdKey}}"
+        local etcdctr=$(podman create ${etcdimg})
```
Prashanth684: is it a fair assumption that the etcd image will always be available and doesn't need to be pulled ?

hexfusion: Should be as this is a master node the relationship is 1:1

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -30,15 +30,13 @@ contents:
...
+        local etcdmnt=$(podman mount "${etcdctr}")
+        cp ${etcdmnt}/bin/etcdctl $ASSET_DIR/bin
```
hexfusion: you tested this? pretty sure the path is `/usr/bin/etcdctl`[1]

[1] https://github.com/openshift/etcd/blob/e30a18f29c8bf1f16507d7983dde2aa51962d6c7/Dockerfile.rhel#L15

hexfusion:
In general, I agree with the approach thanks @cgwalters

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
Prashanth684: yes i did test this - and this worked.

hexfusion: confirmed in manual testing.

hexfusion: > lrwxrwxrwx. 1 root root   7 Sep  2 12:45 bin -> usr/bin

Prashanth684: ah..yep :)

retroflexer: I tested it and it works too.

hexfusion:
/skip
disruptive tests and scaleup are broke

/lgtm
/hold

holding for @retroflexer to verify

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -30,15 +30,13 @@ contents:
...
+        umount "${etcdmnt}"
+        podman rm "${etcdctr}"
+        $ETCDCTL_BIN version
```
retroflexer: I am not sure, why it had worked in the past, but to me, `version` without dashes is giving an error.
[core@ip-10-0-136-108 ~]$ ./assets/bin/etcdctl version
No help topic for 'version'
[core@ip-10-0-136-108 ~]$ ./assets/bin/etcdctl --version
etcdctl version: 3.3.17
API version: 2

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
retroflexer: Just to be consistent, you can put quotes around ${etcdimg} too. 
```
local etcdctr=$(podman create "${etcdimg}")
```

retroflexer:
Installed the patch. And verified some of the DR scripts to be working as expected.

/hold cancel

PR #1425, opened 2020-01-31
Bug 1794824: If PlatformStatus.VSphere is nil do not template the file

jcpowermac:
When installing onto vSphere UPI MCC fails templating the in-service networking
components because there is no vSphere platformstatus.
Add check to each template.

If vSphere platform status does not exist create empty files

Tested with both vSphere IPI and UPI.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -276,7 +276,7 @@ func appendManifestsByPlatform(manifests []manifest, infra configv1.Infrastructu
 			},
 		)
 	}
-	if infra.Status.PlatformStatus.VSphere != nil {
+	if infra.Status.PlatformStatus.VSphere != nil && infra.Status.PlatformStatus.VSphere.APIServerInternalIP != "" {
 		manifests = append(manifests,
 			manifest{
 				name:     "manifests/vsphere/coredns.yaml",
--- a/templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
+++ b/templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
@@ -3,5 +3,9 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-kni.conf"
 contents:
   inline: |
+    {{ if .Infra.Status.PlatformStatus.VSphere -}}
+    {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
     [main]
     dhcp=dhclient
+    {{ end -}}
+    {{ end -}}
--- a/templates/common/vsphere/files/vsphere-coredns-corefile.yaml
+++ b/templates/common/vsphere/files/vsphere-coredns-corefile.yaml
@@ -3,6 +3,8 @@ mode: 0644
 path: "/etc/kubernetes/static-pod-resources/coredns/Corefile.tmpl"
 contents:
   inline: |
+    {{ if .Infra.Status.PlatformStatus.VSphere -}}
+    {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
     . {
         errors
         health :18080
@@ -12,3 +14,5 @@ contents:
         reload
         file /etc/coredns/node-dns-db {{ .EtcdDiscoveryDomain }}
     }
+    {{ end -}}
+    {{ end -}}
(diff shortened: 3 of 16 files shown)


runcom:
/approve

can we get some more eyes from vsphere folks?

jcpowermac:
cc: @patrickdillon 
@mtnbikenc

patrickdillon:
@jcpowermac Would it be better to populate platform status for UPI but leave VIP fields blank? https://github.com/openshift/installer/blob/master/pkg/asset/manifests/infrastructure.go#L146

jcpowermac:
> @jcpowermac Would it be better to populate platform status for UPI but leave VIP fields blank? https://github.com/openshift/installer/blob/master/pkg/asset/manifests/infrastructure.go#L146

Maybe we should do both?  In UPI we don't want the "baremetal networking" to come on line.
So instead check for empty string?

jcpowermac:
/hold
Need to make changes in the installer to set PlatformStatus and its values to empty strings.

patrickdillon:
LGTM. Already approved so I will give other reviewers a chance.

templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
```diff
@@ -3,5 +3,9 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-kni.conf"
 contents:
   inline: |
+    {{ if .Infra.Status.PlatformStatus.VSphere -}}
+    {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
```
abhinavdahiya: hmm there is no way to do AND in templates?

pkg/operator/bootstrap.go
```diff
@@ -277,24 +277,26 @@ func appendManifestsByPlatform(manifests []manifest, infra configv1.Infrastructu
...
-			manifest{
-				name:     "manifests/vsphere/keepalived.yaml",
-				filename: "vsphere/manifests/keepalived.yaml",
-			},
-			manifest{
-				name:     "manifests/vsphere/keepalived.conf.tmpl",
-				filename: "vsphere/static-pod-resources/keepalived/keepalived.conf.tmpl",
-			},
-		)
+		if infra.Status.PlatformStatus.VSphere.APIServerInternalIP != "" {
```
abhinavdahiya: why not move this to since if block ANDed ?

templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
jcpowermac: Templates does have `and` but when I tested them
```
[root@bootstrap-0 bootstrap-secrets]# crictl logs -f 517
I0205 01:41:15.357524       1 bootstrap.go:40] Version: machine-config-daemon-4.3.0-201910280117-238-g1a379174 (1a379174f8a60cbb57436769fa2b33fed30d0a49)
F0205 01:41:15.412264       1 bootstrap.go:47] error running MCC[BOOTSTRAP]: failed to create MachineConfig for role master: failed to execute template: template: /etc/mcc/templates/common/vs
phere/files/vsphere-NetworkManager-kni-conf.yaml:6:57: executing "/etc/mcc/templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml" at <.Infra.Status.PlatformStatus.VSphere.APISer
verInternalIP>: can't evaluate field APIServerInternalIP in type *v1.VSpherePlatformStatus
[root@bootstrap-0 bootstrap-secrets]# Connection to 139.178.89.200 closed by remote host
```

```
  1 filesystem: "root"
  2 mode: 0644
  3 path: "/etc/NetworkManager/conf.d/99-kni.conf"
  4 contents:
  5   inline: |                                                                                                                                                                                
  6     {{ if and .Infra.Status.PlatformStatus.VSphere .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
  7     [main]
  8     dhcp=dhclient
  9     {{ end -}}
```
And from the documentation:

> and
> 	Returns the boolean AND of its arguments by returning the
> 	first empty argument or the last argument, that is,
> 	"and x y" behaves as "if x then y else x". **All the
> 	arguments are evaluated.**

runcom:
> LGTM. Already approved so I will give other reviewers a chance.

/lgtm
/retest

PR #1428, opened 2020-02-01
WIP Bug 1797244: etcd-member.yaml: use etcd db file existence as a condition

hexfusion:
`cluster-etcd-operator` introduced a few additional init containers some of these containers utilize k8s clients to communicate with the apiserver. This becomes a chicken and egg situation during situations where kube may never exist during the life of the container. An example of this is disaster recovery.

By checking for the existence of the etcd db we can conclude if that member has been previously initialized. If it has then we do not require the k8s clients we can utilize the ENV variables populated in the discovery container based on system-level queries (ip address) and SRV lookups.

depends on https://github.com/openshift/cluster-etcd-operator/pull/73

### DIFF
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -111,7 +111,9 @@ contents:
             export ETCDCTL_API=3 ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key) \
               ETCDCTL_ENDPOINTS="$ETCD_ENDPOINTS"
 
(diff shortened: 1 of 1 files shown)


templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -100,8 +102,12 @@ contents:
             #!/bin/sh
             set -euxo pipefail
 
-            source /run/etcd/environment
+            if [ -e  "/var/lib/etcd/member/snap/db" ]; then
+              echo "etcd has previously been initialized as a member"
+              exit 0
```
hexfusion: by exiting early we avoid unbound variable for ETCD_ENDPOINTS.

hexfusion: I think in general we should move this init container logic to go, but for now this moves towards unblocking DR

hexfusion:
reworking this

PR #1429, opened 2020-02-01
Bug 1795696: templates: etcd-member: setup environment variables needed for easy etcdctl execution

retroflexer:
Closes: #1795696

One can easily update .bashrc or .bash_profile using postStart hooks. However, `oc rsh` apparently doesn't use bash, so these are not getting automatically invoked.

https://github.com/openshift/origin/issues/7514
https://github.com/openshift/origin/issues/7496 

https://github.com/openshift/origin/pull/7519

However, standard shell (Bourne shell) uses .profile. But in RHCOS, /bin/sh is symbolically linked to /usr/bin/bash. 

> If bash is invoked with the name sh, it tries to mimic the startup behavior of historical versions of sh as closely as possible, while conforming to the POSIX standard as well. [...] When invoked as an interactive shell with the name sh, bash looks for the variable ENV, expands its value if it is defined, and uses the expanded value as the name of a file to read and execute. Since a shell invoked as sh does not attempt to read and execute commands from any other startup files, the --rcfile option has no effect.

**- What I did**
Added the environment variables needed for straightforward execution of etcdctl into /root/.profile, so that `oc rsh` has the ready-made environment.
**- How to verify it**
1. Install cluster
2. oc rsh -n openshift-etcd _etcd-member-pod_
3. Run etcdctl commands such as `etcdctl member list`

Should work right off the bat.

Provide environment variables for straightfoward execution of etcdctl commands on pods.

### DIFF
--- a/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
+++ b/templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
@@ -176,12 +176,21 @@ contents:
           mountPath: /var/lib/etcd/
         - name: conf
           mountPath: /etc/etcd/
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - echo 'export ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)' >> /root/.profile
 
         env:
         - name: ETCDCTL_API
           value: "3"
         - name: ETCD_DATA_DIR
           value: "/var/lib/etcd"
+        - name: ENV
+          value: "/root/.profile"
         - name: ETCD_NAME
           valueFrom:
             fieldRef:


templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -176,12 +176,21 @@ contents:
           mountPath: /var/lib/etcd/
         - name: conf
           mountPath: /etc/etcd/
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - echo 'export ETCDCTL_API=3 ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)' >> /root/.profile
```
hexfusion: we set API version in env below. we should do one or the other. does this work when if we exec in aswell?

hexfusion:
overall looks good, thanks @retroflexer

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -176,12 +176,21 @@ contents:
...
 
         env:
         - name: ETCDCTL_API
           value: "3"
         - name: ETCD_DATA_DIR
           value: "/var/lib/etcd"
+        - name: ENV
+          value: "/root/.profile"
```
retroflexer: In RHCOS, /bin/sh is symbolically linked to sh, but it tries to mimic the behavior of historical bourne shell.

When bash is invoked as an interactive shell with the name sh, bash looks for the variable ENV, expands its value if it is defined, and uses the expanded value as the name of a file to read and execute. Since a shell invoked as sh does not attempt to read and execute commands from any other startup files (including .profile), the --rcfile option has no effect.
(from the `man` pages of bash)

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
retroflexer: Removed ETCDCTL_API. crictl exec works well too.

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
alaypatel07: @retroflexer thanks, for the crisp explanation. I wonder if it would be a good idea to create a secret/configmap for the env variables [here](https://github.com/openshift/machine-config-operator/pull/1429/files#diff-db87ea55a6cbf22e6bd99d79dc77e943R185) and then mount the [configmap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-container-environment-variables-using-configmap-data)? I think it would be lot cleaner and admins can configure the env variable independent of static pod yaml.

/cc @hexfusion

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
```diff
@@ -176,12 +176,21 @@ contents:
           mountPath: /var/lib/etcd/
         - name: conf
           mountPath: /etc/etcd/
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - echo 'export ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)' >> /root/.profile
```
cgwalters: I really don't think we should have starting any containers affect host content like this by default.

The etcd recovery tools are going to be containerized anyways, so let's just document this stuff over there?

cgwalters:
/hold
I really don't like this and think we should take another approach that doesn't involve changing root's `.profile`.

retroflexer:
> /hold
> I really don't like this and think we should take another approach that doesn't involve changing root's `.profile`.

 What other solution do you recommend, Collins?

Our customers and customer support staff is struggling with typing a mile long commands to achieve simple `member list` or `endpoint status`. 

Michal thought this should be ported to 4.2 and 4.3 as well to help mitigate the pain for our customers.

Cc: @mfojtik, @hexfusion

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
retroflexer: The etcd recovery tools are going to be containerized in 4.5+. And it assumes cluster-etcd-operator to be available, which is not available prior to 4.4. That means, our current customers of 4.2 and 4.3 (and perhaps 4.4) will not have easy execution of etcdctl.

Also .profile is current not used by anyone (that file doesn't exist at all in most pod installations). So, I do not see a reason to oppose it. If you don't like the use of `.profile`, I can put it in a file named `.etcdctl_env`.

mfojtik:
@cgwalters i'm willing to accept this for 4.4, but revisit in 4.5 where we should make the ETCD_DNS_NAME in the cert names go away and one could just set normal pod level env vars that points to ETCDCTL_CERT/etc.
We made this hard to ourself by making the file names non-deterministic, lets drop that legacy and make them deterministic (use a pod name or something).

Also this is important when debugging broken clusters and also come up recently during one escalation call.

/cc @smarterclayton

alaypatel07:
@retroflexer @hexfusion can we maybe do this via the discovery container? for instance, we set the DNS name here[1], can we set the ETCDCTL_CERT* env variables here as well. It will be in the `/run/etcd/environment` and we source that anyway. The etcd container can then always have the env variable. Sorry for the late response, it hit me now.

1. https://github.com/openshift/machine-config-operator/blob/09fe53e2e47bc6f8129376dfe389e98fc151ff48/cmd/setup-etcd-environment/run.go#L189

alaypatel07:
> we set the DNS name here[1], can we set the ETCDCTL_CERT* env variables here as well. It will be in the /run/etcd/environment and we source that anyway. The etcd container can then always have the env variable

Thanks @retroflexer for confirming over slack, this will only work if the user explicitly sources `/run/etcd/environment` after execing or rshing into the container

templates/master/00-master/_base/files/etc-kubernetes-manifests-etcd-member.yaml
cgwalters: Sorry, I mistakenly thought these hooks ran on the host.

cgwalters:
/hold cancel
per https://github.com/openshift/machine-config-operator/pull/1429#discussion_r375492457

hexfusion:
The bottom line here is we are improving user experience. This is a simple and practical approach in my opinion. thanks @retroflexer 

/lgtm

hexfusion:
cc @cgwalters @runcom

PR #1432, opened 2020-02-03
Bug 1797790: [release-4.2] prevent hitting annotation max size limit on nodes (cherry-pick)

ericavonb:
Cherry-pick of #1404 

---

- What I did
Fixes issue where an "annotation not found" error compounds via wrapping the node in the error message, growing the machineconfiguration.openshift.io/reason annotation until it hits the annotation max size limit, thereby preventing updates to that node's annotations by controllers.

    Remove node object from "annotation not found" error and replace with just node name
    Truncate errors used in machineconfiguration.openshift.io/reason node annotation to prevent any that may contain large amounts of data from maxing out annotation size.

- How to verify it

    Stop MCO and MCC controllers
    Delete the machineconfiguration.openshift.io/desiredConfig annotation on a node and wait for the MCD to sync the update
    Check the node object for the annotation"machineconfiguration.openshift.io/reason": "machineconfiguration.openshift.io/desiredConfig annotation not found in node '<name here>'"

- Description for the changelog

    Truncate errors used in machineconfiguration.openshift.io/reason node annotation to prevent hitting the max annotation size

Fixes: Bug 1797790

### DIFF
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -1,6 +1,8 @@
 package daemon
 
 import (
+	"fmt"
+
 	"github.com/golang/glog"
 	"github.com/openshift/machine-config-operator/internal"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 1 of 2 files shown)


ericavonb:
/retitle bug 1797790: prevent hitting annotation max size limit on nodes (cherry-pick)

pkg/daemon/writer.go
```diff
@@ -102,10 +104,14 @@ func (nw *clusterNodeWriter) SetWorking(client corev1client.NodeInterface, liste
...
 	glog.Errorf("Marking Unreconcilable due to: %v", err)
+	// truncatedErr caps error message at a reasonable length to limit the risk of hitting the total
+	// annotation size limit (256 kb) at any point
+	truncatedErr := fmt.Sprintf("%.2000s", err.Error())
 	annos := map[string]string{
 		constants.MachineConfigDaemonStateAnnotationKey:  constants.MachineConfigDaemonStateUnreconcilable,
-		constants.MachineConfigDaemonReasonAnnotationKey: err.Error(),
+		constants.MachineConfigDaemonReasonAnnotationKey: truncatedErr,
 	}
+	MCDState.WithLabelValues(constants.MachineConfigDaemonStateUnreconcilable, truncatedErr).SetToCurrentTime()
```
sinnykumari: This function was introduced in 4.3 release. It seems to me that this is related to telemetry which we added in 4.3, perhaps we can skip these.

runcom: yea we need to drop these calls

ericavonb: dropped. thanks for catching.

PR #1434, opened 2020-02-03
bug 1797797: fix(machineconfig.crd.yaml): add passwd schema

ericavonb:
Cherry-pick of #1234

### DIFF
--- a/manifests/machineconfig.crd.yaml
+++ b/manifests/machineconfig.crd.yaml
@@ -34,3 +34,193 @@ spec:
     singular: machineconfig
     # kind is normally the CamelCased singular type. Your resource manifests use this.
     kind: MachineConfig
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        apiVersion:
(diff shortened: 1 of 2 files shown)


manifests/machineconfig.crd.yaml
```diff
@@ -34,3 +34,202 @@ spec:
...
+                              description: "items describes unit dropin"
+                              type: object
+                              properties:
+                                contents:
+                                  description: "contents is the contents of the drop-in"
+                                  type: string
+                                name:
+                                  description: "name is the name of the drop-in. This must be suffixed with '.conf'."
+                                  type: string
+          kargs:
```
sinnykumari: I don't think we support kargs in 4.1. Not sure about fips though

runcom: both aren't supported in 4.1

ericavonb: thanks for catching that. removed.

ericavonb:
/close

per conversation with @ashcrow, we won't be backporting this to 4.1

Issue #1420, opened 2020-01-29
baremetal: ipv6 non_virtual_ip for workers is empty


hardys:
In https://github.com/openshift/machine-config-operator/commit/0c74358b7d9ca77c78bf23e3cfcc6bcf6c03dbc3#diff-9cbee0e66919c61dc7ca33846ae316ce I reworked the prepender script, but maintained the existing code to check the non_virtual_ip

```
       NON_VIRTUAL_IP=$(/usr/local/bin/non_virtual_ip \
            "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
            "{{.Infra.Status.PlatformStatus.BareMetal.NodeDNSIP}}" \
            "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
```

When running this in an ipv6 environment we see this:

```
[root@worker-0 dispatcher.d]# foo=$(/usr/local/bin/non_virtual_ip "fd2e:6f44:5dd8:c956::5" "fd2e:6f44:5dd8:c956:0:0:0:2" "fd2e:6f44:5dd8:c956::4")
Is fd2e:6f44:5dd8:c956::5 between 172.22.0.0 and 172.22.0.255
Is fd2e:6f44:5dd8:c956::5 between fe80:: and fe80::ffff:ffff:ffff:ffff
Is fd2e:6f44:5dd8:c956::5 between fd2e:6f44:5dd8:c956::17 and fd2e:6f44:5dd8:c956::17
Is fd2e:6f44:5dd8:c956::5 between fe80:: and fe80::ffff:ffff:ffff:ffff
[root@worker-0 dispatcher.d]# echo $foo

```

The APIServerInternalIP references the API VIP, the IngressIP is also a VIP, and the NodeDNSIP resolves to the DNS VIP - so these are *all* VIPs.

I think the IP we want (on the workers, which don't point to the DNS VIP) is `fd2e:6f44:5dd8:c956::17` which is the ULA for the external network on the worker node.

```
3: enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 00:0b:bb:d1:05:56 brd ff:ff:ff:ff:ff:ff
    inet6 fd2e:6f44:5dd8:c956::17/128 scope global dynamic noprefixroute 
       valid_lft 3026sec preferred_lft 3026sec
[...]

Issue #1430, opened 2020-02-03
[FCOS] openstack: workers don't join the cluster due to python scripts


vrutkovs:
Openstack / oVirt are using a python script to find IPs: https://github.com/openshift/machine-config-operator/blob/fcos/templates/worker/00-worker/openstack/files/openstack-non-virtual-ip.yaml#L6

FCOS doesn't provide python (and most likely never will), so it needs to be rewritten

cgwalters:
Today, we also ship `/usr/libexec/machine-config-daemon` in the host by default which comes from this repo.  So, a useful pattern could be to add arbitrary subcommands to that binary.

Note though when submitting PRs to this repo, the binary in CI will *not* be updated - that currently only happens after code lands in master and goes through the main RHCOS build process.

cgwalters:
> Today, we also ship /usr/libexec/machine-config-daemon in the host by default which comes from this repo. So, a useful pattern could be to add arbitrary subcommands to that binary.

And to elaborate this is used today by e.g. [this unit](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/machine-config-daemon-firstboot.service).

> Note though when submitting PRs to this repo, the binary in CI will not be updated - that currently only happens after code lands in master and goes through the main RHCOS build process.

Hence today unfortunately one would have to "ratchet" a change here by landing the Go code in master, wait for it to propagate into an updated `machine-os-content`, then delete the other code and switch the calling code to use the new code.

vrutkovs:
The PR can be tested on `fcos` branch - since FCOS doesn't ship MCD (and RHCOS shouldn't too really) it fetches the binary from the release payload and pivots early - thus MCD is always consistent across all nodes

LorbusChris:
New approach:
[Add a baremetal-runtimecfg-ubi8 image](https://github.com/openshift/release/pull/7537) and use that to run the python scripts in a containerized way, exclusively in OKD.

PR #1436, opened 2020-02-03
Bug 1798146: [baremetal] Ipv6 non virtual ip fix

celebdor:
Closes: #1420 

**- What I did**
Make non_virtual_ip IPv6 compatible by checking the IPv6 route advertisements to complement the check for whether an address in in the VIPs subnet.

**- How to verify it**
Deploy OpenShift Baremetal on IPv6 with addresses provided by a DHCPv6 with Route Advertisement enabled and check that /etc/resolv.conf gets the right non VIP IPv6 address.

### DIFF
--- a/templates/worker/00-worker/baremetal/files/baremetal-non-virtual-ip.yaml
+++ b/templates/worker/00-worker/baremetal/files/baremetal-non-virtual-ip.yaml
@@ -9,7 +9,7 @@ contents:
     import struct
     import subprocess
     import sys
-    from typing import Callable, Iterator, List, Optional, Tuple, Type, TypeVar
+    from typing import Callable, Iterable, Iterator, List, Optional, Tuple, Type, TypeVar
 
 
     class SubnetNotFoundException(Exception):
@@ -21,12 +21,13 @@ contents:
     TA = TypeVar('TA', bound='Address')
 
     class Address:
-        def __init__(self, cidr: str, name: str, family: str, index: int = -1, scope: str = '') -> None:
+        def __init__(self, cidr: str, name: str, family: str, index: int = -1, scope: str = '', deprecated: bool = False) -> None:
             self.index = index
             self.name = name
             self.family = family
             self.cidr = cidr
             self.scope = scope
+            self.deprecated = deprecated
 
         @classmethod
         def from_line(cls: Type[TA], line: str) -> TA:
@@ -35,7 +36,8 @@ contents:
                        name=items[1],
                        family=items[2],
                        cidr=items[3],
-                       scope=items[5])
+                       scope=items[5],
+                       deprecated='deprecated' in items)
 
         def __str__(self) -> str:
             return f'{self.__class__.__name__}({self.cidr}, dev={self.name})'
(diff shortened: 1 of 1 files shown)


hardys:
@russellb we discussed this yesterday in the context of a fix for kublet choosing the wrong IP - also note https://github.com/openshift/machine-config-operator/issues/1430 which I think this means rewriting in golang, perhaps we can create some common library which is consumed in both cases?

hardys:
This seemed to work on first boot of the worker, but then on a subsequent lease renewal I'm seeing:

```[root@worker-0 dispatcher.d]# /usr/local/bin/non_virtual_ip fd2e:6f44:5dd8:c956::5 fd2e:6f44:5dd8:c956:0:0:0:2 fd2e:6f44:5dd8:c956::4
Traceback (most recent call last):
  File "/usr/local/bin/non_virtual_ip", line 192, in <module>
    main()
  File "/usr/local/bin/non_virtual_ip", line 178, in main
    iface_cidrs = list(interface_cidrs((non_host_scope, non_deprecated)))
  File "/usr/local/bin/non_virtual_ip", line 143, in interface_cidrs
    for route in (V6Route.from_line(rline) for rline in route_out.splitlines()):
  File "/usr/local/bin/non_virtual_ip", line 143, in <genexpr>
    for route in (V6Route.from_line(rline) for rline in route_out.splitlines()):
  File "/usr/local/bin/non_virtual_ip", line 61, in from_line
    return cls(**attrs)
TypeError: __init__() missing 1 required positional argument: 'proto'

```

templates/worker/00-worker/baremetal/files/baremetal-non-virtual-ip.yaml
```diff
@@ -4,29 +4,76 @@ path: "/usr/local/bin/non_virtual_ip"
...
+            self.pref: str = pref
+
+        @classmethod
+        def from_line(cls: Type[TR], line: str) -> TR:
+            items = line.split()
+            dest = items[0]
+            if dest == 'default':
+                dest = '::/0'
+            attrs = dict(itertools.zip_longest(*[iter(items[1:])]*2, fillvalue=None))
+            attrs['destination'] = dest
```
hardys: proto isn't always set, so I think we need `attrs.setdefault('proto', '')` or similar here, to avoid a possible traceback with a line like `fd01::/48 via fd01:0:0:4::1 dev k8s-worker-0.os metric 1024 pref medium`

celebdor: Thanks @hardys. Will update

hardys:
I tested this (on ipv6 only so far) and can confirm the latest revision works - this lgtm pending feedback from someone that it all still works OK with ipv4

runcom:
can you link this PR to https://bugzilla.redhat.com/show_bug.cgi?id=1797647 as well?

hardys:
/retitle Bug 1798146: [baremetal] Ipv6 non virtual ip fix

hardys:
> can you link this PR to https://bugzilla.redhat.com/show_bug.cgi?id=1797647 as well?

We now have the following bz's

4.4: https://bugzilla.redhat.com/show_bug.cgi?id=1798146
4.3.z: https://bugzilla.redhat.com/show_bug.cgi?id=1797647

They were initially cloned the wrong way round but we reversed the depends on which hopefully the bot will accept.

celebdor:
@yboaron tested the previous iteration in IPv4.

hardys:
I also tested this in ipv4 and can confirm it works so lgtm

PR #1442, opened 2020-02-05
[fcos] GCP: add a script and a service which ensure GCP routes are set correctly

vrutkovs:
This would ensure masters / workers would report as healthy to GCP LBs

xref https://github.com/openshift/okd/issues/38

### DIFF
--- a/templates/common/gcp/units/gcp-routes-service.yaml
+++ b/templates/common/gcp/units/gcp-routes-service.yaml
@@ -1,0 +1,19 @@
+name: "gcp-routes.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Update GCP routes for forwarded IPs.
+  ConditionKernelCommandLine=|ignition.platform.id=gce
+  ConditionKernelCommandLine=|ignition.platform.id=gcp
+  After=network.target
+
+  [Service]
+  Type=simple
+  ExecStart=/bin/bash /usr/local/bin/gcp-routes.sh start
+  ExecStopPost=/bin/bash /usr/local/bin/gcp-routes.sh cleanup
+  User=root
+  RestartSec=30
+  Restart=always
+
+  [Install]
+  WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


cgwalters:
/approve
Thanks for doing this!

To clarify though I think a next step is shipping this as part of the MCD binary (or at least `/usr/libexec/gcp-routes.sh`) i.e. not part of Ignition (since it's not templated, it's static per node) but lifting it out of of the "internal RHCOS dark matter" here and having it naturally tested with the existing controller is  a great first step.

templates/common/gcp/units/gcp-routes-service.yaml
```diff
@@ -0,0 +1,19 @@
+name: "gcp-routes.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Update GCP routes for forwarded IPs.
+  ConditionKernelCommandLine=|ignition.platform.id=gce
+  ConditionKernelCommandLine=|ignition.platform.id=gcp
+  After=network.target
```
vrutkovs: Is it recommended to use `network.target` or `network-online.target` in Fedora?

vrutkovs:
Not sure if master needs it - currently this script is provided by RHCOS itself.

Lets see if this works on OKD and file an enhancement to have this moved to installer / MCO?

cgwalters:
While I am an official member of the enhancement police :policeman: I think this change doesn't necessarily require one...it's not something an end user would care about, just shuffing around the implementation details of MCO/RHCOS.  (But also, no opposition to making one!)

templates/common/gcp/units/gcp-routes-service.yaml
cgwalters: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/ covers this...I think `network-online.target` is more right but the script retries anyways.

vrutkovs: since we already have `network.target` in vsphere UPI scripts I'll keep `network.target` - and replace these later

LorbusChris:
enhancement police :grin: 

I'm plus one on shuffling this around, ie. landing it in master, and then subsequently removing it from the RHCOS builds in order to align FCOS and RHCOS builds, and reducing the delta we have here between fcos and master branches.

I don't think we need a formal enhancement proposal for this, as it doesn't enhance anything, it just moves an existing implementation around.

An optimization would be to move this into the MCD directly.

PR #1444, opened 2020-02-06
Bug 1798788: Set Kubelet node IP to non-vip

celebdor:
This change prevents:
* Kubelet choosing as node IP an address from a non control plane subnet
  (like the provisioning network)
* Kubelet choosing a deprecated IPv6 address as its node IP that in
  several platform can be a VIP

**- How to verify it**

Deploy on an IPv6 environment on a platform that uses keepalived managed Virtual IPs for its API, DNS and Ingress and see that the node IP is none of those, as those are created by keepalived as deprecated addresses.


celebdor:
I propose the new non_virtual_ip for all of the platforms.
/cc @jcpowermac @mandre @rgolangh @hardys @karmab

celebdor:
Only ran the scripts in the PR in isolation still, I have not deployed with a new MCO yet (it's 2:36am already).

russellb:
This didn't work for me in my first run of this with a backport to 4.3.  The new `kubelet-nodeip-configuration` service never ran.

```
[core@master-1 ~]$ systemctl status kubelet-nodeip-configuration
● kubelet-nodeip-configuration.service - Writes an Environment file for Kubelet to choose a valide node IP address
   Loaded: loaded (/etc/systemd/system/kubelet-nodeip-configuration.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
```

celebdor:
> This didn't work for me in my first run of this with a backport to 4.3. The new `kubelet-nodeip-configuration` service never ran.
> 
> ```
> [core@master-1 ~]$ systemctl status kubelet-nodeip-configuration
> ● kubelet-nodeip-configuration.service - Writes an Environment file for Kubelet to choose a valide node IP address
>    Loaded: loaded (/etc/systemd/system/kubelet-nodeip-configuration.service; disabled; vendor preset: disabled)
>    Active: inactive (dead)
> ```

For some reason I ignore I had the unit template marked as `enabled: false`

templates/worker/00-worker/vsphere/files/vsphere-non-virtual-ip.yaml
```diff
@@ -1,142 +0,0 @@
-filesystem: "root"
-mode: 0755
-path: "/usr/local/bin/non_virtual_ip"
-contents:
-  inline: |
-    {{ if .Infra.Status.PlatformStatus.VSphere -}}
```
jcpowermac: @celebdor this _probably_ won't be a problem but for UPI in vSphere we don't want any "baremetal networking" services coming on line.

russellb:
It looks like you made this change more broad than needed.  I think especially since we need to backport the end result to 4.3, we should limit the changes to the bare metal IPI platform only for now.  It can be expanded later if IPv6 support is enabled elsewhere.

celebdor:
> It looks like you made this change more broad than needed. I think especially since we need to backport the end result to 4.3, we should limit the changes to the bare metal IPI platform only for now. It can be expanded later if IPv6 support is enabled elsewhere.

Agreed. Will post a more targetted version ASAP. Other platforms can copy from it then.

celebdor:
Dropped other platforms. I recommend other platforms that use non_virtual_ip update to the new one though.

templates/worker/00-worker/vsphere/files/vsphere-non-virtual-ip.yaml
celebdor: I removed the other platforms from the change. Still, if you decide to adopt this change for IPv6, I think making the `units` optional but leave the scripts, should be reasonable.

hardys:
I tested this and am seeing errors starting the MCD on the bootstrap VM:

```
I0207 12:12:01.137710       1 bootstrap.go:40] Version: machine-config-daemon-4.3.0-201910280117-152-gfe3b9677 (fe3b9677ee4453533d9357af49b1101bec2bf50f)
F0207 12:12:01.189179       1 bootstrap.go:47] error running MCC[BOOTSTRAP]: failed to create MachineConfig for role master: failed to execute template: template: /etc/mcc/templates/common/baremetal/units/kubelet-nodeip.service:15:57: executing "/etc/mcc/templates/common/baremetal/units/kubelet-nodeip.service" at <.Infra.Status.Platform.BareMetal.APIServerInternalIP>: can't evaluate field BareMetal in type v1.PlatformType
```

russellb:
This is working for me, at least in our 4.3-ipv6 fork

/lgtm

russellb:
@celebdor can you update this to include your crio address fix, too?

kikisdeliveryservice:
@hardys have you retested this?

@celebdor pending his retesting is this good to go? lmk!

russellb:
It looks like this one depends on #1445.  We could either place #1445 with a new PR that includes both changes, or wait for #1445 to go in first

@celebdor

russellb:
opened https://github.com/openshift/machine-config-operator/pull/1466

PR #1446, opened 2020-02-06
Bug 1790823: [baremetal] Verify that MDNS doesn't advertise localhost 

yboaron:
MDNS publisher static pod should start advertising the node
only after node's hostname was set by DHCP.

The idea is to add an init container that will monitor host's hostname
until hostname != localhost.

The issue was how host's hostname should be retrieved by the init container.

Since the MDNS static pod doesn't use host's uts namespace, reading 'hostname'
is not an option.
Reading hostname using dbus didn't work due to security constraints(SElinux).

The suggested solution composed of:
 1. NM dispatcher script will monitor & store hostname in a file
 2. Init container will mount this file, and read the value

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
+++ b/templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
@@ -1,0 +1,22 @@
+filesystem: "root"
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/40-mdns-hostname"
+contents:
+  inline: |
+    #!/bin/bash
+    STATUS=$2
+    case "$STATUS" in
+        up|down|dhcp4-change|dhcp6-change|hostname)
+        logger -s "NM mdns-hostname triggered by ${2}."
+        set +e
+        t_hostname=$(hostname)
+        if [ -z "${t_hostname}" ]; then
+           t_hostname="localhost"
+        fi
+        mkdir -p /etc/mdns
+        echo "${t_hostname}">/etc/mdns/hostname
+        logger -s "Hostname changed: ${t_hostname}"
+        ;;
+        *)
+        ;;
+    esac
--- a/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
+++ b/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
@@ -34,11 +34,23 @@ contents:
         - "-c"
         - |
           #/bin/bash
-          while [ "$(hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
+          function get_hostname()
+          {
+            if [[ -s /etc/mdns/hostname ]]; then
+              cat /etc/mdns/hostname
+            else
+              # if hostname wasn't updated by NM script, read hostname
+              hostname
+            fi
+          }
+          while [ "$(get_hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
           do
             echo "hostname is still ${DEFAULT_LOCAL_HOSTNAME}"
-            sleep 10
+            sleep 1
           done
+        volumeMounts:
+        - name: conf-dir
+          mountPath: "/etc/mdns"
       - name: render-config
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:


yboaron:
I still didn't check e2e deployment (with customized MCO image) with this fix

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
```diff
@@ -0,0 +1,21 @@
...
+path: "/etc/NetworkManager/dispatcher.d/40-mdns-hostname"
+contents:
+  inline: |
+    #!/bin/bash
+    STATUS=$2
+    case "$STATUS" in
+        dhcp4-change|dhcp6-change|hostname)
+        logger -s "NM mdns-hostname triggered by ${2}."
+        set +e
+        t_hostname=$(hostname)
```
celebdor: shouldn't it be the FQDN?

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
```diff
@@ -0,0 +1,21 @@
...
+        if [ -z "${t_hostname}" ]; then
+           t_hostname="localhost"
```
celebdor: `localhost.localdomain` ?

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
```diff
@@ -0,0 +1,21 @@
...
+        fi
+        echo "${t_hostname}">/etc/mdns/hostname.txt
```
celebdor: I think we can safely drop the `.txt` extension

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
```diff
@@ -0,0 +1,21 @@
+filesystem: "root"
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/40-mdns-hostname"
+contents:
+  inline: |
+    #!/bin/bash
+    STATUS=$2
+    case "$STATUS" in
+        dhcp4-change|dhcp6-change|hostname)
```
celebdor: Let's do it in `up|down` as well for good measure.

templates/common/baremetal/files/baremetal-mdns-publisher.yaml
```diff
@@ -34,11 +34,23 @@ contents:
...
         - |
           #/bin/bash
-          while [ "$(hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
+          function get_hostname()
+          {
+            if [[ -s /etc/mdns/hostname.txt ]]; then
+              echo $(</etc/mdns/hostname.txt)
+            else
+              # if hostname wasn't updated by NM script, read hostname
+              hostname
```
celebdor: What's the purpose of reading the cloned uts hostname?

templates/common/baremetal/files/baremetal-mdns-publisher.yaml
```diff
@@ -34,11 +34,23 @@ contents:
         - "-c"
         - |
           #/bin/bash
-          while [ "$(hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
+          function get_hostname()
+          {
+            if [[ -s /etc/mdns/hostname.txt ]]; then
+              echo $(</etc/mdns/hostname.txt)
```
celebdor: why not just `cat` ?

templates/common/baremetal/files/baremetal-mdns-publisher.yaml
```diff
@@ -34,11 +34,23 @@ contents:
...
+            else
+              # if hostname wasn't updated by NM script, read hostname
+              hostname
+            fi
+          }
+          while [ "$(get_hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
           do
             echo "hostname is still ${DEFAULT_LOCAL_HOSTNAME}"
             sleep 10
```
celebdor: I don't think it will kill us to try every second. It is literally just `cat`ting a file.

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
yboaron: Running 'hostname' from shell returns just the hostname without a domain name(e.g: localhost), while GOLANG os.Hostname() returns host+domain-name.
The baremetal-runtimecfg relevant code (utils.ShortHostname [1]) should work OK for both cases.
[1] https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/utils/utils.go#L22

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
yboaron: OK

templates/common/baremetal/files/baremetal-mdns-publisher.yaml
yboaron: If verify-hostname init container start to run before NM dispatcher && If hostname in cloned uts namespace != localhost (meaning that verify-hostname start to run after hostname set by DHCP) - 
there is no reason to wait for NM script, right?

templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
yboaron: Yep, I need to update also baremetal-runtimecfg to consider mdns hostname file

yboaron:
This PR is replaced by https://github.com/openshift/machine-config-operator/pull/1455

PR #1447, opened 2020-02-06
Bug 1794495: [release-4.3] fix ctrcfg and add e2e test

haircommander:
### What I did
ctrcfgs have been broken since [this]() PR, possibly before. There are a couple of reasons behind this:
- https://github.com/openshift/machine-config-operator/pull/1216 and ignition apparently don't play well together. When there are empty values (i.e. commented out ones) in the template, they are rendered as the empty value for whatever type it is (0, "", []). In a lot of cases, CRI-O can't distinguish between a value that is legitimately empty (stream_address is specifically set as so in openshift), and one that is empty due to an error. Thus, we need to populate the template until a better approach to the ContainerRuntimeConfig controller is reached. (this is done [here](https://github.com/openshift/machine-config-operator/pull/1447/commits/8afb65a90b2f6e929adf9f01096fa40e8122f259))
- Next, CRI-O 1.16 dropped support for some config values. The config that was being rendered was from *1.9*. This caused some values to be omitted (`version_file_location`) after the controller did an update. [This](https://github.com/openshift/machine-config-operator/pull/1447/commits/46e3711922bedf9a04cf89bc0e347ab8b890864d) commit updates from vendoring cri-o 1.9 to 1.16 to make sure we're rendering a config file that is compatible with the version of CRI-O. Until we move away from specifying the full template as aformentioned, we'll want to bump CRI-O each openshift version.
- After that was fixed, it was found that comparing quantity.Resource{} against the values sent to the controller caused spurious failures, as an empty Resource structure was sent, yet the comparison between the empty struct didn't succeed. Now, we compare `Resource.Value() == 0`, which is a more deterministic way of finding whether the resource was specified (and prevent a fatally empty value from showing up in the resulting crio.conf). [This](https://github.com/openshift/machine-config-operator/pull/1447/commits/3981914a3b98886c14e124e0e7531d73e827c729) and [this](https://github.com/openshift/machine-config-operator/pull/1447/commits/0d2dc367481048cb0ec7fece4426f176089a43fb) commit make these changes.
- [This](https://github.com/openshift/machine-config-operator/pull/1447/commits/d8a6fa62d4a128e68f7ca78521282e192ccb6b62) was cherry-picked to this branch. Without this patch, a ctrcfg would be rendered, but then fail to be discovered because the controller wasn't looking for the right pool.
- Finally, and perhaps most importantly, an e2e test was added. Now, we will test changes to the MCO against the workflow of a ctrcfg: create, watch and check for an update, remove, watch and check for a reversion.

### Risks
The risk of this PR is relatively low. The biggest worry I have is around cherry-picking [this](https://github.com/openshift/machine-config-operator/pull/1447/commits/d8a6fa62d4a128e68f7ca78521282e192ccb6b62).
However, the existence of an e2e test, and the clear states of "before these changes, the e2e test failed" and "after these changes, the e2e test passes" makes me believe there is no risk to merging this PR mid-z-stream.

### How to verify it
go test -v -run="TestContainerRuntimeConfigPidsLimit" ./test/e2e/

### Description for the changelog
added simple e2e test for ContainerRuntimeConfigs
revendor cri-o to have an updated config structure as well as fix problems with template rendering

cherry-pick of: https://github.com/openshift/machine-config-operator/pull/1414
also includes: https://github.com/openshift/machine-config-operator/pull/1447/commits/d8a6fa62d4a128e68f7ca78521282e192ccb6b62


runcom:
@haircommander there seems to be some issue in op here instead :(

haircommander:
yeah the tests seem legit :confused:

runcom:
> yeah the tests seem legit 😕

maybe something is different between 4.4/master and 4.3 🤔

haircommander:
> > yeah the tests seem legit confused
> 
> maybe something is different between 4.4/master and 4.3 thinking

yeah but it's a pretty confusing error. I get
```
Status:
  Conditions:
    Last Transition Time:  2020-02-06T21:19:35Z
    Message:               could not generate origin ContainerRuntime Configs: could not generate origin ContainerRuntime Configs: %v
    Status:                False
  Observed Generation:     1
```
after deploying a ctrcfg, which tells me nothing about what went wrong

haircommander:
`level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"`

/retest

haircommander:
:sunglasses:

haircommander:
(lol I only just saw you just did that 10 minutes ago @umohnani8)

umohnani8:
@runcom so until https://bugzilla.redhat.com/show_bug.cgi?id=1794493 is not verified (the 4.4 bz), we can't move ahead with this PR?

kikisdeliveryservice:
> @runcom so until https://bugzilla.redhat.com/show_bug.cgi?id=1794493 is not verified (the 4.4 bz), we can't move ahead with this PR?

correct @umohnani8

haircommander:
@eparis @crawford @mfojtik PTAL

cgwalters:
Overall...
/approve
from me.  The thing that makes me most nervous here is that the relationship between the MCO and cri-o version is inherently fragile.  Particularly as we're doing things like cherry-picking across streams.

What we really want again I think is to have `/etc/crio.conf.d/mco.conf` or so, so that the MCO only generates a config for values it changes, and otherwise the defaults live in the crio *code*.

If the possible fallout from this was only confined to clusters with an actually customized `containerruntimeconfig`, that'd be one thing but this will roll out a big `crio.conf` change to everyone.
The changes there look OK, I didn't exhaustively review it but did read through a good bit of it.

kikisdeliveryservice:
> Overall...
> /approve
> from me. The thing that makes me most nervous here is that the relationship between the MCO and cri-o version is inherently fragile. Particularly as we're doing things like cherry-picking across streams.
> 
> What we really want again I think is to have `/etc/crio.conf.d/mco.conf` or so, so that the MCO only generates a config for values it changes, and otherwise the defaults live in the crio _code_.
> 
> If the possible fallout from this was only confined to clusters with an actually customized `containerruntimeconfig`, that'd be one thing but this will roll out a big `crio.conf` change to everyone.
> The changes there look OK, I didn't exhaustively review it but did read through a good bit of it.

@cgwalters I can move this to a new issue/jira to be done as future work you'd like?

haircommander:
> What we really want again I think is to have `/etc/crio.conf.d/mco.conf` or so, so that the MCO only generates a config for values it changes, and otherwise the defaults live in the crio _code_.
> 

we submitted an epic for 4.5 to do just that :smile:

PR #1449, opened 2020-02-06
GCP: Add a script and a service which ensure GCP routes are set correctly 

LorbusChris:
**- What I did**
This PR adds a script and service unit to the GCP templates that ensures GCP routes are set corrently.
These files are already baked into RHCOS by the build system, but we'd like to move them out of there and into installer (for bootstrap nodes) and MCO (for masters/workers).

The analogous installer PR is https://github.com/openshift/installer/pull/3073

**- How to verify it**
CI

GCP: Add a script and a service which ensure GCP routes are set correctly

### DIFF
--- a/templates/common/gcp/files/gcp-routes-sh.yaml
+++ b/templates/common/gcp/files/gcp-routes-sh.yaml
@@ -12,45 +12,47 @@ contents:
 
     get_ifname() {
       sysfs_path="/sys/class/net"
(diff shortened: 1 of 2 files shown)


LorbusChris:
Targeted at 4.5

LorbusChris:
This will probably be moved into afterburn
/close

PR #1450, opened 2020-02-07
Bug 1800319: kubelet: add kube reservation for kubernetes components

rphillips:
**- What I did**
bumps `systemReserved`. Fixes an issue where the kubelet and node could become starved causing the node to hang.

[upstream doc](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#kube-reserved)
[BZ1800319](https://bugzilla.redhat.com/show_bug.cgi?id=1800319)

**- How to verify it**
The BZ has an attached test case.

```NONE```

/cc @smarterclayton @sjenning @mrunalp

### DIFF
--- a/templates/master/01-master-kubelet/_base/files/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/files/kubelet.yaml
@@ -22,8 +22,9 @@ contents:
     serializeImagePulls: false
     staticPodPath: /etc/kubernetes/manifests
     systemReserved:
-      cpu: 500m
-      memory: 500Mi
+      cpu: 768m
+      memory: 1Gi
+      ephemeral-storage: 1Gi
     featureGates:
       IPv6DualStack: true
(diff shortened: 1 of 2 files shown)


rphillips:
ready for review

kikisdeliveryservice:
seems sane leaving LGTM to @smarterclayton @sjenning @mrunalp

/approve

cherylaf:
Hi, @rphillips -- the symptoms of this bug we're seeing in OpenShift on Z are very different from the symptoms on x86.  I had understood from @jaypoulz that a CI system for Z would be available, and since this bug shows up very differently on the two platforms, perhaps it would be prudent to have some CI tests be performed on that.  I'm also pursuing access to a 4.x install on the Power architecture to see what the symptoms of memory stress are there.  Thanks!

rphillips:
768m cpu reservations looks like a requirement. The node dies with a 500m reservation.

PR #1455, opened 2020-02-11
Bug 1790823: [baremetal] Verify that MDNS doesn't advertise localhost

yboaron:
MDNS publisher static pod should start advertising the node
only after node's hostname was set by DHCP.

The idea is to add an init container that will monitor host's hostname
until hostname != localhost.

The issue was how host's hostname should be retrieved by the init container.

Since the MDNS static pod doesn't use host's uts namespace, reading 'hostname'
is not an option.
Reading hostname using dbus didn't work due to security constraints(SElinux).

The suggested solution composed of:
 1. NM dispatcher script will monitor & store hostname in a file
 2. Init container will mount this file, and read the value

This PR is instead of https://github.com/openshift/machine-config-operator/pull/1446

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
+++ b/templates/common/baremetal/files/NetworkManager-mdns-hostname.yaml
@@ -1,0 +1,22 @@
+filesystem: "root"
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/40-mdns-hostname"
+contents:
+  inline: |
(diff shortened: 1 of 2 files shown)


yboaron:
This PR replaced https://github.com/openshift/machine-config-operator/pull/1446

celebdor:
We need to first merge https://github.com/openshift/baremetal-runtimecfg/pull/44

PR #1456, opened 2020-02-11
Bug 1801662: baremetal: all resolvconf editing to NM dispatcher

celebdor:
Before this patch, the search domain modification was dependant on
dhclient. Unfortunately, the interaction of both components sometimes
fails. This patch makes sure that we preserve what comes from the dhcp
client (whichever may be) but we set what we need in VIP managing
environments.


jparrill:
BZ Related: https://bugzilla.redhat.com/show_bug.cgi?id=1801638

kikisdeliveryservice:
this needs a BZ to merge (ie Bz in title etc...)

PR #1458, opened 2020-02-11
[release-4.3] Bug 1801824: kubelet: add more system reservation to protect node

rphillips:
Cherry pick of: https://github.com/openshift/machine-config-operator/pull/1450

Kubelet and Crio are running at around 250-500 MB each (on default installs). This PR bumps the limit to 1 GB to allow for a bit of headroom to preserve some of the Kernel cache as well. If we don't bump the limit then memory pressure on the node could flush some of the kernel cache resulting in the kernel trying to re-read the cache. This flood of IOPS can be throttled  by the cloud providers which results in a kernel pause.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/01-master-kubelet/_base/files/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/files/kubelet.yaml
@@ -25,7 +25,8 @@ contents:
     systemCgroups: /system.slice
     systemReserved:
       cpu: 500m
-      memory: 500Mi
+      memory: 1Gi
+      ephemeral-storage: 1Gi
     featureGates:
       RotateKubeletServerCertificate: true
       NodeDisruptionExclusion: true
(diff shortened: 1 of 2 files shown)


eparis:
/hold
I do not yet understand how this fixes a problem correctly. Nor do I think stealing 500M x EveryServer of schedulable ram is a nice thing to do in a z-stream.

eparis:
The first comment needs to explain why we are doing this. How it fixes what it fixes etc. If I look in a git log I need to be convinced this is the right thing to have done.

rphillips:
@eparis updated

sdodson:
After a lengthy Slack conversation @smarterclayton @jim-minter and @rphillips have made a strong case for this change even if the full details of the telemetry examination didn't make it into the commit message. In the future we should make sure that sort of analysis which justifies clawing back 500MiB of memory is safe to do.

cherry-pick-approved

kikisdeliveryservice:
```
level=error
level=error msg="  on ../tmp/openshift-install-981833611/iam/main.tf line 6, in resource \"google_project_iam_member\" \"worker-compute-viewer\":"
level=error msg="   6: resource \"google_project_iam_member\" \"worker-compute-viewer\" {" 
```
/test e2e-gcp-upgrade

PR #1461, opened 2020-02-11
Bug 1804944: Point etcd SRV record at master hostnames

cybertron:
Our current method of mapping master hostnames to etcd-[0,1,2] records introduces undesirable restrictions on the names of master nodes. Now that cluster-etcd-operator allows us to bootstrap the etcd cluster using the SRV record, we can just point it at the masters directly and drop the extra etcd-specific record. This should eliminate the hostname requirements entirely.

This will need to wait for https://github.com/openshift/coredns-mdns/pull/52 to merge since that is forcing use of the etcd record.

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-mdns-config.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-mdns-config.yaml
@@ -8,7 +8,7 @@ contents:
 
     service {
         name = "{{`{{ .Cluster.Name }}`}} Etcd"
-        host_name = "{{`{{ .EtcdShortHostname }}`}}.local."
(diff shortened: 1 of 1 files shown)


cybertron:
/retest

The coredns-mdns change has merged so this should be good to go.

cybertron:
I was wrong previously, we actually needed to update the vendoring in coredns too. That's now done: https://github.com/openshift/coredns/pull/25

Which should mean this is good to go for real this time.

runcom:
/approve

@celebdor can you give this a look for merging?

runcom:
This is also gonna need a 4.4 cherry-pick

/cherrypick release-4.4

PR #1462, opened 2020-02-11
[WIP] controller: error if node with custom role is missing worker label

kikisdeliveryservice:
WIP: first pass, need to test

TO-DO: I think the unit test helper is creating the custom pool incorrectly (in MCSelector field)... will fix...

-------

When a user does not setup their custom pools correctly and do not label nodes as both worker and custompool, the errors are confusing, for ex:
`kubelet_config_controller.go:303] Error syncing kubeletconfig cluster: GenerateMachineConfigsforRole failed with error failed to read dir "/etc/mcc/templates/infra": open /etc/mcc/templates/infra: no such file or directory`

All nodes using custom roles must belong to both the custom & worker pools as per: https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md  so add an explicit error so user can fix...

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -556,10 +556,13 @@ func (ctrl *Controller) getPoolsForNode(node *corev1.Node) ([]*mcfgv1.MachineCon
 		}
 		// One custom role, let's use its pool
 		pls := []*mcfgv1.MachineConfigPool{custom[0]}
+
 		if worker != nil {
 			pls = append(pls, worker)
+			return pls, nil
 		}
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
@runcom PTAL. this still doesn't allow enqueuing a nil worker MCP (which you added before), but i think makes sure that a worker mcp does exist if we have a custom pool?

kikisdeliveryservice:
Custom roles no longer require worker labels.. see:
https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md#creating-a-custom-pool

PR #1463, opened 2020-02-12
Bug 1797687: docs: Document that MCO doesn't apply custom pool on master node

sinnykumari:
This should help users to not try adding custom pool on master
nodes and wonder why it is not getting applied.

### DIFF
--- a/docs/custom-pools.md
+++ b/docs/custom-pools.md
@@ -4,6 +4,8 @@ Custom pools are pools that inherit from the `worker` pool. They use any Machine
 but add the ability to deploy changes only targeted at the custom pool.
 Since a custom pool inherits from the `worker` pool, any change to the `worker` pool is going to roll out to the custom pool as well (like an OS update during an upgrade).
 
+**Note:** Custom pools on master nodes are not supported by MCO. For more detail, see [Custom pool on master node](#Custom-pool-on-master-node)
+
 ## Creating a custom pool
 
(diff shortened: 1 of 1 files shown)


sinnykumari:
> People tend not to read this far down. So I would add an explicit sentence to the very first para of this page ("Custom pools cannot be applied to master nodes" or something) and link down to this section.

Agree with you. Added a note in the first paragraph.

> Also I would change the title of the new section to : "Custom Pools on master nodes" instead of "Apply custom pools on master nodes"

fixed

docs/custom-pools.md
```diff
@@ -4,6 +4,8 @@ Custom pools are pools that inherit from the `worker` pool. They use any Machine
 but add the ability to deploy changes only targeted at the custom pool.
 Since a custom pool inherits from the `worker` pool, any change to the `worker` pool is going to roll out to the custom pool as well (like an OS update during an upgrade).
 
+**Note:** Custom pools on master nodes are not supported by MCO. For more detail, see [Custom pool on master node](#Custom-pool-on-master-node)
```
kikisdeliveryservice: if you pull this up to view the file normally the link seems to be missing - might be the capitalization? try (#custom-pool-on-master-node)

kikisdeliveryservice:
checking here: https://github.com/openshift/machine-config-operator/blob/61c0588c18f61634cd2343d75c83da3910aa57fc/docs/custom-pools.md

this looks good! thanks sinny!

/lgtm

PR #1472, opened 2020-02-13
Make `mcfgv1` and `ctrlcommon` canonical names for importing

LorbusChris:
Entirely code-style.

Not having agreement on the name to use to import the package can lead to not-so-pretty double imports of the same package (as seen here in `container_runtime_config_controller.go`, where the api is imported as both `v1` and  `mcfgv1`).
The same is true for `pkg/controller/common` which is imported with different names across files, and also imported twice in one (`render_controller.go`).

This PR removes the double imports, and canonicalizes the import names as `mcfgv1` and `ctrlcommon` across all files in this repo.


umohnani8:
LGTM
thanks @LorbusChris!

pkg/server/cluster_server.go
```diff
@@ -13,7 +13,7 @@ import (
 	clientcmd "k8s.io/client-go/tools/clientcmd"
 	clientcmdv1 "k8s.io/client-go/tools/clientcmd/api/v1"
 
-	v1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
```
mtrmac: …/pkg/generated/… is a different package; should it use the same prefix?

LorbusChris: maybe this is the distinction we should make here, call the generated pkg (and only that!) `v1`
Will revert this part.

LorbusChris:
I found another double import of `pkg/controller/common`, which is also imported with different names in different files. Second commit incoming.
/hold

LorbusChris:
@mtrmac the api package will be moved into the api repo eventually -- I believe it is common practice to spell out the names there and not to shorten them.

LorbusChris:
and updated!

/hold cancel

runcom:
loving this - unfortunately we can't merge this right away as we need to wait at least Monday

/hold

runcom:
lifting hold I guess - this has 0 risk

/hold cancel

PR #1465, opened 2020-02-12
bug 1804418: remove the etcd-member pod because we no longer need it

deads2k:
When combined with https://github.com/openshift/cluster-etcd-operator/pull/112, we can get a working install and a working upgrade.

/hold

PR #1473, opened 2020-02-13
bug 1801397: temporarily skip setting failures on degraded machine config pool

deads2k:
per our discussion in slack, we need this PR to skip setting degraded on machineconfigoperator when the pools are bad so that we can land https://github.com/openshift/cluster-etcd-operator/pull/142 and https://github.com/openshift/machine-config-operator/pull/1465.  After that, we will no longer go degraded in 4.4.

You can see an example of the condition here: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/112/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws/597 .  The MCO is completely correct to set it, but we need to break the stalemate.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -20,6 +20,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

holding until we're ready to merge this, but I suspect it's coming.

kikisdeliveryservice:
reported that this test (e2e-aws-scaleup-rhel7) is always broken and but it's not reqd so..

/skip

kikisdeliveryservice:
Also @deads2k your sync.go has some problems and won't pass verify until tweaked:

``` golangci-lint run --build-tags="containers_image_openpgp exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_ostree_stub"
pkg/operator/sync.go:564: File is not `gofmt`-ed with `-s` (gofmt)
				degraded{
make: *** [verify] Error 1 ```

kikisdeliveryservice:
re-adding hold until it's time just as a safeguard.

/hold

deads2k:
https://github.com/openshift/machine-config-operator/pull/1476 is the only commit to merge since this tested green.  it isn't related.  merging to get the train moving.

PR #1474, opened 2020-02-14
mcc: accept ign3 & translate down

kikisdeliveryservice:
Phase 1 of Ign3 transition, the MCC accept Ign3 configs translates them down to version 2 and applies to pool while preserving the original V3 configs. New func IgnParseWrapper() allows us to use one function to parse both configs and can be used for later phases.

Related pre-work : https://github.com/coreos/ign-converter/pull/4


kikisdeliveryservice:
` Feb 14 03:38:05.549: INFO:  > ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource [] `

/retest

runcom:
amazing, I think this has to be a combination of this and https://github.com/openshift/machine-config-operator/pull/996 in order to actually accept Ign v3 and have a smoke e2e that creates a v3 MachineConfig (which is later translated down to v2 internally for the 4.5 timeframe cc @yuqi-zhang )

yuqi-zhang:
If we want https://github.com/openshift/machine-config-operator/pull/996 in first we will probably have to adapt this to convert from v3 and rawextention. We also would need a way to detect versioning in a rawextention object before converting, which 996 currently does not seem to have. https://github.com/openshift/machine-config-operator/pull/996/files#r381541065

kikisdeliveryservice:
Finally rebased to pick up new RawExt changes after being blocked but that PR.

kikisdeliveryservice:
@ashcrow those failures don't have anything to do with this pr. no need to retest just yet! :)

kikisdeliveryservice:
rhel7 job permafailing (already reported)

/skip

kikisdeliveryservice:
```
level=error
level=error msg="Error: Error adding instances to InstanceGroup: googleapi: Error 503: Internal error. Please try again or contact Google Support. (Code: '5A358B9B6FC42.A258483.6F08BE54'), backendError"
level=error
level=error msg="  on ../tmp/openshift-install-267295217/bootstrap/main.tf line 87, in resource \"google_compute_instance_group\" \"bootstrap\":"
level=error msg="  87: resource \"google_compute_instance_group\" \"bootstrap\" {"
```

/test e2e-gcp-upgrade

kikisdeliveryservice:
Upgrade is currently breaking on `no 'cli' image was tagged into the stable-initial stream, that image is required for building a release` across prs.. Reported.

kikisdeliveryservice:
Tests are broken - reported... :frowning_face:

kikisdeliveryservice:
ci is down so im not retesting. this did have passing runs, ill try later

/skip

test/e2e/mcd_test.go
```diff
@@ -533,3 +534,46 @@ func TestCustomPool(t *testing.T) {
...
+
+	// create a dummy MC with an sshKey for user Core
+	mcName := fmt.Sprintf("ign3cfg-worker-%s", uuid.NewUUID())
+	mcadd := &mcfgv1.MachineConfig{}
+	mcadd.ObjectMeta = metav1.ObjectMeta{
+		Name:   mcName,
+		Labels: mcLabelForWorkers(),
+	}
+	// create a new MC that adds a valid user & ssh key
+	Ign3Config := ign3types.Config{}
```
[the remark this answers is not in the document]
kikisdeliveryservice: yeah can do

pkg/controller/common/helpers.go
```diff
@@ -28,16 +32,26 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
...
-		outIgn = igntypes.Config{}
+		outIgn = ign2types.Config{}
 	} else {
-		parsedIgn, report, err := ign.Parse(configs[0].Spec.Config.Raw)
+		parsedIgn, err := IgnParseWrapper(configs[0].Spec.Config.Raw)
 		if err != nil {
-			return nil, errors.Errorf("parsing Ignition config failed with error: %v\nReport: %v", err, report)
+			return nil, err
+		}
+		switch ign := parsedIgn.(type) {
```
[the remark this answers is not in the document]
kikisdeliveryservice: good point, i meant to change it but got lost in the shuffle

pkg/controller/common/helpers.go
```diff
@@ -28,16 +32,26 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
...
+		case ign3types.Config:
+			convertedIgn, err := convertIgnition3to2(ign)
+			if err != nil {
+				return nil, err
+			}
+			outIgn = convertedIgn
+		default:
+			ign2 := parsedIgn.(ign2types.Config)
```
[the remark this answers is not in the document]
kikisdeliveryservice: switch statements need defaults and you need to type assert to get the value.

pkg/controller/common/helpers.go
```diff
@@ -46,15 +60,25 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
...
-			appendIgn = igntypes.Config{}
+			appendIgn = ign2types.Config{}
 		} else {
-			parsedIgn, report, err := ign.Parse(configs[idx].Spec.Config.Raw)
+			parsedIgn, err := IgnParseWrapper(configs[idx].Spec.Config.Raw)
 			if err != nil {
-				return nil, errors.Errorf("parsing appendix Ignition config failed with error: %v\nReport: %v", err, report)
+				return nil, err
+			}
+			switch ign := parsedIgn.(type) {
```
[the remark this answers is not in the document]
kikisdeliveryservice: agree

pkg/controller/common/helpers.go
```diff
@@ -136,14 +182,33 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+// IgnParseWrapper parses rawIgn for both V2 and V3 ignition configs and returns
+// a V2 or V3 Config or an error. This wrapper is necessary since V2 and V3 use different parsers.
+func IgnParseWrapper(rawIgn []byte) (ignconfig interface{}, err error) {
+	ignCfg, rpt, err := ign.Parse(rawIgn)
+
+	// this is an ignv2cfg that was successfully parsed
+	if err == nil {
+		return ignCfg, nil
+	}
+	if err.Error() == "unsupported config version" {
```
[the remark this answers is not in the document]
kikisdeliveryservice: will check on it

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: wait ooh i see.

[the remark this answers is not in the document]
kikisdeliveryservice: yeah i was just following the pattern that existed in the func but yeah. let me take another look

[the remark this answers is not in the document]
runcom: what happens if for any reason this isn't `ign2types.Config` tho? we didn't have such issue before ofc, but we just need to check that this won't cause troubles like a panic. There's:
```go
ign2, ok := parsedIgn.(ign2types.Config)
if !ok {
// gracefully exit
}
```
maybe the above can help (but haven't validated it fully ofc)

runcom:
I think this is getting together pretty well :) and I'd love as well to see the e2e passing once CI gets in a better shape - I've left just a minor comment but architecturally this is good 👍 

/approve

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: done

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: yep. done.

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: done.

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: the function that gives the parsedIgn only returns 2 or 3 or an error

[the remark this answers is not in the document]
kikisdeliveryservice: actually on second thought: ill add to ensure the functions don't change out from under each other.

test/e2e/mcd_test.go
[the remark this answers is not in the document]
kikisdeliveryservice: In progress will ping once it's done and ci is cooperating. :+1:

kikisdeliveryservice:
ok finally got this to run the functionality works IRL, but my tests is wrong..
```-- FAIL: TestIgn3Cfg (0.02s)
    mcd_test.go:557: 
        	Error Trace:	mcd_test.go:557
        	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"MachineConfig.machineconfiguration.openshift.io \"ign3cfg-worker-c570e8d2-1e71-471e-9fe9-54bc4de8ba35\" is invalid: spec.config.ignition.config.replace.source: Invalid value: \"null\": spec.config.ignition.config.replace.source in body must be of type string: \"null\"", Reason:"Invalid", Details:(*v1.StatusDetails)(0xc00012e780), Code:422}}
        	Test:       	TestIgn3Cfg
        	Messages:   	failed to create MC 
```
gonna have to figure it out, im missing something small here..

test/e2e/mcd_test.go
```diff
@@ -533,3 +534,46 @@ func TestCustomPool(t *testing.T) {
...
+	tempUser := ign3types.PasswdUser{Name: "core", SSHAuthorizedKeys: []ign3types.SSHAuthorizedKey{"1234_test_ign3"}}
+	Ign3Config.Passwd.Users = append(Ign3Config.Passwd.Users, tempUser)
+	Ign3Config.Ignition.Version = "3.0.0"
```
runcom, line 552: https://github.com/openshift/machine-config-operator/blob/fcos/pkg/controller/common/helpers.go#L159-L175

test/e2e/mcd_test.go
```diff
@@ -533,3 +534,46 @@ func TestCustomPool(t *testing.T) {
...
+	rawIgnConfig := helpers.MarshalOrDie(Ign3Config)
+	mcadd.Spec.Config.Raw = rawIgnConfig
+
+	_, err := cs.MachineConfigs().Create(context.TODO(), mcadd, metav1.CreateOptions{})
```
runcom, line 556: This is failing because the CRD manifest for MachineConfig is still v2 (manifests/machineconfig.crd.yaml) and it errors out because ignition.config.replace.source is a pointer to string in v3 which can be nullable. I'm not sure what we can do here to allow both as we're now allowing v2 and v3 and they have differences 🤔 @yuqi-zhang do you have any idea if we can allow that? or just make the string nullable as a pointer?

yuqi-zhang: I think this should be fixable by adding a `nullable: true` after https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfig.crd.yaml#L87 , in parallel with `source`. Example usage: https://github.com/openshift/machine-config-operator/blob/master/manifests/machineconfig.crd.yaml#L316

If that does not work we can just drop the field altogether.

To be precise crd validation happens the same way when you oc create -f I think. My best guess is that Create() gets a full v3 object validated whereas oc create -f just assumes its a spec 2 object which is processed later.

yuqi-zhang: We may see other locations with this error since a lot of spec 3 uses pointers, so you may have to fix as we go. We can compare between https://github.com/openshift/machine-config-operator/blob/c8e9ccf5569ae05b2bdaf3a44c5e371ebb5e6ade/vendor/github.com/coreos/ignition/config/v2_3/types/schema.go and https://github.com/openshift/machine-config-operator/blob/c8e9ccf5569ae05b2bdaf3a44c5e371ebb5e6ade/vendor/github.com/coreos/ignition/v2/config/v3_0/types/schema.go

kikisdeliveryservice:
CRD fix did the trick!!
```I0417 19:12:44.377162       1 helpers.go:151] Successfully translated ignition V3 config to ignition V2 config: {{{[] <nil>} {{[]}} {<nil> <nil>} 2.2.0} {[]} {[] [{<nil>  []  core false false false <nil>  [1234_test_ign3]  false <nil>}]} {[] [] [] [] [] []} {[]}}
```
Just have to fix up my test..

test/e2e/mcd_test.go
LorbusChris: Is this really the cause of the problem here? The diff for the CRD in the fcos branch is empty (`git diff fcos manifests/machineconfig.crd.yaml`)

LorbusChris: that is to say the CRD validation works as-is on Spec3 in OKD

LorbusChris: I think we hit a very similar same issue in installer:fcos recently: The marshaling for empty pointer structs didn't behave as desired and they were included in the marshalled data as weird nil fields. Our solution for OKD (which has since also been adopted in the fedora-config-transpiler as well) was to switch out the standard json.Marshal with a drop in replacement from `github.com/clarketm/json` which supports supports zero values of structs
with omitempty markers when marshaling them (it'll drop those fields).
This may be a better solution than regressing on openAPI spec validation.

for reference: https://github.com/coreos/fcct/pull/90

kikisdeliveryservice: @lorbus yours ARE different you don't have jerry's latest chagnes:
https://github.com/openshift/machine-config-operator/commits/fcos/manifests/machineconfig.crd.yaml

https://github.com/openshift/machine-config-operator/commits/master/manifests/machineconfig.crd.yaml

The last commit you have that matches the MCO is from December =)

kikisdeliveryservice: From my count that's 5 commits your branch is missing..

LorbusChris: We don't currently preserve commit history on the fcos branch to include changes with merges instead of rebases. Did you run the diff cmd? (I just double checked and the diff is empty for me)

kikisdeliveryservice: didnt diff, i assumed you pulled commit history. would be good to keep it since to preserve authors..

test/e2e/mcd_test.go
kikisdeliveryservice: yeah i considered doing this but as we don't support ign3 yet I didnt want to create a public function creating them.

test/e2e/mcd_test.go
LorbusChris: it would be good indeed, but it was a compromise we had to make to catch up on the ever growing gap between fcos and master (fixing up every commit's merge errors is much more work than just checking out the state you want on top of current commit. And with the fcos branch being temporary all of that work would be thrown away later anyway. we took that route route.) Hopefully we won't have to maintain that duality much longer now :)

LorbusChris:
Another note: the template renderer still exclusively uses the container-linux-config transpiler to transpile the cl config yamls to spec2 ign. For spec3 this will need Fedora CoreOS Config formatted yaml files, converted to spec3 ign json with https://github.com/coreos/fcct. I guess there could be similar mechanism as the one in this PR to achieve dual-support for them as well - that can probably be done as a follow up though.

kikisdeliveryservice:
According to the logs this totally worked ok for both sshkeys and files (generated MC&wrote all the things), but it's having and issue with:
```         	
Error:      	Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0000eef20), Stderr:[]uint8(nil)}
        	Test:       	TestIgn3Cfg
        	Messages:   	failed to exec cmd [grep 1234_test_ign3 /rootfs/home/core/.ssh/authorized_keys] on node ci-op-5hfm4-w-b-wr4rk.c.openshift-gce-devel-ci.internal 
```
Ran into this once before on someone else's PR and IIRC it was solved with just retesting? 

/retest

kikisdeliveryservice:
Duh, just checked test runs (which are all red) and saw exact same error on a different test:

```
 --- FAIL: TestContainerRuntimeConfigPidsLimit (7.85s)
    utils_test.go:176: 
        	Error Trace:	utils_test.go:176
        	            				ctrcfg_test.go:169
        	            				ctrcfg_test.go:64
        	            				ctrcfg_test.go:23
        	Error:      	Expected nil, but got: &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0003ee560), Stderr:[]uint8(nil)}
        	Test:       	TestContainerRuntimeConfigPidsLimit
        	Messages:   	failed to exec cmd [cat /rootfs/etc/crio/crio.conf.d/01-ctrcfg-pidsLimit] on node ci-op-zgkd8-w-c-wjtj7.c.openshift-gce-devel-ci.internal 
```

So it's not this PR.

kikisdeliveryservice:
Will test a little more tomorrow and clean up commits, but this seems to work well with just `spec.config.igntion.config.replace.source` changed.

test/e2e/mcd_test.go
LorbusChris: I don't think this is resolved -- I believe the problem could be fixed by using a marshaler that excludes the empty pointer struct fields from the data, then a regression on the OpenAPI validation wouldn't be needed (as I mentioned before we do that on OKD already withe exactly the same validation theme used for in OCP right now, compare: https://github.com/openshift/machine-config-operator/blob/fcos/test/helpers/helpers.go#L4)

kikisdeliveryservice:
Cleaned up commits, double checked runs and this is working.  gcp-op still hitting snags..

test/e2e/mcd_test.go
LorbusChris: The above will become necessary when moving the CL config yamls to FCCT format in order to produce valid marshaled data. As this is out of scope for this PR, I'll mark as resolved.

LorbusChris:
Currently, the v3 json raw data contains nil values: If one parses and prints it, `ignCfgV3.Ignition` will look like `{{[] {<nil> {<nil>}}} {{[]}} {<nil> <nil>} 3.0.0}` and this will not pass OpenAPI validation as nil is not a valid string value on the `spec.config.ignition.config.replace.source` field .

A workaround for these pointer struct fields is to use a drop-in replacement json parser that excludes these zero values of structs with omitempty from the marshaled data (i.e. github.com/clarketm/json).

We already do this for OKD (there, we hit an issue with the yaml parser used for FCCT configs that does not support json null values, and therefore failed. This change also enabled us to go back to using the entire, existing open api validation scheme that was created for spec2 unchanged).

That is to say, no regression on the spec validation would be needed when switching the json parser, with the neat side-benefit that the transmitted data on the API is much smaller.

For reference see https://github.com/openshift/machine-config-operator/commit/563275c76cfa23a0bcb63c154f0312adc6542dae

The open api validation regression can otherwise be undone later when moving to FCCT configs in the installer, which will make it required to either switch the parser or walk the struct manually to exclude empty fields here in this repo.

/approve
/retest

runcom:
This should be good now pending CI which hasn't cooperated so far, we'll figure out follow ups then - going to lgtm this as it's definitely done but we need to hear from CI, the lgtm will go away if the code needs changing ofc.

/lgtm

kikisdeliveryservice:
ok so i checked this e2e test on a cluster and it was fine, so this is not failing bc of something being wrong with the test itself.. evidence:
```
=== RUN   TestIgn3Cfg
--- PASS: TestIgn3Cfg (955.14s)
    mcd_test.go:563: Created ign3cfg-worker-c218e32b-4725-4a25-9a35-6b4ca4f33ac0
    utils_test.go:60: Pool worker has rendered config ign3cfg-worker-c218e32b-4725-4a25-9a35-6b4ca4f33ac0 with rendered-worker-59bfe284f243c3e74ea1855e2e375304 (waited 4.084857013s)
    utils_test.go:82: Pool worker has completed rendered-worker-59bfe284f243c3e74ea1855e2e375304 (waited 15m46.042834413s)
    mcd_test.go:580: Node ip-10-0-129-180.us-west-2.compute.internal has SSH key
    mcd_test.go:586: Node ip-10-0-129-180.us-west-2.compute.internal has file
    mcd_test.go:580: Node ip-10-0-136-104.us-west-2.compute.internal has SSH key
    mcd_test.go:586: Node ip-10-0-136-104.us-west-2.compute.internal has file
    mcd_test.go:580: Node ip-10-0-150-32.us-west-2.compute.internal has SSH key
    mcd_test.go:586: Node ip-10-0-150-32.us-west-2.compute.internal has file
```
```
$ oc get mcp

NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT   AGE
master   rendered-master-f9c60da8b65711275a3c1dcbf4cc73f4   True      False      False      3              3                   3                     0                      42m
worker   rendered-worker-59bfe284f243c3e74ea1855e2e375304   True      False      False      3              3                   3                     0                      42m

```
```
I0421 01:30:52.625449       1 render_controller.go:497] Generated machineconfig rendered-worker-59bfe284f243c3e74ea1855e2e375304 from 6 configs: [{MachineConfig  00-worker  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-container-runtime  machineconfiguration.openshift.io/v1  } {MachineConfig  01-worker-kubelet  machineconfiguration.openshift.io/v1  } {MachineConfig  99-worker-a7297686-4a9d-4110-9cda-c2f815663b5d-registries  machineconfiguration.openshift.io/v1  } {MachineConfig  99-worker-ssh  machineconfiguration.openshift.io/v1  } {MachineConfig  ign3cfg-worker-c218e32b-4725-4a25-9a35-6b4ca4f33ac0  machineconfiguration.openshift.io/v1  }]
....
I0421 01:46:37.288479       1 status.go:84] Pool worker: All nodes are updated with rendered-worker-59bfe284f243c3e74ea1855e2e375304
```

kikisdeliveryservice:
I opened a BZ bc I dug for hours, asked around and still have no idea what's happening in these runs..

https://bugzilla.redhat.com/show_bug.cgi?id=1826086

kikisdeliveryservice:
Update: turns out this PR has been failing due to various incarnations of: https://bugzilla.redhat.com/show_bug.cgi?id=1802534

this whole time =(

sinnykumari:
test is passing now :tada:

LorbusChris:
now this PR just needs a BZ :)

ashcrow:
This was done before the cut off. We should have done lgtm before hand. However, I've verified it's OK for us to bypass the bz check. In the future we should be better about ensuring lgtm before cut off :-).

kikisdeliveryservice:
> This was done before the cut off. We should have done lgtm before hand. However, I've verified it's OK for us to bypass the bz check. In the future we should be better about ensuring lgtm before cut off :-).

@ashcrow it had a lgtm given 3 days ago and all the way until at least last night. that's why the bot was retesting.

kikisdeliveryservice:
sigh..now i have to rebase bc another pr slipped in before this one that changed vendor files... =(

kikisdeliveryservice:
Ok rebased onto master... can someone re-lgtm please?

kikisdeliveryservice:
back to the exec failing on the node...

looking closely in the logs i see for the node it's trying to rsh into:
```                "creationTimestamp": "2020-04-22T17:07:27Z",
                "labels": {
                    "beta.kubernetes.io/arch": "amd64",
                    "beta.kubernetes.io/instance-type": "n1-standard-4",
                    "beta.kubernetes.io/os": "linux",
                    "failure-domain.beta.kubernetes.io/region": "us-east1",
                    "failure-domain.beta.kubernetes.io/zone": "us-east1-b",
                    "kubernetes.io/arch": "amd64",
                    "kubernetes.io/hostname": "ci-op-s4kwt-w-b-5bz8l.c.openshift-gce-devel-ci.internal",
                    "kubernetes.io/os": "linux",
                    "node-role.kubernetes.io/worker": "",
                    "node.kubernetes.io/instance-type": "n1-standard-4",
                    "node.openshift.io/os_id": "rhcos",
                    "topology.kubernetes.io/region": "us-east1",
                    "topology.kubernetes.io/zone": "us-east1-b"
                },
                "managedFields": [
                    {
                        "apiVersion": "v1",
                        "fieldsType": "FieldsV1",
                        "fieldsV1": {
                            "f:status": {
                                "f:conditions": {
                                    "k:{\"type\":\"NetworkUnavailable\"}": {
                                        "f:message": {},
                                        "f:reason": {},
                                        "f:status": {}
                                    }
                                }
                            }
                        },
                        "manager": "openshift-sdn-controller",
                        "operation": "Update",
                        "time": "2020-04-22T17:07:27Z"
                    },
```

kikisdeliveryservice:
Another FTR: I have tested this new test on my own clusters on both aws and gcp and the test is able to rsh and check the files correctly, so again it's not this PR.

I've been updating my BZ while digging through the logs: https://bugzilla.redhat.com/show_bug.cgi?id=1826086

I've consistently seen i/o timeouts:```

I0422 16:42:03.757648       1 leaderelection.go:242] attempting to acquire leader lease  openshift-sdn/openshift-network-controller...
E0422 17:06:45.772571       1 leaderelection.go:331] error retrieving resource lock openshift-sdn/openshift-network-controller: Get https://api-int.ci-op-6nw6wy67-1354f.origin-ci-int-gce.dev.openshift.com:6443/api/v1/namespaces/openshift-sdn/configmaps/openshift-network-controller: dial tcp 10.0.0.2:6443: i/o timeout
E0422 17:07:29.993399       1 leaderelection.go:331] error retrieving resource lock openshift-sdn/openshift-network-controller: Get https://api-int.ci-op-6nw6wy67-1354f.origin-ci-int-gce.dev.openshift.com:6443/api/v1/namespaces/openshift-sdn/configmaps/openshift-network-controller: dial tcp 10.0.0.2:6443: i/o timeout
E0422 17:13:54.994370       1 leaderelection.go:331] error retrieving resource lock openshift-sdn/openshift-network-controller: Get https://api-int.ci-op-6nw6wy67-1354f.origin-ci-int-gce.dev.openshift.com:6443/api/v1/namespaces/openshift-sdn/configmaps/openshift-network-controller: dial tcp 10.0.0.2:6443: i/o timeout
```

And also see ovs broken pipe:
```
2020-04-22T18:00:33.580Z|00018|reconnect|WARN|unix#3502: connection dropped (Connection reset by peer)
2020-04-22T18:03:20.709Z|00019|jsonrpc|WARN|unix#3612: send error: Broken pipe
2020-04-22T18:03:20.709Z|00020|reconnect|WARN|unix#3612: connection dropped (Broken pipe)
2020-04-22T18:10:15.711Z|00021|jsonrpc|WARN|unix#3890: send error: Broken pipe
2020-04-22T18:10:15.712Z|00022|reconnect|WARN|unix#3890: connection dropped (Broken pipe)
2020-04-22T18:37:30.737Z|00023|jsonrpc|WARN|unix#4979: send error: Broken pipe
2020-04-22T18:37:30.737Z|00024|reconnect|WARN|unix#4979: connection dropped (Broken pipe)
```

I'm not sure if this is https://bugzilla.redhat.com/show_bug.cgi?id=1802534

@squeed WDYT?

**Comment by runcom** [comment]
found the issue - it's a bug in either the translation code and the update SSH keys routine in MCD:

This is the content of the rendered MC for this ign3 test:
```
passwd: {
users: [
{
name: "core",
sshAuthorizedKeys: [
"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDUq7W38xCZ9WGSWCvustaMGMT04tRohw6AKGzI7P7xql5lhCAReyt72n9qWQRZsE1YiCSQuTfXI1oc8NpSM7+lMLwj12G8z3I1YT31JHr9LLYg/XIcExkzfBI920CaS82VqmKOpI9+ARHSJBdIbKRI0f5Y+u4xbc5UzKCJX8jcKGG7nEiw8zm+cvAlfOgssMK+qJppIbVcb2iZNTsw5i2aX6FDMyC+b17DQHzBGpNbhZYxuoERZVRcnYctgIzuo6fD60gniX0fVvrchlOnubB1sRYbloP2r6UE22w/dpLKOFE5i7CA0ZzNBERZ94cIKumIH9MiJs1a6bMe89VOjjNV devenv "
]
},
{
name: "core",
sshAuthorizedKeys: [
"1234_test_ign3"
]
},
{
name: "core",
sshAuthorizedKeys: [
"1234_test",
"abc_test"
]
}
]
},
```

and you can see the MCD is only using the last one:
```
I0423 09:28:06.841256    2039 update.go:500] user data to be verified before ssh update: {<nil>  []  core false false false <nil>  [1234_test abc_test]  false <nil>}
```

affected code is here:
```
// Update a given PasswdUser's SSHKey
func (dn *Daemon) updateSSHKeys(newUsers []igntypes.PasswdUser) error {
	if len(newUsers) == 0 {
		return nil
	}

	// we're also appending all keys for any user to core, so for now
	// we pass this to atomicallyWriteSSHKeys to write.
	var concatSSHKeys string
	for _, k := range newUsers[len(newUsers)-1].SSHAuthorizedKeys {
		concatSSHKeys = concatSSHKeys + string(k) + "\n"
	}
	if !dn.mock {
		// Note we write keys only for the core user and so this ignores the user list
		if err := dn.atomicallyWriteSSHKey(concatSSHKeys); err != nil {
			return err
		}
	}
	return nil
}
```

I'm fixing it

runcom:
@kikisdeliveryservice why aren't you allowing edits from other maintainers tho? 🤔 I wanted to push my commit to fix this here and couldn't - can you cherry-pick my commit into this PR? commit is here https://github.com/openshift/machine-config-operator/pull/1673/commits/9d9d7e017060c11dfe869a53fe7d6a1fd76f977c - I've also created a PR so you can merge it here directly....

runcom:
Pending my fix

/hold

sinnykumari:
Nice debugging Antonio!

runcom:
I think we can leave this as is now and it'll go in as soon as tests turn green as the override is already taken into account

runcom:
also github is having issues from time to time today

/retest

yuqi-zhang:
The sshkey fix LGTM, it passes tests on https://github.com/openshift/machine-config-operator/pull/1676 as well

runcom:
fixed finally https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1676/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/2005

I'll pick that commit here as that's the one really fixing the issue here.

PR #1477, opened 2020-02-14
*: remove etcd quorum guard

hexfusion:
We are migrating `quorum-guard` to the `openshift-etcd` namespace. The reason it originally lived here was that we did not have an operator to take reference to the cli image it uses.

depends on https://github.com/openshift/cluster-etcd-operator/pull/142


hexfusion:
/hold 

for https://github.com/openshift/cluster-etcd-operator/pull/142

abhinavdahiya:
hmm.. cvo doesn't clean up / prune removed deployments (or any other object) from release images, i think mco / ceo will have to take ownership of that.

hexfusion:
> hmm.. cvo doesn't clean up / prune removed deployments (or any other object) from release images, i think mco / ceo will have to take ownership of that.

yeah good point to manage this we have a controller "guardbudget" that will handle this teardown this on CEO end.

cgwalters:
etcd has left the building

runcom:
> depends on [openshift/cluster-etcd-operator#142](https://github.com/openshift/cluster-etcd-operator/pull/142)

this PR has been closed :thinking: does this still depend on something?

hexfusion:
we are not going to go this route in 4.4

PR #1478, opened 2020-02-14
[FCOS] non-virtual-ip: replace the script with podman wrapper

vrutkovs:
Replace Python script with a wrapper which runs the script in a podman container (using host networking).

Requires https://github.com/openshift/baremetal-runtimecfg/pull/46
Fixes https://github.com/openshift/machine-config-operator/issues/1430

Master counterpart: https://github.com/openshift/machine-config-operator/pull/1521


templates/worker/00-worker/baremetal/files/baremetal-non-virtual-ip.yaml
```diff
@@ -3,136 +3,9 @@ mode: 0755
...
-        except SubnetNotFoundException:
-            sys.exit(1)
-
-    if __name__ == '__main__':
-        main()
+    #!/usr/bin/bash
+    podman run --rm \
+      --net=host \
+      --entrypoint=/usr/bin/non_virtual_ip \
+      -ti {{ .Images.BaremetalRuntimeCfgImage }} \
```
mandre: Have you tried without attaching a TTY? I believe we shouldn't need the `-ti` option.

mandre:
This is going to conflict with https://github.com/openshift/machine-config-operator/pull/1483, can we wait until it merges?
I'd like to keep the port clean in #1483 and resolve the conflict in your PR if you don't mind.

vrutkovs:
>This is going to conflict with #1483, can we wait until it merges?

`fcos` branch is independent of master, so we'll cherry-pick it back to master if the idea works

mandre:
> > This is going to conflict with #1483, can we wait until it merges?
> 
> `fcos` branch is independent of master, so we'll cherry-pick it back to master if the idea works

Ah right, I forgot this PR was against a different branch. All good then.

mandre:
This should be rebased after https://github.com/openshift/machine-config-operator/pull/1483.

Also, after re-syncing with master, you should leave out ovirt platform because they're still using the older version of `non-virtual-ip` script.

Note that there is now another python script `nodeip-finder` that should be moved to `baremetal-runtimecfg` the same way you did in https://github.com/openshift/baremetal-runtimecfg/pull/46.

vrutkovs:
>Also, after re-syncing with master, you should leave out ovirt platform because they're still using the older version of non-virtual-ip script.

Are these incompatible? I was hoping one script could be reused for all platoforms.
cc @rgolangh

vrutkovs:
Updated https://github.com/openshift/baremetal-runtimecfg/pull/46.

I didn't change the wrapper in `fcos` branch, as its still lagging behind master (it would be rebased soon)

vrutkovs:
This now requires https://github.com/openshift/baremetal-runtimecfg/pull/48 (since fcos branch promotes to 4.4 for now)

vrutkovs:
/cc @LorbusChris 

This should fix oVirt / Openstack not joining worker nodes, PTAL

celebdor:
It looks good. I wonder how much delay it is introduced by podman pulling the images, but with a local registry it shouldn't be much.

LorbusChris:
is this a real concern tho? After all we deliver all of OpenShift's operators as container images as well

vrutkovs:
This script is used by NetworkManager, so potentially we may end up with misconfigured networking

cgwalters:
Thanks a ton for doing this!  This also solves the problem of having substantial baremetal code inside the MCO, which is a double win.

LorbusChris:
We should get this into master as well asap. Do we have OpenStack e2e tests here?
/approve

vrutkovs:
/hold

Wrong script names

vrutkovs:
/hold cancel

Update entrypoint, ready for review

LorbusChris:
xref master PR: https://github.com/openshift/machine-config-operator/pull/1521
Let's wait for e2e testing on master to complete before merging here.
/approve

PR #1403, opened 2020-01-22
Add openAPIV3Schema to CRDs

sanchezl:

Issue #1443, opened 2020-02-05
The MCD is stuck and unable to recover from file degradations


yuqi-zhang:
--------------------------------------------------
BUG REPORT INFORMATION
--------------------------------------------------

**Description**
Let's say if I have a file at `/home/core/test`, and then I apply a new machineconfig to write to `/home/core/test/test`, since `/home/core/test` is a file, the MCO properly catches that it is unable to create a directory there, and thus degrades.

If I then delete the machineconfig that introduced this change, the MCC will properly detect that the worker pool should go back to targeting the previous machineconfig for the pool worker. However, the MCD running on the node does not detect this change. It will continuously fail-loop on `Marking Degraded due to: failed to create directory "/home/core/test": mkdir /home/core/test: not a directory`, marking the node as schedulingdisabled and failing to make any progress. In fact, since the annotation on the node never gets updated, even deleting the MCD pod doesn't fix the error, as the new one will attempt the same update and fail on the same error.

To recover, we would need to update the annotation on the node by hand to the previous `desiredConfig`, and the manually `oc adm uncordon node`. This is obviously not desired behaviour, as we should be able to recover automatically when the MC is deleted.

I've not tested every degrade-recovery scenario but I remember we were able to recover from some cases before. Will test to see if other types of degrades exhibit the same behaviour.

**Steps to reproduce the issue:**
1. Create a file with a machineconfig snippet like:
```
    ...
    storage:
      files:
      - contents:
          source: data:,hello%20world%0A
          verification: {}
        filesystem: root
        mode: 420
        path: /home/core/test
```
2. Create a second file with another machineconfig like:
```
    ...
    storage:
      files:
      - contents:
          source: data:,hello%20worlddd%0A
          verification: {}
        filesystem: root
        mode: 420
        path: /home/core/test/test
```
3. Notice that the second machineconfig causes a degrade on one of the nodes
4. Delete the second machineconfig, and notice that the node is unable to recover

**Additional environment details (platform, options, etc.):**
Reproduced so far on 4.4 Azure and AWS

PR #1485, opened 2020-02-18
generate CRD manifests and fix for oc explain

yuqi-zhang:
There are 3 commits in this change that aims to add OpenAPIV3 validation and get `oc explain` to work for MCO repo CRDS. Note that this does NOT include the actual generator code, this that does not current work with the MCO repo manifests. Will open another PR with the generator and explain what needs to be fixed for that to work, but for now this should get `oc explain` going.

Commit 1:
Generate crd manifests with OpenAPIV3 schemas

Generate validation for:

containerruntimeconfig
controllerconfig
kubeletconfig
machineconfig
machineconfigpool

Using update-codegen-crds target from
https://github.com/openshift/build-machinery-go, which will be
included in a separate commit as it is not directly usable with the
MCO repo until https://github.com/coreos/ignition/issues/917 is
resolved. The current generation is done via hacking ignition's
config/v2_2/types/schema.go to have dummy `json:",inline"` for fields
that are causing errors.

Some background on this, there were 2 attempts to add generated
schemas in: https://github.com/openshift/machine-config-operator/pull/1403
and https://github.com/openshift/machine-config-operator/pull/955.
This supercedes those with updated generation methods.

Commit 2:
Fix autogenerated crd schemas for oc explain

Now that we have generated schemas, for oc explain to work we need to
turn on validation via preserveUnknownFields: false in the spec.

There is also one error in the autogenerated machineconfigpool spec
due to IntOrString, see: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#intorstring

Also minor formatting in machineconfig.crd.yaml.

Commit 3:
update generated bindata after crd changes


manifests/containerruntimeconfig.crd.yaml
```diff
@@ -14,9 +14,144 @@ spec:
...
     status: {}
   versions:
-    - name: v1
-      served: true
-      storage: true
+  - name: v1
+    served: true
+    storage: true
+  "validation":
+    "openAPIV3Schema":
```
ericavonb: Any idea why it added quotes around these?

yuqi-zhang: I tested and that doesn't seem to make a difference either way. The generator generates them as such. See for example crds in api: https://github.com/openshift/api/tree/master/config/v1

manifests/machineconfig.crd.yaml
```diff
@@ -39,205 +41,401 @@ spec:
...
+                          gecos:
+                            type: string
+                          groups:
+                            type: array
+                            items:
+                              type: string
+                          homeDir:
+                            type: string
                           name:
-                            description: "name of user. Must be \"core\" user."
```
ericavonb: How can we get these descriptions added back again? Would we need to add proper godoc comments to the types in the ignition repo? For ones like this where we further restrict what ignition we accept, how would be indicate that to users?

manifests/machineconfig.crd.yaml
```diff
@@ -39,205 +41,401 @@ spec:
...
+                              uid:
+                                type: integer
+                          gecos:
+                            type: string
+                          groups:
+                            type: array
+                            items:
+                              type: string
+                          homeDir:
+                            type: string
```
ericavonb: We don’t support all these ignition fields. I think it’s only the ones we had that we allow

runcom:
awesome 👍 

/approve

will leave to the team to further take a look and address comments, well done!

manifests/machineconfig.crd.yaml
yuqi-zhang: Hm, good question. This is obviously a result of the auto generation, and in our case we probably want to make manual fixes for `machineconfig` that differs from general ignition usage. I don't think we should modify ignition directly, but keeping manual modifications may be costly.

manifests/machineconfig.crd.yaml
yuqi-zhang: Same as above, I will look to manually edit for now, unless you know of a better method we can source these MCO-specific use-cases.

manifests/machineconfig.crd.yaml
```diff
@@ -39,205 +41,401 @@ spec:
...
+                            type: boolean
+                          enabled:
+                            type: boolean
+                          mask:
+                            type: boolean
+                          name:
+                            type: string
+            fips:
+              type: boolean
+            kernelArguments:
```
sinnykumari: Similar comment what Erica already mentioned. We should preserve description of these too

sinnykumari:
Interesting error from e2e tests failure
```
 msg="Cluster operator support Disabled is True with NotAuthorized: Reporting was not allowed: your Red Hat account is not enabled for remote support or your token has expired"
```

yuqi-zhang:
/retest

Will fix up the descriptions. In the meantime let's see if that was a flake

yuqi-zhang:
As per review, added back descriptions and removed unused fields. Note that `oc explain` doesn't show these fields anyways.

sinnykumari:
looks good, needs bindata update

yuqi-zhang:
Re-gen'ed bindata and re-ordered commits a bit. Should be good to go.

yuqi-zhang:
Removing BZ as well since this is now targeting master

yuqi-zhang:
/retest

one throttling one api

yuqi-zhang:
`gcp-op` was failing with 
```
ControllerConfig.machineconfiguration.openshift.io \"machine-config-controller\" is invalid: [spec.additionalTrustBundle: Invalid value: \"null\": spec.additionalTrustBundle in body must be of type string: \"null\", spec.proxy: Invalid value: \"null\": spec.proxy in body must be of type object: \"null\", spec.cloudProviderCAData: Invalid value: \"null\": spec.cloudProviderCAData in body must be of type string: \"null\"]"
```
so added `omitempty` to their fields

manifests/machineconfig.crd.yaml
```diff
@@ -79,165 +136,186 @@ spec:
...
+                              When true, the service is masked by symlinking it to /dev/null"
+                            type: boolean
+                          name:
+                            description: Name is the name of the unit. This must be suffixed with a
+                              valid unit type (e.g. 'thing.service')
+                            type: string
+            fips:
+              description: FIPS controls FIPS mode
+              type: boolean
+            kernelArguments:
```
sinnykumari: This should be `kargs` instead of kernelArguments.  From the [log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1485/pull-ci-openshift-machine-config-operator-master-e2e-aws/6547/artifacts/e2e-aws/pods/openshift-machine-config-operator_machine-config-controller-6bcffcc7fb-nmqdw_machine-config-controller.log) that would be reason why mcc pod is failing to render

sinnykumari: spoken a bit early, by looking at code don't think it's required. It could be another case of adding omitempty.

yuqi-zhang:
Ok, I changed it to nullable instead of omitempty and it passed CI.

Also note that there are potentially other fields that need to be nullable. Basically any field that can be null (and is not `omitempty`-able) will fail validation. Are there a lot of potential issues outside what CI can cover?

yuqi-zhang:
Ah, so one more thing, since in `TestReconcileAfterBadMC` we try to give it a bad MC with `networkd` section and wait for it to degrade, that now fails since the validation straight up rejects the MC. I think we should change that test, and then add a test case for e.g. https://github.com/openshift/machine-config-operator/issues/1443 (once we fix that that is).

yuqi-zhang:
I think its not a flake but a change in behaviour, fixing now

yuqi-zhang:
Ok I added ` x-kubernetes-preserve-unknown-fields: true` to the whole Config section, which should fix the test failure.

A bit more background: for oc explain to work, we need to have 

`preserveUnknownFields: false`

which means all fields not defined by the crd schema will be pruned upon generation of the CR. Since we only define ignition sections that are allowed to be machineconfig sections, an invalid ignition spec field will simply get removed when the machineconfig is created. Adding the flag tells validation that anything under Config is allowed, and we will let MCC/MCD do the validation instead, like we do today.

Now it may technically be more inline with the API to just let the fields be pruned, but I feel that disrupts existing user experience. And since we will be switching to RawExtention anyways, that will not be validate-able via the CRD validation, so we will eventually have to do this.

runcom:
/approve

amazing :)

sinnykumari:
Nice work Jerry!
Will wait for the team for any final comments.
/approve

yuqi-zhang:
As a final clarification, this adds schema and schema validation for the manifests that live within our repo. This will enable `oc explain` to work properly, as well as validate new CRs that are applied before they hit the cluster, with the exception of Ignition section items, which we skip and let MCD make the call on whether its valid or not.

yuqi-zhang:
As a final clarification, this adds schema and schema validation for the manifests that live within our repo. This will enable `oc explain` to work properly, as well as validate new CRs that are applied before they hit the cluster, with the exception of Ignition section items, which we skip and let MCD make the call on whether its valid or not.

kikisdeliveryservice:
> As a final clarification, this adds schema and schema validation for the manifests that live within our repo. This will enable `oc explain` to work properly, as well as validate new CRs that are applied before they hit the cluster, with the exception of Ignition section items, which we skip and let MCD make the call on whether its valid or not.

As per our conversation this sounds correct :)
FYI : @rphillips @umohnani8 since this affects kubelet and cr configs

/LGTM

PR #1486, opened 2020-02-18
WIP: add ability to generate crd schemas

yuqi-zhang:
This builds on top of https://github.com/openshift/machine-config-operator/pull/1485, also see context there.

This aims to add the ability to generate CRD schemas by vendoring in `build-machinery-go` https://github.com/openshift/build-machinery-go and following https://github.com/openshift/api#generating-crd-schemas. Note that this does NOT work pending https://github.com/idubinskiy/schematyper/pull/18, so leaving as WIP

Note that also eventually, we'd probably move these crds to openshift/api. We are currently unable to do so pending https://github.com/openshift/machine-config-operator/pull/996

/hold


manifests/machineconfig.crd.yaml
```diff
@@ -39,205 +41,401 @@ spec:
...
+              description: Config is a Ignition Config object.
               type: object
+              required:
+              - ignition
               properties:
                 ignition:
-                  description: "ignition section contains metadata about the configuration itself"
                   type: object
                   properties:
+                    config:
```
LorbusChris: just a note: We won't get any of the CRD schema validation for RawExtension :(
Spec2 and Spec3 are different structs, so we'd either have to choose one for (manually added) validation or none and do validation in the Installer/MCO.

yuqi-zhang: It's probably ok since we will have error reporting via the MCP regarding if a conf is invalid anyways. This generation hopefully won't be needed here and can go directly to the API repo once the conversion is in.

For the time being https://github.com/openshift/machine-config-operator/pull/1485 is good to unblock things.

ashcrow:
Needs a rebase.

yuqi-zhang:
I think this isn't really needed for us anymore.

PR #1488, opened 2020-02-19
bug 1804693: Revert "pkg/daemon: skip /etc/kubernetes/manifests/etcd-member.yaml"

deads2k:
Restores the file consistency check.

@runcom interesting side-node.  I'm not completely sure which operator reports the failure sometimes during upgrade.  If the MCO doesn't upgrade when degraded, you may not be able to pick this into 4.4 without breaking 4.3 to 4.4 upgrades.


kikisdeliveryservice:
procedural note: master is now 4.5 i guess so it would have to be backported to 4.4...:(

kikisdeliveryservice:
@deads2k do you still want this PR open? If so please update the BZ.

PR #1489, opened 2020-02-19
WIP: Move GCP routes service into gcp-routes-controller

LorbusChris:
[This script](https://github.com/LorbusChris/installer/blob/0fbddf1cd735a29282415aa0dae370bad9af5462/data/data/bootstrap/gcp/files/usr/local/bin/gcp-routes.sh) and its service are currently layered into RHCOS at build time. It controls the routes for the cluster on GCP.

This PR moves the functionality into the gcp-routes-controller container.

### DIFF
--- a/cmd/gcp-routes-controller/run.go
+++ b/cmd/gcp-routes-controller/run.go
@@ -283,6 +283,7 @@ func runSetRoutes() error {
 		for devName, fwips := range routes {
 			cmd := fmt.Sprintf("ip route show dev %s table local proto 66 | awk '{print$2}'", devName)
 			ipOutput, err := exec.Command("bash", "-c", cmd).Output()
+			glog.Info("RunSetRoutes CMD: " + cmd + "; Output: " + string(ipOutput))
 			if err != nil {
 				return fmt.Errorf("failed to show routes with command '%s': %v", cmd, err)
 			}
@@ -291,14 +292,16 @@ func runSetRoutes() error {
 
 				if !sliceContainsString(fwips, currentRoute) {
 					glog.Info("Removing stale forwarded IP " + currentRoute + "/32")
-					err := exec.Command("ip", "route", "del", currentRoute+"/32", "dev", devName, "table", "local", "proto", "66").Run()
+					op, err := exec.Command("ip", "route", "del", currentRoute+"/32", "dev", devName, "table", "local", "proto", "66").Output()
+					glog.Info("RunSetRoutes route del Output: " + string(op))
 					if err != nil {
 						return fmt.Errorf("failed to remove stale forwarded IP %s for device %s: %v", currentRoute, devName, err)
 					}
 				}
 			}
 			for _, fwip := range fwips {
-				err := exec.Command("ip", "route", "replace", "to", "local", fwip, "dev", devName, "proto", "66").Run()
+				op, err := exec.Command("ip", "route", "replace", "to", "local", fwip, "dev", devName, "proto", "66").Output()
+				glog.Info("RunSetRoutes route replace Output: " + string(op))
 				if err != nil {
 					return fmt.Errorf("failed to replace route to IP %s for device %s: %v", fwip, devName, err)
 				}
@@ -319,6 +322,7 @@ func RunDelRoutes() error {
 	for devName, fwips := range routes {
 		cmd := fmt.Sprintf("ip route show dev %s table local proto 66 | awk '{print$2}'", devName)
 		ipOutput, err := exec.Command("bash", "-c", cmd).Output()
+		glog.Info("RunDelRoutes CMD: " + cmd + "; Output: " + string(ipOutput))
 		if err != nil {
 			return fmt.Errorf("failed to show routes with command '%s': %v", cmd, err)
 		}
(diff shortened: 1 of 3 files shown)


LorbusChris:
Unsure this already works as-is

/retitle WIP: Move GCP routes service into gcp-routes-controller

runcom:
@sttts could you take a look at this?

LorbusChris:
I may have to move this into the MCD as a subcommand because installer calls the service directly at bootstrap: https://github.com/openshift/installer/blob/master/data/data/bootstrap/gcp/files/usr/local/bin/report-progress.sh#L16

LorbusChris:
actually, no. We can change the installer to run the gcp-routes-controller container with podman at bootstrap time

cgwalters:
/approve
Only skimmed the code but seems sane.  I am *definitely* in favor of less shell script.  
Risky though because we're replacing fragile manually tested code.

But at least we have a CI job that runs in GCP which should cover this.

sttts:
Is this new code or copied?

cmd/gcp-routes-controller/run.go
```diff
@@ -196,3 +196,142 @@ func (sl *healthTracker) OnComplete(state *health.State) {
...
+	ifs, err := getInstances(netPath)
+	if err != nil {
+		return nil, err
+	}
+	for _, vif := range ifs {
+		macs, err := getInstances(netPath + vif + "mac")
+		if err != nil {
+			return nil, err
+		}
+		hwAddr := macs[0]
```
sttts: is this safe? The fields come from a google URL. It can break and return whatever.

LorbusChris:
this is new code - it's a translation to go of the bash script mentioned in the first comment.

cmd/gcp-routes-controller/run.go
```diff
@@ -196,3 +196,142 @@ func (sl *healthTracker) OnComplete(state *health.State) {
 		}
 	}
 }
+
+func getRoutes() (map[string]string, error) {
+	routes := make(map[string]string)
+	const netPath = "network-interfaces/"
+	ifs, err := getInstances(netPath)
+	if err != nil {
+		return nil, err
```
sttts: some error context wouldn't help for all these returns.

cmd/gcp-routes-controller/run.go
```diff
@@ -196,3 +196,142 @@ func (sl *healthTracker) OnComplete(state *health.State) {
...
+	if err != nil {
+		return "", err
+	}
+	for _, dev := range devs {
+		if dev.HardwareAddr.String() == hwAddr {
+			return dev.Name, nil
+		}
+	}
+
+	return "", fmt.Errorf("Interface not found")
```
sttts: lower case errors in Go

cmd/gcp-routes-controller/run.go
LorbusChris: It's done with curl in the original script, see https://github.com/LorbusChris/installer/blob/0fbddf1cd735a29282415aa0dae370bad9af5462/data/data/bootstrap/gcp/files/usr/local/bin/gcp-routes.sh#L57

Suggestions to improve this are welcome ofc

LorbusChris:
@sttts thanks for the review! PR has been updated

LorbusChris:
Bootstrap doesn't seem to complete in `e2e-gcp-op` which may be very well be caused by this change...I'll try once again
/test e2e-gcp-op

LorbusChris:
I still need to tinker with this

/retitle WIP: Move GCP routes service into gcp-routes-controller

cmd/gcp-routes-controller/run.go
```diff
@@ -196,3 +196,152 @@ func (sl *healthTracker) OnComplete(state *health.State) {
...
+		if dev.HardwareAddr.String() == hwAddr {
+			return dev.Name, nil
+		}
+	}
+
+	return "", fmt.Errorf("interface not found")
+}
+
+func getInstances(instanceType string) ([]string, error) {
+	const baseURL = "http://metadata.google.internal/computeMetadata/v1/instance/"
```
juanluisvaladas: How is this resolved? If it's not getting it from /etc/resolv.conf, I see the google libraries, by default, query 169.254.169.254 instead of depending on name resolution: https://github.com/googleapis/google-cloud-go/blob/eed250f8ad01692794e90c5862a7be21b1de2c51/compute/metadata/metadata.go

I'm not quite sure if an scneario where resolution doesn't work due to coreDNS or someting, but I don't see downsides of setting the ip url to use 169.254.169.254 rather depeding on name resolution.

LorbusChris:
/close
MCO isn't the right place for this -- this should really be done by a component of the OS.

BZ for tracking: https://bugzilla.redhat.com/show_bug.cgi?id=1802534

Some more context on the new proposed location of this functionality, `nm-cloud-setup`: https://lists.fedoraproject.org/archives/list/cloud@lists.fedoraproject.org/thread/FSQR6KL4KA37WHTUAXL774SXSWIBSYGI/#FSQR6KL4KA37WHTUAXL774SXSWIBSYGI

PR #1491, opened 2020-02-19
DO NOT MERGE: Delete pods after drain with shutdown annotation

smarterclayton:
Allow pods to request deletion when MCD drains before a reboot to
ensure the pod is up to date after drain. This is testing a different
daemonset update pattern for pods that cannot be safely restarted
while workloads are on the machine (like OVS).

### DIFF
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -33,6 +33,12 @@ const (
 	// InitialNodeAnnotationsBakPath defines the path of InitialNodeAnnotationsFilePath when the initial bootstrap is done. We leave it around for debugging and reconciling.
 	InitialNodeAnnotationsBakPath = "/etc/machine-config-daemon/node-annotation.json.bak"
 
+	// OpenShiftDeleteOnNodeShutdownAnnotationKey is currently specific to OpenShift and will
+	// cause a pod (should be a daemonset) to be explicitly deleted before shutdown.  This
+	// helps ensure that availability is correctly tracked.  See
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1884053 for example.
+	OpenShiftDeleteOnNodeShutdownAnnotationKey = "upgrade.openshift.io/delete-on-node-shutdown"
+
 	// EtcPivotFile is used by the `pivot` command
 	// For more information, see https://github.com/openshift/pivot/pull/25/commits/c77788a35d7ee4058d1410e89e6c7937bca89f6c#diff-04c6e90faac2675aa89e2176d2eec7d8R44
 	EtcPivotFile = "/etc/pivot/image-pullspec"
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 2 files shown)


cgwalters:
One thing I'd like to see podman have more first-class support is for storing containers in `/run` - explicitly transient.  Having transient data stored in a persistent database causes issues around lifecycle binding (podman has problems if e.g. a system is forcibly powered off, kernel crashes etc.)  If we had such a thing, then crio could use it for containers with this annotation - then we wouldn't need to do this dance, we could just reboot and they'd be gone.

smarterclayton:
This is about marking the pod in the apiserver deleted, not the pod in the node.

cgwalters:
Oh right, the inverse problem.  I'm not sure I'm following fully though since won't the daemonset controller try to recreate it?  Is there something that will tell us to ignore that?

smarterclayton:
The behavior we are changing is having node shutdown be what explicitly purges the older daemonset version - the daemonset controller, if it's responsible, disrupts user workloads.

cgwalters:
It sounds like we basically punted on this problem and switched to OVS on the host, but we still want to (hopefully eventually) circle back to moving ovs to a container and using an approach like this?

cgwalters (approved):
Adding approval for the idea to get this off my review queue.

cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1884053#c10

cgwalters:
I rebased :surfer: this.

cgwalters:
So I guess the core problem here is basically: If we are going to do an OS update, then we want to minimize workload disruption by having daemonsets update at the same time.  But if we're *not* going to do an OS update, then we still need the daemonset controller to run, right?

cgwalters:
So after a lot of debate in https://github.com/kubernetes/kubernetes/pull/96129
I think the conclusion is that upstream Kubernetes should have this behavior by default.

cgwalters:
The other half of the design though (telling the daemonset controller not to update the pod at all) is a separate thing.

squeed:
@smarterclayton @cgwalters I don't immediately see how the upstream feature leads to this behavior. We would like the pod deleted, not just shut down gracefully.

We could really use this for ipsec (cc @markdgray).

squeed:
In other words, if we could get this merged, it would be really helpful :-)

cgwalters:
> I don't immediately see how the upstream feature leads to this behavior. We would like the pod deleted, not just shut down gracefully.

OK so to fully xref I came onto this PR while I was looking at https://bugzilla.redhat.com/show_bug.cgi?id=1884053#c12

There are conceptually two parts:

1. Ensuring pod in the apiserver is deleted during node shutdown
2. Avoiding a daemonset controller rescheduling to the node

Now if you follow this chain to https://github.com/kubernetes/kubernetes/pull/96129 that feature helps with a few things in this space, but I think the biggest is that it makes the node go `Ready=false` when the system begins shutdown.  That should help with 2 (though needs testing/validation).  I think we should try cherry picking and enabling that, maybe even for 4.7.

For 1. we could clearly change the MCD to delete the daemonset pods as part of drain as this patch is doing, but...I think it'd be better if we have this taken care of by kubelet and the node lifecycle controller.  Among other things, it would meant that logging into a node and typing `reboot` has the same behavior.

Now that PR was written primarily for the case of preemptible nodes where the node and VM is deleted.  We care about that case, but for us the "in place upgrade" case is much more prominent.  A very tricky thing here is that currently we keep daemonsets and static pods running while we go to do an OS update.  This keeps etcd and the apiserver running, which is good for availability.

But if we went to a model where instead we had the MCD re-exec itself as a systemd unit (like we do for initial pivot) to perform  
all update/upgrade changes, then we could entirely stop kubelet and let all the pods be deleted.  That would solve a bunch of problems (we wouldn't potentially be editing static pod configs while they were running, we wouldn't affect etcd performance during OS update), but the downside would be the update part wouldn't be "kube native" anymore - e.g. the logs would be in the journal and not the pod, etc.

PR #1494, opened 2020-02-21
Bug 1806001: [baremetal] Set hostname when DHCP6_FQDN_FQDN is set

bcrochet:
NetworkManager doesn't always set the hostname when DHCP6 is involved.
This sets the hostname when $DHCP6_FQDN_FQDN is populated and does not
equal localhost.localdomain. This is a workaround while we try to figure
out why NetworkManager is not setting the hostname as it should.


russellb:
@kikisdeliveryservice can you help review this?

cgwalters:
Seems OK to me short term but: Is there a tracker (Bugzilla?) for this?  If so would be great to link in the code and here. If not let's get one.

Because we will soon need to deal with this in the initramfs and this Ignition won't have run there yet.

PR #1496, opened 2020-02-22
Add machineConfigDaemonOSFCOS

LorbusChris:
Prep for supporting Fedora CoreOS as base OS.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -200,8 +200,8 @@ func New(
 		}
 	}
 
-	// Only pull the osImageURL from OSTree when we are on RHCOS
-	if operatingSystem == machineConfigDaemonOSRHCOS {
+	// Only pull the osImageURL from OSTree when we are on RHCOS or FCOS
+	if operatingSystem == machineConfigDaemonOSRHCOS || operatingSystem == machineConfigDaemonOSFCOS {
 		osImageURL, osVersion, err = nodeUpdaterClient.GetBootedOSImageURL()
 		if err != nil {
 			return nil, fmt.Errorf("error reading osImageURL from rpm-ostree: %v", err)
@@ -229,7 +229,7 @@ func New(
 		}
 	}
 
-	// report OS & version (if RHCOS) to prometheus
+	// report OS & version (if RHCOS or FCOS) to prometheus
 	HostOS.WithLabelValues(operatingSystem, osVersion).Set(1)
 
 	return &Daemon{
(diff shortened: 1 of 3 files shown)


pkg/daemon/update.go
```diff
@@ -601,13 +601,22 @@ func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
 	return cmdArgs
 }
 
+func isOSTreeBased(osName string) bool {
```
cgwalters: The canonical way to do this is to check for `/run/ostree-booted`.

LorbusChris: I've removed that func entirely for now and just expanded on the way it's been done so far. Do you think this is sufficient?

cgwalters: Not opposed to landing this patch as is, but I think it'd be a *lot* cleaner to add an `bool osIsCoreOS` to the daemon and conditionalize everything on that.

(`isOSTreeBased` is an intermediate level - for example it would also apply to e.g. RHEL Atomic Host, but that's not "CoreOS" which I think is now the pairing of (Ignition, OSTree) - we could still check that and there *might* be a place that would make sense to use that)

LorbusChris: I agree. Also @runcom for visibilty, this may be something for the MCO backlog? :)

pkg/daemon/daemon.go
```diff
@@ -1257,8 +1257,8 @@ func compareOSImageURL(current, desired string) (bool, error) {
 // Otherwise if `false` is returned, then we need to perform an update.
 func (dn *Daemon) checkOS(osImageURL string) (bool, error) {
 	// Nothing to do if we're not on RHCOS
```
sinnykumari: Let's add "and FCOS"

pkg/daemon/daemon.go
```diff
@@ -201,7 +201,7 @@ func New(
 	}
 
 	// Only pull the osImageURL from OSTree when we are on RHCOS
```
sinnykumari: Let's add "or FCOS"

sinnykumari:
Minor non-blocking comments, otherwise LGTM

pkg/daemon/daemon.go
sinnykumari: Similar update at https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/daemon.go#L1255

LorbusChris:
@sinnykumari thanks! Updated accordingly :)

PR #1497, opened 2020-02-22
Allow tuning mitigation=auto kernel argument

LorbusChris:
This is required for Fedora CoreOS 31 (and greater) hosts

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -36,7 +36,8 @@ const (
 // TODO: fill out the whitelist
 // tuneableArgsWhitelist contains allowed keys for tunable arguments
 var tuneableArgsWhitelist = map[string]bool{
-	"nosmt": true,
+	"nosmt":            true,
+	"mitigations=auto": true,
 }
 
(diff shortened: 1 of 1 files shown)


sinnykumari:
won't this impact RHCOS pivot as I believe it is not using cgroupv2 yet?

yuqi-zhang:
Any reason this is against master and not the fcos branch?

LorbusChris:
This is just the whitelist of allowed args as far as I understand it. They wouldn't be set on RHCOS nodes. As we need to merge fcos and master branches at some point, I'm trying to get non-invasive things into master sooner than later. We could possibly conditionalize those just for `machineConfigDaemonOSFCOS` from https://github.com/openshift/machine-config-operator/pull/1496 if that is preferred.

LorbusChris:
OKD-on-FCOS needs this switch to disable cgroupsv2, we're not using it yet either :)

LorbusChris:
I've updated the PR to only add `mitigations=auto`.
PTAL @runcom @cgwalters @vrutkovs

LorbusChris:
Closing in favour of #1757 
/close

PR #1501, opened 2020-02-23
1790819: [baremetal] Add watcher sidecar container for Coredns static pod

yboaron:
When /etc/resolv.conf file is modified, the forward section of
Coredns Corefile should be updated.
This patch adds sidecar container (similar to HAProxy and Keepalived) that
checks resolv.conf file and updates Coredns Corefile.

This PR should be merged only after https://github.com/openshift/baremetal-runtimecfg/pull/49 is merged.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/baremetal-coredns.yaml
+++ b/templates/common/baremetal/files/baremetal-coredns.yaml
@@ -23,6 +23,9 @@ contents:
       - name: conf-dir
         hostPath:
           path: "/etc/coredns"
+      - name: nm-resolv
+        hostPath:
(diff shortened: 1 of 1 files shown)


sinnykumari:
/skip
/approve
cc  @celebdor PTAL

cybertron:
I don't think we wanted this to merge yet. Without https://github.com/openshift/baremetal-runtimecfg/pull/49 it's going to break baremetal deployments. :-/

I guess we could merge that and address the comments later.

PR #1505, opened 2020-02-25
CRC: compare the correct object

runcom:
Do not compare pointer and struct as that always yields false and forces
the controller to update the MC and spam the logs every ~20m.


runcom:
@umohnani8 ptal

umohnani8:
Should we backport this to 4.2 and 4.3?

runcom:
> Should we backport this to 4.2 and 4.3?

I don't believe that log was so spammy to require this tiny backport unless someone is really yelling - wdyt?

umohnani8:
Yup, agreed

PR #1507, opened 2020-02-26
[release-4.3] Bug 1807279: templates/master/00-master/azure: bump raft tunables

hexfusion:
The reason for this change is that currently, Azure disk performance does not allow for the throughput provided by default raft settings. Under heavier, disk I/O Azure will throttle disks causing massive spikes in I/O 600ms+. By reducing the throughput ETCD_HEARTBEAT_INTERVAL and also extending the timeout for etcd elections ETCD_ELECTION_TIMEOUT we can improve stability in slower disks by reducing leader elections.

### DIFF
--- a/templates/master/00-master/azure/files/etc-etcd-etcd-conf.yaml
+++ b/templates/master/00-master/azure/files/etc-etcd-etcd-conf.yaml
@@ -1,0 +1,18 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/etcd/etcd.conf"
+contents:
+  inline: |
+    #[member]
+    ETCD_SNAPSHOT_COUNT=100000
+    ETCD_HEARTBEAT_INTERVAL=500
+    ETCD_ELECTION_TIMEOUT=2500
+
+    #[storage]
+    ETCD_QUOTA_BACKEND_BYTES=7516192768
+
+    #[logging]
+    ETCD_DEBUG=false
+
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
Reading the BZ I'm a little confused by:
`Please ensure this work goes into 4.3.z.`
and
`This could be backported to 4.4.z but I doubt we will be able to backport to 4.3.z.`
Do we want this in 4.3?

Also we should probably start with a PR to master (4.5) and then backport for 4.4 and 4.3 no?

hexfusion:
> Reading the BZ I'm a little confused by:
> `Please ensure this work goes into 4.3.z.`
> and
> `This could be backported to 4.4.z but I doubt we will be able to backport to 4.3.z.`
> Do we want this in 4.3?
> 
> Also we should probably start with a PR to master (4.5) and then backport for 4.4 and 4.3 no?

yes but as we moved to the operator the source of this change is not MCO for 4.4, 4.5

yuqi-zhang:
Ah got it, the dependent bug https://bugzilla.redhat.com/show_bug.cgi?id=1806700 refers to a closed PR in this repo. I think you also need to fix that first and link to a etcd-operator PR

hexfusion:
cc @smarterclayton

hexfusion:
cc @runcom @kikisdeliveryservice 

please bless.

PR #1514, opened 2020-02-27
docs: Use ocp namespace while creating custom release payload

sinnykumari:


sinnykumari:
@LorbusChris @vrutkovs I haven't tried myself generating custom release payload for OKD. Do you use origin namespace?

vrutkovs:
>Do you use origin namespace?

Yes. It would work the same as previously.

Do we want to mention ocp namespace in HACKING document? Only Red Hat employees have access to that (`origin` is open to all), perhaps its worth keeping `origin` namespace instead?

sinnykumari:
I don't have strong opinion on this, but I prefer doc which is generic and can be used by anyone.

This document is to help folks with MCO . There are reference in the doc that points to OCP and not OKD like release page.

Asking with my limited OKD understanding:
By following this doc, do you think all steps align with hacking MCO on OKD?

vrutkovs:
>By following this doc, do you think all steps align with hacking MCO on OKD?

Yes, the only difference is `fcos` vs. `master` branch - and these would be merged eventually. I don't think HACKING.md needs to be updated to reflect this

sinnykumari:
Awesome, thanks for your input Vadim!
closing this PR

PR #1519, opened 2020-02-27
Bug 1802871: Update to github.com/mtrmac/gpgme v0.1.2

mtrmac:
**- What I did**

Performed by
```console
$ go get github.com/mtrmac/gpgme@v0.1.2 && make go-deps
```
in a `golang:1.12` container.

**- How to verify it**

N/A


kikisdeliveryservice:
This seems to be only updating and existing (but unused) dependency, so seems ok?

Deferring final review/approval to @ericavonb 

/approve

but also hold pending her review/approval..

/hold

PR #1521, opened 2020-02-28
Replace networkmanager scripts with podman wrappers

vrutkovs:
**- What I did**
Replaced python scripts for NetworkManager with podman wrappers

**- How to verify it**
Run vSphere / oVirt / Openstack IPI installs - workers should join cluster as expected

NM helper scripts are running from podman containers

See https://github.com/openshift/baremetal-runtimecfg/pull/46


LorbusChris:
Do we have OpenStack e2e here?
/test e2e-openstack

vrutkovs:
/hold

worker nodes didn't join :/

LorbusChris:
does the container maybe need to be privileged in this case?

vrutkovs:
unprivildged container should work too.

Its odd that workers didn't join as NotReady nodes - lets see if that is consistent

/retest

vrutkovs:
ffs, so `non-virtual-ip` script is python2 and `nodeip-finder` one is python3-only. As a result it crashes when UBI7's python2 is used

kikisdeliveryservice:
As MCO doesn't own these templates, we will ask that approval is given by the baremental/openstack folks as a requirement to merge.

/assign @Fedosin @celebdor @mandre @yboaron

LorbusChris:
This won't work because the UBI7 containers don't have Python 3 included, as noted above by @vrutkovs. We also can't switch over to UBI8 for the baremetal-runtimecfg container base because that won't run on RHEL7, which is still supported as worker nodes in OCP.

If I'm not mistaken @celebdor is working on translating the scripts to Golang, which would we be the preferred solution here anyway.

This is also tracked in https://github.com/openshift/machine-config-operator/issues/1430

I believe this PR can thus be closed?

LorbusChris:
I don't see this as applicable to master at this time.
/close

PR #1525, opened 2020-03-02
Bug 1809372: bump openshift/api to remove IPv6DualStack feature flag

danwinship:
There are bugs with the IPv6DualStack feature flag in 1.17/4.4 even in single-stack clusters. https://bugzilla.redhat.com/show_bug.cgi?id=1794376 removed it from openshift/api and updated cluster-kube-apiserver-operator but didn't update machine-config-operator so we were still setting that flag for kubelet.


kikisdeliveryservice:
Unit tests seem to be failing on kubelet tests... part of this bump is breaking things

danwinship:
> Unit tests seem to be failing on kubelet tests... part of this bump is breaking things

Fixed; I had to update the templates to match the new expected feature flag set

PR #1526, opened 2020-03-02
Bug 1809373: [4.4] bump openshift/api to remove IPv6DualStack feature flag

danwinship:
Backport of #1525

There are bugs with the IPv6DualStack feature flag in 1.17/4.4 even in single-stack clusters. https://bugzilla.redhat.com/show_bug.cgi?id=1794376 removed it from openshift/api and updated cluster-kube-apiserver-operator but didn't update machine-config-operator so we were still setting that flag for kubelet.


kikisdeliveryservice:
@danwinship you need to make a 4.4 BZ dupe of https://bugzilla.redhat.com/show_bug.cgi?id=1809372 and update this PR so it can be backported

PR #1528, opened 2020-03-03
Bug 1809373: kubelet: bump openshift api to remove IPv6DualStack

rphillips:
**- What I did**
Bumps openshift api to remove the IPv6DualStack flag

**- How to verify it**


kikisdeliveryservice:
I think the BZ dependencies are off?

We already have a fix into master via https://bugzilla.redhat.com/show_bug.cgi?id=1809372 / #1525 

We need a 4.4 which this one shoudl be no?

kikisdeliveryservice:
Wait actually we already have: https://github.com/openshift/machine-config-operator/pull/1526 for 4.4 so now I'm confused?

PR #1529, opened 2020-03-04
Bug 1810333: Detect default hostname in its variants

celebdor:
On some environments, `hostname` returns localhost whereas on others it
returns `localhost.localdomain`. This commit ensures that we'll catch
both.

It also increases verbosity of initcontainers to ease debugging of
similar issues in the future.

### DIFF
--- a/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
+++ b/templates/common/baremetal/files/baremetal-mdns-publisher.yaml
@@ -27,15 +27,14 @@ contents:
       - name: verify-hostname
         image: {{ .Images.baremetalRuntimeCfgImage }}
         env:
-          - name: DEFAULT_LOCAL_HOSTNAME
-            value: "localhost"
           - name: RUNTIMECFG_HOSTNAME_PATH
             value: "/etc/mdns/hostname"
         command:
         - "/bin/bash"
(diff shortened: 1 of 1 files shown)


templates/common/baremetal/files/baremetal-mdns-publisher.yaml
```diff
@@ -45,9 +44,9 @@ contents:
               hostname
             fi
           }
-          while [ "$(get_hostname)" == "$DEFAULT_LOCAL_HOSTNAME" ]
+          while [ "$(get_hostname)" =~ ^localhost(.localdomain)?$ ]
```
cybertron: I think you're going to need [[ ]] around this because regexes are a bash thing. If I try this code with single brackets I get: syntax error near unexpected token `('.

However, with double brackets it seems to work and does the right thing.

celebdor: Right. Thanks!

celebdor: This happens to me for testing separately and not paying enough attention on copy paste :-1:

celebdor:
/retitle Bug 1810333: Detect default hostname in its variants

PR #1530, opened 2020-03-04
Bug 1810333: daemon: Always create tempfiles in target dir

cgwalters:
When we go to write a file, we need to create the temporary
file in the exact target directory, not (potentially) `/tmp`. This will
ensure that the right SELinux label is used by default.

Currently the `renameio` library's logic tries to optimize things
by using `/tmp` if possible, otherwise the target directory.
And without SELinux that's a sane optimization.  But we
can't do it.

Force using the target directory by passing it explicitly.

Should fix a bug seen with the baremetal config which
ended up with a `tmp_t` labeled file in `/etc`.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -57,10 +57,11 @@ func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
 // writeFileAtomically uses the renameio package to provide atomic file writing, we can't use renameio.WriteFile
 // directly since we need to 1) Chown 2) go through a buffer since files provided can be big
 func writeFileAtomically(fpath string, b []byte, dirMode, fileMode os.FileMode, uid, gid int) error {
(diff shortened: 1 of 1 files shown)


cgwalters:
Only compile tested.

celebdor:
The change to create in the same dir looks good. Didn't test it though

cgwalters:
Wait, sorry I just read the `renameio` sources more carefully, it does look like it's creating the tempfile in the target dir.  So the bug may be something else.

cgwalters:
OK pushed a much simpler patch that doesn't pull another dep (we should probably teach `renameio` about `O_TMPFILE` instead for that bit).

jlebon:
Wow, that renameio logic is... interesting. Might be worth a comment about the SELinux context here, but either way:

/approve

cgwalters:
flake
/test e2e-gcp-upgrade

celebdor:
/retitle Bug 1810333: daemon: Always create tempfiles in target dir

PR #1532, opened 2020-03-04
Update KubeletConfigDesign.md

madorn:
add a few more instructions so users can avoid the `Error: could not find any MachineConfigPool set for KubeletConfig` error.

### DIFF
--- a/docs/KubeletConfigDesign.md
+++ b/docs/KubeletConfigDesign.md
@@ -61,24 +61,9 @@ KubeletConfig
   [KubeletConfigurationSpec](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L45)
 ```
 
-Example:
+## How-To
 
-```
-apiVersion: machineconfiguration.openshift.io/v1
-kind: KubeletConfig
-metadata:
-  name: set-max-pods
-spec:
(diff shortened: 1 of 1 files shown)


madorn:
@kikisdeliveryservice hi - I agree with you but the problem is that we've referred some folks to these `ContainerRuntimeConfig` and `KubeletConfig` docs and they have issues because they end up creating the CR first and then going to update the machineconfigpool with the label - this does not work - it appears that MCO is not reconciling the update event. 

It only works if the CR is referring to a label that currently exists on the machineconfigpool.

thoughts?

kikisdeliveryservice:
> @kikisdeliveryservice hi - I agree with you but the problem is that we've referred some folks to these `ContainerRuntimeConfig` and `KubeletConfig` docs and they have issues because they end up creating the CR first and then going to update the machineconfigpool with the label - this does not work - it appears that MCO is not reconciling the update event.
> 
> It only works if the CR is referring to a label that currently exists on the machineconfigpool.
> 
> thoughts?

hmm weird i swear i've done it in that order before, let me try. 

if that's the case i think we can come up with some clear instructions... will report back!

kikisdeliveryservice:
The order is correct but i believe the instructions provided are not complete.

Let's work out of https://github.com/openshift/machine-config-operator/pull/1539 

I verified that it does work for ctrcfg and documented what I did there.

PR #1540, opened 2020-03-06
Bug 1806027: Specify cgroups in kubelet.conf so cAdvisor stats will be tracked

joelsmith:
Specifying `runtimeCgroups` will cause the Kubelet to monitor statistics for the specified cgroup so that we have statistics for crio.

Specifying `systemCgroups` will cause the Kubelet to monitor
statistics for system processes and make that data available
in our metrics pipeline.

Specifying `cgroupRoot` explicitly is required when specifying
`systemCgroups`

**- What I did**

Added cgroup parameters so that the eventual
`/etc/kubernetes/kubelet.conf` will be configured for the kubelet
to monitor cgroup statistics for the crio runtime and other system
processes.

**- How to verify it**
```
oc get --raw /api/v1/nodes/node-name-of-one-of-your-nodes/proxy/metrics/cadvisor | grep crio\\.service
```

Before this change, no metrics will be returned, and after, you should see several metrics returned.
Additionally, try with services like `sshd` instead of `crio` and you should also see those.

Kubelet will monitor statistics for system processes and crio

### DIFF
--- a/templates/master/01-master-kubelet/_base/files/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/files/kubelet.yaml
@@ -11,6 +11,7 @@ contents:
       anonymous:
         enabled: false
     cgroupDriver: systemd
+    cgroupRoot: /
     clusterDNS:
       - {{.ClusterDNSIP}}
     clusterDomain: cluster.local
@@ -19,8 +20,10 @@ contents:
     kubeAPIQPS: 50
     kubeAPIBurst: 100
     rotateCertificates: true
+    runtimeCgroups: /system.slice/crio.service
     serializeImagePulls: false
     staticPodPath: /etc/kubernetes/manifests
+    systemCgroups: /system.slice
     systemReserved:
       cpu: 500m
       memory: 1Gi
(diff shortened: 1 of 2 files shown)


smarterclayton:
Yay, lgtm.

Can we verify that the metrics are pulled before we merge though?  With promecius or similar?

This is definite backport to 4.3 after we confirm that it works

rphillips:
@smarterclayton yeah, I'll give it a test.

rphillips:
Metrics are showing up... Nice!

/lgtm
/approve

joelsmith:
This is also gonna need a 4.4 cherry-pick for https://bugzilla.redhat.com/show_bug.cgi?id=1806027
/cherrypick release-4.4

smarterclayton:
This didn't seem to fix anything.  I don't see anything except / and /system.slice/kubelet.service from https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.5/485/artifacts/e2e-gcp/metrics/prometheus.tar which should have this change.

joelsmith:
@smarterclayton either you're looking at a version that doesn't have the fix, or we're testing different things. I don't have any idea what to do with that tarball you linked to above, but on a 4.5 cluster that I just spun up, from the console's metrics tab, I can run the following query and see data.

```
rate(container_cpu_usage_seconds_total{container="",id="/system.slice/crio.service"}[1m])
```

smarterclayton:
It took over a day to get it in master, due to a race condition in builds, which steve and I will track.  Hold cancelled.  I manually verified the data I expected is there now.

PR #1524, opened 2020-03-02
Bug 1809274: crd/kubelet: do not prune kubelet rawExtension fields

yuqi-zhang:
The kubelet config is now validated, but fields are being silently
dropped since its an "unknown field" (rawExtension), which by
default is pruned by the validator.

Signed-off-by: Yu Qi Zhang <jerzhang@redhat.com>

PR #1541, opened 2020-03-06
Bug 1811211: remove validation of kubeletcfg which is breaking our kubeletcfg

kikisdeliveryservice:
Recent changes to kubeletcfg manifests seem to have broken our kubeletcfgs.  This is on a cluster using master.

When I create a new kubelet config from https://github.com/openshift/machine-config-operator/blob/master/docs/KubeletConfigDesign.md#spec 
```
kind: KubeletConfig
metadata:
  name: set-max-pods
spec:
  machineConfigPoolSelector:
    matchLabels:
      custom-kubelet: small-pods
  kubeletConfig:
    maxPods: 100
```

I see errors:
```$ oc apply -f maxpods.yaml
The KubeletConfig "set-max-pods" is invalid: 
* spec.kubeletConfig.apiVersion: Required value: must not be empty
* spec.kubeletConfig.kind: Required value: must not be empty
```

It seems to be validating against something different than our kubeletconfig type (upstream maybe?)

Thoughts?

### DIFF
--- a/manifests/kubeletconfig.crd.yaml
+++ b/manifests/kubeletconfig.crd.yaml
@@ -42,7 +42,6 @@ spec:
           properties:
             kubeletConfig:
               type: object
-              x-kubernetes-embedded-resource: true
               x-kubernetes-preserve-unknown-fields: true
             machineConfigPoolSelector:
               description: A label selector is a label query over a set of resources.
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -1161,7 +1161,6 @@ spec:
           properties:
             kubeletConfig:
               type: object
-              x-kubernetes-embedded-resource: true
               x-kubernetes-preserve-unknown-fields: true
             machineConfigPoolSelector:
               description: A label selector is a label query over a set of resources.


yuqi-zhang:
So this is a result of https://github.com/openshift/machine-config-operator/pull/1524. Since `kubeletConfig` (the object under spec not the overall defined custom object `KubeletConfig`) is a rawExtension object, upstream API requires it also to have a `apiVersion` and `kind`. So something like:
```
kind: KubeletConfig
metadata:
  name: set-max-pods
spec:
  machineConfigPoolSelector:
    matchLabels:
      custom-kubelet: small-pods
  kubeletConfig:
    apiVersion: kubelet.config.k8s.io/v1
    kind: kubeletConfig
    maxPods: 100
```
I believe.

I am ok with us removing this strict check. What you're doing should work fine (doesn't prune unknown fields but also doesn't check for rawExt api/kind). Also ping @rphillips since he would be more knowledgeable on what the correct definition should look like.

kikisdeliveryservice:
> So this is a result of #1524. Since `kubeletConfig` (the object under spec not the overall defined custom object `KubeletConfig`) is a rawExtension object, upstream API requires it also to have a `apiVersion` and `kind`. So something like:
> 
> ```
> apiVersion: machineconfiguration.openshift.io/v1
> kind: KubeletConfig
> metadata:
>   name: set-max-pods
> spec:
>   machineConfigPoolSelector:
>     matchLabels:
>       custom-kubelet: small-pods
>   kubeletConfig:
>     apiVersion: kubelet.config.k8s.io/v1
>     kind: kubeletConfig
>     maxPods: 100
> ```
> 
> I believe.
> 
> I am ok with us removing this strict check. What you're doing should work fine (doesn't prune unknown fields but also doesn't check for rawExt api/kind). Also ping @rphillips since he would be more knowledgeable on what the correct definition should look like.

Yes unfortunately that strict check changes our kubeletcfg spec, which is not good. I'm testing now to make sure that this change now works as expected. I spoke to @rphillips and we are testing currently.

kikisdeliveryservice:
So testing my fix, it works again:
`$ vi maxpods.yaml 

[kg@demo machine-config-operator]$ oc edit machineconfigpool worker
machineconfigpool.machineconfiguration.openshift.io/worker edited

[kg@demo machine-config-operator]$ oc apply -f maxpods.yaml
kubeletconfig.machineconfiguration.openshift.io/set-max-pods created

[kg@demo machine-config-operator]$ oc get kubeletconfigs
NAME           AGE
set-max-pods   17s
`

kikisdeliveryservice:
fix is in flight for our ci so this test will keep failing until we confirm it hit the build.

PR #1543, opened 2020-03-06
Update to Ignition 0.35.0 / Spec 2.4

LorbusChris:
Supports Ignition Config Spec <= v2.4

/hold


kikisdeliveryservice:
Why are we updating to 2.4?

LorbusChris:
HTTPHeaders were added to the spec in 2.4 and Installer already uses that feature in the provisioning of OpenStack nodes, via a shim (https://github.com/openshift/installer/blob/a135734617d76039b9494260975001b58a23c14b/pkg/tfvars/openstack/bootstrap_ignition.go#L9).

Everything still works right now with the Installer using the usual v2.2 for other platforms, but as a prereq for merging fcos and installer branches over there, all functions for generating Ignition configs are gathered and pulled into one location (and later another for all things spec3), so harmonization of spec versions across repos becomes a requirement, see https://github.com/openshift/installer/pull/3248

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -94,7 +95,11 @@ func createNewIgnition(configs map[string][]byte) igntypes.Config {
 }
 
 func findStorageConfig(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
```
mtrmac: These four copies of essentially the same code keep getting larger and more unwieldy (and _only_ two contain the `c := c` copy statement).  Consider replacing it with a `findConfigFile(mc, path)`?

pkg/controller/kubelet-config/helpers.go
```diff
@@ -49,7 +50,11 @@ func createNewDefaultFeatureGate() *osev1.FeatureGate {
 }
 
 func findKubeletConfig(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
-	for _, c := range mc.Spec.Config.Storage.Files {
+	ignCfg, report, err := ign.Parse(mc.Spec.Config.Raw)
+	if err != nil {
+		return nil, fmt.Errorf("parsing Ignition config failed with error: %v\nReport: %v", err, report)
+	}
+	for _, c := range ignCfg.Storage.Files {
```
mtrmac: Another instance of the `findConfigFile(mc, path)` code.

pkg/daemon/update_test.go
```diff
@@ -64,104 +66,184 @@ func TestUpdateOS(t *testing.T) {
...
-			Config: ctrlcommon.NewIgnConfig(),
+			Config: runtime.RawExtension{
+				Raw: helpers.MarshalOrDie(oldIgnCfg),
+			},
 		},
 	}
+	newIgnCfg := ctrlcommon.NewIgnConfig()
+
+	// Set improper version
+	newIgnCfg.Ignition.Version = "3.0.0"
```
mtrmac: ("999.0.0" or something clearly too high?)

pkg/daemon/update_test.go
```diff
@@ -64,104 +66,184 @@ func TestUpdateOS(t *testing.T) {
...
 
 	// Verify Disk changes react as expected
-	oldConfig.Spec.Config.Storage.Disks = []igntypes.Disk{
+	oldIgnCfg.Storage.Disks = []igntypes.Disk{
 		igntypes.Disk{
 			Device: "/one",
 		},
 	}
-
+	oldConfig = &mcfgv1.MachineConfig{
```
mtrmac: This constructor is repeated so often. Make it a shared helper?

Or maybe even a `runReconcilable(oldIgn, newIgn)` that hides the `MachineConfig` intermediaries entirely.

pkg/daemon/update_test.go
```diff
@@ -170,38 +252,50 @@ func TestReconcilable(t *testing.T) {
...
 		ObjectMeta: metav1.ObjectMeta{Name: "newconfig"},
 		Spec: mcfgv1.MachineConfigSpec{
-			Config: ctrlcommon.NewIgnConfig(),
+			Config: runtime.RawExtension{
+				Raw: helpers.MarshalOrDie(newIgnCfg),
+			},
 		},
 	}
-	diff := NewMachineConfigDiff(oldConfig, newConfig)
+	diff, _ := NewMachineConfigDiff(oldConfig, newConfig)
```
mtrmac: Could the error return values be tested as well?

pkg/server/server.go
```diff
@@ -59,15 +61,23 @@ func getAppenders(currMachineConfig string, f kubeconfigFunc, osimageurl string)
...
+	}
+	tmpcfg.Spec.Config.Raw = rawTmpIgnCfg
 	serialized, err := json.Marshal(tmpcfg)
 	if err != nil {
 		panic(err.Error())
 	}
-	appendFileToIgnition(&mccfg.Spec.Config, daemonconsts.MachineConfigEncapsulatedPath, string(serialized))
+	err = appendFileToRawIgnition(&mccfg.Spec.Config, daemonconsts.MachineConfigEncapsulatedPath, string(serialized))
+	if err != nil {
+		panic(err.Error())
```
mtrmac: Why not add an error return instead? Both callers can return errors.

pkg/controller/template/render.go
```diff
@@ -268,15 +271,19 @@ func generateMachineConfigForName(config *RenderConfig, role, name, templateDir,
...
 	// And inject the osimageurl here
 	mcfg.Spec.OSImageURL = config.OSImageURL
 
 	return mcfg, nil
 }
 
-// MachineConfigFromIgnConfig creates a MachineConfig with the provided Ignition config
-func MachineConfigFromIgnConfig(role, name string, ignCfg *igntypes.Config) *mcfgv1.MachineConfig {
+// MachineConfigFromRawIgnConfig creates a MachineConfig with the provided Ignition config
+func MachineConfigFromRawIgnConfig(role, name string, rawIgnCfg []byte) *mcfgv1.MachineConfig {
```
mtrmac: Why not move the `json.Marshal(ignCfg)` call inside this function? Every single caller does that immediately before, and it would be more type-safe.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -546,12 +546,22 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+
+			internalCtrRuntimeConfigIgn := createNewIgnition(map[string][]byte{
 				storageConfigPath: storageTOML,
 				crioConfigPath:    crioTOML,
 			})
+			rawCtrRuntimeConfigIgn, err := json.Marshal(internalCtrRuntimeConfigIgn)
+			if err != nil {
+				glog.V(2).Infoln(internalCtrRuntimeConfigIgn, err, "error marshalling container runtime config Ignition: %v", err)
+			}
+			mc.Spec.Config.Raw = rawCtrRuntimeConfigIgn
```
mtrmac: The “update mc.Spec.Config” operation might also be better with a helper (although saving one line and one variable, it’s less clear cut).

pkg/controller/template/render.go
LorbusChris: This function has to work with Ignition Config Spec 2 as well as 3 (meaning in input cannot be a typed spec 2 Ignition config).

It could probably take any interface as arg and then marshal that internally, but that wouldn't make it any more type safe.

pkg/controller/container-runtime-config/helpers.go
LorbusChris: That is an optimization unrelated to the RawExt conversion and can be done separately if needed.

pkg/daemon/update_test.go
LorbusChris: 3.0.0 is a version that actually exists and that is clearly not yet supported by MCO

pkg/controller/template/render.go
mtrmac: AFAICS there are 5 callers that all start with a `ctrlcommon.NewIgnConfig()` (one of them updating the value), and one caller right above that comes from the updated `transpileToIgn`. All of them are
```go
*igntypes.Config // igntypes "github.com/coreos/ignition/config/v2_4/types"
```
aren’t they?

(And even if it weren’t any more type-safe, the 6 repeated `json.Marshal` calls could be avoided, and if this function _wanted_ to do some kind of sanity check on the contents, it would probably be cheaper on a Go value than on a `[]byte`.)

pkg/daemon/update_test.go
mtrmac: In other words, as soon as MCO will add support for it, this test will break. Or is that the purpose of this test?

pkg/controller/template/render.go
LorbusChris: The RawExt conversion is a prerequisite for supporting both v2 and v3 Ignition configs.
I've changed the function to take `interface{}` as arg and do the marshalling within.
Please refer to #996 (and specifically https://github.com/openshift/machine-config-operator/pull/996/files#diff-7bee216e56f15d868e9558dedc3fda0eR283) for all things related to the RawExt conversion.

pkg/daemon/update_test.go
LorbusChris: This test will indeed have to change once support is added.
However, right now it is important to check v3 is not accepted.

kikisdeliveryservice: agree with @LorbusChris this should stay  as-is

LorbusChris:
This probably won't be needed
/close

PR #1547, opened 2020-03-09
Bug 1811758 :  baremetal: Add missing network-online dependency to kubelet unit

hardys:
This was added in #1206 to the base unit, but not the baremetal
copy of this file

**- What I did**
Added a dependency on the network-online target for the baremetal copy of the kubelet unit file.

**- How to verify it**
Verify the unit file on a IPI baremetal deployment or via `systemctl list-dependencies`

Added missing dependency for kubelet on the network-online target, for the baremetal platform.


hardys:
/retitle Bug 1811758 :  baremetal: Add missing network-online dependency to kubelet unit

cgwalters:
Arrrgh!  We need to dedup this stuff.

PR #1551, opened 2020-03-10
Bug 1811898: cmd/machine-config-operator: log RELEASE_VERSION when starting up

runcom:
Other subcomponents are still following the old version logging when Raw and Hash
are sufficient for us to debug. This patch just changes the MCO itself to print
RELEASE_VERSION at startup. We might want to revisit this for all components and
bootstrap commands later but not critical either.

### DIFF
--- a/cmd/machine-config-operator/start.go
+++ b/cmd/machine-config-operator/start.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"flag"
+	"os"
 
 	"github.com/golang/glog"
 	operatorclientset "github.com/openshift/client-go/operator/clientset/versioned"
@@ -42,7 +43,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	flag.Parse()
 
 	// To help debugging, immediately log version
-	glog.Infof("Version: %+v (%s)", version.Raw, version.Hash)
+	glog.Infof("Version: %s (Raw: %s, Hash: %s)", os.Getenv("RELEASE_VERSION"), version.Raw, version.Hash)
 
 	if startOpts.imagesFile == "" {
 		glog.Fatal("--images-json cannot be empty")


cmd/machine-config-operator/start.go
```diff
@@ -42,7 +43,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	flag.Parse()
 
 	// To help debugging, immediately log version
-	glog.Infof("Version: %+v (%s)", version.Raw, version.Hash)
+	glog.Infof("Version: %s (Raw: %s, Hash: %s)", os.Getenv("RELEASE_VERSION"), version.Raw, version.Hash)
```
yuqi-zhang: So, if I understand this correctly `os.Getenv("RELEASE_VERSION")` is just getting an environment variable. In this case nothing is setting `RELEASE_VERSION` no? I logged in to a node and can see that its empty by default. Wouldn't this just be empty?

runcom:
> So, if I understand this correctly os.Getenv("RELEASE_VERSION") is just getting an environment variable. In this case nothing is setting RELEASE_VERSION no? I logged in to a node and can see that its empty by default. Wouldn't this just be empty?

that env is injected into the MCO pod (not whole node) - you can look at its injection in the MCO manifest itself. When the MCO starts it reads the env and prints it, it's also later used in the code. The version is set by the release payload builder.

Look here https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1551/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1497/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-operator-5cc998d49d-hztvc_machine-config-operator.log

```
I0310 07:23:13.168998       1 start.go:46] Version: 0.0.1-2020-03-10-070742 (Raw: machine-config-daemon-4.5.0-202003100116-2-g3ac4b8ff-dirty, Hash: 3ac4b8ff765a46930ef210cf977531371a803819)
```

it now prints the version (and it does so at bootstrap as well).

/retest

PR #1554, opened 2020-03-10
Bug 1812179: manifests/controllerconfig: make infra nullable

yuqi-zhang:
Infra does not seem mandatory for every platform, and can be null
for some existing clusters. Thus, make it able to be a null object,
otherwise it will fail validation.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1812179

### DIFF
--- a/manifests/controllerconfig.crd.yaml
+++ b/manifests/controllerconfig.crd.yaml
@@ -336,6 +336,7 @@ spec:
                                 as a static pod to serve those hostnames to the nodes
                                 in the cluster.
                               type: string
+              nullable: true
             kubeAPIServerServingCAData:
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
@yuqi-zhang to clarify this is a side affect of the validation turned on before?

if this causes upgrades to fail, wondering if release upgrade tests failed or caught it?

yuqi-zhang:
> to clarify this is a side affect of the validation turned on before?

Yes. API validation unfortunately doesn't handle null objects unless explicitly told to do so

> if this causes upgrades to fail, wondering if release upgrade tests failed or caught it?

No. I think in the vast majority of platforms and scenarios that shouldn't be null. The bug was reported against a long running AWS cluster, which may have done something special

PR #1556, opened 2020-03-11
etcd: remove etcd entries from mdns

mandre:
Similar to https://github.com/openshift/installer/pull/3265, this
removes the etcd entries from mdns config because etcd no longer uses
DNS.

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-mdns-config.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-mdns-config.yaml
@@ -7,15 +7,6 @@ contents:
     collision_avoidance = "hostname"
 
     service {
-        name = "{{`{{ .Cluster.Name }}`}} Etcd"
-        host_name = "{{`{{ .ShortHostname }}`}}.local."
-        type = "_etcd-server-ssl._tcp"
-        domain = "local."
-        port = 2380
-        ttl = 3200
-    }
-
-    service {
         name = "{{`{{ .Cluster.Name }}`}} Workstation"
         host_name = "{{`{{ .ShortHostname }}`}}.local."
         type = "_workstation._tcp"
--- a/templates/master/00-master/openstack/files/openstack-mdns-config.yaml
+++ b/templates/master/00-master/openstack/files/openstack-mdns-config.yaml
@@ -6,26 +6,10 @@ contents:
     bind_address = "{{`{{ .NonVirtualIP }}`}}"
     collision_avoidance = "hostname"
     service {
-        name = "{{`{{ .Cluster.Name }}`}} Etcd"
-        host_name = "{{`{{ .EtcdShortHostname }}`}}.local."
-        type = "_etcd-server-ssl._tcp"
-        domain = "local."
-        port = 2380
-        ttl = 3200
-    }
-    service {
         name = "{{`{{ .Cluster.Name }}`}} Workstation"
         host_name = "{{`{{ .ShortHostname }}`}}.local."
         type = "_workstation._tcp"
         domain = "local."
         port = 42424
         ttl = 3200
-    }
-    service {
-        name = "{{`{{ .Cluster.Name }}`}} EtcdWorkstation"
-        host_name = "{{`{{ .EtcdShortHostname }}`}}.local."
-        type = "_workstation._tcp"
-        domain = "local."
-        port = 42424
-        ttl = 300
     }
--- a/templates/master/00-master/ovirt/files/ovirt-mdns-config.yaml
+++ b/templates/master/00-master/ovirt/files/ovirt-mdns-config.yaml
@@ -7,28 +7,10 @@ contents:
     collision_avoidance = "hostname"
 
     service {
-        name = "{{`{{ .Cluster.Name }}`}} Etcd"
-        host_name = "{{`{{ .EtcdShortHostname }}`}}.local."
-        type = "_etcd-server-ssl._tcp"
-        domain = "local."
-        port = 2380
-        ttl = 3200
-    }
-
-    service {
         name = "{{`{{ .Cluster.Name }}`}} Workstation"
         host_name = "{{`{{ .ShortHostname }}`}}.local."
         type = "_workstation._tcp"
         domain = "local."
         port = 42424
         ttl = 3200
-    }
-
-    service {
-        name = "{{`{{ .Cluster.Name }}`}} EtcdWorkstation"
-        host_name = "{{`{{ .EtcdShortHostname }}`}}.local."
-        type = "_workstation._tcp"
-        domain = "local."
-        port = 42424
-        ttl = 300
     }
--- a/templates/master/00-master/vsphere/files/vsphere-mdns-config.yaml
+++ b/templates/master/00-master/vsphere/files/vsphere-mdns-config.yaml
@@ -11,15 +11,6 @@ contents:
     collision_avoidance = "hostname"
 
     service {
-        name = "{{`{{ .Cluster.Name }}`}} Etcd"
-        host_name = "{{`{{ .ShortHostname }}`}}.local."
-        type = "_etcd-server-ssl._tcp"
-        domain = "local."
-        port = 2380
-        ttl = 3200
-    }
-
-    service {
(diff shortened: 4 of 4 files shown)


templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
```diff
@@ -10,7 +10,7 @@ contents:
     }
 
     vrrp_script chk_dns {
-        script "/usr/bin/host -t SRV _etcd-server-ssl._tcp.{{ .EtcdDiscoveryDomain }} localhost"
+        script "/usr/bin/host api-int.{{ .EtcdDiscoveryDomain }} localhost"
```
abhinavdahiya: I think we should stop using the etcd discovery domain all together...

mandre: I couldn't find where to get the base domain from. Any suggestion?

abhinavdahiya: I think the baremetal-runtime-cfg has a way to provide the cluster domain at runtime.. that might be an option?

deads2k:
@mandre I'd be very happy to see this and backported to 4.4.  It will reduce the risk of people accidentally depending on "naughty" things

templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
mandre: I can change it here to `{{ .Cluster.Domain }}` but many of the templates are not rendered via baremetal-runtimecfg, so I'd rather leave `{{ .EtcdDiscoveryDomain }}` alone until we decide to effectively get rid of it globally.

mandre:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

/cc rgolangh celebdor cybertron jcpowermac

Could you have a look? This PR touches the BM, openStack, ovirt and vsphere mDNS config.

templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
abhinavdahiya: > I can change it here to {{ .Cluster.Domain }} but many of the templates are not rendered via baremetal-runtimecfg, so I'd rather leave {{ .EtcdDiscoveryDomain }} alone until we decide to effectively get rid of it globally.

api-int.<etcd discovery domain> is plain wrong imo, it works because that's the equal today. so i don't see why we would continue to use it..?

cybertron:
I've tested this on baremetal, and while the deployment works, it appears to be breaking our loadbalancing. This is because we repurposed the SRV record for getting the haproxy backends, and if we remove it we end up with no backends and thus no load balancing.

I'm looking into whether there is another way we can handle that, but because baremetal-runtimecfg runs so early in the deployment I'm not sure whether we can use the same methods that etcd is to get the node addresses. If any other platforms are also using haproxy for loadbalancing they should probably verify that this doesn't result in an invalid haproxy config too.

templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
mandre: OK, I misunderstood your first comment - I'll make the change.

jcpowermac:
For vSphere IPI (starting with 4.5) we were following baremetal. So if baremetal works most likely vsphere will as well. The current test is for UPI only.

templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
cybertron: This is just checking that the DNS server is responding. It doesn't matter what record we look up. api-int happens to be a known name that should always be there.

abhinavdahiya: > This is just checking that the DNS server is responding. It doesn't matter what record we look up. api-int happens to be a known name that should always be there.

api-int.<cluster domain> is a known DNS name that should always be there, and not api-int.<etcd discovery domain>

cybertron: Ah, Github ate the domain part of your comment. I thought you were objecting to the use of api-int in general. :-)

rgolangh:
@celebdor oVirt should fail on the same thing.

mandre:
> I've tested this on baremetal, and while the deployment works, it appears to be breaking our loadbalancing. This is because we repurposed the SRV record for getting the haproxy backends, and if we remove it we end up with no backends and thus no load balancing.
> 
> I'm looking into whether there is another way we can handle that, but because baremetal-runtimecfg runs so early in the deployment I'm not sure whether we can use the same methods that etcd is to get the node addresses. If any other platforms are also using haproxy for loadbalancing they should probably verify that this doesn't result in an invalid haproxy config too.

Indeed, I hadn't notice the broken haproxy config. Putting this on hold for now since it breaks on all platforms.
/hold

mandre:
SRV records should no longer be a dependency for haproxy config thanks to https://github.com/openshift/baremetal-runtimecfg/pull/52.

cybertron:
I verified that this works on baremetal now so should be good to go from our end.

mandre:
I've also checked that it worked for OpenStack. I was waiting for https://github.com/openshift/machine-config-operator/pull/1666 to merge to avoid merge conflicts but could push the rebased patch so everyone can try with their platform.

rgolangh:
/lgtm ovirt

Thanks @mandre !

dougsland:
@mandre @rgolangh looks like we also need to update: **templates/worker/00-worker/ovirt/files/ovirt-non-virtual-ip.yaml:121** ?

I see:   api_vip, **dns_vip**, ingress_vip = sys.argv[1:4]

jcpowermac:
@patrickdillon

templates/master/00-master/openstack/files/openstack-keepalived-keepalived.yaml
```diff
@@ -9,7 +9,7 @@ contents:
         weight 50
     }
     vrrp_script chk_dns {
-        script "/usr/bin/host -t SRV _etcd-server-ssl._tcp.{{ .EtcdDiscoveryDomain }} localhost"
+        script "/usr/bin/host api-int.{{ .EtcdDiscoveryDomain }} localhost"
```
abhinavdahiya, line 12: should we use the https://github.com/openshift/api/blob/95abe2d2f4223d5931e418bf8e4d3773d16b42c0/config/v1/types_infrastructure.go#L74-L78

So that we are not assuming the `api-int` subdomain exists on `EtcdDiscoveryDomain`. Because that is technically not true or part of the API.

templates/master/00-master/ovirt/files/ovirt-keepalived-keepalived.yaml
```diff
@@ -10,7 +10,7 @@ contents:
     }
 
     vrrp_script chk_dns {
-        script "/usr/bin/host -t SRV _etcd-server-ssl._tcp.{{ .EtcdDiscoveryDomain }} localhost"
+        script "/usr/bin/host api-int.{{ .EtcdDiscoveryDomain }} localhost"
```
abhinavdahiya, line 13: same as https://github.com/openshift/machine-config-operator/pull/1556/files#r433445452

templates/master/00-master/openstack/files/openstack-keepalived-keepalived.yaml
mandre: This check no longer exists, cf https://github.com/openshift/machine-config-operator/pull/1666/

templates/master/00-master/ovirt/files/ovirt-keepalived-keepalived.yaml
mandre: Same here, this check was removed in https://github.com/openshift/machine-config-operator/pull/1775

sdodson:
/lgtm
@ericavonb PTAL, needs approved

PR #1559, opened 2020-03-12
DO NOT MERGE [fcos] Rebase on master

LorbusChris:
This is a PR to visibly track OKD-on-FCOS development

PR #1560, opened 2020-03-12
[fcos] Merge fcos-devel

LorbusChris:
This branch contains exactly the same code as the `LorbusChris:fcos-devel` branch
from https://github.com/openshift/machine-config-operator/pull/1559
but keeps history of the `fcos` branch intact so CI tests can run.


vrutkovs:
Masters didn't join the cluster :/

PR #1561, opened 2020-03-12
Fcos rebase merge debug

vrutkovs:
Debug version of #1560

/cc LorbusChris


vrutkovs:
masters didn't pivot, odd

vrutkovs:
Now fails with

```
I0312 22:50:23.897868       1 container_runtime_config_controller.go:368] Error syncing image config openshift-config: could not Create/Update MachineConfig: MachineConfig.machineconfiguration.openshift.io "99-master-48508239-6068-48ba-8b27-429f7020cc3f-registries" is invalid: spec.config.ignition.config.replace.source: Invalid value: "null": spec.config.ignition.config.replace.source in body must be of type string: "null"
```

LorbusChris:
/hold
just to make sure we have a cleaned up history with all required fixes in https://github.com/openshift/machine-config-operator/pull/1559 before this branch here (or #1560) is merged.

vrutkovs:
Interesting, so bootstrap's MCC finds `ssh` machineconfig with SSH keys, but it not present in machineconfig's list when the cluster boots

vrutkovs:
Currently we use `json.Marshal` to create Ignition. For configs without append/merge configs this would create
```
ignition:
  config:
    replace:
      source: null
```
due to source being a pointer to a string (thus "null") and Go would find an non-empty field and write the whole stanza. Unfortunately this produces an invalid Ignition config.

Last 3 commit work around that, 

BUT

since installer is not updated to use this branch yet SSH keys machineconfigs would create an MC for SSH key with this stanza and apparently there's still a piece of code where this stanza is not being removed

LorbusChris:
Thank you, @vrutkovs! I'm on it!

PR #1563, opened 2020-03-12
Bug 1812649: split kernel arguments

ericavonb:
**- What I did**
Split multiple kernel parameters in a single string, best matching what the kernel does. This
allows users to provide kernelArguments space delineated as a single string. For
instance, the following will be treated the same:

```yaml
  kernelArguments:
  - a=1 b=2
```

and

```yaml
  kernelArguments:
  - a=1
  - b=2
```
**- How to verify it**

1.
```shell
oc create -f - <<EOF
kind: MachineConfig
metadata:
  labels:
  name: 50-worker-custom
spec:
  kernelArguments:
  - a=1 b=2
EOF
```
2. Watch the node <node> reboot and become Ready.
3. Delete the custom MC
```shell
oc delete mc/50-worker-custom
```
4. Watch the node <node> reboot and become Ready

Split multiple kernel parameters in a single string in best effort attempt. This
allows users to provide kernelArguments space delineated as a single string.

Fixes: [Bug-1812649](https://bugzilla.redhat.com/show_bug.cgi?id=1812649)

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -605,27 +605,76 @@ func checkFIPS(current, desired *mcfgv1.MachineConfig) error {
 	return errors.New("detected change to FIPS flag. Refusing to modify FIPS on a running cluster")
 }
 
+// checks for white-space characters in "C" and "POSIX" locales.
+func isSpace(b byte) bool {
+	return b == ' ' || b == '\f' || b == '\n' || b == '\r' || b == '\t' || b == '\v'
+}
+
+// You can use " around spaces, but can't escape ". See next_arg() in kernel code /lib/cmdline.c
+// Gives the start and stop index for the next arg in the string, beyond the provided `begin` index
+func nextArg(args string, begin int) (int, int) {
+	var (
+		start, stop int
+		inQuote     bool
+	)
+	// Skip leading spaces
+	for start = begin; start < len(args) && isSpace(args[start]); start++ {
+	}
+	stop = start
+	for ; stop < len(args); stop++ {
+		if isSpace(args[stop]) && !inQuote {
+			break
+		}
+
+		if args[stop] == '"' {
+			inQuote = !inQuote
+		}
+	}
+
+	return start, stop
+}
+
+func splitKernelArguments(args string) []string {
+	var (
+		start, stop int
+		split       []string
+	)
+	for stop < len(args) {
+		start, stop = nextArg(args, stop)
+		if start != stop {
+			split = append(split, args[start:stop])
+		}
+	}
+	return split
+}
+
+// parseKernelArguments separates out kargs from each entry and returns it as a map for
+// easy comparison
+func parseKernelArguments(kargs []string) map[string]bool {
+	parsed := make(map[string]bool)
+	for _, k := range kargs {
+		for _, arg := range splitKernelArguments(k) {
+			parsed[strings.TrimSpace(arg)] = true
+		}
+	}
+	return parsed
+}
+
 // generateKargsCommand performs a diff between the old/new MC kernelArguments,
 // and generates the command line arguments suitable for `rpm-ostree kargs`.
 // Note what we really should be doing though is also looking at the *current*
 // kernel arguments in case there was drift.  But doing that requires us knowing
-// what the "base" arguments are.  See https://github.com/ostreedev/ostree/issues/479
+// what the "base" arguments are. See https://github.com/ostreedev/ostree/issues/479
 func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
-	oldKargs := make(map[string]bool)
-	for _, arg := range oldConfig.Spec.KernelArguments {
-		oldKargs[arg] = true
-	}
-	newKargs := make(map[string]bool)
-	for _, arg := range newConfig.Spec.KernelArguments {
-		newKargs[arg] = true
-	}
+	oldKargs := parseKernelArguments(oldConfig.Spec.KernelArguments)
+	newKargs := parseKernelArguments(newConfig.Spec.KernelArguments)
 	cmdArgs := []string{}
-	for _, arg := range oldConfig.Spec.KernelArguments {
+	for arg := range oldKargs {
 		if !newKargs[arg] {
 			cmdArgs = append(cmdArgs, "--delete="+arg)
 		}
 	}
-	for _, arg := range newConfig.Spec.KernelArguments {
+	for arg := range newKargs {
 		if !oldKargs[arg] {
 			cmdArgs = append(cmdArgs, "--append="+arg)
 		}
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -275,6 +275,16 @@ func TestKernelAguments(t *testing.T) {
 			additions: []string{"hello=world"},
 		},
 		{
+			args:      []string{"foo", "bar=1 hello=world"},
+			deletions: []string{"nosmt", "baz=test"},
+			additions: []string{"bar=1", "hello=world"},
+		},
+		{
+			args:      []string{"foo", " baz=test bar=\"hello world\""},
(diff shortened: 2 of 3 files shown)


pkg/daemon/update.go
```diff
@@ -574,27 +574,33 @@ func checkFIPS(current, desired *mcfgv1.MachineConfig) error {
 	return errors.New("detected change to FIPS flag. Refusing to modify FIPS on a running cluster")
 }
 
+// parseKernelArguments separates out kargs from each entry and returns it as a map for
+// easy comparision
+func parseKernelArguments(kargs []string) map[string]bool {
+	parsed := make(map[string]bool)
+	for _, k := range kargs {
+		for _, arg := range strings.Split(k, " ") {
```
jmencak: Isn't this too simplistic?  Is this going to work for quoted kernel parameters with spaces in them such as `param1="hello world!" param2=helllo world-param3`?  I'd suggest looking at kernel code what they do, in particular `next_arg()` function.  I believe something like this might do a better job:
```
// checks for white-space characters in "C" and "POSIX" locales.
func isspace(b byte) bool {
	return b == ' ' || b == '\f' || b == '\n' || b == '\r' || b == '\t' || b == '\v'
}

// You can use " around spaces, but can't escape ".  See next_arg() in kernel code.
func nextArg(args string, start int) (int, int) {
	var (
		i, equals int
		stop int
		inQuote bool
	)

	// Skip leading spaces
	for start < len(args) && isspace(args[start]) {
		start++
	}

	i = start

	if (start < len(args) && args[i] == '"') {
		i++
		inQuote = true
	}

	for ; i < len(args); i++ {
		if isspace(args[i]) && !inQuote {
			break
		}

		if equals == 0 && args[i] == '=' {
			equals = i
		}

		if args[i] == '"' {
			inQuote = !inQuote
		}
	}

	stop = i

	return start, stop
}

// SplitKernelArguments splits kernel parameters into a slice of strings one parameter per string.
func SplitKernelArguments(args string) ([]string) {
	var (
		start, stop int
		kv []string
	)

	for stop < len(args) {
		start, stop = nextArg(args, stop)
		if start != stop {
			kv = append(kv, args[start:stop])
		}
	}

	return kv
}
```

ericavonb: Yes, this is the naive first attempt. I'm looking into copying over the next_arg logic now.

cgwalters:
This is a messy topic...I am not hugely in favor of "automatic splitting" but I totally see where it's convenient for people.  And certainly a lot of places end up with "stringified lists" rather than an actual list; among those is the [BLS files](https://www.freedesktop.org/wiki/Specifications/BootLoaderSpec/) that ostree reads/writes so it does end up needing to be supported.

sinnykumari:
It seems this feature is in WIP, adding hold to avoid accidental merge. @ericavonb Feel free to remove hold whenever it is feature complete
/hold

runcom:
@ericavonb any update on this? do you think we make this into the release as it looks pretty important

ashcrow:
Any update?

pkg/daemon/update.go
ericavonb: Updated with your suggestion. Confirmed the logic matches the kernel code 👍

pkg/daemon/update.go
```diff
@@ -605,27 +605,76 @@ func checkFIPS(current, desired *mcfgv1.MachineConfig) error {
...
+		start, stop = nextArg(args, stop)
+		if start != stop {
+			split = append(split, args[start:stop])
+		}
+	}
+	return split
+}
+
+// parseKernelArguments separates out kargs from each entry and returns it as a map for
+// easy comparision
```
sinnykumari, line 652: verify test complaining about: s/comparision/comparison/

pkg/daemon/update.go
```diff
@@ -605,27 +605,76 @@ func checkFIPS(current, desired *mcfgv1.MachineConfig) error {
 	return errors.New("detected change to FIPS flag. Refusing to modify FIPS on a running cluster")
 }
 
+// checks for white-space characters in "C" and "POSIX" locales.
+func isSpace(b byte) bool {
+	return b == ' ' || b == '\f' || b == '\n' || b == '\r' || b == '\t' || b == '\v'
+}
+
+// You can use " around spaces, but can't escape ".  See next_arg() in kernel code /lib/cmdline.c
```
sinnykumari, line 613: extra space before `See`

pkg/daemon/update_test.go
```diff
@@ -274,6 +274,16 @@ func TestKernelAguments(t *testing.T) {
 			deletions: []string{"nosmt", "baz=test"},
 			additions: []string{"hello=world"},
 		},
+		{
+			args:      []string{"foo", "bar=1 hello=world"},
+			deletions: []string{"nosmt", "baz=test"},
+			additions: []string{"bar=1", "hello=world"},
+		},
+		{
+			args:      []string{"foo", " baz=test bar=\"hello world\""},
```
sinnykumari, line 283: Not sure if space after opening quote `" baz=test` is intentional

runcom:
laeving the hold to check on addressing @sinnykumari comments cc @ericavonb - feel free to unhold once that's ok :+1:

pkg/daemon/update_test.go
ericavonb: Yeah, it was intentional. Wanted to test that leading spaces get trimmed appropriately.

ericavonb:
> It will be nice to add something like `"baz=test bar=\"hello world\"` in our existing e2e test too so that we know these kargs are getting applied correctly.
@sinnykumari do you think this is good enough of a test? https://github.com/openshift/machine-config-operator/pull/1563/commits/e4f77890d46e82ccc022afbbe59dce15b41d0c4c
The `strings.Contains` should be good enough for the quoted kargs since the kernel doesn't do escaped quotes. But I realized, this test could be fooled if a tested karg string is contained in an existing karg like in a quote! Seems unlikely, but maybe deserves a followup pr to fix?

ashcrow:
Infra failres on a few, gcp items are unit tests catching:
```
 --- FAIL: TestKernelArguments (678.99s)
    mcd_test.go:210: Created kargs-c76cb57e-0598-4fa4-9cee-0d6259a29f42
    utils_test.go:60: Pool worker has rendered config kargs-c76cb57e-0598-4fa4-9cee-0d6259a29f42 with rendered-worker-03178d87edd8c8677cddbd3c2f5a867a (waited 2.01149864s)
    utils_test.go:82: Pool worker has completed rendered-worker-03178d87edd8c8677cddbd3c2f5a867a (waited 11m16.007908522s)
    mcd_test.go:225: Missing 'bar="hello world"' in kargs 
```

sinnykumari:
> > It will be nice to add something like `"baz=test bar=\"hello world\"` in our existing e2e test too so that we know these kargs are getting applied correctly.
> 
> @sinnykumari do you think this is good enough of a test? [e4f7789](https://github.com/openshift/machine-config-operator/commit/e4f77890d46e82ccc022afbbe59dce15b41d0c4c)

This looks perfect!

> The `strings.Contains` should be good enough for the quoted kargs since the kernel doesn't do escaped quotes. But I realized, this test could be fooled if a tested karg string is contained in an existing karg like in a quote! Seems unlikely, but maybe deserves a followup pr to fix?

Right, we can fix these later. 

To make e2e-gcp-op happy, let's fix the expected karg to `bar="hello world"` at https://github.com/openshift/machine-config-operator/commit/e4f77890d46e82ccc022afbbe59dce15b41d0c4c#diff-f149a34a66a15b4b5a900e0ab4a001acR222 since this is what gets [appended to kargs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1563/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1275875686459904000/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-88jb8_machine-config-daemon.log)

PR #1564, opened 2020-03-13
MCD: Add node-ip subcommand

celebdor:
node-ip is a subcommand that allows the user to see which IP should the
node use in cases of multiple interface and multiple address nodes. This
is useful to prevent cases where Container Runtime related services bind
to an interface that is not reachable in the control plane.

It has two commands:

* show: Takes one or more Virtual IPs of the control plane and it gives
  you one eligible IP on stdout.

* set: Takes one or more Virtual IPs of the control plane and sets
  systemd service configuration for services like CRI-O or Kubelet that
  need to bind to the control plane.

### DIFF
--- a/cmd/machine-config-daemon/node-ip.go
+++ b/cmd/machine-config-daemon/node-ip.go
@@ -1,0 +1,150 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+
+	// Enable sha256 in container image references
+	_ "crypto/sha256"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/openshift/machine-config-operator/pkg/daemon/nodenet"
+)
+
+const (
+	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
+	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+)
+
+var nodeIPCmd = &cobra.Command{
+	Use:                   "node-ip",
+	DisableFlagsInUseLine: true,
+	Short:                 "Node IP tools",
+	Long:                  "Node IP has tools that aid in the configuration of nodes in platforms that use Virtual IPs",
+}
+
+var nodeIPShowCmd = &cobra.Command{
+	Use:                   "show",
+	DisableFlagsInUseLine: true,
+	Short:                 "Show a configured IP address that directly routes to the given Virtual IPs",
+	Args:                  cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		err := show(cmd, args)
+		if err != nil {
+			glog.Exitf("error in node-ip show: %v\n", err)
+		}
+	},
+}
+
+var nodeIPSetCmd = &cobra.Command{
+	Use:                   "set",
+	DisableFlagsInUseLine: true,
+	Short:                 "Sets container runtime services to bind to a configured IP address that directly routes to the given virtual IPs",
+	Args:                  cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		err := set(cmd, args)
+		if err != nil {
+			glog.Exitf("error in node-ip set: %v\n", err)
+		}
+	},
+}
+
+// init executes upon import
+func init() {
+	rootCmd.AddCommand(nodeIPCmd)
+	nodeIPCmd.AddCommand(nodeIPShowCmd)
+	nodeIPCmd.AddCommand(nodeIPSetCmd)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+}
+
+func show(_ *cobra.Command, args []string) error {
+	vips := make([]net.IP, len(args))
+	for i, arg := range args {
+		vips[i] = net.ParseIP(arg)
+		if vips[i] == nil {
+			return fmt.Errorf("Failed to parse IP address %s", arg)
+		}
+		glog.V(3).Infof("Parsed Virtual IP %s", vips[i])
+	}
+
+	nodeAddrs, err := nodenet.AddressesRouting(vips, nodenet.NonDeprecatedAddress, nodenet.NonDefaultRoute)
+	if err != nil {
+		return err
+	}
+
+	if len(nodeAddrs) > 0 {
+		fmt.Println(nodeAddrs[0])
+	} else {
+		return fmt.Errorf("Failed to find node IP")
+	}
+
+	return nil
+}
+
+func set(_ *cobra.Command, args []string) error {
+	vips := make([]net.IP, len(args))
+	for i, arg := range args {
+		vips[i] = net.ParseIP(arg)
+		if vips[i] == nil {
+			return fmt.Errorf("Failed to parse IP address %s", arg)
+		}
+		glog.V(3).Infof("Parsed Virtual IP %s", vips[i])
+	}
+
+	nodeAddrs, err := nodenet.AddressesRouting(vips, nodenet.NonDeprecatedAddress, nodenet.NonDefaultRoute)
+	if err != nil {
+		return err
+	}
+
+	var chosenAddress net.IP
+	if len(nodeAddrs) > 0 {
+		chosenAddress = nodeAddrs[0]
+	} else {
+		return fmt.Errorf("Failed to find node IP")
+	}
+	glog.Infof("Chosen Node IP %s", chosenAddress)
+
+	// Kubelet
+	glog.V(2).Infof("Opening Kubelet service override path %s", kubeletSvcOverridePath)
+	kOverride, err := os.Create(kubeletSvcOverridePath)
+	if err != nil {
+		return err
+	}
+	defer kOverride.Close()
+
+	kOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"KUBELET_NODE_IP=%s\"\n", chosenAddress)
+	glog.V(3).Infof("Writing Kubelet service override with content %s", kOverrideContent)
+	_, err = kOverride.WriteString(kOverrideContent)
+	if err != nil {
+		return err
+	}
+
+	// CRI-O
+	crioOverrideDir := filepath.Dir(crioSvcOverridePath)
+	err = os.MkdirAll(crioOverrideDir, 0755)
+	if err != nil {
+		return err
+	}
+	glog.V(2).Infof("Opening CRI-O service override path %s", crioSvcOverridePath)
+	cOverride, err := os.Create(crioSvcOverridePath)
+	if err != nil {
+		return err
+	}
+	defer cOverride.Close()
+
+	cOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"CONTAINER_STREAM_ADDRESS=%s\"\n", chosenAddress)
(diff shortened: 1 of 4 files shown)


celebdor:
Remaining:

- Finish *set*
- Once this makes it into RHCOS, make a PR to have the VIP using platforms change to this over the Python implementation.

vrutkovs:
/cc @LorbusChris 

we'll want this in `fcos` branch too

celebdor:
The Set command is working :-)

LorbusChris:
@celebdor please run  `make go-deps` to fix the code-generator scripts file permission change

/cherry-pick fcos

celebdor:
> @celebdor please run `make go-deps` to fix the code-generator scripts file permission change
> (changing 100755 → 100644 is not wanted here)
> 
> /cherry-pick fcos

Done with the `make go-deps`

pkg/daemon/nodenet/utils.go
```diff
@@ -0,0 +1,141 @@
...
+
+// AddressFilter is a function type to filter addresses
+type AddressFilter func(netlink.Addr) bool
+
+// RouteFilter is a function type to filter routes
+type RouteFilter func(netlink.Route) bool
+
+func getAddrs() (addrMap map[netlink.Link][]netlink.Addr, err error) {
+	nlHandle, err := netlink.NewHandle()
+	defer nlHandle.Delete()
```
cybertron: If the previous call errors, is this going to work? Maybe the defer should go after the err check?

pkg/daemon/nodenet/utils.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+		}
+	}
+	glog.V(7).Infof("retrieved Address map %+v", addrMap)
+	return addrMap, nil
+}
+
+func getRouteMap() (routeMap map[int][]netlink.Route, err error) {
+	nlHandle, err := netlink.NewHandle()
+	defer nlHandle.Delete()
```
cybertron: Ditto.

pkg/daemon/nodenet/utils.go
celebdor: you're right! Thanks

sinnykumari:
Do we need unit test added for this?

yuqi-zhang:
This looks sane although I am not the most familiar with regarding node ip. Since this is a new cmd it'd be nice to add some basic unit tests like Sinny said.

LorbusChris:
Not to say it would replace unit testing, but maybe add a commit that enables the use of this new cmd for e2e testing on openstack and baremetal?
E.g. like in https://github.com/openshift/machine-config-operator/pull/1559/commits/a477762ec793542863e9d0b3b70eb0b44103eb92

kikisdeliveryservice:
I'd also like tests added..

@rphillips @umohnani8 @haircommander 

Any thoughts? Is MCD appropriate place for this?  Anything to consider from a crio/kubelet perspective?

kikisdeliveryservice:
Basic question: Is this making changes via commandline?

If so, if the point of the MCO is to make changes via MC/Kubeletcfgs/etc... does it make sense?

I dont know, you tell me =)

cgwalters:
I think I see the rationale for this, but...it's really weird from an architecture perspective to have the MCD go poking at *both* kubelet and crio configuration.

crio already has code to try to detect this, no?  Since we only support crio, any reason why this isnt' a crio command or so?

cgwalters:
This is a generic Kubernetes problem too right?  Feels like we could try to upstream this somewhere in Kube, though I am not plugged in enough there to say.

kikisdeliveryservice:
until this gets sorted out

/hold

kikisdeliveryservice:
Thinking about this more:

The flow for interacting with kubelet/crio via MCO is kubelet/crio cfg -> machine config -> pool.  This seems to bypass this via a  _new_ interface (the MCO's UI is machine configs). I feel uncomfortable using the MCO to make on the fly changes that aren't captured in one of the above cfgs.  I'm not sure I understand why this is bypassing that existing UI? Or maybe this means, as Colin suggested, that these changes are better made elsewhere?

kikisdeliveryservice:
Also ping: @openshift/openshift-team-sdn 

Was told you might have some thoughts on this.

celebdor:
It was suggested by members of the MCO team that, in order to move from Python to golang, we put this functionality as subcommands of MCD. This way it gets propagated to the RHCOS images.

The purpose of this commands is to be available for scripts that ARE created with the MachineConfig interface, namely storage->files that get placed in /etc/NetworkManager dispatcher scripts location.

The reason why this is not all just a MachineConfig placing a couple of /etc/systemd/service.d/ files is that finding out which IP is on the control plane network is something to be done at runtime. This could then be tackled upstream in CRI-O and Kubelet. The latter we know has some code to choose the reporting address and it surely can be enhanced there not to use deprecated IPv6 addresses which is something that is generically wrong. However, that does not mean that it would pick an address in the control plane rather than in the provisioning or the storage networks and it seems rather unlikely that we could convince them to add code for that (although if the consensus here is that we do just that, we can give a shot at making a Kubelet PR that allows you to specify from which subnet the address should be chosen). The CRI-O situation is quite similar, it would mean pushing there the concept of preferred subnets and waiting for CRI-O and Kubelet versions to include this preferred subnets knobs (which they may well say that are unnecessary logic since you can just set a config option with the IP of your choice) before we can drop Python from RHCOS.

danwinship:
> The latter we know has some code to choose the reporting address and it surely can be enhanced there not to use deprecated IPv6 addresses which is something that is generically wrong. However, that does not mean that it would pick an address in the control plane rather than in the provisioning or the storage networks and it seems rather unlikely that we could convince them to add code for that

Indeed, the entire point of having the `--node-ip` argument is that kubelet couldn't possibly guess the right value in *every* possible weird setup.

> The CRI-O situation is quite similar

I don't think it is. The default value of `--stream-address` is not the autodetected node IP, it's `127.0.0.1`. I think the problem there is just that you need to set it to `::1` on IPv6 nodes, not that you need to set it to the correct node IP. (Or, really, we need golang to support being able to listen dual-stack on localhost...)

cmd/machine-config-daemon/node-ip.go
```diff
@@ -0,0 +1,150 @@
...
+const (
+	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
+	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+)
+
+var nodeIPCmd = &cobra.Command{
+	Use:                   "node-ip",
+	DisableFlagsInUseLine: true,
+	Short:                 "Node IP tools",
+	Long:                  "Node IP has tools that aid in the configuration of nodes in platforms that use Virtual IPs",
```
danwinship: It's not really clear in this context what "virtual" means. It might be better to just say "multiple IPs"?

cmd/machine-config-daemon/node-ip.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+			glog.Exitf("error in node-ip show: %v\n", err)
+		}
+	},
+}
+
+var nodeIPSetCmd = &cobra.Command{
+	Use:                   "set",
+	DisableFlagsInUseLine: true,
+	Short:                 "Sets container runtime services to bind to a configured IP address that directly routes to the given virtual IPs",
```
danwinship: So I assume the usage here is that you pass it the IPs of the other masters, in order to find an IP on the interface that routes to those other IPs, rather than an IP on a provisioning/storage network? In that case, the IPs you pass would be *non*-virtual IPs, right?

cmd/machine-config-daemon/node-ip.go
```diff
@@ -0,0 +1,150 @@
...
+	Use:                   "node-ip",
+	DisableFlagsInUseLine: true,
+	Short:                 "Node IP tools",
+	Long:                  "Node IP has tools that aid in the configuration of nodes in platforms that use Virtual IPs",
+}
+
+var nodeIPShowCmd = &cobra.Command{
+	Use:                   "show",
+	DisableFlagsInUseLine: true,
+	Short:                 "Show a configured IP address that directly routes to the given Virtual IPs",
```
danwinship: It's not just *a* configured IP address, it's the *right* one (for some definition of "right")

haircommander:
> > The latter we know has some code to choose the reporting address and it surely can be enhanced there not to use deprecated IPv6 addresses which is something that is generically wrong. However, that does not mean that it would pick an address in the control plane rather than in the provisioning or the storage networks and it seems rather unlikely that we could convince them to add code for that
> 
> Indeed, the entire point of having the `--node-ip` argument is that kubelet couldn't possibly guess the right value in _every_ possible weird setup.
> 
> > The CRI-O situation is quite similar
> 
> I don't think it is. The default value of `--stream-address` is not the autodetected node IP, it's `127.0.0.1`. I think the problem there is just that you need to set it to `::1` on IPv6 nodes, not that you need to set it to the correct node IP. (Or, really, we need golang to support being able to listen dual-stack on localhost...)

yeah, this is the only case I can see that CRI-O would need to be configured like this. For all other IP address choices it makes (choosing hostIP for a pod), it defers to the kubelet.

If that case, or the case of configuring the kubelet alone, is deemed worthy of a new command in the MCD, I don't see anything wrong with this change :+1:

LorbusChris:
I'm wondering, if MCO is not comfortable taking this functionality in, couldn't it also just live in https://github.com/openshift/baremetal-runtimecfg?

celebdor:
> I'm wondering, if MCO is not comfortable taking this functionality in, couldn't it also just live in https://github.com/openshift/baremetal-runtimecfg?

Then it would not be in the RHCOS image (MCD is installed there from an RPM package). Do you suggest to use podman to extract it in ExecStartPre of the units that use it?

LorbusChris:
Couldn't we run it as a containerized command (without extracting)? We we would've done with the python scripts.

cgwalters:
If we move it into the baremetal operator, we're kind of saying that multiple NICs and more sophisticated network is only a baremetal problem, but I don't think that's true.

celebdor:
@cgwalters I think @LorbusChris suggested baremetal-runtimecfg, which now that it is used by several platforms could use a rename (naming is hard :'( ), not the baremetal operator.

kikisdeliveryservice:
> @cgwalters I think @LorbusChris suggested baremetal-runtimecfg, which now that it is used by several platforms could use a rename (naming is hard :'( ), not the baremetal operator.

Hi!
Q:
who is using baremetal-runtimecfg vs who is covered by the use-case for this PR?

vrutkovs:
oVirt, Openstack and Baremetal IPI are using these scripts

LorbusChris:
However, this PR wouldn't actually make any of those platforms use the new commands. To activate, one would need a commit like this: https://github.com/openshift/machine-config-operator/pull/1559/commits/82fa8a72cc904e16f66bb603daa6dfa7386b1ec1 (just with the correct MCD path for OCP)

kikisdeliveryservice:
> oVirt, Openstack and Baremetal IPI are using these scripts

do they also use baremetal-runtimecfg?

kikisdeliveryservice:
> Couldn't we run it as a containerized command (without extracting)? Like we would've done with the python scripts.

@celebdor WDYT? if all of those platforms do indeed use baremetal-runtimecfg (rename pending :wink: )

cmd/machine-config-daemon/node-ip.go
celebdor: You give it the virtual IPs so that then it can find which non virtual IPs are on the virtual IPs network, which will rule out non virtual IPs in other networks (like a storage network).

LorbusChris:
This is superseded by https://github.com/openshift/machine-config-operator/pull/1659

/close

PR #1566, opened 2020-03-13
reconcile relatedObjects and place them in the clusteroperator for CVO

deads2k:
fixes relatedObject reconciliation found in https://github.com/openshift/cluster-version-operator/pull/318 .  This is in service to https://github.com/openshift/enhancements/pull/149 .  The addition of relatedObjects in the CO itself means that when the operator doesn't start during bootstrapping, we can still get good PD data.

### DIFF
--- a/install/0000_80_machine-config-operator_06_clusteroperator.yaml
+++ b/install/0000_80_machine-config-operator_06_clusteroperator.yaml
@@ -9,3 +9,16 @@ status:
   versions:
     - name: operator
       version: "0.0.1-snapshot"
+  relatedObjects:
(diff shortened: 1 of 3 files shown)


deads2k:
huh, 500 error on images.  Doesn't look like me.

runcom:
@deads2k any reason why other CO aren't doing this? e.g. https://github.com/openshift/machine-api-operator/blob/35a3cd0b1d470d024c65d6237b17251b988fbc98/install/0000_30_machine-api-operator_12_clusteroperator.yaml

PR #1571, opened 2020-03-18
pkg/daemon: use upstream kubectl/pkg/drain

runcom:
Get rid of openshift/kubernetes-drain and use the upstream kubectl/pkg/drain pkg.
MAO has switched to that as well and it's the right way forward since the old
drain pkg in openshift and cluster-apis obsolete.

Closes https://github.com/openshift/machine-config-operator/issues/71 also

This is a pre-req to fix https://bugzilla.redhat.com/show_bug.cgi?id=1814241


pkg/daemon/daemon.go
```diff
@@ -288,6 +290,37 @@ func (dn *Daemon) ClusterConnect(
...
 	dn.kubeletHealthzEnabled = kubeletHealthzEnabled
 	dn.kubeletHealthzEndpoint = kubeletHealthzEndpoint
+
+	dn.drainer = &drain.Helper{
+		Client:              dn.kubeClient,
+		Force:               true,
+		IgnoreAllDaemonSets: true,
+		DeleteLocalData:     true,
+		GracePeriodSeconds:  -1,
+		Timeout:             20 * time.Second,
```
runcom, line 300: It seems this isn't enough for certain pods like the router (we bumped that to 90s in master, and MAO still has that at 20s) - this is related to https://bugzilla.redhat.com/show_bug.cgi?id=1843998 and it seems that the router can't be evicted because this drain isn't respecting the 60m graceful termination period.
@michaelgugino @enxebre do you know more about this behavior or what is best? This patch regresses the behavior as we were waiting "forever" before this patch, and then only 20s/90s do you have any inputs here?

cc @danehans

michaelgugino: Timeout here is just how long the drain library will run before giving up (and then drain will be run again on a subsequent requeue).

GracePeriodSeconds is the amount of time that the eviction API will set as grace period if PDBs allow deleting the pod.  -1 (IIRC) means use the pod's specified grace period.

PR #1504, opened 2020-02-25
[baremetal] Change to unicast for keepalived

bcrochet:
This implements a unicast solution for keepalived. Multicast is often not available
so unicast is a more common method.

WIP

**- What I did**

**- How to verify it**

**- Description for the changelog**

PR #1574, opened 2020-03-22
[baremetal] Add support for reading API LB backends from KUBE-API

yboaron:
This PR updates HAProxy static pod manifest to enable reading API LB backend details
from Kubernetes instead of _etcd-server-ssl._tcp SRV.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
@@ -17,9 +17,9 @@ contents:
       - name: resource-dir
         hostPath:
           path: "/etc/kubernetes/static-pod-resources/haproxy"
-      - name: kubeconfig
+      - name: kubeconfigvarlib
         hostPath:
-          path: "/etc/kubernetes/kubeconfig"
+          path: "/var/lib/kubelet"
       - name: run-dir
         empty-dir: {}
       - name: conf-dir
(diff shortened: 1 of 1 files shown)


templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -19,7 +19,10 @@ contents:
           path: "/etc/kubernetes/static-pod-resources/haproxy"
       - name: kubeconfig
         hostPath:
-          path: "/etc/kubernetes/kubeconfig"
+          path: "/etc/kubernetes"
```
celebdor: Why do we do this path change?

yboaron: @celebdor 
Hmm, I think that @bcrochet  reported in keepalived-unicast PR [1] that he needed to mount /etc/kubernetes instead of /etc/kubernetes/kubeconfig - so I did the same also here. 
I'll recheck this PR this time with /etc/kubernetes/kubeconfig, if it's works - I'll update the PR.
[1] 
https://github.com/openshift/machine-config-operator/pull/1504/files#diff-c2a4c0270459a7bea650ed87b237329eR22

yboaron: @celebdor - It appears that we don't need the kubeconfig anymore here, just the '/var/lib/kubelet' , so, I just deleted it and added the 'var/lib/kubelet'

PR #1579, opened 2020-03-24
vsphere ipi: set hostname using vmtoolsd and VM extra config

jcpowermac:
This commit adds two new files:
- A systemd unit that runs a shell script
- A shell script that runs vmtoolsd to retrieve `guestinfo.hostname`
from vm guest. Then uses hostnamectl to set hostname.

### DIFF
--- a/templates/common/vsphere/files/vsphere-hostname.yaml
+++ b/templates/common/vsphere/files/vsphere-hostname.yaml
@@ -1,0 +1,14 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/vsphere-hostname.sh"
+contents:
+  inline: |
+    #!/usr/bin/env bash
+    set -e
+
+    if [ $(hostname -s) = "localhost" ]; then
+      if hostname=$(/bin/vmtoolsd --cmd 'info-get guestinfo.hostname'); then
+          /usr/bin/hostnamectl --transient --static set-hostname ${hostname}
+      fi
+    fi
+
--- a/templates/common/vsphere/units/vsphere-hostname.yaml
+++ b/templates/common/vsphere/units/vsphere-hostname.yaml
@@ -1,0 +1,16 @@
+name: "vsphere-hostname.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=vSphere hostname
+  After=vmtoolsd.service
+  Before=kubelet.service
+
+  [Service]
+  ExecStart=/usr/local/bin/vsphere-hostname.sh
+  Restart=on-failure
+  RestartSec=15
+
+  [Install]
+  WantedBy=multi-user.target
+


templates/common/vsphere/files/vsphere-hostname.yaml
```diff
@@ -0,0 +1,12 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/bin/vsphere-hostname.sh"
+contents:
+  inline: |
+    #!/usr/bin/env bash
+    set -e
+
+    if hostname=$(/bin/vmtoolsd --cmd 'info-get guestinfo.hostname'); then
```
abhinavdahiya: is this binary present in the RHCOS system today? should we inform RHCOS people that please don't remove this binary in later versions as we depend on it.

templates/common/vsphere/files/vsphere-hostname.yaml
```diff
@@ -0,0 +1,12 @@
...
+        /usr/bin/hostnamectl --transient --static set-hostname ${hostname}
```
abhinavdahiya:
```
 hostnamectl set-hostname --help
hostnamectl [OPTIONS...] COMMAND ...

Query or change system hostname.

  -h --help              Show this help
     --version           Show package version
     --no-ask-password   Do not prompt for password
  -H --host=[USER@]HOST  Operate on remote host
  -M --machine=CONTAINER Operate on local container
     --transient         Only set transient hostname
     --static            Only set static hostname
     --pretty            Only set pretty hostname
```

is providing both `transient` and `static` valid option?

jcpowermac: yep...tested and reviewed man page.

templates/common/vsphere/files/vsphere-hostname.yaml
jcpowermac: Yeah its in RHCOS today.  I tested out all the commands prior to scripting and using systemd to execute.

cgwalters: It's not in FCOS though and we do want to eventually containerize it (once vmware stops fighting that).

cgwalters:
OK but...we've had a whole lot of extensive discussions about vmware and networking (which includes hostname settings for static IP cases).

See for example https://github.com/coreos/fedora-coreos-tracker/issues/427 and https://github.com/coreos/ignition-dracut/pull/156 as well as https://github.com/coreos/afterburn/pull/379

A core problem with the hostname is that it's messy who owns it and sets it.

One option is to ensure we only run this if a hostname was not otherwise set (i.e. only set if it's `localhost`).

Does setting this value require the admin to do something specifically, or the installer will be setting it?

jcpowermac:
> 
> Does setting this value require the admin to do something specifically, or the installer will be setting it?
For masters its the installer via terraform for workers its MAO

templates/common/vsphere/files/vsphere-hostname.yaml
jcpowermac: @cgwalters I have an example of running open-vm-tools in a container on fcos.  Using @dav1x blog post as an example: https://developers.redhat.com/blog/2017/03/23/containerizing-open-vm-tools-part-1-the-dockerfile-and-constructing-a-systemd-unit-file/

It would be interesting if I could query guestinfo by using an `exec` but I have yet to try.

jcpowermac: https://gist.github.com/jcpowermac/23ea90e549c02b43758662f90d83a9f1

LorbusChris: iirc we layer `vmtoolsd` into OKD's machine-os-content, so it'll be on master/worker nodes after bootstrapping the cluster.
Is script also present on bootstrap hosts?

jcpowermac: @LorbusChris nope I didn't add it to bootstrap.  bootstrap is fine as `localhost`.

jcpowermac:
@abhinavdahiya I added a check for `localhost` when you have a moment can you please `/lgtm`.

cgwalters:
Since this is IPI, the installer today can just update the Ignition config for each host to include `/etc/hostname` - this is a supported cross-platform way to set the hostname.  I'd *strongly* prefer this unless there's a good reason the installer can't do this.

cgwalters:
/hold
per above

jcpowermac:
> Since this is IPI, the installer today can just update the Ignition config for each host to include `/etc/hostname` - this is a supported cross-platform way to set the hostname. I'd _strongly_ prefer this unless there's a good reason the installer can't do this.

MAO creates workers the installer does not.

cgwalters:
> MAO creates workers the installer does not.

Right, though the installer generates the user data.  But...it's true today we don't support having *dynamic* userdata from the MAO (which would really then directly boil down to machine-specific machineconfigs since the userdata just points to the MCS, which we do want for the MCO, but *that* problem then goes into *identifying* machines...)

OK so I think where we really want this code then is as part of
https://github.com/coreos/afterburn/
which already today supports hostname across various providers but it looks like we don't have VMWare yet.  I think @lucab is working on it?

One continual problem is people add random "grab bag" things to the MCO that might otherwise go into other CoreOS components I think because it's a repository conveniently integrated with OpenShift CI whereas patching other things goes into the world of RPM pain.

OK though short term,
/hold cancel
because we can at least drop this code once the afterburn code lands.

enxebre:
cc @alexander-demichev

jcpowermac:
> 
> OK so I think where we really want this code then is as part of
> https://github.com/coreos/afterburn/
> which already today supports hostname across various providers but it looks like we don't have VMWare yet. I think @lucab is working on it?
> 

Completely agree.  When that is ready please let me know and we can switch it out.

ashcrow:
> OK so I think where we really want this code then is as part of
> https://github.com/coreos/afterburn/
> which already today supports hostname across various providers but it looks like we don't have VMWare yet. I think @lucab is working on it?

I agree :+1: I think `afterburn` makes sense or maybe `nm-cloud-setup` per previous conversations if this is a more general need.

cgwalters:
xref https://github.com/coreos/afterburn/pull/379/files#r399323764

templates/common/vsphere/files/vsphere-hostname.yaml
dav1x: I had to do something similar to this when I converted the container to runc. The exec should work fine:

I changed the systemd unit file to call this shell script instead. 
```
#!/bin/sh
if [ -f /tmp/open-vm-tools-bash-env ]
  then
    source /tmp/open-vm-tools-bash-env
fi

COMMAND=/usr/local/bin/vmware-toolbox-cmd
if [ ! -e $COMMAND ]
  then
    echo 'runc exec -t open-vm-tools vmware-toolbox-cmd "$@"' > /usr/local/bin/vmware-toolbox-cmd
    chmod +x /usr/local/bin/vmware-toolbox-cmd
fi
exec /usr/bin/VGAuthService -s &
exec /usr/bin/vmtoolsd
```

darkmuggle:
To echo what @cgwalters said -- the way hostnames are set in Linux is really messing and its a landmine. If we have NM, Ignition (via a files stanza for /etc/hostname), Afterburn and the MCO, the debug burden starts to grow. Reducing the number of things that can act on the hostname is to our advantage, IMHO.

lucab:
(Sorry for the delay, I was on PTO)

I'm unfortunately missing most context about this and there are no references in PR/commit, can anybody provide that?

In particular, before giving any meaningful feedback, these are my doubts:
 * is this meant for either nodes with DHCP, or with static networking, or both?
 * if this is for nodes with DHCP, why does hostname Option 12 not work? If both are provided, which one is intended to be the valid one?
 * if this is for nodes with static networking, why does hostname kargs setting not work? If both are provided, which one is intended to be the valid one?
 * if the Ignition config specifies some `/etc/hostname` value, which setting is intended to be the valid one?
 * where does `guestinfo.hostname` come from? Is it something we just invented for OCP?
 * is the guestinfo allowed to change while a VM is running?
 * why is MCO a relevant place for this?

In general I agree with all the feedbacks above, this is troublesome because 1) it bypasses the (already too many) possible owners of hostname logic with a random script in an unrelated OCP repo 2) it also reverts the "Ignition as a single source of truth" for file config.

PR #1586, opened 2020-03-26
Bug 1814397: pkg/daemon: fix deletion of stale files

runcom:
We have a serious bug in how we backup "original" files and restore them.
Here, "original" means files that ship with RHCOS. Think of a default Chrony
or another system daemon configuration file. When the MCD kicks in and writes
to those files, we want to backup the original one (the shipped-with-RHCOS) in order
to restore it if a user deletes the MC that modified it (this was the initial bug reported
in GitHub at https://github.com/openshift/machine-config-operator/issues/782).

However, that patch that fixed https://github.com/openshift/machine-config-operator/issues/782
was causing the following; if you shipped a file with just _one_ MC, removing it would wipe it out and that works.
However, if you modified that file later again with another MC, a backup
file will be created for the first MC, and when deleting the file by deleting the second MC, it will
restore the initial file shipped with the first MC instead of wiping it out completely which
it should have since that file was never meant to be backed up because it wasn't on RHCOS from the beginning.

This patch now differentiates between files that are already on RHCOS (on-disk so to speak)
and files that are shipped with an MC. For the former, the MCD will create a backup as it's doing today,
for the latter instead, the MCD creates a placeholder file that tells it to just get rid
of the file altogether (along with adding all the necessary checks and actions in order to create those backup files).

The issue popped up on upgrade paths where the new manifests rendered by the MCO don't contain a certain file.
The MCD notices that and go ahead trying to remove the file. It however notices that a backup file
(which was created for an MC shipped file and later other MC have modified it) is there and tries to restore it (also failing
with invalid cross-link device error, but that's another issue which I'm fixing here as well by using cp directly).

Really hoping all the above makes sense. Otherwise, an e2e can be consulted here https://github.com/openshift/machine-config-operator/pull/1590

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -782,9 +782,17 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
-			if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
+			if _, err := os.Stat(noOrigFileName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting no orig file %q: %v", noOrigFileName(f.Path), err)
+				}
+				glog.V(2).Infof("Removing file %q completely", f.Path)
+			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				if out, err := exec.Command("cp", "-a", "--reflink=auto", origFileName(f.Path), f.Path).CombinedOutput(); err != nil {
+					return errors.Wrapf(err, "restoring %q from orig file %q: %s", f.Path, origFileName(f.Path), string(out))
+				}
+				if err := os.Remove(origFileName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
 				}
 				glog.V(2).Infof("Restored file %q", f.Path)
 				continue
@@ -997,15 +1005,31 @@ func origParentDir() string {
 	return filepath.Join("/etc", "machine-config-daemon", "orig")
 }
 
+func noOrigParentDir() string {
+	return filepath.Join("/etc", "machine-config-daemon", "noorig")
+}
+
 func origFileName(fpath string) string {
 	return filepath.Join(origParentDir(), fpath+".mcdorig")
 }
 
+func noOrigFileName(fpath string) string {
+	return filepath.Join(noOrigParentDir(), fpath+".mcdnoorig")
+}
+
 func createOrigFile(fpath string) error {
-	if _, err := os.Stat(fpath); err != nil {
-		// the file isn't there, no need to back it up
-		// we could check ENOENT only maybe?
+	if _, err := os.Stat(noOrigFileName(fpath)); err == nil {
+		// we already created the no orig file for this default file
 		return nil
+	}
+	if _, err := os.Stat(fpath); err != nil {
+		// create a noorig file that tells the MCD that the file wasn't present on disk in RHCOS
+		// so it can just remove it when deleting stale data, as opposed as restoring a file
+		// that was shipped _with_ RHCOS (e.g. a default chrony config).
+		if err := os.MkdirAll(filepath.Dir(noOrigFileName(fpath)), 0755); err != nil {
+			return errors.Wrapf(err, "creating no orig parent dir: %v", err)
+		}
+		return writeFileAtomicallyWithDefaults(noOrigFileName(fpath), nil)
 	}
 	if _, err := os.Stat(origFileName(fpath)); err == nil {
 		// the orig file is already there and we avoid creating a new one to preserve the real default


runcom:
@cgwalters ptal 🙏

pkg/daemon/update.go
```diff
@@ -997,16 +1005,32 @@ func origParentDir() string {
...
-		// the file isn't there, no need to back it up
-		// we could check ENOENT only maybe?
+	if _, err := os.Stat(noOrigFileName(fpath)); err == nil {
+		// we already created the no orig file for this default file
 		return nil
 	}
+	if _, err := os.Stat(fpath); err != nil {
+		// create a noorig file that tells the MCD that the file wasn't present on disk in RHCOS
+		// so it can just remove it when deleting stale data, as opposed as restoring a file
+		// that was shipped _with_ RHCOS (e.g. a default chrony config).
```
jlebon: This is somewhat of a corner case, but: does this correctly handle an upgrade whose base OSTree introduces a config file which we previously had modified? We'd write the `noorig` file, but then if it becomes stale would just nuke it entirely instead of restoring the base one, right?

This would be a lot easier if we used the `/usr/etc` as the canonical "host" defaults, but yeah I understand we need to support traditional systems too and it's easier to have the same path for both.

pkg/daemon/update.go
```diff
@@ -782,9 +782,17 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
-			if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
+			if _, err := os.Stat(noOrigFileName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting no orig file %q: %v", noOrigFileName(f.Path), err)
+				}
+				glog.V(2).Infof("Removing file %q completely", f.Path)
+			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				if out, err := exec.Command("cp", "-a", "--reflink=auto", origFileName(f.Path), f.Path).CombinedOutput(); err != nil {
```
jlebon: Minor: I see the `exec.Command("cp", ...)` is used elsewhere too. Could probably use a factoring out to make it cleaner to call.

pkg/daemon/update.go
```diff
@@ -997,16 +1005,32 @@ func origParentDir() string {
...
+
 func createOrigFile(fpath string) error {
-	if _, err := os.Stat(fpath); err != nil {
-		// the file isn't there, no need to back it up
-		// we could check ENOENT only maybe?
+	if _, err := os.Stat(noOrigFileName(fpath)); err == nil {
+		// we already created the no orig file for this default file
 		return nil
 	}
+	if _, err := os.Stat(fpath); err != nil {
```
yuqi-zhang: So while this would work with newly created clusters, existing clusters would still run into this problem after we upgrade to this version, since we already lost the fact that the file originally didn't exist

pkg/daemon/update.go
jlebon: Thinking more on this: unless I'm misunderstanding, I'm not sure if using `noorig` is the right approach since it's signaling something that can change. One could imagine the opposite scenario as well: a file we previously shipped in the host and modified with an MC is now dropped out of the host; once you nuke the MC, do we restore the original file instead of deleting it?

yuqi-zhang:
> Why do we backup the file if it's not actually newly modified? Isn't that the root problem here?

No that's not the problem. This issue only occurs when

1. the user specifies a file `/my/file` that didn't originally exist with MC1
2. the user then overwrites `/my/file` with MC2 with new contents (here the MCO makes a backup copy of MC1 `/my/file.orig` when it shouldn't have)
3. the user deletes MC1 and MC2
4. the MCO recognizes that we had a backup, and since the rendered MC doesn't have `/my/file` in it anywhere, that must mean we need to back up from `/my/file.orig`, causing a diff from expected behaviour.

This backup originally is meant to fix scenarios like:

1. the user specifies a file `/my/file` that with MC1, and `/my/file` DID already exist on disk with a default file (here the MCO makes a backup copy of `/my/file.orig` which is correct)
3. the user deletes MC1
4. the MCO recognizes that we had a backup, and since the rendered MC doesn't have `/my/file` in it anywhere, that must mean we need to back up from `/my/file.orig`, which is fine

pkg/daemon/update.go
runcom: This will be covered by patching release-1 (in this case 4.3) - see https://github.com/openshift/machine-config-operator/pull/1587 <- this patch takes care of a best effort to understand which file was on the host/shipped by an rpm like chrony and the like

runcom:
> As a side note do you know why we don't back up systemd units

there's a BZ we need to fix to fix that - it's tracked for now and you're right, we should be doing that for systemd units (and dropins as well)

jlebon:
> the user then overwrites `/my/file` with MC2 with new contents (here the MCO makes a backup copy of MC1 `/my/file.orig` when it shouldn't have)

That's what I mean though. If we fix just that, then the current orig approach still works fine (notwithstanding the issues mentioned [here](https://github.com/openshift/machine-config-operator/pull/1586#discussion_r398572693)) and this whole issue goes away, right?

runcom:
> > the user then overwrites `/my/file` with MC2 with new contents (here the MCO makes a backup copy of MC1 `/my/file.orig` when it shouldn't have)
> 
> That's what I mean though. If we fix just that, then the current orig approach still works fine (notwithstanding the issues mentioned [here](https://github.com/openshift/machine-config-operator/pull/1586#discussion_r398572693)) and this whole issue goes away, right?

that's what this does right? it makes sure to create the backup file only if the file isn't already there

pkg/daemon/update.go
yuqi-zhang: > does this correctly handle an upgrade whose base OSTree introduces a config file which we previously had modified?

Hm, now I think about it, if we are rebooting into a new oscontainer with a new version of a file, how are we consolidating that today?

And yeah this would be another ugly scenario. Correct me if I'm wrong but I think this would only happen if we somehow created a new MC and the OSTree to write the same file right?

cgwalters:
It's worth looking at how ostree handles the base problem here.  The way it works is we always checkout the *new defaults*, then calculate a diff between *current defaults* and *current*, and apply that diff to the new.

It's a lot like `git rebase`.

The most important thing here is we're not explicitly trying to delete files that no longer apply - we're always starting from the new base.

It's also how ostree handles the base `/usr` updates for the readonly files - we checkout the *new* tree, we don't try to diff current/new and apply that.

Whereas this is trying to mutate the *current* state directly into a new state - a lot like how `rpm -Uvh` works.  

On a related topic, this "mutate the current" is also extremely hard to do *transactionally* - neither `rpm -Uvh` nor MCO changes are.

I think the long term what we want to do is capture what I'll call the "local diff" - any changes made on the node that are *not* part of the MCO.  Then on upgrades, we:

- apply OS base defaults
- apply *new* MCO config on top
- apply local diff

Then in no case do we need to even care about a diff versus the *old* MCO state, dealing with `.orig` files and the like.

But this is harder to do while we need to care about !CoreOS.

cgwalters:
It should be really pretty easy to put together a set of unit tests for this right?  Or perhaps better, a more "blackbox" test that runs in a pod:

```
machine-config-daemon ignition-files-update /path/to/basedir /path/to/mc1.yaml /path/to/mc2.yaml
```

Basically `ignition-files-update` would be a CLI entrypoint that just calls into `updateFiles()` without requiring a cluster even.

pkg/daemon/update.go
jlebon: Should check specifically for `ENOENT` here: https://golang.org/pkg/os/#IsNotExist

sinnykumari:
Did we close this PR by mistake :thinking:

yuqi-zhang:
This has been superceded by https://github.com/openshift/machine-config-operator/pull/1593

PR #1589, opened 2020-03-26
[FCOS] gcp templates: update gcp-routes

vrutkovs:
**- What I did**

**- How to verify it**


vrutkovs:
Flakes

/retest

vrutkovs:
This version of `gcp-routes.sh` is exactly the same as RHCOS uses. The issue is that it doesn't pass shellcheck on installer.

/hold

LorbusChris:
I think this is the wrong version of the script, should we close?

LorbusChris:
feel free to reopen if that's not the case :)
/close

PR #1590, opened 2020-03-26
test/e2e: add e2e to test files proper deletion

runcom:
This e2e verifies what's being fixed in https://github.com/openshift/machine-config-operator/pull/1586 and it'll fail w/o that patch (which is why I'm carrying it here right now)

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -425,6 +425,45 @@ func TestReconcileAfterBadMC(t *testing.T) {
 	}
 }
 
+func TestFullyDeleteMCShippedFiles(t *testing.T) {
+	cs := framework.NewClientSet("")
+	nodes, err := getNodesByRole(cs, "worker")
+	require.Nil(t, err)
+	for _, node := range nodes {
+		// this fails if the file is on the node, just a pre-req sanity check to run this test
+		_, err := execCmdOnNodeWithErr(t, cs, node, "test", "!", "-e", "/rootfs/etc/something")
+		require.Nil(t, err)
+	}
+	workerOldMcFirst := getMcName(t, cs, "worker")
+	mcHostFileFirst := createMCToAddFile("modify-host-file-first", "/etc/something", "mco-test", "root")
+	_, err = cs.MachineConfigs().Create(mcHostFileFirst)
+	require.Nil(t, err)
+	renderedConfig, err := waitForRenderedConfig(t, cs, "worker", mcHostFileFirst.Name)
+	require.Nil(t, err)
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", renderedConfig))
+	workerOldMcSecond := getMcName(t, cs, "worker")
+	mcHostFileSecond := createMCToAddFile("modify-host-file-second", "/etc/something", "mco-test-2", "root")
+	_, err = cs.MachineConfigs().Create(mcHostFileSecond)
+	require.Nil(t, err)
+	renderedConfig, err = waitForRenderedConfig(t, cs, "worker", mcHostFileSecond.Name)
+	require.Nil(t, err)
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", renderedConfig))
+	// now delete the MCs file, which must result in the file disappearing from the system
+	require.Nil(t, cs.MachineConfigs().Delete(mcHostFileSecond.Name, &metav1.DeleteOptions{}))
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcSecond))
+	require.Nil(t, cs.MachineConfigs().Delete(mcHostFileFirst.Name, &metav1.DeleteOptions{}))
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcFirst))
+	nodes, err = getNodesByRole(cs, "worker")
+	require.Nil(t, err)
+	for _, node := range nodes {
+		assert.Equal(t, node.Annotations[constants.CurrentMachineConfigAnnotationKey], workerOldMcFirst)
+		assert.Equal(t, node.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
+		// this fails if the file is on the node
+		_, err := execCmdOnNodeWithErr(t, cs, node, "test", "!", "-e", "/rootfs/etc/something")
+		require.Nil(t, err)
+	}
+}
+
 func TestDontDeleteRPMFiles(t *testing.T) {
 	cs := framework.NewClientSet("")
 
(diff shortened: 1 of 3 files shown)


runcom:
Before carrying https://github.com/openshift/machine-config-operator/pull/1590 to turn this to green, I want to watch it fail and validate my theories.

runcom:
confirmed manually that this test covers the scenario in the linked PR and exposes the bug

test/e2e/mcd_test.go
```diff
@@ -425,6 +425,39 @@ func TestReconcileAfterBadMC(t *testing.T) {
...
+	workerOldMcSecond := getMcName(t, cs, "worker")
+	mcHostFileSecond := createMCToAddFile("modify-host-file-second", "/etc/something", "mco-test-2", "root")
+	_, err = cs.MachineConfigs().Create(mcHostFileSecond)
+	require.Nil(t, err)
+	renderedConfig, err = waitForRenderedConfig(t, cs, "worker", mcHostFileSecond.Name)
+	require.Nil(t, err)
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", renderedConfig))
+	// now delete the MCs file, which must result in the file disappearing from the system
+	require.Nil(t, cs.MachineConfigs().Delete(mcHostFileSecond.Name, &metav1.DeleteOptions{}))
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcSecond))
```
jlebon: Would be good to check here that the file still exists and has the contents dictated by the first MC.

test/e2e/mcd_test.go
```diff
@@ -425,6 +425,39 @@ func TestReconcileAfterBadMC(t *testing.T) {
...
+	require.Nil(t, cs.MachineConfigs().Delete(mcHostFileFirst.Name, &metav1.DeleteOptions{}))
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcFirst))
+	nodes, err := getNodesByRole(cs, "worker")
+	require.Nil(t, err)
+	for _, node := range nodes {
+		assert.Equal(t, node.Annotations[constants.CurrentMachineConfigAnnotationKey], workerOldMcFirst)
+		assert.Equal(t, node.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
+		found := execCmdOnNode(t, cs, node, "cat", "/rootfs/etc/something")
```
jlebon: A stronger approach is something like `test -e /rootfs/etc/something` and checking the exit code.

test/e2e/mcd_test.go
```diff
@@ -425,6 +425,39 @@ func TestReconcileAfterBadMC(t *testing.T) {
 	}
 }
 
+func TestFullyDeleteMCShippedFiles(t *testing.T) {
+	cs := framework.NewClientSet("")
```
jlebon: A good sanity check to do upfront here is that `/etc/something` *doesn't* exist (I mean, it shouldn't, but it helps make it clear how this test works :) )

test/e2e/mcd_test.go
runcom: that's the usual MCD workflow tho - here's is just diffing MC1<->MC2 so it's "applying" MC1 and we expect that to work (and tested in other tests as well)

test/e2e/mcd_test.go
runcom: can do that thanks!

test/e2e/mcd_test.go
runcom: I'd add that thanks @jlebon !

test/e2e/mcd_test.go
runcom: i.e. there's not "removal of stale files in this scenario" so there's little value in (re) testing that here

test/e2e/mcd_test.go
runcom: added

test/e2e/mcd_test.go
```diff
@@ -425,6 +425,46 @@ func TestReconcileAfterBadMC(t *testing.T) {
...
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcSecond))
+	require.Nil(t, cs.MachineConfigs().Delete(mcHostFileFirst.Name, &metav1.DeleteOptions{}))
+	require.Nil(t, waitForPoolComplete(t, cs, "worker", workerOldMcFirst))
+	nodes, err = getNodesByRole(cs, "worker")
+	require.Nil(t, err)
+	for _, node := range nodes {
+		assert.Equal(t, node.Annotations[constants.CurrentMachineConfigAnnotationKey], workerOldMcFirst)
+		assert.Equal(t, node.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
+		// this fails if the file is on the node
+		_, err := execCmdOnNodeWithErr(t, cs, node, "test", "-e", "/rootfs/etc/something")
```
jlebon: Sorry, don't mean to drag this out. :)  Can we make this `test ! -e /rootfs/etc/something` instead? That way we can check for a successful exit code. Otherwise it's hard to tell if there's an error because the command exited nonzero or something else went wrong along the way in `oc rsh`.

runcom: done and yeah, definitely better!

runcom:
remember the op test will fail here till we merge a solution (and carry this ofc)

/test e2e-gcp-op
/test e2e-aws-scaleup-rhel7

yuqi-zhang:
I think you will likely have to bump timeouts for this to pass

runcom:
> I think you will likely have to bump timeouts for this to pass

damn, you're right

test/e2e/mcd_test.go
```diff
@@ -425,6 +425,45 @@ func TestReconcileAfterBadMC(t *testing.T) {
...
+	cs := framework.NewClientSet("")
+	nodes, err := getNodesByRole(cs, "worker")
+	require.Nil(t, err)
+	for _, node := range nodes {
+		// this fails if the file is on the node, just a pre-req sanity check to run this test
+		_, err := execCmdOnNodeWithErr(t, cs, node, "test", "!", "-e", "/rootfs/etc/something")
+		require.Nil(t, err)
+	}
+	workerOldMcFirst := getMcName(t, cs, "worker")
+	mcHostFileFirst := createMCToAddFile("modify-host-file-first", "/etc/something", "mco-test", "root")
```
yuqi-zhang, line 438: I wonder if we could hack it a bit to apply both files in the same MC. When we iterate through the new files it will see both instances as separate. Then we can delete both files in one transaction as well, so we go from 4 reboots to 2.

That said this hack will likely not play nicely with ign spec3.

yuqi-zhang: As in, we will see the same functionality being tested whether we update the file through 2 MCs or 1 MC

runcom:
/hold

no way we can keep adding e2e like this indeed, I'll write a unit that fails for this and work from there - unit should still cover the bug - and we may think to move other simple e2e to unit when possible (given that e2e are still a top priority in these scenarios)

PR #1593, opened 2020-03-26
Bug 1814397: fix wrongful backup of files not originally on the system

yuqi-zhang:
Adds to https://github.com/openshift/machine-config-operator/pull/1586

Add a check for backwards compatibility: basically if the file
doesn't exist in /usr/etc/ and no rpm is claming it, we assume
that the orig file came from a wrongful backup of a MachineConfig
file instead of a RHCOS file.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -782,20 +782,33 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
-			if _, err := os.Stat(noOrigFileName(f.Path)); err == nil {
-				if err := os.Remove(noOrigFileName(f.Path)); err != nil {
-					return errors.Wrapf(err, "deleting no orig file %q: %v", noOrigFileName(f.Path), err)
+			if _, err := os.Stat(noOrigFileStampName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileStampName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting noorig file stamp %q: %v", noOrigFileStampName(f.Path), err)
 				}
 				glog.V(2).Infof("Removing file %q completely", f.Path)
 			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if out, err := exec.Command("cp", "-a", "--reflink=auto", origFileName(f.Path), f.Path).CombinedOutput(); err != nil {
-					return errors.Wrapf(err, "restoring %q from orig file %q: %s", f.Path, origFileName(f.Path), string(out))
+				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc
+				// and no rpm is claiming it, we assume that the orig file came from a wrongful backup of a
+				// MachineConfig file instead of a file originally on disk. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
+				if _, err := exec.Command("rpm", "-qf", f.Path).CombinedOutput(); err != nil {
+					if err := os.Remove(origFileName(f.Path)); err != nil {
+						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+					}
+				} else if _, err := os.Stat("/usr" + f.Path); strings.HasPrefix(f.Path, "/etc") && os.IsNotExist(err) {
+					if err := os.Remove(origFileName(f.Path)); err != nil {
+						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+					}
+				} else {
+					if out, err := exec.Command("cp", "-a", "--reflink=auto", origFileName(f.Path), f.Path).CombinedOutput(); err != nil {
+						return errors.Wrapf(err, "restoring %q from orig file %q: %s", f.Path, origFileName(f.Path), string(out))
+					}
+					if err := os.Remove(origFileName(f.Path)); err != nil {
+						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+					}
+					glog.V(2).Infof("Restored file %q", f.Path)
+					continue
 				}
-				if err := os.Remove(origFileName(f.Path)); err != nil {
-					return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
-				}
-				glog.V(2).Infof("Restored file %q", f.Path)
-				continue
 			}
 			glog.V(2).Infof("Deleting stale config file: %s", f.Path)
 			if err := os.Remove(f.Path); err != nil {
@@ -1013,23 +1026,30 @@ func origFileName(fpath string) string {
 	return filepath.Join(origParentDir(), fpath+".mcdorig")
 }
 
-func noOrigFileName(fpath string) string {
+// We use this to create a file that indicates that no original file existed on disk
+// when we write a file via a MachineConfig. Otherwise the MCD does not differentiate
+// between "a file existed due to a previous machineconfig" vs "a file existed on disk
+// before the MCD took over". Also see deleteStaleData() above.
+//
+// The "stamp" part of the name indicates it is not an actual backup file, just an
+// empty file to indicate lack of previous existence.
+func noOrigFileStampName(fpath string) string {
 	return filepath.Join(noOrigParentDir(), fpath+".mcdnoorig")
 }
 
 func createOrigFile(fpath string) error {
-	if _, err := os.Stat(noOrigFileName(fpath)); err == nil {
+	if _, err := os.Stat(noOrigFileStampName(fpath)); err == nil {
 		// we already created the no orig file for this default file
 		return nil
 	}
-	if _, err := os.Stat(fpath); err != nil {
-		// create a noorig file that tells the MCD that the file wasn't present on disk in RHCOS
-		// so it can just remove it when deleting stale data, as opposed as restoring a file
-		// that was shipped _with_ RHCOS (e.g. a default chrony config).
-		if err := os.MkdirAll(filepath.Dir(noOrigFileName(fpath)), 0755); err != nil {
+	if _, err := os.Stat(fpath); os.IsNotExist(err) {
+		// create a noorig file that tells the MCD that the file wasn't present on disk before MCD
+		// took over so it can just remove it when deleting stale data, as opposed as restoring a file
+		// that was shipped _with_ the underlying OS (e.g. a default chrony config).
+		if err := os.MkdirAll(filepath.Dir(noOrigFileStampName(fpath)), 0755); err != nil {
 			return errors.Wrapf(err, "creating no orig parent dir: %v", err)
 		}
-		return writeFileAtomicallyWithDefaults(noOrigFileName(fpath), nil)
+		return writeFileAtomicallyWithDefaults(noOrigFileStampName(fpath), nil)
 	}
 	if _, err := os.Stat(origFileName(fpath)); err == nil {
 		// the orig file is already there and we avoid creating a new one to preserve the real default
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -782,9 +782,17 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
-			if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -788,6 +788,25 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 				}
 				glog.V(2).Infof("Removing file %q completely", f.Path)
 			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc
+				// and no rpm is claming it, we assume that the orig file came from a wrongful backup of a
+				// MachineConfig file instead of a RHCOS file. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
+				if strings.HasPrefix(origFileName(f.Path), "/etc") {
+					if _, err := os.Stat("/usr" + origFileName(f.Path)); err != nil {
```
runcom: should this be just `f.Path`? `/usr/etc` contains paths _in_ `/etc` whether `origFileName` turns `f.Path` into `/etc/machine-config-daemon/orig/f.Name` which won't live by default under `/usr/etc` which stores only RHCOS default

pkg/daemon/update.go
```diff
@@ -788,6 +788,25 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
 				glog.V(2).Infof("Removing file %q completely", f.Path)
 			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc
+				// and no rpm is claming it, we assume that the orig file came from a wrongful backup of a
+				// MachineConfig file instead of a RHCOS file. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
+				if _, err := exec.Command("rpm", "-qf", f.Path).CombinedOutput(); err != nil {
+					if err := os.Remove(origFileName(f.Path)); err != nil {
+						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+					}
+					continue
```
runcom: instead of continuing, should we go ahead at the routine that deletes the file (at https://github.com/openshift/machine-config-operator/pull/1593/commits/7a68edd2d2b87ea481fde86f7c3a5c849565bea4#diff-06961b075f1753956d802ba954d2cfb5R819)

pkg/daemon/update.go
```diff
@@ -788,6 +788,25 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
+				}
+
+				if strings.HasPrefix(f.Path, "/etc") {
+					if _, err := os.Stat("/usr" + f.Path); os.IsNotExist(err) {
+						if err := os.Remove(origFileName(f.Path)); err != nil {
+							return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+						}
+						continue
```
runcom: same q as https://github.com/openshift/machine-config-operator/pull/1593/commits/7a68edd2d2b87ea481fde86f7c3a5c849565bea4#r399142540

runcom:
manually upgraded from 4.2 to 4.4 (with this patch) and it worked just fine, I'm gonna test all the way from 4.1 to fully verify this, otherwise, this replaces https://github.com/openshift/machine-config-operator/pull/1586

/approve

runcom:
The additional patches here are also backported already to https://github.com/openshift/machine-config-operator/pull/1588 - we'd need a 4.3 patch and 4.2 patch anyway but those can wait till we unblock 4.4 as this patch effectively fixes the upgrade.

yuqi-zhang:
Ah, thanks for the fix! Feel free to squash the commits together

pkg/daemon/update.go
```diff
@@ -782,12 +782,35 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
-			if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
+			if _, err := os.Stat(noOrigFileName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting no orig file %q: %v", noOrigFileName(f.Path), err)
```
kikisdeliveryservice: this error message seems confusing to a user

pkg/daemon/update.go
```diff
@@ -997,16 +1020,32 @@ func origParentDir() string {
...
 
+func noOrigParentDir() string {
+	return filepath.Join("/etc", "machine-config-daemon", "noorig")
+}
+
 func origFileName(fpath string) string {
 	return filepath.Join(origParentDir(), fpath+".mcdorig")
 }
 
+func noOrigFileName(fpath string) string {
```
kikisdeliveryservice: i dont really understand this naming?  Can we make it clearer since it's not really clear what's going on here... and the naming of this noOrig doesn't really make it clear

kikisdeliveryservice: ie anyone who will have to maintain this in the future is going to have trouble understanding what this is for...

pkg/daemon/update.go
kikisdeliveryservice: To be clear I'm having trouble understanding what this noOrigFile.... stuff is doing.

yuqi-zhang: `noOrig` just means that there was no original file on disk when we tried to write a file via the MCD. This tells the MCD to not revert a file when the last instance of it is deleted in MachineConfigs.

The reasoning is added to `createOrigFile` function. It's also in the commit message. Where else would you like me to add some comments? Or how else would you like it to be named

pkg/daemon/update.go
yuqi-zhang: I can add some comments above this as well. To be clear this is to unblock upgrades for 4.4 and not the most clean long term solution. We are attempting to fix an insidious error that we didn't catch way back

pkg/daemon/update.go
kikisdeliveryservice: maybe noOrigFileFlag or noOrigFileStamp?

yuqi-zhang:
Pushed new commit:
Squashed fixes
Added comments and changed naming for clarity
Add new fix for files that exist in `/etc`

Manually tested to be good for both new and existing scenarios

pkg/daemon/update.go
```diff
@@ -782,12 +782,33 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
 
 	for _, f := range oldConfig.Spec.Config.Storage.Files {
 		if _, ok := newFileSet[f.Path]; !ok {
-			if _, err := os.Stat(origFileName(f.Path)); err == nil {
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
+			if _, err := os.Stat(noOrigFileStampName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileStampName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting no orig file %q: %v", noOrigFileStampName(f.Path), err)
```
kikisdeliveryservice: could you update error message here?

yuqi-zhang:
Modified error msg

Now I think about it technically I think the logic is wrong.

/hold

yuqi-zhang:
/hold cancel

After some more thought I think this is alright. To be clear if we have an `.orig` file, the conditions it will restore is that it must both be a) managed by an RPM and b) exist in the original ostree (if it was in `/etc`).

Is there a scenario where we have an rpm drop in a file in `/etc` but it would not exist in `/usr/etc`? Conversely, is there a scenario where it exists in `/usr/etc` but no rpm is claiming it?

runcom:
> Is there a scenario where we have an rpm drop in a file in `/etc` but it would not exist in `/usr/etc`? 

I think this is only possible if you install an rpm after initial bootstrap + MCO right?

> Conversely, is there a scenario where it exists in `/usr/etc` but no rpm is claiming it?

I think yes, say you installed an RPM with a spec that doesn't claim ownership of a file (it happens a lot, trust me...) but still ships the file under `/etc`

yuqi-zhang:
I can confirm this passes the tests to be added in https://github.com/openshift/machine-config-operator/pull/1590

I can also add a test to ensure existing `.orig`s are properly pruned as a separate PR. For now I have done manual testing and can confirm this should fix those cases.

runcom:
Tested once more an upgrade from 4.2 up to this PR in 4.4 (https://github.com/openshift/machine-config-operator/pull/1588) and the upgrade succeded and all files were correctly pruned.

/lgtm

pkg/daemon/update.go
```diff
@@ -782,20 +782,33 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
-				if err := os.Remove(origFileName(f.Path)); err != nil {
-					return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc
+				// and no rpm is claming it, we assume that the orig file came from a wrongful backup of a
+				// MachineConfig file instead of a RHCOS file. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
+				if _, err := exec.Command("rpm", "-qf", f.Path).CombinedOutput(); err != nil {
+					if err := os.Remove(origFileName(f.Path)); err != nil {
+						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
+					}
+				} else if _, err := os.Stat("/usr" + f.Path); strings.HasPrefix(f.Path, "/etc") && os.IsNotExist(err) {
```
sinnykumari: I think we should keep /usr/etc/ file content checks only for OSTree based system. On a node containing traditional OS, `else` section would never run because it doesn't have etc content in /usr/etc/ as well.

runcom: right, tho, on normal system /usr/etc isn't containing the base /etc - I agree tho that we need a check just in case of RHCOS-ostree nodes to avoid RHEL hosts misbehaving (remote option but still)

runcom:
/hold

pending Sinny's review

pkg/daemon/update.go
```diff
@@ -997,16 +1018,39 @@ func origParentDir() string {
 	return filepath.Join("/etc", "machine-config-daemon", "orig")
 }
 
+func noOrigParentDir() string {
```
sinnykumari: It will be nice to have description of what this function does.

pkg/daemon/update.go
```diff
@@ -782,12 +782,33 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
-				if err := os.Rename(origFileName(f.Path), f.Path); err != nil {
-					return err
+			if _, err := os.Stat(noOrigFileStampName(f.Path)); err == nil {
+				if err := os.Remove(noOrigFileStampName(f.Path)); err != nil {
+					return errors.Wrapf(err, "deleting noorig file stamp %q: %v", noOrigFileStampName(f.Path), err)
+				}
+				glog.V(2).Infof("Removing file %q completely", f.Path)
+			} else if _, err := os.Stat(origFileName(f.Path)); err == nil {
+				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc
+				// and no rpm is claming it, we assume that the orig file came from a wrongful backup of a
```
sinnykumari: typo s/claming/claiming/

pkg/daemon/update.go
```diff
@@ -782,12 +782,33 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
+				// MachineConfig file instead of a RHCOS file. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
```
sinnykumari: Since [FCOS support has been added](https://github.com/openshift/machine-config-operator/pull/1496), saying CoreOS instead of RHCOS would be more generic

pkg/daemon/update.go
```diff
@@ -997,16 +1018,39 @@ func origParentDir() string {
...
 func createOrigFile(fpath string) error {
-	if _, err := os.Stat(fpath); err != nil {
-		// the file isn't there, no need to back it up
-		// we could check ENOENT only maybe?
+	if _, err := os.Stat(noOrigFileStampName(fpath)); err == nil {
+		// we already created the no orig file for this default file
 		return nil
 	}
+	if _, err := os.Stat(fpath); os.IsNotExist(err) {
+		// create a noorig file that tells the MCD that the file wasn't present on disk in RHCOS
```
sinnykumari: similar to previous comment
s/RHCOS/CoreOS

yuqi-zhang:
Added extra check per comment and fixed up some typos

sinnykumari:
Looks sane to me
/hold cancel
/approve

runcom:
uhm, wondering if something broke the install but weird it's on for e2e-aws

PR #1594, opened 2020-03-26
templates: Use nodeip-finder from baremetalRuntimeCfgImage

LorbusChris:
Another step towards making MCO support running on top of FCOS by replacing Python scripts.

**- What I did**
Replace Python script in Ignition files templates with a containerized nodeip-finder command from baremetalRuntimeCfgImage

**- How to verify it**
CI:
OpenStack e2e
Baremetal e2e

templates: Use nodeip-finder from baremetalRuntimeCfgImage instead of Python scripts.


vrutkovs:
That won't work, iirc nodeip-finder is python3 and the container is rhel7-based

PR #1597, opened 2020-03-27
Bug 1817236: baremetal: Handle unused route fields in non_virtual_ip

cybertron:
In my deployments I am seeing routes that include fields the
non_virtual_ip script can't handle. This causes it to fail and anything
relying on it to function incorrectly.

This change adds a **kwargs parameter to the class constructor so
it will take arbitrary params that we will then ignore. It also
filters out \ characters from the routes because I'm seeing that
as well and we don't want to try to parse it. The \ appears in
multi-line routes that can't be handled correctly by the existing
structure of the class because they have multiple 'via' values and
the class can only handle one per route. However, this is happening
only on the default route in my case, which we ignore anyway, and
this script is being replaced by a Go implementation in the near
future so I don't think it's worth rewriting it to handle multi-line
routes.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/baremetal-non-virtual-ip.yaml
+++ b/templates/common/baremetal/files/baremetal-non-virtual-ip.yaml
@@ -67,7 +67,7 @@ contents:
 
 
     class V6Route:
-        def __init__(self, destination: str, dev: Optional[str] = None, proto: Optional[str] = None, metric: Optional[int] = None, pref: Optional[str] = None, via: Optional[str] = None) -> None:
+        def __init__(self, destination: str, dev: Optional[str] = None, proto: Optional[str] = None, metric: Optional[int] = None, pref: Optional[str] = None, via: Optional[str] = None, **kwargs) -> None:
              self.destination: str = destination
              self.via: Optional[str] = via
              self.dev: Optional[str] = dev
@@ -77,7 +77,7 @@ contents:
 
         @classmethod
         def from_line(cls: Type[TR], line: str) -> TR:
-            items = line.split()
+            items = [i for i in line.split() if i != '\\']
             dest = items[0]
             if dest == 'default':
                 dest = '::/0'


cybertron:
Oops, dropped the commit where I did make update.

cybertron:
Actually that's not why it failed: could not initialize namespace: could not create role binding for: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:ci:ci-operator" cannot create resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "ci-op-z62bwfqv"

Will have to look into that, but I'm guessing there's an issue somewhere else.

sinnykumari:
looks sane and tests are passing as well
/approve

templates/common/baremetal/files/baremetal-non-virtual-ip.yaml
```diff
@@ -77,7 +77,7 @@ contents:
 
         @classmethod
         def from_line(cls: Type[TR], line: str) -> TR:
-            items = line.split()
+            items = [i for i in line.split() if i != '\\']
```
celebdor, line 80: Aren't route entries always a single line? Did you encounter a `\\` ? I only ever encountered it in Address processing.

cybertron: Apparently they aren't. I had some masters with a route that looks like: "default proto ra metric 100 \   nexthop via fe80::69e0:8df7:4351:4eb6 dev enp1s0 weight 1 \     nexthop via fe80::f620:b5b3:612a:8446 dev enp1s0 weight 1 \     nexthop via fe80::b449:f968:42ce:eef5 dev enp1s0 weight 1 pref medium"

kikisdeliveryservice:
assign @celebdor

cybertron:
This merged as part of #1601.

PR #1598, opened 2020-03-31
pkg/operator: clear Degraded on task's success

runcom:
The MCO runs a bunch of sync functions during its sync. Each of these functions has
a name (a task name). When we loop through the sync functions and fail on any of them,
the MCO breaks and reports the Degraded status for that task as Reason: TaskNameFailed.
It then goes and retries all the sync functions (no change in behavior till here).
During some debugging and testing I've noticed that the RenderConfig task failed with the
version mismatch error for osImageURL. That is good, and also right. The MCO went ahead
and set the Degraded condition to RenderConfigFailed (correctly!). The MCO then went again
and re-tried all the sync functions. Now the RenderConfig task succeeded (the version mismatch error
can indeed be temporary). What happened from now on is that the MCO kept rolling and eventually finishes
the upgrade I was running but the RenderConfigFailed Degraded status was never cleared back to False
even if the task _did_ succeed (I had to wait for all the sync functions to finish as expected, and it takes a while still reporting degraded).
This patch introduces a new function which we call after every sync function in order to clear
any previous degraded status when it instead succeeded in later syncs.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -163,6 +163,27 @@ const (
 	asExpectedReason = "AsExpected"
 )
 
+func (optr *Operator) clearDegradedStatus(task string) error {
+	co, err := optr.fetchClusterOperator()
+	if err != nil {
+		return err
+	}
+	if co == nil {
+		return nil
+	}
+	if cov1helpers.IsStatusConditionFalse(co.Status.Conditions, configv1.OperatorDegraded) {
+		return nil
+	}
+	degradedStatusCondition := cov1helpers.FindStatusCondition(co.Status.Conditions, configv1.OperatorDegraded)
+	if degradedStatusCondition == nil {
+		return nil
+	}
+	if degradedStatusCondition.Reason != task+"Failed" {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


sinnykumari:
looks sane, will let other MCO folks for second opinion
/approve

pkg/operator/status.go
```diff
@@ -163,6 +163,27 @@ const (
...
+	}
+	if cov1helpers.IsStatusConditionFalse(co.Status.Conditions, configv1.OperatorDegraded) {
+		return nil
+	}
+	degradedStatusCondition := cov1helpers.FindStatusCondition(co.Status.Conditions, configv1.OperatorDegraded)
+	if degradedStatusCondition == nil {
+		return nil
+	}
+	if degradedStatusCondition.Reason != task+"Failed" {
+		return nil
```
kikisdeliveryservice, line 182: is there anyway a legit degraded message will not have Failed at the end?

kikisdeliveryservice: would it be better to just check that task is contained in Reason? (I dunno, I'm wondering?)

PR #1602, opened 2020-04-01
pkg/daemon: Add event for drain failures

kikisdeliveryservice:
As per request by @smarterclayton , adding an event for drain failures.

Lmk if you'd like anything else.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -161,9 +161,11 @@ func (dn *Daemon) drain() error {
 		if err == wait.ErrWaitTimeout {
 			failMsg := fmt.Sprintf("%d tries: %v", backoff.Steps, lastErr)
 			MCDDrainErr.WithLabelValues(failTime, failMsg).SetToCurrentTime()
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)
 			return errors.Wrapf(lastErr, "failed to drain node (%d tries): %v", backoff.Steps, err)
 		}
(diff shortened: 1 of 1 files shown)


smarterclayton:
Is this the only place we call drain?

kikisdeliveryservice:
> Is this the only place we call drain?

Apologies. We call it 2x: in update() and in checkStateInFirstRun()

kikisdeliveryservice:
@runcom PTAL

kikisdeliveryservice:
These jobs shouldn't be failing... Will check it out in the morning if it's not resolved by then..

pkg/daemon/daemon.go
```diff
@@ -927,6 +927,7 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 	if state.pendingConfig != nil && bootID == dn.bootID {
 		dn.logSystem("drain interrupted, retrying")
 		if err := dn.drain(); err != nil {
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", err.Error())
```
runcom, line 930: it'll avoid duplication if you fire that event directly from `dn.drain()` so every time it fails it fires an event

runcom: and if we add another drain call somewhere else in the future it'll be future proof as well w/o forgetting

kikisdeliveryservice: Good point. Done.

kikisdeliveryservice:
:(

```error: could not run steps: step e2e-gcp-op failed: failed to acquire lease: status 503 Service Unavailable, status code 503```

PR #1604, opened 2020-04-01
[baremetal] Update keepalived Liveness check

yboaron:
Update Liveness probe for keepalived container to check also keepalived process existence, 
with this fix if for some reason keepalived process exits - kubelet will automatically restart the container. 

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -80,12 +80,11 @@ contents:
         livenessProbe:
           exec:
             command:
-            - /bin/sh
+            - /bin/bash
             - -c
             - |
-              [[ -s /etc/keepalived/keepalived.conf ]] || \
               kill -s SIGUSR1 "$(pgrep -o keepalived)" && ! grep -q "State = FAULT" /tmp/keepalived.data
-          initialDelaySeconds: 10
+          initialDelaySeconds: 20
         terminationMessagePolicy: FallbackToLogsOnError
         imagePullPolicy: IfNotPresent
       - name: keepalived-monitor


templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -83,7 +83,7 @@ contents:
             - /bin/sh
             - -c
             - |
-              [[ -s /etc/keepalived/keepalived.conf ]] || \
+              [[ -s /etc/keepalived/keepalived.conf ]] && pgrep -o keepalived || \
```
cybertron, line 86: Isn't the problem here that this should have been && instead of ||? With the || here, it will short-circuit the state check below and return success just if keepalived.conf exists and keepalived is running. That doesn't seem like what we want.

Note that the kill command will already fail if the pgrep fails so that covers the new case being added here.

cybertron: If keepalived.conf doesn't exist, then the liveness probe should fail, right? Keepalived is not alive if it isn't configured yet. As I understand it, three things need to be true for keepalived to be "alive":

1) keepalived.conf exists (the -s check)
2) keepalived is running (the pgrep check)
3) keepalived is not reporting an error (the FAULT check)

If any one of those things is not true then keepalived is not alive. That's why I'm thinking all of those checks should be &&.

yboaron: Not exactly,  keepalived container can start before keepalived-monitor rendered cfg file, and I think we shouldn't fail liveness in that case  (it will wait for 'reload' message from monitor container).
In the other case, if cfg file exists, we need to make sure that keepalived is running.
I wanted to change (add !, to the -s if) it to :

`          [[ **!** -s /etc/keepalived/keepalived.conf ]] || \
          kill -s SIGUSR1 "$(pgrep -o keepalived)" && ! grep -q "State = FAULT" /tmp/keepalived.data
`
But seems that kubelet fail on,
**kill -s SIGUSR1 "$(pgrep -o keepalived)"**   , 
I'm getting -
`sh-4.2# kill -s SIGUSR1  "$(pgrep -o keepalived)"
sh: kill: SIGUSR1: invalid signal specification
sh-4.2# exit
`
Could you please verify if `kill -s SIGUSR1 "$(pgrep -o keepalived)` work for you?

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -83,8 +83,8 @@ contents:
             - /bin/sh
```
cybertron, line 83: Oh, this is naughty when we're using [[]], which is a bash-ism that's not present in a plain POSIX shell. This should be /bin/bash.

templates/common/baremetal/files/baremetal-keepalived.yaml
cybertron: But is unconfigured keepalived "alive" for our purposes? The VIPs it is supposed to be providing won't exist so it's effectively non-functional. If the existence of the process is enough to call it alive then I would argue we don't even need the config file check. Also, the flipped logic will mean we report that keepalived is alive just because the config file _doesn't_ exist, which seems wrong. That could mean neither keepalived is running nor the config file has been created, but we'll report the container alive.

I'm going to comment separately on the kill part because I think that's also an existing bug in this script. :-/

I'll leave details of what I found there.

yboaron: Thanks for the bash tip!

for the other case, let's see, we have two options:
1. include cfg file not exist in liveness check + or keepalived is working (the SIGUSR1 thing)
2. check only if  keepalived is working (the SIGUSR1 thing)
The only difference between the two, that for 2, on startup, kubelet may restart (depends on timing ) keepalived container until monitor container rendered the cfg file.
I assume that your preferred option is 2.
 I think that '1' is more suitable because:
A. Our keepalived container should be considered as healthy if it's waiting for monitor container. 
B. we won't have 'unnecessary' restarts for keepalived container. 

But  I guess that option '2' + increasing initialDelaySeconds value will be a good solution.
Thoughts?

cybertron:
/test e2e-metal-ipi

Also testing locally. Hopefully one of these will pass

cybertron:
/lgtm

Both passed. I should go buy a lottery ticket. ;-)

PR #1606, opened 2020-04-01
ovirt: disable tx checksum offload for workers

Gal-Zaidman:
This patch adds the workaround suggested on [1]
to make nodeport work, instead of ethtool we use
NM to apply the fix for each connction before it is up.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1794714

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
+++ b/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
@@ -1,0 +1,10 @@
+filesystem: "root"
+mode: 0744
+path: "/etc/NetworkManager/dispatcher.d/pre-up.d/disable-tx-checksum-offload.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # This is a workaround for BZ#1794714
+    if [[ -e /var/lib/cni/bin/openshift-sdn ]]; then
+      nmcli con modify ${CONNECTION_UUID} ethtool.feature-tx-checksum-ip-generic off;
+    fi


templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
```diff
@@ -0,0 +1,8 @@
+filesystem: "root"
+mode: 0744
+path: "/etc/NetworkManager/dispatcher.d/pre-up.d/disable-tx-checksum-offload"
+contents:
+  inline: |
+    # This is a workaround for BZ#1794714
```
Gal-Zaidman, line 6: Is there a way to check what is the default network, OpenshiftSDN or OVNkubernetes?
We use OpenshiftSDN by default and the fix is only relevant for OpenshiftSDN.

rgolangh: I think the existence of `vxlan` interface means OpenShiftSDN. If that's true then all you need
is to check $1 name and only if its starts with `vxlan` then apply the change. 
We need to make sure this is really the case.

rgolangh: or more reliably, `nmcli d show $1| awk '/GENERAL.TYPE:.*/ {print $2}'`

templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
```diff
@@ -0,0 +1,8 @@
+filesystem: "root"
+mode: 0744
+path: "/etc/NetworkManager/dispatcher.d/pre-up.d/disable-tx-checksum-offload"
```
rgolangh, line 3: add `.sh` to the script name

templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
```diff
@@ -0,0 +1,8 @@
...
+contents:
+  inline: |
```
rgolangh, line 5: add shebang line

rgolangh:
also, use `/test e2e-ovirt` when you think its ready

templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
Gal-Zaidman: I'll check it locally

templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
Gal-Zaidman: added

cgwalters:
So this is an openshift-sdn bug that only seems to manifest on oVirt, is that right?  What's the timeframe for the SDN fix landing (or rephrased, when can we remove this and how do we track things to ensure it's removed?)

rgolangh:
I suggest to open a bug which depends on 1794714, for the removal of this machine-config item.

Gal-Zaidman:
> So this is an openshift-sdn bug that only seems to manifest on oVirt, is that right?

On ovirt all other platforms according to [comment#8](https://bugzilla.redhat.com/show_bug.cgi?id=1794714#c8)

>What's the timeframe for the SDN fix landing (or rephrased, when can we remove this and how do we track things to ensure it's removed?)

Don't know what is the timeframe but as you can see it is currently handled by RHEL and on high priority

templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
Gal-Zaidman: I checked and the device which is passed is the ethernet dev ens3, I think it should be available on both networks but I will check.
I'll ask the openshift network team what is the best way to distinguish between them

Gal-Zaidman: I've checked and we can check if there is a file /var/lib/cni/bin/openshift-sdn that should do it

Gal-Zaidman:
> I suggest to open a bug which depends on 1794714, for the removal of this machine-config item.
created https://bugzilla.redhat.com/show_bug.cgi?id=1820305

cgwalters:
/approve
because there's a promise/mechanism to ensure this workaround is dropped when it's not necessary anymore.

mandre:
We're likely going to do the same for OpenStack.

kikisdeliveryservice:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1821005

PR #1612, opened 2020-04-03
Bug 1820401: daemon: better way to find installed kernel-rt packages on host

sinnykumari:
Initially we were relying on fixed set of kernel-rt
sub-packages which will be installed on the system. This can cause
issues when machine-os-content ships additional kernel-rt
sub-packages.

To solve this, we now query host with
`rpm-ostree status --json` and look for requested-local-packages
in the 0th index from available deployments list.

Fixes bug - https://bugzilla.redhat.com/show_bug.cgi?id=1820401

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -50,6 +50,38 @@ const (
 	realtimeKernelType = "realtime"
 )
 
+// HostInfo contains information of an OSTree based system
+type HostInfo struct {
+	Deployments []Deployment `json:"deployments"`
+}
+
+// Deployment contains information about a particular OSTree deployment
+type Deployment struct {
+	RequestedLocalPkgs []string `json:"requested-local-packages"`
+}
+
+func installedRTKernelRpmsOnHost() ([]string, error) {
+	var out []byte
+	var err error
+	var rtKernelRpms = []string{}
+	if out, err = exec.Command("rpm-ostree", "status", "--json").Output(); err != nil {
+		return rtKernelRpms, fmt.Errorf("Failed to execute rpm-ostre status --json %v", err)
+	}
+
+	var rpms HostInfo
+	if err := json.Unmarshal(out, &rpms); err != nil {
+		return rtKernelRpms, err
+	}
+
+	rtRegex := regexp.MustCompile("kernel-rt-.*")
+	for _, localPkg := range rpms.Deployments[0].RequestedLocalPkgs {
+		if rtRegex.MatchString(localPkg) {
+			rtKernelRpms = append(rtKernelRpms, localPkg)
+		}
+	}
+	return rtKernelRpms, nil
+}
+
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
 	return writeFileAtomically(fpath, b, defaultDirectoryPermissions, defaultFilePermissions, -1, -1)
 }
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -50,6 +50,39 @@ const (
...
+type Deployment struct {
+	RequestedLocalPkgs []string `json:"requested-local-packages"`
+}
+
+func installedRTKernelRpmsOnHost() ([]string, error) {
+	var out []byte
+	var err error
+	var rtKernelRpms = []string{}
+	if out, err = exec.Command("rpm-ostree", "status", "--json").Output(); err != nil {
+		wrappedErr := fmt.Errorf("Failed to execute rpm-ostre status --json %v", err)
```
runcom, line 68: You can directly return the fmt.Errorf below without the extra wrapperErr variable

runcom:
Just a nit @cgwalters ptal as well if you have some time

/approve

pkg/daemon/update.go
sinnykumari: Right, fixed it.

pkg/daemon/update.go
```diff
@@ -715,11 +755,21 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
 
 	if canonicalizeKernelType(oldConfig.Spec.KernelType) == realtimeKernelType && canonicalizeKernelType(newConfig.Spec.KernelType) == realtimeKernelType {
 		if oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL {
+			var installedRTKernelRpms []string
+			var err error
 			args = []string{"uninstall"}
-			args = append(args, rtKernel...)
+			if installedRTKernelRpms, err = installedRTKernelRpmsOnHost(); err != nil {
+				return fmt.Errorf("Error wihle fetching installed RT kernel on host %v", err)
```
ashcrow, line 762: `s|wihle|while|`

sinnykumari: damm type, fixed

pkg/daemon/update.go
```diff
@@ -50,6 +50,38 @@ const (
...
+	}
+
+	var rpms HostInfo
+	if err := json.Unmarshal(out, &rpms); err != nil {
+		return rtKernelRpms, err
+	}
+
+	rtRegex := regexp.MustCompile("kernel-rt-.*")
+	for _, localPkg := range rpms.Deployments[0].RequestedLocalPkgs {
+		if rtRegex.MatchString(localPkg) {
```
cgwalters, line 78: This could just be `strings.HasPrefix(localPkg, "kernel-rt-")`.  But fine as is too.

pkg/daemon/update.go
```diff
@@ -50,6 +50,38 @@ const (
 	realtimeKernelType = "realtime"
 )
 
+// HostInfo contains information of an OSTree based system
+type HostInfo struct {
```
cgwalters, line 54: We already have `rpmOstreeState` in `rpm-ostree.go`; would be good to share code there.

sinnykumari:
Thanks Colin for review! Agree with you to unblock the CI test first. Will address the review comments in follow-up PR

sinnykumari:
ugh, these failing ci tests

runcom:
/lgtm

let's address Colin's address in a follow up!

runcom:
maybe...

/lgtm

sinnykumari:
Checked into live cluster from e2e-gcp-op CI test run, kernelType test went fine. e2-gcp-op test should finish fine

PR #1614, opened 2020-04-04
Bug 1820822: sdn: ignore new ovn-kubernetes OVS internal port names ovn-k8s-gw0 and ovn-k8s-mp0

dcbw:
To prevent racing with NetworkManager's OVS plugin resulting in IP addresses sometimes being cleared from the OVS internal ports that openshift-sdn and ovn-kubernetes set up, we must tell NM to ignore those ports.

@trozet @danwinship @squeed @openshift/networking

### DIFF
--- a/templates/common/_base/files/nm-ignore-sdn.yaml
+++ b/templates/common/_base/files/nm-ignore-sdn.yaml
@@ -5,5 +5,5 @@ contents:
   inline: |
     # ignore known SDN-managed devices
     [device]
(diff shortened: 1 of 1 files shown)


cgwalters:
/lgtm
That said, I still think we should do this in a better way; something like an extended attribute in `/sysfs` or something that tells the host network software to ignore the interface because it's going to be "owned" by something else.

dcbw:
> That said, I still think we should do this in a better way; something like an extended attribute in `/sysfs` or something that tells the host network software to ignore the interface because it's going to be "owned" by something else.

Yeah, that's a larger conversation and it's been discussed upstream in the kernel periodically. But AFAIK we don't have that kind of "owner" tag for interfaces yet.

PR #1621, opened 2020-04-06
Bug 1821005: oVirt, fix disable tx checksum offload for workers

Gal-Zaidman:
On PR [1] we added a workaround for Bug [2], this
fails when the worker starts for the first time since openshift-sdn
is created only when the sdn pod is starting.
Instead we will disable by default and enable tx-checksum offload\
only when running with OVNkubernetes

[1] On PR https://github.com/openshift/machine-config-operator/pull/1606,
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1794714

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
+++ b/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
@@ -5,6 +5,6 @@ contents:
   inline: |
     #!/bin/bash
     # This is a workaround for BZ#1794714
-    if [[ -e /var/lib/cni/bin/openshift-sdn ]]; then
+    if [[ ! -e /var/lib/cni/bin/ovn-k8s-cni-overlay ]]; then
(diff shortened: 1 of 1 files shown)


templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
```diff
@@ -5,6 +5,8 @@ contents:
   inline: |
     #!/bin/bash
     # This is a workaround for BZ#1794714
-    if [[ -e /var/lib/cni/bin/openshift-sdn ]]; then
+    if [[ -e /var/lib/cni/bin/ovn-k8s-cni-overlay ]]; then
```
rgolangh, line 8: I guess simplify to ! -e path then off

Gal-Zaidman:
Tested manually by logging into the machines and checking the connection

Gal-Zaidman:
/retitle Bug 1821005: oVirt, fix disable tx checksum offload for workers

PR #1626, opened 2020-04-07
RFC: Reboot whitelisting

ondrejmular:
This is a PoC implementation of https://github.com/openshift/enhancements/pull/159

The main goal of this PR is to get feedback on a mechanism responsible for getting files/units changes and running defined actions to avoid reboots. It was only tested manually and automated tests will be added once there is an agreement that this implementation is acceptable.
Filter defined in the code is (obviously) just a placeholder for testing and the final filter definition still needs to be agreed on.


ondrejmular:
cc @beekhof

beekhof:
Replaced by https://github.com/openshift/machine-config-operator/pull/1922

/close

PR #1628, opened 2020-04-07
Bug 1820975: oVirt, fix disable tx checksum offload for workers

Gal-Zaidman:
We need to disable tx checksum offload on worker nodes due to BZ#1794714.
This is currenly a workaround and is tracked for removal on BZ#18.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
+++ b/templates/worker/00-worker/ovirt/files/ovirt-disable-tx-checksum-offload.yaml
@@ -1,0 +1,10 @@
+filesystem: "root"
+mode: 0744
+path: "/etc/NetworkManager/dispatcher.d/pre-up.d/disable-tx-checksum-offload.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # This is a workaround for BZ#1794714
+    if [[ ! -e /var/lib/cni/bin/ovn-k8s-cni-overlay ]]; then
+      nmcli con modify ${CONNECTION_UUID} ethtool.feature-tx-checksum-ip-generic off;
+    fi


Gal-Zaidman:
/retitle Bug 1820975: oVirt, fix disable tx checksum offload for workers

ashcrow:
Is this a backport or was it fixes differently in `master`?

yuqi-zhang:
/hold

So generally the proper way is to fix it in 4.5 first by making a PR against master, and once that merges and passes QA, have the bot backport to 4.4.

I see you instead duplicated the bug for 4.5, and hand modified that to "Verified" to pass the bot check. I am not sure if you should be doing that unless this is a critical urgent fix. For now please open a PR against 4.5 first (unless this is only something that needs to be fixed for 4.4?).

kikisdeliveryservice:
This was already fixed in master via: https://github.com/openshift/machine-config-operator/pull/1606

ashcrow:
> This was already fixed in master via: #1606

@kikisdeliveryservice's right -- there is a fix in master. The BZ being invalid is fine as it should be invalid until the original PR is verified by QE.

ashcrow:
The bugs should be in the right state now:
- 1821005: ON_QA
- 1820975: POST

Once 1821005 is verified by QE then this can be refreshed and move forward.

yuqi-zhang:
Oh I see, the context is in https://github.com/openshift/machine-config-operator/pull/1617
/hold cancel

sdodson:
Why are we only doing this for workers? In compact clusters (no workers) masters would also need this, no?

ashcrow:
@Gal-Zaidman can you answer @sdodson's question above?

Gal-Zaidman:
> Why are we only doing this for workers? In compact clusters (no workers) masters would also need this, no?

This fix is to make NodePort work.
NodePort is used to exposes services which should run on worker nodes.
I know that you can use NodePort with masters as well but this is not the "usual/recommended use case" and disabling checksum by default is a workaround and not a solution.
So we decided to apply this only to worker nodes

PR #1629, opened 2020-04-07
Bug 1820785: [baremetal] Correctly handle both ipv4 and ipv6 for ingress records

cybertron:
The template plugin entries for the ingress *.apps addresses were
hard-coded for ipv4 A records, but we need to also be able to handle
ipv6 AAAA records. Otherwise queries for those addresses in ipv6
environments will fail because the values returned are not valid
ipv4.

This change makes the plugin handle any type of query for those and
templates the type of the query into the response so A queries will
return A records and AAAA queries will return AAAA records.

Closes: #1820785

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -10,9 +10,9 @@
         {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} api.{{ .ControllerConfig.EtcdDiscoveryDomain }}
         fallthrough
     }
-    template IN A {{ .ControllerConfig.EtcdDiscoveryDomain }} {
+    template IN ANY {{ .ControllerConfig.EtcdDiscoveryDomain }} {
         match .*.apps.{{ .ControllerConfig.EtcdDiscoveryDomain }}
-        answer "{{`{{"{{ .Name }}"}}`}} 60 in a {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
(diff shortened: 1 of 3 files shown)


manifests/baremetal/coredns-corefile.tmpl
```diff
@@ -10,9 +10,9 @@
         {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} api.{{ .ControllerConfig.EtcdDiscoveryDomain }}
         fallthrough
     }
-    template IN A {{ .ControllerConfig.EtcdDiscoveryDomain }} {
+    template IN ANY {{ .ControllerConfig.EtcdDiscoveryDomain }} {
         match .*.apps.{{ .ControllerConfig.EtcdDiscoveryDomain }}
-        answer "{{`{{"{{ .Name }}"}}`}} 60 in a {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
+        answer "{{`{{"{{ .Name }}"}}`}} 60 in {{`{{"{{ .Type }}"}}`}} {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
```
celebdor, line 15: Will using ANY and .Type here not mean that when we get an A request, we will answer with A but with an IPv6 address?

cybertron: Yes, but at least when we get the right type of request we'll return the right type. Making this return an AAAA record when an A record was requested is going to be a lot trickier. We'd have to specify the type based on the ip version of the ingress VIP, and I don't think the templating language has a way to do that built in. We'd have to write an external function that did the version detection.

celebdor: We should use runtimecfg to detect the IP family we are using for the VIPs and set the record type accordingly.

cybertron:
I've pushed https://github.com/openshift/machine-config-operator/pull/1694 which makes all of these records behave correctly on either ipv4 or ipv6. We shouldn't need this one now.

PR #1631, opened 2020-04-08
Bug 1821888: controller: do not error on empty Ignition configs

yuqi-zhang:
With the introduction of rawExtension for Ignition config objects
via https://github.com/openshift/machine-config-operator/pull/996,
we also now use ignition.Parse directly to process the validity of
ignition configs. This will cause machineconfigs without a Ignition
section to fail, which we don't want.

Also modify some error messages for clarity.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -28,10 +28,16 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 
 	var fips bool
 	var kernelType string
+	var outIgn igntypes.Config
 
-	outIgn, report, err := ign.Parse(configs[0].Spec.Config.Raw)
-	if err != nil {
-		return nil, errors.Errorf("parsing Ignition config failed with error: %v\nReport: %v", err, report)
+	if configs[0].Spec.Config.Raw == nil {
+		outIgn = igntypes.Config{}
+	} else {
+		parsedIgn, report, err := ign.Parse(configs[0].Spec.Config.Raw)
+		if err != nil {
+			return nil, errors.Errorf("parsing Ignition config failed with error: %v\nReport: %v", err, report)
+		}
+		outIgn = parsedIgn
 	}
 
 	for idx := 1; idx < len(configs); idx++ {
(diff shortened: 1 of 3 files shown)


yuqi-zhang:
Right now any machineconfig we apply, during or after bootstrap, without an ignition section fails since the parse errors with a nil ignition config

yuqi-zhang:
There are also other locations where we introduced ign.Parse directly but those paths shouldn't be affected like this

yuqi-zhang:
/test e2e-gcp-rt-4.5

I wonder if this works...

kikisdeliveryservice:
Can we add a test for these types of configs to your e2e?

yuqi-zhang:
Will take a look at adding a test to catch this regression. Alternatively I can also modify existing tests to have empty ignition sections to catch this

kikisdeliveryservice:
> Will take a look at adding a test to catch this regression. Alternatively I can also modify existing tests to have empty ignition sections to catch this

If I'm understanding correctly, you should be able to just mod this test: https://github.com/openshift/machine-config-operator/blob/219a942746ea04617729c708baed5d2c7dcb2716/test/e2e/mcd_test.go#L229

yuqi-zhang:
Discussed with Kirsten, the test is harder to fix than that because:

Today our tests uses this clientset function to create a machineconfig to post to the server: https://github.com/openshift/machine-config-operator/blob/master/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1/machineconfig.go#L90
this requires a NON-EMPTY config (ignition) section, e.g. what we do here: https://github.com/openshift/machine-config-operator/blob/master/test/e2e/mcd_test.go#L243 which creates a ignition section with just version field. User applications of empty configs sections, however, is perfectly fine. So until we modify that behaviour we are stuck unable to test this case.

So for now we can merge this PR (I manually tested on GCP to fix). I will open another ticket regarding the test discrepancy

kikisdeliveryservice:
I'm swamped.. Can someone else test this pre-merge?

kikisdeliveryservice:
im spinning up cluster just in case..
hold to verify
/hold

kikisdeliveryservice:
I took a look a the gcp-upgrade failure was unrelated to this PR. Will just to just retest..

kikisdeliveryservice:
No degraded pools or errors in mc* logs.

Will just let retest bot handle this

PR #1637, opened 2020-04-09
Bug 1809007: [ctrcfg controller] Use a struct array instead of map when creating new ignitions

umohnani8:
Fixes #1809007 (https://bugzilla.redhat.com/show_bug.cgi?id=1809007)

**- What I did**
A map doesn't guarantee order when we are creating new ignitions.
When we update the image CR with blocked registries, the ctrcfg
controller needs to update two files registries.conf and policy.json.
Since we get an update from the image CR about every 20 mins, we compare
the semantics to see if anything has changed. But since the order is not
guaranteed, the controller might think that the semantics is not equal
even if nothing in the data changed. Hence another MC is created, and
everytime we get an update the MC applied to the nodes keeps flipping
back and forth for the 2 possible orders causing the nodes to reboot a
bunch of times. So move to using a struct array to ensure the order is
always the same and we don't have two similar MCs being created.

**- How to verify it**
Update the image.config.openshift.io CR with blocked registries and watch to ensure
that 2 MCs are not created after a few mins or hours.

Use a struct array instead of map when creating new ignitions for the container
runtime config controller.

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -552,9 +552,9 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 				}
 			}
 
-			ctrRuntimeConfigIgn := createNewIgnition(map[string][]byte{
-				storageConfigPath: storageTOML,
-				crioConfigPath:    crioTOML,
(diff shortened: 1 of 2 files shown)


umohnani8:
Will need to cherry-pick to 4.4. and 4.3 - I have cloned the bug for these.

umohnani8:
@runcom @mrunalp @haircommander @kikisdeliveryservice PTAL

PR #1636, opened 2020-04-09
Start openvswitch and ovsdb-server when network is ovn/ovs

runcom:
Needed by CNO _for now_ - aim is those 2 services enablement are gonna be owned by CNO itself. Left some comments for things I'm not aware of.

Signed-off-by: Antonio Murdaca <runcom@linux.com>

PR #1639, opened 2020-04-10
DO NOT MERGE: Start openvswitch and ovsdb-server when network is ovn/ovs

smarterclayton:
Tweaking #1636, not replacing it


runcom:
now fixed in the original PR as well https://github.com/openshift/machine-config-operator/pull/1636

runcom:
we can close this right @smarterclayton

PR #1552, opened 2020-03-10
etcd quorum guard: don't set hostNetwork

vrutkovs:
**- What I did**

* updated `etcd-quorum-guard` deployment to generate readiness script once and use host IP and name to find certificates in defined locations.

* removed `hostNetwork: true` from this deployment.
  This caused all network traffic to be counted as a container traffic, so some pods had > 4MBps network in/out in console

**- How to verify it**

Run setup / upgrade. 
Check network in/out data for etcd-quorum pods

**- Description for the changelog**

PR #1648, opened 2020-04-14
Bug 1823677:etcdquorumguard_deployment: pass NSS_SDB_USE_CACHE=no to curl

vrutkovs:
**- What I did**
Updated the script created by etcd quroum guard deployment to run curl with `NSS_SDB_USE_CACHE=no`

**- How to verify it**
Observe etcd-quorum-guard memory consumption

Fixed an issue in 4.5/4.4 which caused memory consumption of readyness probes in etcd-quorum-guard

Initial fix: #705, regression occurred in #1552 
Ref: https://github.com/openshift/okd/issues/146


hexfusion:
/lgtm

thanks @vrutkovs

PR #1649, opened 2020-04-14
Bug 1823852: pkg/server: disable weak TLS versions

runcom:
Coming from an user request but it makes sense as we (OpenShift) use and control
that port.

It's not fully clear to me if we _can_ drop tls < 1.2 but I'm leaning toward so for security reasons

@cgwalters also ptal

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -47,6 +47,7 @@ func (a *APIServer) Serve() {
 	mcs := &http.Server{
 		Addr:    fmt.Sprintf(":%v", a.port),
 		Handler: a.handler,
+		// We don't want to allow 1.1 as that's old.  This was flagged in a security audit.
 		TLSConfig: &tls.Config{
 			MinVersion: tls.VersionTLS12,
 		},
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -46,6 +47,9 @@ func (a *APIServer) Serve() {
 	mcs := &http.Server{
 		Addr:    fmt.Sprintf(":%v", a.port),
 		Handler: a.handler,
+		TLSConfig: &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		},
 	}
 
 	glog.Infof("Launching server on %s", mcs.Addr)


pkg/server/api.go
```diff
@@ -46,6 +47,9 @@ func (a *APIServer) Serve() {
 	mcs := &http.Server{
 		Addr:    fmt.Sprintf(":%v", a.port),
 		Handler: a.handler,
+		TLSConfig: &tls.Config{
```
cgwalters, line 50:
```suggestion
	// We don't want to allow 1.1 as that's old.  This was flagged in a security audit.
	TLSConfig: &tls.Config{
```

pkg/server/api.go
```diff
@@ -46,6 +47,9 @@ func (a *APIServer) Serve() {
...
+			MinVersion: tls.VersionTLS12,
```
cgwalters, line 51: So 4 years from now, when the world has moved past TLS 1.2 and  Golang 1.X bumps the minimum to TLS 1.3...we'll start *downgrading* things here.  How do we avoid that?

Maybe something like having a `ClusterTlsMinVersion` constant in openshift/api?

runcom: That will impact openshift as a whole I guess but it sounds like future proof enough to me - there are other components running a server (right?)

runcom: I'll create an issue for this so we don't forget

runcom:
This is good to go now, ptal

pkg/server/api.go
```diff
@@ -46,6 +47,10 @@ func (a *APIServer) Serve() {
 	mcs := &http.Server{
 		Addr:    fmt.Sprintf(":%v", a.port),
 		Handler: a.handler,
+	// We don't want to allow 1.1 as that's old.  This was flagged in a security audit.
```
sinnykumari, line 50: gofmt is complaining, perhaps indentation issue. Otherwise looks sane.

runcom: fixed thanks!

cgwalters:
> It's not fully clear to me if we can drop tls < 1.2 but I'm leaning toward so for security reasons

The only thing hitting this endpoint now should be CoreOS; so the minimum version here is really "maximum TLS supported by CoreOS". I suspect in fact we could bump it up to 1.3.

Although, I guess more precisely it's "maximum TLS supported by RHCOS 4.1...since we don't [update bootimages yet](https://github.com/openshift/enhancements/pull/201).

runcom:
> The only thing hitting this endpoint now should be CoreOS; so the minimum version here is really "maximum TLS supported by CoreOS". I suspect in fact we could bump it up to 1.3.
> 
> Although, I guess more precisely it's "maximum TLS supported by RHCOS 4.1...since we don't [update bootimages yet](https://github.com/openshift/enhancements/pull/201).

awesome, that was my impression as well: since we control the other side of the connection, we can safely assume those are going to be the only proto to be used.

Anyway, since TLS 1.2 doesn't seem to be an issue, what if we stick with that? or do you feel strong about moving to 1.3?

runcom:
from all the e2e tests failing tho, I'm not sure if we can do this lol will keep checking.

sinnykumari:
/approve
will defer final review to @cgwalters

sinnykumari:
we can fix in follow-up pr in case we can move to TLS 1.3. Let's go with 1.2 for now.
/lgtm

kikisdeliveryservice:
PR seems to be hitting a lot of prometheus related erors in e2e-aws.. are these flakes/known bugs?

PR #1655, opened 2020-04-15
Bug 1825976: cherrypick etcd-quorum-guard refactoring to 4.4

vrutkovs:
**- What I did**
Cherrypicked #1552 and #1648 to release-4.4

**- How to verify it**
Run install / upgrade. etcd-quorum-guard pods should not fail or leak memory

Cherrypicked etcd-quorum-guard refactoring on 4.4

### DIFF
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -21,7 +21,6 @@ spec:
         name: etcd-quorum-guard
         k8s-app: etcd-quorum-guard
     spec:
-      hostNetwork: true
       affinity:
         podAntiAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


hexfusion:
/retitle Bug 1825967: cherrypick etcd-quorum-guard refactoring to 4.4

hexfusion:
/retitle Bug 1825976: cherrypick etcd-quorum-guard refactoring to 4.4

eparis:
This PR conflates 2 distinct issue.
1. a cosmetic issue about traffic counts that, as best I can tell, hurts nothing
2. an oddity in NSS/kernel negative dentries which does have a negative impact.

Can we at least get this PR to talk about #2 instead of #1?


Also, how does moving from hostNetwork to SDN affect things if the SDN is down? previously I didn't need SDN to get quorum guard running. Now I will. What affects is that going to have? How do we test is this change is detrimental?


After that you may remove the hold.

Please also ping Bob Relyea and Nikos about the upstream nss/kernel issue to get it back on their radar.

vrutkovs:
>a cosmetic issue about traffic counts that, as best I can tell, hurts nothing

Correct

>an oddity in NSS/kernel negative dentries which does have a negative impact.

There is no negative impact of this in 4.4 (there has been in 4.5). `export NSS...` needs to be moved in the script which is being run during health checks.

>Also, how does moving from hostNetwork to SDN affect things if the SDN is down?

etcd-quorum-guard would be restarted if SDN on master is down. If its down then API container is not reachable anyway

>What affects is that going to have? How do we test is this change is detrimental?

I tested it with running minor and major upgrade - SDN gets disrupted and etcd-quorum-guard survives that.

NSS env var is orthogonal to this PR - it merely moves it in the script which runs curl as this part is being refactored

hexfusion:
Given the above, while I see this as a possible improvement in networking metrics moving quorum-guard to SDN. But I feel like we are missing coverage on full understanding on corner cases. With the CI data that exists, we should be able to better understand this.

So let's spend more time in that validation and this can drop into z-stream if we see the net gain.

sdodson:
/bugzilla refresh
(to pick up retargetting to 4.4.z)

runcom:
@hexfusion bz is now valid, pending your hold and approval to get in for the z stream

sdodson:
My understanding is that our intent is to close this and revert the upstream change. As such not giving cherry-pick-approved despite this bug being marked high severity.

hexfusion:
/close

After conversations with @eparis this is not what we want. Essentially this would run quorum-guard checks through SDN.

PR #1656, opened 2020-04-15
pkg/daemon: reuse rpm-ostree code to get currently booted CoreOS host info

sinnykumari:
Make GetBootedDeployment() function from rpm-ostree.go public
so that it can be reused.

Follow-up from PR https://github.com/openshift/machine-config-operator/pull/1612


runcom:
unit failing but lgtm overall 👍 

/approve

sinnykumari:
unit test is passing now!

PR #1658, opened 2020-04-16
Bug 1825948: cloudConfig: read cloud Config from openshift-config-managed/kube-cloud-config ConfigMap

sinnykumari:
cloudConfig is now generated by kube_cloud_config controller for all supported
platforms. Controller generates kube-cloud-config ConfigMap in openshift-config-managed
namespace where cloud.conf key is stored.

Links:
- https://github.com/openshift/enhancements/blob/master/enhancements/installer/aws-custom-region-and-endpoints.md
- https://github.com/openshift/api/pull/599
- https://github.com/openshift/api/pull/621

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -97,6 +97,50 @@ func (optr *Operator) syncAll(syncFuncs []syncFunc) error {
 	return syncErr.err
 }
 
+// Return true if cloud config is required on a platform.
+func isCloudConfigRequired(infra *configv1.Infrastructure) bool {
+	if infra.Spec.CloudConfig.Name != "" {
+		return true
+	}
+	for _, platform := range []configv1.PlatformType{configv1.AzurePlatformType, configv1.BareMetalPlatformType, configv1.GCPPlatformType, configv1.OpenStackPlatformType,
+		configv1.OvirtPlatformType, configv1.VSpherePlatformType} {
+		if platform == infra.Status.PlatformStatus.Type {
+			return true
+		}
+	}
+	return false
+}
+
+// Sync cloud config on supported platform from cloud.conf available in openshift-config-managed/kube-cloud-config ConfigMap.
+func (optr *Operator) syncCloudConfig(spec *mcfgv1.ControllerConfigSpec, infra *configv1.Infrastructure) error {
+	if _, err := optr.clusterCmLister.ConfigMaps("openshift-config-managed").Get("kube-cloud-config"); err != nil {
+		if apierrors.IsNotFound(err) {
+			if isCloudConfigRequired(infra) {
+				// Return error only if cloud config is required, otherwise prceeds further.
+				return fmt.Errorf("%s/%s configmap is required on platform %s but not found: %v",
+					"openshift-config-managed", "kube-cloud-config", infra.Status.PlatformStatus.Type, err)
+			}
+		} else {
+			return err
+		}
+
+	} else {
+		// Read cloud.conf from openshift-config-managed/kube-cloud-config ConfigMap.
+		cc, err := optr.getCloudConfigFromConfigMap("openshift-config-managed", "kube-cloud-config", "cloud.conf")
+		if err != nil {
+			return err
+		}
+
+		spec.CloudProviderConfig = cc
+
+		caCert, err := optr.getCAsFromConfigMap("openshift-config-managed", "kube-cloud-config", "ca-bundle.pem")
+		if err == nil {
+			spec.CloudProviderCAData = caCert
+		}
+	}
+	return nil
+}
+
 func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 	if err := optr.syncCustomResourceDefinitions(); err != nil {
 		return err
@@ -206,18 +250,8 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 	}
 	spec.AdditionalTrustBundle = trustBundle
 
-	// if the cloudConfig is set in infra read the cloud config reference
-	if infra.Spec.CloudConfig.Name != "" {
-		cc, err := optr.getCloudConfigFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, infra.Spec.CloudConfig.Key)
-		if err != nil {
-			return err
-		}
-		spec.CloudProviderConfig = cc
-
-		caCert, err := optr.getCAsFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, "ca-bundle.pem")
-		if err == nil {
-			spec.CloudProviderCAData = caCert
-		}
+	if err := optr.syncCloudConfig(spec, infra); err != nil {
+		return err
 	}
 
 	//TODO: alaypatel07 remove after cluster-etcd-operator deployed via CVO as Managed
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -504,7 +504,7 @@ func cloudConfigFlag(cfg RenderConfig) interface{} {
 	}
 	flag := "--cloud-config=/etc/kubernetes/cloud.conf"
 	switch cfg.Platform {
-	case platformAzure, platformOpenStack:
+	case platformAWS, platformAzure, platformBaremetal, platformGCP, platformOpenStack, platformOvirt, platformVSphere:
 		return flag
 	default:
 		return ""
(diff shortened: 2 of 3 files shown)


pkg/operator/sync.go
```diff
@@ -34,6 +34,16 @@ import (
...
+
+	// Platforms for which cloudConfig is available
+	// See https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml#L42
+	platformAWS       = "aws"
+	platformAzure     = "azure"
+	platformBaremetal = "baremetal"
+	platformGCP       = "gcp"
+	platformOpenStack = "openstack"
+	platformVSphere   = "vsphere"
+	platformOvirt     = "ovirt"
```
runcom, line 46: can we reuse the const recently added instead? by doing ToLower

sinnykumari: Renamed const as per what mentioned at https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L175 . 

Noticed error in [e2e-gcp-op test](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1658/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1845) which says `Unable to apply 0.0.1-2020-04-17-065403: Unsupported platform GCP`.

But yes, I can look at reusing existing const with doing ToLower. Not sure if it can cause any issue later on.

sinnykumari: Reusing available platform now

sinnykumari:
interesting, since 30 mins jobs are at `waiting for importing pipeline:base ...`

pkg/operator/sync.go
runcom: no wait, sorry, I should have explained myself better, I'm talking about these: https://github.com/openshift/api/blob/c6e8c9b6eb9ad7f72fc048212a56a1124447c55e/config/v1/types_infrastructure.go#L75

sinnykumari: ah ok, I understood other way around :D

LorbusChris:
@sinnykumari @runcom this might be a good opportunity to move to using properly typed PlatformType identifiers instead of the untyped strings we're using (and that are already marked as wrong by the TODO comment).

This commit here from another PR migrates the Controller over to use them, it might serve you as a reference: https://github.com/openshift/machine-config-operator/pull/1624/commits/27502d59c114dcc8c0a084d88527080aef9240ba

The PlatformType fields already defined in ControllerConfig.Infra.Status.PlatformStatus.Type, it's just that the Controller doesn't process them, and looks up that info in the ControllerConfig.Platform strings instead.

pkg/operator/sync.go
LorbusChris: oh this is essentially what I meant with the comment I posted below :)

sinnykumari: Fixed, hopefully I did right this time. ToLower() is not needed `infra.Status.PlatformStatus.Type` gives  result in the was mentioned in https://github.com/openshift/api/blob/c6e8c9b6eb9ad7f72fc048212a56a1124447c55e/config/v1/types_infrastructure.go#L75

abhinavdahiya:
> https://github.com/openshift/api/pull/621/files#diff-cee9360e003319719376a6c9189c2e48R29-R37

so the MCO should be reading from `openshift-managed-config/kube-cloud-config` configmap `cloud.conf` key. that's the expectation of the API.

so, therefore, don't look at the `spec.cloudConfig.name`

so personally this should be
```
get configmap openshift-managed-config/kube-cloud-config
if err 
    if err is NotFound
        if cloud config required for platform
            return err
use the cloud.conf key to configure the kubelet.
```

sinnykumari:
Thanks Abhinav for the feedback, made changes accordingly.
Let me know if I missed something :)

sinnykumari:
e2e-gcp-op won't pass right now, from [mco_pevious log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1658/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1903/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-operator-8ccc6775-gtbks_machine-config-operator_previous.log) , it throws error `[operator.go:331] openshift-config-managed/kube-cloud-config configmap not found: configmap "kube-cloud-config" not found` . 

@abhinavdahiya Is  kube_cloud_config controller available now?

sinnykumari:
Had a chat conversation with Abhinav and it seems there maybe a bug in kube_cloud_config controller because of which cloud.conf is not present in openshift-managed-config/kube-cloud-config configmap. This is being tracked in bug https://bugzilla.redhat.com/show_bug.cgi?id=1825944

sinnykumari:
[PR](https://github.com/openshift/cluster-config-operator/pull/12) referencing bug https://bugzilla.redhat.com/show_bug.cgi?id=1825944 has been merged.
/retest

sinnykumari:
Yay, tests are passing now!

runcom:
@abhinavdahiya ptal

pkg/operator/sync.go
```diff
@@ -206,15 +227,23 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 	}
 	spec.AdditionalTrustBundle = trustBundle
 
-	// if the cloudConfig is set in infra read the cloud config reference
-	if infra.Spec.CloudConfig.Name != "" {
-		cc, err := optr.getCloudConfigFromConfigMap("openshift-config", infra.Spec.CloudConfig.Name, infra.Spec.CloudConfig.Key)
+	// check if openshift-config-managed/kube-cloud-config ConfigMap is available and cloud config is required for given platform
```
abhinavdahiya, line 230: personally, instead of a shallow function like processCloudConfigMapError, i think a function that handled cloud.conf would be better because that keeps the syncRenderConfig simple and also improves the readability.

pkg/operator/sync.go
```diff
@@ -206,15 +227,23 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
+	if _, err = optr.clusterCmLister.ConfigMaps("openshift-config-managed").Get("kube-cloud-config"); err != nil {
+		// processCloudConfigMapError is mainly to keep cyclomatic complexity of syncRenderConfig function in range of 30
+		if ccErr := processCloudConfigMapError(err, infra); ccErr != nil {
+			return ccErr
+		}
```
abhinavdahiya, line 235: processCloudConfigMapError returns nil error if the err are was not `NotFound`, and therefore we are skipping handling of error like NotAuthorized for example.

pkg/operator/sync.go
```diff
@@ -97,6 +97,27 @@ func (optr *Operator) syncAll(syncFuncs []syncFunc) error {
...
+	}
+	return false
+}
+
+func processCloudConfigMapError(err error, infra *configv1.Infrastructure) error {
+	if apierrors.IsNotFound(err) && isCloudConfigRequired(infra) {
+		return fmt.Errorf("%s/%s configmap not found: %v", "openshift-config-managed", "kube-cloud-config", err)
+	}
+	return nil
```
abhinavdahiya, line 118:
```suggestion
	if err != nil && ! apierrors.IsNotFound(err) {return err}
	return nil
```

that helps the caller to not worry about handling err and it can return it as-is better

pkg/operator/sync.go
sinnykumari: fixed! I wasn't sure what others error could be and if it makes sense to return errors in those cases or continue further.

pkg/operator/sync.go
sinnykumari: should be addressed now.

pkg/operator/sync.go
sinnykumari: agree. lots of `if` and error returns made me play extra safe :)
moved everything now in syncCloudConfig() function.

pkg/operator/sync.go
```diff
@@ -97,6 +97,50 @@ func (optr *Operator) syncAll(syncFuncs []syncFunc) error {
 	return syncErr.err
 }
 
+// Return true if cloud config is required on a platform.
+func isCloudConfigRequired(infra *configv1.Infrastructure) bool {
+	if infra.Spec.CloudConfig.Name != "" {
```
sinnykumari, line 102: @abhinavdahiya wondering if we really need this return? Isn't supported platform check and return is enough?

abhinavdahiya: it's just useful in most cases, since if spec is set, we better have a cloud-config regardless of if the platform really needs it.

sinnykumari: got it, thanks!

PR #1660, opened 2020-04-16
Add support for CRIO drop-in config files

umohnani8:
**- What I did**

CRI-O now has support for drop-in files in /etc/crio/crio.conf.d
so add that functionality to the ctrcfg controller.
This will help avoid the issues we run into when we make config
changes in cri-o and don't carry them over to the MCO.
Remove the crio.conf template and put the MCO overrides in 00-default
under the /etc/crio/crio.conf.d directory.
Also update the vendor as we no longer depend on cri-o.

**- How to verify it**

Create ctrcfg CR and the updates should be added to /etc/crio/crio.conf.d/01-newconfig. 


templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -1,262 +1,30 @@
 filesystem: "root"
 mode: 0644
-path: "/etc/crio/crio.conf"
+path: "/etc/crio/crio.conf.d/00-default"
 contents:
   inline: |
-    [crio]
```
haircommander, line 6: do you not still need this?

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -1,262 +1,30 @@
...
-
-    # If true, the runtime will not use pivot_root, but instead use MS_MOVE.
-    no_pivot = false
-
-    # Path to the conmon binary, used for monitoring the OCI runtime.
-    # Will be searched for using $PATH if empty.
     conmon = "/usr/libexec/crio/conmon"
-
-    # Cgroup setting for conmon
-    conmon_cgroup = "pod"
```
haircommander, line 69: I'm pretty sure we still want this, the default is system.slice I think

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -1,262 +1,30 @@
...
-    # profile name is "crio-default-" followed by the version string of CRI-O.
-    apparmor_profile = "crio-default"
-
-    # Cgroup management implementation used for the runtime.
-    cgroup_manager = "systemd"
-
-    # List of default capabilities for containers. If it is empty or commented out,
-    # only the capabilities defined in the containers json file by the user/kube
-    # will be added.
-    default_capabilities = [
```
haircommander, line 95: cri-o default doesn't have NET_RAW and SYS_CHROOT, I think we should work towards dropping those as default, but they should be left in for now

haircommander:
yowzers that's a lot of files, we gotta remember to bump the vendor in 4.6 :upside_down_face: 

looks good to start, I'll do a more thorough review tomorrow :tada: :smile:

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -274,11 +275,20 @@ func (ctrl *Controller) cascadeDelete(cfg *mcfgv1.ContainerRuntimeConfig) error
...
-	mcName := cfg.GetFinalizers()[0]
-	err := ctrl.client.MachineconfigurationV1().MachineConfigs().Delete(context.TODO(), mcName, metav1.DeleteOptions{})
-	if err != nil && !errors.IsNotFound(err) {
+	finalizerName := cfg.GetFinalizers()[0]
+	mcs, err := ctrl.client.MachineconfigurationV1().MachineConfigs().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
 		return err
 	}
+	for _, mc := range mcs.Items {
+		if string(mc.ObjectMeta.GetUID()) == finalizerName || mc.GetName() == finalizerName {
```
runcom, line 284: I think I'm missing someting 🤔 how's the finalizerName ever equal to a MC name ? I guess why this routine got changed

runcom: an MC name is also unique and we're picking only `[0]` which should be just one MC anyway

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: yup, not needed anymore for drop in files

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: ah, I missed it, thanks! Fixed

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: Right, makes sense - fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: The list of Finalizers hold the MC name, and since we only create one MC which we overwrite, the list of Finalizers will have the same name so we can just pick `[0]` and use that to delete the MC. This is for the MC snippet, e.g 99-worker-blahblah-containerruntime

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -84,6 +84,9 @@ func newFixture(t *testing.T) *fixture {
 
 func (f *fixture) validateActions() {
 	actions := filterInformerActions(f.client.Actions())
+	fmt.Println("-----actions-----")
```
haircommander, line 87: left over debug?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -28,10 +27,12 @@ import (
 const (
 	minLogSize           = 8192
 	minPidsLimit         = 20
-	crioConfigPath       = "/etc/crio/crio.conf"
 	storageConfigPath    = "/etc/containers/storage.conf"
 	registriesConfigPath = "/etc/containers/registries.conf"
 	policyConfigPath     = "/etc/containers/policy.json"
+	// CRIODropInFilePath is the path at which changes to the crio config will be dropped in
+	// this is exported so that we can use it in the e2e-tests
+	CRIODropInFilePath = "/etc/crio/crio.conf.d/01-newconfig"
```
haircommander, line 35: can we name this something like `01-ctrcfg` or something more informative of its source?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -211,27 +197,24 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
-func updateCRIOConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error) {
+func createCRIOdropinConfig(internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error) {
 	tomlConf := new(tomlConfigCRIO)
-	if _, err := toml.DecodeReader(bytes.NewBuffer(data), tomlConf); err != nil {
-		return nil, fmt.Errorf("error decoding crio config: %v", err)
-	}
 
 	if internal.PidsLimit > 0 {
 		tomlConf.Crio.Runtime.PidsLimit = internal.PidsLimit
+	} else {
```
haircommander, line 205: I am not a huge fan of coding in the defaults for these values. What would you think about instead creating three new toml structs, each with just one value in it. Then ignition could drop in a overridden, for example, `/etc/crio/crio.conf.d/01-pids-limit` file that only overrides pids_limit.

I am worried about coding in the default for these because we just recently changed log level default, and these values could be left behind, resulting in an incomplete value change.

wdyt

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: yup, fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: Yup, makes sense. I was initially not liking having to do the same thing basically 3 times, but it is better than relying on defaults that might change. Fixed it up :)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -28,10 +27,14 @@ import (
...
 	minPidsLimit         = 20
-	crioConfigPath       = "/etc/crio/crio.conf"
 	storageConfigPath    = "/etc/containers/storage.conf"
 	registriesConfigPath = "/etc/containers/registries.conf"
 	policyConfigPath     = "/etc/containers/policy.json"
+	// CRIODropInFilePathPidsLimit is the path at which changes to the crio config for pids-limit
+	// will be dropped in this is exported so that we can use it in the e2e-tests
+	CRIODropInFilePathPidsLimit  = "/etc/crio/crio.conf.d/02-ctrcfg-pidsLimit"
+	crioDropInFilePathLogLevel   = "/etc/crio/crio.conf.d/01-ctrcfg-logLevel"
+	crioDropInFilePathLogSizeMax = "/etc/crio/crio.conf.d/03-ctrcfg-logSizeMax"
```
haircommander, line 37: nit: I'd probably have these the same priority `01`, having them have different implies (to me) they're of different importance

haircommander:
looks great @umohnani8 one more nit then LGTM :smile:

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

umohnani8:
@mtrmac @runcom PTAL

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
+	}
+	if ctrcfg.LogSizeMax != (resource.Quantity{}) {
+		crioLogSizeMaxTOML, err := createCRIOdropinConfigLogSizeMax(cfg.Spec.ContainerRuntimeConfig)
+		if err != nil {
+			glog.V(2).Infoln(cfg, err, "error updating user changes for log-size-max to crio.conf.d: %v", err)
+		} else {
+			ignConfig = append(ignConfig, ignitionConfig{filePath: crioDropInFilePathLogSizeMax, data: crioLogSizeMaxTOML})
+		}
+	}
+	return ignConfig
```
kikisdeliveryservice, line 301: maybe just call this crioDropInFiles?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -526,25 +528,25 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+			originalStorageIgn, _, _, err := generateOriginalContainerRuntimeConfigs(ctrl.templatesDir, controllerConfig, role)
 			if err != nil {
 				return ctrl.syncStatusOnly(cfg, err, "could not generate origin ContainerRuntime Configs: %v", err)
 			}
 
-			var storageTOML, crioTOML []byte
-			ctrcfg := cfg.Spec.ContainerRuntimeConfig
-			if ctrcfg.OverlaySize != (resource.Quantity{}) {
-				storageTOML, err = ctrl.mergeConfigChanges(originalStorageIgn, cfg, updateStorageConfig)
+			var ignConfig []ignitionConfig
```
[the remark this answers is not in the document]
umohnani8: renamed :)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
+		} else {
+			ignConfig = append(ignConfig, ignitionConfig{filePath: CRIODropInFilePathPidsLimit, data: crioPidsLimitTOML})
+		}
+	}
+	if ctrcfg.LogSizeMax != (resource.Quantity{}) {
+		crioLogSizeMaxTOML, err := createCRIOdropinConfigLogSizeMax(cfg.Spec.ContainerRuntimeConfig)
+		if err != nil {
+			glog.V(2).Infoln(cfg, err, "error updating user changes for log-size-max to crio.conf.d: %v", err)
+		} else {
+			ignConfig = append(ignConfig, ignitionConfig{filePath: crioDropInFilePathLogSizeMax, data: crioLogSizeMaxTOML})
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -46,24 +50,43 @@ type tomlConfigStorage struct {
...
+			LogLevel string `toml:"log_level,omitempty"`
+		} `toml:"runtime"`
 	} `toml:"crio"`
 }
 
-// ignitionConfig is a struct that holds the filepath and date of the various configs
+// tomlConfigCRIOPidsLimit is used for conversions when pids-limit is changed
+// TOML-friendly (it has all of the explicit tables). It's just used for
+// conversions.
+type tomlConfigCRIOPidsLimit struct {
```
mtrmac, line 67: (This is OK, just a sigh:

I can’t say that I’m enthusiastic about maintaining a separate copy of the definitions; the dependency does not go away, it’s just harder to find.

OTOH the -475k lines is an extremely compelling argument, and there’s that comment about handling unset fields.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -46,24 +50,43 @@ type tomlConfigStorage struct {
...
+// conversions.
+type tomlConfigCRIOLogSizeMax struct {
+	Crio struct {
+		Runtime struct {
+			LogSizeMax int64 `toml:"log_size_max,omitempty"`
+		} `toml:"runtime"`
+	} `toml:"crio"`
+}
+
+// fileConfig is a struct that holds the filepath and date of the various configs
```
mtrmac, line 86:
```suggestion
// fileConfig is a struct that holds the filepath and data of the various configs
```

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
 }
 
-// updateCRIOConfig decodes the data rendered from the template, merges the changes in and encodes it
-// back into a TOML format. It returns the bytes of the encoded data
-func updateCRIOConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error) {
-	tomlConf := new(tomlConfigCRIO)
-	if _, err := toml.DecodeReader(bytes.NewBuffer(data), tomlConf); err != nil {
-		return nil, fmt.Errorf("error decoding crio config: %v", err)
-	}
+// createCRIOdropinConfigLogLevel decodes the data rendered from the template, merges the changes in and encodes it
```
mtrmac, line 221: No template is involved now. (In all three instances).

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
+// back into a TOML format. It returns the bytes of the encoded data.
+// This is only for when the log-level field is changed. We have separated it out into 3 different toml structs
```
mtrmac, line 223: This feels like a comment for definitions of the three structures; that’s the place to explain why the CRI-O types are not reused, and why we need three different ones.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
+// and 3 different functions because the toml encoder does not ignore a value that was not set - it will set it as
+// empty. So if we just change field A and not B, it sets B to empty which is usually different from the default it
+// was set to.
+func createCRIOdropinConfigLogLevel(internal *mcfgv1.ContainerRuntimeConfiguration) ([]byte, error) {
```
mtrmac, line 227: These three functions feel rather laborious.

How about a single
```go
func addTOMLFileConfig([]fileConfig, path string, interface{} marshalWithToml) ([]fileConfig, error)
```

called like
```go
	if ctrcfg.LogLevel != "" {
 		conf := tomlConfigCRIOLogLevel{}
                conf.Crio.Runtime.LogLevel = internal.LogLevel
 		fileConfigList, err := addTOMLFileConfig(fileConfigList, crioDropInFilePathLogLevel, conf)
 		if err { … }
      }
```

That would, AFAICS, make `createCRIODropinFiles` almost as large as the current implementation, but without splitting the logic over 4 functions and possibly worrying about whether the “if parameter is set” and “use parameter value” code in different functions can ever get out of sync, and without so much duplication.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -209,39 +218,89 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
...
+// We create different drop-in files for each CRI-O field that can be changed by the ctrcfg CR
+// this ensures that we don't have to rely on hard coded defaults that might cause problems
+// in future if something in cri-o or the templates used by the MCO changes
+func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []fileConfig {
+	var fileConfigList []fileConfig
+	ctrcfg := cfg.Spec.ContainerRuntimeConfig
+	if ctrcfg.LogLevel != "" {
+		crioLogLevelTOML, err := createCRIOdropinConfigLogLevel(cfg.Spec.ContainerRuntimeConfig)
+		if err != nil {
+			glog.V(2).Infoln(cfg, err, "error updating user changes for log-level to crio.conf.d: %v", err)
```
mtrmac, line 280: Is it OK to just log? The previous implementation caused a failure.

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -1,97 +1,20 @@
...
-
-    # Cgroup setting for conmon
     conmon_cgroup = "pod"
-
-    # Environment variable list for the conmon process, used for passing necessary
-    # environment variables to conmon or the runtime.
-    conmon_env = [
-        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+    default_env = [
+        "NSS_SDB_USE_CACHE=no",
```
mtrmac, line 14: This seems to be net new; is that intentional?

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
```diff
@@ -1,97 +1,20 @@
 filesystem: "root"
 mode: 0644
-path: "/etc/crio/crio.conf"
+path: "/etc/crio/crio.conf.d/00-default"
```
mtrmac, line 3: Is there even any difference between the worker/master files, or could this become a single file in `templates/common` now?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -46,24 +50,43 @@ type tomlConfigStorage struct {
...
+			LogSizeMax int64 `toml:"log_size_max,omitempty"`
+		} `toml:"runtime"`
+	} `toml:"crio"`
+}
+
+// fileConfig is a struct that holds the filepath and date of the various configs
 // Using a struct array ensures that the order of the ignition files always stay the same
 // ensuring that double MCs are not created due to a change in the order
-type ignitionConfig struct {
+type fileConfig struct {
```
mtrmac, line 89: (Absolutely non-blocking, I’m sure everybody loves name bike-shedding…: to me this reads as “configuration of a file”. `configFile`? `generatedFile`?)

mtrmac:
(I have only reviewed the mechanism of the thing, not the substance of change to how CRI-O is configured.)

templates/master/01-master-container-runtime/_base/files/crio.yaml
mtrmac: Ah, #1664. Ideally the two PRs should start separate, although I appreciate that’s a hassle.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -903,3 +903,34 @@ func (ctrl *Controller) getPoolsForContainerRuntimeConfig(config *mcfgv1.Contain
...
+		return false, nil
+	}
+
+	for _, pool := range mcpPools {
+		managedKey := getManagedKeyCtrCfg(pool)
+		mc, err := ctrl.client.MachineconfigurationV1().MachineConfigs().Get(context.TODO(), managedKey, metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return false, fmt.Errorf("could not get mc with name %q: %v", managedKey, err)
+		}
+		conf, _, err := ign.Parse(mc.Spec.Config.Raw)
```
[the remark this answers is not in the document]
kikisdeliveryservice: Ah yes we check if it's nil before parsing:
https://github.com/openshift/machine-config-operator/blob/d4267c0ab866323274288d936896097387b57220/pkg/controller/common/helpers.go#L138

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: Ohh those are better suggestions. :smile:

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
kikisdeliveryservice: good catch!

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -378,17 +379,17 @@ func (ctrl *Controller) handleImgErr(err error, key interface{}) {
 }
 
 // generateOriginalContainerRuntimeConfigs returns rendered default storage, and crio config files
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -378,17 +379,17 @@ func (ctrl *Controller) handleImgErr(err error, key interface{}) {
...
-func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.ControllerConfig, role string) (*igntypes.File, *igntypes.File, *igntypes.File, *igntypes.File, error) {
+func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.ControllerConfig, role string) (*igntypes.File, *igntypes.File, *igntypes.File, error) {
 	// Render the default templates
 	rc := &mtmpl.RenderConfig{ControllerConfigSpec: &cc.Spec}
 	generatedConfigs, err := mtmpl.GenerateMachineConfigsForRole(rc, role, templateDir)
 	if err != nil {
-		return nil, nil, nil, nil, fmt.Errorf("generateMachineConfigsforRole failed with error %s", err)
+		return nil, nil, nil, fmt.Errorf("generateMachineConfigsforRole failed with error %s", err)
 	}
 	// Find generated storage.config, and crio.config
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -378,17 +379,17 @@ func (ctrl *Controller) handleImgErr(err error, key interface{}) {
 }
 
 // generateOriginalContainerRuntimeConfigs returns rendered default storage, and crio config files
-func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.ControllerConfig, role string) (*igntypes.File, *igntypes.File, *igntypes.File, *igntypes.File, error) {
+func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.ControllerConfig, role string) (*igntypes.File, *igntypes.File, *igntypes.File, error) {
```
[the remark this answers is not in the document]
umohnani8: Will fix up in a future PR :)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
umohnani8: thanks, good catch. Fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -526,25 +528,25 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
 			}
-			if ctrcfg.LogLevel != "" || ctrcfg.PidsLimit != 0 || ctrcfg.LogSizeMax != (resource.Quantity{}) {
-				crioTOML, err = ctrl.mergeConfigChanges(originalCRIOIgn, cfg, updateCRIOConfig)
-				if err != nil {
-					glog.V(2).Infoln(cfg, err, "error merging user changes to crio.conf: %v", err)
-				}
-			}
+
+			// Create the cri-o drop-in files
+			crioIgnConfig := createCRIODropinFiles(cfg)
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -553,10 +555,8 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 				}
 			}
 
-			ctrRuntimeConfigIgn := createNewIgnition([]ignitionConfig{
-				{filePath: storageConfigPath, data: storageTOML},
-				{filePath: crioConfigPath, data: crioTOML},
-			})
+			fileConfigList = append(fileConfigList, crioIgnConfig...)
```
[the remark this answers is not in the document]

pkg/controller/container-runtime-config/helpers.go
umohnani8: This definitely reduces the complexity with the dependency and all. Since it is only 3 fields, I am inclined to leave it as is for now.

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: yup, fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: Agreed, fixed

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: yeah, we need to backport this in 4.4 as well and no way doing it without getting it in 4.5 first. So ideally #1664 will go in, and then this PR with the new template

templates/worker/01-worker-container-runtime/_base/files/crio.yaml
umohnani8: This could, but I will consolidate it in a future PR

pkg/controller/container-runtime-config/helpers.go
umohnani8: renamed it to `generatedConfigFile` so we can all be happy :D

pkg/controller/container-runtime-config/helpers.go
umohnani8: yes, the previous implementation was also just logging https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/container-runtime-config/container_runtime_config_controller.go#L545

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -413,7 +413,6 @@ func TestContainerRuntimeConfigCreate(t *testing.T) {
 
 			f.expectGetMachineConfigAction(mcs1)
 			f.expectUpdateContainerRuntimeConfig(ctrcfg1)
-			f.expectUpdateContainerRuntimeConfig(ctrcfg1)
```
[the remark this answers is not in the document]
runcom: uhm

runcom:
```
--- FAIL: TestContainerRuntimeConfigPidsLimit (7.74s)
```
the above looks really suspicious given the changes in this PR

test/e2e/ctrcfg_test.go
```diff
@@ -17,9 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-func TestContainerRuntimeConfigPidsLimit(t *testing.T) {
-	runTestWithCtrcfg(t, "pids-limit", `pids_limit = (\S+)`, "12345", &mcfgv1.ContainerRuntimeConfiguration{
-		PidsLimit: 12345,
+func TestContainerRuntimeConfigLogLevel(t *testing.T) {
+	runTestWithCtrcfg(t, "log-level", `log_level = (\S+)`, "debug", &mcfgv1.ContainerRuntimeConfiguration{
+		LogLevel: "debug",
```
runcom, line 22: why aren't we testing PidsLimit anymore?

umohnani8: Because to get the default value of pidsLimit we will have to run `crio config` since we no longer have a crio.conf and that wasn't possible in the MCD pod that the tests use. So moved to LogLevel as we have that set in crio.conf.d/00-default and when we update it we can verify it by reading crio.conf.d/01-ctrcfg-logLevel.

runcom:
This should be good now pending CI which hasn't cooperated so far, we'll figure out follow ups then - going to lgtm this as it's definitely done but we need to hear from CI, the lgtm will go away if the code needs changing ofc.

/lgtm

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -196,7 +209,7 @@ func updateStorageConfig(data []byte, internal *mcfgv1.ContainerRuntimeConfigura
 		return nil, fmt.Errorf("error decoding crio config: %v", err)
 	}
 
-	if internal.OverlaySize != (resource.Quantity{}) {
+	if internal.OverlaySize.Value() != 0 {
```
mtrmac, line 212: Is this related to the topic of this PR? And how does it related to current line 262 in this file (`LogSizeMax`) checking for both?

I would also expect the relationship of conditions in `Controller.syncRuntimeConfig` and {`updateStorageConfig` vs. `createCRIODropInFiles`} to be consistent — either the caller and the callee should be checking for exactly the same condition, or the caller should be checking for “is the value present” and the callee for “is the value (which is known to be present) something that requires a config change”.

Right now `OverlaySize` and `LogSizeMax` are treated differently and I’m not sure whether there is some hard-to-notice reason for that.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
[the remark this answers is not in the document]
mtrmac: @runcom What do you mean?

test/e2e/ctrcfg_test.go
```diff
@@ -104,7 +106,7 @@ func runTestWithCtrcfg(t *testing.T, testName, regexKey, expectedConfValue strin
 	// ensure config rolls back as expected
 	waitForConfigAndPoolComplete(t, cs, poolName, oldMCConfig.Name)
 
-	restoredConfValue := getValueFromCrioConfig(t, cs, node, regexKey)
+	restoredConfValue := getValueFromCrioConfig(t, cs, node, regexKey, false)
```
mtrmac, line 109: AFAICS this no longer tests that rollback actually rolls back — the 00-default configuration is never changed. This should be testing that the override file is not present, or does not contain that key.

test/e2e/ctrcfg_test.go
```diff
@@ -162,16 +163,21 @@ func getMCFromCtrcfg(t *testing.T, cs *framework.ClientSet, ctrcfgName string) (
 // getValueFromCrioConfig jumps onto the node and gets the crio config. It then uses the regexKey to
 // find the value that is being searched for
 // regexKey is expected to be in the form `key = (\S+)` to search for the value of key
-func getValueFromCrioConfig(t *testing.T, cs *framework.ClientSet, node corev1.Node, regexKey string) string {
+func getValueFromCrioConfig(t *testing.T, cs *framework.ClientSet, node corev1.Node, regexKey string, getOverrideValue bool) string {
 	// get the contents of the crio.conf on nodeName
-	out := execCmdOnNode(t, cs, node, "cat", "/rootfs/etc/crio/crio.conf")
+	var out string
+	if getOverrideValue {
+		out = execCmdOnNode(t, cs, node, "cat", filepath.Join("/rootfs", ctrcfg.CRIODropInFilePathLogLevel))
```
mtrmac, line 170: (Non-blocking, this works fine:  If “override” means “one particular override”, it would be a bit clearer to just have the caller provide a node-relative path argument (+ adding a constant for the `00-default` path) instead of a boolean.)

runcom:
@mrunalp I think this should be merged but we're past that date and it's now asking for a bugzilla, do you know how to proceed here?

test/e2e/ctrcfg_test.go
umohnani8: Fixed

test/e2e/ctrcfg_test.go
umohnani8: Fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: Fixed

PR #1661, opened 2020-04-17
Bug 1825221: templates/master:openshift-recovery-tools pass --authfile to podman create

hexfusion:
In the case that a master node is new it is possible that the etcd container has not yet been downloaded. In this case, DR script will fail because we need auth to create the container by first pulling the image.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
@@ -5,7 +5,6 @@ contents:
   inline: |
     #!/usr/bin/env bash
     export ETCDCTL_API=3
-    export ETCD_VERSION=v3.3.17
 
     ETCDCTL_WITH_TLS="$ETCDCTL --cert $ASSET_DIR/backup/etcd-client.crt --key $ASSET_DIR/backup/etcd-client.key --cacert $ASSET_DIR/backup/etcd-ca-bundle.crt"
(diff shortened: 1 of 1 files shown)


retroflexer:
There is a typo in the authfile used, which need to be fixed.

templates/master/00-master/_base/files/usr-local-bin-openshift-recovery-tools-sh.yaml
```diff
@@ -19,10 +19,10 @@ contents:
       fi
     }
 
-    # download and test etcdctl from upstream release assets
+    # extract etcdctl binary from container image
     dl_etcdctl() {
       local etcdimg="{{.Images.etcdKey}}"
-      local etcdctr=$(podman create "${etcdimg}")
+      local etcdctr=$(podman create "${etcdimg}" --authfile=/var/log/kubelet/config.json)
```
retroflexer, line 25: There is a typo. The authfile should be /var/lib/kubelet/config.json.

hexfusion: updated

retroflexer:
Thanks for fixing the typo.

/lgtm

sinnykumari:
This bug doesn't shouldn't depend on bz#1823931 as it is targeted to 4.5

sdodson:
/retitle Bug 1825221: templates/master:openshift-recovery-tools pass --authfile to podman create

PR #1662, opened 2020-04-17
Update vSphere IPI removing DNS VIP

jcpowermac:
Align with current updates to baremetal
See PR #1569

"Now that etcd does not need DNS for clustering, we no longer need to
on the bootstrap coredns so the same method is now used for masters."


jcpowermac:
Tested with installer and custom release image in IPI.

jcpowermac:
@kikisdeliveryservice thanks!

PR #1665, opened 2020-04-20
Bug 1825823: Add support for reading API LB backends from KUBE-API for OpenStack and Ovirt

Fedosin:
This PR updates OpenStack and Ovirt HAProxy static pod manifest to enable reading API LB backend details from Kubernetes instead of _etcd-server-ssl._tcp SRV.

Original baremetal PR: https://github.com/openshift/machine-config-operator/pull/1574


mandre:
It's very likely that oVirt and vsphere need the same fix, related to the https://github.com/openshift/baremetal-runtimecfg/pull/52 change.

/cc @rgolangh @jcpowermac

jcpowermac:
@mandre thanks for the heads up!

PR #1659, opened 2020-04-16
Bug 1826211: drop python usage for node-ip functionality

celebdor:
Recently the non_virtual_ip and nodeip-finder functionality was
implemented in golang on the baremetal-runtimecfg repository. This
commit changes the baremetal and the openstack platforms to use the
golang implementation via podman, so we do not block rhcos dropping
platform-python.

If baremetal IPI clusters, it means it works. This obsoletes https://github.com/openshift/machine-config-operator/pull/1564 but needs https://github.com/openshift/baremetal-runtimecfg/pull/57 to merge first

PR #1666, opened 2020-04-20
Bug 1841066: OpenStack: Remove DNS VIP

mandre:
Now that etcd does not need DNS for clustering, we no longer need to
on the bootstrap coredns so the same method is now used for masters.

This brings the BM change #1569 to OpenStack platform.


mandre:
Depends on https://github.com/openshift/machine-config-operator/pull/1659

mandre:
/hold cancel
/retest

e2e-openstack tests are passing.

PR #1674, opened 2020-04-23
Bug 1825909: baremetal - Limit the number of nameservers to 3

yboaron:
This PR prevents the pods from complaining about the nameserver limits
exceeded:

    Nameserver limits were exceeded, some nameservers have been omitted, the applied nameserver line is: 10.0.128.6 10.0.128.12 10.0.128.11 (207 times)

Original openstack PR: #1573 

**- What I did**

**- How to verify it**


sinnykumari:
one question: are the nameserver entries are unique? If not, it would be good to keep first 3 unique entries.

yboaron:
@sinnykumari thanks for the feedback.
The nodes get their nameserver list from an external DHCP server, so if there are duplicate entries  it's because of misconfiguration in the external DHCP server.
So, I guess we shouldn't handle issues caused by misconfiguration.


I feel like duplicate entries in resolv.conf is a misconfiguration anyway.

yboaron:
@celebdor 
I checked it locally (IPV4 deployment) - looks OK.

PR #1677, opened 2020-04-23
Bug 1827600: Fix crd openAPI validation regression

LorbusChris:
https://github.com/openshift/machine-config-operator/pull/1474 introduced a small regression on the openAPI validation scheme, this fixes that.

**- What I did**
- mco: Use "github.com/clarketm/json" for marshaling Ignition configs
Replaces "encoding/json" with dropin replacement "github.com/clarketm/json"
that supports zero values of structs with omittempty when marshaling
Ignition configs.
In effect, this will exclude empty pointer struct fields from the
marshaled data instead of inserting nil values into them which do not
pass openAPI validation on fields that are supposed to contain strings.
See golang/go#11939 for more context.
- manifests: Update machineconfig crd & bindata
Reverts a regression on the openAPI validation scheme.
- Update vendor

**- How to verify it**
CI

Use "github.com/clarketm/json" for marshaling Ignition configs and revert a regression on the openAPI validation scheme.

@openshift/openshift-team-mco
Please let me know if you want this in 4.5 and therefore need a BZ.


pkg/server/api_test.go
```diff
@@ -29,6 +29,8 @@ type scenario struct {
 	checkResponse checkResponse
 }
 
+const expectedContentLength int = 15
```
yuqi-zhang, line 32: Where did this change originate from?

LorbusChris: the marshaled data is much smaller now due to leaving out all the empty/nulled fields - it's a nice side benefit of using this parser. As the number (114 before, 15 now) is specified a few times, I introdued this constant.

LorbusChris:
> I'm not against using a nullable json parser. Does this mean we can drop all nullable fields in general?

I'm not entirely sure what you mean here, can you elaborate?

We use this code in fcos: https://github.com/openshift/machine-config-operator/pull/1559/commits/563275c76cfa23a0bcb63c154f0312adc6542dae
Unfortunately the history for this on the branch got lost during the latest big sync merge.

LorbusChris:
@yuqi-zhang squashed the two commits.

Wrt above, I think what it means is that we can drop all fields labeled `omitempty`, I'm not sure about the `nullable` part.

PR #1682, opened 2020-04-24
Bug 1765609: templates: add chrony config on Azure for host time synchronization

sinnykumari:
With default config, host time synchronization fails on RHCOS nodes
launched on Azure and results in time drift.
As per Azure best practices we are updating chrony.conf to use
PTP source clock https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1765609

### DIFF
--- a/templates/common/azure/files/chrony.yaml
+++ b/templates/common/azure/files/chrony.yaml
@@ -1,0 +1,51 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/chrony.conf"
+contents:
+  inline: |
+    # This file is shipped by the Machine Config Operator to synchronize host machine time on Azure.
+    # Make sure that chrony.conf content in machine-config-operator (/templates/common/azure/files) and
+    # installer (data/data/bootstrap/azure/files/etc) repo are in sync.
+    #
+    # see https://bugzilla.redhat.com/show_bug.cgi?id=1765609 and https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
+    refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
+
+    # Use public servers from the pool.ntp.org project.
+    # Please consider joining the pool (http://www.pool.ntp.org/join.html).
+    pool 2.rhel.pool.ntp.org iburst
+
+    # Record the rate at which the system clock gains/losses time.
+    driftfile /var/lib/chrony/drift
+
+    # Allow the system clock to be stepped in the first three updates
+    # if its offset is larger than 1 second.
+    makestep 1.0 3
+
+    # Enable kernel synchronization of the real-time clock (RTC).
+    rtcsync
+
+    # Enable hardware timestamping on all interfaces that support it.
+    #hwtimestamp *
+
+    # Increase the minimum number of selectable sources required to adjust
+    # the system clock.
+    #minsources 2
+
+    # Allow NTP client access from local network.
+    #allow 192.168.0.0/16
+
+    # Serve time even if not synchronized to a time source.
+    #local stratum 10
+
+    # Specify file containing keys for NTP authentication.
+    keyfile /etc/chrony.keys
+
+    # Get TAI-UTC offset and leap seconds from the system tz database.
+    leapsectz right/UTC
+
+    # Specify directory for log files.
+    logdir /var/log/chrony
+
+    # Select which information is logged.
+    #log measurements statistics tracking
+


templates/common/azure/files/azure-chrony.yaml
```diff
@@ -0,0 +1,50 @@
...
+mode: 0644
+path: "/etc/chrony.conf"
+contents:
+  inline: |
+    # This file is shipped by the Machine Config Operator to synchronize time RHCOS host time on azure.
+    # It uses PTP source clock, see https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
+    #
+    # Use public servers from the pool.ntp.org project.
+    # Please consider joining the pool (http://www.pool.ntp.org/join.html).
+    pool 2.rhel.pool.ntp.org iburst
```
sinnykumari, line 11: Question: Since this doesn't work , do we want to comment this out?

sinnykumari:
Azure cluster created with this PR  looks fine.  /etc/chrony.conf is overwritten with this change only on Azure and time is updated properly.

```
sh-4.4# date
Fri Apr 24 07:04:22 UTC 2020

sh-4.4# chronyc sources
210 Number of sources = 9
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
#* PHC0                          0   3   377     8    -45us[ -108us] +/- 8403ns
^? inter.tyjo.eu                 0   9     0     -     +0ns[   +0ns] +/-    0ns
^? main24.anyplace-hosting.>     0   9     0     -     +0ns[   +0ns] +/-    0ns
^? ntp1.trans-ix.nl              0   9     0     -     +0ns[   +0ns] +/-    0ns
^? ntp3.doctor.com               0   9     0     -     +0ns[   +0ns] +/-    0ns
^? 2402:1f00:8100:400::3a8       0   6     0     -     +0ns[   +0ns] +/-    0ns
^? entp1.iiasa.ac.at             0   6     0     -     +0ns[   +0ns] +/-    0ns
^? lust.majkl578.cz              0   6     0     -     +0ns[   +0ns] +/-    0ns
^? ntp.ixo.no                    0   6     0     -     +0ns[   +0ns] +/-    0ns

sh-4.4# chronyc tracking
Reference ID    : 50484330 (PHC0)
Stratum         : 1
Ref time (UTC)  : Fri Apr 24 07:05:20 2020
System time     : 0.000000003 seconds slow of NTP time
Last offset     : +0.000047591 seconds
RMS offset      : 0.000040784 seconds
Frequency       : 6.245 ppm fast
Residual freq   : +0.028 ppm
Skew            : 2.423 ppm
Root delay      : 0.000000001 seconds
Root dispersion : 0.000030604 seconds
Update interval : 8.0 seconds
Leap status     : Normal
```

templates/common/azure/files/azure-chrony.yaml
```diff
@@ -0,0 +1,50 @@
+filesystem: "root"
```
runcom, line 1: I'd remove `azure-` prefix as the folders are already namespaced

templates/common/azure/files/azure-chrony.yaml
runcom: if it's not causing issue even if not working but that's something we do ship anyway, I'd leave it there

templates/common/azure/files/azure-chrony.yaml
sinnykumari: sure, will remove it with next fixup!

sinnykumari: fixed

templates/common/azure/files/chrony.yaml
```diff
@@ -0,0 +1,50 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/chrony.conf"
+contents:
+  inline: |
+    # This file is shipped by the Machine Config Operator to synchronize time RHCOS host time on azure.
```
jlebon, line 6: Looks like an extra `time` in there? Also maybe `s/azure/Azure/`.

runcom:
@jlebon ptal

/approve

templates/common/azure/files/chrony.yaml
```diff
@@ -0,0 +1,50 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/chrony.conf"
```
jlebon, line 3: Hmm, it's unfortunate that chrony doesn't seem to support something like `/etc/chrony.conf.d`. Is the only real change here from the default this line:

> `refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0`

?

Might be worth highlighting that better in the config. Maybe e.g. putting it up near the top right after the MCO-specific comment?

runcom:
e2e-gcp-op kernelType failure has been now reported here https://bugzilla.redhat.com/show_bug.cgi?id=1827712

runcom:
> This seems reasonable to me. A few comments. Just to check, files in templates/common also apply to the bootstrap node too, right?

@jlebon nope, that's not the case on the bootstrap host - the bootstrap host initial setup is managed by the installer

jlebon:
> @jlebon nope, that's not the case on the bootstrap host - the bootstrap host initial setup is managed by the installer

Hmm, I think we'd want to do the change there too then, right? See also https://bugzilla.redhat.com/show_bug.cgi?id=1800901 where a mismatch between the bootstrap and masters is causing invalid certificate errors.

runcom:
> Hmm, I think we'd want to do the change there too then, right? See also https://bugzilla.redhat.com/show_bug.cgi?id=1800901 where a mismatch between the bootstrap and masters is causing invalid certificate errors.

This is hacky tho - now we're gonna ship two files from different places, one from MCO and one from Installer and they can then drift if we forget to update either if needed 🤔

templates/common/azure/files/chrony.yaml
sinnykumari: ah yes, will fix it.

templates/common/azure/files/chrony.yaml
sinnykumari: yes,that's the main change. I kept at bottom as this was the addition. I will add it on top if that makes more sense.

sinnykumari: fixed

sinnykumari:
This PR is trying to fix Azure specific time sync issue but I see that we have generic time sync issue as well. 
Does it make sense to merge this PR and discuss generic approach in https://github.com/openshift/machine-config-operator/issues/629 and https://bugzilla.redhat.com/show_bug.cgi?id=1800901 ?

For updating chrony config on bootstrap node, maybe we can open an installer PR and bug if it happens one Azure before we solve the time sync issue in generic way.

runcom:
> For updating chrony config on bootstrap node, maybe we can open an installer PR and bug if it happens one Azure before we solve the time sync issue in generic way.

right, what needs to be done in installer is basically shipping this very same configuration directly under `./data/data/bootstrap/azure/files/etc/chrony.conf` and add a note here and there about keeping them in sync. The issue is that the installer has its own way of configuring the bootstrap host with shipped files from there - maybe it will make sense in the future to have the MCO do that but this is not the time I guess.

Let's merge this and PR the installer as I said above and we can then close the Bugzilla.

/lgtm

sinnykumari:
updated PR with note on keeping chrony.conf content in mco and installer repo in sync. will need lgtm again.

PR to update chrony config on bootstroop node - https://github.com/openshift/installer/pull/3513

runcom:
holding pending some other conversations going on

/hold

sinnykumari:
From conversation on https://github.com/openshift/installer/pull/3513 , it seems we are leaning towards doing  platform specific config in RHCOS itself. WIP PR is at https://github.com/coreos/fedora-coreos-config/pull/367 for first getting tested in FCOS and that would later propagate to RHCOS well. Let's keep this PR on hold and we can close it if we proceed with doing in RHCOS.

cgwalters:
> From conversation on openshift/installer#3513 , it seems we are leaning towards doing platform specific config in RHCOS itself. WIP PR is at coreos/fedora-coreos-config#367 for first getting tested in FCOS and that would later propagate to RHCOS well. Let's keep this PR on hold and we can close it if we proceed with doing in RHCOS.

Yep, I am in process of backporting that.  Sorry about "taking over" this one; the argument that we need synchronized time as early as possible ⏰ I think makes sense as a driver for doing it in the OS by default though.

The FCOS change got merged and I put up the RHCOS update here:
https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/918

Will steal the bug too!

runcom:
@cgwalters can we close this now then?

cgwalters:
Yep
/close

jlebon:
Props to @sinnykumari for finding the relevant Azure docs and getting the ball rolling!

PR #1692, opened 2020-04-28
MCO: grab images from MCO itself

### LINKED ISSUE #988: have the MCO gather bootstrap images
Today, [bootkube inspects the payload](https://github.com/openshift/installer/blob/720b4889bbf34872356dee0a60a49ecf122946c8/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L19) and [pass to the MCO bootstrap the images we need](https://github.com/openshift/installer/blob/720b4889bbf34872356dee0a60a49ecf122946c8/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L193).
This has been an ongoing hassle to have ratchet PRs all around to avoid breaking either component.
What we can do is having the MCO get the full payload image and inspect it ourselves to avoid the back and forth when adding/removing images for us. The [CVO for instance does that itself already](https://github.com/openshift/installer/blob/720b4889bbf34872356dee0a60a49ecf122946c8/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L64).

runcom:
Effectively fix the dependency burden we have everytime someone tries
to add an image to be used in any of our templates. Will require a PR
in the installer to fully drop the flags marked as "drop".

Fix #988
Could further help with PR like this openshift/machine-config-operator #1605

Will require an installer PR once this merges to clean flags up but not a big deal anyway.

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -19,32 +19,33 @@ var (
 	}
 
 	bootstrapOpts struct {
-		baremetalRuntimeCfgImage  string
 		cloudConfigFile           string
 		configFile                string
 		cloudProviderCAFile       string
-		corednsImage              string
 		destinationDir            string
 		etcdCAFile                string
-		etcdImage                 string
 		etcdMetricCAFile          string
-		haproxyImage              string
 		imagesConfigMapFile       string
 		infraConfigFile           string
-		infraImage                string
-		keepalivedImage           string
 		kubeCAFile                string
-		kubeClientAgentImage      string
-		clusterEtcdOperatorImage  string
-		mcoImage                  string
-		mdnsPublisherImage        string
-		oauthProxyImage           string
 		networkConfigFile         string
-		oscontentImage            string
 		pullSecretFile            string
 		rootCAFile                string
 		proxyConfigFile           string
 		additionalTrustBundleFile string
+		// TODO: drop
+		baremetalRuntimeCfgImage string
+		corednsImage             string
+		etcdImage                string
+		haproxyImage             string
+		infraImage               string
+		keepalivedImage          string
+		kubeClientAgentImage     string
+		clusterEtcdOperatorImage string
+		mcoImage                 string
+		mdnsPublisherImage       string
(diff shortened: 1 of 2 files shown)


runcom:
Slated for 4.6 anyway

runcom:
@cgwalters @kikisdeliveryservice w/o any urgency, ptal

runcom:
I'm closing this for now already as I've realized the manifests dir doesn't have "built" images but those are in the payload itself - so, it requires an extra step which I'll do once the time comes :)

runcom:
we're looking at something like `docker run -ti --entrypoint cat {{release-payload}} /release-manifests/0000_80_machine-config-operator_02_images.configmap.yaml  > imagesconfigmap.yaml` to be run in bootkube instead as the MCO image itself doesn't have "substituted" manifests yet but the release payload image does

runcom:
```diff
diff --git a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
index e9d4ee42a..610c18515 100755
--- a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
@@ -266,6 +266,9 @@ then

        rm --recursive --force mco-bootstrap

+        podman run --quiet --rm --net=none --entrypoint "cat" "${RELEASE_IMAGE_DIGEST}" /release-manifests/0000_80_machine-config-operator_02_images.configmap.yaml > /assets/0000_80_machine-config-operator_02_images.configmap.yaml
+        podman run --quiet --rm --net=none --entrypoint "cat" "${RELEASE_IMAGE_DIGEST}" /release-manifests/0000_80_machine-config-operator_05_osimageurl.yaml > /assets/0000_80_machine-config-operator_05_osimageurl.yaml
+
        ADDITIONAL_FLAGS=""
        if [ -f "/opt/openshift/tls/cloud-ca-cert.pem" ]; then
                ADDITIONAL_FLAGS="--cloud-provider-ca-file=/assets/tls/cloud-ca-cert.pem"
```

PR #1693, opened 2020-04-28
Bug 1816178: MCO: have the CVO create CRDs

runcom:
We'll be getting rid of this awful message in bootstrap logs:

unable to get REST mapping for "99_openshift-machineconfig_99-worker-ssh.yaml": no matches for kind "MachineConfig" in version "machineconfiguration.openshift.io/v1"

Which repeats over and over. The reason is the MCO hasn't started yet
by the time the bootstrap process creates those manifests because
the MCO will create the CRDs only once it's running.
With this, we'll have the CVO pre-create our CRDs so they're ready
immediately. The reason it wasn't like this from the MCO birth was that
the CVO was lacking such feature but now that it does have it, we're leveraging it.


kikisdeliveryservice:
```
evel=error msg="Cluster operator authentication Degraded is True with IngressStateEndpoints_MissingEndpoints::RouteStatus_FailedHost: IngressStateEndpointsDegraded: No endpoints found for oauth-server\nRouteStatusDegraded: route is not available at canonical host oauth-openshift.apps.ci-op-15zc46vf-1354f.origin-ci-int-gce.dev.openshift.com: []"
```

/retest

kikisdeliveryservice:
does this affect upgrades/downgrades?

runcom:
Updated to fix `inClusterBringUp` - should now be ready to go

pkg/operator/sync.go
```diff
@@ -286,11 +286,7 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 
 func (optr *Operator) syncCustomResourceDefinitions() error {
 	crds := []string{
-		"manifests/machineconfig.crd.yaml",
 		"manifests/controllerconfig.crd.yaml",
```
LorbusChris, line 289: since you're leaving the controllerconfig here as-is, does that mean it should also be excluded from the move to the api repo?

runcom: I think so, the CC is an internal sync mechanism for the MCO, nothing users/admins should tweak

PR #1694, opened 2020-04-29
Bug 1820785: [baremetal] Correctly handle requests for ipv4/ipv6 records

cybertron:
Previously the ingress record was hard-coded to be an A record, which
does not work when deploying with an ipv6 ingress VIP. While fixing
that, an issue with the api and api-int records was also exposed.

According to RFC 4074, when a query for a record comes in that asks
for the wrong type of record (i.e. asking for an A record when only
a AAAA record exists), the server should return NOERROR with an
empty list of records. This indicates to the client resolver that
a record of that name exists, but not with that type.

For our purposes, it was necessary to convert all of the records to
use the template plugin in order to get this behavior. The hosts
plugin previously used for the api records does not correctly handle
requests for the wrong record type, at least with the fallthrough
option that we also need. The file plugin, which does correctly
handle requests for the wrong record type, has a different issue.
We want to allow external records in the same subdomain (and this is
necessary for development purposes), but there is no way to make
the file plugin forward requests it can't handle.

To make the template plugin behave correctly, two entries are needed
for each record: one with the correct type and the appropriate answer,
and one with the incorrect type and an empty answer. This gets us
RFC-compliant behavior while still allowing fallthrough for requests
that are not present in the local server.

This change requires new fields from baremetal-runtimecfg that are
added in https://github.com/openshift/baremetal-runtimecfg/pull/60.

**- How to verify it**
Requests for the matching record type should return the appropriate record, requests for the wrong record type for the VIP IP version should return NOERROR with an empty record list.

Fixes misbehavior in the internal coredns server when dealing with ipv4 and ipv6 records.

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -5,14 +5,31 @@
     forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}}
     cache 30
     reload
-    hosts {
-        {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} api-int.{{ .ControllerConfig.Infra.Status.EtcdDiscoveryDomain }}
-        {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} api.{{ .ControllerConfig.Infra.Status.EtcdDiscoveryDomain }}
+    template IN {{`{{ .Cluster.IngressVIPRecordType }}`}} {{ .ControllerConfig.Infra.Status.EtcdDiscoveryDomain }} {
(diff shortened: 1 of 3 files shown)


cybertron:
Sorry, one of these days I'm going to remember whether the bug or platform comes first. :-/

/hold

This won't work until the baremetal-runtimecfg PR merges. Need to retest metal-ipi once it does.

cybertron:
/hold cancel
/retest

baremetal-runtimecfg has merged. This should work now.

cybertron:
Just a rebase to pick up the change in EtcdDiscoveryDomain.

celebdor:
/lgtm

so basically we return NOERROR but we also do not return answers. I get it.
We should do that for OpenStack and the other platforms when they add IPv6 support.

Issue #1638, opened 2020-04-09
Support arch specific sysctl variables


Prashanth684:
What I would like to do is have sysctl variables defined but only for specific architectures like ppc64le/s390x.

After having a quick chat with @cgwalters , two options were:
1. Have a `templates/common/$arch` directory with sysctl variables which will only apply to the specific arches
2. Add an `architecture:` keyword in the yaml and parse that to get the architecture to apply the sysctl variables for. No architecture would mean across all arches.

Any thoughts/suggestions on ways to support this? Thank you!

PR #1696, opened 2020-04-29
Bug 1816254: pkg/controller/template: Support architecture specific files and units

Prashanth684:
For a managed baremetal install on ppc64le, there was a need to disable TSO (https://bugzilla.redhat.com/show_bug.cgi?id=1816254). To do this in an architecture specific manner, this PR introduces the ability to specify architecture directories (specified by the architecture uname) which can contain files and systemd units. These can be specified across the different platforms similar to how files and systemd units can be specified today.

Closes: #1638

### DIFF
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"sort"
 	"strings"
 	"text/template"
@@ -19,7 +20,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	runtime "k8s.io/apimachinery/pkg/runtime"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
@@ -35,6 +36,7 @@ type RenderConfig struct {
 const (
 	filesDir = "files"
 	unitsDir = "units"
+	archDir  = runtime.GOARCH
 
 	// TODO: these constants are wrong, they should match what is reported by the infrastructure provider
 	platformAWS       = "aws"
@@ -62,6 +64,11 @@ const (
 //                       /01-worker-kubelet/_base/files/random.conf.tmpl
 //                /master/00-master/_base/units/kubelet.tmpl
 //                                    /files/hostname.tmpl
+//
+// Architecture specific files and units can also be optionally specified. These will be applied only for the specific architecture.
+//
+// ex:
+// 	 templates/common/_base/ppc64le/files/disable-nic-tso.yaml
 //
 func generateTemplateMachineConfigs(config *RenderConfig, templateDir string) ([]*mcfgv1.MachineConfig, error) {
 	infos, err := ioutil.ReadDir(templateDir)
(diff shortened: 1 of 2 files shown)


Prashanth684:
/retitle Bug 1816254: pkg/controller/template: Support architecture specific files and units

kikisdeliveryservice:
cc: @cgwalters

pkg/controller/template/render.go
```diff
@@ -237,6 +244,15 @@ func generateMachineConfigForName(config *RenderConfig, role, name, templateDir,
 			}
 		}
 
+		// check architecture specific files - ignore if not present as they are optional
```
ashcrow, line 247: nit: This is done more than once. Consider using a function in the future.

cgwalters:
This is an operating system bug right?  We keep merging things into the MCO that are really OS bug workarounds I think because the MCO is the only convenient place that has a public Github integrated with OpenShift CI etc.

templates/common/_base/ppc64le/files/disable-nic-tso.yaml
```diff
@@ -0,0 +1,10 @@
+filesystem: "root"
+mode: 0744
+path: "/etc/NetworkManager/dispatcher.d/pre-up.d/disable-nic-tso.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # This is a workaround for BZ#1816254.If the ibmveth driver is used, disable TSO
+    if lsmod|grep -q 'ibmveth'; then
```
cgwalters, line 8: This could also live in RHCOS if we just did:

```
if "$(arch)" != "ppc64le"; then
  exit 0
fi
```
right?

kikisdeliveryservice: this would be great

kikisdeliveryservice:
> This is an operating system bug right? We keep merging things into the MCO that are really OS bug workarounds I think because the MCO is the only convenient place that has a public Github integrated with OpenShift CI etc.

can we decide on an approach, because @cgwalters makes an excellent point here. if it's better suited as an os fix, it makes a lot more sense to do it there.

/hold

Prashanth684:
> This is an operating system bug right? We keep merging things into the MCO that are really OS bug workarounds I think because the MCO is the only convenient place that has a public Github integrated with OpenShift CI etc.

Good point.The issue was  the ibmveth driver's default settings don't work which is sending large MTU packets so it had to be turned off. Not really sure if it is a bug as such. But I guess any Openshift specific setting would be in RHCOS.

cgwalters:
> The issue was the ibmveth driver's default settings don't work which is sending large MTU packets so it had to be turned off. Not really sure if it is a bug as such. But I guess any Openshift specific setting would be in RHCOS.

Wouldn't this problem affect the bootstrap node too (which the MCO doesn't change currently)?  That's the same reason why we [moved time sync](https://github.com/coreos/fedora-coreos-config/pull/367).

I guess the way I think of this is: if something needs to be templated based on *cluster data* then it *must* go in the MCO.  If a template file isn't actually a template (i.e. it's literal, as this is) then the argument for having it in the MCO weakens a lot.  But if it's something we only want to do when joining a cluster, then it still should go in the MCO.

If it affects the bootstrap, and is a sane thing to do always, then that strongly points towards doing it in the OS.

templates/common/_base/ppc64le/files/disable-nic-tso.yaml
ashcrow: True. I'd be good with that myself especially if we don't foresee a need to have a multiarch aware MCO.

Prashanth684:
> > The issue was the ibmveth driver's default settings don't work which is sending large MTU packets so it had to be turned off. Not really sure if it is a bug as such. But I guess any Openshift specific setting would be in RHCOS.
> 
> Wouldn't this problem affect the bootstrap node too (which the MCO doesn't change currently)? That's the same reason why we [moved time sync](https://github.com/coreos/fedora-coreos-config/pull/367).
> 
> I guess the way I think of this is: if something needs to be templated based on _cluster data_ then it _must_ go in the MCO. If a template file isn't actually a template (i.e. it's literal, as this is) then the argument for having it in the MCO weakens a lot. But if it's something we only want to do when joining a cluster, then it still should go in the MCO.
> 
> If it affects the bootstrap, and is a sane thing to do always, then that strongly points towards doing it in the OS.

I find myself agreeing with this argument. The setting change was made on master nodes and that made the packet drops disappear, but i don't see an issue with having it across all nodes. Just in case i'll check with the IBM team to see if they forsee any performance impact having it on worker nodes.

ashcrow:
Needs a rebase.

Prashanth684:
closing this as we don't seem to have a need for this today  -for this specific case, it can be done in rhcos.

PR #1698, opened 2020-04-30
[release-4.4] Bug 1829651: MC CRD: preserve unknown fields

runcom:
TL;DR; if somebody provides fields like disks and other we don't specify in the crd
those get dropped and can cause whatever issue you can think of.
I think adding that will preserve the unknown field.

Signed-off-by: Antonio Murdaca <runcom@linux.com>

holding pending tests

PR #1700, opened 2020-04-30
[release-4.4] crio: Add explicit paths for configuration settings

mrunalp:
There is a window during machine set scale up where a 1.14 crio
uses this configuration and it doesn't understand these empty
settings. Restoring the settings back to full paths till
we also update the boot image for machine sets as part of updates.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1829642

**- What I did**
Restore some crio settings to older values to make them forward compatible with older crio versions.

**- How to verify it**
Test that cri-o 4.2 can still understand the default rendered 4.4 crio.conf.

Add back explicit path for settings to be forwards compatible with older crio versions.

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -80,7 +80,7 @@ contents:
     # Path to the seccomp.json profile which is used as the default seccomp profile
     # for the runtime. If not specified, then the internal default seccomp profile
     # will be used.
-    seccomp_profile = ""
+    seccomp_profile = "/etc/crio/seccomp.json"
 
     # Used to change the name of the default AppArmor profile of CRI-O. The default
     # profile name is "crio-default-" followed by the version string of CRI-O.
@@ -196,7 +196,7 @@ contents:
     # - runtime_root (optional, string): root directory for storage of containers
     #   state.
     [crio.runtime.runtimes.runc]
-    runtime_path = ""
+    runtime_path = "/usr/bin/runc"
     runtime_type = "oci"
     runtime_root = "/run/runc"
 
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -80,7 +80,7 @@ contents:
     # Path to the seccomp.json profile which is used as the default seccomp profile
     # for the runtime. If not specified, then the internal default seccomp profile
     # will be used.
-    seccomp_profile = ""
+    seccomp_profile = "/etc/crio/seccomp.json"
 
     # Used to change the name of the default AppArmor profile of CRI-O. The default
     # profile name is "crio-default-" followed by the version string of CRI-O.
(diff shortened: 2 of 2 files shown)


wking:
New round of CI now that #1698 has landed.

kikisdeliveryservice:
Verified that in 4.2:
```
 seccomp_profile = "/etc/crio/seccomp.json"
runtime = "/usr/bin/runc"
```

wking:
Seems like this PR is narrowly reverting some portions of 606bd2bc8 (#1414), which was backported to 4.3 in 8afb65a90b2 (#1447).  And apparently not backported to 4.2 (possibly because of this issue).

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -80,7 +80,7 @@ contents:
     # Path to the seccomp.json profile which is used as the default seccomp profile
     # for the runtime. If not specified, then the internal default seccomp profile
     # will be used.
-    seccomp_profile = ""
+    seccomp_profile = "/etc/crio/seccomp.json"
```
wking, line 83: Oh, wait.  This isn't a revert of 606bd2bc8a (#1414).  Before #1414, this line was `# seccomp_profile...` (commented out).  In #1414 it became master's current empty string.  With this commit it is becoming an uncommented, nonempty value.  So was there a default baked into CRI-O at some point, but no longer?  And now we need to provide the old default explicitly here to support CRI-O that are recent enough to have dropped the baked-in-default?  But bug 1829642 is about supporting older CRI-O?  I'm confused about how these pieces fit together.

kikisdeliveryservice: it was commented out further back: https://github.com/openshift/machine-config-operator/commit/69025e8e8c82ed6d188eb0e409e8148da09ac3b2#diff-5c79cb2bed55d5971e8178e82d18cdc9

wking: Ah, ok.  And looks like #1216 went out in 4.3 and was never backported to 4.2.

kikisdeliveryservice: pretty sure that 4.2/1/.12 fails if these are empty so it cant be backported

kikisdeliveryservice:
just adding for safety.. 

/hold

sdodson:
It seems like we should include this on master branch until such a time that we're managing the machineset image on updates. There's no assurances that will happen before 4.5 so we may run into 4.2 to 4.3 to 4.4 to 4.5 upgrades with a 4.2 bootimage is still in use during scaleup.
/cherrypick master

templates/master/01-master-container-runtime/_base/files/crio.yaml
umohnani8: Yeah this is needed, so that cri-o 1.14 can use this config and not fail to start up like it was.

kikisdeliveryservice:
unlinked from https://bugzilla.redhat.com/show_bug.cgi?id=1829642 as we have #1706

ashcrow:
Closed per https://github.com/openshift/machine-config-operator/pull/1707#issuecomment-622383876

PR #1702, opened 2020-04-30
Bug 1817455: pkg/daemon: best effort restore or delete

runcom:
Turns out files under /usr/bin for instance, or whatever not caught
by rpm -qf or /usr/etc/ still causes upgrade failures.
I'm chaing this whole backup story to be a best effort or delete
mechanism as I think an upgrade should be able to fix this issue.

More info in the attached bug on what went wrong lately

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -779,6 +779,20 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error 
 	return nil
 }
 
+func (dn *Daemon) deleteStaleConfigFile(path string) error {
+	glog.V(2).Infof("Deleting stale config file: %s", path)
+	if err := os.Remove(path); err != nil {
+		newErr := fmt.Errorf("unable to delete %s: %s", path, err)
+		if !os.IsNotExist(err) {
+			return newErr
+		}
+		// otherwise, just warn
+		glog.Warningf("%v", newErr)
+	}
+	glog.Infof("Removed stale file %q", path)
+	return nil
+}
+
 // deleteStaleData performs a diff of the new and the old config. It then deletes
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -826,16 +850,9 @@ func (dn *Daemon) deleteStaleData(oldConfig, newConfig *mcfgv1.MachineConfig) er
...
 			}
-			glog.V(2).Infof("Deleting stale config file: %s", f.Path)
-			if err := os.Remove(f.Path); err != nil {
-				newErr := fmt.Errorf("unable to delete %s: %s", f.Path, err)
-				if !os.IsNotExist(err) {
-					return newErr
-				}
-				// otherwise, just warn
-				glog.Warningf("%v", newErr)
+			if err := dn.deleteStaleConfigFile(f.Path); err != nil {
```
ashcrow, line 853: +1 to making this a function

pkg/daemon/update.go
```diff
@@ -779,6 +779,20 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	glog.V(2).Infof("Deleting stale config file: %s", path)
+	if err := os.Remove(path); err != nil {
+		newErr := fmt.Errorf("unable to delete %s: %s", path, err)
+		if !os.IsNotExist(err) {
+			return newErr
+		}
+		// otherwise, just warn
+		glog.Warningf("%v", newErr)
+	}
+	glog.Infof("Removed stale file %q", path)
```
ashcrow, line 792: nit: Technically this should say something like "Stale file %q no longer exists" since it may not have been removed, but that's not a big deal.

runcom:
Not needed. Keep holding.

PR #1670, opened 2020-04-21
Bug 1802534: gcp-routes: move to MCO, implement downfile, tweak timing

squeed:
Move gcp-routes to the MCO, call it openshift-gcp-routes. Also, improve the behavior significantly:

1. always accept connections from external sources
2. use downfiles instead of stopping the service
3. implement the equivalent in gcp-routes-controller.

PR #1703, opened 2020-04-30
Ignition spec v2/v3 dual support

LorbusChris:
**- What I did**
- Added functionality to parse Ignition spec version <= 2.2, 3.0 and 3.1
- Added transforming function to the MCS to allow serving different config specs for different Ignition User Agents.

**- How to verify it**
- OCP: CI
- OKD: manual testing

Added Ignition spec v2/v3 dual support

/cc @runcom @yuqi-zhang @ashcrow


LorbusChris:
I'll work on fixing `verify` next week.
Once this is all green I'd like to use this as base for the next subsequent OKD Beta release.
(there are 3 or 4 more small commits that will need to be rebased on top of this for that)
/cc @vrutkovs

vrutkovs:
Seems we're hitting https://bugzilla.redhat.com/show_bug.cgi?id=1815799 here
/retest

LorbusChris:
Looks like this code works in OCP.

The last commit is optional -- the migration of the types used internally to spec v3.1 is not strictly necessary at this time as the MCS only supports a subset of Ignition anyway.

For this to work with OKD on GCP, the gcp-routes script needs to be moved into the MCO first (https://github.com/openshift/machine-config-operator/pull/1670)

vrutkovs:
Verified that this pr would provision FCOS nodes too - great work!

LorbusChris:
@vrutkovs I've updated the PR with support for spec 3.0 by chaining the parsers

vrutkovs:
Tried the latest version on OKD - I get
```
I0528 21:50:53.917132       1 bootstrap_server.go:84] reading file "/etc/mcs/bootstrap/machine-configs/rendered-master-49c9ddc9c1c829905226d5a1be650a38.yaml"
E0528 21:51:01.817912       1 api.go:131] failed to translate {master} config: failed to convert Ignition config spec v2 to v3: unable to convert Ignition spec v2 config to v3: error at $.systemd.units.7: duplicate entry defined
E0528 21:51:02.000549       1 api.go:131] failed to translate {master} config: failed to convert Ignition config spec v2 to v3: unable to convert Ignition spec v2 config to v3: error at $.systemd.units.7: duplicate entry defined
E0528 21:51:02.036623       1 api.go:131] failed to translate {master} config: failed to convert Ignition config spec v2 to v3: unable to convert Ignition spec v2 config to v3: error at $.systemd.units.7: duplicate entry defined
```
Maybe additional patches on my side broke it, but its unlikely. Unit #8 in the list is `kubelet.service` - perhaps kubelet dropin needs to be renamed?

LorbusChris:
I think the issue I hit here earlier is that changes to MCD aren't available in the CI env.

I have removed the last commit for migrating the internal state to use spec v3.
For it to pass CI, the v3 parsing support added to the MCD in this PR has to land in an RHCOS image first.

This showcases one of the problems that version skew between MCO and MCD can create -- IMO it would be goo to move to pulling the MCD from the MCO image instead of baking it into the OS image compose.

ashcrow:
```
level=error msg="Error: Unable to find matching route for Route Table and destination CIDR block (0.0.0.0/0)."
level=error
level=error msg="  on ../tmp/openshift-install-619500295/vpc/vpc-public.tf line 34, in resource \"aws_route\" \"igw_route\":"
level=error msg="  34: resource \"aws_route\" \"igw_route\" {" 
```

LorbusChris:
hm that looks like a Terraform issue (throttling).
Let's see if it reproduces
/retest

pkg/server/api.go
```diff
@@ -136,6 +148,72 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	v1 := semver.New("1.0.0")
+	v2 := semver.New("2.0.0")
+	v3 := semver.New("3.0.0")
+	var reqCfgVersion semver.Version
+	if ignSemver.LessThan(*v1) {
+		// Ignition v0.x supprts config spec v2.x
+		reqCfgVersion = *v2
+	} else if !ignSemver.LessThan(*v2) && ignSemver.LessThan(*v3) {
```
yuqi-zhang, line 169: minor: since there is no such thing as ignition 1.x I suppose we can just check `ignSemver.LessThan(*v3)`

pkg/server/api.go
```diff
@@ -115,11 +123,15 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	data, err := json.Marshal(conf)
+	data, translated, err := translateConfigForIgnitionMiddleware(conf, useragent)
```
yuqi-zhang, line 126: To make sure I am understanding this bit correctly, this is what happens when a node's ignition makes a request towards the MCS port: we hit this function, which passes effectively the rendered-config for the pool in ignition format to the node. This translation will essentially compare the ignition version being requested and the ignition version of the rendered config, and translate to match them. Today the rendered config is v2 so we won't see `ConvertIgnition2to3` being used

pkg/controller/common/helpers.go
```diff
@@ -204,22 +217,54 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+		return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3, rptV3)
+	}
+
+	return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v2 failed with error: %v\nReport: %v", err, rpt)
+}
+
+// ParseAndConvertConfig parses rawIgn for both V2 and V3 ignition configs and returns
+// a V2 or an error.
+func ParseAndConvertConfig(rawIgn []byte) (ign2types.Config, error) {
+	ignconfigi, err := IgnParseWrapper(rawIgn)
```
yuqi-zhang, line 251: what does the `i` at the end of the var stand for?

pkg/controller/common/helpers.go
```diff
@@ -146,13 +125,47 @@ func WriteTerminationError(err error) {
...
+
+	glog.V(4).Infof("Successfully translated Ignition spec v2 config to Ignition spec v3 config: %v", converted3)
+	return converted3, nil
+}
+
+// ConvertIgnition3to2 takes an ignition spec v3 config and returns a v2 config
+func ConvertIgnition3to2(ign3config ign3types.Config) (ign2types.Config, error) {
+	// Bad hack to convert spec 3.1 config to spec 3.0
+	// ign-convert doesn't currently support spec v3.1
+	// TODO(lorbus)
```
yuqi-zhang, line 152: So, I'm tempted to say we should just be on 3.0 for now to not add confusion, and wait until we're sure what we want for https://github.com/coreos/ign-converter/pull/6. Then we can add a hack to convert from 3.0 to 3.1 instead of the 3.1  to 3.0 here. I could be missing something, is there functionality you need from 3.1 or just trying to update to latest stable?

pkg/controller/common/helpers.go
yuqi-zhang: Oh nvm, `interface`

ashcrow:
Rebase needed.

pkg/server/api.go
LorbusChris: yes, that's exactly right! (we can't update the rendered ign to v3 right away, as the MCD changes need to land in the RPM in RHCOS first). Should I add commentary here?

pkg/server/api.go
LorbusChris: I have no real preference here, but I though it would be best to be as concise as possible

pkg/server/api.go
runcom: > as the MCD changes need to land in the RPM in RHCOS first

This might not be true anymore right? #1766 cc @cgwalters

yuqi-zhang: Sorry I'm not sure if I completely follow. The request should have come from ignition itself, and we're passing the data for the ignition binary to parse. So the MCD RPM doesn't touch this flow, right?

yuqi-zhang:
I'm :+1: to the changes. I'm basing the ignition spec 3 translation for the MCC on this PR, so we can either merge this first, or if you don't mind, I can put up that work as WIP hopefully by the end of this week, and we can see if the whole method makes sense

pkg/server/api.go
LorbusChris: Right, in thta case MCD is not involved: Serving spec 3 to nodes requesting it will work with this PR.

However the internal state/rendered MC will still be saved as spec v2 here, as our MCD doesn't understand spec 3 yet (for Day 2 ops). If we can merge in #1766 first, it will make things much clearer and saner.

ashcrow:
> I'm +1 to the changes. I'm basing the ignition spec 3 translation for the MCC on this PR, so we can either merge this first, or if you don't mind, I can put up that work as WIP hopefully by the end of this week, and we can see if the whole method makes sense

@LorbusChris do you have a preference?

LorbusChris:
Now that #1766 has merged, the follow-up commit I had split out into #1778 that actually "flips the switch" and migrates MCC to rendering spec v3, could now be part of this PR again.

@ashcrow
@yuqi-zhang I see no need to rush a merge here, additional sanity testing is always appreciated :) feel free to re-use this ofc!
/hold

pkg/controller/common/helpers.go
```diff
@@ -150,13 +129,47 @@ func WriteTerminationError(err error) {
...
+	converted3 := translate3.Translate(ign3_0config)
+
+	glog.V(4).Infof("Successfully translated Ignition spec v2 config to Ignition spec v3 config: %v", converted3)
+	return converted3, nil
+}
+
+// ConvertIgnition3to2 takes an ignition spec v3 config and returns a v2 config
+func ConvertIgnition3to2(ign3config ign3types.Config) (ign2types.Config, error) {
+	// Bad hack to convert spec 3.1 config to spec 3.0
+	// ign-convert doesn't currently support spec v3.1
```
kikisdeliveryservice, line 155: why do we even need to support 3.1?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -144,9 +144,9 @@ func findStorageConfig(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
 }
 
 func findRegistriesConfig(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
-	ignCfg, report, err := ign.Parse(mc.Spec.Config.Raw)
+	ignCfg, err := ctrlcommon.ParseAndConvertConfig(mc.Spec.Config.Raw)
 	if err != nil {
-		return nil, fmt.Errorf("parsing Registries Ignition config failed with error: %v\nReport: %v", err, report)
```
kikisdeliveryservice, line 149: Just to confirm (bc i can't follow the new github diffs very well) : does this new returned error still return a report?

pkg/daemon/daemon.go
```diff
@@ -1275,9 +1276,9 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) error
 		return errors.Errorf("expected target osImageURL %s", currentConfig.Spec.OSImageURL)
 	}
 	// And the rest of the disk state
-	currentIgnConfig, report, err := ign.Parse(currentConfig.Spec.Config.Raw)
+	currentIgnConfig, err := ctrlcommon.ParseAndConvertConfig(currentConfig.Spec.Config.Raw)
```
kikisdeliveryservice, line 1279: same Q re: report

pkg/server/api.go
kikisdeliveryservice: If there's no ignition 1, i think @yuqi-zhang suggestion is less confusing (I was confused).

pkg/controller/common/helpers.go
```diff
@@ -150,13 +129,47 @@ func WriteTerminationError(err error) {
...
+	}
+	ign3_0config, rep, err := ign3_0.Parse(raw3_0)
+	if rep.IsFatal() || err != nil {
+		return ign2types.Config{}, fmt.Errorf("SHOULD NEVER HAPPEN: failed to parse Ignition spec v3.1 to v3.0 config: %s\nreport: %v", err, rep)
 	}
-	glog.V(4).Infof("Successfully translated ignition V3 config to ignition V2 config: %v", converted2)
+
+	converted2, err := ignconverter.Translate3to2(ign3_0config)
+	if err != nil {
+		return ign2types.Config{}, errors.Errorf("unable to convert Ignition spec v3 config to v2: %v", err)
```
kikisdeliveryservice, line 169: Does Translate3to2 not return a report? Does the error contain the report?

kikisdeliveryservice:
> Now that #1766 has merged, the follow-up commit I had split out into #1778 that actually "flips the switch" and migrates MCC to rendering spec v3, could now be part of this PR again.
> 
> @ashcrow
> @yuqi-zhang I see no need to rush a merge here, additional sanity testing is always appreciated :) feel free to re-use this ofc!
> /hold

If this PR can be merged safely (after pending comments addressed) it makes much more sense to merge sooner rather than later to let it soak, instead of delaying. Jerry's PR can then just be based off of master.

pkg/controller/common/helpers.go
yuqi-zhang: I think support 3.1 itself isn't a hack, just that 3.0 > 3.1 doesn't translate very nicely. I suppose while we're developing this it'd be good to keep to latest stable

pkg/controller/common/helpers.go
```diff
@@ -208,22 +221,54 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+		return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3, rptV3)
+	}
+
+	return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v2 failed with error: %v\nReport: %v", err, rpt)
+}
+
+// ParseAndConvertConfig parses rawIgn for both V2 and V3 ignition configs and returns
+// a V2 or an error.
+func ParseAndConvertConfig(rawIgn []byte) (ign2types.Config, error) {
+	ignconfigi, err := IgnParseWrapper(rawIgn)
```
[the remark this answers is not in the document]
yuqi-zhang: stands for `interface`

pkg/controller/common/helpers.go
kikisdeliveryservice: Would make sense to just support latest stable instead of translating down to it (from a messaging perspective & from a support perspective).

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: i 100% would not have guess that at all :laughing:

pkg/controller/common/helpers.go
yuqi-zhang: Right we are supporting 3.1 directly via this change. The hack in place is for internal purposes to support 3.0 (which is used for the translator)

pkg/controller/container-runtime-config/helpers.go
yuqi-zhang: Not 100% sure on the reports but from what I recall they were mostly empty and the error message captured the reason

yuqi-zhang: On actually the report is now captured via the error message that the parse function gives

yuqi-zhang:
I think its a bit safer to merge this, and then do a `flip the switch` PR. My WIP for the switch flipping is in https://github.com/openshift/machine-config-operator/pull/1873.

/approve

kikisdeliveryservice:
rhel7 and ovn registry tests are meh rn

/skip

kikisdeliveryservice:
would like to see this pass tho

/test e2e-aws

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: ok perfect!

kikisdeliveryservice: wait.. which parse function?

kikisdeliveryservice: ok both IgnParseWrapper and ParseAndConvertConfig wrap the report in the error! :+1:

pkg/server/api.go
kikisdeliveryservice: > Should I add commentary here?

YES PLEASE!

kikisdeliveryservice:
I like this but would like the few outstanding questions/requests addressed first then we can merge! :)

pkg/server/api.go
kikisdeliveryservice: wait that pr merged...

pkg/controller/common/helpers.go
LorbusChris: The main reason to go directly to v3.1 is that installer uses (or will use rather) some 3.1 functionality (it currently shims to v2.4 on some platforms for lack of those features in v2.2). It would be nice if we could get rid of that shim there and agree on spec v3.1 being used universally. (v3.1 also has vastly simplified and improved typing).

It would be cleanest here to wait for https://github.com/coreos/ign-converter/pull/6 and use that to avoid this hack here.

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
LorbusChris: This seems to cause confusion, I may've been a little bit too lazy here :smile:.
Should I call it `iface` instead?

pkg/controller/common/helpers.go
kikisdeliveryservice: Ahh gotcha. So if we really want 3.1 and are just waiting for that PR I think we could wait a little bit. Seems like it's basically finished and just needs a final review? If it gets stuck we can stop waiting ofc but seems really close to merging?

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
kikisdeliveryservice: names are hard.. iface wfm

pkg/controller/common/helpers.go
kikisdeliveryservice: (also didnt realize 3.1 _was_ stable)

pkg/server/api.go
LorbusChris: simplified the conditional here

pkg/server/api.go
LorbusChris: added a comment here for clarification

pkg/controller/common/helpers.go
LorbusChris: updated ign-converter and removed the hack

pkg/controller/common/helpers.go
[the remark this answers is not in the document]
LorbusChris: changed to `ignConfigIface`

LorbusChris:
@kikisdeliveryservice @yuqi-zhang I think I've addressed all remaining comments. PTAL :)

LorbusChris:
@yuqi-zhang @kikisdeliveryservice I've simplified and deduped the code a bit more to make it easier to reason with.
The IgnParseWrapper function now parses _and_ translates to the canonical version that is rendered (still v2.2 with this PR).

pkg/controller/common/helpers.go
kikisdeliveryservice: yay!

yuqi-zhang:
Hm, I actually prefer having the `IgnParseWrapper` return ignition with the version in the raw config, and have the extra `parseAndConvertConfig` on top. We may still want ign2 for upgrading purposes between spec 2 and spec 3, such as validating on-disk state with the version with the rendered config ignition version originally rendered. This decreases chances of upgrade failures.

If you feel like having the canonical ignition version at all times, I'm not opposed to revising the above approach either. Or we can merge this PR and I can adapt out a separate function to give the corresponding version back.

LorbusChris:
@yuqi-zhang I'm not opposed to what you're saying, but I thought it wasn't really necessary in the scope of this PR.

If you don't mind merging this as-is, I'd prefer having the addition of a more flexible converter function in your follow-up.

pkg/server/api.go
```diff
@@ -138,6 +170,28 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 type healthHandler struct{}
 
+func getConfigSpecVersionFromUseragent(useragent string) (*semver.Version, error) {
```
LorbusChris, line 173: this func should really be called `deriveSpecVersionFromUseragent`

yuqi-zhang:
Ok did some manual tests, this should be good, let's merge

/lgtm

LorbusChris:
@yuqi-zhang updated the ign-converter dep again to include another fix.
Also made renamed the `getConfigSpecVersionFromUseragent` to `deriveSpecVersionFromUseragent` and made it stricter so we only serve config to user agents that actually support either v2.2 or v3.1

@vrutkovs fyi, I've now cherrypicked the commit from https://github.com/openshift/machine-config-operator/pull/1889 here to de-dupe the Ign v2 files.

LorbusChris:
reworked the deduping logic a bit again (see the 4th commit: https://github.com/openshift/machine-config-operator/pull/1703/commits/fe3b794b7bc94c85cd4c865cce75d622af14dc82). Although I think this should probably reside in the ign-converter itself

LorbusChris:
created https://github.com/coreos/ign-converter/pull/9 to ensure deduplication of v3 files and units

yuqi-zhang:
/hold

whoops there was a merge conflict

LorbusChris:
rebased. should be good to go now, just needs another `lgtm`
/hold cancel

pkg/server/api.go
```diff
@@ -138,6 +169,27 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// or an error for unsuported user agents.
+// Only Ignition versions that come with support for either spec v2.2 or spec v3.1 are supported
+func deriveSpecVersionFromUseragent(useragent string) (*semver.Version, error) {
+	ignVersionString := strings.SplitAfter(useragent, "/")[1]
+	ignSemver, err := semver.NewVersion(ignVersionString)
+	if err != nil {
+		return ignSemver, errors.Errorf("invalid version in useragent: %s", useragent)
+	}
+
+	if !ignSemver.LessThan(*semver.New("0.22.0")) && ignSemver.LessThan(*semver.New("1.0.0")) {
```
yuqi-zhang, line 182: Shouldn't this be Greater than 0.22.0

pkg/server/api.go
```diff
@@ -138,6 +169,27 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		// versions [0.22.0;1) support config spec v2.2, and
+		return semver.New("2.2.0"), nil
+	} else if !ignSemver.LessThan(*semver.New("2.3.0")) && ignSemver.LessThan(*semver.New("3.0.0")) {
```
yuqi-zhang, line 185: Same as above, greater than 2.3.0

pkg/server/api.go
yuqi-zhang: oh wait nvm I mis-read

PR #1708, opened 2020-05-01
[FCOS] Control amount of replicas in etcd-quorum-guard deployment

vrutkovs:
**- What I did**
* Moved etcd-quorum-guard deployment under MCO control (instead of CVO). 
* Updated the code which applies MCD daemonset to apply etcd-quorum-guard deployment
* Updated amount of replicas in etcd-quorum-guard deployment - if CEO's unsupported non-HA option is enabled MCO would create a deployment with one replica.
* etcd-quorum-guard now uses `pod` image instead of cli

This allows provisioning single node clusters.

**- How to verify it**
Create a cluster with 0 workers, 1 master and apply CEO's manifest to allow less than 3 workers(https://github.com/openshift/cluster-etcd-operator/pull/279).

TODO:
* [x] Check if `cliImage` can be avoided and `infraImage` can be reused instead

This is submitted for `fcos` branch only, as master's implementation would depend on scalable control plane enhancement

/cc @LorbusChris 

Fixes https://github.com/openshift/okd/issues/162


LorbusChris:
I'd rather not introduce this now that were so close to merging `fcos` with `master` and instead wait for implementation of https://github.com/openshift/enhancements/pull/292

vrutkovs:
That's for 4.4, the enhancement won't be backported there anyway - and we can branch `fcos-4.4` later

ashcrow:
/cc'ing @cgwalters @dustymabe as well for review

LorbusChris:
this is only needed for OKD CRC, right?

vrutkovs:
That's not required for CRC, but with this PR you can setup a single-node cluster (still having bootstrap node), but that would be a full blown cluster, no operators disabled / ripped. This cluster, of course, cannot be updated

LorbusChris:
Just noting that this will need to be handled differently in master from 4.6 on, as we won't have an fcos fork anymore by then. Otherwise we'll have to regress on this in OKD 4.6.
/lgtm

vrutkovs:
/hold

```
level=fatal msg="failed to initialize the cluster: Cluster operator machine-config is reporting a failure: Failed to resync 0.0.1-2020-05-11-102411 because: failed to execute template: template: manifests/etcdquorumguard_deployment.yaml:54:24: executing \"manifests/etcdquorumguard_deployment.yaml\" at <.Images.KubeClientAgent>: can't evaluate field KubeClientAgent in type *operator.RenderConfigImages"
```

pkg/operator/sync.go
```diff
@@ -424,6 +425,16 @@ func (optr *Operator) syncMachineConfigDaemon(config *renderConfig) error {
 		return err
 	}
 
+	etcdQuorumGuardBytes, err := renderAsset(config, "manifests/etcdquorumguard_deployment.yaml")
+	if err != nil {
+		return err
+	}
+	etcdQuorumGuard := resourceread.ReadDeploymentV1OrDie(etcdQuorumGuardBytes)
```
LorbusChris, line 432: looks like this function can panic, maybe better use `resourceread.ReadDeploymentV1` and return an error?

vrutkovs: This is expected - operator should crash if the rendered version is not a valid deployment. Throwing a panic here is probably too harsh, but that's what other functions here do

vrutkovs:
Lets keep the hold for now, still testing this one

vrutkovs:
/hold cancel
/retest

Imagestream import issues

vrutkovs:
/hold

PDB is not valid :(

vrutkovs:
/retest

Storage flakes

vrutkovs:
/hold cancel

@LorbusChris ready for another review (added a default setting  for etcd-quorum-guard replicas)

PR #1711, opened 2020-05-04
Bug 1809345: templates: add etc-networkmanager-dispatcher.d-90-long-hostname.yaml

darkmuggle:
On GCP it is not uncommon to have DNS names that are too long.
Linux restricts the kernel hostname to being less than 63 characters.
The new template simply ensures that a hostname is set in the event that
NetworkManager is unable to do so.

Bug: 1809345

### DIFF
--- a/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
+++ b/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
@@ -1,0 +1,40 @@
+filesystem: "root"
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/90-long-hostname"
+contents:
+  inline: |
+    #!/bin/bash
+    #
+    # On Google Compute Platform (GCP) the hostname may be too long (>63 chars).
+    # During firstboot the hostname is set in the initramfs before NetworkManager
+    # runs; on reboot affect nodes use 'localhost'. This hook is a simple work
+    # around: if the host name is longer than 63 characters, then the hostname
+    # is truncated at the _first_ dot.
+    #
+    # Additionally, this hook does not break DNS or cluster DNS resolution,
+    # since NetworkManager sets the appropriate /etc/resolv.conf settings.
+
+    IF=$1
+    STATUS=$2
+
+    log() { logger --tag "network-manager/$(basename $0)" "${@}"; }
+
+    # capture all eligible hostnames
+    if [[ ! "$(/bin/hostname)" =~ (localhost|localhost.local) ]]; then
+        log "hostname is already set"
+        exit 0
+    fi
+
+    if [[ ! "$STATUS" =~ (up|hostname|dhcp4-change|dhcp6-change) ]]; then
+        exit 0
+    fi
+
+    default_host="${DHCP4_HOST_NAME:-$DHCP6_HOST_NAME}"
+    # truncate the hostname to the first dot and than 64 characters.
+    host=$(echo ${default_host} | cut -f1 -d'.' | cut -c -63)
+
+    if [ "${#default_host}" -gt 63 ]; then
+        log "discovered hostname is longer than than 63 characters"
+        log "truncating ${default_host} => ${host}"
+        /bin/hostnamectl --transient set-hostname "${host}"
+    fi


templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
```diff
@@ -0,0 +1,40 @@
...
+    STATUS=$2
+
+    log() { logger --tag "network-manager/$(basename $0)" "${@}"; }
+
+    if [ "${HOSTAME}" == "localhost" ]; then
+        log "hostname is already set"
+        exit 0
+    fi
+
+    if [[ ! "$STATUS" =~ (up|hostname|dhcp4-change|dhcp4-change) ]]; then
```
tedyu, line 29: dhcp4-change seems to be repeated.
Is this intentional ?

darkmuggle:
I'm going to file a bug with NM and start the ball rolling over there to get this fixed properly. In the meantime, we need something like this to fix GCP

cgwalters:
But per discussion in https://github.com/coreos/ignition-dracut/pull/156 it's required that the hostnames are routable.  Will GCP also handle lookups for these similarly truncated names?

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
darkmuggle: Ah, thank you. That was supposed to be `dhcp6`. Changed and repushed.

cgwalters:
For at least IPI GCP we should be in control of the network setup and ideally able to control what the DHCP server is sending the instance.

I only briefly glanced at https://cloud.google.com/compute/docs/internal-dns and haven't looked at what the installer terraform is doing, but I'd assume we could configure the infrastructure to fix this.

darkmuggle:
> But per discussion in [coreos/ignition-dracut#156](https://github.com/coreos/ignition-dracut/pull/156) it's required that the hostnames are routable. Will GCP also handle lookups for these similarly truncated names?

As I understood that discussion -- and plenty of slack commentaries -- the hostname needs to be resolvable. The hostname is still resolvable within the search path which is properly set in `/etc/resolv.conf`

@abhinavdahiya Can you confirm that truncating the hostname that the node level (NOT cluster) will work?

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
tedyu: I don't see difference after refreshing several times.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
```diff
@@ -0,0 +1,40 @@
...
+        exit 0
+    fi
+
+    default_host="${DHCP4_HOST_NAME:-$DHCP6_HOST_NAME}"
+    host=$(echo ${default_host} | cut -f1 -d'.')
```
yuqi-zhang, line 34: Although VERY unlikely the first dot may be over 63 characters, so maybe we should hard truncate at 63 chars or something, since the hostname will likely not make sense anyways

darkmuggle:
> For at least IPI GCP we should be in control of the network setup and ideally able to control what the DHCP server is sending the instance.
> 
> I only briefly glanced at https://cloud.google.com/compute/docs/internal-dns and haven't looked at what the installer terraform is doing, but I'd assume we could configure the infrastructure to fix this.

You are correct. Using internal DNS will work for new bootstraps; the request was for an OS solution  for upgrades. @abhinavdahiya argued strongly against using it as the solution.

cgwalters:
> The hostname is still resolvable within the search path which is properly set in /etc/resolv.conf

If that's the case, then GCP must be using the same truncation algorithm.


> You are correct. Using internal DNS will work for new bootstraps; the request was for an OS solution for upgrades. 

Ahhh...yes, that is an issue.

OK to be clear, I am totally :+1: on this if we've tested it to work (did you?).  And I guess that's the tricky aspect; we'd need to roll a new release image with this PR in it, and then try to create a cluster with openshift-install with a cluster name that provokes name length issues.

Or maybe it'd be possible to test this with an existing GCP cluster by editing the DNS setup to trigger it?

darkmuggle:
I tested this outside the scope of a cluster by:
- creating 90 character FQDN
- dropping the file in manually
- rebooting

Before the new dispatcher, I would get "localhost" as the hostname with a resolv.conf containing the search path and dns resolvers. After this new dispatcher, the hostname was set to the truncated named the same resolv.conf. 

:+1: to the testing path.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
ashcrow: Possibly adding `| cut -c -63` or ` | head -c 63` or  at the end to ensure it's no more than 63 characters.

abhinavdahiya:
Few things:

1. Does it make sense to add this to all platforms or just GCP? Because currently it does it for all.

2. Is this workaround, something we should be configure in MCO or should this be in rhcos

yuqi-zhang:
> Does it make sense to add this to all platforms or just GCP? Because currently it does it for all.

Since the issue itself arises from Linux kernel restrictions I it might be fine living under common. That said,

> If that's the case, then GCP must be using the same truncation algorithm.

Might not be the case for every platform.

> Is this workaround, something we should be configure in MCO or should this be in rhcos

Today we have this: https://github.com/coreos/afterburn/blob/master/dracut/30afterburn/afterburn-hostname.service

Maybe that would be a better place for this to live?

cgwalters:
> Might not be the case for every platform.

Right, we can only truncate the names if the DNS server is doing the same thing.  And I'm sure that's not the default.

> Today we have this: https://github.com/coreos/afterburn/blob/master/dracut/30afterburn/afterburn-hostname.service

Hmm...I think that's only used when we need to use an out-of-band (non-DHCP) mechanism to get the hostname.

darkmuggle:
I went back and forth on where to carry this I ultimately chose the MCO for a couple of reasons:
- changing this NM is _way_ harder and has a wider blast surface
- providing this via the MCO means that we don't have the FCOS/RHCOS delta problem
- if/when we get a proper "fix" elsewhere is easy to drop
- Afterburn doesn't handle the discovered hostname which is where this is coming from.

The bigger issue that OCP is relying on implicit hostnames via the PTR records when nothing else sets the hostname. But per [1] we have a collision: 
- we MUST handle hostnames up to 63 characters
- we SHOUILD handle hostnames up to 255 characters. 

The kernel and Systemd are compliant with the must. The issue is _how_ NM is setting the hostname by not ensuring that the components under it handle the assignment. Since NM is using PTR records to discover the hostname (I'm not sure if this behavior is covered by a spec).  

IMO we'll need this fix for pre-4.6 OCP and until we get proper NM/Systemd/Kernel support for the "should" part of the RFC. 

Finally, the reason I didn't scope the patch at GCP specifically, is that this condition can happen generically and is most likely to happen on bare-metal: just use the PTR record hostname with a character count more than 63. 

[1] https://tools.ietf.org/html/rfc1123#page-13

darkmuggle:
> Few things:
> 
> 1. Does it make sense to add this to all platforms or just GCP? Because currently it does it for all.

Taxonomically speaking, this condition is most often encountered on all GCP, but it NOT unique to GCP. Any environment where the host is using PTR record hostname discovery can result in this situation (such as UPI).  

I seriously considered putting within the scope of GCP, but per RFC 1123, anything less than 255 is valid, and no software is NOT required to handle it.  Fix this just for GCP seems a bit short-sighted. 

> 2. Is this workaround, something we should be configure in MCO or should this be in rhcos

Ultimately, I would like to see the upstreams of RHCOS (FCOS, NetworkManager or Systemd) come up with a proper fix, but given the blast radius and release times, its unlikely to happen soon. This fixes the immediate pain in a way that can be reversed quickly without baking it into the bootimages.

abhinavdahiya:
@darkmuggle 
https://github.com/openshift/machine-config-operator/pull/1711#issuecomment-623677329
and https://github.com/openshift/machine-config-operator/pull/1711#issuecomment-623696800 answer all my concerns from https://github.com/openshift/machine-config-operator/pull/1711#issuecomment-623629067

Thanks! i'm fine with the change. I can try and test this longer hostnames locally to verify.

abhinavdahiya:
Here's a test that is testing the long names using cluster-bot
```
/msg @cluster-bot launch openshift/installer#3544,openshift/machine-config-operator#1711 gcp
/msg @cluster-bot test e2e openshift/installer#3544,openshift/machine-config-operator#1711 gcp
```

here are jobs currently running https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/703 and https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/704

these should help figure out if the change is going to work..

abhinavdahiya:
both the CI jobs
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/704
and 
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/703

failed to complete installation because the works did not join the clusters, so maybe this fix is not yet working...

abhinavdahiya:
Okay i think I know what's up, the CSRs for the worker nodes are not approved because the machine-api doesn't add the vm name to the InternalDNS, therefore it doesn't get approved by the in-cluster-approver.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
```diff
@@ -0,0 +1,40 @@
...
+    # Additionally, this hook does not break DNS or cluster resolution,
+    # since NetworkManager sets the appropriate /etc/resolv.conf settings.
+
+
+    IF=$1
+    STATUS=$2
+
+    log() { logger --tag "network-manager/$(basename $0)" "${@}"; }
+
+    if [ "${HOSTAME}" == "localhost" ]; then
```
wking, line 24: is this supposed to be `!=`, or am I just misunderstanding?

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
wking: I still see `dhcp4-change` twice as well.  Forgot to commit or push?

abhinavdahiya:
https://github.com/openshift/cluster-api-provider-gcp/pull/88
okay, let's see how this new combination works,

```
/msg @cluster-bot test e2e openshift/installer#3544,openshift/machine-config-operator#1711,openshift/cluster-api-provider-gcp#88 gcp
```

kikisdeliveryservice:
> /retest ci/prow/e2e-metal-ipi

just fyi: the e2e-metal-ipi seems kind of flaky rn across prs (and not required) unless you think _this_ failure is directly related to your PR..

/test e2e-metal-ipi

abhinavdahiya:
Okay, nice this works with change from cluster-api-provider-gcp

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/708
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/707

Both succeeded.. so we need to merge https://github.com/openshift/cluster-api-provider-gcp/pull/88 with this change and we should be good.

NautiluX:
Nice! Thanks for following up on this issue.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
darkmuggle: Yikes....good catch. Fixed and pushed. Retetsting that now.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
darkmuggle: Added, and tested.

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
```diff
@@ -0,0 +1,40 @@
...
+        exit 0
+    fi
+
+    if [[ ! "$STATUS" =~ (up|hostname|dhcp4-change|dhcp6-change) ]]; then
+        exit 0
+    fi
+
+    default_host="${DHCP4_HOST_NAME:-$DHCP6_HOST_NAME}"
+    # truncate the hostname to the first dot and than 64 characters.
+    host=$(echo ${default_host} | cut -f1 -d'.' | cut -c -63)
```
ashcrow, line 34:
```
$ echo "averylonghostnamewhichshouldprobablybelessbecausereallyitsprettycrazy.to.have.such.long.names" | cut -f1 -d'.' | cut -c -63
averylonghostnamewhichshouldprobablybelessbecausereallyitsprett
$ echo -n "averylonghostnamewhichshouldprobablybelessbecausereallyitsprettycrazy.to.have.such.long.names" | cut -f1 -d'.' | cut -c -63 | wc -c
64
```
Enforces 64 char limit. However, the next check on default host is >63. Should this check really be limiting to 63 instead?

ashcrow: Other than this, LGTM!

darkmuggle: Ha! I did the same thing:
```
$ echo hi | cut -c -2 | wc -c
3
```
`wc -c` counts from 1

ashcrow: Ahh :+1:

PR #1712, opened 2020-05-05
Bug 1781575: Remove rejects from icmp ratemask

juanluisvaladas:
Fixes 1781575

**- What I did**:
Clears the 4th bit of the icmp_ratemask

**- How to verify it**:
cat /proc/sys/net/ipv4/icmp_ratemask must return a number which has the 4th bit set to 0.
Do a curl request to a service without endpoints. It should consistently fail with a connection rejected rather than a timeout.

Remove icmp rejects rate limit. This is necessary so that kubernetes services without endpoints are timely rejected.

### DIFF
--- a/templates/common/_base/files/sysctl-icmp-conf.yaml
+++ b/templates/common/_base/files/sysctl-icmp-conf.yaml
@@ -1,0 +1,6 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/sysctl.d/00-icmp.conf"
+contents:
+  inline: |
+    net.ipv4.icmp_ratemask = 6160


templates/common/_base/files/sysctl-icmp-conf.yaml
```diff
@@ -0,0 +1,6 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/sysctl.d/icmp.conf"
+contents:
+  inline: |
+    net.ipv4.icmp_ratemask = {{ .icmpRateMask }}
```
cgwalters, line 6: That's...going to have weird semantics.  This file ends up getting rendered based on the dynamic contents of the host that happens to be running the MCC.

It's probably simplest to just write a fixed value for now.

If we really can't write a fixed value, then this is going to need to be special cased in the MCD or so.

juanluisvaladas: I can set it to 6180 which will be correct for *our* SDNs, not quite sure about third parties. @danwinship any inputs here?

cgwalters: I don't mean to bounce this around (and I know I mentioned the MCO in the other PR) but I hadn't realized that the value needed to be *dynamic* in that sense - the MCO templates just don't work that way.

Isn't there a container that's run as part of the SDN that could do the equivalent of https://github.com/openshift/sdn/pull/135 but would run regardless of the SDN type?  Just an init container or something?

cgwalters:
Prior discussion in https://github.com/openshift/sdn/pull/135

templates/common/_base/files/sysctl-icmp-conf.yaml
danwinship: It's not that it needs to be dynamic, it's just that in theory, someone might be setting this sysctl for some other reason, and so one of us would end up stomping on the other one now. Whatever. Just give it an early name ("00-icmp.conf" or whatever) and if someone else wants to overwrite our value, let them.

danwinship: or do it via a script that can do "get/modify/write" rather than just "write"

danwinship:
Actually, I'm increasingly unconvinced of this whole thing. If the test is a problem, we should just skip it for now rather than rushing into a bad solution.

(I think the right fix may be that we change the test to make multiple attempts, so that the test will pass even if ICMP replies are being rate-limited.)

cgwalters:
> Actually, I'm increasingly unconvinced of this whole thing. If the test is a problem, we should just skip it for now rather than rushing into a bad solution.

> (I think the right fix may be that we change the test to make multiple attempts, so that the test will pass even if ICMP replies are being rate-limited.)

Sounds good to me!

juanluisvaladas:
> Actually, I'm increasingly unconvinced of this whole thing. If the test is a problem, we should just skip it for now rather than rushing into a bad solution.

Agreed.

> (I think the right fix may be that we change the test to make multiple attempts, so that the test will pass even if ICMP replies are being rate-limited.)

I don't think we should go that way. We can easily reproduce this issue a dozen failures in a row with curl. Adding retries may not make the test pass and even if it does it won't fix the issue which a customer has also reported on 4.2.

templates/common/_base/files/sysctl-icmp-conf.yaml
cgwalters: > or do it via a script that can do "get/modify/write" rather than just "write"

Right, so if we want to that path we can inject a systemd unit, but that then gets into "OK now how do I run code from the MCO" which is a messy topic; we have mostly been moving this type of stuff to pull containers from elsewhere.

Which goes to my comment earlier - the SDN runs as a daemonset already, isn't there a variant of openshift/sdn#135 that would run everywhere?

> Whatever. Just give it an early name ("00-icmp.conf" or whatever) and if someone else wants to overwrite our value, let them.

That also seems sane to me.

juanluisvaladas:
> Curious why we just disable the limit as opposed to reduce the rate limit. The default is 1000 which amounts to 1 per second. Would changing this to maybe 200, which would limit to 5 per second, get the needed results without opening us to a flood?

Because that may fix the issue but I can't be certain... I tried sniffing icmp traffic from literally every network interface on every single net namespace for several minutes, but I haven't caught a single ICMP packet  even after entirely disabling the ratelimit (except the ones I forced to be created).

Therefore I don't know how many packets are normally being created per second or if that number scales proportionally to something like the amount of pods in the node, pods in the cluster or nodes in the cluster. I don't  even know if there are that many rejects or if the packets that are being rate limited are other kind of icmp traffic. So because I can't be sure this seems like the only safe option I have.

We'll probably need some systemtap script or some sort of kernel tracing in order to understand the problem.

templates/common/_base/files/sysctl-icmp-conf.yaml
juanluisvaladas: > Which goes to my comment earlier - the SDN runs as a daemonset already, isn't there a 
variant of openshift/sdn#135 that would run everywhere?

We don't have a daemonset that would run on everywhere for every cluster, we'd need to patch it on kuryr or OVN as well, or add yet another daemonset.

> That also seems sane to me.
Does it seem sane to you in its current state? Note it's been modified to have a static value.

danwinship:
> I tried sniffing icmp traffic from literally every network interface on every single net namespace for several minutes, but I haven't caught a single ICMP packet even after entirely disabling the ratelimit (except the ones I forced to be created).

Given that we now believe that this happened because we bumped RHCOS to RHEL 8.2, that sounds like the rate-limiting code is just busted and is blocking everything in 8.2...

cgwalters:
/approve
based on the fact we now have a simple static value
https://github.com/openshift/machine-config-operator/pull/1712#discussion_r420342312

cgwalters:
> Given that we now believe that this happened because we bumped RHCOS to RHEL 8.2, that sounds like the rate-limiting code is just busted and is blocking everything in 8.2...

This is possible.

templates/common/_base/files/sysctl-icmp-conf.yaml
danwinship: > the SDN runs as a daemonset already, isn't there a variant of [openshift/sdn#135](https://github.com/openshift/sdn/pull/135) that would run everywhere?

(No part of https://github.com/openshift/sdn/ is every deployed into any cluster that is not `networkType: OpenShiftSDN`. Cluster-network-operator could do something but we don't currently have any daemonset that is deployed in 100% of configurations.)

juanluisvaladas:
> Given that we now believe that this happened because we bumped RHCOS to RHEL 8.2, that sounds like the rate-limiting code is just busted and is blocking everything in 8.2...

This affects RHEL 7 nodes too and older RHCOS versions with older kernel.

danwinship:
/hold
This is being fixed by backing out of RHEL 8.2; see link in bugzilla.

cgwalters:
CI failures look like CI infra issues.
/retest

danwinship:
> This affects RHEL 7 nodes too and older RHCOS versions with older kernel.

Is it *flaking* there, or failing 100% reliably? People haven't been very clear. I didn't realize until like an hour ago that this was literally failing all the time in 4.4.

juanluisvaladas:
> Is it flaking there, or failing 100% reliably? People haven't been very clear. I didn't realize until like an hour ago that this was literally failing all the time in 4.4.

I tested it manually and I can't say it was 100%, but it was definitely over 95%

mkumatag:
> > This affects RHEL 7 nodes too and older RHCOS versions with older kernel.
> 
> Is it _flaking_ there, or failing 100% reliably? People haven't been very clear. I didn't realize until like an hour ago that this was literally failing all the time in 4.4.

I have seen this test failing on ppc64le post `4.3.18` release continuously and even on s390x platform. I feel this is not a flaky test.

pecameron:
@juanluisvaladas eliminating the rate limit opens us to a DOS attack. The limits are there for a reason.

juanluisvaladas:
/close
we discarded this approach

PR #1714, opened 2020-05-05
Bug 1784201: lib/resourcemerge: set AdditionalTrustBundle

runcom:
Proxy day2 wasn't working because the MCO wasn't rolling the additionalTrustBundle to nodes. Now it does and day2 proxy is deployed.


runcom:
cc @danehans

runcom:
Tested on a cluster, working as expected now, `kube-controller-manager` is stuck at `Available=True,Progressing=True,Degraded=True` which looks totally unrelated to this as all other operators are doing just fine after the proxy rollout. This patch might have exposed a bug there (or it's just slow rn) or since my test uses 4.5.latest something else is going on. Let's see if tests pass here also.

runcom:
> @runcom: The specified target(s) for `/test` were not found.
> The following commands are available to trigger jobs:

right.... we need to enable that perhaps

PR #1715, opened 2020-05-06
Bug 1764001: pkg/daemon: rollback dropins and units

runcom:
Restarted from https://github.com/openshift/machine-config-operator/pull/1203 and that's why we have the dropins rename as well (which is something we should do anyway as we ship those).

Backup and restore (when needed) units and dropins

Also, we need https://github.com/openshift/machine-config-operator/pull/1716


runcom:
cc @yuqi-zhang

templates/common/_base/units/etc-systemd-system-crio.service.d-10-default-env.conf.yaml
```diff
@@ -0,0 +1,16 @@
+name: crio.service
```
runcom, line 1: I think these dropins are making installation fails 🤔 I'll dig into this

runcom:
something funny is happening, dropins files seem like they don't get written when empty - causing the cluster to fail. I'm looking into this.

/hold

runcom:
> something funny is happening, dropins files seem like they don't get written when empty - causing the cluster to fail. I'm looking into this.

this looks like the case, really funny.

runcom:
Tested the proxy functionality with this (since this is touching that area) and I can enable proxy day2 just fine 👍

pkg/daemon/update.go
```diff
@@ -1008,7 +1043,9 @@ func (dn *Daemon) writeUnits(units []igntypes.Unit) error {
 
 		if u.Contents != "" {
 			glog.Infof("Writing systemd unit %q", u.Name)
-
+			if err := createOrigFile(fpath); err != nil {
```
yuqi-zhang, line 1046: Hm, there is a regressive behaviour here:

Let's say for example, we have a systemd unit deployed to the system via a MCO that had no backups attached to it (call it MC1). At some point we upgrade to this PR, and then write a new systemd unit file to overwrite it (call it MC2).

When this createOrigFile is called, it sees that:
1. this file is present on disk
2. there is no orig/no noorig associated with it

It then believes that the MC1 created unit belongs to the originally shipped MCO and creates an orig file. and thus if we were ever to delete both MCs it would "restore" a systemd unit that it thinks was on disk but it wasn't, much like what happened with etcd for files.

I think we have to harden it by checking whether the file existed on disk prior to restore, like the fix we had for the files case.

runcom: Dang, right I think since units and dropins are under /etc we only need one check from the preiovus fix ti files. I’ll fix this thanks for catching it!

runcom: alrighty, so I've modified the code to instead create the orig for units/dropins _only_ if the file is in /usr/etc instead of the logic we have in delete which is definitely more complex. It should serve the same tho as we won't create the orig for units/dropins if the file isn't shipped by rhcos/fcos

ptal @yuqi-zhang

pkg/daemon/update.go
```diff
@@ -975,11 +1007,21 @@ func (dn *Daemon) disableUnit(unit igntypes.Unit) error {
...
+		return errors.Wrapf(err, "checking operating system")
+	}
 	for _, u := range units {
 		// write the dropin to disk
 		for i := range u.Dropins {
 			glog.Infof("Writing systemd unit dropin %q", u.Dropins[i].Name)
 			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
+			if _, err := os.Stat("/usr" + dpath); err == nil &&
+				(operatingSystem == machineConfigDaemonOSRHCOS || operatingSystem == machineConfigDaemonOSFCOS) {
+				if err := createOrigFile(dpath); err != nil {
```
yuqi-zhang, line 1021: So by doing this instead we're running into another erroneous (although no longer regressive?) behaviour:

If we look at the workflow I commented above (Create MC1, upgrade to this PR, create MC2), but with a slight difference: there originally exists a version of the service on disk.

So now when we get here, we realize that there was an original version and attempt to create a backup. Instead of backing up from what existed on disk at the beginning, we backup what MC1 dropped since we're copying from `$dpath` directly.

I think we need to modify behaviour so that we create the orig file from the `/usr/$dpath` location instead.

yuqi-zhang: We would probably need to modify the `createOrigFile` function to take a "from" location and a "to" location

runcom: so in this case, we always want the original from `/usr/$dpath` right? should I just `createOrigFile("/usr"+$dpath)`?

runcom:
```diff
diff --git a/pkg/daemon/update.go b/pkg/daemon/update.go
index 492e0f79..cadfc7d1 100644
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1018,7 +1018,7 @@ func (dn *Daemon) writeUnits(units []igntypes.Unit) error {
                        dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
                        if _, err := os.Stat("/usr" + dpath); err == nil &&
                                (operatingSystem == machineConfigDaemonOSRHCOS || operatingSystem == machineConfigDaemonOSFCOS) {
-                               if err := createOrigFile(dpath); err != nil {
+                               if err := createOrigFile("/usr"+dpath); err != nil {
                                        return err
                                }
                        }
@@ -1052,7 +1052,7 @@ func (dn *Daemon) writeUnits(units []igntypes.Unit) error {
                        glog.Infof("Writing systemd unit %q", u.Name)
                        if _, err := os.Stat("/usr" + fpath); err == nil &&
                                (operatingSystem == machineConfigDaemonOSRHCOS || operatingSystem == machineConfigDaemonOSFCOS) {
-                               if err := createOrigFile(fpath); err != nil {
+                               if err := createOrigFile("/usr" + fpath); err != nil {
                                        return err
                                }
                        }
```

yuqi-zhang: Wouldn't it back up `/usr/$dpath` so it writes back to `/usr/$dpath` when it comes time to restore? Then we don't restore `$dpath`

Basically we currently have a forced 1-to-1 between the location of file backed up and the origin location of it. What we want is to back up `/usr/$dpath` to `$dpath` when we restore, and not restore `/usr/$dpath` to `/usr/$dpath`

runcom: alrighty, I think I've modified createOrigFile to take a fromPath arg and backup _from_ there once we're backing up dropins and unit

runcom:
> Now one thing to note is that if a systemd unit was overwritten before this patch, we will not restore that file even with this patch in place. This is not a regressive behaviour but is something we can fix if we really wanted to (I don't mind either way). If you are ok with the way it is I can give it the lgtm.

@yuqi-zhang right, it's not a regression but probably something worth fixing for people coming from past releases. We'd have to implement what we did for files in `deleteStaleData` right? and it'll be easier somehow cause systems units/dropins live in `/etc` and we can lookup `/usr/etc` there as well, right?

pkg/daemon/update.go
```diff
@@ -906,6 +914,18 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *igntypes.Config) e
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
 			if _, ok := newDropinSet[path]; !ok {
+				if _, err := os.Stat(noOrigFileStampName(path)); err == nil {
```
ashcrow, line 917: nit: Done more than once. Consider making a function in the future.

yuqi-zhang:
> let's follow up on that as it's not a regression unless this patch makes it harder to fix that in a follow up even between releases

Yes, I think this is fine.

/lgtm

PR #1716, opened 2020-05-06
Bug 1764001: templates: move proxy dropins to dropins

runcom:
This will make sure we don't ship files that are dropins. If we do that
(like we're doing), someone could ship a dropin for the same service/same droping name
and the validation will fail as the "file" is validate first. See https://bugzilla.redhat.com/show_bug.cgi?id=1764001


runcom:
This is now in https://github.com/openshift/machine-config-operator/pull/1715

PR #1717, opened 2020-05-06
Bug 1832180: Exclude ovirt from syncing cloud-config sync

rgolangh:


runcom:
Pending BZ

/hold

sinnykumari:
we need to fix https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/template/render.go#L508 as well.

sinnykumari:
Valid bug exists
/hold cancel

PR #1719, opened 2020-05-08
Bug 1810680: make file diffs more user friendly

kikisdeliveryservice:
This is most straightforward fix as users are super confused by A/B instead of Actual/Expected - This can be better and imo isn't worth wrapping. 

Am considering using `go-cmp`(https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff) which diffs line by line (like a traditional diff) instead of just huge blocks as it's hard for users to see the differences between the files.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20191002201903-404acd9df4cc // indirect
 	github.com/golang/mock v1.3.1 // indirect
 	github.com/golangci/golangci-lint v1.18.0
+	github.com/google/go-cmp v0.3.1
 	github.com/google/renameio v0.1.0
 	github.com/googleapis/gnostic v0.3.1 // indirect
 	github.com/gostaticanalysis/analysisutil v0.0.3 // indirect
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -22,6 +22,7 @@ import (
 	ign "github.com/coreos/ignition/config/v2_2"
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
 	"github.com/golang/glog"
+	"github.com/google/go-cmp/cmp"
 	"github.com/openshift/machine-config-operator/lib/resourceread"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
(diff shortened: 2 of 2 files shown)


runcom:
> Am considering using go-cmp(https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff) which diffs line by line (like a traditional diff) instead of just huge blocks as it's hard for users to see the differences between the files.

I’d go for it as well 👍

kikisdeliveryservice:
weird changing an error message shouldn't make this test fail this:
```
		if strings.Contains(kernelInfo, "PREEMPT RT") {
			t.Fatalf("Node %s did not rollback successfully", node.Name)
}
```

sigh

/test e2e-gcp-op

kikisdeliveryservice:
```
evel=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-1dpnzvq9-28de9.origin-ci-int-gce.dev.openshift.com:6443 to initialize..."
level=info msg="Cluster operator insights Disabled is False with : "
level=info msg="Cluster operator kube-apiserver Progressing is True with NodeInstaller: NodeInstallerProgressing: 1 nodes are at revision 5; 2 nodes are at revision 7"
level=fatal msg="failed to initialize the cluster: Working towards 0.0.1-2020-05-28-181044: 87% complete, waiting on authentication"
2020/05/28 18:57:02 Container setup in pod e2e-gcp-upgrade failed, exit code 1, reason Error 
```

:frowning_face: 

/test e2e-gcp-upgrade

kikisdeliveryservice:
Ok cool, we now have a line by line diff we used to have and error message and then printing out the two files without indicating which lines mismatched, now we have an integrated diff. Will definitely help people see where the issue is a lot quicker!:
```
E0617 22:10:56.569787  134867 daemon.go:1446] content mismatch for file /etc/dummy.txt (-want +got):
  bytes.Join({
+ 	"kittycat",
  	"server foo.example.net maxdelay 0.4 offline",
-  	"server bar.example.net maxdelay 0.4 offline",
+ 	"doggo",
+ 	"",
  	"server baz.example.net maxdelay 0.4 offline",
  	"",
  }, "\n")
```
Yes I am terrible at picking out words to test.

kikisdeliveryservice:
So in order to test this, we have to create an on-disk mismatch.

First roll out a cluster. Then roll out an MC creating some file with some content. Then oc debug/node onto that node (master is easier) and manually change that same file content (creating the ondisk mismatch). Upgrade to a payload containing this fix (I just pulled it from ci `oc adm upgrade --force --allow-upgrade-with-warnings --allow-explicit-upgrade --to-image registry.svc.ci.openshift.org/ci-op-mnlq0h3y/release:latest` ) The upgrade should stall once it gets to the machine-config-operator, when it does `oc get mcp -oyaml` then find the daemon that matches the node you changed.. You should see the new diff in the mcd logs.

kikisdeliveryservice:
metal job is completely red and n/a

/override ci/prow/e2e-metal-ipi

PR #1721, opened 2020-05-10
[Baremetal, Openstack,Vsphere, Ovirt] Update fall and rise values in keepalived check scripts

yboaron:
Keepalived runs a check script every interval (1 Sec in our case) to calculate VRRP instance's status.
The check script must succeed 'rise' times for the instance to be considered  as 'healthy and fail 'fall'
time to be considered as 'non healthy'.

This change updates 'rise' and 'fall' values from the default (1 for both of them) to 3 and 2.

**- What I did**

**- How to verify it**


jcpowermac:
cc: @patrickdillon @mtnbikenc

pierreprinetti:
/test e2e-openstack
Our CI is currently unable to give reliable results, however putting this here ready for a /retest

yboaron:
This Keepalived failure detection timing being handled by https://github.com/openshift/machine-config-operator/pull/1744/ , so we can close this PR.

PR #1729, opened 2020-05-12
pkg/daemon: Add IgnitionVersion to Daemon

LorbusChris:
Populate with output of `ignition --version` command
from locally installed binary.

This might be useful.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -52,8 +52,12 @@ import (
 type Daemon struct {
 	// name is the node name.
 	name string
+
 	// OperatingSystem the operating system the MCD is running on
 	OperatingSystem string
+
+	// IgnitionVersion is the version of the installed Ignition binary on the system
+	IgnitionVersion string
 
 	// mock is set if we're running as non-root, probably under unit tests
 	mock bool
@@ -115,6 +119,8 @@ type Daemon struct {
 }
 
 const (
+	// pathIgnition is the path where the ignition binary resides
+	pathIgnition = "/usr/lib/dracut/modules.d/30ignition/ignition"
 	// pathSystemd is the path systemd modifiable units, services, etc.. reside
 	pathSystemd = "/etc/systemd/system"
 	// wantsPathSystemd is the path where enabled units should be linked
@@ -190,6 +196,7 @@ func New(
 	var (
 		osImageURL string
 		osVersion  string
+		ignVersion string
 		err        error
 	)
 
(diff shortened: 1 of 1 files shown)


sinnykumari:
Not against this but curious if we have any particular usecase where this will be useful? Knowing version of a binary should also be possible by mounting the desired OSContainer.

pkg/daemon/daemon.go
```diff
@@ -209,6 +216,13 @@ func New(
 			return nil, fmt.Errorf("error reading osImageURL from rpm-ostree: %v", err)
 		}
 		glog.Infof("Booted osImageURL: %s (%s)", osImageURL, osVersion)
+
+		ignVersionBytes, err := exec.Command(pathIgnition, "--version").Output()
+		if err != nil {
+			return nil, fmt.Errorf("error getting installed Ignition version: %v", err)
+		}
+		ignVersion = string(ignVersionBytes)
+		glog.Infof("Installed Ignition binary version: %s", ignVersion)
```
ashcrow, line 225: nit: I think this will read as follows:
```
Installed Ignition binary version: Ignition 0.35.1
```

That isn't really a problem, though if we expect other things to key off this as a version number it may be worth dropping the prefix. Example:

```go
ignVersion = strings.SplitAfter(string(ignVersionBytes), " ")[1]
```

LorbusChris: :+1:

LorbusChris:
CI already shows that this works:
```
daemon.go:225] Installed Ignition binary version: Ignition 0.35.1
```
(from: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1729/pull-ci-openshift-machine-config-operator-master-e2e-aws/7324/artifacts/e2e-aws/gather-extra/pods/openshift-machine-config-operator_machine-config-daemon-52q8c_machine-config-daemon.log )

I'll update per the suggestion above to only save the version number in the string.

LorbusChris:
> Not against this but curious if we have any particular usecase where this will be useful? Knowing version of a binary should also be possible by mounting the desired OSContainer.

@sinnykumari my foremost reasoning for this is the following:
In `RunFirstbootCompleteMachineconfig()` we call `canonicalizeEmptyMC(nil)`. In the ignition spec v2/v3 dual support case (from PR https://github.com/openshift/machine-config-operator/pull/1703) we wouldn't know what the canonical and supported one for the running system is that it has to create.

And additionally:
Over in https://github.com/openshift/machine-config-operator/pull/1703 I'm planning to add a field on the ControllerConfig to specify the requested spec version. If the requested version isn't the supported one, it would fail. However this is could actually be extended to fall back to the supported version no matter what is requested now that we have the ability to translate between versions.

PR #1730, opened 2020-05-13
cmd: Configure klog to match glog

wking:
To try and understand why compute nodes sometimes have a long delay between informer initialization and the initial MachineConfig retrieval in getStateAndConfigs.

https://bugzilla.redhat.com/show_bug.cgi?id=1834895

### DIFF
--- a/cmd/gcp-routes-controller/run.go
+++ b/cmd/gcp-routes-controller/run.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"crypto/tls"
-	"flag"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -18,7 +17,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 
-	"github.com/openshift/machine-config-operator/pkg/version"
+	"github.com/openshift/machine-config-operator/internal"
 )
 
 var (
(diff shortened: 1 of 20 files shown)


wking:
Repurposed this branch to just be about klog/glog compat.  This gives us the ability to use `-v=10` when calling our executables and get all of the klog goodness out of our vendored Kubernetes libraries.

I also drop some flag parsing which had been used by `version` commands which print the version and exist, and so never actually needed the logging stack to be configured.

wking:
[rhel7 job][1] failed, but shows that [MCD logs][2] still have the `Version: ...` string and such, and don't appear to be any more chatty than they used to be.  Presumably bumping `-v=..` on them (like dd9cfe02 or some subset of that previous commit) would give us all the klog volume we want for debugging tricky vendored-code behavior.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1730/pull-ci-openshift-machine-config-operator-master-e2e-aws-scaleup-rhel7/1925
[2]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1730/pull-ci-openshift-machine-config-operator-master-e2e-aws-scaleup-rhel7/1925/artifacts/e2e-aws-scaleup-rhel7/pods/openshift-machine-config-operator_machine-config-daemon-6z9zt_machine-config-daemon.log

cgwalters (approved):
I am not intimately familiar with glog vs klog, this overall looks like a nice cleanup though!

Adding 4.7 label.

PR #1732, opened 2020-05-14
Bug 1825967: revert: etcd quorum guard: don't set hostNetwork

hexfusion:
After a conversation with @eparis about the fact that this change adds a dependency on SDN to quorum-guard health checks for etcd I feel this should be reverted. The fear would be false negatives on health. Although the cluster would have other problems if SDN were down we don't want to triage that layer if we don't have to.

cc @vrutkovs @sdodson 

This PR reverts

- https://github.com/openshift/machine-config-operator/pull/1552
- https://github.com/openshift/machine-config-operator/pull/1648

## Note this change was not made in 4.4

### DIFF
--- a/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
+++ b/install/0000_80_machine-config-operator_07_etcdquorumguard_deployment.yaml
@@ -61,41 +61,27 @@ spec:
           name: kubecerts
         command:
         - /bin/bash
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
/assign @sdodson @vrutkovs 

PTAL

vrutkovs:
LGTM

Issue #1578, opened 2020-03-24
Machine config daemon times out quickly on drain leading to node degrade


yanirq:
**Description**
Description of problem:

When applying machine configuration changes the machine-config-daemon tries to train pods from the node updated.
The drain timeout seems to be too low or the retry mechanism is not sufficient enough to allow pods to be terminated.

Looking at the changes with using the k8s drain lib I can see that a global timeout was added with 20s:
https://github.com/openshift/machine-config-operator/blob/9428caeffffa9e02663e5e1f02100c1071558f92/pkg/daemon/daemon.go#L300 

After applying a config change on ip-10-0-138-19.us-east-2.compute.internal:

```
$oc get nodes
NAME                                         STATUS                        ROLES              AGE   VERSION
ip-10-0-129-202.us-east-2.compute.internal   Ready                         master             85m   v1.17.1
ip-10-0-133-119.us-east-2.compute.internal   Ready                         master             85m   v1.17.1
ip-10-0-138-19.us-east-2.compute.internal    NotReady,SchedulingDisabled   worker,worker-rt   72m   v1.17.1
ip-10-0-140-42.us-east-2.compute.internal    Ready                         worker             73m   v1.17.1
ip-10-0-152-39.us-east-2.compute.internal    Ready                         worker             72m   v1.17.1
ip-10-0-155-52.us-east-2.compute.internal    Ready                         master             85m   v1.17.1
```

```
$oc get mcp
NAME        CONFIG                                                UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT   AGE
master      rendered-master-11c876e2263782349eddf836a9aa4bf2      True      False      False      3              3                   3                     0                      104m
worker      rendered-worker-9b8741b610aba145903baa6a62e640f5      True      False      False      2              2                   2                     0                      104m
worker-rt   rendered-worker-rt-05d30a7542756a28cdd085fde085886e   False     True       True       1              0                   0                     1    

$oc describe mcp worker-rt
..
    Type:                  Updating
    Last Transition Time:  2020-03-24T11:03:26Z
[...]

PR #1739, opened 2020-05-19
Bug 1834948: increase eviction time to avoid preventable timeouts

kikisdeliveryservice:
Time of 20s was added when we switched to kubectl drain lib.  This results in consistent error-retries for router pod, which takes ~ 90s each time. Upped timeout to 90s to avoid these preventable and predictable errors.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -298,7 +298,7 @@ func (dn *Daemon) ClusterConnect(
 		IgnoreAllDaemonSets: true,
 		DeleteLocalData:     true,
 		GracePeriodSeconds:  -1,
-		Timeout:             20 * time.Second,
+		Timeout:             90 * time.Second,
 		OnPodDeletedOrEvicted: func(pod *corev1.Pod, usingEviction bool) {
 			verbStr := "Deleted"
 			if usingEviction {


kikisdeliveryservice:
looking at https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1739/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/2310/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-nwsnn_machine-config-daemon.log

I still see router timing out even at 45s less often but still happening...

sinnykumari:
Looks like 60 second is not enough as well for some pods to finish eviction https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1739/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/2315/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-twlk5_machine-config-daemon.log .

Not sure if it is good idea to keep increasing timeouts. Do we know which all pods could impact upgrade if eviction failed? It could be possible that underlying issue in https://github.com/openshift/machine-config-operator/issues/1578 is something different which is not captured by oc describe.

kikisdeliveryservice:
> Looks like 60 second is not enough as well for some pods to finish eviction https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1739/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/2315/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-twlk5_machine-config-daemon.log .
> 
> Not sure if it is good idea to keep increasing timeouts. Do we know which all pods could impact upgrade if eviction failed? It could be possible that underlying issue in #1578 is something different which is not captured by oc describe.

yes i noted this on the bug as i think this _is_ in fact a router issue not an mco issue as even increasing up to 60 resulted in same failures.

kikisdeliveryservice:
I don't  think that this PR should be used to close the bz as still seeing the issue, though I do think it's correct to increase to 45s as that is expected.

sinnykumari:
> I don't think that this PR should be used to close the bz as still seeing the issue, though I do think it's correct to increase to 45s as that is expected.

makes sense to me unless there is any other thoughts from the team

ashcrow:
> > I don't think that this PR should be used to close the bz as still seeing the issue, though I do think it's correct to increase to 45s as that is expected.
> 
> makes sense to me unless there is any other thoughts from the team

I also agree ... upping to 45s, which is what is currently expected, is a good change though it doesn't fix the issue at hand which, due to the great debugging here, looks like it may reside in the router, etc...

kikisdeliveryservice:
```
 level=info msg="Cluster operator authentication Progressing is True with _WellKnownNotReady: Progressing: got '404 Not Found' status while trying to GET the OAuth well-known https://10.0.0.4:6443/.well-known/oauth-authorization-server endpoint data"
level=info msg="Cluster operator authentication Available is False with : "
level=error msg="Cluster operator console Degraded is True with RouteHealth_StatusError: RouteHealthDegraded: route not yet available, https://console-openshift-console.apps.ci-op-42r0zyz8-1354f.origin-ci-int-gce.dev.openshift.com/health returns '503 Service Unavailable'" 
```
/test e2e-gcp-op

kikisdeliveryservice:
Something seems weird with our ci job.. masters not coming up (obviously not related to this pr).. will investigate.

kikisdeliveryservice:
seems good now with no affect on overall drain time and no weird router eviction errors that would cause a users/anyone to be confused...

/assign @runcom

PR #1740, opened 2020-05-20
Bug 1804067: Prevent etcd-member-add.sh from running on an active master

retroflexer:
Closes: #1804067

**- What I did**
Added an extra check to make sure etcd-member container is not already running on the master.

**- How to verify it**
Try to run etcd-member-add.sh script on an active etcd-member. It should print an error and exit.

Prevent users from running the etcd-member-add.sh script on a wrong master.

### DIFF
--- a/templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
+++ b/templates/master/00-master/_base/files/usr-local-bin-etcd-member-add-sh.yaml
@@ -41,6 +41,11 @@ contents:
     source "/usr/local/bin/openshift-recovery-tools"
     
     function run {
+      if [ ! -z "$(crictl ps -name etcd-member --state running -q)" ]; then
+        echo "etcd-member container is already running and active on this master."
+        echo "Please make sure you run this script on the new master host to be added to the etcd cluster. Exiting!"
+        exit 1
+      fi
       init
       dl_etcdctl
       backup_manifest


retroflexer:
Can we override the dependency on 4.4 and 4.5 bugs? This bug is limited to 4.3, as the script etcd-member-add.sh is no longer extant in 4.4 and 4.5 releases, and therefore there is no dependent bug targeting a release in 4.4.0, 4.4.z.

ashcrow:
> Can we override the dependency on 4.4 and 4.5 bugs? This bug is limited to 4.3, as the script etcd-member-add.sh is no longer extant in 4.4 and 4.5 releases, and therefore there is no dependent bug targeting a release in 4.4.0, 4.4.z.

Was there any similar fixes done in 4.5/4.4 just possibly not in the same area or is this literally a 4.3 bug _only_?

hexfusion:
> > Can we override the dependency on 4.4 and 4.5 bugs? This bug is limited to 4.3, as the script etcd-member-add.sh is no longer extant in 4.4 and 4.5 releases, and therefore there is no dependent bug targeting a release in 4.4.0, 4.4.z.
> 
> Was there any similar fixes done in 4.5/4.4 just possibly not in the same area or is this literally a 4.3 bug _only_?

Scaling up, which is the focus of this script is handled by the operator in 4.4+.

kikisdeliveryservice:
Argh, the bot removed the 4.3.z target...

ashcrow:
> > Was there any similar fixes done in 4.5/4.4 just possibly not in the same area or is this literally a 4.3 bug _only_?
> 
> Scaling up, which is the focus of this script is handled by the operator in 4.4+.

BZ overridden.

sdodson:
/retitle Bug 1804067: Prevent etcd-member-add.sh from running on an active master

PR #1745, opened 2020-05-20
DNM: Test keepalived without api healthcheck

cybertron:
To see if unnecessary failovers are triggering our api flakes. The
    haproxy healthcheck should cover most cases where we actually need
    to failover.


kikisdeliveryservice:
im assuming DNM = do not merge?

/hold

cybertron:
This didn't work at all. haproxy comes up well before the api, and if we move the vip at that point it completely breaks the deployment.

PR #1753, opened 2020-05-26
[release-4.2] Bug 1840461: Update the crio.conf template to have log_dir

umohnani8:
**- What I did**
Update the crio.conf template to add log_dir and its
value. This ensures that we don't break on an empty value
when a ctrcfg is created and the cluster is upgraded to 4.3.

**- How to verify it**
Create a ctrcfg CR and then upgrade to 4.3.x

Update the crio.conf template to include log_dir


umohnani8:
@runcom PTAL

haircommander:
LGTM

umohnani8:
@ashcrow this is also ready for the next backport window, PTAL

PR #1754, opened 2020-05-26
WIP: Bug 1840881: KubeletConfigController: rework to sync overall state for pool

yuqi-zhang:
Currently the KubeletConfigController operates as follows: When the
controller receives a add/update/delete operation:

 - if add/update, re-generate the template confs and merge the added/
   updated conf into it (only the latest one) and create/update the
   corresponding MC for pool
 - if delete, delete the whole MC for pool

This means it will only allow 1 kubeletconfig per pool. This also
means that updates to an existing kubeletconfig that modifies a
pool will break the logic, and is an upgrade blocker. See:
https://bugzilla.redhat.com/show_bug.cgi?id=1829116

With this rework, we aim to make kubeletconfigs re-sync the overall
state every event, much like we do for MachineConfigs. This means
that all kubeletconfigs belonging to a pool gets synced and the
MachineConfig is updated. When the KC is deleted, the sync will
catch that and, if that was the last KC associated with a pool,
also delete the corresponding MC.

Tested so far to work with add/adding multiple/deleting/adding
to multiple pools simultaneously, and changing pools for existing
configs (see https://bugzilla.redhat.com/show_bug.cgi?id=1829116)

There are 2 observed issues:
1. I've currently commented out the old deleting path which makes
use of finalizers. I think I need to modify that to not be a
regression in one case: if the user, for some reason, delete a
MachineConfigPool. This is because currently the MC that KCC
manages is named based on UID and is not managed otherwise.
For normal operations (upgrades, etc.) this will not be a problem
as the MCP UID does not change. The only time it would change is
if the object is deleted and re-created. This would cause us
to essentially lose a generated MC related to KC, which is
an issue obviously. Probably should rework finalizers to account
for this case.

2. Rarely, an update to just one pool causes the other to also
trigger updates for some nodes. BUT the machineconfigpool doesn't
change desired config, so I'm not sure what is triggering it.
This does not cause errors but obviously we don't want to incur
an unnecessary reboot.

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -17,7 +17,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/jsonmergepatch"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
Marking WIP due to the issues. PTAL @rphillips

runcom:
@umohnani8 this is what we talked about ~2 weeks ago on Slack - ptal

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -211,32 +210,38 @@ func (ctrl *Controller) deleteKubeletConfig(obj interface{}) {
 			return
 		}
 	}
-	if err := ctrl.cascadeDelete(cfg); err != nil {
-		utilruntime.HandleError(fmt.Errorf("couldn't delete object %#v: %v", cfg, err))
-	} else {
-		glog.V(4).Infof("Deleted KubeletConfig %s and restored default config", cfg.Name)
-	}
+	glog.V(4).Infof("Deleting KubeletConfig %s", cfg.Name)
+	ctrl.enqueueKubeletConfig(cfg)
```
runcom, line 214: I think the overall approach work but if this is not as urgent (as it's not a regression) maybe we want to have a better mapping between KC and MC? Right now we have N KC -> 1 MC being re-generated everytime. We really want to have N KC -> 1 MC in order to make this MC-like and in order for the correct deletion of the MCs using finalizers

PR #1678, opened 2020-04-23
templates: Migrate from Container Linux to Fedora CoreOS config format

LorbusChris:
Migrate from Container Linux to Fedora CoreOS config format
~~Right now this retains support for CL configs although I think this might not be necessary.~~

@openshift/openshift-team-mco

~~Blocked on: https://github.com/openshift/machine-config-operator/pull/1677~~
/hold
until master/4.6 opens

**- What I did**
- templates: Merge MCD host service and dropin into one config
- templates: Name all systemd unit configs *.service.yaml
- pkgs/controller: Add support for transpiling Fedora CoreOS configs
Adds support to the template renderer for transpiling Fedora CoreOS
configs to Ignition spec v2 config
- Update vendor
- templates: Migrate to FC config format
Drops the the `filesystem` field on all templates to make them
adhere to FC config format.

**- How to verify it**
CI e2e

**- Description for the changelog**
templates: Migrate to FC config format

PR #1755, opened 2020-05-27
templates: Merge MCD host service and dropin into one config

LorbusChris:
/cc @runcom 

I'm not entirely sure whether the template needs a `.yaml` suffix for the template replacement to work. Either way, it might be cleaner to rename those `.service` files to `.service.yaml` to reflect the fact that they actually contain yaml.

**- What I did**
In order to not create two separate configs for the
`machine-config-daemon-host.service` unit's contents and dropins,
merge them into one.

Having separate config files for a systemd unit and its dropins works
with the Container Linux Config Transpiler, but it won't work with the
Fedora CoreOS Config Transpiler as Ignition Config Spec v3 does not
allow for nor merge duplicate unit definitions.

**- How to verify it**
- CI
- verify the `machine-config-daemon-host.service` dropin contains the right proxy config

templates: Merge MCD host service and dropin into one config


LorbusChris:
This is included in https://github.com/openshift/machine-config-operator/pull/1678 now
/close

PR #1756, opened 2020-05-27
Bug 1826150: Bootstrap Machine Config ordering issue

runcom:
This is a funny bug...and it's tricky to explain as well but I'll try.

Scenario: installation with custom chrony conf user provided with name (remember it):

99-master-etc-chrony-conf

Install the cluster with the above and 1/(probably an high number) fails with the usual:

Node master-0-0.ocp-edge08-0.qe.lab.redhat.com is reporti
ng: \"machineconfig.machineconfiguration.openshift.io \\\"rendered-master-a2dacf21a0e43d86e8495317ad36ae98\\\" not found\"

So master and workers go up, but masters somehow have a diff between what the bootstrap generated
for rendered-master and what the MCO did once it was up and running in the cluster.

Now, taking a look at the diff between what's on bootstrap and what the MCO generated yields
nothing in file contents diff, everything is same EXCEPT a file name ordering.
File name ordering means that the slice that contains the files and their names is different.
Upon inspecting the renderd-master on the bootstrap, I found the following:

        path: /etc/containers/registries.conf
        path: /etc/chrony.conf

So, on the bootstrap I had the above list (registries comes first, then chrony).
Upon inspecting the rendered master on the MCO, I found:

        path: /etc/chrony.conf
        path: /etc/containers/registries.conf

Boom, it's clear now that the real diff is how the files are ordered within the rendered machine config itself.

Let's look at what we do in `RunBootstrap.generateRenderedMachineConfig` _before_ my patches:

- get all the machine configs
- sort them by name
- merge the result together to create the rendered config

and then we do this:

		pool.Spec.Configuration.Name = generated.Name
		pool.Status.Configuration.Name = generated.Name

So what we're doing above in the bootstrap is just setting the generated config name. That yields
the following manifest on bootstrap:

[root@localhost bootstrap]# cat machine-pools/master.yaml
kind: MachineConfigPool
metadata:
  creationTimestamp: null
  labels:
    machineconfiguration.openshift.io/mco-built-in: ""
    operator.machineconfiguration.openshift.io/required-for-upgrade: ""
  name: master
spec:
  configuration:
    name: rendered-master-a2dacf21a0e43d86e8495317ad36ae98
  machineConfigSelector:
    matchLabels:
  nodeSelector:
    matchLabels:
      node-role.kubernetes.io/master: ""
  paused: false
status:
  conditions: null
  configuration:
    name: rendered-master-a2dacf21a0e43d86e8495317ad36ae98
  degradedMachineCount: 0
  machineCount: 0
  readyMachineCount: 0
  unavailableMachineCount: 0
  updatedMachineCount: 0

Now let's look at the above and we can see that the main difference from the master MCP
in the cluster is that:

- it's missing the Source field which contains the configs it was generated from
- it's missing the UID (!!!)

Now, let's dive into how the registries MC name is constructed:

	return fmt.Sprintf("99-%s-%s-registries", pool.Name, pool.ObjectMeta.UID)

Ok, so for the registries MC on the bootstrap, since we don't have an UUID we get this name:

99-master--registries

So, when that MCP manifest it's applied to kube, it auto gets an UUID, and the MCO registies controller generates:

99-master-f7818c63-cd84-4ca6-954c-274fbdc5f71d-registries

So, those are different but they 1/{high number} it's ok and I'll tell you why.

Remember the user provided the chrony conf? the file name was `99-master-etc-chrony-conf`.
So, on the boostrap, following how `generateRenderedMachineConfig` sort the names, we get this order:

99-master--registries
99-master-etc-chrony-conf

On the MCO we get:

99-master-etc-chrony-conf
99-master-f7818c63-cd84-4ca6-954c-274fbdc5f71d-registries

So they're flipped!!! and that's because the generated UUID for the registries conf starts with "f" which is > than "e" (ofc)

So, everything should be clear now, and the bug is that we're not providing an UUID at bootstrap
which COULD lead to ordering issues and thus, masters won't be able to re-construct what the boostrap did.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -295,8 +295,8 @@ func MachineConfigFromRawIgnConfig(role, name string, rawIgnCfg []byte) (*mcfgv1
 }
 
 // GetManagedKey returns the managed key for sub-controllers, handling any migration needed
-func GetManagedKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, suffix, deprecatedKey string) (string, error) {
-	managedKey := fmt.Sprintf("99-%s-%s", pool.Name, suffix)
+func GetManagedKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, prefix, suffix, deprecatedKey string) (string, error) {
+	managedKey := fmt.Sprintf("%s-%s-%s", prefix, pool.Name, suffix)
 	// if we don't have a client, we're installing brand new, and we don't need to adjust for backward compatibility
 	if client == nil {
 		return managedKey, nil
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -296,7 +296,7 @@ func MachineConfigFromRawIgnConfig(role, name string, rawIgnCfg []byte) (*mcfgv1
 
 // GetManagedKey returns the managed key for sub-controllers, handling any migration needed
 func GetManagedKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, prefix, suffix, deprecatedKey string) (string, error) {
-	managedKey := fmt.Sprintf("%s-%s-%s", prefix, pool.Name, suffix)
+	managedKey := fmt.Sprintf("%s-%s-generated-%s", prefix, pool.Name, suffix)
 	// if we don't have a client, we're installing brand new, and we don't need to adjust for backward compatibility
 	if client == nil {
 		return managedKey, nil
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -33,7 +33,7 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 	if len(configs) == 0 {
 		return nil, nil
 	}
-	sort.Slice(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
+	sort.SliceStable(configs, func(i, j int) bool { return configs[i].Name < configs[j].Name })
 
 	var fips, ok bool
 	var kernelType string
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"reflect"
@@ -19,9 +20,12 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+	kerr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 )
 
 // MergeMachineConfigs combines multiple machineconfig objects into one object.
@@ -260,4 +264,63 @@ func TranspileCoreOSConfigToIgn(files, units []string) (*ign2types.Config, error
 	}
 
 	return &convertedIgnCfgV2, nil
+}
+
+// MachineConfigFromIgnConfig creates a MachineConfig with the provided Ignition config
+func MachineConfigFromIgnConfig(role, name string, ignCfg interface{}) (*mcfgv1.MachineConfig, error) {
+	rawIgnCfg, err := json.Marshal(ignCfg)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling Ignition config: %v", err)
+	}
+	return MachineConfigFromRawIgnConfig(role, name, rawIgnCfg)
+}
+
+// MachineConfigFromRawIgnConfig creates a MachineConfig with the provided raw Ignition config
+func MachineConfigFromRawIgnConfig(role, name string, rawIgnCfg []byte) (*mcfgv1.MachineConfig, error) {
+	labels := map[string]string{
+		mcfgv1.MachineConfigRoleLabelKey: role,
+	}
+	return &mcfgv1.MachineConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: labels,
+			Name:   name,
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			OSImageURL: "",
+			Config: runtime.RawExtension{
+				Raw: rawIgnCfg,
+			},
+		},
+	}, nil
+}
+
+// GetManagedKey returns the managed key for sub-controllers, handling any migration needed
+func GetManagedKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, suffix, deprecatedKey string) (string, error) {
+	managedKey := fmt.Sprintf("99-%s-%s", pool.Name, suffix)
+	// if we don't have a client, we're installing brand new, and we don't need to adjust for backward compatibility
+	if client == nil {
+		return managedKey, nil
+	}
+	if _, err := client.MachineconfigurationV1().MachineConfigs().Get(context.TODO(), managedKey, metav1.GetOptions{}); err == nil {
+		return managedKey, nil
+	}
+	old, err := client.MachineconfigurationV1().MachineConfigs().Get(context.TODO(), deprecatedKey, metav1.GetOptions{})
+	if err != nil && !kerr.IsNotFound(err) {
+		return "", fmt.Errorf("could not get MachineConfig %q: %v", deprecatedKey, err)
+	}
+	// this means no previous CR config were here, so we can start fresh
+	if kerr.IsNotFound(err) {
+		return managedKey, nil
+	}
+	// if we're here, we'll grab the old CR config, dupe it and patch its name
+	mc, err := MachineConfigFromRawIgnConfig(pool.Name, managedKey, old.Spec.Config.Raw)
+	if err != nil {
+		return "", err
+	}
+	_, err = client.MachineconfigurationV1().MachineConfigs().Create(context.TODO(), mc, metav1.CreateOptions{})
+	if err != nil {
+		return "", err
+	}
+	err = client.MachineconfigurationV1().MachineConfigs().Delete(context.TODO(), deprecatedKey, metav1.DeleteOptions{})
+	return managedKey, err
 }
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
@@ -407,7 +407,10 @@ func TestContainerRuntimeConfigCreate(t *testing.T) {
 			mcp2 := helpers.NewMachineConfigPool("worker", nil, helpers.WorkerSelector, "v0")
 			mcp2.ObjectMeta.Labels["custom-crio"] = "storage-config"
 			ctrcfg1 := newContainerRuntimeConfig("set-log-level", &mcfgv1.ContainerRuntimeConfiguration{LogLevel: "debug", LogSizeMax: resource.MustParse("9k"), OverlaySize: resource.MustParse("3G")}, metav1.AddLabelToSelector(&metav1.LabelSelector{}, "custom-crio", "my-config"))
-			mcs1 := helpers.NewMachineConfig(getManagedKeyCtrCfg(mcp), map[string]string{"node-role": "master"}, "dummy://", []igntypes.File{{}})
+			ctrCfgKey, _ := getManagedKeyCtrCfg(mcp, nil)
+			mcs1 := helpers.NewMachineConfig(getManagedKeyCtrCfgDeprecated(mcp), map[string]string{"node-role": "master"}, "dummy://", []igntypes.File{{}})
+			mcs2 := mcs1.DeepCopy()
+			mcs2.Name = ctrCfgKey
 
 			f.ccLister = append(f.ccLister, cc)
 			f.mcpLister = append(f.mcpLister, mcp)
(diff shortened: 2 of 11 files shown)


runcom:
Testing it out, requires the ContainerRuntimeConfig to be fixed as well.

runcom:
@umohnani8 ptal as well, not sure why we were relying on UIDs but I can't see why so just dropped that as it's just something not fully reliable between bootstrap and MCO

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -172,10 +176,54 @@ func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool) string {
 	return fmt.Sprintf("99-%s-%s-containerruntime", pool.Name, pool.ObjectMeta.UID)
 }
 
+func getStableManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface) (string, error) {
+	// if we don't have a client, we're installing brand new, and we don't need to adjust for backward compatibility
```
cgwalters, line 180: "installing brand new" means "on the bootstrap node"?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -172,10 +176,54 @@ func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool) string {
...
+	oldRegCopy.Name = managedKey
+	_, err = client.MachineconfigurationV1().MachineConfigs().Update(context.TODO(), oldRegCopy, metav1.UpdateOptions{})
+	return managedKey, err
+}
+
 func getManagedKeyReg(pool *mcfgv1.MachineConfigPool) string {
 	return fmt.Sprintf("99-%s-%s-registries", pool.Name, pool.ObjectMeta.UID)
 }
 
+func getStableManagedKeyReg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface) (string, error) {
```
cgwalters, line 205: Optional: Looks like this could be de-duped `getStableManagedKeyCtrCfg` with an internal helper that takes the string suffix.

cgwalters:
Really nice debugging and description!  Fun bug.

pkg/controller/container-runtime-config/helpers.go
runcom: yeah, fresh installation

pkg/controller/container-runtime-config/helpers.go
runcom: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -411,6 +411,7 @@ func TestContainerRuntimeConfigCreate(t *testing.T) {
 			f.mccrLister = append(f.mccrLister, ctrcfg1)
 			f.objects = append(f.objects, ctrcfg1)
 
+			f.expectGetMachineConfigAction(mcs1)
```
ashcrow, line 414: nit: I didn't really dig in to see why this test requires this to be run twice, but it would be nice to denote why in a comment.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -484,9 +490,10 @@ func TestContainerRuntimeConfigUpdate(t *testing.T) {
 				t.Errorf("syncHandler returned: %v", err)
 			}
 
-			f.expectGetMachineConfigAction(mcs)
+			f.expectGetMachineConfigAction(mcsUpdate)
```
ashcrow, line 493: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -555,9 +568,11 @@ func TestImageConfigUpdate(t *testing.T) {
 			f.cvLister = append(f.cvLister, cvcfg1)
 			f.imgObjects = append(f.imgObjects, imgcfg1)
 
+			f.expectGetMachineConfigAction(mcs1)
```
ashcrow, line 571: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -601,16 +621,18 @@ func TestImageConfigUpdate(t *testing.T) {
 				t.Errorf("syncImgHandler returned: %v", err)
 			}
 
-			f.expectGetMachineConfigAction(mcs1)
-			f.expectUpdateMachineConfigAction(mcs1)
-			f.expectGetMachineConfigAction(mcs2)
-			f.expectUpdateMachineConfigAction(mcs2)
+			f.expectGetMachineConfigAction(mcs1Update)
```
ashcrow, line 624: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -601,16 +621,18 @@ func TestImageConfigUpdate(t *testing.T) {
...
+			f.expectGetMachineConfigAction(mcs1Update)
+			f.expectUpdateMachineConfigAction(mcs1Update)
+			f.expectGetMachineConfigAction(mcs2Update)
```
ashcrow, line 627: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -644,9 +668,11 @@ func TestICSPUpdate(t *testing.T) {
 			f.imgObjects = append(f.imgObjects, imgcfg1)
 			f.operatorObjects = append(f.operatorObjects, icsp)
 
+			f.expectGetMachineConfigAction(mcs1)
```
ashcrow, line 671: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -644,9 +668,11 @@ func TestICSPUpdate(t *testing.T) {
...
 			f.expectGetMachineConfigAction(mcs1)
 			f.expectCreateMachineConfigAction(mcs1)
 			f.expectGetMachineConfigAction(mcs2)
+			f.expectGetMachineConfigAction(mcs2)
```
ashcrow, line 675: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -694,16 +725,18 @@ func TestICSPUpdate(t *testing.T) {
 				t.Errorf("syncImgHandler returned: %v", err)
 			}
 
-			f.expectGetMachineConfigAction(mcs1)
-			f.expectUpdateMachineConfigAction(mcs1)
-			f.expectGetMachineConfigAction(mcs2)
-			f.expectUpdateMachineConfigAction(mcs2)
+			f.expectGetMachineConfigAction(mcs1Update)
```
ashcrow, line 728: nit: Same nit about running twice.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -694,16 +725,18 @@ func TestICSPUpdate(t *testing.T) {
...
+			f.expectGetMachineConfigAction(mcs1Update)
+			f.expectUpdateMachineConfigAction(mcs1Update)
+			f.expectGetMachineConfigAction(mcs2Update)
```
ashcrow, line 731: nit: Same nit about running twice.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -168,14 +172,64 @@ func findPolicyJSON(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
 	return nil, fmt.Errorf("could not find Policy JSON")
 }
 
+func getManagedKey(pool *mcfgv1.MachineConfigPool, suffix string) string {
+	return fmt.Sprintf("99-%s-%s", pool.Name, suffix)
+}
+
+// Deprecated: use getStableManagedKeyCtrCfg
```
ashcrow, line 179: Should we add a warning log on this so it's caught when called? If so, I the actual functionality in this function is small enough to be used in `getStableManagedKeyCtrCfg` directly.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -168,14 +172,64 @@ func findPolicyJSON(mc *mcfgv1.MachineConfig) (*igntypes.File, error) {
...
+		return managedKey, nil
+	}
+	// if we're here, we'll grab the old CR config, dupe it and patch its name
+	oldRegCopy := oldReg.DeepCopy()
+	oldRegCopy.Name = managedKey
+	_, err = client.MachineconfigurationV1().MachineConfigs().Update(context.TODO(), oldRegCopy, metav1.UpdateOptions{})
+	return managedKey, err
+}
+
+// Deprecated: use getStableManagedKeyReg
```
ashcrow, line 206: Same question as `getManagedKeyCtrCfg`

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
runcom: it's the standard unit pattern in the unit tests that mimick a controller - it's like a framework and we call `expect*` function everytime we want to check that a controller is doing what is supposed to do. I think yes, we could refactor this someday and better document it but adding a comment here doesn't buy us anything now :(

pkg/controller/container-runtime-config/helpers.go
runcom: if you're using Code it'll tell you that it's deprecated when referring to it :) as for adding logs in the running cluster, I think we as code maintainers we should know the code well enough (if we're doing reviews) to catch that something it's old and deprecated. So I'm gonna do something and just rename `getManagedKeyCtrCfg` to `getManagedKeyCtrCfgDeprecated` as it's still used in unit to make sure the upgrade that we're doing in this code still works.

umohnani8:
LGTM

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
ashcrow: I see. Viewing it _is_ confusing from the outside, but I accept your response :slightly_smiling_face:.

kikisdeliveryservice:
```
 May 27 15:40:28.056 W ns/openshift-cluster-machine-approver pod/machine-approver-5f569497cc-hxsg9 node/ci-op-jjt2zrvd-28de9-lcg6g-master-1 reason/FailedMount MountVolume.SetUp failed for volume "config" : failed to sync configmap cache: timed out waiting for the condition 
```
/test e2e-gcp-upgrade

runcom:
> ```
> machine-approver pod/machine-approver-5f569497cc-hxsg9 node/ci-op-jjt2zrvd-28de9-lcg6g-master-1 reason/FailedMount MountVolume.SetUp failed for volume "config" : failed to sync configmap cache: timed out waiting for the condition 
> ```
> 
> 
> /test e2e-gcp-upgrade

there's an actual bug in the PR from the MCC logs:

```
I0527 16:03:49.246607       1 container_runtime_config_controller.go:370] Error syncing image config openshift-config: Operation cannot be fulfilled on machineconfigs.machineconfiguration.openshift.io "99-master-registries": StorageError: invalid object, Code: 4, Key: /kubernetes.io/machineconfiguration.openshift.io/machineconfigs/99-master-registries, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 0e9bccba-828b-441c-9cd9-6527e836fe64, UID in object meta: 
```

runcom:
holding again pending further optimization fixes, creating the new MC and deleting the old one has the awful side effect of kicking the render_controller to re-generate a rendered config. While that's ok, it's wasted time (a lot) if a node picks that up and starts the upgrade. I'm patching the code to fix that by pausing the render controller somehow.

runcom:
It should be good now even if it could use some more optimization to avoid re-rendering while we create+delete. This is now scheduled for 4.6, so we can merge this as soon as master opens and have it soak for a while in CI to fully assess how it is doing.

/hold cancel
/bugzilla refresh

runcom:
holding cause kubelet config has the same issue (thanks @yuqi-zhang !)

/hold

runcom:
> holding cause kubelet config has the same issue (thanks @yuqi-zhang !)
> 
> /hold

now ready to go, let's get this asap so it can have time to soak

/hold cancel

runcom:
/skip

can I get some eyes on this? the sooner we get this in and let it soak into master, the better

sinnykumari:
/approve
will let another set of eye to review the change

runcom:
> Just a final confirmation, with this patch the naming will be e.g. `99-worker-kubeletconfig`, are we sure a naming like this will not conflict with any existing MCs on the system? The old UID naming scheme was a bit error-prone but one of the upsides was that it was unlikely to conflict with existing MCs.

The thing is we can't rely on something that changes between installer and MCO - it's probably generic, I think it won't conflict with anything user provided but I'm happy to explore alternatives

pkg/controller/common/helpers.go
```diff
@@ -261,3 +265,62 @@ func TranspileCoreOSConfigToIgn(files, units []string) (*ign2types.Config, error
...
+			Config: runtime.RawExtension{
+				Raw: rawIgnCfg,
+			},
+		},
+	}, nil
+}
+
+// GetManagedKey returns the managed key for sub-controllers, handling any migration needed
+func GetManagedKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, prefix, suffix, deprecatedKey string) (string, error) {
+	managedKey := fmt.Sprintf("%s-%s-generated-%s-", prefix, pool.Name, suffix)
```
yuqi-zhang, line 299: there's an extra dash at the end ;)

runcom: damn :joy:

runcom: fixed :)

runcom:
See also https://bugzilla.redhat.com/show_bug.cgi?id=1851965#c8

PR #1757, opened 2020-05-27
WIP Allow tuning cgroupv2 and mitigation arguments

vrutkovs:
This is required for Fedora CoreOS installs, which set `mitigations=auto`
and may enable unified cgroups

**- What I did**
Updated tuneable whitelist to include params necessary for OKD-on-FCOS:
* `mitigations=auto,nosmt`
* `systemd.unified_cgroup_hierarchy=0`

**- How to verify it**
Modify `systemd.unified_cgroup_hierarchy` or `mitigations` kernel args

Whitelist `systemd.unified_cgroup_hierarchy` and `mitigations=auto,nosmt` kernel args

cc @LorbusChris 

TODO:
* Ensure these arguments are allowed in !rhcos only

Supersedes #1497

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -37,7 +37,9 @@ const (
 // TODO: fill out the allowlists
 // tuneableRHCOSArgsAllowlist contains allowed keys for tunable kernel arguments on RHCOS
 var tuneableRHCOSArgsAllowlist = map[string]bool{
-	"nosmt": true,
+	"nosmt":                              true,
+	"systemd.unified_cgroup_hierarchy=0": true,
+	"mitigations=auto,nosmt":             true,
 }
 
 // tuneableFCOSArgsAllowlist contains allowed keys for tunable kernel arguments on FCOS


ashcrow:
Change seems sane.`nosmt` is in the MCO docs. Should the new whitelisted items be added as well noting when they should be (or not be) used?

LorbusChris:
Is addition of `systemd.unified_cgroup_hierarchy` required at this time? FCOS still defaults to cgroups v1.
Other than that looks good (I'll close #1497)
/approve

vrutkovs:
We're going to need to control cgroupsv2, as later we'll want to enable it - and disable if required

vrutkovs:
Tested this on FCOS - `mitigations=auto,nosmt` is required, it won't work for `mitigations=auto` for some reason

LorbusChris:
Is this good to go?
/cc @runcom @cgwalters

ashcrow:
Needs a rebase. @runcom PTAL.

cmd/machine-config-daemon/pivot.go
```diff
@@ -36,7 +36,9 @@ const (
 // TODO: fill out the allowlist
 // tuneableArgsAllowlist contains allowed keys for tunable arguments
 var tuneableArgsAllowlist = map[string]bool{
-	"nosmt": true,
+	"nosmt":                              true,
+	"systemd.unified_cgroup_hierarchy=0": true,
+	"mitigations=auto,nosmt":             true,
```
runcom, line 41: Pivot does understand which platform is being run on - maybe to be picky and cleaner we can allow these two only when !rhcos for now and expand it later when okd grows more platform?

LorbusChris: sgtm!

vrutkovs: We could reset it to be FCOS specific, but I don't see a reason for that - RHCOS users should be able to modify that parameter if it happens to be set/unset

runcom: I think this is now https://github.com/openshift/machine-config-operator/pull/1925/files which merged right?

vrutkovs: oh, right, so its not urgent for OKD then.

Do we want to allow this for RHCOS? I don't think there's need right now so we can keep it close and reopen if necessary

LorbusChris: I think that means we can close this PR for now? I'll do that now, but feel free to reopen if this merits further discussion!

PR #1758, opened 2020-05-27
Bug 1840222: [baremetal] Mount /var/run/NetworkManager in init containers

cybertron:
For some reason our early containers are sometimes getting empty
/etc/resolv.conf files. This is problematic because it means the
coredns init container will generate an invalid configuration with
no upstream servers. The init container resolv.conf never seems to
update either unless we mount in /var/run/NetworkManager (which also
contains resolv.conf).

This mounts /var/run/NetworkManager in both the coredns and
mdns-publisher init containers to allow them to generate valid
configurations even when they get an initially empty resolv.conf.

Note that this is more of a workaround than a fix. The empty resolv.conf
issue is likely to cause other issues, but hopefully this will at least unblock
ci for baremetal.

### DIFF
--- a/templates/common/baremetal/files/baremetal-coredns.yaml
+++ b/templates/common/baremetal/files/baremetal-coredns.yaml
@@ -48,6 +48,8 @@ contents:
           mountPath: "/config"
         - name: conf-dir
           mountPath: "/etc/coredns"
+        - name: nm-resolv
+          mountPath: "/var/run/NetworkManager"
         imagePullPolicy: IfNotPresent
       containers:
       - name: coredns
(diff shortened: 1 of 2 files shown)


cybertron:
/hold

After some more investigation I'm not sure this actually helped. It may be that adding the bind mount just triggered a container restart which pulled in the new resolv.conf. We're still in discussion with other teams about what could be causing this empty resolv.conf.

cybertron:
Yeah, this didn't help. The ci job failed with one node that didn't come up. must-gather is empty so no logs, but it's a pretty good bet that it was the resolv.conf bug.

PR #1759, opened 2020-05-27
[baremetal & friends] Give coredns and mdns-publisher init containers unique names

cybertron:
Currently both the coredns and mdns-publisher init containers are
named "render-config". This makes it impossible to tell at a glance
which belongs to which service. This change appends the name of the
service to the init container name so it is clear which service the
container is configuring.

Give coredns and mdns-publisher init containers unique names so it
is clear which service they belong to.

### DIFF
--- a/templates/common/baremetal/files/baremetal-coredns.yaml
+++ b/templates/common/baremetal/files/baremetal-coredns.yaml
@@ -27,7 +27,7 @@ contents:
         hostPath:
           path: "/var/run/NetworkManager"
       initContainers:
-      - name: render-config
+      - name: render-config-coredns
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:
         - runtimecfg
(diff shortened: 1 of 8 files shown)


cybertron:

Since this touches files from other platforms. Should be safe since it's just container names though.

rgolangh:
lgtm ovirt

mandre:
/test e2e-openstack

LGTM from an OpenStack point of vue :+1: 
Let's just make sure the CI agrees.

kikisdeliveryservice:
something seems wrong with the ovn test do we actually expect it to pass? it took 4 hours to fail? :woman_shrugging: 

trying one last time...

/retest

cybertron:
Do you know anything about the aws-proxy job? It's failed every time on this PR, and nothing in AWS should be affected.

kikisdeliveryservice:
It's a new job that was added and accidentally run on all prayers. It doesn't seem to have passing runs and isn't required so..

/skip

PR #1760, opened 2020-05-28
Bug 1780318: MCD: tolerate all taints

runcom:
This is supposed to fix the BZ in $subject and it's the same as BZs like https://bugzilla.redhat.com/show_bug.cgi?id=1813479. The impact of a change like this isn't fully clear as mentioned in the DNS PR as well https://github.com/openshift/cluster-dns-operator/pull/171#issuecomment-622110777

I'm opening this PR to get some more eyes and testing on this - @derekwaynecarr @smarterclayton ptal

The BZ has been pushed to 4.6 following what the DNS team did as well.

### DIFF
--- a/manifests/machineconfigdaemon/daemonset.yaml
+++ b/manifests/machineconfigdaemon/daemonset.yaml
@@ -63,13 +63,6 @@ spec:
       hostPID: true
       serviceAccountName: machine-config-daemon
       terminationGracePeriodSeconds: 600
-      tolerations:
-      - key: node-role.kubernetes.io/master
-        operator: Exists
-        effect: NoSchedule
-      - key: node-role.kubernetes.io/etcd
-        operator: Exists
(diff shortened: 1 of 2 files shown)


runcom:
@stbenjam ping for the metal-ipi e2e sort of constantly failing, can you help assess if this PR is breaking it or it's transient?

yuqi-zhang:
In theory I think this is correct. Should we skip 4.5 and put this into early 4.6 so to let the change of behaviour soak? I don't think this will break any existing clusters anyways.

runcom:
> In theory I think this is correct. Should we skip 4.5 and put this into early 4.6 so to let the change of behaviour soak? I don't think this will break any existing clusters anyways.

yep, BZ is already 4.6 and now that master is open we can get this in to test it out

runcom:
> /bugzilla refresh

all good now then

PR #1706, opened 2020-04-30
Bug 1829642: templates: Add a special machine-config-daemon-firstboot-v42.service

cgwalters:
This is aiming to fix:
https://bugzilla.redhat.com/show_bug.cgi?id=1829642
AKA
https://github.com/openshift/machine-config-operator/pull/1215#issuecomment-622081019

Basically we have our systemd units dynamically differentiate between
"4.2" and "4.3 or above" by looking at the aleph version.

---

# https://bugzilla.redhat.com/show_bug.cgi?id=1829642

## Background

In the beginning, we had this concept of `pivot.service` whose whole job was to pull a new `machine-os-content` and reboot, running `Before=kubelet.service`.
More: [OSUpgrades](https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md).

That was simple, and worked well!  Life was good.  Everyone was relaxing on the beach.

Then, we needed to [handle kernel arguments](https://github.com/openshift/installer/pull/1392) (specifically adding `nosmt`).

But a core problem here is that we have [no mechanism to update bootimages](https://github.com/openshift/enhancements/pull/201) by default - so every change we make that needs to happen on this "firstboot" becomes an upgrade hazard.

Further, we had a strong split between the `pivot` code (day 0) and what happened in the MCO.  So we started baking in the MCD to the host; https://github.com/openshift/machine-config-operator/pull/868 is a notable commit here.

But still today, the MCO needs to handle having even very old bootimages (from 4.1) scale up and join the cluster.

https://github.com/openshift/machine-config-operator/pull/891 is a notable commit here; basically the MCS serves Ignition which contains a pair of systemd units:

[machine-config-daemon-firstboot.service](https://github.com/openshift/machine-config-operator/blob/5396c45464fc1658c65bad980de0a386958fbf9a/templates/common/_base/units/machine-config-daemon-firstboot.service):  Replaces `pivot.service` - this handles everything the MCO knows how to handle day 1/2, such as kernel arguments too.

[machine-config-daemon-host.service](https://github.com/openshift/machine-config-operator/blob/5396c45464fc1658c65bad980de0a386958fbf9a/templates/common/_base/units/machine-config-daemon-host.service): Run by rpm-ostree to perform upgrades.  This is not enabled by default, but is started by the MCD (both via the `-firstboot.service` as well as "day 2" when the MCD runs as a daemonset).

[...]

PR #1761, opened 2020-05-28
machine-config-daemon-host: Also Before=crio.service

cgwalters:
This came up in discussions around here:
https://github.com/openshift/machine-config-operator/pull/1706#issuecomment-622091201

There's no reason to start crio either, and it's confusing
to do so because in that bug it made us think there was a
crio problem when there wasn't.

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-host.service
+++ b/templates/common/_base/units/machine-config-daemon-host.service
@@ -10,6 +10,7 @@ contents: |
   # If pivot exists, defer to it.  Note similar code in update.go
   ConditionPathExists=!/usr/lib/systemd/system/pivot.service
   After=ignition-firstboot-complete.service
+  Before=crio.service crio-wipe.service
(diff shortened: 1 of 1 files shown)


runcom:
This won't merge right away this week but lgtm'ing it 

/approve
/lgtm

cgwalters:
Hmm I'm seeing `crio-wipe.service` start pretty early on...I am wondering if we actually also should be doing `Before=crio-wipe.service` in more places like this too.

haircommander:
> Hmm I'm seeing `crio-wipe.service` start pretty early on...I am wondering if we actually also should be doing `Before=crio-wipe.service` in more places like this too.

yeah I agree, I would say it makes most sense to have crio-wipe happen right before crio, if possible

cgwalters:
I rolled this into https://github.com/openshift/machine-config-operator/pull/1762

PR #1766, opened 2020-05-29
Use MCD binary from container in /run/bin

vrutkovs:
(PR updated by Colin Walters <walters@verbum.org>, 
Co-authored-by: Vadim Rutkovsky <vrutkovs@redhat.com> )

---

Use MCD binary from container in /run/bin

And drop all dependencies at firstboot and cluster time on the presence
of the MCD baked into the host.

This solves a whole pile of problems:

 - We need this for OKD (with FCOS)
 - Now pull requests with our CI naturally use
   the updated code, so we get (important) coverage of firstboot paths
 - Relatedly, we don't have to worry about "drift" between
   the on-host version and the in-container version
 - ART can stop building the MCD and RHCOS can drop it

Since we're now injecting our updated *code* into the bootimage,
we can drop the special 4.1 handling with `pivot.service` and
the special `4.2` unit.

Tweak the code run via `machine-config-daemon-firstboot` to
run rpm-ostree directly rather than via the `-host` unit, which
just adds confusion.

And finally
With this the MCD (when run as a pod) stops using `machine-config-daemon-host.service`.
and creates a dynamic unit instead.

With the combination of both, `machine-config-daemon-host.service` is on the
path to not being used by default and migrating to a "4.1 bootimage aid".

The systemd-run model of creating a unit dynamically is much clearer for what we want here;
conceptually the service is just a dynamic child of this pod (if we could we'd
tie the lifecycle together).  Further:

- Let's shorten our systemd unit names by using the `mco-` prefix
- Inject the `RPMOSTREE_CLIENT_ID`, see https://github.com/coreos/rpm-ostree/commit/016c1c5e627fc2a8cd3266ccda3a47a5f8992594

Co-authored-by: Vadim Rutkovsky <vrutkovs@redhat.com>


cgwalters:
When I was looking at this in the past, one problem is that all of the OCP containers are built against UBI7 (i.e. RHEL7).  And while the MCD is Go and hence all statically linked...there's `openssl` which does come into play with FIPS and differs between the two (we carry `openssl-deprecated` in RHCOS solely because people are doing this type of "extract binary from a container" thing).

See  https://github.com/coreos/fedora-coreos-tracker/issues/354

So based on that one approach would be to change our container image to build binaries for both RHEL7/RHEL8 and then extract the rhel8 one here.

templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -0,0 +1,28 @@
+name: "machine-config-daemon-pull.service"
+enabled: true
+contents: |
+  [Unit]
+  Description=Machine Config Daemon Pull
+  # Make sure it runs only on OSTree booted system
+  ConditionPathExists=/run/ostree-booted
+  BindsTo=ignition-firstboot-complete.service
```
cgwalters, line 8: This was just removed in https://github.com/openshift/machine-config-operator/pull/1762

templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -0,0 +1,28 @@
...
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  # Extract machine-config-daemon binary from MCD container
+  ExecStart=/usr/bin/sh -c "while ! /usr/bin/podman pull --authfile=/var/lib/kubelet/config.json '{{ .Images.setupEtcdEnvKey }}'; do sleep 5; done"
+  ExecStart=/usr/bin/sh -c "/usr/bin/podman run --authfile=/var/lib/kubelet/config.json --quiet --net=host --entrypoint=cat '{{ .Images.setupEtcdEnvKey }}' /usr/bin/machine-config-daemon > /usr/local/bin/machine-config-daemon"
+  ExecStart=/usr/bin/chmod +x /usr/local/bin/machine-config-daemon
+  ExecStart=/usr/sbin/restorecon /usr/local/bin/machine-config-daemon
+
+  [Install]
+  WantedBy=multi-user.target
```
cgwalters, line 28: It might be simplest to do:
```
RequiredBy=machine-config-daemon-firstboot.service
```

Then we don't need to copy all of the conditionals here - we'll only run if that unit is going to run, and also crucially it means if this unit fails then `firstboot` won't run.

cgwalters: See https://github.com/openshift/installer/pull/3683 for a similar "dependency enforcement" change.

templates/common/_base/units/machine-config-daemon-host.service
```diff
@@ -16,7 +16,7 @@ contents: |
   # Need oneshot to delay kubelet
   Type=oneshot
   # TODO add --from-etc-pullspec after ratcheting
-  ExecStart=/usr/libexec/machine-config-daemon pivot
+  ExecStart=/usr/local/bin/machine-config-daemon pivot
```
sinnykumari, line 19: I think we would need to create symlink to usr/local/bin/machine-config-daemon here as well to cover cluster upgrade scenario.

templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -0,0 +1,44 @@
...
+  # Make sure it runs only on OSTree booted system
+  ConditionPathExists=/run/ostree-booted
+  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
+  # We only want to run on 4.3 clusters and above; this came from
+  # https://github.com/coreos/coreos-assembler/pull/768
+  ConditionPathExists=/sysroot/.coreos-aleph-version.json
+  # Skip this unit if MCD binary already exists at libexec
+  ConditionPathExists=!/usr/libexec/bin/machine-config-daemon
+  After=ignition-firstboot-complete.service
+  Before=machine-config-daemon-host.service
```
sinnykumari, line 15: This is not required because we don't run machine-config-daemon-host.service on firstboot.

cmd/machine-config-daemon/pivot.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/sh
-exec /usr/libexec/machine-config-daemon pivot "$@"
+exec /usr/local/machine-config-daemon pivot "$@"
```
sinnykumari, line 2: `bin` is missing in the m-c-d path

sinnykumari:
We would also need to create symlink and fix m-c-d path in https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/machine-config-daemon-firstboot-v42.service#L18

cmd/machine-config-daemon/pivot.sh
vrutkovs: Updated. Is it being used at all in 4.5?

LorbusChris:
@vrutkovs thanks for including the proxy settings here!

Before this can go in, we need to create UBI8 based MCO images and pull MCD from there (as Colin mentioned).

cmd/machine-config-daemon/pivot.sh
sinnykumari: Don't think so but I believe this is mainly there to maintain compatibility with cluster when [pivot ](https://github.com/openshift/pivot) was maintained separately. @cgwalters  would have more insights.

templates/common/_base/units/machine-config-daemon-firstboot.service
```diff
@@ -10,14 +10,16 @@ contents: |
...
+  After=machine-config-daemon-pull.service
   Before=crio.service crio-wipe.service
   Before=kubelet.service
 
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/libexec/machine-config-daemon firstboot-complete-machineconfig
+  # Create /usr/local/bin symbolic link unless it exists
+  ExecStart=-/usr/bin/ln --symbolic /usr/libexec/bin/machine-config-daemon /usr/local/bin/machine-config-daemon
```
sinnykumari, line 21: Just curious, why did we add `-` prefix to /usr/bin/ln/ ?

vrutkovs: This would make systemd ignore the error if the file already exists, https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=

LorbusChris:
@vrutkovs instead of having another container in the payload, we could also build the MCD in a UBI8 env and stick it into the UBI7 container

vrutkovs:
Right, `golang-1.13` is centos7 based. Do we have a centos8-based builder image?

templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -0,0 +1,43 @@
...
+  ConditionPathExists=!/usr/libexec/bin/machine-config-daemon
+  After=ignition-firstboot-complete.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  # Extract machine-config-daemon binary from MCD container
+  ExecStart=/usr/bin/sh -c "while ! /usr/bin/podman pull --authfile=/var/lib/kubelet/config.json '{{ .Images.setupEtcdEnvKey }}'; do sleep 5; done"
+  ExecStart=/usr/bin/sh -c "/usr/bin/podman run --authfile=/var/lib/kubelet/config.json --quiet --net=host --entrypoint=cat '{{ .Images.setupEtcdEnvKey }}' /usr/bin/machine-config-daemon > /usr/local/bin/machine-config-daemon"
+  ExecStart=/usr/bin/chmod +x /usr/local/bin/machine-config-daemon
```
cgwalters, line 22: I was going to say to write this in `/run` instead per https://github.com/coreos/fedora-coreos-tracker/issues/354# but...then I remembered we still need this for `machine-config-daemon-host.service`.  Ugg.

Ideally at some point we get rid of `machine-config-daemon-host.service` - I think that only exists because of `spc_t` issues but https://bugzilla.redhat.com/show_bug.cgi?id=1839065 should let us fix that.

Alternatively, we could also dynamically inject the binary into the host from the *pod* case too (i.e. not firstboot).  That would make everything symmetric.

The concern here is "what updates that binary" - the semantic here is we only write it on firstboot, which is going to mean it becomes stale.

So if we dynamically inject it in both cases, ownership and lifecycle is clear.

cmd/machine-config-daemon/pivot.sh
cgwalters: I believe this was simply for humans to use interactively via `oc debug node/` or `ssh`.  It's OK if we remove it I guess, but it'd also be nice to keep it working.

templates/common/_base/units/machine-config-daemon-pull.service.yaml
sinnykumari: Indeed it will be more consistent if we pull m-c-d binary from image during subsequent upgrades as well. This also mean that we will no longer need to maintain m-c-d rpm package.

LorbusChris:
just a thought: Do we actually need to pull the MCD binary out of the container, or could we run it as a privileged DaemonSet from the container?

cgwalters:
> just a thought: Do we actually need to pull the MCD binary out of the container, or could we run it as a privileged DaemonSet from the container?

The reason we do this is https://github.com/openshift/machine-config-operator/blob/05539381b9b10d7f5147aa838df64b0be21b61d3/pkg/daemon/rpm-ostree.go#L303

cgwalters:
@vrutkovs mind if I take this PR?  I have a pile of changes to debug
https://bugzilla.redhat.com/show_bug.cgi?id=1842906
and I'm realizing that the lack of CI coverage for the on-host MCD is increasingly problematic for our sanity.

This would fix that and other things too.

vrutkovs:
>mind if I take this PR?

That would be great! Feel free to make a new one or push to this branch

pkg/daemon/update.go
```diff
@@ -1280,9 +1280,25 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	}
 
 	glog.Infof("Updating OS to %s", newURL)
-	if err := dn.NodeUpdaterClient.RunPivot(newURL); err != nil {
-		MCDPivotErr.WithLabelValues(newURL, err.Error()).SetToCurrentTime()
-		return fmt.Errorf("failed to run pivot: %v", err)
+	// In the cluster case, for now we run indirectly via machine-config-daemon-host.service
```
sinnykumari, line 1283: s/machine-config-daemon-host.service/mco-pivot unit/?

pkg/daemon/update.go
```diff
@@ -1280,9 +1280,25 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
...
+	// for SELinux reasons, see https://bugzilla.redhat.com/show_bug.cgi?id=1839065
+	if dn.kubeClient != nil {
```
sinnykumari, line 1285: does it make sense to instead check for /etc/ignition-machine-config-encapsulated.json file existance?

pkg/daemon/rpm-ostree.go
```diff
@@ -305,40 +303,29 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
...
+	// We used to start machine-config-daemon-host here, but now we make a dynamic
+	// unit because that service was started in too many ways, and the systemd-run
+	// model of creating a unit dynamically is much clearer for what we want here;
+	// conceptually the service is just a dynamic child of this pod (if we could we'd
+	// tie the lifecycle together).  Further, let's shorten our systemd unit names
+	// by using the mco- prefix, and we also inject the RPMOSTREE_CLIENT_ID now.
+	unitName := "mco-pivot"
+	glog.Infof("Executing OS update (pivot) on host via systemd-run unit=%s", unitName)
+	err := exec.Command("systemd-run", "--wait", "--collect", "--unit="+unitName,
+		"-E", "RPMOSTREE_CLIENT_ID=mco", constants.HostSelfBinary, "pivot", osImageURL).Run()
```
sinnykumari, line 326: Curious, what is the use of passing osImageURL, does it get processed?

sinnykumari:
Overall this looks great, thanks Vadim and Colin! 

Before getting this merged, we definitely need to test and ensure that we are not breaking anything like day1/day2 MachineConfig operation, upgrades, machine scale-up from older upgraded cluster.

Few additional comments:
1. I think we can also remove https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/pivot.service.yaml as part of this PR?
2. We still need to address https://github.com/openshift/machine-config-operator/pull/1766#issuecomment-636101355 to get it working on RHCOS, right?

pkg/daemon/rpm-ostree.go
cgwalters: It does, the `pivot` CLI accepts an argument too.  And in this combined code we stopped writing `/etc/pivot/image-pullspec` (and unlink it even if it exists).

pkg/daemon/update.go
cgwalters: Should be an equivalent condition - `kubeClient` is only set when we're running as the MCD and are connected to the cluster.  

Though I think `kubeClient` is *also* unset in the "once from" mode used for Ansible joining rhel7 workers, but for rhel7 workers we won't try running rpm-ostree either.

But...we are definitely inconsistent about these checks; some look for `dn.kubeClient` some look for `dn.recorder` etc.  It might be better to have a single `dn.runningAsPod` or something?

cgwalters:
Cool, here's a successful e2e-gcp-op run: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1766/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/2526

Looking at the journal, you can see `-pull` and `-firstboot` going.

And in the pod logs,
`I0611 03:08:20.864049    2147 start.go:97] Copied self to /run/bin/machine-config-daemon on host`

But unfortunately we're not covering OS updates right now in this CI.  We'll need to add that.  I did test this locally though in a libvirt cluster by editing the `configmap/machine-config-osimageurl` manually and I saw `mco-pivot` running.

cgwalters:
> We still need to address #1766 (comment) to get it working on RHCOS, right?

I'm wavering on that to be honest.  We definitely want that long term...it's kind of a hard requirement for using golang w/FIPS.  But it'lll be annoying to do and this is such a big cleanup :smile:

cgwalters:
> I think we can also remove https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/pivot.service.yaml as part of this PR?

Done, thanks!

pkg/daemon/update.go
sinnykumari: Anything is fine by me as far as it works for all conditions. `ignition-machine-config-encapsulated.json` is just easier for me to understand/debug when something goes wrong :laughing:

pkg/daemon/rpm-ostree.go
sinnykumari: ah yes, I just looked at pivot options. Should have checked into the run().
I maybe missing something but since we are no longer looking at image-pullspec, perhaps we can do more clean-up in pivot.go related to image-pullspec?

sinnykumari:
> > We still need to address #1766 (comment) to get it working on RHCOS, right?
> 
> I'm wavering on that to be honest. We definitely want that long term...it's kind of a hard requirement for using golang w/FIPS. But it'lll be annoying to do and this is such a big cleanup smile

Agree, my intention was not to fix everything in this PR :) My concern is that if we would like to get this PR landed into 4.6, won't using UBI7 built container is going to block us?

templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -0,0 +1,36 @@
...
+  Description=Machine Config Daemon Pull
+  # Make sure it runs only on OSTree booted system
+  ConditionPathExists=/run/ostree-booted
+  # This "stamp file" is unlinked when we complete
+  # machine-config-daemon-firstboot.service
+  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
+  Wants=network-online.target
+  After=network-online.target
+  After=rhcos-growpart.service
+
```
sinnykumari, line 14: wondering if we need `After=ignition-firstboot-complete.service` too

cgwalters: Good question!  We've discussed in CoreOS moving that service to the initramfs (because that's where Ignition runs), so the ordering would become redundant at some point anyways.

I can't think of any reason we ordered relating to the firstboot complete before either.  Conceptually what we're doing is a "second stage", but it doesn't require the boot partition to be updated.

cgwalters:
> My concern is that if we would like to get this PR landed into 4.6, won't using UBI7 built container is going to block us?

If we ever decide for some reason that the MCD needs to be built in FIPS mode it would absolutely break us.  OK I updated [this issue](https://github.com/coreos/fedora-coreos-tracker/issues/354#issue-555696546) to link directly to the multus Dockerfile that we can copy.

kikisdeliveryservice:
@cgwalters As mentioned in https://github.com/openshift/machine-config-operator/pull/1826 Can we please move this new e2e-test to it's own e2e and callable as you proposed? Our current MCO e2e is already very long and the gcp-op run on your PR was 2:44 which is unmanageable.  I think it might make sense to do that via your other (now closed) pr and then call it on this PR to verify it is running as expected.

cgwalters:
> Can we please move this new e2e-test to it's own e2e and callable as you proposed? Our current MCO e2e is already very long and the gcp-op run on your PR was 2:44 which is unmanageable. 

Right, two points:
First we're not running the test in an automated fashion right now, so no concerns about delaying gcp-op.
And it can't be part of gcp-op because it relies on AWS bootimages to conveniently test 4.1 (we only shipped GCP in 4.2).

> I think it might make sense to do that via your other (now closed) pr and then call it on this PR to verify it is running as expected.

OK I reopened it and will look at adding bits to openshift/release to run it.

kikisdeliveryservice:
> > Can we please move this new e2e-test to it's own e2e and callable as you proposed? Our current MCO e2e is already very long and the gcp-op run on your PR was 2:44 which is unmanageable.
> 
> Right, two points:
> First we're not running the test in an automated fashion right now, so no concerns about delaying gcp-op.
> And it can't be part of gcp-op because it relies on AWS bootimages to conveniently test 4.1 (we only shipped GCP in 4.2).
> 
> > I think it might make sense to do that via your other (now closed) pr and then call it on this PR to verify it is running as expected.
> 
> OK I reopened it and will look at adding bits to openshift/release to run it.

Thank you for explaining!!

cgwalters:
Reopened test as https://github.com/openshift/machine-config-operator/pull/1836

cgwalters:
e2e failure looks like https://bugzilla.redhat.com/show_bug.cgi?id=1836376
/retest

cgwalters:
OK so the bootimages test successfully failed - looks like the 4.2 bootimage joined the cluster but 4.1 didn't.  Will investigate.

At least now we can debug that working in a single pull request!

cgwalters:
OK some debugging later, the current flaw with 4.1 is:
```
[root@ip-10-0-130-191 ~]# systemctl status machine-config-daemon-pull
● machine-config-daemon-pull.service - Machine Config Daemon Pull
   Loaded: loaded (/etc/systemd/system/machine-config-daemon-pull.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2020-06-23 19:17:57 UTC; 5min ago
  Process: 1271 ExecStart=/bin/sh -c /usr/bin/podman run --rm --authfile=/var/lib/kubelet/config.json --quiet --net=host --entrypoint=cat 'registry.svc.ci.openshift.org/rhcos-devel/walters-mco@sha256:c38a3b21cf53b8376d4929d13eff7871368ec7dfa482887cd1d2bd02fb464489' /usr/bin/machine-config-daemon > /run/bin/machine-co>
  Process: 1247 ExecStart=/bin/sh -c /bin/mkdir -p /run/bin && chcon --reference=/usr/bin /run/bin (code=exited, status=0/SUCCESS)
 Main PID: 1271 (code=exited, status=125)
      CPU: 67ms

Jun 23 19:17:52 ip-10-0-130-191 systemd[1]: Starting Machine Config Daemon Pull...
Jun 23 19:17:57 ip-10-0-130-191 sh[1271]: flag provided but not defined: -authfile
Jun 23 19:17:57 ip-10-0-130-191 sh[1271]: See 'podman run --help'.
Jun 23 19:17:57 ip-10-0-130-191 systemd[1]: machine-config-daemon-pull.service: Main process exited, code=exited, status=125/n/a
Jun 23 19:17:57 ip-10-0-130-191 systemd[1]: machine-config-daemon-pull.service: Failed with result 'exit-code'.
Jun 23 19:17:57 ip-10-0-130-191 systemd[1]: Failed to start Machine Config Daemon Pull.
Jun 23 19:17:57 ip-10-0-130-191 systemd[1]: machine-config-daemon-pull.service: Consumed 67ms CPU time
```

I think I saw a similar thing go by for one of the other containers we're pulling.

cgwalters:
/approve
(I can do that right?  The rules are funny when co-authoring a PR with someone else :smile: )

cgwalters:
OK and I also hacked up my test cluster's 4.1 bootimage node manually with the change in this PR, did `systemctl restart machine-config-daemon-pull machine-config-daemon-firstboot`, node rebooted and joined the cluster; the MCD for that node went degraded because I'd edited the systemd unit; reverted that, MCD is happy and so is MCO.

So let's give this another try
/test cluster-bootimages

cgwalters:
> ci/prow/cluster-bootimages — Job succeeded. 

:tada: !

cgwalters:
OK I'm lifting WIP on this.

I think the remaining issue is to determine whether or not we want to block on building a RHEL8 MCD binary into the container image: https://github.com/openshift/machine-config-operator/pull/1766#issuecomment-644448816

I'm uncertain...it is something we can do later, and try this out in master?  If we need to back this out, then the work of adding another binary into the image wouldn't be useful either.

vrutkovs:
>I'm uncertain...it is something we can do later, and try this out in master? 

+1 to that. I haven't hit rhel7/rhel8 issues yet, so the fix for this can be postponed

cgwalters:
e2e-gcp-upgrade looks like
https://bugzilla.redhat.com/show_bug.cgi?id=1791162
/test e2e-gcp-upgrade

cgwalters:
/assign sinnykumari
for review/approval (hope that's OK MCO team!)

I know this is a big change but...we all going to be *way* happier debugging this in the future I think.  The "CI covers changes to firstboot" *alone* is worth it for all of our sanity.

cgwalters:
Followup to drop `machine-config-daemon` from RHCOS: https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1044

cgwalters:
As of currently in 4.6, `/usr/libexec/machine-config-daemon` was dropped from RHCOS.

PR #1767, opened 2020-05-29
[baremetal] Set NonVirtualIP in mdns plugin config

cybertron:
In order to avoid sending multicast traffic on all configured
interfaces on a system, we need to specify the address of the
interface we want to send the traffic on.

Because this is a high priority problem on baremetal and (at least
to my knowledge) not on other platforms, I'm doing it for baremetal
only for now. We can do the other platforms as a followup, but I
don't want to hold it up waiting for everyone to sign off.

Configure coredns-mdns to only send multicast traffic on one interface
to avoid sending it to interfaces where it is undesirable.

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -1,7 +1,7 @@
 . {
     errors
     health :18080
-    mdns {{ .ControllerConfig.Infra.Status.EtcdDiscoveryDomain }} {{`{{.Cluster.MasterAmount}}`}} {{`{{.Cluster.Name}}`}}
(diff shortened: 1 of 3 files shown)


cybertron:
/hold

This will need https://github.com/openshift/coredns-mdns/pull/60 to merge and be vendored into coredns, and then new coredns images built before it will work. I'm testing everything together locally in the meantime to verify the fix.

cybertron:
Vendoring proposed in https://github.com/openshift/coredns/pull/31 Once that merges and we have new images this should be good to go.

cybertron:
/hold cancel

Looks like we have the new coredns image to support this.

PR #1768, opened 2020-06-01
[baremetal] Update keepalived to support both multicast and unicast modes (default=multicast)

yboaron:
The current deployment workflow relies on the network allowing multicast traffic for VRRP management of the VIP addresses, with this PR keepalived supports both unicast and multicast modes.

To enable unicast keepalived, keepalived.conf in each node should include all the relevant nodes
(including the bootstrap node ) IP addresses, to support this requirement the bootstrap node will run
a simple server that will return its IP address, rest of nodes details will be retrieved from kube-apiserver.

The keepalived mode (unicast or multicast) controlled by 'ENABLE_UNICAST' environment variable, it's set by default to multicast.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/baremetal/keepalived.conf.tmpl
+++ b/manifests/baremetal/keepalived.conf.tmpl
@@ -4,17 +4,29 @@
 # For more information, see installer/data/data/bootstrap/baremetal/README.md
 # in the installer repo.
 
-vrrp_instance {{`{{.Cluster.Name}}`}}_API {
+{{`{{$nonVirtualIP := .NonVirtualIP}}`}}
+
+{{`vrrp_instance {{.Cluster.Name}}_API {
     state BACKUP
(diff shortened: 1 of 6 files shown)


yboaron:
should be merged after  https://github.com/openshift/baremetal-runtimecfg/pull/65

yboaron:
https://github.com/openshift/baremetal-runtimecfg/pull/65 was merged, hold can be removed.

/hold cancel

yboaron:
/retitle [baremetal] Update keepalived to support both multicast and unicast modes (default=multicast)

kikisdeliveryservice:
@yboaron is there a reason why you are committing under a different github account?

yboaronn:
> @yboaron is there a reason why you are committing under a different github account?
No, sorry about that. I'll fix that for the next commits.

PR #1769, opened 2020-06-01
Bug 1834194: vsphere: 4.1 to 4.4 upgrade bug

jcpowermac:
v4.1 did not include the `.Infra` see: https://github.com/openshift/machine-config-operator/blob/master-4.1/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L111-L136

Add a check to make sure that it is not nil.

### DIFF
--- a/templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
+++ b/templates/common/vsphere/files/vsphere-NetworkManager-kni-conf.yaml
@@ -3,12 +3,14 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-kni.conf"
 contents:
   inline: |
+    {{ if .Infra -}}
     {{ if .Infra.Status -}}
     {{ if .Infra.Status.PlatformStatus -}}
     {{ if .Infra.Status.PlatformStatus.VSphere -}}
     {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
     [main]
     dhcp=dhclient
+    {{ end -}}
     {{ end -}}
     {{ end -}}
     {{ end -}}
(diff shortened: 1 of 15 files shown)


jcpowermac:
Tested with release image:
https://quay.io/repository/jcallen/origin-release?tab=tags

```
$ oc adm upgrade                                               
Cluster version is 0.0.1-2020-06-01-160915
```
```
$ oc get clusteroperators.config.openshift.io machine-config                                                                   
NAME             VERSION                   AVAILABLE   PROGRESSING   DEGRADED   SINCE                        
machine-config   0.0.1-2020-06-01-160915   True        False         False      106m     
```

jcpowermac:
cc: @patrickdillon

sdodson:
/hold
This change is only relevant to 4.4 or does this need to go to master, release-4.5 first? I can help expedite if that's the case so we don't break process.

jcpowermac:
> /hold
> This change is only relevant to 4.4 or does this need to go to master, release-4.5 first? I can help expedite if that's the case so we don't break process.

It won't cleanly cherry-pick because of template changes between 4.4 and 4.5.  Do you want a corresponding PR for master/4.5 before we go foward on this?

sdodson:
> It won't cleanly cherry-pick because of template changes between 4.4 and 4.5. Do you want a corresponding PR for master/4.5 before we go foward on this?

Yes if an equivalent fix is required in master and release-4.5 then we should do that. I wonder why QE didn't find the problem the new problem in 4.5?

sdodson:
/hold cancel
Upstream PRs have been opened in #1772, however RHCOS in 4.6 is still 8.2 and thus failing all CI jobs preventing its merge. I'm fine with this merging as long as we agree to ensure the upstream changes are 4.5 blocking. I've fixed up the bugzilla side of things. Hopefully by the time patch manager gets to this upstream will have merged but I'd like to consider merging this ahead of that if not because this is blocking 4.3 to 4.4 upgrades.

jcpowermac:
> hmm `e2e-vsphere` is UPI not IPI. Thats not right.

nm this is for `release-4.4`.  No IPI

sdodson:
ci/prow/e2e-aws

PR #1770, opened 2020-06-01
Bug 1829453: check for localhost fqdn before kubelet starts up

rphillips:
**- What I did**
Adds a kubelet-hostname-check bash script to check to see if the hostname on the machine is a relative of `localhost`. This will prevent the kubelet from registering a 'localhost' machine, and wait for the proper fqdn to be applied to the machine.

**- How to verify it**

### DIFF
--- a/templates/common/_base/files/kubelet-hostname.yaml
+++ b/templates/common/_base/files/kubelet-hostname.yaml
@@ -1,0 +1,21 @@
+filesystem: "root"
+mode: 0744
+path: "/usr/local/bin/kubelet-hostname-check"
+contents:
+  inline: |
+    #!/bin/bash
+    # This is a workaround for BZ#1829453
+    name=$(hostname -f)
+    if [[ "$name" =~ ^localhost.* ]] ; then
+      cat << EOF
+    The machine's fully qualified domain name ($name) is set to localhost.
+    Openshift is not supported with a localhost FQDN, but this should
+    automatically resolve within a short period of time with a proper network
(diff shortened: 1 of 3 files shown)


cgwalters:
I'm not opposed to this...but, I think it'd be best if we had a canonical document about this, and the error message included a URL to it or so.

The failure here is going to be not so obvious for people to debug.

https://docs.openshift.com/container-platform/4.4/installing/installing_bare_metal/installing-bare-metal-network-customizations.html maybe, as well as link to the BZ?

rphillips:
@cgwalters great suggestions.... I added them in. ptal

rphillips:
We can also ask @mburke5678 to add this to the docs as well.

rphillips:
rebased to the latest master

rphillips:
the EOF was two spaces off

/facepalm

Fixed

cgwalters:
Do we still need this after  https://github.com/openshift/machine-config-operator/pull/1914 ?
(I mentioned this PR there but we didn't follow up on it AFAICS)

sjenning:
fyi @rphillips is on paternity level for a while

I'm not sure if we still need this or not.  As long as the kubelet can't start before this NetworkManager check passes, we should be ok.  Said another way, the kubelet should not start if `node-valid-hostname.service` did not succeed. `node-valid-hostname.service` is Before `network-online.target` but that doesn't mean that service needs to succeed in order to reach `network-online.target` right?

PR #1781, opened 2020-06-04
Bug 1843974: daemon: fix restoring file logic to avoid unnecessary deletion

sinnykumari:
This fixes restoring original file content into /etc/
when file is available in /usr/etc/ and not owned by any rpm.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -869,21 +869,32 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *igntypes.Config) e
 				// Add a check for backwards compatibility: basically if the file doesn't exist in /usr/etc (on FCOS/RHCOS)
 				// and no rpm is claiming it, we assume that the orig file came from a wrongful backup of a MachineConfig
 				// file instead of a file originally on disk. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
-				if _, err := exec.Command("rpm", "-qf", f.Path).CombinedOutput(); err != nil {
-					if err := os.Remove(origFileName(f.Path)); err != nil {
-						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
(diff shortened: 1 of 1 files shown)


runcom:
op failure is known and still worked on afaict

@yuqi-zhang ptal

/retest

pkg/daemon/update.go
```diff
@@ -869,12 +869,15 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *igntypes.Config) e
...
 				// file instead of a file originally on disk. See https://bugzilla.redhat.com/show_bug.cgi?id=1814397
+				var notOwnedByRpm, missingFromUsr bool
 				if _, err := exec.Command("rpm", "-qf", f.Path).CombinedOutput(); err != nil {
-					if err := os.Remove(origFileName(f.Path)); err != nil {
-						return errors.Wrapf(err, "deleting orig file %q: %v", origFileName(f.Path), err)
-					}
-				} else if _, err := os.Stat("/usr" + f.Path); strings.HasPrefix(f.Path, "/etc") && os.IsNotExist(err) &&
+					notOwnedByRpm = true
+				}
+				if _, err := os.Stat("/usr" + f.Path); strings.HasPrefix(f.Path, "/etc") && os.IsNotExist(err) &&
```
yuqi-zhang, line 876: So the thing is, if a wrongfully backed up file exists somewhere other than `etc`, we never hit this condition since `strings.HasPrefix(f.Path, "/etc")` is false, thus `missingFromUsr` is false and we end up restoring it still (now, I don't think there are scenarios where we'd hit this since the original bug that caused this fix in the first place was etcd, but its still technically wrong logic).

sinnykumari: true, didn't strike me much about non /etc cases.
Updated the conditions, hopefully it addresses missing cases.

sinnykumari:
Thanks Jerry for review and catching the issue!

PR #1785, opened 2020-06-05
Bug 1844387: Fail healthz/readyz curls on non-200 http errors

sttts:
Before this `curl -s´ only failed when the tcp connection failed, not on http non-200 errors.


runcom:
@yboaron ptal
/approve

mandre:
/test e2e-openstack
LGTM

PR #1788, opened 2020-06-05
Update keepalived ingress check's endpoint

mandre:
Based on [1] and [2] we should check for the router's readiness rather than for its liveliness.

This updates the keepalived ingress check query accordingly.

[1] https://github.com/openshift/installer/blob/master/docs/dev/kube-apiserver-health-check.md#load-balancer-health-check-probe
[2] https://github.com/openshift/router/blob/df645b1bb2c4085468845a526ac21cadb54b375d/deploy/router.yaml#L47-L48

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
@@ -3,7 +3,7 @@ path: "/etc/kubernetes/static-pod-resources/keepalived/keepalived.conf.tmpl"
 contents:
   inline: |
     vrrp_script chk_ocp {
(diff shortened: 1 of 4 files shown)


yboaron:
@mandre ,  
I crictl execed into router pod in one of my worker nodes, I curled simultaneously  **/healthz** and **healthz/ready** while I repeatedly suspend & resume the HAProxy process inside the router container, and it seems to me that both of them fail and recover on the same time.
Do we know what the difference between  **/healthz** and **/healthz/ready**  ?

yboaron:
@mandre this is [1] the answer I received from network-edge folks about the difference between  /healthz and healthz/ready , so I guess we should use healthz/ready.

[1] 
Better to use /healthz/ready.  /healthz just means the router has started, but /healthz/ready tells you that it has synched routes and is ready to serve, and /healthz/ready also stops reporting ready when the router starts shutting down, so using /healthz/ready allows the router to drain and terminate gracefully.

rgolangh:
lgtm ovirt.
@mandre thanks for ovirt change.

PR #1789, opened 2020-06-05
templates: Write kubelet proxy env settings as file instead of unit dropin

LorbusChris:
**- What I did**
The unit definition in `_base` is overwritten by platform-specific units.
In order to not require defining the dropin settings in each of the
overwrite units, and thus duplicating them, write out the dropin as an
Ignition file instead of an Ignition systemd dropin.

**- How to verify it**
- CI (is there a CI run that tests proxy settings?)

templates: Write kubelet proxy env settings as file instead of unit dropin

/cc @vrutkovs @runcom


cgwalters:
/approve
Though I think the real fix is to stop carrying multiple kubelet units and instead teach kubelet about `/etc/kubelet.conf.d` or so.

LorbusChris:
/hold

I haven't been able to repro this issue the issue with plain Ignition configs. Might have something to do with the FCC transpiler, though.

@vrutkovs do you have a CI run at hand where you hit this?

vrutkovs:
>@vrutkovs do you have a CI run at hand where you hit this?

The issue manifested itself on Ignition v3 only, so lets keep it on hold for now, until RHCOS switches - or OKD tests would be added to master

LorbusChris:
we won't need this with proper ignition systemd unit deduplication
/close

PR #1792, opened 2020-06-08
Manage the ignition stub config

runcom:
~~Nothing to see here yet, just code, and testing going on elsewhere~~

~~current testing plan~~

- ~~install with this PR (covered in CI)~~
- ~~upgrade to this PR (covered in CI)~~
- ~~test an install with https://github.com/openshift/installer/pull/3730~~
- ~~test and upgrade with the above~~

Related issue https://github.com/openshift/machine-config-operator/issues/683
Enhancement https://github.com/openshift/enhancements/pull/368

The key points here are:

- if you're installing you'll get the new user-data managed by the MCO, scale up continues working
- if you're installing with ignition v3 (4.6) this keeps working and serves the new ignition stub to new machines as the code will be typed with the new ignition structs
- if you're upgrading from pre-ign-v3 (this is gonna be the case in our release cycle so let's not bother about v2->v3 upgrade other than in this PR I guess), then your old machine(sets) will still be able to grab the ignition v2 endpoint because we don't delete the old user-data even if the code is now ignition v3 typed

What this PR does is basically:

- we let the installer take care of installing the initial user-data secret
- we then take over with our managed secret
- if we are upgrading (and thus no installer at play), we just create the new (managed) secret


manifests/master.userdata_secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: master-user-data
+  namespace: openshift-machine-config-operator
```
runcom, line 5: I actually think we should be creating this in the MAO ns, leaving a note for me to remember later

manifests/master.userdata_secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: master-user-data
```
runcom, line 4: this will be named `-new` or something to update the installer later on

runcom:
This works as expected, both alone and with https://github.com/openshift/installer/pull/3730

```
$ oc get secrets -n openshift-machine-api
NAME                                              TYPE                                  DATA   AGE
...
master-user-data-managed                          Opaque                                2      19m
worker-user-data-managed                          Opaque                                2      19m
```

The new user-data is created just fine from the MCO and read correctly by machine(sets) in MAO. If you create a cluster with both this PR and https://github.com/openshift/installer/pull/3730 you'll get a fully functional cluster using the new managed-by-MCO user-data:

```
$ oc describe machine -n openshift-machine-api ci-ln-0s13scb-d5d6b-ln8c4-master-0
...
      User Data Secret:
        Name:  master-user-data-managed
...
```
same for machinesets

Scaling up the worker machinesets works as expected as well

Next I'm going to bring up a cluster and upgrade to a payload that contains this to check what goes on (upgrade tests pass in this PR so I just want to check what's going on)

I'm going to create create the enhacement and cleanup both PRs - the MCO PR has to go in first ofc.

pkg/controller/common/helpers.go
```diff
@@ -130,6 +132,33 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 	}, nil
 }
 
+// PointerConfig ... TODO(runcom)
+func PointerConfig(ignitionHost, role string, rootCA []byte) (ign2types.Config, error) {
```
runcom, line 136: we might want to unify this with the very same function the installer has

pkg/operator/sync.go
```diff
@@ -329,6 +379,22 @@ func (optr *Operator) syncMachineConfigPools(config *renderConfig) error {
 		}
 	}
 
+	userdatas := []string{
+		"manifests/master.userdata_secret.yaml",
+		"manifests/worker.userdata_secret.yaml",
```
runcom, line 384: we'll need a watcher on those secrets so if they get changed manually, the MCO can reconcile them (much like what we do for pools or any other resource the MCO manages)

runcom:
have just confirmed on an upgraded cluster that we effectively re-generate what the installer generates so it's safe to assume we'll generate the correct ign v3 version once we upgrade :muscle:

manifests/master.userdata_secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: master-user-data-managed
```
cgwalters, line 4: (Not to totally derail this but we *do* have an opportunity now to call this `controlplane-user-data-managed` or so; the [installer config](https://github.com/openshift/installer/blob/master/docs/user/customization.md) uses the terms "controlPlane" and "compute" and not master/worker...just noting this)

pkg/controller/common/helpers.go
cgwalters: Would be nice, but doesn't seem required.  (I often follow a rule of thumb that having something in 2 places isn't a serious problem, but if it ever needs to become 3 that's when it's really required to de-duplicate)

That said let's at least add a comment in both places like:
```
// NOTE: If you change this, you also need to change the pointer configuration in openshift/installer, see
// <installer git link>
```

pkg/operator/sync.go
```diff
@@ -143,6 +146,7 @@ func (optr *Operator) syncCloudConfig(spec *mcfgv1.ControllerConfigSpec, infra *
 	return nil
 }
 
+//nolint:gocyclo
```
cgwalters, line 149: 🚳

pkg/operator/sync.go
```diff
@@ -281,11 +285,57 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
 	return nil
 }
 
+func getIgnitionHost(infraStatus *configv1.InfrastructureStatus) (string, error) {
+	internalURL := infraStatus.APIServerInternalURL
+	internalURLParsed, err := url.Parse(internalURL)
+	if err != nil {
+		return "", err
+	}
+	ignitionHost := fmt.Sprintf("%s:22623", internalURLParsed.Hostname())
```
cgwalters, line 321: Let's share port as a constant with the MCS?

pkg/operator/sync.go
```diff
@@ -281,11 +285,57 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
+	switch infraStatus.PlatformStatus.Type {
+	case configv1.BareMetalPlatformType:
+		// Baremetal needs to point directly at the VIP because we don't have a
+		// way to configure DNS before Ignition runs.
```
cgwalters, line 325: (Actually they do since https://github.com/openshift/enhancements/pull/210 )

manifests/master.userdata_secret.yaml
runcom: awesome :star_struck:

pkg/operator/sync.go
runcom: it was merely copied over from the installer, will drop!

runcom:
the secret should be syncing now - basically, our bad libgo fork wasn't properly working so the diff you see here is basically importing the right libgo pkgs to make secrets sync correctly

runcom:
> the secret should be syncing now - basically, our bad libgo fork wasn't properly working so the diff you see here is basically importing the right libgo pkgs to make secrets sync correctly

tested this out and it works :+1: if you edit the secrets (much like what happens if you edit an MCP) the MCO will rewrite the object with the correct data

runcom:
> The overall implementation makes sense. I do have a question: how are kubeMAOSharedInformer and maoSecretInformer actually being used? I don't quite see how they are currently triggering syncMachineConfigPools which as far as I understand is where the actual application of the secrets is

You can see the maoSecretInformer is added to the event handler so every time something changes there (secrets in the Mao namespace) the controller enqueues a sync and that’s when we reapply our computed manifest. kubeMAOSharedInformer Is where we grab the secret informer from (it’s like a builder like the others we have)

pkg/operator/operator.go
```diff
@@ -165,6 +167,7 @@ func New(
 		mcpInformer.Informer(),
 		proxyInformer.Informer(),
 		oseKubeAPIInformer.Informer(),
+		maoSecretInformer.Informer(),
```
runcom, line 170: Jerry this is where we use the actual informer adding it to the event handler so we get a resync every time something changes (which calls syncMachineConfigPools)

cgwalters:
Would like to see the info from the PR in the commit mesage too.  Otherwise, I think this looks good.

To be clear...this will be a no-op until the installer PR lands to start using the new managed secrets?

runcom:
> To be clear...this will be a no-op until the installer PR lands to start using the new managed secrets?

yeah, in the upgrade scenarios this is gonna be a no-op for all 4.6-dev as machines still points to the _old_ userdata name. When 4.6 goes out (upgrade scenario) we simply create the new userdata secret which users can manually use in machinesets and we can leverage to be used in a future update-bootimages feature.

In the installation case it's the same as we won't be using this until installer switches naming as well

does it make sense? 🤔 

(I'll update the commit message also good point 👍 )

runcom:
I've pushed a change to allow the creation of the userdata for every pool in the cluster (which I still it's correct as Michael pointed out in the enhancement) - it shouldn't be disruptive, we just do what we always had to do 🤷‍♂️ 

I'll test this out again and report back quickly

PR #1802, opened 2020-06-09
daemon: Run rpm-ostree directly in -firstboot.service

cgwalters:
Currently `machine-config-daemon-host` is started by all 3 cases of:

- directly via presence of `/etc/pivot/image-pullspec`
- via `machine-config-daemon-firstboot`
- via the MCD pod

This is unnecessary in the second case (we're already running as a systemd unit)
and it's quite confusing to have one systemd unit start another and
watch/replicate its logs.

Motivated by debugging https://bugzilla.redhat.com/show_bug.cgi?id=1842906

In theory this actually *widens* the race in allowing two things to try to
upgrade at the same time, but it will help us turn `machine-config-daemon-host.service`
into something more clearly targeted for handling 4.1 bootimages on firstboot, much like
how `machine-config-daemon-firstboot-v42.service` exists.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1280,9 +1280,25 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	}
 
 	glog.Infof("Updating OS to %s", newURL)
-	if err := dn.NodeUpdaterClient.RunPivot(newURL); err != nil {
-		MCDPivotErr.WithLabelValues(newURL, err.Error()).SetToCurrentTime()
-		return fmt.Errorf("failed to run pivot: %v", err)
+	// In the cluster case, for now we run indirectly via machine-config-daemon-host.service
+	// for SELinux reasons, see https://bugzilla.redhat.com/show_bug.cgi?id=1839065
+	if dn.kubeClient != nil {
+		if err := dn.NodeUpdaterClient.RunPivot(newURL); err != nil {
+			MCDPivotErr.WithLabelValues(newURL, err.Error()).SetToCurrentTime()
+			return fmt.Errorf("failed to run pivot: %v", err)
+		}
+	} else {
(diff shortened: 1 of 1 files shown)


cgwalters:
Also let me apologize for the horrendous mess around all of this code...in my defense there are some confusing external constraints like the SELinux BZ and handling 4.1 bootimages.

Once both of those get knocked out a whole lot of stuff will get cleaner.

cgwalters:
/hold
I sent this up as a cleanup, and I want to have CI test it, but let's focus on https://github.com/openshift/machine-config-operator/pull/1804

cgwalters:
And likewise for this, CI isn't telling us anything here because this code only runs in the bootimage and we're not overriding that w/our CI right now.

(This would also be fixed by https://github.com/openshift/machine-config-operator/pull/1804#issuecomment-641614026 )

pkg/daemon/update.go
```diff
@@ -1280,9 +1280,25 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig) error {
 	}
 
 	glog.Infof("Updating OS to %s", newURL)
-	if err := dn.NodeUpdaterClient.RunPivot(newURL); err != nil {
-		MCDPivotErr.WithLabelValues(newURL, err.Error()).SetToCurrentTime()
-		return fmt.Errorf("failed to run pivot: %v", err)
+	// In the cluster case, for now we run indirectly via machine-config-daemon-host.service
+	// for SELinux reasons, see https://bugzilla.redhat.com/show_bug.cgi?id=1839065
+	if dn.kubeClient != nil {
```
sinnykumari, line 1285: wondering instead of checking for kubeClient if we should check for `/etc/ignition-machine-config-encapsulated.json` file existence.

cgwalters:
Closing in favor of piling onto doing this in "one shot" with https://github.com/openshift/machine-config-operator/pull/1766
(And thanks Sinny for reviewing the code here!)

PR #1803, opened 2020-06-09
daemon: Use systemd-run, not machine-config-daemon-host.service

cgwalters:
This is the mirror/dual of
https://github.com/openshift/machine-config-operator/pull/1802

With this the MCD (when run as a pod) stops using `machine-config-daemon-host.service`.
and creates a dynamic unit instead.

With the combination of both, `machine-config-daemon-host.service` is on the
path to not being used by default and migrating to a "4.1 bootimage aid".

The systemd-run model of creating a unit dynamically is much clearer for what we want here;
conceptually the service is just a dynamic child of this pod (if we could we'd
tie the lifecycle together).  Further:

- Let's shorten our systemd unit names by using the `mco-` prefix
- Inject the `RPMOSTREE_CLIENT_ID`, see https://github.com/coreos/rpm-ostree/commit/016c1c5e627fc2a8cd3266ccda3a47a5f8992594

For example, one weird semantic of `systemctl start` is that it "joins" if
somehow the service is started for another reason.  But here, if somehow
two instances of the MCD were running then `systemd-run` will say e.g.:
`Failed to start transient service unit: Unit mco-pivot.service already exists.`

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -322,23 +322,16 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
 		return errors.Wrap(err, "deleting pivot reboot-needed file")
 	}
 
-	service := "machine-config-daemon-host.service"
-	// We need to use pivot if it's there, because machine-config-daemon-host.service
-	// currently has a ConditionPathExists=!/usr/lib/systemd/system/pivot.service to
-	// avoid having *both* pivot and MCD try to update.  This code can be dropped
-	// once we don't need to care about compat with older RHCOS.
-	var err error
-	_, err = os.Stat("/usr/lib/systemd/system/pivot.service")
+	// We used to start machine-config-daemon-host here, but now we make a dynamic
+	// unit because that service was started in too many ways, and the systemd-run
+	// model of creating a unit dynamically is much clearer for what we want here;
+	// conceptually the service is just a dynamic child of this pod (if we could we'd
+	// tie the lifecycle together).  Further, let's shorten our systemd unit names
+	// by using the mco- prefix, and we also inject the RPMOSTREE_CLIENT_ID now.
+	err := exec.Command("systemd-run", "--wait", "--collect", "--unit=mco-pivot",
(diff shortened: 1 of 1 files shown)


cgwalters:
Another reason to do this is that the current
`Before=kubelet.service` in `machine-config-daemon-host.service` is just wrong when we're running as part of the cluster.  systemd doesn't complain but still - we want to clearly separate the flow of firstboot vs node updates while having them share code.  So basically, separate systemd units running the same binaries.

cgwalters:
/hold
I sent this up as a cleanup, and I want to have CI test it, but let's focus on https://github.com/openshift/machine-config-operator/pull/1804

cgwalters:
Looking at the logs for this one...we have no coverage of this in current CI :cry:  (the upgrade test is testing *from* an older revision, so we won't see the new logs).

pkg/daemon/rpm-ostree.go
```diff
@@ -322,23 +322,16 @@ func (r *RpmOstreeClient) RunPivot(osImageURL string) error {
...
-	var err error
-	_, err = os.Stat("/usr/lib/systemd/system/pivot.service")
+	// We used to start machine-config-daemon-host here, but now we make a dynamic
+	// unit because that service was started in too many ways, and the systemd-run
+	// model of creating a unit dynamically is much clearer for what we want here;
+	// conceptually the service is just a dynamic child of this pod (if we could we'd
+	// tie the lifecycle together).  Further, let's shorten our systemd unit names
+	// by using the mco- prefix, and we also inject the RPMOSTREE_CLIENT_ID now.
+	err := exec.Command("systemd-run", "--wait", "--collect", "--unit=mco-pivot",
+		"-E", "RPMOSTREE_CLIENT_ID=mco", "/usr/libexec/machine-config-daemon", "pivot", osImageURL).Run()
```
sinnykumari, line 332: umm, so we are directly calling  m-c-d pivot here which means in case of an update pivot will run `systemctl reboot` (https://github.com/openshift/machine-config-operator/blob/master/cmd/machine-config-daemon/pivot.go#L253) . Can this cause system to reboot without executing remaining control flow?

cgwalters:
Closing in favor of piling onto doing this in "one shot" with https://github.com/openshift/machine-config-operator/pull/1766
(And thanks Sinny for reviewing the code here!)

PR #1804, opened 2020-06-09
machine-config-daemon-host.service: Stop enabling by default

cgwalters:
I haven't verified this yet but I think this is the core fix
for https://bugzilla.redhat.com/show_bug.cgi?id=1842906

Basically the MCS injects both `/etc/pivot/image-pullspec` which
4.1 bootimages need, *and* we get `/etc/ignition-machine-config-encapsulated.json`
which starts `machine-config-daemon-firstboot.service`.

We should only have one running.  I think we have a setup like this:

4.1: Uses `pivot.service` which lives on the host already
4.2: Uses `machine-config-daemon-firstboot-v42.service` from
     https://github.com/openshift/machine-config-operator/pull/1706
>= 4.3: Uses `machine-config-daemon-firstboot.service`

So let's stop starting this by default.


cgwalters:
I'll also say here I am hopeful for https://github.com/openshift/machine-config-operator/pull/1766 - basically rather than creating a mess of systemd units written by Ignition with tricky conditionals ("programming with systemd units") and where we exec a separate copy of the MCD, we dynamically inject *the same code* and run it e.g. as a systemd generator and have it generate units at runtime.

That way e.g. we can do things like "detect if this is a 4.1 or 4.2 or 4.3 bootimage" at early boot on the node, rather than using `ConditionPathExists=/sysroot/.coreos-aleph-version.json` and things like that.

runcom:
Let’s get this in after tests pass I think, I would be highly surprised if we were relying on this service always starting and this effectively helps us narrow down or fix the issue we’re seeing

/approve
/lgtm

cgwalters:
e2e-aws is https://bugzilla.redhat.com/show_bug.cgi?id=1835371
e2e-gcp-op failure is https://bugzilla.redhat.com/show_bug.cgi?id=1827712
(And we're still on track to land that fix)

Didn't bother to look at scaleup since it can't be related.

/retest

cgwalters:
OK now to be honest I'm super confused because in current 4.4 (before this PR)
```
[root@osiris-s625c-master-0 ~]# ls -al /etc/systemd/system/multi-user.target.wants/machine-config-daemon-host.service
ls: cannot access '/etc/systemd/system/multi-user.target.wants/machine-config-daemon-host.service': No such file or directory
[root@osiris-s625c-master-0 ~]# systemctl is-enabled machine-config-daemon-host
disabled
[root@osiris-s625c-master-0 ~]# 
```

Yet...we have that `[Install]` the same as the others so it should be there...

But, it turned out https://bugzilla.redhat.com/show_bug.cgi?id=1842906 was something entirely different so we can chase this down at our leisure...

templates/common/_base/units/machine-config-daemon-host.service.yaml
```diff
@@ -18,8 +18,6 @@ contents: |
   # TODO add --from-etc-pullspec after ratcheting
   ExecStart=/usr/libexec/machine-config-daemon pivot
 
-  [Install]
-  WantedBy=multi-user.target
```
sinnykumari, line 22: +1 to remove this as it may not adding up in this service.

Speaking out loud:
From the documentation https://www.freedesktop.org/software/systemd/man/systemd.unit.html ,  it says `WantedBy=  option may be used more than once, or a space-separated list of unit names may be given. A symbolic link is created in the .wants/ or .requires/ directory of each of the listed units when this unit is installed by systemctl enable.` By default m-c-d-host.service is in disabled state, so I hope this wasn't causing any issue so far.

PR #1807, opened 2020-06-10
Bug 1842906: daemon: Remove encapsulated config when joining the cluster

cgwalters:
https://bugzilla.redhat.com/show_bug.cgi?id=1842906

The removal of `BindsTo=` in https://github.com/openshift/machine-config-operator/pull/1762/commits/75dbab9c54c6cb3470075af1da1b139ecea02d38
I think was completely correct for *new* installs, but creates
an upgrade hazard if we happened to be interrupted when
the cluster was originally installed as 4.4 or below.

Basically the scenario is:

- 1st boot: 4.4 install, machine boots the first time
- restart.service kills machine-config-daemon-firstboot before it upgrades, or before it unlinks the file
- 2nd boot: We stumble on without having pivoted, node joins cluster
  MCD starts a reboot into 44.81.202006062030-0 which we *should* have updated to before joining the cluster
- Cluster upgrade to 4.5 starts, new MCO starts an updated OS rollout
- 3rd boot: Prepared new root with 4.5, and the new machine-config-daemon-firstboot without BindsTo= gets laid down in /etc
- 4th boot (current): In 4.5, and -firstboot triggers again, trying to finally apply the original config

When the MCD (running as a pod) is going we shouldn't run `-firstboot` again,
the MCD should handle everything.  So just rename the file to
signal completion.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -914,6 +914,24 @@ func (dn *Daemon) storeCurrentConfigOnDisk(current *mcfgv1.MachineConfig) error 
 	return writeFileAtomicallyWithDefaults(dn.currentConfigPath, mcJSON)
 }
 
+// https://bugzilla.redhat.com/show_bug.cgi?id=1842906
+// If we didn't successfully complete -firstboot.service, because
+// 4.5 and newer removed the BindsTo=, the service may start downgrading
+// things.  At this point we should have already applied all target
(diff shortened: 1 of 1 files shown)


sinnykumari:
e2e-gcp-op seems to be hitting https://bugzilla.redhat.com/show_bug.cgi?id=1827712
/retest

ashcrow:
Continued infra flakes. Waiting to retry.

PR #1809, opened 2020-06-10
Remove pivot invocations of systemctl reboot

cgwalters:
While chasing https://bugzilla.redhat.com/show_bug.cgi?id=1842906
I thought at one point this code might be invoked and was causing
a reboot that shouldn't be happening.

That turned out not to be the case.  But since nothing uses
the `--reboot` or stamp file anymore (we use the `-firstboot`)
code, the MCD code is always controlling reboots, delete this code.

Code that doesn't exist can't have (potential) bugs.

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -22,12 +22,12 @@ import (
 
 // flag storage
 var keep bool
-var reboot bool
 var fromEtcPullSpec bool
 
 const (
-	etcPivotFile       = "/etc/pivot/image-pullspec"
-	runPivotRebootFile = "/run/pivot/reboot-needed"
+	// etcPivotFile is used for 4.1 bootimages and is how the MCD
+	// currently communicated with this service.
+	etcPivotFile = "/etc/pivot/image-pullspec"
 	// File containing kernel arg changes for tuning
 	kernelTuningFile = "/etc/pivot/kernel-args"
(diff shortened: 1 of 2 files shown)


cmd/machine-config-daemon/pivot.go
```diff
@@ -22,12 +22,12 @@ import (
...
 // flag storage
 var keep bool
-var reboot bool
 var fromEtcPullSpec bool
 
 const (
-	etcPivotFile       = "/etc/pivot/image-pullspec"
-	runPivotRebootFile = "/run/pivot/reboot-needed"
+	// etcPivotFile is used for 4.1 bootimages and is how the MCD
+	// currently communicated with this service.
```
LorbusChris, line 29: nit: `currently communicated` would've made my English teacher's stomach turn :laughing: 
Is it `currently communicates` or `previously communicated` (previously as in before this PR)?

cgwalters:
Using this as a victim PR to test https://github.com/openshift/release/pull/9835
/test cluster-bootimages

cgwalters:
Hmm, maybe it takes a few minutes to roll out after the Prow configmap updates.  Will retry later.

cgwalters:
/test cluster-bootimages
Hmm, seems to have hung

cgwalters:
Seems like the test is using the wrong `KUBECONFIG`?  I need to debug that...

Issue #1585, opened 2020-03-26
nodeip-configuration 'Failed to find suitable node ip'


e-minguez:
In recent OCP4 versions (4.4.0-202003230106 at least), nodeip-configuration service is failing:

```
[root@kni1-worker-0 ~]# systemctl status nodeip-configuration.service
● nodeip-configuration.service - Writes IP address configuration so that kubelet and crio services select a valid node IP
   Loaded: loaded (/etc/systemd/system/nodeip-configuration.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2020-03-25 13:47:46 UTC; 1h 10min ago
  Process: 2092 ExecStart=/usr/local/bin/nodeip-finder 10.19.138.14 (code=exited, status=1/FAILURE)
 Main PID: 2092 (code=exited, status=1/FAILURE)
      CPU: 106ms
 
Mar 25 13:47:46 localhost.localdomain systemd[1]: Starting Writes IP address configuration so that kubelet and crio services select a valid node IP...
Mar 25 13:47:46 localhost.localdomain nodeip-finder[2092]: Filtering out Address(127.0.0.1/8, dev=lo) due to it having host scope
Mar 25 13:47:46 localhost.localdomain nodeip-finder[2092]: Filtering out Address(::1/128, dev=lo) due to it having host scope
Mar 25 13:47:46 localhost.localdomain nodeip-finder[2092]: Is 10.19.138.14 between 172.22.0.0 and 172.22.0.255
Mar 25 13:47:46 localhost.localdomain nodeip-finder[2092]: Is 10.19.138.14 between fe80:: and fe80::ffff:ffff:ffff:ffff
Mar 25 13:47:46 localhost.localdomain nodeip-finder[2092]: Failed to find suitable node ip
Mar 25 13:47:46 localhost.localdomain systemd[1]: nodeip-configuration.service: Main process exited, code=exited, status=1/FAILURE
Mar 25 13:47:46 localhost.localdomain systemd[1]: nodeip-configuration.service: Failed with result 'exit-code'.
Mar 25 13:47:46 localhost.localdomain systemd[1]: Failed to start Writes IP address configuration so that kubelet and crio services select a valid node IP.
Mar 25 13:47:46 localhost.localdomain systemd[1]: nodeip-configuration.service: Consumed 106ms CPU time
```

It seems the problem is the service was executed before an address in the vip subnet was set.

e-minguez:
[...]

PR #1813, opened 2020-06-10
Bug 1845885: templates: add node-valid-hostname.service for hostname check

darkmuggle:
This seeks to address the problem with "localhost" nodes attempting to
join a cluster by holding up network-online.target until the node gets a
valid hostname (not localhost).

Closes 1845885: race condition during installation between nodes getting their
                            hostnames and crio+kubelet starting

### DIFF
--- a/templates/common/_base/units/node-valid-hostname.service
+++ b/templates/common/_base/units/node-valid-hostname.service
@@ -1,0 +1,24 @@
+name: node-valid-hostname.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Ensure hostname is not localhost
+  # Only run when the host has a localhost name.
+  ConditionHost=|localhost
+  ConditionHost=|localhost.localdomain
+  Before=network-online.target
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStartPre=/bin/echo "Node has localhost hostname. Waiting for new hostname."
+  # Get the short hostname. This is more reliable than a regex.
+  ExecStartPre=/bin/bash -c 'while [ `hostname -s` == "localhost" ]; do sleep 1; done;'
+  ExecStart=/bin/sh -c "echo Node changed hostname to `hostname`"
+  # Wait up to 5min for the node to get a real hostname.
+  TimeoutSec=300
+
+  [Install]
+  WantedBy=multi-user.target
+  # Ensure that network-online.target will not complete until the node has a real hostname.
+  RequiredBy=network-online.target


cgwalters:
> This unit would fail since it would be required by network-online.target. 

Ah, true.  OK.  That definitely addresses the concern then.

darkmuggle:
To follow up on the idea of making this a hard-fail, https://github.com/darkmuggle/machine-config-operator/commit/73ea1ce3a4526532a6e75c6f469786d74da8cfed makes `network-online.target` a hard requirement. 

Frankly, I'm unsure if this is a good idea or not; it seems reasonable, but I'd need upstack confirmation before I *would even think* of proposing it.

templates/common/_base/units/node-valid-hostname.service
```diff
@@ -0,0 +1,24 @@
+name: node-valid-hostname.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Ensure hostanme is not localhost
```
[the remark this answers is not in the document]
darkmuggle: Great catch! Fixed.

darkmuggle:
> I'm not super knowledgeable but was going to suggest that platform ci needs to run and this pr should also be looked at by the affected platforms to ensure this doesn't somehow break them... (I'm really not sure as they manage their own templates, but worth asking).

The effect of this would be move all the failures related to `localhost` into one spot, instead of the various components. I thought long and hard about whether we want the individual units to do the check and it seemed cleaner to just prevent those things from starting in the first place. 

But yes, putting this through platform CI should be done. I did look at the all the hostname setting components (intiramfs, NetworkManager, Ignition, and the MCO code) and I didn't see anything that would break per se.

The change does change make the hostname requirement of anything but localhost more or less explicit.

kikisdeliveryservice:
@mandre  should i expect this e2e-openstack job to pass? 
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1813/pull-ci-openshift-machine-config-operator-master-e2e-openstack/82

sinnykumari:
Since these are templates, I think service filename should be node-valid-hostname.service.yaml

e-minguez:
I guess this will fix this as well? https://github.com/openshift/machine-config-operator/issues/1585

pierreprinetti:
ShiftStack's CI is having a hard time right now. We are investigating what is going on. In the meantime, I am triggering a
/retest

cgwalters:
One thing is we have random lists of hardcoded systemd units in various places, and this will be an important one that might not be captured today, see e.g.

https://github.com/openshift/installer/pull/3736
https://github.com/openshift/must-gather/pull/158

Also how about e.g. `kubelet-valid-hostname.service`?

darkmuggle:
> Also how about e.g. `kubelet-valid-hostname.service`?

I went with `node-..` instead of `kublet-...` since its blocking on `network-online.target`. The binding is not specific to the `kublet`.

cgwalters:
Hmm right.  But actually, it *is* just specific to the kubelet I think.  Or well, I guess crio too potentially.

But we also have other units that are `After=network-online.target` like the WIP https://github.com/openshift/machine-config-operator/pull/1766# that don't care what the hostname is.  On the other hand, in all sane cases we're going to get the hostname anyways either as part of a sane DHCP response or it'll be static, so it doesn't matter.

So OK as is too!

kikisdeliveryservice:
this job does have a recent pass so trying again

/test e2e-vsphere

cybertron:
I would expect this to be fine from a baremetal perspective. We already have workarounds for this[0] in place so moving where the check happens won't affect us much. It should just mean we can drop our hostname checks.

The one concern I have is that on baremetal the deployer controls the hostnames, so it's possible they could pick something that would match the localhost regex. In our version we explicitly checked for localhost and localhost.localdomain to avoid that.

0: https://github.com/openshift/machine-config-operator/blob/650e0009a28e9cc27ef3e7a19273a858cfd76021/templates/common/baremetal/files/baremetal-mdns-publisher.yaml#L46

darkmuggle:
> The one concern I have is that on baremetal the deployer controls the hostnames, so it's possible they could pick something that would match the localhost regex. In our version we explicitly checked for localhost and localhost.localdomain to avoid that.
 
Based on when the new template would run -- before `network-online.target` -- I don't think there will be a conflict and it only prevents services requiring `network-online.target` from starting until a valid node name is determined; something could change it to `localhost` later.

kikisdeliveryservice:
@pierreprinetti what is the status of the openstack job?

cybertron:
> > The one concern I have is that on baremetal the deployer controls the hostnames, so it's possible they could pick something that would match the localhost regex. In our version we explicitly checked for localhost and localhost.localdomain to avoid that.
> 
> Based on when the new template would run -- before `network-online.target` -- I don't think there will be a conflict and it only prevents services requiring `network-online.target` from starting until a valid node name is determined; something could change it to `localhost` later.

My concern is that the valid node name might be something like "localhost-abc". Our hostnames are deployer-assigned so they can use anything they want. If they chose something with localhost in it then this check would never pass.

I don't expect this to happen since I've never seen anyone use "localhost" as part of their hostname format, but technically it is possible.

darkmuggle:
> I don't expect this to happen since I've never seen anyone use "localhost" as part of their hostname format, but technically it is possible.

I changed the regex to `'while [[ "$(< /proc/sys/kernel/hostname)" =~ ^localhost(|\.localdomain)$ ]]; do sleep 1; done;'` which should eliminate this concern.

/retest

darkmuggle:
/hold 

Doing the extra-due diligence testing on with the new regex to be doubly sure.

darkmuggle:
/unhold

@runcom @cgwalters if this looks sane can I get an LGTM/approval? I've put the last version through the paces and this should resolve a bunch of the `localhost` problems.

darkmuggle:
In the last push, I changed from reading the kernel hostname to using `hostname -s` since this is an exact test for any combination of `localhost` or `localhost.localdomain`.

templates/common/_base/units/node-valid-hostname.service
```diff
@@ -0,0 +1,25 @@
...
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStartPre=/bin/echo "Node has localhost hostname. Waiting for new hostname."
+  # Get the short hostname. This is more reliable than a regex.
+  ExecStartPre=/bin/bash -c 'while [ `hostname -s` == "localhost" ]; do sleep 1; done;'
+  ExecStart=/bin/sh -c "echo Node changed hostname to `hostname`"
+  # Wait up to 5min for the node to get a real hostname.
+  KillMode=process
```
cgwalters, line 19: Why?  This is asking to leak processes and it's not clear to me why we want that.

templates/common/_base/units/node-valid-hostname.service
```diff
@@ -0,0 +1,25 @@
...
+  TimeoutSec=300
```
cgwalters, line 20: You can also write `TimeoutSec=5m` FWIW.

templates/common/_base/units/node-valid-hostname.service
```diff
@@ -0,0 +1,25 @@
...
+
+  [Install]
+  WantedBy=multi-user.target
```
cgwalters, line 23: I don't think we need this - if `network-online.target` isn't requested then we shouldn't run.

cgwalters:
BTW there is also https://github.com/openshift/machine-config-operator/pull/1770 which I even reviewed and then forgot about...

darkmuggle:
Logs showing this working on GCP
```
-- Logs begin at Tue 2020-06-16 02:19:38 UTC, end at Tue 2020-06-16 03:15:10 UTC. --
Jun 16 02:22:33 localhost systemd[1]: Starting Ensure hostname is not localhost...
Jun 16 02:22:33 localhost echo[1260]: Node has localhost hostname. Waiting for new hostname.
Jun 16 02:22:34 ci-ln-s5vcym2-f76d1-b75vt-master-2 sh[1435]: Node changed hostname to ci-ln-s5vcym2-f76d1-b75vt-master-2
Jun 16 02:22:34 ci-ln-s5vcym2-f76d1-b75vt-master-2 systemd[1]: Started Ensure hostname is not localhost.
Jun 16 03:12:37 ci-ln-s5vcym2-f76d1-b75vt-master-2 systemd[1]: Stopped Ensure hostname is not localhost.
Jun 16 03:12:37 ci-ln-s5vcym2-f76d1-b75vt-master-2 systemd[1]: node-valid-hostname.service: Consumed 18ms CPU time
-- Reboot --
Jun 16 03:13:31 localhost systemd[1]: Starting Ensure hostname is not localhost...
Jun 16 03:13:31 localhost echo[1260]: Node has localhost hostname. Waiting for new hostname.
Jun 16 03:13:32 ci-ln-s5vcym2-f76d1-b75vt-master-2 sh[1438]: Node changed hostname to ci-ln-s5vcym2-f76d1-b75vt-master-2
Jun 16 03:13:32 ci-ln-s5vcym2-f76d1-b75vt-master-2 systemd[1]: Started Ensure hostname is not localhost.
```

cybertron:
Thanks for the change. metal-ipi is likely failing on an IPv6 DHCP issue that we're currently investigating.

darkmuggle:
@cybertron is the failure on `e2e-metal-ipi` caused by the aforementioned DHCPv6 issue? If so, if it okay to skip the `e2e-metal-ipi` test?

ashcrow:
I'll give a little bit for @cybertron to chime in, but it doesn't look like `e2e-metal-ipi` is a blocking test which means we could merge with the normal tag. I'll do that in a bit if we don't hear back.

cybertron:
Yeah, looks like it's failing on the known issue. Since we already have these checks in place I wouldn't anticipate this being a problem for us so I'm +1 on going ahead with it.

cgwalters:
Did you see [my review comments above](https://github.com/openshift/machine-config-operator/pull/1813/files/0d887f03d46c74ba063ca6322a82186c539997eb#r440553228) ?  Not a hard blocker but would be nice to address/reply if you can!

templates/common/_base/units/node-valid-hostname.service
darkmuggle: https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml#L7 

The kublet is started _after_ network-online.target.

templates/common/_base/units/node-valid-hostname.service
darkmuggle: Removed the `KillMode` and repushed.

darkmuggle:
> Did you see [my review comments above](https://github.com/openshift/machine-config-operator/pull/1813/files/0d887f03d46c74ba063ca6322a82186c539997eb#r440553228) ? Not a hard blocker but would be nice to address/reply if you can!

done

Waiting on retesting.

templates/common/_base/units/node-valid-hostname.service
cgwalters: Right, I'm talking about removing the
`WantedBy=multi-user.target`

PR #1823, opened 2020-06-15
[MCO] Ignore Windows Nodes

ravisantoshgudimetla:
As of now, MCC is watching for all the nodes in the cluster. While this is not
a problem, with the introduction of worker Windows Nodes in the cluster, MCC
will react to changes happening to the Windows Node objects. This commit ensures
that MCC ignores the changes for the Windows node objects by not assigning them
to the worker pool

**- What I did**
Make MCC ignore Windows worker node events.

**- How to verify it**
Updated the unit tests to exercise those scenarios.

MCO should watch only for Linux node events

 I couldn't write a e2e to validate the changes as it requires bringing a Windows Worker node. Here is the ref for the [label](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#kubernetes-io-os)
xref: https://issues.redhat.com/browse/GRPA-2259, https://issues.redhat.com/browse/GRPA-2213

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -118,6 +118,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 
 // isNodeManaged checks whether the MCD has ever run on a node
 func isNodeManaged(node *corev1.Node) bool {
+	if isWindows(node) {
+		glog.V(4).Infof("Node %v is a windows node so won't be managed by MCO", node.Name)
+		return false
+	}
 	if node.Annotations == nil {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/node/status.go
```diff
@@ -118,6 +118,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 
 // isNodeManaged checks whether the MCD has ever run on a node
 func isNodeManaged(node *corev1.Node) bool {
+	if isWindows(node) {
```
ravisantoshgudimetla, line 121: Ideally we might not need this as [this](https://github.com/openshift/machine-config-operator/pull/1823/files#diff-3bd49282ca65e5525da44a27caaa55c7L124) should fail but I wanted to be explicit and fail early.

runcom:
@cgwalters ptal - I think this is the safest thing to do as the MCD runs on linux only anyway (for now)

/approve

PR #1824, opened 2020-06-15
docs: custom-pools work w/o the worker role label

runcom:
Update the docs for custom pools showing that you can drop the worker role label to have "infra-only" nodes (which is valid for any custom pool also).

### DIFF
--- a/docs/custom-pools.md
+++ b/docs/custom-pools.md
@@ -25,6 +25,14 @@ ip-10-0-151-146.us-west-1.compute.internal   Ready     master         43m   v1.1
 ip-10-0-152-59.us-west-1.compute.internal    Ready     worker         37m   v1.14.0+e020ea5b3
 ```
 
+When we create infra nodes, we apply them to existing worker nodes first. If you want to use the node as a purely infra node, you can remove the worker pool as follows:
+
+```
+oc label node ip-10-0-130-218.us-west-1.compute.internal node-role.kubernetes.io/worker-
+```
+
+This will not change the files on the node itself: the infra pool inherits from workers by default. This means that if you add a new MachineConfig to update workers, a purely infra node will still get updated. However this will mean that workloads scheduled for workers will no longer be scheduled on this node, as it no longer has the worker label.
+
 Next you need to create a MachineConfigPool that contains both the `worker` role and your custom one as MachineConfig selector as follows:
 
 ```console
@@ -117,6 +125,9 @@ Removing a custom pool requires first to un-label each node:
 ```console
 $ oc label node ip-10-0-130-218.us-west-1.compute.internal node-role.kubernetes.io/infra-
 ```
+
+Note: a node must have a role at any given time to be properly functioning. If you have infra-only nodes,
+you should first relabel the node as worker and only then proceed to unlabel it from the infra role.
 
 ```console
 $ oc get nodes


docs/custom-pools.md
```diff
@@ -25,6 +25,12 @@ ip-10-0-151-146.us-west-1.compute.internal   Ready     master         43m   v1.1
 ip-10-0-152-59.us-west-1.compute.internal    Ready     worker         37m   v1.14.0+e020ea5b3
 ```
 
+You can drop the worker role if you need infra-only nodes:
+
+```
+oc label node node ip-10-0-130-218.us-west-1.compute.internal node-role.kubernetes.io/worker-
```
yuqi-zhang, line 31: I think you have an extra `node` in `oc label node node`

docs/custom-pools.md
```diff
@@ -118,6 +124,9 @@ Removing a custom pool requires first to un-label each node:
 $ oc label node ip-10-0-130-218.us-west-1.compute.internal node-role.kubernetes.io/infra-
 ```
 
+Note: a node must have a role at any given time to be properly functioning. If you had infra-only nodes,
```
yuqi-zhang, line 127: Nits: "If you had" -> "If you have", "relabel then node" -> "relabel the node"

pkg/controller/node/node_controller.go
```diff
@@ -464,7 +464,9 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 		return
 	}
 
-	pools, err := ctrl.getPoolsForNode(curNode)
+	// enqueue all pools from the "old" node object as this update call might
+	// include a change where a role has been dropped.
+	pools, err := ctrl.getPoolsForNode(oldNode)
```
yuqi-zhang, line 469: Sorry this is slightly confusing, wouldn't this ONLY sync pools for `oldNode`? In the proposed scenario it would work since oldNode has everything in curNode + more, but wouldn't the opposite be true (and cause problems) if you're trying to add more labels (more pools) to a node?

`pools` should be a merge of the two lists instead, right?

kikisdeliveryservice:
>     1. if that isn't the case, if we drop a `worker` label and only have infra, does that mean that the rendered-infra-xxx would only contain MCs targeting infra? Wouldn't it lose all the base template files/etc. that comes with the worker node

Yeah I'm wondering the same. Are we saying that infra nodes would start with the worker base templates (bc you have to start somewhere), then apply the infra mc config and never pick up any worker mc changes from there on out (once it's an infra only node)?

runcom:
The confusion here is between the role of the node and our mapping to MCPs.
The infra or custom mcp does inherit from the worker one because of the machineSelector and that makes sure that any MC that targets worker and custom mcps are taken into consideration when generating the rendered config for the custom pool. That works besides having the node labeled as a worker as well - so yeah, the worker role is redundant.

> Yeah I'm wondering the same. Are we saying that infra nodes would start with the worker base templates (bc you have to start somewhere), then apply the infra mc config and never pick up any worker mc changes from there on out (once it's an infra only node)?

No, if you spin up a cluster, create an infra mcp and upgrade the cluster, infra gets updated and that’s because the infra mcp also always contain worker machine configs (machineSelector). (Same answer for Kirsten and Jerry 1b questions)

> 1a) Is that assessment correct? If it is, even if we removed that labelling, the node would still effective be a worker node no?

Yes, the configuration for that node labeled just infra is a merge of workers configurations (this is a must) and additional custom configurations.

pkg/controller/node/node_controller.go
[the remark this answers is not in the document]
runcom: Good point, I’ll grab both labels thanks!

yuqi-zhang:
Ok I think I understand, this should be an UI only change (I'm going to verify locally). I think we should clarify this in the docs too, instead of:

`You can drop the worker role if you need infra-only nodes`

maybe we can say:

```
When we create infra nodes, we apply them to existing worker nodes first. If you want to use the node as a purely infra node, you can remove the worker pool as follows:

...

This will not change the files on the node itself: the infra pool inherits from workers by default. This means that if you add a new MachineConfig to update workers, a purely infra node will still get updated. However this will mean that workloads scheduled for workers will no longer be scheduled on this node, as it no longer has the worker label.

```

runcom:
> maybe we can say:

that is indeed the perfect wording for this - I'm gonna change it to that :+1:

runcom:
alrighty, this is now a docs-only PR with all your suggestions addressed :+1:

PR #1825, opened 2020-06-15
Bug 1841358: kubelet: retry kubeletconfigs if the MCP pool is not found

rphillips:
**- What I did**
We ignored retrying KubeletConfigs if the MCP pool was not available. This fix will retry applying the KubeletConfig if the MCP pool is not present.

**- How to verify it**
We did not retry by returning a ForgettableError. This change removes the forgettableerror and returns the standard Error object.

```
Retry applying KubeletConfigs if the MachineConfigPool is not present.
```

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -61,7 +61,7 @@ var updateBackoff = wait.Backoff{
 	Jitter:   1.0,
 }
 
-var errCouldNotFindMCPSet = newForgetError(errors.New("could not find any MachineConfigPool set for KubeletConfig"))
+var errCouldNotFindMCPSet = errors.New("could not find any MachineConfigPool set for KubeletConfig")
 
 // Controller defines the kubelet config controller.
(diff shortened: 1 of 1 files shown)


runcom:
Thanks Ryan
/approve
/lgtm

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -61,7 +61,7 @@ var updateBackoff = wait.Backoff{
 	Jitter:   1.0,
 }
 
-var errCouldNotFindMCPSet = newForgetError(errors.New("could not find any MachineConfigPool set for KubeletConfig"))
+var errCouldNotFindMCPSet = errors.New("could not find any MachineConfigPool set for KubeletConfig")
```
runcom, line 64: I didn’t take a stab in the BZ as to why this was originally a forgettable error but maybe you remember (non blocking ofc)

rphillips: I remember writing this code, I think it was just an oversight on my part.

PR #1826, opened 2020-06-15
test/old-bootimages: New test

cgwalters:
Let's validate that old bootimages can join the cluster.
I started this in shell because it seemed simpler, then it got
messy.

Still need to finish the bits around validating the nodes join,
but I did test this manually and it worked!

### DIFF
--- a/test/e2e/old-bootimages.sh
+++ b/test/e2e/old-bootimages.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+set -euo pipefail
+# Validate that workers starting from old (4.1, 4.2) bootimages can join an AWS cluster.
+# Motivated by https://github.com/openshift/machine-config-operator/pull/1766
+# Also xref https://github.com/openshift/enhancements/pull/201
+
+oldreleases=('4.1' '4.2')
+
+fatal() {
+    echo "error: $@" >&2
+    exit 1
+}
+
+runverbose() {
+    (set -xeuo pipefail && "$@")
+}
+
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
Were you thinking of adding this as a `/hack` script or an actual e2e test within the golang framework?

cgwalters:
Tentatively thinking of creating a new `e2e-aws-bootimages` job that is optional (i.e. run on demand) mostly when we are changing code that might affect this.

Could also make sense to do as part of a periodic (this is the "run MCO's tests regularly" thread).

Or we could make `e2e-gcp-op` become `e2e-aws-op` (again IIRC) and glom this on the end there.

Open for discussion!

kikisdeliveryservice:
i think the existing e2e is already too long so i'd prefer it to be it's own test.

kikisdeliveryservice:
But would also prefer this to be in go not shell :smile:

cgwalters:
> But would also prefer this to be in go not shell

Yeah, I regretted it about halfway through but I'm also not likely to rewrite it myself right now.

Anyways I've validated this test passes with `4.4.0-0.nightly-2020-06-14-142924` and am now starting a cluster with https://github.com/openshift/machine-config-operator/pull/1766/ to validate it against that.

kikisdeliveryservice:
just adding a hold for now 

/hold

cgwalters:
I merged this into https://github.com/openshift/machine-config-operator/pull/1766/

PR #1828, opened 2020-06-16
WIP [FCOS] templates/common: nodeip-configuration should run after node has pivoted

vrutkovs:
/cc @LorbusChris 

This should fix vSphere IPI setup


vrutkovs:
That doesn't seem to fix it

PR #1829, opened 2020-06-16
vsphere nodeip-configuration.service: run node-ip binary from container

vrutkovs:
**- What I did**
Replaced python `nodeip-finder` script with a baremetal's version running in container. This is required on FCOS, which doesn't have python interpreter. Also, this would deduplicate fixes for the same functionality

**- How to verify it**
vSphere IPI install should pass

Container is used to determine node IP for vSphere

/cc @jcpowermac @patrickdillon


vrutkovs:
https://github.com/openshift/release/pull/9702 would add IPI tests. UPI flow is not affected with this change

jcpowermac:
Let's retry IPI - bootstrap didn't complete
/test e2e-vsphere

jcpowermac:
cri-o is unhappy, that is the wrong ip for the node

```
2: ens192: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:50:56:b9:5e:70 brd ff:ff:ff:ff:ff:ff
    inet 139.178.94.234/25 brd 139.178.94.255 scope global dynamic noprefixroute ens192
```

```
Jun 17 01:55:43 ci-op-jgt1ifhj-6b38f-ss4gt-master-0 crio[2805]: time="2020-06-17 01:55:43.015599455Z" level=fatal msg="Failed to start streaming server: listen tcp 139.178.94.235:10010: bind: cannot assign requested address"
Jun 17 01:55:43 ci-op-jgt1ifhj-6b38f-ss4gt-master-0 systemd[1]: crio.service: Main process exited, code=exited, status=1/FAILURE
Jun 17 01:55:43 ci-op-jgt1ifhj-6b38f-ss4gt-master-0 systemd[1]: crio.service: Failed with result 'exit-code'.
```

vrutkovs:
Ah, the service runs right after the firstboot, records that IP - and then machine-config-daemon-host.service pivots to new machine-os-content, reboots - and VM gets another IP

vrutkovs:
Cleaned up the commits and added `systemctl daemon-reload` so that cri-o would catch new settings if node IP changes between first boot and post-pivot boot

jcpowermac:
/lgtm
Thanks @vrutkovs !

cgwalters:
/approve

But I would really like to see us rethink this stuff at some point after 4.6 because we have much more powerful abilities since https://github.com/openshift/enhancements/pull/291 particularly for VSphere to control exactly what the node is doing from the first boot.

PR #1834, opened 2020-06-16
Bug 1822945: templates: add a file to load legacy iptables kernel modules

danwinship:
**- What I did**
Added a template file to create `/etc/modules-load.d/iptables.conf` to force the legacy iptables module to be loaded; this guarantees that pods that try to create iptables rules in their own network namespace will be able to, regardless of whether they use iptables-legacy or iptables-nft.

(Specifically, this is about making legacy and third-party pods work. We can't just say "change the pod to do X instead" because this is about fixing the behavior of pods that we don't control.)

**- How to verify it**
Create a privileged, non-hostNetwork pod, verify that "iptables-legacy -L" (or just "iptables -L" with older iptables) works rather than giving an error

The legacy iptables kernel module is now loaded by default on OCP nodes, to allow pods to use legacy iptables in their own network namespace.

### DIFF
--- a/templates/common/_base/files/iptables-modules.yaml
+++ b/templates/common/_base/files/iptables-modules.yaml
@@ -1,0 +1,6 @@
+mode: 0644
+path: "/etc/modules-load.d/iptables.conf"
+contents:
+  inline: |
+    # Force-load legacy iptables so it is usable from pod network namespaces
+    ip_tables


cgwalters:
I'm not opposed to this, but note that any privileged pod can also do e.g.:
`systemd-run modprobe ip_tables`
too.

templates/common/_base/files/iptables-modules.yaml
```diff
@@ -0,0 +1,6 @@
+mode: 0644
+path: "/etc/modules-load.d/iptables.conf"
+contents:
+  inline: |
+    # Force-load legacy iptables so it is usable from pod network namespaces
+    ip_tables
```
cgwalters, line 6: Is just this sufficient?  Aren't there a ton of other kernel modules related to netfilter?

danwinship:
> I'm not opposed to this, but note that any privileged pod can also do e.g.:
> `systemd-run modprobe ip_tables`
> too.

They could also upgrade to use iptables-nft; this is a compatibility fix for existing pods. Existing third-party pods expect to be able to use iptables(-legacy) in their own network namespace and have it work.

templates/common/_base/files/iptables-modules.yaml
danwinship: There are, but they get autoloaded as needed. The top-level module needs to be explicitly modprobed. (`/sbin/iptables` will try to do this itself if it sees that the module needs to be loaded, but that won't work when it's running from a container).

danwinship:
/hold
I tested this by modprobing manually but I have to use cluster-bot to bring up a cluster to test that it works this way...

templates/common/_base/files/iptables-modules.yaml
dcbw: @can you make sure nftables is modprobed first? I don't know about the ordering here, but we need to make sure nftables gets loaded before iptables for some corner cases, and if we're force-loading iptables, that may not happen in the right order. ICBW though.

danwinship: > we need to make sure nftables gets loaded before iptables for some corner cases

explain?

dcbw: @erig0 can you describe the issues around iptables/nftables? Is it module load order that matters, or is it just whatever process runs the first command that adds the hooks?

erig0: I think the issue @dcbw is recalling is actually about NAT. In old kernels, if `iptables_nat` is loaded it would either a) cause nftables NAT to never work or b) cause nftables nat hook registration to fail - all depending on how old the kernel is. But [it was fixed back in 4.18](https://patchwork.ozlabs.org/project/netfilter-devel/cover/20180514214659.1757-1-fw@strlen.de/) (not 100% sure about kernel version).

The current behavior is:
  - both iptables_nat and nftables may register NAT hooks
  - the one that process first (i.e. highest priority hook) "wins" and the later NAT does nothing

danwinship: (Further clarifying from IRC: to the extent that any problem still exists, it happens if you create iptables-legacy NAT rules before creating iptables-nft NAT rules, _not_ if you just load `ip_tables` before `nf_tables`. So this PR is fine.)

PR #1844, opened 2020-06-18
Bug 1842906: pkg/daemon: run 4.4 upgrade hack on MCD cluster boot

runcom:
It seems `upgradeHackFor44AndBelow` was running inside an if statement
that only runs on cluster installation. The bug we got happens on
upgrade so what we really want is to run the hack once the MCD is
running from within the cluster (e.g. during an upgrade).

This PR still targets the now-verified 4.6 BZ - if we confirm this fix (and the verification was then flacky) we'll have to move 1842906 back to POST

@cgwalters @sinnykumari ptal

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -419,6 +419,10 @@ func (dn *Daemon) syncNode(key string) error {
 	// and then proceeds to check the state of the node, which includes
 	// finalizing an update and/or reconciling the current and desired machine configs.
 	if dn.booting {
+		// Be sure only the MCD is running now, disable -firstboot.service
+		if err := upgradeHackFor44AndBelow(); err != nil {
+			return err
+		}
 		if err := dn.checkStateOnFirstRun(); err != nil {
 			return err
 		}
@@ -990,11 +994,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 
 	// Bootstrapping state is when we have the node annotations file
 	if state.bootstrapping {
-		// Be sure only the MCD is running now, disable -firstboot.service
-		if err := upgradeHackFor44AndBelow(); err != nil {
-			return err
-		}
-
 		targetOSImageURL := state.currentConfig.Spec.OSImageURL
 		osMatch, err := dn.checkOS(targetOSImageURL)
 		if err != nil {


pkg/daemon/daemon.go
```diff
@@ -990,11 +994,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 
 	// Bootstrapping state is when we have the node annotations file
 	if state.bootstrapping {
```
runcom, line 996: i.e. this runs only at cluster installation

cgwalters: Oh duh, yep.

We should be able to actually test this too by injecting a `MachineConfig` with the `reboot.service` in question, I was lazy the first time and didn't do that but it's probably worth doing now to avoid another potential QE cycle?

Or it should replicate by just - on a cluster with this pr, do `mv /etc/ignition-machine-config-encapsulated.{bak,json}` and kill the MCD pod.

cgwalters:
Thank you for debugging this and sorry I didn't test the PR the first time!

pkg/daemon/daemon.go
runcom: > We should be able to actually test this too by injecting a MachineConfig with the reboot.service in question

So the test would be:

- spin up a 4.4 cluster with a MachineConfig with restart.service
- upgrade to a 4.5/4.6 that does have this fix

right? I'm just worried about the fact the bug doesn't happen 100% of the times and we would waste everybody's cycles

runcom: I mean, if the intention of 1822 was this fix but this is just fixing the actual call to the hack, I think we can merge this and let QE test again right?

runcom:
not fully sure what's gonig on with CI and or this PR - tide isn't even in the checks as of now :man_shrugging: 

/retest

yuqi-zhang:
/hold

I think there's a small problem with our approach. Even if this is in a new 4.5 payload, existing 4.4 versions still cannot be upgraded because the MCD-firstboot service would have ran before this check. i.e. this codepath will not run in time for clusters that didn't already have an MCD on it before upgrading to 4.5

The alternative is to backport this to 4.4 and block ALL upgrade edges from 4.4 before this PR, which is not feasible. I will manually test whether my assessment is correct here

runcom:
> I think there's a small problem with our approach. Even if this is in a new 4.5 payload, existing 4.4 versions still cannot be upgraded because the MCD-firstboot service would have ran before this check. i.e. this codepath will not run in time for clusters that didn't already have an MCD on it before upgrading to 4.5

I'm xrefing this https://github.com/openshift/machine-config-operator/pull/1807#issue-432440615 as I think it's related and the reason why the MCD will have a chance to run and delete the file during an upgrade

yuqi-zhang:
/hold cancel

Actually, since during an upgrade to 4.5, the 4.5 MCD should have ran before the 4.5 machine-os-content gets booted, so it should be fine regardless. Sorry for the noise and thanks to @runcom for walking me through it :D 

(still will check manually though)

runcom:
let's keep the hold while @sinnykumari and @yuqi-zhang finish their testing to be extra sure about this fix (and thanks a lot!)

/hold

PR #1848, opened 2020-06-18
Add retry loop to nodeip-configuration service

cybertron:
We had a report that podman failed to pull the image, possibly because
the service came up before the interface that had access to the
registry, and as a result the service failed. This results in crio
and kubelet starting without an explicit IP set, which may cause
problems in some environments.

Ideally we would set Restart=on-failure in the unit file, but it looks
like the version of systemd we have right now is too old to support
that. Instead, add an infinite loop around the podman command so it
will retry indefinitely and prevent the deployment from continuing
without proper configuration. Once podman completes successfully it
will break the loop and end the oneshot.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/units/nodeip-configuration.service.yaml
+++ b/templates/common/baremetal/units/nodeip-configuration.service.yaml
@@ -13,14 +13,22 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/bin/podman run --rm \
+  # Would prefer to do Restart=on-failure instead of this bash retry loop, but
+  # the version of systemd we have right now doesn't support it. It should be
+  # available in systemd v244 and higher.
+  ExecStart=/bin/bash -c " \
+    until \
+    /usr/bin/podman run --rm \
     --authfile /var/lib/kubelet/config.json \
     --net=host \
     --volume /etc/systemd/system:/etc/systemd/system:z \
     {{ .Images.baremetalRuntimeCfgImage }} \
     node-ip \
     set --retry-on-failure \
-    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}
+    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}; \
+    do \
+    sleep 5; \
+    done"
 
   [Install]
   WantedBy=multi-user.target
--- a/templates/common/openstack/units/nodeip-configuration.service.yaml
+++ b/templates/common/openstack/units/nodeip-configuration.service.yaml
@@ -13,14 +13,22 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/bin/podman run --rm \
+  # Would prefer to do Restart=on-failure instead of this bash retry loop, but
+  # the version of systemd we have right now doesn't support it. It should be
+  # available in systemd v244 and higher.
+  ExecStart=/bin/bash -c " \
+    until \
+    /usr/bin/podman run --rm \
     --authfile /var/lib/kubelet/config.json \
     --volume /etc/systemd/system:/etc/systemd/system:z \
     --net=host \
     {{ .Images.baremetalRuntimeCfgImage }} \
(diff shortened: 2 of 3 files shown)


cybertron:
/hold

If we decide to go with this approach I need to apply it to the other platforms that use this service.

templates/common/baremetal/units/nodeip-configuration.service.yaml
```diff
@@ -13,14 +13,20 @@ contents: |
...
+    while :; do \
+    /usr/bin/podman run --rm \
     --authfile /var/lib/kubelet/config.json \
     --net=host \
     --volume /etc/systemd/system:/etc/systemd/system:z \
     {{ .Images.baremetalRuntimeCfgImage }} \
     node-ip \
     set --retry-on-failure \
-    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}
+    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} && break; \
```
crawford, line 28: Can you add a pause in here, just in case this is a rapid failure? Maybe something like:

```bash
until <command>
do
  sleep 5
done
```

cybertron: Yep, good idea. Done.

crawford:
Overall approach looks good to me.

cgwalters:
/approve

And note to self that this code also needs to be in https://github.com/openshift/machine-config-operator/pull/1766 - so we'll probably want to generate it from a shared helper instead of inline shell in the future.

cybertron:
I applied the logic to the other platforms. I was planning to wait until metal ci finished, but then I remembered it's still broken so it was going to fail anyway. :-/

celebdor:
I would rather the until only took care of pulling the image. So that, until systemd provides us with its own retry, we have the retry pulling in bash but the retry nodeip logic in the command.

That being said, this looks good to me.

cybertron:
I think in practice that's what this will end up doing. We still have --retry-on-failure for the command itself, so once it has started it should never go back to the bash loop.

templates/common/baremetal/units/nodeip-configuration.service.yaml
```diff
@@ -13,14 +13,22 @@ contents: |
...
     --authfile /var/lib/kubelet/config.json \
     --net=host \
     --volume /etc/systemd/system:/etc/systemd/system:z \
     {{ .Images.baremetalRuntimeCfgImage }} \
     node-ip \
     set --retry-on-failure \
-    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}
+    {{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}; \
+    do \
+    sleep 5; \
```
romfreiman, line 30: The service will be stuck in 'Active' state but the user (or whoever gonna troubleshoot) will have no idea that it's in a loop. Maybe to add log at least?
Another option is do podman pull in a loop with ExecStartPre phase - it will be more clear to the system - although there might be issues with failures, phase timeout and other systemd affects for oneshot services.

cybertron: Podman should be outputting an error message when it fails already. That's going to contain the information you need to debug the problem. I'm not sure adding a generic message in the service would gain us much.

templates/common/baremetal/units/nodeip-configuration.service.yaml
```diff
@@ -13,14 +13,22 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
-  ExecStart=/usr/bin/podman run --rm \
+  # Would prefer to do Restart=on-failure instead of this bash retry loop, but
```
mandre, line 16: `Restart=on-failure` has been in systemd forever, what feature do we need from systemd v244?

https://github.com/systemd/systemd/blob/master/NEWS

celebdor: Restart=on-failure was not available for `oneshot` until v244 IIRC, only for the other kinds of [Service]

PR #1850, opened 2020-06-19
Support installing extensions shipped by RHCOS

sinnykumari:
enhancement doc- https://github.com/openshift/enhancements/pull/317

- Save old and new MachineConfig into json file and process later on host
- Rework the existing PullAndRebase() to run extensions in host context
- Add m-c-d subcommand mount-container which we run in
  host context. Here we pull OSContainer, create a container and mount it.
  We also save the created container name and mount location in /run
  which we will be used later by MCD to rebase the OS and apply extensions
- Install kernel-rt packages from coreos-extensions repo we created earlier.
 This saves us from searching explicitly for kernel-rt packages in the OSContainer.
 Also simplifies updating kernel-rt packages

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -280,7 +280,7 @@ func run(_ *cobra.Command, args []string) (retErr error) {
 	}
 
 	if !changed {
-		glog.Info("No changes; already at target oscontainer, no kernel args provided")
+		glog.Info("No changes; already at target oscontainer, no kernel args provided, no kernelType switch, no extensions applied")
 	}
 
 	return nil
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/golang/glog"
 	daemon "github.com/openshift/machine-config-operator/pkg/daemon"
+	"github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	"github.com/openshift/machine-config-operator/pkg/daemon/pivot/types"
 	errors "github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -214,8 +215,10 @@ func run(_ *cobra.Command, args []string) (retErr error) {
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
-	var container string
+	var containerName string
 	if fromEtcPullSpec || len(args) == 0 {
+		// In this case we will just rebase to OSImage provided in etcPivotFile.
+		// Extensions won't apply. This should be consistent with old behavior.
 		fromEtcPullSpec = true
 		data, err := ioutil.ReadFile(etcPivotFile)
 		if err != nil {
@@ -224,14 +227,34 @@ func run(_ *cobra.Command, args []string) (retErr error) {
 			}
 			return errors.Wrapf(err, "failed to read from %s", etcPivotFile)
 		}
-		container = strings.TrimSpace(string(data))
+		container := strings.TrimSpace(string(data))
+		unitName := "mco-mount-container"
+		glog.Infof("Pulling in image and mounting container on host via systemd-run unit=%s", unitName)
+		err = exec.Command("systemd-run", "--wait", "--collect", "--unit="+unitName,
+			"-E", "RPMOSTREE_CLIENT_ID=mco", constants.HostSelfBinary, "mount-container", container).Run()
+		if err != nil {
+			return errors.Wrapf(err, "failed to create extensions repo")
+		}
+		var containerName string
+		if containerName, err = daemon.ReadFromFile(constants.MountedOSContainerName); err != nil {
+			return err
+		}
+
+		defer func() {
+			// Ideally other than MCD pivot, OSContainer shouldn't be needed by others.
+			// With above assumption, we should be able to delete OSContainer image as well as associated container with force option
+			exec.Command("podman", "rm", containerName).Run()
+			exec.Command("podman", "rmi", container).Run()
+			glog.Infof("Deleted container %s and corresponding image %s", containerName, container)
+		}()
+
 	} else {
-		container = args[0]
+		containerName = args[0]
 	}
 
 	client := daemon.NewNodeUpdaterClient()
 
-	_, changed, err := client.PullAndRebase(container, keep)
+	changed, err := client.PerformRpmOSTreeOperations(containerName, keep)
 	if err != nil {
 		return err
 	}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -46,7 +46,8 @@ const (
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsFile is the file to check if FIPS is enabled
-	fipsFile = "/proc/sys/crypto/fips_enabled"
+	fipsFile       = "/proc/sys/crypto/fips_enabled"
+	extensionsRepo = "/etc/yum.repos.d/coreos-extensions.repo"
 )
 
 func installedRTKernelRpmsOnHost() ([]string, error) {
@@ -255,6 +256,31 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
 	return true, nil
 }
 
+func writeMachineConfigs(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	oldMCFile, err := os.Create(constants.OldMachineConfigPath)
+	if err != nil {
+		return fmt.Errorf("Can't open file: %v", err)
+	}
+	defer oldMCFile.Close()
+
+	encoder := json.NewEncoder(oldMCFile)
+	if err := encoder.Encode(oldConfig); err != nil {
+		return fmt.Errorf("Error while encoding: %v", err)
+	}
+
+	newMCFile, err := os.Create(constants.NewMachineConfigPath)
+	if err != nil {
+		return fmt.Errorf("Can't open file: %v", err)
+	}
+	defer newMCFile.Close()
+
+	encoder = json.NewEncoder(newMCFile)
+	if err := encoder.Encode(newConfig); err != nil {
+		return fmt.Errorf("Error while encoding: %v", err)
+	}
+	return nil
+}
+
 // update the node to the provided node configuration.
 func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr error) {
 	oldConfig = canonicalizeEmptyMC(oldConfig)
@@ -297,6 +323,11 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 			dn.recorder.Eventf(mcRef, corev1.EventTypeWarning, "FailedToReconcile", wrappedErr.Error())
 		}
 		return errors.Wrapf(errUnreconcilable, "%v", wrappedErr)
+	}
+
+	// Serialize old and new MachineConfigs in file to be processed later by rpm-ostree during pivot
+	if err := writeMachineConfigs(oldConfig, newConfig); err != nil {
+		return fmt.Errorf("Error serialzing MachineConfigs on host: %v", err)
 	}
 
 	dn.logSystem("Starting update from %s to %s: %+v", oldConfigName, newConfigName, diff)
@@ -395,6 +426,7 @@ type machineConfigDiff struct {
 	files      bool
 	units      bool
 	kernelType bool
+	extensions bool
 }
 
 // isEmpty returns true if the machineConfigDiff has no changes, or
@@ -429,6 +461,7 @@ func newMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineC
 	// Both nil and empty slices are of zero length,
 	// consider them as equal while comparing KernelArguments in both MachineConfigs
 	kargsEmpty := len(oldConfig.Spec.KernelArguments) == 0 && len(newConfig.Spec.KernelArguments) == 0
+	extensionsEmpty := len(oldConfig.Spec.Extensions) == 0 && len(newConfig.Spec.Extensions) == 0
 
 	return &machineConfigDiff{
 		osUpdate:   oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL,
@@ -438,6 +471,7 @@ func newMachineConfigDiff(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineC
 		files:      !reflect.DeepEqual(oldIgn.Storage.Files, newIgn.Storage.Files),
 		units:      !reflect.DeepEqual(oldIgn.Systemd.Units, newIgn.Systemd.Units),
 		kernelType: canonicalizeKernelType(oldConfig.Spec.KernelType) != canonicalizeKernelType(newConfig.Spec.KernelType),
+		extensions: !(extensionsEmpty || reflect.DeepEqual(oldConfig.Spec.Extensions, newConfig.Spec.Extensions)),
 	}, nil
 }
 
@@ -698,8 +732,8 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 	return exec.Command("rpm-ostree", args...).Run()
 }
 
-// mountOSContainer mounts the container and returns the mountpoint
-func (dn *Daemon) mountOSContainer(container string) (mnt, containerName string, err error) {
+// MountOSContainer mounts the container and returns the mountpoint
+func MountOSContainer(container string) (mnt, containerName string, err error) {
 	var authArgs []string
 	if _, err = os.Stat(kubeletAuthFile); err == nil {
 		authArgs = append(authArgs, "--authfile", kubeletAuthFile)
@@ -728,6 +762,7 @@ func (dn *Daemon) mountOSContainer(container string) (mnt, containerName string,
 		return
 	}
 	mnt = strings.TrimSpace(string(mntBuf))
+	glog.Infof("Container ID,  mnt  and name %s   %s   %s\n", cid, mnt, containerName)
 	return
 }
 
@@ -771,7 +806,7 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
 
 	var mnt, containerName string
 	var err error
-	if mnt, containerName, err = dn.mountOSContainer(newConfig.Spec.OSImageURL); err != nil {
+	if mnt, containerName, err = MountOSContainer(newConfig.Spec.OSImageURL); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/daemon/update.go
```diff
@@ -792,6 +807,94 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	return nil
 }
 
+func addExtRepo(mnt string) error {
+	if err := os.MkdirAll(filepath.Dir(extensionsRepo), 0755); err != nil {
+		return fmt.Errorf("Error creating dir %s: %v", filepath.Dir(extensionsRepo), err)
```
kikisdeliveryservice, line 812: maybe "Error created extensions repo:.."

pkg/daemon/update.go
```diff
@@ -792,6 +807,94 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	}
+	repo, err := os.Create(extensionsRepo)
+	if err != nil {
+		repo.Close()
+		return err
```
kikisdeliveryservice, line 817: do we want this error wrapped with a message?

pkg/daemon/update.go
```diff
@@ -792,6 +807,94 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	}
+	defer repo.Close()
+	if _, err := repo.WriteString("[rhcos-extensions]\nenabled=1\nmetadata_expire=1m\nbaseurl=" + mnt + "/extensions/\ngpgcheck=0\nskip_if_unavailable=False\n"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
```
kikisdeliveryservice, line 826: would like a comment here just so it's obvious what this is

pkg/daemon/update.go
```diff
@@ -792,6 +807,94 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	}
+	for _, ext := range removed {
+		extArgs = append(extArgs, "--uninstall", ext)
+	}
+
+	return extArgs
+}
+
+// applyExtensions proesses specified extensions if it is supported on RHCOS.
+// It deletes an extension if it is no logner available in new rendered MachineConfig.
```
kikisdeliveryservice, line 862: love this comments!

pkg/daemon/update.go
sinnykumari: Fixed with "error creating yum repo directory" as we are creating standard `/etc/yum.repos.d` where rpm-ostree looks for repos.

pkg/daemon/update.go
sinnykumari: Elaborate error, don't think we need error to be wrapped. Should be easy to track from error message.

pkg/daemon/update.go
sinnykumari: Thanks for the suggestion!
Until this is feature complete, there is possibility of changes in the logic if needed. Will be adding comments at needed places once we cover all scenarios.

sinnykumari:
To install extensions on RHCOS host, I am pulling mounting OSContainer and then creating a repo in /etc/yum.repos.d/ which points to mounted OSContainer path (https://github.com/openshift/machine-config-operator/pull/1850/files#diff-06961b075f1753956d802ba954d2cfb5R820) . And then run `rpm-ostree update --install <pkg>` but it gives error:
```
Checking out tree ee22f82... done
Enabled rpm-md repositories: rhcos-extensions
Updating metadata for 'rhcos-extensions'... done
error: Updating rpm-md repo 'rhcos-extensions': /var/lib/containers/storage/overlay/<sha>/merged/extensions/ was not found
```

Looking at audit message on the applied host, I don't see any selinux denial message`grep -R "denied" /var/log/audit/audit.log` , perhaps selinux is not playing any role here. 

Do we need to implement something like https://github.com/coreos/rpm-ostree/pull/1732/ for `rpm-ostree install/update/remove` as well or there is any alternative way?

Hacky option I see is copy the extensions/ from mounted container into somewhere (like /var/tmp/) and delete after applying extensions.

@cgwalters @jlebon thoughts?

cgwalters:
OK I think the answer here is that currently for the ostree repo, what we're doing is opening a file descriptor and sending it to the daemon over DBus.  That will work regardless of any mount namespaces in effect.

When we invoked `podman mount` here it only affects the mount namespace of the MCD itself.

Possible fixes:

- Add runtime-only rpm-md repos (i.e. `/run/yum.repos.d`) and support referencing them by file descriptor (the best fix but also the hardest)
- Expose the repos via a temporary local webserver (kind of eww but would work short term)
- Invoke `podman mount` in the namespace of the host - and actually if we want to unroll this farther, once we have https://github.com/openshift/machine-config-operator/pull/1766/ we can just run most of this code in the context of the host instead of the MCD pod

jlebon:
> When we invoked `podman mount` here it only affects the mount namespace of the MCD itself.

Hmm, I'm confused. Doesn't the `rpm-ostree rebase` already run directly on the host via `machine-config-daemon-host.service`?

sinnykumari:
> OK I think the answer here is that currently for the ostree repo, what we're doing is opening a file descriptor and sending it to the daemon over DBus. That will work regardless of any mount namespaces in effect.
> 
> When we invoked `podman mount` here it only affects the mount namespace of the MCD itself.
> 
> Possible fixes:
> 
>     * Add runtime-only rpm-md repos (i.e. `/run/yum.repos.d`) and support referencing them by file descriptor (the best fix but also the hardest)
> 
>     * Expose the repos via a temporary local webserver (kind of eww but would work short term)
> 
>     * Invoke `podman mount` in the namespace of the host - and actually if we want to unroll this farther, once we have #1766 we can just run most of this code in the context of the host instead of the MCD pod

3rd option sound promising to me. I have limited knowledge, so I am going to ask some silly questions -
1. `podman mount` documentation says ` it mounts the specified containers' root file system in a location which can be accessed from the host, and returns its location` . `rpm-ostree`  binary which we will run is part of host and mounted container is accessible to host so shouldn't this already work?
2. For my info, how do we specify namespace during podman mount? Don't see anything in `podman mount` man page related to namespace.

cgwalters:
> Hmm, I'm confused. Doesn't the rpm-ostree rebase already run directly on the host via machine-config-daemon-host.service? (And that same service also does the podman create and podman mount dance.)

We have two copies of that code now, one which runs in the MCD and one which doesn't that was added for kernel-rt.

I think we need to deduplicate, and we can do so conveniently after https://github.com/openshift/machine-config-operator/pull/1766/ lands!

jlebon:
> We have two copies of that code now, one which runs on the host and one which runs in the MCD that was added for kernel-rt.

Ahh gotcha, I see it now. I was just grepping for `rebase` and following the trail from there.

Re. possible fixes suggested in https://github.com/openshift/machine-config-operator/pull/1850#issuecomment-649522828, my vote personally is for (2) short-term until (3).

pkg/daemon/update.go
```diff
@@ -46,7 +46,8 @@ const (
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsFile is the file to check if FIPS is enabled
-	fipsFile = "/proc/sys/crypto/fips_enabled"
+	fipsFile       = "/proc/sys/crypto/fips_enabled"
+	extensionsRepo = "/etc/yum.repos.d/rhcos-extensions.repo"
```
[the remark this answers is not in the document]
sinnykumari: yep, sound more generic. I wasn't sure whether OKD has plans to support extensions.

pkg/daemon/update.go
```diff
@@ -792,6 +807,94 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	return extArgs
+}
+
+// applyExtensions processes specified extensions if it is supported on RHCOS.
+// It deletes an extension if it is no logner available in new rendered MachineConfig.
+// Newly requested extension will be installed and existing extensions gets updated
+// if we have a new version of the extension available in machine-os-content.
+func (dn *Daemon) applyExtensions(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// We support installing extensions only on RHCOS nodes.
+	if dn.OperatingSystem != machineConfigDaemonOSRHCOS {
```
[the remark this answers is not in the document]
sinnykumari: should be doable, will keep this in mind. kernel-rt may make validations a bit tricky(unless we have plan to ship in OKD)

sinnykumari:
>     * Invoke `podman mount` in the namespace of the host - and actually if we want to unroll this farther, once we have #1766 we can just run most of this code in the context of the host instead of the MCD pod

Since #1766 is already in, I am going to give more thoughts and rework extensions feature + existing kernel-rt work to run in the context of host

cgwalters:
> Since #1766 is already in, I am going to give more thoughts and rework extensions feature + existing kernel-rt work to run in the context of host

Yeah I think having the kernel-rt logic run consistently on the host is really a pre-requisite for this.  I think actually today what we can do is serialize the `MachineConfig` object into JSON and pass it to ourself on the host and do the upgrade/kernel-rt/extensions stuff all there.

Hmm right I just realized the reason the kernel-rt code is working today is because the `rpm-ostree` client call opens file descriptors for each passed RPM and sends them to the daemon, but we can't do that with rpm-md repositories today.

Once we have https://bugzilla.redhat.com/show_bug.cgi?id=1839065 (which might even happen in 8.2.z) I think we can drop all of the logic for writing our binary to the host in the MCD case and just run everything from the MCD which would obviously be a huge cleanup.

sinnykumari:
> > Since #1766 is already in, I am going to give more thoughts and rework extensions feature + existing kernel-rt work to run in the context of host
> 
> Yeah I think having the kernel-rt logic run consistently on the host is really a pre-requisite for this. I think actually today what we can do is serialize the `MachineConfig` object into JSON and pass it to ourself on the host and do the upgrade/kernel-rt/extensions stuff all there.

+1

I was also thinking about early validation of extensions args i.e. when we render the applied MachineConfig. I believe this would require another podman mount of OSContainer to look into available extensions :/


> Hmm right I just realized the reason the kernel-rt code is working today is because the `rpm-ostree` client call opens file descriptors for each passed RPM and sends them to the daemon, but we can't do that with rpm-md repositories today.

right

> Once we have https://bugzilla.redhat.com/show_bug.cgi?id=1839065 (which might even happen in 8.2.z) I think we can drop all of the logic for writing our binary to the host in the MCD case and just run everything from the MCD which would obviously be a huge cleanup.

that will be nice. If I understand correctly we will still need to pull m-c-d binary during firstboot, right?

cgwalters:
> I was also thinking about early validation of extensions args i.e. when we render the applied MachineConfig. I believe this would require another podman mount of OSContainer to look into available extensions :/

Oh yeah, that's a good point.  Hmm.  Probably the MCC container could pull the oscontainer once when it starts up and then and cache a mapping of `oscontainer url => allowed extensions`, so we're not re-pulling the container each time we validate.

Or I guess we could just maintain a static list of allowed extensions in the MCO too.  There shouldn't be too many to start.

>  If I understand correctly we will still need to pull m-c-d binary during firstboot, right?

Yeah, because we don't support really ancient kubelets joining the cluster, we don't want any workloads to land before OS updates etc.

pkg/daemon/update.go
kikisdeliveryservice: sounds perf!

miabbott:
> Oh yeah, that's a good point. Hmm. Probably the MCC container could pull the oscontainer once when it starts up and then and cache a mapping of oscontainer url => allowed extensions, so we're not re-pulling the container each time we validate.

Maybe we should be including a JSON file or the like in machine-os-content that provides the list of supported extensions?  Since we are responsible for sticking the right RPMs and deps into the container, seems like we are in the best position to enumerate the supported extensions per oscontainer/release.

Could even go as far as providing a supported and unsupported list, if we were to ever decide to drop support for an extension.

sinnykumari:
> > Oh yeah, that's a good point. Hmm. Probably the MCC container could pull the oscontainer once when it starts up and then and cache a mapping of oscontainer url => allowed extensions, so we're not re-pulling the container each time we validate.
> 
> Maybe we should be including a JSON file or the like in machine-os-content that provides the list of supported extensions? Since we are responsible for sticking the right RPMs and deps into the container, seems like we are in the best position to enumerate the supported extensions per oscontainer/release.
> 
> Could even go as far as providing a supported and unsupported list, if we were to ever decide to drop support for an extension.

agree, if we want to extend this to supported/unsupported list then machine-os-content seems to be the right place to provide that detail.

For now I think it should be fine for MCO to consume extensions as it is because we ship only those extra packages in machine-os-content which we want users to install. https://github.com/openshift/machine-config-operator/pull/1766 makes things really flexible to update MCD behavior whenever neededd

cgwalters:
The new v2 layout https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/952 aims to have "enumerate extensions" simply use the filesystem.  Basically:

- List all sub-directories of `extensions/`
- Skip `dependencies/`
- If the directory has any `.rpm` files in it, it's an extension

Not opposed to a JSON list, but keep in mind the rpm-md repodata is already a "list of packages" (just in XML form).  We'd be inventing "rpm metadata in JSON form"?

Anything we change inside the container though isn't solving the issue that validating a MachineConfig fragment requires fetching and unpacking the container.

miabbott:
>     * If the directory has any `.rpm` files in it, it's an extension

I missed this association when that MR went by.  That kind of clear relationship is what I was thinking about when I suggested the JSON file approach.

jlebon:
> Not opposed to a JSON list, but keep in mind the rpm-md repodata is already a "list of packages" (just in XML form). We'd be inventing "rpm metadata in JSON form"?

I think this is fine to start, though IMO there's definitely an argument for having the flexibility that some metadata layer affords us. For example, if we want extension "foobar" to actually result in the installation of two related, but not necessarily dependent packages. Or if we want to change what extension "foobar" means across an update.

I think actually this might help with the [Recommends](https://github.com/coreos/rpm-ostree/issues/718) issue as well. If we solve that and turn it off by default both server-side and client-side, then we'll need a way to pull in recommended packages which no longer get pulled in when it makes sense (and without globally turning on weak deps again client-side).

miabbott:
I wanted to revive the discussion about providing some kind of metadata about available extensions in `machine-os-content`

In a real-time meeting with @zvonkok @mrunalp @darkmuggle (re: https://github.com/openshift/enhancements/pull/357) we discussed adding additional annotations to the `machine-os-content` about the kernel version (*and possibly the glibc version) used by RHCOS.  This would help the users doing out-of-tree module builds by providing the information in the metadata, rather than having to pull down the entire container image.

It was brought up that we could go a step further and also enumerate the available extensions in an annotation on the `machine-os-content`.  This could provide similar benefits to allowing users (or tooling) to inspect the metadata of `machine-os-content` to determine which extensions are supported/available in a particular image.

sinnykumari:
> I wanted to revive the discussion about providing some kind of metadata about available extensions in `machine-os-content`

This sounds like it needs more thoughts and discussion. To avoid getting this lost in this lengthy PR update, I think creating a new issue with relevant information would be nice.

miabbott:
> This sounds like it needs more thoughts and discussion. To avoid getting this lost in this lengthy PR update, I think creating a new issue with relevant information would be nice.

Moved to https://github.com/openshift/os/issues/409

sinnykumari:
In the latest PR, I have moved kernel-rt and extensions processing into host context. It installs needed packages from coreos-extensions repo which we again mount in the host context so that it is accessible to rpm-ostree.

It seems rpm-ostree still has some issues accessing the extensions repo path which points to already mounted OSContainer. Journal log of rpm-ostreed service while installing usbguard extensions at cluster install time -

```
$ journalctl -u rpm-ostreed.service
-- Logs begin at Thu 2020-07-16 03:13:57 UTC, end at Thu 2020-07-16 07:21:34 UTC. --
Jul 16 03:22:48 ip-10-0-163-33 systemd[1]: Starting rpm-ostree System Management Daemon...
Jul 16 03:22:48 ip-10-0-163-33 rpm-ostree[1967]: Reading config file '/etc/rpm-ostreed.conf'
Jul 16 03:22:49 ip-10-0-163-33 rpm-ostree[1967]: In idle state; will auto-exit in 63 seconds
Jul 16 03:22:49 ip-10-0-163-33 systemd[1]: Started rpm-ostree System Management Daemon.
Jul 16 03:22:49 ip-10-0-163-33 rpm-ostree[1967]: client(id:cli dbus:1.14 unit:machine-config-daemon-firstboot.service uid:0) added; new total=1
Jul 16 03:22:49 ip-10-0-163-33 rpm-ostree[1967]: client(id:cli dbus:1.14 unit:machine-config-daemon-firstboot.service uid:0) vanished; remaining=0
Jul 16 03:22:49 ip-10-0-163-33 rpm-ostree[1967]: In idle state; will auto-exit in 62 seconds
Jul 16 03:23:21 ip-10-0-163-33 rpm-ostree[1967]: client(id:cli dbus:1.19 unit:machine-config-daemon-firstboot.service uid:0) added; new total=1
Jul 16 03:23:21 ip-10-0-163-33 rpm-ostree[1967]: Initiated txn UpdateDeployment for client(id:cli dbus:1.19 unit:machine-config-daemon-firstboot.service uid:0): /org/projectatomic/rpmostree1/rhcos
Jul 16 03:23:41 ip-10-0-163-33 rpm-ostree[1967]: Librepo version: 1.11.0 with CURL_GLOBAL_ACK_EINTR support (libcurl/7.61.1 OpenSSL/1.1.1c zlib/1.2.11 brotli/1.0.6 libidn2/2.2.0 libpsl/0.20.2 (+libidn2/2.0.5) libssh/0.9.0/openssl/zlib nghttp2/1.33.0)
----> Jul 16 03:23:42 ip-10-0-163-33 rpm-ostree[1967]: Txn UpdateDeployment on /org/projectatomic/rpmostree1/rhcos failed: Updating rpm-md repo 'coreos-extensions': /var/lib/containers/storage/overlay/d73cc16e4b5906d0ce2bc67544cc744370bee8c9300d2b172245b985476644fe/merged/extensions/ was not found
Jul 16 03:23:42 ip-10-0-163-33 rpm-ostree[1967]: client(id:cli dbus:1.19 unit:machine-config-daemon-firstboot.service uid:0) vanished; remaining=0
Jul 16 03:23:42 ip-10-0-163-33 rpm-ostree[1967]: In idle state; will auto-exit in 63 seconds
Jul 16 03:24:45 ip-10-0-163-33 rpm-ostree[1967]: In idle state; will auto-exit in 62 seconds
```

Above issue has been seen while using latest installer where bootimage contains OSTree version 46.82.202007152240-0, rpm-ostree-2020.2-2.el8.x86_64, ostree-2020.3-3.el8.x86_64)

Before moving to latest installer, I have been using older installer where bootimage contains OSTree version 45.81.202005200134-0, rpm-ostree-2019.6-8.el8.x86_64, ostree-2019.6-2.el8.x86_64 . With this OS update, extensions apply and kernel-rt switch works fine.

Don't see any selinux denial in audit log. @cgwalters @jlebon Can it be related to some recent rpm-ostree update?

sinnykumari:
Had a chat with Colin and Jonathan about recent issue https://github.com/openshift/machine-config-operator/pull/1850#issuecomment-659228078.

Issue is happening because:
- Mounted OSContainer container is private
```$ findmnt -o TARGET,PROPAGATION | grep var/lib/containers
| `-/var/lib/containers/storage/overlay                                                                           private
|   `-/var/lib/containers/storage/overlay/0fe3227f6f6f10dae3a43ae0ba951c8d94097d3fb2f2ee3d21e396d136fe3d1b/merged private
```
-  rpm-ostreed isn't seeing it because it's running in its own mount namespace due to MountFlags=slave which was introduced recently https://github.com/coreos/rpm-ostree/commit/75c676715ae5c76f35602e3533bf4c084791babd

Solution:
1. It should work fine if rpm-ostreed.service is started after the mount. We can stop rpm-ostreed service, mount container and then start rpm-ostreed.
2. Extract the container content using `oc image extract` to fetch container content and perform OS update and extensions from there. Related https://hackmd.io/WeqiDWMAQP2sNtuPRul9QA

We will explore option 2 first and if it doesn't work we will go for option 1

Until we fix this
/hold

sinnykumari:
While going down the road of `oc extract image`, it was considerable deviation to this PR. Have reworked in a fresh PR https://github.com/openshift/machine-config-operator/pull/1941 which should supersede this one.

We can close this PR later on.

cmd/machine-config-daemon/pivot.go
```diff
@@ -224,14 +227,34 @@ func run(_ *cobra.Command, args []string) (retErr error) {
...
+		glog.Infof("Pulling in image and mounting container on host via systemd-run unit=%s", unitName)
+		err = exec.Command("systemd-run", "--wait", "--collect", "--unit="+unitName,
+			"-E", "RPMOSTREE_CLIENT_ID=mco", constants.HostSelfBinary, "mount-container", container).Run()
+		if err != nil {
+			return errors.Wrapf(err, "failed to create extensions repo")
+		}
+		var containerName string
+		if containerName, err = daemon.ReadFromFile(constants.MountedOSContainerName); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
LorbusChris: meh.. I just saw there's a new PR :smile:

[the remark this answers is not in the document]
sinnykumari: will close this once we merge the other one.

sinnykumari:
Alternative implementation of extensions in https://github.com/openshift/machine-config-operator/pull/1941 has been merged, closing this PR.

PR #1851, opened 2020-06-19
Minimal DHCP fix to unblock baremetal ci

cybertron:
The full fix for all platforms is #1840, but as this is completely
blocking baremetal ci and dev right now and we're having ci issues
with the other platforms on the full fix, this is just the bit that
is needed to get baremetal working again. We'll still want to get
1840 in eventually, but that will be less critical if we merge this
in the meantime.

### DIFF
--- a/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
+++ b/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
@@ -3,7 +3,7 @@ path: "/etc/NetworkManager/conf.d/99-kni.conf"
 contents:
   inline: |
     [main]
-    dhcp=dhclient
     rc-manager=unmanaged
     [connection]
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
/approve
/lgtm

Do you need to remove `templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml` and `templates/master/00-master/baremetal/files/dhcp-dhclient-conf.yaml` as well in this? Or would those just not be used?

cybertron:
> Do you need to remove `templates/worker/00-worker/baremetal/files/dhcp-dhclient-conf.yaml` and `templates/master/00-master/baremetal/files/dhcp-dhclient-conf.yaml` as well in this? Or would those just not be used?

They just won't be used since we're not using dhclient anymore. It should be fine for them to be there until they get cleaned up in #1840.

PR #1855, opened 2020-06-20
Bug 1802880: Update to github.com/mtrmac/gpgme v0.1.2 [4.3]

mtrmac:
**- What I did**

Performed by
```console
$ GOPROXY=https://proxy.golang.org GO111MODULE=on go get github.com/mtrmac/gpgme@v0.1.2 && make go-deps
```
in a `golang:1.12` container.

**- How to verify it**

N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7
 	github.com/joho/godotenv v1.3.0
+	github.com/mtrmac/gpgme v0.1.2 // indirect
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/openshift/api v3.9.1-0.20191111211345-a27ff30ebf09+incompatible
 	github.com/openshift/client-go v0.0.0-20191001081553-3b0e988f8cb0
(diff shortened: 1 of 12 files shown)


mtrmac:
Compare https://github.com/openshift/machine-config-operator/pull/1519 for the master branch.

runcom:
I think the 4.2 backport isn't necessary - we know this cve doesn't impact MCO, we should just close it for 4.2/4.3 IMO

kikisdeliveryservice:
> I think the 4.2 backport isn't necessary - we know this cve doesn't impact MCO, we should just close it for 4.2/4.3 IMO

@runcom I'm down :+1: 

@mtrmac WDYT? can we close?

mtrmac:
The cost/benefit decision of dealing with this PR and the backport process vs. the minimal risk of introducing new users of that code is entirely up to repo maintainers who will be paying the cost; I’m fine with whatever you decide.

runcom:
> The cost/benefit decision of dealing with this PR and the backport process vs. the minimal risk of introducing new users of that code is entirely up to repo maintainers who will be paying the cost; I’m fine with whatever you decide.

awesome, so what we know is that bumping this library and introducing more lines of code doesn't buy anything as we're not impacted by the CVE, so let's close this for 4.3

runcom:
The only issue is automatic tools that look for CVEs with static analysis but we can't help with that - we're again not impacted by that CVE

PR #1856, opened 2020-06-20
Bug 1802883: Update to github.com/mtrmac/gpgme v0.1.2 [4.2]

mtrmac:
**- What I did**

Performed by updating `Gopkg.toml` and
```console
$ dep ensure
```

**- How to verify it**

N/A

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -432,12 +432,12 @@
   version = "1.0.1"
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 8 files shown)


mtrmac:
Compare #1519 for the master branch.

runcom:
I think the 4.2 backport isn't necessary - we know this cve doesn't impact MCO, we should just close it for 4.2/4.3 IMO

mtrmac:
Closing per the discussion in #1855 .

PR #1857, opened 2020-06-22
pkg/daemon: use Golang idioms and bubble up errors

runcom:
One of the fundamental flaw of the MCO is its inability to properly
bubble up errors to the node object and pools (and later, cluster
operator).

One of the code flaw that we have is around Go idioms. The functions
I've modified here were just logging the error and returning a boolean.
That makes thing very tricky when debugging as there's no other way to
read those errors other than asking for a full must-gather...

This patch converts that code to properly return errors and bubble up
any. The only thing we don't report is the diff when file content
mismatch which is ok for debugging purposes at least we can immediately
check "which" file is diffing (and later check what's diffing with
must-gather if needed, at least this helps us debug and triage duplicate
bugs if possible).

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1049,8 +1049,8 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		expectedConfig = state.currentConfig
 	}
 	if _, err := os.Stat(constants.MachineConfigDaemonForceFile); err != nil {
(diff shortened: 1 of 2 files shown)


runcom:
need to fix unit I guess

/hold

runcom:
dropping the hold, this is ok now to get in

/hold cancel

PR #1858, opened 2020-06-22
pkg/operator: report pools statuses

runcom:
This func is called only from within `setOperatorStatusExtension` which
in turns is called in some other operator routines. The thing is, this
doesn't work, the `Status.Extension` field of the CO is never set and
I'm looking at why as we speak. Besides that, this function should
report pools statuses regardless of the correctness of the pool
(controller version mismatch anyone? :troll:).
This will ensure, once we fix the CO issue, that we'll always have
something to immediately look at when looking at the MCO cluster
operator object.

This is meant to be the "improve controller version mismatch errors" as
from now on, we'll have users differentiate on the exact error by
looking at the CO.Status.Extension field (which, as you can see from the
code, reports the pools statuses, including any MCD error, see https://github.com/openshift/machine-config-operator/pull/1857 also).

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -17,7 +17,6 @@ import (
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
-	"github.com/openshift/machine-config-operator/pkg/version"
 )
 
 // syncVersion handles reporting the version to the clusteroperator
@@ -332,13 +331,7 @@ func (optr *Operator) allMachineConfigPoolStatus() (map[string]string, error) {
 	}
 	ret := map[string]string{}
 	for _, pool := range pools {
-		p := pool.DeepCopy()
-		err := isMachineConfigPoolConfigurationValid(p, version.Hash, optr.mcLister.Get)
-		if err != nil {
-			glog.V(4).Infof("Skipping status for pool %s because %v", p.GetName(), err)
-			continue
-		}
-		ret[p.GetName()] = machineConfigPoolStatus(p)
+		ret[pool.GetName()] = machineConfigPoolStatus(pool)
 	}
 	return ret, nil
 }
@@ -347,7 +340,7 @@ func (optr *Operator) allMachineConfigPoolStatus() (map[string]string, error) {
 func isMachineConfigPoolConfigurationValid(pool *mcfgv1.MachineConfigPool, version string, machineConfigGetter func(string) (*mcfgv1.MachineConfig, error)) error {
 	// both .status.configuration.name and .status.configuration.source must be set.
 	if pool.Spec.Configuration.Name == "" {
-		return fmt.Errorf("configuration spec for pool %s is empty", pool.GetName())
+		return fmt.Errorf("configuration spec for pool %s is empty: %v", pool.GetName(), machineConfigPoolStatus(pool))
 	}
 	if pool.Status.Configuration.Name == "" {
 		// if status is empty, it means the node controller hasn't seen any node at the target configuration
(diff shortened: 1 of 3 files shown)


ashcrow:
Odd:
```
 error: build error: running 'make install DESTDIR=./instroot && tar -C instroot -cf instroot.tar .' failed with exit code 2
2020/06/22 13:12:46 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/06/22 13:12:47 Ran for 1m27s
error: some steps failed:
  * could not run steps: step machine-config-operator failed: could not wait for build: the build machine-config-operator failed after 57s with reason DockerBuildFailed: Docker build strategy has failed. 
```

ashcrow:
Ah!
```
 pkg/operator/status.go:20:2: imported and not used: "github.com/openshift/machine-config-operator/pkg/version" 
```

runcom:
build error on my side :man_shrugging:

test/e2e/mco_test.go
```diff
@@ -92,3 +94,26 @@ func waitForAllMastersUpdate(cs *framework.ClientSet, mastersSchedulable bool) e
 		return true, nil
 	})
 }
+
+func TestClusterOperatorStatusExtension(t *testing.T) {
```
runcom, line 98: this was broken in master/4.5/4.4 until https://github.com/openshift/cluster-config-operator/pull/136 and since there was no gating on PRs breaking the Status.Extension field, this test will do it since we rely on that field to be set when something fails in the MCO

yuqi-zhang: To make sure, if someone else breaks the extension field, it'll show up in our e2e test as a failure? I feel like we should be gating it somewhere else as well

pkg/operator/status.go
```diff
@@ -374,13 +367,13 @@ func isMachineConfigPoolConfigurationValid(pool *mcfgv1.MachineConfigPool, versi
...
 		if !ok && pool.Status.Configuration.Name == mcName {
-			return fmt.Errorf("%s must be created by controller version %s", mcName, version)
+			return fmt.Errorf("%s must be created by controller version %s: %v", mcName, version, machineConfigPoolStatus(pool))
 		}
 		// user provided MC fragments do not have the annotation, so we just skip the version check there.
 		// The check below is for: 1) the generated MC for the pool, and 2) the bootstrapped fragments
 		// that do have this annotation set with a version.
 		if ok && v != version {
-			return fmt.Errorf("controller version mismatch for %s expected %s has %s", mcName, version, v)
+			return fmt.Errorf("controller version mismatch for %s expected %s has %s: %v", mcName, version, v, machineConfigPoolStatus(pool))
```
yuqi-zhang, line 376: Clarification question, you say we're not reporting the error up because the extension field is broken, once that's fixed, does the pool status here report the same error as before?

test/e2e/mco_test.go
runcom: yep, working on having a proper test but still under discussion - this is the fastest way to get something in - what we care anyway is about having this field work so whoever breaks it, it breaks us so it kind of make sense that we test our logic (we're not testing the field per-se)

pkg/operator/status.go
runcom: it'll have the additional `, machineConfigPoolStatus(pool)` which is also what we expose in the Extension field - kind of redundant maybe. So right now w/o this fix, it doesn't work because extension wasn't working, now with this fix, besides extension working, we'll always have the pool status in the MCO degraded reason

runcom: > you say we're not reporting the error up because the extension field is broken,

also, not fully, see #1857 for mcd errors

yuqi-zhang: Yeah, ok, I'm fine with the redundancy, I just wanted to check if my understanding is correct, thanks!

PR #1860, opened 2020-06-22
Configure OVS NIC for OVN

trozet:
OVN-k8s needs to be able to move the primary NIC of the host into the
OVS bridge, making the physical NIC essentially a Layer 2 port. This
patch uses nmcli to configure OVS with the NIC on it, as well as bring
up the Layer 3 OVS interface. The configuration can be auto-detected
based on the existing default gateway interface or NM configuration key
files may be provided during ignition and placed on the host.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -1,0 +1,116 @@
+mode: 0755
+path: "/usr/local/bin/configure-ovs.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    set -eux
+
+    # Configures NICs onto OVS bridge "br-ex"
+    # Configuration is either auto-detected or provided through a config file written already in Network Manager
+    # key files under /etc/NetworkManager/system-connections/
+    # Managing key files is outside of the scope of this script
+
+    iface=""
+    counter=0
+    # find default interface
+    while [ $counter -lt 12 ]; do
+      iface=$(ip -j route show default | jq -r '.[0].dev')
+      if [ -n "$iface" ]; then
+        echo "Default gateway interface found: ${iface}"
+        break
+      fi
+      counter=$((counter+1))
+      echo "No default route found on attempt: ${counter}"
+      sleep 5
+    done
+
+    if [ "$iface" = "br-ex" ]; then
+      echo "Networking already configured and up for br-ex!"
+      exit 0
+    fi
+
+    if [ -z "$iface" ]; then
+      echo "ERROR: Unable to find default gateway interface"
+      exit 1
+    fi
+
+    # find the MAC from OVS config or the default interface to use for OVS internal port
+    # this prevents us from getting a different DHCP lease and dropping connection
+    if ! iface_mac=$(<"/sys/class/net/${iface}/address"); then
+      echo "Unable to determine default interface MAC"
+      exit 1
+    fi
+
+    echo "MAC address found for iface: ${iface}: ${iface_mac}"
+
+    # find MTU from original iface
+    iface_mtu=$(ip -j link show "$iface" | jq -r '.[0].mtu')
+    if [ -z "$iface_mtu" ]; then
+      echo "Unable to determine default interface MTU, defaulting to 1500"
+      iface_mtu=1500
+    else
+      echo "MTU found for iface: ${iface}: ${iface_mtu}"
+    fi
+
+    # create bridge
+    if ! nmcli connection show br-ex &> /dev/null; then
+      nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+    fi
+
+    # store old conn for bringing down later
+    old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
+
+    # find default port to add to bridge
+    if ! nmcli connection show ovs-port-phys0 &> /dev/null; then
+      nmcli c add type ovs-port conn.interface ${iface} master br-ex con-name ovs-port-phys0
+    fi
+
+    if ! nmcli connection show ovs-port-br-ex &> /dev/null; then
+      nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
+    fi
+
+    # bring down any old iface
+    nmcli conn down $old_conn
+
+    if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
+      nmcli c add type 802-3-ethernet conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
+        connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu}
+    fi
+
+    if ! nmcli connection show ovs-if-br-ex &> /dev/null; then
+      nmcli c add type ovs-interface slave-type ovs-port conn.interface br-ex master ovs-port-br-ex con-name \
+        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+    fi
+
+    # wait for DHCP to finish, verify connection is up
+    counter=0
+    while [ $counter -lt 5 ]; do
+      sleep 5
+      # check if connection is active
+      if nmcli --fields GENERAL.STATE conn show ovs-if-br-ex | grep -i "activated"; then
+        echo "OVS successfully configured"
+        ip a show br-ex
+        exit 0
+      fi
+      counter=$((counter+1))
+    done
+
+    echo "WARN: OVS did not succesfully activate NM connection. Attempting to bring up connections"
+    counter=0
+    while [ $counter -lt 5 ]; do
+      if nmcli conn up ovs-if-br-ex; then
+        echo "OVS successfully configured"
+        ip a show br-ex
+        exit 0
+      fi
+      sleep 5
+      counter=$((counter+1))
+    done
+
+    echo "ERROR: Failed to activate ovs-if-br-ex NM connection"
+    # if we made it here networking isnt coming up, revert for debugging
+    set +e
+    nmcli conn down ovs-if-br-ex
+    nmcli conn down ovs-if-phys0
+    nmcli conn up $old_conn
+    exit 1
--- a/templates/common/_base/units/ovs-configuration.service.yaml
+++ b/templates/common/_base/units/ovs-configuration.service.yaml
@@ -1,0 +1,19 @@
+name: ovs-configuration.service
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else}}false{{end}}
+contents: |
+  [Unit]
+  Description=Configures OVS with proper host networking configuration
+  # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
+  Requires=NetworkManager-wait-online.service openvswitch.service
+  After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
+  Before=network-online.target kubelet.service crio.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  ExecStart=/usr/local/bin/configure-ovs.sh
+  StandardOutput=journal+console
+  StandardError=journal+console
+
+  [Install]
+  WantedBy=network-online.target
--- a/templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
+++ b/templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
@@ -5,7 +5,8 @@ contents: |
   Description=Update GCP routes for forwarded IPs.
   ConditionKernelCommandLine=|ignition.platform.id=gce
   ConditionKernelCommandLine=|ignition.platform.id=gcp
-  After=network.target
+  Wants=network-online.target
+  After=network-online.target
 
   [Service]
   Type=simple


trozet:
@runcom FYI

trozet:
@dcbw fyi

kikisdeliveryservice:
Is there any documentation/summary about this work? Jira? BZ? Enhancement?

trozet:
> Is there any documentation/summary about this work? Jira? BZ? Enhancement?

https://issues.redhat.com/browse/SDN-1030 and more specifically https://issues.redhat.com/browse/SDN-1032

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,62 @@
+mode: 0755
+path: "/usr/local/bin/configure-ovs.sh"
+contents:
+  inline: |
+    #!/usr/bin/env bash
```
cgwalters, line 5: BTW after https://github.com/openshift/machine-config-operator/pull/1766 lands, it will become possible to add Go code into the MCD binary itself and have it execute on the host reliably.

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,18 @@
+name: ovs-configuration.service
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else}}false{{end}}
+contents: |
+  [Unit]
+  Description=Configures OVS with proper host networking configuration
+  # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
+  Wants=network-online.target
+  After=openvswitch.service network-online.target ignition-firstboot-complete.service
```
cgwalters, line 8: Drop the `ignition-firstboot-complete.service` here; you want `machine-config-daemon-firstboot.service` now probably.

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,18 @@
...
+  Before=kubelet.service crio.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  ExecStart=/usr/local/bin/configure-ovs.sh
+  StandardOutput=journal
+
+  [Install]
+  WantedBy=multi-user.target
```
cgwalters, line 18: In other units more recently we made this e.g.
`RequiredBy=kubelet.service`
if it *must* run before kubelet.

cgwalters:
> https://issues.redhat.com/browse/SDN-1030 and more specifically https://issues.redhat.com/browse/SDN-1032

This is the problem with using a private JIRA instance to track work on public Free/Open Source software...

templates/common/_base/units/openvswitch.service
```diff
@@ -0,0 +1,2 @@
+name: "openvswitch.service"
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
```
cgwalters, line 2: I think you can drop this and instead just have:
```
After=openvswitch.service
Requires=openvswitch.service
```
in `ovs-configuration.service` - then it will pull it in as a dependency only if that service itself is enabled.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,62 @@
...
+    if ! nmcli connection show ovs-port-br-ex; then
+      nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex connection.autoconnect no
+    fi
+
+    iface=""
+    counter=0
+    # find default interface
+    while [ $counter -lt 12 ]; do
+      if ip route | grep default; then
+        iface=$(ip route |grep default | grep -Eo "dev\s[^ ]+" | sed 's/dev\s//')
```
miabbott, line 21: In the event there are multiple interfaces connected to the same network, you could end up with multiple entries for the default gw.  Using my laptop as an example:

```
$ ip route | grep default
default via 192.168.1.1 dev enp0s31f6 proto dhcp metric 100 
default via 192.168.1.1 dev wlp58s0 proto dhcp metric 600 
$ ip route |grep default | grep -Eo "dev\s[^ ]+" | sed 's/dev\s//'
enp0s31f6
wlp58s0
```

miabbott: Probably want to pick the the interface with the lower metric after getting the ifaces

templates/common/_base/files/configure-ovs-network.yaml
trozet: thanks. I had planned on moving to go after I got the bash working. I didn't know about this option.

templates/common/_base/files/configure-ovs-network.yaml
trozet: nice catch, thanks

trozet:
I think I'm going to take a different approach to this PR and try to configure all of the NICs before networking comes up.

templates/common/_base/units/openvswitch.service
trozet: that's part of @runcom patch not mine, it was already merged

kikisdeliveryservice:
> /test e2e-ovn-step-registry

Is this PR intended to fix the ovn-step-registry job? Bc it has a very spotty passage rate.

trozet:
> > /test e2e-ovn-step-registry
> 
> Is this PR intended to fix the ovn-step-registry job? Bc it has a very spotty passage rate.

No, this PR is not intended to fix anything. If you have some links to other jobs where it is failing @dcbw or I can take a look.

kikisdeliveryservice:
> > > /test e2e-ovn-step-registry
> > 
> > 
> > Is this PR intended to fix the ovn-step-registry job? Bc it has a very spotty passage rate.
> 
> No, this PR is not intended to fix anything. If you have some links to other jobs where it is failing @dcbw or I can take a look.

I just looked in the job history for this e2e: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-machine-config-operator-master-e2e-ovn-step-registry

templates/common/_base/files/configure-ovs-network.yaml
trozet: looks like the routes are always sorted by lowest metric so I can just grab the first 1.
[root@test-q9gz9-bootstrap ~]# ip route
default via 192.168.126.1 dev br-ex 
default via 127.0.0.2 dev lo metric 50 
default via 192.168.126.1 dev br-ex metric 200 
default via 192.168.126.1 dev br-ex proto dhcp metric 800 
default via 192.168.126.1 dev br-ex metric 1299 
default via 127.0.0.2 dev lo metric 2000

trozet:
There is still some issue where in some deployments CNO detects not all nodes are up:
Failed to reconcile platform networking resources: Unable to bootstrap OVN, expected amount of control plane nodes (3) do not match found (1): timed out waiting for the condition

Trying to determine the root cause of that so holding this for now.

trozet:
> There is still some issue where in some deployments CNO detects not all nodes are up:
> Failed to reconcile platform networking resources: Unable to bootstrap OVN, expected amount of control plane nodes (3) do not match found (1): timed out waiting for the condition
> 
> Trying to determine the root cause of that so holding this for now.

@alexanderConstantinescu @dcbw 
I finally was able to reproduce the problem with my own AWS deployment. The nodes do indeed lose network connectivity. I was able to finally get into one of the machines and get the journal. The problem is sometimes the DHCP will fail on br-ex when it's first coming online:

Jul 03 05:14:42 localhost.localdomain NetworkManager[1361]: <info>  [1593753282.4301] device (br-ex): Activation: connection 'ovs-if-br-ex' enslaved, continuing activation
Jul 03 05:14:42 localhost.localdomain NetworkManager[1361]: <info>  [1593753282.4373] dhcp4 (br-ex): activation: beginning transaction (timeout in 45 seconds)
Jul 03 05:15:27 localhost.localdomain NetworkManager[1361]: <warn>  [1593753327.3774] dhcp4 (br-ex): request timed out
Jul 03 05:15:27 localhost.localdomain NetworkManager[1361]: <info>  [1593753327.3776] dhcp4 (br-ex): state changed unknown -> timeout
Jul 03 05:15:27 localhost.localdomain NetworkManager[1361]: <info>  [1593753327.3776] device (br-ex): state change: ip-config -> failed (reason 'ip-config-unavailable', sys-iface-state: 'managed')
Jul 03 05:15:27 localhost.localdomain NetworkManager[1361]: <info>  [1593753327.3785] device (br-ex): releasing ovs interface br-ex

This is not a problem, because there is logic to retry the DHCP in case the wrong mac is used or something else fails. However, the nmcli conn up command blocks until the DHCP fails and gets back an error code, which fails the script. At that point all is lost. I think this latest patch should fix things.

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  ExecStart=/usr/local/bin/configure-ovs.sh
+  StandardOutput=journal+console
+  StandardError=journal+console
+
+  [Install]
+  WantedBy=network-online.target
```
trozet, line 19: debating making this RequiredBy

trozet:
Looks like there is another case where networking does not come up correctly. This time it looks like Network Manager fails to get the connection up:

```
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2618] device (ens3): Activation: connection 'ovs-port-phys0' enslaved, continuing activation
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2620] device (ens3): state change: ip-config -> secondaries (reason 'ip-config-unavailable', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2624] device (br-ex): state change: config -> ip-config (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2627] device (br-ex): Activation: connection 'ovs-port-br-ex' enslaved, continuing activation
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2628] device (br-ex): state change: ip-config -> secondaries (reason 'ip-config-unavailable', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2631] device (br-ex): state change: secondaries -> activated (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2645] policy: set-hostname: set hostname to 'localhost.localdomain' (no default device)
Jul 03 17:06:47 localhost NetworkManager[1360]: <info>  [1593796007.2649] device (br-ex): Activation: successful, device activated.
Jul 03 17:06:47 localhost.localdomain systemd-hostnamed[1365]: Changed host name to 'localhost.localdomain'
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.2660] device (ens3): state change: config -> ip-config (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.2680] device (ens3): Activation: connection 'ovs-if-phys0' enslaved, continuing activation
Jul 03 17:06:47 localhost.localdomain chronyd[1180]: Source 169.254.169.123 offline
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3123] device (ens3): state change: ip-config -> ip-check (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3141] device (ens3): state change: secondaries -> activated (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3151] device (ens3): Activation: successful, device activated.
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3157] device (br-ex): state change: config -> ip-config (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3165] device (br-ex): Activation: connection 'ovs-if-br-ex' enslaved, continuing activation
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3167] device (br-ex): state change: secondaries -> activated (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain kernel: ixgbevf 0000:00:03.0: NIC Link is Up 10 Gbps
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3185] device (br-ex): Activation: successful, device activated.
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3219] device (ens3): state change: ip-check -> secondaries (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3222] device (ens3): state change: secondaries -> activated (reason 'none', sys-iface-state: 'managed')
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3236] device (ens3): Activation: successful, device activated.
Jul 03 17:06:47 localhost.localdomain kernel: device br-ex entered promiscuous mode
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3262] device (br-ex): carrier: link connected
Jul 03 17:06:47 localhost.localdomain NetworkManager[1360]: <info>  [1593796007.3274] dhcp4 (br-ex): activation: beginning transaction (timeout in 45 seconds)
Jul 03 17:06:47 localhost.localdomain systemd-udevd[1416]: link_config: autonegotiation is unset or enabled, the speed and duplex are not writable.
Jul 03 17:06:48 localhost.localdomain NetworkManager[1360]: <info>  [1593796008.2592] device (ens3): carrier: link connected
Jul 03 17:06:57 localhost.localdomain systemd[1]: NetworkManager-dispatcher.service: Consumed 116ms CPU time
Jul 03 17:07:16 localhost.localdomain systemd[1]: NetworkManager-wait-online.service: Main process exited, code=exited, status=1/FAILURE
Jul 03 17:07:16 localhost.localdomain systemd[1]: NetworkManager-wait-online.service: Failed with result 'exit-code'.
Jul 03 17:07:16 localhost.localdomain systemd[1]: Failed to start Network Manager Wait Online.
Jul 03 17:07:16 localhost.localdomain systemd[1]: Dependency failed for Configures OVS with proper host networking configuration.
Jul 03 17:07:16 localhost.localdomain systemd[1]: ovs-configuration.service: Job ovs-configuration.service/start failed with result 'dependency'.
Jul 03 17:07:16 localhost.localdomain systemd[1]: NetworkManager-wait-online.service: Consumed 62ms CPU time
```

Full log here:
https://chunk.io/trozet/47f9b5e9022f435db06f3a045267e760?lang=tar

trozet:
After further investigation I believe this is a different bug in NM. Filed https://bugzilla.redhat.com/show_bug.cgi?id=1853750

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+    if [ -z "$iface_mtu" ]; then
+      echo "Unable to determine default interface MTU"
+      exit 1
+    fi
+
+    echo "MTU found for iface: ${iface}: ${iface_mtu}"
+
+    # store ipv4 address
+    iface_ip=""
+    if ip -4  addr show ${iface} |grep -Eo "([0-9]{1,3}[\.]){3}[0-9]{1,3}"; then
```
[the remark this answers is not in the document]
celebdor: You can tell keepalived addresses from regular dhcp addresses by the fact that they are deprecated addresses (preferred_lft of 0 seconds)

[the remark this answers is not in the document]
celebdor:
```
[core@master-2 ~]$ ip -j -6 address show dev enp2s0 | jq '.[0].addr_info[] | select(.preferred_life_time!=0 and .scope=="global").local'
"fd2e:6f44:5dd8:c956::16"
```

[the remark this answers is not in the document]
celebdor: In IPv4, instead of checking for lifetimes, you check for primary addresses

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+        connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu}
+    fi
+
+    if ! nmcli connection show ovs-if-br-ex &> /dev/null; then
+      nmcli c add type ovs-interface slave-type ovs-port conn.interface br-ex master ovs-port-br-ex con-name \
+        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu}
+    fi
+
+    # setting MTU on ovs iface does not seem to propagate to OVS
+    # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/206
```
celebdor, line 87: Since it should be fixed in 1.22.12, maybe we should request rhcos NM to be bumped (currently at 1.22.8 afaict).

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+    if [ ! -z "$iface_ip" ]; then
+      echo "checking iface IP matches original DHCP leased IP"
+      # sleep for a few seconds to allow DHCP to finish
+      sleep 10
+      counter=0
+      while [ $counter -lt 30 ]; do
+        new_iface_ip=$(ip -4  addr show br-ex |grep -Eo "([0-9]{1,3}[\.]){3}[0-9]{1,3}" | head -n 1)
+        if [ "$new_iface_ip" = "$iface_ip" ]; then
+          echo "DHCP successful on OVS iface"
+          exit 0
```
celebdor, line 119: Maybe we could check connectivity to the gateway for extra points and rollback if we cant.

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  [Unit]
+  Description=Configures OVS with proper host networking configuration
+  # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
+  Requires=NetworkManager-wait-online.service openvswitch.service
+  After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
+  Before=network-online.target kubelet.service crio.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
```
celebdor, line 13: Oneshot does not yet (in the slightly older systemd shipped in rhcos) support retry on failure. Should we put the logic here in the execstart?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+contents:
+  inline: |
+    #!/bin/bash
+    set -eux
+
+    iface=""
+    counter=0
+    # find default interface
+    while [ $counter -lt 12 ]; do
+      iface=$(ip route | grep default | head -n 1 | grep -Eo "dev\s[^ ]+" | sed 's/dev\s//')
```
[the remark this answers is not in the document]
trozet: thanks, i'll change it to use json output/parsing. I'm not handling IPv6 in this patch. Will do that in a follow up (along with specifying config). Just need to get basic things working with this first PR.

templates/common/_base/files/configure-ovs-network.yaml
trozet: https://bugzilla.redhat.com/show_bug.cgi?id=1820052

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+    if [ -z "$iface_mac" ]; then
+      echo "Unable to determine default interface MAC"
+      exit 1
+    fi
+
+    echo "MAC address found for iface: ${iface}: ${iface_mac}"
+
+    # find MTU on original iface
+    iface_mtu=$(ip link show ${iface} | grep -Eo "mtu\s[0-9]+" | sed 's/mtu\s//')
+    if [ -z "$iface_mtu" ]; then
```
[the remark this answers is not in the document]
trozet: done

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+    if [ "$iface" = "br-ex" ]; then
+      echo "Networking already configured and up for br-ex!"
+      exit 0
+    fi
+
+    # find the MAC of default interface to use for OVS internal port
+    # this prevents us from getting a different DHCP lease and dropping connection
+    iface_mac=$(cat /sys/class/net/${iface}/address)
+    if [ -z "$iface_mac" ]; then
```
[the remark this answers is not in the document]
trozet: done

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,134 @@
...
+  inline: |
+    #!/bin/bash
+    set -eux
+
+    iface=""
+    counter=0
+    # find default interface
+    while [ $counter -lt 12 ]; do
+      iface=$(ip route | grep default | head -n 1 | grep -Eo "dev\s[^ ]+" | sed 's/dev\s//')
+      if [ ! -z "$iface" ]; then
```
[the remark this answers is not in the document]
trozet: done

templates/common/_base/files/configure-ovs-network.yaml
trozet: this whole thing should go away when the bug is fixed, which is now verified:
https://bugzilla.redhat.com/show_bug.cgi?id=1852106

templates/common/_base/units/ovs-configuration.service.yaml
trozet: i would rather build the script to be more robust rather than add logic here

templates/common/_base/units/ovs-configuration.service.yaml
trozet: i think for now will not make this RequiredBy

templates/common/_base/files/configure-ovs-network.yaml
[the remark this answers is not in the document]
trozet: done (for ipv4)

templates/common/_base/files/configure-ovs-network.yaml
cgwalters: #1766 landed.  Now as I understand things this PR is about controlling the SDN, which we do *not* need during "firstboot" which is when we [perform os updates](https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md), only after kubelet is ready to start.

I think all of this bash code could become a `machine-config-daemon initialize-ovs` subcommand or so, then the MCD just invokes `systemd-run --unit mco-ovs /run/bin/machine-config-daemon initialize-ovs` when it starts up the first time on a node.  (Note that you will need to handle the case of the MCD being restarted without a node restart, e.g. new daemonset rollout, so your code should  be idempotent)

cgwalters:
I know we have time pressure here to deliver this feature, but we have that for every feature too :smile:  This one could really use some sort of brief public description of what we're trying to accomplish, a brief outline of the proposed architecture etc.

That said from my PoV
/approve

templates/common/_base/files/configure-ovs-network.yaml
cgwalters: OK after some real-time discussion, it sounds like we need this before kubelet, which does not fit neatly into the current architecture.  Today we have:

- Code shipped as part of the OS that runs before and up to kubelet (podman, crio, kubelet)
- The special MCD code introduced for #1766 - but that *only* applies on the firstboot, after that the MCD pod takes over (i.e. it's after kubelet)

(We're also implicitly having a larger debate about how much "host management" should happen in a pod)

One thing we could do is ship this code as part of a container image (actually, it could be the SDN container), and build upon the "host binary" trick we're already doing there, but instead of extracting it from the pod, have our systemd unit pull and extract it directly and run `Before=kubelet.service`.  Or, another variant of this is to basically just `podman run --privileged -v /:/host` and have the container perform the same trick the MCD does and extract its binary to the host, where it can then be run via `systemd-run`.

abhat:
> _No description provided._

nit: can you please add some description in the PR? I know the individual commits have detailed commit messages.

templates/common/_base/units/ovs-configuration.service.yaml
abhat, line 13: Do we need to set `RemainAfterExit=true` for this?

Also, what happens when we stop the openvswitch service? Are we supposed to deconfigure what this script configured?

trozet:
> > _No description provided._
> 
> nit: can you please add some description in the PR? I know the individual commits have detailed commit messages.

Done

templates/common/_base/units/ovs-configuration.service.yaml
trozet: I don't think we need it. Are you suggesting having it so we can check its status? Or are you saying we should implement an exec stop as well?

If someone stops OVS they will lose network connectivity. If they are stopping it for some administrative reason on the host, then there's no reason to remove the config that this script has done in NM. The NM connections will remain configured, but we also leave the original connections configured in NM with a lower priority. So the admin could deactivate the OVS connection and bring up the old connection if he/she desired.

trozet:
With the updated NetworkManager package landing with related OVS fixes, this patch now works. I've deployed 8 times today with cluster bot and every time has been successful. Removing the hold. @alexanderConstantinescu @abhat @dcbw PTAL

/hold cancel

templates/common/_base/units/ovs-configuration.service.yaml
abhat: I was thinking whatever we do to move the primary interface under OVS should be undone when OVS is stopped and moved back outside the ovs bridge as a net device of it's own on the stack.

abhat: So in that sense, yes, implement a `ExecStop`.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+      local s
+      local w
+      local fs
+
+      s='[[:space:]]*'
+      w='[a-zA-Z0-9_.-]*'
+      fs="$(echo @|tr @ '\034')"
+
+      (
+        sed -e '/- [^\“]'"[^\']"'.*: /s|\([ ]*\)- \([[:space:]]*\)|\1-\'$'\n''  \1\2|g' |
```
abhat, line 51: Use `$s` instead of `[[:space:]]*`

abhat: Also, can you add some comments on what this sed/awk black-magic does so someone trying to grok this later can have an easier time?

templates/common/_base/units/ovs-configuration.service.yaml
trozet: Hmm so in that case if OVS stops, we get a stop on this service, and then we nmcli down the OVS connection and bring up the old connection? I'm not sure of the implications there because I think NM will try to access ovsdb to delete the config (which won't be up anymore). I think we can address this as a follow up.

templates/common/_base/files/configure-ovs-network.yaml
trozet: It's magic to me as well :) I've used this in the past on another project for yaml parsing in bash, so it works. The hope is this is short-lived and replaced by something in golang when we can figure out how to do that with MCD. Or if we could get yq installed we could get rid of these functions if we stick with bash.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+      if [ -n "$iface" ]; then
+        echo "Default gateway interface found: ${iface}"
+        break
+      fi
+      counter=$((counter+1))
+      echo "No default route found on attempt: ${counter}"
+      sleep 5
+    done
+
+    if [ "$iface" = "br-ex" ]; then
```
abhat, line 119: can the name of the bridge change to anything else other than `br-ex`? Is this something that should be passed in as an argument to ExecStart?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+
+    create_variables() {
+      local yaml_file="$1"
+      local prefix="$2"
+      local yaml_string="$(parse_yaml "$yaml_file" "$prefix")"
+      unset_variables ${yaml_string[@]}
+      eval "${yaml_string}"
+    }
+
+    CONFIG_FILE="/etc/sysconfig/network-scripts/ovs-config.yaml"
```
abhat, line 103: Should users be allowed to pass the `CONFIG_FILE` as a command-line arg to the script?

templates/common/_base/files/configure-ovs-network.yaml
trozet: It will always be br-ex. We have to pass this name into ovn-k8s via CNO, and if we can stick to a single name it will make debugging easier.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+
+    phys_port_opts=""
+    CONFIG_PROVIDED=false
+    BONDING=false
+    # check if config file is provided at /etc/sysconfig/network-scripts/ovs-config.yaml
+    if [ -f "$CONFIG_FILE" ]; then
+      create_variables ${CONFIG_FILE} ovs_
+      echo "Variables parsed from OVS config file: $(compgen -A variable | grep ovs_)"
+
+      if [ ${#ovs_interfaces__name[@]} -gt 2 ]; then
```
abhat, line 132: Should this be validating that it has one of each i.e. the physical NIC and the bridge name? It just checks for the length of the array being > 2. But doesn't validate the two entries per-se? Am I missing something?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+        echo "ERROR: multiple interfaces specified. Only one (br-ex bridge and port) may be specified in the config file"
+        exit 1
+      fi
+      CONFIG_PROVIDED=true
+
+      # Add VLAN args if they exist in config
```
abhat, line 138: Consider: refactoring handling of different config options (like vlan, bonding, lacp, mtu etc.) into their own bash functions for readability.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+    fi
+
+    # create bridge
+    if ! nmcli connection show br-ex &> /dev/null; then
+      nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+    fi
+
+    old_conns=()
+    if "$BONDING"; then
+      iface=bond0
```
abhat, line 230: why hard-code to bond0? should this be `${bonded_ifaces[0]}` or whatever is the primary bond interface?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+    fi
+
+    if ! nmcli connection show ovs-if-br-ex &> /dev/null; then
+      nmcli c add type ovs-interface slave-type ovs-port conn.interface br-ex master ovs-port-br-ex con-name \
+        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+    fi
+
+    # wait for DHCP to finish, verify connection is up
+    counter=0
+    while [ $counter -lt 5 ]; do
```
abhat, line 275: Why is this hard-coded to 5 retries 5s apart? What if it takes longer to bring up the interfaces?

templates/common/_base/units/ovs-configuration.service.yaml
abhat: Sure. Please have a github issue/BZ/JIRA to track this.

abhat: Shouldn't a systemd directive be sufficient to take care of the ordering that is when OVS is stopped, this configuration service should get a chance to delete the config (while ovsdb is still up) and then set itself as stopped?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+      if nmcli conn up ovs-if-br-ex; then
+        echo "OVS successfully configured"
+        ip a show br-ex
+        exit 0
+      fi
+      sleep 5
+      counter=$((counter+1))
+    done
+
+    echo "ERROR: Failed to activate ovs-if-br-ex NM connection"
```
abhat, line 298: Isn't this roughly what you would partially do for `ExecStop`?

abhat: If so, consider refactoring this as a separate bash function which can be reused when the `ExecStop` is implemented.

kikisdeliveryservice:
adding hold until outstanding comments are addressed

/hold

templates/common/_base/files/configure-ovs-network.yaml
trozet: I should parse this from the config file. It will be under the ovs_interfaces__bridge_port__name array.

templates/common/_base/files/configure-ovs-network.yaml
trozet: Yeah will do in a follow up

templates/common/_base/files/configure-ovs-network.yaml
trozet: it wasn't really intended to be something users would run. The config file should be placed on disk during initial install.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,313 @@
...
+      fs="$(echo @|tr @ '\034')"
+
+      (
+        sed -e '/- [^\“]'"[^\']"'.*: /s|\([ ]*\)- \([[:space:]]*\)|\1-\'$'\n''  \1\2|g' |
+
+        sed -ne '/^--/s|--||g; s|\"|\\\"|g; s/[[:space:]]*$//g;' \
+            -e "/#.*[\"\']/!s| #.*||g; /^#/s|#.*||g;" \
+            -e "s|^\($s\)\($w\)$s:$s\"\(.*\)\"$s\$|\1$fs\2$fs\3|p" \
+            -e "s|^\($s\)\($w\)${s}[:-]$s\(.*\)$s\$|\1$fs\2$fs\3|p" |
+
```
danwinship, line 57: so, this is all kind of terrifying

Why do we have to support defining bonding and vlans in _this_ script? Can't we just require the admin to set that up via other means (eg, writing NetworkManager config files directly) and then have this script just only do the "create an OVS bridge and move the default interface to it" part? (And then if RHCOS gets support for nmstate or whatever later, then people can start using that feature then.)

templates/common/_base/files/configure-ovs-network.yaml
trozet: it's just a simple validation to ensure the number of entries are 2. I.E. someone doesn't write an nmstate config file expecting this to also handle other bonds or linux interfaces. You are right there could be more validation here, but given the time constraint and NM bugs this is what I got to finally pass. We can enhance it more in a follow up patch.

templates/common/_base/files/configure-ovs-network.yaml
trozet: So you are suggesting the user writes the NM key files themselves and places them into the right folder during bring up, and then what, this script just sees the files there and does nothing and allows NM to do its thing?

trozet: I've removed the NM state config reading and only left the auto detection bits. For custom network configuration the user will need to provide NM key files via ignition.

templates/common/_base/files/configure-ovs-network.yaml
trozet: NM will attempt a certain number of tries to DHCP (it's either 4 or 5) and then give up for 5 minutes. I feel like if we don't have a dhcp response in 25 seconds something is wrong. Then it will try for another 25 seconds on line 100 to bring up the connection with nmcli conn up. This is a blocking operation, so it will not return until NM gives up, and then we will try 5 more times there. tl;dr we are trying for a while here.

trozet:
@abhat @danwinship PTAL again thanks

kikisdeliveryservice:
The metal-ipi job is completely red.

/skip

abhat:
/lgtm

Other review comments can be handled later in a follow-up PR.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,116 @@
...
+    # Configures NICs onto OVS bridge "br-ex"
+    # Configuration is either auto-detected or provided through a config file written already in Network Manager
+    # key files under /etc/NetworkManager/system-connections/
+    # Managing key files is outside of the scope of this script
+
+    iface=""
+    counter=0
+    # find default interface
+    while [ $counter -lt 12 ]; do
+      iface=$(ip -j route show default | jq -r '.[0].dev')
```
danwinship, line 17: Isn't the loop unnecessary since this is `After=NetworkManager-wait-online.service`? It should be guaranteed that a default route already exists.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,116 @@
...
+    fi
+
+    if [ -z "$iface" ]; then
+      echo "ERROR: Unable to find default gateway interface"
+      exit 1
+    fi
+
+    # find the MAC from OVS config or the default interface to use for OVS internal port
+    # this prevents us from getting a different DHCP lease and dropping connection
+    if ! iface_mac=$(<"/sys/class/net/${iface}/address"); then
```
danwinship, line 39: (might want to use `ip -j link show "$iface" | jq -r '.[0].address'` for consistency?)

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,116 @@
...
+      nmcli c add type 802-3-ethernet conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
+        connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu}
+    fi
+
+    if ! nmcli connection show ovs-if-br-ex &> /dev/null; then
+      nmcli c add type ovs-interface slave-type ovs-port conn.interface br-ex master ovs-port-br-ex con-name \
+        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+    fi
+
+    # wait for DHCP to finish, verify connection is up
```
danwinship, line 85: I think you can do this with `nm-online --timeout 30` or whatever; that will wait until there is a default route again

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -0,0 +1,116 @@
...
+        ip a show br-ex
+        exit 0
+      fi
+      counter=$((counter+1))
+    done
+
+    echo "WARN: OVS did not succesfully activate NM connection. Attempting to bring up connections"
+    counter=0
+    while [ $counter -lt 5 ]; do
+      if nmcli conn up ovs-if-br-ex; then
```
danwinship, line 101: Is there some situation where we think the connection won't correctly activate automatically but will correctly activate if we poke at it?

templates/common/_base/files/configure-ovs-network.yaml
trozet: I don't think so because its a Wants+After and not a Requires. Therefore if the unit fails during start up (never gets online) this unit will still run.

templates/common/_base/files/configure-ovs-network.yaml
trozet: I think this is fine.

templates/common/_base/files/configure-ovs-network.yaml
trozet: oh cool, didn't know about nm-online command. We can fix that up later.

templates/common/_base/files/configure-ovs-network.yaml
trozet: Ideally no, but I've hit some inconsistencies in NM + OVS so this is a fail safe to try to address those if we hit them. In general, adding a connection doesn't mean that the connection will automatically connect. For example, if I add a new connection for eth0 that has higher priority than an already activated connection using eth0, it will not bring down the old connection and activate the new.

Doing an nmcli conn up on ovs-if-br-ex (the interface connection) is key because it will bring up all of the other master connections like the OVS port and the bridge and trigger those things to come up as well. I did see a case where doing an nmcli up on ovs-if-br-ex will not bring up ovs-if-phys0 and I had to manually bring that one up. Perhaps in a follow up we can add a check for that as well.

templates/common/_base/files/configure-ovs-network.yaml
danwinship: but in that case, waiting another 60 seconds isn't going to help; we can just run `ip route show default` once, and if it fails, exit immediately

trozet: well it might, because NM will still be trying to DHCP that interface even though wait online has already failed. But we can take this to a follow up.

trozet:
@kikisdeliveryservice can you please remove the hold?

kikisdeliveryservice:
> @kikisdeliveryservice can you please remove the hold?

:+1: 

/hold cancel

trozet:
@kikisdeliveryservice is there anyway we can override e2e-gcp-upgrade? This patch only affects the ovn-kubernetes path (not openshift-sdn). The e2e-gcp-upgrade job is failing due to a known bug in openshift-sdn:

https://bugzilla.redhat.com/show_bug.cgi?id=1828858

The only job actually testing this this patch is e2e-ovn-step-registry which passed.

runcom:
As explained, freeing up some jobs, this can't affect e2e-gcp-upgrade (maybe some other future upgarde job will)

templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
```diff
@@ -5,7 +5,8 @@ contents: |
   Description=Update GCP routes for forwarded IPs.
   ConditionKernelCommandLine=|ignition.platform.id=gce
   ConditionKernelCommandLine=|ignition.platform.id=gcp
-  After=network.target
+  Wants=network-online.target
+  After=network-online.target
```
runcom, line 9: triple checking on this service - @squeed you wrote the original https://github.com/openshift/machine-config-operator/blob/release-4.5/templates/master/00-master/gcp/units/openshift-gcp-routes.service#L8 w/o the wants but it seems Tim found an issue requiring it for gcp+ovn - just triple checking as this service caused some weird issues in the past

runcom: as a follow up, looks like Casey is on PTO also - talked to Stefan, we're not fully sure but Tim pointed out this is needed anyway in OVN+GCP

trozet: Yes from what I found the gcp routes thing relies on network connectivity and finding an interface to use. With these changes that interface will change sometime after network.target but before network-online.target...meaning there could be a race where gcp routes uses the wrong interface. I see GCP just passed with this patch on

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ovn-kubernetes/216/pull-ci-openshift-ovn-kubernetes-master-e2e-gcp-ovn/1288509043568021504

So I think we are OK.

PR #1867, opened 2020-06-23
[fcos] Control amount of replicas in etcd-quorum-guard

vrutkovs:
MCO should apply etcd-quorum-guard deployment instead of CVO.
It also controls the number of replicas in this deployment: it would
scale 1 replica if CEO's useUnsupportedUnsafeNonHANonProductionUnstableEtcd
option is enabled. This allows creating single node clusters.

Cherry-pick from OKD 4.4 on 4.5 branch.
/cc @LorbusChris


LorbusChris:
Just noting again that me may loose this in 4.6 if no other way is implemented in master by then

/approve
/lgtm

vrutkovs:
>Just noting again that me may loose this in 4.6 if no other way is implemented in master by then

Right, I'm going to submit this to master - but its relatively a huge change, so I'd prefer to wait for things to settle first

PR #1869, opened 2020-06-24
pkg/operator/status: gate major version upgrades on pools being ready

runcom:
This is a long standing issue so let's give this a try and check if
Upgradeable=False really gates at least major version upgrades...

**Note**: opening this for testing, I suspect Upgradeable gates major versions upgrades but not within a z stream (which is ok, minor version upgrades aren't the issue here). Also, this would effectively prevents any upgrade from starting when someone is in the middle of say rolling a machineconfig (which is kind of desiderable anyway right)

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
@@ -162,7 +163,8 @@ func (optr *Operator) updateStatus(co *configv1.ClusterOperator, status configv1
 }
 
 const (
-	asExpectedReason = "AsExpected"
+	asExpectedReason    = "AsExpected"
+	poolsNotReadyReason = "PoolsNotReady"
 )
 
 func (optr *Operator) clearDegradedStatus(task string) error {
@@ -254,6 +256,22 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		Status: configv1.ConditionTrue,
 		Reason: asExpectedReason,
 	}
+
+	allReady, notReadyPools, err := optr.arePoolsReadyToUpgrade()
+	if err != nil {
+		return err
+	}
+	var message string
+	if len(notReadyPools) != 0 {
+		message = fmt.Sprintf("Please check the MachineConfigPool: %q with: oc describe mcp/<pool-name>", strings.Join(notReadyPools, ","))
+	}
+	coStatus.Message = message
+	if allReady {
+		return optr.updateStatus(co, coStatus)
+	}
+	coStatus.Status = configv1.ConditionFalse
+	coStatus.Reason = poolsNotReadyReason
+
 	return optr.updateStatus(co, coStatus)
 }
 
@@ -322,6 +340,35 @@ func (optr *Operator) setOperatorStatusExtension(status *configv1.ClusterOperato
 		return
 	}
 	status.Extension.Raw = raw
+}
+
+func isPoolReadyToUpgrade(pool *mcfgv1.MachineConfigPool) bool {
+	return mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) && mcfgv1.IsMachineConfigPoolConditionFalse(pool.Status.Conditions, mcfgv1.MachineConfigPoolDegraded)
+}
+
+func (optr *Operator) arePoolsReadyToUpgrade() (bool, []string, error) {
+	pools, err := optr.mcpLister.List(labels.Everything())
+	if err != nil {
+		return false, nil, err
+	}
+	var (
+		ready = true
+		ret   []string
+	)
+	for _, pool := range pools {
+		poolReady := isPoolReadyToUpgrade(pool)
+		if !poolReady {
+			ret = append(ret, pool.GetName())
+		}
+		if pool.GetLabels() != nil {
+			_, requiredPool := pool.GetLabels()[requiredForUpgradeMachineConfigPoolLabelKey]
+			if !requiredPool {
+				continue
+			}
+		}
+		ready = ready && poolReady
+	}
+	return ready, ret, nil
 }
 
 func (optr *Operator) allMachineConfigPoolStatus() (map[string]string, error) {


cgwalters:
Doesn't the Cincinnati graph generally have upgrades going serially between majors at least?  Did you do this while thinking of a specific bug?

This also means we start blocking on the worker pool when we didn't before.

runcom:
> This also means we start blocking on the worker pool when we didn't before.

Right, I think what we really want is to block master at least from being able to upgrade if something is rolling mcp wise, I'll change this to account just for masters even if I do think that not waiting on workers during major version upgrades it's a huge bug that we have since day0

runcom:
> Doesn't the Cincinnati graph generally have upgrades going serially between majors at least? Did you do this while thinking of a specific bug?

also yes, there are plenty of bugs where we have clusters with workers at 4.1 and control plane at 4.4 just because we didn't wait/signal on workers being stuck :( 
About Cincinnati, not sure, but this is about blocking if the MCO is in bad shape (which it is if control plane is Degraded or Updating right?)

pkg/operator/status.go
```diff
@@ -254,6 +255,18 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+	allReady, notReadyPools, err := optr.arePoolsReadyToUpgrade()
+	if err != nil {
+		return err
+	}
+	if allReady {
+		return optr.updateStatus(co, coStatus)
+	}
+	coStatus.Status = configv1.ConditionFalse
+	coStatus.Reason = "PoolsNotReadyToUpgrade" // FIXME(runcom)
+	coStatus.Message = fmt.Sprintf("The following pools are still not ready to upgrade, please check them out: %s", strings.Join(notReadyPools, ","))
```
yuqi-zhang, line 268: Instead of `please check them out` I think it'd be more valuable to make it a bit more precise, since maybe not all sysadmin would know what a `pool` is, maybe specify: `please check the MachineConfigPool: %v with: oc describe mcp/${pool}` for which pool, etc.

Mostly because the CVO states:
`An operator reports Upgradeable as false when it wishes to prevent an upgrade for an admin-correctable condition. The component should include a message that describes what must be fixed.`

So maybe its worth it to bubble up the error here too? So the CVO can report it.

runcom: Sure, please check them out was for testing as I want to see that before polishing this out and addressing all questions :joy:

runcom:
> Also regarding: https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md , I don't see a reference to this only blocking major upgrades... wouldn't all upgrades in the graph be treated similarly regardless of z-stream or major version?

So, I saw that as well but unless my memory fails me I clearly remember Trevor saying that we only block major upgrades (I never verified yet but that’s what I want to test for this). 

And yeah, agree on blocking for control plane only 👍

runcom:
Alrighty, I've reworked this to set Upgradeable=False when control plane isn't ready but also tweak the message in case other non-blocking pools are not ready.

ptal again

runcom:
So, finally tested this out

workers are not blocking but we're loud about it:

```
...
Status:
  Conditions:
    Last Transition Time:  2020-07-01T12:18:58Z
    Message:               Cluster version is 0.0.1-2020-07-01-103322
    Status:                False
    Last Transition Time:  2020-07-01T12:18:58Z
    Status:                False
    Last Transition Time:  2020-07-01T12:18:58Z
    Message:               Cluster has deployed 0.0.1-2020-07-01-103322
    Status:                True
    Last Transition Time:  2020-07-01T12:18:58Z
    Message:               Please check the MachineConfigPool: "worker" with: oc describe mcp/<pool-name>
    Reason:                AsExpected
    Status:                True
    Type:                  Upgradeable
  Extension:
    Master:  all 3 nodes are at latest configuration rendered-master-5efde12999e1afc7fb41d6ed3cb90f2b
    Worker:  pool is degraded because nodes fail with "1 nodes are reporting degraded status on sync": "Node ip-10-0-156-111.us-west-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-worker-d23158fc1e0c0dae24287e6591415598: content mismatch for file \\\"/etc/crio/crio.conf.d/00-default\\\"\""
```

for masters tho, the flip of upgradeable isn't immediate because we'll have to wait 20m for the usual sync - will fix that

runcom:
The logic for masters is correct, but as I said, we take ~20m to report Upgradeable=False in masters case which is wrong:
```
Status:
  Conditions:
    Last Transition Time:  2020-07-01T12:18:58Z
    Message:               Cluster version is 0.0.1-2020-07-01-103322
    Status:                False
    Last Transition Time:  2020-07-01T13:01:18Z
    Message:               Failed to resync 0.0.1-2020-07-01-103322 because: timed out waiting for the condition during syncRequiredMachineConfigPools: error pool master is not ready, retrying. Status: (pool degraded: true total: 3, ready 2, updated: 2, unavailable: 1)
    Reason:                RequiredPoolsFailed
    Status:                True
    Last Transition Time:  2020-07-01T13:01:18Z
    Message:               Cluster not available for 0.0.1-2020-07-01-103322
    Status:                False
    Last Transition Time:  2020-07-01T13:01:18Z
    Message:               Please check the MachineConfigPool: "master,worker" with: oc describe mcp/<pool-name>
    Reason:                PoolsNotReady
    Status:                False
    Type:                  Upgradeable
  Extension:
    Last Sync Error:  error pool master is not ready, retrying. Status: (pool degraded: true total: 3, ready 2, updated: 2, unavailable: 1)
    Master:           pool is degraded because nodes fail with "1 nodes are reporting degraded status on sync": "Node ip-10-0-177-22.us-west-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-master-5efde12999e1afc7fb41d6ed3cb90f2b: content mismatch for file \\\"/etc/crio/crio.conf.d/00-default\\\"\""
    Worker:           pool is degraded because nodes fail with "1 nodes are reporting degraded status on sync": "Node ip-10-0-156-111.us-west-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-worker-d23158fc1e0c0dae24287e6591415598: content mismatch for file \\\"/etc/crio/crio.conf.d/00-default\\\"\""
```

yuqi-zhang:
> we take ~10m to report Upgradeable=False in masters case which is wrong

That's a result of the timing between syncs right? So unless we have a way to forcing a sync to happen and finish before an upgrade is allowed to start (when one is triggered) the other option seems to be for the MCC to set this flag when it's doing something with the master pool.

That said the only scenario I can think of where 10 min isn't enough is when someone triggers 2 updates in a row, so maybe we're fine to add the gate as is?

/retest

runcom:
> That said the only scenario I can think of where 10 min isn't enough is when someone triggers 2 updates in a row, so maybe we're fine to add the gate as is?

Uhm not sure, right now if you make the master pool Updating the mco sync starts. If that goes to Degraded we’ll notice only after 10 minutes (we do report the degradation but we don’t set Degraded=True until 10m).
I think we can either have the MCC set this flag or we can finally make the MCO control this by the mean of understanding what is an perma error and what not. In the latter case we can immediately switch Upgradeable to false (and also degraded immediately), in the former we’ll behave as usual waiting. But I think you’re right we should switch upgradeable very soon, as soon as the pool starts doing something (an upgrade in this case)

cgwalters (approved):
I know there's a lot of pending design around this (that would be nice if it happened on Github and not somewhere internal :wink: ).  

Agree with the overall idea.

PR #1870, opened 2020-06-24
Bug 1851541: Correctly handle requests for ipv4/ipv6 records

cybertron:
Previously the ingress record was hard-coded to be an A record, which
does not work when deploying with an ipv6 ingress VIP. While fixing
that, an issue with the api and api-int records was also exposed.

According to RFC 4074, when a query for a record comes in that asks
for the wrong type of record (i.e. asking for an A record when only
a AAAA record exists), the server should return NOERROR with an
empty list of records. This indicates to the client resolver that
a record of that name exists, but not with that type.

For our purposes, it was necessary to convert all of the records to
use the template plugin in order to get this behavior. The hosts
plugin previously used for the api records does not correctly handle
requests for the wrong record type, at least with the fallthrough
option that we also need. The file plugin, which does correctly
handle requests for the wrong record type, has a different issue.
We want to allow external records in the same subdomain (and this is
necessary for development purposes), but there is no way to make
the file plugin forward requests it can't handle.

To make the template plugin behave correctly, two entries are needed
for each record: one with the correct type and the appropriate answer,
and one with the incorrect type and an empty answer. This gets us
RFC-compliant behavior while still allowing fallthrough for requests
that are not present in the local server.

This change requires new fields from baremetal-runtimecfg that are
added in https://github.com/openshift/baremetal-runtimecfg/pull/60.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -5,14 +5,31 @@
     forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}}
     cache 30
     reload
-    hosts {
-        {{ .ControllerConfig.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }} api-int.{{ .ControllerConfig.EtcdDiscoveryDomain }}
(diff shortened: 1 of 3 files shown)


cybertron:
/hold

This needs https://github.com/openshift/baremetal-runtimecfg/pull/69

sdodson:
/hold cancel
/retest 
https://github.com/openshift/baremetal-runtimecfg/pull/69 has merged

PR #1872, opened 2020-06-25
Bug 1849432: [baremetal] verify resolv.conf in HAProxy static pod synced with host resolv.conf

yboaron:
In some cases, we noticed that HAProxy static pod starts running before NM resolv prepend script[1] was applied,
as a result of that the pod's resolv.conf file doesn't point to the local Coredns instance.

In this case, HAProxy pod (actually it's haproxy-monitor container) will fail to retrieve information
from api-int:kube-apiserver (because local Coredns instance his the one that resolves api-int).

With this PR the resolv.conf used by haproxy-monitor should be always synced with node's resolv.conf

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
@@ -109,12 +109,11 @@ contents:
           privileged: true
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |            
+            cp /host/etc/resolv.conf /etc/resolv.conf
+            monitor /var/lib/kubelet/kubeconfig  /config/haproxy.cfg.tmpl  /etc/haproxy/haproxy.cfg  --api-vip {{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}
         resources:
           requests:
             cpu: 100m
@@ -130,6 +129,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
+                cmp /host/etc/resolv.conf /etc/resolv.conf
         terminationMessagePolicy: FallbackToLogsOnError
         imagePullPolicy: IfNotPresent
       hostNetwork: true
--- a/templates/master/00-master/openstack/files/openstack-haproxy.yaml
+++ b/templates/master/00-master/openstack/files/openstack-haproxy.yaml
@@ -107,12 +107,11 @@ contents:
           privileged: true
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.OpenStack.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |            
+            cp /host/etc/resolv.conf /etc/resolv.conf
+            monitor /var/lib/kubelet/kubeconfig  /config/haproxy.cfg.tmpl  /etc/haproxy/haproxy.cfg  --api-vip {{ .Infra.Status.PlatformStatus.OpenStack.APIServerInternalIP }}
         resources:
           requests:
             cpu: 100m
@@ -128,6 +127,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
+                cmp /host/etc/resolv.conf /etc/resolv.conf
         terminationMessagePolicy: FallbackToLogsOnError
         imagePullPolicy: IfNotPresent
       hostNetwork: true
--- a/templates/master/00-master/ovirt/files/ovirt-haproxy.yaml
+++ b/templates/master/00-master/ovirt/files/ovirt-haproxy.yaml
@@ -87,12 +87,11 @@ contents:
           privileged: true
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.Ovirt.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            cp /host/etc/resolv.conf /etc/resolv.conf
+            monitor /var/lib/kubelet/kubeconfig  /config/haproxy.cfg.tmpl  /etc/haproxy/haproxy.cfg  --api-vip {{ .Infra.Status.PlatformStatus.Ovirt.APIServerInternalIP }}
         resources:
           requests:
             cpu: 100m
@@ -108,6 +107,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
+                cmp /host/etc/resolv.conf /etc/resolv.conf
         terminationMessagePolicy: FallbackToLogsOnError
         imagePullPolicy: IfNotPresent
       hostNetwork: true
--- a/templates/master/00-master/vsphere/files/vsphere-haproxy.yaml
+++ b/templates/master/00-master/vsphere/files/vsphere-haproxy.yaml
@@ -114,12 +114,11 @@ contents:
           privileged: true
         image: {{ .Images.baremetalRuntimeCfgImage }}
         command:
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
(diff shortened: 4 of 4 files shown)


yboaron:
I'll wait first for feedback on the suggested fix for baremetal (and see also the results of CI e2e-metal-ipi) and after that will adopt this change to other platforms.

yboaron:
The prepend NM script being used also by the rest of the platforms (Vsphere, Ovirt and Openstack), I guess I should update the PR.

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -109,12 +109,12 @@ contents:
...
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            #/bin/bash
+            cp /host/etc/resolv.conf /etc/resolv.conf
```
rgolangh, line 116: what about this non-atomic copy? that's another race

yboaron: @rgolangh  I don't think that atomic cp is required here,  the monitor code runs periodically so even if it fails once or twice  (till the CP operation completed) it should be OK.

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -109,12 +109,12 @@ contents:
...
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            #/bin/bash
```
bcrochet, line 115: This is just a comment. I think you meant to have #!

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
```diff
@@ -130,6 +130,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
```
bcrochet, line 139: You used #!/bin/bash everywhere else. Why not here?

templates/master/00-master/openstack/files/openstack-haproxy.yaml
```diff
@@ -107,12 +107,12 @@ contents:
...
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.OpenStack.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            #/bin/bash
```
bcrochet, line 113: Ditto on #!

templates/master/00-master/openstack/files/openstack-haproxy.yaml
```diff
@@ -128,6 +128,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
```
bcrochet, line 137: #!/bin/bash?

templates/master/00-master/ovirt/files/ovirt-haproxy.yaml
```diff
@@ -87,12 +87,12 @@ contents:
...
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.Ovirt.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            #/bin/bash
```
bcrochet, line 93: Ditto

templates/master/00-master/ovirt/files/ovirt-haproxy.yaml
```diff
@@ -108,6 +108,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
```
bcrochet, line 117: Ditto.

templates/master/00-master/vsphere/files/vsphere-haproxy.yaml
```diff
@@ -114,12 +114,12 @@ contents:
...
-        - monitor
-        - "/var/lib/kubelet/kubeconfig"
-        - "/config/haproxy.cfg.tmpl"
-        - "/etc/haproxy/haproxy.cfg"
-        - "--api-vip"
-        - "{{ .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP }}"
+          - "/bin/bash"
+          - "-c"
+          - |
+            #/bin/bash
```
bcrochet, line 120: Ditto.

templates/master/00-master/vsphere/files/vsphere-haproxy.yaml
```diff
@@ -135,6 +135,14 @@ contents:
           mountPath: "/host"
         - name: kubeconfigvarlib
           mountPath: "/var/lib/kubelet"
+        livenessProbe:
+          initialDelaySeconds: 10
+          exec:
+            command:
+              - /bin/bash
+              - -c
+              - |
```
bcrochet, line 144: Ditto

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
rgolangh: I'm not worried about the monitor, but on the dns resolver that could have a resolv.conf bits in-flight. I'm quite sure this is what happened on the old prepender.

cybertron:
I wouldn't think you even need to specify the interpreter when you're already calling /bin/bash -c. It should be fine to exclude that completely.

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
yboaron: I think we should run bash command in this way, we had an issue with Keepalived Liveness [1] command, and at the end that was the 'winning' syntax.

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/baremetal-keepalived.yaml#L81-#L85

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
yboaron: Do you mean the resolv.conf in the host or in the pod?

mandre:
> I wouldn't think you even need to specify the interpreter when you're already calling /bin/bash -c. It should be fine to exclude that completely.

Definitely.
The `#/bin/bash` mistake is copied in a few places in our templates and should probably be cleaned up at some point.

templates/master/00-master/baremetal/files/baremetal-haproxy.yaml
celebdor: better to do it atomically

yboaron: @celebdor @rgolangh 
I tried to change cp to be atomic ( use the same trick as [1] ) , but mv command fails with EBUSY error because /etc/rsolv.conf is a mount point.

Do you have any idea how to make cp atomic?

[1] https://github.com/openshift/machine-config-operator/pull/1763/files

cybertron: I looked into this a bit more. First, I don't think we can make operations on resolv.conf atomic because it is its own tmpfs filesystem inside the container. Even mv is not atomic across filesystems (it degrades into a cp+unlink).

However, I don't think this is a concern inside the container. As Yossi pointed out, if we get an incomplete resolv.conf from the copy then the liveness probe will immediately fail because of the mismatch with the host resolv.conf. Furthermore, nothing is running in the container at the point where the copy is made so there's no possibility of in-flight DNS resolutions using a bad resolv.conf. The monitor doesn't start until after the file has been copied. The important thing here is that we're not modifying the system-wide resolv.conf, just this one container's, so there's no potential problems with concurrent access to the file.

mandre:
I do not know about the state of ovirt and vsphere CI, but openstack one did succeed earlier. +1 from me.

rgolangh:
ovirt is unfortunately broken by the latest cluster-etcd-operator changes that I hope https://github.com/openshift/cluster-etcd-operator/pull/388 will solve.

yboaron:
@rgolangh do you think we can merge this PR without green Ovirt CI?
because this PR fixes release (4.6) blocker bug.

rgolangh:
> @rgolangh do you think we can merge this PR without green Ovirt CI?
> because this PR fixes release (4.6) blocker bug.
This fix can potentially brake ovirt so I prefer not. The thing is that we have this fix to unblock the CI https://github.com/openshift/release/pull/10197/ 
@ironcladlou Can you please unhold the above fix, and then after CEO fix is merged we will revert?

yboaron:
> > @rgolangh do you think we can merge this PR without green Ovirt CI?
> > because this PR fixes release (4.6) blocker bug.
> 
> This fix can potentially brake ovirt so I prefer not. The thing is that we have this fix to unblock the CI [openshift/release#10197](https://github.com/openshift/release/pull/10197)
> @ironcladlou Can you please unhold the above fix, and then after CEO fix is merged we will revert?

@rgolangh  Until Ovirt CI will be fixed, will it be possible for you to update locally haproxy manifest under /etc/kubernetes/manifest in one your master nodes and check that it doesn't break anything?

rgolangh:
/test e2e-ovirt

@yboaron ovirt ci should be green now. lets see.

yboaron:
@patrickdillon  from VSPhere perspective,  do you think we can merge this fix?

yboaron:
@mandre Do you think we can remove the hold ?

rgolangh:
@yboaron @mandre lgtm for ovirt. no hold from our side

patrickdillon:
/lgtm
for vsphere

yboaron:
@runcom  @ericavonb  Can we merge this PR? seems that we have green light from all platforms

PR #1778, opened 2020-06-03
DNM: Migrate internals to Ignition spec v3.1

LorbusChris:
This migrates all internal on-disk state representation over to
Ignition config spec v3.1.

Compatibilty with older boot images is retained:
In the MCS, the User Agent on the request is inspected so it can
respond with a supported version, translated in needed.

Contains all 3 commits from #1703.

Will be superseded by https://github.com/openshift/machine-config-operator/pull/1873

PR #1873, opened 2020-06-25
*: Ign spec 3 upgrade

yuqi-zhang:
Based on https://github.com/openshift/machine-config-operator/pull/1703, this is a (messy and hacky) MCO that renders all existing configs to spec 3. Currently tested to work with a default install upgrading to this.

Unit and e2e tests not included

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -119,7 +119,7 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 // PointerConfig generates the stub ignition for the machine to boot properly
 // NOTE: If you change this, you also need to change the pointer configuration in openshift/installer, see
 // https://github.com/openshift/installer/blob/master/pkg/asset/ignition/machine/node.go#L20
-func PointerConfig(ignitionHost string, rootCA []byte) (ign2types.Config, error) {
+func PointerConfig(ignitionHost string, rootCA []byte) (ign3types.Config, error) {
 	configSourceURL := &url.URL{
 		Scheme: "https",
 		Host:   ignitionHost,
@@ -128,20 +128,21 @@ func PointerConfig(ignitionHost string, rootCA []byte) (ign2types.Config, error)
 	// we do decoding here as curly brackets are escaped to %7B and breaks golang's templates
 	ignitionHostTmpl, err := url.QueryUnescape(configSourceURL.String())
 	if err != nil {
-		return ign2types.Config{}, err
+		return ign3types.Config{}, err
 	}
-	return ign2types.Config{
-		Ignition: ign2types.Ignition{
-			Version: ign2types.MaxVersion.String(),
-			Config: ign2types.IgnitionConfig{
-				Append: []ign2types.ConfigReference{{
-					Source: ignitionHostTmpl,
+	CASource := dataurl.EncodeBytes(rootCA)
+	return ign3types.Config{
+		Ignition: ign3types.Ignition{
+			Version: ign3types.MaxVersion.String(),
+			Config: ign3types.IgnitionConfig{
+				Merge: []ign3types.Resource{{
+					Source: &ignitionHostTmpl,
 				}},
 			},
-			Security: ign2types.Security{
-				TLS: ign2types.TLS{
-					CertificateAuthorities: []ign2types.CaReference{{
-						Source: dataurl.EncodeBytes(rootCA),
+			Security: ign3types.Security{
+				TLS: ign3types.TLS{
+					CertificateAuthorities: []ign3types.Resource{{
+						Source: &CASource,
 					}},
 				},
 			},
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -48,13 +48,13 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 
 	var fips bool
 	var kernelType string
-	var outIgn ign2types.Config
+	var outIgn ign3types.Config
 	var err error
 
 	if configs[0].Spec.Config.Raw == nil {
-		outIgn = ign2types.Config{}
+		outIgn = ign3types.Config{}
 	} else {
-		outIgn, err = IgnParseWrapper(configs[0].Spec.Config.Raw)
+		outIgn, err = ParseAndConvertConfig(configs[0].Spec.Config.Raw)
 		if err != nil {
 			return nil, err
 		}
@@ -66,16 +66,16 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 			fips = true
 		}
 
-		var appendIgn ign2types.Config
+		var mergedIgn ign3types.Config
 		if configs[idx].Spec.Config.Raw == nil {
-			appendIgn = ign2types.Config{}
+			mergedIgn = ign3types.Config{}
 		} else {
-			appendIgn, err = IgnParseWrapper(configs[idx].Spec.Config.Raw)
+			mergedIgn, err = ParseAndConvertConfig(configs[idx].Spec.Config.Raw)
 			if err != nil {
 				return nil, err
 			}
 		}
-		outIgn = ign2.Append(outIgn, appendIgn)
+		outIgn = ign3.Merge(outIgn, mergedIgn)
 	}
 	rawOutIgn, err := json.Marshal(outIgn)
 	if err != nil {
@@ -150,10 +150,10 @@ func PointerConfig(ignitionHost string, rootCA []byte) (ign2types.Config, error)
 }
 
 // NewIgnConfig returns an empty ignition config with version set as latest version
-func NewIgnConfig() ign2types.Config {
-	return ign2types.Config{
-		Ignition: ign2types.Ignition{
-			Version: ign2types.MaxVersion.String(),
+func NewIgnConfig() ign3types.Config {
+	return ign3types.Config{
+		Ignition: ign3types.Ignition{
+			Version: ign3types.MaxVersion.String(),
 		},
 	}
 }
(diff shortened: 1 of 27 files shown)


pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -96,6 +96,8 @@ type ControllerConfigSpec struct {
 	// Nobody is also changing this once the cluster is up and running the first time, so, disallow
 	// regeneration if this changes.
 	NetworkType string `json:"networkType,omitempty"`
+
+	IgnitionSpecVersion int32 `json:"ignitionSpecVersion"`
```
yuqi-zhang, line 100: Not needed, will remove

pkg/controller/common/helpers.go
```diff
@@ -250,30 +252,145 @@ func IgnParseWrapper(rawIgn []byte) (interface{}, error) {
...
+			return ign3types.Config{}, errors.Wrapf(err, "failed to convert Ignition config spec v2 to v3")
+		}
+		return convertedIgnV3, nil
+	default:
+		return ign3types.Config{}, errors.Errorf("unexpected type for ignition config: %v", typedConfig)
+	}
+}
+
+// Test: have some situations de-duplicate and see what happens, will merge into main function
+func ParseAndConvertConfigWithDeduplication(rawIgn []byte) (ign3types.Config, error) {
```
yuqi-zhang, line 277: will probably merge this into the main function, just trying to think if there's scenarios we won't want to de-duplicate

pkg/controller/common/helpers.go
```diff
@@ -250,30 +252,145 @@ func IgnParseWrapper(rawIgn []byte) (interface{}, error) {
...
+			glog.Infof("Found duplicate file: %v, older file will not be rendered.", duplicate)
+			continue
+		}
+		outFiles = append(outFiles, files[i])
+		filePathMap[path] = true
+	}
+
+	// TODO jerzhang check if you can have contents and dropins in the same file
+	unitNameMap := map[string]bool{}
+	var outUnits []ign2types.Unit
```
yuqi-zhang, line 327: so units is somewhat wonky since it needs to consider dropins when doing a merge, now consider this scenario:
```
Unit A:
  Content: AC
  Dropins:
    Dropin: AD
```
```
Unit B:
    Content: BC
  Dropins:
    Dropin: BD
```
Should the final look like:
```
  Unit rendered:
    Content: BC
  Dropins:
    Dropin: AD
    Dropin: BD
```
or 
```
Unit rendered:
    Content: BC
  Dropins:
    Dropin: BD
```

pkg/controller/render/render_controller.go
```diff
@@ -481,6 +481,23 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
...
+	// For now before we generate the rendered configs, let's also update the MCs here
+	// TODO should move this
+	for _, config := range configs {
+
+		needUpdate, err := ctrlcommon.NeedUpdateConfigToSpec3(config)
+		if err != nil {
+			return err
+		}
+		if needUpdate {
+			_, err = ctrl.client.MachineconfigurationV1().MachineConfigs().Update(context.TODO(), config, metav1.UpdateOptions{})
```
yuqi-zhang, line 493: For non-rendered configs I probably need a better location to update them, there is also the Boostrap path to consider

pkg/daemon/daemon.go
```diff
@@ -1276,17 +1277,33 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) error
...
 		return errors.Errorf("Failed to parse Ignition for validation: %s", err)
 	}
-	if err := checkFiles(currentIgnConfig.Storage.Files); err != nil {
-		return err
-	}
-	if err := checkUnits(currentIgnConfig.Systemd.Units); err != nil {
-		return err
+
+	switch typedConfig := ignconfigi.(type) {
+	case ign3types.Config:
```
yuqi-zhang, line 1288: for validation I'm thinking we should keep validating in the config that it was created in

pkg/controller/render/render_controller.go
yuqi-zhang: that said with this update in place we will no longer be able to rollback, so maybe for user-created configs I should keep it in the version that it was on?

pkg/daemon/daemon.go
kikisdeliveryservice: doesnt the ign translator func also validate?

pkg/controller/common/helpers.go
LorbusChris: I'd be in favor of the former, having both drop ins added.

Another case that may also be considered is if there is a unit in one template, and a dropin only in another template.

yuqi-zhang: In the case of a unit in one and a dropin in another, both the existing spec 2 behaviour and the ignition spec 3 merge function keeps both, so I think we should keep that.

And agreed, having both dropins probably makes it more consistent

pkg/daemon/daemon.go
yuqi-zhang: Right in this case I meant validating the on-disk config based on the rendered-config

If the cluster is still on a spec 2 rendered-config version (which shouldn't happen that often, only should happen if a) templates changed causing a race or b) a rendered-config was created during a middle of an update or c) a previous rendered-config update was failing), I'm thinking we should continue to validate in spec 2, instead of translating up and then validating in spec 3.

Maybe we could get away with translating first, just have to be a bit more careful

yuqi-zhang:
With https://github.com/openshift/machine-config-operator/pull/1766 rebased this also works on first installs (take a 4.6 installer and override release image with this MCO), which brings up everything in spec 3 automatically.

Next steps:
 - consolidate with new changes in https://github.com/openshift/machine-config-operator/pull/1703
 - clean up code
 - fix tests

LorbusChris:
@yuqi-zhang I'll leave https://github.com/openshift/machine-config-operator/pull/1778 open for now, as it might be helpful as a reference. The upgrade path isn't handled there yet, because there is no deduplication of units happening (we'll probably also want this for files).

yuqi-zhang:
Updated tests in reference to https://github.com/openshift/machine-config-operator/pull/1778. Probably need to add a few more but hopefully the existing ones pass now.

Also cleaned up the code a bit. Rebased on https://github.com/openshift/machine-config-operator/pull/1703 again and tested manually to pass upgrades + install.

yuqi-zhang:
/test e2e-aws

flaking on various other tests, but just to make sure...

yuqi-zhang:
Rebased, fixed some tests and fixed unit deduplication a bit more

pkg/controller/common/helpers.go
```diff
@@ -271,9 +296,107 @@ func IgnParseWrapper(rawIgn []byte) (ign2types.Config, error) {
...
+// Function to remove duplicated files/units from a V2 MC, since the translator
+// (and ignition spec V3) does not allow for duplicated entries in one MC.
+// This should really not change the actual final behaviour, since it keeps
+// ordering into consideration and has contents from the highest alphanumeric
+// MC's final version of a file.
+// Note:
+// Append is not considered since we do not allow for appending
+// Units have one exception: dropins are concat'ed
+
+func removeIgnDuplicateFilesAndUnits(ignConfig ign2types.Config) ign2types.Config {
```
runcom, line 331: I think this function would benefit from a couple of unit tests scenarios 👼 just to make sure I don't go and change something and break some cases - what do you think?

pkg/controller/common/helpers.go
```diff
@@ -271,9 +296,107 @@ func IgnParseWrapper(rawIgn []byte) (ign2types.Config, error) {
...
+	}
+
+	return ignConfig
+}
+
 // TranspileCoreOSConfigToIgn transpiles Fedora CoreOS config to ignition
-// internally it transpiles to Ign spec v3 config and translates to spec v2
-func TranspileCoreOSConfigToIgn(files, units []string) (*ign2types.Config, error) {
+// internally it transpiles to Ign spec v3 config
+func TranspileCoreOSConfigToIgn(files, units []string) (*ign3types.Config, error) {
```
LorbusChris, line 399: Now that we don't convert the FCCT templates to spec v2 anymore, we must ensure that the spec v3 config resulting from this function does not contain duplicate unit and file entries, i.e. they cannot be blindly appended like it has been done until now.

LorbusChris: wondering why we're not hitting this with `kubelet.service` here in CI already (there two templates for it: one that defines a common dropin, and one each for masters/workers that defines the contents)

pkg/controller/common/helpers.go
yuqi-zhang: Added a unit test, thanks!

pkg/controller/common/helpers.go
yuqi-zhang: Edited so we use the updated fcct parser, and parse each file to igntion first and then merge them

runcom:
Holding to get https://github.com/openshift/machine-config-operator/pull/1792 first (and I’ll be away tomorrow)

/lgtm
/hold

yuqi-zhang:
Squashed commits and added better message

yuqi-zhang:
I don't expect any of those tests to pass except `/test e2e-openstack` , which would be nice to make sure

azure we're out of quota, and the other ones we haven't passed in awhile.

LorbusChris:
:shipit:

kikisdeliveryservice:
> I don't expect any of those tests to pass except `/test e2e-openstack` , which would be nice to make sure
> 
> azure we're out of quota, and the other ones we haven't passed in awhile.

Can you run the e2e-azure test on your own azure cluster? or is quota gone for everything?

LorbusChris:
now that #1792 has merged, does this need an update to switch the pointer config to v3 as well?
ref: https://github.com/openshift/machine-config-operator/pull/1792/files#diff-5926e0cec606d949b66f823310790298R142

yuqi-zhang:
Rebased again and modified stub secret generation to v3

yuqi-zhang:
/hold cancel

This is ready to go

> Can you run the e2e-azure test on your own azure cluster? or is quota gone for everything?

Unfortunately the CI account and my account is the same namespace I think

Issue #1875, opened 2020-06-26
mastersSchedulable=false does not take effect if worker node label and master taint are missing


damemi:
**Description**
If `mastersSchedulable: false` is set in the scheduler, the [`reconcileMaster()` function](https://github.com/openshift/machine-config-operator/blob/915ca7a11fead41d75ffa0ed7e310f7284645a13/pkg/controller/node/node_controller.go#L359-L380) checks for the [presence of a `Worker` label and no `Master` taint](https://github.com/openshift/machine-config-operator/blob/915ca7a11fead41d75ffa0ed7e310f7284645a13/pkg/controller/node/node_controller.go#L256-L265) to determine if the node is `alreadySchedulable`

However, in the event that the Master taint was removed (and no worker label is present), this function will not reconcile that a Master taint should be reapplied when missing and `mastersSchedulable: false` due to the checks in its conditional:

```
if mastersSchedulable && !alreadySchedulable {
  ... remove master taint, apply worker label ...
} else if !mastersSchedulable && alreadySchedulable {
  ... apply master taint, remove worker label
}
```

so in the case where `mastersSchedulable: false` and the node simply had its master taint removed, this evaluates to
```
if false && true {
} else if true && false {
}
```

**Steps to reproduce the issue:**
1. Remove the master taint from a master node
2. Set `mastersSchedulable: false` in the scheduler config
3. Create several pods, some will still schedule on the master node without the taint

**Describe the results you received:**
Pods are still scheduled onto master nodes even with `mastersSchedulable: false`

**Describe the results you expected:**
User expects `mastersSchedulable: false` to be the source of truth over whether masters are schedulable.

**Additional information you deem important (e.g. issue happens only occasionally):**
Could this be intentional, to allow users to set only *some* masters schedulable manually?

**Output of `oc adm release info --commits | grep machine-config-operator`:**

```
(paste your output here)
```

**Additional environment details (platform, options, etc.):**

runcom:
@ravisantoshgudimetla can you take a quick look at this?

PR #1878, opened 2020-06-29
Bug 1851226: Refactor mastersSchedulable reconciliation checks

damemi:
**- What I did**
Fixes https://github.com/openshift/machine-config-operator/issues/1875
The `CheckMasterIsAlreadySchedulable` check is redundant in these cases, because both `makeMasterNode(Un)Schedulable` functions already check if the worker label exists before adding/removing it, and `make..Schedulable` checks for the taint, so we just need to add the taint check to `make...UnSchedulable`.

This also cleans up the logic where if a master taint was removed (effectively making the master nodes schedulable), it could conflict with a `mastersSchedulable: false` setting in the schedule, and would never be reconciled (because `CheckMasterIsAlreadySchedulable` always returned `false` without the master taint.

This makes the `mastersSchedulable: false` setting the ultimate source-of-truth for whether these taints should be applied to nodes, and while this removes any use cases for making only certaint master nodes schedulable it will also resolve some confusion from users.

**- How to verify it**
1. Set `mastersSchedulable: false` in the scheduler
2. Remove the master taint from a master 
3. The master taint should be re-applied by node controller

`mastersSchedulable: false` will now re-apply the master NoSchedule taint to any master node missing it

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -40,6 +40,12 @@ import (
 )
 
 const (
+	// MasterLabel defines the label associated with master node. The master taint uses the same label as taint's key
+	MasterLabel = "node-role.kubernetes.io/master"
+
+	// WorkerLabel defines the label associated with worker node.
+	WorkerLabel = "node-role.kubernetes.io/worker"
+
 	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
 	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
 	// a machineconfig pool is going to be requeued:
@@ -51,14 +57,8 @@ const (
 	// https://github.com/openshift/machine-config-operator/issues/301
 	updateDelay = 5 * time.Second
 
-	// masterLabel defines the label associated with master node. The master taint uses the same label as taint's key
-	masterLabel = "node-role.kubernetes.io/master"
-
 	// osLabel is used to identify which type of OS the node has
 	osLabel = "kubernetes.io/os"
-
-	// workerLabel defines the label associated with worker node.
-	workerLabel = "node-role.kubernetes.io/worker"
 
 	// schedulerCRName that we're interested in watching.
 	schedulerCRName = "cluster"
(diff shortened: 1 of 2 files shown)


ravisantoshgudimetla:
> This also cleans up the logic where if a master taint was removed (effectively making the master nodes schedulable), it could conflict with a mastersSchedulable: false setting in the schedule, and would never be reconciled (because CheckMasterIsAlreadySchedulable always returned false without the master taint.


How did the master taint get removed. It shouldn't have been the case, we need to identify which component is unsetting it or if the customer did that. In either case, if this is blocking customer let's get this merged with a TODO or comment which mentions the use-case for which we are doing reconciliation and saying that this needs to be moved elsewhere.

To give everyone some more background, we did not do this reconciliation in the code previously because we wanted to have another controller that is capable of reconciling label/taints.

damemi:
> How did the master taint get removed. It shouldn't have been the case, we need to identify which component is unsetting it or if the customer did that.

@ravisantoshgudimetla I agree that this could be the bigger problem, if there is some component that is removing the taint (or not properly re-applying it). 

In this customer's case they indicated that they were seeing this behavior (`mastersSchedulable: false` and pods still being scheduled to master) before (and after) an upgrade. So it seems to be unrelated to the upgrade itself, but they included that information.

This fix would be a bandaid to the problem of, "why are pods still being scheduled to masters when I set `mastersSchedulable: false`?"

ravisantoshgudimetla:
> This fix would be a bandaid to the problem of, "why are pods still being scheduled to masters when I set mastersSchedulable: false?"

I am fine if this is a band-aid but please don't close the bug associated and ask more information from the author of the BZ. This way we can identify if there is any other entity which removed the taint.

/lgtm

damemi:
@ravisantoshgudimetla agree, we need to find out more from the customer about how this could have happened. Still waiting on a response from them
/retest

kikisdeliveryservice:
> > This fix would be a bandaid to the problem of, "why are pods still being scheduled to masters when I set mastersSchedulable: false?"
> 
> I am fine if this is a band-aid but please don't close the bug associated and ask more information from the author of the BZ. This way we can identify if there is any other entity which removed the taint.
> 
> /lgtm

For clarity if we don't want to close the associated BZ, should we remove from title of this PR?

runcom:
> For clarity if we don't want to close the associated BZ, should we remove from title of this PR?

I think it's ok to keep it there for tracking purpose maybe - Steve explained we can attach how many PRs we want to a BZ, the automation will move the BZ to POST yeah, I'll move it back once this merges

/approve
/retest

PR #1879, opened 2020-06-29
Bug 1848094: [baremetal] Set NonVirtualIP in mdns plugin config

cybertron:
In order to avoid sending multicast traffic on all configured
interfaces on a system, we need to specify the address of the
interface we want to send the traffic on.

Because this is a high priority problem on baremetal and (at least
to my knowledge) not on other platforms, I'm doing it for baremetal
only for now. We can do the other platforms as a followup, but I
don't want to hold it up waiting for everyone to sign off.

### DIFF
--- a/manifests/baremetal/coredns-corefile.tmpl
+++ b/manifests/baremetal/coredns-corefile.tmpl
@@ -1,7 +1,7 @@
 . {
     errors
     health :18080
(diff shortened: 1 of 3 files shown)


sdodson:
/cc @hardys @stbenjam 
PTAL

sdodson:
@cybertron Why aren't the MCO changes necessary in the 4.6 bug?

cybertron:
> @cybertron Why aren't the MCO changes necessary in the 4.6 bug?

They were: https://github.com/openshift/machine-config-operator/pull/1767 It didn't backport cleanly due to other changes in 4.6.

PR #1880, opened 2020-06-30
Bug 1805019: KubeletConfig: Verify EvictionSoftGracePeriod is set when EvictionSof…

harche:
**- What I did**
Fixes - https://bugzilla.redhat.com/show_bug.cgi?id=1805019

**- How to verify it**
Try to create a KubeletConfig with `EvictionSoft` but without `EvictionSoftGracePeriod` and  while describing the kubeletconfig, `oc describe KubeletConfig worker-kubeconfig-fix` you should be able to see the error,
```sh
      Custom - Kubelet:  small-pods
Status:
  Conditions:
    Last Transition Time:  2020-06-30T11:57:34Z
    Message:               Error: KubeletConfiguration: EvictionSoftGracePeriod must be set when evictionSoft is defined, evictionSoft: map[memory.available:90% nodefs.available:90% nodefs.inodesFree:90%]
    Status:                False

```

Making sure `EvictionSoftGracePeriod` is specified when `EvictionSoft` in KubeletConfig

### DIFF
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -120,6 +120,18 @@ func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
 		return fmt.Errorf("KubeletConfiguration: staticPodPath is not allowed to be set, but contains: %s", kcDecoded.StaticPodPath)
 	}
 
+	if kcDecoded.EvictionSoft != nil && len(kcDecoded.EvictionSoft) > 0 {
+		if kcDecoded.EvictionSoftGracePeriod == nil || len(kcDecoded.EvictionSoftGracePeriod) == 0 {
+			return fmt.Errorf("KubeletConfiguration: EvictionSoftGracePeriod must be set when evictionSoft is defined, evictionSoft: %v", kcDecoded.EvictionSoft)
+		}
+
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
I think this check works but that we should also check when it's created so that we don't create an incorrect KC in the first place. Probably like here in validateUserKubeletConfig: https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/helpers.go#L90

@rphillips WDYT?

runcom:
> I think this check works but that we should also check when it's created so that we don't create an incorrect KC in the first place. Probably like here in validateUserKubeletConfig: https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/helpers.go#L90

I think this PR is indeed adding validation in `validateUserKubeletConfig` right?

harche:
> > I think this check works but that we should also check when it's created so that we don't create an incorrect KC in the first place. Probably like here in validateUserKubeletConfig: https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/helpers.go#L90
> 
> I think this PR is indeed adding validation in `validateUserKubeletConfig` right?

Yes

runcom:
leaving to Ryan to lgtm

/approve

PR #1889, opened 2020-07-01
[fcos] pkg/controller/common/helpers: replace duplicate files in ConvertIgnition2to3

vrutkovs:
When Ignition2 files are converted to Ignition3 we have to ensure no files are duplicated. This commit would rewrite previous versions of the file with the latest content.

Fixes https://github.com/openshift/okd/issues/237

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -125,12 +125,30 @@ func WriteTerminationError(err error) {
 	glog.Fatal(msg)
 }
 
+func replaceOrAppend(files []ign2types.File, file ign2types.File) []ign2types.File {
+	for i, f := range files {
+		if f.Node.Path == file.Node.Path {
+			files[i] = file
+			return files
+		}
+	}
+	return append(files, file)
+}
+
 // ConvertIgnition2to3 takes an ignition spec v2 config and returns a v3 config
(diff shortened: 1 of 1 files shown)


pkg/controller/common/helpers.go
```diff
@@ -125,13 +125,31 @@ func WriteTerminationError(err error) {
...
 }
 
+func replaceOrAppend(files []ign2types.File, file ign2types.File) {
+	for i, f := range files {
+		if f.Node.Path == file.Node.Path {
+			files[i] = file
+			return
+		}
+	}
+	files = append(files, file)
```
LorbusChris, line 135: I think we're missing a `return` after this line

LorbusChris: actually, it needs look like this: https://github.com/openshift/machine-config-operator/pull/1703/commits/db4a802817e7339da5ae74d751069e1d3e67614d
anoalog to the way it's done in installer

vrutkovs: I was mutating `files` there, reworked to return the array

vrutkovs:
`failed to acquire lease: resources not found`

/retest

LorbusChris:
@vrutkovs compare with https://github.com/coreos/ign-converter/pull/9

vrutkovs:
I'd prefer ignconverter solution in the long term, however this patch would unblock users using mirroring quickier

PR #1890, opened 2020-07-01
[fcos] disable zincati on masters/workers

vrutkovs:
This ensures Zincati is disabled early on master pivot.

Fixes https://github.com/openshift/okd/issues/215


cgwalters:
/lgtm
(Though see https://github.com/coreos/fedora-coreos-tracker/issues/392#issuecomment-606670949 for a stronger way to disable it)

PR #1891, opened 2020-07-02
pkg/daemon: make force use the force

runcom:
force does 2 things mainly: skip validation and "force" update to the desiredConfig. That doesn't mean the MCD can't fail anymore, we're just forcing a rollout that wouldn't have normally happened. The usual cases are: pivot not booting into the right ostree, someone writes the force file and the update is retryed (but still can fail!) or somebody touches files on disk, use the force file to bring the system back to desiredConfig w/o validating (and degrading)

cc @cgwalters related to https://github.com/openshift/machine-config-operator/pull/1086#issuecomment-652592586

Testing with:

- `oc debug node/<worker>`
- edit `/etc/crio/crio.conf.d/00-default` and add a dummy comment
- check MCD degraded because of the above
- touch the file, reboot the MCD
- watch it recover to the correct file

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1027,7 +1027,7 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 	if currentOnDisk != nil && state.currentConfig.GetName() != currentOnDisk.GetName() {
 		// The on disk state (if available) is always considered truth.
 		// We want to handle the case where etcd state was restored from a backup.
-		dn.logSystem("Disk currentConfig %s overrides node annotation %s", currentOnDisk.GetName(), state.currentConfig.GetName())
+		dn.logSystem("Disk currentConfig %s overrides node's currentConfig annotation %s", currentOnDisk.GetName(), state.currentConfig.GetName())
 		state.currentConfig = currentOnDisk
 	}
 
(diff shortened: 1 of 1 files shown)


runcom:
Holding while testing this out.

runcom:
> Holding while testing this out.

looks like it's doing what I thought, testing with:

- `oc debug node/<worker>`
- edit `/etc/crio/crio.conf.d/00-default` and add a dummy comment
- check MCD degraded because of the above
- touch the file, reboot the MCD
- watch it recover to the correct file

Tha above works when current==desired but the force check in this PR is valid for situations where there's a pendingConfig as well (and an ostree mismatch like the BZ we debugged)

runcom:
testing this out, I've realized we proceed to reconcile two machine configs that have no diff 🤦‍♂️ I'll work something out for that as a follow up and/or a general fix

cgwalters:
May be simpler to understand if we just do:

```
if force {
	return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
}
```
or so?

runcom:
> May be simpler to understand if we just do:

definitely!

runcom:
> Makes sense. If I'm reading this correctly `triggerUpdateWithMachineConfig` does an update regardless of `currentConfig == desiredConfig` right? Which... I guess is behaviour we want

yep, that's what we intended this force file to be (I guess)

/retest

PR #1892, opened 2020-07-02
[wip] DEBUG add atop monitor unit

hexfusion:
use atop to understand pivot writes during an upgrade.

### DIFF
--- a/templates/common/_base/units/atop-monitor.service
+++ b/templates/common/_base/units/atop-monitor.service
@@ -1,0 +1,17 @@
+name: atop-monitor.service
+enabled: true
+contents: |
+  [Unit]
+  Description=atop performance monitor
+  [Service]
+  Type=simple
+  Restart=always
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh, that's where the question about upgrade iops came from.

So ostree is definitely going to do a lot of IO quickly (it does multithreaded writes) and could probably saturate the queue.  Hmm, we could probably use `IOSchedulingClass=idle` by default.

If you want to gather stats on this it's probably easier to use a container like this and just directly run e.g. `pivot quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:8bc8fa4fc8494035e43e359702bdcd2b220e924c585f3c5d5b0e51d1c3bf2846` or so from inside an `oc debug node/` or `ssh`.

hexfusion:
thanks for the ideas @cgwalters !

PR #1893, opened 2020-07-02
"Bug 1823950: [baremetal] Don't failover api vip if loadbalanced endpoint is responding"

cybertron:
We're having some issues in ci with api flakiness that appear to be
related to VIP failover. In order to reduce the need to do failovers
during normal circumstances, we want to leave the VIP alone as long
as the loadbalanced api endpoint on the node is responding. Currently
if the local api service on a node stops responding it will trigger
a failover. This is unnecessary since in most cases haproxy will
continue to distribute the traffic while the local api restarts.

In order to get this behavior, the chk_ocp vrrp_script is split in
two. The reason for this is that we still want to handle the case
where all haproxy instances in the cluster go down, but at least
one api service is still functional. One check looks at the haproxy
endpoint only. This one has a higher weight since it's the preferred
situation. The other check looks for either the haproxy endpoint or
the local endpoint. This means that if haproxy is up then both
checks will succeed and the node will have maximum priority regardless
of the state of its local api. However, if haproxy goes down but
the local api is still working then at least the priority will be
higher than the minimum because the local api is responding.

It was also necessary to add a check that the haproxy firewall rule
is in place. It does us no good to have the loadbalancer working if
traffic isn't being routed to it.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -30,6 +30,29 @@ contents:
       - name: chroot-host
         hostPath:
           path: "/"
+      initContainers:
+      - name: render-config-keepalived
+        image: {{ .Images.baremetalRuntimeCfgImage }}
+        command:
+        - runtimecfg
+        - render
+        - "/etc/kubernetes/kubeconfig"
+        - "--api-vip"
+        - "{{ .Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP }}"
+        - "--ingress-vip"
+        - "{{ .Infra.Status.PlatformStatus.BareMetal.IngressIP }}"
+        - "/config"
+        - "--out-dir"
+        - "/etc/keepalived"
+        resources: {}
+        volumeMounts:
+        - name: kubeconfig
+          mountPath: "/etc/kubernetes"
+        - name: resource-dir
+          mountPath: "/config"
+        - name: conf-dir
+          mountPath: "/etc/keepalived"
+        imagePullPolicy: IfNotPresent
       containers:
       - name: keepalived
         securityContext:
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -27,6 +27,9 @@ contents:
           path: "/etc/keepalived"
       - name: run-dir
         empty-dir: {}
+      - name: chroot-host
+        hostPath:
+          path: "/"
       containers:
       - name: keepalived
         securityContext:
@@ -120,6 +123,8 @@ contents:
           mountPath: "/etc/keepalived"
         - name: run-dir
           mountPath: "/var/run/keepalived"
+        - name: chroot-host
+          mountPath: "/host"
         readinessProbe:
           httpGet:
             path: /readyz
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -16,6 +16,9 @@ contents:
       - name: resource-dir
         hostPath:
           path: "/etc/kubernetes/static-pod-resources/keepalived"
+      - name: script-dir
+        hostPath:
+          path: "/etc/kubernetes/static-pod-resources/keepalived/scripts"
       - name: kubeconfig
         hostPath:
           path: "/etc/kubernetes"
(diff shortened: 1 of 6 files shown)


cybertron:
/hold

Depends on https://github.com/openshift/baremetal-runtimecfg/pull/70

templates/master/00-master/baremetal/files/baremetal-haproxy-haproxy.yaml
```diff
@@ -20,7 +20,7 @@ contents:
     listen health_check_http_url
       bind :::50936 v4v6
       mode http
-      monitor-uri /readyz
+      monitor-uri /haproxy_ready
```
celebdor, line 23: what checks this endpoint?

cybertron: The liveness probe for haproxy.

cybertron:
This is also going to need a rebase to pull in the changes in https://github.com/openshift/machine-config-operator/pull/1909

I may retarget this for 4.5 since #1909 means anything after that won't backport cleanly (unless we end up needing to backport that too, which is possible).

In any case, it will need to be reworked.

pkg/operator/bootstrap.go
```diff
@@ -205,7 +205,7 @@ func RenderBootstrap(
 		if err := os.MkdirAll(dirname, 0655); err != nil {
 			return err
 		}
-		if err := ioutil.WriteFile(path, b, 0655); err != nil {
```
cybertron, line 208: Need clarification on whether there was a reason for this file mode. It seems weird to set the executable bit for only group and other. I guess we'll see what ci says.

I had to change this because keepalived won't run with an executable config file, and I needed to maintain permissions when baremetal-runtimecfg templates files since we're using external scripts for keepalived healthchecks now.

cybertron: Sounds like this is fine, but I had to move it to its own PR because it creates a circular dependency with the baremetal-runtimecfg PR.

https://github.com/openshift/machine-config-operator/pull/1934

kikisdeliveryservice:
do we expect e2e-metal-ipi to pass with this PR?

just gonna hit 
/retest

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -24,6 +24,32 @@ contents:
           path: "/etc/keepalived"
       - name: run-dir
         empty-dir: {}
+      - name: chroot-host
+        hostPath:
+          path: "/"
+      initContainers:
```
yboaron, line 30: Why do we need this init container?

I think that this change will break the unicast keepalived deployment.
In unicast mode the keepalived conf should be rendered in a synchronized manner (we use etcd for that purpose), otherwise, each master will render first a cfg file with empty **unicast_peers** section, so we'll end up with multiple nodes holding the VIP.

cybertron: This was added to template the check scripts. The monitor only regenerates the keepalived.conf file, but to make the check ports dynamic I had to template the scripts as well. I'll have to look at whether we can put the scripts in a separate location so we can template only them. Looks like I need to rebase this anyway so I'll look at it then.

cybertron:
> do we expect e2e-metal-ipi to pass with this PR?

Not until https://github.com/openshift/baremetal-runtimecfg/pull/70 merges and we have a new image with it. e2e-metal-ipi is just broken right now too, but that's unrelated. :-/

templates/common/baremetal/files/baremetal-keepalived.yaml
cybertron: I tried the current version of this with unicast enabled and it worked correctly as far as I can tell. It's only templating the scripts in the init container, so it shouldn't affect the unicast configuration.

yboaron:
I checked this PR locally together with https://github.com/openshift/machine-config-operator/pull/1943 , and I think we have a problem here.I can see that for a period of time both master-1 and the bootstrap node hold the API-VIP

cybertron:
/retest
/hold cancel

I think all the dependencies have merged now. Let's see what ci says.

cybertron:
/retest

I don't know if those jobs are supposed to be passing, but let's give them one more try.

cybertron:
Of the two failing jobs, one doesn't appear to be running on this repo anymore and the other is consistently red.

PR #1895, opened 2020-07-02
pkg/daemon: unexport pkg functions/structs

runcom:
No need for these to be exported out of the package since the package
itself is the only user.

### DIFF
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -76,12 +76,12 @@ func TestReconcilable(t *testing.T) {
 	// newConfig is the config that is being requested to apply to the system
 	newConfig := helpers.CreateMachineConfigFromIgnition(newIgnCfg)
 	// Verify Ignition version mismatch react as expected
-	_, isReconcilable := Reconcilable(oldConfig, newConfig)
+	_, isReconcilable := reconcilable(oldConfig, newConfig)
 	checkIrreconcilableResults(t, "Ignition", isReconcilable)
 	//reset to proper Ignition version
 	newIgnCfg.Ignition.Version = igntypes.MaxVersion.String()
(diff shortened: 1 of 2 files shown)


pkg/daemon/update_test.go
```diff
@@ -161,30 +161,30 @@ func TestReconcilable(t *testing.T) {
...
 	oldIgnCfg := ctrlcommon.NewIgnConfig()
 	oldConfig := helpers.CreateMachineConfigFromIgnition(oldIgnCfg)
 	oldConfig.ObjectMeta = metav1.ObjectMeta{Name: "oldconfig"}
 	newIgnCfg := ctrlcommon.NewIgnConfig()
 	newConfig := helpers.CreateMachineConfigFromIgnition(newIgnCfg)
 	newConfig.ObjectMeta = metav1.ObjectMeta{Name: "newconfig"}
-	diff, err := NewMachineConfigDiff(oldConfig, newConfig)
+	diff, err := newMachineConfigDiff(oldConfig, newConfig)
 	assert.Nil(t, err)
 	assert.True(t, diff.IsEmpty())
```
LorbusChris, line 184: missed in `isEmpty` in this func

pkg/daemon/update_test.go
```diff
@@ -64,7 +64,7 @@ func TestUpdateOS(t *testing.T) {
 
 // TestReconcilable attempts to verify the conditions in which configs would and would not be
 // reconcilable. Welcome to the longest unittest you've ever read.
-func TestReconcilable(t *testing.T) {
+func Testreconcilable(t *testing.T) {
```
LorbusChris, line 67: nit: You might want to leave this func and `testmachineConfigDiff` in upper case (if not, this func's comment should be adapted)

PR #1898, opened 2020-07-04
[fcos] templates/common/_base/units: disable docker

vrutkovs:
`docker` service is not used and takes ~150MB of RAM on all nodes

Fixes https://github.com/openshift/okd/issues/243

### DIFF
--- a/templates/common/_base/units/docker.socket.yml
+++ b/templates/common/_base/units/docker.socket.yml
@@ -1,0 +1,6 @@
+name: docker.socket
+dropins:
+- name: mco-disabled.conf
+  contents: |
+    [Unit]
+    ConditionPathExists=/enoent


LorbusChris:
I’m all for disabling docker, but this looks like a real hacky way of doing it, so I think this should be noted in the commit message. (Properly disabling services is blocked by an ignition/systemd issue <@LorbusChris insert links here once back at desktop>)

vrutkovs:
OVN failed to rollout on one node

/test e2e-aws

cgwalters:
Let's put this on `docker.socket` instead - if you do it on the `.service`, the socket will still exist, it will just fail to work, which might confuse clients.

cgwalters:
(Also FWIW I don't see any harm to having this on the master branch too, but we can deal with that when we get to merging the `fcos` branch)

vrutkovs:
/override ci/prow/e2e-aws

unrelated test failures

PR #1899, opened 2020-07-06
Bug 1853890: pkg/daemon: preserve kargs order

runcom:
Basically what the title says, ~~allow args with same key value and~~ preserve the order we get the args from the MC. Not fully sure if rpm-ostree does some magic there but this should avoid random ranging over a map.

Note: we lost the beauty of checking in O(1) using a map but this isn't gonna be a big deal as the kargs array shouldn't be thousands of elements afaict.

On a second thought, I think we were allowing duplicates already, I'm gonna triple check that.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -650,14 +650,23 @@ func splitKernelArguments(args string) []string {
 
 // parseKernelArguments separates out kargs from each entry and returns it as a map for
 // easy comparison
-func parseKernelArguments(kargs []string) map[string]bool {
-	parsed := make(map[string]bool)
+func parseKernelArguments(kargs []string) []string {
+	parsed := []string{}
 	for _, k := range kargs {
 		for _, arg := range splitKernelArguments(k) {
-			parsed[strings.TrimSpace(arg)] = true
+			parsed = append(parsed, strings.TrimSpace(arg))
 		}
 	}
 	return parsed
+}
+
+func inArray(elem string, array []string) bool {
+	for _, k := range array {
+		if k == elem {
+			return true
+		}
+	}
+	return false
 }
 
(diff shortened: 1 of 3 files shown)


runcom:
I wanna test this out wrt multiple same keys with different values

/hold

runcom:
Ok, this seems to have worked as expected and preserves order as well (adding a further check to allow duplicates)
```
# I0706 13:18:52.184871    2100 update.go:1478] Running rpm-ostree [kargs --append=skew_tick=1 --append=nohz=on --append=rcu_nocbs=1-3 --append=tuned.non_isolcpus=00000001 --append=intel_pstate=disable --append=nosoftlockup --append=tsc=nowatchdog --append=intel_iommu=on --append=iommu=pt --append=systemd.cpu_affinity=0 --append=default_hugepagesz=1G --append=test=foo --append=test=bar --append=nmi_watchdog=0 --append=audit=0 --append=mce=off --append=processor.max_cstate=1 --append=idle=poll --append=intel_idle.max_cstate=0]
```

```
kind: MachineConfig
metadata:
  labels:
  name: 50-examplecorp-kargs
spec:
  kernelArguments:
    - "skew_tick=1"
    - "nohz=on"
    - "rcu_nocbs=1-3"
    - "tuned.non_isolcpus=00000001"
    - "intel_pstate=disable"
    - "nosoftlockup"
    - "tsc=nowatchdog"
    - "intel_iommu=on"
    - "iommu=pt"
    - "systemd.cpu_affinity=0"
    - "default_hugepagesz=1G"
    - "test=foo"
    - "test=bar"
    - "nmi_watchdog=0"
    - "audit=0"
    - "mce=off"
    - "processor.max_cstate=1"
    - "idle=poll"
    - "intel_idle.max_cstate=0"
```

sinnykumari:
makes sense.
/lgtm

PR #1866, opened 2020-06-23
Bug 1835233: Add support for configuring registries at install

umohnani8:
Fixes #1835233 (https://bugzilla.redhat.com/show_bug.cgi?id=1835233)
Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

**- What I did**
Add support to bootstrap to configure the registries in the
cluster-wide Image CR. Ensure that the MCC can consume the
Image CR and generate /etc/containers/registries.conf correctly.

**- How to verify it**
1. Run `openshift-install create manifests`
2. Add the following Image CR yaml to the manifests directory created from step 1
```
apiVersion: config.openshift.io/v1
kind: Image
metadata:
  creationTimestamp: null
  name: cluster
spec:
  registrySources:
    insecureRegistries:
    - blah.io
    - foo.io
```
3. Run `openshift-install create cluster`
4. Watch to ensure that the cluster was successfully installed and check the `/etc/containers/registries.conf` file on all the nodes to see that it has the configuration from the Image CR in step 2.

**- Description for the changelog**

Add day 1 support for registries configured in the cluster-wide Image CR

PR #1903, opened 2020-07-06
Add --release-image flag to bootstrap options

umohnani8:
Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1835233

**- What I did**
We want to be able to access the release-image used for
cluster installation so we can avoid setting that registry
as blocked when there is an Image CR at installation. For
this, we need to pass down the release-image info from the
installer to the controller at bootstrap.
This is step 1. After this is merged, will update bootkube.sh
to set the --release-image flag with the information needed.

**- How to verify it**
Same way as mentioned in https://github.com/openshift/machine-config-operator/pull/1866 once that is merged in.

Add --release-image as a bootstrap option flag.

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -32,6 +32,7 @@ var (
 		imagesConfigMapFile       string
 		infraConfigFile           string
 		infraImage                string
+		releaseImage              string
 		keepalivedImage           string
 		kubeCAFile                string
 		kubeClientAgentImage      string
@@ -68,6 +69,8 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.clusterEtcdOperatorImage, "cluster-etcd-operator-image", "", "Image for Cluster Etcd Operator. An empty string here means the cluster boots without CEO.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraImage, "infra-image", "", "Image for Infra Containers.")
 	bootstrapCmd.MarkFlagRequired("infra-image")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.releaseImage, "release-image", "", "Release image used for cluster installation.")
+	// bootstrapCmd.MarkFlagRequired("release-image") - will uncomment this after updating bootkube.sh to use the --release-image flag
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.configFile, "config-file", "", "ClusterConfig ConfigMap file.")
 	bootstrapCmd.MarkFlagRequired("config-file")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.infraConfigFile, "infra-config-file", "/assets/manifests/cluster-infrastructure-02-config.yml", "File containing infrastructure.config.openshift.io manifest.")


pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -96,6 +96,9 @@ type ControllerConfigSpec struct {
 	// Nobody is also changing this once the cluster is up and running the first time, so, disallow
 	// regeneration if this changes.
 	NetworkType string `json:"networkType,omitempty"`
+
+	// ReleaseImage is the image used for cluster installation
+	ReleaseImage string `json:"releaseImage"`
```
runcom, line 101: since we're not setting this - nor using it, I think the changes in the bootstrap file are what we need till we'll be able to play with the image name itlsef

pkg/operator/sync.go
```diff
@@ -267,6 +267,7 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 	spec.RootCAData = bundle
 	spec.PullSecret = &corev1.ObjectReference{Namespace: "openshift-config", Name: "pull-secret"}
 	spec.OSImageURL = imgs.MachineOSContent
+	spec.ReleaseImage = imgs.ReleaseImage
```
runcom, line 270: following offline discussions - this isn't needed as we'll have this from the CVO crd

pkg/operator/bootstrap.go
```diff
@@ -142,6 +142,7 @@ func RenderBootstrap(
 	spec.RootCAData = bundle
 	spec.PullSecret = nil
 	spec.OSImageURL = imgs.MachineOSContent
+	spec.ReleaseImage = imgs.ReleaseImage
```
runcom, line 145: this and pkg/operator/images.go changes can be dropped for now

cmd/machine-config-operator/bootstrap.go
```diff
@@ -111,6 +114,7 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 			MdnsPublisher:       bootstrapOpts.mdnsPublisherImage,
 			Haproxy:             bootstrapOpts.haproxyImage,
 			BaremetalRuntimeCfg: bootstrapOpts.baremetalRuntimeCfgImage,
+			ReleaseImage:        bootstrapOpts.releaseImage,
```
runcom, line 117: this isn't an "image" as we mean about the others here (I get the confusion tho :( ) - we'll need to pass this down to RenderBootstrap directly - so, I think this PR is enough by just having the dummy flag added and we can proceed once we have a meaningful value there

runcom: let's get this in w/o changing this also :) this is ok for no 👍

umohnani8: Fixed as well :)

ashcrow:
Let's make sure this stuff is documented as well ... but :+1:

runcom:
I want to get a conversation going with @abhinavdahiya @wking maybe to understand how the CVO can render a manifest with status set with the correct release image payload so we can use it the same way as we do with other manifests. The thing is basically we don't have a CVO manifest with Status and release image set at bootstrap time (we do have the correct object later on ofc like we do in https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/container-runtime-config/container_runtime_config_controller.go#L632-L642)

So question is: at MCO/CVO bootstrap in bootkube - can the CVO drop a manifest clusterversion/version which also contains the release image name (as it would do when in the cluster) so we can use it early at bootstrap time? We need this to allow day1 image registries configuration. This PR/way of going is "hacky" compared to what we do today with stuff like network and proxy where a manifest _is_ available.

kikisdeliveryservice:
scaleup-rhel7 & metal-ipi are broken

/skip

umohnani8:
/hold cancel
Got a green light to go ahead with this approach for now.

PR #1904, opened 2020-07-06
Bug 1854160: server: serve v2 if no user-agent is defined

yuqi-zhang:
To not break existing behaviour, if no user-agent is defined in
the request, just serve spec v2 as we used to.

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -173,6 +173,12 @@ type healthHandler struct{}
 // or an error for unsuported user agents.
 // Only Ignition versions that come with support for either spec v2.2 or spec v3.1 are supported
 func deriveSpecVersionFromUseragent(useragent string) (*semver.Version, error) {
+	// for now, serve v2 if we receive a request without a useragent header with Ignition
+	// this happens if the user pings the endpoint directly (e.g. with curl)
+	// and we don't want to break existing behaviour
+	if !strings.HasPrefix(useragent, "Ignition/") {
+		return semver.New("2.2.0"), nil
+	}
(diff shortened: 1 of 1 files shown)


pkg/server/api.go
```diff
@@ -173,6 +173,12 @@ type healthHandler struct{}
 // or an error for unsuported user agents.
 // Only Ignition versions that come with support for either spec v2.2 or spec v3.1 are supported
 func deriveSpecVersionFromUseragent(useragent string) (*semver.Version, error) {
+	// for now, serve v2 if we receive a request without a useragent header
+	// this happens if the user pings the endpoint directly (e.g. with curl)
+	// and we don't want to break existing behaviour
+	if useragent == "" {
```
cgwalters, line 179: `curl` and most well behaved clients set a `User-Agent` by default.

```
curl -L https://httpbin.org/get
{
  "args": {}, 
  "headers": {
    "Accept": "*/*", 
    "Host": "httpbin.org", 
    "User-Agent": "curl/7.66.0", 
    "X-Amzn-Trace-Id": "Root=1-5f03680f-2484bc4082ab921c451ee35f"
  }, 
  "origin": "24.61.188.223", 
  "url": "https://httpbin.org/get"
}
```

So we really need to do this for non-Ignition units.

cgwalters: IOW something more like
```
if !strings.StartsWith(useragent, "Ignition/") {
   return "2.2.0"
}
```

yuqi-zhang: Fixed, thanks!

PR #1909, opened 2020-07-07
Bug 1854249: [On-Prem] Workaround issues with keepalived v2.0.10

mandre:
**- What I did**

This commit extracts the chk_ocp script into a separate file and wraps
the check scripts with the timeout command to workaround
acassen/keepalived#1364.

This PR supersedes #1908.

Co-Authored-By: Yossi Boaron <yboaron@redhat.com>
Co-Authored-By: Ben Nemec <bnemec@redhat.com>

**- How to verify it**

Installer can deploy OCP on on-prem platform again.

Wrap keepalived checks with timeout to workaround known issue with keepalived v2.0.10.

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
@@ -3,7 +3,7 @@ path: "/etc/kubernetes/static-pod-resources/keepalived/keepalived.conf.tmpl"
 contents:
   inline: |
     vrrp_script chk_ocp {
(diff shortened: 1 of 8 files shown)


Fedosin:
Am I right that the command is the same, but you just created a one-liner script instead of calling it directly?

kikisdeliveryservice:
Q: do we expect vsphere job to pass with this PR @mandre ?

cybertron:
/lgtm

This worked for me locally. I think the flapping of the VIP will be helped by https://github.com/openshift/machine-config-operator/pull/1893 which adds rise and fall values to the check so they won't immediately cause failover because of a single timeout. This should get us back to where we were though.

I think @patrickdillon is the vsphere contact, so maybe he can comment on the status of the vsphere job.

Issue #507, opened 2019-02-27
docs: Provide Machine Config Spec


kikisdeliveryservice:
**Description**

We should add a full machine config spec to the documentation showing the definitively supported, optional and required fields, etc.. This will go a long way to reducing users from submitting bad machine configs and provides one place to explicitly and clearly state what's in a valid machine config. 

Some ideas for format: 
https://github.com/coreos/container-linux-config-transpiler/blob/master/doc/configuration.md
https://github.com/coreos/ignition/blob/master/doc/configuration-v3_0-experimental.md

kikisdeliveryservice:
Add: not documenting this makes us rely solely on Ignitions documentation, which is confusing since we don't even support all of the config fields.

I think for 4.1 we should start doing this.

PR #1910, opened 2020-07-07
docs: more details + examples for Ignition spec support

miabbott:
Users could benefit for some more clear details and examples about how
to use the Ignition spec in their MachineConfigs.  This attempts to
codify the parts of the Ignition spec that are supported and provide
sample MachineConfigs.

### DIFF
--- a/docs/IgnitionSupport.md
+++ b/docs/IgnitionSupport.md
@@ -1,0 +1,334 @@
+# Ignition Support in MachineConfigs
+
+`MachineConfig` objects are the defacto way of customizing the nodes in an
+OpenShift cluster. Customizations can be applied "day 1" during the install of
+the cluster or "day 2" after the cluster has been installed.
+
+The way the customizations are defined are mostly through the use of the
+Ignition specification. However, not all of the specification is supported.
+This attempts to codify which parts of the [Ignition 3.2.0 specification](https://github.com/coreos/ignition/blob/master/docs/configuration-v3_2.md)
+are supported and provide examples for **day 2** customizations.
+
+---
+**WARNING** These examples are valid as of OpenShift 4.7; they are not guaranteed to
+work with older or newer versions of OpenShift.  It is recommended that users
+reference the official [OpenShift documentation](https://docs.openshift.com/container-platform/4.7/post_installation_configuration/machine-configuration-tasks.html)
+for up to date information about how to use `MachineConfig` objects.
+
+---
+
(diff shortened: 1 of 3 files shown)


miabbott:
This partially addresses #507

Also, big thanks to @yuqi-zhang for helpfully answering all of my questions around this topic and catching all of my user errors along the way.

yuqi-zhang:
In general I think the doc direction is good, there is a couple of clarification points I'd like to make though:

1. since we are in the middle of the spec 3 transition, perhaps it would be good to consider using spec 3 as the default doc
2. there's a distinction between what MCs are supported day 1 vs day 2. We only allow **updates** to files/units/sshkeys via the MCO but there is no such strict restrictions during provisioning

kikisdeliveryservice:
> In general I think the doc direction is good, there is a couple of clarification points I'd like to make though:
> 
>     1. since we are in the middle of the spec 3 transition, perhaps it would be good to consider using spec 3 as the default doc
> 
>     2. there's a distinction between what MCs are supported day 1 vs day 2. We only allow **updates** to files/units/sshkeys via the MCO but there is no such strict restrictions during provisioning

Jinx.. Agree with Jerry on both points.

miabbott:
> In general I think the doc direction is good, there is a couple of clarification points I'd like to make though:
> 
>     1. since we are in the middle of the spec 3 transition, perhaps it would be good to consider using spec 3 as the default doc

Agreed; it was easier for me to spin up a cluster with spec v2, so that's how I got here.  I think *most* of the existing content should translate, but I can follow up.

>     2. there's a distinction between what MCs are supported day 1 vs day 2. We only allow **updates** to files/units/sshkeys via the MCO but there is no such strict restrictions during provisioning

Excellent distinction.  I will add wording to scope this to day 2.

(Though this was a hack day effort, so this PR might sit stale for a while)

kikisdeliveryservice:
just adding a hold for updates to the docs as per comments above

/hold

miabbott:
Well, I didn't make as much progress on this as I would have liked, but I at least updated the examples to 3.1.0 and verified they work.

miabbott:
/remove-lifecycle rotten
/assign @kikisdeliveryservice 

Hey folks, I finally came back to this and updated it for Ignition spec v3.2.0 and OCP 4.7.

Let me know if y'all are still interested in landing this or if we should just close it in favor of the official docs.

PR #1914, opened 2020-07-08
Bug 1853584: templates: node hostnames: ensure RHCOS hostnames are mostly valid

darkmuggle:
This aims to fix all the edge-cases around invalid hostnames. This can
include:
 - localhost as the hostname

Previously, there was a node-valid-hostname.service, but it only ensured
that the hostname was not localhost. When running in the initramfs,
NetworkManager will truncate the hostname at 64 characters resulting in
the `node-valid-hostname.service` checks succeeding (!localhost) but
still invalid for the cluster. This change consolidates the logic used by
the dispatcher script _and_ the systemd unit into a library script that
is used by both.

This fix is not ideal, but it seems like the best path that be
backported to 4.3 and 4.5.

Fixes BZs 1844613, 1845885, 1853400, 1853584 (and probably more)

### DIFF
--- a/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
+++ b/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
@@ -1,0 +1,83 @@
+mode: 0755
+path: "/usr/local/sbin/set-valid-hostname.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # On some platforms the hostname may be too long (>63 chars).
+    #  - On firstboot the hostname is set in the initramfs before NetworkManager
+    #    And it may be truncated at 64 characters (too long)
+    #  - On reboot affect nodes use 'localhost'.
+    #
+    # This script is a simple workaround for hostname woes, including
+    #  - NOT a localhost name
+    #  - NOT longer than 63 characters. Names will be truncated at the
+    #    first dot, and then capped at 63 char (which ever is less).
+    #  - Race conditions between truncated hostnames by the dhclient
+    #    and NetworkManager.
+    #
+    # Finally, this script is invoked via:
+    #  - /etc/NetworkManager/dispatcher.d/90-long-hostnames
+    #  - on boot via node-valid-hostname.service
+
+    export PATH="/usr/bin:/usr/local/bin:/sbin:/usr/local/sbin:/bin:${PATH}"
+    log() { logger --tag "$(basename $0)" "${@}"; }
+
+    # wait_localhost waits until the host gets a real hostname.
+    # This will wait indefinately. node-valid-hostname.service will terminate
+    # this after 5m.
+    wait_localhost() {
+        log "waiting for non-localhost hostname to be assigned"
+        while [[ "$(< /proc/sys/kernel/hostname)" =~ (localhost|localhost.localdomain) ]];
+        do
+            sleep 1
+        done
+        log "node identified as $(</proc/sys/kernel/hostname)"
+        exit 0
+    }
+
+    set_valid_hostname() {
+        local host_name=${1}
+        local type_arg="transient"
+
+        # /etc/hostname is used for static hostnames and is authorative.
+        # This will check to make sure that the static hostname is the
+        # less than or equal to 63 characters in length.
+        if [ -f /etc/hostname ] && [ "$(cat /etc/hostname | wc -m)" -gt 0 ]; then
+            etc_name="$(< /etc/hostname)"
+            type_arg="static"
+            if [ "${etc_name}" != "${host_name}" ]; then
+                log "/etc/hostname is set to ${etc_name} but does not match ${host_name}"
+                log "using /etc/hostname as the authorative name"
+                host_name="${etc_name}"
+            fi
+        fi
+
+        # Only mutate the hostname if the length is longer than 63 characters. The
+        # hostname will be the lesser of 63 characters after the first dot in the
+        # FQDN.
+        if [ "${#host_name}" -gt 63 ]; then
+            alt_name=$(printf "${host_name}" | cut -f1 -d'.' | cut -c -63)
+            log "${host_name} is longer than 63 characters, using trunacated hostname"
+            host_name="${alt_name}"
+        fi
+        log "setting ${type_arg} hostname to ${host_name}"
+        /bin/hostnamectl "--${type_arg}" set-hostname "${host_name}"
+        exit 0
+    }
+
+    cli_run() {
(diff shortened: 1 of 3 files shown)


darkmuggle:
I ran this through the paces attempting to re-create all the Linux valid but invalid cluster names. If/when the PR is backported, it should fix 4.3 and 4.5 GCP, race conditions on initramfs-truncated hostnames and kublet startup, the CI problems with long GCP hostnames, and close 6 or so bugs. The goal is of this is to make the customer, engineer and CI pain end regarding hostnames.

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,82 @@
...
+    #!/bin/bash
+    # On some platforms the hostname may be too long (>63 chars).
+    #  - On firstboot the hostname is set in the initramfs before NetworkManager
+    #    And it may be truncated at 64 characters (too long)
+    #  - On reboot affect nodes use 'localhost'.
+    #
+    # This script is a simple workaround for hostname woes, including
+    #  - NOT a localhost name
+    #  - NOT longer than 63 characters. Names will be truncated at the
+    #    first dot, and then capped at 63 char (which ever it less).
```
miabbott, line 14: nit:  `which ever is less`

darkmuggle: indeed, fixed.

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,82 @@
...
+        exit 0
+    }
+
+    cli_run() {
+        mode="${1:?mode must be the first argument}"; shift;
+        case "${mode}" in
+                wait_localhost) wait_localhost;;
+            set_valid_hostname) hname="${1:?hostname is a required last argument}";
+                                set_valid_hostname "${hname}";;
+                            *) log "unknown mode ${mode}"; exit 10;;
```
miabbott, line 73: I'm not familiar with the use of `exit 10`; could you explain why it is used over `exit 1` for example?

darkmuggle: Ah, that was a typo. It should have been `exit 1`. However, any exit code between 1 and 255 is valid for a general error so long as its not reserved (http://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF). A `10` would but a custom code. `man curl` shows how it uses custom exist codes.

miabbott: Ah ok, this matched my understanding of exit codes in general.  Just wanted to be sure you were using a custom code over standard `1`

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+        # /etc/hostname is used for static hostnames and is authorative.
+        # This will check to make sure that the static hostname is the
+        # less than or equal to 63 characters in length.
+        if [ -f /etc/hostname ] && [ "$(cat /etc/hostname | wc -m)" -gt 0 ]; then
+            etc_name="$(< /etc/hostname)"
+            type_arg="static"
+            if [ "${etc_name}" != "${host_name}" ]; then
+                log "/etc/hostname is set to ${etc_name} but does not match ${host_name}"
+                log "using /etc/hostname as the authorative name"
```
miabbott, line 50: Should `host_name` be set to the value of `etc_name` here?  Otherwise, when `hostnamectl` is called, it looks you are going to end up using the hostname argument passed to `set_valid_hostname`

ashcrow:
e2e-aws failure looks unrelated:
```
"Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:acquiring_lease:aws-quota-slice\"
```

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
darkmuggle: Good catch, fixed.

darkmuggle:
Checking GCP shows expected output:
```
Jul 09 16:15:17.707393 localhost systemd[1]: Started Network Manager Wait Online.
Jul 09 16:15:17.734274 localhost 90-long-hostname[1422]: ci-op-j57n6ly4-1354f-dcfrz-master-0.c.openshift-gce-devel-ci.internal is longer than 63 characters, using trunacated hostname
Jul 09 16:15:17.737135 localhost 90-long-hostname[1424]: setting transient hostname to ci-op-j57n6ly4-1354f-dcfrz-master-0
Jul 09 16:15:17.744589 ci-op-j57n6ly4-1354f-dcfrz-master-0 systemd-hostnamed[1355]: Changed host name to 'ci-op-j57n6ly4-1354f-dcfrz-master-0'
Jul 09 16:15:18.673774 ci-op-j57n6ly4-1354f-dcfrz-master-0 bash[1431]: node identified as ci-op-j57n6ly4-1354f-dcfrz-master-0
```

We still see the

kikisdeliveryservice:
@abhinavdahiya @cgwalters PTAL

kikisdeliveryservice:
@darkmuggle if this is going to be backported you will have to have a BZ in this master PR so pick one out of your list :smile:

darkmuggle:
I picked 1853584 as the "lead bug" for the backporting.

cgwalters:
>  When running in the initramfs, NetworkManager will truncate the hostname at 64 characters resulting in the node-valid-hostname.service checks succeeding (!localhost) but still invalid for the cluster. 

Right; in the initramfs we haven't applied the Ignition config that contains these DHCP hooks yet.  I think longer term we want NM to do the "right thing" out of the box rather than installing these dispatcher scripts.   What's less clear to me is the implemetation of that because I think this "truncate hostname" path is somewhat GCP-specific right?  (I still don't understand why GCP's DHCP server does this).  But since in CoreOS we know our platform via the kernel argument, I could imagine we wire things up so we have some small conditional that's "if on GCP, --truncate-hostname-short" or so.  Or maybe we could do it unconditionally.

Hmm but, don't we tear down networking we set up in the initramfs?  Though I guess we might not be "tearing down" the hostname which is probably a bug. 

Also, we always do the OS upgrade and reboot before anything kubelet is involved, which should also as a side effect ensure that we *do* use these dispatcher scripts when kubelet is starting.  So I don't quite understand the reference to the initramfs.

Anyways
/lgtm

PR #1918, opened 2020-07-09
Bug 1842603: Fix runLoginMonitor

runcom:
_This PR contains the following commits:_

__pkg/daemon: skip runLoginMonitor if node isn't set__

This whole function runs too early for the node object to be set in the daemon struct, resulting
in:

```
W0709 14:35:40.458777  237119 daemon.go:556] Got an error from auxiliary tools: error: cannot apply annotation for SSH access due to: unable to update node "nil": node "ci-ln-t6mdh42-f76d1-9qqj4-worker-c-l75p7" not found
```

just ignore that, we can't do anything until we have the node set anyway, which we always do eventually

__pkg/daemon: workaround new sessions detection__

First, this may be a journalctl bug (systemd 239 on RHCOS)

Let me explain with a simple reproducer:

- deploy a bastion to be able to SSH to nodes
- jump on a (random) target node with `oc debug` (this won't trigger anything ssh related)
- run `journalctl -f -o cat -b -u systemd-logind.service MESSAGE_ID=8d45620c1a4348dbb17410da57c60c66` (this is what the MCD uses to follow new ssh sessions)
- in another tab, do run an ssh login command (core@...) against that target node
- notice nothing showed up from step 3
- keep ssh'ing and keep noticing nothing shows up in that journalctl command
- now, ctrl^c from step 3, re-run that command (now it has the login sessions we did before)
- in the other tab, do run an ssh login command (core@...) against that target node
- notice the log follow is _only now_ working and streaming the new ssh sessions

If step 3 is stuck, we don't get anything until we restart the MCD __after__ a successful ssh session,
otherwise we'll be perpetually stuck from detecting new sessions.

See BZ https://bugzilla.redhat.com/show_bug.cgi?id=1842603#c9

__Note__

Notice the second commit is a stopgap/workaround - I'm digging the journalctl behavior, if anybody knows more, please let me know :) (everybody dislikes strings matching right?)

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -615,7 +615,7 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 }
 
 func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
-	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "-u", logindUnit, "MESSAGE_ID="+sdMessageSessionStart)
+	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "-u", logindUnit)
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		exitCh <- err
@@ -639,14 +639,24 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 				l, err := stdout.Read(buf)
 				if err != nil {
 					if err == io.EOF {
+						glog.Warning("runLoginMonitor got EOF while reading the journal, exiting")
 						return
 					}
 					exitCh <- err
 					return
 				}
 				if l > 0 {
-					line := strings.Split(string(buf), "\n")[0]
-					glog.Infof("Detected a new login session: %s", line)
+					lines := strings.Split(string(buf), "\n")
+					var newSession bool
+					for _, line := range lines {
+						if strings.HasPrefix(line, "New session ") {
+							newSession = true
+						}
+					}
+					if !newSession {
+						continue
+					}
+					glog.Info("Detected a new login session")
 					glog.Infof("Login access is discouraged! Applying annotation: %v", machineConfigDaemonSSHAccessAnnotationKey)
 					if err := dn.applySSHAccessedAnnotation(); err != nil {
 						exitCh <- err
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -632,6 +632,9 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
 			case <-worker:
 				return
 			default:
+				if dn.node == nil {
+					continue
+				}
 				buf := make([]byte, 1024)
 				l, err := stdout.Read(buf)
 				if err != nil {


runcom:
testing this out (I know how it breaks, not sure if the workaround is valid)

/hold

pkg/daemon/daemon.go
```diff
@@ -615,7 +615,7 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 }
 
 func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
-	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "-u", logindUnit, "MESSAGE_ID="+sdMessageSessionStart)
+	cmd := exec.Command("journalctl", "-b", "-f", "-o", "cat", "-u", logindUnit)
```
runcom, line 618: this effectively removes the filter for "New session .." entries and that's why we check them below. This seems to be working as grabbing the journal for the logind service _always_ has something (not sure if that's 100% valid all the time tho). Here's an example:

```
[root@ci-ln-t6mdh42-f76d1-9qqj4-master-1 /]# journalctl -f -o cat -b -u systemd-logind.service
Starting Login Service...
Watching system buttons on /dev/input/event0 (Power Button)
Watching system buttons on /dev/input/event1 (Sleep Button)
Watching system buttons on /dev/input/event2 (AT Translated Set 2 keyboard)
New seat seat0.
Started Login Service.
New session 1 of user core.
Session 1 logged out. Waiting for processes to exit.
Removed session 1.
New session 3 of user core.
Session 3 logged out. Waiting for processes to exit.
Removed session 3.
New session 5 of user core.
Session 5 logged out. Waiting for processes to exit.
Removed session 5.

```
So, the above is at least guaranteed to keep streaming messages and we can read them

pkg/daemon/daemon.go
```diff
@@ -632,18 +632,31 @@ func (dn *Daemon) runLoginMonitor(stopCh <-chan struct{}, exitCh chan<- error) {
...
+						glog.Warning("runLoginMonitor got EOF while reading the journal, exiting")
 						return
 					}
 					exitCh <- err
 					return
 				}
 				if l > 0 {
-					line := strings.Split(string(buf), "\n")[0]
-					glog.Infof("Detected a new login session: %s", line)
+					lines := strings.Split(string(buf), "\n")
```
ashcrow, line 649: Is iterating over the lines and checking for "New session" faster/more efficient than a regexp check on the buffer (Match with bytes or Find with string)?

ashcrow: Simple example (again -- may not be better or worse)

```go

package main

import (
        "fmt"
        "regexp"
)

var RE = regexp.MustCompile(`New session`)


func newSession(d []byte) {
        fmt.Printf("%t: %s", RE.Match(d), d)
        fmt.Println("\n-")
}


func main() {
        newSession([]byte(`Starting Login Service...
Watching system buttons on /dev/input/event0 (Power Button)
New seat seat0.
New session 5 of user core.
Removed session 5.`))
        newSession([]byte(`Starting Login Service...
Watching system buttons on /dev/input/event0 (Power Button)
Removed session 5.`))
}
```

Output:
```
$ ./main 
true: Starting Login Service...
Watching system buttons on /dev/input/event0 (Power Button)
New seat seat0.
New session 5 of user core.
Removed session 5.
-
false: Starting Login Service...
Watching system buttons on /dev/input/event0 (Power Button)
Removed session 5.
-

```

runcom: Right, uhm, I don’t really know, the regexp has to match “starting with that pattern” also. Not sure, I think we can tweak this with a regexp for sure, I don’t think this buffers grow enough to worry about the for loop over those lines - I think benchmarking this would be great to do, but we don’t have much data to run the benchmark against lol 
I don’t really have a preference tho, I’m ok changing to a regexp that works 👍

ashcrow: Na, keep it as is -- the above can be an option to try if we see a performance hit.

ashcrow:
Flake:
```
"Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:acquiring_lease:aws-quota-slice\""
```

/test e2e-aws

runcom:
> testing this out (I know how it breaks, not sure if the workaround is valid)
> 
> /hold

(tested not working in 4.4, tested it works with this in 4.6, and also tested it doesn't work in 4.6 w/o this PR)

it works as expected and we can now see new sessions as they happen and we'll annotate the node - not optimal I know, but at least it now works. I'll leave this here for further thoughts/solutions.

/hold cancel

cgwalters (commented):
So recently jlebon figured out why monitoring the journal can be racy; it's because clients may only see content when journald syncs to disk.  We started doing this in tests: https://github.com/coreos/fedora-coreos-config/blob/dbaa174e2a0bbe3122359cf44668b6d2e76a1f29/overlay.d/05core/usr/lib/dracut/modules.d/99emergency-timeout/ignition-virtio-dump-journal.sh#L7

But in general our plan of using the journal for *timely* event notifications may be problematic - we may have to e.g. talk DBus to logind.

(I still need to dig into this more because it does seem like I get notifications quickly sometimes too)

runcom:
> But in general our plan of using the journal for _timely_ event notifications may be problematic - we may have to e.g. talk DBus to logind.

we used to use dbus via the coreos/systemd integration for dropped it and rewrote it shelling out to journalctl - I'll take a further look at the golang dbus integration

runcom:
> But in general our plan of using the journal for _timely_ event notifications may be problematic - we may have to e.g. talk DBus to logind.

for this specific case I don' t think we're time sensitive so this PR would just allow the monitor to keep working and notify/label the node. Any reason not to merge this in 4.7?

PR #1919, opened 2020-07-09
templates/common/_base/units: disable Zincati service

LorbusChris:
This disables Zincati updates on FCOS nodes - no-op on RHCOS.

### DIFF
--- a/templates/common/_base/units/zincati.service.yaml
+++ b/templates/common/_base/units/zincati.service.yaml
@@ -1,0 +1,6 @@
+name: zincati.service
+dropins:
+- name: mco-disabled.conf
+  contents: |
+    [Unit]
+    ConditionPathExists=/enoent


cgwalters:
/lgtm
(Personally I would say disable the service entirely, but this is fine too)

LorbusChris:
We should be able to disable the service entirely once https://github.com/coreos/fedora-coreos-tracker/issues/392 / https://github.com/systemd/systemd/pull/15205 is fixed

cgwalters:
> We should be able to disable the service entirely once coreos/fedora-coreos-tracker#392 / systemd/systemd#15205 is fixed

I don't think we need to block on that, see
https://github.com/openshift/machine-config-operator/pull/1907
(Certainly having systemd fixed would be really nice)

LorbusChris:
Oh right, let's do it that way :)
/hold

PR #1921, opened 2020-07-11
Bug 1855821: pkg/controller/render: log actions on machine configs

runcom:
It was a pain to support our claim here https://bugzilla.redhat.com/show_bug.cgi?id=1855821
regarding the fact that _somebody/something_ deleted the relevant machine configs.
These logs shouldn't be spammy, I can't remember why they were behind a 4 log level.
It makes sense to make them more visible to back up our claims on MC actions.

### DIFF
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -165,6 +165,8 @@ func (ctrl *Controller) deleteMachineConfigPool(obj interface{}) {
 
 func (ctrl *Controller) addMachineConfig(obj interface{}) {
 	mc := obj.(*mcfgv1.MachineConfig)
+	glog.Infof("MachineConfig %s added", mc.Name)
+
 	if mc.DeletionTimestamp != nil {
 		ctrl.deleteMachineConfig(mc)
 		return
@@ -173,7 +175,6 @@ func (ctrl *Controller) addMachineConfig(obj interface{}) {
 	controllerRef := metav1.GetControllerOf(mc)
 	if controllerRef != nil {
 		if pool := ctrl.resolveControllerRef(controllerRef); pool != nil {
-			glog.V(4).Infof("MachineConfig %s added", mc.Name)
 			ctrl.enqueueMachineConfigPool(pool)
 			return
 		}
@@ -185,7 +186,6 @@ func (ctrl *Controller) addMachineConfig(obj interface{}) {
 		return
 	}
 
-	glog.V(4).Infof("MachineConfig %s added", mc.Name)
 	for _, p := range pools {
 		ctrl.enqueueMachineConfigPool(p)
 	}
@@ -194,6 +194,8 @@ func (ctrl *Controller) addMachineConfig(obj interface{}) {
 func (ctrl *Controller) updateMachineConfig(old, cur interface{}) {
 	oldMC := old.(*mcfgv1.MachineConfig)
 	curMC := cur.(*mcfgv1.MachineConfig)
+
+	glog.Infof("MachineConfig %s updated", curMC.Name)
 
 	curControllerRef := metav1.GetControllerOf(curMC)
 	oldControllerRef := metav1.GetControllerOf(oldMC)
@@ -205,7 +207,6 @@ func (ctrl *Controller) updateMachineConfig(old, cur interface{}) {
 
 	if curControllerRef != nil {
 		if pool := ctrl.resolveControllerRef(curControllerRef); pool != nil {
-			glog.V(4).Infof("MachineConfig %s updated", curMC.Name)
 			ctrl.enqueueMachineConfigPool(pool)
 			return
 		}
@@ -217,7 +218,6 @@ func (ctrl *Controller) updateMachineConfig(old, cur interface{}) {
 		return
 	}
 
-	glog.V(4).Infof("MachineConfig %s updated", curMC.Name)
 	for _, p := range pools {
 		ctrl.enqueueMachineConfigPool(p)
 	}
@@ -239,10 +239,11 @@ func (ctrl *Controller) deleteMachineConfig(obj interface{}) {
 		}
 	}
 
+	glog.Infof("MachineConfig %s deleted", mc.Name)
+
 	controllerRef := metav1.GetControllerOf(mc)
 	if controllerRef != nil {
 		if pool := ctrl.resolveControllerRef(controllerRef); pool != nil {
-			glog.V(4).Infof("MachineConfig %s deleted", mc.Name)
 			ctrl.enqueueMachineConfigPool(pool)
 			return
 		}
@@ -254,7 +255,6 @@ func (ctrl *Controller) deleteMachineConfig(obj interface{}) {
 		return
 	}
 
-	glog.V(4).Infof("MachineConfig %s deleted", mc.Name)
 	for _, p := range pools {
 		ctrl.enqueueMachineConfigPool(p)
 	}


pkg/controller/render/render_controller.go
```diff
@@ -165,6 +165,8 @@ func (ctrl *Controller) deleteMachineConfigPool(obj interface{}) {
 
 func (ctrl *Controller) addMachineConfig(obj interface{}) {
 	mc := obj.(*mcfgv1.MachineConfig)
+	glog.Infof("MachineConfig %s added", mc.Name)
```
sinnykumari, line 168: Not superconfident but wondering if moving log statement here may lead to false positives logs. Perhaps we can keep the earlier logging while removing verbosity.

pkg/daemon/daemon.go
```diff
@@ -1221,6 +1224,24 @@ func (dn *Daemon) prepUpdateFromCluster() (*mcfgv1.MachineConfig, *mcfgv1.Machin
...
+// the MC isn't there, it'll go and grab it from disk.
+// see https://bugzilla.redhat.com/show_bug.cgi?id=1855821#c2
+func (dn *Daemon) getCurrentConfig(name string) (*mcfgv1.MachineConfig, error) {
+	currentConfig, err := dn.mcLister.Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			currentConfig, err = dn.getCurrentConfigOnDisk()
+			if err != nil && !os.IsNotExist(err) {
+				return nil, err
+			}
```
sinnykumari, line 1237: Perhaps adding a log here in `else` would be useful to help us debug that something may have gone wrong due to which we are fetching the currentconfig from disk

pkg/controller/render/render_controller.go
runcom: I think we won't have this issue as, if we got here we were "informed" that an add action has been performed at the api level (same for update/delete) so it should always be safe to log right?

pkg/daemon/daemon.go
runcom: right, good point, I'll wrap the `err` variables to be more explicit

pkg/controller/render/render_controller.go
sinnykumari: great!

pkg/daemon/daemon.go
```diff
@@ -1194,21 +1199,19 @@ func (dn *Daemon) prepUpdateFromCluster() (*mcfgv1.MachineConfig, *mcfgv1.Machin
...
 		return nil, nil, err
 	}
-
 	currentOnDisk, err := dn.getCurrentConfigOnDisk()
 	if err != nil && !os.IsNotExist(err) {
 		return nil, nil, err
 	}
-
-	if currentOnDisk != nil && currentOnDisk.GetName() != currentConfig.GetName() {
+	if currentOnDisk != nil && currentOnDisk.GetName() != currentConfigName {
```
ashcrow, line 1214: Nit: In the future it may make sense to functionize this check as it happens multiple times.

pkg/daemon/daemon.go
ashcrow: Following on @sinnykumari's logging note I think it would be helpful to log that we've successfully gotten the config from disk (since [dn.getCurrentConfigOnDisk](https://github.com/openshift/machine-config-operator/blob/24e24e8e0bd2dcc79e5c429909a99d7049171bfa/pkg/daemon/daemon.go#L901-L912) doesn't log, understandably so). That way if something was, for some reason out of sync, it would be easier to trace.

cgwalters:
>  otherwise, we're stuck.

One thing I think we should recommend more often is - for MAO managed clusters, simply deleting the machine object and letting the MAO reprovision.  Particularly for workers.  In the future hopefully that will all work for the control plane too.

We could even consider doing that automatically.  But let's just at least keep it in mind as an option.  (And in the future we can more cleanly support "reprovision in place" too with https://github.com/coreos/fedora-coreos-tracker/issues/399 )

runcom:
> One thing I think we should recommend more often is - for MAO managed clusters, simply deleting the machine object and letting the MAO reprovision. Particularly for workers. In the future hopefully that will all work for the control plane too.

right, I think that would be good too - does that block this PR to ease the reconciliation today tho?

cgwalters:
That comment was just an aside, not blocking this.

But now that I did actually look at this beyond just skimming the PR description briefly - are you really sure of this diagnosis?  

/hold
for a reply to this (but feel free to lift if you disagree).

You say:
> ... it means that MC isn't in the cluster anymore and we know MCO doesn't delete it. It also means by the time you've got the must-gather, something else _happened_ in the cluster (likely a machine config cleanup...and that's why the MC isn't found anymore).

Nothing GCs machineconfig objects today - isn't this much more likely to be "bootstrap rendering vs cluster" drift that's happened a few times?

I think the reason we didn't do this before is it will cause an additional reboot uncoordinated by the MCO.  See some related discussion in https://github.com/openshift/machine-config-operator/pull/926

Debugging this is incredibly painful for sure, and I see the temptation to just automatically reconcile.  I think we need to be *very* loud if this ever happens because we really should fix the drift between bootstrap and cluster rendering.

And also I am not sure we should do this until we've addressed the reboot coordination problem.  

(I only spent a few minutes on looking at this, so I may be wrong here)

runcom:
> But now that I did actually look at this beyond just skimming the PR description briefly - are you really sure of this diagnosis?

There's a reproducer here that I've written: https://bugzilla.redhat.com/show_bug.cgi?id=1855821#c3
Also, pretty sure we're not even at the bootstrap vs cluster drift since we're _in cluster_ based on the BZ.

> Nothing GCs machineconfig objects today - isn't this much more likely to be "bootstrap rendering vs cluster" drift that's happened a few times?

So yeah, the drift symptom is clear tho: the bootstrap MC is different than the cluster MC - here, we're in the cluster, no bootstrap anymore, and that's because the _user_ manually deleted that MC - no bootstrap in the picture here.

Also, I don't think this patch is making anything reboot more than it should 🤔 can you explain that given we're not in the bootstrap vs cluster case? Even in that bad drift scenario, how can this cause a reboot? In the drift scenario currentOnDisk != currentConfig|desiredConfig - this case should only cover currentOnDisk == currentConfig to be able to reconciles the only cases where somebody delete an in-use rendered machine config

runcom:
> Nothing GCs machineconfig objects today

I wanna re-state that _this is about somebody manually deleting the rendered config_ (manually deletion is the only guess that makes sense unless etcd starts having consistency issues, which I hope it doesn't)

cgwalters (changes requested):
I'd just like to be *really* really sure this isn't ye olde "bootstrap rendering different" problem.  Or I'd like to see strong evidence that somehow MCs are being deleted.

runcom:
> I'd just like to be really really sure this isn't ye olde "bootstrap rendering different" problem. Or I'd like to see strong evidence that somehow MCs are being deleted.

so, looking at the bz, there's no doubt this isn't bootstrap rendering different. The BZ has a cluster up and running and that was the way to repro the behavior in https://bugzilla.redhat.com/show_bug.cgi?id=1855821#c3

I think one way to move this forward since there should be a manual workaround is to land a fix to log exactly the lifecycle of any machine config so we're able to look at that, compare and debug if it happens again with 4.6.

I've dropped the main commit in favor of just the one that does the enhanced logging and this should be good to go

runcom:
holding retesting as this shouldn't impact upgrades

cgwalters (approved):
Sure, more logging is good by me.  (I think it's possible to dig through the apiserver audit logs for some of this stuff, but TBH I've never done that)

yuqi-zhang:
We've closed the underlying bug as we've determined it is not an actual bug, and we will be having more discussions on the idea of "safety" in the MCO. Correspondingly I will close this PR as it is not high priority. If you feel otherwise feel free to reopen. Thanks!

PR #1922, opened 2020-07-13
Avoid reboots for ICSP changes

beekhof:
**- What I did**

Cleaned up the PoC code for https://github.com/openshift/enhancements/pull/159, replaces PR #1626

**- How to verify it**

e2e and unit tests included

Provide an alternative to node reboots in the specific cases that we know it is safe to do so


pkg/daemon/apply.go
```diff
@@ -0,0 +1,177 @@
...
+	if !reflect.DeepEqual(oldIgnConfig.Storage.Files, newIgnConfig.Storage.Files) {
+		actions := []actionResult{}
+
+		added, removed, modified := getFileChanges(oldIgnConfig.Storage.Files, newIgnConfig.Storage.Files)
+		if len(added) > 0 || len(removed) > 0 {
+			return []actionResult{RebootPostAction{Reason: fmt.Sprintf("Files were added or removed")}}
+		}
+		for _, file := range modified {
+			if file.Node.Path != "/etc/containers/registry.conf" {
+				return []actionResult{RebootPostAction{Reason: fmt.Sprintf("Storage file %v changed", file.Node.Path)}}
```
runcom, line 156: s/Storage/Registy/g perhaps? Also, I think this must be abstracted out, we won't add if/else(s) to this

pkg/daemon/apply.go
```diff
@@ -0,0 +1,177 @@
...
+
+			// This could be simpler but prepares for the
+			// eventuality that other file changes might be handled
+			// without reboots
+			if len(actions) == 0 {
+				actions = append(actions, DrainPostAction{})
+			}
+			actions = append(actions, ServicePostAction{
+				Reason: fmt.Sprintf("change to %v", file.Node.Path),
+				ServiceName: "crio.service", ServiceAction: "restart"})
```
runcom, line 167: uhm, same as above - don't have a solution rn, will think about it!

pkg/daemon/update.go
```diff
@@ -397,7 +418,34 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+	// been successfully updated
+	if err := dn.nodeWriter.SetDone(
+		dn.kubeClient.CoreV1().Nodes(),
+		dn.nodeLister,
+		dn.name,
+		newConfigName,
+	); err != nil {
+		glog.Errorf("Setting node's state to Done failed, node will reboot: %v", err)
+		return dn.drainAndReboot(newConfig)
+	}
```
runcom, line 445: will play with this PR a little bit - not fully sure about the consequences of having this here, what happens when rebooting? 🤔 - Also, as safe as any changes can be, should we cordon/drain/uncordon anyway?

pkg/daemon/apply.go
beekhof: I'm also in two minds about have flexible vs. simple this should be.
Definitely open to suggestions.

pkg/daemon/update.go
beekhof: I wouldn't be averse to always performing a cordon/drain/uncordon cycle, it would make the code a little simpler.

I didn't look closely enough at the reboot case to know why this block isn't needed there.

pkg/daemon/apply.go
beekhof: At one point there was basically a lookup table that mapped filenames to actions, with reboot used for anything not present.
How would you feel about re-introducing that?

pkg/daemon/apply.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	output := <-outputChannel
+	switch output {
+	case "done":
+		fallthrough
+	case "skipped":
+		glog.Infof("Systemd action successful: %s", output)
+	default:
+		return fmt.Errorf("Systemd action %s", output)
```
mhrivnak, line 107: This error message could use a little more context. Is the action unexpected? Is it invalid? What about this action is bad and/or erroneous?

pkg/daemon/apply.go
```diff
@@ -0,0 +1,213 @@
...
+type actionResult interface {
+	Describe(dn *Daemon) string
+	Execute(dn *Daemon, newConfig *mcfgv1.MachineConfig) error
+}
+
+type NoOpPostAction struct {
+	actionResult
+}
+
+func (a NoOpPostAction)Describe(dn *Daemon) string {
```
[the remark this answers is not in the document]
beekhof: Have now, thanks for the reminder

pkg/daemon/update.go
beekhof: Updated PR unconditionally drains and automatically uncordons if not rebooting.

pkg/daemon/apply.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	defer systemdConnection.Close()
+
+	var err error
+	outputChannel := make(chan string)
+	switch a.ServiceAction {
+	case "restart":
+		glog.Infof("Restarting unit %q", a.ServiceName)
+		_, err = systemdConnection.RestartUnit(a.ServiceName, "replace", outputChannel)
```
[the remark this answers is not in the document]
beekhof: According to the code comments for RestartUnit():

// If the provided channel is non-nil, a result string will be sent to it upon
// job completion: one of done, canceled, timeout, failed, dependency, skipped.

So I think we're safe here.

pkg/daemon/apply.go
beekhof: Messages have been improved and `skipped` is now treated as an error until there is reason not to

ashcrow:
> All of the user stories on this are baremetal, why not only roll this out to baremetal clusters (at least first)?

That makes a lot of sense to me :+1:

beekhof:
I'd be happy to do that... any hints on how?

beekhof:
//nogocyclo

kikisdeliveryservice:
I think that the verify error needs to be fixed instead of overriding.  As we need make verify to pass on all current and future PRs.

kikisdeliveryservice:
also this needs a hold to prevent merging

/hold

beekhof:
> I think that the verify error needs to be fixed instead of overriding.  As we need make verify to pass on all current and future PRs.

I'm a little dubious but will investigate 😉

Is there a way to get more detail about the cycle it's finding?

kikisdeliveryservice:
> > I think that the verify error needs to be fixed instead of overriding.  As we need make verify to pass on all current and future PRs.
> 
> I'm a little dubious but will investigate wink
> 
> Is there a way to get more detail about the cycle it's finding?

Nothing to be dubious about all PRs must pass this test, so this failure is not normal and is related to your changes.
You will will need to look at the function you modified. I believe someone in slack suggested factoring out helper functions so you can start there.

beekhof:
Successful demonstration as an e2e test:

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1922/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1286479423402414080/artifacts/e2e-gcp-op/pods/openshift-machine-config-operator_machine-config-daemon-7676p_machine-config-daemon.log

> I0724 03:08:38.343723    2040 daemon.go:1115] In desired config rendered-worker-b61393fec68dafeb18db70548418da2f
> I0724 03:13:23.558956    2040 update.go:284] Checking Reconcilable for config rendered-worker-b61393fec68dafeb18db70548418da2f to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a
> I0724 03:13:23.588915    2040 update.go:1511] Starting update from rendered-worker-b61393fec68dafeb18db70548418da2f to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:true units:false kernelType:false}
> I0724 03:13:23.592190    2040 update.go:1511] Update prepared; beginning drain 

[...]

> I0724 03:14:12.881026    2040 update.go:1511] drain complete
> I0724 03:14:12.884272    2040 update.go:190] Successful drain took 49.289095485 seconds
> E0724 03:14:12.899678    2040 daemon.go:1507] **content mismatch for file "/etc/containers/registries.conf**" (-want +got):

```
  bytes.Join({
- 	`unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]`,
- 	"",
- 	"[[registry]]",
- 	`  prefix = ""`,
- 	`  location = "blocked.com/ns-b/ns2-b"`,
- 	"  mirror-by-digest-only = true",
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "other.com/ns-o2"`,
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "insecure.com/ns-i2"`,
- 	"",
- 	"[[registry]]",
- 	`  prefix = ""`,
- 	`  location = "insecure.com/ns-i1"`,
- 	"  mirror-by-digest-only = true",
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "blocked.com/ns-b1"`,
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "other.com/ns-o1"`,
- 	"",
- 	"[[registry]]",
- 	`  prefix = ""`,
- 	`  location = "other.com/ns-o3"`,
- 	"  mirror-by-digest-only = true",
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "insecure.com/ns-i2"`,
- 	"",
- 	"  [[registry.mirror]]",
- 	`    location = "blocked.com/ns-b/ns3-b"`,
+ 	"unqualified-search-registries = ['registry.access.redhat.com', 'docker.io']",
  	"",
  }, "\n")

```
> I0724 03:14:12.901433    2040 update.go:915] Updating files

[...]

> I0724 03:14:13.018724    2040 update.go:1315] Writing SSHKeys at "/home/core/.ssh/authorized_keys"
> I0724 03:14:13.025340    2040 update.go:1511] **Performing Restarting service Change to /etc/containers/registries.conf**
> I0724 03:14:13.030881    2040 apply.go:69] **Restarting unit "crio.service"**
> I0724 03:14:13.999594    2040 apply.go:85] **Systemd action "restart" for "crio.service" completed successful: done**
> I0724 03:14:14.035069    2040 update.go:430] **In desired config rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a**
> I0724 03:14:14.055179    2040 update.go:284] Checking Reconcilable for config rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a
> I0724 03:14:14.087933    2040 update.go:1511] Starting update from rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false}
>

beekhof:
I don't believe the remaining test failures are related to this PR

runcom:
> I0724 03:14:14.087933 2040 update.go:1511] Starting update from rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false}

From your previous comment, that seems an update to the very same machine config - is that a result of this PR? is it causing it to restart again?

runcom:
> All of the user stories on this are baremetal, why not only roll this out to baremetal clusters (at least first)?

@beekhof do you still need help with the above suggestion from @kikisdeliveryservice ? it seems the current PR isn't accounting for that - I was thinking a raw call to the infra object, checking if it's BM would be sufficient to gate this, or maybe more robust would also be a label on the MCP indicating you want to opt-in for this new behavior (still on BM only to start with).

beekhof:
Yes, if that is still wanted. Although it would break/prevent e2e testing since that only runs on gcp

beekhof:
> > I0724 03:14:14.087933 2040 update.go:1511] Starting update from rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false}
> 
> From your previous comment, that seems an update to the very same machine config - is that a result of this PR? is it causing it to restart again?

Hmm, I didn't pick up on that.
If it's a result of the PR its not an intentional or expected one.
I'll investigate.

runcom:
> Yes, if that is still wanted. Although it would break/prevent e2e testing since that only runs on gcp

I think we can target an e2e to run in an env within e2e-metal-ipi perhaps? or if we gate this behind a feature flag we would easily test it out on any platform as well - I guess this would need input from MCO team @openshift/openshift-team-mco-maintainers  as well as @cgwalters @crawford (re: special label at the MCP level to gate a feature like this on any platform)

beekhof:
> > Yes, if that is still wanted. Although it would break/prevent e2e testing since that only runs on gcp
> 
> I think we can target an e2e to run in an env within e2e-metal-ipi perhaps? or if we gate this behind a feature flag we would easily test it out on any platform as well - I guess this would need input from MCO team @openshift/openshift-team-mco-maintainers as well as @cgwalters @crawford (re: special label at the MCP level to gate a feature like this on any platform)

IMHO it sounds more complicated/risky than the code path we'd be keeping out

runcom:
> IMHO it sounds more complicated/risky than the code path we'd be keeping out

uhm, can you expand on your concerns? what I'm proposing is a special label acting as an opt-in mechanism for this feature, the MCD could just query the MCP that it is tied on and make the decision whether to reboot or not. It's effectively the same as a gate, instead of platforms, we're querying on users actively setting that label (code wise, it's again as easy afaict)

cgwalters:
Exposing features per platform feels very risky to me.  Today OpenShift (and Kubernetes in general) is extensively used in cloud environments and we get a lot of benefit from the constant testing from that.  However, it also means that more "metal specific" things like having multiple NICs tends not to be exercised as much.

Only exposing a feature on metal is minimizing our testing...without getting us much in return AFAICS.

We do have [feature gates](https://github.com/openshift/api/blob/e7fa4b871a25985ef0cc36c2fbd9f2cb4445dc9c/config/v1/0000_10_config-operator_01_featuregate.crd.yaml) and adding one for this could make sense.

crawford:
What is the reason for needing to introduce this config flag? To further @cgwalters point above, adding another flag introduces another variant for our clusters, so I'm hesitant to go that route. Is this too risky to enable across the board?

cgwalters:
Can you split off a "prep patches" PR from this?  For example I split off https://github.com/openshift/machine-config-operator/pull/1947  from https://github.com/openshift/machine-config-operator/pull/1946

ashcrow:
> What is the reason for needing to introduce this config flag? To further @cgwalters point above, adding another flag introduces another variant for our clusters, so I'm hesitant to go that route. Is this too risky to enable across the board?

Generally I agree that having more configuration flag adds to the supportability matrix. However, in this case it may be warranted. Previously in talking with @runcom and members of the MCO team there was a lot of research they wanted to do before making a fully fledged feature allowing for limited rebooting. If this is still the case I support limiting the scope of this feature to just the area that this is needed sooner rather than later with the understanding it will likely change with output of the MCO teams design work. If this is _not_ the case and the above is now as far as we are taking this work, then I'd tend to agree that limiting may not be helpful.

cgwalters:
> uhm, can you expand on your concerns? what I'm proposing is a special label acting as an opt-in mechanism for this feature, the MCD could just query the MCP that it is tied on and make the decision whether to reboot or not. It's effectively the same as a gate, instead of platforms, we're querying the MCP to understand if users actively set that label (code wise, it's again as easy afaict)

Right; if we're doing gating, doing it on the pool makes a lot more sense to me, because it will allow us to easily test the code on cloud platforms (there's nothing actually metal specific about it).

(I'm in the "reboots are normal" camp mostly, but I also do think it's useful for us to look strategically at the common changes that users might make even in cloud environments and try to avoid reboots where they happen; they're still disruptive, just to a much lesser degree)

beekhof:
> > > I0724 03:14:14.087933 2040 update.go:1511] Starting update from rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false}
> > 
> > 
> > From your previous comment, that seems an update to the very same machine config - is that a result of this PR? is it causing it to restart again?

A second go around the update was a result of this PR. 
Looks like the finalise flow was incomplete, I'm testing a followup patch that should address this.

beekhof:
> Can you split off a "prep patches" PR from this? For example I split off #1947 from #1946

I'll leave this PR as-is for now, so that I can look into feature sets, but I've created #1953 which is a squashed version that still always performs a reboot.

runcom:
> What is the reason for needing to introduce this config flag? To further @cgwalters point above, adding another flag introduces another variant for our clusters, so I'm hesitant to go that route. Is this too risky to enable across the board?

@crawford this is a shift of almost 2y on this pattern of "always reboot", now we're going the route of "avoid reboots" which, given how important reboots are for upgrades, warrants more time to finalize and stabilize. In the enhancement we've clarified that this ICSP case is definitely a start in line with the plan of avoiding reboots and we can evolve from here but we haven't gauged and thought about other cases where maybe this needs further changes and maybe some break here and there (which will make us support another migration? or added logic to handle different cases for compatibility?).
What we're proposing sounds aligned with how Kube introduces features. Sure, it creates another variant but we know this feature targets a specific use case already whether we won't have any enabler on most clouds :) Narrowing this down to whoever opt-in for this feature gives us some space in the future to discuss and come back at this, if we enable it everywhere, we're going to rush fixing whatever cases we missed I think.

There are ~4 ways this can be gated:

- featuregates like @cgwalters suggested, the MCD would just read that and act accordingly
- gate on the platform we know is seeking this feature like @kikisdeliveryservice suggested (from within the MCD again)
- gate  at the MCP level (for better testing)
- a combination of the above? platform+feature gate somehow?

> Generally I agree that having more configuration flag adds to the supportability matrix. However, in this case it may be warranted. Previously in talking with @runcom and members of the MCO team there was a lot of research they wanted to do before making a fully fledged feature allowing for limited rebooting. If this is still the case I support limiting the scope of this feature to just the area that this is needed sooner rather than later with the understanding it will likely change with output of the MCO teams design work. If this is _not_ the case and the above is now as far as we are taking this work, then I'd tend to agree that limiting may not be helpful.


> Right; if we're doing gating, doing it on the pool makes a lot more sense to me, because it will allow us to easily test the code on cloud platforms (there's nothing actually metal specific about it).

@cgwalters @ashcrow I've replied above for those comments as well

beekhof:
> > > > I0724 03:14:14.087933 2040 update.go:1511] Starting update from rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a to rendered-worker-ae9b36c934fcb6457ef7cd6a2165608a: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false}
> > > 
> > > 
> > > From your previous comment, that seems an update to the very same machine config - is that a result of this PR? is it causing it to restart again?
> 
> A second go around the update was a result of this PR.
> Looks like the finalise flow was incomplete, I'm testing a followup patch that should address this.

@runcom fixed in the latest patch

beekhof:
Is there a final decision about whether to gate, and if so, how?

PR #1923, opened 2020-07-13
pkg/daemon: refresh the node object

runcom:
Somehow, and it's not fully clear why as I'm not an expert on the k8s API side of the world, I imagine the indexers getting confused somehow and bringing us here:

```
"metadata": {
        "annotations": {
            "machine.openshift.io/machine": "openshift-machine-api/ci-ln-zic4ylt-f76d1-68jfg-worker-b-46p64",
            "machineconfiguration.openshift.io/currentConfig": "rendered-worker-84b8fd86e2cb32a272b2ea9871bbfa87",
            "machineconfiguration.openshift.io/desiredConfig": "rendered-worker-e8f5eab633c34a53702b0efc7abaaa50",
            "machineconfiguration.openshift.io/reason": "failed to run pivot: failed to run pivot: exit status 1: failed to run command podman (6 tries): timed out waiting for the condition: running podman pull -q --authfile /var/lib/kubelet/config.json quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8 failed: Error: error pulling image \"quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8\": unable to pull quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8: unable to pull image: Error initializing source docker://quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8: Error reading manifest sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8 in quay.io/openshift-release-dev/ocp-v4.0-art-dev: manifest unknown: manifest unknown\n: exit status 125",
            "machineconfiguration.openshift.io/state": "Working",
            "volumes.kubernetes.io/controller-managed-attach-detach": "true"
        },
```

The above is totally unexpected unless something goes wrong when managing the object
at the kube level. Our code makes sure that "if we're degraded, we never switch to working".
So the above, I think, must be the result of some weird Strategic Patch edge case
where, despite being told to update Degraded and Reason, it only does it for Reason.
[...]

PR #1924, opened 2020-07-14
pkg/daemon: drop nodeWriter interface + set resourceVersion when patching

runcom:
tl;dr; refer to https://github.com/openshift/machine-config-operator/pull/1923#issuecomment-657735033 for a detailed diagnosis of the issue

These patches do two (major) things:

- remove the nodeWriter interface, not needed now that the daemon runs as a real controller and we can afford to fail and retry
- set node's resourceVersion when patching, making sure we get a 409 on conflict and retry

Minor changes in different commits:

- react to Reason annotation changes (why we never did that is a unknown to me, so far, we're sticking to an old degraded reason if the reason changes but we're still degraded, it's just wrong...)
- use a faster workqueue for the render controller (see commit message specifically)

Long story:

The second point above is crucial as it's the source of a bug, let me try to explain what happens:

- resync on an informer means: replay the in-memory state that I have.  It does not mean relist from the apiserver
- the lister.get is based on the cached state, so if the watch hasn't observed the change, the lister.Get won't have it

Given the assumptions above (thanks deads for explaining):

- the daemon update the node object when it degrades
- it immediately retries the action, lister.get yielding an _old_ cached state
- that defeats guards like "switch to working only when not degraded"

Explanation to last above above: because of how we patch node objects, we may end up checking
the state annotation with an _old_ object (passing the guard) but then updating the right object
with states that don't make sense anymore (like a Working state and a real Degraded reason).

The fix for all that madness is pretty simple (after speaking to deads):

- use resourceVersion when patching the node object

That makes sure that if we try to patch an object using an old resouce version we'll get a conflict error,
we then exit the guard we entered, and retry with a fresher object, all good.

This can be tested with (credits to deads again)

- oc get ns/foo -ojson
- build your patch with RV
- oc annotate ns/foo something-to-change-rv=true
- oc patch with RV

step four should get a conflict error

### DIFF
--- a/internal/kubeutils.go
+++ b/internal/kubeutils.go
@@ -22,33 +22,37 @@ import (
 func UpdateNodeRetry(client corev1client.NodeInterface, lister corev1lister.NodeLister, nodeName string, f func(*corev1.Node)) (*corev1.Node, error) {
 	var node *corev1.Node
 	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		n, err := lister.Get(nodeName)
-		if err != nil {
-			return err
-		}
-		oldNode, err := json.Marshal(n)
-		if err != nil {
-			return err
-		}
-
-		nodeClone := n.DeepCopy()
-		f(nodeClone)
-
-		newNode, err := json.Marshal(nodeClone)
-		if err != nil {
-			return err
-		}
-
-		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldNode, newNode, corev1.Node{})
-		if err != nil {
-			return fmt.Errorf("failed to create patch for node %q: %v", nodeName, err)
-		}
-
-		node, err = client.Patch(context.TODO(), nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
-		return err
+		var innerErr error
+		node, innerErr = UpdateNode(client, lister, nodeName, f)
+		return innerErr
 	}); err != nil {
 		// may be conflict if max retries were hit
-		return nil, fmt.Errorf("unable to update node %q: %v", node, err)
+		return nil, fmt.Errorf("unable to update node %q: %v", nodeName, err)
 	}
 	return node, nil
+}
+
+// UpdateNode calls f to update a node object in Kubernetes.
+func UpdateNode(client corev1client.NodeInterface, lister corev1lister.NodeLister, nodeName string, f func(*corev1.Node)) (*corev1.Node, error) {
+	n, err := lister.Get(nodeName)
+	if err != nil {
+		return nil, err
+	}
+	oldNodeClone := n.DeepCopy()
+	oldNodeClone.ResourceVersion = ""
+	oldNode, err := json.Marshal(oldNodeClone)
+	if err != nil {
+		return nil, err
+	}
+	nodeClone := n.DeepCopy()
+	f(nodeClone)
+	newNode, err := json.Marshal(nodeClone)
+	if err != nil {
+		return nil, err
+	}
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldNode, newNode, corev1.Node{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create patch for node %q: %v", nodeName, err)
+	}
+	return client.Patch(context.TODO(), nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
 }
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -15,6 +15,7 @@ import (
 	mcfginformersv1 "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions/machineconfiguration.openshift.io/v1"
 	mcfglistersv1 "github.com/openshift/machine-config-operator/pkg/generated/listers/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/version"
+	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -40,6 +41,12 @@ const (
 	// renderDelay is a pause to avoid churn in MachineConfigs; see
 	// https://github.com/openshift/machine-config-operator/issues/301
 	renderDelay = 5 * time.Second
+
+	// updateDelay is the baseline speed at which we react to changes.
+	updateDelay = 5 * time.Second
+	// maxDelay is the maximum time to react to a change as we back off
+	// in the face of errors.
+	maxDelay = 60 * time.Second
 )
 
 var (
@@ -84,8 +91,10 @@ func New(
 	ctrl := &Controller{
 		client:        mcfgClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "machineconfigcontroller-rendercontroller"}),
-		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "machineconfigcontroller-rendercontroller"),
 	}
+	ctrl.queue = workqueue.NewNamedRateLimitingQueue(workqueue.NewMaxOfRateLimiter(
+		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(updateDelay), 1)},
+		workqueue.NewItemExponentialFailureRateLimiter(1*time.Second, maxDelay)), "machineconfigcontroller-rendercontroller")
 
 	mcpInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    ctrl.addMachineConfigPool,
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -95,8 +95,6 @@ type Daemon struct {
 	updateActive     bool
 	updateActiveLock sync.Mutex
 
-	nodeWriter NodeWriter
-
 	// channel used by callbacks to signal Run() of an error
 	exitCh chan<- error
 
@@ -277,9 +275,6 @@ func (dn *Daemon) ClusterConnect(
 	dn.name = name
 	dn.kubeClient = kubeClient
 
-	dn.nodeWriter = newNodeWriter()
-	go dn.nodeWriter.Run(dn.stopCh)
-
 	// Other controllers start out with the default controller limiter which retries
 	// in milliseconds; since any change here will involve rebooting the node
 	// we don't need to react in milliseconds.  See also updateDelay above.
@@ -372,12 +367,18 @@ func (dn *Daemon) handleErr(err error, key interface{}) {
 }
 
 func (dn *Daemon) updateErrorState(err error) {
-	switch errors.Cause(err) {
-	case errUnreconcilable:
-		dn.nodeWriter.SetUnreconcilable(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
-	default:
-		dn.nodeWriter.SetDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
+	cause := errors.Cause(err)
+	if cause == errUnreconcilable {
+		dn.setUnreconcilable(err)
+		return
 	}
+	if apierrors.IsConflict(cause) {
+		// skip conflict errors, those are retried and should never result in a permanent
+		// failure worth reporting to the node annoation reasons
+		glog.Infof("API conflict error: %v, retrying", err)
+		return
+	}
+	dn.setDegraded(err)
 }
 
 func (dn *Daemon) syncNode(key string) error {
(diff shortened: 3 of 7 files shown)


runcom:
And here's the race being taken care of (and we'll just retry):
```
I0714 09:45:48.563897    2028 kubeutils.go:58] runcom patch bytes: {"metadata":{"annotations":{"machineconfiguration.openshift.io/reason":"failed to run pivot: failed to run pivot: exit status 1: failed to run command podman (6 tries): timed out waiting for the condition: running podman pull -q --authfile /var/lib/kubelet/config.json registry.svc.ci.openshift.org/ocp/4.6@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8 failed: Error: error pulling image \"registry.svc.ci.openshift.org/ocp/4.6@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8\": unable to pull registry.svc.ci.openshift.org/ocp/4.6@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8: unable to pull image: Error initializing source docker://registry.svc.ci.openshift.org/ocp/4.6@sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8: Error reading manifest sha256:d746ebe14faaa6d1a1847c0d939373b74306c38c39f77315e559a7ec676fbbb8 in registry.svc.ci.openshift.org/ocp/4.6: manifest unknown: manifest unknown\n: exit status 125","machineconfiguration.openshift.io/state":"Degraded"},"resourceVersion":"40664"}}
I0714 09:45:48.577978    2028 kubeutils.go:58] runcom patch bytes: {"metadata":{"resourceVersion":"40664"}}
E0714 09:45:48.587757    2028 writer.go:55] Marking Degraded due to: error setting node's state to Working: Operation cannot be fulfilled on nodes "ci-ln-s0rhzmb-f76d1-nlfzr-worker-b-57j9w": the object has been modified; please apply your changes to the latest version and try again
I0714 09:45:48.589432    2028 kubeutils.go:58] runcom patch bytes: {"metadata":{"annotations":{"machineconfiguration.openshift.io/reason":"error setting node's state to Working: Operation cannot be fulfilled on nodes \"ci-ln-s0rhzmb-f76d1-nlfzr-worker-b-57j9w\": the object has been modified; please apply your changes to the latest version and try again"},"resourceVersion":"43416"}}
```
/hold cancel

internal/kubeutils.go
```diff
@@ -22,33 +22,37 @@ import (
...
-		}
-
-		node, err = client.Patch(context.TODO(), nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
-		return err
+		var innerErr error
+		node, innerErr = UpdateNode(client, lister, nodeName, f)
+		return innerErr
 	}); err != nil {
 		// may be conflict if max retries were hit
 		return nil, fmt.Errorf("unable to update node %q: %v", node, err)
```
ericavonb, line 30: Can we use `nodeName` rather than the node object in the error message here? Since `node` could be nil if we hit max retries (also putting the whole node object into the error has caused issues before)

internal/kubeutils.go
```diff
@@ -22,33 +22,37 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	nodeClone := n.DeepCopy()
+	f(nodeClone)
+	newNode, err := json.Marshal(nodeClone)
+	if err != nil {
+		return nil, err
+	}
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldNode, newNode, corev1.Node{})
```
ericavonb, line 53: Wait, why do we marshal to json if we immediately unmarshal it? Something to do with schemas and versioning?

pkg/controller/render/render_controller.go
```diff
@@ -84,8 +85,10 @@ func New(
 	ctrl := &Controller{
 		client:        mcfgClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "machineconfigcontroller-rendercontroller"}),
-		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "machineconfigcontroller-rendercontroller"),
 	}
+	ctrl.queue = workqueue.NewNamedRateLimitingQueue(workqueue.NewMaxOfRateLimiter(
+		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(5*time.Second), 1)},
```
ericavonb, line 90: Can we make these rate limits a const with a descriptive comment about why e.g. 5 sec was chosen? like we do for `renderDelay` above

internal/kubeutils.go
runcom: great catch! thanks Erica!

internal/kubeutils.go
runcom: pretty sure that's the case when dealing with `CreateTwoWayMergePatch` yeah 🤷‍♂️

pkg/daemon/daemon.go
```diff
@@ -372,11 +367,18 @@ func (dn *Daemon) handleErr(err error, key interface{}) {
...
 func (dn *Daemon) updateErrorState(err error) {
-	switch errors.Cause(err) {
+	cause := errors.Cause(err)
+	switch cause {
 	case errUnreconcilable:
-		dn.nodeWriter.SetUnreconcilable(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
+		dn.setUnreconcilable(err)
 	default:
-		dn.nodeWriter.SetDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
+		if apierrors.IsConflict(cause) {
```
ericavonb, line 375: nit: rather than nest an if clause in the switch, remove the switch.
e.g.
```go
	cause := errors.Cause(err)
	if cause == errUnreconcilable {
		return dn.setUnreconcilable(err)
	}
	if apierrors.IsConflict(cause) {
		// skip conflict errors, those are retried and should never result in a permanent
		// failure worth reporting to the node annoation reasons
		glog.Infof("API conflict error: %v, retrying", err)
		return
	}
	dn.setDegraded(err)
}
```

runcom:
@ericavonb addressed your comments, can you ptal 👼

runcom:
I really want a test for this to make sure the error wrapping + conditionals to avoid reporting degraded always work (bear with me)

/hold

pkg/daemon/node.go
```diff
@@ -43,7 +43,7 @@ func (dn *Daemon) loadNodeAnnotations(node *corev1.Node) (*corev1.Node, error) {
 	}
 
 	glog.Infof("Setting initial node config: %s", initial[constants.CurrentMachineConfigAnnotationKey])
-	n, err := setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, node.Name, initial)
+	n, err := dn.setNodeAnnotations(initial)
```
yuqi-zhang, line 46: hm, do you know why for some instances we don't use the retry version of setNodeAnnotations (e.g. here)?

pkg/daemon/daemon.go
```diff
@@ -372,12 +367,18 @@ func (dn *Daemon) handleErr(err error, key interface{}) {
...
-	default:
-		dn.nodeWriter.SetDegraded(err, dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name)
+	cause := errors.Cause(err)
+	if cause == errUnreconcilable {
+		dn.setUnreconcilable(err)
+		return
+	}
+	if apierrors.IsConflict(cause) {
+		// skip conflict errors, those are retried and should never result in a permanent
+		// failure worth reporting to the node annoation reasons
```
yuqi-zhang, line 377: nit `annoation` -> `annotation`

internal/kubeutils.go
```diff
@@ -22,33 +22,37 @@ import (
...
 }
+
+// UpdateNode calls f to update a node object in Kubernetes.
+func UpdateNode(client corev1client.NodeInterface, lister corev1lister.NodeLister, nodeName string, f func(*corev1.Node)) (*corev1.Node, error) {
+	n, err := lister.Get(nodeName)
+	if err != nil {
+		return nil, err
+	}
+	oldNodeClone := n.DeepCopy()
+	oldNodeClone.ResourceVersion = ""
```
yuqi-zhang, line 42: So in the PR you mentioned that:
```
That makes sure that if we try to patch an object using an old resouce version we'll get a conflict error,
we then exit the guard we entered, and retry with a fresher object, all good.
```

I'm a little confused as to how we're actually performing this. By setting the old `ResourceVersion` to empty, the merge should set the ResourceVersion to whatever is in newNode, right? Is that different from before? I don't really see a difference in how ResourceVersion is used I guess is my question.

runcom: Right, let me explain better the situation here and maybe you can help me word this better Jerry:

Before (just an example, annotations is a map, but I'm doing array for simplicity):

- grab the node object (we'll call it `instance{number: 1, resourceVersion: 42, annotations: ["test1"]}`)
- clone it (now we have `instance{number: 2, resourceVersion: 42, annotations: ["test1"]}`)
- modify the annotations on the clone (now we have `instance{instance{number: 2, resourceVersion: 42, annotations: ["test1", "test2"]}`)
- run the instances through `CreateTwoWayMergePatch`

What `CreateTwoWayMergePatch` does and how Patch works is crucial here. When `CreateTwoWayMergePatch` is run above the following (truncated) `diff` is generated:

```
node
  annotations: ["test1", "test2"]
```

You can notice the above generated a diff, meaning it's giving you just the diff object that we'll use to patch. The result is that Patching with the diff above will only ever change annotations, as that's effectively the only changes `CreateTwoWayMergePatch` can see from the two objects.

The fix I did here forces us to always generate **also** a resourceVersion diff so that it can be noticed at Patch time and the apiserver complains if it's old.

One other thing to note here is that we're not playing with the original `n` node instance, only `oldNodeClone` and `oldNode` to generate the diff (`n` remains untouched)

By setting resourceVersion to empty, we'll go like this:

- grab the node object (we'll call it `instance{number: 1, resourceVersion: 42, annotations: ["test1"]}`)
- make a first clone and empty out the resourceVersion `instance{number: 1, resourceVersion: "", annotations: ["test1"]}`
- clone the original object again (now we have `instance{number: 2, resourceVersion: 42, annotations: ["test1"]}`)
- modify the annotations on this last clone (now we have `instance{instance{number: 2, resourceVersion: 42, annotations: ["test1", "test2"]}`)
- run the instances through `CreateTwoWayMergePatch`

Now, the diff we get is:

```
node
  annotations: ["test1", "test2"]
  resourceVersion: 42
``` 

And we'll patch with the above. If we send `42` but the apiserver already knows about `43` we'll get a conflict

pkg/daemon/node.go
runcom: Yep, for the critical path (from `syncNode` to reboot) we don't really want to retry, it's the controller mechanism which will retry for us. For things like setting the SSH annotation we don't really care, I could have _not_ add a retry as well as it's retried anyway, but it looked more consistent there to retry 🤷‍♂️

internal/kubeutils.go
runcom: Right, let me explain better the situation here and maybe you can help me word this better Jerry:
Before (just an example, annotations is a map, but I'm doing array for simplicity):
- grab the node object (we'll call it `instance{number: 1, resourceVersion: 42, annotations: ["test1"]}`)
- clone it (now we have `instance{number: 2, resourceVersion: 42, annotations: ["test1"]}`)
- modify the annotations on the clone (now we have `instance{instance{number: 2, resourceVersion: 42, annotations: ["test1", "test2"]}`)
- run the instances through `CreateTwoWayMergePatch`
What `CreateTwoWayMergePatch` does and how Patch works is crucial here. When `CreateTwoWayMergePatch` is run above the following (truncated) `diff` is generated:
```
node
  annotations: ["test1", "test2"]
```
You can notice the above generated a diff, meaning it's giving you just the diff object that we'll use to patch. The result is that Patching with the diff above will only ever change annotations, as that's effectively the only changes `CreateTwoWayMergePatch` can see from the two objects.

The fix I did here forces us to always generate **also** a resourceVersion diff so that it can be noticed at Patch time and the apiserver complains if it's old.
One other thing to note here is that we're not playing with the original `n` node instance, only `oldNodeClone` and `oldNode` to generate the diff (`n` remains untouched)
By setting resourceVersion to empty, we'll go like this:
- grab the node object (we'll call it `instance{number: 1, resourceVersion: 42, annotations: ["test1"]}`)
- make a first clone and empty out the resourceVersion `instance{number: 1, resourceVersion: "", annotations: ["test1"]}`
- clone the original object again (now we have `instance{number: 2, resourceVersion: 42, annotations: ["test1"]}`)
- modify the annotations on this last clone (now we have `instance{instance{number: 2, resourceVersion: 42, annotations: ["test1", "test2"]}`)
- run the first old clone and the other clone instances through `CreateTwoWayMergePatch`
Now, the diff we get is:
```
node
  annotations: ["test1", "test2"]
  resourceVersion: 42
``` 

It is ultimately making sure that if we're trying to patch an object, at least the changes we're applying are applied to the very last version of an object in the apiserver, otherwise, we'll retry.

pkg/daemon/node.go
runcom: Yep, for the critical path (from `syncNode` to reboot) we don't really want to retry, it's the controller mechanism which will retry for us. For things like setting the SSH annotation we don't really care, I could have _not_ add a retry as well as it's retried anyway, but it looked more consistent there to retry 🤷‍♂️

internal/kubeutils.go
yuqi-zhang: Ah, I see, no that makes perfect sense. Awesome explanation!

Maybe for future purposes it'd be nice to add a comment like:

`// Add empty resourceVersion, so CreateTwoWayMergePatch returns a diff with resourceVersion of the node object when comparing.`

runcom: Adding that, that's really helpful Jerry, thanks for pointing it out!

cgwalters:
Wouldn't it be sufficient to solve the bug to do *either* of these things (drop nodeWriter or retry on resourceVersion) - basically the core problem is we're racing against our own process, so if we patched the node synchronously there wouldn't be a race anymore.  Right?

PR #1925, opened 2020-07-14
Add kernel arg tuneable allowlist for FCOS

LorbusChris:
This is required for Fedora CoreOS installs, which set
`mitigations=auto,nosmt` and may enable cgroups v2
in the future so it has to be allowed to disable it explicitly.

/cc @runcom @vrutkovs

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -34,10 +34,16 @@ const (
 	cmdLineFile      = "/proc/cmdline"
 )
 
-// TODO: fill out the allowlist
-// tuneableArgsAllowlist contains allowed keys for tunable arguments
-var tuneableArgsAllowlist = map[string]bool{
+// TODO: fill out the allowlists
+// tuneableRHCOSArgsAllowlist contains allowed keys for tunable kernel arguments on RHCOS
+var tuneableRHCOSArgsAllowlist = map[string]bool{
 	"nosmt": true,
+}
+
+// tuneableFCOSArgsAllowlist contains allowed keys for tunable kernel arguments on FCOS
+var tuneableFCOSArgsAllowlist = map[string]bool{
+	"systemd.unified_cgroup_hierarchy=0": true,
+	"mitigations=auto,nosmt":             true,
 }
 
 var pivotCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


cgwalters:
This is a super messy topic.  First, I think we should recommend using `kernelArgs` in MachineConfig today via additional manifests provided to the installer instead of the `/etc/pivot` interface.  The installer should switch to doing that instead.

However...what you really want for *this specific* case is the ability to *remove* kernel arguments (I am not sure that one can easily re-enable SMT by appending).  We could add such a thing to MachineConfig...or we could add a high level `hyperThreading: true` value just like the install config has.

Another path would be for OKD to always remove the `migitations=auto,nosmt` value by default (i.e. making OKD match OCP/RHCOS) and then the existing interfaces exposed for this by the installer and the MCO to disable it again would work.

LorbusChris:
@cgwalters what we've been doing so far is this: https://github.com/openshift/installer/commit/5814b133153c45377ff72ba2c58f59b6917d95cc#diff-41963c416a97dfbe789cafbd4a347a78R217 -- is what you're saying that the removal doesn't work?

I agree that we should switch to using MC `kernelArgs` for this instead of pivot - will look into it.

cgwalters:
> @cgwalters what we've been doing so far is this: openshift/installer@5814b13#diff-41963c416a97dfbe789cafbd4a347a78R217 -- is what you're saying that the removal doesn't work?

That code might work indeed.  I'm not opposed to this PR to be clear, but I'd like to consider the `/etc/pivot` interface *very deprecated*, and would prefer [one of the options above](https://github.com/openshift/machine-config-operator/pull/1925#issuecomment-658244328).

LorbusChris:
@cgwalters @runcom this PR is keeping the status quo, and adding to master what we do on the `fcos` branch already.
If you're not entirely opposed, I'd like to merge this, as it's among the last pieces missing before we can drop the fork entirely.
And I'll file a card for moving from pivot to MC's kernelArguments.

runcom:
I agree with Colin about deprecating that interface and also ok with this PR as well to allow a safe iteration towards porting the working FCOS branch patches here. As long as we track the deprecation with a card I’m fine, that card will likely drop nosmt for rhcos as well, I’ll leave to Colin to approve if the plan sounds good 👍

sinnykumari:
I don't have full context of how fcos branch of installer and MCO works today. For master branch we have kargs support added for both day1 and day2 . It should be easy to validate for OKD with the doc https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfiguration.md#KernelArguments. As Colin mentioned, applying a karg through MachineConfigs has advantage of deleting them further as well.

While working on PR https://github.com/openshift/machine-config-operator/pull/1850 which involves lot of changes during pivot, I was thinking of the possibility of cleaning up tunable kargs which we support during pivot.

LorbusChris:
@sinnykumari OCP and OKD installer currently both use the pivot interface to add the kargs (they don't differ in the way they do it, they just set different ones).

cgwalters:
For now
/approve
and code looks good to me.

sinnykumari:
> @sinnykumari OCP and OKD installer currently both use the pivot interface to add the kargs (they don't differ in the way they do it, they just set different ones).

right. I was saying that same should be then possible to achieve through MachineConfig during cluster install. But yeah, we can discuss this in the card  suggested in https://github.com/openshift/machine-config-operator/pull/1925#issuecomment-658584046

PR #1926, opened 2020-07-14
templates: Make all templates into yaml files

LorbusChris:
**- What I did**

**- How to verify it**


ericavonb:
Should this PR have a WIP on it? Not sure what's going on or what the context is here.

yuqi-zhang:
All our template files are written in yaml but we're indifferent to the actual naming scheme when parsing them. I think this is just a minor cleanup to make sure they are named with a `.yaml` suffix (As opposed to a .service suffix since they're not "service files"). Shouldnt error either way

kikisdeliveryservice:
> All our template files are written in yaml but we're indifferent to the actual naming scheme when parsing them. I think this is just a minor cleanup to make sure they are named with a `.yaml` suffix (As opposed to a .service suffix since they're not "service files"). Shouldnt error either way

If this is indeed the case, @LorbusChris can you please add this rationale to the commit message just so it's clear?

ericavonb:
Got it. I don't see any problems with the change. @LorbusChris ping when you've updated the commit message and I can add the lgtm

kikisdeliveryservice:
Just threw a WIP on this since it needs a rebase/update commit as requested. Take it off when ready :smiley_cat:

LorbusChris:
@kikisdeliveryservice retitles per your comment, should be good to go :)

/retest

kikisdeliveryservice:
Thanks @LorbusChris 

/lgtm

PR #1927, opened 2020-07-15
Bug 1854801: Make sure openvswitch related services are always restarted

abhat:
**- What I did**
Set Restart=always in service files

**- How to verify it**
By manually killing the ovs-vswitchd process on rhcos and verifying
it gets started automatically by systemd.

OVS services (openvswitch and ovsd-bserver) are run without the
--monitor option. As a result, when the underlying daemon gets
killed by SIGTERM, the process is not automatically restarted
unless the Restart tag is set to always. This is critical for
system services such as OVS which need to be always up for traffic
forwarding. This fixes BZ: 1854801

### DIFF
--- a/templates/common/_base/units/ovs-vswitchd.service.yaml
+++ b/templates/common/_base/units/ovs-vswitchd.service.yaml
@@ -1,0 +1,6 @@
+name: ovs-vswitchd.service
+dropins:
+  - name: 10-ovs-vswitchd-restart.conf
+    contents: |
+      [Service]
+      Restart=always
--- a/templates/common/_base/units/ovsdb-server.service
+++ b/templates/common/_base/units/ovsdb-server.service
@@ -1,2 +1,7 @@
 name: "ovsdb-server.service"
 enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
+dropins:
+  - name: 10-ovsdb-restart.conf
+    contents: |
+      [Service]
+      Restart=always


templates/common/_base/units/ovsdb-server.service.yaml
```diff
@@ -0,0 +1,8 @@
+name: ovsdb-server.service
+dropins:
+  - name: 10-ovsdb-restart.conf
+    contents: |
+      [Service]
+      {{ if or (eq {{.NetworkType -}} "OVNKubernetes") (eq {{.NetworkType -}} "OpenshiftSDN") }}
+      Restart=always
+      {{end}}
```
runcom, line 8: since this drop-in is for `ovsdb-server.service` which has a service as well, you can collapse the two right?

```
name: ovsdb-server.service
enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
dropins:
  - name: 10-ovsdb-restart.conf
    contents: |
      [Service]
      {{ if or (eq {{.NetworkType -}} "OVNKubernetes") (eq {{.NetworkType -}} "OpenshiftSDN") }}
      Restart=always
      {{end}}
```

templates/common/_base/units/ovs-vswitchd.service.yaml
```diff
@@ -0,0 +1,8 @@
+name: ovs-vswitchd.service
```
runcom, line 1: is this a standalone service or is it `openvswitch.service`?

templates/common/_base/units/ovs-vswitchd.service.yaml
```diff
@@ -0,0 +1,8 @@
...
+dropins:
+  - name: 10-ovs-vswitchd-restart.conf
+    contents: |
+      [Service]
+      {{- if or (eq .NetworkType "OVNKubernetes") (eq .NetworkType "OpenshiftSDN") }}
```
runcom, line 6: I'm not sure why we need conditionals on the dropins, the services are enabled under the very same conditions, so if network isn't sdn|ovn, the service doesn't even get enabled and doesn't matter if it has restart always or not

templates/common/_base/units/ovsdb-server.service.yaml
runcom: Also, related to https://github.com/openshift/machine-config-operator/pull/1927/files#r455735208, this can just be:

```
name: ovsdb-server.service
enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
dropins:
  - name: 10-ovsdb-restart.conf
    contents: |
      [Service]
      Restart=always
```

The dropin doesn't have any effect anyway as the whole service isn't enabled if not ovn|sdn

templates/common/_base/units/ovs-vswitchd.service.yaml
runcom: if it's openvswitch, this won't work as you're not dropping an override for the correct service, and you'd need something like https://github.com/openshift/machine-config-operator/pull/1927/files#r455735851 anyway

abhat: ovs-vswitchd.service is a separate service. it's not a drop-in for openvswitch.service.

templates/common/_base/units/ovsdb-server.service.yaml
abhat: fixed.

templates/common/_base/units/ovs-vswitchd.service.yaml
abhat: ack

abhat:
@runcom thank you for the review. Appreciate it. PTAL I made the changes as you requested.

abhat:
@runcom looks like this isn't cutting it. The dropin files exist, but the service files doesn't have `Restart=always` set. So somehow the default config is overriding the dropin.

abhat:
Correct, so while this isn't MCO related per-se, I am looking for help on how things get put together for the service file.

runcom:
> Correct, so while this isn't MCO related per-se, I am looking for help on how things get put together for the service file.

I think that's something we might want to ask to whoever owns those services, other than the usual check we could do with systemd 🤷‍♂️ if the services and dropins are there this PR may just be putting things in the wrong place? or something else I don't know

PR #1935, opened 2020-07-17
Bug 1858026: add nil check for infra.Status.PlatformStatus in isCloudConfigReqd

kikisdeliveryservice:
infra.Status.PlatformStatus is *PlatformStatus and in some <4.5.x setups this entire thing is empty and only platform is set to none. So when we hit the switch statement in the MCO we panic bc pointers, so check that it's really there before you do the switch stmt case comparisions.  Before my fix, the unit test I added failed with same panic, now it passes and func returns false.  This behavior was seen in bm clusters updating from 4.4.12->4.5.2, which is also when Platform was deprecated in favor of PlatformStatus and the MCO was missing the check. The checks do exist in the MCC transitioning them to the new type.

As for why we saw this with users but not in CI, AFAIK there isn't an e2e metal upgrade job accounting for this anywhere and I believe the existing metal job would just install a 4.5.x cluster with the new PlatformStatus.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -112,13 +112,18 @@ func isCloudConfigRequired(infra *configv1.Infrastructure) bool {
 	if infra.Spec.CloudConfig.Name != "" {
 		return true
 	}
-	switch infra.Status.PlatformStatus.Type {
-	case configv1.AzurePlatformType,
-		configv1.GCPPlatformType,
-		configv1.OpenStackPlatformType,
-		configv1.VSpherePlatformType:
-		return true
-	default:
+
+	if infra.Status.PlatformStatus != nil {
+		switch infra.Status.PlatformStatus.Type {
+		case configv1.AzurePlatformType,
+			configv1.GCPPlatformType,
+			configv1.OpenStackPlatformType,
+			configv1.VSpherePlatformType:
+			return true
+		default:
+			return false
+		}
+	} else {
 		return false
 	}
 }
--- a/lib/resourcemerge/machineconfig.go
+++ b/lib/resourcemerge/machineconfig.go
@@ -76,9 +76,11 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
 	setBytesIfSet(modified, &existing.KubeAPIServerServingCAData, required.KubeAPIServerServingCAData)
 	setBytesIfSet(modified, &existing.CloudProviderCAData, required.CloudProviderCAData)
 
-	if !equality.Semantic.DeepEqual(existing.Infra.Status.PlatformStatus.Type, required.Infra.Status.PlatformStatus.Type) {
-		*modified = true
-		existing.Infra.Status.PlatformStatus.Type = required.Infra.Status.PlatformStatus.Type
+	if existing.Infra.Status.PlatformStatus != nil && required.Infra.Status.PlatformStatus != nil {
+		if !equality.Semantic.DeepEqual(existing.Infra.Status.PlatformStatus.Type, required.Infra.Status.PlatformStatus.Type) {
+			*modified = true
+			existing.Infra.Status.PlatformStatus.Type = required.Infra.Status.PlatformStatus.Type
+		}
 	}
 
 	if !equality.Semantic.DeepEqual(existing.Proxy, required.Proxy) {
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -422,12 +422,16 @@ func etcdMetricCertCommand(cfg RenderConfig) (interface{}, error) {
 }
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
-	switch cfg.Infra.Status.PlatformStatus.Type {
-	case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.OpenStackPlatformType, configv1.VSpherePlatformType:
-		return strings.ToLower(string(cfg.Infra.Status.PlatformStatus.Type)), nil
-	case configv1.GCPPlatformType:
-		return "gce", nil
-	default:
+	if cfg.Infra.Status.PlatformStatus != nil {
+		switch cfg.Infra.Status.PlatformStatus.Type {
+		case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.OpenStackPlatformType, configv1.VSpherePlatformType:
+			return strings.ToLower(string(cfg.Infra.Status.PlatformStatus.Type)), nil
+		case configv1.GCPPlatformType:
+			return "gce", nil
+		default:
+			return "", nil
+		}
+	} else {
 		return "", nil
 	}
 }
--- a/pkg/operator/sync_test.go
+++ b/pkg/operator/sync_test.go
@@ -1,0 +1,14 @@
+package operator
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestIsCloudConfigRequired(t *testing.T) {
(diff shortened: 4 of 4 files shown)


kikisdeliveryservice:
Replicated underlying bug with a missing struct field:
```
 --- FAIL: TestIsCloudConfigRequired (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14e7959] 
```

cgwalters:
Code LGTM, but I wonder if there are other bugs lurking.  Doing a quick grep in the code I also see e.g.

`	if !equality.Semantic.DeepEqual(existing.Infra.Status.PlatformStatus.Type, required.Infra.Status.PlatformStatus.Type) {`

which also doesn't seem to be accounting for a possibly `nil` status.

(If this was Rust it'd be `Option<PlatformStatus>` and the compiler would warn error on this...[billion dollar mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/) etc.)

kikisdeliveryservice:
> Code LGTM, but I wonder if there are other bugs lurking. Doing a quick grep in the code I also see e.g.
> 
> ` if !equality.Semantic.DeepEqual(existing.Infra.Status.PlatformStatus.Type, required.Infra.Status.PlatformStatus.Type) {`
> 
> which also doesn't seem to be accounting for a possibly `nil` status.
> 
> (If this was Rust it'd be `Option<PlatformStatus>` and the compiler would warn error on this...[billion dollar mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/) etc.)

good point!!! ill check those out

kikisdeliveryservice:
Ok updated the remaining ones that didnt' have a nil check.

pkg/operator/sync.go
```diff
@@ -112,13 +112,18 @@ func isCloudConfigRequired(infra *configv1.Infrastructure) bool {
...
+		switch infra.Status.PlatformStatus.Type {
+		case configv1.AzurePlatformType,
+			configv1.GCPPlatformType,
+			configv1.OpenStackPlatformType,
+			configv1.VSpherePlatformType:
+			return true
+		default:
+			return false
+		}
+	} else {
```
runcom, line 126: a nest can be removed here and above with:
```
if cfg.Infra.Status.PlatformStatus == nil {
    return ...
}
switch cfg.Infra.Status.PlatformStatus.Type {
...
```
Thinking twice on this, the change in `lib/resourcemerge/machineconfig.go` LGTM but following the trails for the panic indicates that our code isn't that robust to changes (and we're human and miss things in reviews, all good ofc!).
The original change w/o the nil check is here https://github.com/openshift/machine-config-operator/pull/1658/commits/e7455dcb4e0150e00f78e0ae4954b73047d1bf75#diff-fa45321336db7ad1cedc28bf643a4f97R107
Then, here's another (!!!) https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L322 (in master, doesn't affect 4.5, but as we can notice, we aren't checking for nil there)
Then we can instead notice https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/render.go#L80-L91 within `createDiscoveredControllerConfigSpec` which is the first thing we do after grabbing the infra object. In those lines we make sure that if the PlatformStatus is nil, we "adjust" it for compatibility reasons and then we store the infra object in the ControllerConfig object (there's another bug there tho, unrelated, we're modifying an object from the indexers which is a no-no).
Anyway, it seems that after grabbing the initial infra object (https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L228), we do run it through a compatibility fix to make sure PlatformStatus is never nil from there (https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/render.go#L80-L91), perhaps a forward looking fix would be to:

- grab the infra object (we do this already)
- run it through `createDiscoveredControllerConfigSpec` (we do this already)
- pass the resulting `ControllerConfig.Spec.Infra...` object in funcs like this and `getIgnitionHost`, instead of the `infra` global one, as we're sure that the infra in the CC already went through this very same nil check (we need to do this)

The above allows us to avoid adding specific nil checks for the same thing and help us tomorrow when the next PR adds something that access PlatformStatus w/o checking it

pkg/operator/sync.go
```diff
@@ -112,13 +112,18 @@ func isCloudConfigRequired(infra *configv1.Infrastructure) bool {
...
 	}
-	switch infra.Status.PlatformStatus.Type {
-	case configv1.AzurePlatformType,
-		configv1.GCPPlatformType,
-		configv1.OpenStackPlatformType,
-		configv1.VSpherePlatformType:
-		return true
-	default:
+
+	if infra.Status.PlatformStatus != nil {
```
runcom, line 116: Great job narrowing this down!

the change in `lib/resourcemerge/machineconfig.go` LGTM but following the trails for the panic indicates that our code isn't that robust to changes and we don't have units for all the functions that access the field being potentially nil (and we're human and miss things in reviews, it happens).
The original change w/o the nil check is here https://github.com/openshift/machine-config-operator/pull/1658/commits/e7455dcb4e0150e00f78e0ae4954b73047d1bf75#diff-fa45321336db7ad1cedc28bf643a4f97R107
Then, here's another (!!!) https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L322 (in master, doesn't affect 4.5, but as we can notice, we aren't checking for nil there and we're missing it again)

Then we can instead notice https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/render.go#L80-L91 within `createDiscoveredControllerConfigSpec` which is the first thing we do after grabbing the infra object. In those lines we make sure that if the PlatformStatus is nil, we "adjust" it for compatibility reasons and then we store the infra object in the ControllerConfig object which makes sure we can safely check that field as we made sure it's always non-nil (there's another bug there tho, unrelated, we're modifying an object from the indexers which is a no-no).
Anyway, it seems that after grabbing the initial infra object (https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L228), we do run it through a compatibility fix to make sure PlatformStatus is never nil from there (https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/render.go#L80-L91), perhaps a forward-looking fix for this BZ would be to:
- grab the infra object (we do this already)

- run it through `createDiscoveredControllerConfigSpec` (we do this already)
- pass the resulting `ControllerConfig.Spec.Infra...` object in funcs like this and `getIgnitionHost`, instead of the `infra` global one, as we're sure that the infra in the CC already went through this very same nil check (we need to do this)

With the above we:

- centralize the nil check (also there's a comment telling us why it could be nil, the key thing for this BZ seems to be an upgrade all the way from 4.1 which didn't have PlatformStatus)
- adjust it so it's never nil in our functions

The above allows us to avoid adding specific nil checks for the same thing over and over and help us tomorrow when the next PR adds something that access PlatformStatus w/o checking it

kikisdeliveryservice: sounds good!

cgwalters: Me imagining what Antonio means when he says on Twitter he's going to take some time to do some writing: Antonio sitting at his computer, sipping an espresso and writing a gripping Italian crime novel :book: 

The actual result: Antonio sitting at his computer, sipping an espresso and writing longer pull request review comments :wink:

runcom: @cgwalters 😂 😂 😂 😂 😂 😂 😂 😂 😂 😂 😂 😂  it was fun to imagine myself wiring a crime novel in from of my laptop sipping an expresso tho! thanks for that Colin! 😂

PR #1937, opened 2020-07-20
pools: add default label for master/worker pool

yuqi-zhang:
Today the master/worker MCP does not have any different labels.
This means that if e.g. a user wants to add a kubeletconfig, they
must first create a label for the pool and target the kubeletconfig
against the label. This PR creates a default label for master/worker
pools to skip that step.

### DIFF
--- a/manifests/master.machineconfigpool.yaml
+++ b/manifests/master.machineconfigpool.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     "operator.machineconfiguration.openshift.io/required-for-upgrade": ""
     "machineconfiguration.openshift.io/mco-built-in": ""
+    "pools.operator.machineconfiguration.openshift.io/master": ""
 spec:
   machineConfigSelector:
     matchLabels:
--- a/docs/ContainerRuntimeConfigDesign.md
+++ b/docs/ContainerRuntimeConfigDesign.md
@@ -68,33 +68,35 @@ metadata:
 spec:
  machineConfigPoolSelector:
    matchLabels:
+     pools.operator.machineconfiguration.openshift.io/worker: ""
+ containerRuntimeConfig:
+   pidsLimit: 2048
+```
+Save your `ctrcfg` locally, for example as highpids.yaml
+
+The label in the above example corresponds to the worker MachineConfigPool. By default the master/worker
+MachineConfigPool has labels pools.operator.machineconfiguration.openshift.io/{worker|master}: "" in OCP 4.6 and later. If you have a custom pool, or have an earlier OCP version, you can instead create a label youself as follows:
+
+```
+apiVersion: machineconfiguration.openshift.io/v1
+kind: ContainerRuntimeConfig
+metadata:
+ name: set-pids-limit
+spec:
+ machineConfigPoolSelector:
+   matchLabels:
      custom-crio: high-pid-limit
  containerRuntimeConfig:
    pidsLimit: 2048
 ```
-Save your `ctrcfg` locally, for example as highpids.yaml and create the ctrcfg:
 
-```
-oc create -f highpids.yaml
-```
-
-Check that it was created:
-
-```
-$ oc get ctrcfg
-NAME             AGE
-set-pids-limit   50s
-```
-
-Use the label you created in `matchLabels` in the MachineConfigPool config that you want the changes rolled out to. From the example above, that label would be `custom-crio: high-pid-limit`
-
-To roll out the pids limit changes to all the worker nodes (can switch this to master for the master nodes), add `custom-crio: high-pid-limit` under labels in the machineConfigPool config.  
+To roll out the pods limit changes to all the worker nodes (can switch this to master for the master nodes), add the label that you created, here: `custom-crio: high-pid-limit` under labels in the MachineConfigPool config:
 
 ```
 oc edit machineconfigpool worker
 ```
 
-Snippet of the machineConfigPool config with the matching label added:
+Snippet of the MachineConfigPool config with the matching label added:
 
 ```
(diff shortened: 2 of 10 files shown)


cgwalters:
Hmm, the kubeletconfig can't target a name directly?  It seems weird to add labels that just match our names.

yuqi-zhang:
> Hmm, the kubeletconfig can't target a name directly? It seems weird to add labels that just match our names.
by design it searches the MCPs for a label that it wants to match

2 notes:
1. it seems that our labels are all quoted (`"machineconfiguration.openshift.io/mco-built-in": ""` instead of `machineconfiguration.openshift.io/mco-built-in: ""`), whereas many other repos are not. Was there a reason for this?

2. this only considers master/worker mcp and does not consider custom pools, which isn't synced like master/worker and probably needs a minor rework if we want to do that too

yuqi-zhang:
To be more specific on my previous point, I mean that we use labels to select the pool here: https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/kubelet_config_controller.go#L607 . It's not a direct kubeletconfig-pool matching. Would you say we should redesign that?

yuqi-zhang:
/test e2e-aws-proxy

Just doing this for an environment, please ignore

runcom:
> 3 notes:
> 
> 1. it seems that our labels are all quoted (`"machineconfiguration.openshift.io/mco-built-in": ""` instead of `machineconfiguration.openshift.io/mco-built-in: ""`), whereas many other repos are not. Was there a reason for this?

uhm, I can't remember anything there :thinking: is it causing any specific issue?

> 2\. this only considers master/worker mcp and does not consider custom pools, which isn't synced like master/worker and probably needs a minor rework if we want to do that too

I think we can target custom pools as well as we're growing further support as first-class citizen (stub ignition creation and the like)

yuqi-zhang:
Adding it as a template instead, so it should get applied to all pools. Tested with master/worker, will test with custom pools to make sure

install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml
```diff
@@ -69,6 +69,10 @@ spec:
     # shortNames allow shorter string to match your resource on the CLI
     shortNames:
     - mcp
+  template:
+    metadata:
+      labels:
+          pools.operator.machineconfiguration.openshift.io/{{ status.configuration.name }}: ""
```
runcom, line 75: omg niceee

yuqi-zhang:
I can confirm that this works for installs/upgrades for master/worker pools, however the template doesn't seem to apply for custom pools by default. Maybe I'm mis-understanding the templating mechanism?

install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml
deads2k: > omg niceee

hmm..... What do you think this does?

yuqi-zhang: Having a default label applied to each MCP object being created, do you know what the correct way to do this is?

runcom:
alrighty, the master/worker MCP are still something managed by MCO and it makes sense to have default labels applied. Custom pools on the other hand are user-provided and it makes sense to follow the pattern in the default pool manifests (maybe we'll find a way to do that). This is an optimization for the default pools anyway, the old behavior is untouched as per acceptance criteria in the jira card

runcom:
@openshift/openshift-team-mco-maintainers ptal

/approve

yuqi-zhang:
I've changed it back to just adding a label for master/workers based on their manifests for now. In the long term we may want to rework the controller to manage the label

manifests/master.machineconfigpool.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   labels:
     "operator.machineconfiguration.openshift.io/required-for-upgrade": ""
     "machineconfiguration.openshift.io/mco-built-in": ""
+    pools.operator.machineconfiguration.openshift.io/master: ""
```
ericavonb, line 8: for consistency, and because I don't trust yaml, should we put the key in quotes like the other labels?

yuqi-zhang:
Added quotes to labels. Although I feel like we should remove them like other openshift repos, for now lets keep them consistent

ericavonb:
> To be more specific on my previous point, I mean that we use labels to select the pool here: https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/kubelet_config_controller.go#L607 . It's not a direct kubeletconfig-pool matching. Would you say we should redesign that?

Wait, KubeletConfigs specify MachineConfigPools via label selectors to create MachineConfigs selected by that pool via label selectors? 😱  Isn't this backwards?

Except that it sorta assumes this'll work based on the pool name. A lot of indirection going on. KubeletConfigs should not be aware of MachineConfigPools at all, nor should they rely on any sort of naming or labeling, and they definitely shouldn't be aware of templates or rendering logic. Gah, this deserves an issue at least

yuqi-zhang:
We can add that as part of the KCC/CRCC rework items. What would you recommendation be in terms of hooking up KC/CRC to MCs?

yuqi-zhang:
This PR just aims to add a default. We can always remove it if we rework the behaviour

ericavonb:
> This PR just aims to add a default. We can always remove it if we rework the behaviour

👍 sorry, yes this PR seems fine by me. 

Shouldn't have polluted the PR like that! Sorry again.

/lgtm

yuqi-zhang:
No worries at all! Thanks for the review :D

PR #1938, opened 2020-07-20
[release 4.4] Bug 1855878:  cherry-pick hostname fixes

darkmuggle:

     - localhost as the hostname

Fixes BZs 1844613, 1845885, 1853400, 1853584 (and probably more)

Includes commits:

### DIFF
--- a/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
+++ b/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
@@ -1,0 +1,84 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/sbin/set-valid-hostname.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # On some platforms the hostname may be too long (>63 chars).
+    #  - On firstboot the hostname is set in the initramfs before NetworkManager
+    #    And it may be truncated at 64 characters (too long)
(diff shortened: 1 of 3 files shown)


darkmuggle:

darkmuggle:
I'm not sure I can make Bugzilla happy with the various bugs that are referenced in the clone. I might just have create a "backport all the hostname fixes" that is not a cloned bug.

kikisdeliveryservice:
> I'm not sure I can make Bugzilla happy with the various bugs that are referenced in the clone. I might just have create a "backport all the hostname fixes" that is not a cloned bug.

Backports need a master BZ so that probably won't work, can you not just backport them individually?

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,83 @@
+mode: 0755
```
kikisdeliveryservice, line 1: need filesystem :)

darkmuggle: Does it? The other templates lack it. Digging around in the other templates, most -- if not all -- lack it too.

cgwalters:
/lgtm
On one hand: As far as we know this equivalent code has been working in 4.5/4.6 
But, on the other hand: this is also not a trivial patch to send to an existing stream.

PR #1939, opened 2020-07-20
[release 4.5] Bug 1855877: cherry-pick hostname fixes

darkmuggle:

     - localhost as the hostname

Fixes BZ 1855878

Includes commits:

### DIFF
--- a/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
+++ b/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
@@ -1,0 +1,84 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/sbin/set-valid-hostname.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # On some platforms the hostname may be too long (>63 chars).
+    #  - On firstboot the hostname is set in the initramfs before NetworkManager
+    #    And it may be truncated at 64 characters (too long)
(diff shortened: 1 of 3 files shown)


darkmuggle:

kikisdeliveryservice:
Pretty sure this is failing unit tests bc in master we had ignition spec3 but 4.5 and less have ign 2. So you probably have  to figure out what changed in these folders you touched between master/4.5 to make sure it can work with the lower version(that change should hold for 4.3/4.4/4.5).

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,83 @@
+mode: 0755
```
kikisdeliveryservice, line 1: You need a filesystem here, it's reqd in ign spec2

darkmuggle: fixed :)

kikisdeliveryservice:
```
 level=fatal msg="failed to initialize the cluster: Working towards 0.0.1-2020-07-21-142823: 87% complete, waiting on authentication"
---
time="2020-07-21T15:30:13Z" level=info msg="Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:executing_test:executing_template\"" 
```

Let's try again

/retest

PR #1940, opened 2020-07-20
[release-4.3] Bug 1855879: cherry-pick hostname fixes

darkmuggle:

     - localhost as the hostname

Fixes BZ 1855879

Includes commits:

This is needed due to the guard condition in the installer being rolled back at https://github.com/openshift/installer/pull/3644.

### DIFF
--- a/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
+++ b/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
@@ -1,0 +1,84 @@
+filesystem: "root"
+mode: 0755
+path: "/usr/local/sbin/set-valid-hostname.sh"
+contents:
+  inline: |
+    #!/bin/bash
+    # On some platforms the hostname may be too long (>63 chars).
+    #  - On firstboot the hostname is set in the initramfs before NetworkManager
+    #    And it may be truncated at 64 characters (too long)
+    #  - On reboot affect nodes use 'localhost'.
+    #
+    # This script is a simple workaround for hostname woes, including
+    #  - NOT a localhost name
+    #  - NOT longer than 63 characters. Names will be truncated at the
+    #    first dot, and then capped at 63 char (which ever is less).
+    #  - Race conditions between truncated hostnames by the dhclient
+    #    and NetworkManager.
+    #
+    # Finally, this script is invoked via:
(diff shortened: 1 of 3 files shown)


darkmuggle:

kikisdeliveryservice:
Is this backport necessary for 4.3 bc the BZ is only severity medium?

darkmuggle:
> Is this backport necessary for 4.3 bc the BZ is only severity medium?

We either take this backport, rollback openshift/installer#3644 or accept that GCP clusters are *likely* to be uninstallable. The number of 4.3 bugs that have come into RHCOS related to hostnames has increased with he 4.4 and 4.5 rollback of the guard condition.

templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml
```diff
@@ -0,0 +1,83 @@
+mode: 0755
```
kikisdeliveryservice, line 1: need filesystem

darkmuggle: And fixed :)

cgwalters (approved):
Approval really comes down to the patch queue manager, I'm adding tentative approval but I think a bottom line here is: Are there really organizations out there doing *new* 4.3 installations that can't use 4.4 or 4.5 for some reason?  That's what I'd like to see.  We are running this code in 4.4 and above already, but the flip side is customers can get that code by upgrading.

cgwalters:
And now that I look https://github.com/openshift/installer/pull/3644 was closed so it seems like we should close this too?

miabbott:
4.3 entered Maintenance Support phase on June 5

From https://access.redhat.com/support/policy/updates/openshift

> During the Maintenance Support phase, qualified Critical and Important Security Advisories (RHSAs) and Urgent and Selected High Priority Bug Fix Advisories (RHBAs) may be released as they become available. Other Bug Fix (and Enhancement (RHEA) Advisories may be released at Red Hat’s discretion, but should not be expected.

I think this PR and associated BZ can be closed

darkmuggle:
There was some question as to whether this would qualify for a maintenance release for 4.3. The backport was requested by support, but this has been sitting for a while without cherry-pick approval. I concur with closing this now since it has not been approved for cherry-picking.

PR #1941, opened 2020-07-21
Support installing extensions and rework OS update

sinnykumari:
Earlier to perform OS update we were pulling OS image
and mounting the content using podman. We were performing
OS update in host context because of selinux constraints
on mounted container.

Also, for rt-kernel switch we were pulling OS image again.
With coreos extensions support, we require extensions rpm
which is available in os container.

We tried different approach to solve problems
like minimizing container image pull, host/container
context switch, selinux permission on mounted
container and rpm-ostreed behavior.
See https://github.com/openshift/machine-config-operator/pull/1850

Finally, using `oc image extract` and skopeo to inspect container
image solves our problems. With this we are also getting
rid of using mco-pivot systemd service to rebase OS.


sinnykumari:
Note that with this PR we don't need to do `m-c-d pivot` . So, if we require tunable args then we can  either update update()->updateKernelArguments() accordingly or update our docs to perform tunable args via MachineConfigs.

pkg/daemon/update.go
```diff
@@ -255,7 +224,41 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
...
+func addExtensionsRepo() error {
+	if err := os.MkdirAll(filepath.Dir(extensionsRepo), 0755); err != nil {
+		return fmt.Errorf("error creating yum repo directory %s: %v", filepath.Dir(extensionsRepo), err)
+	}
+	repo, err := os.Create(extensionsRepo)
+	if err != nil {
+		return fmt.Errorf("error creating extensions repo %s: %v", extensionsRepo, err)
+	}
+	defer repo.Close()
+	if _, err := repo.WriteString("[coreos-extensions]\nenabled=1\nmetadata_expire=1m\nbaseurl=" + osImageContentDir + "/extensions/\ngpgcheck=0\nskip_if_unavailable=False\n"); err != nil {
```
runcom, line 236: you could use `writeFileAtomically`

pkg/daemon/update.go
```diff
@@ -698,37 +749,60 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
...
+	if dn.OperatingSystem != MachineConfigDaemonOSRHCOS {
+		return fmt.Errorf("Extensions is not supported on non-RHCOS nodes ")
 	}
-	mnt = strings.TrimSpace(string(mntBuf))
-	return
+
+	args := generateExtensionsArgs(oldConfig, newConfig)
+	glog.Infof("Applying extensions : %+q", args)
+	if err := exec.Command("rpm-ostree", args...).Run(); err != nil {
+		return fmt.Errorf("Failed to execute rpm-ostree %+q : %v", args, err)
```
runcom, line 801: nit: errors start lowercase (glog uppercase)

pkg/daemon/rpm-ostree.go
```diff
@@ -292,12 +237,6 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
 		return
 	}
 
-	// By default, delete the image.
-	if !keep {
-		// Related: https://github.com/containers/libpod/issues/2234
-		exec.Command("podman", "rmi", imgid).Run()
-	}
```
runcom, line 299: are we leaving `osImageContentDir` around? is there any safe place we could get rid of it? (just thinking out loud)

sinnykumari: osImageContentDir is keeping machine-os-content extracted using `oc extract image` in  a temporary dir /run/mco-machine-os-content/ . This will go away once system reboots after update.

runcom: Nice! I missed that, thanks Sinnny :)

sinnykumari: np :)

pkg/daemon/rpm-ostree.go
```diff
@@ -164,87 +154,41 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
...
+	inspectArgs = append(inspectArgs, authArgs...)
 
-	inspectArgs := []string{"inspect", "--type=image"}
-	inspectArgs = append(inspectArgs, fmt.Sprintf("%s", container))
+	// Map container image name with transport so that skopeo understands it
+	imageName := "docker://" + container
+	inspectArgs = append(inspectArgs, imageName)
 	var output []byte
-	output, err = runGetOut("podman", inspectArgs...)
+	output, err = runGetOut("skopeo", inspectArgs...)
```
cgwalters, line 164: This is going to re-introduce https://github.com/openshift/pivot/pull/51

pkg/daemon/rpm-ostree.go
```diff
@@ -164,87 +154,41 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
...
-	}
-	defer func() {
-		// Kill our dummy container
-		podmanRemove(containerName)
-	}()
-
-	cid := strings.TrimSpace(string(cidBuf))
-	// Use the container ID to find its mount point
-	var mntBuf []byte
-	mntBuf, err = runGetOut("podman", "mount", cid)
```
cgwalters, line 239: Aren't we missing a call to the extract function here?

sinnykumari: Hopefully I understood your question correctly.
We are now not mounting container image. https://github.com/openshift/machine-config-operator/pull/1941/files#diff-06961b075f1753956d802ba954d2cfb5R251 is where we extract OS container content in /run/mco-machine-os-content/ and use same for rpm-ostree rebase.

pkg/daemon/rpm-ostree.go
sinnykumari: I went skopeo route to save pulling container image as it is required by podman inspect. Is skopeo inspect more expensive than doing `podman pull containerImg && podman inspect` ?

I was also wondering if we can just use https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/rpm-ostree.go#L257,L280 to get checksum and rebase. It works but not sure if we hit issue in some situation.

cgwalters: I think we actually have the containers/image stack vendored into the MCO for other reasons, so we could just directly do that from our process.  Another alternative looks like `oc image info -o json`.

cgwalters: Or we could try to figure out if there's a way to get skopeo to not try to enumerate all the tags.

pkg/daemon/update.go
```diff
@@ -353,6 +356,37 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+			return err
+		}
+		if dn.OperatingSystem == MachineConfigDaemonOSRHCOS {
+			if err := addExtensionsRepo(); err != nil {
+				return err
+			}
+		}
+	}
+
+	if mcDiff.osUpdate {
```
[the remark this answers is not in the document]
sinnykumari: I feel moving this with earlier section is going to be tricky and may lead to un-necessary bugs. Both section are doing different things and right order is really important here.

pkg/daemon/update.go
```diff
@@ -353,6 +356,37 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 		}
 	}()
 
+	// Extract image and add coreos-extensions repo if we have either OS update or package layering to perform
+	mcDiff, err := newMachineConfigDiff(oldConfig, newConfig)
+	if err != nil {
+		return err
+	}
+
+	if mcDiff.osUpdate || mcDiff.extensions || mcDiff.kernelType {
```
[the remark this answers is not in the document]
sinnykumari: Definitely, there is scope of moving OS update/layering section into a function.

pkg/daemon/update.go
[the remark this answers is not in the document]
sinnykumari: Actually `if` is not needed here as we also compare OSImageURL in updateOS()

sinnykumari:
Thank you all for reviewing!
Have pushed changes which should address all comments other than skopeo issue(digging into it)

pkg/daemon/update.go
```diff
@@ -255,7 +224,41 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
...
+		return fmt.Errorf("error creating extensions repo %s: %v", extensionsRepo, err)
+	}
+	defer repo.Close()
+	if _, err := repo.WriteString("[coreos-extensions]\nenabled=1\nmetadata_expire=1m\nbaseurl=" + osImageContentDir + "/extensions/\ngpgcheck=0\nskip_if_unavailable=False\n"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func extractOSImage(imgURL string) error {
```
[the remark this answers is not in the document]
sinnykumari: I feel that this is one of the things which we can discuss collectively to cleanup and refactor MCO code.

pkg/daemon/daemon.go
```diff
@@ -1004,7 +1004,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		if !osMatch {
 			glog.Infof("Bootstrap pivot required to: %s", targetOSImageURL)
 			// This only returns on error
```
jlebon, line 1006: Should move this comment down to the `dn.finalizeAndReboot` now I suppose?

pkg/daemon/update.go
```diff
@@ -255,6 +207,76 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
 	return true, nil
 }
 
+// addExtensionsRepo adds a repo into /etc/yum.repos.d/ which we use later to
+// install extensions and rt-kernel
+func addExtensionsRepo() error {
+	repoConetnt := "[coreos-extensions]\nenabled=1\nmetadata_expire=1m\nbaseurl=" + osImageContentDir + "/extensions/\ngpgcheck=0\nskip_if_unavailable=False\n"
```
jlebon, line 213: Why `gpgcheck=0` here? Shouldn't all the RPMs we ship via extensions be signed?

jlebon, line 213: Minor typo: `repoContent`

sinnykumari: Ah yes, good catch. Don't know why I kept it like that, most likely copy paste error.

pkg/daemon/daemon.go
sinnykumari: umm, we will return back here also when updateOS() fails?

pkg/daemon/rpm-ostree.go
sinnykumari: By looking at skopeo code, it doesn't seem like we can ask skopeo to skip looking for tags(have also asked on skopeo irc channel to get a second opinion)
Not very comfortable with us maintaining the code to inspect images, feels like it is better to use already tested tool `oc image info`

pkg/daemon/update.go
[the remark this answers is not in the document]
kikisdeliveryservice: Nice!

pkg/daemon/rpm-ostree.go
sinnykumari: After discussing with skopeo team it appears skipping to fetch RepoTags is not possible right now. With inspiration from skopeo inspect code, wrote an independent program which uses containers/image stack(already vendored in MCO) https://github.com/sinnykumari/container-inspect/blob/master/container_inspect.go and doesn't fetch RepoTags . If this looks ok I will integrate this in MCO, otherwise we can go for `oc image info`

pkg/daemon/update.go
sinnykumari: We had an offline discussion, but putting a comment here for future reference:
It is possible that some packages may not be signed, especially when dealing with nightly images. We are anyway shipping extensions packages in machine-os-content(which is source of truth for us). So it should be ok to install them with gpgcheck off.

pkg/daemon/rpm-ostree.go
sinnykumari: Pushed changes to use containers/image stack directly for inspecting image.

sinnykumari:
@mtrmac Thank you for the help with implementing container inspect. https://github.com/openshift/machine-config-operator/pull/1941/commits/28742e302dbf34e4fdad98dcb95ade524c47330b is how currently I have implemented in MCO, let us know if you have any additional feedback.

sinnykumari:
Tests seems to be flaky
/retest

pkg/daemon/container-inspect.go
```diff
@@ -0,0 +1,77 @@
+package daemon
```
mtrmac, line 1: Shouldn’t the file be named `image-inspect.go`?

pkg/daemon/container-inspect.go
```diff
@@ -0,0 +1,77 @@
...
+			break
+		case <-ctx.Done():
+			return err
+		}
+		err = operation()
+	}
+	return err
+}
+
+// parseImageSource converts image URL-like string to an ImageSource.
```
mtrmac, line 31: With `docker.ParseReference` it is no longer very URL-like; and I would expect this function to handle the `"//"` prefix instead of callers having to care. Ideally, this should “convert an image reference”, or “create an `ImageSource` from an image reference”.

pkg/daemon/container-inspect.go
```diff
@@ -0,0 +1,77 @@
...
+// The caller must call .Close() on the returned ImageSource.
+func parseImageSource(ctx context.Context, name string) (types.ImageSource, error) {
```
mtrmac, line 33: Absolutely non-blocking: `newDockerImageSource` maybe; the “parse” name was used in context of a fairly general CLI that is not applicable in this call tree.

pkg/daemon/container-inspect.go
```diff
@@ -0,0 +1,77 @@
...
+	ref, err := docker.ParseReference(name)
+	if err != nil {
+		return nil, err
+	}
+	sys := &types.SystemContext{AuthFilePath: kubeletAuthFile}
```
mtrmac, line 38: The caller can pass this down instead of having two copies that can get out of sync.

go.mod
```diff
@@ -8,23 +8,20 @@ require (
 	github.com/Masterminds/goutils v1.1.0 // indirect
 	github.com/Masterminds/semver v1.4.2 // indirect
 	github.com/Masterminds/sprig v2.20.0+incompatible
-	github.com/Microsoft/go-winio v0.4.14 // indirect
 	github.com/OpenPeeDeeP/depguard v1.0.1 // indirect
 	github.com/apparentlymart/go-cidr v1.0.0
 	github.com/ashcrow/osrelease v0.0.0-20180626175927-9b292693c55c
 	github.com/clarketm/json v1.14.1
 	github.com/containers/image v3.0.2+incompatible
```
mtrmac, line 15: Hum, container-runtime-config (and openshift/runtime-utils) still use the old version. That should probably be updated to decrease duplication… but it’s also rather out of scope of this PR.

pkg/daemon/rpm-ostree.go
```diff
@@ -159,96 +133,19 @@ func (r *RpmOstreeClient) PullAndRebase(container string, keep bool) (imgid stri
...
-	inspectArgs = append(inspectArgs, fmt.Sprintf("%s", container))
-	var output []byte
-	output, err = runGetOut("podman", inspectArgs...)
-	if err != nil {
-		return
-	}
-	var imagedataArray []imageInspection
-	err = json.Unmarshal(output, &imagedataArray)
+	var imageData *types.ImageInspectInfo
+	imageData, err = imageInspect(fmt.Sprintf("//%s", container))
```
mtrmac, line 137: As mentioned elsewhere, I’d expect the `"//"` syntax to be hidden in the implementation.

go.mod
sinnykumari: yeah, we can clean this up separately.

sinnykumari:
> (Reading only the image inspect part.) ACK, just trivial nits.

Thanks for the review, pushed changes addressing the feedback.

sinnykumari:
e2e-gcp-op test is green! I have tested day1 operations locally which works as expected.
Unless there is something broken, let's get this reviewed and merged and we can address minor changes in another PR.

pkg/controller/common/helpers.go
```diff
@@ -285,12 +298,41 @@ func ValidateIgnition(ignconfig interface{}) error {
...
+	for _, k := range slice {
+		if k == elem {
+			return true
+		}
+	}
+	return false
+}
+
+func validateExtensions(exts []string) error {
+	supportedExtensions := []string{"usbguard", "kernel-devel"}
```
ericavonb, line 312: Could these be constants exported alongside the MC api types? If we could even add them to the CRD's openapi schema that'd be even better, so that users get immediate feedback on invalid extensions.

pkg/daemon/image-inspect.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+	"time"
+
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/image"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+)
+
+func retryIfNecessary(ctx context.Context, operation func() error) error {
```
ericavonb, line 16: Any reason not to use [k8s.io/client-go/util/retry](https://pkg.go.dev/k8s.io/client-go/util/retry)?

pkg/daemon/update.go
```diff
@@ -255,6 +207,76 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
...
+	if err := dn.updateOS(newConfig); err != nil {
+		return err
+	}
+
+	// Apply kargs
+	if err := dn.updateKernelArguments(oldConfig, newConfig); err != nil {
+		return err
+	}
+
+	// Apply extensions
```
ericavonb, line 266: Wrong comment :) (needs to be switched with https://github.com/openshift/machine-config-operator/pull/1941/files#diff-06961b075f1753956d802ba954d2cfb5R271 )

pkg/daemon/update.go
```diff
@@ -255,6 +207,76 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
...
+	if err != nil {
+		return err
+	}
+
+	if mcDiff.osUpdate || mcDiff.extensions || mcDiff.kernelType {
+		if err := extractOSImage(newConfig.Spec.OSImageURL); err != nil {
+			return err
+		}
+		if dn.OperatingSystem == MachineConfigDaemonOSRHCOS {
+			if err := addExtensionsRepo(); err != nil {
```
ericavonb, line 250: Why do we want to do this here rather than if/when we apply extensions in [`applyExtensions`](https://github.com/openshift/machine-config-operator/pull/1941/files#diff-06961b075f1753956d802ba954d2cfb5R747)

cgwalters:
> Could these be constants exported alongside the MC api types? If we could even add them to the CRD's openapi schema that'd be even better, so that users get immediate feedback on invalid extensions.

I'm a bit uncertain about requiring an MCO patch when an extension is added.  The immediate validation would be nice though for sure.

pkg/daemon/update.go
ericavonb: I don't know that it's ok to turn off signature verification. It seems like it leaves open a pretty significant security hole that attackers could exploit to run untrusted code on the host. While we check the machine-os-content image signature, we don't have any checks on the integrity of the files in the extensions directory ensuring that the RPMs that get installed are same ones shipped in the machine-os-image without alteration. There are likely other security gaps that could be enabled by installing unverified RPMs.

At the very least, I'm pretty sure many customers in highly-regulated industries require every package to be signed and validated to run on a host.

ericavonb:
> I'm a bit uncertain about requiring an MCO patch when an extension is added. The immediate validation would be nice though for sure.

Isn't that what we're requiring with this validation function anyways? https://github.com/openshift/machine-config-operator/pull/1941/files#diff-5926e0cec606d949b66f823310790298R311

pkg/daemon/update.go
cgwalters: > While we check the machine-os-content image signature, we don't have any checks on the integrity of the files in the extensions directory ensuring that the RPMs that get installed are same ones shipped in the machine-os-image without alteration. 

We do, because we pull-by-digest and the container stack verifies that the digest matches.

cgwalters:
> Isn't that what we're requiring with this validation function anyways? https://github.com/openshift/machine-config-operator/pull/1941/files#diff-5926e0cec606d949b66f823310790298R311

Yep definitely, but adding it to the openAPI more strongly commits us to that.  I'm uncertain.  I think we would be fine not doing openapi validation and instead failing later (ideally in the render controller, but failing that in the MCD).  But I'd also be fine adding openapi, we'd just have to live with the tradeoff of needing to re-roll the MCO every time an extension is added.

pkg/daemon/update.go
```diff
@@ -46,26 +43,11 @@ const (
 	// SSH Keys for user "core" will only be written at /home/core/.ssh
 	coreUserSSHPath = "/home/core/.ssh/"
 	// fipsFile is the file to check if FIPS is enabled
-	fipsFile = "/proc/sys/crypto/fips_enabled"
+	fipsFile          = "/proc/sys/crypto/fips_enabled"
+	extensionsRepo    = "/etc/yum.repos.d/coreos-extensions.repo"
+	osImageContentDir = "/run/mco-machine-os-content/"
```
cgwalters, line 48: I don't see any code removing this directory when we're done with it.  I know right now we're rebooting, but extracting to `/run` is going to take up RAM until it's deleted.  Let's use an `ioutil.TempDir()` underneath there or so?

pkg/daemon/update.go
```diff
@@ -255,6 +207,76 @@ func (dn *Daemon) compareMachineConfig(oldConfig, newConfig *mcfgv1.MachineConfi
...
+func extractOSImage(imgURL string) error {
+	var registryConfig []string
+	if _, err := os.Stat(kubeletAuthFile); err == nil {
+		registryConfig = append(registryConfig, "--registry-config", kubeletAuthFile)
+	}
+	if err := os.MkdirAll(osImageContentDir, 0755); err != nil {
+		return fmt.Errorf("error creating directory %s: %v", osImageContentDir, err)
+	}
+	// Extract the image
+	args := []string{"image", "extract", "--path", "/:" + osImageContentDir}
```
cgwalters, line 229: In the future with https://github.com/openshift/machine-config-operator/pull/1947 I think we're going to want to wrap this in a `ionice -c 3` or so.  No need to do so in this patch, just noting.  This also argues for keeping the extraction in a separate process going forward (which could just be us re-exec'ing our own binary of course).

pkg/daemon/image-inspect.go
sinnykumari: I followed the convention of what skopeo has been doing to retry https://github.com/containers/skopeo/blob/153f18dc0ac34e1adfc700806ef30af33187908b/cmd/skopeo/utils.go#L330

pkg/daemon/update.go
sinnykumari: right :P

pkg/daemon/update.go
sinnykumari: Because extensions repo is also used to switch to or update kernel-rt packages. If you extract OSContainer, there is a extensions/ directory which has all additional rpm packages we ship and that includes kernel-rt packages as well. Maybe we can name the repo to something else but the actual repo directory which contains rpm has been named as extensions, so the name is in sync with that :)

pkg/controller/common/helpers.go
sinnykumari: Not very sure. The list of supported extensions is going to change whenever we add more extensions support in upcoming releases. In future, we are also exploring to add metadata for supported extension list in machine-os-content https://github.com/openshift/os/issues/409 which MCO can read from and validate before rendering.

pkg/daemon/update.go
sinnykumari: yeah, we can do that. Reason of not deleting explicitly was we extract the OSContainer just before we start applying OS changes. Once OS changes get applied, we reboot the node.

pkg/daemon/update.go
sinnykumari: will keep this in mind to fix in a follow-up PR

pkg/daemon/update.go
sinnykumari: Pushed changes to extract content under a temp dir and delete them explicitly.

pkg/daemon/update.go
sinnykumari: fixed

cmd/machine-config-daemon/pivot.go
```diff
@@ -231,7 +231,12 @@ func run(_ *cobra.Command, args []string) (retErr error) {
 
 	client := daemon.NewNodeUpdaterClient()
 
-	_, changed, err := client.PullAndRebase(container, keep)
+	var osImageContentDir string
+	var err error
+	if osImageContentDir, err = daemon.ExtractOSImage(container); err != nil {
+		return err
+	}
```
runcom, line 238: any reason not to do:
```go
osImageContentDir, err = daemon.ExtractOSImage(container)
if err != nil {
    return err
}
```

pkg/controller/common/helpers.go
```diff
@@ -285,12 +298,41 @@ func ValidateIgnition(ignconfig interface{}) error {
 	}
 }
 
+// InSlice search for an element in slice and return true if found, otherwise return false
+func InSlice(elem string, slice []string) bool {
```
runcom, line 302: can be unexported `inSlice` till we find a use for it (if any)

pkg/daemon/daemon.go
```diff
@@ -1004,7 +1004,18 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
-			return dn.updateOSAndReboot(state.currentConfig)
+			var osImageContentDir string
+			var err error
+			if osImageContentDir, err = ExtractOSImage(targetOSImageURL); err != nil {
+				return err
+			}
```
runcom, line 1011: same as https://github.com/openshift/machine-config-operator/pull/1941/files#r461643637

pkg/controller/common/helpers.go
sinnykumari: thanks for catching this, this was carried from work done in https://github.com/openshift/machine-config-operator/pull/1850. I missed updating inArray() call in update.go to InSlice(). Fixed https://github.com/openshift/machine-config-operator/pull/1941/files#diff-06961b075f1753956d802ba954d2cfb5R694

cmd/machine-config-daemon/pivot.go
sinnykumari: Not really, missed to refactor at the end in proper way. fixed

pkg/daemon/daemon.go
sinnykumari: fixed

kikisdeliveryservice:
@sinnykumari the cluster-bootimages failure should be fixed by:https://github.com/openshift/ci-tools/pull/1035

/test cluster-bootimages

runcom:
e2e-gcp-upgrade looks unrelated, the log shows the MCO is just fine https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1941/pull-ci-openshift-machine-config-operator-master-e2e-gcp-upgrade/1288353058928988160/artifacts/e2e-gcp-upgrade/clusteroperators.json

I'll check e2e-aws but no logs fwict

/retest
/skip

runcom:
e2e-aws actually passed yesterday on this very last commit as well `Test started yesterday at 5:17 PM passed after 2h17m3s. (more info)`

runcom:
okd-e2e-aws never passed also :thinking: wondering why it's not skippable also

sinnykumari:
> okd-e2e-aws never passed

@vrutkovs can you please take a look at why it is failing, don't find anything useful in the ci logs. If this PR is breaking somehow OKD, will look at  fixing it in next PR

>  also thinking wondering why it's not skippable also

ok, my thinking was right. It seems we can't skip while test is still running.

cmd/machine-config-daemon/pivot.go
vrutkovs: This breaks OKD as it doesn't have `oc` in FCOS image. From `release-image.log` on bootstrap node: 
```
Jul 29 08:25:43 ip-10-0-7-95 release-image-download.sh[1450]: I0729 08:25:43.638761    1450 run.go:18] Running: oc image extract --path /:/run/mco-machine-os-content/os-content-779087097 registry.svc.ci.openshift.org/ci-op-lx6rlf8w/stable@sha256:f5283b939e47e2158602d2954f9f1ce1e18ffafb2e711d4187f5954cacc050a6
Jul 29 08:25:43 ip-10-0-7-95 release-image-download.sh[1450]: W0729 08:25:43.638849    1450 run.go:44] oc failed: running oc image extract --path /:/run/mco-machine-os-content/os-content-779087097 registry.svc.ci.openshift.org/ci-op-lx6rlf8w/stable@sha256:f5283b939e47e2158602d2954f9f1ce1e18ffafb2e711d4187f5954cacc050a6 failed: : exec: "oc": executable file not found in $PATH; retrying...
```

vrutkovs:
>It seems we can't skip while test is still running.

It would automatically start on very PR, but its skippable via `/skip`.

This change however needs to have `pivot` change reverted, since it expects the bootimage to contain `oc`. Not sure if I should `/hold` it due to OKD

cmd/machine-config-daemon/pivot.go
runcom: decided to follow up thanks Vadim :pray: https://github.com/openshift/machine-config-operator/pull/1941#issuecomment-665538414

runcom:
> Jul 29 11:37:50.340 E clusteroperator/machine-config changed Degraded to True: MachineConfigPoolsFailed: Failed to resync 0.0.1-2020-07-29-104715 because: etcdserver: request timed out

Spotted the above in the last upgrade run - the MCO reconciles just fine anyway, but the above isn't the first time we see it and may be an indication of some other bug (somewhere else perhaps?)

/retest

sinnykumari:
gcp-upgrade test seems flaky
```
Jul 29 11:33:19.834: INFO: lookupDiskImageSources: gcloud error with [[]string{"instance-groups", "list-instances", "", "--format=get(instance)"}]; err:exit status 1
Jul 29 11:33:19.834: INFO:  > ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource [] 
```
/test e2e-gcp-upgrade

sinnykumari:
we need mutex here :D

cgwalters:
In looking at the PR logs for this I noticed something (unrelated but this PR changes a bit): https://github.com/openshift/machine-config-operator/issues/1964

sinnykumari:
e2e-gcp-op test is failing because of timeout of e2e-gcp-op has reached, see https://github.com/openshift/machine-config-operator/pull/1961#issuecomment-666543150 .  From the test run https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1941/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1289038515597742080/build-log.txt, we can see that Extensions test has passed successfully.

This has also been fixed in PR https://github.com/openshift/machine-config-operator/pull/1961/

Not sure what is going on with e2e-aws as they are failing today consistently across various MCO PRs.

Since these tests has already been succeeded once for given commit, does it make sense to skip them?

sinnykumari:
I have updated the pr to run Extensions tests using infra pool . Not other changes has been made.

runcom:
> I have updated the pr to run Extensions tests using infra pool . Not other changes has been made.

thanks Sinny, re-applying lgtm

/lgtm

runcom:
Thanks Sinny for the quick test fix also

/lgtm

runcom:
> /lgtm

nevermind tho - we'd still need a fix for the RT kernel test and a rebase here anyway

sinnykumari:
Rebased to get e2e-gcp-op fixes from https://github.com/openshift/machine-config-operator/pull/1969 included

sinnykumari:
needs lgtm blessing :)

sinnykumari:
e2e-aws is failing consistently today across various PRs.
If it makes sense let's override it.

kikisdeliveryservice:
> e2e-aws is failing consistently today across various PRs.
> If it makes sense let's override it.

The PR is already approved by the cutoff, we usually don't override tests just to merge, they usually end up merging over the weekend. This PR hasn't passed this test since its initial run on Tuesday so I think we should just let the bot retest so it can pass.

PR #1942, opened 2020-07-22
Update kernel args on firstboot-complete

vrutkovs:
This commits ensures kernel args modifications from /etc/pivot/kernel-args
are being applied on both pivot and firstboot_complete_machineconfig. Currently its only applied on bootstrap (which uses `pivot`), but not on masters/workers. In order to archieve that kargs tuning logic is moved from `cmd/` package to `pkg/daemon`.

OKD requires this so that "mitigations" karg could be removed from cmdline,
as MachineConfig's `kernelArguments` doesn't support removing existing arguments. This should also affect OCP with HyperThreading off, as machineconfigs for that still use `/etc/pivot/kernel-args`

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -544,6 +544,14 @@ func (dn *Daemon) RunFirstbootCompleteMachineconfig() error {
 		return errors.Wrap(err, "failed to rename encapsulated MachineConfig after processing on firstboot")
 	}
 
+	tuningChanged, err := UpdateTuningArgs(KernelTuningFile, CmdLineFile)
+	if err != nil {
+		return err
+	}
+	if !tuningChanged {
+		glog.Info("No changes; already at target oscontainer, no kernel args provided")
+	}
+
 	dn.skipReboot = false
 	return dn.reboot(fmt.Sprintf("Completing firstboot provisioning to %s", mc.GetName()))
 }
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -1,12 +1,10 @@
 package main
 
 import (
-	"bufio"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


cgwalters:
This is strongly related to https://github.com/openshift/machine-config-operator/pull/1925

OK right, I understand this now.

I think what would be a lot simpler and get us away from the `/etc/pivot` stuff is to just hardcode in the MCD:

```
if os == FCOS {
  run(rpm-ostree kargs --remove mitigations=auto,nosmt)
}
```

But for now
/approve
/lgtm

vrutkovs:
>I think what would be a lot simpler and get us away from the /etc/pivot stuff is to just hardcode in the MCD

We'd need to include hyperthreading off case for RHCOS too (its not widely used though)

pkg/daemon/daemon.go
```diff
@@ -544,6 +544,14 @@ func (dn *Daemon) RunFirstbootCompleteMachineconfig() error {
 		return errors.Wrap(err, "failed to rename encapsulated MachineConfig after processing on firstboot")
 	}
 
+	tuningChanged, err := UpdateTuningArgs(KernelTuningFile, CmdLineFile)
```
sinnykumari, line 547: A bit late to see this PR. Having this functionality here is going to work only during firstboot. Isn't it that users who maybe using this functionality would want to add/delete these kargs later on as well? Since, these are not applied using MachineConfig, there is no standard way to undo changes done during firstboot.

vrutkovs: Right, removing default kargs via MCs should be solved in https://issues.redhat.com/browse/GRPA-2490

sinnykumari: sorry for the noise, I missed that we already do it at https://github.com/openshift/machine-config-operator/pull/1942/files#diff-06b532a2a5a5d3ad1d5341b3e00945efR547

PR #1943, opened 2020-07-22
[Baremetal] Set Keepalived default mode to unicast

yboaron:
Set Keepalived default mode to unicast for Baremetal case


yboaron:
@kikisdeliveryservice  Do you think we can merge this PR?

cybertron:
/lgtm

I tried this locally and it worked, so let's do it.

PR #1945, opened 2020-07-23
Bug 1859230: operator/sync: update getIgnitionHost to account for nil PlatformStatus

kikisdeliveryservice:
We will be updating this with a more robust nil check later tp avoid these piecemeal checks in the next Sprint I believe, but this fixes the last outstanding platformStatus nil check.

No backport this func is new in 4.6.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -324,18 +324,21 @@ func getIgnitionHost(infraStatus *configv1.InfrastructureStatus) (string, error)
 	}
 	securePortStr := strconv.Itoa(server.SecurePort)
 	ignitionHost := fmt.Sprintf("%s:%s", internalURLParsed.Hostname(), securePortStr)
-	switch infraStatus.PlatformStatus.Type {
-	case configv1.BareMetalPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.BareMetal.APIServerInternalIP, securePortStr)
-	case configv1.OpenStackPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.OpenStack.APIServerInternalIP, securePortStr)
-	case configv1.OvirtPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.Ovirt.APIServerInternalIP, securePortStr)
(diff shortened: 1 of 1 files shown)


pkg/operator/sync.go
```diff
@@ -324,18 +324,21 @@ func getIgnitionHost(infraStatus *configv1.InfrastructureStatus) (string, error)
...
-	case configv1.BareMetalPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.BareMetal.APIServerInternalIP, securePortStr)
-	case configv1.OpenStackPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.OpenStack.APIServerInternalIP, securePortStr)
-	case configv1.OvirtPlatformType:
-		ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.Ovirt.APIServerInternalIP, securePortStr)
-	case configv1.VSpherePlatformType:
-		if infraStatus.PlatformStatus.VSphere.APIServerInternalIP != "" {
-			ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.VSphere.APIServerInternalIP, securePortStr)
+	if infraStatus.PlatformStatus != nil {
```
abhinavdahiya, line 327: Wouldn't it be easier to read if return early on nil platform status instead of switch block inside

cgwalters:
LGTM though I also agree with https://github.com/openshift/machine-config-operator/pull/1945#discussion_r459858286

Also I think a lesson here is in the MCO let's give extra review to any code which touches the "baremetal and friends" code; the ones with special tweaks that we aren't doing on the azure/aws/gcp clouds.

cgwalters:
Eh I'm just going to 
/lgtm
since I'd really like to get the Ignition work in and this is just a nit.  We can always do cleanups as followups!

kikisdeliveryservice:
> Eh I'm just going to 
> /lgtm
> since I'd really like to get the Ignition work in and this is just a nit.  We can always do cleanups as followups!
> 

Thanks @cgwalters! I'm going to overhaul all of the nil checks as @runcom suggested (have jira card) in the near future. Just felt more expedient to do this to unblock quickly when I saw bz late yesterday. 🙂

PR #1946, opened 2020-07-24
[WIP] Bug 1850057: update etcd followers first, use bfq on control plane

cgwalters:
This PR rolls up a few others, including notably:

- https://github.com/openshift/machine-config-operator/pull/1957
- https://github.com/openshift/machine-config-operator/pull/1962

Then adds: WIP: Update etcd followers first 

The way we're talking to etcd is a bit hacky, I ended up cargo
culting some code.  This would be much cleaner if the etcd operator
did it.

But it's critical that we update the etcd followers first, because
leader elections are disruptive events and we can easily minimize
that.

Closes: https://github.com/openshift/machine-config-operator/issues/1897

---

### Updated release images for testing (v2)

I think the previous results were misleading/broken due to using the same release image as base, new images:

`test upgrade registry.svc.ci.openshift.org/coreos/walters-mco-upgrade-release@sha256:25deaaf3074cf984f352ab92fca5f61c4663c9b506ce05c50db8f723a5b386b7 registry.svc.ci.openshift.org/coreos/walters-mco-upgrade-release-target@sha256:1ad5e2fb2df0e4af4abd3c1e59461a7977924201eb0eada524ce7af34d7ac1c4`

- 🔴 https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1291715700414484480 
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1291842967132180480
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-azure/1291842988791566336
- https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1291842941274296320

Issue #1897, opened 2020-07-02
Bug 1850057: stage OS updates (nicely) while etcd is still running


cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1850057

(This content is canonically stored at https://github.com/cgwalters/workboard/tree/master/openshift/bz1850057-etcd-osupdate )

# OS upgrade I/O competes with etcd

Currently the MCD does:

- drain
- apply updates
- reboot

Now "drain" keeps both daemonsets and static pods running.  Of those two, etcd is a static pod today.  When we're applying OS updates, that can be a lot of I/O and (reportedly) compete with etcd.

We have two options:

1. kill etcd after draining (or even more strongly, stop kubelet)
1. "stage" updates gracefully while everything is running

I like option 2) better because we've put a *whole lot* of work into making the ostree stack support this "stage updates while system is running" and it'd be cool if OpenShift used it :smile:   Another way to say this is - I think we want to minimize the time window in which the etcd cluster is missing a member, so the more work we can do while etcd is still running the better!

## Links

Workboard: 

Pull request *in progress*: https://github.com/openshift/machine-config-operator/pull/1957

### Example failing jobs:

- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-azure-upgrade-4.4-stable-to-4.5-ci/1275367195391561728
- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.4-stable-to-4.5-ci/1275416807540264960

Note that both of those jobs jumped from RHEL 8.1 to 8.2.

### Prometheus queries

```
histogram_quantile(0.99, sum(irate(etcd_disk_wal_fsync_duration_seconds_bucket[5m])) by (instance, le)) 
histogram_quantile(0.99, sum(irate(etcd_disk_backend_commit_duration_seconds_bucket[5m])) by (instance, le))
```

---

cgwalters:
We also had problems in the past with applying config changes to `/etc` (including the static pod manifest) while the static pod was still running; I tried looking at `git log --grep=etcd` but didn't see it.  If we change the flow to ensure we're stopping etcd before we apply changes on the filesystem in `/etc`, that would also have fixed that earlier problem.

[...]

PR #1947, opened 2020-07-24
Controlplane upgrades prep

cgwalters:
Prep work for  https://github.com/openshift/machine-config-operator/pull/1946 - basically stubbing out infrastructure.  See patches for details.

### DIFF
--- a/pkg/controller/common/constants.go
+++ b/pkg/controller/common/constants.go
@@ -12,4 +12,7 @@ const (
 
 	// KernelTypeRealtime denominates the realtime kernel type
 	KernelTypeRealtime = "realtime"
+
+	// MasterLabel defines the label associated with master node. The master taint uses the same label as taint's key
+	MasterLabel = "node-role.kubernetes.io/master"
 )
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"github.com/openshift/machine-config-operator/internal"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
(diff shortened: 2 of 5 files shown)


ashcrow:
e2e-gcp-op failure is an infra flake (hitting limits)

cgwalters:
The baseline cluster tests aren't really covering any of the new code here; the upgrade and `e2e-gcp-op` tests are most useful.

Also man...we really need to solve the problem of having controller pod logs go away when we upgrade and they get rescheduled.

cgwalters:
OK cool, seeing the expected logs in e2e-gcp-op; this one should be good to go!
[e2e-gcp-upgrade](https://prow.ci.openshift.org/?job=*e2e-gcp-upgrade*) is quite red across the board which...a lot of that will hopefully what will be fixed by https://github.com/openshift/machine-config-operator/pull/1946 !  We'll see.

Dunno about metal-ipi.

runcom:
/approve

@openshift/openshift-team-mco-maintainers ptal

kikisdeliveryservice:
Metal-ipi job is dead. Let's try these 2 again.

/test e2e-gcp-op
/test e2e-aws-scaleup-rhel7

kikisdeliveryservice:
Gah I meant

/test e2e-gcp-upgrade

cgwalters:
Ooh hey, I had a logic error there causing us to try to update all controlplane nodes at once...hooray for CI tests.

cgwalters:
OK e2e-gcp-upgrade is unbroken.

Hmm interesting, some of Prometheus test failures in e2e-gcp-op seem to be unique.

cgwalters:
This one should be good to go, can I get a lgtm?

cgwalters:
Can't affect rhel7 nodes really
/override e2e-aws-scaleup-rhel7

cgwalters:
This is prep for fixing the upgrade problems, and I'm confident now we're not breaking anything new here.

ericavonb:
Catching up on all the issues behind this PR. I don't like adding new etcd-specific logic to the MCO. It feels like this moves us backwards from separating out etcd concerns into its own operator. Can we do this in a more component-agnostic way?

Something like, a node label or annotation that specifies upgrade priority classes maybe?

ericavonb:
Similar requirement came up here: https://github.com/openshift/machine-config-operator/issues/662#issuecomment-510057616

cgwalters:
> Catching up on all the issues behind this PR. I don't like adding new etcd-specific logic to the MCO. It feels like this moves us backwards from separating out etcd concerns into its own operator. Can we do this in a more component-agnostic way?

Let's debate in https://github.com/openshift/machine-config-operator/issues/1897 ?

PR #1951, opened 2020-07-27
vsphere: check for nil platform in getIgnitionHost

abhinavdahiya:
For UPI workflow installer not set the vSphere platform [1] when no API / Ingress VIP are provided
by the user as they intend to setup their own load balancing etc.

Therefore we should make sure we check for nil vSphere platform.

[1]: https://github.com/openshift/installer/blob/c72e5026b06e8d810fb8f911c6ece8ada7d4c957/pkg/asset/manifests/infrastructure.go#L150-L157


cgwalters:
Can't affect non-vsphere
/override ci/prow/e2e-aws
/hop
/skip
/jump

cgwalters:
Also can't affect GCP or our e2e tests
/override ci/prow/e2e-gcp-op

PR #1953, opened 2020-07-28
Prepare for rebooting to be an optional part of applying updates

beekhof:
**- What I did**

Cleaned up the PoC code for openshift/enhancements#159, replaces PR #1626 and most of PR #1922
Prepares the way for config changes that do not require reboots to be applied.

**- How to verify it**

Exsiting e2e tests and additional unit tests

Refactoring to support future alternatives to node reboots in the specific cases that we know it is safe to do so.


beekhof:
I can also go one further and strip out the inert systemd struct and code

PR #1956, opened 2020-07-29
Bug 1879409: docs: Update for Ignition spec 2/3 dual support

travier:
**- What I did**

Additional documentation updates for Ignition spec 2/3 dual support.

### DIFF
--- a/docs/MachineConfigDaemon.md
+++ b/docs/MachineConfigDaemon.md
@@ -24,11 +24,22 @@ If the updated Ignition config contains changes compatible with the current conf
 
 Not all Ignition config sections are supported; see the following table:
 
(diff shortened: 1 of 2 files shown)


travier:
This is a docs only change so I don't think the CI is relevant here.

travier:
Rebased. This will also require backporting to 4.6 & 4.7.

cgwalters:
/approve
/lgtm

> Rebased. This will also require backporting to 4.6 & 4.7.

Currently while there are `release-4.6` git branches they're just automatically fast forwarded to match `master`.

This technically requires a BZ to merge unfortunately.

cgwalters (approved):

PR #1957, opened 2020-07-29
Bug 1850057: Use bfq scheduler on control plane, idle I/O for rpm-ostreed

cgwalters:
Part of solving #1897
A lot more details in https://hackmd.io/WeqiDWMAQP2sNtuPRul9QA

The TL;DR is that the `bfq` I/O scheduler better respects IO priorities,
and also does a better job of handling latency sensitive processes
like `etcd` versus bulk/background I/O .

### DIFF
--- a/pkg/daemon/controlplane.go
+++ b/pkg/daemon/controlplane.go
@@ -1,0 +1,79 @@
+package daemon
+
+// This file provides changes that we make to the control plane
+// only.
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+)
+
+// setRootDeviceSchedulerBFQ switches to the `bfq` I/O scheduler
+// for the root block device to better share I/O between etcd
+// and other processes.  See
+// https://github.com/openshift/machine-config-operator/issues/1897
+// Note this is the current systemd default in Fedora, but not RHEL8,
+// except for NVMe devices.
+func setRootDeviceSchedulerBFQ() error {
+	sched := "bfq"
+
+	rootDevSysfs, err := getRootBlockDeviceSysfs()
+	if err != nil {
+		return err
+	}
+
+	schedulerPath := filepath.Join(rootDevSysfs, "/queue/scheduler")
+	schedulerContentsBuf, err := ioutil.ReadFile(schedulerPath)
+	if err != nil {
+		return err
+	}
+	schedulerContents := string(schedulerContentsBuf)
+	if strings.Contains(schedulerContents, fmt.Sprintf("[%s]", sched)) {
+		glog.Infof("Device %s already uses scheduler %s", rootDevSysfs, sched)
+		return nil
+	}
+
+	f, err := os.OpenFile(schedulerPath, os.O_WRONLY|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, err = f.Write([]byte(sched))
+	if err != nil {
+		return err
+	}
+	glog.Infof("Set root blockdev %s to use scheduler %v", rootDevSysfs, sched)
+
+	return nil
+}
+
+// updateOstreeObjectSync enables "per-object-fsync" which helps avoid
+// latency spikes for etcd; see https://github.com/ostreedev/ostree/pull/2152
+func updateOstreeObjectSync() error {
+	if err := exec.Command("ostree", "--repo=/sysroot/ostree/repo", "config", "set", "core.per-object-fsync", "true").Run(); err != nil {
+		return errors.Wrapf(err, "Failed to set per-object-fsync for ostree")
+	}
+	return nil
+}
+
+// initializeControlPlane performs setup for the node that should
+// only occur on the control plane.  Currently this switches the IO
+// scheduler and starts a goroutine acting as a small controller
+// for reflecting the etcd leader status in the node object to help
+// the MCC coordinate control plane updates.
+func (dn *Daemon) initializeControlPlane() error {
+	if err := setRootDeviceSchedulerBFQ(); err != nil {
+		return err
+	}
+	if err := updateOstreeObjectSync(); err != nil {
+		return err
+	}
+	return nil
+}
--- a/pkg/daemon/coreos.go
+++ b/pkg/daemon/coreos.go
@@ -1,0 +1,52 @@
+package daemon
+
+// This file provides routines that apply on Fedora CoreOS style systems,
+// including deriviatives like RHEL CoreOS.
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// byLabel returns the udev generated symlink to the block device with the given label
+func byLabel(label string) string {
+	return fmt.Sprintf("/dev/disk/by-label/%s", label)
+}
+
+// getParentDeviceSysfs returns e.g. /sys/devices/pci0000:00/0000:00:05.0/virtio2/block/vda from /dev/vda4, though
+// it can be more complex than that with e.g. NVMe.
+func getParentDeviceSysfs(device string) (string, error) {
+	target, err := os.Readlink(device)
+	if err != nil {
+		return "", errors.Wrapf(err, "reading %s", device)
+	}
+	sysfsDevLink := fmt.Sprintf("/sys/class/block/%s", filepath.Base(target))
+	sysfsDev, err := filepath.EvalSymlinks(sysfsDevLink)
+	if err != nil {
+		return "", errors.Wrapf(err, "parsing %s", sysfsDevLink)
+	}
+	if _, err := os.Stat(filepath.Join(sysfsDev, "partition")); err == nil {
+		sysfsDev = filepath.Dir(sysfsDev)
+	}
+	return sysfsDev, nil
+}
+
+// getRootBlockDeviceSysfs returns the path to the block
+// device backing the root partition on a FCOS system
+func getRootBlockDeviceSysfs() (string, error) {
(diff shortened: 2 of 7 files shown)


cgwalters:
xref https://github.com/coreos/rpm-ostree/pull/2164

sdodson:
/cc @fabianofranz @hexfusion 
FYI, I know you'd done extensive looks into I/O scheduler and its effects on etcd performance

cgwalters:
I rolled in https://github.com/openshift/machine-config-operator/pull/1962 to this PR because I really want to be able to see the effect of the change by looking at events.

cgwalters:
Rebased :surfing_man:

cgwalters:
OK looking at some Prometheus queries mentioned in https://github.com/openshift/machine-config-operator/issues/1897 - this does seem to help.   I see lower p99 and p999 spikes in etcd fsync latency, as expected.  Staging an update took maybe 10s before, would now become 30s or more, but that's obviously totally fine.  

[Before](https://fedorapeople.org/~walters/Screenshot_2020-08-06%20Prometheus%20Time%20Series%20Collection%20and%20Processing%20Server.png)
[After](https://fedorapeople.org/~walters/Screenshot_2020-08-06%20Prometheus%20Time%20Series%20Collection%20and%20Processing%20Server(1).png)

I've mostly done investigation with the [larger PR](https://github.com/openshift/machine-config-operator/pull/1946) that also prefers etcd followers, but I haven't done a huge amount of investigation as to how much of a difference that makes.  I lean towards getting this one in at least relatively soon and get more data from all of the PR jobs and periodics.

cgwalters:
Though, most test runs won't produce interesting data around this until we change the default test to [synthesize a nontrivial OS update](https://github.com/coreos/coreos-assembler/pull/1635).

hexfusion:
I did a very basic test of bfq that was inconclusive. One thing with ionice I was under the impression was it was only honored with `CFQ` scheduler per manpage. But maybe I am missing something.

cc @cgwalters 

>NOTES
>       Linux supports I/O scheduling priorities and classes since  2.6.13  with  the  CFQ  I/O
>       scheduler.

ironcladlou:
@cgwalters's testing seems to be making the claim that with `--per-object-fsync` (https://github.com/ostreedev/ostree/pull/2152) the priorities are being respected with bfq — is that an accurate statement? It doesn't make sense to me why that would be true, and I'd like to understand.

cgwalters:
> I did a very basic test of bfq that was inconclusive. One thing with ionice I was under the impression was it was only honored with CFQ scheduler per manpage. But maybe I am missing something.

The man page is out of date, see the [BFQ docs](https://www.kernel.org/doc/html/latest/block/bfq-iosched.html) which talk about using this, and see also the actual code: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/block/bfq-iosched.c#n4988

cgwalters:
Previously I'd been trying to add information in the PRs but I found I really wanted to write scripts and track nontrivial data there.

I ended up creating a "workboard" repo which has a subdirectory for this: https://github.com/cgwalters/workboard/tree/master/openshift/bz1850057-etcd-osupdate

I'm increasingly confident that this is a good fix - comparing some metrics before after proves out that we are seeing much better p999 latency for example.  Please see the bottom half of this file: https://github.com/cgwalters/workboard/blob/master/openshift/bz1850057-etcd-osupdate/bz185007.md
for links to release images that you can use for your own tests, as well as Prow jobs that have run you can inspect.

cgwalters:
Reminder: the jobs run from this PR, and really almost all of the OpenShift jobs in general are not very useful for gathering any data because *currently* they are all basically noops for OS upgrades.  The only exception is the main "upgrade-4.4-to-4.5" type tests but *only* if they happen to e.g. do a RHEL rebase.  Otherwise they are mostly no-ops too from the OS update perspective.

runcom:
@sinnykumari @yuqi-zhang @kikisdeliveryservice @ericavonb PTAL (and the linked internal Slack thread)

cgwalters:
OK rebased this, dropping out https://github.com/openshift/machine-config-operator/pull/1962 since it's not strictly necessary even though I'd really like to get that one in too.

I think we have good enough results to ship this!

cgwalters:
We seem to have used a BZ clone originally accidentally, switching to use the main one.

cgwalters:
https://github.com/cgwalters/workboard/tree/master/openshift/bz1850057-etcd-osupdate is now updated with some screenshots as well as links to more Prow jobs you can use for your own inspection (and to launch clusters).  For example:

- [Before](https://github.com/cgwalters/workboard/blob/master/openshift/bz1850057-etcd-osupdate/before-grafana.png)
- [After](https://github.com/cgwalters/workboard/blob/master/openshift/bz1850057-etcd-osupdate/after-grafana.png)

cgwalters:
What's interesting is the before/after between `OSUpdateStarted` and `OSUpdateStaged` is still about 30s in both cases; we're not making things slower.  So the problem seems to more be the *pattern* of how we were invoking `fsync` before causing a latency spike.

(That said I do think the effect here could be much more pronounced in clusters that are running a real workload too)

sdodson:
> (That said I do think the effect here could be much more pronounced in clusters that are running a real workload too)

If we feel that's critical we could work with the perf-scale team to replicate their 250 node 4000 namespace + $bignum pod workload upgrade tests where we could readily reproduce this problem. level of effort for that is measured in days though.

pkg/daemon/controlplane.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	glog.Infof("Set root blockdev %s to use scheduler %v", rootDevSysfs, sched)
+
+	return nil
+}
+
+// updateOstreeObjectSync enables "per-object-fsync" which helps avoid
+// latency spikes for etcd; see https://github.com/ostreedev/ostree/pull/2152
+func updateOstreeObjectSync() error {
+	if err := exec.Command("ostree", "--repo=/sysroot/ostree/repo", "config", "set", "core.per-object-fsync", "true").Run(); err != nil {
```
hexfusion, line 60: question: do you think` ionice` or the above has the most dramatic effect on reducing latency?

cgwalters: If you think it's important enough I can try to make new release images with just this change and not the `bfq` default.  But to answer your question, both do matter, see:

https://github.com/cgwalters/workboard/blob/master/openshift/bz1850057-etcd-osupdate/ionice.md#concurrent-updates-with-none-scheduler

cgwalters:
> If we feel that's critical we could work with the perf-scale team to replicate their 250 node 4000 namespace + $bignum pod workload upgrade tests where we could readily reproduce this problem. level of effort for that is measured in days though.

I don't think we *actually* need 250 real nodes, just a simulated workload hitting the apiserver/etcd right?

pkg/daemon/controlplane.go
hexfusion: I don't think it's important enough to change anything I am just curious. You have done enough work :)

hexfusion:
this is amazing wor,  thank you very much for the high level of effort and professionalism in seeing this through for 4.6.

LGTM!

cgwalters:
All green here on the core jobs, just needs a lgtm.

pkg/daemon/controlplane.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+)
+
+// setRootDeviceSchedulerBFQ switches to the `bfq` I/O scheduler
+// for the root block device to better share I/O between etcd
+// and other processes.  See
+// https://github.com/openshift/machine-config-operator/issues/1897
+// Note this is the current systemd default in Fedora, but not RHEL8,
+// except for NVMe devices.
```
smarterclayton, line 23: So why wouldn't we do this for all root devices on all nodes in the cluster?

cgwalters: Mostly to limit the "blast radius" of this PR - changing the control plane only minimizes risk.  But, it does also mean more divergence.

It probably does make sense to do an across-the-board switch though...maybe as a followup to this?  Or would you rather do it in one go?

cgwalters: The other aspect is that currently the OS update vs etcd happens because we're updating the OS while etcd is still running.  But for regular worker nodes that don't have static pods, everything will have been drained before we start the update - so there's less need for bfq inherently.

(Long term I think we really want cgroups v2 + a filesystem that honors IO priorities - it's the only way to get real balance, bfq is mostly heuristics + support for baseline io priorities)

sinnykumari (approved):
This is really amazing work, thanks Colin for doing such a deep analysis.
It looks good to me and considering all other positive feedback let's get this in!

/lgtm

pkg/daemon/controlplane.go
zvonkok: @cgwalters @smarterclayton I would suggest to sync up with the perf-dept, there were some concerns regarding BFQ as the default scheduler in RHEL. There are several workloads where BFQ did not perform the best. MQ-deadline did better. There are probably more workloads that need to be strobed. 

Even though BFQ is great on the low end, past experience shows it doesn't do as well on the high end.  It would need to be thoroughly tested if RHCOS considered changing it.  See http://post-office.corp.redhat.com/archives/kscale-list/2019-September/msg00010.html

zvonkok: The BFQ may work great for the control-plane where we're only running "one" workload (etcd), on the worker nodes we have a variety of workloads that may or not benefit from BFQ. Would also be interesting to see how the infra pods are reacting to such a change.

sinnykumari: Thank you for raising the concern. This PR uses BFQ only for control plane nodes. We will keep in mind and make sure to talk to perf team if we plan to use BFQ for worker/custom pools.

fabianofranz:
+1 this is a big win, thanks @cgwalters!

PR #1958, opened 2020-07-29
[OKD] api: Serve spec v3.1 by default for non-Ignition user agents

LorbusChris:
This is for people using e.g. curl to retrieve Ignition config from the MCS (a few people have had this issue, e.g. here: https://github.com/openshift/okd/issues/239#issuecomment-665737901)

/cc @vrutkovs


cgwalters:
I'm OK with this, but since we have the patient open wouldn't it make sense to support [the `Accept` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) too?

LorbusChris:
> I'm OK with this, but since we have the patient open wouldn't it make sense to support [the `Accept` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) too?

indeed, I'm actually working on this rn :)

PR #1223, opened 2019-10-29
install/telemetry: add prometheus alerts

kikisdeliveryservice:
added alerts for following metrics:
MCDRebootErr(critical), MCDDrainErr(critical),
MCDPivotErr(warning), KubeletHealthState(warning)

PR #1959, opened 2020-07-29
Bug 1861876: telemetry: change MCDDrainErr from critical to warning

kikisdeliveryservice:
This was initially made critical, but that seems to be overkill so move to warning (probably should have been from the start). Prom rules are going to be refactored, so will bake in more nuance at that time.

### DIFF
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -22,7 +22,7 @@ spec:
           expr: |
             mcd_drain > 0
           labels:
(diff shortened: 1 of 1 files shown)


wking:
Fixup for #1223, which was new in 4.3.  Does this need an attached bug for backporting, or did we want to wait and backport once the additional rerolling is completed?

yuqi-zhang:
/lgtm

I think we can wait for the refactor to backport if needed

kikisdeliveryservice:
Added this as a BZ to backport to give a better & less confusing UX until telemetry refactor.

/cherry-pick release-4.5

PR #1960, opened 2020-07-29
Bug 1866743: api: Detect spec version to serve from Accept header

LorbusChris:
**- What I did**
Makes the API inspect the Accept header instead of the User-Agent to
derive the Ignition config spec version it has to serve.

**- How to verify it**
e2e CI

api: Derive spec version to serve from Accept header instead of User-Agent

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/clarketm/json"
@@ -113,13 +115,14 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	useragent := r.Header.Get("User-Agent")
-	glog.Infof("Pool %s requested by address:%q User-Agent:%q", cr.machineConfigPool, r.RemoteAddr, useragent)
+	acceptHeader := r.Header.Get("Accept")
+	glog.Infof("Pool %s requested by address:%q User-Agent:%q Accept-Header: %q", cr.machineConfigPool, r.RemoteAddr, useragent, acceptHeader)
 
-	reqConfigVer, err := deriveSpecVersionFromUseragent(useragent)
+	reqConfigVer, err := detectSpecVersionFromAcceptHeader(acceptHeader)
 	if err != nil {
 		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusBadRequest)
-		glog.Errorf("invalid useragent: %v", err)
+		glog.Error(err)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/server/api.go
```diff
@@ -186,31 +187,33 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// For Ignition v0.x, the accept header looks like:
+	// "application/vnd.coreos.ignition+json; version=2.4.0, application/vnd.coreos.ignition+json; version=1; q=0.5, */*; q=0.1".
+	// For v2.x, it looks like:
+	// "application/vnd.coreos.ignition+json;version=3.1.0, */*;q=0.1".
+	if !strings.HasPrefix(acceptHeader, "application/vnd.coreos.ignition+json") {
 		return semver.New("2.2.0"), nil
 	}
-	ignVersionString := strings.SplitAfter(useragent, "/")[1]
+	ignVersionString := strings.SplitAfter(acceptHeader, "=")[1]
+	ignVersionString = ignVersionString[0:5]
```
cgwalters, line 204: Might be more robust to do something more like taking the whole string until we hit a `,`?  But eh.  Fine for now.

pkg/server/api.go
```diff
@@ -186,31 +187,33 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// For Ignition v0.x, the accept header looks like:
+	// "application/vnd.coreos.ignition+json; version=2.4.0, application/vnd.coreos.ignition+json; version=1; q=0.5, */*; q=0.1".
+	// For v2.x, it looks like:
+	// "application/vnd.coreos.ignition+json;version=3.1.0, */*;q=0.1".
+	if !strings.HasPrefix(acceptHeader, "application/vnd.coreos.ignition+json") {
 		return semver.New("2.2.0"), nil
 	}
-	ignVersionString := strings.SplitAfter(useragent, "/")[1]
+	ignVersionString := strings.SplitAfter(acceptHeader, "=")[1]
+	ignVersionString = strings.Split(ignVersionString, ",")[0]
```
LorbusChris, line 204: @cgwalters PTAL

pkg/server/api.go
```diff
@@ -186,31 +187,33 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-// Only Ignition versions that come with support for either spec v2.2 or spec v3.1 are supported
-func deriveSpecVersionFromUseragent(useragent string) (*semver.Version, error) {
-	// for now, serve v2 if we receive a request without a useragent header with Ignition
-	// this happens if the user pings the endpoint directly (e.g. with curl)
-	// and we don't want to break existing behaviour
-	if !strings.HasPrefix(useragent, "Ignition/") {
+// deriveSpecVersionFromAcceptHeaderUseragent returns a supported Ignition config spec version for a given Accept header.
+// For non-Ignition Accept headers it defaults to config spec v2.2.0
+func deriveSpecVersionFromAcceptHeader(acceptHeader string) (*semver.Version, error) {
+	// serve v3 if we receive a request without an Ignition accept header.
```
LorbusChris, line 193: whoops, shouldn't have changed - we're still serving spec 2 by default.
@runcom @yuqi-zhang now that everything in 4.6 has made the switch to v3, shouldn't we serve it here by default now, too? (might be a bit weird ergonomically if you curl for the config, expecting you get what your new nodes will get, but instead you'll receive the down-translated conf.)

yuqi-zhang: Hmm the "serve by default" is there to maintain backwards compatibility. If someone is not using ignition directly, we're giving them time to make the transition from spec 2 to spec 3 by not breaking existing behaviour. In that sense I think its better to serve spec2 here so as not to break existing CI, etc.

LorbusChris: But don't we want to encourage folks to migrate their custom Ignition to v3 too? If we keep serving v2 we may not encourage that. (If you'd really like it to stay at v2 however, then I'm OK with it :)

LorbusChris: This'll be discussed in a Cabal. Let's leave current behaviour (v2) for now

LorbusChris:
fixed the comment, needs another lgtm
/hold cancel

bgilbert:
This will work for the Accept header currently generated by Ignition, but isn't really a valid Accept parser.  I don't know how much that matters to us.  If third-party clients (curl scripts, etc.) will also be generating headers, they may find this parser brittle.

Proposal:

1. split on commas,
1. stable descending sort by `q` factor,
1. pick the first acceptable `application/vnd.coreos.ignition+json` version from the list,
1. if we only see unacceptable `application/vnd.coreos.ignition+json` versions, fail,
1. if we don't see `application/vnd.coreos.ignition+json` at all, return default version.

Steps 4 and 5 violate the stated intent of the header (there's a wildcard match at the end, implying that Ignition wants to receive whatever it can get, and takes responsibility for failing if it doesn't like it), but avoid having to parse wildcards.

LorbusChris:
@bgilbert I've changed this according to your proposal. PTAL

kikisdeliveryservice:
just adding hold as we are post FF

/hold

LorbusChris:
It was approved before FF, does that mean it's ok to go in? Or would it have needed the lgtm as well before?

kikisdeliveryservice:
> It was approved before FF, does that mean it's ok to go in? Or would it have needed the lgtm as well before?

A PR needs to be final and fully approved (approved and lgtm) by the FF deadline. This only had an approved tag and also was updated/pushed to today (the 3rd day after the deadline).

LorbusChris:
This fixes wrong but working behavior. I'll create a BZ.

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+type acceptHeaderValue struct {
+	MIMEType    string
+	MIMESubtype string
+	Version     *semver.Version
+	QValue      float32
+}
+
+// Parse an accept header, ignoring any extensions that aren't
+// either version or relative quality factor q.
+func parseAcceptHeader(input string) ([]acceptHeaderValue, error) {
```
cgwalters, line 201: There's a lot going on in this parser - a few unit tests would help me feel confident in it.

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			if strings.Contains(ext, "=") {
+				keyval := strings.SplitN(ext, "=", 2)
+				if keyval[0] == "version" && !versionIsSet {
+					var err error
+					v, err = semver.NewVersion(keyval[1])
+					if err != nil {
+						// This is not a valid version
+						continue
+					}
+					versionIsSet = true
```
cgwalters, line 235: We know it's set if `v != nil`.

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				} else if keyval[0] == "q" && !qIsSet {
+					q64, err := strconv.ParseFloat(keyval[1], 32)
+					if err != nil {
+						// This is not a valid relative quality factor
+						continue
+					}
+					q = float32(q64)
+					qIsSet = true
```
cgwalters, line 243: And similarly we can simulate Rust's `Option<T>` type here by doing `var q *float32` or so.

pkg/server/api.go
```diff
@@ -113,13 +115,14 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	acceptHeader := r.Header.Get("Accept")
+	glog.Infof("Pool %s requested by address:%q User-Agent:%q Accept-Header: %q", cr.machineConfigPool, r.RemoteAddr, useragent, acceptHeader)
 
-	reqConfigVer, err := deriveSpecVersionFromUseragent(useragent)
+	reqConfigVer, err := detectSpecVersionFromAcceptHeader(acceptHeader)
 	if err != nil {
 		w.Header().Set("Content-Length", "0")
 		w.WriteHeader(http.StatusBadRequest)
-		glog.Errorf("invalid useragent: %v", err)
+		glog.Error(err)
```
kikisdeliveryservice, line 125: just to confirm this error will be informative enough on its own?

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// Parse an accept header, ignoring any extensions that aren't
+// either version or relative quality factor q.
+func parseAcceptHeader(input string) ([]acceptHeaderValue, error) {
+	var header []acceptHeaderValue
+
+	values := strings.Split(input, ",")
+	for i, value := range values {
+		// remove spaces
+		values[i] = strings.TrimSpace(values[i])
+		parts := strings.Split(value, ";")
```
cgwalters, line 208: It's confusing how we're mutating `values[i]` and then refer to `value` when they're the same thing.  Maybe just:
```
parts := strings.Split(strings.TrimSpace(value), ";")
```
Do we need to mutate `values`?

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		}
+		// mtype[0] is the main MIME type, mtype[1] is the sub MIME type
+		mtype := strings.SplitN(parts[0], "/", 2)
+
+		// check value extensions for version and q parameters, ignore other extensions
+		var v *semver.Version
+		var q float32
+		versionIsSet := false
+		qIsSet := false
+		for _, ext := range parts[1:] {
```
cgwalters, line 225: This is parsing the `q=` value right?  Can't there only be one of those?  IOW do we need a loop?

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		return semver.New("2.2.0"), nil
-	} else if !ignSemver.LessThan(*semver.New("2.3.0")) && ignSemver.LessThan(*semver.New("3.0.0")) {
-		// versions [2.3.0;3) support config spec v3.1
-		return semver.New("3.1.0"), nil
+	if len(header) == 0 {
+		return nil, errors.New("no valid accept header detected")
+	}
+
+	// sort headers by descending q factor value
+	sort.SliceStable(header, func(i, j int) bool { return header[i].QValue > header[j].QValue })
```
kikisdeliveryservice, line 270: question for my understanding : why sort?

pkg/server/api_test.go
```diff
@@ -32,21 +32,21 @@ type scenario struct {
 
 const expectedContentLength int = 32
 
-func setUserAgentOnReq(req *http.Request) *http.Request {
-	req.Header.Set("User-Agent", "Ignition/0.35.0")
+func setAcceptHeaderOnReq(req *http.Request) *http.Request {
```
kikisdeliveryservice, line 35: not knowing much here do these unit tests adequately test the helper funcs you modified above?

pkg/server/api.go
```diff
@@ -186,31 +189,123 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+					}
+					q = float32(q64)
+					qIsSet = true
+				}
+			}
+		}
+
+		// Default version to nil pointer
+		if !versionIsSet {
+			v = nil
```
bgilbert, line 250: Should already be nil in this case.

pkg/server/api.go
bgilbert: We want to select the acceptable Ignition version with the highest client `q` value, and the client isn't required to present them in order.

kikisdeliveryservice: thanks @bgilbert ! :smile_cat:

pkg/server/api.go
LorbusChris: I wasn't aware this was agreed with the Windows team. Sticking v2.2 default for now :)

pkg/server/api.go
LorbusChris: We do not. Fixed!

pkg/server/api_test.go
LorbusChris: the header is set on all (most) requests in the tests below, confirming it works. I've added a separate tests for the two header funcs for even more confidence.

pkg/server/api.go
LorbusChris: yes, it'll only enter here if we get an unsupported ignition version in the accept header. For other headers the default is served.

pkg/server/api.go
LorbusChris: Added a test for the two funcs consuming the accept header string!

LorbusChris:
@bgilbert @cgwalters @kikisdeliveryservice I think I've addressed all comments, PTAL!

kikisdeliveryservice:
> @bgilbert @cgwalters @kikisdeliveryservice thank you for the reviews! I think I've addressed all comments, PTAL :)

looks good from my perspective @LorbusChris i'll let the others verify

just please add the BZ if you'd like the hold on this removed to merge. :smile_cat:

cgwalters:
I don't see Prow enforcing the BZs right now, so IMO let's not go through the overhead of filing a new bug - this is just a bugfix to Ignition spec 3 migration right?

kikisdeliveryservice:
> I don't see Prow enforcing the BZs right now, so IMO let's not go through the overhead of filing a new bug - this is just a bugfix to Ignition spec 3 migration right?

This is not about prow, Christian already said he'd be attaching a BZ. We are asking everyone to follow the rules on this.

yuqi-zhang:
LGTM as well! As a follow up PR maybe it'd be nice to have a snippet in the docs to show what an example request would look like if you were to use curl yourself.

The minimum accepted header seems to be `Accept: application/vnd.coreos.ignition+json; version=2.4.0` / `Accept: application/vnd.coreos.ignition+json; version=3.1.0`. Is that what we'd like to document this as?

PR #1962, opened 2020-07-30
Bug 1852047: controller: Emit events

cgwalters:
A while ago I'd invested some time in tweaking the
node controller to have useful logs around what it's
doing; my first "point of contact" when looking at
upgrades was its pod logs. But...we lose most
those on upgrade since the pod gets killed.

Add events to the node controller too.
Currently the MCD emits useful events which
can be queried afterwards (in our CI runs we
dump `events.json`).

With this we can create a "journal/history"
for upgrade/update events just by querying the
event stream.

### DIFF
--- a/manifests/machineconfigcontroller/events-clusterrole.yaml
+++ b/manifests/machineconfigcontroller/events-clusterrole.yaml
@@ -1,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: machine-config-controller-events
+  namespace: {{.TargetNamespace}}
+rules:
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
--- a/manifests/machineconfigcontroller/events-rolebinding-default.yaml
+++ b/manifests/machineconfigcontroller/events-rolebinding-default.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: machine-config-controller-events
+  namespace: default
+roleRef:
+  kind: ClusterRole
+  name: machine-config-controller-events
+subjects:
+- kind: ServiceAccount
+  namespace: {{.TargetNamespace}}
+  name: machine-config-controller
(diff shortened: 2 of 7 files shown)


cgwalters:
Demo `jq` invocation [over here](https://github.com/openshift/machine-config-operator/pull/1946#issuecomment-666329728) - I plan to stick that some place better but need to decide where that is (this repo?  some shared openshift repo?)

cgwalters:
This one is passing tests and should help us a lot trace through upgrades in the future, can I get a lgtm?

cgwalters:
This one also relates to https://bugzilla.redhat.com/show_bug.cgi?id=1852047

manifests/machineconfigcontroller/events-clusterrole.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: machine-config-controller-events
+  namespace: {{.TargetNamespace}}
+rules:
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
```
runcom, line 9: I don't know enough but is patch needed 🤔 is firing an event involved into patch?

cgwalters: Probably not, but I just copied this from the MCD ones.

cgwalters: Updated to remove `patch` from both (though I didn't test locally).

cgwalters:
I think that `patch` permission may have been necessary, it looks like the run without it just ended up printing them in the logs which isn't useful.

cgwalters:
It'd be really useful to me to have these changes in master, so I can start gathering more "baseline" data across all the clusters being launched for 4.6.

cgwalters:
Yep, same answer as https://github.com/openshift/machine-config-operator/pull/1977#issuecomment-669349800

cgwalters:
We're nearing 3 weeks to get this pretty simple patch in...

cgwalters:
I want to emphasize the value of getting this patch in soon - https://github.com/openshift/machine-config-operator/pull/1962#issuecomment-669169073

sinnykumari:
lgtm.
@runcom any final comment before we get it merged?

PR #1965, opened 2020-07-31
Add missing IPv6 support for OVN/OVS config

trozet:
Previously we were only looking at IPv4 for default routes. This also
looks for IPv6 if IPv4 is not found. This patch also includes a fix for
jq returning null when it fails to read, which we were not previously
checking for.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -14,9 +14,16 @@ contents:
     counter=0
     # find default interface
     while [ $counter -lt 12 ]; do
+      # check ipv4
       iface=$(ip -j route show default | jq -r '.[0].dev')
-      if [ -n "$iface" ]; then
-        echo "Default gateway interface found: ${iface}"
+      if [[ -n "$iface"  && "$iface" != "null" ]]; then
+        echo "IPv4 Default gateway interface found: ${iface}"
+        break
+      fi
+      # check ipv6
+      iface=$(ip -6 -j route show default | jq -r '.[0].dev')
+      if [[ -n "$iface"  && "$iface" != "null" ]]; then
+        echo "IPv6 Default gateway interface found: ${iface}"
         break
       fi
       counter=$((counter+1))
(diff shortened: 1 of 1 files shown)


templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -14,9 +14,16 @@ contents:
     counter=0
     # find default interface
     while [ $counter -lt 12 ]; do
+      # check ipv4
       iface=$(ip -j route show default | jq -r '.[0].dev')
-      if [ -n "$iface" ]; then
-        echo "Default gateway interface found: ${iface}"
+      if [[ -n "$iface"  && "$iface" != "null" ]]; then
```
danwinship, line 19: fwiw you can do `jq -r '.[0].dev // empty'` to avoid the "`null`" output
(you can also use `jq -e` to exit with non-0 status on "`null`", but I don't think that would help here)

kikisdeliveryservice:
@trozet do we expect this e2e-ovn-step-registry test to pass?

trozet:
> @trozet do we expect this e2e-ovn-step-registry test to pass?

I expect it to only fail 1 case for that job:
https://github.com/openshift/origin/pull/25356

However, I can see this patch worked in the must-gather journal for the node:
```
Jul 31 15:47:10.998720 ip-10-0-152-53 configure-ovs.sh[1638]: OVS successfully configured
Jul 31 15:47:10.998960 ip-10-0-152-53 configure-ovs.sh[1638]: + ip a show br-ex
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]: 4: br-ex: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc noqueue state UNKNOWN group default qlen 1000
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]:     link/ether 0a:d9:b7:51:75:71 brd ff:ff:ff:ff:ff:ff
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]:     inet 10.0.152.53/18 brd 10.0.191.255 scope global dynamic noprefixroute br-ex
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]:        valid_lft 3595sec preferred_lft 3595sec
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]:     inet6 fe80::319b:c1cf:8fc1:a9bc/64 scope link noprefixroute
Jul 31 15:47:11.002879 ip-10-0-152-53 configure-ovs.sh[1638]:        valid_lft forever preferred_lft forever
Jul 31 15:47:11.012168 ip-10-0-152-53 configure-ovs.sh[1638]: + exit 0
Jul 31 15:47:11.007248 ip-10-0-152-53 systemd[1]: Started Configures OVS with proper host networking configuration
```

PR #1967, opened 2020-07-31
Capturing steps to run customized kubelet on Openshift worker nodes

swatisehgal:
This is to enable developers looking to test/run customized kubelet on an openshift cluster

Signed-off-by: Swati Sehgal <swsehgal@redhat.com>

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -63,7 +63,7 @@ All tests (unit and e2e) can be executed with:
 
 # Managing Go Dependencies
 
-Dependencies are managed with [go modules](https://github.com/golang/go/wiki/Modules) but committed directly to the repository. 
+Dependencies are managed with [go modules](https://github.com/golang/go/wiki/Modules) but committed directly to the repository.
 
 Please ensure that you are using Go v1.13.  Using a different version may result in unexpected behavior.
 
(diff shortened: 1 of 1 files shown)


docs/HACKING.md
```diff
@@ -300,3 +300,30 @@ It will fail if you use e.g. `:latest`.
 
 At the time of this writing, the [kubelet](https://github.com/smarterclayton/origin/blob/4de957b019aee56931b1a29af148cf64865a969b/images/os/Dockerfile)
 has code to do this in CI, and work is in progress to [replicate that for cri-o](https://github.com/openshift/release/pull/4030).
+
+### Running worker nodes with customized kubelet
+
+1) Ssh to the node (use github.com/eparis/ssh-bastion)
```
kikisdeliveryservice, line 306: we don't encourage sshing in to nodes

kikisdeliveryservice:
@swatisehgal what is the use case for this?

kikisdeliveryservice:
This doesn't feel like something that should be in the MCO repo?

cgwalters:
The rest of the doc is about making changes to nodes for development or testing, right?  We need to enable developers, testers, support organizations etc.  to do this kind of thing. 

Seems fine to me overall (though it does duplicate some of the bits above but eh).

docs/HACKING.md
```diff
@@ -300,3 +300,30 @@ It will fail if you use e.g. `:latest`.
...
+``
+$ ssh core@<worker-node-ip>
+``
+2) Use scp to copy it the customised kubelet binary
+``
+$ scp root@<remote-host-with-kubelet-binary>:<path/to/kubelet/binary> /home/core
+``
+3) Make the host writable
+``
+$ mount -o remount,rw /usr
```
cgwalters, line 316: It's better to use `rpm-ostree usroverlay` as is mentioned above.

docs/HACKING.md
```diff
@@ -300,3 +300,30 @@ It will fail if you use e.g. `:latest`.
...
+``
+4) ``$ systemctl stop kubelet``
+5) Copy the binary to /usr/bin/kubelet
+``
+$ cp /home/core/kubelet /usr/bin/kubelet
```
cgwalters, line 321: Because among other things, `cp` does `open(O_TRUNC)` which will corrupt the ostree object store, so rolling back won't work.  So you want the `rpm-ostree usroverlay`.

docs/HACKING.md
swatisehgal: Sure, I can update this to use `rpm-ostree usroverlay`

docs/HACKING.md
swatisehgal: Ack

swatisehgal:
@kikisdeliveryservice  The use case of introducing this is exactly what @cgwalters mentioned, to enable developers to test/run customized kubelet on openshift clusters.

runcom:
> to enable developers to test/run customized kubelet on openshift clusters.

I think this is good to have in our HACKING.md :+1:

docs/HACKING.md
swatisehgal: We need elevated privileges to stop and restart kubelet, that's the reason I have this step

kikisdeliveryservice:
> > to enable developers to test/run customized kubelet on openshift clusters.
> 
> I think this is good to have in our HACKING.md +1 perhaps adding that to your first PR comment would shed some light in case anybody else has the same doubt :)

Sounds good, @swatisehgal if you could squash your commits and update the message as @runcom suggested that would be great.

swatisehgal:
@kikisdeliveryservice Done!

kikisdeliveryservice:
doc updates are allowed at any time in the cycle so removing the hold and giving final approval, thanks for this @swatisehgal 

/hold cancel
/lgtm

PR #1969, opened 2020-07-31
test/e2e: check that kernel-rt-core pacakge is installed

sinnykumari:
It seems the kernel-rt-core package may change uname -a
content. In older package it included substring PREEMPT RT
which we were using to determine that current booted kernel
in rt kernel.

This changed recently to PREEMPT_RT, due to which tests started
failing. Hopefully just checking for PREEMPT is sufficient.

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -187,8 +187,8 @@ func TestKernelType(t *testing.T) {
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], renderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
 
-	kernelInfo := execCmdOnNode(t, cs, infraNode, "uname", "-a")
-	if !strings.Contains(kernelInfo, "PREEMPT RT") {
+	kernelInfo := execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "kernel-rt-core")
+	if !strings.Contains(kernelInfo, "kernel-rt-core") {
 		t.Fatalf("Node %s doesn't have expected kernel", infraNode.Name)
 	}
 	t.Logf("Node %s has expected kernel", infraNode.Name)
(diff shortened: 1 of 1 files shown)


runcom:
let's wait and see how it goes with this

/approve

test/e2e/mcd_test.go
```diff
@@ -187,8 +187,8 @@ func TestKernelType(t *testing.T) {
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], renderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
 
-	kernelInfo := execCmdOnNode(t, cs, infraNode, "uname", "-a")
-	if !strings.Contains(kernelInfo, "PREEMPT RT") {
+	kernelInfo := execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "kernel-rt-core")
```
cgwalters, line 190: The `-a` there is unnecessary (but fine as is, we can address it in a followup).  We should also probably make `execCmdOnNode` do the `chroot /rootfs` internally because basically everything else in this function is targeting the real root too.

sinnykumari: -a is not necessary but `-a` explicitly says `query/verify all packages` and that made me use it :)
yeah, you are right. We can look at execCmdOnNode() to do  `chroot /rootfs`

test/e2e/mcd_test.go
```diff
@@ -210,8 +210,8 @@ func TestKernelType(t *testing.T) {
 
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], oldInfraRenderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
-	kernelInfo = execCmdOnNode(t, cs, infraNode, "uname", "-a")
-	if strings.Contains(kernelInfo, "PREEMPT RT") {
+	kernelInfo = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "kernel-rt-core")
+	if strings.Contains(kernelInfo, "kernel-rt-core") {
```
celebdor, line 214:
```suggestion
	if strings.hasPrefix(kernelInfo, "kernel-rt-core") {
```

sinnykumari: Thanks for the feedback! let's fix this in a followup PR to unblock tests and pending prs.

test/e2e/mcd_test.go
runcom: the funny thing is the other uses of `execCmdOnNode` play from outside the chroot by just string matching on files within `/rootfs` - but I agree we can implement chroot in the function to make it transparent that commands are executed on the node's root directly

cgwalters:
This can't affect anything except our own tests, no reason to waste resources on testing the install with this
/override ci/prow/e2e-aws
/override ci/prow/e2e-aws-scaleup-rhel7

sinnykumari:
can someone add lgtm label

sinnykumari:
skip isn't friendly when test is running

sinnykumari:
> /lgtm

Thanks :tada:

PR #1970, opened 2020-07-31
IgnParseWrapper: return better errors

yuqi-zhang:
Default to spec 3 for the return type, and clarify error messages,
so the user has a better sense of what went wrong in MCC logs.

Mostly useful for MachineConfig manifests provided during install
that have errors in them.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -306,30 +306,32 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
 // IgnParseWrapper parses rawIgn for both V2 and V3 ignition configs and returns
 // a V2 or V3 Config or an error. This wrapper is necessary since V2 and V3 use different parsers.
 func IgnParseWrapper(rawIgn []byte) (interface{}, error) {
-	ignCfg, rpt, err := ign2.Parse(rawIgn)
-	if err == nil && !rpt.IsFatal() {
-		// this is an ignv2cfg that was successfully parsed
-		return ignCfg, nil
+	ignCfgV3_1, rptV3_1, errV3_1 := ign3.Parse(rawIgn)
+	if errV3_1 == nil && !rptV3_1.IsFatal() {
+		return ignCfgV3_1, nil
 	}
-	if err.Error() == ign2error.ErrUnknownVersion.Error() {
-		ignCfgV3, rptV3, errV3 := ign3.Parse(rawIgn)
-		if errV3 == nil && !rptV3.IsFatal() {
-			return ignCfgV3, nil
+	// unlike spec v2 parsers, v3 parsers aren't chained by default so we need to try parsing as spec v3.0 as well
+	if errV3_1.Error() == ign3error.ErrUnknownVersion.Error() {
+		ignCfgV3_0, rptV3_0, errV3_0 := ign3_0.Parse(rawIgn)
+		if errV3_0 == nil && !rptV3_0.IsFatal() {
+			return translate3.Translate(ignCfgV3_0), nil
 		}
-		// unlike spec v2 parsers, v3 parsers aren't chained by default so we need to try parsing as spec v3.0 as well
-		if errV3.Error() == ign3error.ErrUnknownVersion.Error() {
-			ignCfgV3_0, rptV3_0, errV3_0 := ign3_0.Parse(rawIgn)
-			if errV3_0 == nil && !rptV3_0.IsFatal() {
-				return translate3.Translate(ignCfgV3_0), nil
+
+		if errV3_0.Error() == ign3error.ErrUnknownVersion.Error() {
+			ignCfgV2, rptV2, errV2 := ign2.Parse(rawIgn)
+			if errV2 == nil && !rptV2.IsFatal() {
+				return ignCfgV2, nil
 			}
 
-			return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3_0, rptV3_0)
+			// If the error is still UnknownVersion it's not a 3.1/3.0 or 2.x config, thus unsupported
+			if errV2.Error() == ign2error.ErrUnknownVersion.Error() {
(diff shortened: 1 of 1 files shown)


pkg/controller/common/helpers.go
```diff
@@ -306,30 +306,36 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+			if errV2 == nil && !rptV2.IsFatal() {
+				return ignCfgV2, nil
 			}
 
-			return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3_0, rptV3_0)
-		}
+			// If the error is still UnknownVersion it's not a 3.1/3.0 or 2.x config, thus unsupported
+			if errV2.Error() == ign2error.ErrUnknownVersion.Error() {
+				return ign3types.Config{}, errors.Errorf("parsing Ignition config failed: unknown version:"+
+					"v3 error %v\nv3 Report: %v\nv2 error %v\nv2 Report: %v\n", errV3, rptV3, errV2, rptV2)
```
kikisdeliveryservice, line 329: This seems confusing, wouldn't it be useful to tell the user what we _do_ support? In this case do we need to give them the error and the report?

pkg/controller/common/helpers.go
```diff
@@ -306,30 +306,36 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+			}
 
-		return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3, rptV3)
+			// Parsing all 3 types failed, let's return both 3.1 and 2.x errors
+			return ign3types.Config{}, errors.Errorf("parsing Ignition config failed: could not parse config in v2 or v3,"+
+				"v3 error %v\nv3 Report: %v\nv2 error %v\nv2 Report: %v\n", errV3, rptV3, errV2, rptV2)
```
kikisdeliveryservice, line 334: have you run these errors to see wht they look like? bc it sounds like it could be confusing?

pkg/controller/common/helpers.go
```diff
@@ -306,30 +306,36 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+		}
+		return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3, rptV3)
```
kikisdeliveryservice, line 336: are we treating 3.0/3.1 as completely different? if we support 3.2 will we need another return statement? maybe just v3?

pkg/controller/common/helpers.go
```diff
@@ -306,30 +306,36 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
-	ignCfg, rpt, err := ign2.Parse(rawIgn)
-	if err == nil && !rpt.IsFatal() {
-		// this is an ignv2cfg that was successfully parsed
-		return ignCfg, nil
-	}
-	if err.Error() == ign2error.ErrUnknownVersion.Error() {
-		ignCfgV3, rptV3, errV3 := ign3.Parse(rawIgn)
-		if errV3 == nil && !rptV3.IsFatal() {
-			return ignCfgV3, nil
+	ignCfgV3, rptV3, errV3 := ign3.Parse(rawIgn)
```
kikisdeliveryservice, line 309: for clarity ign3_1 since you're distinguishing it from 3_0 below

pkg/controller/common/helpers.go
yuqi-zhang: I will double check this

yuqi-zhang: Yeah so basically if we have 3.2 we will have to chain parse with another if clause, this is a ignition design decision

pkg/controller/common/helpers.go
yuqi-zhang: Clarified this since it can only be spec 2

yuqi-zhang:
Fixed up a bit. Basically following https://github.com/coreos/ignition/blob/ff94f8727d8558b38456cb3eb84ec7d1048d74e4/config/shared/errors/errors.go, non versioning errors will fall under "spec3" failing, which should be our default anyways

pkg/controller/common/helpers.go
```diff
@@ -306,30 +306,34 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
 
-			return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3_0, rptV3_0)
-		}
+			// If the error is still UnknownVersion it's not a 3.1/3.0 or 2.x config, thus unsupported
+			if errV2.Error() == ign2error.ErrUnknownVersion.Error() {
+				return ign3types.Config{}, errors.Errorf("parsing Ignition config failed: unknown version. Supported spec versions: 2.2, 3.0, 3.1")
+			}
 
-		return ign2types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3, rptV3)
+			// Parsing all 3 types failed, let's return both 3.1 and 2.x errors
```
kikisdeliveryservice, line 331: this comment seem outdated now?

yuqi-zhang: whoops, missed it, fixed

kikisdeliveryservice:
```   * could not run steps: step e2e-aws failed: failed to acquire lease: Post "https://boskos-ci.apps.ci.l2s4.p1.openshiftapps.com/acquire?dest=leased&owner=ci-op-kjybwvkd-c4a31&request_id=1976235410884491574&state=free&type=aws-quota-slice": net/http: TLS handshake timeout
time="2020-07-31T21:06:42Z" level=info msg="Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:acquiring_lease:aws-quota-slice\"" 
```
/retest

Issue #1964, opened 2020-07-30
inconsistency between osimageurl and digest comparisons


cgwalters:
I was looking at some recent e2e-gcp-upgrade jobs (e.g. [this one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1941/pull-ci-openshift-machine-config-operator-master-e2e-gcp-upgrade/1288897683385225216)) and I noticed:

```
I0730 19:53:35.758111   94185 update.go:1430] Starting update from rendered-worker-427e9f6f5393fffc56bdeddc50f8a33f to rendered-worker-e95dec12126dfcd3937bbfde60933d8c: &{osUpdate:true kargs:false fips:false passwd:false files:true units:true kernelType:false extensions:false}
```

Notice `osUpdate: true`, because the way the job works is that it makes a "synthetic" upgrade by changing the URL for most things but not the digest.  For regular containers in Kube that will still trigger a redeployment, but a while ago we added special casing for "matching digest".

The [MCD in this PR](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/1941/pull-ci-openshift-machine-config-operator-master-e2e-gcp-upgrade/1288897683385225216/artifacts/e2e-gcp-upgrade/pods/openshift-machine-config-operator_machine-config-daemon-9lscc_machine-config-daemon.log) for example still runs:

```
I0730 19:56:06.879041   94185 run.go:18] Running: oc image extract --path /:/run/mco-machine-os-content/os-content-227367051 --registry-config /var/lib/kubelet/config.json registry.svc.ci.openshift.org/ci-op-hzcly6m9/stable@sha256:3a53c55db7ba5b45732ca410bf3edbbc277b03fc2044aa977b2e341e63f1d996
```

But then we do nothing with it because the digest matches.

What would probably make the most sense is to change that CI job to actually synthesize an OS update.  And we should debate whether we keep this special "matching digest" code. 

I'd have to dig through git history but I think it was trying to optimize the "avoid initial reboot if bootimage == target" case but that never happens today.

cgwalters:
Another option of course could be
```
diff --git a/pkg/daemon/update.go b/pkg/daemon/update.go
index c052e099..924bfc62 100644
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
[...]

PR #1971, opened 2020-07-31
Bug 1852047: daemon: Remove "compare digest" code for OS updates

cgwalters:
We don't need to do this special casing anymore.  This
corner case only arises for the synthetic `e2e-gcp-upgrade`
test, and we actually want to test the OS update path there.

Closes: https://github.com/openshift/machine-config-operator/issues/1964

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -18,7 +18,6 @@ import (
 	"syscall"
 	"time"
 
-	imgref "github.com/containers/image/docker/reference"
 	ign2types "github.com/coreos/ignition/config/v2_2/types"
 	ign3types "github.com/coreos/ignition/v2/config/v3_1/types"
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -1351,7 +1351,10 @@ func getRefDigest(ref string) (string, error) {
 	return canon.Digest().String(), nil
 }
 
-// compareOSImageURL is the backend for checkOS.
+// compareOSImageURL is the backend for checkOS.  This function
+// used to have more sophistication around comparing the digests,
+// but we removed that.  The dummy:// could also likely go away.
+// https://github.com/openshift/machine-config-operator/issues/1964
 func compareOSImageURL(current, desired string) (bool, error) {
```
ericavonb, line 1358: We can remove the error return from the function altogether at this point.
```golang
func compareOSImageURL(current, desired string) bool {
	return desired == "" || desired == "://dummy" || current == desired
```
("No target" log omitted for demonstration only)

cgwalters: Yep you're right, it turns into a bigger patch but as long as we're here we might as well do the bigger cleanup.

ericavonb:
yay for simpilified code!
/lgtm

sinnykumari:
I am not 100% sure but there is lot  of possibility that this PR merge will create  merge conflict to the extensions feature https://github.com/openshift/machine-config-operator/pull/1941/files#diff-06961b075f1753956d802ba954d2cfb5L1319 . It will be good to fix this on top of #1941.

Adding hold to get Pr#1941 merged.
/hold

sinnykumari:
#1941 merged
/hold cancel

cgwalters:
Turned out the relevant bit here was already removed in https://github.com/openshift/machine-config-operator/pull/1941

PR #1972, opened 2020-07-31
Bug 1907671: Handle keepalived hangs in liveness probe

cybertron:
Currently, if keepalived stops responding for any reason but the
process is still there, we will never notice because the previous
version of /tmp/keepalived.data will still be present in /tmp.

This change writes the fault string to keepalived.data before doing
the kill -s SIGUSR1 so that we assume failure unless keepalived
overwrites the file as expected. A sleep is added as well to give
it time to do so. This way if keepalived stops responding to the
kill command we will see the default failure and restart keepalived.

### DIFF
--- a/templates/common/on-prem/files/keepalived.yaml
+++ b/templates/common/on-prem/files/keepalived.yaml
@@ -114,7 +114,7 @@ contents:
             - /bin/bash
             - -c
             - |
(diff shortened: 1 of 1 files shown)


cybertron:
/hold cancel
/cc @celebdor 

I think this version addresses some discussion we had about this back when it was first proposed. It may also help with some odd keepalived issues.

celebdor:
/retitle Bug 1907671: Handle keepalived hangs in liveness probe

PR #1973, opened 2020-07-31
Bug 1864677: operator/bootstrap: use correct key for generated cloud conf

abhinavdahiya:
Previously the cloud provider config map was sourced from the `.spec.cloudConf` of the infrastructure object.
but to enable user input and stitching explained in [1] and [2], we added a controller to generate a cloud configuraration using the spec
for use by kubelet and kube cloud controller manager. The new API defined in [3] creates a config map `openshift-config-managed/kube-cloud-config` with
the configuration in `cloud.conf` key. We updated the in-cluster MCO to use the new API [4] but the MCO on bootstrap host was left untouched because the controller
could not generated on the bootstrap host yet.

But [5] and [6] intend to provide the same generated config map on the bootstrap host for MCO, and therefore we need the bootstrap MCO to read the key defined by
the generated API but fallback to he spec key (old behavior) for backward compatibility.

/cc @sinnykumari 
since this is an extension to [4]

[1]: https://github.com/openshift/enhancements/blob/master/enhancements/installer/aws-custom-region-and-endpoints.md
[2]: https://github.com/openshift/enhancements/blob/master/enhancements/installer/azure-support-known-cloud-environments.md
[3]: https://github.com/openshift/api/blob/e21882127f24772e3b5388fe1fdc37669e8e1d04/config/v1/types_infrastructure.go#L26-L40
[4]: https://github.com/openshift/machine-config-operator/pull/1658
[5]: https://github.com/openshift/cluster-config-operator/pull/140
[6]: https://github.com/openshift/installer/pull/3831

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -110,19 +110,11 @@ func RenderBootstrap(
 
 	// if the cloudConfig is set in infra read the cloudConfigFile
 	if infra.Spec.CloudConfig.Name != "" {
-		data, err := ioutil.ReadFile(cloudConfigFile)
+		cloudConf, err := loadBootstrapCloudProviderConfig(infra, cloudConfigFile)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to load the cloud provider config: %v", err)
 		}
-		obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(corev1.SchemeGroupVersion), data)
-		if err != nil {
-			return err
-		}
-		cm, ok := obji.(*corev1.ConfigMap)
-		if !ok {
-			return fmt.Errorf("expected *corev1.ConfigMap found %T", obji)
-		}
-		spec.CloudProviderConfig = cm.Data[infra.Spec.CloudConfig.Key]
+		spec.CloudProviderConfig = cloudConf
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/operator/bootstrap.go
```diff
@@ -122,7 +122,11 @@ func RenderBootstrap(
 		if !ok {
 			return fmt.Errorf("expected *corev1.ConfigMap found %T", obji)
 		}
-		spec.CloudProviderConfig = cm.Data[infra.Spec.CloudConfig.Key]
+		cloudConf, ok := cm.Data["cloud.conf"]
+		if !ok {
+			cloudConf = cm.Data[infra.Spec.CloudConfig.Key]
```
sinnykumari, line 127: Adding info message when we are doing fallback will be nice for debugging.

sinnykumari:
Overall lgtm, one minor comment. Also, Verify test needs to be fixed.

/retest

kikisdeliveryservice:
adding hold since we are post-FF

if this is a bug @abhinavdahiya please open and attach BZ

/hold

abhinavdahiya:
> adding hold since we are post-FF
> 
> if this is a bug @abhinavdahiya please open and attach BZ
> 
> /hold

added the BZ
/hold cancel

PR #1975, opened 2020-08-03
WIP pkg/daemon: extract oc from CLI image if no oc found on host

vrutkovs:
**- What I did**
`ExtractOSImage` now checks that `oc` is present on the host before attempting to run `oc image extract`. If not it would extract it from `cli` image using `podman run` similar to mcd binary bootstrap.

**- How to verify it**
Run pivot on the host which doesn't have `/usr/bin/oc` (FCOS)

MCD now extracts OC from `cli` container if necessary

TODO:
* [ ] pass real CLI image to daemon

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -46,6 +46,9 @@ const (
 	fipsFile              = "/proc/sys/crypto/fips_enabled"
 	extensionsRepo        = "/etc/yum.repos.d/coreos-extensions.repo"
 	osImageContentBaseDir = "/run/mco-machine-os-content/"
+	cliBinPath            = "/usr/bin/oc"
+	cliRunBinPath         = "/usr/local/bin/oc"
+	podmanPath            = "/usr/bin/podman"
 )
 
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -46,6 +46,9 @@ const (
 	fipsFile              = "/proc/sys/crypto/fips_enabled"
 	extensionsRepo        = "/etc/yum.repos.d/coreos-extensions.repo"
 	osImageContentBaseDir = "/run/mco-machine-os-content/"
+	cliBinPath            = "/usr/bin/oc"
+	cliRunBinPath         = "/run/bin/oc"
```
cgwalters, line 50: Though something that will become important with this is that `/run/bin` isn't in `$PATH`.  We can fix that pretty easily but if we take this approach out farther we'll likely want to do something more sophisticated.

pkg/daemon/update.go
```diff
@@ -217,6 +220,37 @@ func addExtensionsRepo(osImageContentDir string) error {
 	return nil
 }
 
+func getOCBinaryPath() (string, error) {
+	if _, err := os.Stat(cliBinPath); err == nil {
+		return cliBinPath, nil
+	}
+
+	cliImage := "registry.svc.openshift.org/origin/4.6:cli"
```
cgwalters, line 228: We need to pull the from the manifest, the same way as we template all the other images.

pkg/daemon/update.go
```diff
@@ -217,6 +220,37 @@ func addExtensionsRepo(osImageContentDir string) error {
...
+	// Pull cli image
+	args := []string{"pull", "--authfile=/var/lib/kubelet/config.json", "--quiet", cliImage}
+	if _, err := pivotutils.RunExt(numRetriesNetCommands, podmanPath, args...); err != nil {
+		return "", err
+	}
+	// Extract oc from cli Container to tmp path
+	cliTmpPath := fmt.Sprintf(cliRunBinPath)
```
cgwalters, line 235: Did you mean something like:
`cliTmpPath := fmt.Sprintf("%s.tmp", cliRunBinPath)`
?

pkg/daemon/update.go
vrutkovs: Right, I'll fix that once OKD tests pass

LorbusChris:
looks good (except for the hardcoded image ref)
/approve

vrutkovs:
/hold

`oc image extract` is probably going to be reverted, so I'll help including these commits to a rework PR

Issue #283, opened 2019-01-10
support inline data in MCs


cgwalters:
Ignition was designed to be a *target* for higher level config languages (e.g. `ct`).  We are currently exposing "Ignition JSON as YAML" in MCs.  I think there's no real reason for Ignition not to support true inline data (as opposed to `data:` URLs which aren't friendly to edit directly), e.g.:

```
          files:
          - contents:
              inline: |
                some inline file data here
              verification: {}
            filesystem: root
            mode: 420
            path: /etc/somefile
```

rather than requiring the `data:` URL encoding.

ashcrow:
Would this be an enhancement to Ignition to allow the functionality or do you envision MCD adding a translation layer?

ashcrow:
@cgwalters it sounds like a good addition, but I'm not sure how it lands in the Ignition spec. We may want to move this over to the Ignition issue tracker or pull in some maintainers into this discussion.

kikisdeliveryservice:
@ashcrow I checked and this seems to not conform to the Ignition spec found here: https://coreos.com/ignition/docs/0.20.1/configuration-v2_1.html
files' contains:
contents (object): options related to the contents of the file.
- compression (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
- source (string): the URL of the file contents. Supported schemes are http, https, tftp, s3, and data. When using http, it is advisable to use the verification option to ensure the contents haven't been modified.
- verification (object): options related to the verification of the file contents.
- hash (string): the hash of the config, in the form <type>-<value> where type is sha512.

Because of this, I don't think that we should be diverging from the actual Ignition config specs in our use of Ignition, and would rather we open this as an issue/feature request in the Ignition repo.

kikisdeliveryservice:
Ha ignore me because actually this is already possible in Ignition!
See our _own_ documentation: https://github.com/openshift/machine-config-operator/blob/master/HACKING.md
```
files:
      - contents:
          source: data:,hello%20world%0A
```

Use source: data

ajeddeloh:
Ignition supports inline data via data URLs. CT has sugar to do this for you. I don't know the history of why this is the case (perhaps @crawford knows) but I'd guess it was to avoid a major spec bump. I'd be open to removing the data URL and switching to allow inline contents in 3.0.0. For one it'd make the two config structs (ct/ignition) more isomorphic which makes translation easier.

cgwalters:
[...]

PR #1980, opened 2020-08-04
*: Support CoreOS configs in MCs

LorbusChris:
**- What I did**
Add support for parsing CoreOS configs from RawExtension in MC

This is one possible way to go about the issue discussed in #283 

**- How to verify it**
I'll add a test
Support CoreOS configs in MCs

/hold


kikisdeliveryservice:
hi @LorbusChris is there any problem definition/background for this?

LorbusChris:
@kikisdeliveryservice this is one way to solve https://github.com/openshift/machine-config-operator/issues/283
It has come up a few times. Just threw this up here to open up the discussion.

cgwalters:
I like the idea but I think we should do this after https://github.com/coreos/fcct/pull/79

Also MCO team: I added a 4.7 label we can use for PRs like this - "things to look at/merge after we branch for 4.6".

cgwalters (changes requested):
Moving off my review queue for now

cgwalters:
So we could probably land this one pretty quickly if we agree to conditionalize it first on being on FCOS/OKD right?  There's low-to-zero risk for that and then we'd be able to tackle the "how does fcc work with RHCOS" as a separate phase or so.

LorbusChris:
I've created an enhancement proposal for this: https://github.com/openshift/enhancements/pull/525

@cgwalters I'm not sure we need to conditionalize it for OKD - or are there any drawbacks to having it in OCP as a tech preview of sorts that is not used by any component by default, and not recommended to admin users yet?

kikisdeliveryservice:
Agree on gating and also there's an open enhancement so it feels like that should be finished first..

/hold

kikisdeliveryservice:
Also: is this okd-e2e test dead?

LorbusChris:
@kikisdeliveryservice I don't think master and 4.6 okd tests will succeed just, yet as machine-os-content for 4.6+ still has some issues.
/retest

PR #1982, opened 2020-08-05
[release-4.4] Bug 1855895: templates: add a file to load legacy iptables

knobunc:
We need to load the iptables modules so that containers can use the legacy iptables interface.  Since 4.4 uses v2 Ignition, I needed to add fields to conform to the older format when I did the backport.

**- What I did**

Backport of a change to make RHCOS load the iptables compatibility modules.

**- How to verify it**

With this, the egress router pods should work.

We need to load the iptables modules so that containers can use the legacy iptables interface.


knobunc:
This supersedes https://github.com/openshift/machine-config-operator/pull/1955

PR #1984, opened 2020-08-06
Bug 1858277: Check the version hash so we resync on an upgrade

umohnani8:
Fixes #1858277 (https://bugzilla.redhat.com/show_bug.cgi?id=1858277)

**- What I did**
Add a check for the version hash to ensure that we resync
and generate a new controller version during an upgrade.

Update the way we append the status for a ctrcfg as well -
ensure that we don't have mutliple of the same status recorded.

Add a check to avoid adding the same mc name more than once in
the list of finalizers. When we update an existing ctrcfg, the
generation number changes causing a resync and we don't need to add
the same name to the list of finalizers every time this happens.
And remove the check for cri-o drop-in files.

**- How to verify it**
1. Create a 4.5 cluster
2. Create a ctrcfg CR on it and apply it to both master and worker pools.
e.g:
```
kind: ContainerRuntimeConfig
metadata:
  name: set-log
spec:
  machineConfigPoolSelector:
    matchLabels:
      custom-crio: set-log
  containerRuntimeConfig:
    logLevel: debug
```
3. Upgrade the cluster to a version that has the changes from this PR. The upgrade should be successful with the ctrcfg CR changes intact. Check the controller version for the MCs, it should have been updated to the new one.

Check the generated controller version to ensure that we resync and create a new one on upgrades.

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -423,17 +423,25 @@ func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.Cont
 
 func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.ContainerRuntimeConfig, err error, args ...interface{}) error {
 	statusUpdateErr := retry.RetryOnConflict(updateBackoff, func() error {
-		if cfg.GetGeneration() != cfg.Status.ObservedGeneration {
-			cfg.Status.ObservedGeneration = cfg.GetGeneration()
-			cfg.Status.Conditions = append(cfg.Status.Conditions, wrapErrorWithCondition(err, args...))
-		} else if cfg.GetGeneration() == cfg.Status.ObservedGeneration && err == nil {
-			// If the CR was created before a matching label was added, the CR would be in failure state
-			// However the observed generation would be the same, so check if err is nil as well
-			// Which means that, the ctrcfg was finally successfully able to sync. In that case update the status
-			// to success and clear the previous failure status
-			cfg.Status.Conditions = []mcfgv1.ContainerRuntimeConfigCondition{wrapErrorWithCondition(err, args...)}
+		newcfg, getErr := ctrl.mccrLister.Get(cfg.Name)
+		if getErr != nil {
+			return getErr
 		}
-		_, updateErr := ctrl.client.MachineconfigurationV1().ContainerRuntimeConfigs().UpdateStatus(context.TODO(), cfg, metav1.UpdateOptions{})
+		// Update the observedGeneration
+		if newcfg.GetGeneration() != newcfg.Status.ObservedGeneration {
+			newcfg.Status.ObservedGeneration = newcfg.GetGeneration()
+		}
+		// To avoid a long list of same statuses, only append a status if it is the first status
+		// or if the status message is different from the message of the last status recorded
+		// If the last status message is the same as the new one, then update the last status to
+		// reflect the latest time stamp from the new status message.
+		newStatusCondition := wrapErrorWithCondition(err, args...)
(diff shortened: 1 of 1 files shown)


umohnani8:
@runcom @kikisdeliveryservice PTAL

kikisdeliveryservice:
I'll test this out next week. 👍 

/assign @kikisdeliveryservice

runcom:
cc @yuqi-zhang

umohnani8:
@kikisdeliveryservice @yuqi-zhang PTAL when you have a few :)

kikisdeliveryservice:
> @kikisdeliveryservice @yuqi-zhang PTAL when you have a few :)

ah thanks for the reminder!

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -423,15 +423,23 @@ func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.Cont
...
-		if cfg.GetGeneration() != cfg.Status.ObservedGeneration {
-			cfg.Status.ObservedGeneration = cfg.GetGeneration()
-			cfg.Status.Conditions = append(cfg.Status.Conditions, wrapErrorWithCondition(err, args...))
-		} else if cfg.GetGeneration() == cfg.Status.ObservedGeneration && err == nil {
-			// If the CR was created before a matching label was added, the CR would be in failure state
-			// However the observed generation would be the same, so check if err is nil as well
-			// Which means that, the ctrcfg was finally successfully able to sync. In that case update the status
-			// to success and clear the previous failure status
-			cfg.Status.Conditions = []mcfgv1.ContainerRuntimeConfigCondition{wrapErrorWithCondition(err, args...)}
+		newcfg, getErr := ctrl.mccrLister.Get(cfg.Name)
```
yuqi-zhang, line 426: I'm a bit confused here, are you fetching the current config again here to ensure you have the latest mccr config?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -523,71 +517,79 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
-				if err != nil {
-					glog.V(2).Infoln(cfg, err, "error merging user changes to storage.conf: %v", err)
-				} else {
-					configFileList = append(configFileList, generatedConfigFile{filePath: storageConfigPath, data: storageTOML})
-				}
+		mc, err := ctrl.client.MachineconfigurationV1().MachineConfigs().Get(context.TODO(), managedKey, metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return ctrl.syncStatusOnly(cfg, err, "could not find MachineConfig: %v", managedKey)
+		}
+		isNotFound := errors.IsNotFound(err)
```
yuqi-zhang, line 524: since we're defining this anyways might as well move it above the if so we can use the variable there too

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -523,71 +517,79 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+		// If we have seen this generation and the sync didn't fail, then skip
+		if !isNotFound && cfg.Status.ObservedGeneration >= cfg.Generation && cfg.Status.Conditions[len(cfg.Status.Conditions)-1].Type == mcfgv1.ContainerRuntimeConfigSuccess {
+			// But we still need to compare the generated controller version because during an upgrade we need a new one
+			mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
```
yuqi-zhang, line 528: So this check is the crux of the change for the bug right? To still sync even if we've observed a generation?

umohnani8:
@runcom @kikisdeliveryservice PTAL for a final review please

PR #1987, opened 2020-08-07
Bug 1867080: Revert "[Baremetal] Set Keepalived default mode to unicast"

hardys:
We're seeing bootstrapping issues in CI, ref:

https://bugzilla.redhat.com/show_bug.cgi?id=1867080

This reverts commit b22920202aa07fba2a2882581a986f12cb7dd4ff.
This reverts commit 37fa01d.

### DIFF
--- a/manifests/baremetal/keepalived.yaml
+++ b/manifests/baremetal/keepalived.yaml
@@ -95,7 +95,7 @@ spec:
     image: {{ .Images.BaremetalRuntimeCfgBootstrap }}
     env:
       - name: ENABLE_UNICAST
-        value: "yes"
+        value: "no"
       - name: IS_BOOTSTRAP
         value: "yes"
     command:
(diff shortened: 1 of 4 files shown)


hardys:
/cc @yboaron @celebdor @stbenjam 

Lets see if this helps with the CI issues, marked WIP until we confirm if we need to revert or if there's another fix we can do quickly to unblock CI.

hardys:
/retitle Bug 1867080: Revert "[Baremetal] Set Keepalived default mode to unicast"

hardys:
@celebdor and team are looking at the fix for the unicast issues, but in the meantime we need to unblock CI so I think it's reasonable to go ahead with this revert.

I re-triggered e2e-metal-ipi (which passed last week) just to be sure there aren't other issues, so we can be confident that landing this revert will improve the CI pass rate.

hardys:
https://github.com/openshift/baremetal-runtimecfg/pull/84 landed instead

PR #1990, opened 2020-08-07
Bug 1861095: daemon: Log when we validate that FIPS is on

cgwalters:
I was surprised we weren't logging anything related to this when
I went to double-check the node state while working on
https://github.com/openshift/origin/pull/25362

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1861095

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -629,11 +629,14 @@ func checkFIPS(current, desired *mcfgv1.MachineConfig) error {
 		return errors.Wrapf(err, "Error parsing FIPS file at %s", fipsFile)
 	}
 	if desired.Spec.FIPS == nodeFIPS {
+		if desired.Spec.FIPS {
+			glog.Infof("FIPS is configured and enabled")
+		}
 		// Check if FIPS on the system is at the desired setting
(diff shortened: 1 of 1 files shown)


cgwalters:
Also looking at this code...two other things:

- I really think we need to kill the "silently continue if we don't find the file" code and refactor it so that only applies when building this on !Linux to run the unit tests.  Silently continuing if FIPS is specified but the file doesn't exist (maybe we typo'd the constant or whatever) would be evil
- Why is this function checking the node state at all?  Talking about "reconciliation" between two MCs should be independent of the node; instead, we should do that in `dn.validateOnDiskState(expectedConfig)` right?

cgwalters:
CI outage is hopefully over
/retest

ericavonb:
logging fips seems useful to me.
/lgtm

kikisdeliveryservice:
> /test e2e-aws-scaleup-rhel7

@sinnykumari : FYI that test is now named  `e2e-aws-workers-rhel7` :smiley:

sinnykumari:
@kikisdeliveryservice ah ok. Thanks!

PR #1992, opened 2020-08-08
Bug 1867608: ds/machine-config-daemon: Set maxUnavailable 10%

sdodson:
This daemonset is not critical to availability like SDN or DNS pods therefore
allow maxUnavailable to scale with cluster size. This roughly increases rollout
speed by 10x on a 250 node cluster.

How to verify this, provision a cluster with 20 or more hosts. Confirm that updates
to the machine-config-daemonset allows for up to 10% to update at once but otherwise
operates normally.

### DIFF
--- a/manifests/machineconfigdaemon/daemonset.yaml
+++ b/manifests/machineconfigdaemon/daemonset.yaml
@@ -7,6 +7,9 @@ spec:
   selector:
     matchLabels:
       k8s-app: machine-config-daemon
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 10%
   template:
     metadata:
       name: machine-config-daemon
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -1411,6 +1411,9 @@ spec:
   selector:
     matchLabels:
       k8s-app: machine-config-daemon
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 10%
   template:
     metadata:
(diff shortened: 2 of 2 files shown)


sdodson:
This is similar to others I'm updating as well, but for some reason this one *feels* riskier than those just given the criticality of MCO but I don't think it really is. Feedback welcome though.

https://github.com/openshift/cluster-monitoring-operator/pull/902
https://github.com/openshift/cluster-node-tuning-operator/pull/149

sdodson:
/retitle Bug 1867608: ds/machine-config-daemon: Set maxUnavailable 10%

runcom:
@sdodson given how critical the MCO is, what do you think if we defer this to the next release to give it more and more time to soak? the issue is known (around timing) but it'll make us more comfortable and find us ready to debug if anything goes south. Would that work?

cgwalters:
/approve
Agree with the rationale here - in fact *most* of the time the MCD is effectively idle I believe.  We watch for ssh events from journald but I'm not thinking offhand of anything else we react to *other* than when the MCC changes the `desiredConfig` annotation.

sdodson:
Yeah, I'm fine with deferring this as long as we're open to backporting to 4.6.z before it transitions to maintenance phase. This is just an optimization.

cgwalters (approved):
/lgtm

sdodson:
This has sat in master for a month now and as far as I know there's been no ill effect.
/cherry-pick release-4.6

PR #1986, opened 2020-08-06
Bump golangci-lint to v1.42.1 and switch from golint to revive 

LorbusChris:
See individual commit messages for what's happening here.

/hold
til after freeze

PR #1993, opened 2020-08-08
tools.go: Bump go-bindata to module-aware v3.1.3

LorbusChris:
and regenerate bindata.

/hold
til after freeze


LorbusChris:
Folded into https://github.com/openshift/machine-config-operator/pull/1986

PR #1995, opened 2020-08-10
baremetal: bootstrap api vip until down

celebdor:
Now that the cluster is grown from boostrap to the master nodes, it
should make the clustering easier if we just keep the API vip in the
bootstrap until we're done bootstrapping.

**- What I did**
We used to transition the API VIP to the master nodes as soon as
we could. This was necessary with the way Etcd used to cluster. Now
that the clustering starts with the bootstrap node and grows to the
master nodes, it should be possible to reduce deployment instability
by keeping the VIP in the bootstrap until the bootstrap is taken down.

**- How to verify it**
Make a deployment and see that the clustering succeeds.

Keep API VIP in the bootstrap while it exists.

### DIFF
--- a/manifests/baremetal/keepalived.conf.tmpl
+++ b/manifests/baremetal/keepalived.conf.tmpl
@@ -4,21 +4,18 @@
 # For more information, see installer/data/data/bootstrap/baremetal/README.md
 # in the installer repo.
 
-{{`{{$nonVirtualIP := .NonVirtualIP}}`}}
-
 {{`vrrp_instance {{.Cluster.Name}}_API {
     state BACKUP
     interface {{.VRRPInterface}}
     virtual_router_id {{.Cluster.APIVirtualRouterID }}
     priority 50
     advert_int 1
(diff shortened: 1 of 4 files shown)


manifests/baremetal/keepalived.conf.tmpl
```diff
@@ -4,21 +4,18 @@
...
 
-{{`{{$nonVirtualIP := .NonVirtualIP}}`}}
-
 {{`vrrp_instance {{.Cluster.Name}}_API {
     state BACKUP
     interface {{.VRRPInterface}}
     virtual_router_id {{.Cluster.APIVirtualRouterID }}
     priority 50
     advert_int 1
+    nopreempt
```
yboaron, line 13: After testing deployment with this PR, it seems that configuring 'nopreempt' only at bootstrap side doesn't change the behavior, bootstrap moves to one of the master nodes as soon as API is up.
So, I would suggest to delete the nopreempt

PR #1996, opened 2020-08-10
Bug 1866693: docs: drop kernel-devel from supported extensions

sinnykumari:
We don't want to encourage users to install kernel-devel
as an extension in BaseOS. Preferred way is to install
it in a development build container.


kikisdeliveryservice:
Doc fix only, no retest required.

PR #1999, opened 2020-08-10
Bug 1867392: Fixes multiple NICs OVN/OVS config

trozet:
Deploying with multiple NICs will end up placing all NICs on the same
default NM connection. If one of these interfaces fails to DHCP or does
not DHCP in time during boot, NM-wait-online will fail (even though 1
NIC may have DHCP'ed correctly). Therefore the ovs-configuration service
should not require NM-wait-online, but we should wait until it is
completed and we do want it. However, if it fails we still try to to
check the system if a default gateway is present.

Additionally, since multiple NICs may share the same NM connection, we
cannot always just bring it down. Instead, change the behavior to just
disconnect the NIC and then bring up the higher priority connection
later to preserve the default NM connection.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -77,12 +77,14 @@ contents:
     fi
 
     # bring down any old iface
-    nmcli conn down $old_conn
+    nmcli device disconnect $iface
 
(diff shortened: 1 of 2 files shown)


trozet:
@danwinship PTAL got confirmation it fixed the reporter's issue

yuvalk:
yep, I can attest this PR solved the issue in our BM environment.

PR #2001, opened 2020-08-11
HACKING: add example for talking to MCS via curl

miabbott:
I ran into this question while verifying #1960, so let's document it


kikisdeliveryservice:
here's your confirmation it indeed works :smile: 

/assign @yuqi-zhang

kikisdeliveryservice:
Doc only retest not required

PR #2002, opened 2020-08-12
[FCOS] Rebase Aug 12

vrutkovs:
Rebase fcos on latest commits from release-4.5

TARGET_COMMIT=$(git describe --tags fcos | sed 's/.*-g//')
git checkout -b fcos-rebase-aug12 origin/release-4.5
git rebase --onto fcos $TARGET_COMMIT

/cc @LorbusChris


templates/common/_base/files/iptables-modules.yaml
```diff
@@ -0,0 +1,7 @@
+filesystem: "root"
```
LorbusChris, line 1: this line must be gone for FCCT configs^

PR #2005, opened 2020-08-17
Bug 1865839: daemon: cleanup pending deployment on an error

sinnykumari:
We handle rpm-ostree operations such as OS update,
karg update, kernel switch and additional package
overlay in applyOSChanges(). When an error occurs
during any one of these operations, we would want
to rollback any changes we have performed on booted
OS so that we can cleanly retry again to apply
the changes. rpm-ostree provides us `cleanup -p`
options which removes pending deployment.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -306,7 +306,14 @@ func ExtractOSImage(imgURL string) (osImageContentDir string, err error) {
 	return
 }
 
-func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+// Remove pending deployment on OSTree based system
+func removePendingDeployment() error {
+	args := []string{"cleanup", "-p"}
+	_, err := runGetOut("rpm-ostree", args...)
+	return err
+}
+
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -306,7 +306,17 @@ func ExtractOSImage(imgURL string) (osImageContentDir string, err error) {
...
 
-func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+// Remove pending deployment on OSTree based system
+func removePendingDeployment() error {
+	args := []string{"cleanup", "-p"}
+	_, err := runGetOut("rpm-ostree", args...)
+	if err != nil {
+		return err
+	}
+	return nil
```
runcom, line 316: nit you can `return err` directly if you don't plan to wrap it

runcom:
💯 thanks Sinny for the quick fix!

pkg/daemon/update.go
sinnykumari: right, fixed.

PR #2006, opened 2020-08-17
Configure OVS after firstboot is done

vrutkovs:
Run OVS configuration only after the machine has pivoted. This is crucial for OKD installs which uses standard FCOS image with no OVS pre-installed.

### DIFF
--- a/templates/common/_base/units/ovs-configuration.service.yaml
+++ b/templates/common/_base/units/ovs-configuration.service.yaml
@@ -3,6 +3,8 @@ enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else}}false{{end}}
 contents: |
   [Unit]
   Description=Configures OVS with proper host networking configuration
+  # Removal of this file signals firstboot completion
+  ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
   Wants=NetworkManager-wait-online.service


LorbusChris:
/approve
I think this should also just work on OCP

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -6,8 +6,8 @@ contents: |
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
   Wants=NetworkManager-wait-online.service
-  After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
-  Before=network-online.target kubelet.service crio.service
```
cgwalters, line 10: Was dropping `Before=network-online.target` intentional?

How about adding  `ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json` to this service instead?

And we should do that for `kubelet.service` and `crio.service` too probably - basically everything we don't want to run until we've pivoted, rather than implicitly relying on having the firstboot service initiate a reboot before they've started.

vrutkovs: >Was dropping Before=network-online.target intentional?

Yes, after `network-online` mcd-pull and mcd-firstboot started, so ovs-configuration would always happen after network-online anyway

> How about adding ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json to this service instead?

Good idea, added that

LorbusChris: I think it wouldn't hurt to leave it, though? Just in case we shuffle things around again...

cgwalters:
I think this is a clear bugfix, and we should avoid starting services before pivoting in general.
cc @trozet - do you agree?

vrutkovs:
OKD is expected to fail for now - it needs an updated installer.
RHEL7 test fails due to missing `jq` - https://bugzilla.redhat.com/show_bug.cgi?id=1867992

templates/common/_base/units/ovs-configuration.service.yaml
vrutkovs: Removed unnecessary After/Before/Wanted changed for now - this can be re-iterated later

trozet:
> OKD is expected to fail for now - it needs an updated installer.
> RHEL7 test fails due to missing `jq` - https://bugzilla.redhat.com/show_bug.cgi?id=1867992

It's not just jq, even without jq I don't think NM OVS will work.

trozet:
> I think this is a clear bugfix, and we should avoid starting services before pivoting in general.
> cc @trozet - do you agree?

looks OK to me and ovn-step-registry passed

ashcrow:
@vrutkovs is there a bug for this?

vrutkovs:
@ashcrow  I didn't file one (as it primarily affects OKD)

PR #2007, opened 2020-08-18
Bug 1869886: Fix to configure vmxnet3 driver in multicast mode

abhat:
When the vmxnet3 driver gets added to ovs, it does not
honor the promiscuous mode setting for multicast traffic.
This PR adds support to temporarily fix that by configuring
the allmulti flag on the interface when it's moved to ovs.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -10,6 +10,17 @@ contents:
     # key files under /etc/NetworkManager/system-connections/
     # Managing key files is outside of the scope of this script
 
+    # if the interface is of type vmxnet3 add multicast capability for that driver
+    # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
+    function configure_driver_options {
+      intf=$1
+      driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
+      echo "Driver name is" $driver
+      if [ "$driver" = "vmxnet3" ]; then
+        ifconfig "$intf" allmulti
+      fi
+    }
+    
     iface=""
     counter=0
     # find default interface
@@ -32,6 +43,10 @@ contents:
     done
 
     if [ "$iface" = "br-ex" ]; then
+      # handle vlans and bonds etc if they have already been
+      # configured via nm key files and br-ex is already up
+      ifaces=$(ovs-vsctl list-ifaces ${iface})
+      for intf in $ifaces; do configure_driver_options $intf; done
       echo "Networking already configured and up for br-ex!"
       exit 0
     fi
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
@abhat please add a BZ to this PR

/hold

trozet:
@kikisdeliveryservice this is a workaround until the driver is actually fixed in the kernel, which is being tracked by:
https://bugzilla.redhat.com/show_bug.cgi?id=1854355

@abhat we should probably have a more clear comment in the code with "REMOVEME" or something after the fix is done

I didn't see a BZ mentioned in a PR that touched this code yesterday:
https://github.com/openshift/machine-config-operator/pull/2006

Is it ok to not include one for this workaround?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -10,6 +10,17 @@ contents:
     # key files under /etc/NetworkManager/system-connections/
     # Managing key files is outside of the scope of this script
 
+    # if the interface is of type vmxnet3 add multicast capability for that driver
+    # BZ: 1854355
+    function configure_driver_options {
```
trozet, line 15: I think you also need to call this at line 47 when br-ex is already configured. This has to be done on every boot.

kikisdeliveryservice:
> I didn't see a BZ mentioned in a PR that touched this code yesterday:
> #2006
> 
> Is it ok to not include one for this workaround?

That PR should have had a BZ as well.  Please make one so we can track these changes as we are well past codefreeze.

templates/common/_base/files/configure-ovs-network.yaml
abhat: Ack.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -32,6 +43,8 @@ contents:
     done
 
     if [ "$iface" = "br-ex" ]; then
+      intf=$(ovs-vsctl list-ports ${iface})
```
abhat, line 46: is this good enough? how do we uniquely get the interface after br-ex is already configured? will there be other ports that br-ex might have?

trozet: I think you want to do `ovs-vsctl list-ifaces br-ex` here. Then loop through the interfaces and call configure_driver_options on each one. The port could be a bond, with underlying interfaces attached to it. I'm not sure if that will ever happen in a vsphere deployment, but better to be safe than sorry.

abhat: We currently aren't handling bond and vlans anyway during the initial bring-up right? If there were indeed sub-interfaces in the config that get moved to br-ex, yes, that makes sense.

trozet:
@runcom @kikisdeliveryservice PTAL

yuqi-zhang:
This shouldn't affect AWS and I don't know how to personally verify this, but if you're confident it works I am LGTM in terms of the code

PR #2008, opened 2020-08-18
Bug 1869876: Cluster upgrades silently stall when machine config pools are paused

rvanderp3:
**- What I did**
Added a log message that is written when the machine config pool is Paused, but is also updating.  This message is written when the machine config pool status is being calculated .

**- How to verify it**
1. Configure `worker` machine config to be 'Paused' 
2. Update the `99-worker-ssh` machine config with a new SSH key
3. Wait for `worker` pool to progress to `Updating`
4. Check machine-config-controller logs for an entry like: W0818 19:49:38.503997      10 status.go:94] Pool worker is Paused, but is 'Updating'.  Pool will not progress

Logs a warning when a machine config pool is Paused, but also updating

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -89,6 +89,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdated)
 		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, "", fmt.Sprintf("All nodes are updating to %s", pool.Spec.Configuration.Name))
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+
+		if pool.Spec.Paused == true { // if the pool is Paused and updating, the pool will never progress.
+			glog.Warningf("Pool %s is Paused, but is 'Updating'.  Pool will not progress", pool.Name)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


rvanderp3:
will do :)

pkg/controller/node/status.go
```diff
@@ -89,6 +89,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdated)
 		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, "", fmt.Sprintf("All nodes are updating to %s", pool.Spec.Configuration.Name))
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+
+		if pool.Spec.Paused == true { // if the pool is Paused and updating, the pool will never progress.
```
kikisdeliveryservice, line 93: Q: as per https://github.com/openshift/machine-config-operator/blob/fe05cca6e8784f60dfc03faa9684672b30e24711/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L222

Is this the correct place for this? 

Or is https://github.com/openshift/machine-config-operator/blob/cd925586f71d709a36a11945a7abf8104e3d2046/pkg/controller/node/node_controller.go#L723 better?

cgwalters:
Tentative
/approve

But I thought there was discussion about adding a Prometheus metric for this?

I also worry a bit about the level of "spam" in this case, we might want rate-limiting warnings but eh, not a blocker.

PR #2009, opened 2020-08-18
Bug 1869876: Cluster upgrades silently stall when machine config pools are paused

rvanderp3:
**- What I did**
Added a log message that is written when the machine config pool is Paused, but is also updating. This message is written when the machine config pool status is being calculated .

**- How to verify it**
1. Configure worker machine config to be 'Paused'
2. Update the 99-worker-ssh machine config with a new SSH key
3. Wait for worker pool to progress to Updating
4. Check machine-config-controller logs for an entry like: W0818 19:49:38.503997 10 status.go:94] Pool worker is Paused, but is 'Updating'. Pool will not progress

Logs a warning when a machine config pool is Paused, but also updating

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -721,6 +721,9 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	}
 
 	if pool.Spec.Paused {
+		if mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
+			glog.Infof("Pool %s is paused and will not update.", pool.Name)
+		}
 		return ctrl.syncStatusOnly(pool)
 	}
 


rvanderp3:
apologies on the churn on this PR

pkg/controller/node/status.go
```diff
@@ -89,6 +89,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdated)
 		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, "", fmt.Sprintf("All nodes are updating to %s", pool.Spec.Configuration.Name))
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+
+		if pool.Spec.Paused == true { // if the pool is Paused and updating, the pool will never progress.
```
kikisdeliveryservice, line 93: delete comment it's just repeating the glog below

pkg/controller/node/status.go
```diff
@@ -89,6 +89,10 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
...
+			glog.Warningf("Pool %s is Paused, but is 'Updating'.  Pool will not progress", pool.Name)
```
kikisdeliveryservice, line 94: Is there a reason you put it here? Why not just put it where we already check for paused?: https://github.com/openshift/machine-config-operator/blob/cd925586f71d709a36a11945a7abf8104e3d2046/pkg/controller/node/node_controller.go#L723

kikisdeliveryservice: Also I don't think this should be glog.Warning(), could just be glog.Info

kikisdeliveryservice: Can we get this comment into 1 sentence? "Pool %s is paused and will not update/progress/<insert word of choice"."

runcom:
One day we could perhaps avoid setting Updating when the pool is paused. Those 2 seems conflicting somehow. It's Paused but shows updating, maybe it's better if it's either Paused or Updating (just my own confusion with this). I think this log makes sense tho pending Kirsten's comments

pkg/controller/node/status.go
rvanderp3: @kikisdeliveryservice thanks for that feedback.  I went ahead and committed those changes.  I did not consider putting this change around where we checked for 'Paused'.  That would make sense too and can certainly make that change if you want.

kikisdeliveryservice: if you can move & delete comment that would be great.

rvanderp3: sure thing.  done!

rvanderp3:
@kikisdeliveryservice I moved the check in to node_controller where we check for paused.  I also tested the PR with cluster-bot and it looks good.

pkg/controller/node/node_controller.go
```diff
@@ -721,6 +721,9 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	}
 
 	if pool.Spec.Paused {
+		if mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
+			glog.Infof("Pool %s is paused and will not update.", pool.Name)
+		}
 		return ctrl.syncStatusOnly(pool)
```
runcom, line 727: on a second thought - while this log looks good, it's still gonna be pretty hidden within our logs which is again not giving any useful information to the admin

one possible solution would be to enhance `syncStatusOnly` with logic to show a message in the Updating condition (or somewhere else) saying it's Paused (we ultimately need something at the MCP <-> MCO objects level to be admin-visible)

install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml
```diff
@@ -25,6 +25,10 @@ spec:
       configuration is failing.
     name: Degraded
     type: string
+  - JSONPath: .spec.paused
+    description: When paused, updates to the machineconfigpool will not be processed.
```
kikisdeliveryservice, line 29: please remove this

install/0000_80_machine-config-operator_01_machineconfigpool.crd.yaml
```diff
@@ -25,6 +25,10 @@ spec:
       configuration is failing.
     name: Degraded
     type: string
+  - JSONPath: .spec.paused
```
kikisdeliveryservice, line 28: please remove this - this extends the output to be way too large and we can deal with this in a different way. once this is reverted we can merge, we will add further changes as a follow-on pr.

rvanderp3: just reverted that change

pkg/controller/node/node_controller.go
kikisdeliveryservice: Makes sense I think we can do this via a follow-on pr.

PR #2011, opened 2020-08-19
Bug 1868158: gcp, azure: Handle azure vips similar to GCP

squeed:
This PR does the following things:

- Rename gcp-routes-controller to apiserver-watcher, since it is generic
- Remove obsolete service-management mode from gcp-routes-controller
- Change downfile directory to /run/cloud-routes from /run/gcp-routes
- Write $VIP.up as well as $VIP.down
- Add an azure routes script that fixes hairpin.

Background: Azure hosts cannot hairpin back to themselves over a load balancer. Thus, we need to redirect traffic to the apiserver vip to ourselves via iptables. However, we should only do this when our local apiserver is running.

The apiserver-watcher drops a $VIP.up and $VIP.down file, accordingly, depending on the state of the apiserver. Then, we add or remove iptables rules that short-circuit the load balancer.

Unlike GCP, we don't need to do this for external traffic, only local clients.

**- How to verify it**
Install on azure, ensure connections to the internal API load balancer are reliable - both when the local apiserver process is running and stopped.

Masters on azure can now reliably connect to the apiserver service, without encountering hairpin issues


cmd/apiserver-watcher/README.md
```diff
@@ -0,0 +1,65 @@
...
+ |  +---------+  |          |                 |
+ |  | kubelet +------------->  layer-3        |
+ |  +---------+  |          |  load balancer  |
+ |               |     +----+                 |
+ |  +---------+  |     |    +-----------------+
+ |  |apiserver+<-------+
+ |  +---------+  |
+ |               |
+ +---------------+
+```
```
sttts, line 21: ascii figures 😍

squeed:
oops, typo'd the image key. Good thing the tests mostly failed...

squeed:
/refresh
(azure results have disappeared)

cmd/apiserver-watcher/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Functionality
+
+The apiserver-watcher is installed on all the masters and monitors the
+apiserver process /readyz.
+
+When /readyz fails,  write `/run/cloud-routes/$VIP.down`, which tells the
+provider-specific service to update iptables rules.
+
+Separately,
```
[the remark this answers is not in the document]
squeed: Hah. Maybe?

cmd/apiserver-watcher/run.go
```diff
@@ -151,83 +140,62 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
-			if err != nil {
-				return nil, fmt.Errorf("failed to lookup host %s: %v", uri.Hostname(), err)
-			}
-			if len(addrs) != 1 {
-				return nil, fmt.Errorf("hostname %s has %d addresses, expected 1 - aborting", uri.Hostname(), len(addrs))
-			}
-			h.vip = addrs[0]
-			glog.Infof("Using VIP %s", h.vip)
+		if len(addrs) != 1 {
+			return nil, fmt.Errorf("hostname %s has %d addresses, expected 1 - aborting", uri.Hostname(), len(addrs))
```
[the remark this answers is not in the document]
squeed: (This isn't a behavior change, just removing some dead code and a corresponding re-indentation)

It could happen if we somehow switch to RRDNS.

cmd/apiserver-watcher/run.go
```diff
@@ -151,83 +140,62 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+	}
+	glog.Infof("healthcheck succeeded, removed downfile %s.down", h.vip)
+	if err := writeFile(h.vip, "up"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func writeFile(vip, state string) error {
+	file := path.Join(runOpts.rootMount, downFileDir, fmt.Sprintf("%s.%s", vip, state))
```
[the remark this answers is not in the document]
squeed: duh, good point.

templates/master/00-master/azure/files/etc-kubernetes-manifests-apiserver-watcher.yaml
```diff
@@ -1,17 +1,17 @@
 mode: 0644
-path: "/etc/kubernetes/manifests/gcp-routes-controller.yaml"
+path: "/etc/kubernetes/manifests/apiserver-watcher.yaml"
```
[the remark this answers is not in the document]
squeed: I could also keep it as the same filename; the filename definitely doesn't matter.

templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
```diff
@@ -0,0 +1,180 @@
+mode: 0755
```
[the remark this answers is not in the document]
squeed: It's pretty different from GCP, so it needs a review.

templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
```diff
@@ -0,0 +1,180 @@
...
+path: "/opt/libexec/openshift-azure-routes.sh"
+contents:
+  inline: |
+    #!/bin/bash
```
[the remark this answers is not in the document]
squeed: I agree; I don't like adding all this bash. If it helps, I extract it and run it through shellcheck automatically. I could probably add that to `make verify`.

For 4.7, should we add an item to rewrite all this in go?

templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
```diff
@@ -0,0 +1,16 @@
+name: openshift-azure-routes.path
+enabled: true
+contents: |
+  [Unit]
+  Description=Watch for downfile changes
+  Wants=network-online.target
+  After=network-online.target
+  Before=kubelet.service
```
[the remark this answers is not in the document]
squeed: Sure, I can add those. This isn't as critical, since it won't do anything until the apiserver-watcher starts, but it's good to be consistent.

squeed:
azure CI failure is confusing; at this point in time, my script is a no-op. However, etcd seems to be down (but must-gather ran??). Confusing all the same. re-running, but investigating interactively.

cmd/apiserver-watcher/README.md
```diff
@@ -0,0 +1,65 @@
...
+So, an agent exists on the node. It programs the node (either via iptables or routing tables) to
+accept traffic destined for the VIP. However, this has a problem: all hairpin traffic
+to the balanced servce is *always* handled by that backend, even if it is down
+or otherwise out of rotation.
+
+We want to withdraw the internal API service from google-routes redirection when
+it's down, or else the node (i.e. kubelet) loses access to the apiserver VIP
+and becomes unmanagable.
+
+### Azure
```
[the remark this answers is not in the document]
squeed: Not sure what you mean exactly; apiserver-watcher is identical on azure and gzp. I did add pointers to the cloud-provider-specific scripts, so maybe that's helpful?

templates/master/00-master/azure/files/etc-kubernetes-manifests-apiserver-watcher.yaml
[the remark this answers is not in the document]
squeed: The old static pod doesn't matter; it writes to `/run/gcp-routes` while the new one is `/run/cloud-routes`, so they can happily coexist (and should, until the service is swapped).

squeed:
quota
/test e2e-azure

squeed:
Still out of quota
`autorest/azure: Service returned an error. Status=<nil> Code=\"OperationNotAllowed\" Message=\"Operation could not be completed as it results in exceeding approved standardDSv3Family Cores quota. Additional details - Deployment Model: Resource Manager, Location: centralus, Current Limit: 1000, Current Usage: 998, Additional Required: 8, (Minimum) New Limit Required: 1006.`

squeed:
Azure CI still throttling to death :-(

kikisdeliveryservice:
Maybe it will work now?

/retest

templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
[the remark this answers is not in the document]
kikisdeliveryservice: I would love this to be in go!  defer to @cgwalters / @runcom on whether waiting to 4.7 makes sense.

squeed:
whoops, bug
`openshift-azure-routes[31554]: /opt/libexec/openshift-azure-routes.sh: line 94: v4vips[${ipt_vip}]: unbound variable`

squeed:
quota
/test e2e-azure

squeed:
/test e2e-azure
still infra issues

kikisdeliveryservice:
azure quota. trying again for good luck :crossed_fingers: 

/retest

squeed:
rebased.
/test e2e-azure

deads2k:
azure quota

/retest

deads2k:
Azure has re-opened, so this may produce results today

sttts:
No nodes are available. Digging....

templates/master/00-master/azure/units/openshift-azure-routes.service.yaml
```diff
@@ -0,0 +1,11 @@
+name: openshift-azure-routes.service
+enabled: true
```
cgwalters, line 2: I think this is going to start it unconditionally instead of just triggering via the `.path` unit - but would need to do some testing to verify.

templates/master/00-master/gcp/files/etc-kubernetes-manifests-apiserver-watcher.yaml
```diff
@@ -0,0 +1,38 @@
+mode: 0644
+path: "/etc/kubernetes/manifests/apiserver-watcher.yaml"
```
cgwalters, line 2: Hmm...does really need to be a static pod versus part of the existing MCD?  I guess since the previous code was a static pod too it's OK to keep as is for now.

sttts: Yes, that's probably the only reason.

templates/master/00-master/azure/units/openshift-azure-routes.service.yaml
cgwalters: Fixed this.

templates/master/00-master/gcp/files/etc-kubernetes-manifests-apiserver-watcher.yaml
cgwalters: I wrote some small code to teach the MCD to behave differently on the control plane as part of https://github.com/openshift/machine-config-operator/pull/1946 FWIW.

cgwalters:
OK rebased :surfer:  on #2061
(Really we could change the pull to use the renamed APIServerWatcherKey but this way is IMO much less confusing for the future)

sttts:
Two runs of e2e-azure terminated in a similar way. So there is still something wrong.

cgwalters:
Hm at least that Azure run hit https://github.com/systemd/systemd/issues/9656

cgwalters:
The previous azure failure looks like the kube apiserver never came up.  I don't have expertise in debugging that.
/test e2e-azure
just to gather more data.

sttts:
Here I only see logs of two masters. Kube-apiserver does not come up because `connection refused` contacting etcd:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2011/pull-ci-openshift-machine-config-operator-master-e2e-azure/1303466099257380864

nodes.json and all other resource jsons are empty. But as the masters have pods running, I assume that the kubelet succeeded to connect to the API.

sttts:
Another run: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2011/pull-ci-openshift-machine-config-operator-master-e2e-azure/1302952710894522368

This one **does not** show the unit issues systemd/systemd#9656.

But again only two masters have logs. Resources jsons are all empty. And I don't see anything about etcd-operator (could be on the third master node though).

hexfusion:
After reviewing a cluster built with this change I noticed 2 of the 3 master nodes booted after pivot without recieving IP address from the provider. Looking at the serial console for the node showed it booted but with no address for eth0. After rebooting the nodes it came up fine and the cluster reconciled.

```
Red Hat Enterprise Linux CoreOS 46.82.202009080940-0 (Ootpa) 4.6
SSH host key: SHA256:z5wmNZChPOPAP+QzI+eENhz3NJW3g1bX8Fhk+jtfOKE (RSA)
SSH host key: SHA256:GIE2+/I0IXTU62qUozpVDJC/n7NrtWsjrWuSywVbN2Q (ED25519)
SSH host key: SHA256:fc5Z5hSlcb9ylTj+XYudXMXJxrm7YABb/hWvKRmOhK4 (ECDSA)
eth0:
sbatsche-dhcp-2020-09-jkfrz-master-0 login:
```

hexfusion:
An additional test with the latest ci release did not reproduce the issue.

hexfusion:
/test e2e-azure

retest with same payload as https://github.com/openshift/machine-config-operator/pull/2011#issuecomment-689677001 did not reproduce the condition. Possible race?

deads2k:
azure is fixed, removing my hold

/hold cancel

hexfusion:
/hold
I can't reproduce the issue outside of this PR. once I can lets merge.

hexfusion:
/hold cancel

we will monitor net gain is large

templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
```diff
@@ -0,0 +1,17 @@
+name: openshift-azure-routes.path
+enabled: true
+contents: |
+  [Unit]
+  Description=Watch for downfile changes
+  Wants=network-online.target
+  After=network-online.target
+  Before=kubelet.service crio.service
```
jlebon, line 8: Path units get started with `paths.target` by default. So this is creating an ordering cycle because `network-online.target` wants to happen after `basic.target`:

```
[  324.981226] systemd[1]: network-online.target: Found ordering cycle on network.target/start
[  325.000115] systemd[1]: network-online.target: Found dependency on ovs-vswitchd.service/start
[  325.013025] systemd[1]: network-online.target: Found dependency on basic.target/start
[  325.027061] systemd[1]: network-online.target: Found dependency on paths.target/start
[  325.035592] systemd[1]: network-online.target: Found dependency on openshift-azure-routes.path/start
[  325.049473] systemd[1]: network-online.target: Found dependency on network-online.target/start
[  325.058497] systemd[1]: network-online.target: Job network.target/start deleted to break ordering cycle starting with network-online.target/start
[ SKIP ] Ordering cycle found, skipping Network
```

(this is from the serial console logs from https://bugzilla.redhat.com/show_bug.cgi?id=1877483)

Depending on which unit systemd decides to delete, networking might not come up.

Hmm, I think we probably need to add `DefaultDependencies=false` here?

cgwalters: Also related to this, we should patch the MCO to add an alert if it detects a systemd ordering cycle.

cgwalters:
/lgtm cancel
per https://github.com/openshift/machine-config-operator/pull/2011#discussion_r486346793

I'll look at fixing it.

templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
cgwalters: Looking at this I think those dependencies are just wrong - we will only write files in there when the static pod comes up, so kubelet will be up for that.  IOW I think we want:

```
diff --git a/templates/master/00-master/azure/units/openshift-azure-routes.path.yaml b/templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
index 75298150..a3795a30 100644
--- a/templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
+++ b/templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
@@ -3,9 +3,6 @@ enabled: true
 contents: |
   [Unit]
   Description=Watch for downfile changes
-  Wants=network-online.target
-  After=network-online.target
-  Before=kubelet.service crio.service
   ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json
 
   [Path]
@@ -14,4 +11,4 @@ contents: |
   MakeDirectory=true
 
   [Install]
-  WantedBy=multi-user.target
+  WantedBy=paths.target
```

cgwalters: Force pushed an update similar to that.

cgwalters:
I'm holding the mutex :lock: around force pushing updates here.

deads2k:
> I'm holding the mutex  around force pushing updates here.

Thanks

templates/master/00-master/azure/units/openshift-azure-routes.path.yaml
miabbott: > Also related to this, we should patch the MCO to add an alert if it detects a systemd ordering cycle.

https://github.com/openshift/machine-config-operator/issues/2074

cgwalters:
OK we have a green azure run here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2011/pull-ci-openshift-machine-config-operator-master-e2e-azure/1304091886448807936
I'm digging into the node logs now to make sure we fixed the systemd unit issue.

cgwalters:
Confirmed we fixed the ordering cycle by looking at the journal from the current run versus the previous:

```
walters@toolbox /tmp> grep 'systemd.*Found ordering cycle' journal-old
Sep 09 17:46:41.436748 ci-op-llmvs4bh-9bd41-b2fmn-master-0 systemd[1]: openvswitch.service: Found ordering cycle on ovsdb-server.service/start
Sep 09 17:46:41.436884 ci-op-llmvs4bh-9bd41-b2fmn-master-0 systemd[1]: basic.target: Found ordering cycle on paths.target/start
Sep 09 17:50:22.102635 ci-op-llmvs4bh-9bd41-b2fmn-master-0 systemd[1]: network-online.target: Found ordering cycle on network.target/start
Sep 09 17:50:22.102706 ci-op-llmvs4bh-9bd41-b2fmn-master-0 systemd[1]: NetworkManager.service: Found ordering cycle on openvswitch.service/start
Sep 09 17:50:22.102793 ci-op-llmvs4bh-9bd41-b2fmn-master-0 systemd[1]: NetworkManager.service: Found ordering cycle on basic.target/start
walters@toolbox /tmp> grep 'systemd.*Found ordering cycle' journal
walters@toolbox /tmp [1]> 
```

/approve

cgwalters:
Eh we had prior approvals on the old code and the new one just fixes systemd ordering issues so
/lgtm

deads2k:
YAY!

PR #1773, opened 2020-06-02
Bug 1853298: [openstack/vsphere] Move dispatcher script to common

mandre:
This alignes OpenStack and vSphere platforms with with the BM one after
#1763 and #1765.

PR #2012, opened 2020-08-19
Bug 1870285: OpenStack: Ensure resolv.conf exists before running podman

mandre:
Manual backport of
https://github.com/openshift/machine-config-operator/pull/1773 to
release-4.5 because of merge conflict. We can't share the prepender
script between control plane and compute nodes in OpenStack due to
dependency on the DNS VIP.

This basically boils down to porting
https://github.com/openshift/machine-config-operator/pull/1765 to the
openstack worker prepender script that was forgotten in the initial
commit.

VShpere counterpart doesn't run podman and hence doesn't need the patch.


Fedosin (commented):
/lgtm

PR #2013, opened 2020-08-20
Bug 1872444: Run ovs-configuration.service after the node has hostname stabilized

vrutkovs:
**- What I did**
If ovs-configuration is enabled it should block crio/kubelet from starting until the node has hostname set. Currently it [races with the kubelet](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/promote-release-openshift-okd-machine-os-content-e2e-aws-4.6/1296266834999250944) in OKD, so often tests fails due to crashlooping OVN/DNS pods and/or worker nodes not joining the cluster

**- How to verify it**
Run OVN-enabled cluster.

OKD has OVN enabled by default, so now results are looking much better: 
* [1](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2013/pull-ci-openshift-machine-config-operator-master-okd-e2e-aws/1296362803300405248) - storage flakes
* [2](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2013/pull-ci-openshift-machine-config-operator-master-okd-e2e-aws/1296398459334561792) - minor flakes
* [3](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2013/pull-ci-openshift-machine-config-operator-master-okd-e2e-aws/1296427427123171328) - pass

Configure OVS after hostname has been set

### DIFF
--- a/templates/common/_base/units/ovs-configuration.service.yaml
+++ b/templates/common/_base/units/ovs-configuration.service.yaml
@@ -7,6 +7,7 @@ contents: |
   ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
+  Requires=node-valid-hostname.service
   Wants=NetworkManager-wait-online.service
   After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
   Before=network-online.target kubelet.service crio.service


cgwalters:
Right, same as the change we made for other services, `-firstboot.service` has
```
  [Install]
  WantedBy=multi-user.target
  RequiredBy=crio.service kubelet.service
```

So LGTM.

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -20,3 +20,5 @@ contents: |
 
   [Install]
   WantedBy=network-online.target
+  RequiredBy=crio.service
+  RequiredBy=kubelet.service
```
trozet, line 24: This is in the install section, we already have on line 12, which will prevent kubelet or crio from starting until this service is complete.

vrutkovs: It seems to be a common pattern to include both - https://www.freedesktop.org/software/systemd/man/systemd.unit.html#%5BUnit%5D%20Section%20Options. Although I'm not sure why `Before=` is not sufficient here

cgwalters: If `foo.service` is `After=bar.service`, but `bar.service` fails, then `foo.service` will still start.

`foo.service` must have `Requires=bar.service` in order for systemd to not start `foo` if `bar` fails.

trozet: Right, but I thought the issue here is that you are seeing a race between ovs-configuration service and kubelet. Using After/Before should ensure that kubelet does not start until ovs-configuration is finished (regardless of if ovs-configuration succeeds or fails).

Are we sure the issue is a race between the two services, instead of an issue that ovs-configuration failed and then kubelet started anyway?

vrutkovs: Yes :( In [this OKD test](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/promote-release-openshift-okd-machine-os-content-e2e-aws-4.6/1296266834999250944) `ip-10-0-158-138` node started kubelet before ovs-configure has completed (see [worker logs](https://storage.googleapis.com/origin-ci-test/logs/promote-release-openshift-okd-machine-os-content-e2e-aws-4.6/1296266834999250944/artifacts/e2e-aws/nodes/workers-journal). That caused SDN container crashloop and failed the test.

Restarted OKD test twice and nodes come up as expected.

trozet: @vrutkovs looking at your log I think I see you are concerned about node ip-10-0-158-138. I think the problem is the way the logs are munged. If you look at:
https://gist.github.com/trozet/232688e0ce7bd853cb720ee65b0cc357#file-ovs-config-kubelet-worker-race-L42

We see there kubelet is started without a message before it saying:
Aug 20 02:45:05.478151 ip-10-0-146-26 systemd[1]: ovs-configuration.service: Succeeded.

However, if you look at the log messages with no ip hostname before them:
https://gist.github.com/trozet/232688e0ce7bd853cb720ee65b0cc357#file-ovs-config-kubelet-worker-race-L46

You can see that ovs-configuration failed right before kubelet was started:
Aug 20 02:47:28.524450 localhost.localdomain systemd[1]: ovs-configuration.service: Failed with result 'exit-code'.
Aug 20 02:47:29.363963 ip-10-0-158-138 systemd[1]: Starting Kubernetes Kubelet...

vrutkovs: Oh, I see, so perhaps `ovs-configuration` should depend on `systemd-hostnamed` instead?

vrutkovs: Seems `After` is not sufficient, see [worker logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2013/pull-ci-openshift-machine-config-operator-master-okd-e2e-aws/1297192628235800576/artifacts/e2e-aws/gather-extra/nodes/workers-journal):
```
Aug 22 16:02:27.090063 localhost.localdomain systemd[1]: ovs-configuration.service: Main process exited, code=exited, status=1/FAILURE
Aug 22 16:02:27.090224 localhost.localdomain systemd[1]: ovs-configuration.service: Failed with result 'exit-code'.                                                                            Aug 22 16:02:27.091858 localhost.localdomain systemd[1]: Failed to start Configures OVS with proper host networking configuration.
....
Aug 22 16:02:27.127077 localhost.localdomain systemd[1]: Starting Container Runtime Interface for OCI (CRI-O)...
```
and the node fails. Re-added `RequiredBy` to install section

cgwalters:
/approve
Looks good to me, just needs a BZ.

templates/common/_base/units/ovs-configuration.service.yaml
trozet: I don't think the bug description is accurate. ovs-configuration is not racing with kubelet. Kubelet is simply running after it. Requires will ensure that kubelet only starts if ovs-configuration was successful. Since kubelet did not require network-online.target before I didn't want to make that a requirement here. I'm not convinced it should be a require.

ovs-configuration can race with setting the hostname. However, I don't see how that is a problem. ovs-configuration doesn't rely on the hostname being set. Alternatively, sometimes DHCP can offer a hostname. I'm not sure how the current hostname service works, but if it ever looks to use a hostname offered by DHCP then we should maybe hold off on the hostname service until ovs-configuration is done.

@vrutkovs were you able to determine why the ovs-configuration script was failing in the first place?

vrutkovs: > ovs-configuration is not racing with kubelet

Kubelet may start before ovs-configuration completes successfully. `RequiredBy` ensures that kubelet would wait for it complete.

>  ovs-configuration can race with setting the hostname

Correct, that seems to make it crash. But the real issue is that kubelet start regardless of the ovs-configuration exit status

trozet: @vrutkovs in the logs you shared:
Aug 22 16:02:27.090224 localhost.localdomain systemd[1]: ovs-configuration.service: Failed with result 'exit-code'.
Aug 22 16:02:27.127077 localhost.localdomain systemd[1]: Starting Container Runtime Interface for OCI (CRI-O)...

CRIO is starting *after* ovs-configuration service. I'm still not seeing any issue here.

trozet:
There is no evidence part of this patch fixes anything, so holding this for now until the evidence is produced. Making kubelet depend on ovs-configuration is akin to making kubelet depend on networking-online target. If kubelet requires networking to start, kubelet should require network-online.target.

cgwalters:
> Making kubelet depend on ovs-configuration is akin to making kubelet depend on networking-online target.

It does: https://github.com/cri-o/cri-o/blob/9c637e0c1c102fc01ea6f37d3203bf0a30aebe3e/contrib/systemd/crio.service#L5

trozet:
> > Making kubelet depend on ovs-configuration is akin to making kubelet depend on networking-online target.
> 
> It does: https://github.com/cri-o/cri-o/blob/9c637e0c1c102fc01ea6f37d3203bf0a30aebe3e/contrib/systemd/crio.service#L5

@cgwalters That's crio (not kubelet), and its using After not Requires. Therefore it does not depend on network-online.target success, but it does ensure that CRIO **must run after** ovs-configuration.service.

I don't think when I checked kubelet that it had any Requirement on network-online.target. If it did this change would still be unnecessary as ovs-configuration.service is part of network-online.target.

templates/common/_base/units/ovs-configuration.service.yaml
vrutkovs: Its starting after first run of ovs-configuration, but its doesn't wait for it to complete successfully.

vrutkovs:
>There is no evidence part of this patch fixes anything

OKD jobs are no longer failing w/ "pod ovn-node... is crashlooping" - and its pretty common in [current AWS tests](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/release-openshift-okd-installer-e2e-aws-4.6)

dcbw:
Why are we doing this in the ovs configuration script, when this has nothing to do with OVS/OVN itself?

This is a *host* issue, not a networking issue. Kubelet/CRIO require the hostname to be set before they can operate. That has nothing to do with networking, except that networking provides the hostname. The thing that requires the hostname (kubelet/crio) must block, not networking.

vrutkovs:
>The thing that requires the hostname (kubelet/crio) must block, not networking.

ovs-configuration crashes before hostname is set and that makes ovn-node pods crashloop even after ovs-configuration has passed

dcbw:
> ovs-configuration crashes before hostname is set and that makes ovn-node pods crashloop even after ovs-configuration has passed

That's cool, and  Requires=node-valid-hostname.service is ok. But making kubelet/crio depned on the service isn't correct. kubelet/crio should depend on a valid hostname, since that's something we require in OpenShift *regardless* of network plugin.

vrutkovs:
@trozet @dcbw removed RequiredBy for kubelet, I think its a different bug. I think we all agree that `Requires=node-valid-hostname.service` is necessary, right?

dcbw:
> @trozet @dcbw removed RequiredBy for kubelet, I think its a different bug (as kubelet requires a hostname, not OVS bridge). I think we all agree that `Requires=node-valid-hostname.service` is necessary, right?

Probably? but I don't see anywhere in ovs-configuration that cares about hostname, so is the problem really in ovn-kube? (and you've pointed out at least one)

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -7,6 +7,7 @@ contents: |
   ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
+  Requires=node-valid-hostname.service
   Wants=NetworkManager-wait-online.service
```
abhat, line 11: should we also change `NetworkManager-wait-online.service` to `Requires` instead of `Wants`

trozet: No, if networkamnager-wait-online fails it can be because other NICs failed to DHCP. That does not mean we do not have a default gateway out of the host, and thus we should still run.

cybertron:
I think the ordering of this was backward to begin with, and that's what causes the race. ovs-configuration doesn't actually depend on the hostname being set, but the changes made in ovs-configuration _can_ affect the hostname, which makes everything thereafter racy. I have a more detailed breakdown of what I'm seeing in https://github.com/openshift/machine-config-operator/pull/2058 

I don't believe this would fix the behavior I'm seeing because the services were already running in the order specified.

PR #2016, opened 2020-08-20
[release-4.5] Bug 1870665: Refactor mastersSchedulable reconciliation checks

damemi:
Backport of https://github.com/openshift/machine-config-operator/pull/1878

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -40,6 +40,12 @@ import (
 )
 
 const (
+	// MasterLabel defines the label associated with master node. The master taint uses the same label as taint's key
+	MasterLabel = "node-role.kubernetes.io/master"
+
+	// WorkerLabel defines the label associated with worker node.
+	WorkerLabel = "node-role.kubernetes.io/worker"
+
 	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
 	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
 	// a machineconfig pool is going to be requeued:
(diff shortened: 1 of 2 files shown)


damemi:
/cc @ravisantoshgudimetla @runcom 
ptal, these are the same changes that set `mastersSchedulable` as the source of truth for unschedulable taints as were added in https://github.com/openshift/machine-config-operator/pull/1878

kikisdeliveryservice:
Bugzillas look fine btw the 4.6 bz just needs to be verified first before this can be merged.

Deferring to runcom on the final approval.

damemi:
@kikisdeliveryservice this will depend on the 4.6 bug (https://bugzilla.redhat.com/show_bug.cgi?id=1851226) being verified by QA. There is some discussion in that BZ on whether the fix here is enough to verify the bug or if more of a root cause should be identified

damemi:
@kikisdeliveryservice sorry, didn't see your new comment after the refresh :)

cgwalters (commented):
I'm uncertain we actually *need* this in 4.5.  The core bug was drift between the default kubelet config omitting the taint on bare metal, which is now fixed by d044c74ea4b9900c269ee8de8131ed49ba6fedc8 -  Anyone affected by this has a simple workaround: add the taint manually for the 3 master nodes.

There may even be "compact" 3 node bare metal clusters out there that this unintentionally *breaks* if they happened to be relying on us omitting that?

I'm not fully opposed to landing this in 4.5 but just want to analyze the risk/benefits a bit more carefully.

damemi:
@cgwalters +1, I was working on this on a different BZ in parallel with the solution you linked (I was unaware of that work at the time). If we're confident that the root cause has been addressed then this change is probably not necessary, and I'm fine with closing it if that's the consensus.

eparis:
/hold
and I'm marking this as awaiting more evidence. If you guys want to close this or decide the right path and remove the hold,either is fine! Thanks all!

sdodson:
(patch manager) It's been ~ 6 weeks since a hold was put on this, are we moving forward with this or not?

damemi:
Sorry this got lost, based on how long it's been without any more evidence that this is necessary I'll close it. This has been referenced in the related bugs that if any evidence does arise we can always revive it. Thanks for the patience everyone!
/close

PR #2017, opened 2020-08-20
[baremetal] Create and enable overlay mount point for NetworkManager

bcrochet:
This does 4 things for the baremetal platform:

1. Configures NetworkManager to look for it's keyfiles in /etc/NetworkManager/system-connections-merged
2. Creates that directory
3. Adds two temporary directories for systemd-tmpfiles to process. These are the storage and work locations for the overlay.
4. Mount an overlay filesystem at /etc/NetworkManager/system-connections-merged that uses /etc/NetworkManager/system-connections as the lower filesystem, and the temporary directory as the upper.

Any changes made outside of a MachineConfig will result in it being ephemeral. This maintains the idea that MCO will own the configuration of a machine before kubelet runs. MCO changes would go into /etc/NetworkManager/system-connections.

**- What I did**

**- How to verify it**

Mount an overlay directory to make NetworkManager changes ephemeral.

### DIFF
--- a/templates/common/baremetal/files/baremetal-kni-tmpfiles.yaml
+++ b/templates/common/baremetal/files/baremetal-kni-tmpfiles.yaml
@@ -1,0 +1,8 @@
+mode: 0644
+path: "/etc/tmpfiles.d/kni.conf"
+contents:
+  inline: |
+    D /run/nm-system-connections 0755 root root - -
+    D /run/nm-system-connections-work 0755 root root - -
+    d /etc/NetworkManager/system-connections-merged 0755 root root - -
+
--- a/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
+++ b/templates/common/baremetal/files/baremetal-NetworkManager-kni-conf.yaml
@@ -7,3 +7,5 @@ contents:
     [connection]
     ipv6.dhcp-duid=ll
     ipv6.dhcp-iaid=mac
+    [keyfile]
+    path=/etc/NetworkManager/system-connections-merged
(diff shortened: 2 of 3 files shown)


cybertron:
MCO looks unhappy with this: 

lstat /etc/NetworkManager/system-connections-merged/.dummy: no such file or directory

bcrochet:
/retitle [baremetal] Create and enable overlay mount point for NetworkManager

templates/common/baremetal/files/baremetal-kni-tmpfiles.yaml
```diff
@@ -0,0 +1,8 @@
+mode: 0644
+path: "/etc/tmpfiles.d/kni.conf"
+contents:
+  inline: |
+    D /tmp/nm-system-connections 0755 root root - -
+    D /tmp/nm-system-connections-work 0755 root root - -
```
cgwalters, line 6: Avoid hardcoded paths in `/tmp` - see <https://cwe.mitre.org/data/definitions/377.html>. A better choice for this is `/run` that only privileged components can write into.

cgwalters: And that gets into the fact that IMO really this should be a NetworkManager feature - it should by default look for and parse `/run/NetworkManager/system-connections` too.  We'd just be "backporting" that feature or so.

Did you run this by the NM team?

bcrochet: NetworkManager will eventually have the ability for ephemeral connections. That will not come until RHEL 8.3. This is a stop-gap until we get there. But thanks for the advice on /run. I will incorporate that.

cybertron:
/lgtm

Looks like all of the comments have been addressed and it's passing ci.

sinnykumari:
lgtm, implements what has been described in https://github.com/openshift/enhancements/pull/399 in Option C
/approve

PR #2021, opened 2020-08-24
Bug 1854009: remove err cardinality from mcd_kubelet_state metrics

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1854009

Using the err message as a label introduces a bug in resetting the metrics to 0 when the kubelet health returns to normal.

Test by ssh'nig into node, stopping kubelet, waiting for alert to fire, restart kubelet, ensure the alert goes away and the `mcd_kubelet_state` metric returns to 0.

@runcom

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -138,9 +138,8 @@ const (
 	// also retrieve the pending config after a reboot
 	pendingStateMessageID = "machine-config-daemon-pending-state"
 
-	kubeletHealthzPollingInterval  = 30 * time.Second
-	kubeletHealthzTimeout          = 30 * time.Second
-	kubeletHealthzFailureThreshold = 3
+	kubeletHealthzPollingInterval = 30 * time.Second
+	kubeletHealthzTimeout         = 30 * time.Second
 
 	// updateDelay is the baseline speed at which we react to changes.  We don't
 	// need to react in milliseconds as any change would involve rebooting the node.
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -692,19 +691,15 @@ func (dn *Daemon) runKubeletHealthzMonitor(stopCh <-chan struct{}, exitCh chan<-
 		case <-stopCh:
 			return
 		case <-time.After(kubeletHealthzPollingInterval):
-			if err := dn.getHealth(); err != nil {
-				glog.Warningf("Failed kubelet health check: %v", err)
+			err := dn.getHealth()
+			if err != nil {
 				failureCount++
-				KubeletHealthState.WithLabelValues(err.Error()).Set(float64(failureCount))
-				if failureCount >= kubeletHealthzFailureThreshold {
```
kikisdeliveryservice, line 699: do you not want to keep the kubeletHealthzFailureThreshold just to gate when an error is sent to the exitCh?

sjenning: we were logging it before on old line 696 and i don't see any side effect of just sending it to the channel for logging.  that way we don't log it twice when we are over the threshold.

kikisdeliveryservice: fair enough :+1:

pkg/daemon/metrics.go
```diff
@@ -49,11 +49,11 @@ var (
 		}, []string{"state", "reason"})
 
 	// KubeletHealthState logs kubelet health failures and tallys count
-	KubeletHealthState = prometheus.NewGaugeVec(
```
kikisdeliveryservice, line 52: this has to stay NewGaugeVec I think

as per the build errors

kikisdeliveryservice: ah going the other way works too :+1:

PR #2022, opened 2020-08-25
Bug 1867992: Support RHEL7 workers by removing 'jq' commands from ovs setup

JacobTanenbaum:
RHEL7 does not have 'jq' by default or a version of the ip command that supports the
"-j" option. To support RHEL7 worker nodes it is required to remove the unsupported actions
from the setup scripts

Signed-off-by: Jacob Tanenbaum <jtanenba@redhat.com>

Fixes: 1867992

**- What I did**

**- How to verify it**
Run the script in the yaml on a RHEL7 system with ovs installed and an updated network manager (so far not in an official release https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=30784728#)

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -15,14 +15,14 @@ contents:
     # find default interface
     while [ $counter -lt 12 ]; do
       # check ipv4
-      iface=$(ip -j route show default | jq -r '.[0].dev')
-      if [[ -n "$iface"  && "$iface" != "null" ]]; then
+      iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')
+      if [[ -n "$iface" ]]; then
         echo "IPv4 Default gateway interface found: ${iface}"
         break
       fi
       # check ipv6
-      iface=$(ip -6 -j route show default | jq -r '.[0].dev')
-      if [[ -n "$iface"  && "$iface" != "null" ]]; then
+      iface=$(ip -6 route show default | awk '{if ($4 == "dev") print $5; exit}')
+      if [[ -n "$iface" ]]; then
         echo "IPv6 Default gateway interface found: ${iface}"
         break
       fi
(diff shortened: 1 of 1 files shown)


JacobTanenbaum:
@trozet PTAL

cgwalters:
To repeat though we should really be writing this stuff in "not-bash" (we can use the MCO binary on the host, or we can pull a container image).

trozet:
> To repeat though we should really be writing this stuff in "not-bash" (we can use the MCO binary on the host, or we can pull a container image).

I totally agree, but I thought we had this discussion a few weeks ago and the conclusion was for now its not possible due to limitations of not being able to run in the host binary, and complications with trying to run it in a container.. Maybe we should have another discussion on this.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -15,13 +15,13 @@ contents:
...
-      iface=$(ip -j route show default | jq -r '.[0].dev')
+      iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')
       if [[ -n "$iface"  && "$iface" != "null" ]]; then
         echo "IPv4 Default gateway interface found: ${iface}"
         break
       fi
       # check ipv6
-      iface=$(ip -6 -j route show default | jq -r '.[0].dev')
+      iface=$(ip -6 route show default | awk '{if ($4 == "dev") print $5; exit}')
       if [[ -n "$iface"  && "$iface" != "null" ]]; then
```
trozet, line 25: dont need the null check anymore since we remove jq

cgwalters:
> not possible due to limitations of not being able to run in the host binary, and complications with trying to run it in a container.. 

There is `/run/bin/machine-config-daemon` on the host that is extracted by the MCD.  It'd require having systemd units start after the MCD pod has landed on the host; there might be some bootstrapping if the default interface is ovs and we need to configure that before pulling a container.  If we have to solve that then we could probably change things to also persist the previous MCD binary across updates and use the old version - this would introduce some [hysteresis](https://en.wikipedia.org/wiki/Hysteresis) though.

templates/common/_base/files/configure-ovs-network.yaml
JacobTanenbaum: done

trozet:
> > not possible due to limitations of not being able to run in the host binary, and complications with trying to run it in a container..
> 
> There is `/run/bin/machine-config-daemon` on the host that is extracted by the MCD. It'd require having systemd units start after the MCD pod has landed on the host; there might be some bootstrapping if the default interface is ovs and we need to configure that before pulling a container. If we have to solve that then we could probably change things to also persist the previous MCD binary across updates and use the old version - this would introduce some [hysteresis](https://en.wikipedia.org/wiki/Hysteresis) though.

Right now I remember you mentioned this, but I think we need to do this before at least kubelet is up. Maybe we can talk more on slack about how to approach this.

PR #2025, opened 2020-08-25
Bug 1872001: Stop using EtcdDiscoveryDomain in templates

cybertron:
Etcd is no longer using EtcdDiscoveryDomain and as a result the parameter should be removed. However, the baremetal and friends templates reference it to get the cluster domain.  This information is also available from the DNS structure, which is not currently wired into the templates.

This change adds the DNS struct to the resources in the controller config and switches the templates to get the cluster domain from there.


cybertron:
/hold

This is not complete and may not be working at all.

pkg/operator/sync.go
```diff
@@ -324,21 +324,18 @@ func getIgnitionHost(infraStatus *configv1.InfrastructureStatus) (string, error)
...
-			ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.BareMetal.APIServerInternalIP, securePortStr)
-		case configv1.OpenStackPlatformType:
-			ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.OpenStack.APIServerInternalIP, securePortStr)
-		case configv1.OvirtPlatformType:
-			ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.Ovirt.APIServerInternalIP, securePortStr)
-		case configv1.VSpherePlatformType:
-			if infraStatus.PlatformStatus.VSphere != nil && infraStatus.PlatformStatus.VSphere.APIServerInternalIP != "" {
-				ignitionHost = net.JoinHostPort(infraStatus.PlatformStatus.VSphere.APIServerInternalIP, securePortStr)
-			}
+	switch infraStatus.PlatformStatus.Type {
```
cgwalters, line 327: This looks like it's reverting f6b693805f4fa941a74096ccad7e393878014105 - do you have a reason for that?

retroflexer: @cgwalters see openshift/installer#4067.

templates/common/baremetal/files/baremetal-coredns-corefile.yaml
```diff
@@ -5,7 +5,7 @@ contents:
     . {
         errors
         health :18080
-        mdns {{ .Infra.Status.EtcdDiscoveryDomain }} 0 {{`{{.Cluster.Name}}`}} {{`{{.NonVirtualIP}}`}}
+        mdns {{ .ControllerConfig.DNS.Spec.BaseDomain }} 0 {{`{{.Cluster.Name}}`}} {{`{{.NonVirtualIP}}`}}
```
retroflexer, line 8: There are more places in this file that uses `.Infra.Status.EtcdDiscoveryDomain.EtcdDiscoveryDomain`. They should all be replaced with `.ControllerConfig.DNS.Spec.BaseDomain`

pkg/operator/sync.go
cybertron: Argh, no. I must have had an old copy of the file open.

templates/common/baremetal/files/baremetal-coredns-corefile.yaml
celebdor: Agreed

cybertron:
/hold cancel

I need to fix the unit tests, but otherwise I believe this is ready for review.

templates/common/baremetal/files/baremetal-coredns-corefile.yaml
cybertron: This should all be taken care of now.

mandre:
LGTM for OpenStack :+1:

cmd/machine-config-operator/bootstrap.go
```diff
@@ -75,6 +76,7 @@ func init() {
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.networkConfigFile, "network-config-file", "/assets/manifests/cluster-network-02-config.yml", "File containing network.config.openshift.io manifest.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.cloudConfigFile, "cloud-config-file", "", "File containing the config map that contains the cloud config for cloudprovider.")
 	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.proxyConfigFile, "proxy-config-file", "/assets/manifests/cluster-proxy-01-config.yaml", "File containing proxy.config.openshift.io manifest.")
+	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.dnsConfigFile, "dns-config-file", "/assets/manifests/cluster-dns-02-config.yml", "File containing dns.config.openshift.io manifest.")
```
cgwalters, line 79: OK and I just looked, this has existed since https://github.com/openshift/installer/commit/dd6d6d5d4de207cdb00b40ff4af332ff00646808 so that's clearly fine.

cgwalters (approved):
LGTM over all; will defer to mco team for final lgtm.  This is a nontrivial patch that doesn't *directly* fix anything, it's just completing the removal of etcd from the MCO right?  I think that would be good to get done fully for 4.6, but just worth noting.

cybertron:
> LGTM over all; will defer to mco team for final lgtm. This is a nontrivial patch that doesn't _directly_ fix anything, it's just completing the removal of etcd from the MCO right? I think that would be good to get done fully for 4.6, but just worth noting.

Yes, I was asked to fix this because it was blocking the installer team from removing EtcdDiscoveryDomain, which was work targeted for 4.6. We could defer it, but that would require them to defer their part as well.

cybertron:
So it looks like ovirt hasn't passed on this repo in a couple of weeks, and vsphere is red for over a month. ovirt at least appears to have successfully rendered the coredns templates that use this new field. The logs show it is generating records correctly. vsphere is harder to tell. mdns-publisher has errors in its log that suggest it isn't able to even publish records for coredns. coredns did start up successfully though, so it's likely that its config was correctly rendered too.

Given that, and the fact that the template changes were scripted, it's highly unlikely that this broke one platform but not the others. I think it should be good to go.

/cc @kikisdeliveryservice

cybertron:
/test e2e-ovirt

I see a few recent passes of this job, so let's give it another shot.

cybertron:
/test e2e-ovirt

Failure to create release image, which seems unrelated.

kikisdeliveryservice (commented):
@cgwalters doesn't this need to also update https://github.com/openshift/machine-config-operator/blob/bec98e9598b6560b8e6cdafddee92559e2abda7b/lib/resourcemerge/machineconfig.go#L68

cgwalters:
> @cgwalters doesn't this need to also update 

Looks like it should to me yep

cybertron:
/hold

Okay, I'll get that changed too.

cybertron:
/hold cancel

Okay, I added a merge implementation based on the infra merge. I'm not exactly sure how that all works so let me know if it's not right.

lib/resourcemerge/machineconfig.go
```diff
@@ -100,5 +100,10 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
 		*modified = true
 	}
 
+	if existing.DNS != nil && required.DNS != nil && !equality.Semantic.DeepEqual(existing.DNS, required.DNS) {
```
kikisdeliveryservice, line 103: since DNS is *:
if existing.DNS != nil && required.DNS != nil {
    !equality.Semantic.DeepEqual(existing.DNS, required.DNS) {
}

kikisdeliveryservice (dismissed):

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: unless someone can confirm the order of the &&& eval?

retroflexer: Yes. This is typical from the days of good old `C` and `C++`. 
You can see code like this in the Unix OS code:

```
if (ptr != nil && ptr->data == blah) {
}
```

cybertron: Yeah, and the short-circuit logic is relied upon by some of the other checks in this function too.

kikisdeliveryservice: thanks just double checking!

PR #1933, opened 2020-07-17
 Bug 1868117: cleanup: remove legacy etcd bits

hexfusion:
This PR removes various etcd related assets and functions that are no longer used. Once these changes merge we will followup with the installer removing the related etcd flags against MCO render command.

Then to close this effort out one for PR will be needed to remove kube-client-agent from manifests. The reason for this is we need to remove the references to kube-client-agent before we remove reference to the image or installer will fail.

PR #2027, opened 2020-08-26
bug 1872802: revert of Bug 1868117: cleanup: remove legacy etcd bits

deads2k:
revert https://github.com/openshift/machine-config-operator/pull/1933 which appears to break upgrade 4.5 to 4.6

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -51,8 +51,8 @@ var (
 
 func init() {
 	rootCmd.AddCommand(bootstrapCmd)
-	bootstrapCmd.PersistentFlags().StringVar(&bootstrapOpts.etcdCAFile, "etcd-ca", "", "path to etcd CA certificate")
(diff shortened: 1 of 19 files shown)


kikisdeliveryservice:
cc: @hexfusion

cgwalters:
That's unfortunate :cry: 
/approve

Hm, I guess we don't have a job hooked up to repos for 4.5 upgrades yet.

kikisdeliveryservice:
@deads2k do we have any data points showing that this broke upgrades?  can we get a bz with a summary laying this out?

deads2k:
one better fix that isn't yet proven here: https://github.com/openshift/cluster-etcd-operator/pull/429

deads2k:
https://bugzilla.redhat.com/show_bug.cgi?id=1872802

kikisdeliveryservice:
Asked in https://github.com/openshift/cluster-etcd-operator/pull/429 but are we trying that one in lieu of this PR? Or are we merging both?

kikisdeliveryservice:
Looks like after merging the other PR upgrades were successful see: https://github.com/openshift/cluster-etcd-operator/pull/429#issuecomment-681163157

So I think we don't need this PR anymore? But I'll let @deads2k / @hexfusion give the official word.

deads2k:
upgrades are green now

/close

PR #2029, opened 2020-08-27
Bug 1861026: daemon: perform other rpm-ostree operations after OS rebase

sinnykumari:
We perform operations like rt-kernel switch from packages
available in the latest machine-os-content to which we are going
to rebase OS. It makes sense to overlay additional
packages after performing OS rebase. This also prevents issues
like missing dependencies which may be now already part of BaseOS.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1861026

### DIFF
--- a/cmd/machine-config-daemon/pivot.go
+++ b/cmd/machine-config-daemon/pivot.go
@@ -22,12 +22,12 @@ import (
 
 // flag storage
 var keep bool
-var reboot bool
 var fromEtcPullSpec bool
 
 const (
-	etcPivotFile       = "/etc/pivot/image-pullspec"
-	runPivotRebootFile = "/run/pivot/reboot-needed"
+	// etcPivotFile is used for 4.1 bootimages and is how the MCD
+	// currently communicated with this service.
+	etcPivotFile = "/etc/pivot/image-pullspec"
 	// File containing kernel arg changes for tuning
 	kernelTuningFile = "/etc/pivot/kernel-args"
 	cmdLineFile      = "/proc/cmdline"
@@ -51,7 +51,6 @@ var pivotCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(pivotCmd)
 	pivotCmd.PersistentFlags().BoolVarP(&keep, "keep", "k", false, "Do not remove container image")
-	pivotCmd.PersistentFlags().BoolVarP(&reboot, "reboot", "r", false, "Reboot if changed")
 	pivotCmd.PersistentFlags().BoolVarP(&fromEtcPullSpec, "from-etc-pullspec", "P", false, "Parse /etc/pivot/image-pullspec")
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 }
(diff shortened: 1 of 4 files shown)


sinnykumari:
Going to test it locally to ensure it fixes the bug
/hold

pkg/daemon/update.go
```diff
@@ -353,34 +383,20 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
-	if err := dn.switchKernel(oldConfig, newConfig); err != nil {
+	if err := dn.applyOSChanges(oldConfig, newConfig); err != nil {
 		return err
 	}
 
 	defer func() {
 		if retErr != nil {
-			if err := dn.switchKernel(newConfig, oldConfig); err != nil {
-				retErr = errors.Wrapf(retErr, "error rolling back Real time Kernel %v", err)
+			if err := dn.applyOSChanges(newConfig, oldConfig); err != nil {
```
cgwalters, line 392: Not strictly related to this change but we're now grouping the "defer rollback" of the kargs/kernel type switch - I was going to comment about how this might cause us to incorrectly roll kargs back if we failed at the kernel type switch but then I realized:  

The only defer rollback we need is the `removePendingDeployment()` - that reverts everything.

(I think that defer will run last and hence win, so this code is fine as is)

pkg/daemon/update.go
```diff
@@ -109,13 +109,43 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
 	}
 }
 
-// updateOSAndReboot is the last step in an update(), and it can also
-// be called as a special case for the "bootstrap pivot".
-func (dn *Daemon) updateOSAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error) {
+// Remove pending deployment on OSTree based system
+func removePendingDeployment() error {
+	args := []string{"cleanup", "-p"}
+	_, err := runGetOut("rpm-ostree", args...)
```
cgwalters, line 115:
```suggestion
	_, err := runGetOut("rpm-ostree", "cleanup", "-p")
```

sinnykumari:
looks fine locally, I miss https://github.com/openshift/machine-config-operator/pull/1766 and at the same time I am happy that at least we have this in 4.6+ 

Should be good to go.

/hold cancel

sinnykumari:
@ashcrow This doesn't have corresponding 4.6 bug because we already fixed this as part of extensions epic https://github.com/openshift/machine-config-operator/pull/1941/

pkg/daemon/update.go
sinnykumari: fixed

ashcrow:
> @ashcrow This doesn't have corresponding 4.6 bug because we already fixed this as part of extensions epic #1941

The referenced bug is set to target 4.6. Should it target 4.5 instead? I'm happy to override the check if 4.6 was fixed due to a change in how things worked in extensions.

ashcrow:
Overriding BZ check since 4.6 was fixed due to underlying, non bug related changes.

cmd/machine-config-daemon/pivot.go
```diff
@@ -22,12 +22,12 @@ import (
...
 // flag storage
 var keep bool
-var reboot bool
 var fromEtcPullSpec bool
 
 const (
-	etcPivotFile       = "/etc/pivot/image-pullspec"
-	runPivotRebootFile = "/run/pivot/reboot-needed"
+	// etcPivotFile is used for 4.1 bootimages and is how the MCD
+	// currently communicated with this service.
```
ashcrow, line 29: nit: Unclear as to if this means now or previously. Currently would indicate it's in use now while communicated would indicate the past.

ashcrow: It was noted this is a direct pull in from https://github.com/openshift/machine-config-operator/pull/1809/files#r442404093

sinnykumari:
tests are passing, can we have lgtm :)

cgwalters:
One note about this: Anyone affected by this will need to apply this update *before* they try an OS update.  From this comment:
> @Sinny, any plans to backport it into 4.4? I see the same issue during minor upgrades of OCP 4.4.5 -> 4.4.17 (node with RT kernel becomes degraded)

Since 4.4.17 has already shipped this requires manual intervention:

- Create a custom release image with new MCO from this patch based on e.g. 4.4.5
- Upgrade to custom
- Upgrade to 4.4.18 or a new release that still has this patch

PR #2030, opened 2020-08-27
Bug 1872806: [baremetal] Wait for NM dhcp nameserver population

celebdor:
In some cases, the dhcp event may come before resolv.conf has nameserver
entries. In these cases, we'll wait.


templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -22,10 +22,20 @@ contents:
...
-        up|down|dhcp4-change|dhcp6-change)
-        logger -s "NM resolv-prepender triggered by ${1} ${2}."
+        up|dhcp4-change|dhcp6-change)
+        >&2 echo "NM resolv-prepender triggered by ${1} ${2}."
 
         # Ensure resolv.conf exists before we try to run podman
+        # In DHCP connections, the resolv.conf content may be late, thus we wait for nameservers
+        timeout 45s /bin/bash <<EOF
+            if [[ "$STATUS" == dhcp* ]]; then
+                >&2 echo  "NM resolv-prepender: Checking for nameservers" /var/run/NetworkManager/resolv.conf
```
cybertron, line 32: Nit: This line is formatted/worded a bit oddly. Might be better as:

\>&2 echo  "NM resolv-prepender: Checking for nameservers in /var/run/NetworkManager/resolv.conf"

celebdor:
/hold

I will address Ben's comment

cybertron:
/lgtm
/retest

This is a baremetal-specific file, it didn't break aws. We don't need to block on it, but let's see if we can get a green test run.

mandre:
Does it need to be ported to other on-prem platforms?

celebdor:
@mandre that's a good question. In multiple interface and dual stack deployments it could be necessary

PR #2031, opened 2020-08-27
Bug 1851540: OpenStack: set hostnames on early stages of machine provisioning

Fedosin:
To prevent possible race conditions with ovn-kubernetes we need to set hostnames before network-online.target and node-valid-hostname.service

### DIFF
--- a/templates/common/openstack/units/afterburn-hostname.service.yaml
+++ b/templates/common/openstack/units/afterburn-hostname.service.yaml
@@ -3,11 +3,11 @@ enabled: true
 contents: |
   [Unit]
   Description=Afterburn Hostname
-  Before=kubelet.service
+  Before=node-valid-hostname.service
 
   [Service]
   ExecStart=/usr/bin/afterburn --provider openstack-metadata --hostname=/etc/hostname
   Type=oneshot
 
   [Install]
-  WantedBy=multi-user.target
+  WantedBy=network-online.target


mandre:
Could you add OpenStack in the commit title?

templates/common/openstack/units/afterburn-hostname.service.yaml
```diff
@@ -3,7 +3,7 @@ enabled: true
 contents: |
   [Unit]
   Description=Afterburn Hostname
-  Before=kubelet.service
+  Before=network-online.target kubelet.service crio.service
```
mandre, line 6: If afterburn is getting it's info from nova-metadata service, shouldn't it have it's networking stack ready?

Fedosin: yeah, for sure. but afterburn can also get these values from the config drive. in addition to this PR I want to restore my patches for the config drive in the installer.

cgwalters: We also tear down the network before exiting the initramfs - this allows us to support configuring the network via Ignition.

cgwalters:
Potentially afterburn upstream should ship this as a unit file, just disabled by default.  Then the MCO would just enable it.

cgwalters:
cc @lucab

lucab:
@darkmuggle has been working on openshift side to make sure that the whole "hostnames are hard" topic is covered in a homogeneous way. I think this bypasses that, which seems worrisome, but I don't know all the details.

From my side, in general we do prefer to have the hostname coming from DHCP, so this may be something to tweak in the OpenStack provisioner to make sure the infrastructure is providing that. For those platforms where the DHCP offers does not carry the hostname option (which I think is not the case for OpenStack), we do set it statically via Afterburn on first boot [in initramfs](https://github.com/coreos/afterburn/blob/master/dracut/30afterburn/afterburn-hostname.service) to avoid this kind of races.

templates/common/openstack/units/afterburn-hostname.service.yaml
darkmuggle: Since multi-user.target wants `network-online.target` and that requires networking to be up [1], changing this to be
```
[Unit]
Before=node-valid-hostname.service
```

And 
```
[Install]
```
Would accomplish the goal and work with the "hardnames are hard work." 

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/node-valid-hostname.service.yaml

PR #2032, opened 2020-08-27
ovs/ovn: sync systemd service name with script name

dcbw:
Makes grepping for issues show all relevant hits, instead of
needing two different greps, one for service and one for the
script itself.

@vrutkovs @trozet


cgwalters:
Sure; we can call it a bugfix for the previous PR.
/approve

kikisdeliveryservice:
Does this have a BZ?

dcbw:
No, and it'll take a CNO change too. So I'll drop it for now.

PR #2033, opened 2020-08-27
Bug 1881208: add alert for system.slice rss going over system-reserved

sjenning:
@smarterclayton @mrunalp @nee1esh @derekwaynecarr 

xref https://issues.redhat.com/browse/OCPNODE-402

### DIFF
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -43,3 +43,12 @@ spec:
             severity: warning
           annotations:
             message: "Kubelet health failure threshold reached"
+    - name: system-memory-exceeds-reservation
+      rules:
+        - alert: SystemMemoryExceedsReservation
+          expr: |
+            sum by (node) (container_memory_rss{id="/system.slice"}) > ((sum by (node) (kube_node_status_capacity{resource="memory"} - kube_node_status_allocatable{resource="memory"})) * 0.9)
+          labels:
+            severity: warning
+          annotations:
+            message: "System memory usage of {{ $value | humanize }} on {{ $labels.node }} exceeds 90% of the reservation. Reserved memory ensures system processes can function even when the node is fully allocated and protects against workload out of memory events impacting the proper functioning of the node. The reservation may be increased (https://docs.openshift.com/container-platform/latest/nodes/nodes/nodes-nodes-managing.html) when running nodes with high numbers of pods."


mrunalp:
I think we want to get a number that accounts for pages that we aren't able to evict from cri-o on drop caches. Also, the alert could be at 90 -95% of the reservation.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -43,3 +43,12 @@ spec:
             severity: warning
           annotations:
             message: "Kubelet health failure threshold reached"
+    - name: system-memory-exceeds-reservation
+      rules:
+        - alert: SystemMemoryExceedsReservation
+          expr: |
+            container_memory_rss{id="/system.slice"} > 1073741824
```
smarterclayton, line 50: Hrm, this needs to come from the reservation the system makes, which is exposed as a metric

smarterclayton: constant should be kube_node_status_capacity{... mem} - kube_node_status_allocatable{... mem}, and you need to figure out how to make that number show up in the annotation somehow

smarterclayton: is there anything on a node outside of system and kubepods?  should the first value be `container_memory_rss{id="/"} - container_memory_rss{id="/kubepods.slice"} > (capacity - allocatable)`

sjenning: all the kernel threads are in `/`. no userspace process though.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -43,3 +43,12 @@ spec:
...
             message: "Kubelet health failure threshold reached"
+    - name: system-memory-exceeds-reservation
+      rules:
+        - alert: SystemMemoryExceedsReservation
+          expr: |
+            sum by (node) (container_memory_rss{id="/system.slice"}) > ((sum by (node) (kube_node_status_capacity{resource="memory"} - kube_node_status_allocatable{resource="memory"})) * 0.9)
+          labels:
+            severity: warning
+          annotations:
+            message: "System memory usage of {{ $value | humanize }} on {{ $labels.node }} exceeds 90% of the reservation"
```
smarterclayton, line 54: Messages should be acitonable.  So you need to recommend next steps.  I would suggest playing around with something like the following additions:

> Reserved memory ensures system processes can function even when the node is fully allocated and protects against workload out of memory events impacting the proper functioning of the node. The reservation may be increased (link to docs) when running nodes with high numbers of pods.

Or similar

sjenning: @smarterclayton better now?

smarterclayton:
/lgtm

Attach the bug and this is gtg for 4.6

kikisdeliveryservice (approved):

PR #2034, opened 2020-08-27
 Bug 1873321: Check the version hash so we resync on an upgrade

umohnani8:
Backport of https://github.com/openshift/machine-config-operator/pull/1984

Fixes #1873321 (https://bugzilla.redhat.com/show_bug.cgi?id=1873321)

**- What I did**
Add a check for the version hash to ensure that we resync
and generate a new controller version during an upgrade.

Update the way we append the status for a ctrcfg as well -
ensure that we don't have mutliple of the same status recorded.

Add a check to avoid adding the same mc name more than once in
the list of finalizers. When we update an existing ctrcfg, the
generation number changes causing a resync and we don't need to add
the same name to the list of finalizers every time this happens.
And remove the check for cri-o drop-in files.

**- How to verify it**
1. Create a 4.5 cluster
2. Create a ctrcfg CR on it and apply it to both master and worker pools.
e.g:
```
kind: ContainerRuntimeConfig
metadata:
  name: set-log
spec:
  machineConfigPoolSelector:
    matchLabels:
      custom-crio: set-log
  containerRuntimeConfig:
    logLevel: debug
```
3. Upgrade the cluster to a version that has the changes from this PR. The upgrade should be successful with the ctrcfg CR changes intact. Check the controller version for the MCs, it should have been updated to the new one.

Check the generated controller version to ensure that we resync and create a new one on upgrades.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -218,3 +218,13 @@ func IgnParseWrapper(rawIgn []byte) (ignconfig interface{}, err error) {
 	}
 	return ign2types.Config{}, errors.Errorf("parsing Ignition config failed with error: %v\nReport: %v", err, rpt)
 }
+
+// InSlice search for an element in slice and return true if found, otherwise return false
+func InSlice(elem string, slice []string) bool {
+	for _, k := range slice {
+		if k == elem {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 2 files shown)


umohnani8:
@kikisdeliveryservice @runcom @yuqi-zhang PTAL

yuqi-zhang (approved):
tested to work fine with upgrades
/lgtm

Issue #1619, opened 2020-04-06
enhance flow for custom machineconfigs for specific machinesets


cgwalters:
The CI team is trying to use new AWS m5d.xlarge instances which have two NVMe disks attached.  We crafted a [custom RAID partition machineconfig](https://github.com/openshift/release/pull/8102) to enable that.

We added this as part of the main `worker` pool - the MCD will fail to roll out the partitioning on the *existing* workers, but that's fine because the plan was to "roll" the worker pool.  Basically get the new MC in the pool, have new workers come online with that config, then scale down the old workers.

However, there are a few issues here.

First, this whole thing would obviously be a *lot* better if we had machineset-specific machineconfigs.  That would solve a bunch of races and be much more elegant.

What we're seeing right now is that one new m5d node went `OutOfDisk=true` because it was booted with just a 16G root volume from the *old* config.  That unschedulable node then blocks rollout of further changes.

I think we can unstick ourselves here by deleting that node and getting the MCO to roll out the new config.

cgwalters:
And the real problem here is that the MCS serves:

```
currConf := mp.Status.Configuration.Name
```

So new nodes will only get the last config that successfully rolled out, so this attempt deadlocks.
We really need machineset-specific MCs.

cgwalters:
Another idea: some sort of "force MCS to serve `mp.Spec.Configuration.Name`" config option?

cgwalters:
> First, this whole thing would obviously be a lot better if we had machineset-specific machineconfigs. 

This though gets into the "node identity" problem - see also https://github.com/openshift/machine-config-operator/pull/784
Which could be e.g. a secret value passed through user data and then provided to the MCS, which can link that to a machine object, and from there a machineset.

cgwalters:
If someone wants to change the machine type of their workers *today* I think it'd require a [custom `worker-new` pool](https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md) *and* also creating a custom user-data that does `s/worker/worker-new` that's referenced from the new machineset.

Alternatively, scaling down to 0 workers then back up should work, but obviously that's a lot more disruptive.

cgwalters:
The other alternative would be having something like a `spec.machineSet` in a pool - then the MCO could understand how to say "okay, the new config has partitioning changes, so let me destroy the machine objects and provide new configs to newly provisioned nodes".  That'd be the most elegant but would require tightest integration between MCO and machineAPI.

[...]

PR #2035, opened 2020-08-27
Bug 1873288: server: Target the spec configuration if we have at least one node

cgwalters:
The CI cluster hit an issue where a pull secret was broken, and
then we hit a deadlock because the MCO failed to drain nodes on
the old config, because other nodes on the old config couldn't
schedule the pod.

It just generally makes sense for new nodes to use the new config;
do so as long as at least one node has successfully joined the
cluster at that config.  This way we still avoid breaking
the cluster (and scaleup) with a bad config.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1873288

### DIFF
--- a/pkg/server/cluster_server.go
+++ b/pkg/server/cluster_server.go
@@ -65,7 +65,16 @@ func (cs *clusterServer) GetConfig(cr poolRequest) (*runtime.RawExtension, error
 		return nil, fmt.Errorf("could not fetch pool. err: %v", err)
 	}
 
-	currConf := mp.Status.Configuration.Name
+	// For new nodes, we roll out the latest if at least one node has successfully updated.
+	// This avoids deadlocks in situations where the old configuration broke somehow
+	// (e.g. pull secret expired)
+	// and also avoids provisioning a new node, only to update it not long thereafter.
+	var currConf string
+	if mp.Status.UpdatedMachineCount > 0 {
+		currConf = mp.Spec.Configuration.Name
+	} else {
+		currConf = mp.Status.Configuration.Name
+	}
 
 	mc, err := cs.machineClient.MachineConfigs().Get(context.TODO(), currConf, metav1.GetOptions{})
 	if err != nil {


cgwalters:
(only compile tested)

cgwalters:
Also see https://github.com/openshift/machine-config-operator/issues/1619 that touched on this a bit.

cgwalters:
It looks like actually in https://bugzilla.redhat.com/show_bug.cgi?id=1873288 we didn't have any nodes successfully on the new config, so this wouldn't have helped - but I think that's mostly bad luck - the new config *could* have rolled out but we just happened to pick a node that couldn't drain.

Perhaps the other alternative of a `newNodesUseSpec bool` value that an admin can flip on in these situations is simplest.

cgwalters:
Now this PR also would have greatly mitigated the issue we hit in https://github.com/openshift/machine-config-operator/pull/2167
I'd like to consider this for 4.7 (probably needs an e2e, a bit annoying to write but doable and we need the excuse to start testing our interaction with machineAPI)..

As is right now, scaling up *during* an upgrade is just painful and basically just makes things slower and worse because:

- node boots
- does the firstboot pivot
- joins the cluster
- node now needs to be targeted by the MCO for upgrade to the new config
- node will be drained, a new OS update applied and rebooted *again*

runcom:
This, related to the BZ just attached is very much welcome in 4.7 if we can fix the bz and have something that works in those scenarios

yuqi-zhang (approved):
I am in favour of this approach. A use case is e.g. for CI where we dynamically spin up and down new nodes, upgrades could be stalled much longer as all new nodes will have to update (and drain workloads which will take hours). Will do some manual testing

sinnykumari:
lgtm. Will wait for Jerry's test result.

alvaroaleman:
@cgwalters @yuqi-zhang any chance this could get backported to 4.6?

PR #2040, opened 2020-08-30
Bug 1866265: [baremetal] Stop Keepalived on bootstrap after bootstrap completed

yboaron:
When OPENSHIFT_INSTALL_PRESERVE_BOOTSTRAP deployment configuration parameter is set to true the
bootstrap node won't be destroyed and infrastructure components including Keepalived on the bootstrap continue to run.

With this PR the Keepalived process is stopped when a stop message is received in Keepalived control socket.

This change is important for platforms that run Keepalived in unicast mode (baremetal).

This PR should be merged after [this baremetal-runtimecfg PR](https://github.com/openshift/baremetal-runtimecfg/pull/92) is merged

### DIFF
--- a/manifests/baremetal/keepalived.yaml
+++ b/manifests/baremetal/keepalived.yaml
@@ -52,13 +52,23 @@ spec:
             /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
         fi
       }
+      stop_keepalived()
+      {
+        echo "Keepalived process stopped" >> /var/run/keepalived/stopped
+        if pid=$(pgrep -o keepalived); then
+            kill -s TERM "$pid"
+        fi
+      }
+
       msg_handler()
       {
         while read -r line; do
           echo "The client sent: $line" >&2
-          # currently only 'reload' msg is supported
+          # currently only 'reload' and 'stop' msgs are supported
           if [ "$line" = reload ]; then
               reload_keepalived
+          elif  [ "$line" = stop ]; then
+              stop_keepalived
           fi
         done
       }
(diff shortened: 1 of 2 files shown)


yboaron:
/hold 
until [baremetal-runtimecfg PR](https://github.com/openshift/baremetal-runtimecfg/pull/92) is merged

manifests/baremetal/keepalived.yaml
```diff
@@ -86,7 +97,7 @@ spec:
         - /bin/bash
         - -c
         - |
-          [[ -s /etc/keepalived/keepalived.conf ]] || \
+          [[ -s /etc/keepalived/stopped ]] || \
```
cybertron, line 100: Should this be an added check instead of replacing the keepalived.conf check? We still don't want to start liveness checks until after the config has been rendered.

manifests/baremetal/keepalived.yaml
```diff
@@ -52,20 +52,31 @@ spec:
             /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
         fi
       }
+      stop_keepalived()
+      {
+        echo "Keepalived process stopped" >> /etc/keepalived/stopped
```
cybertron, line 57: Not a big deal, but since the control socket is in /var/run/keepalived, maybe we put the stop file there too?

cybertron (commented):
A couple of thoughts inline. Overall this looks similar to what I'm doing on the masters though, so +1 on the approach.

manifests/baremetal/keepalived.yaml
yboaron: Yep, maybe it's better to keep also this check.
Will update
10x

yboaron:
/hold cancel 
 [baremetal-runtimecfg PR](https://github.com/openshift/baremetal-runtimecfg/pull/92)  got merged.

yuqi-zhang:
/test e2e-metal-ipi

Do we expect this to pass with this PR?

yboaron:
> /test e2e-metal-ipi
> 
> Do we expect this to pass with this PR?

Yep

yboaron:
Hmmm, this PR already pushed with #2107 , but I can still see diff between this PR and master code.

yboaron:
/hold
Code was merged with  #2107 ,  though I can still see diff between this PR and master code. meanwhile set hold

PR #2042, opened 2020-08-31
[WIP] Bug 1872238: oVirt: Introduce qemu-guest-agent container

eslutsky:
ocp VMs  needs to run oVirt QEMU agent container.

Bug: 1872238
Signed-off-by: Evgeny Slutsky <eslutsky@redhat.com>

### DIFF
--- a/manifests/ovirt/qemu-guest-agent.yaml
+++ b/manifests/ovirt/qemu-guest-agent.yaml
@@ -1,0 +1,55 @@
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: qemu-guest-agent
+  namespace: openshift-ovirt-infra
+  creationTimestamp:
+  deletionGracePeriodSeconds: 65
+  labels:
+    app: qemu-guest-agent
+spec:
+  volumes:
+  - name: dev
+    hostPath:
+      path: "/dev"
+  - name: virtio-ports
+    hostPath:
+      path: "/dev/virtio-ports"
+  - name: dbus
+    hostPath:
+      path: "/run/dbus"
+  - name: systemd
+    hostPath:
+      path: "/run/systemd"
+  - name: zoneinfo
+    hostPath:
+      path: "/usr/share/zoneinfo"
+
+  containers:
+  - name: qemu-guest-agent
+    securityContext:
+      privileged: true
+    image: registry.svc.ci.openshift.org/ovirt/qemu-guest-agent:4.2
+    resources:
+      requests:
+        cpu: 100m
+        memory: 200Mi
+    volumeMounts:
+    - name: dev
+      mountPath: "/dev"
+    - name: virtio-ports
+      mountPath: "/dev/virtio-ports"
+    - name: dbus
+      mountPath: "/run/dbus"
+    - name: systemd
+      mountPath: "/run/systemd"
+    - name: zoneinfo
+      mountPath: "/usr/share/zoneinfo"
+
+    terminationMessagePolicy: FallbackToLogsOnError
+  hostNetwork: true
+  tolerations:
+  - operator: Exists
+  priorityClassName: system-node-critical
+status: {}
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -40,6 +40,7 @@
 // manifests/ovirt/coredns.yaml
 // manifests/ovirt/keepalived.conf.tmpl
 // manifests/ovirt/keepalived.yaml
+// manifests/ovirt/qemu-guest-agent.yaml
 // manifests/userdata_secret.yaml
 // manifests/vsphere/coredns-corefile.tmpl
 // manifests/vsphere/coredns.yaml
@@ -2466,6 +2467,78 @@ func manifestsOvirtKeepalivedYaml() (*asset, error) {
 	return a, nil
 }
 
+var _manifestsOvirtQemuGuestAgentYaml = []byte(`---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: qemu-guest-agent
+  namespace: openshift-ovirt-infra
+  creationTimestamp:
+  deletionGracePeriodSeconds: 65
+  labels:
+    app: qemu-guest-agent
+spec:
+  volumes:
+  - name: dev
+    hostPath:
+      path: "/dev"
+  - name: virtio-ports
+    hostPath:
+      path: "/dev/virtio-ports"
+  - name: dbus
+    hostPath:
+      path: "/run/dbus"
+  - name: systemd
+    hostPath:
+      path: "/run/systemd"
+  - name: zoneinfo
+    hostPath:
+      path: "/usr/share/zoneinfo"      
+
+  containers:
+  - name: qemu-guest-agent
+    securityContext:
+      privileged: true
+    image: registry.svc.ci.openshift.org/ovirt/qemu-guest-agent:4.2
+    resources:
+      requests:
+        cpu: 100m
+        memory: 200Mi
+    volumeMounts:
+    - name: dev
+      mountPath: "/dev"
+    - name: virtio-ports
+      mountPath: "/dev/virtio-ports"
+    - name: dbus
+      mountPath: "/run/dbus"
+    - name: systemd
+      mountPath: "/run/systemd"
+    - name: zoneinfo
+      mountPath: "/usr/share/zoneinfo"
+
+    terminationMessagePolicy: FallbackToLogsOnError
+  hostNetwork: true
+  tolerations:
+  - operator: Exists
+  priorityClassName: system-node-critical
+status: {}
+`)
+
+func manifestsOvirtQemuGuestAgentYamlBytes() ([]byte, error) {
+	return _manifestsOvirtQemuGuestAgentYaml, nil
+}
+
+func manifestsOvirtQemuGuestAgentYaml() (*asset, error) {
+	bytes, err := manifestsOvirtQemuGuestAgentYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "manifests/ovirt/qemu-guest-agent.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
+	a := &asset{bytes: bytes, info: info}
+	return a, nil
+}
+
 var _manifestsUserdata_secretYaml = []byte(`apiVersion: v1
 kind: Secret
 metadata:
@@ -2892,6 +2965,7 @@ var _bindata = map[string]func() (*asset, error){
 	"manifests/ovirt/coredns.yaml":                                           manifestsOvirtCorednsYaml,
 	"manifests/ovirt/keepalived.conf.tmpl":                                   manifestsOvirtKeepalivedConfTmpl,
 	"manifests/ovirt/keepalived.yaml":                                        manifestsOvirtKeepalivedYaml,
+	"manifests/ovirt/qemu-guest-agent.yaml":                                  manifestsOvirtQemuGuestAgentYaml,
 	"manifests/userdata_secret.yaml":                                         manifestsUserdata_secretYaml,
 	"manifests/vsphere/coredns-corefile.tmpl":                                manifestsVsphereCorednsCorefileTmpl,
 	"manifests/vsphere/coredns.yaml":                                         manifestsVsphereCorednsYaml,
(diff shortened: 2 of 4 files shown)


cgwalters:
Hmm did you see https://github.com/coreos/afterburn/issues/458 ? We are actively working on implementing the protocol in Afterburn.

lucab:
> We are actively working on implementing the protocol in Afterburn.

Slight correction to the above: we are exploring whether we can re-use the oVirt protocol/virtio-port to signal that a node booted into initramfs (for the sake of not reinveinting a wheel if that protocol is already good).
I don't think it overlaps much with this PR, which instead wants to run the qemu-agent (not the oVirt one) in the real rootfs as a long-running podman service (not sure why not as a daemonset).

cgwalters:
Is there a reason this is running via podman instead of as a daemonset?

cgwalters:
Also are you really aiming this at 4.6?

templates/common/ovirt/units/qemu-guest-agent.service.yaml
```diff
@@ -0,0 +1,34 @@
...
+    --privileged \
+    --volume /dev:/dev \
+    --volume /dev/virtio-ports:/dev/virtio-ports \
+    --volume /etc:/etc \
+    --volume /run/dbus:/run/dbus \
+    --volume /run/systemd:/run/systemd  \
+    --volume /usr/share/zoneinfo:/usr/share/zoneinfo \
+    --net=host \
+    --pull missing \
+    docker://registry.svc.ci.openshift.org/ovirt/qemu-guest-agent:4.2
```
cgwalters, line 27: We cannot pull arbitrary container images with OpenShift 4 by default.  Anything we run by default must be part of the release image.

cgwalters: And once it's part of the release image, this would need to have its value substituted by templating, same as other containers the MCO references.

cgwalters:
/hold
The BZ you linked is about creating the container.  I think you need a separate BZ about shipping this with OpenShift by default.  And this needs more prep work, like being added to the release image.
And I'd like to see discussion about podman vs daemonset.

templates/common/ovirt/units/qemu-guest-agent.service.yaml
eslutsky: agreed, we build this container in ovirt NS just for testing purposes, we want to see it running and docume

eslutsky:
> /hold
> The BZ you linked is about creating the container. I think you need a separate BZ about shipping this with OpenShift by default. And this needs more prep work, like being added to the release image.
> And I'd like to see discussion about podman vs daemonset.

sure, we've run it quickly as podman service for evaluation only, just to see how it behaves against ovirt CI  (without any external dependencies).

eslutsky:
> Also are you really aiming this at 4.6?

no .. this is for 4.7

rgolangh:
> /hold
> The BZ you linked is about creating the container. I think you need a separate BZ about shipping this with OpenShift by default. And this needs more prep work, like being added to the release image.
> And I'd like to see discussion about podman vs daemonset.

- release image
Early in the process we had a discussion about that starting [here](https://bugzilla.redhat.com/show_bug.cgi?id=1764804#c25). This container image is platform-specific and is considered 'extra'. It falls in the same category as CSI drivers I guess.

- podman vs. daemonset
The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures. 
The main disadvantage is that we can't monitor its resources and it's a hidden process from OpenShift pov.

Its possible that having an IP as early as possible for debug purposes is not a that important because we have other means to get it (console motd?). Let me know what you think.

eslutsky:
@cgwalters , can we  a get a feedback on this ,
so we ene

> > /hold
> > The BZ you linked is about creating the container. I think you need a separate BZ about shipping this with OpenShift by default. And this needs more prep work, like being added to the release image.
> > And I'd like to see discussion about podman vs daemonset.
> 
> * release image
>   Early in the process we had a discussion about that starting [here](https://bugzilla.redhat.com/show_bug.cgi?id=1764804#c25). This container image is platform-specific and is considered 'extra'. It falls in the same category as CSI drivers I guess.
> * podman vs. daemonset
>   The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures.
>   The main disadvantage is that we can't monitor its resources and it's a hidden process from OpenShift pov.
> 
> Its possible that having an IP as early as possible for debug purposes is not a that important because we have other means to get it (console motd?). Let me know what you think.
@roy
by looking a

> > /hold
> > The BZ you linked is about creating the container. I think you need a separate BZ about shipping this with OpenShift by default. And this needs more prep work, like being added to the release image.
> > And I'd like to see discussion about podman vs daemonset.
> 
> * release image
>   Early in the process we had a discussion about that starting [here](https://bugzilla.redhat.com/show_bug.cgi?id=1764804#c25). This container image is platform-specific and is considered 'extra'. It falls in the same category as CSI drivers I guess.
> * podman vs. daemonset
>   The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures.
>   The main disadvantage is that we can't monitor its resources and it's a hidden process from OpenShift pov.
> 
> Its possible that having an IP as early as possible for debug purposes is not a that important because we have other means to get it (console motd?). Let me know what you think.

@rgolangh , I've checked  your CSI drivers reference , and it appears to be also be part of the release image: `
oc adm release info -a=/tmp/pull-secret.yaml --pullspecs  registry.svc.ci.openshift.org/ocp/release:4.7.0-0.nightly-2020-10-27-051128 | grep ovirt-csi                                                
  ovirt-csi-driver                               quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3d6d6d5a7b2bfb797cf521d9d188447584a53e9543bb3e2cb6ba8decb675c45c
  ovirt-csi-driver-operator                      quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ae8f9bbc986b38951b466415984661603dc55ecdb627040c2c9852e6312b6079`

cgwalters:
> podman vs. daemonset
> The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures.

I think the MCO should take care of this generically across platforms.  How would qemu-guest-agent help with that anyways?

Stated more strongly, exactly what functionality do you want from the agent?

We have some work on implementing boot checkins in afterburn here: https://github.com/coreos/afterburn/issues/458

It sounds like the guest agent has some protocol for the guest to report its IP address - what are you using that for?

thecreed:
> > podman vs. daemonset
> > The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures.
> 
> I think the MCO should take care of this generically across platforms. How would qemu-guest-agent help with that anyways?
> 
> Stated more strongly, exactly what functionality do you want from the agent?
> 
> We have some work on implementing boot checkins in afterburn here: [coreos/afterburn#458](https://github.com/coreos/afterburn/issues/458)
> 
> It sounds like the guest agent has some protocol for the guest to report its IP address - what are you using that for?

on each oVirt host we have a vdsm component which use libvirt  to communicate with the agent over the virtio serial port .
vdsm relay this information to the oVirt engine.

eslutsky:
> > podman vs. daemonset
> > The main advantage of systemd+podman is that its decoupled from kubelet and will allow log gathering in case kubelet or other k8s failures.
> 
> I think the MCO should take care of this generically across platforms. How would qemu-guest-agent help with that anyways?
> 
> Stated more strongly, exactly what functionality do you want from the agent?
> 
> We have some work on implementing boot checkins in afterburn here: [coreos/afterburn#458](https://github.com/coreos/afterburn/issues/458)
> 
> It sounds like the guest agent has some protocol for the guest to report its IP address - what are you using that for?

on each oVirt host we have a vdsm component which use libvirt  to communicate with the agent over the virtio serial port .
vdsm relay this information to the oVirt engine.
our IPI installer using oVirt API to query the engine for the VM IP address.
this info will be used by the  cluster-api-provider-ovirt (to retrieve VMs ip address) , and also for collecting installation logs  when the installation fails (no API available) [0]

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1810438

cgwalters:
There's of course actually *three* options for running this:

- via extensions/package layering: We ship `open-vm-tools` today baked into RHCOS today because vSphere refused to support containerzed agents and some of the things like live migration (fs freezing) are critical
- podman
- daemonset

They all have tradeoffs.  (I'd actually like to switch the `open-vm-tools` case to be an extension that we dynamically layer only on vSphere platforms so we don't look dumb by shipping it aws/gcp/bare metal/etc; anyways somewhat tangential)

If the oVirt team feels strongly that it'd significantly help things now to just add the guest agent into the set of extensions and change the MCO to do exactly the above (dynamically layer just on oVirt platforms) - well...okay.

But the thing is every time we do this it hinders our long term goals to get people to move to containers - and also to minimize the "agent footprint" on the host.

cgwalters:
> our IPI installer using oVirt API to query the engine for the VM IP address.
> this info will be used by the cluster-api-provider-ovirt (to retrieve VMs ip address) , 

On other IPI virt platforms (AWS/GCP/Azure/OpenStack) we get by without this.  Why does oVirt need it but those don't?  The hypervisor is in a position to know what DHCP address it served to an instance.

> , and also for collecting installation logs when the installation fails (no API available) [0]

Again I strongly think this one is a *generic* problem; see related discussion in https://github.com/coreos/ignition/issues/585

eslutsky:
> > our IPI installer using oVirt API to query the engine for the VM IP address.
> > this info will be used by the cluster-api-provider-ovirt (to retrieve VMs ip address) ,
> 
> On other IPI virt platforms (AWS/GCP/Azure/OpenStack) we get by without this. Why does oVirt need it but those don't? The hypervisor is in a position to know what DHCP address it served to an instance.
well, the hypervisor doesn't act as a DHCP server,
usually, the VMs get their IP from the external network DHCP server which is transparent to the hypervisor.
that's why the agent is required here.
OpenStack for example controls its DHCP allocations using neutron, 
in oVirt, we are delegating this task to an external DHCP service which makes oVirt unique in some way.



> > , and also for collecting installation logs when the installation fails (no API available) [0]
> 
> Again I strongly think this one is a _generic_ problem; see related discussion in [coreos/ignition#585](https://github.com/coreos/ignition/issues/585)

Gal-Zaidman:
> On other IPI virt platforms (AWS/GCP/Azure/OpenStack) we get by without this. Why does oVirt need it but those don't? The hypervisor is in a position to know what DHCP address it served to an instance.

oVirt needs the guest agent running inside each VM that it manages (ocp or not ocp).
The gust agent enables the oVirt Engine to manage the VM, show detail of the VM in the oVirt Engine UI, and from the perspective of OCP, it will help us get information on the VM by using the oVirt API (like IP address of the VM).

Currently, we need this container to be available for our RHCOS VMs those are the only VMs that doesn't have the guest agent.
So we need to understand:
1. How to get the container image into the release? 
2. What is the best way to run the container on the VM, we want it available and running at the earliest stage possible?

As you can see this PR is old and we have been dragging this for a while now, and we need it for 4.7.
We can setup a meeting talking about how to get this in, I think it will be faster

runcom:
shipping the daemonset would still be backing this within the MCO repo (which doesn't help us with code that the MCO team isn't an expert about). Hina is setting up a meeting to discuss this as I also think it'll help.

cgwalters:
The list above missed the 4th option, which is:

- Extend https://github.com/coreos/afterburn/ with the core functionality you need.  As I noted we already had some work on implementing boot time check in there; implementing the bit to send networking information might not be really hard, although that's obviously a notable increase in scope from basic checkins.

So for clarity, all the options:

1. via extensions/package layering
2. Extend afterburn
3. via podman
4. via daemonset

Now for option 3 - personally I am a bit skeptical that this needs to run before kubelet - we already have too many of these "special before kubelet" containers and they make things a lot more complicated. 

My preferred order looks like:

- Investigate difficulty of afterburn path, if easy: preferred
- daemonset
- extension

cgwalters:
Can you clarify, is the desired functionality in the `qemu-guest-agent` package or do we really need https://github.com/oVirt/ovirt-guest-agent ?  At least the former seems to be zero additional dependencies on the host which is nice.

nyoxi:
> Can you clarify, is the desired functionality in the `qemu-guest-agent` package or do we really need https://github.com/oVirt/ovirt-guest-agent ? At least the former seems to be zero additional dependencies on the host which is nice.

You don't want oVirt Guest Agent as that one is deprecated. What you really want is QEMU Guest Agent.

cgwalters:
OK this turned into https://bugzilla.redhat.com/show_bug.cgi?id=1900759
/close

cgwalters:
Though after that meeting I did think of yet another option - stick the code inside https://github.com/openshift/cluster-api-provider-ovirt
It's already oVirt/RHV specific etc. - The thing that's architecturally ugly about this though is cluster API is supposed to be about managing the infrastructure (possibly for multiple clusters, though AFAIK OpenShift doesn't use that) - it's not supposed to be about managing stuff *inside* the machines.   But...eh, I can't imagine it'd be too hard to carry a special case there to say something like "if (running_in_openshift) deploy_daemonset()".

Anyways though, fine continuing the angle of including in the OS by default, but leaving this here in case you guys are like "ah that sounds great!".

mandre:
> Though after that meeting I did think of yet another option - stick the code inside https://github.com/openshift/cluster-api-provider-ovirt
> It's already oVirt/RHV specific etc.

OpenStack would also benefit from having the qemu-guest-agent available somehow. If we were to go down the "daemonset in the platform specific cluster API providers" path, that would mean duplicating the code from the ovirt provider to the openstack one. I can see it be rather hard to maintain on the long run.

PR #2044, opened 2020-08-31
Bug 1866873: update MCDDrainErr to reduce cardinality & fix nodename

kikisdeliveryservice:
This fix does the following:

- fixes bug where node name was missing for MCDDrainErr
- reduces cardinality of MCDDrainErr via set err messages & value
- adds a duration to the metric (15m) to give time for the failed drain to resolve
- fixes labels on other metrics (minor)

This fixes some Drain metric pain points have have been going on forever..
see: https://bugzilla.redhat.com/show_bug.cgi?id=1866873#c3 for screenshot

### DIFF
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -20,7 +20,7 @@ spec:
       rules:
         - alert: MCDDrainError
           expr: |
-            mcd_drain > 0
+            mcd_drain_err > 0
           labels:
             severity: warning
           annotations:
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -23,6 +23,7 @@ spec:
             mcd_drain_err > 0
           labels:
             severity: warning
+          for: 15m
           annotations:
             message: "Drain failed on {{ $labels.node }} , updates may be blocked. For more details:  oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon"
     - name: mcd-pivot-error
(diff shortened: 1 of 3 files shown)


install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -20,9 +20,10 @@ spec:
       rules:
         - alert: MCDDrainError
           expr: |
-            mcd_drain > 0
+            mcd_drain_err > 0
```
wking, line 23: How does this react to labels changing?  Does that reset the clock?  Do we alert if at least one node has been reporting an error for every time in the past 15m, or does it need a single node to report an error for the full 15m?  I've reread [the alert docs][1], but am still not clear on this :/

[1]: https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/

kikisdeliveryservice: Nothing here has changed how our alerts fundamentally work. 

These are alerts from mcd which are per node. So any mcd alert is based on the status of the underlying node. 

In this case if the drain error for the specific node is still present after the 15 minutes (which means failing after many, many retries) it will fire instead of immediately firing after very the first failure. This means that we have a better signal as to whether or not the error is transient/self-recovering before we send anything.

If another node somehow also started failing to drain, there would be a separate alert associated with that node, etc...

We plan on adding more in the future this is intended to fix current behavior.

yuqi-zhang (approved):
Overall lgtm, just one question: since you removed the "failing drain time" from the reporting directly, will users see that the timeout was 15m somehow? In the screenshot I just see the name. If not maybe its worth reporting somewhere that 15m is when we report this alert.

kikisdeliveryservice:
It's 

> Overall lgtm, just one question: since you removed the "failing drain time" from the reporting directly, will users see that the timeout was 15m somehow? In the screenshot I just see the name. If not maybe its worth reporting somewhere that 15m is when we report this alert.

It's actually super standard to have a delay before firing an alert to make sure the alert is true other ocp operators also do so without further info as that's avail in the mcd logs.. it was my mistake to not have one initially.  I can add docs to this to clarify, but I dont think explicitly saying so in the alert makes sense. Failing drain time was a timestamp (yikes) so wasn't providing any signal and was causing the alert to actually be super confusing to users as the value of the alert was always changing.

yuqi-zhang:
ack, thanks!
/lgtm

PR #2050, opened 2020-09-01
[release-4.5] Bug 1856529: Create a drop-in file for cri-o's default_capabilities

umohnani8:
Fixes #1856529 (https://bugzilla.redhat.com/show_bug.cgi?id=1856529)

**- What I did**

Split out the default_capabilities list from the default.conf
file to a spearate drop-in file under /etc/crio/crio.conf.d.
cri-o enables the NET_RAW and SYS_CHROOT capabilities by default
in the config the MCO lays down. We want to drop these capabilities,
but need to make sure that we don't break existing clusters by doing so.
Thus, split out the default capabilities list to its own drop-in file and
create an MC for it. This way, the user has the ability to delete this MC
once they believe that removing these capabilities will not affect their
cluster.

We will drop this logic in the following version as the capabilities
MC would have already been created. It will then be up to the user on
when they want to drop these capabilities.

The upgrade flow will be:
4.5.x --> 4.5.y+capabilities MC logic (this will be mandatory upgrade point)
4.5.y+capabilities MC logic --> 4.5.z

Also drop the default_capabilities list from the crio template.

**- How to verify it**

1. Start a 4.5 cluster
2. Create a custom release image with the changes from this PR and upgrade your cluster to it.
3. Check that you have 2 MCs created `00-worker/master-capabilities` and a `crio-capabilities` config map is create in the openshift-machine-config-operator namespace. Log onto the nodes and verify that the `/etc/crio/crio.conf.d/01-mc-defaultCapabilities` file exists.
4. Delete one of the capabilities MC and reboot a node, restart the MCC pod etc and check that the capabilities MC you deleted wasn't created again. Log on the node for which you deleted the MC and run `crio config`, check the `default_capabilities` list in the output, it should not have the NET_RAW or SYS_CHROOT capabilities in it.
5. Upgrade your cluster to 4.5.x or 4.6 and check that the capabilities MC that you did not delete still exists. Log onto a node and look at the `/etc/crio/crio.conf.d/00-default` file to ensure that the `default_capabilities` list is no longer there.

Split out the default_capabilities list to a separate MC and drop-in file at crio.conf.d.

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -68,10 +68,11 @@ var updateBackoff = wait.Backoff{
 
 // Controller defines the container runtime config controller.
 type Controller struct {
-	templatesDir string
+	namespace, templatesDir string
 
 	client        mcfgclientset.Interface
 	configClient  configclientset.Interface
+	kubeClient    clientset.Interface
 	eventRecorder record.EventRecorder
 
 	syncHandler                   func(mcp string) error
@@ -106,7 +107,7 @@ type Controller struct {
 
 // New returns a new container runtime config controller
 func New(
-	templatesDir string,
+	namespace, templatesDir string,
 	mcpInformer mcfginformersv1.MachineConfigPoolInformer,
 	ccInformer mcfginformersv1.ControllerConfigInformer,
 	mcrInformer mcfginformersv1.ContainerRuntimeConfigInformer,
@@ -122,9 +123,11 @@ func New(
 	eventBroadcaster.StartRecordingToSink(&coreclientsetv1.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
 
 	ctrl := &Controller{
+		namespace:     namespace,
 		templatesDir:  templatesDir,
 		client:        mcfgClient,
 		configClient:  configClient,
+		kubeClient:    kubeClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "machineconfigcontroller-containerruntimeconfigcontroller"}),
 		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "machineconfigcontroller-containerruntimeconfigcontroller"),
 		imgQueue:      workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
@@ -171,6 +174,10 @@ func New(
 	ctrl.clusterVersionListerSynced = clusterVersionInformer.Informer().HasSynced
 
 	ctrl.patchContainerRuntimeConfigsFunc = ctrl.patchContainerRuntimeConfigs
+	// Add to the queue to trigger a sync when an upgrade happens
+	// this ensures that the capabilities MC is created on an upgrade
+	// This will be removed in the next version
+	ctrl.queue.Add("force-sync-on-upgrade")
 
 	return ctrl
 }
@@ -461,6 +468,17 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 		glog.V(4).Infof("Finished syncing ContainerRuntimeconfig %q (%v)", key, time.Since(startTime))
 	}()
 
+	// First let's create the MC for the default capabilities file
+	// This will be removed in the next version
+	if err := ctrl.createCapabilitiesMC(); err != nil {
+		return fmt.Errorf("failed to create the crio capabilitie MC: %v", err)
+	}
+	// If the key is set to force-sync-on-upgrade, then we can return after creating
+	// the capabilities MC.
+	if key == "force-sync-on-upgrade" {
+		return nil
+	}
+
 	_, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		return err
@@ -504,12 +522,6 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 		return ctrl.syncStatusOnly(cfg, err)
 	}
 
-	if len(mcpPools) == 0 {
-		err := fmt.Errorf("containerRuntimeConfig %v does not match any MachineConfigPools", key)
-		glog.V(2).Infof("%v", err)
-		return ctrl.syncStatusOnly(cfg, err)
-	}
-
 	for _, pool := range mcpPools {
 		role := pool.Name
 		// Get MachineConfig
@@ -522,9 +534,8 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 		// If we have seen this generation and the sync didn't fail, then skip
 		if !isNotFound && cfg.Status.ObservedGeneration >= cfg.Generation && cfg.Status.Conditions[len(cfg.Status.Conditions)-1].Type == mcfgv1.ContainerRuntimeConfigSuccess {
 			// But we still need to compare the generated controller version because during an upgrade we need a new one
-			mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
-			if mcCtrlVersion == version.Hash {
-				return nil
+			if mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey] == version.Hash {
+				continue
 			}
 		}
 		// Generate the original ContainerRuntimeConfig
@@ -605,6 +616,95 @@ func (ctrl *Controller) mergeConfigChanges(origFile *igntypes.File, cfg *mcfgv1.
 		return nil, ctrl.syncStatusOnly(cfg, err, "could not update container runtime config with new changes: %v", err)
 	}
 	return cfgTOML, ctrl.syncStatusOnly(cfg, nil)
+}
+
+// createCapabilitiesMC creates a Machine Config for the crio drop-in capabilities file.
+// We are dropping the NET_RAW and SYS_CHROOT capabilitie by default and in order to not break
+// existing clusters, we are creating a drop-in file at /etc/crio/crio.conf.d that has NET_RAW
+// and SYS_CHROOT as part of the default capabilities list. Users then have the option to delete the
+// capabilities MC that creates this drop-in file.
+// Create a crio-capabilities config map to keep track of whether this capabilities MC was already created
+// once. This ensures that if the user deletes the capabilities MC, the controller doesn't recreate it on a
+// resync.
+func (ctrl *Controller) createCapabilitiesMC() error {
+	var configMapName = "crio-capabilities"
+
+	// Check if the crio-capabilities config map exists in the openshift-machine-config-operator namespace
+	capCM, err := ctrl.kubeClient.CoreV1().ConfigMaps(ctrl.namespace).Get(context.TODO(), configMapName, metav1.GetOptions{})
+	capCMIsNotFound := errors.IsNotFound(err)
+	if err != nil && !capCMIsNotFound {
+		return fmt.Errorf("error checking for %s config map: %v", configMapName, err)
+	}
+	// If the crio-capabilities config map exists, that means the capabilities MC was already created
+	// so we should not create this MC again and return
+	if capCM != nil && !capCMIsNotFound {
+		return nil
+	}
+
+	sel, err := metav1.LabelSelectorAsSelector(metav1.AddLabelToSelector(&metav1.LabelSelector{}, builtInLabelKey, ""))
+	if err != nil {
+		return err
+	}
+	// Find all the MachineConfigPools
+	mcpPoolsAll, err := ctrl.mcpLister.List(sel)
+	if err != nil {
+		return err
+	}
+
+	// Create the capabilities MC for all the available pools
+	for _, pool := range mcpPoolsAll {
+		managedKey := getManagedKeyCaps(pool)
+		mc, err := ctrl.client.MachineconfigurationV1().MachineConfigs().Get(context.TODO(), managedKey, metav1.GetOptions{})
+		isNotFound := errors.IsNotFound(err)
+		if err != nil && !isNotFound {
+			return fmt.Errorf("error checking for %s machine config: %v", managedKey, err)
+		}
+		// continue to the next MC if this already exists
+		if mc != nil && !isNotFound {
+			continue
+		}
+
+		tempIgnCfg := ctrlcommon.NewIgnConfig()
+		mc, err = mtmpl.MachineConfigFromIgnConfig(pool.Name, managedKey, tempIgnCfg)
+		if err != nil {
+			return fmt.Errorf("could not create capabilities MachineConfig from new Ignition config: %v", err)
+		}
+		rawCapsIgnition, err := json.Marshal(createNewIgnition(createDefaultCapsFile()))
+		if err != nil {
+			return fmt.Errorf("error marshalling capabilities config ignition: %v", err)
+		}
+		mc.Spec.Config.Raw = rawCapsIgnition
+		// Create the capabilities MC
+		if err := retry.RetryOnConflict(updateBackoff, func() error {
+			_, err = ctrl.client.MachineconfigurationV1().MachineConfigs().Create(context.TODO(), mc, metav1.CreateOptions{})
+			return err
+		}); err != nil {
+			return fmt.Errorf("could not create MachineConfig for capabilities: %v", err)
+		}
+		glog.Infof("Applied Capabilities MC %v on MachineConfigPool %v", managedKey, pool.Name)
+	}
+
+	// Create the config map for capabilities so we know that the capabilities MC has been created
+	capCM.Name = configMapName
+	capCM.Namespace = ctrl.namespace
+	if _, err := ctrl.kubeClient.CoreV1().ConfigMaps(ctrl.namespace).Create(context.TODO(), capCM, metav1.CreateOptions{}); err != nil {
+		return fmt.Errorf("error creating %s config map: %v", configMapName, err)
+	}
+	return nil
+}
+
+// RunCapsBootstrap creates the capabilities mc on bootstrap
+func RunCapsBootstrap(mcpPools []*mcfgv1.MachineConfigPool) ([]*mcfgv1.MachineConfig, error) {
+	var res []*mcfgv1.MachineConfig
+	for _, pool := range mcpPools {
+		capsIgn := createNewIgnition(createDefaultCapsFile())
+		mc, err := mtmpl.MachineConfigFromIgnConfig(pool.Name, getManagedKeyCaps(pool), capsIgn)
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, mc)
+	}
+	return res, nil
 }
 
 func (ctrl *Controller) syncImageConfig(key string) error {
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/vincent-petithory/dataurl"
 
+	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -60,6 +61,7 @@ type fixture struct {
 	client         *fake.Clientset
 	imgClient      *fakeconfigv1client.Clientset
 	operatorClient *fakeoperatorclient.Clientset
+	kubeClient     *k8sfake.Clientset
 
 	ccLister   []*mcfgv1.ControllerConfig
 	mcpLister  []*mcfgv1.MachineConfigPool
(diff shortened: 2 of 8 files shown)


umohnani8:
verify should stop complaining about lint once https://github.com/openshift/machine-config-operator/pull/2034 is merged

yuqi-zhang (commented):
So if I understand this correctly, the crux of the PR wants to add a new containerruntimeconfigcontroller (henceforth CRCC) controlled MC that splits out a sub functionality of cri-o configuration, right?

 1. today, could you modify these capabilities with a CRC?
 2. I'm not really sure how I feel about a completely new default MC being generated for this purpose
 3. I don't really know about the idea of using a one-time config map that acts as a token to be deleted to signal its completion (if I understand correctly). I feel like this is somewhat anti-design to how the rest of the MCC works, so in light of that, how would you feel about:

Have some version of the MCO (with this PR) perform the following:

- consider if we should block modification of this via a CRC (if that's possible today), but otherwise remove its management through the CRCC
- have this PR create a dropin file, as you were planning to do
- if a user wants to modify the capabilities, they could remove parts of that MC

Now this doesn't really fix the problem of this being a security vulnerability, since itd still be up to the user to apply said fix. So I'm considering forcing this to be false by default, after we check what scenarios will fail if we do so

Edit: oh shoot I missed the 4.6 PR somehow, will comment there

umohnani8:
The list of capabilities cannot be modified by the ctrcfg CR today. We had the default_capabilities list in the template in MCO because we dropped the NET_RAW and SYS_CHROOT capabilities in the cri-o code defaults, and wanted these to be enabled in the cluster. We later realized that doing this was actually a regression from 3.x.

But since we have had these capabilities enabled by default from the beginning of 4.x, there could be clusters that have workloads that are actually using the privileges given by these 2 capabilities and just dropping them out of the blue would cause those clusters to break. So we decided to create a version of 4.5.x that splits out the capabilities (with NET_RAW and SYS_CHROOT) into a drop-in file under crio.conf.d and this is done by creating a MC for it for all the pools available in the cluster. This way, we get rid of these 2 capabilities from the default crio template in the MCO, but don't break clusters when they upgrade to this version of 4.5.x - we will force that every cluster go through this upgrade path btw.

The plan is to document what we have done and encourage customers to make their workloads run without having to enable these 2 capabilities. Once they think that they are ready, they can simply delete this MC and it will remove the capabilities drop-in file. The config map is a check we use to ensure that we don't recreate this capabilities MC when a sync is triggered and it is found that the MC no longer exists - basically we make sure we create the MC only once for each pool when we upgrade.

In the next version of 4.5, let's say 4.5.y, the default_capabilities list will be removed from the MCO template and this commit that splits out the capabilities to its own drop-in file and MC will be removed as well. So this is the version where we just drop this 2 capabilities like in https://github.com/openshift/machine-config-operator/pull/2051. We can be sure that this won't cause any issues as the clusters would have been forced to upgrade to 4.5.x (with the capabilities MC logic) first. We leave it up to the user to decide when they want to drop this capabilities and delete the MC. So if they want to keep it for a while, the capabilities MC will persist across upgrades.

The upgrade path would be:
4.5.a --> 4.5.x (has the MC capabilities logic)
4.5.x --> 4.5.y (default_capabilities is removed from the MCO template and no longer have the MC capabilities logic)

Fresh clusters from 4.5.y will no longer have the security regressions.

yuqi-zhang:
> The list of capabilities cannot be modified by the ctrcfg CR today. We had the default_capabilities list in the template in MCO because we dropped the NET_RAW and SYS_CHROOT capabilities in the cri-o code defaults, and wanted these to be enabled in the cluster. We later realized that doing this was actually a regression from 3.x.

Sounds good, I just wanted to double check

> The plan is to document what we have done and encourage customers to make their workloads run without having to enable these 2 capabilities. Once they think that they are ready, they can simply delete this MC and it will remove the capabilities drop-in file. The config map is a check we use to ensure that we don't recreate this capabilities MC when a sync is triggered and it is found that the MC no longer exists - basically we make sure we create the MC only once for each pool when we upgrade. And needed to ensure that this "check" persisted across reboots and all, so that is why I went with the config map.

Right, I think I phrased my initial point poorly. I was trying to think if there was a better way to do it such that clusters installed from 4.5 and before (which have this capability on) can spit out a MC that's not being synced by the controller, much like if a user had installed a conf, so it can be removed later without having a config map tracking, thereby bypassing 1. the need to force all upgrades to go through 4.5.x before upgrading further to 4.6 and 2. the creation of a configmap that will (as far as I can tell) live on the system forever unless you want to force remove it in the 4.6 patch. Offhand though I have not yet thought of a good way to do this.

Actually now I think about it the capability will be a controller-created MC right? If it goes to 4.6, since the 4.6 branch doesn't track it, wouldn't it throw an error since the MC wouldn't get re-rendered by anyone else (controller version mismatch) if the user opted to keep it around before upgrading to 4.6?

> In the next version of 4.5, let's say 4.5.y, the default_capabilities list will be removed from the MCO template and this commit that splits out the capabilities to its own drop-in file and MC will be removed as well. So this is the version where we just drop this 2 capabilities like in #2051. We can be sure that this won't cause any issues as the clusters would have been forced to upgrade to 4.5.x (with the capabilities MC logic) first. We leave it up to the user to decide when they want to drop this capabilities and delete the MC. So if they want to keep it for a while, the capabilities MC will persist across upgrades.

Right so new clusters in 4.6 doesn't have this, but it seems that with this patch 4.5 installs after this patch will still have the capabilities. Do we want all 4.5 to keep default capabilities and 4.6 to not? Or 4.5 before this patch to keep existing defaults and 4.5 after this patch + 4.6 to not? The former will be true with this current PR right?

yuqi-zhang:
To unwind that a little, I'm tentatively ok with this method (need to think a bit more), but support for this needs to exist in 4.6 as well for upgrades if the machineconfig exists when an upgrade happens, since we don't force the user to delete the extra machineconfig created to handle capabilities. This dependency is what makes me nervous.

umohnani8:
>  Right so new clusters in 4.6 doesn't have this, but it seems that with this patch 4.5 installs after this patch will still have the capabilities. Do we want all 4.5 to keep default capabilities and 4.6 to not? Or 4.5 before this patch to keep existing defaults and 4.5 after this patch + 4.6 to not? The former will be true with this current PR right?

We want 4.5 before this patch to keep the exiting defaults and 4.5 after this patch plus 4.6 to not. And you are right about the controller version mismatch happening, for that we could add a patch to track if the caps MC exists and if they do, we update the controller version on upgrades. I will add that to https://github.com/openshift/machine-config-operator/pull/2051.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -171,6 +174,10 @@ func New(
 	ctrl.clusterVersionListerSynced = clusterVersionInformer.Informer().HasSynced
 
 	ctrl.patchContainerRuntimeConfigsFunc = ctrl.patchContainerRuntimeConfigs
+	// Add to the queue to trigger a sync when an upgrade happens
+	// this ensures that the capabilities MC is created on an upgrade
+	// This will be removed in the next version
+	ctrl.queue.Add("force-sync-on-upgrade")
```
umohnani8, line 180: @yuqi-zhang, this line adds an item to the controller queue to force it to resync on an upgrade/start up. For a fresh installation, this was causing failures as the bootstrap didn't have the capabilities MC, but was expecting it I guess and was complaining about an expected rendered MC not being found. So I added the `RunCapBootstrap` function at line https://github.com/openshift/machine-config-operator/pull/2050/files#diff-22412a83973655ff84984dc7d9969013R699. 
I tested this out and had a successful installation, but want to double check if there is a case where having this force sync on controller start up could cause issues with the bootstrap logic. wdyt?

cgwalters:
We often differentiate today from "new install" and "upgrade" by having openshift-install do something; that's how we're handling things like the https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/user-data-secret-managed.md

Another alternative is parsing the `clusterversion` history and seeing if the first one is `4.6` although I'm not sure if that's a blessed path.

Another path is having the MCO bootstrap phase create an object.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
yuqi-zhang: Hmm sorry I missed this sync when I did the initial review. Is this sync needed to get the capabilities MC to trigger?

When an upgrade happens the new MCO with this new CRCC should be deployed first. It will then (when running the sync) call `createCapabilitiesMC` and split out the new MC before it attempts an update. The bootstrap should also be running a MCC which should also have the same workflow. Thus as soon as you install/update to a version with this patch, it would create the MC + configmap without the need for a sync again right? Maybe I am confused there

umohnani8: So when you don't have any ctrcfg CRs created in your cluster and you upgrade, the controller does not resync and that is why I added an item to the queue to force this sync to happen. So when the cluster is upgraded and the controller starts up, it has an item in its queue and we sync. This sync splits creates the capabilities MC + config map.
For the bootstrap, we create the capabilities MC but not the config map, because it looks like when the cluster is ready and the controller starts up, a sync is forced by this line. In my test run everything worked fine, the capabilities MCs and config map were created.
I am not very familiar with the bootstrap process, so wanted to see if there may be any scenarios where having the controller sync up immediately after the bootstrap/installation could cause any issues.

yuqi-zhang:
Ok so since we basically can't force existing clusters to drop the capabilities, I think we should carry the support in the controller forwards into 4.6.

If we do that, in 4.5 we can create the configmap as you are doing here. Maybe in 4.6 and beyond the configmap is generated as part of the bootstrap process so the CRCC sees it when it installs, and thus assumes a "previous version" has generated the MC and will not attempt to. Otherwise if the MC still exists it will still sync it to the new controller version.

I'd really like to meet and discuss a bit on the design overall. I guess I'm nervous since this is an unprecedented workflow in a few aspects:
1. we're introducing a concept of a "one-time cluster-generated machineconfig"
2. we're letting users delete a cluster-generated machineconfig without the controller trying to regenerate it

This will also potentially introduce a diff between 4.5->4.6 clusters and 4.6-borne clusters (although this exists today elsewhere e.g. bootimage compatibility so I'm not as worried about this point).

Maybe I'm overthinking it :D . The design you have now I can get behind as well (so long as we also carry support in 4.6). I just want to make sure we're setting a good precedent on how to deal with these issues.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
yuqi-zhang: Ah I see, hmm, I almost feel like either
a) the crcc is updated to syncs this
b) this checking lives somewhere else instead (maybe the MCC directly? Not sure if that'd be a good idea either)

Offhand I don't know if this will cause bootstrap problems but since the installation is successful now I think it shouldnt? This sync really just affects the CRCC anyways right

umohnani8: Yeah, from my tests the sync right after installation hasn't affected anything with the bootstrap stuff, so we should be good.

umohnani8:
This should be ready now. Updated the PR so that we are not assigning a controller version to the capabilities MC created, this way we won't run into any out of sync issues during upgrades. @yuqi-zhang PTAL
These are the cases I have tested so far with this PR:
1) Fresh cluster installation
2) Upgrade to version with this patch
     - with ctrcfg CR on one pool
     - with ctrcfg CR on all the pools
     - with no ctrcfg CR
3) After upgrade tried the following
     - if no ctrcfg CR was created prior to upgrade, create one
     - if ctrcfg CR was created prior to upgrade, delete it
     - delete one or both of the capabilities MCs
     - restart the MCC
     - restart the MCO
     - reboot nodes (make sure that the delete capabilities MCs are not created again)
     - check that the config map exists in the openshift-machine-config-operator namespace
3) Upgrade to next version that has https://github.com/openshift/machine-config-operator/pull/2051 patch
     - upgrade with capabilities MC existing
     - upgrade with no capabilities MC

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -176,6 +188,10 @@ func getManagedKeyReg(pool *mcfgv1.MachineConfigPool) string {
 	return fmt.Sprintf("99-%s-%s-registries", pool.Name, pool.ObjectMeta.UID)
 }
 
+func getManagedKeyCaps(pool *mcfgv1.MachineConfigPool) string {
+	return fmt.Sprintf("99-%s-crio-capabilities", pool.Name)
```
yuqi-zhang, line 192: Minor: maybe we should a "-generated" or something like that as a suffix so users know that it was the controller who created it (since this won't have a controller version attached)

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -524,7 +535,16 @@ func TestImageConfigCreate(t *testing.T) {
 			f.expectGetMachineConfigAction(mcs2)
 			f.expectCreateMachineConfigAction(mcs2)
 
-			f.run("cluster")
+			c := f.newController()
```
yuqi-zhang, line 538: sorry, could you clarify what this test change is trying to do?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -607,6 +618,94 @@ func (ctrl *Controller) mergeConfigChanges(origFile *igntypes.File, cfg *mcfgv1.
...
+			return err
+		}); err != nil {
+			return fmt.Errorf("could not create MachineConfig for capabilities: %v", err)
+		}
+		glog.Infof("Applied Capabilities MC %v on MachineConfigPool %v", managedKey, pool.Name)
+	}
+	return nil
+}
+
+// RunCapsBootstrap creates the capailities mc on bootstrap
```
yuqi-zhang, line 695: capailities > capabilities

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -607,6 +618,94 @@ func (ctrl *Controller) mergeConfigChanges(origFile *igntypes.File, cfg *mcfgv1.
...
+		return err
+	}
+	// Find all the MachineConfigPools
+	mcpPoolsAll, err := ctrl.mcpLister.List(sel)
+	if err != nil {
+		return err
+	}
+
+	// Create the capabilities MC for all the available pools
+	for _, pool := range mcpPoolsAll {
```
yuqi-zhang, line 661: Hmm, today all custom pools must inherit from the `worker` pool. By looping through all pools aren't you effectively double-adding capabilities as a drop-in to a custom pool twice? Once as a worker MC and once as a custom pool MC.

Maybe we should just add it to master/worker.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -607,6 +618,94 @@ func (ctrl *Controller) mergeConfigChanges(origFile *igntypes.File, cfg *mcfgv1.
...
+		}); err != nil {
+			return fmt.Errorf("could not create MachineConfig for capabilities: %v", err)
+		}
+		glog.Infof("Applied Capabilities MC %v on MachineConfigPool %v", managedKey, pool.Name)
+	}
+	return nil
+}
+
+// RunCapsBootstrap creates the capailities mc on bootstrap
+func RunCapsBootstrap(mcpPools []*mcfgv1.MachineConfigPool) ([]*mcfgv1.MachineConfig, error) {
```
yuqi-zhang, line 696: Do we need this function? I was under the impression that since you added a forced sync it will generate the capabilities MC that way.

This also looks like it just creates the machineconfig. What about the associated configmap?

yuqi-zhang (commented):
Overall I think I'm more comfortable with this approach. A few questions:

pkg/controller/container-runtime-config/container_runtime_config_controller.go
yuqi-zhang: Oh I just tested this and it seems to only apply select master/worker even with a custom pool, maybe this isn't an issue

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: Yeah, with `f.run("cluster")` it was calling the sync function for both containerRuntimeConfig and Image. This was causing the test to fail as it was expecting the actions for creating the capabilities. But since this function is only testing changes to an Image CR, I changed it to just run the Image sync function, similar to how the other Image tests are doing. I think this was accidentally left over from when the Image tests were modified.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: We want to apply the capabilities MC to all the available pools in the cluster. Some customers have more than master and worker pool and I thought using the `LabelSelectorAsSelector` function gets you all the available pools in the cluster. I might be wrong, but so far I have only tested it on clusters that have master and worker pools only.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Yeah, we do need this. From what I understand, forcing the controller to sync is rendering an MC that has the capabilities stuff in it, but if we don't explicitly create this on bootstrap the cluster fails to start as the master nodes are trying to find a rendered MC that does not exist. The config map is created when the controller starts up, so there is no need to create it at bootstrap.

pkg/controller/container-runtime-config/helpers.go
umohnani8: yup, fixed!

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Fixed!

pkg/controller/container-runtime-config/container_runtime_config_controller.go
yuqi-zhang: Yeah I created a cluster, added an `infra` pool, and upgraded to this PR, and it only created capabilities MC for master/worker.

Today in the MCO we only allow one custom pool, and it MUST inherit from worker if I recall correctly. So in essence in this case the `infra` node would have the capabilities MC through the worker MC, so I guess in that sense it will work. Although I am not very sure why selecting the way you did didnt generate a infra MC. I'll take a look but it sounds like the behaviour is what we want anyways?

yuqi-zhang:
/retest

failures unrelated

yuqi-zhang (approved):
Functionality wise I think this is good to go. I did some upgrade testing and it all seems to work.

For this to work properly we have to make sure that 4.6 on release drops the default_capabilities section. If I upgrade to a 4.6 nightly today without the master PR the capabilities would get re-added (which is expected).

Let's also make sure we've communicated this and we get a +1 from someone regarding blocking the upgrade edges cc @mrunalp

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: Does this mean that if the user deletes the worker capabilities MC, the changes will roll out to both worker and infra nodes? I think we want each pool to have its own capabilities MC so that users can delete the MC for the specific pools where they want to drop these capabilities.
And is this behaviour only for the infra pool, or it will be for all other custom pools the user creates? If this is the way it is done now, we could probably document that deleting the worker pool capabilities MC will drop these capabilities on nodes that are in the custom pools as well as the worker pool.

umohnani8:
@yuqi-zhang @kikisdeliveryservice can I please get a final review here.

yuqi-zhang (approved):
Ok I think I'm +1 with this PR. Functionality should be good. An ack regarding the edge creation was given as well. It's very important to communicate this change properly but we can handle that post-merge.

One last minor concern: there exists a path in the code that the config map is created, but something fails right after and the MC is not correctly created. This might leave the cluster in a somewhat limbo state. Perhaps creating the MC should come first? Either case shouldn't error too often.

If anyone else could take a double check would be great too, otherwise I will just go ahead and lgtm

yuqi-zhang:
/lgtm

I'm going to give the LGTM. Let's move this forward. The other labels I think might be to be overridden

umohnani8:
> One last minor concern: there exists a path in the code that the config map is created, but something fails right after and the MC is not correctly created. This might leave the cluster in a somewhat limbo state. Perhaps creating the MC should come first? Either case shouldn't error too often.

@yuqi-zhang updated the code so that we create the configMap after we create the capabilities MCs. There is a check in place that moves on if the capabilities MC already exists, so in the event where creating the configMap fails for some reason the first time, the next resync should try and create the configMap again, but it won't try to create a capabilities MC if it already exists.

yuqi-zhang (approved):
/lgtm

yuqi-zhang:
So the BZ can only handle 4.6 backporting to 4.5 but not a reverse dependency. When we get the cherry-pick-approved label, we're also going to have to ask them to override the BZ label (at least I don't think I can)

yuqi-zhang:
/lgtm cancel
/hold

pending discussions

umohnani8:
We are getting ready to merge this, can I get a final look here and lgtm + approve please. Just waiting on the rehearse jobs in https://github.com/openshift/release/pull/12061 to pass, so we can hold till then.
@yuqi-zhang PTAL

yuqi-zhang:
Are we still good on the blocking upgrade edge?

umohnani8:
@yuqi-zhang yup, we are good on that. Started a thread in slack to get that going and cc'ed you on it :)

PR #2051, opened 2020-09-02
Bug 1874671:  Drop the default_capabilities list from the crio template

umohnani8:
Fixes #1874671 (https://bugzilla.redhat.com/show_bug.cgi?id=1874671)

**- What I did**
Dropping this list will also drop the NET_RAW and SYS_CHROOT
capabilities from the default list.

**- How to verify it**
Log onto a node and run `crio config`, the list of default_capabilities returned should not have NET_RAW and SYS_CHROOT in it.

Drop default_capabilities list from the crio template

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -14,18 +14,8 @@ contents:
     ]
     log_level = "info"
     cgroup_manager = "systemd"
-    default_capabilities = [
-        "CHOWN",
-        "DAC_OVERRIDE",
-        "FSETID",
-        "FOWNER",
-        "NET_RAW",
-        "SETGID",
-        "SETUID",
-        "SETPCAP",
-        "NET_BIND_SERVICE",
-        "SYS_CHROOT",
-        "KILL",
+    default_sysctls = [
+        "net.ipv4.ping_group_range=0 2147483647",
     ]
     hooks_dir = [
         "/etc/containers/oci/hooks.d",
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -14,18 +14,8 @@ contents:
     ]
     log_level = "info"
     cgroup_manager = "systemd"
-    default_capabilities = [
-        "CHOWN",
-        "DAC_OVERRIDE",
-        "FSETID",
-        "FOWNER",
-        "NET_RAW",
-        "SETGID",
-        "SETUID",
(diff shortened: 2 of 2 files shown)


cgwalters:
I'm tentatively OK with this plan but (I've said this before) IMO *by far* the biggest security improvement we could make would be landing [user namespaces](https://github.com/kubernetes/enhancements/issues/127).  And specifically, that covers this because e.g. `CAP_SYS_CHROOT` becomes user namespace specific for example.  `CAP_NET_RAW` is a bit more special though.

umohnani8:
Yeah, user namespace work is underway in cri-o. We enabled NET_RAW and SYS_CHROOT by default in 4.x causing a regression from 3.x. This PR and https://github.com/openshift/machine-config-operator/pull/2050 is the attempt at fixing this regression without breaking clusters that need these capabilities.

umohnani8:
@kikisdeliveryservice @yuqi-zhang PTAL
/retest

yuqi-zhang:
I left a comment on the 4.5 branch since I didn't see this first, https://github.com/openshift/machine-config-operator/pull/2050#pullrequestreview-481373351

that said, I'm a bit confused as to why the 4.6 PR only does the dropping of those capabilities, what would the default look like for a new install then?

umohnani8:
I commented on https://github.com/openshift/machine-config-operator/pull/2050. But to answer here, a new install would not have the NET_RAW and SYS_CHROOT capabilities enabled by default and we will no longer have the security regression. The list of default_capabilities in the cri-o code already drops the 2 capabilities mentioned above https://github.com/cri-o/cri-o/blob/6e571d7a4de62105cba293ebe6292043401e1b15/internal/config/capabilities/capabilities.go#L15.

umohnani8:
@yuqi-zhang added a commit with the logic to check for any existing capabilities mc and update the generated controller version whenever there is a change.
Will work on fixing the test failures tomorrow.

umohnani8:
This PR is also ready. We will need to get this in before https://github.com/openshift/machine-config-operator/pull/2050

cgwalters:
The OKD test is known failing and not gating right now, it's being worked on.

Where were we on the discussion around this?  I had been arguing for this "hard drop" approach but the fact that OpenShift 3.11 (right?) and docker/containerd do enable these capbilities by default seems to weight towards keeping them.

cgwalters (commented):
/hold
for clarification

umohnani8:
/hold cancel
This is ready once tests pass, can I get the approve and lgtm labels here please @yuqi-zhang @cgwalters

yuqi-zhang (approved):
/lgtm

PR #2052, opened 2020-09-02
Bug 1866117: pkg/server: reduce calls to ignition parse

runcom:
There's no need to parse the whole ignition section from the MC every
time we're about to append a file in the MCS. This leads to a performance
regression when the ignition section contains a lot of files as
discovered in https://bugzilla.redhat.com/show_bug.cgi?id=1866117
compared to pre 4.5.
Also removed some parse call where it's obvious that in code we're
dealing with a particular version and there's no need to test against
something we know it can't be.

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -138,21 +138,11 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
-
-	// the internally saved config should be v3.1, but to eliminate the
-	// potential of race conditions, translate to the requested version
-	// to make sure
+	// we know we're at 3.1 in code.. serve directly, parsing is expensive...
+	// we're doing it during an HTTP request, and most notably before we write the HTTP headers
 	var serveConf *runtime.RawExtension
 	if reqConfigVer.Equal(*semver.New("3.1.0")) {
-		converted3, err := ctrlcommon.ConvertRawExtIgnitionToV3(conf)
-		if err != nil {
-			w.Header().Set("Content-Length", "0")
-			w.WriteHeader(http.StatusInternalServerError)
-			glog.Errorf("couldn't convert config for req: %v, error: %v", cr, err)
-			return
-		}
-
-		serveConf = &converted3
+		serveConf = conf
 	} else {
 		// Can only be 2.2 here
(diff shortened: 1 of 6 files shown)


runcom:
going through some testing...

pkg/server/api.go
```diff
@@ -138,21 +138,10 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
-
-	// the internally saved config should be v3.1, but to eliminate the
-	// potential of race conditions, translate to the requested version
-	// to make sure
+	// we know we're at 3.1 in code.. serve directly, parsing is expensive...
```
cgwalters, line 141: It's not *just* that parsing is expensive, the core problem is that we're doing it during an HTTP request, and most notably before we write the HTTP headers.

runcom: I'll add that you're right thanks Colin

runcom:
> I think this looks ok, but another alternative is to have the MCS watch the pool and pre-create the full spec2&spec3 configs whenever the pool's target config changes.

I really like this, I'm creating a tracker to move into that direction in the next release as I want to change this to a minimum set of changes w/o refactoring too much at this stage. The MCS watching the pull requires changing it to become sort of a controller and I don't want to introduce that _now_. Next release is a perfect candidate for this.

Focusing on data:
```
4.5 w/o this fix

real	0m2.379s
user	0m0.013s
sys	0m0.008s

4.5 with this

real	0m0.489s
user	0m0.018s
sys	0m0.007s
```
So it went to less than a second again even if still slower than 4.4 https://bugzilla.redhat.com/show_bug.cgi?id=1866117#c14 - I think this is acceptable for 4.5 and 4.6

pkg/server/server_test.go
```diff
@@ -53,44 +53,6 @@ func TestStringEncode(t *testing.T) {
 	}
 }
 
-func TestMachineConfigToRawIgnition(t *testing.T) {
```
cgwalters, line 56: This could be a unit test for `appendEncapsulated()` now though right?

runcom: yep I'm cleaning the branches while still testing this in 4.6, I'll sure add a translated unit

cgwalters:
/lgtm
It is actually pretty amazing how many times we were parsing and re-serializing the config before!

PR #2053, opened 2020-09-02
Bug 1874815: pkg/server: reduce calls to ignition parse

runcom:
There's no need to parse the whole ignition section from the MC every
time we're about to append a file in the MCS. This leads to a performance
regression when the ignition section contains a lot of files as
discovered in https://bugzilla.redhat.com/show_bug.cgi?id=1866117
compared to pre 4.5.

### DIFF
--- a/pkg/server/bootstrap_server.go
+++ b/pkg/server/bootstrap_server.go
@@ -1,11 +1,13 @@
 package server
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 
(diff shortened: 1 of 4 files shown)


runcom:
going through some testing...

runcom:
cross-referencing the result of testing https://github.com/openshift/machine-config-operator/pull/2052#issuecomment-685763214

kikisdeliveryservice:
Underlying BZ was finally approved & verified so let's get this in for our 4.5 users.

PR #2055, opened 2020-09-02
Bug 1870343: machine-config-daemon-pull: Retry pulling image

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1870343

Pulling the image can flake, we need to retry.  Today the
bootstrap node will retry indefinitely, see
https://github.com/openshift/installer/blob/18e286bfc0120142a480f6239191e90c6bf69d4f/data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template

Let's do the same.  It's not elegant but eh.

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-pull.service.yaml
+++ b/templates/common/_base/units/machine-config-daemon-pull.service.yaml
@@ -16,7 +16,7 @@ contents: |
   RemainAfterExit=yes
   # See https://github.com/coreos/fedora-coreos-tracker/issues/354
   ExecStart=/bin/sh -c '/bin/mkdir -p /run/bin && chcon --reference=/usr/bin /run/bin'
-  ExecStart=/bin/sh -c "/usr/bin/podman pull --authfile=/var/lib/kubelet/config.json --quiet '{{ .Images.gcpRoutesControllerKey }}'"
+  ExecStart=/bin/sh -c "while ! /usr/bin/podman pull --authfile=/var/lib/kubelet/config.json --quiet '{{ .Images.gcpRoutesControllerKey }}'; do sleep 1; done"
   ExecStart=/bin/sh -c "/usr/bin/podman run --rm --quiet --net=host --entrypoint=cat '{{ .Images.gcpRoutesControllerKey }}' /usr/bin/machine-config-daemon > /run/bin/machine-config-daemon.tmp"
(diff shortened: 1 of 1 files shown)


templates/common/_base/units/machine-config-daemon-pull.service.yaml
```diff
@@ -16,7 +16,7 @@ contents: |
   RemainAfterExit=yes
   # See https://github.com/coreos/fedora-coreos-tracker/issues/354
   ExecStart=/bin/sh -c '/bin/mkdir -p /run/bin && chcon --reference=/usr/bin /run/bin'
-  ExecStart=/bin/sh -c "/usr/bin/podman pull --authfile=/var/lib/kubelet/config.json --quiet '{{ .Images.gcpRoutesControllerKey }}'"
+  ExecStart=/bin/sh -c "while ! /usr/bin/podman pull --authfile=/var/lib/kubelet/config.json --quiet '{{ .Images.gcpRoutesControllerKey }}'; do sleep 1; done"
```
yuqi-zhang, line 19: Question: since this loops infinitely the service itself technically never fails. Do we want that? Or do we want to retry a certain amount of times and then have the service fail, so something can report it back up

cgwalters: Oh we definitely could improve this and should gather failures somewhere (also related to https://github.com/coreos/ignition/issues/585 ).

It'd also make sense to do at least bounded exponential backoff.

But that said: this is what the installer is doing for the release image today too.  I'm open to doing something more sophisticated but it feels like that can come as a followup - this is a one line change that's going to fix the problem and is "obviously correct".

I offhand can't think of a scenario where we fail to pull the oscontainer that wouldn't have failed elsewhere earlier.  Anyone who does hit some sort of actually permanent scenario would be able to ssh into the node and debug - the same thing that happens with the bootstrap today.

yuqi-zhang (approved):

runcom:
```
extension: {
master: "all 3 nodes are at latest configuration rendered-master-0b808f983f6c8795482d8c481341d3fc",
worker: "all 0 nodes are at latest configuration rendered-worker-d583060ddf3b75e6a28f25afaf74564c"
},
```
doesn't seem directly related to this PR but it's the second time I'm seeing this in the upgrade jobs, wondering where the compute nodes hang out..

PR #2056, opened 2020-09-02
Bug 1875181: [4.4] crio: bump to v1.17.5

haircommander:
**- What I did**
when the crio version falls out of sync with the verison in openshift, we end up rendering the incorect ignition config

This fixes a problem where, upon creating a ctrcfg, we lose our default_env value

**- How to verify it**
run `crio config` on a node
apply any ctrcfg
run `crio config` again
check the diff, noting the changes didn't strictly include that which came in the ctrcfg

note: this would be a good e2e test to add, however, running `crio config` on a node has proven to be challenging. One has to chroot, or else there will be ldd issues. However, a combination of the way the go os/exec package works, along with the need to run `chroot/ host /bin/bash -c 'crio config` has caused it to just fail for me. not sure why yet.


haircommander:
@runcom @kikisdeliveryservice @yuqi-zhang I believe I've done the correct vendoring. I ran `make go-deps` in a  go 1.12 container, but it took along more changes than I expected. please advise if I should do something differently

cgwalters:
Hmm this is a nontrivial PR targeted for 4.4.  First: is this already fixed in 4.5?  Does this really have the severity necessary to ship to 4.4?

Are there any easy workarounds a 4.4 user could do instead?

cgwalters (changes requested):
Also *long term* I think we need to better solve the fundamental structural problem that we have crio config in two places:

 - Embedded in RHCOS as defaults
 - Vendored in the MCO

And what would be much better IMO is to teach the MCO's rendering to use the drop-ins to *only write values that it changed*.

haircommander:
> Hmm this is a nontrivial PR targeted for 4.4. First: is this already fixed in 4.5? Does this really have the severity necessary to ship to 4.4?
> 
Yes, it's not an issue in 4.5 afaict, as 4.5 has a version of cri-o vendored that has this field

> Are there any easy workarounds a 4.4 user could do instead?
either don't use a ctrcfg, or allow `NSS_SDB_USE_CACHE=yes` (the default) which caused issues in the past (I cannot remember what, though, cc @mrunalp )

runcom:
> And what would be much better IMO is to teach the MCO's rendering to use the drop-ins to only write values that it changed.

we recently started doing that if I'm not entirely mistaken, cc @umohnani8

umohnani8:
Yup, in 4.5 we use drop-in files at crio.conf.d that only write the values have been changed. The drop-in files definitely avoid losing any information as everything else falls back to the cri-o defaults.

haircommander:
@cgwalters do you still want changes here?

@runcom PTAL

haircommander:
@cgwalters @runcom @kikisdeliveryservice @yuqi-zhang @umohnani8 @mrunalp PTAL

umohnani8:
LGTM

kikisdeliveryservice:
LGTM but deferring to @cgwalters

kikisdeliveryservice:
Ping @runcom @cgwalters

cgwalters:
Tentative
/approve
but I'd hope most users have upgraded instead to 4.5 at least.

I think someone else on the node team can add the `/lgtm` right?

/retest
Since we're now on agnostic.

OK and it does look like there's at least an e2e that covers the container runtime configs in 4.4: https://github.com/openshift/machine-config-operator/blob/release-4.4/test/e2e/ctrcfg_test.go
so that's good.

sdodson:
@haircommander Why are we backporting a low severity bug to 4.4 which is in maintenance phase of its lifecycle? Either severity needs to be increased or we should close this.

haircommander:
> @haircommander Why are we backporting a low severity bug to 4.4 which is in maintenance phase of its lifecycle which generally only accepts high and urgent severity bugs? Either severity needs to be increased or we should close this.

it's been waiting since 4.4 was not in maintenance phase. However, after talking to @mrunalp and @umohnani8 we've decided it should be at least medium. Losing the NSS_SDB_USE_CACHE env variable in containers can cause instability in nodes

sdodson:
/hold
@haircommander ptal at these failing tests

sdodson:
Clearing stale cherry-pick-approved. I wish there were a way to stop retests without leaving the hold on here but I don't know of such a mechanism. Next Patch Manager please consult with the author to see what can be done to get this to merge successfully next week.

haircommander:
let's see if a rebase fixes it!

haircommander:
also carrying https://github.com/openshift/machine-config-operator/pull/2229

kikisdeliveryservice:
Are we still trying to get this into 4.4?
 /test e2e-gcp-op

haircommander:
yes but I have no idea why the test is failing and haven't had the capacity to check 
/retest

haircommander:
4.4 is EOL, good bye PR

PR #2057, opened 2020-09-03
Bug 1861404: sync Infra in ControllerConfigSpec

kikisdeliveryservice:
Born in 4.1 clusters have Platform and EtcdDiscoveryDomain in their ControllerConfigSpec both deprecated: Platform(Infra.Status.PlatformStatus.Type) and EtcdDiscovery(Infra.Status.EtcdDiscoveryDomain)

As such on upgrade to 4.6 the upgrade was failing as Infra was empty.

This PR syncs Infra in the ControllerConfigSpec to allow the 4.1 clusters to update and upgrade successfully.

### DIFF
--- a/lib/resourcemerge/machineconfig.go
+++ b/lib/resourcemerge/machineconfig.go
@@ -78,6 +78,11 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
 	setBytesIfSet(modified, &existing.KubeAPIServerServingCAData, required.KubeAPIServerServingCAData)
 	setBytesIfSet(modified, &existing.CloudProviderCAData, required.CloudProviderCAData)
 
+	if required.Infra != nil && !equality.Semantic.DeepEqual(existing.Infra, required.Infra) {
+		*modified = true
+		existing.Infra = required.Infra
+	}
+
 	if existing.Infra.Status.PlatformStatus != nil && required.Infra.Status.PlatformStatus != nil {
 		if !equality.Semantic.DeepEqual(existing.Infra.Status.PlatformStatus.Type, required.Infra.Status.PlatformStatus.Type) {
 			*modified = true


lib/resourcemerge/machineconfig.go
```diff
@@ -78,6 +79,37 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
...
 	setBytesIfSet(modified, &existing.CloudProviderCAData, required.CloudProviderCAData)
 
+	// First off: do I need to generally account for Infra changes? Seems like we should?
+	if required.Infra != nil && !equality.Semantic.DeepEqual(existing.Infra, required.Infra) {
+		*modified = true
+		existing.Infra = required.Infra
+	}
+
+	// Moving on... if all Infras are empty (ie born in 4.1 clusters), maybe update Infra with values we _do_ have(?)
+	if required.Infra == nil && existing.Infra == nil {
```
runcom, line 89: `required` can't be empty here as it's the one that we bake into the CC in the MCO sync - `existing` OTOH can indeed be nil if you're coming from 4.1

runcom: in other words, the check at line 83 makes sure, at this point, that none of those 2 are empty anymore. If CC.infra (existing) was nil, line 85 populated it since required is never nil when this code runs

lib/resourcemerge/machineconfig.go
```diff
@@ -78,6 +79,37 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
...
+		// but alsooo from types.go: "etcdDiscoveryDomain is deprecated, use Infra.Status.EtcdDiscoveryDomain instead"
+		if existing.EtcdDiscoveryDomain != "" {
+			*modified = true
+			required.Infra = &configv1.Infrastructure{
+				Status: configv1.InfrastructureStatus{
+					EtcdDiscoveryDomain: existing.EtcdDiscoveryDomain,
+				},
+			}
+		}
+	}
```
runcom, line 111: I'm thinking the infra sync at line 83 should just be enough :thinking: when this code runs the MCO sync function makes sure we have a "good" `required` infra object in the CC and that should be enough to avoid the panic and keep this working and in sync - a live test on a cluster coming from 4.1 will confirm this :woman_dancing:

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: Ahh yes I had these in a diff order and didn't reread them again 😂

lib/resourcemerge/machineconfig.go
runcom: I.E. if we don't have anything at the API level for CC.infra (cause it's nil and never sync), it should be just enough to sync it here with required (which we populate in sync.go) and that's it

lib/resourcemerge/machineconfig.go
```diff
@@ -78,6 +79,37 @@ func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfi
...
+
```
runcom, line 112: we can get rid of the "PlatformStatus.Type" check just below this also - we'll just sync the whole infra object as that's the source of truth anyway

runcom:
Tested with:

- 4.1 to 4.6+this (w/o the two extra conditionals): works like a charm
- 4.1 to 4.6.0-fc.3: panic as expected

pending those comments
/approve

kikisdeliveryservice:
Thanks @runcom will cleanup and re-push!!

lib/resourcemerge/machineconfig.go
```diff
@@ -1,6 +1,7 @@
 package resourcemerge
 
 import (
+	configv1 "github.com/openshift/api/config/v1"
```
runcom, line 4: failing cause this

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: done

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: done

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: done

lib/resourcemerge/machineconfig.go
kikisdeliveryservice: i knoooooo :grimacing: 

Fixed!

PR #2058, opened 2020-09-03
Bug 1875388: Have ovs-configuration.service run before node-valid-hostname

cybertron:
In ipv6 deployments on baremetal we're seeing the following sequence
when deploying nodes:

* node-valid-hostname runs
* node gets a hostname, so node-valid-hostname succeeds
* ovs-configuration runs
* it unconfigures the physical interface so it can be added to the
  bridge. When this happens, NetworkManager removes the hostname that
  was set using the DHCP address on that interface, which puts us
  back at localhost.localdomain
* kubelet starts, sees localhost.localdomain and madness ensues
* the node gets its hostname back using the address now on br-ex

This means all nodes try to register as localhost.localdomain, which
obviously does not work. This change switches the ordering of
ovs-configuration and node-valid-hostname so that ovs-configuration
has to run before node-valid-hostname starts. This means that the
whole "hostname,no hostname,hostname again" dance is done before
node-valid-hostname ever starts, and so the hostname should be
stable from that point on. Since ovs-configuration is not dependent
on the hostname, there shouldn't be any need for it to follow
node-valid-hostname.

### DIFF
--- a/templates/common/_base/units/ovs-configuration.service.yaml
+++ b/templates/common/_base/units/ovs-configuration.service.yaml
@@ -8,8 +8,8 @@ contents: |
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
   Wants=NetworkManager-wait-online.service
-  After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
-  Before=network-online.target kubelet.service crio.service
+  After=NetworkManager-wait-online.service openvswitch.service
+  Before=network-online.target kubelet.service crio.service node-valid-hostname.service
 
   [Service]
   # Need oneshot to delay kubelet


templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -8,8 +8,8 @@ contents: |
   # This service is used to move a physical NIC into OVS and reconfigure OVS to use the host IP
   Requires=openvswitch.service
   Wants=NetworkManager-wait-online.service
-  After=NetworkManager-wait-online.service openvswitch.service node-valid-hostname.service
-  Before=network-online.target kubelet.service crio.service
+  After=NetworkManager-wait-online.service openvswitch.service
+  Before=network-online.target kubelet.service crio.service node-valid-hostname.service
```
trozet, line 12: since ovs-configuration is part of network-online.target, and really node-valid-hostname needs DHCP to be done (regardless of if ovs-configuration is present or not), it seems like it would be better in the node-valid-hostname.service file to do:

After=network-online.target
Wants=network-online.target

cgwalters:
This seems strongly related to https://github.com/openshift/machine-config-operator/pull/2013

cybertron:
> This seems strongly related to #2013

Agreed, thanks for the heads up. I'll comment there.

cgwalters (approved):
TBH what's going on with ovs makes my head spin.  But your commit message is *very* useful and makes sense.

(It does seem to me like a better solution would be teaching NM how to do this operation "transactionally" so that there's no de-configuration phase but clearly we're not doing that in shell script)

cgwalters:
Green CI!  :tada:

templates/common/_base/units/ovs-configuration.service.yaml
cybertron: That reasoning makes sense to me, but we'd also need to change the kubelet service then because it currently waits for network-online too. To avoid reintroducing the race we'd need to have kubelet explicitly depend on node-valid-hostname.

I went back and looked at the original PR for node-valid-hostname, and using network-online was the mechanism chosen to block other services from starting if it failed. I don't know if there was a reason we didn't just add a dependency on node-valid-hostname directly. Maybe @darkmuggle can chime in.

danwinship:
/lgtm
need to get this working again. Feel free to repush before it merges if you want to change the exact way we order the different services

templates/common/_base/units/ovs-configuration.service.yaml
darkmuggle: > I don't know if there was a reason we didn't just add a dependency on node-valid-hostname directly.  

The reason I proposed using `network-online.target` was to provide a blocking mechanism for all the other things that changing the hostname breaks -- logging, cri-o, kubelet, ssh, etc....  The goal was to put to bed all the `localhost` problems by blocking signaling that networking is not really up until we have a valid hostname. In evaluating the problem, `network-online.target` was the lowest blast radius and it didn't require changing all the up-stream services.  The other big reason was for debugging: we block things from coming up before the hostname is valid and that makes it really clear why the boot failed. The pantheon of bugs related to the hostname being cluster-invalid was a bit dizzying.

PR #2060, opened 2020-09-04
Bug 1877448: pkg/daemon/update: allow installing os-extensions on FCOS

vrutkovs:
Support installing OS Extensions on FCOS nodes too.

TODO:
* [x] Rework `validateExtensions` to allow any extensions on FCOS

/cc @LorbusChris @sinnykumari

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -302,29 +302,10 @@ func InSlice(elem string, slice []string) bool {
 	return false
 }
 
-func validateExtensions(exts []string) error {
-	supportedExtensions := []string{"usbguard", "kernel-devel"}
-	invalidExts := []string{}
-	for _, ext := range exts {
-		if !InSlice(ext, supportedExtensions) {
-			invalidExts = append(invalidExts, ext)
-		}
-	}
-	if len(invalidExts) != 0 {
-		return fmt.Errorf("invalid extensions found: %v", invalidExts)
-	}
-	return nil
-
-}
-
 // ValidateMachineConfig validates that given MachineConfig Spec is valid.
 func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
 	if !(cfg.KernelType == "" || cfg.KernelType == KernelTypeDefault || cfg.KernelType == KernelTypeRealtime) {
 		return errors.Errorf("kernelType=%s is invalid", cfg.KernelType)
-	}
-
-	if err := validateExtensions(cfg.Extensions); err != nil {
-		return err
 	}
 
 	if cfg.Config.Raw != nil {
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -338,7 +338,7 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 		// Delete extracted OS image once we are done.
 		defer os.RemoveAll(osImageContentDir)
 
-		if dn.OperatingSystem == MachineConfigDaemonOSRHCOS {
+		if dn.OperatingSystem == MachineConfigDaemonOSRHCOS || dn.OperatingSystem == MachineConfigDaemonOSFCOS {
 			if err := addExtensionsRepo(osImageContentDir); err != nil {
 				return err
 			}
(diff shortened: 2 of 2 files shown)


LorbusChris (approved):

pkg/controller/common/helpers.go
```diff
@@ -302,31 +302,12 @@ func InSlice(elem string, slice []string) bool {
...
-
-}
-
 // ValidateMachineConfig validates that given MachineConfig Spec is valid.
 func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
 	if !(cfg.KernelType == "" || cfg.KernelType == KernelTypeDefault || cfg.KernelType == KernelTypeRealtime) {
 		return errors.Errorf("kernelType=%s is invalid", cfg.KernelType)
 	}
 
-	if err := validateExtensions(cfg.Extensions); err != nil {
```
sinnykumari, line 326: We are validating extensions early enough so that MCC catches it while rendering the new rendered MC and on error prevents from applying it on nodes and getting degraded. We will loose that with this change.

I was thinking we can get runningOS information during MC rendering on MCC pod by looking at /rootfs/etc/os-release and /usr/lib/os-release. @cgwalters @runcom what do you think?

vrutkovs: MCP would become degraded in both cases - this change would make it happen later, as you've mentioned. So I don't see a significant difference from current situation (except that a misconfigured rendered- MC is being generated).

LorbusChris: While I do think it would be better to check early and not generate the rendered MC if invalid, the controller doesn't know what OS it runs on, which makes this a bit hard to do with just a small change.

On a side note: The check for kernelArgs also happens in the MCD, and not at the controller when the MC is rendered.

sinnykumari: @vrutkovs No, MCP doesn't get degraded because a new config doesn't get rendered in case of invalid extensions passed. Verified same.

@LorbusChris we don't validate kernelArgs, we simply apply what user provides in MachineConfig https://github.com/openshift/machine-config-operator/blob/fdf00c3c4c5186d82f77e73a5e04d846436437c6/pkg/daemon/update.go#L828 .

https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/common/helpers.go#L321 is where we do possible MachineConfig related validations.

Since controller runs on one of  the master node and if I am right a cluster won't be running both FCOS and RHCOS nodes on a single cluster, ideally checking the OS on which controller pod is running should give us right OS.

LorbusChris: Right, the kargs are rendered, no matter what they are, but they are checked against the Allowlist with `isArgTunable` before they're getting applied.

That said, I have nothing against making the controller aware of the OS.

sinnykumari: yeah, isArgTunable is something we anyway would like to get away in future and rely fully on applying kargs using MC.

runcom: I think the MCC shouldn't be directly aware of the host itself, sure, it can communicate with the MCD through annotations to inform about the OS but the MCC should really just check the MC at an high level (and go RenderDegraded). Whether an extension is supported on rhcos should be the MCD telling that once it tries to apply it

vrutkovs:
While this approach works, it may accidentally break nodes. Extension list verification happens on MCD, so if invalid extension is found the node is drained and unscheduled already. Its not trivial to recover - one has to change desiredConfig on affected nodes.

Instead we should use MCC host os-release. FCOS/RHCOS nodes cannot be combined, so worker extensions list would be verified according master OS.

sinnykumari:
Based on arguments in the PR and side discusison on slack:
We are leaning towards extensions validation should be done by MCD for each node. We understand that this change can lead to node going degraded if user creates an MC with invalid extensions.  Right now, recovering from a degraded node is painful process but there is an area of improvement in MCO to deal with recovering a degraded node due to bad MC being applied.

We can go as it is with the PR.
/lgtm

sinnykumari:
Another good point from Colin:
we can do validation in the MCC in the future but it does require us to pull the oscontainer right now, OTOH in the future too I think we should serve the oscontainer via HTTP so that would become free

sinnykumari:
Corresponding bug is required

vrutkovs:
/skip

OKD is not installable yet (requires this change + installer PR + machine-os-content update)

PR #2061, opened 2020-09-04
Bug 1868158: machine-config-daemon-pull: Use the MCO image

cgwalters:
Splitting this out from https://github.com/openshift/machine-config-operator/pull/2011 so we can test/review it independently.

I *think* this will work although I find the split between `RenderConfigImages` and `ControllerConfigImages` confusing because it certainly looks like we will use both of them when rendering both `manifests/` and `templates/`.

### DIFF
--- a/pkg/controller/template/constants.go
+++ b/pkg/controller/template/constants.go
@@ -1,6 +1,9 @@
 package template
 
 const (
+	// MachineConfigOperatorKey is our own image used by e.g. machine-config-daemon-pull.service
+	MachineConfigOperatorKey string = "machineConfigOperator"
+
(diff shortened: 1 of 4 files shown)


cgwalters:
Ohh wait I see there's this whole indirection through these constants in the template controller.

cgwalters:
OK rebased :surfer: on https://github.com/openshift/machine-config-operator/pull/2061
(Really we could change the pull to use the renamed `APIServerWatcherKey` but this way is IMO *much* less confusing for the future)

cgwalters:
Linked this one to the BZ from https://github.com/openshift/machine-config-operator/pull/2011

PR #2062, opened 2020-09-06
Bug 1876216: Check if /etc/resolv.conf is present before overriding

Gal-Zaidman:
We copy /var/run/NetworkManager/resolv.conf to /etc/resolv.conf for the case that /etc/resolv.conf is not present, but if it is present we override it.
On oVirt this causes the api-int to not be resolvable for a short period of time, causing failures and events.
This PR adds a check to override cp only if it is not present.

Notice that at the end of this script we replace it with a new file anyway but that is ok.

** Sent this fix to all the platforms that have similar logic to be consistent

Signed-off-by: Gal-Zaidman <gzaidman@redhat.com>

**- What I did**

Added a check to see if /etc/resolv.conf exist before overriding it

Check if /etc/resolv.conf is present before overriding at the beginning of NetworkManager-resolv-prepender

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
@@ -36,7 +36,11 @@ contents:
                 done
             fi
     EOF
-        cp /var/run/NetworkManager/resolv.conf /etc/resolv.conf
+        # Ensure resolv.conf exists before we try to run podman
+        if [[ ! -e /etc/resolv.conf ]]; then
+            cp /var/run/NetworkManager/resolv.conf /etc/resolv.conf
+        fi
+
 
         NAMESERVER_IP=$(/usr/bin/podman run --rm \
             --authfile /var/lib/kubelet/config.json \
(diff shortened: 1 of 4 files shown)


mandre:
All good for OpenStack :+1:

cybertron:
+1 from baremetal too.

cybertron:
/lgtm

We have three of the four affected platforms signing off on this, so I think we're good to go.

cgwalters:
I'm not so sure about this...effectively it's making the generation of `/etc/resolv.conf` happen once.  That's going to cause "drift" if e.g. DHCP gives us a new result right?

The BZ says:

> that means that in those 2 seconds we have the wrong DNS and that will lead to unexpected problems.

Sure but, the platform should be resilient to transient failures right?

cgwalters:
Oh nevermind I see, we do unconditionally overwrite at the end of the script as you mentioend.
/approve
but I feel like this all needs more...engineering and less shell script.  We'll get there.

hardys:
It looks like this breaks some baremetal IPI deployments, since it's possible for resolv.conf to exist like:

```
[core@localhost ~]$ cat /etc/resolv.conf
# Generated by NetworkManager
search local <somedomain>
```

This won't work for external resolution though, so the podman then fails to pull the image and we reintroduce the bug fixed via https://github.com/openshift/machine-config-operator/pull/1765

So I think we either need to partially revert this (for the baremetal platform), or do some smarter check to still overwrite the resolve.conf when it only contains a local search domain and no upstream DNS?

/cc @celebdor @bcrochet @stbenjam

yuqi-zhang:
This appears to break rhel workers: https://bugzilla.redhat.com/show_bug.cgi?id=1879322

PR #2066, opened 2020-09-09
Bug 1854306: Initialize host ovs differently for Openshift-SDN and Ovn-kubernetes by ovs-configuration.service

pliurh:
**- What I did**
Fixes: BZ 1854306
1. Clean NM objects created for ovn-kube shared gateway mode by configure-ovs.sh, when cluster network is openshift-sdn.
2. Render ovs-configuration.service differently for openshift-sdn and ovn-kube.
3. Remove unneeded ovs bridges.

**- How to verify it**
Executing the SDN migration process

Initialize host ovs differently for Openshift-SDN and Ovn-kubernetes

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -4,199 +4,228 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
+    if [ "$1" == "OVNKubernetes" ]; then
+      # Configures NICs onto OVS bridge "br-ex"
+      # Configuration is either auto-detected or provided through a config file written already in Network Manager
+      # key files under /etc/NetworkManager/system-connections/
+      # Managing key files is outside of the scope of this script
 
-    # Configures NICs onto OVS bridge "br-ex"
-    # Configuration is either auto-detected or provided through a config file written already in Network Manager
-    # key files under /etc/NetworkManager/system-connections/
-    # Managing key files is outside of the scope of this script
+      # if the interface is of type vmxnet3 add multicast capability for that driver
+      # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
+      function configure_driver_options {
+        intf=$1
+        driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
+        echo "Driver name is" $driver
+        if [ "$driver" = "vmxnet3" ]; then
+          ifconfig "$intf" allmulti
+        fi
+      }
 
-    # if the interface is of type vmxnet3 add multicast capability for that driver
-    # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
-    function configure_driver_options {
-      intf=$1
-      driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
-      echo "Driver name is" $driver
-      if [ "$driver" = "vmxnet3" ]; then
-        ifconfig "$intf" allmulti
+      NM_CONN_PATH="/etc/NetworkManager/system-connections"
+      iface=""
+      counter=0
+      # find default interface
+      while [ $counter -lt 12 ]; do
+        # check ipv4
+        iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')
+        if [[ -n "$iface" ]]; then
+          echo "IPv4 Default gateway interface found: ${iface}"
+          break
+        fi
+        # check ipv6
+        iface=$(ip -6 route show default | awk '{if ($4 == "dev") print $5; exit}')
+        if [[ -n "$iface" ]]; then
+          echo "IPv6 Default gateway interface found: ${iface}"
+          break
+        fi
+        counter=$((counter+1))
+        echo "No default route found on attempt: ${counter}"
+        sleep 5
+      done
+
+      if [ "$iface" = "br-ex" ]; then
+        # handle vlans and bonds etc if they have already been
+        # configured via nm key files and br-ex is already up
+        ifaces=$(ovs-vsctl list-ifaces ${iface})
+        for intf in $ifaces; do configure_driver_options $intf; done
+        echo "Networking already configured and up for br-ex!"
+        # remove bridges created by openshift-sdn
+        ovs-vsctl --timeout=30 --if-exists del-br br0
+        exit 0
       fi
-    }
 
-    NM_CONN_PATH="/etc/NetworkManager/system-connections"
-    iface=""
-    counter=0
-    # find default interface
-    while [ $counter -lt 12 ]; do
-      # check ipv4
-      iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')
-      if [[ -n "$iface" ]]; then
-        echo "IPv4 Default gateway interface found: ${iface}"
-        break
+      if [ -z "$iface" ]; then
+        echo "ERROR: Unable to find default gateway interface"
+        exit 1
       fi
-      # check ipv6
-      iface=$(ip -6 route show default | awk '{if ($4 == "dev") print $5; exit}')
-      if [[ -n "$iface" ]]; then
-        echo "IPv6 Default gateway interface found: ${iface}"
-        break
+
+      # find the MAC from OVS config or the default interface to use for OVS internal port
+      # this prevents us from getting a different DHCP lease and dropping connection
+      if ! iface_mac=$(<"/sys/class/net/${iface}/address"); then
+        echo "Unable to determine default interface MAC"
+        exit 1
       fi
-      counter=$((counter+1))
-      echo "No default route found on attempt: ${counter}"
-      sleep 5
-    done
 
-    if [ "$iface" = "br-ex" ]; then
-      # handle vlans and bonds etc if they have already been
-      # configured via nm key files and br-ex is already up
-      ifaces=$(ovs-vsctl list-ifaces ${iface})
-      for intf in $ifaces; do configure_driver_options $intf; done
-      echo "Networking already configured and up for br-ex!"
-      exit 0
-    fi
+      echo "MAC address found for iface: ${iface}: ${iface_mac}"
 
-    if [ -z "$iface" ]; then
-      echo "ERROR: Unable to find default gateway interface"
-      exit 1
-    fi
+      # find MTU from original iface
+      iface_mtu=$(ip link show "$iface" | awk '{print $5; exit}')
+      if [[ -z "$iface_mtu" ]]; then
+        echo "Unable to determine default interface MTU, defaulting to 1500"
+        iface_mtu=1500
+      else
+        echo "MTU found for iface: ${iface}: ${iface_mtu}"
+      fi
 
-    # find the MAC from OVS config or the default interface to use for OVS internal port
-    # this prevents us from getting a different DHCP lease and dropping connection
-    if ! iface_mac=$(<"/sys/class/net/${iface}/address"); then
-      echo "Unable to determine default interface MAC"
-      exit 1
-    fi
+      # create bridge
+      if ! nmcli connection show br-ex &> /dev/null; then
+        nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+      fi
 
-    echo "MAC address found for iface: ${iface}: ${iface_mac}"
+      # store old conn for later
+      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
 
-    # find MTU from original iface
-    iface_mtu=$(ip link show "$iface" | awk '{print $5; exit}')
-    if [[ -z "$iface_mtu" ]]; then
-      echo "Unable to determine default interface MTU, defaulting to 1500"
-      iface_mtu=1500
-    else
-      echo "MTU found for iface: ${iface}: ${iface_mtu}"
-    fi
+      # find default port to add to bridge
+      if ! nmcli connection show ovs-port-phys0 &> /dev/null; then
+        nmcli c add type ovs-port conn.interface ${iface} master br-ex con-name ovs-port-phys0
+      fi
 
-    # create bridge
-    if ! nmcli connection show br-ex &> /dev/null; then
-      nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
-    fi
+      if ! nmcli connection show ovs-port-br-ex &> /dev/null; then
+        nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
+      fi
 
-    # store old conn for later
-    old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
+      # bring down any old iface
+      nmcli device disconnect $iface
 
-    # find default port to add to bridge
-    if ! nmcli connection show ovs-port-phys0 &> /dev/null; then
-      nmcli c add type ovs-port conn.interface ${iface} master br-ex con-name ovs-port-phys0
-    fi
+      if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
+        nmcli c add type 802-3-ethernet conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
+          connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu}
+      fi
 
-    if ! nmcli connection show ovs-port-br-ex &> /dev/null; then
-      nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
-    fi
+      nmcli conn up ovs-if-phys0
 
-    # bring down any old iface
-    nmcli device disconnect $iface
-
-    if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
-      nmcli c add type 802-3-ethernet conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
-        connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu}
-    fi
-
-    nmcli conn up ovs-if-phys0
-
-    if ! nmcli connection show ovs-if-br-ex &> /dev/null; then
-      if nmcli --fields ipv4.method,ipv6.method conn show $old_conn | grep manual; then
-        echo "Static IP addressing detected on default gateway connection: ${old_conn}"
-        # find and copy the old connection to get the address settings
-        if egrep -l --include=*.nmconnection $old_conn ${NM_CONN_PATH}/*; then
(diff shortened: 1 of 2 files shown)


pliurh:
@trozet @cybertron @abhat PTAL

cgwalters:
What would help here I think is to teach NM to have a way to make *transient* changes to the network setup i.e. don't persist into `/etc`.  Then we'd run `configure-ovs.sh` at each boot, and wouldn't need a cleanup script.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -175,6 +175,7 @@ contents:
         echo "OVS successfully configured"
         ip a show br-ex
         configure_driver_options ${iface}
+        mv /etc/keepalived/keepalived.conf /etc/keepalived/keepalived.conf.bak
```
cybertron, line 178: I think this will fail on platforms that don't use keepalived.

cybertron (changes requested):
> What would help here I think is to teach NM to have a way to make _transient_ changes to the network setup i.e. don't persist into `/etc`. Then we'd run `configure-ovs.sh` at each boot, and wouldn't need a cleanup script.

@bcrochet was working on a way to make transient network config changes that might be relevant.

One concern with the keepalived workaround inline. Ideally we would remove that once my fix is in anyway, but if it unblocks things in the meantime I guess it's fine.

templates/common/_base/files/configure-ovs-network.yaml
bcrochet: > > What would help here I think is to teach NM to have a way to make _transient_ changes to the network setup i.e. don't persist into `/etc`. Then we'd run `configure-ovs.sh` at each boot, and wouldn't need a cleanup script.
> 
> @bcrochet was working on a way to make transient network config changes that might be relevant.
> 
> One concern with the keepalived workaround inline. Ideally we would remove that once my fix is in anyway, but if it unblocks things in the meantime I guess it's fine.

Proposal here: openshift/enhancements#399
PR here: #2017

pliurh: > I think this will fail on platforms that don't use keepalived.

fixed.

pliurh:
> What would help here I think is to teach NM to have a way to make _transient_ changes to the network setup i.e. don't persist into `/etc`. Then we'd run `configure-ovs.sh` at each boot, and wouldn't need a cleanup script.

This is not only about NM objects. In ovs-configuration.sh, we use NM to create ovs bridge and interfaces, which are stored in ovsdb and also need to be removed properly.

pliurh:
PR: #2017 cannot help this case. 
As we need to configure the OVS before kubelet is up, we cannot wait for nm-state to do the job.

templates/common/_base/files/clean-ovs-network.yaml
```diff
@@ -0,0 +1,60 @@
...
+    #!/bin/bash
+    set -eux
+
+    # Revert changes made by /usr/local/bin/configure-ovs.sh.
+    # Remove OVS bridge "br-ex". Use the default NIC for cluster network.
+
+    iface=""
+    counter=0
+    # find default gateway interface
+    while [ $counter -lt 12 ]; do
```
danwinship, line 14: This should not be necessary; the setup side needs to deal with the possibility that the first time it runs, NM might be partway through bringing the network up for the first time, and so we don't know which interface will end up being the default. But in the cleanup case, either `br-ex` and `ovs-port-phys0` and the rest already exist as NetworkManager objects, and you know what needs to be done with them; or else they don't exist, and there's no cleanup to be done.

templates/common/_base/files/clean-ovs-network.yaml
```diff
@@ -0,0 +1,60 @@
+mode: 0755
+path: "/usr/local/bin/clean-ovs.sh"
```
danwinship, line 2: Given how the scripts are closely connected (needing to know the same interface names, etc) it seems like it might be better to put the setup and cleanup both into a single file, and run it as either `configure-ovs.sh setup` or `configure-ovs.sh cleanup` depending on which mode you need?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -175,6 +175,11 @@ contents:
         echo "OVS successfully configured"
         ip a show br-ex
         configure_driver_options ${iface}
+        if [ -e /etc/keepalived/keepalived.conf ]; then
+          # Stop keepalived after switching the default interface.
```
danwinship, line 179: Would it be better to make keepalived be `After: network-online.target` ?

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -1,5 +1,5 @@
 name: ovs-configuration.service
-enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else}}false{{end}}
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
```
danwinship, line 2: according to the template docs, you should be able to say

    enabled: {{if eq .NetworkType "OVNKubernetes" "OpenShiftSDN"}}true{{else}}false{{end}}

templates/common/_base/files/clean-ovs-network.yaml
danwinship: or, if you're going to merge https://github.com/openshift/cluster-network-operator/pull/782 into this, then maybe just `configure-ovs.sh {{.NetworkType}}`, where `configure-ovs.sh OpenShiftSDN` would delete `br-ex` if it existed, and `configure-ovs.sh OVNKubernetes` would delete `br0` if it existed and set up `br-ex`.

templates/common/_base/files/clean-ovs-network.yaml
```diff
@@ -0,0 +1,60 @@
...
+    if nmcli connection show ovs-port-br-ex &> /dev/null; then
+      nmcli c del ovs-port-br-ex
+    fi
+
+    if nmcli connection show ovs-if-br-ex &> /dev/null; then
+      nmcli c del ovs-if-br-ex
+    fi
+
+    nmcli c del br-ex
+    nmcli device connect $iface
```
trozet, line 60: is this part necessary? The other connections that are lower priority should automatically become active with iface after ovs-if-phys0 gets deleted.

templates/common/_base/files/configure-ovs-network.yaml
pliurh: keepalived runs in static pod. So it is `After: network-online.target`. However, the `keepalived-monitor` depends on `kube-apiserver` to generate the correct `keepalived.conf`.  We cannot control the order of the starting of static pods. So when keepalived is up before kube-apiserver without a up-to-date config, it will mistakenly set the VIP to the old interface and break the network connectivity. Then kube-apiserver cannot come up as expected. We end up in a  dead lock.

The logic here is just a workaround for this case, so that keepalived can not start before kube-apiserver. Ideally, keepalived side shall be able to handle this situation properly after fixing https://bugzilla.redhat.com/show_bug.cgi?id=1873955. But before that is available, this workaround can work.

templates/common/_base/files/clean-ovs-network.yaml
pliurh: fixed.

templates/common/_base/files/clean-ovs-network.yaml
pliurh: fixed.

templates/common/_base/units/ovs-configuration.service.yaml
pliurh: fixed.

templates/common/_base/files/clean-ovs-network.yaml
pliurh: fixed.

pliurh:
/hold
wait until https://github.com/openshift/machine-config-operator/pull/2085 merged

pliurh:
/unhold
The keepalived issue has been fixed by https://github.com/openshift/baremetal-runtimecfg/pull/100

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -4,199 +4,230 @@ contents:
...
+      # Managing key files is outside of the scope of this script
+
+      # if the interface is of type vmxnet3 add multicast capability for that driver
+      # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
+      function configure_driver_options {
+        intf=$1
+        driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
+        echo "Driver name is" $driver
+        if [ "$driver" = "vmxnet3" ]; then
+          ifconfig "$intf" allmulti
```
rbbratta, line 20: Isn't `ifconfig` deprecated?

```bash
ip link set dev "$intf" allmulticast on
```

rbbratta:
could we run scripts through shellcheck?   https://www.shellcheck.net

```

In configure-ovs-network.sh line 14:
    echo "Driver name is" $driver
                          ^-----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    echo "Driver name is" "$driver"


In configure-ovs-network.sh line 45:
    ifaces=$(ovs-vsctl list-ifaces ${iface})
                                   ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    ifaces=$(ovs-vsctl list-ifaces "${iface}")


In configure-ovs-network.sh line 46:
    for intf in $ifaces; do configure_driver_options $intf; done
                                                     ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    for intf in $ifaces; do configure_driver_options "$intf"; done


In configure-ovs-network.sh line 76:
    nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
                                                                                                                                      ^----------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address "${iface_mac}"


In configure-ovs-network.sh line 80:
  old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
                                                                  ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep "${iface}" | awk '{print $1}')


In configure-ovs-network.sh line 84:
    nmcli c add type ovs-port conn.interface ${iface} master br-ex con-name ovs-port-phys0
                                             ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    nmcli c add type ovs-port conn.interface "${iface}" master br-ex con-name ovs-port-phys0


In configure-ovs-network.sh line 92:
  nmcli device disconnect $iface
                          ^----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  nmcli device disconnect "$iface"


In configure-ovs-network.sh line 95:
    nmcli c add type 802-3-ethernet conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
                                                   ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    nmcli c add type 802-3-ethernet conn.interface "${iface}" master ovs-port-phys0 con-name ovs-if-phys0 \


In configure-ovs-network.sh line 102:
    if nmcli --fields ipv4.method,ipv6.method conn show $old_conn | grep manual; then
                                                        ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    if nmcli --fields ipv4.method,ipv6.method conn show "$old_conn" | grep manual; then


In configure-ovs-network.sh line 105:
      if egrep -l --include=*.nmconnection $old_conn ${NM_CONN_PATH}/*; then
         ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.
                                           ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      if egrep -l --include=*.nmconnection "$old_conn" ${NM_CONN_PATH}/*; then


In configure-ovs-network.sh line 106:
        old_conn_file=$(egrep -l --include=*.nmconnection $old_conn ${NM_CONN_PATH}/*)
                        ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.
                                                          ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        old_conn_file=$(egrep -l --include=*.nmconnection "$old_conn" ${NM_CONN_PATH}/*)


In configure-ovs-network.sh line 111:
        nmcli conn clone ${old_conn} ${old_conn}-clone
                         ^---------^ SC2086: Double quote to prevent globbing and word splitting.
                                     ^---------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        nmcli conn clone "${old_conn}" "${old_conn}"-clone


In configure-ovs-network.sh line 124:
      cp -f ${old_conn_file} ${new_conn_file}
            ^--------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      cp -f "${old_conn_file}" ${new_conn_file}


In configure-ovs-network.sh line 126:
        nmcli conn delete ${old_conn}-clone
                          ^---------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        nmcli conn delete "${old_conn}"-clone


In configure-ovs-network.sh line 127:
        rm -f ${old_conn_file}
              ^--------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        rm -f "${old_conn_file}"


In configure-ovs-network.sh line 161:
        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
                                                                                       ^----------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        ovs-if-br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address "${iface_mac}"


In configure-ovs-network.sh line 173:
      configure_driver_options ${iface}
                               ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      configure_driver_options "${iface}"


In configure-ovs-network.sh line 187:
      configure_driver_options ${iface}
                               ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      configure_driver_options "${iface}"


In configure-ovs-network.sh line 201:
  nmcli conn up $old_conn
                ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  nmcli conn up "$old_conn"

```
For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2196 -- egrep is non-standard and depreca...

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -4,199 +4,230 @@ contents:
...
-          sed -i '/^\[connection\]$/,/^\[/ s/^interface-name=.*$/interface-name=br-ex/' ${new_conn_file}
+          echo "old connection file found at: ${old_conn_file}"
+          new_conn_file=${NM_CONN_PATH}/ovs-if-br-ex.nmconnection
+          if [ -f "$new_conn_file" ]; then
+            echo "WARN: existing br-ex interface file found: $new_conn_file, which is not loaded in NetworkManager...overwriting"
+          fi
+          cp -f ${old_conn_file} ${new_conn_file}
+          if $cloned; then
+            nmcli conn delete ${old_conn}-clone
+            rm -f ${old_conn_file}
```
rbbratta, line 131: use guards and always quote args to `rm`


bad
```bash
old_conn_file='-r .*'
rm -f ${old_conn_file}
```
better
```bash
old_conn_file='-r .*'
rm -f -- "${old_conn_file}"
```

pliurh:
@rbbratta The diff result of Github is a bit misleading here. You can have more clear diff result with vscode. Actually, this patch didn't modify the lines you comment on. This patch only adds the following logic to this file. I prefer we focus on the new code in this PR and address your comments in a separate one.

```
if [ "$1" == "OVNKubernetes" ]; then
  <old logic - unchanged>
elif [ "$1" == "OpenShiftSDN" ]; then
  <new logic of removing NM objects>
fi
```

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -4,199 +4,230 @@ contents:
...
+      # Configures NICs onto OVS bridge "br-ex"
+      # Configuration is either auto-detected or provided through a config file written already in Network Manager
+      # key files under /etc/NetworkManager/system-connections/
+      # Managing key files is outside of the scope of this script
+
+      # if the interface is of type vmxnet3 add multicast capability for that driver
+      # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
+      function configure_driver_options {
+        intf=$1
+        driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
```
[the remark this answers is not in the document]
trozet: this isn't new code. He's just moving it into the if statement so this PR isn't really fixing that

trozet (approved):
/lgtm

trozet:
@kikisdeliveryservice or @ashcrow can you please approve?

templates/common/_base/units/ovs-configuration.service.yaml
```diff
@@ -1,5 +1,5 @@
 name: ovs-configuration.service
-enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else}}false{{end}}
+enabled: {{if eq .NetworkType "OVNKubernetes" "OpenShiftSDN"}}true{{else}}false{{end}}
```
cgwalters, line 2: Is this ever `false` now?

cgwalters (approved):
I really hope after 4.7 starts we can approach this in a much more structured fashion that involves more testing, much more observability/debuggability and much less shell script.

templates/common/_base/units/ovs-configuration.service.yaml
pliurh: CNO also supports kuryr and other 3-party CNIs.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -4,199 +4,228 @@ contents:
...
-    nmcli conn down ovs-if-phys0
-    nmcli conn up $old_conn
-    exit 1
+
+      if nmcli connection show br-ex &> /dev/null; then
+        nmcli c del br-ex
+      fi
+
+      # remove bridges created by ovn-kubernetes, try to delete br-ex again in case NM fail to talk to ovsdb
+      ovs-vsctl --timeout=30 --if-exists del-br br-int -- --if-exists del-br br-local -- --if-exists del-br br-ex
```
trozet, line 230: @pliurh one thing I just thought of is that, if you are using openshift-sdn and you reboot a node, this will run again and delete these bridges. Does openshift-sdn also use br-int bridge? Would this cause problems for openshift-sdn to delete the bridge on each reboot?

@danwinship ^ ?

danwinship: openshift-sdn does not use `br-int`, it only uses `br0`. And it will end up destroying and recreating br0 if OVS brings it up after a reboot anyway. It doesn't want OVS preserving its state across reboots.

PR #2067, opened 2020-09-09
Bug 1865839: daemon/update: print error causing rollback

jlebon:
Right now, we `cleanup -p` if we hit an error. But if we hit another
error during the cleanup, we lose what the original error was. Let's log
it to help debugging.

Related: rhbz#1865839


ashcrow (approved):

ashcrow:
Errors were due to the registry throwing 500s.

/test e2e-gcp-upgrade

kikisdeliveryservice:
@jlebon If you need this for the BZ can you move the BZ back to modified (it's on QA now)

PR #2068, opened 2020-09-09
Bug 1865839: daemon: Set RPMOSTREE_CLIENT_ID again

cgwalters:
This way we know in the systemd journal that it's the MCO making
changes, as distinct from any other process running `rpm-ostree`
as a binary.

I added this in 1e08e0c02a692d7749651636e23d6f56c0191be9
but then it was removed in a99a12315e422bc46cf95d3b9dcceeb87bd1f69e

Came up as part of https://bugzilla.redhat.com/show_bug.cgi?id=1865839
so we can more easily identify which other project is calling rpm-ostree.

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -107,6 +107,9 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	// To help debugging, immediately log version
 	glog.Infof("Version: %+v (%s)", version.Raw, version.Hash)
 
+	// See https://github.com/coreos/rpm-ostree/pull/1880
+	os.Setenv("RPMOSTREE_CLIENT_ID", "machine-config-operator")
+
 	onceFromMode := startOpts.onceFrom != ""
 	if !onceFromMode {
 		// in the daemon case


sinnykumari:
Thanks Colin for finding the cause.
I missed the importance of why we were setting RPMOSTREE_CLIENT_ID and missed adding back in https://github.com/openshift/machine-config-operator/commit/a99a12315e422bc46cf95d3b9dcceeb87bd1f69e

/lgtm

jlebon:
Hmm, I think the bot thinks that this will close rhbz#1865839. Should we demote the bz reference here to a "Related" instead?

cgwalters:
Just as a sanity check, you can see this in the node journal for the workers in the e2e-gcp-op tests, looks like:
`Sep 09 23:27:08.986843 ci-op-9mslp6yb-1354f-qb5vt-worker-b-kjfnw rpm-ostree[2600]: client(id:machine-config-operator dbus:1.80 unit:crio-b2f39d633469e8b088fd58df152f763a2c040cbc4deb58de0fee5d45a003cb43.scope uid:0) added; new total=1`

PR #2070, opened 2020-09-10
[WIP] Bug 1875005: On-Prem: No longer check for apiserver in keepalived

mandre:
There is no good reason for keepalived to query the kube-apiserver for
its liveliness since all requests are proxied by haproxy anyway that
balances the requests among backends in a roundrobin fashion. Keepalived
should only care about the status of the local HAProxy instance, and
HAProxy should be the one to care about the backend status.

Having keepalived check for API status causes frequent failovers and
connection drops when the API is a bit busy.

### DIFF
--- a/templates/master/00-master/openstack/files/openstack-keepalived-keepalived.yaml
+++ b/templates/master/00-master/openstack/files/openstack-keepalived-keepalived.yaml
@@ -8,7 +8,7 @@ contents:
     }
 
     vrrp_script chk_ocp {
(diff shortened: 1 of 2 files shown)


mandre:
Hmm, we still need to look at the api status otherwise we may move the VIP to the master nodes too early and bootstrap doesn't finish.

I just found out it might have already been addressed on BM with https://github.com/openshift/machine-config-operator/pull/1893/.

/close

PR #2071, opened 2020-09-10
Dedupe on-prem bootstrap manifests

bcrochet:
**- What I did**
Currently, all of the on-prem platforms duplicate the manifests needed for
bootstrap. There is a small difference in all of them, and these can be
gotten rid of by moving some of the template code into functions in Go.

**- How to verify it**
Verify that on-prem platforms can still deploy.

Dedupe bootstrap manifests for on-prem platforms.


cgwalters (approved):
/approve
the idea in general (and we should likely do the same with the kubelet systemd unit).

But this is a nontrivial change without an immediate bug it's fixing (right?) so I'm adding the 4.7 label.

bcrochet:
> /approve
> the idea in general (and we should likely do the same with the kubelet systemd unit).
> 
> But this is a nontrivial change without an immediate bug it's fixing (right?) so I'm adding the 4.7 label.
> 

You are correct. Thanks!

pkg/operator/render.go
```diff
@@ -156,3 +160,79 @@ func isSingleStackIPv6(serviceCIDRs []string) (bool, error) {
...
+		case configv1.BareMetalPlatformType:
+			return cfg.Infra.Status.PlatformStatus.BareMetal.NodeDNSIP, nil
+		case configv1.OvirtPlatformType:
+			return cfg.Infra.Status.PlatformStatus.Ovirt.NodeDNSIP, nil
+		case configv1.OpenStackPlatformType:
+			return cfg.Infra.Status.PlatformStatus.OpenStack.NodeDNSIP, nil
+		case configv1.VSpherePlatformType:
+			return cfg.Infra.Status.PlatformStatus.VSphere.NodeDNSIP, nil
+		default:
+			return nil, fmt.Errorf("invalid platform for API VIP")
```
cybertron, line 214: s/API VIP/DNS IP/

pkg/operator/render.go
```diff
@@ -156,3 +160,79 @@ func isSingleStackIPv6(serviceCIDRs []string) (bool, error) {
...
+		case configv1.BareMetalPlatformType:
+			return cfg.Infra.Status.PlatformStatus.BareMetal.IngressIP, nil
+		case configv1.OvirtPlatformType:
+			return cfg.Infra.Status.PlatformStatus.Ovirt.IngressIP, nil
+		case configv1.OpenStackPlatformType:
+			return cfg.Infra.Status.PlatformStatus.OpenStack.IngressIP, nil
+		case configv1.VSpherePlatformType:
+			return cfg.Infra.Status.PlatformStatus.VSphere.IngressIP, nil
+		default:
+			return nil, fmt.Errorf("invalid platform for API VIP")
```
cybertron, line 195: s/API VIP/Ingress VIP/

pkg/operator/render.go
```diff
@@ -156,3 +160,79 @@ func isSingleStackIPv6(serviceCIDRs []string) (bool, error) {
...
+		}
+	} else {
+		return nil, fmt.Errorf("")
+	}
+}
+
+func onPremPlatformNodeDNSIP(cfg mcfgv1.ControllerConfigSpec) (interface{}, error) {
```
cybertron, line 202: Is this used? I'm not sure we need it now that we removed the DNS VIP.

cybertron (changes requested):
Cool, this is a big improvement! A couple of things inline.

pkg/operator/render.go
bcrochet: Yeah. I haven't seen it in any manifests/templates. I'll remove it.

ravidbro:
I'm thrilled to see that.
Are we planning to do the same also for the templates?

bcrochet:
> I'm thrilled to see that.
> Are we planning to do the same also for the templates?

See #2079

cybertron (approved):
/lgtm

cybertron:
/lgtm cancel

I don't think we want the bot to sit here and retest until 4.7 opens.

cgwalters:
Thanks for pushing this through!

PR #2072, opened 2020-09-10
feat: block upgrades until worker pool completes

ericavonb:
Wait to report an upgrade is complete to the CVO until the worker
pool has also finished updating its node. Currently, we only wait
on the master pool. This has caused confusion for users and led
to problems when admins try to upgrade later.

This change adds the "required-for-upgrade" annotation to the
worker pool manifests. This will cause the MCO to wait until it
has completed before reporting done, like it does for the master
pool currently.

The MCO will not wait on custom pools unless users add a similar
annotation.

Issue: [GRPA-2779](https://issues.redhat.com/browse/GRPA-2779)

----

This PR is more of an exploration of the issue. Ideally, we'd like to
block further upgrades until the worker pool completes or only
when it is marked degraded, while reporting `progressing=false`
after the control-plane completes.


kikisdeliveryservice:
superseded by: https://github.com/openshift/machine-config-operator/pull/2231

PR #2073, opened 2020-09-10
[baremetal and friends] Drop unnecessary readiness probes

cybertron:
Now that we are using the loadbalanced endpoint for keepalived
healthchecks, it is not correct to be using the local apiserver for
readiness checks. Also, I don't think there's any need for readiness
checks on the monitor. Kubernetes isn't routing any traffic to this
pod, so its ready status is meaningless.

### DIFF
--- a/manifests/on-prem/coredns.yaml
+++ b/manifests/on-prem/coredns.yaml
@@ -63,16 +63,6 @@ spec:
     volumeMounts:
     - name: conf-dir
       mountPath: "/etc/coredns"
-    readinessProbe:
-      httpGet:
-        path: /health
-        port: 18080
-        scheme: HTTP
-      initialDelaySeconds: 10
-      periodSeconds: 10
-      successThreshold: 1
-      failureThreshold: 3
-      timeoutSeconds: 10
     livenessProbe:
       httpGet:
         path: /health
--- a/manifests/on-prem/keepalived.yaml
+++ b/manifests/on-prem/keepalived.yaml
@@ -136,16 +136,6 @@ spec:
       mountPath: "/var/run/keepalived"
     - name: manifests
       mountPath: "/opt/openshift/manifests"
-    readinessProbe:
-      httpGet:
-        path: /readyz
-        port: 6443
-        scheme: HTTPS
-      initialDelaySeconds: 10
-      periodSeconds: 10
-      successThreshold: 1
-      failureThreshold: 3
-      timeoutSeconds: 10
     imagePullPolicy: IfNotPresent
   hostNetwork: true
   tolerations:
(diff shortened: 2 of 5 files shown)


cybertron:
This can probably wait until 4.7 opens up. It's not causing any problems that I know of, I just noticed it as I was doing some other work with keepalived.

yboaron:
I think you can delete also delete the readiness probe for baremetal Keepalived in the bootstrap https://github.com/openshift/machine-config-operator/blob/master/manifests/baremetal/keepalived.yaml#L129 , isn't it ?

I'm not sure if there's a K8S service that abstract the coredns static pods (I think the answer is no), so if there's no such a service I guess we can delete also the readiness probes from coredns static pod in all the platforms: 
https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/baremetal-coredns.yaml#L66

cybertron:
I extended this change to all platforms and included coredns in the latest revision.That gets rid of all the readinessProbes in the templates and manifests directory AFAICT.

cybertron:
One of these days I will remember to run make update before pushing MCO patches...

jcpowermac:
openshift-tests might not pass but installed correctly
/lgtm

kikisdeliveryservice:
waiting on lgtm from @rgolangh

mandre (commented):
Just a heads up that the cloud powering OpenStack CI is down for maintenance and it's expected that e2e-openstack fails.

Patch looks good to me.

kikisdeliveryservice:
Do we need this ovn job passing for this pr?

cybertron:
I don't think so. It looks like that runs on AWS, which wouldn't be affected by this patch.

sjenning:
@cybertron what is holding this up? Other than a rebase.  I'm very much wanting the warning on my web console to go away :smile:

cybertron:
It was waiting for Roy's lgtm, but given that it's not really a platform-specific change I think we can probably go ahead at this point. However, I see now that it is in merge conflict so I'll need to get that addressed today.

sjenning:
huge failures across CI due to new docker hub rate limiting on busybox image pull
```
docker.io/library/busybox: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

mandre (commented):
/lgtm

kikisdeliveryservice:
one more try, test seems red

/test e2e-ovirt

kikisdeliveryservice (approved):

kikisdeliveryservice:
@rgolangh the ovirt job seems very flaky, any insight?

mandre:
/retest
The ovirt job passed, can we merge this?

PR #2075, opened 2020-09-10
Bug 1823950: Reverse haproxy and keepalived check timings

cybertron:
Since we moved to keepalived healthchecking against haproxy, we want
haproxy to handle most failures so the VIP doesn't have to move.
However, previously the time it took for haproxy to recognize an
outage on a node was longer than it was for keepalived, which resulted
in the VIP moving before haproxy removed the failing backend.

This change makes the haproxy interval 1 second, so it should notice
outages in 2 seconds or less (because it has a fall value of 2).
The keepalived interval is changed to 2, which means it will detect
failures in 2 to 4 seconds (also a fall value of 2). This means
haproxy should deal with api outages before keepalived does and
allow the VIP to stay on the same node.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
+++ b/templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
@@ -17,16 +17,16 @@ contents:
     # api to take the VIP. This isn't preferred because it means all api
     # traffic will go through one node, but at least it keeps the api available.
     vrrp_script chk_ocp_lb {
-        script "/usr/bin/timeout 0.9 /etc/keepalived/chk_ocp_script.sh"
-        interval 1
+        script "/usr/bin/timeout 1.9 /etc/keepalived/chk_ocp_script.sh"
+        interval 2
         weight 20
         rise 3
         fall 2
     }
 
     vrrp_script chk_ocp_both {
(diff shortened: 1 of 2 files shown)


templates/master/00-master/baremetal/files/baremetal-keepalived-keepalived.yaml
```diff
@@ -18,15 +18,15 @@ contents:
     # traffic will go through one node, but at least it keeps the api available.
     vrrp_script chk_ocp_lb {
         script "/usr/bin/timeout 0.9 /etc/keepalived/chk_ocp_script.sh"
```
cybertron, line 20: Hmm, I should change this timeout too.

yboaron:
@cybertron  Don't we need this fix for 4.6?  I can see that the PR was labeled with 4.7

cybertron:
> @cybertron Don't we need this fix for 4.6? I can see that the PR was labeled with 4.7

Yes. I think the label got added because I pushed it without a bug reference initially. @kikisdeliveryservice Are you okay with dropping the 4.7 label? This is needed to complete a 4.6 bug fix.

kikisdeliveryservice:
go for it! :smile: 

(there originally wasn't a bz attached to this)

kikisdeliveryservice (approved):

PR #2076, opened 2020-09-10
Bug 1849543: Use mc name as finalizer for kubelet config

umohnani8:
Fixes #1849543 (https://bugzilla.redhat.com/show_bug.cgi?id=1849543) 

**- What I did**
The uid was being used as the entry for the finalizer list
but everytime an update happens it generates a new uid, which
is why we were ending up with multiple uids in the finalizers
list. So use the mc name for the list of finalizer instead as
the mc name stays the same for each kubelet config.

**- How to verify it**
1) Start a cluster with this patch and create a kubeletconfig CR. The description of the CR should not have duplicate entries in the list of finalizers.
2) Start a cluster without this patch and create a kubeletconfig CR - there will be multiple entries in the list of finalizers. Upgrade the cluster to a version with this patch, the list of finalizers should only have entries for the pools that the CR is applied to and no duplicates.

Use MC name for the list of finalizers to avoid duplicate and multiple entries for the same MCP

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/clarketm/json"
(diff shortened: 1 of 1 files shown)


umohnani8:
@rphillips @yuqi-zhang PTAL

rphillips:
Nice!

/lgtm

yuqi-zhang (approved):

PR #2079, opened 2020-09-11
Dedupe on-prem templates

bcrochet:
There are a number of duplicated files across the on-prem platforms.
This patch begins the process of de-duplicating them. This is going
to be a bigger task than the bootstrap manifests, as not all of the
files are duplicated, and some have differences that would not lend
to being deduped. But this at least starts small, and proves the
concept.

**- What I did**

**- How to verify it**

Dedupe templates for on-prem platforms


cgwalters (approved):
Only skimmed this I think the lines deleted count looks awesome!

ravidbro:
I put /hold to stop the bot from testing every hour until the conflict will be resolved

PR #2081, opened 2020-09-14
bug 1878129: add node collection for oc adm inspect

deads2k:
MCO holds kubelet config, which leads to node creation.  This adds node collection for related resources because the MCO configuration has a direct impact on the node resources.

### DIFF
--- a/install/0000_80_machine-config-operator_06_clusteroperator.yaml
+++ b/install/0000_80_machine-config-operator_06_clusteroperator.yaml
@@ -23,3 +23,5 @@ status:
       resource: kubeletconfigs
     - group: machineconfiguration.openshift.io
       resource: containerruntimeconfigs
+    - group: ""
+      resource: nodes
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -70,6 +70,8 @@ func (optr *Operator) syncRelatedObjects() error {
 		{Group: "machineconfiguration.openshift.io", Resource: "kubeletconfigs"},
 		{Group: "machineconfiguration.openshift.io", Resource: "containerruntimeconfigs"},
 		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigs"},
+		// gathered because the machineconfigs created container bootstrap credentials and node configuration that gets reflected via the API and is needed for debugging
+		{Group: "", Resource: "nodes"},
 	}
 
 	if !equality.Semantic.DeepEqual(coCopy.Status.RelatedObjects, co.Status.RelatedObjects) {


yuqi-zhang:
Just to make sure, this should effectively only affect must-gather (via `oc adm inspect clusteroperators`) and not any other MCO functionality right?

Does this also mean the MCO is now the effective "owner" of a node object? In normal cluster operation does anyone "sync" node objects?

kikisdeliveryservice:
It doesn't really make sense to me to have the nodes grouped under MCO instead of a first class resource like @cgwalters saidin the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1878129#c6

I think this needs more discussion bc this doesn't feel like the right choice.

/hold

deads2k:
> Does this also mean the MCO is now the effective "owner" of a node object? In normal cluster operation does anyone "sync" node objects?

It doesn't mean that at all.  This just means the data is gathered so that the node team can debug.


> Just to make sure, this should effectively only affect must-gather (via `oc adm inspect clusteroperators`) and not any other MCO functionality right?

correct.

deads2k:
> It doesn't really make sense to me to have the nodes grouped under MCO instead of a first class resource like @cgwalters saidin the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1878129#c6

Resources are inspected and related resources gathered. The method to hardcode collection is to add them to a relatedresource.  Doing it in generic client handling (used by RHACM for instance or for inspecting a customer namespace) does not make sense.

/hold cancel

yuqi-zhang (approved):
If its only for collection purposes I am +1

Also the BZ is targetted 4.7, did you want this in 4.6

deads2k:
> If its only for collection purposes I am +1
> 
> Also the BZ is targetted 4.7, did you want this in 4.6

Yes. Looks like someone may not have understood the impact and ease of the change and moved it out.

cgwalters:
Are there written down rules around what data `oc adm inspect` should gather?  Node objects today include the pulled images, which could indirectly leak some customer data potentially.  Maybe that happens elsewhere too.

Anyways tentative
/approve
but I still wonder whether this should just be something that `oc adm inspect` gathers on its own by default, particularly if it's just `oc adm inspect clusteroperator` and not specifically `oc adm inspect clusteroperator/machine-config`.

(On the flip side though debugging a lot of MCO problems is going to really want the node objects too)

deads2k:
> Are there written down rules around what data `oc adm inspect` should gather? Node objects today include the pulled images, which could indirectly leak some customer data potentially. Maybe that happens elsewhere too.

The enhancement describes what is collected: https://github.com/openshift/enhancements/blob/master/enhancements/oc/inspect.md#proposal 

It also includes information about what data is elided from output.  Since the images would also be listed in the pods themselves, I don't see image metadata as different.

deads2k:
> but I still wonder whether this should just be something that `oc adm inspect` gathers on its own by default, particularly if it's just `oc adm inspect clusteroperator` and not specifically `oc adm inspect clusteroperator/machine-config`.

Inspect is a tool for pulling information related to what is requested. That relationship should be described by the resources themselves, especially since usage like `oc adm inspect ns/foo` is a valid, and very useful use-case for someone trying to get information about an NS.

We will not add information about particular resources for particular operators, since the operators have a structured spot to include the relationship.

If the MCO isn't able to add this information, group-b can provide a stop-gap, but it would be really disappointing if an additional piece of debugging information, most tightly coupled to the MCO wasn't able to be added where it is needed

cgwalters:
> It also includes information about what data is elided from output. Since the images would also be listed in the pods themselves, I don't see image metadata as different.

I don't quite see this listed in the enhancement but basically - we're not gathering *user* projects (i.e. non-`openshift-*` namespaces) and the pods from those right?  But enumerating nodes includes images from *user* pods too (images I think would generally be less concerning than pods but still)

(This data probably leaks into other operators or pod logs; it's not quite clear to me that `oc adm inspect` explicitly has this as a goal, but I think it should)

cgwalters:
/lgtm
based on above and Jerry's approval
/bugzilla refresh

PR #2084, opened 2020-09-14
Bug 1872885: always check hostname char count

darkmuggle:
hostnamectl previously refused to assign a long hostname. Since
hostnamectl now allows for longer hostnames, we need to alway check if
the hostname is too long.

Fixes BZ #1872885

### DIFF
--- a/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
+++ b/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
@@ -15,7 +15,8 @@ contents:
         exit 0
     fi
 
-    if [[ ! "$(< /proc/sys/kernel/hostname)" =~ (localhost|localhost.localdomain) ]]; then
+    kn="$(< /proc/sys/kernel/hostname)"
+    if [[ ! "${kn}" =~ (localhost|localhost.localdomain) ]] && [ "${#kn}" -le 63 ]; then
         log "hostname is already set"
         exit 0
     fi
@@ -29,4 +30,3 @@ contents:
     if [ -n "${host_name}" ]; then
         set_valid_hostname "${host_name}"
     fi
-


miabbott:
@darkmuggle where do you think would the best place for a test would be?  maybe [RHCOS exttest](https://github.com/coreos/coreos-assembler/blob/master/mantle/kola/README-kola-ext.md)?

miabbott:
Change looks sane to me.
/approve

templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
```diff
@@ -15,7 +15,8 @@ contents:
         exit 0
     fi
 
-    if [[ ! "$(< /proc/sys/kernel/hostname)" =~ (localhost|localhost.localdomain) ]]; then
+    kn="$(< /proc/sys/kernel/hostname)"
+    if [[ ! "${kn}" =~ (localhost|localhost.localdomain) ]] && [ "${kn}" -le 63 ]; then
```
yuqi-zhang, line 19: I think bash requires that you're comparing an int when doing the `-le` check, so you want `${#kn}` not the actual variable here I believe.

darkmuggle: Good catch....the `#` was intended.

yuqi-zhang (approved):
The code itself seems fine. Maybe for future debugging purposes it'd be nice to differentiate the truncate condition and log here. If you've tested this to work I can lgtm

cgwalters:
> hostnamectl previously refused to assign a long hostname. Since hostnamectl now allows for longer hostnames, 

Do you have a reference for this?  This is a systemd change?  Was it a change from RHEL 8.1 to 8.2?

yuqi-zhang:
`master` is currently 4.6 there is no need to cherry pick

ashcrow:
`Create IAM Members roles` is failing in the last two tests. Could be a flake. Let's give it one more run.

/retest

darkmuggle:
/retest

Still hitting the IAM limits -- I think that's a flake and the change passes when the flake isn't "active".

PR #2085, opened 2020-09-14
Bug 1873955: Add support for stopping and starting keepalived

cybertron:
There are circumstances where keepalived can cause issues with the
networking on a node, notably when bridging a physical interface.
After the address has been moved to the bridge, it is possible for
old routes to exist that cause problems talking to other nodes, which
breaks the apiserver and prevents us from updating the keepalived
config to reflect the networking change. This leaves us in a situation
where the code can't recover properly from the bad configuration.
In short, the apiserver is waiting for keepalived to update its
configuration, but keepalived needs the apiserver in order to do so.

This change addresses the problem by stopping keepalived if the
monitor fails to update the config more than 3 times in a row. That
will unconfigure any VIPs on the node, which should fix the error
described above. Once the bad routes related to the VIP(s) are gone,
the apiserver will recover and we'll be able to update the keepalived
config again. After that happens, keepalived is restarted.

This is one half of the fix. The other half will be in
baremetal-runtimecfg to call the control socket with stop and start
commands as appropriate.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/on-prem/keepalived.yaml
+++ b/manifests/on-prem/keepalived.yaml
@@ -42,42 +42,61 @@ spec:
         if pid=$(pgrep -o keepalived); then
             kill -s SIGHUP "$pid"
         else
-            /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+            start_keepalived
         fi
       }
+
       stop_keepalived()
       {
-        echo "Keepalived process stopped" >> /var/run/keepalived/stopped
         if pid=$(pgrep -o keepalived); then
-            kill -s TERM "$pid"
+          kill -s SIGTERM "$pid"
+          # The monitor runs every 10 seconds
+          sleep 9
+          if pid=$(pgrep -o keepalived); then
+            kill -s SIGKILL "$pid"
+          fi
+          touch /var/run/keepalived/stopped
         fi
+      }
+
+      start_keepalived()
+      {
+        if ! pgrep -o keepalived > /dev/null; then
+          /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+        fi
+        rm -f /var/run/keepalived/stopped
       }
 
       msg_handler()
       {
         while read -r line; do
+          # These get sent a lot, don't spam the logs with them
+          if [ "$line" = stop ]; then
+              stop_keepalived
+              continue
+          elif [ "$line" = start ]; then
+              start_keepalived
+              continue
+          fi
           echo "The client sent: $line" >&2
-          # currently only 'reload' and 'stop' msgs are supported
           if [ "$line" = reload ]; then
               reload_keepalived
-          elif  [ "$line" = stop ]; then
-              stop_keepalived
+          else
+              echo "Unrecognized command: $line" >&2
           fi
         done
       }
+
       set -ex
       declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
       export -f msg_handler
       export -f reload_keepalived
       export -f stop_keepalived
-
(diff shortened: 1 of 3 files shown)


manifests/baremetal/keepalived.yaml
```diff
@@ -49,26 +49,50 @@ spec:
...
         else
-            /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
+            start_keepalived
         fi
       }
+
+      stop_keepalived()
+      {
+        if pid=$(pgrep -o keepalived); then
+          kill "$pid"
```
celebdor, line 59: Two suggestions:

- Be specific about the signal we are sending
- Wait for the graceful termination.

cybertron: Do you mean add a kill -9 if it doesn't terminate gracefully (which seems like a good thing to do)? Otherwise there isn't much need to wait since we aren't going to do anything else after it exits.

celebdor: Yeah, I meant wait a reasonable time for SIGTERM to do its thing (probably we can use the same systemd uses) and then SIGKILL it.

cybertron: Yeah, that's possible. I'll make the change.

cybertron: Okay, this should be done. I set the timeout to 9 seconds since the monitor runs every 10 and I doubt we want to have multiple stop commands going at once.

yuqi-zhang:
/retest

Do we expect metal tests to pass on this one?

bcrochet (commented):
/lgtm
/retest

manifests/baremetal/keepalived.yaml
```diff
@@ -49,26 +49,57 @@ spec:
...
         done
       }
+
       set -ex
       declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
       export -f msg_handler
       export -f reload_keepalived
+      export -f stop_keepalived
+      export -f start_keepalived
       if [ -s "/etc/keepalived/keepalived.conf" ]; then
```
yboaron, line 99: Do we want to cover also the following (corner) case ? : 
1. Monitor container send STOP request 
   1.1 Keepalived container stop Keepalived process
2. After sometime Kubelet restart Kaapalived container for some reason

yboaron:
/hold 

[This BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1873955) was already resolved by the VIP mask change,  do we still want to push the workers/masters Keepalived start/stop mechanism for 4.6 ?

templates/common/baremetal/files/baremetal-keepalived.yaml
```diff
@@ -74,17 +74,44 @@ contents:
...
+                start_keepalived
             fi
           }
 
+          stop_keepalived()
+          {
+            if pid=$(pgrep -o keepalived); then
+              kill -s SIGTERM "$pid"
+              # The monitor runs every 10 seconds
+              sleep 9
```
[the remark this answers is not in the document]
cybertron: It's to give the process time to exit before we kill -9 it below.

manifests/baremetal/keepalived.yaml
cybertron: I'm inclined to say no. If the whole keepalived container gets restarted, I think we should let it attempt to run normally. If the monitor continues to fail it will stop it again anyway.

cybertron:
Sorry, somehow I missed Yossi's comments on this before. I've rebased it and tested it locally and it still seems to be working fine.

kikisdeliveryservice:
Underlying BZ was closed by https://github.com/openshift/baremetal-runtimecfg/pull/100/

Closing this PR, reopen if necessary

PR #2087, opened 2020-09-15
templates: Disable SSH keys lookup from authorized_keys.d on FCOS

LorbusChris:
**- What I did**
On FCOS, this sshd config dropin ensures that only SSH keys from the
`/home/core/.ssh/authorized_keys` file are picked up, and not ones
present in the `/home/core/.ssh/authorized_keys.d/` directory,
which might be written by Ignition and/or Afterburn.

On RHCOS this is a no-op, as it already looks up SSH keys from the
authorized_keys file only.

**- How to verify it**
CI e2e testing

templates: Disable SSH keys lookup from authorized_keys.d on FCOS

/hold
for testing

### DIFF
--- a/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
+++ b/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
@@ -1,0 +1,6 @@
+mode: 0644
+path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
+contents:
+  inline: |
+    # disable key lookup from ~/.ssh/authorized_keys.d/ on FCOS
+    AuthorizedKeysCommand none


ashcrow (commented):
The code makes sense :+1: . My only question is should this be done whenever `update` is called or should it be done once (bootstrap/firstboot)? I don't think it matters right now from the functionality perspective but still worth thinking about.

bgilbert (commented):
> My only question is should this be done whenever update is called or should it be done once (bootstrap/firstboot)? I don't think it matters right now from the functionality perspective but still worth thinking about.

On some platforms, Afterburn may run on subsequent boots and recreate its fragment file.

---

Note that this PR makes a tradeoff: to avoid having to programmatically restart sshd, it disables `authorized_keys.d` at provisioning time.  As a result, after an FCOS node first boots but before the MCO first syncs SSH keys, it will be impossible to SSH into the machine.

If we're comfortable with that tradeoff, there are a couple consequences:
- Since we're disabling `authorized_keys.d` from a template, we don't actually need to remove the fragment files at all, except perhaps as documentation that they're not being used.  On FCOS they'll never be read because of the `sshd_config.d` fragment.  On RHCOS, we can simply not re-enable reading of `.ssh/authorized_keys.d/ignition` in `sshd_config`.
- Furthermore, we could consider having Ignition on RHCOS continue to write the `authorized_keys` file directly.  That would enable SSHing into an RHCOS node before the MCO SSH key sync runs, at the cost of divergent behavior between the OSes.

ashcrow:
> On some platforms, Afterburn may run on subsequent boots and recreate its fragment file.

Ah, great point :+1:

LorbusChris:
@bgilbert updated, PTAL.

templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
```diff
@@ -0,0 +1,6 @@
+mode: 0644
+path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
+contents:
+  inline: |
+    # disable key lookup from ~/.ssh/authorized_keys.d/
```
bgilbert, line 5: ...on FCOS.

bgilbert (commented):
@LorbusChris Any thoughts on https://github.com/openshift/machine-config-operator/pull/2087#pullrequestreview-489234956?  I see this version is still deleting the fragment files.

LorbusChris:
@bgilbert if we're ok with Ignition continuing to write directly to `/home/core/.ssh/authorized_keys` on RHCOS we could indeed leave the files in the `.d` dir. I'll update.

bgilbert (approved):
LGTM, provided we're okay with the tradeoffs in https://github.com/openshift/machine-config-operator/pull/2087#pullrequestreview-489234956.

cgwalters:
Random aside, a lot of our thinking now is "rhcos == rhel8" but that *will* change - it's likely rhcos9 would potentially inherit any default changes we make to OpenSSH for example, so down the line ideally we'd do more "feature detection" rather than hardcode expectations that "rhcos" means a whole set of things.

ashcrow (approved):
LGTM too ... cc @yuqi-zhang @runcom

bgilbert:
> Random aside, a lot of our thinking now is "rhcos == rhel8" but that will change - it's likely rhcos9 would potentially inherit any default changes we make to OpenSSH for example

In this case there shouldn't be an issue.  If future RHEL versions start reading `sshd_config.d`, this config fragment will be a no-op without the fragment it's overriding, and the latter is added by the `ssh-key-dir` package in `fedora-coreos-base.yaml`.

Issue #944, opened 2019-07-10
Allow to set --node-ip in kubelet.service


FlorinPeter:
**Description**

We are facing currently some issues on AWS when adding additional network interfaces to a master or worker node. Some time it happens that the first InternalIP is not the IP from the first network interface.
To avoid this issue we tried to add the --node-ip parameter by overwriting the kubelet.service as described here https://github.com/openshift/installer/blob/master/docs/user/customization.md but it it turns out that this customization is not longer possible.

**Steps to reproduce the issue:**
1. Install cluster on AWS
2. add additional network interfaces

**Describe the results you received:**
Node ip-192-168-43-78.eu-central-1.compute.internal is reporting: "during
      bootstrap: unexpected on-disk state validating against rendered-worker-9480d1bae1f38b6a6efcbd8152249e7e"

**Describe the results you expected:**
A possibility to overwrite the kubelet.service or a possibility set the --node-ip parameter via environments variables 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `oc adm release info --commits | grep machine-config-operator`:**

```
 machine-config-controller                     https://github.com/openshift/machine-config-operator                       02c07496ba0417b3e12b78fb32baf6293d314f79
  machine-config-daemon                         https://github.com/openshift/machine-config-operator                       02c07496ba0417b3e12b78fb32baf6293d314f79
  machine-config-operator                       https://github.com/openshift/machine-config-operator                       02c07496ba0417b3e12b78fb32baf6293d314f79
  machine-config-server                         https://github.com/openshift/machine-config-operator                       02c07496ba0417b3e12b78fb32baf6293d314f79
  setup-etcd-environment                        https://github.com/openshift/machine-config-operator                       02c07496ba0417b3e12b78fb32baf6293d314f79
```

**Additional environment details (platform, options, etc.):**

runcom:
This probably needs to live in the kubelet Config? @rphillips

FlorinPeter:
The current workaround we found is to overwrite the https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-kubelet/_base/units/kubelet.yaml but this only for testing purpose. 
```yaml
name: "kubelet.service"
enabled: true
contents: |
  [Unit]
  Description=Kubernetes Kubelet
  Wants=rpc-statd.service crio.service
  After=crio.service
  [Service]
  Type=notify
  ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
[...]

PR #2088, opened 2020-09-15
Bug 1872632: allow overriding kubelet --node-ip in the _base templates

danwinship:
This is not a complete fix for bug 1872632. However, regardless of the approach taken there, it seems like there needs to be some way for platform-none UPI users to override an incorrect guess by kubelet regarding the node IP. Currently we allowing overriding it (based on `nodeip-configuration.service`) in bare metal IPI, openstack, and vsphere, but not anywhere else.

This just patches the _base kubelet service files so that if `KUBELET_NODE_IP` has been set, somewhere, we pass it to kubelet. It does not actually provide any simple way to set it yet. (`nodeip-configuration.service` won't work on UPI because it requires there to be an apiserver VIP.)

The idea of making `--node-ip` configurable was somewhat rejected in #944 although in that case it was clear that the user only needed it to work around a bug somewhere else (which may or may not also be true for bz 1872632, but it's certainly the case that in _some_ platform-none UPI installs, kubelet will not be able to correctly guess the node IP). But we could add `{{ if eq .Infra.Status.PlatformStatus.Type "None" }}` around the new code if we wanted to not allow overriding `--node-ip` on cloud platforms...

(I believe I'm correctly differentiating "bare metal platform" / "no platform" / IPI / UPI here but I may have gotten some of the details wrong)

cc @kikisdeliveryservice

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -25,7 +25,9 @@ contents: |
         --runtime-cgroups=/system.slice/crio.service \
         --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
 {{- if .KubeletIPv6}}
-        --node-ip :: \
+        --node-ip=${KUBELET_NODE_IP:-::} \
+{{- else}}
+        --node-ip=${KUBELET_NODE_IP:-} \
 {{- end}}
         --minimum-container-ttl-duration=6m0s \
         --cloud-provider={{cloudProvider .}} \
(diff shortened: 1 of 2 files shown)


dcbw:
ovn-step-registry seems to be the CI test pod not correctly starting, not even getting to the point of AWS setup
/retest

cgwalters:
> This just patches the _base kubelet service files so that if KUBELET_NODE_IP has been set, somewhere, we pass it to kubelet. It does not actually provide any simple way to set it yet. 

Presumably it'd work to write a systemd-drop in like `/etc/systemd/system/kubelet.service.d/80-nodeip.conf`
```
[Service]
Environment=KUBELET_NODE_IP=10.x.x.x
```
?

cgwalters (approved):
OK

danwinship:
> Presumably it'd work to write a systemd-drop in like `/etc/systemd/system/kubelet.service.d/80-nodeip.conf`

I was thinking "edit `/etc/kubernetes/kubelet-env`" since the unit file sources that. But anyway, I just meant that this doesn't provide any way to _automatically_ set the node IP. The user would still have to fiddle with machine configs or whatever to do that, at this point

PR #2089, opened 2020-09-15
Bug 1878794: handle azure vips for pods

aojea:
Azure hosts cannot hairpin back to themselves over a load
balancer. Thus, we need to redirect traffic to the apiserver vip to
ourselves via iptables. However, we should only do this when our local
apiserver is running.

Previously it was added support for host network only, adding the
same iptables rules in PREROUTING we apply the same rules for pods.

### DIFF
--- a/templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
+++ b/templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
@@ -80,24 +80,29 @@ contents:
 
     # set the chain, ensure entry rules, ensure ESTABLISHED rule
     initialize() {
-        ensure_chain4 nat "${CHAIN_NAME}-local"
-        ensure_chain6 nat "${CHAIN_NAME}-local"
+        ensure_chain4 nat "${CHAIN_NAME}"
+        ensure_chain6 nat "${CHAIN_NAME}"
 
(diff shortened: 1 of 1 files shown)


templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
```diff
@@ -80,35 +80,40 @@ contents:
 
     # set the chain, ensure entry rules, ensure ESTABLISHED rule
     initialize() {
-        ensure_chain4 nat "${CHAIN_NAME}-local"
```
abhat, line 83: why are we removing the `-local` suffix?

aojea: because the previous chain was referring only to local clients, that is no longer the case

sinnykumari:
Overall PR looks good and azure ci test is passing as well.
/approve
/bugzilla refresh

aojea:
/retest
the jobs seems unrelated to the PR, none of them are using azure

aojea:
/retest

e2e-gcp-o failed with 

level=fatal msg="failed to fetch Master Machines: failed to load asset \"Install Config\": platform.gcp.project: Invalid value: \"openshift-gce-devel-ci\": invalid project ID"

PR #2092, opened 2020-09-16
Bug 1884165: firstboot.service: disable existing repos before pivot

vrutkovs:
On FCOS boot OS extensions are being installed, so in order to use
only shipped RPMs and avoid pulling in updates from public repos existing
repos need to be removed.

TODO:
* [x] Disable repos instead of removing

Keeping this on hold until updated OKD machine-os-content is promoted

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
+++ b/templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
@@ -14,6 +14,8 @@ contents: |
   [Service]
   Type=oneshot
   RemainAfterExit=yes
+  # Disable existing repos (if any) so that OS extensions would use embedded RPMs only
+  ExecStartPre=-/usr/bin/sh -c "sed -i 's/enabled=1/enabled=0/' /etc/yum.repos.d/*.repo"
   ExecStart=/run/bin/machine-config-daemon firstboot-complete-machineconfig
 
   {{if .Proxy -}}


jlebon:
Looks sane to me, though agreed disabling would be better. Something like

```
ExecStart=/usr/bin/sh -c "sed -i 's/enabled=0/enabled=1/' /etc/yum.repos.d/*.repo"
```

?

templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
```diff
@@ -14,6 +14,7 @@ contents: |
   [Service]
   Type=oneshot
   RemainAfterExit=yes
+  ExecStart=/usr/bin/sh -c "sed -i 's/enabled=1/enabled=0/' /etc/yum.repos.d/*.repo"
```
sinnykumari, line 17: comment will be nice for future reference

vrutkovs: Good idea, fixed

vrutkovs:
/hold

Keeping this on hold until updated OKD machine-os-content is promoted

vrutkovs:
/retest

OKD tests pass

vrutkovs:
OKD tests are flaking (I'll have another look at that later).

So WDYT, does this look good enough to make it in 4.6?

cgwalters:
Wait but is something going to enable them then after this?

It feels cleaner to me to make this change as part of the OKD build process - that should be equally easy right?

cgwalters (changes requested):
See above

vrutkovs:
>Wait but is something going to enable them then after this?

Why would they need to be enabled? `-firstboot` would reboot the node so we can't run it in `ExecPost`.

Re-enabling the repos is trivial and I don't think we should be ensuring these are enabled post-firstboot

cgwalters:
> We're installing extensions when we're booted in plain FCOS, so we're using repos from initial deployment, not staged.

But isn't the FCOS build process generating a custom `machine-os-content` today?  

Wait I'm confused how this intersects with 
https://github.com/vrutkovs/machine-os-content-builder/blob/97bc83264e56393a3e45ca634c75a02fce35b9f6/entrypoint.sh#L107
Doesn't that already do what you're doing here?

cgwalters:
Oh wait sorry I do understand now, this is before the pivot.  Sorry about misunderstanding for so long.

Hmm...well
/approve
But...it feels like we could do this in Go code in the MCD rather than shell.  (More verbose but ultimately more reliable)

sinnykumari:
@vrutkovs Feel free to cancel hold
/lgtm

PR #2094, opened 2020-09-16
Bug 1879156: Check that resolv.conf actually has nameservers

bcrochet:
If /etc/resolv.conf existed but didn't have nameservers, the prepender
wouldn't copy the resolv.conf.

**- What I did**
Check that either resolv.conf does not exist OR does not contain nameservers
**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
@@ -25,7 +25,6 @@ contents:
         up|dhcp4-change|dhcp6-change)
         >&2 echo "NM resolv-prepender triggered by ${1} ${2}."
 
-        # Ensure resolv.conf exists before we try to run podman
         # In DHCP connections, the resolv.conf content may be late, thus we wait for nameservers
         timeout 45s /bin/bash <<EOF
             if [[ "$STATUS" == dhcp* ]]; then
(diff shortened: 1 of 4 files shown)


templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -36,8 +35,8 @@ contents:
                 done
             fi
     EOF
-        # Ensure resolv.conf exists before we try to run podman
-        if [[ ! -e /etc/resolv.conf ]]; then
+        # Ensure resolv.conf exists and contains nameservers before we try to run podman
+        if [[ ! -e /etc/resolv.conf || -z $(grep -q nameserver /etc/resolv.conf) ]]; then
```
cybertron, line 39: Could check the return code from grep directly by doing this:

if [[ ! -e /etc/resolv.conf ]] || ! grep -q nameserver /etc/resolv.conf; then

cybertron (approved):
/lgtm

We may want this for the other platforms too. At the very least it shouldn't hurt anything.

cybertron:
I think the title still isn't right. The : goes after the bug number.

bcrochet:
> I think the title still isn't right. The : goes after the bug number.

Apparently, I can't read.

mandre:
> /lgtm
> /cc @mandre @rgolangh @patrickdillon
> 
> We may want this for the other platforms too. At the very least it shouldn't hurt anything.

Agreed. @bcrochet, would you mind updating the patch to fix for all platforms?

cybertron:
/test e2e-vsphere

Note that ovirt and vsphere have been pretty unstable lately, so I don't know if we'll get ci passes from them.

kikisdeliveryservice (approved):

PR #2095, opened 2020-09-16
controller: Reflect pool paused status in Updating

cgwalters:
If a pool is paused, we can't be `Updating=True`.  Fixes a UX
issue as administrators want to use pool pausing to more
finely control updates.

### DIFF
--- a/pkg/controller/node/status.go
+++ b/pkg/controller/node/status.go
@@ -93,8 +93,13 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	} else {
 		supdated := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdated, corev1.ConditionFalse, "", "")
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdated)
-		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, "", fmt.Sprintf("All nodes are updating to %s", pool.Spec.Configuration.Name))
-		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+		if pool.Spec.Paused {
+			supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionFalse, "", fmt.Sprintf("Pool is paused; will not update to %s", pool.Spec.Configuration.Name))
+			mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+		} else {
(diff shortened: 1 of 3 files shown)


cgwalters:
Looking at my 4.7 pending work, this one should be an easy review I think, any takers?

pkg/controller/node/status.go
```diff
@@ -93,8 +93,13 @@ func calculateStatus(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) mcfgv
 	} else {
 		supdated := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdated, corev1.ConditionFalse, "", "")
 		mcfgv1.SetMachineConfigPoolCondition(&status, *supdated)
-		supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionTrue, "", fmt.Sprintf("All nodes are updating to %s", pool.Spec.Configuration.Name))
-		mcfgv1.SetMachineConfigPoolCondition(&status, *supdating)
+		if pool.Spec.Paused {
+			supdating := mcfgv1.NewMachineConfigPoolCondition(mcfgv1.MachineConfigPoolUpdating, corev1.ConditionFalse, "", fmt.Sprintf("Pool is paused; would update to %s", pool.Spec.Configuration.Name))
```
kikisdeliveryservice, line 97: Feel like message would be clearer as:

"Pool paused, will not update to xxxxx"

kikisdeliveryservice (commented):
I'm wondering how this affects the CVO status though... if master pool is paused for some reason?

cgwalters:
> I'm wondering how this affects the CVO status though... if master pool is paused for some reason?

The clusteroperator status?  Hmm...good question.  As best I can tell the operator doesn't look at the pool's `Upgrading` status.   The core logic currently is
```
			degraded := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolDegraded)
			if pool.Generation <= pool.Status.ObservedGeneration &&
				isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated) &&
				!degraded {
				continue
			}
			lastErr = fmt.Errorf("error pool %s is not ready, retrying. Status: (pool degraded: %v total: %d, ready %d, updated: %d, unavailable: %d)", pool.Name, degraded, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount)
			return false, nil
```

So the operator will go degraded if the master pool is paused while it's not updated.

That...seems OK to me?

pkg/controller/node/status.go
cgwalters: Sure, will ~not~ update to that wording instead.

kikisdeliveryservice:
Just dealt with this the other day and it's pretty confusing to have:
Paused:True, Updating:True, Updated:False

This PR makes more intuitive sense to me, bc it's not updating...it's paused:
Paused: True, Updating:False, Updated:False 

Then once you unpause you will have:
Paused: False, Updating:True, Updated:False

which is correct and clear.

/lgtm

PR #2097, opened 2020-09-17
Bug 1865839: daemon: better error reporting for rpm-ostree operations

sinnykumari:
When an error occurs during rpm-ostree operation, we get
returned error as exit status 1 which is not much useful.
Actual failure gets printed on the stdout which we capture
in runGetOut() function.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -260,8 +260,7 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 func runGetOut(command string, args ...string) ([]byte, error) {
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	cmd.Stderr = os.Stderr
-	rawOut, err := cmd.Output()
+	rawOut, err := cmd.CombinedOutput()
 	if err != nil {
 		return nil, errors.Wrapf(err, "error running %s %s: %s", command, strings.Join(args, " "), string(rawOut))
 	}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -837,7 +837,8 @@ func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConf
 
 	args := append([]string{"kargs"}, diff...)
 	dn.logSystem("Running rpm-ostree %v", args)
-	return exec.Command("rpm-ostree", args...).Run()
+	_, err := runGetOut("rpm-ostree", args...)
+	return err
 }
 
 func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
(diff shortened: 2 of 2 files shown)


pkg/daemon/rpm-ostree.go
```diff
@@ -260,10 +261,11 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 func runGetOut(command string, args ...string) ([]byte, error) {
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	cmd.Stderr = os.Stderr
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
 	rawOut, err := cmd.Output()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error running %s %s: %s", command, strings.Join(args, " "), string(rawOut))
+		return nil, errors.Wrapf(err, "error running %s %s: %s%s", command, strings.Join(args, " "), string(rawOut), stderr.string())
```
ashcrow, line 268:
```suggestion
		return nil, errors.Wrapf(err, "error running %s %s: %s%s", command, strings.Join(args, " "), string(rawOut), stderr.String())
```

CI errors are from a missing capital letter :slightly_smiling_face:. The `String` function on `bytes.Buffer` should [start with a capital letter](https://golang.org/pkg/bytes/#Buffer.String).

ashcrow (changes requested):
One change requested (see comment).

pkg/daemon/rpm-ostree.go
sinnykumari: sigh, i fixed it locally and didn't do git add before ammend

ashcrow (approved):

pkg/daemon/rpm-ostree.go
```diff
@@ -260,10 +261,11 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 func runGetOut(command string, args ...string) ([]byte, error) {
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	cmd.Stderr = os.Stderr
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
 	rawOut, err := cmd.Output()
```
cgwalters, line 266: I think it'd be both simpler and better to use https://golang.org/pkg/os/exec/#Cmd.CombinedOutput in this case - this way stdout/stderr will be ordered correctly, and we only have one buffer to look at.

cgwalters (changes requested):

pkg/daemon/rpm-ostree.go
sinnykumari: agree. Fixed it in latest push.

cgwalters (approved):
/lgtm
as is
but a followup: I think it'd be even simpler if we introduced a basic `run()` that just didn't return the output, since almost all of the callers you change here are ignoring it.

sinnykumari:
> but a followup: I think it'd be even simpler if we introduced a basic `run()` that just didn't return the output, since almost all of the callers you change here are ignoring it.

we can. I thought it is better approach(in terms of maintenance as well) to use existing function as we are just ignoring some of return type.

PR #2100, opened 2020-09-18
Bug 1872632: templates: fix --node-ip override

danwinship:
#2088 was broken because I tried to use shell syntax in a systemd unit file where shell syntax isn't parsed _and_ I forgot that kubelet will just treat a bogus `--node-ip` value (eg, passing literally `--node-ip='${KUBELET_NODE_IP:-}'`) the same as not passing any value so my assumption that it would break CI if it was wrong failed.

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -11,6 +11,9 @@ contents: |
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
   Environment="KUBELET_LOG_LEVEL=4"
+{{- if .KubeletIPv6}}
+  Environment="KUBELET_NODE_IP=::"
+{{- end}}
   EnvironmentFile=/etc/os-release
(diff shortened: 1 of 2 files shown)


danwinship:
OK, so this now does the right thing by default, and it does the right thing if you create an override file as Colin suggested in https://github.com/openshift/machine-config-operator/pull/2088#issuecomment-692829731, but it _doesn't_ work if you just add the variable to `/etc/kubernetes/kubelet-env`. Is there some weird "running systemd unit files out of a chroot" thing going on or something? Not only does it not see things when I add them to `/etc/kubernetes/kubelet-env`, but if I remove the `-` from `EnvironmentFile=-/etc/kubernetes/kubelet-env` after creating that file, then kubelet fails to start with `systemd[1]: kubelet.service: Failed to load environment files: No such file or directory`...

But anyway, that's either a pre-existing problem or me doing something stupid; it doesn't work to override `KUBELET_LOG_LEVEL` either (and you _can_ override `KUBELET_NODE_IP` by setting it in `/etc/os-release`... seriously WTF is going on here? I'm not typo'ing the filename, I actually copied and pasted from the unit file...)

/assign @cgwalters

cgwalters:
Hopefully once we branch we can refactor this code to generate the kubelet config specifically rather than hacking it with these templates...

/approve

cgwalters (approved):

PR #2101, opened 2020-09-18
Bug 1877984: Convert networkType from status to lowercase

rcarrillocruz:
**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/operator/render.go
+++ b/pkg/operator/render.go
@@ -109,7 +109,7 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 	}
 	if network.Status.NetworkType == "" {
 		// At install time, when CNO has not started, status is unset, use the value in spec.
-		ccSpec.NetworkType = network.Spec.NetworkType
+		ccSpec.NetworkType = strings.ToLower(network.Spec.NetworkType)
 	} else {
 		// After installation, the MCO should not assume the network is changing just because the spec changed, it needs to wait until CNO updates the status.
 		ccSpec.NetworkType = network.Status.NetworkType


pkg/operator/render.go
```diff
@@ -109,7 +109,7 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 	}
 	if network.Status.NetworkType == "" {
 		// At install time, when CNO has not started, status is unset, use the value in spec.
-		ccSpec.NetworkType = network.Spec.NetworkType
+		ccSpec.NetworkType = strings.ToLower(network.Spec.NetworkType)
```
danwinship, line 112: The template files currently check for it in the "canonical" case; so you either need to case-insensitively check for `OpenShiftSDN` and `OVNKubernetes` and then set `ccSpec.NetworkType` to the canonically-cased value. Or else set `ccSpec.NetworkType` to the lowercase value both here and in the case below, and update the templates to use the lowercase value.

I feel like forcing templates to use a nonstandard "spelling" would be weird, so probably the former is preferable...

danwinship: (we could also potentially "fix" this the same way in the installer...)

abhat: +1 to @danwinship 's comment. I think fixing this in the installer makes more sense. If the install-config takes `openshiftsdn`, `OpenShiftSDN`, or `OpenshiftSDN`, it should all get parsed in as canonical case `OpenShiftSDN` and passed to the templates.

abhat: Likewise for `ovnkubernetes`

wking: Downcasing outside of the network operator is easy.  Canonical casing outside of the network operator would be more elegant from a user perspective, but requires teaching the canonicalizing repo about the canonical values.  Are there more than [these][1]?

[1]: https://github.com/openshift/installer/blob/79510119c59212ed1021a4bf96233df646667b2e/pkg/types/validation/installconfig.go#L42

kikisdeliveryservice: I think it makes more sense to fix this where it's originally coming from rather than changing it in the controller config ie fix in installer.

kikisdeliveryservice:
I don't think CC is the place to do this and there seems to be agreement above so..

/hold

pkg/operator/render.go
danwinship: `OpenShiftSDN` and `OVNKubernetes` are the only two network types that CNO canonicalizes.

(IIRC the reason we canonicalize `OpenShiftSDN` is because it was originally defined incorrectly in openshift/api, and even though we never shipped a release where `OpenshiftSDN` was considered correct, it was simpler to accept it to preserve compatibility with old tools that were generating install-configs with the incorrect version. Presumably the reason we canonicalize `OVNKubernetes` is because someone saw that we canonicalized `OpenShiftSDN` and copied it.)

rcarrillocruz: the deeper the better, so +1 from me to fix this in installer.
Discussing tihs internally, if agree we may move the BZ to installer team

rcarrillocruz: > Downcasing outside of the network operator is easy. Canonical casing outside of the network operator would be more elegant from a user perspective, but requires teaching the canonicalizing repo about the canonical values. Are there more than [these](https://github.com/openshift/installer/blob/79510119c59212ed1021a4bf96233df646667b2e/pkg/types/validation/installconfig.go#L42)?

I think "Kuryr' should also be taken into account

rcarrillocruz: We agreed on team meeting to move this to installer, thus closing PR

PR #2102, opened 2020-09-18
Bug 1874696: Add ovs-notification.service

juanluisvaladas:
In ovnkubernetes and openshift-sdn we're unable to reliably identify
whether the openvswitch.service is enabled or not though systemctl or
the systemd's DBus API.

To avoid issues we have decided to create a unit that launches before
the SDN and creates an empty file that we can check easily and reliably.
This is part of a fix for BZ#1874696

### DIFF
--- a/templates/common/_base/units/ovs-notification.yaml
+++ b/templates/common/_base/units/ovs-notification.yaml
@@ -1,0 +1,19 @@
+name: ovs-notification.service
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
+contents: |
+  [Unit]
+  Description=Creates a file to let the OVS pod know OVS is running on systemd.
+  Wants=NetworkManager-wait-online.service
+  After=NetworkManager-wait-online.service openvswitch.service
+  Before=network-online.target kubelet.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  # Create the file in a tmpfs so that the file gets deleted on restart
+  ExecStart=/usr/bin/touch /var/run/openvswitch/ovs-notification
+  StandardOutput=journal+console
+  StandardError=journal+console
+
+  [Install]
+  WantedBy=network-online.target


templates/common/_base/units/ovs-notification.yaml
```diff
@@ -0,0 +1,20 @@
...
+  # Don't write the notification before OVS starts
+  Requires=openvswitch.service
+  Wants=NetworkManager-wait-online.service
+  After=NetworkManager-wait-online.service openvswitch.service
+  Before=network-online.target kubelet.service crio.service node-valid-hostname.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  ExecStart=/usr/bin/touch /var/run/openvswitch/ovs-notification
```
dcbw, line 15: Might be worth a comment that we put this in /var/run/ (which we assume to be a tmpfs) so that it gets deleted automatically on restart.

dcbw:
lgtm but we should get some more review @abhat @mccv1r0 @danwinship @trozet

templates/common/_base/units/ovs-notification.yaml
juanluisvaladas: done

juanluisvaladas:
Tested it manually everything looks fine in my manual testing.

kikisdeliveryservice:
> lgtm but we should get some more review @abhat @mccv1r0 @danwinship @trozet

Was going to say the same thing as they have the expertise to review this PR as they've been working on this :+1:

dcbw:
@juanluisvaladas when you confirm that a cluster-bot multi-PR run does the right thing, please note that in this PR. Thanks!

templates/common/_base/units/ovs-notification.yaml
```diff
@@ -0,0 +1,21 @@
+name: ovs-notification.service
+enabled: {{if eq .NetworkType "OVNKubernetes"}}true{{else if eq .NetworkType "OpenShiftSDN"}}true{{else}}false{{end}}
+contents: |
+  [Unit]
+  Description=Creates a file to let the OVS pod know OVS is running on systemd.
+  # Don't write the notification before OVS starts
```
danwinship, line 6: Why not? The goal of the service is to tell the sdn-ovs pod that system OVS is enabled. If this service is running at all, then system OVS is enabled, so it doesn't matter if this runs before or after OVS starts

templates/common/_base/units/ovs-notification.yaml
```diff
@@ -0,0 +1,21 @@
...
+  Requires=openvswitch.service
+  Wants=NetworkManager-wait-online.service
+  After=NetworkManager-wait-online.service openvswitch.service
+  Before=network-online.target kubelet.service crio.service node-valid-hostname.service
```
danwinship, line 10: Why should we care about most of those services? If we run before kubelet then we run before the sdn-ovs pod and we're good, right?

templates/common/_base/units/ovs-notification.yaml
```diff
@@ -0,0 +1,21 @@
...
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  # Create the file in a tmpfs so that the file gets deleted on restart
+  ExecStart=/usr/bin/touch /var/run/openvswitch/ovs-notification
+  StandardOutput=journal+console
+  StandardError=journal+console
+
+  [Install]
+  WantedBy=network-online.target
```
danwinship, line 21: why?

templates/common/_base/units/ovs-notification.yaml
mccv1r0: Is the goal to have a `hostPath` `type File` constraint in [ovn-kubernetes/006-ovs-node.yaml ](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/006-ovs-node.yaml) to keep ovs-node from starting until it sees this file?

juanluisvaladas:
> @juanluisvaladas when you confirm that a cluster-bot multi-PR run does the right thing, please note that in this PR. Thanks!
Yes, I confirm that

templates/common/_base/units/ovs-notification.yaml
juanluisvaladas: @danwinship yes I guess that's reasonable.
@mccv1r0 yes

templates/common/_base/units/ovs-notification.yaml
juanluisvaladas: Did that to be consistent with the ovs-configuration.service.yaml, but you're right.

I'm pretty sure we don't need the node-valid-hostname.service and crio. Regarding the network-online.target, I think it's necessary so that the network-online.target isn't complete until this is finished.

templates/common/_base/units/ovs-notification.yaml
juanluisvaladas: This is part of the network so I'd say we can't consider the network completely setup until this unit is complete. I would also make sense in the network.target but I think network-online.target is the most logical place.

Anyway I don't feel strongly enough about this to block the PR while we discuss where it should go, so if you want it somewhere else I'll move it.

juanluisvaladas:
@danwinship resolved some of your concerns, but not entirely. Please let me know if you want it outside the network-online.target where should it go. network.target and I'll change that?

templates/common/_base/units/ovs-notification.yaml
trozet: This isn't right to say wanted by network online target, but then in runtime this is running before network online target

trozet (changes requested):
I dont really see why this PR is necessary if we already have ovs-confguration service. You can just  mount  that path and check that the file exists. As long as the service is part of a target, its fine to check the target file path. I think we already do this in ovn: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/ovnkube-node.yaml#L229

danwinship:
> I dont really see why this PR is necessary if we already have ovs-confguration service. You can just mount that path and check that the file exists.

Yeah... we don't need to answer the question "Is OVS _actually_ running as a system service on this node?", we just need to answer the question "Is this node running a version of OCP which would configure OVS to run as a system service if OpenShiftSDN was the configured network plugin?" which is a fact that is either true or false at install/upgrade time, not something that needs to be computed at any particular time during boot.

`ovs-configuration.service` isn't quite the right thing to check for though; it's a _slightly_ different thing. (At the moment we don't even install it on openshift-sdn nodes anyway, and it's possible that nmstate or something would result in that all being done differently in the future.)

We could check for one of the `.conf` files created as side effects of `templates/common/_base/units/ovs-vswitchd.service.yaml` or `ovsdb-server.service.yaml`, but those are only specified as relative paths and I'm not sure it's 100% guaranteed that they end up as specific absolute paths? (In the same way we don't want to try to figure out if OVS is enabled based on file paths.)

If we can't check for those, then we can add a new file to `templates/common/_base/files/` to just statically create some file, and then openshift-sdn can then check for that.

trozet:
@danwinship I think ovs-configuration exists even in openshift-sdn deployments; it just is not enabled as a service. Peng was changing this though in another PR to also enable it for openshift-sdn I think: #2066

danwinship:
Ah... ok. And actually "it's possible that nmstate or something would result in that all being done differently in the future" doesn't matter because this is only needed for the 4.6 cycle anyway right? (The 4.6 sdn-ovs pod needs to be able to DTRT when started on a 4.5 node, but the 4.7 sdn-ovs pod will only ever be started on 4.6 or 4.7 nodes and so can assume system OVS is always enabled.)

juanluisvaladas:
> Peng was changing this though in another PR to also enable it for openshift-sdn I think: #2066
@pliurh Can this be merged before code freeze? If it will be merged the best thing we can do is closing this PR and add a line with `touch /usr/bin/touch /var/run/openvswitch/ovs-notification` in configure-ovs.sh

pliurh:
> @pliurh Can this be merged before code freeze? If it will be merged the best thing we can do is closing this PR and add a line with `touch /usr/bin/touch /var/run/openvswitch/ovs-notification` in configure-ovs.sh

I think we have to get that merge before code freeze. The keepalived issue has been fixed. I'm testing #2066 with the latest build. Could you add a comment in #2066 where you want to add this line?

danwinship:
> If it will be merged the best thing we can do is closing this PR and add a line with `touch /usr/bin/touch /var/run/openvswitch/ovs-notification` in configure-ovs.sh

No, you don't need to test if the service has _run_, you only need to test if it's _there_. `test -f /usr/local/bin/configure-ovs.sh`. If it's there, then this is a 4.6 node and OVS is run on the system. If it's not there, then this is a 4.5 node and OVS should be run in the pod.

dcbw:
> No, you don't need to test if the service has _run_, you only need to test if it's _there_. `test -f /usr/local/bin/configure-ovs.sh`. If it's there, then this is a 4.6 node and OVS is run on the system. If it's not there, then this is a 4.5 node and OVS should be run in the pod.

hmm, good point; if we do this, we're not yet sure if OVS is actually *running*, so we'll need to remove all the log tailing stuff from the OVS container or maybe have it use `-F` to continously retry?

juanluisvaladas:
/close
Not necessary, we'll use https://github.com/openshift/machine-config-operator/pull/2066 instead.

PR #2105, opened 2020-09-21
[baremetal, openstack, ovirt, vsphere] Change keepalived VIP subnet mask to host mask

yboaron:
To work around [1] bug in network manager the subnet mask of Keepalived VIP address was set to the
value of control plane netmask.
So, after the network manager bug was resolved we can safely change VIP subnet mask to a host subnet mask.

Additionally, we noticed that sometimes after moving the control plane IP from the physical interface to
a bridge there were old routes still pointing to the physical interface, and that messed up the node's connectivity.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1700415

### DIFF
--- a/manifests/on-prem/keepalived.conf.tmpl
+++ b/manifests/on-prem/keepalived.conf.tmpl
@@ -24,6 +24,6 @@
         auth_pass {{.Cluster.Name}}_api_vip
     }
     virtual_ipaddress {
-        {{ .Cluster.APIVIP }}/{{ .Cluster.VIPNetmask }}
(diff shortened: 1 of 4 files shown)


yboaron:
/retitle bug 1873955: Change keepalived VIP subnet mask to host mask

yboaron:
/retitle Change keepalived VIP subnet mask to host mask

mandre:
Triggering the optional e2e jobs for other on-prem platforms:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

PR #2108, opened 2020-09-21
Bug 1881057: ignore the IPv6DualStack feature gate for the kubelet config

danwinship:
To install a dual-stack cluster you must set the IPv6DualStack feature gate at install time. MCO currently does not process feature gates provided as install-time manifests and so the bootstrap configs will not match the post-install configs and the operator will become Degraded because the masters are stuck unable to move to the expected config.

As it happens, kubelet does not use the IPv6DualStack feature gate unless you are using dockershim, so to work around this problem for now, just ignore that feature gate when generating the kubelet config, so that the "real" generated config will match the config expected by the bootstrap code.

(Not yet tested, but I think this will fix things in 4.6, and then it can be reverted and fixed correctly in 4.7, since kubelet will need the feature gate then.)

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_features.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features.go
@@ -204,7 +204,11 @@ func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[st
 		return &rv, fmt.Errorf("enabled FeatureSet %v does not have a corresponding config", features.Spec.FeatureSet)
 	}
 	for _, featEnabled := range set.Enabled {
-		rv[featEnabled] = true
+		if featEnabled == "IPv6DualStack" {
+			glog.Infof("Ignoring IPv6DualStack feature gate for kubelet config")
+		} else {
+			rv[featEnabled] = true
+		}
 	}
 	for _, featDisabled := range set.Disabled {
 		rv[featDisabled] = false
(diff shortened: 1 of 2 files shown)


cgwalters:
Hmm...so the feature gate isn't implemented in crio?  Or is it more we're "abusing" the kubelet config flag in other places in the platform - i.e. are there operators that are reading this flag in the platform right now?

cgwalters (approved):
Tentative approval

danwinship:
> Hmm...so the feature gate isn't implemented in crio? 

AFAIK crio doesn't know about feature gates, and doesn't treat "dual stack" as a configurable option; if the CNI plugin returns dual-stack IPs to crio then crio returns dual-stack IPs to kubelet. (And kubelet likewise will just set dual-stack IPs on the Pod if it gets dual-stack IPs from crio.)

> Or is it more we're "abusing" the kubelet config flag in other places in the platform - i.e. are there operators that are reading this flag in the platform right now?

No, nothing else is looking at kubelet's config. The issue is that OCP only lets you specify a single set of feature gates which then gets used by every component that cares about feature gates. So we can't enable `IPv6DualStack` in kube-apiserver without also causing it to be enabled in kubelet, which then triggers the MCO bug. So this just works around the problem by having kubelet's FeatureGate listener ignore it.

danwinship:
With this PR, machine-config becomes Available and not Degraded.

cgwalters:
https://github.com/openshift/machine-config-operator/pull/2108#issuecomment-696225841
is great info, makes sense.
/approve
/lgtm

PR #2112, opened 2020-09-21
Bug 1879099: Inject version into controllerconfig, refuse mismatches

cgwalters:

Per Jerry's comment in Bugzilla:

> 1. The new MCO rolls out. It first syncs the controllerconfig to the new one as part of the RenderConfig step above.
> 2. The running (old) MCC's template controller has an informer on the controllerconfig, which since the previous step updated the config, calls a syncControllerConfig. This re-generates the base master config (00-master MC)
> 3. The running (old) MCC now sees a MC change which calls another sync call to generate a new rendered config, and instructs the master pool to start an update
> 4. The new MCC rolls out, and does the whole rendering process, creating the actual updated MC for the master pool, which it waits until the previous one is finished, and starts the new (real) update.

After this patch, the old MCC will refuse to render from a controllerconfig generated by the new MCO.

### DIFF
--- a/manifests/controllerconfig.crd.yaml
+++ b/manifests/controllerconfig.crd.yaml
@@ -441,6 +441,9 @@ spec:
                 uid:
                   description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                   type: string
+            rendererVersion:
+              description: operator version
+              type: string
             rootCAData:
               description: rootCAData specifies the root CA data
               type: string
--- a/lib/resourcemerge/machineconfig.go
+++ b/lib/resourcemerge/machineconfig.go
@@ -66,6 +66,7 @@ func ensureMachineConfigSpec(modified *bool, existing *mcfgv1.MachineConfigSpec,
 }
 
 func ensureControllerConfigSpec(modified *bool, existing *mcfgv1.ControllerConfigSpec, required mcfgv1.ControllerConfigSpec) {
+	setStringIfSet(modified, &existing.RendererVersion, required.RendererVersion)
 	setStringIfSet(modified, &existing.ClusterDNSIP, required.ClusterDNSIP)
 	setStringIfSet(modified, &existing.CloudProviderConfig, required.CloudProviderConfig)
 	setStringIfSet(modified, &existing.Platform, required.Platform)
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -31,6 +31,11 @@ type ControllerConfig struct {
 
 // ControllerConfigSpec is the spec for ControllerConfig resource.
 type ControllerConfigSpec struct {
+	// RendererVersion is the operator version that generated this object;
+	// readers (e.g. machine-config-controller) whose version does not
+	// match should ignore this file.
+	RendererVersion string `json:"rendererVersion"`
+
 	// clusterDNSIP is the cluster DNS IP address
 	ClusterDNSIP string `json:"clusterDNSIP"`
 
--- a/pkg/controller/bootstrap/testdata/bootstrap/machineconfigcontroller-controllerconfig.yaml
+++ b/pkg/controller/bootstrap/testdata/bootstrap/machineconfigcontroller-controllerconfig.yaml
@@ -3,6 +3,7 @@ kind: ControllerConfig
 metadata:
   name: machine-config-controller
 spec:
+  rendererVersion: "v0.0.0-was-not-built-properly"
   additionalTrustBundle: null
   cloudProviderConfig: ""
   clusterDNSIP: 172.30.0.10
--- a/pkg/controller/render/render_controller.go
+++ b/pkg/controller/render/render_controller.go
@@ -529,6 +529,12 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 
 // generateRenderedMachineConfig takes all MCs for a given pool and returns a single rendered MC. For ex master-XXXX or worker-XXXX
 func generateRenderedMachineConfig(pool *mcfgv1.MachineConfigPool, configs []*mcfgv1.MachineConfig, cconfig *mcfgv1.ControllerConfig) (*mcfgv1.MachineConfig, error) {
+	// Suppress rendered config generation until a corresponding new controller can roll out too.
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1879099
+	if cconfig.Spec.RendererVersion != version.Raw {
+		return nil, fmt.Errorf("Ignoring controller config generated from %s (my version: %s)", cconfig.Spec.RendererVersion, version.Raw)
+	}
+
 	// Before merging all MCs for a specific pool, let's make sure MachineConfigs are valid
 	for _, config := range configs {
 		if err := ctrlcommon.ValidateMachineConfig(config.Spec); err != nil {
--- a/pkg/controller/render/render_controller_test.go
+++ b/pkg/controller/render/render_controller_test.go
@@ -28,6 +28,7 @@ import (
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
 	informers "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/machine-config-operator/pkg/version"
 	"github.com/openshift/machine-config-operator/test/helpers"
 )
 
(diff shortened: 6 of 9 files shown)


yuqi-zhang:
As we discussed the issue with this approach is that it must be ported back to 4.5, and then we'd have to gate upgrade paths. The 4.5 controller is the one generating the extra config and it would need to have this check to stop the generation

cgwalters:
> As we discussed the issue with this approach is that it must be ported back to 4.5, and then we'd have to gate upgrade paths. The 4.5 controller is the one generating the extra config and it would need to have this check to stop the generation

Yep, you're right.  When we were debating this I was saying we should do fix that handles 4.6, but then reading the code and previous changes more carefully I realized that we had a lot of similar version checking patterns in the code, and it makes sense to follow those.

I think my proposal is:

- Land this in master (debate 4.6 versus 4.7...I lean slightly to 4.6 but eh)
- Add a CI test that validates both number of reboots *and* number of rendered MCs for master in e2e-*-upgrade jobs
- Backport to 4.5
- If we get more reports of this, consider an additional fix that has the master/4.6 code explicitly upgrade the controller first or so

cgwalters:
```
E0921 21:42:42.546872       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 245 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x19f3140, 0x2abbff0)
	/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa6
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/github.com/openshift/machine-config-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x89
panic(0x19f3140, 0x2abbff0)
	/usr/lib/golang/src/runtime/panic.go:969 +0x175
github.com/openshift/machine-config-operator/pkg/operator.(*Operator).syncRenderConfig(0xc0002fae00, 0x0, 0xc020321a40, 0x29ced5a23f2)
```

Ah, I see `syncRenderConfig` takes a pointer we can't use yet initially because it's generating it.  Fixed.

cgwalters:
Test in https://github.com/openshift/origin/pull/25533

cgwalters:
cluster-bot: job [test upgrade 4.6 registry.svc.ci.openshift.org/ci-ln-nrl1hdk/release:latest succeeded](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1308400900636676096)
This validates that we can upgrade from the current 4.6 (without this PR, and hence with a `controllerconfig` object without a version) to this code.

cgwalters:
/bugzilla refresh
I lean towards landing this in 4.6 (and backporting to 4.5) - based on the above testing I have some confidence in it, but leaving up to MCO team.

yuqi-zhang:
After some more consideration I am still leaning towards pushing this out to 4.7, or at least not adding this to the initial 4.6 release:

1.The underlying upgrade issue will not be solved unless we backport to 4.5 and gate upgrades off that version
2. I would like to consider this design to make it better
3. This shouldn't block upgrades

I'm leaning towards fixing in z-streams in the future instead.

kikisdeliveryservice:
Given the above comments I'm inclined to wait until a z-fix. We can review with @runcom when he's back on Monday as well, for a final check. Does that work?

kikisdeliveryservice:
We can also create the backport to test the upgrade situation too since that would be required for the fix.

cgwalters:
> We can also create the backport to test the upgrade situation too since that would be required for the fix.

The inverse of https://github.com/openshift/machine-config-operator/pull/2112#issuecomment-696796781 - right.  I *think* everything would work if the upgrade source had this change even if the target didn't, but it'd probably be saner to test with both.

> Given the above comments I'm inclined to wait until a z-fix. We can review with @runcom when he's back on Monday as well, for a final check. Does that work?

Sure, it's up to you guys.  I'd say there's a low risk and medium benefit here, but low != zero for sure.

I don't have a strong sense of how often this occurs; it seems like it'd be a rare race but OTOH at scale, even rare races have a tendency to pop up a lot more than you'd think.

cgwalters:
4.5 backport in https://github.com/openshift/machine-config-operator/pull/2118 (some small trivial conflicts) - will test an upgrade from that to 4.6 w/cluster-bot soon.

ashcrow (approved):

runcom:
we really needed to merge all controllers :man_facepalming: instead we'll have this version check in all our crds :( 

guess /approve - let's see when this will go, I'll update this PR (thanks @cgwalters )

runcom:
One thing that this PR isn't really putting an accent on is the "extra reboot" that a master node does during a 4.5 to 4.6 upgrade with this bug. If it wasn't for that, I'm agreeing with the team that this should go to a z stream (or a better solution would be to have a better design for 4.7 that also fixes this).
However the extra reboot is a regression. We had bugs like this before sure, and we carried them for more than a release which makes them not regressions anymore but this is introducting a regression.
If it's just 1 master node rebooting (Jerry is verifying it) I think 4.6 can go w/o this and we can fix this immediately after in a z stream with this PR and a 4.5 backport (and again, maybe bake something better for future releases).
@crawford @cgwalters the extra reboot, assuming it's just on a master node should take 5-10m extra - would you deem that acceptable overall?

crawford:
Yes, the extra reboots are acceptable. This is purely a performance issue and not one of correctness. Is it truly a regression? It sounds like this race has always existed.

crawford:
As for the PR itself, this is going to need much more detail in the commit message. I'm not going to remember why we needed to make this change a few months from now.

runcom:
> Is it truly a regression? It sounds like this race has always existed.

the race yeah, as I said, the extra reboot wasn't there but seems like that's acceptable thanks Alex

runcom:
so, last word is this is going to _past_ 4.6 :)

yuqi-zhang:
So I have a bit of a bad news regarding this. The extra reboot does in fact happen for all nodes in the master pool. This might have something to do with MCC versions?

Normally if I apply a MC, and then while the nodes are updating, apply a second MC, the first rendered-config will be skipped for nodes that have not yet updated. i.e.
node1: 2 reboots
node2, 3: 1 reboot

but for the upgrade from 4.5->4.6 you actually get all the new MCDs applying the first MC and then the second MC. This should be controlled by the MCC but since the nodes reboot the "new" MCC gets booted around the master nodes (restarting many times as the MCD restarts the node, losing the logs while its doing so), so in my test I did not monitor the logs closely enough to see why it doesn't just instruct the nodes to immediately go to the newest rendered config.

Again, this doesn't fail CI and as far as the user is concerned, all they should see is that the 4.5->4.6 upgrade takes a bit longer than older versions.

crawford:
@yuqi-zhang thanks for digging into that. For the record, I still don't think this is a blocker for 4.6.

cgwalters:
> As for the PR itself, this is going to need much more detail in the commit message. I'm not going to remember why we needed to make this change a few months from now.

Added Jerry's comment from Bugzilla into the commit message.

cgwalters:
Note: when trying to test https://github.com/openshift/machine-config-operator/pull/2167 in a 4.5 :arrow_right: 4.6 upgrade I hit this issue.

cgwalters:
Rebased :surfer:

kikisdeliveryservice:
Actually I hit a wedged update once that (among other problems it had) that I think hit this as it was rendering mismatched CC with a deprecated /removed field (I think? It was a while ago), but the MCC also couldnt update due to the other cluster issues and it was stuck with the wrong config and a mess..

manifests/controllerconfig.crd.yaml
```diff
@@ -441,6 +441,9 @@ spec:
                 uid:
                   description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                   type: string
+            rendererVersion:
+              description: operator version
+              type: string
```
runcom, line 446: can we just use the metadata.Annotations field instead of adding a new field to the spec? I don't remember why we need this in the past but it does make sense to leverage the already existing field instead of adding a new one

runcom (changes requested):
Just a I think tiny request and ready to test and merge then

cgwalters:
OK so I agree using annotations would be technically cleaner.  I'm not sure how much it really matters given the controllerconfig CRD is an internal implementation detail of the MCO, but I took a crack at the change.

Where I got stuck is that in the operator code it passes around a `ControllerConfigSpec` - i.e. not an API object with annotations.  Once created it gets stashed in a `renderConfig` via `getRenderConfig()` and...after that I can't find the bit of code that is actually updating the API object.  Anyone know where that is?

yuqi-zhang:
Hmm if I understand your question correctly, it does look a bit confusing to me as well. I think we sync the object internally in syncRenderConfig and then a bit later:
https://github.com/openshift/machine-config-operator/blob/f2489433faacc32a7ead1caafcf076b943ffc9d8/pkg/operator/sync.go#L394

We sync the API object here?

cgwalters:
Oh wow I understand now, we are parsing the YAML and then doing:
```
{{toYAML .ControllerConfig | toString | indent 2}}
```

But why the heck are we doing that versus just creating the struct in Go?

cgwalters:
OK I took a swing at this as a separate PR https://github.com/openshift/machine-config-operator/pull/2206

yuqi-zhang:
Looking at the history individual fields were used until Abhinav decided to just render the whole config into yaml as he added more fields, and that hasn't been touched since then.

I agree we should just do it in go.

runcom:
so should we close this one in favor of #2206

cgwalters:
> so should we close this one in favor of #2206

Well; this one is known working, the other isn't.  It's not clear to me that reworking it is worth it, but I'll keep trying a bit.

cgwalters:
OK https://github.com/openshift/machine-config-operator/pull/2206 looks like it's working now.

PR #2113, opened 2020-09-21
Bug 1875237: Add etcdDiscoveryDomain back to controllerconfig crd

cybertron:
This was only deprecated in 4.6, and removing it completely in the
same release breaks upgrades of platforms that use the value in
their 4.5 templates. It appears that the controllerconfig is getting
upgraded before the templates, so when MCO goes to re-render the
old templates with the new controllerconfig we end up with a blank
value for EtcdDiscoveryDomain references. This value should stay in
the crd until 4.7 when it can be safely removed because the 4.6
templates no longer have any references to it.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/controllerconfig.crd.yaml
+++ b/manifests/controllerconfig.crd.yaml
@@ -68,6 +68,9 @@ spec:
             clusterDNSIP:
               description: clusterDNSIP is the cluster DNS IP address
               type: string
+            etcdDiscoveryDomain:
+              description: etcdDiscoveryDomain is deprecated, use infra.status.etcdDiscoveryDomain instead
+              type: string
             images:
               description: images is map of images that are used by the controller
                 to render templates under ./templates/
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -618,6 +618,9 @@ spec:
             clusterDNSIP:
               description: clusterDNSIP is the cluster DNS IP address
               type: string
+            etcdDiscoveryDomain:
+              description: etcdDiscoveryDomain is deprecated, use infra.status.etcdDiscoveryDomain instead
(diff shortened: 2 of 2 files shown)


cybertron:
@kikisdeliveryservice I think this should fix the upgrade problem on baremetal and friends. I haven't managed to test upgrades locally yet because I'm not sure how to build a custom release that will work for upgrade, but after this I can see that the value is back in the controllerconfig on 4.6, which I'm 99% sure will fix the problem.

kikisdeliveryservice:
> @kikisdeliveryservice I think this should fix the upgrade problem on baremetal and friends. I haven't managed to test upgrades locally yet because I'm not sure how to build a custom release that will work for upgrade, but after this I can see that the value is back in the controllerconfig on 4.6, which I'm 99% sure will fix the problem.

Ah so it is what we found Friday!

You can use `clusterbot build` from this PR to make an image check it out via @clusterbot help

cgwalters:
> It appears that the controllerconfig is getting upgraded before the templates, 

This would be fixed by https://github.com/openshift/machine-config-operator/pull/2112 I believe.

cgwalters:
Needs a `make bindata` at least.

Not entirely sure I understand the fix - is it that the new MCO is updating the CRD first which drops the now-invalid field?  The more I think about this the more I feel we really want https://github.com/openshift/machine-config-operator/pull/2112

That said this PR seems clearly safe at least.

cgwalters (approved):
/approve
based on above

ashcrow (approved):

cybertron:
Ran make update so it should pass the verify job now.

#2112 would also probably fix this and would make it easier to reason about the upgrade process, so +1 to that as well.

PR #2118, opened 2020-09-23
Inject version into controllerconfig, refuse mismatches

cgwalters:

(cherry picked from commit 1239f2580316cd65838349195fa053665d561959)
Cherry pick from: https://github.com/openshift/machine-config-operator/pull/2112


cgwalters:
Upgrade from this PR to 4.6 (without this patch): https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1308892489008025600

Though any potentially relevant logs would be in the previous mcc pod which have been gc'd unfortunately.

yuqi-zhang (approved):
Will do some testing but the general idea LGTM

cgwalters:
Going to do proper :cherries: picks for this.

PR #2120, opened 2020-09-25
Bug 1882667: daemon: Unlink encapsulated config in once-from mode

cgwalters:
Unconditionally remove this file in the once-from (classic RHEL)
case.  We use this file to suppress things like kubelet and SDN
starting on CoreOS during the firstboot/pivot boot, but there's
no such thing on classic RHEL.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -517,6 +517,17 @@ func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
 func (dn *Daemon) RunOnceFrom(onceFrom string, skipReboot bool) error {
 	dn.skipReboot = skipReboot
 
+	// Unconditionally remove this file in the once-from (classic RHEL)
+	// case.  We use this file to suppress things like kubelet and SDN
+	// starting on CoreOS during the firstboot/pivot boot, but there's
+	// no such thing on classic RHEL.
+	_, err := os.Stat(constants.MachineConfigEncapsulatedPath)
+	if err == nil {
+		if err := os.Remove(constants.MachineConfigEncapsulatedPath); err != nil {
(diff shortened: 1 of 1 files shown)


cgwalters:
xref https://github.com/openshift/openshift-ansible/blob/2cf2144dff1241702a0f2506f6dd66334341a919/roles/openshift_node/tasks/apply_machine_config.yml

kikisdeliveryservice:
Makes sense to me!! As this file isn't removed and can break other things :+1: 

/lgtm

kikisdeliveryservice:
We hit this in : BZ 1882667 so I'll link but im not sure if this fix closes the BZ alone @dcbw

pkg/daemon/daemon.go
```diff
@@ -517,6 +517,17 @@ func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
...
+	// Unconditionally remove this file in the once-from (classic RHEL)
+	// case.  We use this file to suppress things like kubelet and SDN
+	// starting on CoreOS during the firstboot/pivot boot, but there's
+	// no such thing on classic RHEL.
+	_, err := os.Stat(constants.MachineConfigEncapsulatedPath)
+	if err == nil {
+		if err := os.Remove(constants.MachineConfigEncapsulatedPath); err != nil {
+			return errors.Wrapf(err, "failed to remove %s", constants.MachineConfigEncapsulatedPath)
+		}
+	}
```
runcom, line 529: this should be run after the once from is run - this is a RHEL node so when this run we're effectively writing the encapsulated mc at the end of this, which results, obviously, in the file being there. We should move this at the end.

runcom:
@cgwalters ptal https://github.com/openshift/machine-config-operator/pull/2125

PR #1527, opened 2020-03-03
Bug 1809611: OpenStack: Set coredns forward policy to sequencial

mandre:
The default policy for CoreDNS forward plugin being random [1], the
cluster may not be able to resolve the OpenStack API hostname when the
user sets multiple external DNS resolvers such as
['172.31.8.1','8.8.8.8'], where 172.31.8.1 knows to resolve internal
hostnames and 8.8.8.8 is a public resolver that doesn't know how to
resolve internal hostnames.

Switch to the sequencial policy to simulate the libc behavior and not
surprise users who expect the DNS servers specified via `externalDNS`
option to have a hierarchy.

[1] https://coredns.io/plugins/forward/

PR #2122, opened 2020-09-28
Bug 1882209: baremetal & friends: Set coredns forward policy to sequential

cybertron:
This was already done for openstack in #1527, but not for baremetal
because we don't have any separation between external and internal
hostnames. However, in environments where the DNS servers are
misconfigured such that the second one listed in resolv.conf does
not properly resolve external names, this can cause intermittent
resolution failures from our coredns because it randomly chooses to
use the broken server.

Switching the forward policy to sequential will avoid potentially
confusing issues where a set of DNS servers works fine standalone
but breaks when configured as the forwarding upstreams in coredns.

**- What I did**

**- How to verify it**

### DIFF
--- a/manifests/on-prem/coredns-corefile.tmpl
+++ b/manifests/on-prem/coredns-corefile.tmpl
@@ -2,7 +2,9 @@
     errors
     health :18080
     mdns {{ .ControllerConfig.DNS.Spec.BaseDomain }} {{`{{.Cluster.MasterAmount}}`}} {{`{{.Cluster.Name}}`}} {{`{{.NonVirtualIP}}`}}
-    forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}}
+    forward . {{`{{- range $upstream := .DNSUpstreams}} {{$upstream}}{{- end}}`}} {
+        policy sequential
+    }
     cache 30
     reload
     template IN {{`{{ .Cluster.IngressVIPRecordType }}`}} {{ .ControllerConfig.DNS.Spec.BaseDomain }} {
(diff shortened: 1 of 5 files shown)


runcom:
ping @cybertron

cybertron:
Sorry, I missed that this was in merge conflict. Should be ready to go now.

kikisdeliveryservice:
@yboaron @celebdor PTAL

dougsland:
This patch make sense for ovirt as well. Going to give a try.

cybertron:
Okay, I've rebased this to pick up the dedupe changes and to apply it to the other on-prem platforms. It's causing confusing issues on all platforms.

@mandre @rgolangh @jcpowermac @patrickdillon @bcrochet I think with this we can dedupe the corefile template too. I believe the only differences between them were this sequential change and whether they use the template or a db file. The latter is an implementation detail that yields effectively the same results so it shouldn't block us.

Is anyone aware of any other differences I may have missed that would cause issues if we consolidated the coredns config?

mandre:
> Is anyone aware of any other differences I may have missed that would cause issues if we consolidated the coredns config?

OpenStack would be fine switching over to the on-prem template for corefile, at least I don't foresee any issue.

cybertron:
Looks like the ovirt job has been red for nearly two weeks. It's unlikely that failure is related to this patch.

dougsland:
tested in my lab, it seems resolved the issue reported on ovirt
/lgtm

kikisdeliveryservice:
Are any other sign offs needed based on the affected platforms?

cybertron:
Maybe @jcpowermac or @patrickdillon . This is a pretty safe change though. It's just making coredns behave the same way the system resolver would.

rgolangh:
lgtm for kubevirt platform

dougsland:
Would be nice to merge this one soon.

kikisdeliveryservice:
/assign @jcpowermac @patrickdillon 

PTAL

/retest

cybertron:
@kikisdeliveryservice I think we have what we need to merge this now.

dougsland:
just a friendly ping for the merge :) @mandre @kikisdeliveryservice

sandrobonazzola:
@ashcrow, @cgwalters, @darkmuggle, @kikisdeliveryservice , @runcom , @sinnykumari , @yuqi-zhang anything pending for getting this in?

sinnykumari (approved):
LGTM
All tests are green and we have agreement between different platform teams

Issue #2086, opened 2020-09-15
panic after nil pointer dereference in pkg/daemon/update.go


mhrivnak:
**Description**

A particular arrangement of data in a File object in ignition will cause MCO to panic.

OCP build: `4.6.0-0.nightly-2020-08-31-220837`

**Steps to reproduce the issue:**
1. use openshift-baremetal-install to generate ignition files for bare metal provisioning
2. boot a host with bootstrap.ign

The `/etc/motd` file in this case is the one causing the panic, and its entry in the ignition file looks like this:

```
      {
        "group": {},
        "overwrite": false,
        "path": "/etc/motd",
        "user": {
          "name": "root"
        },
        "append": [
          {
            "source": "data:text/plain;charset=utf-8;base64,VGhpcyBpcyB0aGUgYm9vdHN0cmFwIG5vZGU7IGl0IHdpbGwgYmUgZGVzdHJveWVkIHdoZW4gdGhlIG1hc3RlciBpcyBmdWxseSB1cC4KClRoZSBwcmltYXJ5IHNlcnZpY2VzIGFyZSByZWxlYXNlLWltYWdlLnNlcnZpY2UgZm9sbG93ZWQgYnkgYm9vdGt1YmUuc2VydmljZS4gVG8gd2F0Y2ggdGhlaXIgc3RhdHVzLCBydW4gZS5nLgoKICBqb3VybmFsY3RsIC1iIC1mIC11IHJlbGVhc2UtaW1hZ2Uuc2VydmljZSAtdSBib290a3ViZS5zZXJ2aWNlCg==",
            "verification": {}
          }
        ],
        "contents": {
          "verification": {}
        },
        "mode": 420
      },
```

**Describe the results you received:**
```
time=\"2020-09-14T21:48:21Z\" level=info msg=\"I0914 21:48:21.264243       1 update.go:1265] Writing file \\\"/etc/ignition-machine-config-encapsulated.json\\\"\\n\"\n
time=\"2020-09-14T21:48:21Z\" level=info msg=\"I0914 21:48:21.265137       1 update.go:1265] Writing file \\\"/etc/motd\\\"\\n\"\n
time=\"2020-09-14T21:48:21Z\" level=info msg="panic: runtime error: invalid memory address or nil pointer dereference\n[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x16833ea]\n\ngo>
[...]

PR #2123, opened 2020-09-28
Bug 1879283: *: add nil check when decoding File.Contents.Source

yuqi-zhang:
In ignition spec 3, the File.Contents.Source was changed to a pointer,
thus causing a panic when it is nil when we dereference during decodes.

Also explicitly reject Append sections of Ignition File sections, since
they are not supported anyways and should fail eventually.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1448,13 +1448,20 @@ func checkV2Units(units []ign2types.Unit) error {
 // no need to check for overwrites.
 func checkV3Files(files []ign3types.File) error {
 	for _, f := range files {
+		if len(f.Append) > 0 {
+			return fmt.Errorf("found an append section when checking files. Append is not supported")
+		}
 		mode := defaultFilePermissions
 		if f.Mode != nil {
 			mode = os.FileMode(*f.Mode)
 		}
-		contents, err := dataurl.DecodeString(*f.Contents.Source)
-		if err != nil {
-			return errors.Wrapf(err, "couldn't parse file %q", f.Path)
+		contents := &dataurl.DataURL{}
+		if f.Contents.Source != nil {
+			var err error
+			contents, err = dataurl.DecodeString(*f.Contents.Source)
+			if err != nil {
+				return errors.Wrapf(err, "couldn't parse file %q", f.Path)
+			}
 		}
 		if err := checkFileContentsAndMode(f.Path, contents.Data, mode); err != nil {
 			return err
@@ -1470,6 +1477,9 @@ func checkV2Files(files []ign2types.File) error {
 		// skip over checked validated files
 		if _, ok := checkedFiles[f.Path]; ok {
 			continue
+		}
+		if f.Append {
+			return fmt.Errorf("found an append section when checking files. Append is not supported")
 		}
 		mode := defaultFilePermissions
 		if f.Mode != nil {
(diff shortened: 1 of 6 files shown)


yuqi-zhang:
fixes https://github.com/openshift/machine-config-operator/issues/2086

pkg/daemon/update.go
```diff
@@ -1287,9 +1287,20 @@ func (dn *Daemon) writeFiles(files []ign3types.File) error {
 	for _, file := range files {
 		glog.Infof("Writing file %q", file.Path)
 
-		contents, err := dataurl.DecodeString(*file.Contents.Source)
-		if err != nil {
-			return err
+		// We don't support appends in the file section, so instead of waiting to fail validation,
+		// let's explicitly fail here.
+		if len(file.Append) > 0 {
+			return fmt.Errorf("found an append section when writing files. Append is not supported")
```
runcom, line 1293: to not regress in 4.6 we should just move this to a warning right?

pkg/daemon/daemon.go
```diff
@@ -1448,13 +1448,20 @@ func checkV2Units(units []ign2types.Unit) error {
 // no need to check for overwrites.
 func checkV3Files(files []ign3types.File) error {
 	for _, f := range files {
+		if len(f.Append) > 0 {
+			return fmt.Errorf("found an append section when checking files. Append is not supported")
+		}
```
runcom, line 1453: I think all this errors (in other files as well) should be just warning till we fix the wrong behavior - isn't erroring out a change in behavior in itself even if it's correct

yuqi-zhang: So I thought about this a bit more. The main thing here is that even today, the files check should be failing if you were to use append. In 4.5 https://github.com/openshift/machine-config-operator/blob/release-4.5/pkg/daemon/daemon.go#L1384 if you have a MC with append, it will interpret as empty, which the file itself obviously shouldn't be. This should cause an immediate unexpected on-disk state. I can test this but I'm pretty sure this should fail. Instead of being vague about it I thought we might as well make it explicit.

pkg/daemon/update.go
yuqi-zhang: See below, I don't think this should have ever worked. Just moving error detection.

pkg/daemon/daemon.go
runcom: I see - so yeah, if in 4.5 we're still going degraded this makes total sense - just checking to avoid any break in behavior (even if this would be correct anyway)

yuqi-zhang: I tested just now on 4.5, if you try to install with a `append` section this will show up:

```
I0929 15:01:14.463348    2250 daemon.go:1013] Validating against pending config rendered-worker-1fbc77069595714b1690fb1060794617
E0929 15:01:14.694195    2250 daemon.go:1412] content mismatch for file /etc/motd (-want +got):
```

degrading the pool during installation

yuqi-zhang:
Added another commit:

Ignition spec 2 has append as a bool, whereas spec 3 has it as a
array, so the validation will fail either or. Remove that validation
and instead have the MCD fail validation as we would expect for
other ignition errors.

Tested with adding a config with append pre-install and post-install. Pre-install fails with: 
`unexpected on-disk state validating against rendered-worker-d466ffba1412d60c8f0ffe1d19a6a1ae: found an append section when checking files. Append is not supported`
Post-install fails with: 
`Marking Unreconcilable due to: can't reconcile config rendered-worker-343463c8965575bc9a0a7a89ab374428 with rendered-worker-2e380f3dd3bd887a9a5065279ee36e76: ignition file /etc/motd includes append: unreconcilable`

So errors are more clear and no panics.

yuqi-zhang:
/retest

test failure looks a bit odd, checking if I can replicate

PR #2124, opened 2020-09-29
[WIP] Bug 1879565: workaround for NM not parsing DHCPv6 hostname

celebdor:
Current RHCOS NetworkManager does not parse IPv6 hostname options, so on environments without reverse DNS record for the node address, the node name would never be the configured one in DHCPv6.

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
@@ -3,7 +3,7 @@ path: "/etc/NetworkManager/dispatcher.d/30-resolv-prepender"
 contents:
   inline: |
     #!/bin/bash
-    set -eo pipefail
+    set -eEo pipefail
     IFACE=$1
     STATUS=$2
 
@@ -19,8 +19,6 @@ contents:
     {{end -}}
     {{end -}}
 
-    # If $DHCP6_FQDN_FQDN is not empty and is not localhost.localdomain
-    [[ -n "$DHCP6_FQDN_FQDN" && "$DHCP6_FQDN_FQDN" != "localhost.localdomain" && "$DHCP6_FQDN_FQDN" =~ "." ]] && hostnamectl set-hostname --static --transient $DHCP6_FQDN_FQDN
     case "$STATUS" in
         up|dhcp4-change|dhcp6-change)
         >&2 echo "NM resolv-prepender triggered by ${1} ${2}."
@@ -49,6 +47,39 @@ contents:
             show \
             "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
             "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
+        # check if we got DHCPv6 hostname. The value would usually be in
+        # DHCP6_FQDN_FQDN but NM internal client does not parse the option
+        # NM only parses it for IPv4
+        if [[ "$NAMESERVER_IP" =~ ":" ]]; then
+            LEASE_FILE=$(mktemp)
+            readonly CONT_NAME="dhcpv6-hostname-workaround"
+
+            trap "podman rm -f $CONT_NAME" ERR
+            /usr/bin/podman run --rm -d --name "$CONT_NAME" \
+                --authfile /var/lib/kubelet/config.json \
+                --net=host \
+                --entrypoint /usr/sbin/dhclient \
+                -v "${LEASE_FILE}:/tmp/workaround.lease:Z" \
+                {{ .Images.baremetalRuntimeCfgImage }} \
+                -d  -sf /bin/true -lf /tmp/workaround.lease -6 -v -D LL "$IFACE"
+            DHCPv6_DHCLIENT_TIMEOUT=15
+            while : ; do
+                mapfile -t NAME_DATA < <(awk '/(host|domain)-name / {print substr($3, 2, length($3) - 3)}' "$LEASE_FILE")
+                if [[ ${#NAME_DATA[@]} -eq 2 ]]; then
+                    hostnamectl set-hostname --static --transient "${NAME_DATA[0]}.${NAME_DATA[1]}"
+                    break
+                else
+                    if [[ DHCPv6_DHCLIENT_TIMEOUT -gt 0 ]]; then
+                        sleep 0.5
+                        DHCPv6_DHCLIENT_TIMEOUT=$((DHCPv6_DHCLIENT_TIMEOUT - 1))
+                    else
+                        break
+                    fi
+                fi
+            done
+            podman rm -f "$CONT_NAME"
+        fi
+
         DOMAIN="{{.Infra.Status.EtcdDiscoveryDomain}}"
         if [[ -n "$NAMESERVER_IP" ]]; then
             >&2 echo "NM resolv-prepender: Prepending 'nameserver $NAMESERVER_IP' to /etc/resolv.conf (other nameservers from /var/run/NetworkManager/resolv.conf)"


templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,37 @@ contents:
             show \
             "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
             "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
+        # check if we got DHCPv6 hostname. The value would usually be in
+        # DHCP6_FQDN_FQDN but NM internal client does not parse the option
+        # Only if it is IPv4
+        if grep -q ":" "$NAMESERVER_IP"; then
```
cybertron, line 53: Is it possible this could be an issue in dual stack? I realize we don't care about that yet, but we will. I'm wondering if it's possible we'll have ipv6 addresses where the nameserver is ipv4. Maybe we should look at IP6_NUM_ADDRESSES to detect an ipv6 env?

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,37 @@ contents:
...
+                -d  -sf /bin/true -lf /tmp/workaround.lease -6 -v "$IFACE"
+            DHCPv6_DHCLIENT_TIMEOUT=15
+            while : ; do
+                mapfile -t NAME_DATA < <(awk '/(host|domain)-name / {print substr($3, 2, length($3) - 3)}' "$LEASE_FILE")
+                if [[ ${#NAME_DATA[@]} -eq 2 ]]; then
+                    hostnamectl set-hostname --static --transient "${NAME_DATA[0]}.${NAME_DATA[1]}"
+                    break
+                else
+                    if [[ DHCPv6_DHCLIENT_TIMEOUT -gt 0 ]]; then
+                        sleep 0.5
```
cybertron, line 71: This means we'll only wait for 7.5 seconds, right? That feels a bit short to wait for a DHCP response.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,37 @@ contents:
...
+                        DHCPv6_DHCLIENT_TIMEOUT=$((DHCPv6_DHCLIENT_TIMEOUT - 1))
+                    else
+                        break
+                    fi
+                fi
+            done
+            podman rm -f "$CONT_NAME"
```
cybertron, line 78: I wonder if we should put this in a trap so it will still get run even if an earlier command fails. set -e means we could potentially exit the script before this point.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,37 @@ contents:
...
+            /usr/bin/podman run --rm -d --name "$CONT_NAME" \
+                --authfile /var/lib/kubelet/config.json \
+                --net=host \
+                --entrypoint /usr/sbin/dhclient \
+                -v "${LEASE_FILE}:/tmp/workaround.lease:Z" \
+                {{ .Images.baremetalRuntimeCfgImage }} \
+                -d  -sf /bin/true -lf /tmp/workaround.lease -6 -v "$IFACE"
+            DHCPv6_DHCLIENT_TIMEOUT=15
+            while : ; do
+                mapfile -t NAME_DATA < <(awk '/(host|domain)-name / {print substr($3, 2, length($3) - 3)}' "$LEASE_FILE")
```
cybertron, line 65: Just for reference for anyone else looking at this, the part of the lease file that we care about looks something like this:

  option host-name "my-hostname";
  option domain-name "my.domain";

That helps make sense of some of the string processing that happens here.

cybertron (commented):
Overall looks reasonable. A few comments inline, but I think this should do the trick.

cybertron:
Also, I'm doing a test locally with rdns hostname resolution disabled to see if this fixes the problem.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,37 @@ contents:
             show \
             "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
             "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
+        # check if we got DHCPv6 hostname. The value would usually be in
+        # DHCP6_FQDN_FQDN but NM internal client does not parse the option
+        # Only if it is IPv4
```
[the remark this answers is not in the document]
celebdor: I meant "It only parses it if it is IPv4"

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
celebdor: I'll increase. Thanks!

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
celebdor: That's a very good point.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
celebdor: Yup!

celebdor:
@cybertron it has a big bug. Which is that I need to generate the DUID-LL after mktemp so that dhclient does not generate a random one. Otherwise we get extra (unused) leases per node.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
celebdor: When we support dual stack, I assume that there will be a PR to drop this hack and we can handle this case then :-)

celebdor:
> @cybertron it has a big bug. Which is that I need to generate the DUID-LL after mktemp so that dhclient does not generate a random one. Otherwise we get extra (unused) leases per node.

passing -D LL should be enough.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
cybertron: Oh, good point. We should have the NM fix by then.

cybertron: That being said, this doesn't actually do what you want. grep treats the second string as a filename, not the thing to be grepped. I think you want:

if [[ "$NAMESERVER_IP" =~ ":"]]; then

This caused my test to fail. I'm retesting with it fixed.

templates/common/baremetal/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -49,6 +47,39 @@ contents:
             show \
             "{{.Infra.Status.PlatformStatus.BareMetal.APIServerInternalIP}}" \
             "{{.Infra.Status.PlatformStatus.BareMetal.IngressIP}}")
+        # check if we got DHCPv6 hostname. The value would usually be in
+        # DHCP6_FQDN_FQDN but NM internal client does not parse the option
+        # NM only parses it for IPv4
+        if [[ "$NAMESERVER_IP" =~ ":" ]]; then
+            LEASE_FILE=$(mktemp)
+            const CONT_NAME="dhcpv6-hostname-workaround"
```
cybertron, line 55: const is not a thing in bash. I think you want readonly instead.

celebdor: Damn it! This happens to me for helping somebody fix some javascript earlier this week.

PR #2125, opened 2020-09-29
Bug 1882667: pkg/daemon: rm the encapsulated mc at the right time

runcom:
Bringing up a RHEL7 node requires the ansible playbook to fetch the
ignition config from the MCS and the run the MCD in onceFrom mode.
We want the encapsulated MC file to be unconditionally removed on RHEL
nodes, thus in onceFrom we can just rm the file.
The problem is we were deleting the file before the onceFrom mode wrote
the file itself, resulting in the file always being there.
This patch moves the removal _after_ the onceFrom writes all files from
ignition, including the encapsulated path, that we'll later now remove.

Just a note that leveraging that file isn't something the MCO advertises or exposes as an API to derive any kind of logic like the OVN components imply. That's just wrong and from now on, unsupported. I'll work with them to find another way.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -516,18 +516,6 @@ func (dn *Daemon) detectEarlySSHAccessesFromBoot() error {
 // RunOnceFrom is the primary entrypoint for the non-cluster case
 func (dn *Daemon) RunOnceFrom(onceFrom string, skipReboot bool) error {
 	dn.skipReboot = skipReboot
-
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
> The problem is we were deleting the file before the onceFrom mode wrote
> the file itself, resulting in the file always being there.

Oooh

kikisdeliveryservice (approved):
/lgtm

cgwalters (commented):
Yep, thanks for fixing this!

Issue #1720, opened 2020-05-08
machine-specific machineconfigs


cgwalters:
Ironically today MachineConfig objects target a *pool* - not a specific machine.

Today people who want to e.g. perform static IP addressing are sometimes running `openshift-install generate ignition` and then hand-editing the result to include e.g. `/etc/sysconfig/network-scripts/ifcfg-foo` configs. 

Whereas if we had a way to provide "machine specific machineconfigs" then the admins could provide those MCs as [additional manifests](https://github.com/openshift/machine-config-operator/issues/1619) and it'd all Just Work.

But...this gets into the "node identity" problem.  We'd need to define a way for the MCS to identify the node it's serving the config to.

Perhaps the simplest thing is to assume anyone who wants this is statically assigning hostnames, and we change Ignition to include a header with the node's hostname.

(I guess we could do something like a reverse lookup of the requester's IP address too)

One messy aspect of this too is that we can't include these bits configs in the main `rendered-$pool` configs which means the MCS needs to generate it internally.  (I guess we could create a separate `rendered-node-$x` config too?)

michaelgugino:
> Perhaps the simplest thing is to assume anyone who wants this is statically assigning hostnames, and we change Ignition to include a header with the node's hostname.

Definitely we should do this with or without machine-specific machineconfigs.

> (I guess we could do something like a reverse lookup of the requester's IP address too)

This is not guaranteed to work in all environments, and I think we discovered during troubleshooting 4.4 release that the IPs that show up in the MCS during first boot are VIPs, not instance IPs.

Including the hostname and IP on every request to the MCS would greatly aid in determining when a machine failed to boot/request an ignition vs failing to join cluster after getting ignition file.

cgwalters:
One way to do this today is - for each machine one wants to have a custom config for, manually set things up to pass it a custom derived *pointer* config.

For example in a PXE install scenario, take the Ignition output from `openshift-install create ignition-configs` and edit that, rather than trying to customize the full Ignition config returned from the MCS.

cgwalters:
Perhaps one simple approach here is:

- user injects `/etc/hostname` in the pointer Ignition config
- Ignition detects this and adds it as a header when doing requests for merged configs (like the MCO)
- MCO reads this header and adds in to the rendered Ignition (from all machineconfig objects with a label for that hostname)

[...]

PR #2126, opened 2020-09-29
Bug 1881703: Revert https://github.com/openshift/machine-config-operator/pull/1792

runcom:
This PR just reverts Manage the ignition stub config #1792
More information can be found at https://docs.google.com/document/d/1TnnfPgFim-e895MD0msOGN9tyPBk3CM8cmSewFmgzDQ/edit#

TL;DR; users can customize the pointer ignition config adding files and whatever ignition supports. We weren't aware of that and assumed only the MCS endpoint and the cert would be managed by the installer - given where we are and the assessment that it's safe to just revert, let's do it and re-think the work in light of the new findings (as it'll require a translation as well..)


yuqi-zhang (commented):
I think given the timelines I'm +1 for the revert. Since this isn't an exact one-to-one I just want to make sure the diffs are good:
1. We did not seem to have removed the `KubeMAOSharedInformer` in controller_context.go
2. We did not revert `renderAsset` func typing in render.go
3. We didn't remove one line from go.sum
4. We didn't remove variable definitions (this is fine)
Those are the diffs I saw

cgwalters:
> TL;DR; users can customize the pointer ignition config adding files and whatever ignition supports.

Right, we do need to support that.  It is actually one of the only sane ways to have *per machine* data (see also https://github.com/openshift/machine-config-operator/issues/1720 ).

cgwalters:
Are you proposing to revert https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/user-data-secret-managed.md fully for 4.6?

The more I look at this BZ the more I feel like being able to hand-edit the Ignition *in a machineAPI managed setup* was somewhat of an accidental feature.  As I commented in the BZ I think they should be using MachineConfig which would transparently just work.

(Actually at a higher level I think they could be taking better advantage of all the networking enhancements in 4.6 but it looks like they're using baremetal IPI which doesn't use those yet)

runcom:
> Are you proposing to revert https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/user-data-secret-managed.md fully for 4.6?

yes, while the BZ can be "fixed" by shipping an MC I'm more worried about the change in behavior that we've effectively introduced with an half-baked feature like that - we should really take into account the translation from 4.5 or just v2 and also we don't really know if users are using this but since it's kind of advertised it's a safe assumption.

The thing is there could be clusters where we would just overwrite and delete any pointer ignition config customization done at installation so I think it's safe to revert and perfection the implementation in a future release, does it sound good?

kikisdeliveryservice:
+1 on removing and pushing a redo later

cgwalters:
There's definitely risk to trying to back this out at the last minute - if the issue can be addressed by having the user provide MCs, it also seems reasonable to me to mark this as a known issue and move on.

kikisdeliveryservice:
> There's definitely risk to trying to back this out at the last minute - if the issue can be addressed by having the user provide MCs, it also seems reasonable to me to mark this as a known issue and move on.

@cgwalters just to clarify: you're saying leave it as-is and say "the supported way to do this is via a machine-config" if we receive a bz? But later fix or not ?

runcom:
> There's definitely risk to trying to back this out at the last minute - if the issue can be addressed by having the user provide MCs, it also seems reasonable to me to mark this as a known issue and move on.

right, there's definitely some risk at reverting as well after ~1 month that this is soaking into master... I think I'd be open writing something in bold that says "if you were doing this, use an MC from now on when installing a cluster" - but then, what do we do about already installed clusters? we're still screwing up their scale-up but I think we can also document a way out of that (if there's one?)?

cgwalters:
> therefore we can have clusters upgrading from 4.5 with customized pointer configuration that we’ll just leave behind, effectively breaking any scale up scenario.

Hmmm.  That is indeed more of a potential problem.  The BZ reporter was talking about new installs, but we also clearly want to support them upgrading in-place to 4.6.  

So far we only have evidence that 4.5 baremetal users are customizing the pointer config, but not for other platforms.  I strongly suspect the percentage of cloud/IaaS users (azure/gcp/aws/etc) doing this is effectively zero.

> @cgwalters just to clarify: you're saying leave it as-is and say "the supported way to do this is via a machine-config" if we receive a bz? But later fix or not ?

Honestly I'm mostly arguing for a bit more consideration/debate, I just want to be sure we have explored the alternatives and weighed the drawbacks a bit more.

cgwalters:
So if we do the revert...it means that the MCO won't manage the worker userdata, which shouldn't break anything in new 4.6 installs or upgrades, all that we have lost is *future* support for updating bootimages per
> when we'll grow the ability to update bootimages, the new bootimages will understand v3 and they'll grab the managed secret which is v3
from the enhancement.  Right?

runcom:
> Honestly I'm mostly arguing for a bit more consideration/debate, I just want to be sure we have explored the alternatives and weighed the drawbacks a bit more.

oh yeah, I think I didn't mean to push the revert (and sorry if I did) - I opened the reverts to have something immediately actionable given where we are time-wise in the release - let me put an hold on this meanwhile we discuss 

/hold

runcom:
> So if we do the revert...it means that the MCO won't manage the worker userdata, which shouldn't break anything in new 4.6 installs or upgrades, all that we have lost is _future_ support for updating bootimages per

yes, when doing the bootimages upgrade feature, we'd need to re-take this into account

cgwalters:
OK I'm leaning towards revert indeed, let's just be "on point" to make sure that everything works afterwards.

On this specific PR - we could also just comment out the code doing the secret sync right?  That'd make it less of a conflict-fest to reintroduce later in 4.7.  But I'm fine as is too.

yuqi-zhang:
I'm still +1 on the revert here too. Just to make sure:
1. comments in https://github.com/openshift/machine-config-operator/pull/2126#pullrequestreview-498617482
2. we should revert machine-api first, then installer, then this

yuqi-zhang:
cross linking https://github.com/openshift/machine-api-operator/pull/715 and https://github.com/openshift/installer/pull/4228

machine-api is merged, I will approve installer, and we can lgtm this after we're sure no regressions happen

runcom:
> * We did not seem to have removed the `KubeMAOSharedInformer` in controller_context.go

removed

> * We did not revert `renderAsset` func typing in render.go

this shouldn't affect anything, we use the type as interface and it made more sense to type the func with that - reverted tho

> * We didn't remove one line from go.sum

I've run `make go-deps` trusting it did all the things so that won't affect runtime right?

> * We didn't remove variable definitions (this is fine)
>   Those are the diffs I saw

should be ok to go!

PR #2127, opened 2020-09-29
[WIP]Allow all processes to make core dumps

astoycos:
Currently we cannot see ovs coredumps when it segfaults

This should fix that and allow us to solve
https://bugzilla.redhat.com/show_bug.cgi?id=1875534

Signed-off-by: Andrew Stoycos <astoycos@redhat.com>

**- What I did**

**- How to verify it**


astoycos:
@trozet @JacobTanenbaum If we see an `ovs` segfault in one of these tests we should be able to get coredumps

astoycos:
well @trozet we had a segfault but it wasn't ovs and there's still no core dump in the must gather

PR #2131, opened 2020-09-30
Bug 1882667: Make ovs-configuration wait for network.service

trozet:
With RHEL 7 worker nodes, network service and NetworkManager will both
exist. network service runs after NetworkManager and may interfere with
ovs-configuration service. Therefore ensure we run ovs-configuration
after network.service is done.


trozet:
@danwinship @dcbw PTAL

ashcrow (approved):

PR #2132, opened 2020-09-30
Bug 1872885: add template for NM to not manage hostname on GCP

darkmuggle:
In cases where NetworkManager can set an otherwise invalid cluster
hostname, but valid for NM (between 63 and 65 characters),
NetworkManager will bypass the dispatcher hook.

This config tells NetworkManager to not set the transient hostname.
Instead, the dispatcher.d hook script will set the hostname.

### DIFF
--- a/templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml
+++ b/templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml
@@ -1,0 +1,12 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # On GCP the hostname is likely to be longer than 63 chars.
+    # /etc/networkmanager/dispatcher.d/90-long-hostname.sh will set
+    # the hostname. The following configuration allows 90-long-hostname.sh
+    # to manage setting transient hostname instead of NetworkManager itself.
+    # See: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+    #      https://bugzilla.redhat.com/show_bug.cgi?id=1872885
+    [main]
+    hostname-mode=none


darkmuggle:
I want to dig to see if there is a better way, but this might be the only option for 4.6.

darkmuggle:
/retest

the changes only affect GCP

templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml
```diff
@@ -0,0 +1,10 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # On GCP the hostname is likely to be longer than 63 chars.
+    # /etc/networkmanager/dispatcher.d/90-long-hostname.sh will
+    # the hostname. The following configuration allows 90-long-hostname.sh
+    # to manage setting transient hostname instead of NetworkManager itself.
```
miabbott, line 8: nit: drop a link to the NM docs for this setting?  https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html

templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml
```diff
@@ -0,0 +1,10 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # On GCP the hostname is likely to be longer than 63 chars.
+    # /etc/networkmanager/dispatcher.d/90-long-hostname.sh will
+    # the hostname. The following configuration allows 90-long-hostname.sh
```
miabbott, line 7:
```suggestion
    # set the hostname. The following configuration allows 90-long-hostname.sh
```

miabbott:
Two nits, but otherwise this approach looks sound for GCP.

/approve

ashcrow (approved):
Change looks good to me -- and I agree with @miabbott's suggestions

templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml
darkmuggle: Ha. I wouldn't call this a nit -- the doc llink helps explain this. Also, I added the BZ ref.

darkmuggle:
> I'm OK with this but would like to revisit this for 4.7.

@cgwalters indeed...I think this is an ugly workaround. I'm raising this with the Pillar to see it we can't get the Kubelet to deal with this in a sane manner.

darkmuggle:
@ashcrow all tested and ready to go. I launched a cluster and put this through the paces and tried every way possible to trigger this bug...so this good to go. Are there any special rules to land this for 4.6 since its a release-blocker?

PR #2135, opened 2020-10-01
docs/custom-pools: Describe a little about nodes and pools and updates

cgwalters:
This came up in chat, let's explain the "one pool per node" bit
as well as the concurrency.


cgwalters:
/bugzilla refresh
(shouldn't be required now that we've branched)

kikisdeliveryservice:
love this!

/lgtm

PR #2136, opened 2020-10-01
Bug 1880259: ovs-configuration: use NM default ethernet route metric

dcbw:
Instead of the default OVS interface route metric, which is much lower,
and causes problems when secondary interfaces are added to the machine.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1880259

@trozet @knobunc

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -77,9 +77,15 @@ contents:
         echo "MTU found for iface: ${iface}: ${iface_mtu}"
       fi
 
-      # create bridge
+      # create bridge; use NM's ethernet device default route metric (100)
       if ! nmcli connection show br-ex &> /dev/null; then
-        nmcli c add type ovs-bridge conn.interface br-ex con-name br-ex 802-3-ethernet.mtu ${iface_mtu} 802-3-ethernet.cloned-mac-address ${iface_mac}
+        nmcli c add type ovs-bridge \
+            con-name br-ex \
+            conn.interface br-ex \
+            802-3-ethernet.mtu ${iface_mtu} \
+            802-3-ethernet.cloned-mac-address ${iface_mac} \
+            ipv4.route-metric 100 \
+            ipv6.route-metric 100
       fi
 
       # store old conn for later


templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -77,9 +77,15 @@ contents:
         echo "MTU found for iface: ${iface}: ${iface_mtu}"
       fi
 
-      # create bridge
+      # create bridge; use NM's ethernet device default route metric (100)
```
trozet, line 80: I'm wondering if we should use the metric of the interface we take instead.

dcbw: Yes, we technically should. But even that doesn't solve the real problem; bonds default to 300. So if you have a bond as your default interface, set up OVS, and then hot-plug another NIC, you're still screwed.

The real fix is to make sure that when you hotplug a NIC that you don't want to be the default route, you somehow tell the guest OS that it shouldn't be the default route. THat's way out of scope for us here.

trozet: if that's the argument, then shouldn't we make our route even lower? if someone hot plugs and gets the same metric coudln't we have the same problem? Maybe we should make it 99.

dcbw: @trozet NM dynamically adjusts default route priorities to make sure they don't conflict, and since we're hardcoding one, it will make sure other devices dont' conflict. So yeah, we could do 99, but don't need to. Since we've claimed 100, all other interfaces will be assigned higher priorites (eg >= 101)

dcbw:
@mandre does this look OK?

mandre:
> @mandre does this look OK?

I didn't get a chance to try this patch or check that it really fixes https://bugzilla.redhat.com/show_bug.cgi?id=1880259 but forcing the metric for br-ex to a lower value than any interface that would be added later makes sense to me.

kikisdeliveryservice:
Does @trozet need to sign off before merging this?

trozet (approved):
/lgtm

trozet:
@kikisdeliveryservice we should see e2e-ovn-step-registry pass first before we approve it

kikisdeliveryservice:
> @kikisdeliveryservice we should see e2e-ovn-step-registry pass first before we approve it

Gotcha, thanks!

trozet:
@runcom @kikisdeliveryservice can you please approve?

PR #2137, opened 2020-10-01
Bug 1890797: Set NonVirtualIP in mdns plugin config

cybertron:
This is the same change as #1879, but applied to the non-baremetal platforms that use coredns-mdns.

### DIFF
--- a/manifests/openstack/coredns-corefile.tmpl
+++ b/manifests/openstack/coredns-corefile.tmpl
@@ -1,7 +1,7 @@
 . {
     errors
     health :18080
-    mdns {{ .ControllerConfig.EtcdDiscoveryDomain }} {{`{{.Cluster.MasterAmount}}`}} {{`{{.Cluster.Name}}`}}
+    mdns {{ .ControllerConfig.EtcdDiscoveryDomain }} {{`{{.Cluster.MasterAmount}}`}} {{`{{.Cluster.Name}}`}} {{`{{.NonVirtualIP}}`}}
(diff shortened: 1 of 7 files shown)


cybertron:
/bugzilla refresh

I'm going to have to figure out why the dependent bug is still ON_QA, but I want to verify that everything else is ready to go.

cybertron:
Note that this is not touching any baremetal templates, so the metal-ipi job failure is irrelevant.

iamemilio:
/hold
e2e-openstack was unable to reach the api

iamemilio:
/hold cancel 

never mind

mandre (commented):
/test e2e-ovirt
/test e2e-vsphere

OpenStack is green already, so :+1: from me.

kikisdeliveryservice:
@cybertron is there anyone else who should review this?

cybertron:
CI has passed on all of the affected platforms (metal-ipi was done in a previous PR, so it's not touched by this one). We should be safe to go ahead with this.

/cc @jcpowermac @rgolangh

cybertron:
/bugzilla refresh 

I had to create a new clone of the bug, but otherwise this should be ready to go now.

I will note again that the metal-ipi job failure is irrelevant because this patch doesn't touch the metal-ipi templates.

kikisdeliveryservice:
@mandre metal-ipi job has been failing since sept. is it ok to approve this without a passing run?

cybertron:
This doesn't touch any of the metal templates, so we don't need to block it on that test.

PR #2077, opened 2020-09-11
Bug 1875005: OpenStack: Don't failover api vip if loadbalanced endpoint is responding

mandre:
This ports #1893 to OpenStack platform, together with the dependent #1124 PR, and fixes from #1508 and #1604, and a sprinkle of #1768 on top.

PR #2138, opened 2020-10-02
[wip] vsphere - excessive vip migration

jcpowermac:
trying to use nopreempt to resolve vip migration


yboaron:
Because we use two track scripts to set Keepalived instance priority  I think we might see unexpected behavior when **nopreempt**  is set, for example an instance with 2 successful scripts won't preempt an instance with a single successful script .
So I'm not sure if nopreempt is a good idea with the current Keepalived track scripts config.

see [1]

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1870177

mandre:
We were also experiencing high failover rate with keepalived on openstack platform and ended up solving it with https://github.com/openshift/machine-config-operator/pull/2077 and https://github.com/openshift/machine-config-operator/pull/2077.

jcpowermac:
Thanks @mandre and @yboaron.

Just going to close this pr and deal with the issue in CI.

PR #2140, opened 2020-10-06
Bug 1880591: Make ovs-configuration write a file

trozet:
In CNO, we have no reliable way to detect if OVS should be running in
systemd or not. Using systemd or dbus is unreliable from the containers.
Checking for file existence is not foolproof because MCO lays the files
down before it reboots into 4.6. This patch makes ovs-configuration
write a file when it executes. We can use this to detect if we should
expect OVS to be running in systemd.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -4,6 +4,10 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
+    # Create file to signal that ovs-configuration has executed (we are booted into 4.6)
+    # This is necessary because systemd calls from SDN containers are unreliable, so need to look for a file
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
I don't think I can do it in this repo but:
/lgtm

danwinship:
/lgtm
too. Note that there's no corresponding git master PR because this is only needed for 4.5→4.6 upgrade

runcom:
Thanks @dcbw and @trozet :pray: hopefully this flow is now validated and should be the correct behavior on both ovn/mco sides

/approve

runcom:
those two jobs seem broken by something happened for sure elsewhere in the past - we should investigate (cc @cgwalters for the cluster-bootimages)

/override cluster-bootimages
/override e2e-aws-proxy

PR #2141, opened 2020-10-06
Bug 1882191: Add GODEBUG=x509ignoreCN=0 to systemd DefaultEnvironment

sdodson:
This sets GODEBUG=x509ignoreCN=0 system wide.

To verify install a cluster with this patch, confirm that your environment variables contain the expected value. Something like this.
```
$ oc debug node/ip-10-0-141-152.us-west-2.compute.internal
Starting pod/ip-10-0-141-152us-west-2computeinternal-debug ...
Pod IP: 10.0.141.152
sh-4.4# chroot /
sh-4.4# systemctl show-environment
GODEBUG=x509ignoreCN=0

sh-4.4# cat /etc/systemd/system/test.service
[Service]
ExecStart=echo $GODEBUG
sh-4.4# systemctl start test.service
sh-4.4# journalctl -lu test.service
-- Logs begin at Tue 2020-10-06 20:58:26 UTC, end at Tue 2020-10-06 23:47:25 UTC. --
Oct 06 23:47:23 ip-10-0-141-152 systemd[1]: Started test.service.
Oct 06 23:47:23 ip-10-0-141-152 echo[109184]: x509ignoreCN=0
Oct 06 23:47:23 ip-10-0-141-152 systemd[1]: test.service: Consumed 1ms CPU time
```

This will also need to be rendered into bootstrap ignition assets as well.

### DIFF
--- a/templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-godebug.conf.yaml
+++ b/templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-godebug.conf.yaml
@@ -1,0 +1,7 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/systemd/system.conf.d/10-default-env-godebug.conf"
+contents:
+  inline: |
+    [Manager]
+    DefaultEnvironment=GODEBUG=x509ignoreCN=0


templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-godebug.conf.yaml
```diff
@@ -0,0 +1,7 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/systemd/system.conf.d/10-default-env-godebug.conf"
+contents:
+  inline: |
+    [Manager]
+    DefaultEnvironMent=GODEBUG=x509ignoreCN=0
```
kikisdeliveryservice, line 7: EnvironMent -> Environment

cgwalters: Yeah, that's what he `Ment` clearly.

cgwalters:
Not opposed but - did the option of just patching our golang default come up?  I guess it'd affect customer generated binaries too.

BTW I think the best way to deprecate things is to add both a warning print *and* a  `sleep(5 * time.Second)` or so, if they'd done that we probably would have noticed and acted on this sooner.

Anyways weak
/approve

sdodson:
> Not opposed but - did the option of just patching our golang default come up? I guess it'd affect customer generated binaries too.

Briefly yes but it wasn't immediately obvious how to do it and given this was the documented method to fall back to previous behavior we chose the environment variable. I'm definitely not happy about setting this host wide, the other option I thought about was setting it in all the places we're setting HTTP_PROXY when proxy is enabled but that may've missed some tools which weren't deemed to need proxy but may still be connecting to TLS endpoints.

sinnykumari:
Tested with this PR, behaving as expected:

```
sh-4.4# systemctl show-environment
GODEBUG=x509ignoreCN=0

sh-4.4# cat /etc/systemd/system/test.service
[Service]
ExecStart=echo $GODEBUG

sh-4.4# systemctl start test.service

sh-4.4# journalctl -lu test.service
-- Logs begin at Wed 2020-10-07 09:18:16 UTC, end at Wed 2020-10-07 09:38:43 UTC. --
Oct 07 09:38:34 ci-ln-ybhhl32-f76d1-78ssn-worker-b-mr8sw systemd[1]: Started test.service.
Oct 07 09:38:34 ci-ln-ybhhl32-f76d1-78ssn-worker-b-mr8sw echo[48082]: x509ignoreCN=0
Oct 07 09:38:34 ci-ln-ybhhl32-f76d1-78ssn-worker-b-mr8sw systemd[1]: test.service: Consumed 1ms CPU time
```

Do we plan to revert this change in future once it stabilizes upstream? If so, adding a comment in config file will remind  us in future to revisit the change.

sdodson:
> Do we plan to revert this change in future once it stabilizes upstream? If so, adding a comment in config file will remind us in future to revisit the change.

I don't know that the Golang behavior will be reverted upstream. We'll be publishing deprecation notices in 4.6 and will consider reverting this in 4.7 or a later release depending on how confident we are that everyone has updated certificates. There's some effort to generate metrics so that we can monitor how pervasive these broken certificates are but that hasn't been done yet.

bgilbert:
Note that this won't affect Ignition itself.  We'd need to add a corresponding file to the initramfs to do that.

sdodson:
> Note that this won't affect Ignition itself. We'd need to add a corresponding file to the initramfs to do that.

Should we open a separate bug on that? That hole seems really small to me, thus far we've closed all OCP build images, now we're closing on host binaries. I don't fully understand how likely it is that Ignition will be dealing with faulty TLS endpoints, I know MCS should have valid certs and in general cloud providers are keen to adhere to standards.

bgilbert:
We should probably have an RHCOS bug, yeah.  Any customer-supplied custom Ignition fragment with a `source` attribute pointing to an HTTPS server is potentially affected.

sdodson:
@bgilbert Thanks for the input, filed https://bugzilla.redhat.com/show_bug.cgi?id=1886134

yuqi-zhang:
Given that there are no objections, and this is mostly aiming to preserve existing behaviour before we publish the deprecation, I am
/lgtm

PR #2142, opened 2020-10-07
WIP: Bug 1885365: update.go: use systemd presets to enable/disable units

yuqi-zhang:
Change unit enable/disable to write a mco systemd preset file, so
that we can properly handle units not targetting multi-user.target.wants.
The code is essentially copied from Ignition/units and emulates
Ignition behaviour for this.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -47,8 +47,17 @@ const (
 	// fipsFile is the file to check if FIPS is enabled
 	fipsFile              = "/proc/sys/crypto/fips_enabled"
 	extensionsRepo        = "/etc/yum.repos.d/coreos-extensions.repo"
+	systemdPresetFile     = "/etc/systemd/system-preset/15-mco.preset"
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
I think I should perhaps just shell out directly to `systemctl enable/disable`, Benjamin noted that presets are really just used for defaults on firstboot.

bgilbert:
The officially recommended approach for programmatic enable/disable operations at runtime is to use the [systemd D-Bus API](https://www.freedesktop.org/wiki/Software/systemd/dbus/), but I don't know if that's feasible here.

cgwalters:
> I think I should perhaps just shell out directly to systemctl enable/disable, Benjamin noted that presets are really just used for defaults on firstboot.

Yeah, I think so.  It'd be a lot simpler of a patch too.

yuqi-zhang:
closing in favour of https://github.com/openshift/machine-config-operator/pull/2145

PR #2143, opened 2020-10-07
Drop kernel-devel from supported extensions

cgwalters:
We only ship this in the oscontainer so that other tooling can
pull the oscontainer and use it to build kernel modules in a
container image.  People shouldn't install it to the host
system because that would lead to a desire for installing e.g.
`gcc` and `make` on the host system too.  Builds should
be done in a container image.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -106,13 +106,6 @@ func MergeMachineConfigs(configs []*mcfgv1.MachineConfig, osImageURL string) (*m
 		extensions = append(extensions, cfg.Spec.Extensions...)
 	}
 
-	// Ensure that kernel-devel extension is applied only with default kernel.
-	if kernelType != KernelTypeDefault {
-		if InSlice("kernel-devel", extensions) {
-			return nil, fmt.Errorf("installing kernel-devel extension is not supported with kernelType: %s", kernelType)
-		}
-	}
-
 	return &mcfgv1.MachineConfig{
 		Spec: mcfgv1.MachineConfigSpec{
 			OSImageURL:      osImageURL,
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -876,7 +876,7 @@ func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string
 }
 
 func validateExtensions(exts []string) error {
-	supportedExtensions := []string{"usbguard", "kernel-devel"}
+	supportedExtensions := []string{"usbguard"}
 	invalidExts := []string{}
 	for _, ext := range exts {
 		if !ctrlcommon.InSlice(ext, supportedExtensions) {


cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1885408

runcom:
cc @miabbott @ashcrow

runcom:
I think this needs just further discussion so holding just to have time to discuss as it changes the behavior

/hold

dunnevan:
@cgwalters how can you mount oscontainer as a pod volume?  

If the kernel module has to match and be built to match a pod container's software version - how would you go about that?  Currently, I think an initContainer makes sense for that use case (make, gcc inside the initContainer).

I think what needs to be ironed out is where and how kernel-devel can be accessed.  I.e we say container, but that could be a podman build container? a container running in a pod? etc, etc.

Since the kmod is to interact with userspace software inside a container running in a pod, I think the kernel dependencies needs to be accessible all the way up there.

cgwalters:
> @cgwalters how can you mount oscontainer as a pod volume?

Currently I'd do it the same way the MCO does - use `oc image extract` explicitly.

The other way is to run the oscontainer with a webserver as an entrypoint and access it as a service.  I think we may go that route for the MCO at some point.

cgwalters:
Well I may accept defeat on this temporarily - I added a 4.8 label for this.
But long term we really need to get away from installing kernel headers on the host.  It goes away from our containerization principle.

And further it's one of two things dragging `perl` into the host and I'd *really* like to not ship that.

jlebon:
@dunnevan Can we discuss how to move forward with this? This would be a very nice cleanup conceptually for RHCOS and should help streamline the kmod workflow as well because it doesn't require using the MCO/host as an intermediary (which costs at least one reboot).

dunnevan:
> @dunnevan Can we discuss how to move forward with this? This would be a very nice cleanup conceptually for RHCOS and should help streamline the kmod workflow as well because it doesn't require using the MCO/host as an intermediary (which costs at least one reboot).

Sure, we can discuss.  It is definitely tricky when a container needs to reach down into and interact with the kernel (or kernel modules).

Today we have a filesystem packaged inside a container.  It has two pieces, a daemon that runs inside the container and a kernel module that needs to be loaded onto the host.  The container's daemon and the kernel module versions must match.  But the kernel module also must match the kernel version.  For the most part the kernel module is a shim bridging GPL-code in the kernel and proprietary code in the daemon.

We deploy inside a pod using first an initContainer with privilege to load the module.  It has our source and small build environment, but it requires kernel-headers and kernel-devel files.  This is where we rely on the MCO kernel-devel extension (to install kernel-header and kernel-devel packages).

We read-only hostPath the following directories into the initContainer, which contains the **kernel version specific** headers and source:
```
/usr/src/kernels
/usr/include/asm
/usr/include/asm-generic
/usr/include/linux
```
The initContainer entrypoint is to build and then load the kernel module (which takes ~20 seconds).  Eventually, our daemon container starts and the filesystem can be mounted, etc etc.

When we update our software it is seamless because a new kernel module will be built during pod startup.  When the underlying kernel is updated likewise we will be rebuilt during pod startup.  We like this approach because we do not own the kernel, nor do we own it's updates, we just require it bring along the headers and symbols we need to interface with it - whether they are accessible on the host (as with MCO extensions), or through some other mechanism (such as mounting oscontainer).

We realize that an alternative approach might be pre-built kernel module binaries (for the 3 architectures we support), but we think that would be an untenable number of combinations.  We are targeting not just kernels from CoreOS, but also from other distributions that run vanilla kubernetes.

jlebon:
Your workflow of using an initContainer to build and load the kernel module on-the-fly makes sense.

So one idea is to have two initContainers: one is the machine-os-content image itself, and the other is yours. We expose an HTTP server on the machine-os-content and then your init code can fetch the needed RPMs from there. This is obviously more heavyweight though because it'd pull down the m-o-c image (and because of how the MCD fetches the m-o-c, it wouldn't hit any cache).

Honestly, maybe we should make this workflow more official and have an actual `machine-os-kernel-dev` container which has the devel and headers in its filesystem tree and then the init container (or really anything which wants to build kernel modules or build against the headers) could just use `oc image extract` to copy everything on top of its filesystem.

miabbott:
> Honestly, maybe we should make this workflow more official and have an actual `machine-os-kernel-dev` container which has the devel and headers in its filesystem tree and then the init container (or really anything which wants to build kernel modules or build against the headers) could just use `oc image extract` to copy everything on top of its filesystem.

@zvonkok has been tirelessly pursuing a similar idea of a "driver container image" that would be shipped async from OCP release payloads.  It would contain the things necessary for building out of tree modules (`kernel-devel`, `kernel-headers`, `elf-utils`, etc).  Worth getting his thoughts here.

jlebon:
Still hoping we eventually do this.

/remove-lifecycle stale

sinnykumari:
@cgwalters haven't seen any discussion happening around it, should we close this for now? We can reopen it or create a new one when we will able to drop it.

PR #2145, opened 2020-10-07
Bug 1885365: daemon: properly handle unit enable/disables

yuqi-zhang:
Remove the hardlink to multi-user.target and instead invoke systemctl
to directly enable/disable units as needed, so [Install] - wantedby
sections are properly parsed.

Also call systemctl preset on units that are deleted/changed, to
ensure "rollback" of unit enablement to system defaults. Remove
the Ignition-written preset file to account for this.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -132,8 +132,6 @@ const (
 	pathIgnition = "/usr/lib/dracut/modules.d/30ignition/ignition"
 	// pathSystemd is the path systemd modifiable units, services, etc.. reside
 	pathSystemd = "/etc/systemd/system"
-	// wantsPathSystemd is the path where enabled units should be linked
-	wantsPathSystemd = "/etc/systemd/system/multi-user.target.wants/"
 	// pathDevNull is the systems path to and endless blackhole
 	pathDevNull = "/dev/null"
 	// currentConfigPath is where we store the current config on disk to validate
@@ -455,6 +453,9 @@ func (dn *Daemon) syncNode(key string) error {
 	if dn.booting {
 		// Be sure only the MCD is running now, disable -firstboot.service
 		if err := upgradeHackFor44AndBelow(); err != nil {
+			return err
+		}
+		if err := removeIgnitionArtifacts(); err != nil {
 			return err
 		}
 		if err := dn.checkStateOnFirstRun(); err != nil {
@@ -960,6 +961,16 @@ func upgradeHackFor44AndBelow() error {
 		if err := os.Rename(constants.MachineConfigEncapsulatedPath, constants.MachineConfigEncapsulatedBakPath); err != nil {
 			return errors.Wrap(err, "failed to rename encapsulated MachineConfig after processing on firstboot")
 		}
+	}
+	return nil
+}
+
+// Remove artifacts used by ignition, that the MCO should no longer
+// use since the machine is up.
+// Currently removes the systemd preset file written by Ignition.
+func removeIgnitionArtifacts() error {
+	if err := os.Remove(constants.IgnitionSystemdPresetFile); err != nil && !os.IsNotExist(err) {
+		return errors.Wrap(err, "failed to remove Ignition-written systemd preset file")
 	}
 	return nil
 }
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1398,8 +1398,12 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config) 
 		}
 		path := filepath.Join(pathSystemd, u.Name)
 		if _, ok := newUnitSet[path]; !ok {
-			if err := dn.disableUnit(u); err != nil {
-				glog.Warningf("Unable to disable %s: %s", u.Name, err)
+			// since the unit doesn't exist anymore within the MachineConfig,
+			// look to restore defaults here, so that symlinks are removed first
+			// if the system has the service disabled
+			// writeUnits() will catch units that still have references in other MCs
+			if err := dn.presetUnit(u); err != nil {
+				glog.Infof("Did not restore preset for %s (may not exist): %s", u.Name, err)
 			}
 			if _, err := os.Stat(noOrigFileStampName(path)); err == nil {
 				if err := os.Remove(noOrigFileStampName(path)); err != nil {
@@ -1429,38 +1433,34 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config) 
 	return nil
 }
 
-// enableUnit enables a systemd unit via symlink
+// enableUnit enables a systemd unit via systemctl
 func (dn *Daemon) enableUnit(unit ign3types.Unit) error {
-	// The link location
-	wantsPath := filepath.Join(wantsPathSystemd, unit.Name)
-	// sanity check that we don't return an error when the link already exists
-	if _, err := os.Stat(wantsPath); err == nil {
-		glog.Infof("%s already exists. Not making a new symlink", wantsPath)
-		return nil
-	}
-	// The originating file to link
-	servicePath := filepath.Join(pathSystemd, unit.Name)
-	err := renameio.Symlink(servicePath, wantsPath)
+	_, err := exec.Command("systemctl", "enable", unit.Name).CombinedOutput()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "enabling unit")
 	}
-	glog.Infof("Enabled %s", unit.Name)
-	glog.V(2).Infof("Symlinked %s to %s", servicePath, wantsPath)
+	glog.Infof("Enabled systemd unit %s", unit.Name)
 	return nil
 }
 
-// disableUnit disables a systemd unit via symlink removal
+// disableUnit disables a systemd unit via systemctl
 func (dn *Daemon) disableUnit(unit ign3types.Unit) error {
-	// The link location
-	wantsPath := filepath.Join(wantsPathSystemd, unit.Name)
-	// sanity check so we don't return an error when the unit was already disabled
-	if _, err := os.Stat(wantsPath); err != nil {
-		glog.Infof("%s was not present. No need to remove", wantsPath)
-		return nil
+	_, err := exec.Command("systemctl", "disable", unit.Name).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "disabling unit")
 	}
-	glog.V(2).Infof("Disabling unit at %s", wantsPath)
+	glog.Infof("Disabled systemd unit %s", unit.Name)
+	return nil
+}
 
-	return os.Remove(wantsPath)
+// presetUnit resets a systemd unit to its preset via systemctl
+func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
+	_, err := exec.Command("systemctl", "preset", unit.Name).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "running preset on unit")
+	}
+	glog.Infof("Preset systemd unit %s", unit.Name)
+	return nil
 }
 
 // writeUnits writes the systemd units to disk
@@ -1520,22 +1520,26 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 		}
 
 		// if the unit doesn't note if it should be enabled or disabled then
-		// skip all linking.
-		// if the unit should be enabled, then enable it.
-		// otherwise the unit is disabled. run disableUnit to ensure the unit is
-		// disabled. even if the unit wasn't previously enabled the result will
-		// be fine as disableUnit is idempotent.
+		// honour system presets. This to account for an edge case where you
+		// deleted a MachineConfig that enabled/disabled the unit to revert,
+		// but the unit itself is referenced in other MCs. deleteStaleData() will
+		// catch fully deleted units.
+		// if the unit should be enabled/disabled, then enable/disable it.
+		// this is a no-op if the system wasn't change this iteration
 		if u.Enabled != nil {
 			if *u.Enabled {
 				if err := dn.enableUnit(u); err != nil {
 					return err
 				}
-				glog.V(2).Infof("Enabled systemd unit %q", u.Name)
 			} else {
 				if err := dn.disableUnit(u); err != nil {
 					return err
 				}
-				glog.V(2).Infof("Disabled systemd unit %q", u.Name)
+			}
+		} else {
+			if err := dn.presetUnit(u); err != nil {
+				// Don't fail here, since a unit may have a dropin referencing a nonexisting actual unit
+				glog.Infof("Could not reset unit preset for %s, skipping. (Error msg: %v)", u.Name, err)
 			}
 		}
 	}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1433,31 +1433,34 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config) 
 	return nil
 }
 
-// enableUnit enables a systemd unit via systemctl
-func (dn *Daemon) enableUnit(unit ign3types.Unit) error {
-	_, err := exec.Command("systemctl", "enable", unit.Name).CombinedOutput()
+// enableUnits enables a set of systemd units via systemctl, if any fail all fails.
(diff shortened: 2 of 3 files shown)


yuqi-zhang:
need to fix up a bit more but general idea is up for review

pkg/daemon/daemon.go
```diff
@@ -964,6 +965,20 @@ func upgradeHackFor44AndBelow() error {
 	return nil
 }
 
+// Remove artifacts used by ignition, that the MCO should no longer
+// use since the machine is up.
+// Currently removes the systemd preset file written by Ignition.
+func removeIgnitionArtifacts() error {
+	_, err := os.Stat(constants.IgnitionSystemdPresetFile)
+	if err != nil && os.IsNotExist(err) {
+		return errors.Wrap(err, "failed to open Ignition-written systemd preset file")
```
bgilbert, line 974: To be clear, your intent is to fail if the preset file doesn't exist?

In any event, the `os.Stat()` isn't necessary; you can just check for `os.IsNotExist()` from the `os.Remove()` call.

bgilbert (commented):
LGTM generally.

pkg/daemon/daemon.go
yuqi-zhang: No, I was going to do a rename so we keep a backup and then decided against it. It should not fail regardless. Fixed, thanks!

yuqi-zhang:
Tested with the following:

```
kind: MachineConfig
metadata:
  labels:
  name: 98-test-service
spec:
  config:
    ignition:
      version: 3.1.0
    systemd:
      units:
        - name: test.service
          contents: |
            [Unit]
            Description=Hello Test Service
            After=network-online.target
            [Service]
            Type=simple
            ExecStart=/usr/bin/echo hello
            [Install]
```
and
```
kind: MachineConfig
metadata:
  labels:
  name: 98-test-service-2
spec:
  config:
    ignition:
      version: 3.1.0
    systemd:
      units:
        - name: test.service
          enabled: true
```

Adding the first correctly links to network-online.target,
Adding the second correctly enables and runs the service upon reboot
Removing the second correctly disables the service

pkg/daemon/update.go
```diff
@@ -1259,22 +1259,26 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
 			} else {
 				if err := dn.disableUnit(u); err != nil {
 					return err
 				}
-				glog.V(2).Infof("Disabled systemd unit %q", u.Name)
+			}
+		} else {
+			if err := dn.presetUnit(u); err != nil {
+				// Don't fail here, since a unit may have a dropin referencing a nonexisting actual unit,
+				glog.Infof("Could not reset unit preset for %s, skipping", u.Name)
```
bgilbert, line 1281: It'd be good to log the error.

yuqi-zhang: will add that :+1:

yuqi-zhang:
So there's a small issue with this approach: there appears to be roughly a ~35% increase in operation time due to the extra writes and calls to systemctl. At the very least this times out our GCP-OP tests.

Our options then seem to be:
1. incur the performance penalty and bump test timeouts
2. go back to writing the files by hand, and add a parser ourselves to find the `install` section of the unit, when its enabled, and link based on that instead of hardcoding links

cgwalters:
Today the MCO rewrites files even if they didn't change; if we fixed that it'd be a general improvement and also fix that bug I think.

yuqi-zhang:
> Today the MCO rewrites files even if they didn't change; if we fixed that it'd be a general improvement and also fix that bug I think.

I was under the impression that this was somewhat intentional. Although inefficient, this has some benefits e.g. allowing us to use the `forcefile` option to re-write all the configs, in case of a drift-degraded situation.

We could consider that separately. I also don't see how it would fix this issue, as the fundamental problem is a wrong hardcode + no restoration of system defaults

bgilbert:
Hand-implementing `[Install]` would require handling or intentionally ignoring various [corner cases](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#%5BInstall%5D%20Section%20Options): `Alias`, `Also`, and `DefaultInstance` for templated units.  I wouldn't recommend it.

`systemctl {enable,disable,preset}` support specifying multiple units on the command line.  You could accumulate three lists of units to configure, and then at the end, run `systemctl` exactly once per category of change.

pkg/daemon/update.go
```diff
@@ -1169,23 +1169,25 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config)
 	return nil
 }
 
-// enableUnit enables a systemd unit via systemctl
-func (dn *Daemon) enableUnit(unit ign3types.Unit) error {
-	_, err := exec.Command("systemctl", "enable", unit.Name).CombinedOutput()
+// enableUnits enables a set of systemd units via systemctl, if any fail all fails.
+func (dn *Daemon) enableUnits(units []string) error {
+	unitString := strings.Join(units, " ")
+	_, err := exec.Command("systemctl", "enable", unitString).CombinedOutput()
```
bgilbert, line 1175: Shouldn't this be passing each unit as a separate argument?

pkg/daemon/update.go
```diff
@@ -1169,23 +1169,25 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config)
 	return nil
 }
 
-// enableUnit enables a systemd unit via systemctl
-func (dn *Daemon) enableUnit(unit ign3types.Unit) error {
-	_, err := exec.Command("systemctl", "enable", unit.Name).CombinedOutput()
+// enableUnits enables a set of systemd units via systemctl, if any fail all fails.
+func (dn *Daemon) enableUnits(units []string) error {
```
bgilbert, line 1173: Either here or in the caller, we should skip the `exec.Command` if the list is empty.

bgilbert (approved):
LGTM

yuqi-zhang:
Also fixed the preset command, sorry for the torrent of pushes

pkg/daemon/update.go
```diff
@@ -1165,38 +1169,37 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config)
...
+	return nil
+}
 
-	return os.Remove(wantsPath)
+// presetUnit resets a systemd unit to its preset via systemctl
+func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
+	args := []string{"preset", unit.Name}
+	_, err := exec.Command("systemctl", args...).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "running preset on unit")
```
bgilbert, line 1199: What's the reasoning for not reporting stdout/stderr here?  I notice that you were doing that and then removed it.

yuqi-zhang: added back in, was testing something

yuqi-zhang:
I can confirm that with the updates, this PR still works. However the performance enhancement from doing the enable/disable as one command is somewhat negligible. GCP-op will still timeout.

yuqi-zhang:
Hmm so I tried this myself and looked at the test run logs a bit. It seems that the old file+unit section operations takes around 0.3s, now it takes around 1.5s.

On the tests we reboot around 25 times total, which should have cost us less than a minute extra time overall. So this is kinda odd... retesting a few times to see

kikisdeliveryservice:
This is super weird bc passing test runs have a 15 min cushion before timeout... so somehow runs on this PR are picking up an extra 15+ min which... doesn't make any sense at all..

kikisdeliveryservice:
Ok so looking and comparing simlar logs between this pr and other PRs the added time seems to be:

slower on this pr:
```
I1013 23:48:41.725121    1954 update.go:1581] initiating reboot: Node will reboot into config rendered-node-log-level-55a28ad7-c4b8-42a9-a6e5-be617ae79d97-0d86d89c5687a7dfaab339c73ff552cb
I1013 23:48:44.875129    1954 daemon.go:642] Shutting down MachineConfigDaemon
I1013 23:50:46.706624    2062 start.go:108] Version: machine-config-daemon-4.6.0-202006240615.p0-341-g28d6e6a9-dirty (28d6e6a933bda8f44c1ca8a1640f3df0e0af003e)
```
vs normal:
```

I1014 13:19:24.060603    1750 update.go:1558] initiating reboot: Node will reboot into config rendered-node-log-level-fd357080-dac4-421f-b330-19f3ac1da875-83b0e131821c8dc7ea723c37f62b6c9a
I1014 13:19:24.178783    1750 daemon.go:641] Shutting down MachineConfigDaemon
I1014 13:20:24.422098    2028 start.go:108] Version: machine-config-daemon-4.6.0-202006240615.p0-343-gc94d7de8-dirty (c94d7de8d3b41dea6e91c8596cfe8effd6151b4a)

```

It's an extra minute each time..

yuqi-zhang:
Thanks for digging in! So I pulled up a cluster with this PR on GCP to dig into the journal logs, and it seems that the extra reboot time is... the node itself. As in if I looked at `journactl -b` vs `journalctl -b -1` vs `journalctl -b -2` there is a 2 minute gap between the node going down and coming back up. This also does not seem to be the case for the FIRST reboot post-install, which is much faster (30s), but only for further boots when the tests are running. So... I'm at a bit of a loss why this is happening. Let's see how reproducible this is.

yuqi-zhang:
/hold

so its not presets thats causing the slowdown as dropping it doesn't seem to make a difference. I realized we have console logs on AWS jobs so for example, the latest run: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2145/pull-ci-openshift-machine-config-operator-master-e2e-aws/1316497123818082304/artifacts/e2e-aws/gather-aws-console/i-06474dcdf74706166 a worker node

nothing looks out of the ordinary and seems to be consistent with what I've observed in other jobs. Also tried manually before and at least the logs look fine.

At this point my only remaining suspicion is that something is happening during the shutdown process that is slowing down the reboot, after journal has already died but before the system fully shuts down. Since AWS does not seem to have streaming console access I've requested access to GCP to see if I can find something there.

yuqi-zhang:
Ok so now I'm even more confused. I spun up 3 clusters with this PR on GCP, and ran e2e tests on them. Each time the tests completed in relatively normal time (sometimes even faster than successful runs from other PRs). Why would that be the case when they should be basically the same environment?

The GCP console did yield something of interest, though. In my test clusters I can see that this happens sometimes:

```         Starting Reboot...
[  486.341143] systemd-shutdown[1]: Syncing filesystems and block devices.
[  486.360034] systemd-shutdown[1]: Sending SIGTERM to remaining processes...
[  486.374257] systemd-journald[877]: Received SIGTERM from PID 1 (systemd-shutdow).
[  576.370509] systemd-shutdown[1]: Sending SIGKILL to remaining processes...
[  576.379894] systemd-shutdown[1]: Sending SIGKILL to PID 2300 (openshift-tuned).
[  576.387491] systemd-shutdown[1]: Sending SIGKILL to PID 4679 (tuned).
[  576.449078] device veth1a671224 left promiscuous mode
[  576.454423] kauditd_printk_skb: 7 callbacks suppressed
[  576.454425] audit: type=1700 audit(1603484031.537:315): dev=veth1a671224 prom=0 old_prom=256 auid=4294967295 uid=0 gid=0 ses=4294967295
[  576.497077] systemd-shutdown[1]: Unmounting file systems.
[  576.503758] [24452]: Remounting '/var' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  576.528226] [24453]: Unmounting '/var'.
[  576.543172] [24454]: Remounting '/etc' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  576.553298] [24455]: Unmounting '/etc'.
[  576.563056] [24456]: Remounting '/' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  576.572766] systemd-shutdown[1]: All filesystems unmounted.
[  576.578500] systemd-shutdown[1]: Deactivating swaps.
[  576.583619] systemd-shutdown[1]: All swaps deactivated.
[  576.588967] systemd-shutdown[1]: Detaching loop devices.
[  576.594565] systemd-shutdown[1]: All loop devices detached.
[  576.600257] systemd-shutdown[1]: Detaching DM devices.
[  576.627255] kvm: exiting hardware virtualization
[  576.632206] sd 0:0:1:0: [sda] Synchronizing SCSI cache
[  576.782626] reboot: Restarting system
[  576.786422] reboot: machine restart
```
whereas a faster reboot:
```
         Starting Reboot...
[  466.111093] systemd-shutdown[1]: Syncing filesystems and block devices.
[  466.133854] systemd-shutdown[1]: Sending SIGTERM to remaining processes...
[  466.148008] systemd-journald[880]: Received SIGTERM from PID 1 (systemd-shutdow).
[  466.678982] device vethd662d8f2 left promiscuous mode
[  466.729116] systemd-shutdown[1]: Sending SIGKILL to remaining processes...
[  466.741486] systemd-shutdown[1]: Unmounting file systems.
[  466.747979] [26078]: Remounting '/var' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  466.780715] [26079]: Unmounting '/var'.
[  466.798323] [26080]: Remounting '/etc' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  466.808452] [26081]: Unmounting '/etc'.
[  466.819307] [26082]: Remounting '/' read-only in with options 'seclabel,attr2,inode64,prjquota'.
[  466.829049] systemd-shutdown[1]: All filesystems unmounted.
[  466.834792] systemd-shutdown[1]: Deactivating swaps.
[  466.839920] systemd-shutdown[1]: All swaps deactivated.
[  466.845263] systemd-shutdown[1]: Detaching loop devices.
[  466.850846] systemd-shutdown[1]: All loop devices detached.
[  466.856584] systemd-shutdown[1]: Detaching DM devices.
[  466.887225] kvm: exiting hardware virtualization
[  466.892162] sd 0:0:1:0: [sda] Synchronizing SCSI cache
[  467.038417] reboot: Restarting system
[  467.042226] reboot: machine restart
```

so it seems that tuned sometimes is slowing the whole shutdown process. Maybe that's something to look into as well.

pkg/daemon/update.go
```diff
@@ -1252,23 +1258,40 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
+		// as an error, but it would cause other presets that would have succeeded
+		// to not go through.
+
 		if u.Enabled != nil {
 			if *u.Enabled {
-				if err := dn.enableUnit(u); err != nil {
-					return err
-				}
-				glog.V(2).Infof("Enabled systemd unit %q", u.Name)
+				enabledUnits = append(enabledUnits, u.Name)
```
cgwalters, line 1275: I think it'd probably be better to do a "diff" here - only enable/disable units that changed between the two configs.

Otherwise I worry that this may cause unexpected side effects.

yuqi-zhang: So this doesn't change any existing behaviour. Today we re-write all enable/disables anyways (via the hardlink to multi-user.target.wants) just like how we re-write any file/unit again. I'm not sure if we want to make the exception

cgwalters: Right, but I suspect part of the problem here is that going via `systemctl` may touch the DBus API doing a bunch of synchronous requests; that's going to slow things down.

Another angle is to try `SYSTEMD_OFFLINE=1` [in the environment](https://github.com/systemd/systemd/blob/master/docs/ENVIRONMENT.md)?  

Of course if we implemented https://github.com/openshift/machine-config-operator/issues/1190 this issue would also go away because we'd only be enabling in the new root, so shouldn't be talking to the running systemd either.

yuqi-zhang:
Rebased, will see whether slowdowns are still seen in gcp-op tests

yuqi-zhang:
Comparing runtimes of most recent gcp-op tests from other PRs, I don't see a significant difference in runtime for the tests that passed. Will retest a few more times after gcp-op issues are fixed.

yuqi-zhang:
Tested again with the steps in https://github.com/openshift/machine-config-operator/pull/2145#issuecomment-705796418. Once we're confident in GCP-OP not being slowed down by this, I'm ready to have this merge

darkmuggle:
/approve
/lgtm

Reading over the PR and the code, this LGTM.

yuqi-zhang:
One more test to be safe

/test ci/prow/e2e-gcp-op

yuqi-zhang:
Times look fine
/hold cancel

PR #2148, opened 2020-10-08
node: always set current/desired/state

yuqi-zhang:
In rare cases where we lose node annotations, set all annotations
as if we were reading from the initial file. Otherwise the MCD
will likely fail right after when it attempts to access the other
variables.

See https://bugzilla.redhat.com/show_bug.cgi?id=1885921 as an
example of us only setting current config.

I would also not be against failing outright here, as we are likely
in a bad state for the node anyways. In that case we should
explicitly fail here and not waiting for a later error for desired config.

### DIFF
--- a/pkg/daemon/node.go
+++ b/pkg/daemon/node.go
@@ -32,7 +32,13 @@ func (dn *Daemon) loadNodeAnnotations(node *corev1.Node) (*corev1.Node, error) {
 		currentOnDisk, err := dn.getCurrentConfigOnDisk()
 		if err == nil {
 			glog.Infof("Setting initial node config based on current configuration on disk: %s", currentOnDisk.GetName())
-			return setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, node.Name, map[string]string{constants.CurrentMachineConfigAnnotationKey: currentOnDisk.GetName()})
+			// this state shouldn't really happen, but since we're here, we're just going to set desired == current and state == done,
+			// otherwise the MCD will complain it cannot find those annotations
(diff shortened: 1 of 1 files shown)


pkg/daemon/node.go
```diff
@@ -32,7 +32,13 @@ func (dn *Daemon) loadNodeAnnotations(node *corev1.Node) (*corev1.Node, error) {
 		currentOnDisk, err := dn.getCurrentConfigOnDisk()
 		if err == nil {
 			glog.Infof("Setting initial node config based on current configuration on disk: %s", currentOnDisk.GetName())
-			return setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, node.Name, map[string]string{constants.CurrentMachineConfigAnnotationKey: currentOnDisk.GetName()})
+			// this state shouldn't really happen, but since we're here, we're just going to set desired == current and state == done,
+			// otherwise the MCD will complain it cannot find those annotations
+			return setNodeAnnotations(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, node.Name, map[string]string{
+				constants.CurrentMachineConfigAnnotationKey:     currentOnDisk.GetName(),
+				constants.DesiredMachineConfigAnnotationKey:     currentOnDisk.GetName(),
```
cgwalters, line 39: Currently the node controller is the only thing setting the desired config.

I think this will introduce a race condition if a node completes an update and at the same time the node controller wants it to update to a different configuration.

The race might be benign - I think the node controller would probably just set it back.

But probably what we want is to only set it if it's not currently set.

cgwalters (changes requested):

PR #2149, opened 2020-10-12
Fix example to include directives required for alerting

matt-simons:
**- What I did**
Updated the example to include the directives documented here: https://docs.openshift.com/container-platform/4.5/installing/install_config/installing-customizing.html#installation-special-config-crony_installing-customizing

Without rtcsync in particular an alert called NodeClockNotSynchronising will trigger in Openshift, as discussed in Case 02763639 

**- How to verify it**

Apply the example as is (with a valid NTP server) and the alert will fire.

Fix example to include directives required for alerting

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -87,7 +87,7 @@ spec:
     storage:
       files:
       - contents:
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Thanks!
/approve
/lgtm
/ok-to-test
/skip

PR #2152, opened 2020-10-13
Bug 1887545: Fix ovs-configuration detecting bond and vlan interfaces

trozet:
When we add the bond port OVS connection it needs to use type bond
instead of ethernet. Similarly VLAN needs type vlan as well as some
extra arguments.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -94,12 +94,33 @@ contents:
         nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
       fi
 
+      extra_phys_args=""
+      # check if this interface is a vlan, bond, or ethernet type
+      if [ $(nmcli --get-values connection.type conn show ${old_conn}) == "vlan" ]; then
+        iface_type=vlan
+        vlan_id=$(nmcli --get-values vlan.id conn show ${old_conn})
+        if [ -z "$vlan_id"]; then
+          echo "ERROR: unable to determine vlan_id for vlan connection: ${old_conn}"
+          exit 1
+        fi
+        vlan_parent=$(nmcli --get-values vlan.parent conn show ${old_conn})
+        if [ -z "$vlan_parent" ]; then
+          echo "ERROR: unable to determine vlan_parent for vlan connection: ${old_conn}"
+          exit 1
+        fi
+        extra_phys_args="dev ${vlan_parent} id ${vlan_id}"
+      elif [ $(nmcli --get-values connection.type conn show ${old_conn}) == "bond" ]; then
+        iface_type=bond
+      else
+        iface_type=802-3-ethernet
+      fi
+
       # bring down any old iface
       nmcli device disconnect $iface
 
       if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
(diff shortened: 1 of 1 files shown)


trozet:
@dcbw PTAL

trozet:
/hold

while @mcornea verifies this

trozet:
ovn-step-registry failure is caused by metrics issues and has nothing to do with this PR

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -94,12 +94,33 @@ contents:
         nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
       fi
 
+      extra_phys_args=""
+      # check if this interface is a vlan, bond, or ethernet type
+      if nmcli --fields connection.type conn show ${old_conn} | grep -q vlan; then
```
danwinship, line 99: if [[ $(nmcli --get-values connection.type conn show ${old_conn}) == "vlan" ]]; then

? Especially, then it won't be fooled by `connection.type: macvlan`

(Likewise with `--get-values` below)

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -94,12 +94,33 @@ contents:
...
+          echo "ERROR: unable to determine vlan_id for vlan connection: ${old_conn}"
+          exit 1
+        fi
+        vlan_parent=$(nmcli --fields vlan.parent conn show ${old_conn} | awk '{print $2}')
+        if [ -z "$vlan_parent" ]; then
+          echo "ERROR: unable to determine vlan_parent for vlan connection: ${old_conn}"
+          exit 1
+        fi
+        extra_phys_args="dev ${vlan_parent} id ${vlan_id}"
+      elif nmcli --fields connection.type conn show ${old_conn} | grep -q bond || [ -d /sys/class/net/${iface}/bonding/ ]; then
```
danwinship, line 112: why are both checks needed rather than just one or the other?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -94,12 +94,33 @@ contents:
...
+        iface_type=bond
+      else
+        iface_type=802-3-ethernet
```
danwinship, line 115: should we explicitly validate that and properly error out if it's any other type?

templates/common/_base/files/configure-ovs-network.yaml
trozet: ah didn't know about --get-values, will change to use that and direct compare. Thanks.

templates/common/_base/files/configure-ovs-network.yaml
trozet: I originally tried just the second check, but then found through testing that if the interface goes down that goes away. I included both checks just to be safe. I guess the second check really isn't needed, the script already assumes that the nm connection must exist for a default gateway.

templates/common/_base/files/configure-ovs-network.yaml
trozet: I'm kind of hesitant to do that. If we don't explicitly know the type I think its ok to attempt with ethernet. If it doesn't work we script will end up failing and fall back to the original connection.

trozet:
@runcom @sinnykumari PTAL

PR #2154, opened 2020-10-13
Bug 1887040: OVS config: check if OVS is installed

trozet:
With UPI installs, openvswitch package is not installed until after
upgrade. In the ovs-configuration.service there is a Requires
declaration on openvswitch.service. However due to a bug in systemd,
ovs-configuration is started even when openvswitch.service is not
present:

https://bugzilla.redhat.com/show_bug.cgi?id=1888017

As a workaround, this patch checks to ensure OVS is installed in
configure-ovs.sh.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -4,6 +4,13 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
+    # Workaround to ensure OVS is installed due to bug in systemd Requires:
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1888017
+    if ! rpm -qa | grep -q openvswitch; then
+      echo "Warning: Openvswitch package is not installed!"
+      exit 1
+    fi
+
     if [ "$1" == "OVNKubernetes" ]; then
       # Configures NICs onto OVS bridge "br-ex"
       # Configuration is either auto-detected or provided through a config file written already in Network Manager


zhaozhanqi:
@trozet I saw 4.6 branch create the file at the begginging `touch /var/run/ovs-config-executed`
https://github.com/openshift/machine-config-operator/commit/e16d4cdcdfca0af0ee495072602012e1c1ff2367#diff-afb45a3711a77d94f26471d9d94a7f7a03d931d9e72bdf849f2e26e2711d6fd7
this no need for 4.7 version?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -4,6 +4,13 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
+    # Workaround to ensure OVS is installed due to bug in systemd Requires:
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1888017
+    if ! rpm -qa | grep -q openvswitch; then
```
dcbw, line 9: @trozet let's do `openvswitch2.13`

squeed, line 9: how about `if ! command-v ovs-vsctl; then` ?

dcbw: missing space, right? `if ! command -v ovs-vsctl; then`

trozet: why is checking for a binary better than checking if the RPM is installed?

cybertron: I suppose there's an argument that if the package got split and the binary moved to a different package it might be more robust to check the binary itself. That said, if I understand this correctly it's a workaround that only needs to exist for one release, right? After that openvswitch will have been installed and we don't need the check for future upgrades. We shouldn't need to worry about binaries moving around in that case.

I guess the one thorny issue would be if skip-level upgrades need to be supported. I'm pretty sure our IPI templates wouldn't handle that right now, but I don't know what UPI requires.

trozet: if the binary gets split then its even more of an argument not to check for the binary. We need ovs-vswitchd up and running here not just ovs-vsctl binary. I'm going to stick with the ensuring the package is installed since that is actually what we care about for UPI. If openvswitch for some split into multiple packages, then we will need to update this check to look for those packages as well.

cybertron: Oh, yeah if the binary is not the thing you care about then that wouldn't make sense. I suppose you could check for ovs-vswitchd instead, but like I said it's a temporary workaround so I'm not inclined to get too hung up on the details.

That said, although I got assigned to this I don't have approval on this repo so this is just my 2 cents anyway. :-)

vrutkovs:
OKD 4.6 is known to be broken right now (https://github.com/openshift/release/pull/12683 to fix it), so feel free to skip optional tests

templates/common/_base/files/configure-ovs-network.yaml
sinnykumari: maybe just use `rpm -qa openvswitch`  ? using grep here can also give a match for packages which has openvswitch as sub-string

dcbw: If we go for rpm -q, it would have to be `rpm -q openvswitch2.13` FWIW

trozet: "using grep here can also give a match for packages which has openvswitch as sub-string"

Yeah that was intentional. What if we move to openvswitch2.14 and the package changes names? I don't think that's better. Really openvswitch package should not have the version in its package name, but to workaround this we only look for openvswitch here.

sinnykumari: sure, if this is preferred way by the OVN team :)

sinnykumari (approved):

PR #2156, opened 2020-10-13
Bug 1886636: update ctrcfg crd to make spec & containerruntimeconfig required

kikisdeliveryservice:
All ctrcfgs must have a spec and containerruntimeconfig all kubelet configs and all kubeletconfigs should have a spec (but as the new loglevel  was added as same level, kubeletruntimeconfig not required)

### DIFF
--- a/install/0000_80_machine-config-operator_01_containerruntimeconfig.crd.yaml
+++ b/install/0000_80_machine-config-operator_01_containerruntimeconfig.crd.yaml
@@ -30,6 +30,8 @@ spec:
       description: ContainerRuntimeConfig describes a customized Container Runtime
         configuration.
       type: object
+      required:
+      - spec
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
testing this:
with a bad ctrcfg like:
```
kind: ContainerRuntimeConfig
metadata:
 name: set-pids-limit
spec:
 machineConfigPoolSelector:
   matchLabels:
     pools.operator.machineconfiguration.openshift.io/worker: ""
```
Yields:
```
$ oc create -f high-pids.yaml
The ContainerRuntimeConfig "set-pids-limit" is invalid: spec.containerRuntimeConfig: Required value
```

Also some other bad ctrcfgs and kubeletcfgs yield:
```
$ oc create -f maxpods.yaml
The KubeletConfig "set-max-pods" is invalid: spec: Required value
$ oc create -f high-pids.yaml
The ContainerRuntimeConfig "set-pids-limit" is invalid: spec: Required value
```

yuqi-zhang (approved):
Functionality wise LGTM. cc @umohnani8 to see if the changes match expected behaviour

umohnani8:
So when you make `containerRuntimeConfig` required, does it not error anymore? What is the difference in the error reported with a bad config with the changes from this PR?

umohnani8:
Nvm, I misread, ignore my comment above. LGTM
/lgtm
/retest

PR #2157, opened 2020-10-14
Bug 1872001: Stop using EtcdDiscoveryDomain in templates

retroflexer:
Etcd is no longer using EtcdDiscoveryDomain and as a result the parameter should be removed. However, the baremetal and friends templates reference it to get the cluster domain. This information is also available from the DNS structure, which is not currently wired into the templates.

This is a rebased replica of #2025. 

**- What I did**
This change adds the DNS struct to the resources in the controller config and switches the templates to get the cluster domain from there.
**- How to verify it**
Install a new cluster without etcdDiscoveryDomain (https://github.com/openshift/installer/pull/4067). It should successfully install the cluster.

### DIFF
--- a/cmd/machine-config-operator/bootstrap.go
+++ b/cmd/machine-config-operator/bootstrap.go
@@ -41,6 +41,7 @@ var (
 		rootCAFile                string
 		proxyConfigFile           string
 		additionalTrustBundleFile string
+		dnsConfigFile             string
 	}
 )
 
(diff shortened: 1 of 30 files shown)


kikisdeliveryservice:
Baremetal & friends need to review/test this before merging.

/hold

retroflexer:
Since the original PR is reactivated and rebased, I am closing this PR in favor of #2025 .

/close

PR #2158, opened 2020-10-14
Update hostname when transient hostname is fedora

bdurrow:
A recent change, perhaps the adoption of fcos 33 broke hostname assignment (at least in GCP) because the transient hostname was set to fedora on boot.  We need to use the dhcp provided hostname in this case.

**- What I did**
Modified templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml so that the string fedora is treated like localhost when we check to see if the hostname was previously set.

**- How to verify it**
Build a 4.6 cluster in GCP.

Use dhcp provided hostname when transient hostname is fedora

### DIFF
--- a/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
+++ b/templates/common/_base/files/etc-networkmanager-dispatcher.d-90-long-hostname.yaml
@@ -15,7 +15,7 @@ contents:
         exit 0
     fi
 
-    if [[ ! "$(< /proc/sys/kernel/hostname)" =~ (localhost|localhost.localdomain) ]]; then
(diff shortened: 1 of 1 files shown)


vrutkovs:
Sorry for confusion, but `fcos` branch is for OKD 4.5. Please re-submit this for master (or `release-4.6` branch) to this out on OKD 4.6

vrutkovs:
Verified that this fix works on OKD 4.6 in GCP. However its odd that default hostname is `fedora`, it would be easier to keep it `localhost` on FCOS side

bdurrow:
Rebased in https://github.com/openshift/machine-config-operator/pull/2160

Issue #2059, opened 2020-09-03
Allow administrators to guide upgrade ordering


cgwalters:
Today the MCO makes no attempt to apply any ordering to which nodes it updates from the candidates.  One problem we're thinking about is (particularly on bare metal scenarios where there might be a lot of pods on a node, and possibly pods expensive to reschedule like CNV) that it's quite possible that workloads are disrupted multiple times for an OS upgrade.

When we go to drain a node, its pods will be rescheduled across the remaining nodes...and then we will upgrade one of those, quite possibly moving one of the workload pods *again* etc.

One idea here is to add the minimal hooks such that a separate controller could influence this today.

If for example we supported a label `machineconfig.openshift.io/upgrade-weight=42` and the node controller picked the highest weight node, then the separate controller could also e.g. mark $number nodes which are next in the upgrade ordering as unschedulable, ensuring that the drain from the current node doesn't land on them.

Without excess capacity or changing the scheduler to more strongly prefer packing nodes it seems hard to avoid multiple disruption, but the label would allow this baseline integration.

cgwalters:
Also strongly related to this is https://github.com/openshift/machine-config-operator/pull/2035 - if anyone wants to e.g. avoid multiple upgrade disruption by scaling up new nodes - today those nodes start at the *old* config and so will then shortly get upgraded and rebooted.  I think other kube distros tend to do upgrades by just creating a new node at the *new* config and deleting the old node, so they're less subject to this (but that model can't easily apply on bare metal).  If we had that PR we could at least have a phase like:

- upgrade one node
- some sort of linkup between MCO and machineAPI says "OK I want some upgrade burst capacity" and scales up the worker machineset, which now has *new* nodes at the new config
- old nodes could be targeted for deletion instead of being upgraded

In combination with the above we could allow admins to choose between in-place updates and the burst upgrade strategy at least in cloud.

smarterclayton:
I will spawn this as a separate issue but openshift should bias the scheduler to prefer nodes with capacity that are “up to date” over nodes that might have more capacity but are “not up to date”, because a not up to date node will eventually need to be drained.  That implies there should be a label or other marker on nodes that by convention indicates a node that is “not up to date and likely to be drained in the near future”, and the scheduler should avoid that without completely ruining other scheduling criteria.  This is a soft criteria because it is better to disrupt a workload twice than let the workload dip below the preferred availability of the cluster.

cgwalters:
[...]

PR #2162, opened 2020-10-15
node-controller: Support an annotation to hold/prioritize updates

cgwalters:
Today the MCO arbitrarily chooses a node to update from the candidates.
We want to allow admins to avoid specific nodes entirely (for as long
as they want) as well as guide upgrade ordering.

This replaces the defunct etcd-specific code with support for a generic
annotation `machineconfiguration.openshift.io/update-order` that allows
an external controller (and/or human) to do both of these.

Setting it to `0` will entirely skip that node for updates.  Otherwise,
higher values are preferred.

Closes: https://github.com/openshift/machine-config-operator/issues/2059

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sort"
+	"strconv"
 	"time"
 
 	"github.com/golang/glog"
@@ -866,7 +868,7 @@ func getAllCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*core
 	}
 	capacity -= failingThisConfig
 
-	return nodes, uint(capacity)
+	return prioritizeCandidateNodes(nodes), uint(capacity)
 }
 
 // getCandidateMachines returns the maximum subset of nodes which can be updated to the target config given availability constraints.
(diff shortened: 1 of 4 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -878,46 +880,45 @@ func getCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.
...
-			glog.Infof("Deferring update of etcd leader: %s", node.Name)
-			continue
+		orderVal, ok := node.Annotations[daemonconsts.MachineUpdateOrderingAnnotationKey]
+		if ok {
+			order, err := strconv.Atoi(orderVal)
+			if err != nil {
+				glog.Warningf("Failed to parse %s %s: %v", node.Name, daemonconsts.MachineUpdateOrderingAnnotationKey, err)
+				continue
+			}
+			// order 0 means "skip this node"
```
cgwalters, line 904: This is the simple PoC, but if we go this route I think we need to make this more observable; something like a count of "held/skipped" nodes in the pool as well, etc.

crawford:
/hold

As mentioned in https://github.com/openshift/machine-config-operator/issues/2059#issuecomment-692166046, I'm against this direction. Its potential for misuse is too high and it's not clear to me what problem this solves. As for the implementation, assigning special function to `0` is no longer considered good API. I'd prefer an algebraic type or a separate field entirely.

If the intention of this PR is to provide a mechanism for pausing updates for a particular node, then let's specifically tackle that. I'm in favor of defining an annotation whose presence is the signal to MCO that this node should be skipped.

cgwalters:
> As mentioned in #2059 (comment), I'm against this direction. Its potential for misuse is too high and it's not clear to me what problem this solves. 

I replied here on that concern: https://github.com/openshift/machine-config-operator/issues/2059#issuecomment-692178438

I completely agree that OpenShift should *by default* be more intelligent about how we upgrade nodes, but I can't imagine we hardcode all of that logic into the node controller.  An update ordering system seems like it really needs to be a separate controller with a higher level view (including of machinesets, etc.).  And on UPI metal admins are just going to want full control.  So I don't see how we can avoid a low-level API like this at least eventually.

> If the intention of this PR is to provide a mechanism for pausing updates for a particular node, then let's specifically tackle that.

That's fair, yeah we can make that separate.

eparis:
I asked colin to look at the problem of designing an API to allow explicit pause that wouldn't constrain us too much from doing more in the future. i don't have a strong opinion of how much further to go than just a 'paused/unpaused' bit.

cgwalters:
OK holding only is https://github.com/openshift/machine-config-operator/pull/2163

kikisdeliveryservice:
To be clear this PR is now the prioritize updates PR and #2163 is the hold updates pr?

cgwalters:
> To be clear this PR is now the prioritize updates PR and #2163 is the hold updates pr?

Yeah they're related but conceptually orthogonal.  Since it seems we want #2163 more we can rebase this on that when it merges, or close this if we decide to take another direction. 

(I guess in fact a controller could implement update priority by simply adding a hold to everything it didn't want to update, would be crude but...)

PR #2164, opened 2020-10-17
Bug 1887607: OVS config: check if OVS is installed

vrutkovs:
Cherry-pick of https://github.com/openshift/machine-config-operator/pull/2154 on 4.6

With UPI installs, openvswitch package is not installed until after
upgrade. In the ovs-configuration.service there is a Requires
declaration on openvswitch.service. However due to a bug in systemd,
ovs-configuration is started even when openvswitch.service is not
present:

https://bugzilla.redhat.com/show_bug.cgi?id=1888017

As a workaround, this patch checks to ensure OVS is installed in
configure-ovs.sh.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -4,6 +4,14 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
+
+    # Workaround to ensure OVS is installed due to bug in systemd Requires:
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1888017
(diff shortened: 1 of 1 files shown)


sdodson:
Patch manager, I've overridden the bugzilla/valid-bug flag because this is really only testable in 4.5 to 4.6 and I have personally tested that upgrade path with a release image built from this PR. This affects upgrade success in clusters with RHEL workers which is a small set but non zero.

vrutkovs:
OKD is currently broken on GCP (fix: https://github.com/openshift/machine-config-operator/pull/2160)

/skip

PR #2167, opened 2020-10-21
Bug 1890250: mcs: Ensure that the encapsulated config is spec 2 if requested

cgwalters:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1890250

Basically adding new workers during the window when:

- control plane has upgraded to 4.6 (actually more precisely, the machine-config-server)
- worker pool is still rolling out

Will fail for clusters whose bootimage is older than 4.6 because
the MCD-on-host will find the `3.1.0` version in
`/etc/ignition-machine-config-encapsulated.json`
and barf.

PR #2168, opened 2020-10-21
Bug 1890362: mcs: Ensure that the encapsulated config is spec 2 if requested

cgwalters:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1890250

Basically adding new workers during the window when:

- control plane has upgraded to 4.6
- worker pool is still rolling out

Will fail for clusters whose bootimage is older than 4.6 because
the MCD-on-host will find the `3.1.0` version in
`/etc/ignition-machine-config-encapsulated.json`
and barf.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -153,6 +153,7 @@ github.com/coreos/ignition v0.35.0/go.mod h1:WJQapxzEn9DE0ryxsGvm8QnBajm/XsS/Pkr
 github.com/coreos/ignition/v2 v2.1.1/go.mod h1:RqmqU64zxarUJa3l4cHtbhcSwfQLpUhv0WVziZwoXvE=
 github.com/coreos/ignition/v2 v2.3.0 h1:TK+STbzVe6KZp4tQ2IaNSRMiWX4/diNngep1F7tP7Zk=
 github.com/coreos/ignition/v2 v2.3.0/go.mod h1:85dmM/CERMZXNrJsXqtNLIxR/dn8G9qlL1CmEjCugp0=
+github.com/coreos/ignition/v2 v2.7.0 h1:JCKxJllVtnk1lQY1uisxrtFSHG5L2NI1LRzc8wBEk84=
 github.com/coreos/pkg v0.0.0-20160727233714-3ac0863d7acf/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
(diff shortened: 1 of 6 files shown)


cgwalters:
4.6 version of https://github.com/openshift/machine-config-operator/pull/2167 - I'm using this so I can create a release image from it and get early e2e testing.

cgwalters:
Release image for this PR: `registry.svc.ci.openshift.org/ci-ln-sgx9mwt/release:latest`

go.sum
```diff
@@ -153,6 +153,7 @@ github.com/coreos/ignition v0.35.0/go.mod h1:WJQapxzEn9DE0ryxsGvm8QnBajm/XsS/Pkr
 github.com/coreos/ignition/v2 v2.1.1/go.mod h1:RqmqU64zxarUJa3l4cHtbhcSwfQLpUhv0WVziZwoXvE=
 github.com/coreos/ignition/v2 v2.3.0 h1:TK+STbzVe6KZp4tQ2IaNSRMiWX4/diNngep1F7tP7Zk=
 github.com/coreos/ignition/v2 v2.3.0/go.mod h1:85dmM/CERMZXNrJsXqtNLIxR/dn8G9qlL1CmEjCugp0=
+github.com/coreos/ignition/v2 v2.7.0 h1:JCKxJllVtnk1lQY1uisxrtFSHG5L2NI1LRzc8wBEk84=
```
yuqi-zhang, line 156: do we need to update this? let's not if we don't have to

yuqi-zhang (commented):
I can confirm this does fix the problem, the encapsulated json is spec2, and scaleups during and after the upgrade works fine.
/approve
One minor question:

go.sum
cgwalters: Probably not but do you really want to respin for that?

eparis:
I'm manually adding valid-bug. QE will look at both this 4.6 and the 4.7 bug in parallel. If there turns out to be an issue we will have time to fix it.

eparis:
I'm going to step on Jerry's toes because this is exactly the same bits (modulo _test.go differences) as he signed off in 4.7. If this turns out to be a bad build I will take on the personal responsibility to get it reverted and Jerry can make fun of me at Tracy's next all hands.

/lgtm

yuqi-zhang:
Well Eric beat me to it but I'll sign off on it too so we can both get made fun of if needed

PR #2170, opened 2020-10-22
Bug 1890074: daemon: allow an extension to install group of packages

sinnykumari:
In the previous implementation, an extension was one to one
mapping with the package it would install on the host. This can
become confusing if enabling an extension would require
installing multiple sets of packages which are not install
time dependency.

Also added e2e test for kernel-devel extensions

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -865,22 +865,43 @@ func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string
 		}
 	}
 
+	// Supported extensions has package list info that is required
+	// to enable an extension
+	extensions := getSupportedExtensions()
+
 	extArgs := []string{"update"}
 	for _, ext := range added {
-		extArgs = append(extArgs, "--install", ext)
+		for _, pkg := range extensions[ext] {
+			extArgs = append(extArgs, "--install", pkg)
+		}
 	}
 	for _, ext := range removed {
-		extArgs = append(extArgs, "--uninstall", ext)
+		for _, pkg := range extensions[ext] {
+			extArgs = append(extArgs, "--uninstall", pkg)
+		}
 	}
 
 	return extArgs
 }
 
+// Returns list of extensions possible to install on a CoreOS based system.
+func getSupportedExtensions() map[string][]string {
+	// In future when list of extensions grow, it will make
+	// more sense to populate it in a dynamic way.
+
+	// These are RHCOS supported extensions.
+	// Each extension keeps a list of packages required to get enabled on host.
+	return map[string][]string{
+		"usbguard":     {"usbguard"},
+		"kernel-devel": {"kernel-devel", "kernel-headers"},
+	}
+}
+
 func validateExtensions(exts []string) error {
-	supportedExtensions := []string{"usbguard", "kernel-devel"}
+	supportedExtensions := getSupportedExtensions()
 	invalidExts := []string{}
 	for _, ext := range exts {
-		if !ctrlcommon.InSlice(ext, supportedExtensions) {
+		if _, ok := supportedExtensions[ext]; !ok {
 			invalidExts = append(invalidExts, ext)
 		}
 	}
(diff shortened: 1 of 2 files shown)


test/e2e/mcd_test.go
```diff
@@ -269,10 +269,14 @@ func TestExtensions(t *testing.T) {
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], renderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
 
-	installedExtesnions := execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "usbguard")
-	if !strings.Contains(installedExtesnions, "usbguard") {
-		t.Fatalf("Node %s doesn't have expected extensions", infraNode.Name)
+	installedExtesnions := execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel", "kernel-headers")
```
cgwalters, line 272: Not new but tyop here: `installedExtensions`

cgwalters (commented):
/approve
LGTM as is

test/e2e/mcd_test.go
sinnykumari: fixed it. variable has been also renamed to installedPackages which I missed to update here. Somehow, my VS code didn't catch the error(which I need to fix).

runcom:
awesome Sinny! does this just work as rhcos base does have kernel-headers then? if it does I think this is good to get in (I think we spoke about a test for this somewhere but I'm probably just missing the PR rn)

sinnykumari:
> awesome Sinny! does this just work as rhcos base does have kernel-headers then? if it does I think this is good to get in (I think we spoke about a test for this somewhere but I'm probably just missing the PR rn)

machine-os-content already ships kernel-headers package along with (kernel-{devel,core,modules}) in extensions/ directory. So, we can just install it by MCO by explicitly specifying that as args to `rpm-ostree install`.

kernel-devel extensions test is covered in this PR https://github.com/openshift/machine-config-operator/pull/2170/files#diff-6324e53eb8b87fa591d0dd87722e4dff1200a2bc57d7c83f042a90f4be53f68bR255 :smile:

pkg/daemon/update.go
```diff
@@ -865,22 +865,43 @@ func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string
...
+		for _, pkg := range extensions[ext] {
+			extArgs = append(extArgs, "--uninstall", pkg)
+		}
 	}
 
 	return extArgs
 }
 
+// Returns list of extensions possible to install on a CoreOS based system.
+func getSupportedExtensions() map[string][]string {
```
kikisdeliveryservice, line 888: love that you add helpful notes :heart:

pkg/daemon/update.go
```diff
@@ -865,22 +865,43 @@ func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string
 		}
 	}
 
+	// Supported extensions has package list info that is required
+	// to enable an extension
+	extensions := getSupportedExtensions()
```
vrutkovs, line 870: This breaks OKD 4.7 :(

cgwalters: Can you link me to that code?   Wait...I think I found it:
https://github.com/openshift/release/blob/959c2954344438c4eed3ec7f52a5e099e8335516/ci-operator/jobs/openshift/release/openshift-release-release-4.7-periodics.yaml#L586

That's...terrifying :smile: 

I guess it's hard to do much here other than continue to entirely remove validation on OKD short term.

vrutkovs: Right, previously MCC verified that extensions list is valid (on FCOS it skipped this check). Guess its time to rework it so that MCD could mutate it (and leave as is on OKD)

sinnykumari: damm, missed this usecase. It we will good to have okd-gcp-op test running so that we don't break it.
Will give some thought to fix it.

PR #2174, opened 2020-10-23
Bug 1887596: Fix ovs-configuration detecting bond and vlan interfaces

trozet:
When we add the bond port OVS connection it needs to use type bond
instead of ethernet. Similarly VLAN needs type vlan as well as some
extra arguments.

(cherry picked from commit 64b79dfb1247e480478633f3f383653a6f787d3b)

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -98,12 +98,33 @@ contents:
         nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
       fi
 
+      extra_phys_args=""
+      # check if this interface is a vlan, bond, or ethernet type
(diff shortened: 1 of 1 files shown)


trozet:
@danwinship PTAL

cgwalters (approved):
At some point soon I may get motivated to demo how this could use the MCD code in Go...

/approve

sinnykumari:
both OKD tests are known to be failed due to https://github.com/coreos/fedora-coreos-tracker/issues/649
/skip

danwinship:
/hold
all the required tests are passing but the bot is going to keep retrying the optional broken test until this gets `cherrypick-approved`

kikisdeliveryservice:
> /hold
> all the required tests are passing but the bot is going to keep retrying the optional broken test until this gets `cherrypick-approved`

should be good now

trozet:
@kikisdeliveryservice can we override e2e-gcp-op? These changes only affect OVN and that is an openshift-sdn deployment.

PR #2175, opened 2020-10-23
[Baremetal] Update keepalived liveness probe to check also CPU usage

yboaron:
We noticed sometimes that Keepalived container consumes ~100% of CPU resources,
the root cause for that is [1].

This commit updates Keepalived liveness probe to monitor also CPU usage to work around
the high CPU usage bug until U/S fix will be inherited to D/S repo.

[1]  https://bugzilla.redhat.com/show_bug.cgi?id=1868077

### DIFF
--- a/templates/common/baremetal/files/baremetal-keepalived.yaml
+++ b/templates/common/baremetal/files/baremetal-keepalived.yaml
@@ -114,7 +114,7 @@ contents:
             - /bin/bash
             - -c
             - |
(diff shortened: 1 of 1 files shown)


yboaron:
/hold

Seems that Keepalived U/S fix will be inherited in RHEL8.2 [1], so no need for the work around

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1890626

yboaron:
U/S Keepalived fix was inherited to RHEL8.2 [1] , so this workaround isn't needed.
/close 


[1] https://bugzilla.redhat.com/show_bug.cgi?id=1890626

PR #2176, opened 2020-10-23
templates: no longer manipulate the hostnames length

darkmuggle:
Using templates to manipulate the hostname on GCP has proven to be an
extraordinarily annoying bug.

This change requires that Afterburn on GCP set the hostname.


darkmuggle:
Withdrawing. This code is NOT upgrade-safe.

PR #2177, opened 2020-10-23
crio: drop infra container when possible

haircommander:
**- What I did**
in cri-o 1.19 we introduced a feature to drop the infra container when a pod doesn't have a pod level PID namespace
We have been running with this option in cri-o internal tests and have seen no issues.

Let's enable it by default in 4.7. In addition to being more performant, it also should aid in a race with the kubelet
win-win!

**- How to verify it**

Have CRI-O drop the infra container in some cases, which performs better

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -22,6 +22,7 @@ contents:
         "/run/containers/oci/hooks.d",
     ]
     manage_ns_lifecycle = true
+    drop_infra_ctr = true
 
     [crio.image]
     global_auth_file = "/var/lib/kubelet/config.json"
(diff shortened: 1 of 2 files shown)


mrunalp:
We should do a few runs before merging this.

cgwalters (approved):
/lgtm

Is there an e2e covering this?

mrunalp:
@cgwalters The regular e2e test runs should catch regressions, if any.

rphillips:
We want to let this back for a few runs in CI before merging, thus the hold.

The e2e-agnostic-upgrade failed on a GCE error.

haircommander:
[hm doesn't look related](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2177/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1320773273037836288)
/test all

PR #2178, opened 2020-10-24
Bug 1890074: pkg/daemon: avoid duplicate extension verification

vrutkovs:
When package aliases (or groups) were introduced [recently](https://github.com/openshift/machine-config-operator/pull/2170),
this list is now being verified twice - during `validateExtensions` on
MCC and in `generateExtensionsArgs` on MCD. This breaks OKD installs,
as we'd like to avoid verifying a pre-approved list of RPMs on FCOS.

However, MCC can't know which systems it would manage, so it applies
RHCOS allowlist on all systems unknowingly.

This commit refactors extensions verification and application:

* available packages are listed in `constants.SupportedPackages`
* groups/aliases MCD knows about are listed in `constants.SupportedPackageAliases`
* `validateExtensions` ensures that `spec.extensions` items are in either list
* `generateExtensionsArgs` expands aliases into a list of RPMs

On OKD `validateExtensions` is not called, however `generateExtensionsArgs`
would still apply (so `kernel-devel` would automatically bring in `kernel-headers`).
This might bite us in the future, but at least it won't block
installs, so its safe to merge now.

/cc @cgwalters @sinnykumari @kikisdeliveryservice

### DIFF
--- a/pkg/daemon/constants/constants.go
+++ b/pkg/daemon/constants/constants.go
@@ -56,3 +56,12 @@ const (
 	// to proceed and attempt to "reconcile" to the new "desiredConfig" state regardless.
 	MachineConfigDaemonForceFile = "/run/machine-config-daemon-force"
 )
+
+var (
+	// SupportedPackages lists RPMs available in machine-os-content for RHCOS
+	SupportedPackages = []string{"kernel-headers", "usbguard"}
+	// SupportedPackageAliases expands allowed aliases to a list of RPMs to install
+	SupportedPackageAliases = map[string][]string{
+		"kernel-devel": {"kernel-headers", "kernel-devel"},
+	}
+)
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -865,18 +865,15 @@ func generateExtensionsArgs(oldConfig, newConfig *mcfgv1.MachineConfig) []string
 		}
 	}
 
-	// Supported extensions has package list info that is required
-	// to enable an extension
-	extensions := getSupportedExtensions()
-
+	// Extension may be an RPM name or an alias for a list of RPMs
 	extArgs := []string{"update"}
 	for _, ext := range added {
-		for _, pkg := range extensions[ext] {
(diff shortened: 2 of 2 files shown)


vrutkovs:
/retest

rhel7/gcp-op flakes

sinnykumari:
Thank you for creating the PR to fix the OKD issue.
I have created an alternate PR to fix this issue https://github.com/openshift/machine-config-operator/pull/2181 . Once OKD base package stabilizes, it encourages to group all the packages as a single extension which OKD can have it installed by default.

vrutkovs:
>Once OKD base package stabilizes, it encourages to group all the packages as a single extension which OKD can have it installed by default

I don't think we want that. Some packages may be shipped in OS image, but some may be installed from Fedora repos, so we'd prefer to avoid restricting OKD users to a hardcoded list of images

sinnykumari:
> > Once OKD base package stabilizes, it encourages to group all the packages as a single extension which OKD can have it installed by default
> 
> I don't think we want that. Some packages may be shipped in OS image, but some may be installed from Fedora repos, so we'd prefer to avoid restricting OKD users to a hardcoded list of images

Hmm, right. Using Fedora repo is another possibility for users to layer packages. 
Just curious, won't this make hard to keep track of what user is installing and issues OKD cluster may get during upgrade?

https://github.com/openshift/machine-config-operator/pull/2181 preserves the previous implementation of extension on FCOS, so OKD existing package layering approach should stay as it is.

vrutkovs:
>Just curious, won't this make hard to keep track of what user is installing and issues OKD cluster may get during upgrade?

It certainly would, that would be user responsibility though. In order to boot its sufficient to avoid removing default list of packages added by installer.

#2181 would work, we might want to investigate unrestricted package list later.

/hold

sinnykumari:
> #2181 would work, we might want to investigate unrestricted package list later.

definitely!

vrutkovs:
Closed in favor of #2181

PR #2183, opened 2020-10-27
templates/master/00-master/_base/files: remove etcd-root-ca

hexfusion:
This CA is not used by the product and causes confusion.


hexfusion:
> infra failure 
/test e2e-agnostic-upgrade

yuqi-zhang:
Value should be still be in `/etc/kubernetes/ca.crt`. Given that tests pass and Sam believes this is not used,
/lgtm

PR #2185, opened 2020-10-27
[Baremetal] Remove BootstrapIP from Keepalived template

yboaron:
Since we eliminate the need for the master nodes to know the bootstrap IP address
for unicast Keepalived, we can delete BootstrapIP from Keepalived template.


cybertron:
/retest

The okd job looks pretty red, but the other one seems to be passing about 50%.

Not that this affects either of those platforms though. Green CI just gives me warm fuzzies. ;-)

PR #2188, opened 2020-10-28
[baremetal] Add dispatcher script to statically configure DHCP address

cybertron:
We have a feature request for nodes to be able to initially get
assigned their IP addresses via infinite DHCP leases, but then to
have those addresses statically configured on the node so that if
the DHCP server is unavailable for any reason they don't lose their
network configuration.

This adds a dispatcher script for NetworkManager that does so. It
looks at the DHCP assigned network settings and adds static
configuration with the same values.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
+++ b/templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
@@ -1,0 +1,63 @@
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/30-static-dhcp"
+contents:
+  inline: |
+    #!/bin/bash
+    set -ex -o pipefail
+
+    if [[ "{{ .NetworkType }}" == "OVNKubernetes" && "$CONNECTION_ID" == "Wired Connection" ]]
+    then
+        >&2 echo "Refusing to modify default connection."
+        exit 0
+    fi
+
+    if [ -z ${DHCP4_IP_ADDRESS:-} ]
+    then
+        >&2 echo "Not a DHCP4 address. Ignoring."
+        exit 0
+    fi
+
+    if [ ${DHCP4_DHCP_LEASE_TIME:-0} -lt 4294967295 ]
+    then
+        >&2 echo "Not an infinite DHCP4 lease. Ignoring."
+        exit 0
+    fi
+
+    IPS=($IP4_ADDRESS_0)
+    CIDR=${IPS[0]}
+    GATEWAY=${IPS[1]}
+
+    TYPE=$(nmcli --get-values connection.type connection show "$CONNECTION_ID")
+
+    if ! nmcli con show inf-lease-to-static
+    then
+        nmcli con add type "$TYPE" con-name inf-lease-to-static
+    fi
+    nmcli con mod inf-lease-to-static \
+      conn.interface "$1" \
+      connection.autoconnect yes \
+      ipv4.addresses "$CIDR" \
+      ipv4.method manual \
+      ipv4.gateway "$GATEWAY" \
+      ipv4.dns "$IP4_NAMESERVERS"
+
+    if [ -n "$IP4_DOMAINS" ]; then
+        nmcli con mod inf-lease-to-static ipv4.dns-search "$IP4_DOMAINS"
+    fi
+    plus=''
+    for i in $(seq 0 $(($IP4_NUM_ROUTES-1)) )
+    do
+        varname="IP4_ROUTE_$i"
+        nmcli con mod inf-lease-to-static ${plus}ipv4.routes "${!varname}"
+        plus='+'
+    done
+
+    nmcli con up inf-lease-to-static
+
+    # Copy it from the OverlayFS mount to the persistent lowerdir
+    cp "/etc/NetworkManager/system-connections-merged/inf-lease-to-static.nmconnection" /etc/NetworkManager/system-connections
+
+    if [ -n "${DHCP4_HOST_NAME:-}" ]
+    then
+        hostnamectl set-hostname --static --transient "$DHCP4_HOST_NAME"
+    fi


cybertron:
/hold

This is a PoC. If we decide to pursue this it will almost certainly need some refinement.

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
```diff
@@ -0,0 +1,30 @@
+mode: 0755
+path: "/etc/NetworkManager/dispatcher.d/30-static-dhcp"
+contents:
+  inline: |
+    #!/bin/bash
+    set -ex
+
+    if [ ${DHCP4_DHCP_LEASE_TIME:-0} -lt 4294967295 ]
```
celebdor, line 8: Does this not exist in IPv6?

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
```diff
@@ -0,0 +1,30 @@
...
+    then
+        echo "Not an infinite lease. Ignoring."
```
celebdor, line 10: Should probably go to stderr

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
```diff
@@ -0,0 +1,30 @@
...
+        exit 0
+    fi
+
+    IPS=($IP4_ADDRESS_0)
+    CIDR=${IPS[0]}
+    GATEWAY=${IPS[1]}
```
celebdor, line 16: Is there some way to retrieve all of the dhcp configuration data and easily add it? Things like hostname would get probably persisted by our other dispatcher script, but we may be missing things like the search domain.

celebdor (changes requested):

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
celebdor: Maybe using dbus?

celebdor: Getting the active connections
```
busctl get-property --json=pretty org.freedesktop.NetworkManager /org/freedesktop/NetworkManager org.freedesktop.NetworkManager ActiveConnections
{
        "type" : "ao",
        "data" : [
                "/org/freedesktop/NetworkManager/ActiveConnection/802",
                "/org/freedesktop/NetworkManager/ActiveConnection/2"
        ]
}
```

Checking the name (which I suspect is what we get from dispatcher scripts)
```
❯ busctl get-property --json=pretty org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/ActiveConnection/802 org.freedesktop.NetworkManager.Connection.Active Id
{
        "type" : "s",
        "data" : "MYWIFI"
}
```

Retrieving the DHCP4 object path:
```
busctl get-property --json=pretty org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/ActiveConnection/802 org.freedesktop.NetworkManager.Connection.Active Dhcp4Config
{
        "type" : "o",
        "data" : "/org/freedesktop/NetworkManager/DHCP4Config/821"
}
```

Retrieving the DHCPv4 properties
```
busctl get-property --json=pretty org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/DHCP4Config/821 org.freedesktop.NetworkManager.DHCP4Config Options
{
	"type" : "a{sv}",
	"data" : {
		"dhcp_lease_time" : {
			"type" : "s",
			"data" : "3600"
		},
		"domain_name" : {
			"type" : "s",
			"data" : "home.myhome.cz"
		},
		"domain_name_servers" : {
			"type" : "s",
			"data" : "192.168.2.1"
		},
		"domain_search" : {
			"type" : "s",
			"data" : "home.myhome.cz"
		},
		"expiry" : {
			"type" : "s",
			"data" : "1603931004"
		},
		"host_name" : {
			"type" : "s",
			"data" : "bucek.home.myhome.cz"
		},
		"ip_address" : {
			"type" : "s",
			"data" : "192.168.2.37"
		},
		"requested_broadcast_address" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_domain_name" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_domain_name_servers" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_domain_search" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_host_name" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_interface_mtu" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_ms_classless_static_routes" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_nis_domain" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_nis_servers" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_ntp_servers" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_rfc3442_classless_static_routes" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_root_path" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_routers" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_static_routes" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_subnet_mask" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_time_offset" : {
			"type" : "s",
			"data" : "1"
		},
		"requested_wpad" : {
			"type" : "s",
			"data" : "1"
		},
		"routers" : {
			"type" : "s",
			"data" : "192.168.2.1"
		},
		"subnet_mask" : {
			"type" : "s",
			"data" : "255.255.255.0"
		}
	}
}
```

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
cybertron: I only implemented ipv4 in this version. Based on my discussion with the NM team, I would probably argue that for this use case in ipv6 we should just use SLAAC addressing. If there's a reason we can't do that then we can always add support for DHCPv6 here too.

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
cybertron: Ah, yeah I'll update this to match the logging in the other dispatcher scripts.

templates/common/baremetal/files/NetworkManager-static-dhcp.yaml
cybertron: Hmm, I'll have to look into this. I don't see the search domain in the dispatcher env vars, but I also didn't specify one on the DHCP server side so it may just not have been populated.

cybertron: If the DHCP server sets a search domain it will be available in the dispatcher env vars. We just need to map it to nmcli, which I will do in my next update to this. If there's anything else you think we should capture, let me know.

cybertron:
/hold cancel

I've tested this locally with infinite DHCP leases and it worked fine.

celebdor (approved):

cybertron:
/hold

We discovered that this would cause problems on reboot. Fix coming.

celebdor:
Just tested it with a small modification on one of my masters with openshift-sdn. It worked well. Here some snippets:

```
mode: 0755
path: "/etc/NetworkManager/dispatcher.d/30-static-dhcp"
contents:
  inline: |
    #!/bin/bash
    set -ex -o pipefail
    if [[ "{{ .NetworkType }}" == "OVNKubernetes" && "$CONNECTION_ID" == "Wired Connection" ]]
    then
        >&2 echo "Refusing to modify default connection."
        exit 0
    fi

    if [ -z ${DHCP4_IP_ADDRESS:-} ]; then
        >&2 echo "Not a DHCP4 address. Ignoring."
        exit 0
    fi
    if [ ${DHCP4_DHCP_LEASE_TIME:-0} -lt 4294967295 ]; then
        >&2 echo "Not an infinite DHCP4 lease. Ignoring."
        exit 0
    fi
    IPS=($IP4_ADDRESS_0)
    CIDR=${IPS[0]}
    GATEWAY=${IPS[1]}
    TYPE=$(nmcli --get-values connection.type connection show "$CONNECTION_ID")
    if ! nmcli con show inf-lease-to-static
    then
        nmcli con add type "$TYPE" con-name inf-lease-to-static
    fi
    nmcli con mod inf-lease-to-static \
      conn.interface "$1" \
      connection.autoconnect yes \
      ipv4.addresses "$CIDR" \
      ipv4.method manual \
      ipv4.gateway "$GATEWAY" \
      ipv4.dns "$IP4_NAMESERVERS"
    if [ -n "$IP4_DOMAINS" ]; then
        nmcli con mod inf-lease-to-static ipv4.dns-search "$IP4_DOMAINS"
    fi
    plus=''
    for i in $(seq 0 $(($IP4_NUM_ROUTES-1)) )
    do
        varname="IP4_ROUTE_$i"
        nmcli con mod inf-lease-to-static ${plus}ipv4.routes "${!varname}"
        plus='+'
    done
    nmcli con up inf-lease-to-static
    # Copy it from the OverlayFS mount to the persistent lowerdir
    cp "/etc/NetworkManager/system-connections-merged/inf-lease-to-static.nmconnection" /etc/NetworkManager/system-connections

    if [ -n "${DHCP4_HOST_NAME:-}" ]
    then
        hostnamectl set-hostname --static --transient "$DHCP4_HOST_NAME"
    fi
```

After making these changes, that make sure that it works for openshift SDN and for OVN kubernetes, you can see in the following image that there is a reboot of master-0 (I also rebooted master-1) after I had killed the dnsmasq that provided DHCP functionality.
You can also see that the nodes boot up with the hostname they are supposed and become ready again.

image

cybertron:
/hold cancel

I also had success rebooting nodes with the latest version of this. I think it's good to go.

celebdor:
/lgtm

The version in this PR now matches what I tested successfully in https://github.com/openshift/machine-config-operator/pull/2188#issuecomment-737629752

sinnykumari:
overall lgtm.
/approve

PR #2189, opened 2020-10-29
Bug 1892448: daemon: add back metrics for pivot error

sinnykumari:
It accidentally got removed in https://github.com/openshift/machine-config-operator/commit/6757381f09358f05fe5eefedfa85480ce09b3720#diff-349c0748c3f52201852d3027c29daf618b45300b949482728df6666a3f9ba245L1325 . Thanks to @kikisdeliveryservice for noticing this.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -347,6 +347,7 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 
 	// Update OS
 	if err := dn.updateOS(newConfig, osImageContentDir); err != nil {
+		MCDPivotErr.WithLabelValues(dn.node.Name, newConfig.Spec.OSImageURL, err.Error()).SetToCurrentTime()
 		return err
 	}
 


pkg/daemon/update.go
```diff
@@ -1484,6 +1484,7 @@ func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig, osImageContentDir strin
 	glog.Infof("Updating OS to %s", newURL)
 	client := NewNodeUpdaterClient()
 	if _, err := client.Rebase(newURL, osImageContentDir); err != nil {
+		MCDPivotErr.WithLabelValues(dn.node.Name, newURL, err.Error()).SetToCurrentTime()
```
sinnykumari, line 1487: updateOS() unit test segfaults if it reaches here. Made unsuccessful attempt to start a fake metrics listener.

kikisdeliveryservice:
Need to look and fix cardinality

/hold

sinnykumari:
> Need to look and fix cardinality

About cardinality, do we want to wait for some duration before firing MCDPivotErr alert?
Also, was wondering if severity of Pivot failure be moved from warning to critical as node hasn't been updated.

pkg/daemon/update.go
kikisdeliveryservice: since this is the only place returning an error we could just move this to capture the return value of updateOS and leave your unit tests alone?

like here: https://github.com/openshift/machine-config-operator/blob/dd9050c7faf5400cde55608a0393f90bfc213e53/pkg/daemon/update.go#L350

sinnykumari: yes, we can. There is one more place which calls updateOS() https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/daemon.go#L1036 but I really don't know if it is ever called. I updated this code to maintain the compatibility with previous code.

kikisdeliveryservice:
> > Need to look and fix cardinality
> 
> About cardinality, do we want to wait for some duration before firing MCDPivotErr alert?
> Also, was wondering if severity of Pivot failure be moved from warning to critical as node hasn't been updated.

Talked to Sinny and we'll sync on this, in a follow-on pr  when I fix the remaining cardinality issues in metrics. For now, we will just let this PR restore the missing metric to get it back into 4.6 ASAP

pkg/daemon/update.go
sinnykumari: moved it on level up

kikisdeliveryservice (approved):
Thanks Sinny!!

/lgtm

kikisdeliveryservice:
some known e2e-aws problems today... =/

PR #2190, opened 2020-10-30
Use hostNetwork: true for operator and controller

cgwalters:
We want to support switching cluster network types "day 2".
This may involve a MachineConfig rollout, so using the
host network will avoid potential deadlocks.

### DIFF
--- a/install/0000_80_machine-config-operator_04_deployment.yaml
+++ b/install/0000_80_machine-config-operator_04_deployment.yaml
@@ -44,6 +44,10 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+      # We want to support switching cluster network types "day 2".
+      # This may involve a MachineConfig rollout, so using the
+      # host network will avoid potential deadlocks.
+      hostNetwork: true
       tolerations:
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
--- a/manifests/machineconfigcontroller/deployment.yaml
+++ b/manifests/machineconfigcontroller/deployment.yaml
@@ -26,6 +26,10 @@ spec:
             memory: 50Mi
         terminationMessagePolicy: FallbackToLogsOnError
       serviceAccountName: machine-config-controller
+      # We want to support switching cluster network types "day 2".
+      # This may involve a MachineConfig rollout, so using the
+      # host network will avoid potential deadlocks.
+      hostNetwork: true
       nodeSelector:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -793,6 +793,10 @@ spec:
             memory: 50Mi
         terminationMessagePolicy: FallbackToLogsOnError
       serviceAccountName: machine-config-controller
+      # We want to support switching cluster network types "day 2".
+      # This may involve a MachineConfig rollout, so using the
+      # host network will avoid potential deadlocks.
+      hostNetwork: true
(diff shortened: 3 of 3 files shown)


cgwalters:
It's not 100% clear to me that we don't *also* need to change the MCO with this.  I *think* the only potential consequence if the MCO is temporarily unable to network is that the clusteroperator status won't update; otherwise...once a control plane rollout is done and the node hosting the MCO is drained, if the MCO lands on an updated node then status should start working again.

But maybe it would just overall be easier to understand if we made *all* containers in the MCO `hostNetwork: true` and hence fully severed any dependency on the SDN.

sinnykumari:
> But maybe it would just overall be easier to understand if we made _all_ containers in the MCO `hostNetwork: true` and hence fully severed any dependency on the SDN.

+1

I am curious to know why MCO pods didn't opt for host networking in the beginning?

pliurh:
> But maybe it would just overall be easier to understand if we made all containers in the MCO hostNetwork: true and hence fully severed any dependency on the SDN.

+1

cgwalters:
Updated to make the operator use `hostNetwork: true` too.

pliurh:
@cgwalters If we want to decouple the MCO with the clusterNetwork, using hostNetwork may not be sufficient. Currently, MCO uses the in-cluster kube client configuration which talks to the apiserver VIP of `172.30.0.1`. The route to this VIP is also handled by the clusterNetwork providers on hostnetwork. So if we want MCO has no dependency on the clusterNetwork, we shall also use the URI of `api-int.<cluster-domain>` to communicate with the apiserver.

crawford:
Is this part of a larger epic or enhancement?

darkmuggle:
/retest

The SDN Team is asking that we include this in 4.8. AFAIK, this looks fine.  
@skinny or  @kikisdeliveryservice  give the +1, would one of you feel comfortable with an LGTM?

sinnykumari:
With my limited cluster networking knowledge can't think of any issue, this PR looks good to me as it is.
Do we have CI coverage somewhere to look at where we perform Day2 networking switch?

kikisdeliveryservice:
> Is this part of a larger epic or enhancement?

@pliurh could you answer the above?

pliurh:
Here is the story of 4.8 https://issues.redhat.com/browse/SDN-1546

Currently, we don't have CI for the SDN migration. Because during the migration, a manual reboot step is required, which I haven't figure out how it can be done in openshift-ci. We used to have a baremetal CI environment for that, but it was put to other uses. :(

My two cents: Put the SDN migration requirement aside, considering MCO's position, I think it still makes sense to decouple the MCO from the cluster network.

cgwalters:
> Because during the migration, a manual reboot step is required, which I haven't figure out how it can be done in openshift-ci. 

Just cordon/drain the node and then use e.g. `oc debug node/$node chroot /host reboot`.

danwinship:
> I am curious to know why MCO pods didn't opt for host networking in the beginning?

> +1 As someone who be confused later... I think this is the way to go unless there are good reasons not to?

For pods that don't expose any services, the argument against being hostNetwork is mostly just that it's more privilege (eg, now you can't restrict the pods with NetworkPolicy, etc). Especially, pods that are both `privileged` and `hostNetwork` can do a lot of damage if they are compromised. (But neither MCO nor MCC is privileged.)

(If either pod exposes any ports [for metrics?] you will need to update https://github.com/openshift/enhancements/blob/master/enhancements/network/host-port-registry.md)

> So if we want MCO has no dependency on the clusterNetwork, we shall also use the URI of `api-int.<cluster-domain>` to communicate with the apiserver.

cf #2232

PR #2194, opened 2020-11-01
templates: convert kubelet proxy config into a dropin

vrutkovs:
This ensures OKD 4.5 -> 4.6 upgrade can complete. Currently kubelet config is considered to be a file, but in 4.6 config its a drop in, so MCD can't properly create it.

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -40,3 +40,20 @@ contents: |
 
   [Install]
   WantedBy=multi-user.target
+dropins:
+- name: 10-mco-default-env.conf
+  contents: |
+    [Unit]
+    Description=MCO environment configuration
+    {{if .Proxy -}}
+    [Service]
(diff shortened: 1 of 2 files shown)


vrutkovs:
/hold

Testing 4.5 latest -> this PR -> 4.6 CI update

vrutkovs:
Flakes

/retest

vrutkovs:
4.5 -> this PR upgrade passes, but 4.5z -> 4.6 still fails. Apparently MCD needs to be fixed

cgwalters:
Ahh this is because 6f5b3c52f671da3230a40686adc2deaa66d3e5db

vrutkovs:
Right, that uncovered a bug in MCD stale file handling:
1. Create systemd dropin via a file
2. Realize that its suboptimal, convert into a dropin
3. Ignition will rewrite the file, but MCD's `deleteStaleData` would remove the file
4. Machine would reboot and MCD would complain that file is not found.

I'll file a bug and fix it in MCD master/release-4.6

vrutkovs:
https://github.com/openshift/machine-config-operator/pull/2196 to fix this

PR #2197, opened 2020-11-03
Bump to the latest Ignition 2.7.0

cgwalters:
On general principle; that's what's in current RHCOS 4.7 and we
should support that version of configs.


ashcrow:
cc @miabbott

ashcrow (approved):

jlebon:
I think this is pretty safe because it's purely a vendor update and parsing-related imports are versioned to the spec, so this isn't on its own causing us to accept spec 3.2.0 configs, but it sets us up for that (and we'll need that for 4.7).

/lgtm

PR #2198, opened 2020-11-03
WIP [fcos] pkg/daemon: don't delete a file if its replaced with a dropin

vrutkovs:
Cherry-pick of https://github.com/openshift/machine-config-operator/pull/2196 on FCOS

Some systemd service settings may be defined in .Storage.Files and in .Systemd.Units.Dropins. Dropins is preferable, so users may want to migrate to the way.

However if a file representing a dropin was defined in .Storage.Files and converted into .Systemd.Units.Dropin in new version of MC, it would be placed by Ignition's dropin implementation and then garbage-collected by MCD's deleteStaleData.

This change ensures we're checking Systemd.Units.Dropins before attempting to remove this file. This is crucial for OKD 4.5 -> 4.6 upgrades - kubelet MCO config was defined as a file in 4.5 and converted into a dropin in 4.6

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -843,20 +843,27 @@ func restorePath(path string) error {
 // parse path to find out if its a systemd dropin
 // Returns is dropin (true/false), service name, dropin name
 func isSystemdDropin(path string) (bool, string, string) {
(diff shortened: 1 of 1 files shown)


vrutkovs:
not necessary, a fix needs to go in MCO 4.6

PR #2202, opened 2020-11-05
Restrict default file mode for /etc/kuberentes/kubeconfig

JAORMX:
**- What I did**

The file `/etc/kubernetes/kubeconfig` which was handled by the MCO now should
have a more restritive file mode

**- How to verify it**

On a host, the file mode for the aforementioned file should be 0640

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -123,7 +123,8 @@ func appendKubeConfig(conf *igntypes.Config, f kubeconfigFunc) error {
 	if err != nil {
 		return err
 	}
-	err = appendFileToIgnition(conf, defaultMachineKubeConfPath, string(kcData))
+	// Use 0640 as mode for security, so only root:root can read this.
+	err = appendFileToIgnitionWithMode(conf, defaultMachineKubeConfPath, string(kcData), 416)
 	if err != nil {
 		return err
 	}
@@ -157,6 +158,10 @@ func getNodeAnnotation(conf string) (string, error) {
 
 func appendFileToIgnition(conf *igntypes.Config, outPath, contents string) error {
 	fileMode := int(420)
+	return appendFileToIgnitionWithMode(conf, outPath, contents, fileMode)
+}
+
+func appendFileToIgnitionWithMode(conf *igntypes.Config, outPath, contents string, fileMode int) error {
 	overwrite := true
 	source := getEncodedContent(contents)
 	file := igntypes.File{


cgwalters (approved):
Yep, makes sense to me.

cgwalters:
One thing to double check is that the various privileged pods like the apiserver run as `uid 0` - I believe they do.  I also think existing e2e testing should cover this just fine.

JAORMX:
> One thing to double check is that the various privileged pods like the apiserver run as `uid 0` - I believe they do. I also think existing e2e testing should cover this just fine.

They do run with uid `0`

cgwalters:
Hmm CNO is crashlooping: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2202/pull-ci-openshift-machine-config-operator-master-e2e-aws/1324357713387327488/artifacts/e2e-aws/gather-extra/oc_cmds/pods

cgwalters:
And yep [there's the smoking gun](https://github.com/openshift/cluster-network-operator/blob/1ee0aaf1bdb838d6cdeef221915b32237aba86f8/manifests/0000_70_cluster-network-operator_03_deployment.yaml#L77):

```
      volumes:
        - name: host-kubeconfig
          hostPath:
            path: /etc/kubernetes/kubeconfig
      restartPolicy: Always
      securityContext:
        runAsNonRoot: true
        runAsUser: 65534
      tolerations:
```

cgwalters:
/hold
until above is fixed.

runcom:
not sure gcp-op is gonna pass, yet. But let's see

JAORMX:
@cgwalters thanks for checking this out! Seems we gotta fix the CNO then! Would be better if they use their own kubeconfig... or else they'll have to mount it as root.

JAORMX:
/remove-lifecycle stale

I'll get back to this soon....

JAORMX:
No dice. Seems I'll have to reach out to the team that works on the SNO to get this fixed.

kikisdeliveryservice:
@JAORMX any update on this?

JAORMX:
@kikisdeliveryservice This is blocked as this file is used by the CNO... that needs an update but I haven't found time to fix that. If someone else can pick that up, it's required to get this working. Also, I'm on PTO until middle of August.

kikisdeliveryservice:
> @kikisdeliveryservice This is blocked as this file is used by the CNO... that needs an update but I haven't found time to fix that. If someone else can pick that up, it's required to get this working. Also, I'm on PTO until middle of August.

Thanks for the update and sorry for interrupting your time off :heart:

PR #2204, opened 2020-11-05
Revert "crio: drop infra container when possible"

deads2k:
This reverts commit 6eb37894974b71080648a92bbb71fd201c5a84f4.

reverts https://github.com/openshift/machine-config-operator/pull/2177, one of two commits that merged before the permafail of e2e-serial on OOMs

PR #2205, opened 2020-11-05
Revert "Dedupe on-prem templates"

fabianvf:
Reverts openshift/machine-config-operator#2079

One of two PRs that were merged before the first serial test failure.


fabianvf:
looks like #2204 is the one

PR #2206, opened 2020-11-05
Bug 1879099: Inject version into controllerconfig, refuse mismatches

cgwalters:

Per Jerry's comment in Bugzilla:

> 1. The new MCO rolls out. It first syncs the controllerconfig to the new one as part of the RenderConfig step above.
> 2. The running (old) MCC's template controller has an informer on the controllerconfig, which since the previous step updated the config, calls a syncControllerConfig. This re-generates the base master config (00-master MC)
> 3. The running (old) MCC now sees a MC change which calls another sync call to generate a new rendered config, and instructs the master pool to start an update
> 4. The new MCC rolls out, and does the whole rendering process, creating the actual updated MC for the master pool, which it waits until the previous one is finished, and starts the new (real) update.

After this patch, the old MCC will refuse to render from a controllerconfig generated by the new MCO.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift/machine-config-operator/lib/resourceread"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	templatectrl "github.com/openshift/machine-config-operator/pkg/controller/template"
+	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	"github.com/openshift/machine-config-operator/pkg/operator/assets"
 	"github.com/openshift/machine-config-operator/pkg/version"
 )
@@ -391,6 +392,11 @@ func (optr *Operator) syncMachineConfigController(config *renderConfig) error {
 		return err
 	}
 	cc := resourceread.ReadControllerConfigV1OrDie(ccBytes)
+	// Propagate our binary version into the controller config to help
+	// suppress rendered config generation until a corresponding
+	// new controller can roll out too.
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1879099
+	cc.Annotations[daemonconsts.GeneratedByVersionAnnotationKey] = version.Raw
 	_, _, err = resourceapply.ApplyControllerConfig(optr.client.MachineconfigurationV1(), cc)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


cgwalters:
This is an alternative to https://github.com/openshift/machine-config-operator/pull/2112 that uses an annotation

cgwalters:
So in this code I didn't actually adjust the version in the bootstrap path because that requires special casing the controllerconfig rendering there, and I'm not 100% sure it's required.  (If it's not I'll delete the first hunk too)

yuqi-zhang:
hm, I did try using this on 4.5, and I get:
```
F1109 21:34:21.714659       1 bootstrap.go:47] error running MCC[BOOTSTRAP]: Ignoring controller config generated without machineconfiguration.openshift.io/generated-by-version annotation (my version: machine-config-daemon-4.5.0-202006231303.p0-55-gaaef53f0)
```

cgwalters:
OK tried to fix the bootstrap

cgwalters:
OK e2e-aws got a cluster up, this only failed on some test flakes.
/test e2e-aws
and e2e-gcp-op is also known broken for other reasons, this can't affect it.

pkg/operator/sync.go
```diff
@@ -395,6 +396,11 @@ func (optr *Operator) syncMachineConfigController(config *renderConfig) error {
 	if err != nil {
 		return err
 	}
+	// Propagate our binary version into the controller config to help
+	// suppress rendered config generation until a corresponding
+	// new controller can roll out too.
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1879099
+	cc.Annotations[daemonconsts.GeneratedByVersionAnnotationKey] = version.Raw
```
runcom, line 403: wait, why isn't this CC created in the cluster? should this be when we create the CC in the operator (sync.go.syncRenderConfig)? when this code runs we're sure that whatever we're upgrading from doesn't have this version and the check is going to fail

cgwalters: > wait, why isn't this CC created in the cluster?

"In the cluster"?  What do you mean by that?

> when this code runs we're sure that whatever we're upgrading from doesn't have this version and the check is going to fail

We're reading a file that lives inside our own binary via bindata.
As far as we can tell it was just done this way originally because most of the rest of the code here is templating stuff too.

I mean conceptually all of this could be inline Go structs too, I think it would be clearer in *this* case but eh.

> what's happening right now is that this never gets synced in etcd cause it's after the actual sync to the api

Good catch!   I think in most cases it was working because we'd end up re-syncing again and at that point we'd get the version set, but we clearly want to ensure we sync to N and not N-1.

Fixed.

runcom: > "In the cluster"? What do you mean by that?

I guess I just meant what I wrote in the last sentence lol sorry about the confusion

PR #2207, opened 2020-11-05
Bug 1895099: Fix VSphere UPI not populating PlatformStatus

bcrochet:
The VSphere UPI install does not fully populate PlatformStatus. When we deduped
the templates, we didn't account for this being nil. This should render these
templates empty when that PlatformStatus is nil.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/vsphere/files/etc-systemd-system-crio-stream-address.conf.yaml
+++ b/templates/common/vsphere/files/etc-systemd-system-crio-stream-address.conf.yaml
@@ -2,11 +2,7 @@ mode: 0644
 path: "/etc/systemd/system/crio.service.d/20-stream-address.conf"
 contents:
   inline: |
-    {{ if .Infra -}}
-    {{ if .Infra.Status -}}
-    {{ if .Infra.Status.PlatformStatus -}}
-    {{ if .Infra.Status.PlatformStatus.VSphere -}}
-    {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
+    {{ if (onPremPlatformAPIServerInternalIP .) -}}
     [Service]
     ExecStart=
     ExecStart=/usr/bin/crio \
@@ -14,8 +10,4 @@ contents:
           $CRIO_STORAGE_OPTIONS \
           $CRIO_NETWORK_OPTIONS \
           $CRIO_METRICS_OPTIONS
-    {{ end -}}
-    {{ end -}}
-    {{ end -}}
-    {{ end -}}
     {{ end -}}
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -410,7 +410,12 @@ func onPremPlatformIngressIP(cfg RenderConfig) (interface{}, error) {
 		case configv1.OpenStackPlatformType:
 			return cfg.Infra.Status.PlatformStatus.OpenStack.IngressIP, nil
 		case configv1.VSpherePlatformType:
-			return cfg.Infra.Status.PlatformStatus.VSphere.IngressIP, nil
+			if cfg.Infra.Status.PlatformStatus.VSphere != nil {
+				return cfg.Infra.Status.PlatformStatus.VSphere.IngressIP, nil
+			}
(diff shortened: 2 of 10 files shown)


patrickdillon:
cc @jcpowermac

patrickdillon:
Not sure if these are setup:
/test e2e-vsphere
/test e2e-vsphere-upi

templates/common/vsphere/files/etc-systemd-system-crio-stream-address.conf.yaml
```diff
@@ -2,20 +2,10 @@ mode: 0644
 path: "/etc/systemd/system/crio.service.d/20-stream-address.conf"
 contents:
   inline: |
-    {{ if .Infra -}}
-    {{ if .Infra.Status -}}
-    {{ if .Infra.Status.PlatformStatus -}}
-    {{ if .Infra.Status.PlatformStatus.VSphere -}}
-    {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
```
patrickdillon, line 9: don't we need to replace this with the simpler conditional used in other templates? `{{ if (onPremPlatformAPIServerInternalIP .) -}}`

patrickdillon:
https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/NetworkManager-mdns-hostname.yaml also needs to be checked for `onPremPlatformAPIServerInternalIP`

patrickdillon:
as does: 
https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/NetworkManager-onprem.conf.yaml

https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/mdns-publisher.yaml

https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/units/nodeip-configuration.service.yaml

templates/common/vsphere/files/etc-systemd-system-crio-stream-address.conf.yaml
bcrochet: There is no reference to the PlatformStatus in that file. Are you saying you don't want it at all?

patrickdillon:
CI is consistently failing with `Failed pulling builder image` not sure what's going on there. I do wonder if the issue fixed by this pr could be causing part of the problem, as 4.7 images are not being promoted.

/lgtm

jcpowermac:
MCO is still configured for template vs. multistep at least for UPI.

bcrochet:
This may not pass until #2209 is merged. It might need a rebase after.

jcpowermac:
@bcrochet looking at job results the installation succeeded the tests failed. I think this could move forward as-is.

@patrickdillon wdyt?

bcrochet:
> @bcrochet looking at job results the installation succeeded the tests failed. I think this could move forward as-is.
> 
> @patrickdillon wdyt?

If you're ok with the order, it doesn't matter to me. This patch would probably allow #2209 to pass.

yuqi-zhang (approved):

patrickdillon:
> > @bcrochet looking at job results the installation succeeded the tests failed. I think this could move forward as-is.
> > @patrickdillon wdyt?
> 
> If you're ok with the order, it doesn't matter to me. This patch would probably allow #2209 to pass, or vice versa. This one is probably failing because of the problem fixed in the #2209.

All vSphere installs fail without this PR. Merging this PR will move us closer to stability in CI and I don't think we should wait for a clean CI run to merge. I'm fine waiting a bit if it's better to merge #2209 first but I'd rather get the periodic back on track and use that as a barometer.

bcrochet:
e2e-gcp-op seems to be failing across the board.

jcpowermac:
Any chance we can get an override for gcp-op? This PR is needed for vSphere UPI to install and CI to pass.

yuqi-zhang:
Ok given the AWS still slowly recovering, and the fact that this has passed all the other tests in the past + it should not affect cloud platforms, I will override this.

/override ci/prow/e2e-gcp-op

PR #1672, opened 2020-04-22
Bug 1828622: bumps kubelet log level to 4

rphillips:
**- What I did**
Bumping the kubelet log level to 4 to aid in debugging an issue.

**- How to verify it**

**- Description for the changelog**

PR #2211, opened 2020-11-09
Bug 1895385: Drop kubelet logging back down to level 3

kikisdeliveryservice:
KUBELET_LOG_LEVEL was bumped to level 4 to aid in
debugging some issue by node team (see https://github.com/openshift/machine-config-operator/pull/1672).
However customers upgrading to 4.5 in prod, are seeing this bump
causing extra GBs of data to be saved.

Closes: BZ 1895385

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -10,7 +10,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
-  Environment="KUBELET_LOG_LEVEL=4"
+  Environment="KUBELET_LOG_LEVEL=3"
 {{- if .KubeletIPv6}}
   Environment="KUBELET_NODE_IP=::"
 {{- end}}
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
Also some question if 3 is too high? Or is 3 the necessary level for proper debugging of kubelet?

mrobson:
From a large production cluster point of view, 3 is still too high for everyday running in production. OCP 3.x ran at log level 2.

Scott Worthington did some sampling on the issue on 4.5.17 in a 3 masters / 9 workers node cluster:
loglevel=2  for masters and workers: 14000 documents per 5 minutes in ES 
loglevel=3  for masters and workers: 42000 documents per 5 minutes in ES
loglevel=4  for masters and workers: 126000 documents per 5 minutes in ES

There is also a big different in the level of detail / verbosity of each log line at level 4 versus level 2 which has additional storage implications.

kikisdeliveryservice:
> From a large production cluster point of view, 3 is still too high for everyday running in production. OCP 3.x ran at log level 2.
> 
> Scott Worthington did some sampling on the issue on 4.5.17 in a 3 masters / 9 workers node cluster:
> loglevel=2 for masters and workers: 14000 documents per 5 minutes in ES
> loglevel=3 for masters and workers: 42000 documents per 5 minutes in ES
> loglevel=4 for masters and workers: 126000 documents per 5 minutes in ES
> 
> There is also a big different in the level of detail / verbosity of each log line at level 4 versus level 2 which has additional storage implications.

I'm comfortable dropping down to 3 as that was the previous default in OCP 4.x. We can see if 2 makes sense to node team as we do need sufficient logging for troubleshooting clusters and dropping down to 2 may make it harder from a support perspective.

PR #2212, opened 2020-11-09
[WIP] Use same namespace name for kni-networking services on all platforms.

ravidbro:
Instead of always creating (on each platform) all the namespaces for irrelevant platforms and use only one of them.

Signed-off-by: Ravid Brown <ravid@redhat.com>


kikisdeliveryservice:
There's a very long discussion here re: namespaces: https://github.com/openshift/enhancements/pull/412

so this should be held until that is resolved

/hold

PR #2214, opened 2020-11-10
Testing fix for dualstack job

derekhiggins:
/hold

Testing dualstack job

### DIFF
--- a/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
@@ -20,7 +20,7 @@ contents:
     {{end -}}
 
     # If $DHCP6_FQDN_FQDN is not empty and is not localhost.localdomain
-    [[ -n "$DHCP6_FQDN_FQDN" && "$DHCP6_FQDN_FQDN" != "localhost.localdomain" && "$DHCP6_FQDN_FQDN" =~ "." ]] && hostnamectl set-hostname --static --transient $DHCP6_FQDN_FQDN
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
this touches some of the same files as https://github.com/openshift/machine-config-operator/pull/2207/files

So i'd encourage you to coordinate

derekhiggins:
> this touches some of the same files as https://github.com/openshift/machine-config-operator/pull/2207/files
> 
> So i'd encourage you to coordinate

Thanks, will do if I get the PR worth merging, at the moment its more of a test to find a problem.

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -20,7 +20,7 @@ contents:
     {{end -}}
 
     # If $DHCP6_FQDN_FQDN is not empty and is not localhost.localdomain
-    [[ -n "$DHCP6_FQDN_FQDN" && "$DHCP6_FQDN_FQDN" != "localhost.localdomain" && "$DHCP6_FQDN_FQDN" =~ "." ]] && hostnamectl set-hostname --static --transient $DHCP6_FQDN_FQDN
+    [[ "$DHCP6_FQDN_FQDN" =~ "." && ! "$DHCP6_FQDN_FQDN" =~ localhost ]] && hostnamectl set-hostname --static --transient "${DHCP6_FQDN_FQDN/.*}"
```
derekhiggins, line 23: This worked where I tested it locally, although I'm not sure its correct, it forces the setting on the hostname to the short version, but I'm starting to think we used to expect it to be the long version.

derekhiggins:
Fixed in packaging

PR #2210, opened 2020-11-09
Revert "Revert "crio: drop infra container when possible""

haircommander:
This reverts commit 4554713bef99dcfc80ef9fc48788f4042881083e.

Attempt to drop infra container again, now with a test that verifies we don't regress in the same way that
caused us to revert before

Signed-off-by: Peter Hunt <pehunt@redhat.com>

**- What I did**

**- How to verify it**

**- Description for the changelog**

PR #2215, opened 2020-11-10
  recycler pod: drop serviceAccountName

haircommander:
**- What I did**

this static pod fails on creation because static pods can't have an attached service account.
This would normally not be that problematic (the mirror pod creation fails in the API server, but continues in kubelet),
however, Multus queries the API-server for pod status, and fails if it can't be found.
This, this pod is never actually created.

**- How to verify it**
```
kubelet_service.log:Nov 10 03:07:32.757094 ip-10-0-143-173 hyperkube[1522]: E1110 03:07:32.757070    1522 kubelet.go:1583] Failed creating a mirror pod for "recyler-pod-ip-10-0-143-173.ec2.internal_openshift-infra(c4a6d0bdd7ae69ac5efa10a12ad33bc0)": pods "recyler-pod-ip-10-0-143-173.ec2.internal" is forbidden: a mirror pod may not reference service accounts
...
crio_service.log:Nov 10 03:07:32.798133 ip-10-0-143-173 crio[1487]: time="2020-11-10 03:07:32.798004563Z" level=error msg="Error adding network: Multus: [openshift-infra/recyler-pod-ip-10-0-143-173.ec2.internal]: error getting pod: pods \"recyler-pod-ip-10-0-143-173.ec2.internal\" not found" file="ocicni/ocicni.go:771"
```
does not appear

### DIFF
--- a/templates/master/00-master/_base/files/recycler_pod.yaml
+++ b/templates/master/00-master/_base/files/recycler_pod.yaml
@@ -10,7 +10,6 @@ contents:
     spec:
       activeDeadlineSeconds: 60
       restartPolicy: Never
-      serviceAccountName: pv-recycler-controller
       containers:
         -
           name: recycler-container
--- a/templates/master/00-master/_base/files/recycler_pod.yaml
+++ b/templates/master/00-master/_base/files/recycler_pod.yaml
@@ -5,7 +5,7 @@ contents:
     apiVersion: v1
     kind: Pod
     metadata:
-      name: recyler-pod
+      name: recycler-pod
       namespace: openshift-infra
     spec:
       activeDeadlineSeconds: 60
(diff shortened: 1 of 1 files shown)


haircommander:
@bertinatto PTAL

haircommander:
(I happened upon this by poking through the logs of https://github.com/openshift/machine-config-operator/pull/2210 trying to figure out why it failed, my understanding is incomplete but hopefully near-correct)

kikisdeliveryservice (approved):
This seems sane to me but I defer to @bertinatto to LGTM as they originally added this template.

bertinatto:
/lgtm

This makes sense to me. This template is a copy of the one used in 3.x, so I'm not sure why we haven't seen this before.

kikisdeliveryservice:
lots of retesting will remove hold when they start to pass..

/hold

sjenning:
I don't believe we need this.

This pod was not intended to be a static pod in the first place.  See https://github.com/openshift/machine-config-operator/pull/2238

@haircommander if you agree, can you close this?

sjenning:
@haircommander I also pulled in the typo fix for `recyler`

haircommander:
agreed that this can be closed!

PR #2220, opened 2020-11-12
daemon: Remove rollback deployment

cgwalters:
Came out of concerns from compliance about having bootloader
entries that may not have expected parameters.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -512,6 +512,16 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 	return dn.finalizeAndReboot(newConfig)
 }
 
+// removeRollback removes the rpm-ostree rollback deployment.  It
+// takes up space, and we don't generally expect administrators to
+// use this versus e.g. removing broken configuration.  We only
+// remove the rollback once the MCD pod has landed on a node, so
+// we know kubelet is working.
+func (dn *Daemon) removeRollback() error {
+	_, err := runGetOut("rpm-ostree", "cleanup", "-r")
+	return err
+}
+
 // machineConfigDiff represents an ad-hoc difference between two MachineConfig objects.
 // At some point this may change into holding just the files/units that changed
 // and the MCO would just operate on that.  For now we're just doing this to get
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1022,6 +1022,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		return fmt.Errorf("error detecting previous SSH accesses: %v", err)
 	}
 
+	if err := dn.removeRollback(); err != nil {
+		return errors.Wrapf(err, "Failed to remove rollback")
+	}
+
 	// Bootstrapping state is when we have the node annotations file
 	if state.bootstrapping {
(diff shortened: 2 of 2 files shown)


JAORMX (approved):

pkg/daemon/update.go
```diff
@@ -512,6 +512,16 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 	return dn.finalizeAndReboot(newConfig)
 }
 
+/// removeRollback removes the rpm-ostree rollback deployment.  It
+/// takes up space, and we don't generally expect administrators to
+/// use this versus e.g. removing broken configuration.  We only
+/// remove the rollback once the MCD pod has landed on a node, so
+/// we know kubelet is working.
```
JAORMX, line 519: oops, gotta replace `///` with `//`

JAORMX (changes requested):

sinnykumari:
I am not sure how relevant it is, but I find it useful to have previous deployment. When things goes wrong, it is quick to look at `rpm-ostree status` output from MCD log which is very easy to read the difference between booted and rollback deployment. Since this PR deletes the rollback deployment, I believe we won't see it now.

pkg/daemon/update.go
```diff
@@ -512,6 +512,16 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+func (dn *Daemon) removeRollback() error {
+	_, err := runGetOut("rpm-ostree", "cleanup", "-r")
```
sinnykumari, line 521: haven't checked myself, does rpm-ostree return success if there is no rollback deployment available to delete?

cgwalters: Yep, it's a no-op if not present.

cgwalters:
> I am not sure how relevant it is, but I find it useful to have previous deployment. When things goes wrong, it is quick to look at rpm-ostree status output from MCD log which is very easy to read the difference between booted and rollback deployment. Since this PR deletes the rollback deployment, I believe we won't see it now.

Agree with that concern!  I think it's addressed though because today we call `dn.LogSystemData()` very early on which will add the previous deployment information into the pod logs before we prune the rollback.

cgwalters:
Another reason to do this is that until we implement https://github.com/openshift/machine-config-operator/issues/1190 rolling back will make the MCD go degraded and require the force flag to fix, so it's rarely going to be useful except as a last ditch recovery on bare metal if kubelet fails to come up.

pkg/daemon/update.go
cgwalters: Context switching between Rust and Go is hard :smile:

sinnykumari:
> Agree with that concern! I think it's addressed though because today we call `dn.LogSystemData()` very early on which will add the previous deployment information into the pod logs before we prune the rollback.

just checked the code, you are right!

JAORMX (approved):

sinnykumari:
> Another reason to do this is that until we implement #1190 rolling back will make the MCD go degraded and require the force flag to fix, so it's rarely going to be useful except as a last ditch recovery on bare metal if kubelet fails to come up.

agree.

Before adding lgtm, will give some time to see if anyone in the team has came across case where user have used rpm-ostree rollback feature to recover a cluster node.

PR #2222, opened 2020-11-12
crio: add builder runtime class

haircommander:
**- What I did**
to allow unprivileged builds in openshift, we need to give pods a way to pass in /dev/fuse and allocate user user namespaces.
However, we don't want every pod to have that behavior.

We've created an option for admins to specify runtime classes that interpret pod annotations for each of these fields.
For the builder team to take advantage of it, we need to add it to the crio templates.

Note: if we omit the plain "runc" entry, then the runtimes map will only container the builder class. We need to add both to append the builder class

**- How to verify it**


haircommander:
cc @mrunalp @nalind

haircommander:
/hold

we should wait until 4.8 for this

PR #2223, opened 2020-11-12
MCS: Validate provisioning token

cgwalters:
Implements: https://github.com/openshift/enhancements/pull/443
Requires: https://github.com/openshift/installer/pull/4372

Basically we need to protect the Ignition config since it
contains secrets (pull secret, kubeconfig) and we can't rely
solely on network firewalling in all cases.

The installer will generate a secret into both the pointer
config and as a secret in our namespace, the MCS checks it.


cgwalters:
Note since the installer PR hasn't merged, this will effectively test the upgrade (no secret configured) path.

ashcrow:
/hold

Pending https://github.com/openshift/installer/pull/4372

PR #2227, opened 2020-11-13
Bug 1897361: ctrcfg_test: Wait for our prior target config

cgwalters:
Waiting for the previous MC isn't right here; we're deleting
a config and so what we want to wait for is the previous pool
target.

Drop the racy `sleep()`.

### DIFF
--- a/test/e2e/ctrcfg_test.go
+++ b/test/e2e/ctrcfg_test.go
@@ -84,7 +84,7 @@ func runTestWithCtrcfg(t *testing.T, testName, regexKey, expectedConfValue strin
 		err := cs.MachineConfigs().Delete(context.TODO(), oldMCConfig.Name, metav1.DeleteOptions{})
 		require.Nil(t, err, "machine config deletion failed")
 	})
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
Yesterday I was also confused about why they didn't just waitForPoolComplete like we do in MCD tests so this makes sense to me. Obvi would like test run to confirm it but yep. :+1: !

rphillips:
ci/prow/e2e-agnostic-upgrade -- gcp IAM error, unrelated

haircommander:
haha it would help to write tests for components you understand... LGTM, good find @cgwalters :smile:

kikisdeliveryservice:
testing another pr without second commit.

/hold

kikisdeliveryservice:
Closing in favor of https://github.com/openshift/machine-config-operator/pull/2229 which is the first commit by Colin and the minimal commit necessary to fix the ctrcfg test:

https://github.com/openshift/machine-config-operator/pull/2229

PR #2228, opened 2020-11-13
Bug 1896898: Fix awk syntax in configure-ovs.sh

danwinship:
The script did

    iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')

which is to say:

    iface=$(ip route show default | awk '{
        if ($4 == "dev")
            print $5;
        exit
    }')

ie, it was unconditionally exiting after the first line (which actually worked most of the time...)

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -34,13 +34,13 @@ contents:
       # find default interface
       while [ $counter -lt 12 ]; do
         # check ipv4
-        iface=$(ip route show default | awk '{if ($4 == "dev") print $5; exit}')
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
How...awkward.

danwinship:
can someone `/override ci/prow/e2e-gcp-op` since it seems mostly broken?

cgwalters:
Yes, I think we should use some strategic overrides to clear the queue.  This one looks like a good candidate; I can't see how it could break just our operator's tests - either ovs works or it doesn't really.
/override ci/prow/e2e-gcp-op

danwinship:
sigh. looks like e2e-aws-serial is mostly-broken too

PR #2231, opened 2020-11-16
operator: prevent upgrades on degraded pools

kikisdeliveryservice:
Finally using our Upgradeable operator status which is currently always at true..

One of the biggest problems and compounding issues we see are upgrades rolled out by users when there are already degraded pools. This makes troubleshooting more difficult, makes users think that a previous upgrade is finished an successful and compounds any existings bugs/problems.

This pr finally leverages the upgradeable status so that users can fix their degraded pools before upgrading (which is the easier time) and have the MCO's status be a better reflection of the pools that it manages.

So we end up with Upgradeable = False when pools are degraded and toggle back to Upgradeable = True when no pools are degraded.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -560,14 +560,6 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 	if err != nil {
 		return err
 	}
-	isPoolStatusConditionTrue := func(pool *mcfgv1.MachineConfigPool, conditionType mcfgv1.MachineConfigPoolConditionType) bool {
-		for _, condition := range pool.Status.Conditions {
-			if condition.Type == conditionType {
-				return condition.Status == corev1.ConditionTrue
-			}
-		}
-		return false
-	}
 
 	var lastErr error
 	if err := wait.Poll(time.Second, 10*time.Minute, func() (bool, error) {
@@ -842,4 +834,13 @@ func mergeCertWithCABundle(initialBundle, newBundle []byte, subject string) []by
 		initialBundle = next
 	}
 	return mergedBytes
+}
+
+func isPoolStatusConditionTrue(pool *mcfgv1.MachineConfigPool, conditionType mcfgv1.MachineConfigPoolConditionType) bool {
+	for _, condition := range pool.Status.Conditions {
+		if condition.Type == conditionType {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
WIP last commit (9d8ca3b) untested.  up to and incl commit (d8cf99c) gets upgradeable false on master pool
```
  - lastTransitionTime: "2020-11-17T02:51:01Z"
    reason: One or more machine config pool is degraded, please see `oc get mcp` for
      further details and resolve before upgrading
    status: "False"
    type: Upgradeable
  extension:
    master: 'pool is degraded because rendering fails with "": "Failed to render configuration
      for pool master: parsing Ignition config failed: unknown version. Supported
      spec versions: 2.2, 3.0, 3.1"'
    worker: all 3 nodes are at latest configuration rendered-worker-e28c7ac562ab60d143ec4afbde4e8393
```

```
  - lastTransitionTime: "2020-11-17T02:52:48Z"
    reason: AsExpected
    status: "True"
    type: Upgradeable
  extension:
    master: all 3 nodes are at latest configuration rendered-master-77519c04d1536ccb2cb303acb5a7106c
    worker: all 3 nodes are at latest configuration rendered-worker-e28c7ac562ab60d143ec4afbde4e8393
```

pkg/operator/sync.go
```diff
@@ -587,28 +577,46 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
...
+
+				if pool.Generation <= pool.Status.ObservedGeneration &&
+					isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated) {
+					continue
+				}
 			}
-			lastErr = fmt.Errorf("error pool %s is not ready, retrying. Status: (pool degraded: %v total: %d, ready %d, updated: %d, unavailable: %d)", pool.Name, degraded, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount)
-			return false, nil
+		}
+		syncstatuserr := optr.syncUpgradeableStatus()
```
runcom, line 612: is this extra as we're syncing upgradeable in the loop above? (maybe my eyes are tricking me lol)

runcom:
:100: @sinnykumari @yuqi-zhang ptal
/approve

pkg/operator/sync.go
kikisdeliveryservice: no we're only syncing if it's degraded above (it syncs then returns false), this needs to be here to pick up with it's not degraded, without it here it won't toggle back when it fixed.

runcom: won't it? :thinking: this function returns to `syncAll` which in turns calls `syncUpgradeableStatus` again right?

runcom:
just a not harmful comment https://github.com/openshift/machine-config-operator/pull/2231#discussion_r530261697 which we can drop if that's the case later on

/lgtm

pkg/operator/sync.go
runcom: postponed :) not harmful

PR #2232, opened 2020-11-17
templates: add apiserver-url.env file

squeed:
A few components need to know the URL of the internal apiserver load balancer. Most notably, the network operator does.

Right now, the only way to get this is to parse the kubelet's kubeconfig, but this is fragile. So, write down an environment file that just sets the "usual" service environment variables. Then we can just seamlessly source the environment file.

Ref: [SDN-1347](https://issues.redhat.com/browse/SDN-1347)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -1,0 +1,15 @@
+package constants
+
+// constants defines some file paths that are shared outside of the
+// MCO package; and thus consumed by other users
+
+const (
+	// APIServerURLFile is the path to the apiserver url environment file.
+	// See templates/master/00-master/_base/files/apiserver-url-env.yaml
+	APIServerURLFile = "/etc/kubernetes/apiserver-url.env"
+)
+
+// ConstantsByName is a map of constants for ease of templating
+var ConstantsByName = map[string]string{
+	"APIServerURLFile": APIServerURLFile,
+}
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
@@ -111,6 +111,8 @@ func newControllerConfig(name string, platform apicfgv1.PlatformType) *mcfgv1.Co
 					PlatformStatus: &apicfgv1.PlatformStatus{
 						Type: platform,
 					},
+					APIServerURL:         fmt.Sprintf("https://api.%s.tt.testing:6443", name),
+					APIServerInternalURL: fmt.Sprintf("https://api-int.%s.tt.testing:6443", name),
 				},
 			},
 			ReleaseImage: "release-reg.io/myuser/myimage:test",
--- a/pkg/controller/kubelet-config/kubelet_config_controller_test.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller_test.go
@@ -106,6 +106,8 @@ func newControllerConfig(name string, platform osev1.PlatformType) *mcfgv1.Contr
 					PlatformStatus: &osev1.PlatformStatus{
 						Type: platform,
 					},
+					APIServerURL:         fmt.Sprintf("https://api.%s.tt.testing:6443", name),
+					APIServerInternalURL: fmt.Sprintf("https://api-int.%s.tt.testing:6443", name),
 				},
 			},
 		},
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -16,6 +17,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/constants"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/version"
 )
@@ -24,6 +26,9 @@ import (
 type RenderConfig struct {
 	*mcfgv1.ControllerConfigSpec
 	PullSecret string
+
+	// no need to set this, will be automatically configured
+	Constants map[string]string
 }
 
 const (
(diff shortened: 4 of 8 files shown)


squeed:
Nice, CI seems basically green.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -111,6 +111,8 @@ func newControllerConfig(name string, platform apicfgv1.PlatformType) *mcfgv1.Co
 					PlatformStatus: &apicfgv1.PlatformStatus{
 						Type: platform,
 					},
+					APIServerURL:         fmt.Sprintf("https://api.%s.tt.testing:6443", name),
```
danielmellado, line 114: How about also allowing to change the port using a var? I mean, it'll probably never change but just in case.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -111,6 +111,8 @@ func newControllerConfig(name string, platform apicfgv1.PlatformType) *mcfgv1.Co
...
+					APIServerInternalURL: fmt.Sprintf("https://api-int.%s.tt.testing:6443", name),
```
danielmellado, line 115: same there

pkg/controller/kubelet-config/kubelet_config_controller_test.go
```diff
@@ -108,6 +108,8 @@ func newControllerConfig(name string, platform osev1.PlatformType) *mcfgv1.Contr
 					PlatformStatus: &osev1.PlatformStatus{
 						Type: platform,
 					},
+					APIServerURL:         fmt.Sprintf("https://api.%s.tt.testing:6443", name),
```
danielmellado, line 111: ditto

pkg/controller/kubelet-config/kubelet_config_controller_test.go
```diff
@@ -108,6 +108,8 @@ func newControllerConfig(name string, platform osev1.PlatformType) *mcfgv1.Contr
...
+					APIServerInternalURL: fmt.Sprintf("https://api-int.%s.tt.testing:6443", name),
```
danielmellado, line 112: ditto

pkg/controller/template/template_controller_test.go
```diff
@@ -67,6 +67,8 @@ func newControllerConfig(name string) *mcfgv1.ControllerConfig {
 					PlatformStatus: &configv1.PlatformStatus{
 						Type: configv1.LibvirtPlatformType,
 					},
+					APIServerURL:         fmt.Sprintf("https://api.%s.tt.testing:6443", name),
```
danielmellado, line 70: ditto

pkg/controller/template/template_controller_test.go
```diff
@@ -67,6 +67,8 @@ func newControllerConfig(name string) *mcfgv1.ControllerConfig {
...
+					APIServerInternalURL: fmt.Sprintf("https://api-int.%s.tt.testing:6443", name),
```
danielmellado, line 71: ditto

pkg/controller/template/render.go
```diff
@@ -462,3 +465,37 @@ func onPremPlatform(platformString configv1.PlatformType) bool {
...
+		return port, nil
+	}
+
+	// default port
+	switch parsed.Scheme {
+	case "https":
+		return "443", nil
+	case "http":
+		return "80", nil
+	default:
```
danielmellado, line 498: What would happen if this were to fall to any non-standard port? Is there a way to dynamically get this?

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
squeed: this is just test data; it doesn't matter in this case.

pkg/controller/template/render.go
squeed: yeah, the above line handles this case (since 6443 is, of course, non-standard)

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
danielmellado: oh, fair enough for the tests!

pkg/controller/template/render.go
danielmellado: so it would return unknown scheme for 6443? am I missing something around here?

squeed: you've got it backwards - what this code does is ensure `https://foo.com` returns port 443, but `https://foo.com:1234` returns 1234.

templates/master/00-master/_base/files/apiserver-url-env.yaml
```diff
@@ -0,0 +1,7 @@
+mode: 0644
+path: "/etc/kubernetes/apiserver-url.env"
+# used by the cluster network operator to learn the URL to the internal apiserver load balancer
```
cgwalters, line 3: For things like the binaries extracted to the host?  Or code running in a container?  Both?

templates/master/00-master/_base/files/apiserver-url-env.yaml
```diff
@@ -0,0 +1,7 @@
+mode: 0644
+path: "/etc/kubernetes/apiserver-url.env"
```
cgwalters, line 2: Maybe we should add something like a "MCO internals" package that things like the SDN could vendor in that would have this as a constant.

cgwalters (approved):

templates/master/00-master/_base/files/apiserver-url-env.yaml
squeed: oh, that's a decent idea - as long as it won't annoy `go mod`.

templates/master/00-master/_base/files/apiserver-url-env.yaml
squeed: The motivating use case is the CNO, which is in a container (and thus bind-mounts it), but I suspect we'll find a few more uses for it.

templates/master/00-master/_base/files/apiserver-url-env.yaml
squeed: That said, this is certainly not changeable, so parameterizing it isn't really useful.

Up to you.

squeed:
@cgwalters I updated to add a `pkg/constants` reference. Let me know what you think. I'm not super happy with the naming.

pkg/constants/constants.go
```diff
@@ -0,0 +1,15 @@
+package constants
+
+// constants defines some file paths that are shared outside of the
+// MCO package; and thus consumed by other users
```
cgwalters, line 4: Thanks!  This should help us in the future in doing things like searching across the OpenShift repos to see which other projects are using bits from the MCO.

cgwalters (approved):
Thanks!

squeed:
@cgwalters could I get a lgtm again?

PR #2234, opened 2020-11-17
[on-prem] Move MDNS config file to common directory

yboaron:
Since etcd does not rely on DNS entries anymore,
there shouldn't be any difference between masters and workers MDNS configuration.


mandre (commented):
/lgtm

PR #2237, opened 2020-11-18
Bug 1899057: configure-ovs-network: fix spurious OVS warnings

danwinship:
We see spurious errors in ovs-vswitchd.log when using OVN:

    2020-11-17T12:20:50.895Z|00033|bridge|ERR|interface br-ex: ignoring mac in Interface record (use Bridge record to set local port's mac)

This is caused by how we're creating `br-ex`; we specify the (same) MAC on the bridge record and the local port record but OVS doesn't want you do specify it on the latter (since the port isn't allowed to have a different MAC from the bridge itself anyway).

/assign @trozet

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -174,11 +174,7 @@ contents:
           else
             sed -i '/^\[connection\]$/a interface-name=br-ex' ${new_conn_file}
           fi
-          if ! grep 'cloned-mac-address=' ${new_conn_file} &> /dev/null; then
-            sed -i '/^\[ethernet\]$/,/^\[/ s/^cloned-mac-address=.*$/cloned-mac-address='"$iface_mac"'/' ${new_conn_file}
-          else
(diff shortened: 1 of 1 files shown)


cybertron (approved):

trozet:
@danwinship this is present due to a nuance in NetworkManager-OVS:
https://bugzilla.redhat.com/show_bug.cgi?id=1853750#c3

tl;dr NM-OVS will use the mac from the interface for dhclient and not the bridge. We probably need to loop back with the NM team and possibly reopen/retest this bug.

danwinship:
Hm. OK, filed https://bugzilla.redhat.com/show_bug.cgi?id=1899745 against NetworkManager. Let's see if that's likely to get fixed quickly and then we can decide whether to leave this open or not

kikisdeliveryservice:
> Hm. OK, filed https://bugzilla.redhat.com/show_bug.cgi?id=1899745 against NetworkManager. Let's see if that's likely to get fixed quickly and then we can decide whether to leave this open or not

Any update on this?

danwinship:
/close
still need to figure out with NM team what the right fix is

PR #1687, opened 2020-04-27
Bug 1805908: Add template for recycler pod

bertinatto:
**- What I did**

Added template for recycler pod.

**- How to verify it**

This should create a new file in `/etc/kubernetes/manifests/recycler-pod.yaml`.

CC @openshift/storage

PR #2238, opened 2020-11-18
templates: move recycler pod out of kubelet manifests directory

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1896226

This is the first step in resolving a bug introduced by https://github.com/openshift/machine-config-operator/pull/1687 and https://github.com/openshift/cluster-kube-controller-manager-operator/pull/405

`/etc/kubernetes/manifests` is the kubelet's static pod manifests directory.  The kubelet will try to start pods defined in this location as static pods.  The recycler pod is not a static pod and the kubelet continuously fails to start it.

The solution is to move the recycler pod manifests to a different directory.

The fix is a 3-step process (as I see it):
- Change location of the recycler pod in the MCO template (this PR)
- Change KCM to use the new location
- Delete (or empty out) the file in the old location

@runcom @sttts @rphillips

### DIFF
--- a/templates/master/00-master/_base/files/recycler_pod_bad.yaml
+++ b/templates/master/00-master/_base/files/recycler_pod_bad.yaml
@@ -1,5 +1,5 @@
 mode: 0644
-path: "/etc/kubernetes/recycler-pod.yaml"
+path: "/etc/kubernetes/manifests/recycler-pod.yaml"
 contents:
   inline: |
     apiVersion: v1


sjenning:
also @haircommander

sjenning:
Indeed the old file still needs to be there
```
F1118 21:07:22.756911       1 plugins.go:123] Could not create hostpath recycler pod from file /etc/kubernetes/manifests/recycler-pod.yaml: failed to read file path /etc/kubernetes/manifests/recycler-pod.yaml: open /etc/kubernetes/manifests/recycler-pod.yaml: no such file or directory
```

JAORMX (approved):
/retest

cgwalters:
OK I see, and it's then configured by the other pod.
/approve
/lgtm

cgwalters:
It'd be good I think to change https://github.com/openshift/cluster-kube-controller-manager-operator to detect the path dynamically - that would avoid a version interlock between the these two codebases and make backporting easier.

Issue #1592, opened 2020-03-26
move rhel7 worker node updates entirely to openshift-ansible


cgwalters:
It's really confusing today the role and split between [openshift-ansible](https://github.com/openshift/openshift-ansible/) and the MCO (we need a FAQ for this).

Anyways here's an idea: what if we stopped the MCD running on !CoreOS systems, and instead openshift-ansible always needed to be run to apply changes from MachineConfig.  It'd be openshift-ansible that'd somehow trigger drain/reboot etc.

The downside of this is e.g. if a user wants to change the kubelet config, they `oc edit` but then need to bounce out of Kube down to Ansible+ssh.

But the massive benefit is that it's super clear how things work - the MCO can optimize for CoreOS, openshift-ansible optimizes fully for traditional.  

And with PRs like https://github.com/openshift/machine-config-operator/pull/1586/ for example - we can *rely* on `/usr/etc` for original files, and do things like [make better use of OSTree](https://github.com/openshift/machine-config-operator/issues/119).

jlebon:
A random idea I discussed with @runcom which is less radical but more cumbersome is we detect the case where the cluster is only RHCOS nodes and activating "OSTree-optimized" mode where we can use things like `/usr/etc`. But yeah, not fun to maintain multiple paths to do the same thing.

kikisdeliveryservice:
I like the clear "separation of powers" as I think it would allow us to optimize and maintain a tight and consistent focus - would also seem to help from a troubleshooting (fixing/docs/ux) perspective going forward.

How fussy this would be to implement: ¯\_(ツ)_/¯¯

sdodson:
> Anyways here's an idea: what if we stopped the MCD running on !CoreOS systems, and instead openshift-ansible always needed to be run to apply changes from MachineConfig. It'd be openshift-ansible that'd somehow trigger drain/reboot etc.

Do you specifically mean running MCD as a daemon or would even the `once-from` mode be eliminated outside of CoreOS?

/cc @crawford

cgwalters:
> Do you specifically mean running MCD as a daemon or would even the once-from mode be eliminated outside of CoreOS?

Just the MCD as daemon.

cgwalters:
One thing I could imagine is that openshift-ansible owns a "translator" from MachineConfig into Ansible...e.g. if `kernelArguments` in a MC are set that could be a playbook that calls out to `grubby`.

cgwalters:
[...]

PR #2241, opened 2020-11-19
Bug 1899603: daemon: do not roll back on non-*COS machines

yuqi-zhang:
This breaks RHEL7 jobs as they do not have rpm-ostree.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -518,6 +518,10 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 // remove the rollback once the MCD pod has landed on a node, so
 // we know kubelet is working.
 func (dn *Daemon) removeRollback() error {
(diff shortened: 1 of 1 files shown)


sinnykumari:
lgtm, let's wait for aws-workers-rhel7 to go green

cgwalters (approved):
We really want to do https://github.com/openshift/machine-config-operator/issues/1592 ...

mtnbikenc:
Scaleup succeeded, which is where it was stuck before.  Hopefully e2e tests pass as well.
`Pod e2e-aws-workers-rhel7-workers-rhel7-scaleup succeeded after 13m28s`

mtnbikenc:
203 e2e tests failed on workers-rhel7, which is not normal.  However, I also don't think that is related to this PR.  Since it passed the scaleup portion I'm fine with this being merged to unblock other queues.

kikisdeliveryservice:
> 203 e2e tests failed on workers-rhel7, which is not normal. However, I also don't think that is related to this PR. Since it passed the scaleup portion I'm fine with this being merged to unblock other queues.

@mtnbikenc can you do the final lgtm then

/assign @mtnbikenc

PR #2242, opened 2020-11-19
 daemon: Rework operating system conditionals

cgwalters:
A whole lot of our conditionals were "is coreos or not".  Let's rework
things so that we explicitly use that.

Detecting based on the operating system ID is also generally wrong;
e.g. we had an "is this traditional rhel or centos" check that really wanted to be
"is this a rhel7-like system" because both "traditional" RHEL8 and CentOS8 exist.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -54,8 +54,8 @@ type Daemon struct {
 	// name is the node name.
 	name string
 
-	// OperatingSystem the operating system the MCD is running on
-	OperatingSystem string
+	// os the operating system the MCD is running on
+	os OperatingSystem
 
 	// IgnitionVersion is the version of the installed Ignition binary on the system
 	IgnitionVersion string
@@ -208,9 +208,9 @@ func New(
 		err        error
 	)
 
-	operatingSystem := "mock"
+	os := OperatingSystem{}
 	if !mock {
-		operatingSystem, err = GetHostRunningOS()
+		os, err = GetHostRunningOS()
 		if err != nil {
 			HostOS.WithLabelValues("unsupported", "").Set(1)
 			return nil, errors.Wrapf(err, "checking operating system")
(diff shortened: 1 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -1367,7 +1382,7 @@ func compareOSImageURL(current, desired string) bool {
 // Otherwise if `false` is returned, then we need to perform an update.
 func (dn *Daemon) checkOS(osImageURL string) bool {
 	// Nothing to do if we're not on RHCOS or FCOS
-	if dn.OperatingSystem != MachineConfigDaemonOSRHCOS && dn.OperatingSystem != MachineConfigDaemonOSFCOS {
+	if !dn.isOstree {
```
cgwalters, line 1385: Avoiding the need to mentally apply a [DeMorgan rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) when reading this code is IMO a clear win.

pkg/daemon/daemon.go
```diff
@@ -208,6 +212,16 @@ func New(
 		err        error
 	)
 
+	isOstree := false
+	_, err = os.Stat(pathOstreeBooted)
```
kikisdeliveryservice, line 216: I keep reading this a Pathos Tree :laughing:

kikisdeliveryservice:
> This way if someone creates a CentOS CoreOS most of this code would
> just work

Just as a clarifying : this isn't Supported, it's just a nice fringe effect?

cgwalters:
/retest
Hmm I don't think the failures are related offhand, though not sure how to debug that metal-ipi failure.

cgwalters:
> Just as a clarifying : this isn't Supported, it's just a nice fringe effect?

Right.

cgwalters:
Hmm in that e2e-aws failure a single node was in `NotReady`, similar failure I saw in the DNS PR.  We need to track this down.

cgwalters:
OK cool, got passing CI on this one!

pkg/daemon/daemon.go
```diff
@@ -56,6 +56,8 @@ type Daemon struct {
 
 	// OperatingSystem the operating system the MCD is running on
 	OperatingSystem string
+	// isOstree is true if OperatingSystem uses rpm-ostree
```
crawford, line 59: Rather than adding a new field, why not just give `OperatingSystem` its own type with an `isOstree` method? Written as is, there is the possibility to encode an invalid state into this object.

cgwalters: Good suggestion, thanks!  I did a larger rework along those lines.

sinnykumari:
This PR makes OS type handling much cleaner.
/approve
LGTM

sinnykumari:
This is good stuff, getting this in!
/lgtm

PR #2243, opened 2020-11-19
Bug 1899600: daemon: Only switch to bfq scheduler when we have an OS update

cgwalters:
Using bfq by default reportedly causes etcd latency regressions
when we're not doing an OS update.  Change the logic so we
only do the switch when we want to balance update speed while running
etcd.

(I still think schedulable masters may want this but they can
 configure it manually with a systemd unit)

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -327,6 +327,12 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 
 	var osImageContentDir string
 	if mcDiff.osUpdate || mcDiff.extensions || mcDiff.kernelType {
+		// When we're going to apply an OS update, switch the block
+		// scheduler to BFQ to apply more fairness between etcd
+		// and the OS update.
+		if err := setRootDeviceSchedulerBFQ(); err != nil {
+			return err
+		}
 		// We emitted this event before, so keep it
 		if dn.recorder != nil {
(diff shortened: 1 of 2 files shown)


hexfusion:
This makes a lot of sense, thanks @cgwalters

hexfusion:
/rename Bug1899600: daemon: Only switch to bfq scheduler when we have an OS update

hexfusion:
/retitle Bug 1899600: daemon: Only switch to bfq scheduler when we have an OS update

pkg/daemon/update.go
```diff
@@ -327,6 +327,12 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 
 	var osImageContentDir string
 	if mcDiff.osUpdate || mcDiff.extensions || mcDiff.kernelType {
+		// When we're going to apply an OS update, switch the block
+		// scheduler to BFQ to apply more fairness between etcd
+		// and the OS update.
+		if err := setRootDeviceSchedulerBFQ(); err != nil {
```
ironcladlou, line 333: What restores the original scheduler implementation following an update? A byproduct of restarting the node?

ironcladlou:
@cgwalters who would you like to review and tag this for merging?

/lgtm

pkg/daemon/update.go
cgwalters: Yep.

cgwalters:
I belatedly realized that this lost the "only on control plane" logic.  Which...well, I guess it doesn't matter very much probably.  It's likely we'll rework this as part of something like https://github.com/openshift/machine-config-operator/pull/1491#issuecomment-737252150

PR #2248, opened 2020-11-23
*: switch to v3.2.0 Ignition configs instead of v3.1.0

arithx:
**- What I did**

Switch the underlying Ignition spec 3 configs to use the `3.2.0` types (and update the underlying translation for spec 3 configs to uptranslate into 3.2.0).

**- How to verify it**

Spec 3 configs should remain working. Additionally new config fields that are only present in the Ignition 3.2.0 spec ([see here for a list of new features to the 3.2.0 spec](https://github.com/coreos/ignition/blob/master/docs/migrating-configs.md#from-version-310-to-320)) can be specified.

vendor commit: Bump `ign-converter` to pick up `v32tov2x` translation packages & pull in spec 3.2.0 Ignition packages
other commit: update the `ign3types` import to use the `config/v3_2/types` package instead of `config/v3_1/types`


pkg/server/api.go
```diff
@@ -283,8 +283,8 @@ func detectSpecVersionFromAcceptHeader(acceptHeader string) (*semver.Version, er
 
 	for _, header := range headers {
 		if header.MIMESubtype == "vnd.coreos.ignition+json" && header.SemVer != nil {
-			if !header.SemVer.LessThan(*v3_1) && header.SemVer.LessThan(*semver.New("4.0.0")) {
-				return v3_1, nil
+			if !header.SemVer.LessThan(*v3_2) && header.SemVer.LessThan(*semver.New("4.0.0")) {
```
cgwalters, line 286: In both the API and the tests we need to also continue to handle 3.1 right?

cgwalters (commented):
Mainly concerned about the MCS parts.

If every spec bump involves this level of churn that's a bit unfortunate but, solving that seems hard without reworking the API.

pkg/server/api.go
arithx: Ah yep, that was an oversight. I've got https://github.com/coreos/ign-converter/pull/19 up to add `v32tov31` down-translation. After that merges I'll bump the `ign-converter` vendor again & update to check for `v3_1` semver, down-translate in `server/api` & add tests for `3.1.0` in `server/api_test`.

arithx: Ok, I think I've addressed this.

I've added back support for `v3_1` headers, added `3.2.0` ->` 3.1.0` translation, and re-added unit tests in `pkg/server/api_test`.

cgwalters:
Hmm not sure what went wrong with that e2e-aws run; installer ran fine but then this failed which seems like something in CI glue:
> 2020/11/23 23:11:26 Failed to upload $KUBECONFIG: timed out waiting for the condition

kikisdeliveryservice:
throwing a hold while i review

/hold

kikisdeliveryservice:
Also okd-e2e is currently red and it looks like rhel7 is flaky, so you don't have to retest @arithx ill confirm we can skip the rhel7 one and deal with that later.

darkmuggle:
@kikisdeliveryservice in light of your hold, have you had a chance to review?

The update to spec is needed for finalizing complex disks (confirmed by @bgilbert via Slack)

I looked over the code, and nothing stands out. 
/approve

kikisdeliveryservice:
I believe @yuqi-zhang is looking it over too, wanted to give MCO team the opp to review this as it's non-trivial.

ashcrow:
@yuqi-zhang PTAL

yuqi-zhang:
Did some testing for this, upgraded fine, and all generated configs are on 3.2, applied 2.2 config looks fine, scaling node looks fine

/lgtm

yuqi-zhang:
Will let @kikisdeliveryservice cancel the hold when comfortable

pkg/controller/common/helpers.go
```diff
@@ -210,16 +249,16 @@ func convertIgnition2to3(ign2config ign2types.Config) (ign3types.Config, error)
 	if err != nil {
 		return ign3types.Config{}, errors.Errorf("unable to convert Ignition spec v2 config to v3: %v", err)
 	}
-	// Workaround to get a v3.1 config as output
-	converted3 := translate3.Translate(ign3_0config)
+	// Workaround to get a v3.2 config as output
+	converted3 := translate3.Translate(translate3_1.Translate(ign3_0config))
```
yuqi-zhang, line 253: Actually, this would translate to 3.1 still. Did you want it to translate to 3.2?

kikisdeliveryservice:
will lift hold once jerry's q above is addressed.

pkg/controller/common/helpers.go
kikisdeliveryservice: i read this as  3.0->3.1->3.2? am i wrong?

yuqi-zhang: Oh whoops I misread completely, ignore me. LGTM!

kikisdeliveryservice:
Q as a consumer of ignition : are you planning on chaining parsers? bc as you can see moving forward with future versions this is going to cause a mess...

bgilbert:
@kikisdeliveryservice We chained parsers in spec 2.x but removed that functionality for spec 3.  I agree that we should think about adding it back.

kikisdeliveryservice:
this is fully approved and made the deadline, e2e is still hitting some well documented ci resource issues:
```
   * could not run steps: step e2e-agnostic-upgrade failed: failed to acquire lease: resources not found
time="2020-12-03T17:10:30Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:step_failed:utilizing_lease'" 
```

which are still being worked on by admins...

PR #2252, opened 2020-11-23
WIP: Bug 1898169: pkg/controller/template: handle >=2048 units line size for noProxy

runcom:
The noProxy value coming from proxy/cluster can be lenghty and it
happened that it resulted in more than 2048 bytes. The value itself is
comma separated list of cidrs/domains/ips for noProxy. This list can
indeed grow large. When we format that value to a systemd unit or dropin
we're gonna incur into a known systemd line limit (LINE_MAX from glibc
== 2048). The error we're getting on the cluster is:

      failed to syncing towards (13) generation using controller version v4.5.0-202010160047.p0-dirty: failed to create MachineConfig for role master: error transpiling ct config to Ignition config: failed to convert config to Ignition config error: invalid unit content: line too long (max 2048 bytes)

So we're not even at the systemd level but the config transpiler has a
built-in check that fails with the above.

Now, we know splitting the line with backslashes is a workaround, and
that's what this PR does which will get rid (effectively) of the limit.

How to test:

- spin up a cluster with proxy
- `oc edit proxy/cluster`
- under `spec.noProxy` add a bunch of (even dummy) domains/ips to reach >= 2048 bytes
- the configuration resulting from the noProxy additions should roll out just fine (it fails today, see bz)

### DIFF
--- a/templates/common/_base/files/proxy.yaml
+++ b/templates/common/_base/files/proxy.yaml
@@ -1,0 +1,15 @@
+mode: 0600
+path: "/somewhere/proxy"
+contents:
+  inline: |
+      {{if .Proxy -}}
+      {{if .Proxy.HTTPProxy -}}
+      HTTP_PROXY={{.Proxy.HTTPProxy}}
+      {{end -}}
+      {{if .Proxy.HTTPSProxy -}}
+      HTTPS_PROXY={{.Proxy.HTTPSProxy}}
+      {{end -}}
+      {{if .Proxy.NoProxy -}}
+      NO_PROXY={{.Proxy.NoProxy}}
+      {{end -}}
+      {{end -}}
--- a/templates/common/_base/units/crio.service.yaml
+++ b/templates/common/_base/units/crio.service.yaml
@@ -3,15 +3,5 @@ dropins:
   - name: 10-mco-default-env.conf
     contents: |
       [Unit]
-      {{if .Proxy -}}
       [Service]
-      {{if .Proxy.HTTPProxy -}}
-      Environment=HTTP_PROXY={{.Proxy.HTTPProxy}}
-      {{end -}}
-      {{if .Proxy.HTTPSProxy -}}
-      Environment=HTTPS_PROXY={{.Proxy.HTTPSProxy}}
-      {{end -}}
-      {{if .Proxy.NoProxy -}}
-      Environment=NO_PROXY={{.Proxy.NoProxy}}
-      {{end -}}
-      {{end -}}
+      EnvironmentFile=/somewhere/proxy
--- a/templates/common/_base/units/kubelet.service.yaml
+++ b/templates/common/_base/units/kubelet.service.yaml
@@ -3,15 +3,6 @@ dropins:
   - name: 10-mco-default-env.conf
     contents: |
       [Unit]
-      {{if .Proxy -}}
       [Service]
-      {{if .Proxy.HTTPProxy -}}
-      Environment=HTTP_PROXY={{.Proxy.HTTPProxy}}
-      {{end -}}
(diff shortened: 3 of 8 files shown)


LorbusChris:
Couldn't this happen with too-long `HTTPProxy` and `HTTPSProxy` as well? (AFAICT they're also saved to env vars by the MCD and then used in systemd units).

runcom:
> Couldn't this happen with too-long `HTTPProxy` and `HTTPSProxy` as well? (AFAICT they're also saved to env vars by the MCD and then used in systemd units).

there's only one/two endpoints for http/https afaict, the noproxy value is the only one that can grow because it excludes the things that'll go through the proxy (but somebody with more expertise can shed a light there)

pkg/controller/template/render.go
```diff
@@ -295,6 +307,7 @@ func renderTemplate(config RenderConfig, path string, b []byte) ([]byte, error)
 	funcs["onPremPlatformIngressIP"] = onPremPlatformIngressIP
 	funcs["onPremPlatformShortName"] = onPremPlatformShortName
 	funcs["onPremPlatformKeepalivedEnableUnicast"] = onPremPlatformKeepalivedEnableUnicast
+	funcs["noProxySplit"] = noProxySplit // TODO: use this to properly format noProxy values in templates
```
runcom, line 310: this is just for testing it out - ideally, now that we know that trying to render a unit/dropin with more than 2048 bytes per line results in a failure, we should put something in place that takes care of not doing that while rendering

runcom:
Good to go now - there's an obvious "let's do this better"-card here https://issues.redhat.com/browse/GRPA-3091 but the fix here is simple enough to get in and fix the issue for the proxy while investigating other (numerous) cases

pkg/controller/template/render.go
runcom: after investigating this resulted in https://issues.redhat.com/browse/GRPA-3091

pkg/operator/render.go
```diff
@@ -123,6 +123,9 @@ func createDiscoveredControllerConfigSpec(infra *configv1.Infrastructure, networ
 			glog.V(2).Info("Not setting proxy config because Proxy status is empty")
 		} else {
 			ccSpec.Proxy = &proxy.Status
+			// Not doing this will result into LINE_MAX limit errors from the ignition transpiler
+			// see https://bugzilla.redhat.com/show_bug.cgi?id=1898169
+			ccSpec.Proxy.NoProxy = fmt.Sprintf("%s", strings.Join(strings.Split(ccSpec.Proxy.NoProxy, ","), ",\\\n"))
```
yuqi-zhang, line 128: To be safe here perhaps we should add a nil check for NoProxy?

yuqi-zhang (approved):
One minor question

pkg/operator/render.go
runcom: sorry, changed my mind and re-pushed what I think it's a better fix (noProxy is a string tho :thinking: )

runcom:
WIP again, I may have a found a better fix to strings splitting that I'd like to test :grimacing:

templates/common/_base/files/proxy.yaml
```diff
@@ -0,0 +1,15 @@
+mode: 0600
+path: "/somewhere/proxy"
```
runcom, line 2: this isn't obviously working (yet) @cgwalters any good location to drop this file at? :thinking:

LorbusChris: If you were to put it at `/etc/environment` you might not even have to explicitly specify `EnvironmentFile` on the units.

Maybe it's better to be explicit here though, and reference it clearly visibly, though.

PR #2254, opened 2020-11-24
daemon: skip or perform node reboot based on rebootAction

sinnykumari:
Today MCO defaults to rebooting node in order to apply successfully
changes to the node. This is the safest way to apply a change on the
node as we don't have to worry about things like which change is safe
to skip reboot.

In certain environments, rebooting is expensive and due to complex
hardware setup sometimes can cause problems and node won't boot up.

This will help to avoid rebooting nodes for certain cases where
MCO thinks it is safe to do so.

See - https://github.com/openshift/enhancements/pull/159

**Note** In the current state, this will default to rebooting node. When MCO learns what rebootAction to be taken based on old and new rendered MC diff, we can add the logic.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -165,6 +165,14 @@ const (
 	onceFromRemoteConfig
 )
 
+type rebootAction int
+
+const (
+	rebootActionReboot rebootAction = iota
+	rebootActionNone
+	rebootActionReloadCrio
+)
+
 var (
 	defaultRebootTimeout = 24 * time.Hour
 )
@@ -1015,7 +1023,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		if err := dn.drain(); err != nil {
 			return err
 		}
-		return dn.finalizeAndReboot(state.pendingConfig)
+		if err := dn.finalizeBeforeReboot(state.pendingConfig); err != nil {
+			return err
+		}
+		return dn.reboot(fmt.Sprintf("Node will reboot into config %v", state.pendingConfig.GetName()))
 	}
 
 	if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
@@ -1043,7 +1054,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 			if err := os.RemoveAll(osImageContentDir); err != nil {
 				return err
 			}
-			return dn.finalizeAndReboot(state.currentConfig)
+			if err := dn.finalizeBeforeReboot(state.currentConfig); err != nil {
+				return err
+			}
+			return dn.reboot(fmt.Sprintf("Node will reboot into config %v", state.currentConfig.GetName()))
 		}
 		glog.Info("No bootstrap pivot required; unlinking bootstrap node annotations")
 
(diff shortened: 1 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -93,9 +93,56 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+// In the end uncordon node to schedule workload.
+// If at any point an error occurs, we reboot the node so that node has correct configuration.
+func (dn *Daemon) performRebootAction(action rebootAction, configName string) error {
+	switch action {
+	case rebootActionNone:
+		dn.logSystem("Node has Desired Config %s, skipping reboot", configName)
+	case rebootActionReloadCrio:
+		if err := reloadCrioConfig(); err != nil {
+			dn.logSystem("Reloading crio configuration failed, rebooting: %v", err)
+			dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
```
yuqi-zhang, line 112: One thing I'm debating is whether we want to hard-reboot here or block + degrade. It seems to me that if we are explicitly supporting "rebootless updates" that we should not reboot unexpectedly because of a failure. We should probably do e.g. what we do when a file write fails. This might also allow us not to drain (although recovery is another thing we need to consider) What do you think @sinnykumari ?

kikisdeliveryservice: f you degrade, the customer can't really recover from it and _do_ something to make it work on a retry... so degrading might mean that it can't proceed. not sure how that cuts but since it's an optimization it feels like rebooting should be the fall back.

kikisdeliveryservice: (at least to me rebootless upgrades isn't a promise, it's a best effort optimization in certain cases where it can be safely done)

sinnykumari: We can but isn't it better to reboot (crio will restart on reboot) and fix the problem by itself rather than degrade? I think we are not promising rebootless update rather it is a best effort basis i.e. MCO tries its best not to reboot for certain config change but if an error occur it reboots. As this is new feature, we can frame documentation accordingly.

sinnykumari: we can definitely become smarter in future.

kikisdeliveryservice: ha jinx! sinny and i both wrote our responses at the same time :laughing:

pkg/daemon/daemon.go
```diff
@@ -1103,33 +1117,44 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
 		if out, err := dn.storePendingState(state.pendingConfig, 0); err != nil {
-			return errors.Wrapf(err, "failed to reset pending config: %s", string(out))
+			return true, errors.Wrapf(err, "failed to reset pending config: %s", string(out))
 		}
 
 		state.currentConfig = state.pendingConfig
 	}
 
-	if state.bootstrapping {
-		if err := dn.storeCurrentConfigOnDisk(state.currentConfig); err != nil {
```
yuqi-zhang, line 1128: is this no longer needed?

pkg/daemon/update.go
```diff
@@ -93,9 +93,56 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
 	}
 }
 
-// finalizeAndReboot is the last step in an update(), and it can also
-// be called as a special case for the "bootstrap pivot".
-func (dn *Daemon) finalizeAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error) {
+func reloadCrioConfig() error {
+	_, err := runGetOut("pkill", "-HUP", "crio")
```
crawford, line 97: It would be nice if we could use `systemctl reload crio.service` instead. That way, crio can specify how exactly it needs to be reloaded (`ExecReload`) without the MCD needing to know the details. I don't think this will work today, since crio.service doesn't define an ExecReload (systemd will restart the service in this case).

pkg/daemon/update.go
```diff
@@ -93,9 +93,56 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+	return err
+}
+
+// performRebootAction takes action based on what rebootAction has been asked.
+// For non-reboot action, it applies configuration, updates node's config and state.
+// In the end uncordon node to schedule workload.
+// If at any point an error occurs, we reboot the node so that node has correct configuration.
+func (dn *Daemon) performRebootAction(action rebootAction, configName string) error {
```
crawford, line 105: I would rename this so that it doesn't include "reboot" in the name (since we might not be rebooting). I'm also not sure that an enum is going to be expressive enough (feel free to tell me to stop over-complicating things and to come back in six months). I had envisioned this being expressed as a list of actions that needed to be taken (e.g. `[ ]` (rebootActionNone), `[ reboot ]` (rebootActionReboot), `[ reload crio.service, reload foobar.service]` (rebootActionReloadCrio + reload foobar) so that we could build up the list as we enumerate the diffed config (e.g. any changes to a particular set of paths triggers a reload of crio, any unrecognized change triggers a reboot).

pkg/daemon/update.go
yuqi-zhang: Agreed but I think for this release we are ok with this operation as it stands

pkg/daemon/update.go
yuqi-zhang: I've changed it to "PostConfigChangeAction" in an attempt to be more descriptive but that doesn't quite feel right either. Naming open to suggestions.

As for the enum, I've changed it to a list. I initially created the enum as a skeleton structure since we were handling very limited cases, but I agree that parsing a list will be needed in the future. I've updated that at: https://github.com/openshift/machine-config-operator/pull/2259

yuqi-zhang:
PR that builds on this: https://github.com/openshift/machine-config-operator/pull/2259

pkg/daemon/update.go
sinnykumari: As Jerry said we can but we are trying to keep impact as minimal as possible. Since, crio supports [live configuration reload for container registry](https://github.com/cri-o/cri-o/blob/master/docs/crio.conf.5.md), we are making use of it in our usecase.

pkg/daemon/daemon.go
sinnykumari: good catch, seems like it got lost during function split up.

pkg/daemon/update.go
sinnykumari: Agree with better regrouping of actions.

sinnykumari:
For better co-ordination, review and testing, we have updated changes from this PR into Jerry's PR https://github.com/openshift/machine-config-operator/pull/2259 . Closing this as further chanes will be pushed directly into https://github.com/openshift/machine-config-operator/pull/2259

PR #2255, opened 2020-11-25
Bug 1901472: Only enable unicast mode on platform that asked for it

mandre:
The bootstrap node was unconditionally setting the ENABLE_UNICAST
environment variable to yes on all on-prem platforms, causing it to use
unicast_src_ip and unicast_peer settings in keeepalived.conf, while it
should have used the output of the
`onPremPlatformKeepalivedEnableUnicast` function as done on the control
plane nodes.

### DIFF
--- a/manifests/on-prem/keepalived.yaml
+++ b/manifests/on-prem/keepalived.yaml
@@ -105,7 +105,7 @@ spec:
     image: {{ .Images.BaremetalRuntimeCfgBootstrap }}
     env:
       - name: ENABLE_UNICAST
-        value: "yes"
+        value: "{{ onPremPlatformKeepalivedEnableUnicast .ControllerConfig }}"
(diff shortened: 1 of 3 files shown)


mandre:
/hold

```
error rendering bootstrap manifests: failed to parse asset manifests/on-prem/keepalived.yaml: template: manifests/on-prem/keepalived.yaml:108: function "onPremPlatformKeepalivedEnableUnicast" not defined
```
This needs a bit more work :)

mandre:
Great, this passes CI and has an LGTM from the Brad who worked on the de-dup patch. We just need an approval now.
/assign @runcom

mandre:
/test e2e-ovirt
/test e2e-vsphere

@Gal-Zaidman @jcpowermac FYI, this patch should help fixing 4.7 deployment on oVirt and vSphere platforms too.

PR #2181, opened 2020-10-26
Bug 1890074: daemon: allow one to one mapping of extension on OKD

sinnykumari:
FCOS does one to one mapping of extension to package
to be installed on FCOS node. This is needed as OKD
layers additional packages on top of official FCOS shipped,
See https://github.com/openshift/release/blob/959c2954344438c4eed3ec7f52a5e099e8335516/ci-operator/jobs/openshift/release/openshift-release-release-4.7-periodics.yaml#L586

In future, once the package list has been stabilized,
we can make use of the group and add all the packages required
to enable OKD as a single extension.

PR #2256, opened 2020-11-25
Bug 1908648: [test] Skip TestKernelType on OKD

LorbusChris:
Adds an isOKDCluster function to determine whether the current cluster
is an OKD cluster.

Skips TestKernelType on OKD because the realtime kernel is not available
on OKD clusters.

### DIFF
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -156,6 +156,12 @@ func TestKernelArguments(t *testing.T) {
 func TestKernelType(t *testing.T) {
 	cs := framework.NewClientSet("")
 
+	isOKD, err := isOKDCluster(cs)
+	require.Nil(t, err)
+	if isOKD {
+		t.Skip("skipping test on OKD")
+	}
+
 	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", "node-role.kubernetes.io/infra")
 
 	oldInfraRenderedConfig := getMcName(t, cs, "infra")
@@ -174,7 +180,7 @@ func TestKernelType(t *testing.T) {
 		},
 	}
 
-	_, err := cs.MachineConfigs().Create(context.TODO(), kernelType, metav1.CreateOptions{})
+	_, err = cs.MachineConfigs().Create(context.TODO(), kernelType, metav1.CreateOptions{})
 	require.Nil(t, err)
 	t.Logf("Created %s", kernelType.Name)
 	renderedConfig, err := waitForRenderedConfig(t, cs, "infra", kernelType.Name)
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -275,8 +275,19 @@ func TestExtensions(t *testing.T) {
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], renderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
 
-	installedPackages := execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel", "kernel-headers")
-	expectedPackages := []string{"usbguard", "kernel-devel", "kernel-headers"}
+	isOKD, err := isOKDCluster(cs)
+	require.Nil(t, err)
+
+	var installedPackages string
+	var expectedPackages []string
+	if isOKD {
+		// OKD does not support grouped extensions yet, so installing kernel-devel will not also pull in kernel-headers
+		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel")
+		expectedPackages = []string{"usbguard", "kernel-devel"}
+	} else {
+		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel", "kernel-headers")
+		expectedPackages = []string{"usbguard", "kernel-devel", "kernel-headers"}
+	}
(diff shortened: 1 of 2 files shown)


LorbusChris:
couldn't pull images from quay due to an outage - should be resolved now
/retest

test/e2e/utils_test.go
```diff
@@ -183,6 +183,13 @@ func execCmdOnNode(t *testing.T, cs *framework.ClientSet, node corev1.Node, args
 	return string(b)
 }
 
+func getOsReleaseFromNode(t *testing.T, cs *framework.ClientSet, node corev1.Node) (string, string) {
+	osId := execCmdOnNode(t, cs, node, "bash", "-c", "cat /rootfs/etc/os-release | grep '^ID=' | sed -e 's/^ID=//'")
```
cgwalters, line 187: This is going to be fragile around shell quoting handling.  The `os-release` files are designed instead to be sourced by the shell.  So what you really want to do is more like
```
bash -c ". /rootfs/etc/os-release; echo $ID"
```

test/e2e/mcd_test.go
```diff
@@ -158,6 +158,13 @@ func TestKernelType(t *testing.T) {
 
 	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", "node-role.kubernetes.io/infra")
 
+	infraNode := getSingleNodeByRole(t, cs, "infra")
+
+	osId, osVariant := getOsReleaseFromNode(t, cs, infraNode)
+	if osId == "fedora" && osVariant == "coreos" {
```
cgwalters, line 164: Isn't there a convenient cluster-wide way to check for OKD (maybe the `clusterversion` CRD?) instead of querying the node to see if it's FCOS?  (We don't support joining FCOS workers to an OCP cluster AFAIK right?)

LorbusChris:
I'll rework this
/hold
/retitle WIP: [test] Skip TestKernelType on OKD

LorbusChris:
@vrutkovs this now fails at [TestExtensions](https://github.com/openshift/machine-config-operator/blob/master/test/e2e/mcd_test.go#L256) since we don't ship `usbguard` and `kernel-devel` in the extensions yum repo in OKD's machine-os-content. 

Let's ship them there instead of skipping this test as well (while not installing them by default).

vrutkovs:
>package kernel-headers is not installed

iirc `kernel-devel` in MCO really means `kernel-devel and kernel-headers` - can't find a reference on top of my head right now

LorbusChris:
yeah it does look that way: https://github.com/openshift/os/blob/master/extensions.yaml
Should we fix the test to install `kernel-headers` explicitly as an extension, or not check for it to be installed?
In the former case we'd have to add it to okd-machine-os' extension repo.

sinnykumari:
> > package kernel-headers is not installed
> 
> iirc `kernel-devel` in MCO really means `kernel-devel and kernel-headers` - can't find a reference on top of my head right now

https://github.com/openshift/machine-config-operator/pull/2170
Since OKD uses https://github.com/openshift/machine-config-operator/pull/2181 , extensions test for OKD will need to be written down accordingly.

vrutkovs:
I think we should stay close to rhcos here, so lets a) add kernel-headers to okd-machine-os and b) revert OKD hack in rhcos

sinnykumari (commented):
For correctness, can you also update conditional run at https://github.com/openshift/machine-config-operator/blob/master/test/e2e/mcd_test.go#L301 .
Since this PR is already making changes to this test, can you please remove https://github.com/openshift/machine-config-operator/blob/master/test/e2e/mcd_test.go#L302 as this seems duplicate.

vrutkovs:
https://github.com/openshift/okd-machine-os/pull/44 would add this to OKD's machine-os-content

sinnykumari:
awesome!

LorbusChris:
> [openshift/okd-machine-os#44](https://github.com/openshift/okd-machine-os/pull/44) would add this to OKD's machine-os-content

we'd still have to adapt the test here to explicitly install it

sinnykumari:
LGTM
/approve

sinnykumari:
e2e-gcp-op will hopefully get fixed by end of the day

sinnykumari:
Need newer m-o-c  with https://github.com/coreos/coreos-assembler/pull/1971 fixes

vrutkovs (approved):
/lgt,

:partying_face:

sinnykumari:
@LorbusChris you can remove hold when you think it is good to go

sinnykumari:
weird lgtm label didn't get applied.

/lgtm

sinnykumari:
umm, a bug also needs to get created as we are in feature freeze

vrutkovs:
/retitle Bug 1908648: [test] Skip TestKernelType on OKD

LorbusChris:
this works, lifting the hold
/hold cancel

kikisdeliveryservice:
So okd-e2e-gcp-op is now passing okd-e2e-aws is passing e2e-gcp-op is passing and this PR only changes files in /test/e2e so I'm overriding e2e-aws-serial/e2e-aws/e2e-agnostic-upgrade bc they are unaffected by this pr and okd having functioning ci is important.

/override ci/prow/e2e-aws
/override ci/prow/e2e-aws-serial

kikisdeliveryservice:
Also ftr modified tests were run in the e2e-gcp-op job:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2256/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1339634970741706752/artifacts/e2e-gcp-op/test/container-logs/test.log

kikisdeliveryservice:
One more time bot:

/override ci/prow/e2e-aws

LorbusChris:
Thank you @kikisdeliveryservice!

/cherry-pick release-4.6
so we can get OKD tests on 4.6 working again..

PR #2258, opened 2020-11-25
[baremetal & friends] Move api-int to /etc/hosts

cybertron:
Instead of relying on coredns to provide the api-int record, which
requires our coredns to be deployed as a static pod so it will be
available before the node tries to register with the api, we can
put the api-int record in /etc/hosts. This way the node can register
without coredns running and then coredns can be started at a later
point in the deployment. This makes our coredns deployment much more
flexible.

Since I'm told we should not try to template all of /etc/hosts because
other service may be trying to add entries to it as well, this
change adds a systemd service and timer to ensure that the correct
entry is in /etc/hosts. It will run once per minute and append the
api-int line if it doesn't already find one there.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/on-prem/files/coredns-corefile.yaml
+++ b/templates/common/on-prem/files/coredns-corefile.yaml
@@ -27,13 +27,4 @@ contents:
             match api.{{ .DNS.Spec.BaseDomain }}
             fallthrough
         }
-        template IN {{`{{ .Cluster.APIVIPRecordType }}`}} {{ .DNS.Spec.BaseDomain }} {
-            match api-int.{{ .DNS.Spec.BaseDomain }}
-            answer "{{`{{"{{ .Name }}"}}`}} 60 in {{`{{"{{ .Type }}"}}`}} {{ onPremPlatformAPIServerInternalIP . }}"
-            fallthrough
-        }
-        template IN {{`{{ .Cluster.APIVIPEmptyType }}`}} {{ .DNS.Spec.BaseDomain }} {
-            match api-int.{{ .DNS.Spec.BaseDomain }}
-            fallthrough
-        }
     }
--- a/templates/common/on-prem/units/api-int-service.yaml
+++ b/templates/common/on-prem/units/api-int-service.yaml
@@ -8,4 +8,4 @@ contents: |
 
   [Service]
   Type=oneshot
-  ExecStart=/bin/bash -c "grep -q {{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }} /etc/hosts || echo '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' >> /etc/hosts"
+  ExecStart=/bin/bash -c "grep -q -e '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' -F /etc/hosts || echo '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' >> /etc/hosts"
--- a/templates/common/on-prem/units/api-int-service.yaml
+++ b/templates/common/on-prem/units/api-int-service.yaml
@@ -1,0 +1,11 @@
+name: api-int.service
+# This service is run by a timer
+enabled: false
+contents: |
+  [Unit]
+  Description=Ensure api-int entry exists in /etc/hosts
+  Before=kubelet.service crio.service
+
+  [Service]
+  Type=oneshot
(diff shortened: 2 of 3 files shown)


templates/common/on-prem/units/api-int-service.yaml
```diff
@@ -0,0 +1,11 @@
...
+# This service is run by a timer
+enabled: false
+contents: |
+  [Unit]
+  Description=Ensure api-int entry exists in /etc/hosts
+  Before=kubelet.service crio.service
+
+  [Service]
+  Type=oneshot
+  ExecStart=/bin/bash -c "grep -q {{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }} /etc/hosts || echo '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' >> /etc/hosts"
```
Miciah, line 11: The search pattern should be quoted so that `api-int` and  `api-int.{{ .DNS.Spec.BaseDomain }}` are parsed as part of the pattern and not as names of files to search:
```suggestion
  ExecStart=/bin/bash -c "grep -q -e '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' /etc/hosts || echo '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' >> /etc/hosts"
```
Is there any chance the internal IP address or base domain could change?  If so, presumably you need to clean up the old entries.

Miciah: Also, you might as well use `-F`:
```suggestion
  ExecStart=/bin/bash -c "grep -q -e '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' -F /etc/hosts || echo '{{ onPremPlatformAPIServerInternalIP . }} api-int api-int.{{ .DNS.Spec.BaseDomain }}' >> /etc/hosts"
```

templates/common/on-prem/files/coredns-corefile.yaml
```diff
@@ -27,13 +27,4 @@ contents:
             match api.{{ .DNS.Spec.BaseDomain }}
             fallthrough
         }
-        template IN {{`{{ .Cluster.APIVIPRecordType }}`}} {{ .DNS.Spec.BaseDomain }} {
```
yboaron, line 30: The overall idea LGTM, I have one Q.

I understand that you deleted api-int from CoreDNS to verify that we can break the Kubelet->coredns->Kubelet circular dependency and have a stable deployment without relying on CoreDNS for resolving api-int. 
I wonder if it will be safer (for 4.7) to keep also api-int in CoreDns ? in case something went wrong with /etc/hosts .

templates/common/on-prem/units/api-int-service.yaml
cybertron: Good suggestions, thanks!

I'm not sure whether we support changing the api vip or the domain, but I think we can make this easy to clean up, so I'll make that change as well.

templates/common/on-prem/files/coredns-corefile.yaml
cybertron: 1. I considered that, but then we have multiple definitions of the api-int record which can get out of sync. If we're going to do this, I think I'd rather just make a clean switch over. I'm open to being convinced otherwise though.

2. Good point. I didn't want to introduce differences between the platforms when this should work on all of them.

kikisdeliveryservice:
@mandre @jcpowermac : friends.. PTAL

jcpowermac:
/assign @patrickdillon 

cc: @rvanderp3

cgwalters:
> Since I'm told we should not try to template all of /etc/hosts because
other service may be trying to add entries to it as well, this
change adds a systemd service and timer to ensure that the correct
entry is in /etc/hosts. It will run once per minute and append the
api-int line if it doesn't already find one there.

OK a few things here.  First just a note: we can use inotify to watch the file and avoid wakeups.  (In Kubernetes this doesn't really matter because there's so much polling going on anyways...but it's exactly the kind of problem that becomes important in e.g. small scale IoT/edge type systems where you just want to run some containers on a single system in a CPU/power efficient way)

Second...why does this have to be in shell script at all?  Can't it be part of the MCD Go code?  If we need this `api-int` record before kubelet comes up, this process could be part of the existing `machine-config-daemon-firstboot.service` - basically this is part of a general pattern where we can include things in the MCD and run them either before or after kubelet.

Third, as implemented today if the internal IP changes won't this result in *multiple* lines ending up in `/etc/hosts`?  I think the logic probably needs to be something more like writing a section of the file that looks like

```
127.0.0.1 localhost blah blah the default /etc/hosts content
...
(stuff here might be injected manually by custom Ignition)
...

# Managed by machine-config-operator; do not edit
10.x.x.x api-int
# End machine-config-operator managed section
```

and the MCO code carefully only edits the latter section.

cybertron:
After further discussion, we don't think /etc/hosts is a good place for this after all. If we pursue this, https://github.com/openshift/machine-config-operator/pull/2374 is likely a better option.

PR #2259, opened 2020-11-26
daemon: selectively reboot based on diffs of applied MCs

yuqi-zhang:
Adds to/supercedes https://github.com/openshift/machine-config-operator/pull/2254

This is the overall PR for the reboot epic work, which aims to allow "rebootless updates" when certain changes happen.

Currently supported:
No action (drain, update files, uncordon)
 - ssh keys
 - pull secret

Reload crio
 - registries.conf changes

Reboot
 - all other scenarios

Also add corresponding unit and e2e tests

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -165,14 +165,6 @@ const (
 	onceFromRemoteConfig
 )
 
-type rebootAction int
-
-const (
-	rebootActionReboot rebootAction = iota
-	rebootActionNone
-	rebootActionReloadCrio
-)
-
 var (
 	defaultRebootTimeout = 24 * time.Hour
 )
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -165,6 +165,14 @@ const (
 	onceFromRemoteConfig
 )
 
+type rebootAction int
+
+const (
+	rebootActionReboot rebootAction = iota
+	rebootActionNone
+	rebootActionReloadCrio
+)
+
 var (
 	defaultRebootTimeout = 24 * time.Hour
 )
@@ -1015,7 +1023,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		if err := dn.drain(); err != nil {
 			return err
 		}
-		return dn.finalizeAndReboot(state.pendingConfig)
+		if err := dn.finalizeBeforeReboot(state.pendingConfig); err != nil {
+			return err
+		}
+		return dn.reboot(fmt.Sprintf("Node will reboot into config %v", state.pendingConfig.GetName()))
 	}
 
 	if err := dn.detectEarlySSHAccessesFromBoot(); err != nil {
@@ -1043,7 +1054,10 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 			if err := os.RemoveAll(osImageContentDir); err != nil {
 				return err
 			}
-			return dn.finalizeAndReboot(state.currentConfig)
+			if err := dn.finalizeBeforeReboot(state.currentConfig); err != nil {
+				return err
+			}
+			return dn.reboot(fmt.Sprintf("Node will reboot into config %v", state.currentConfig.GetName()))
 		}
 		glog.Info("No bootstrap pivot required; unlinking bootstrap node annotations")
 
@@ -1103,33 +1117,44 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
 	}
 
-	// We've validated our state.  In the case where we had a pendingConfig,
-	// make that now currentConfig.  We update the node annotation, delete the
-	// state file, etc.
-	//
-	// However, it may be the case that desiredConfig changed while we
-	// were coming up, so we next look at that before uncordoning the node (so
-	// we don't uncordon and then immediately re-cordon)
+	// We've validated state. Now, ensure that node is in desired state
+	var inDesiredConfig bool
+	if inDesiredConfig, err = dn.updateConfigAndState(state); err != nil {
+		return err
+	}
+	if inDesiredConfig {
+		return nil
+	}
+
+	if dn.recorder != nil {
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "BootResync", fmt.Sprintf("Booting node %s, currentConfig %s, desiredConfig %s", dn.node.Name, state.currentConfig.GetName(), state.desiredConfig.GetName()))
+	}
+	// currentConfig != desiredConfig, and we're not booting up into the desiredConfig.
+	// Kick off an update.
+	return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
+}
+
+// updateConfigAndState updates node to desired state, labels nodes as done and uncordon
+func (dn *Daemon) updateConfigAndState(state *stateAndConfigs) (bool, error) {
+	// In the case where we had a pendingConfig, make that now currentConfig.
+	// We update the node annotation, delete the state file, etc.
 	if state.pendingConfig != nil {
 		if dn.recorder != nil {
 			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "NodeDone", fmt.Sprintf("Setting node %s, currentConfig %s to Done", dn.node.Name, state.pendingConfig.GetName()))
 		}
 		if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.pendingConfig.GetName()); err != nil {
-			return errors.Wrap(err, "error setting node's state to Done")
+			return true, errors.Wrap(err, "error setting node's state to Done")
 		}
 		if out, err := dn.storePendingState(state.pendingConfig, 0); err != nil {
-			return errors.Wrapf(err, "failed to reset pending config: %s", string(out))
+			return true, errors.Wrapf(err, "failed to reset pending config: %s", string(out))
 		}
 
 		state.currentConfig = state.pendingConfig
 	}
 
-	if state.bootstrapping {
-		if err := dn.storeCurrentConfigOnDisk(state.currentConfig); err != nil {
-			return err
-		}
-	}
-
+	// In case of node reboot, it may be the case that desiredConfig changed while we
+	// were coming up, so we next look at that before uncordoning the node (so
+	// we don't uncordon and then immediately re-cordon)
 	inDesiredConfig := state.currentConfig.GetName() == state.desiredConfig.GetName()
 	if inDesiredConfig {
 		if state.pendingConfig != nil {
@@ -1138,7 +1163,7 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 			glog.Infof("Completing pending config %s", state.pendingConfig.GetName())
 			if err := dn.completeUpdate(dn.node, state.pendingConfig.GetName()); err != nil {
 				MCDUpdateState.WithLabelValues("", err.Error()).SetToCurrentTime()
-				return err
+				return inDesiredConfig, err
 			}
 		}
 		// If we're degraded here, it means we got an error likely on startup and we retried.
@@ -1147,7 +1172,7 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 			if err := dn.nodeWriter.SetDone(dn.kubeClient.CoreV1().Nodes(), dn.nodeLister, dn.name, state.currentConfig.GetName()); err != nil {
 				errLabelStr := fmt.Sprintf("error setting node's state to Done: %v", err)
 				MCDUpdateState.WithLabelValues("", errLabelStr).SetToCurrentTime()
-				return errors.Wrap(err, "error setting node's state to Done")
+				return inDesiredConfig, errors.Wrap(err, "error setting node's state to Done")
 			}
 		}
 
@@ -1155,14 +1180,8 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		MCDUpdateState.WithLabelValues(state.currentConfig.GetName(), "").SetToCurrentTime()
 
 		// All good!
-		return nil
 	}
-	if dn.recorder != nil {
-		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "BootResync", fmt.Sprintf("Booting node %s, currentConfig %s, desiredConfig %s", dn.node.Name, state.currentConfig.GetName(), state.desiredConfig.GetName()))
-	}
-	// currentConfig != desiredConfig, and we're not booting up into the desiredConfig.
-	// Kick off an update.
-	return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
+	return inDesiredConfig, nil
 }
 
 // runOnceFromMachineConfig utilizes a parsed machineConfig and executes in onceFrom
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1144,6 +1144,12 @@ func (dn *Daemon) updateConfigAndState(state *stateAndConfigs) (bool, error) {
 		state.currentConfig = state.pendingConfig
 	}
 
+	if state.bootstrapping {
+		if err := dn.storeCurrentConfigOnDisk(state.currentConfig); err != nil {
+			return false, err
+		}
+	}
+
 	// In case of node reboot, it may be the case that desiredConfig changed while we
 	// were coming up, so we next look at that before uncordoning the node (so
 	// we don't uncordon and then immediately re-cordon)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1176,9 +1176,9 @@ func (dn *Daemon) updateConfigAndState(state *stateAndConfigs) (bool, error) {
 
 		glog.Infof("In desired config %s", state.currentConfig.GetName())
 		MCDUpdateState.WithLabelValues(state.currentConfig.GetName(), "").SetToCurrentTime()
-
-		// All good!
 	}
+
+	// No errors have occurred. Returns true if currentConfig == desiredConfig, false otherwise (needs update)
 	return inDesiredConfig, nil
 }
 
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1103,9 +1103,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		glog.Info("Validated on-disk state")
 	} else {
 		glog.Infof("Skipping on-disk validation; %s present", constants.MachineConfigDaemonForceFile)
-		if err := os.Remove(constants.MachineConfigDaemonForceFile); err != nil {
-			return errors.Wrap(err, "failed to remove force validation file")
-		}
 		return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
 	}
 
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -117,20 +117,27 @@ func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string
 		dn.logSystem("Rebooting node")
 		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
 	}
+
 	if ctrlcommon.InSlice(postConfigChangeActionNone, postConfigChangeActions) {
 		if dn.recorder != nil {
-			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot.")
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "SkipReboot", "Config changes do not require reboot.")
 		}
 		dn.logSystem("Node has Desired Config %s, skipping reboot", configName)
 	}
+
 	if ctrlcommon.InSlice(postConfigChangeActionReloadCrio, postConfigChangeActions) {
 		serviceName := "crio"
+
 		if err := reloadService(serviceName); err != nil {
+			if dn.recorder != nil {
+				dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedServiceReload", fmt.Sprintf("Reloading %s service failed. Error: %v", serviceName, err))
+			}
 			dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)
 			dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
 		}
+
 		if dn.recorder != nil {
-			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot. Service %s was reloaded.", serviceName)
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "SkipReboot", "Config changes do not require reboot. Service %s was reloaded.", serviceName)
 		}
 		dn.logSystem("%s config reloaded successfully! Desired config %s has been applied, skipping reboot", serviceName, configName)
 	}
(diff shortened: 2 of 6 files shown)


pkg/daemon/update.go
```diff
@@ -440,6 +446,91 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
...
+
+	// Now check if any files were added/changed
+	for path, newFile := range newFileSet {
+		oldFile, ok := oldFileSet[path]
+		if !ok {
+			// debug: remove
+			glog.Infof("File diff: %v was added", path)
+			diffFileSet = append(diffFileSet, path)
+		}
+		if !reflect.DeepEqual(oldFile, newFile) {
```
sinnykumari, line 485: This should run only when file was found in both old and new MC.

pkg/daemon/update.go
```diff
@@ -486,6 +577,13 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 
 	dn.logSystem("Starting update from %s to %s: %+v", oldConfigName, newConfigName, diff)
 
+	// TODO: consider how we should honor "force" flag here. Maybe if force, always reboot?
+	// TODO: consider if we should not cordon if no action needs to be taken
```
sinnykumari, line 581: We are doing multiple things when we call updateFiles() irrespective of what rebootAction we perform like deleting stale file, writing units from newIgnConfig, etc. There are chances of getting errors in these steps. Cordon and drain will ensure that we don't end up with degraded node that has running workload.

pkg/daemon/update.go
yuqi-zhang: ah good catch! this works as it today I think but will cause a duplicate add of a file. Will fix

pkg/daemon/update.go
yuqi-zhang: Hmm ok, I think my consideration was that drain on some workloads (e.g. CI) take quite awhile so it would be a nice value add. And if we were doing an update that does not require a reboot, it was unlikely to break the state of the actual node (at worst, it would e.g. fail a file write, but would not commit any changes to disk and the node itself should still be working even as we're degrading the pool).
For safety purposes though perhaps we should do this drain method for now, and consider improving in the future?

sinnykumari: You are right, even I get tempted to skip drain which will work fine during happy path. This is probably an extra precaution to avoid coroner cases bugs that bite us later on.
+1 on revisiting drain optimization later on (probably in 4.8?)

sinnykumari:
Tested all scenarios we are handling in this PR including node scaleup, working perfectly fine.

yuqi-zhang:
Fixed above error case and added some initial unit tests. Will expand.

Manual testing looks good so far.

sinnykumari:
Added e2e test

pkg/daemon/update.go
```diff
@@ -48,6 +48,12 @@ const (
 	fipsFile              = "/proc/sys/crypto/fips_enabled"
 	extensionsRepo        = "/etc/yum.repos.d/coreos-extensions.repo"
 	osImageContentBaseDir = "/run/mco-machine-os-content/"
+
+	// These are the actions for a node to take after applying config changes.
+	// Defaults to reboot.
+	postConfigChangeActionNone       = "none"
```
cgwalters, line 54: What change would result in no action?

pkg/daemon/update.go
```diff
@@ -93,9 +99,56 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
 	}
 }
 
-// finalizeAndReboot is the last step in an update(), and it can also
-// be called as a special case for the "bootstrap pivot".
-func (dn *Daemon) finalizeAndReboot(newConfig *mcfgv1.MachineConfig) (retErr error) {
+func reloadCrioConfig() error {
+	_, err := runGetOut("pkill", "-HUP", "crio")
```
cgwalters, line 103: One of the massive benefits of systemd over baseline Unix is that by using cgroups, it applies a rigorous *structure* to the prior total Wild West of Unix processes you could identify by grepping their name or maybe racy PID files.

In particular, this invocation will send SIGHUP to e.g. a process running in a regular user pod if it happens to be named "crio".

What you want instead is `systemctl reload crio` - systemd is actively tracking the PID that it launched and can send the signal in a race-free way to exactly the correct process.

pkg/daemon/update.go
yuqi-zhang: SSH keys + pull secret. I'll add some comments here.

"No action" is perhaps not the right way to frame it, we still cordon, drain + write files/units/etc. We just don't reload any unit and don't reboot.

pkg/daemon/update.go
yuqi-zhang: Ah for some reason I thought crio didn't have it defined. I can confirm that it indeed should be doing the same thing:

```
# systemctl cat crio.service
...
ExecReload=/bin/kill -s HUP $MAINPID
...
```

And running a reload:
```
Nov 30 23:23:17 ip-10-0-131-15 systemd[1]: Reloading Open Container Initiative Daemon.
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.304012464Z" level=info msg="Reloading configuration"
Nov 30 23:23:17 ip-10-0-131-15 systemd[1]: Reloaded Open Container Initiative Daemon.
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.309633738Z" level=info msg="Updating config from file /etc/crio/crio.conf.d/0-default"
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.309845337Z" level=info msg="Updating config from path /etc/crio/crio.conf.d"
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.310093463Z" level=info msg="Applied new registry configuration: &{Registries:] UnqualifiedSearchRegistries:[registry.access.redhat.com docker.io]}"
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.310131904Z" level=info msg="No seccomp profile specified, using the internal efault"
Nov 30 23:23:17 ip-10-0-131-15 crio[1507]: time="2020-11-30 23:23:17.310144510Z" level=info msg="Set config seccomp_profile to \"\""
```

Seems like its doing what we want. I'll defer to @sinnykumari since she is more knowledgeable on whether that is correct.

yuqi-zhang:
/retest

`uptimeOld, err := strconv.ParseFloat(oldTime, 64)` seems to have failed in the run but works locally, trying again just in case

sinnykumari:
> /retest
> 
> `uptimeOld, err := strconv.ParseFloat(oldTime, 64)` seems to have failed in the run but works locally, trying again just in case

yeah, I don't know why it is failing in ci and works fine when running `make test-e2e` on a local cluster. This is making debugging difficult. From one of the ci run where I added debug message, it printed:
```
 mcd_test.go:193: DEBUG: Uptime file content: I1130 18:29:05.151521    8238 request.go:645] Throttling request took 1.019506334s, request: GET:https://api.ci-op-fpxq6893-1354f.origin-ci-int-gce.dev.openshift.com:6443/apis/storage.k8s.io/v1beta1?timeout=32s
```

pkg/daemon/update.go
sinnykumari: Ah, didn't know `crio reload` already does the same thing, definitely nicer way to go.
Useful learning and something I will keep in mind :)

test/e2e/mcd_test.go
```diff
@@ -153,6 +154,128 @@ func TestKernelArguments(t *testing.T) {
...
 
+func TestNoReboot(t *testing.T) {
+	cs := framework.NewClientSet("")
+	unlabelFunc := labelRandomNodeFromPool(t, cs, "worker", "node-role.kubernetes.io/infra")
+	oldInfraRenderedConfig := getMcName(t, cs, "infra")
+
+	infraNode := getSingleNodeByRole(t, cs, "infra")
+
+	output := execCmdOnNode(t, cs, infraNode, "cat", "/rootfs/proc/uptime")
+	oldTime := strings.Split(output, " ")[0]
```
kikisdeliveryservice, line 165: Looking at the test failure are you sure this oldTime is correct? maybe it should be a different index?

sinnykumari: command will return correct value, but it seems to go into api timeout  error when we call `cat /rootfs/uptime` through execCmdOnNode() . See the error message in https://github.com/openshift/machine-config-operator/pull/2259#issuecomment-736298449 from one of the previous gcp-op run on this [job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2259/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1333453313349259264)

sinnykumari: While replying to you, I understand what is happening. execCmdOnNode() combines both stdout and error and returns it which is causing the problem.

sinnykumari:
gcp-op test is green now :tada: 
/retest

yuqi-zhang:
seems like a flake
/retest

kikisdeliveryservice:
Going to pull up a cluster with this PR just to test it out :woman_dancing:

pkg/daemon/daemon.go
```diff
@@ -1103,33 +1109,50 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
+	if dn.recorder != nil {
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "BootResync", fmt.Sprintf("Booting node %s, currentConfig %s, desiredConfig %s", dn.node.Name, state.currentConfig.GetName(), state.desiredConfig.GetName()))
+	}
+	// currentConfig != desiredConfig, and we're not booting up into the desiredConfig.
+	// Kick off an update.
+	return dn.triggerUpdateWithMachineConfig(state.currentConfig, state.desiredConfig)
+}
+
+// updateConfigAndState updates node to desired state, labels nodes as done and uncordon
+func (dn *Daemon) updateConfigAndState(state *stateAndConfigs) (bool, error) {
```
kikisdeliveryservice, line 1130: im a little confused about why we are returning bool here when we seem to just care about whether we return an error or not? where this is called we are always first checking error , so im not clear on why we need to return the bool bc it seems to always come down to: is there an error or not (and there's only one case where err is nil)

pkg/daemon/update.go
```diff
@@ -93,9 +99,57 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+func reloadService(name string) error {
+	_, err := runGetOut("systemctl", "reload", name)
+	return err
+}
+
+// performPostConfigChangeAction takes action based on what postConfigChangeAction has been asked.
+// For non-reboot action, it applies configuration, updates node's config and state.
+// In the end uncordon node to schedule workload.
+// If at any point an error occurs, we reboot the node so that node has correct configuration.
+func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string, configName string) error {
```
kikisdeliveryservice, line 111: wondering if there should be an event here?

test/e2e/mcd_test.go
```diff
@@ -153,6 +154,129 @@ func TestKernelArguments(t *testing.T) {
 	require.Nil(t, err)
 }
 
+func TestNoReboot(t *testing.T) {
```
kikisdeliveryservice, line 157: why break up the 2 kernel tests and put this in between?

pkg/daemon/update.go
```diff
@@ -93,9 +99,57 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+		dn.logSystem("Rebooting node")
+		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+	}
+	if ctrlcommon.InSlice(postConfigChangeActionNone, postConfigChangeActions) {
+		dn.logSystem("Node has Desired Config %s, skipping reboot", configName)
+	}
+	if ctrlcommon.InSlice(postConfigChangeActionReloadCrio, postConfigChangeActions) {
+		serviceName := "crio"
+		if err := reloadService(serviceName); err != nil {
+			dn.logSystem("Reloading %s configuration failed, rebooting: %v", serviceName, err)
```
kikisdeliveryservice, line 122: feel like this would make more sense as:
`"Reloading %s configuration failed with error: %v"

pkg/daemon/update.go
```diff
@@ -93,9 +99,57 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+		}
+		dn.logSystem("%s config reloaded successfully! Desired config %s has been applied, skipping reboot", serviceName, configName)
+	}
+
+	// We are here, which means reboot was not needed to apply the configuration.
+
+	// Get current state of node, in case of an error reboot
+	state, err := dn.getStateAndConfigs(configName)
+	if err != nil {
+		glog.Errorf("Error processing state and configs, rebooting: %v", err)
```
kikisdeliveryservice, line 133: same as above..

"Error processing state ang configs: %v, err"

pkg/daemon/update.go
```diff
@@ -93,9 +99,57 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+	}
+
+	var inDesiredConfig bool
+	if inDesiredConfig, err = dn.updateConfigAndState(state); err != nil {
+		glog.Errorf("Setting node's state to Done failed, rebooting: %v", err)
```
kikisdeliveryservice, line 139: "Setting node's state to Done failed with error: %v"

pkg/daemon/daemon.go
kikisdeliveryservice: is there are case where you return false, nil? The "All good!" comment below implies no?

kikisdeliveryservice (commented):
some questions/comments

pkg/daemon/daemon.go
yuqi-zhang: There is one case where we get `false, nil`, if `inDesiredConfig := state.currentConfig.GetName() == state.desiredConfig.GetName()` evaluates as false, it skips the whole if block and just returns false.

yuqi-zhang: The naming is somewhat confusing, we probably should have named it `isInDesiredConfig` or something originally. But `inDesiredConfig` can be false

pkg/daemon/update.go
yuqi-zhang: Hmm, as in, add events for non-reboot scenarios? (We do have reboot events I believe)

test/e2e/mcd_test.go
yuqi-zhang: Will move

pkg/daemon/update.go
yuqi-zhang: Hmm I'd like to highlight the fact that we need to reboot, how about something like: `Reloading %s configuration failed, node will reboot instead. Error: %v` ?

kikisdeliveryservice: that works too :smile:

pkg/daemon/update.go
kikisdeliveryservice: yes exactly wondering if they are covered sufficiently  (sorry for not being clear!)

pkg/daemon/daemon.go
kikisdeliveryservice: could we delete that "All good!" comment below? since that's back from when it just returned nil?

pkg/daemon/update.go
kikisdeliveryservice: maybe postConfigChangeActionSkipReboot? tho i think the none also works

pkg/daemon/update.go
kikisdeliveryservice: (i can take or leave this was just wondering)

yuqi-zhang: Added some in the specific scenarios below:
`dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot.")`
and
`dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot. Service %s was reloaded.", serviceName)`

Should I also add some warnings for the failure cases? Or what would be a good way to handle those (they would get a reboot event regardless)

pkg/daemon/update.go
yuqi-zhang: I think e.g. reloading crio also "skips reboot", so I defaulted to "none" to mean "no extra action needed"

kikisdeliveryservice:
Played around with this pre- jerry fixing up comments and it works fine for me. i tested that things work to not reboot but reload crio, etc... then made a mc that changed the ssh key and a file and it did indeed reboot.

I think rolling it out with no reboot was ~6min for 3 nodes vs 11 min to roll out for the same nodes with reboot so some time savings there!

Also the comment clarifications from the last push are really helpful for me and addressed my questions.

Will let runcom/walters take final looks but lgtm.

pkg/daemon/update.go
kikisdeliveryservice: good point, im fine with none :+1:

pkg/daemon/update.go
kikisdeliveryservice: we have events for failsdrain and fails to reconcile for ex:
`dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)`
maybe a similar one near?: `dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)`

yuqi-zhang: Did you mean this line: `dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)` ?
So basically for what I added, instead of `Normal` I added it as a warning instead?

kikisdeliveryservice: oh duh yes sorry for the bad copypasta, yes that is the example, and yeah i thought maaybe put a warning near the reloading failed logging you already have to show it failed

pkg/daemon/update.go
```diff
@@ -93,9 +103,63 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+	if ctrlcommon.InSlice(postConfigChangeActionNone, postConfigChangeActions) {
+		if dn.recorder != nil {
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot.")
+		}
+		dn.logSystem("Node has Desired Config %s, skipping reboot", configName)
+	}
+	if ctrlcommon.InSlice(postConfigChangeActionReloadCrio, postConfigChangeActions) {
+		serviceName := "crio"
+		if err := reloadService(serviceName); err != nil {
+			dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)
```
kikisdeliveryservice, line 129: event here is what i mean @yuqi-zhang :laughing:

pkg/daemon/update.go
```diff
@@ -93,9 +103,63 @@ func getNodeRef(node *corev1.Node) *corev1.ObjectReference {
...
+// In the end uncordon node to schedule workload.
+// If at any point an error occurs, we reboot the node so that node has correct configuration.
+func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string, configName string) error {
+	if ctrlcommon.InSlice(postConfigChangeActionReboot, postConfigChangeActions) {
+		dn.logSystem("Rebooting node")
+		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+	}
+	if ctrlcommon.InSlice(postConfigChangeActionNone, postConfigChangeActions) {
+		if dn.recorder != nil {
+			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Reboot", "Config changes do not require reboot.")
```
sinnykumari, line 122: Shouldn't the reason of event here be "SkipReboot" instead of "Reboot"?

pkg/daemon/update.go
sinnykumari: added warning event

sinnykumari:
flake due to VpcLimitExceeded
/retest

sinnykumari:
Need final round of review and approval get it merged.
/assign @cgwalters @runcom

cgwalters (approved):
I gave this a medium-level review and I think it looks good, but there are a lot of subtleties here and it touches code paths that we don't necessarily have CI coverage.

On the flip side, I think this is pretty safe to land because it will only kick in in very specialized circumstances.  We're very unlikely to break any default installs or CI.

That said, I do have one request (I wouldn't call this a blocker but a very nice to have): Let's add a new `machineconfiguration.openshift.io/bootedConfig` annotation that is set the first time we do a "live" update.

If you look at the recent [rpm-ostree livefs work](https://github.com/coreos/rpm-ostree/pull/2293) which is quite analogous to this, if you type `rpm-ostree status` you can clearly see that a "live" update was applied, as distinct from the booted commit.

Now currently nothing in rpm-ostree tries to capture the full "live history", and I don't think we need to do that here either.  The goal is that it should be *immediately obvious* looking at a node if it's in a "live applied" state.  That will help us debug problems in the future.

To clarify...anyone else should feel free to add a LGTM as is, but hopefully you agree with the above and we can try to add a separate `bootedConfig` as a followup soon.

cgwalters:
If you haven't tried `rpm-ostree ex livefs`, here's a demo in current Fedora CoreOS.  There's a strong analogy between:

- "booted deployment" == "currentConfig"
- "staged deployment" == "desiredCommit"

Then when a livefs is applied, there's a new "livefs commit" (which is the same as staged, but note that staged could change without live changing later, so they need to be separately tracked).

```
```
walters@toolbox /v/s/w/b/fcos> cosa run
...
[root@cosa-devsh ~]# rpm-ostree install strace
...
Will download: 1 package (1.2?MB)
...
Added:
  strace-5.9-1.fc33.x86_64
Run "systemctl reboot" to start a reboot
[root@cosa-devsh ~]# rpm-ostree status
State: idle
Deployments:
  ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 33.20201119.dev.0 (2020-11-19T22:04:26Z)
                BaseCommit: 6d5b62bd02745cffbc4f3b470196eed8207fffc674c6602c428f632c226e5f2e
              GPGSignature: (unsigned)
                      Diff: 1 added
           LayeredPackages: strace

* ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 33.20201119.dev.0 (2020-11-19T22:04:26Z)
                    Commit: 6d5b62bd02745cffbc4f3b470196eed8207fffc674c6602c428f632c226e5f2e
              GPGSignature: (unsigned)
[root@cosa-devsh ~]# rpm-ostree ex livefs
[root@cosa-devsh ~]# rpm-ostree status
State: idle
Deployments:
  ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 33.20201119.dev.0 (2020-11-19T22:04:26Z)
                BaseCommit: 6d5b62bd02745cffbc4f3b470196eed8207fffc674c6602c428f632c226e5f2e
                    Commit: 0c398101141bfbf6c3661dd0d237622b865e2b85966007960e05a71cc7fcf1df
              GPGSignature: (unsigned)
                      Diff: 1 added
           LayeredPackages: strace

* ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 33.20201119.dev.0 (2020-11-19T22:04:26Z)
              BootedCommit: 6d5b62bd02745cffbc4f3b470196eed8207fffc674c6602c428f632c226e5f2e
                LiveCommit: 0c398101141bfbf6c3661dd0d237622b865e2b85966007960e05a71cc7fcf1df
                  LiveDiff: 1 added
              GPGSignature: (unsigned)
                  Unlocked: transient
[root@cosa-devsh ~]# 
```

File: pkg/daemon/update.go (hunk 12 above)
Comment (kikisdeliveryservice) [reply pkg/daemon/update.go:135]: perf!

**Comment by kikisdeliveryservice** [comment]
I think @cgwalters suggestion could be a good followup for debugability (is that a word?)

@runcom any final thoughts?

File: pkg/daemon/update.go [hunk 14]
```diff
@@ -519,6 +526,17 @@ func calculatePostConfigChangeActionFromFileDiffs(oldIgnConfig, newIgnConfig ign
 }
 
 func calculatePostConfigChangeAction(oldConfig, newConfig *mcfgv1.MachineConfig) ([]string, error) {
+	// If a machine-config-daemon-force file is present, it means the user wants to
+	// move to desired state without additional validation. We will reboot the node in
+	// this case regardless of what MachineConfig diff is.
+	if _, err := os.Stat(constants.MachineConfigDaemonForceFile); err == nil {
+		if err := os.Remove(constants.MachineConfigDaemonForceFile); err != nil {
```
yuqi-zhang, line 533: So this does change the behaviour of how the forcefile works slightly, but I guess its probably for the better, since this way it stays on the system until an update happens. Off the top of my head I don't think this should cause a problem. So +1

yuqi-zhang:
Ok I think this is good to go! Let's aim to improve documentation and debugability in follow up PRs. Thanks everyone for the reviews!

/retest

kikisdeliveryservice:
let's do this.

/lgtm

kikisdeliveryservice:
everything passed! but now... tide is gonna... retest :neutral_face:

runcom:
> To clarify...anyone else should feel free to add a LGTM as is, but hopefully you agree with the above and we can try to add a separate bootedConfig as a followup soon.

That is indeed as @kikisdeliveryservice commented a great thing to have for debuggability - we'll still must-gather in the cases we're investigating our usual bugs but it'll definitely help direct the investigation the right way. We can definitely follow up!

/lgtm

sinnykumari:
>  hopefully you agree with the above and we can try to add a separate `bootedConfig` as a followup soon.

Nice and useful suggestion, definitely we will discuss and work on adding it.

PR #2261, opened 2020-11-30
templates: change additional kubelet service files with new log-level

sjenning:
@rphillips 

https://github.com/openshift/machine-config-operator/pull/2211 did not change the log level for all platforms

### DIFF
--- a/templates/master/01-master-kubelet/on-prem/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/on-prem/units/kubelet.service.yaml
@@ -10,7 +10,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
-  Environment="KUBELET_LOG_LEVEL=4"
+  Environment="KUBELET_LOG_LEVEL=3"
   EnvironmentFile=/etc/os-release
   EnvironmentFile=-/etc/kubernetes/kubelet-workaround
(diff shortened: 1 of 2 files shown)


templates/master/01-master-kubelet/baremetal/units/kubelet.service.yaml
```diff
@@ -0,0 +1,41 @@
+name: kubelet.service
```
rphillips, line 1: adding a baremetal config?

cgwalters:
Ah yep, we still really need to de-dup the templates.  Probably handle these explicitly in the code.

templates/master/01-master-kubelet/on-prem/units/kubelet.service.yaml
```diff
@@ -10,7 +10,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
-  Environment="KUBELET_LOG_LEVEL=4"
+  Environment="KUBELET_LOG_LEVEL=3"
```
darkmuggle, line 13: Honest question: why not move this its own dropin to `/etc/systemd/system/kubelet.service.d/kubelog.conf` with
```
[Unit]
Environment="KUBELET_LOG_LEVEL=3"
```
That would set the log level in _one_ stop.

sjenning:
/hold
might do what Ben suggests instead.  @rphillips can you do that?

rphillips:
yep... on it

sjenning:
close in favor of https://github.com/openshift/machine-config-operator/pull/2262

PR #2262, opened 2020-11-30
Bug 1903290: kubelet: refactor KUBELET_LOG_LEVEL into KubeletConfig field

rphillips:
**- What I did**
Takes Ben's [suggestion](https://github.com/openshift/machine-config-operator/pull/2261#discussion_r532807948) and refactors the Kubelet log level into a systemd environment file.

Obsoletes: https://github.com/openshift/machine-config-operator/pull/2261

This also changes the default of the log level to 2, since stale logs are printed into the kubelet log at level 3: [BZ1903290](https://bugzilla.redhat.com/show_bug.cgi?id=1903290).

Allows the logLevel to be configured on the KubeletConfig CRD:

```
kind: KubeletConfig
metadata:
  name: set-kubelet-loglevel
spec:
  logLevel: 2
  machineConfigPoolSelector:
    matchLabels:
      pools.operator.machineconfiguration.openshift.io/worker: ""
```

**- How to verify it**


templates/common/_base/files/kubelet-log-level.yaml
```diff
@@ -0,0 +1,6 @@
+mode: 0644
+path: "/etc/systemd/system/kubelet.service.d/kubelog.conf"
+contents:
+  inline: |
+[Unit]
+Environment="KUBELET_LOG_LEVEL=3"
```
darkmuggle, line 6:
```suggestion
  inline: |
      [Unit]
      Environment="KUBELET_LOG_LEVEL=3"
```

Missing indentation. Otherwise, this look much cleaner :)

darkmuggle (commented):
Other than indentation, this looks good.

templates/common/_base/files/kubelet-log-level.yaml
rphillips: thanks. fixed :)

kikisdeliveryservice:
vpc exceeded error, reported in 4-dev

darkmuggle:
/approve 
But I'll leave the LGTM to a member of the MCO team.

kikisdeliveryservice:
aws/gcp are having issues across the board today would like to see passing runs before approving

templates/common/_base/files/kubelet-log-level.yaml
```diff
@@ -3,4 +3,4 @@ path: "/etc/systemd/system/kubelet.service.d/kubelog.conf"
 contents:
   inline: |
     [Unit]
-    Environment="KUBELET_LOG_LEVEL=3"
+    Environment="KUBELET_LOG_LEVEL=2"
```
kikisdeliveryservice, line 6: i thought level 2 was not sufficient for debugging?

rphillips: Problem is that level 3 is logging dead pod logs into the kubelet log. Instead of chasing all level 3 logs... I'm proposing just setting it to level 2. This new approach with a kubelog.conf would allow us to manage the kubelet log level within the kubeletconfig perhaps.

[Log Line](https://github.com/openshift/origin/blob/release-4.4/vendor/k8s.io/kubernetes/pkg/kubelet/status/status_manager.go#L434)

rphillips: I'm open to changing the log line to be 4, but it seems like whack-a-mole.

kikisdeliveryservice: If there's consensus that level work works between the experts (ie you and @sjenning) I'm ok with it, just want to make sure this isn't going to hurt us later.

@sjenning wdyt?

kikisdeliveryservice:
uh since ci is still underwater could you update: https://github.com/openshift/machine-config-operator/blob/master/docs/KubeletConfigDesign.md

kikisdeliveryservice:
unit and verify failures don't look to be flakes..

rphillips:
Updated to fix the unit tests and allow support for only setting the logLevel

rphillips:
Vpc limit

/retest

kikisdeliveryservice:
@rphillips brought this up again in 4-dev no aws resolution yet :frowning_face:

kikisdeliveryservice:
It's getting late but I'll test this on a cluster tomorrow.

templates/common/_base/files/kubelet-log-level.yaml
sjenning: @kikisdeliveryservice I'm on board with this

kikisdeliveryservice (approved):
Tested & seems to work ok - made kubelet config applied, then changed kubelet logging seemed to change appropriately.

sjenning:
still hitting VPC limits issues :disappointed:

pkg/controller/kubelet-config/helpers.go
```diff
@@ -22,7 +22,33 @@ import (
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 )
 
-func createNewKubeletIgnition(jsonConfig []byte) ign3types.Config {
+func createNewKubeletLogLevelIgnition(level int32) *ign3types.File {
+	config := fmt.Sprintf(`
+[Service]
+Environment="KUBELET_LOG_LEVEL=%d"
+	`, level)
```
sjenning, line 29: This is creating some weird whitespace issues
```
sh-4.4# cat 20-logging.conf 

[Service]
Environment="KUBELET_LOG_LEVEL=3"
	sh-4.4#
```

sjenning: actually, this is more than a nit because if I go from having no `KubeletConfig` to having one that specifies log-level 2 (the default), I get an unexpected rollout because the default `20-logging.conf` doesn't match the rendered one.

kikisdeliveryservice: Given that this is a be probably should just fix it now instead of creating another bz to fix later.

kikisdeliveryservice:
Since Seth found an issue, I think we should fix it first then merge

/hold

pkg/controller/kubelet-config/helpers.go
rphillips: fixed!

rphillips:
/hold cancel

fixed the issue.

sjenning:
Trying it out now. Will lgtm when verified.

PR #2264, opened 2020-11-30
Bug 1899350: Include dhcp client-id and bond options during ovs-configuration

trozet:
Adds copying over the dhcp client-id as well as bond options from the interface being migrated over to OVS.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -85,7 +85,7 @@ contents:
       fi
 
       # store old conn for later
-      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
+      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | awk "/\s${iface}\s*\$/ {print \$1}")
 
       extra_brex_args=""
       # check for dhcp client ids
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -137,6 +137,11 @@ contents:
         extra_phys_args="dev ${vlan_parent} id ${vlan_id}"
       elif [ $(nmcli --get-values connection.type conn show ${old_conn}) == "bond" ]; then
         iface_type=bond
+        # check bond options
+        bond_opts=$(nmcli --get-values bond.options conn show ${old_conn})
+        if [ -n "$bond_opts" ]; then
+          extra_phys_args+="bond.options ${bond_opts} "
+        fi
       else
         iface_type=802-3-ethernet
       fi
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -84,6 +84,21 @@ contents:
         echo "MTU found for iface: ${iface}: ${iface_mtu}"
       fi
 
+      # store old conn for later
+      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
+
+      extra_brex_args=""
+      # check for dhcp client ids
+      dhcp_client_id=$(nmcli --get-values ipv4.dhcp-client-id conn show ${old_conn})
+      if [ -n "$dhcp_client_id" ]; then
+        extra_brex_args+="ipv4.dhcp-client-id ${dhcp_client_id} "
+      fi
+
+      dhcp6_client_id=$(nmcli --get-values ipv6.dhcp-duid conn show ${old_conn})
+      if [ -n "$dhcp6_client_id" ]; then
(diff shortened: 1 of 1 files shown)


trozet:
@danwinship @dcbw

cgwalters:
I think it'd be pretty easy BTW to:

- Move this logic into a container in the SDN
- Change the MCO templates to pull down and execute that container via podman before starting kubelet, much like we do with `machine-config-daemon-firstboot.service`

danwinship:
(Filed a jira card about moving the logic out of MCO: https://issues.redhat.com/browse/SDN-1374)

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -124,6 +132,11 @@ contents:
         extra_phys_args="dev ${vlan_parent} id ${vlan_id}"
       elif [ $(nmcli --get-values connection.type conn show ${old_conn}) == "bond" ]; then
         iface_type=bond
+        # check bond options
+        bond_opts=$(nmcli --get-values bond.options conn show ${old_conn})
+        if [ -n "$bond_opts" ]; then
+          extra_phys_args+="bond.options ${bond_opts} "
```
danwinship, line 138: This is going to have quoting issues. The `bond.options` value can definitely have spaces in it, and maybe other shell metacharacters.

danwinship (changes requested):

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -84,6 +84,16 @@ contents:
         echo "MTU found for iface: ${iface}: ${iface_mtu}"
       fi
 
+      # store old conn for later
+      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | grep ${iface} | awk '{print $1}')
```
celebdor, line 88: No need to both grep and awk
```suggestion
      old_conn=$(nmcli --fields UUID,DEVICE conn show --active | awk "/${iface}/ {print \$1}")
```

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -84,6 +84,16 @@ contents:
...
+
+      extra_brex_args=""
+      # check for dchp-client-id
+      dhcp_client_id=$(nmcli --get-values ipv4.dhcp-client-id conn show ${old_conn})
+      if [ -n "$dhcp_client_id" ]; then
+        extra_brex_args+="ipv4.dhcp-client-id ${dhcp_client_id} "
+      fi
+
```
celebdor, line 96: Do we also need to preserve `ipv6.dhcp-duid` ?

templates/common/_base/files/configure-ovs-network.yaml
trozet: I don't think bond.options can have spaces:
```
[trozet@trozet ~]$ nmcli conn add type bond con-name bond1 bond.options "mode=active-backup miimon=100 fail_over_mac=follow"
Error: failed to modify bond.options: 'active-backup miimon=100 fail_over_mac=follow' not among [balance-rr, active-backup, balance-xor, broadcast, 802.3ad, balance-tlb, balance-alb].
[trozet@trozet ~]$ 
[trozet@trozet ~]$ nmcli conn add type bond con-name bond1 bond.options mode=active-backup miimon=100 fail_over_mac=follow
Error: invalid <setting>.<property> 'miimon=100'.
[trozet@trozet ~]$ nmcli conn add type bond con-name bond3 bond.options mode=active-backup,miimon=100,fail_over_mac=follow
Connection 'bond3' (4fefd985-fbb9-4d16-a51d-fd28361e5cae) successfully added.
```

ifcfg bond options can have spaces:
BONDING_OPTS="mode=active-backup miimon=100 fail_over_mac=follow"

however when the ifcfg is loaded into NM, NM is smart enough to convert it to csv

danwinship: after some experimentation it looks like this should be ok

templates/common/_base/files/configure-ovs-network.yaml
trozet: yeah will add

templates/common/_base/files/configure-ovs-network.yaml
trozet: this made me realize we probably need to be more deterministic here how we match the interface. If iface was eth1, and we had eth10 on the system, we would accidentally grab the wrong one. Going to change it to something like:
```
nmcli --fields UUID,DEVICE conn show --active | awk "/\s${iface}\s*\$/ {print \$1}"
```

kikisdeliveryservice:
is this ovn test supposed to pass @trozet 

/hold

kikisdeliveryservice (approved):
approving, tim if you don't need that test to pass remove the hold

trozet:
the test got far enough that it made it past what this patch invokes, so the failure isn't due to this

/hold cancel

PR #2265, opened 2020-11-30
FAQ: Document /run/machine-config-daemon-force

cgwalters:
This fits in with the other questions.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -54,6 +54,11 @@ If a file that *is* managed by MachineConfig is changed, the MCD will detect thi
 
 In the future, we would like to harden things more so that these things are more controlled, and ideally avoid having any persistent "unmanaged" state.  But it will take significant work to get there; and the status quo means that we can support other operators such as SDN (and e.g. [nmstate](https://github.com/nmstate/kubernetes-nmstate)) that may control parts of the host without the MCO's awareness.
 
+## Q: I changed something, how do I make the MCO revert it?
+
+First, be sure you've read the above question.  If it's a file that is managed by the MCO, then
+use e.g. `oc debug node/$nodename chroot /host touch /run/machine-config-daemon-force`.  See [mco#1086](https://github.com/openshift/machine-config-operator/pull/1086/) and [mco#1891](https://github.com/openshift/machine-config-operator/pull/1891).  Wait for the next MCD sync.
+
 ## Q: How do I debug a node failing to join the cluster?
 
 In clusters that are managed by the [Machine API](https://github.com/openshift/machine-api-operator/), see


docs/FAQ.md
```diff
@@ -54,6 +54,11 @@ If a file that *is* managed by MachineConfig is changed, the MCD will detect thi
 
 In the future, we would like to harden things more so that these things are more controlled, and ideally avoid having any persistent "unmanaged" state.  But it will take significant work to get there; and the status quo means that we can support other operators such as SDN (and e.g. [nmstate](https://github.com/nmstate/kubernetes-nmstate)) that may control parts of the host without the MCO's awareness.
 
+## Q: I changed something, how do I make the MCO revert it?
+
+First, be sure you've read the above question.  If it's a file that is managed by the MCO, then
+use e.g. `oc debug node/$nodename chroot /host touch /run/machine-config-daemon-force`.  See [mco#1086](https://github.com/openshift/machine-config-operator/pull/1086/) and [mco#1891](https://github.com/openshift/machine-config-operator/pull/1891).  The next MCD sync should honor this; you can also kill the MCD pod on the node to force it to restart and find the file.
```
yuqi-zhang, line 60: I don't think you need to kill the pod. The check happens when syncNode runs so it should reconcile on the next run

wking:
This seems like useful information.  Can it go somewhere in the official openshift-docs as well?

docs/FAQ.md
yuqi-zhang: In fact killing the pod could stop it in the middle of performing a forced update, losing the forcefile and not completing the update

cgwalters: Yeah, removed.

docs/FAQ.md
```diff
@@ -54,6 +54,11 @@ If a file that *is* managed by MachineConfig is changed, the MCD will detect thi
 
 In the future, we would like to harden things more so that these things are more controlled, and ideally avoid having any persistent "unmanaged" state.  But it will take significant work to get there; and the status quo means that we can support other operators such as SDN (and e.g. [nmstate](https://github.com/nmstate/kubernetes-nmstate)) that may control parts of the host without the MCO's awareness.
 
+## Q: I changed something, how do I make the MCO revert it?
+
+First, be sure you've read the above question.  If it's a file that is managed by the MCO, then
+use e.g. `oc debug node/$nodename chroot /host touch /run/machine-config-daemon-force`.  See [mco#1086](https://github.com/openshift/machine-config-operator/pull/1086/) and [mco#1891](https://github.com/openshift/machine-config-operator/pull/1891).  Wait for the next MCD sync.
```
yuqi-zhang, line 60: Sorry one more thing. To be more specific, you can only use this to "skip validation". The MCD itself cannot revert anything, it can only overwrite when an update is triggered. So this would only really work if you're seeing:
`unexpected on-disk state validating against rendered-xxx`
and in the node annotations, the `currentConfig` != `desiredConfig`
then this would cause the validation to get passed, and the new config would get applied without asking questions, most likely overwriting the old file.
If there was no update and you want to revert the changes you did manually, doing a forcefile would cause nothing to happen. You would instead have to go manually edit the desiredConfig, again something we don't want people to do without knowing the correct background.

I'm hesitant to add this to the openshift docs for the reason that the hard part is determining the state you're in and what you can do to fix it. Force doesn't really fix anything and could make the cluster worse, since the fundamental cause of the drift might be something else. (we've had problems of community operators overwriting crio conf, for example)

Maybe similarly, we should instead just say "the MCO can't reprovision nodes today. If you are seeing a validation error and just want to force an update to a new render config, do this".?

kikisdeliveryservice:
this is a docs change

/override ci/prow/e2e-aws-serial

docs/FAQ.md
kikisdeliveryservice: agree about jerry's point adding this to openshift docs..

kikisdeliveryservice:
I think Jerry makes very valid points above and agree that if we update the info to mention force, we should narrow the usage as he suggested. We generally only suggest it when we've looked into the user's state to the point that we're confident that this will not compound issue.  Deferring to him to review and finalize on this. (I think he also mentioned adding not killing the pod in the slack convo)?

/assign @yuqi-zhang

cgwalters:
Eh, not worth a battle to get it in.

PR #2266, opened 2020-11-30
Bug 1873556: [on-prem] inject the proxy into the env for NetworkManager.service

EmilienM:
In order to have the proxy variables (HTTP_PROXY, HTTPS_PROXY and
NO_PROXY), we need to `export` them otherwise then don't end up being
loaded in the environment and it causes issues if a proxy is used,
when pulling an image from a registry for example.

### DIFF
--- a/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
@@ -10,13 +10,13 @@ contents:
 
     {{if .Proxy -}}
     {{if .Proxy.HTTPProxy -}}
-    HTTP_PROXY={{.Proxy.HTTPProxy}}
+    export HTTP_PROXY={{.Proxy.HTTPProxy}}
     {{end -}}
     {{if .Proxy.HTTPSProxy -}}
-    HTTPS_PROXY={{.Proxy.HTTPSProxy}}
+    export HTTPS_PROXY={{.Proxy.HTTPSProxy}}
     {{end -}}
     {{if .Proxy.NoProxy -}}
-    NO_PROXY={{.Proxy.NoProxy}}
+    export NO_PROXY={{.Proxy.NoProxy}}
     {{end -}}
     {{end -}}
 


cgwalters:
Oh heh, that's an embarrassing bug.
/approve

cgwalters:
(Unrelated to this PR's code, I love the stack trace from our Github automation trying to talk to Bugzilla which is trying to talk to JIRA...)

mandre:
Perhaps it's a good time to start setting the proxy configuration system-wide rather than per script. Not sure if this has been considered previously, but is there any reason not to do it?

For our use, it should be enough to drop relevant files to `/etc/systemd/system.conf.d/` and `/etc/profile.d/`. Then we could clean up the many places where we set the proxy in MCO.

We could even inject them via ignition, as in the example at https://coreos.com/os/docs/latest/using-environment-variables-in-systemd-units.html#other-examples

EmilienM:
> Perhaps it's a good time to start setting the proxy configuration system-wide rather than per script. Not sure if this has been considered previously, but is there any reason not to do it?
> 
> For our use, it should be enough to drop relevant files to `/etc/systemd/system.conf.d/` and `/etc/profile.d/`. Then we could clean up the many places where we set the proxy in MCO.
> 
> We could even inject them via ignition, as in the example at https://coreos.com/os/docs/latest/using-environment-variables-in-systemd-units.html#other-examples

+1, if maintainers think it's a good idea I would be happy to take a look in this direction.

mandre:
> > Perhaps it's a good time to start setting the proxy configuration system-wide rather than per script. Not sure if this has been considered previously, but is there any reason not to do it?
> 
> +1, if maintainers think it's a good idea I would be happy to take a look in this direction.

Sorry I should have been a bit more specific. IMO making the setting system-wide is the next logical step but should not necessarily be handled in this PR as it would extend the scope quite a bit.

Thanks for the bug fix.

/lgtm

cgwalters:
Yeah, I argued to do system wide but...I forget why we didn't.

Hmm.  One thing we could probably do is render an `/etc/mco/proxy.env` file and change our units to source it via `EnvironmentFile=/etc/mco/proxy.env` instead. 

Rather than patching our dispatcher script we could also inject the proxy into the env for `NetworkManager.service` instead (so all dispatcher scripts would inherit).

EmilienM:
> Yeah, I argued to do system wide but...I forget why we didn't.
> 
> Hmm. One thing we could probably do is render an `/etc/mco/proxy.env` file and change our units to source it via `EnvironmentFile=/etc/mco/proxy.env` instead.
> 
> Rather than patching our dispatcher script we could also inject the proxy into the env for `NetworkManager.service` instead (so all dispatcher scripts would inherit).

sounds like a plan!

cgwalters:
Did you test this?  It looks correct...but it's a larger change.  I agree with [this comment](https://github.com/openshift/machine-config-operator/pull/2266#issuecomment-736597038) - let's merge the obviously-correct original code and do this as a separate followup.

(But if you've tested it I'm OK to merge as is too)

EmilienM:
> Did you test this? It looks correct...but it's a larger change. I agree with [this comment](https://github.com/openshift/machine-config-operator/pull/2266#issuecomment-736597038) - let's merge the obviously-correct original code and do this as a separate followup.

I agree it would be safer to proceed with the initial change request and move that refactor into another PR.
I'm still in in the process of testing your proposal, so I'll split PRs now so bug can be closed soon.

kikisdeliveryservice:
who exactly needs to sign off on this PR since it affects onprem? bc so far i only see openstack represented? in general, we don't put an approval on prs that affect multiple platforms until we clarify and get the signoff..

assigning some others for review, please assign whoever else needs to look

/approve cancel
/assign @bcrochet 
/assign @yboaron

mandre:
> /cherry-pick release-4.6

The patch doesn't apply cleanly because we de-duped the on prem platforms templates into the `on-prem` directory in 4.7. They were previously in platform specific directories, `openstack`, `baremetal`, `ovirt`, and `vsphere`.

PR #2221, opened 2020-11-12
Add IBM Cloud managed annotations to CVO manifests

csrwng:
This matches https://github.com/openshift/enhancements/pull/445 and does
not change behavior.

PR #2268, opened 2020-12-01
Annotate manifests for single-node-developer cluster profile

guillaumerose:
This implements phase 1 of openshift/enhancements#482

### DIFF
--- a/install/0000_80_machine-config-operator_00_clusterreader_clusterrole.yaml
+++ b/install/0000_80_machine-config-operator_00_clusterreader_clusterrole.yaml
@@ -6,6 +6,7 @@ metadata:
     rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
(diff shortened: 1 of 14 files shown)


cgwalters:
This is going to conflict with https://github.com/openshift/machine-config-operator/pull/2221 but I guess whoever merges second will just win the conflict resolution game :wink:   Well to be fair to him since it was first I added lgtm.

I thought CodeReady wasn't using the MCO because it wasn't doing in-place updates?  In light of https://github.com/openshift/enhancements/pull/504 though perhaps that is changing?

kikisdeliveryservice:
This was discussed in the team meeting and @runcom and @crawford are both aware said that we can merge this for 4.7 for CRC.

kikisdeliveryservice:
ugh labelling isn't getting picked up... reported in 4-dev =(

guillaumerose:
@cgwalters It runs once when we bootstrap the cluster and then we disable it. But the long term goal is to keep it activated (and handle reboot, etc.).

PR #2272, opened 2020-12-01
[common] inject proxy into system wide environment

### LINKED ISSUE #2271: Configure HTTP PROXY variables permanently into environment
**Description**

Proxy variables (HTTP_PROXY, HTTPS_PROXY and
NO_PROXY) are manually set/exported into some bash scripts while it would be beneficial for them to be widely exported on the system; since any application would need access to it (e.g. podman, etc).

**Note**
It is mainly a follow-up of https://github.com/openshift/machine-config-operator/pull/2266.

**Reference**
[Related BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1873556)

EmilienM:
Following the official manual:
https://coreos.com/os/docs/latest/using-environment-variables-in-systemd-units.html#other-examples

In order to load the proxy variables (HTTP_PROXY, HTTPS_PROXY and
NO_PROXY), we now load them at systemd level and in a profile for system
wide impact.

By doing so, we allow the proxy to be used generally by all the services
or commands which look for them.

Closes #2271
Signed-off-by: Emilien Macchi <emilien@redhat.com>

### DIFF
--- a/templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-proxy.conf.yaml
+++ b/templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-proxy.conf.yaml
@@ -1,0 +1,9 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/systemd/system.conf.d/10-default-env-proxy.conf"
+contents:
+  inline: |
+    {{if .Proxy -}}
+    [Manager]
+    DefaultEnvironment=HTTP_PROXY={{.Proxy.HTTPProxy}} HTTPS_PROXY={{.Proxy.HTTPSProxy}} NO_PROXY={{.Proxy.NoProxy}}
+    {{end -}}
--- a/templates/common/_base/files/etc-profile.d-proxy.env.yaml
+++ b/templates/common/_base/files/etc-profile.d-proxy.env.yaml
@@ -1,0 +1,10 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/profile.d/proxy.env"
+contents:
+  inline: |
+    {{if .Proxy -}}
+    export HTTP_PROXY={{.Proxy.HTTPProxy}}
+    export HTTPS_PROXY={{.Proxy.HTTPSProxy}}
+    export NO_PROXY={{.Proxy.NoProxy}}
+    {{end -}}


EmilienM:
Waiting for https://github.com/openshift/machine-config-operator/pull/2266 first to land.

kikisdeliveryservice:
@EmilienM #2266 landed but this needs a rebase.

EmilienM:
> @EmilienM #2266 landed but this needs a rebase.

Thanks. Yeah, I'm actually still testing that chance, hence the WIP. I'll rebase it of course.

templates/common/on-prem/files/proxy-env.yaml
```diff
@@ -0,0 +1,15 @@
+mode: 0644
+path: "/etc/mco/proxy-env"
```
EmilienM, line 2: I tested it, this works fine, I can see the file in my env.

templates/common/on-prem/units/networkmanager.service.yaml
```diff
@@ -0,0 +1,6 @@
+name: NetworkManager.service
```
EmilienM, line 1: this doesn't work, I'll investigate further. I see no change in NetworkManager.service.

templates/common/_base/files/etc-systemd-system.conf.d-10-default-env-proxy.conf.yaml
```diff
@@ -0,0 +1,9 @@
+filesystem: "root"
+mode: 0644
+path: "/etc/systemd/system.conf.d/10-default-env-proxy.conf"
+contents:
+  inline: |
+    {{if .Proxy -}}
+    [Manager]
+    DefaultEnvironment={{if .Proxy.HTTPProxy}}HTTP_PROXY={{.Proxy.HTTPProxy}} {{end}}{{if .Proxy.HTTPSProxy}}HTTPS_PROXY={{.Proxy.HTTPSProxy}} {{end}}{{if .Proxy.NoProxy}}NO_PROXY={{.Proxy.NoProxy}}{{end}}
```
bcrochet, line 8: Do you need to also check {{ if .Proxy.HTTPProxy }} like the other file?

bcrochet, line 8: Rather than wrap each in the if, it might make more sense to just put that immediately after the if .Proxy once.

EmilienM: Like this?

```suggestion
    {{if .Proxy -}}
    [Manager]
    DefaultEnvironment=HTTP_PROXY={{.Proxy.HTTPProxy}} HTTPS_PROXY={{.Proxy.HTTPSProxy}} NO_PROXY={{.Proxy.NoProxy}}
```

PR #2276, opened 2020-12-02
Create drop in file for ContainerRuntimeSearchRegistries

umohnani8:

**- What I did**
A new option, containerRuntimeSearchRegistries, has been added
to the images.config.openshift.io CRD. Implement this new option
in the ctrcfg controller so that when users set this in the cluster
wide Image CR, we create a drop-in file at /etc/containers/registries.conf.d
that sets unqualified-search-registries to the given list. This overwrites
the unqualified-search-registries list in registries.conf and lets the user
decide specify which registries to check when using short names for an image.

This PR is for the UnqualifiedSearchRegistries support epic.

**- How to verify it**
Start a 4.7 cluster and add the `containerRuntimeSearchRegistries` field to the cluster wide Image CR. Once that deploys, you should see a drop-in file at `/etc/containers/registries.conf.d` with the `unqualified-search-registries` field set to the list from the CR. When you try to pull an image using short name, it should only try to pull from the registries specified in that list.

Create drop in file for ContainerRuntimeSearchRegistries option when set in the cluster wide Image CR

### DIFF
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -26,11 +26,12 @@ import (
 )
 
 const (
-	minLogSize           = 8192
-	minPidsLimit         = 20
-	storageConfigPath    = "/etc/containers/storage.conf"
-	registriesConfigPath = "/etc/containers/registries.conf"
-	policyConfigPath     = "/etc/containers/policy.json"
+	minLogSize              = 8192
+	minPidsLimit            = 20
+	storageConfigPath       = "/etc/containers/storage.conf"
+	registriesConfigPath    = "/etc/containers/registries.conf"
+	searchRegDropInFilePath = "/etc/containers/registries.conf.d/01-image-searchRegistries.conf"
+	policyConfigPath        = "/etc/containers/policy.json"
 	// CRIODropInFilePathLogLevel is the path at which changes to the crio config for log-level
 	// will be dropped in this is exported so that we can use it in the e2e-tests
 	CRIODropInFilePathLogLevel   = "/etc/crio/crio.conf.d/01-ctrcfg-logLevel"
@@ -244,7 +245,7 @@ func addTOMLgeneratedConfigFile(configFileList []generatedConfigFile, path strin
 	return configFileList, nil
 }
 
-// createCRIODropinFiles gets the data from the CRD and creates the respective drio in file in /etc/crio/crio.conf.d
+// createCRIODropinFiles gets the data from the CRD and creates the respective drop in file in /etc/crio/crio.conf.d
 // We create different drop-in files for each CRI-O field that can be changed by the ctrcfg CR
 // this ensures that we don't have to rely on hard coded defaults that might cause problems
 // in future if something in cri-o or the templates used by the MCO changes
@@ -277,6 +278,22 @@ func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []generatedConfig
 		if err != nil {
 			glog.V(2).Infoln(cfg, err, "error updating user changes for log-size-max to crio.conf.d: %v", err)
 		}
+	}
+	return generatedConfigFileList
+}
+
+// updateSearchRegistriesConfig gets the ContainerRuntimeSearchRegistries data from the Image CRD
+// and creates a drop-in file for it at /etc/containers/registries.conf.d
+func updateSearchRegistriesConfig(searchRegs []string) []generatedConfigFile {
+	var (
+		generatedConfigFileList []generatedConfigFile
+		err                     error
+	)
+	tomlConf := sysregistriesv2.V2RegistriesConf{}
+	tomlConf.UnqualifiedSearchRegistries = searchRegs
+	generatedConfigFileList, err = addTOMLgeneratedConfigFile(generatedConfigFileList, searchRegDropInFilePath, tomlConf)
+	if err != nil {
+		glog.Warningln("error updating user changes for containerRuntimeSearchRegistries to registries.conf.d: ", err)
 	}
 	return generatedConfigFileList
 }
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -682,7 +682,8 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		}
 		if err := retry.RetryOnConflict(updateBackoff, func() error {
 			registriesIgn, err := registriesConfigIgnition(ctrl.templatesDir, controllerConfig, role,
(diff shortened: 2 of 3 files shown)


umohnani8:
@yuqi-zhang @kikisdeliveryservice @runcom PTAL

kikisdeliveryservice:
Q: does this need an update to https://github.com/openshift/machine-config-operator/blob/master/docs/ContainerRuntimeConfigDesign.md at all?

umohnani8:
@kikisdeliveryservice no it does not, that doc is only for the ctrcfg CRD. This change is for the images.config.openshift.io CRD, which lives in openshift/api and the ctrcfg controller watches it for changes.

kikisdeliveryservice (approved):

haircommander:
LGTM, chances we could get an e2e test? (I'm not sure what tests already exist for registires.conf)

umohnani8:
We have the mock tests for this, which I have updated for this new option. Will work on adding an e2e for the Image CRD in a follow up PR

haircommander:
/approve

waiting on @mrunalp or @mtrmac for a more knowledgeable review

mtrmac (commented):
ACK to the little extent I understand MCO.

And to demonstrate how little I understand it: Why is this fully Go code and not just one of the file templates generated by pkg/controller/template? Is it undesirable to change the user-visible `ControllerConfigSpec` object, is the extra step generating that object undesirable, is it just that this is by far less code, or some other reason?

umohnani8:
The users want to be able to overwrite the `unqualified-search-registries` list with their own list and we are exposing this option in the images.config.openshift.io CRD. So this PR checks for this new option now in the cluster wide Image CR and creates a drop-in file at `/etc/containers/registries.conf.d` with the new `unqualified-search-registries` list. Hence the go code similar to what we do for the crio.conf drop-in files. The template approach wouldn't work here if I understand your comment correctly.

We renamed the user facing option to be `containerRuntimeSearchRegistries` as `unqualifiedSearchRegistries` was a bit confusing plus this list only works for the container runtime, builds and imagestream imports won't be using it for short names.

mtrmac:
> The users want to be able to overwrite the `unqualified-search-registries` list with their own list and we are exposing this option in the images.config.openshift.io CRD. So this PR checks for this new option now in the cluster wide Image CR and creates a drop-in file at `/etc/containers/registries.conf.d` with the new `unqualified-search-registries` list. 

Sure.

> The template approach wouldn't work here if I understand your comment correctly.

Why not? It seems to work for things like `AdditionalTrustBundle`. Is it the conditional existence of the drop-in file? (I understand why `policy.json` and `registries.conf` are not templates, the data has to be fairly non-trivially modified; still, why is the logic in `pkg/controller` and not `pkg/operator`?)

Or alternatively, why do things like `AdditionalTrustBundle` (or https://github.com/openshift/machine-config-operator/pull/2025 ) need to go through the extra `ControllerConfigSpec` CRD step and it isn’t all done in `pkg/controller`?

I’m not at all saying this PR is wrong (it’s certainly the lower-risk approach, given all the existing infrastructure accessing all the relevant CRDs, and given the past pain dealing with the bootstrap process and render differences during that time), I just don’t understand the difference in approach.

mtrmac:
> I just don’t understand the difference in approach.

… which makes me unqualified to be reviewing PRs here; just to be very explicit, this is _not_ a request to reimplement.

kikisdeliveryservice:
> > I just don’t understand the difference in approach.
> 
> … which makes me unqualified to be reviewing PRs here; just to be very explicit, this is _not_ a request to reimplement.

SBT, you came up as suggested by bots

umohnani8:
@yuqi-zhang @sinnykumari would appreciate a review here when you have a few please

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -281,6 +282,22 @@ func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []generatedConfig
...
+func updateSearchRegistriesConfig(searchRegs []string) []generatedConfigFile {
+	var (
+		generatedConfigFileList []generatedConfigFile
+		err                     error
+	)
+	tomlConf := sysregistriesv2.V2RegistriesConf{}
+	tomlConf.UnqualifiedSearchRegistries = searchRegs
+	generatedConfigFileList, err = addTOMLgeneratedConfigFile(generatedConfigFileList, searchRegDropInFilePath, tomlConf)
+	if err != nil {
+		glog.V(2).Infoln("error updating user changes for containerRuntimeSearchRegistries to registries.conf.d: ", err)
```
yuqi-zhang, line 296: If we fail here should we be a bit more vocal about the fact that this failed? Maybe return back up to `registriesConfigIgnition` and fail like all the other cases?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -281,6 +282,22 @@ func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []generatedConfig
 	return generatedConfigFileList
 }
 
+// updateSearchRegistriesConfig gets the ContainerRuntimeSearchRegistries data from the Image CRD
+// and creates a drop-in file for it at /etc/containers/registries.conf.d
+func updateSearchRegistriesConfig(searchRegs []string) []generatedConfigFile {
+	var (
```
yuqi-zhang, line 288: minor nit: could just change below line to `generatedConfigFileList, err :=` and skip the extra definition lines

yuqi-zhang (commented):
Mostly LGTM, 1 concern regarding failure reporting

pkg/controller/container-runtime-config/helpers.go
umohnani8: We don't want to block the rest of the config updates because the drop-in file creation failed, that is why we just log it. The user can always try again if it doesn't work, but that should be a very rare case.

pkg/controller/container-runtime-config/helpers.go
umohnani8: The `addTOMLgeneratedConfigFile` function needs a `[]generatedConfigFile` to be passed in, that is why I have the definition earlier on.

yuqi-zhang: ah I missed that, :+1:

pkg/controller/container-runtime-config/helpers.go
yuqi-zhang: Hmm ok, can we bump the log up to verbosity 1 and maybe have it as a warning? So its easier to find.

umohnani8: Sure, bumped it up to 1, but looks like `Info` is the only available option for some reason

umohnani8: Fixed now :)

yuqi-zhang (approved):
/lgtm

PR #2277, opened 2020-12-02
Bug 1907639: Pass dual-stack node IPs to kubelet in dual-stack clusters

danwinship:
kubelet 1.20 supports, eg,  `--node-ip 10.0.0.5,fd01::5` for bare-metal dual-stack, so pass that when available.

https://github.com/openshift/baremetal-runtimecfg/pull/110 changes runtimecfg to set both `KUBELET_NODE_IP` and `KUBELET_NODE_IPS`, where the latter contains either a single IP (if the node only has one) or dual-stack IPs (if available). Note that we don't tell runtimecfg whether we actually want dual-stack IPs or not, so we have to be careful to only use `KUBELET_NODE_IPS` in the case where we actually expect dual-stack IPs (because kubelet will complain if you pass dual `--node-ip`s on a cloud platform). (Also, restricting the use of `KUBELET_NODE_IPS` to the dual-stack case means we don't have to worry about MCO/runtimecfg version skew during 4.6->4.7 upgrades, because we don't support upgrading from 4.6 dual-stack to 4.7 dual-stack anyway.)

`manifests/controllerconfig.crd.yaml` needs to be regenerated but I couldn't figure out how to do that...

In addition to depending on the runtimecfg PR, this PR should be held until kubelet is rebased to 1.20(-beta), since the 1.19 kubelet won't handle dual-stack `--node-ip`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	k8s.io/code-generator v0.19.2
 	k8s.io/kubectl v0.0.0-20201023045331-0a68e0d30fe9
 	k8s.io/kubelet v0.19.0
+	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
 )
 
 replace (
--- a/manifests/controllerconfig.crd.yaml
+++ b/manifests/controllerconfig.crd.yaml
@@ -389,15 +389,15 @@ spec:
                                 in the cluster.
                               type: string
               nullable: true
+            ipFamilies:
+              description: ipFamilies indicates the IP families in use by the cluster
+                network
+              type: string
             kubeAPIServerServingCAData:
               description: kubeAPIServerServingCAData managed Kubelet to API Server
                 Cert... Rotated automatically
               type: string
               format: byte
-            kubeletIPv6:
-              description: kubeletIPv6 is true to force a single-stack IPv6 kubelet
-                config
-              type: boolean
             networkType:
               description: networkType holds the type of network the cluster is using
               type: string
--- a/pkg/apis/machineconfiguration.openshift.io/v1/types.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/types.go
@@ -88,8 +88,8 @@ type ControllerConfigSpec struct {
 	// +nullable
 	DNS *configv1.DNS `json:"dns"`
 
-	// kubeletIPv6 is true to force a single-stack IPv6 kubelet config
-	KubeletIPv6 bool `json:"kubeletIPv6,omitempty"`
+	// ipFamilies indicates the IP families in use by the cluster network
+	IPFamilies IPFamiliesType `json:"ipFamilies"`
 
 	// networkType holds the type of network the cluster is using
 	// XXX: this is temporary and will be dropped as soon as possible in favor of a better support
@@ -98,6 +98,15 @@ type ControllerConfigSpec struct {
 	// regeneration if this changes.
 	NetworkType string `json:"networkType,omitempty"`
 }
+
+// IPFamiliesType indicates whether the cluster network is IPv4-only, IPv6-only, or dual-stack
+type IPFamiliesType string
+
+const (
+	IPFamiliesIPv4      IPFamiliesType = "IPv4"
+	IPFamiliesIPv6      IPFamiliesType = "IPv6"
+	IPFamiliesDualStack IPFamiliesType = "DualStack"
+)
 
 // ControllerConfigStatus is the status for ControllerConfig
 type ControllerConfigStatus struct {
--- a/pkg/controller/kubelet-config/kubelet_config_features.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features.go
@@ -207,11 +207,7 @@ func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[st
 		return &rv, fmt.Errorf("enabled FeatureSet %v does not have a corresponding config", features.Spec.FeatureSet)
 	}
 	for _, featEnabled := range set.Enabled {
-		if featEnabled == "IPv6DualStack" {
-			glog.Infof("Ignoring IPv6DualStack feature gate for kubelet config")
-		} else {
-			rv[featEnabled] = true
-		}
+		rv[featEnabled] = true
 	}
 	for _, featDisabled := range set.Disabled {
 		rv[featDisabled] = false
--- a/pkg/controller/kubelet-config/kubelet_config_features_test.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features_test.go
@@ -72,34 +72,3 @@ func TestFeaturesDefault(t *testing.T) {
 		})
 	}
 }
-
-func createNewDualStackFeatureGate() *configv1.FeatureGate {
-	return &configv1.FeatureGate{
-		Spec: configv1.FeatureGateSpec{
-			FeatureGateSelection: configv1.FeatureGateSelection{
-				FeatureSet: configv1.IPv6DualStackNoUpgrade,
-			},
-		},
-	}
-}
-
-// Hack for 4.6; don't pass the IPv6DualStack feature gate to kubelet
-func TestFeaturesDualStack(t *testing.T) {
-	f := newFixture(t)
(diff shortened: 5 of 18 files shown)


danwinship:
This is ready for review but not merging yet. Does the approach look OK? In particular, the baremetal-runtimecfg part has to merge first, so if you don't like how it ends up looking on the MCO side we might have to do something different there.

kikisdeliveryservice:
/assign @rphillips 

and also a baremetal person

/assign @celebdor 

I'm going to just add a hold to this since the other pr has to merge first

/hold

kikisdeliveryservice (commented):
doesn't this needs to be changed too?

https://github.com/openshift/machine-config-operator/blob/3add4b96e379d34aa953c8bde82bf0c61054be6d/manifests/controllerconfig.crd.yaml#L397

danwinship:
> doesn't this needs to be changed too?

yes, I mentioned that in passing above; I couldn't figure out how to regenerate it... or is it maintained by hand?

kikisdeliveryservice:
@danwinship here you go: https://github.com/openshift/machine-config-operator/blob/master/docs/HACKING.md#updating-the-manifests

cybertron:
This looks reasonable to me, pending the CRD changes.

danwinship:
Updated the manifest.

Unfortunately, this doesn't work, because https://bugzilla.redhat.com/show_bug.cgi?id=1881213 never got implemented, so we still have the problem from #2108, except now we can't ignore the feature gate any more because kubelet actually checks it (and will refuse to start if you pass a dual-stack `--node-ip` without `--feature-gate IPv6DualStack=true`).

I tried reverting the patch from #2108 and adding a new patch to add `IPv6DualStack` to the kubelet bootstrap when the config is dual-stack, and it resulting in the kubelets coming up correctly, but I still get the

`master: 'pool is degraded because nodes fail with "3 nodes are reporting degraded status on sync": "Node master-1.ostest.test.metalkube.org is reporting: \"machineconfig.machineconfiguration.openshift.io \\\"rendered-master-cd4a8e9b05f4433a15ed36422335b5f7\\\" not found\", Node master-2.ostest.test.metalkube.org is reporting: \"machineconfig.machineconfiguration.openshift.io \\\"rendered-master-cd4a8e9b05f4433a15ed36422335b5f7\\\" not found\", Node master-0.ostest.test.metalkube.org is reporting: \"machineconfig.machineconfiguration.openshift.io \\\"rendered-master-cd4a8e9b05f4433a15ed36422335b5f7\\\" not found\""'`

so I must not have done it right... need help figuring this out

danwinship:
> Updated the manifest.
> 
> Unfortunately, this doesn't work

Ha! It didn't work because I hadn't updated the CRD in the manifest yet when I tested it! Works now:

```
[root@dwinship1 dev-scripts]# oc get clusterversion
NAME      VERSION                                           AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.7.0-0.ci.test-2020-12-04-145949-ci-ln-9twhfl2   True        False         17m     Cluster version is 4.7.0-0.ci.test-2020-12-04-145949-ci-ln-9twhfl2
[root@dwinship1 dev-scripts]# oc get nodes -o template='{{range .items}}{{range .status.addresses}}{{.type}}: {{.address}}, {{end}}{{"\n"}}{{end}}'
InternalIP: 192.168.111.20, InternalIP: fd2e:6f44:5dd8:c956::14, Hostname: master-0.ostest.test.metalkube.org, 
InternalIP: 192.168.111.21, InternalIP: fd2e:6f44:5dd8:c956::15, Hostname: master-1.ostest.test.metalkube.org, 
InternalIP: 192.168.111.22, InternalIP: fd2e:6f44:5dd8:c956::16, Hostname: master-2.ostest.test.metalkube.org, 
InternalIP: 192.168.111.23, InternalIP: fd2e:6f44:5dd8:c956::17, Hostname: worker-0.ostest.test.metalkube.org, 
InternalIP: 192.168.111.24, InternalIP: fd2e:6f44:5dd8:c956::18, Hostname: worker-1.ostest.test.metalkube.org, 
```

So this is ready for review, but still on `/hold` until the kube rebase merges (https://github.com/openshift/kubernetes/pull/471)

kikisdeliveryservice:
Since this affects baremetal, could @celebdor (or another baremetal person) test this once the k8s rebase merges?

danwinship:
It looks like `/test e2e-metal-ipi-ovn-dualstack` is available in this repo, it's just not run by default. So you can test by running that and looking at the `nodes.yaml` in the must-gather (and seeing that each node has both IPv4 and IPv6 `.status.addresses`)

danwinship:
oops
/test e2e-metal-ipi-ovn-dualstack

danwinship:
OK, so the e2e-metal tests don't do a "must-gather" when the test succeeds, so you can't actually verify that the cluster was configured correctly from the result of that test run.

So I brought up a cluster by hand and tested, but apparently we're still getting a 1.19 kubelet so it still doesn't pass. Does there need to be a machine-os-content rebase to the new openshift/kubernetes or something like that? Where is that happening?

(The reason running this PR against a 1.19 kubelet did not cause the e2e-metal-ipi-ovn-dualstack run to fail is apparently because kubelet 1.19 sees `--node-ip=192.168.111.20,fd2e:6f44:5dd8:c956::14` and calls `net.ParseIP("192.168.111.20,fd2e:6f44:5dd8:c956::14")`, which returns `nil`, but it doesn't notice that there was an error and so behaves as though you just hadn't passed `--node-ip` at all, but in the clusters we generate for CI, kubelet is able to guess the same IPv4 IP as nodeip-configuration would have chosen, so it all ends up working.)

/hold
because we're still waiting for kubelet 1.20...

cgwalters:
> because we're still waiting for kubelet 1.20...

FYI recent 4.7 RHCOS promotions are failing https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/?job=promote-release-openshift-machine-os-content-e2e-aws-4.7

which might be related to this.

danwinship:
Latest CI builds have kubelet 1.20 now
/retest
/test e2e-metal-ipi-ovn-dualstack

danwinship:
infra/ci-script flake? (It failed before it even got to anything affected by this PR)
/test e2e-metal-ipi-ovn-dualstack

danwinship:
/hold

From `namespaces/openshift-cluster-machine-approver/pods/machine-approver-546c47bbbc-rjx46/machine-approver-controller/machine-approver-controller/logs/current.log`:
```
2020-12-21T17:09:46.956608646Z I1221 17:09:46.956560       1 main.go:182]
CSR csr-drh4g not authorized: IP address 'fd2e:6f44:5dd8:c956::17' not in machine addresses:
fd00:1101::db7a:2a77:dc13:634b 192.168.111.23
```

and from `namespaces/openshift-machine-api/machine.openshift.io/machines/ostest-cpfrs-worker-0-rpg9q.yaml`:
```
status:
  addresses:
  - address: fd00:1101::db7a:2a77:dc13:634b
    type: InternalIP
  - address: 192.168.111.23
    type: InternalIP
  - address: worker-0.ostest.test.metalkube.org
    type: Hostname
  - address: worker-0.ostest.test.metalkube.org
    type: InternalDNS
```

(`fd00:1101::/64` is dev-scripts "`PROVISIONING_NETWORK`". I don't know what that means, but it shouldn't be putting that IP into the `machine` object. So, we need fixes in the baremetal/ironic stuff somewhere.)

danwinship:
ok, should work now that https://github.com/openshift/baremetal-operator/pull/118 has merged
/test e2e-metal-ipi-ovn-dualstack

danwinship:
/hold cancel
@kikisdeliveryservice @celebdor This is ready to go now. (e2e-metal-ipi-ovn-dualstack passes again, and you can't actually see that it's working from the logs that get gathered [since e2e-metal doesn't include the mustgather output unless the test fails] but I tested by hand as well and confirmed that the nodes are getting dual-stack addresses.)

kikisdeliveryservice:
@celebdor ptal

cybertron (approved):
/lgtm

kikisdeliveryservice (approved):

PR #2283, opened 2020-12-03
Revert "templates: Disable SSH keys lookup from authorized_keys.d"

LorbusChris:
This reverts commit 5ed6fa3a68acd572fd69be8531012807db59e241.

Not having SSH access to the bootstrap node/before MCO starts successfully on OKD is hindering the ability to debug installation related failures immensely.

We'll need to find another way to solve this properly.
/cc @bgilbert

### DIFF
--- a/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
+++ b/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
@@ -1,6 +1,0 @@
-mode: 0644
-path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
-contents:
-  inline: |
-    # disable key lookup from ~/.ssh/authorized_keys.d/ on FCOS
-    AuthorizedKeysCommand none


cgwalters:
Didn't we agree there's an important security issue being fixed by that change?  I think we need a plan other than just reverting.

I don't fully understand the problem here honestly - actually, how can the MCO affect SSH access to the bootstrap?  Isn't all the Ignition for that generated by the installer?

Would it somehow address the problem to move writing this file into the MCD-as-daemonset logic instead of Ignition?

LorbusChris:
@cgwalters the MCO doesn't affect the bootstrap node - it's just that we do not currently have SSH access to any node before MCO starts to manage the key, as Ignition on FCOS writes it to `authorized_keys.d` which is now disabled.
Only the MCO will write it to `authorized_keys` where it can then be looked up by SSH.

LorbusChris:
I wonder, how is ignition configured on RHCOS to write to `authorized_keys` instead of `authorized_keys.d`, and could we configure it to do the same on OKD's FCOS?

bgilbert:
I think @cgwalters's point is that `authorized_keys.d` is only disabled because of the MCO template being removed by this PR, and that template isn't applied on the bootstrap node.

Ignition on RHCOS writes to `authorized_keys` because of a compile-time flag in the Ignition RHCOS package.

LorbusChris:
> I think @cgwalters's point is that `authorized_keys.d` is only disabled because of the MCO template being removed by this PR, and that template isn't applied on the bootstrap node.

oh that's right :man_facepalming:, I'll close this PR. 

 > Ignition on RHCOS writes to `authorized_keys` because of a compile-time flag in the Ignition RHCOS package.
@bgilbert does Ignition have any runtime config options? Could we make this into one?

bgilbert:
> does Ignition have any runtime config options? Could we make this into one?

We generally avoid them, and there's very little opportunity to invoke one anyway, since the initramfs environment is mostly not customizable by the user.  Perhaps we should revisit writing the `sshd_config` fragment after Ignition runs.

PR #2275, opened 2020-12-02
[baremetal] Add ipv6 support to static DHCP configuration

cybertron:
This is essentially an ipv6 version of #2188.

Co-authored-by: Antoni Segura Puimedon <antoni@redhat.com>

**- What I did**

**- How to verify it**

**- Description for the changelog**

PR #2285, opened 2020-12-03
Bug 1904124: [baremetal] Add check for default connection id to static DHCPv6 script

cybertron:
If infinite leases are used on initial boot, we will attempt to
modify the default connection because OVNKubernetes hasn't created
the br-ex bridge yet. We want to wait for that to happen, so if
we see that we're operating on the default "Wired Connection" id
then we want to exit the script.

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-static-dhcpv6.yaml
+++ b/templates/common/baremetal/files/NetworkManager-static-dhcpv6.yaml
@@ -19,6 +19,14 @@ contents:
         exit 0
     fi
 
+    # We don't want this to run before OVNKubernetes creates its bridge. If we
+    # see the default CONNECTION_ID we know to wait.
+    if [ "$CONNECTION_ID" == "Wired Connection" ]
+    then
+        >&2 echo "Refusing to modify default connection."
+        exit 0
+    fi
+
     IPS=($IP6_ADDRESS_0)
     # For ipv6 we have two addresses: DHCP and link-local. Make sure we're using the right one.
     CHECK_STR="^${DHCP6_IP6_ADDRESS}/"


celebdor:
To give a bit more context. For IPv6, the way I was testing was after doing an automated deployment (that sets up the DHCP and DNS as well as the virtual baremetal machines), then swapping the DHCP config and rebooting the machines. The original PR was successful as shown in https://github.com/openshift/machine-config-operator/pull/2275#issuecomment-737366047 .

However, when testing IPv4 for https://github.com/openshift/machine-config-operator/pull/2188 we realized that in the case the DHCP server gives infinite leases from the first node boot, we'll encounter the situation where we persist the configuration for simple eth connection and then (if OVN Kubernetes is in use), that will change the connection to an ovs bridge and that will get persisted as well, ending up with two configured persistent static IP addressing NM connections, which is definitely not what we want.

Seeing that, we set to address the likely case that DHCP is set to infinite leases from start on IPv6.

templates/common/baremetal/files/NetworkManager-static-dhcpv6.yaml
```diff
@@ -19,6 +19,14 @@ contents:
         exit 0
     fi
 
+    # We don't want this to run before OVNKubernetes creates its bridge. If we
+    # see the default CONNECTION_ID we know to wait.
+    if [ "$CONNECTION_ID" == "Wired Connection" ]
```
celebdor, line 24: Should we check if the networking sdn is ovn kubernetes like we do in IPv4 (in case somebody uses a third party SDN, since we know that openshift-sdn is not an option)?

cybertron: I'm thinking not. We know that things break if we modify the Wired Connection because unlike the v4 version we don't create a new connection here. I think it's better for us to refuse to run this with untested SDNs than to hope it works.

celebdor: alright

sinnykumari:
Should we get a review from OVN team to make sure we are not missing any corner case?
/retest

cybertron:
Couldn't hurt.
/cc @trozet 
/retest

PR #2290, opened 2020-12-09
OCPNODE-454: kubelet: ignore reboots on kubelet CA change

rphillips:
**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -477,6 +477,7 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 func calculatePostConfigChangeActionFromFileDiffs(oldIgnConfig, newIgnConfig ign3types.Config) (actions []string) {
 	filesPostConfigChangeActionNone := []string{
 		"/var/lib/kubelet/config.json",
+		"/etc/kubernetes/kubelet-ca.crt",
 	}
 	filesPostConfigChangeActionReloadCrio := []string{
 		"/etc/containers/registries.conf",


romfreiman:
If you specify jira ticket in the commit message it will be linked automatically

pkg/daemon/update.go
```diff
@@ -467,6 +467,7 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 func calculatePostConfigChangeActionFromFileDiffs(oldIgnConfig, newIgnConfig ign3types.Config) (actions []string) {
 	filesPostConfigChangeActionNone := []string{
 		"/var/lib/kubelet/config.json",
+		"/etc/kubernetes/kubelet-ca.crt",
```
yuqi-zhang, line 470: This by itself probably won't do anything. I've never observed this to change without `/etc/kubernetes/ca.crt` also changing.

rphillips: `/etc/kubernetes/ca.crt` gets rendered out from the bootstrap to the masters... I do not see any processes reading ca.crt on a fully installed master. The kubelet does not use ca.crt.

I added /etc/kubernetes/ca.crt into the list.

rphillips:
removed the ca.crt ignore.

David's PR fixes that situation: https://github.com/openshift/machine-config-operator/pull/2294

pkg/daemon/update.go
rphillips: With David's patch we can just keep kubelet-ca.crt...

rphillips:
rebased, ready for 4.8

/hold cancel

yuqi-zhang:
I am :+1: on the change generally, just curious, is there a good way for us to test the kubelet change being applied live properly? Is this covered somewhere in the e2e testing (the general live reload functionality)?

kikisdeliveryservice (changes requested):
can you update the commit to contain a little more detail?

yuqi-zhang:
I've included this in https://github.com/openshift/machine-config-operator/pull/2398 and updated the commit message (https://github.com/openshift/machine-config-operator/pull/2398/commits/3aabfec2ece0d7bd3487399e615e27b8efbfe808) while keeping you as the original author. If you are ok with it @rphillips can we close this in favour of 2398?

rphillips:
@yuqi-zhang thanks! closing.

PR #2291, opened 2020-12-09
Bug 1888565: daemon: Explicitly start rpm-ostreed

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1888565
for one instance of this.  Since the DBus activation timeout is 25s
but the systemd timeout is 90s, let's defer to the latter by explicitly
starting the service.  (This is slightly racy because rpm-ostreed
auto-exits on idle, but when we go to send the DBus message we'll
auto-activate again if really necessary).

Also tell systemd we're going to need the service via the
firstboot unit file.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -60,6 +60,7 @@ type imageInspection struct {
 // NodeUpdaterClient is an interface describing how to interact with the host
 // around content deployment
 type NodeUpdaterClient interface {
+	Start() error
 	GetStatus() (string, error)
 	GetBootedOSImageURL() (string, string, error)
 	Rebase(string, string) (bool, error)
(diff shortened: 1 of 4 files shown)


sinnykumari:
Few observation:
1. We are explicitly starting rpm-ostreed daemon only before fetching OSImageURL. Won't we have to ensure same at other places where we perform rpm-ostree operations?
2. This seems like a workaround to me. For long term solution, are we planning to make changes in rpm-ostree? Perhaps when user calls rpm-ostree functionality, rpm-ostree first checks and starts rpm-ostreed if not already running.

cgwalters:
> We are explicitly starting rpm-ostreed daemon only before fetching OSImageURL. Won't we have to ensure same at other places where we perform rpm-ostree operations?

I think the main problem is during OS boot, there's a lot going on, a lot of services starting at once competing for I/O on the disk.

> This seems like a workaround to me. For long term solution, are we planning to make changes in rpm-ostree? Perhaps when user calls rpm-ostree functionality, rpm-ostree first checks and starts rpm-ostreed if not already running.

That already happens today - that's how it's all working without this!  The problem as noted in the commit message is the disparity between the DBus activation timeout and the systemd one.

But yes, we will likely work around this in rpm-ostree upstream too.  It'll just take much longer to land.  (And in any case I think the last bit of `Wants=` is better regardless because it tells systemd explicitly what's happening)

sinnykumari:
Thanks for the explanation, makes sense then.

pkg/daemon/rpm-ostree.go
```diff
@@ -99,6 +100,18 @@ func (r *RpmOstreeClient) GetBootedDeployment() (*RpmOstreeDeployment, error) {
 	return nil, fmt.Errorf("not currently booted in a deployment")
 }
 
+// Start ensures the daemon is running; the DBus activation timeout
+// is shorter than the systemd timeout.  xref
+// https://bugzilla.redhat.com/show_bug.cgi?id=1888565#c3
```
sinnykumari, line 105: Just wanted to confirm, this solves referenced bug but it doesn't solve bug https://bugzilla.redhat.com/show_bug.cgi?id=1888565 where we are seeing `rpm-ostree cleanup -r` failing because of slow disk, right?

PR #2293, opened 2020-12-10
Remove instructions around removing the worker node label 

briantward:
...when creating a custom infra pool.

This particular solution, when removing the node label manually like this, is incorrect.  Every new "infra" member of this infra custom pool will begin with a worker node label because of the nature of the MachineConfigs it inherits from the worker pool itself, and this is not desirable behavior (it is inconsistent since it requires manual intervention).  So there is always a manual step to remove the worker node label.  I think it also has been shown to have consequences during upgrades (came across BZs but I'll have to go back looking for them for evidence).  In addition, during some reboot sequences on cloud that use the _base template, the kubelet will reapply the worker label anyways:

https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml#L28

I may not have the analysis 100% correct, but it would be good to review this.

### DIFF
--- a/docs/custom-pools.md
+++ b/docs/custom-pools.md
@@ -25,14 +25,6 @@ ip-10-0-151-146.us-west-1.compute.internal   Ready     master         43m   v1.1
 ip-10-0-152-59.us-west-1.compute.internal    Ready     worker         37m   v1.14.0+e020ea5b3
 ```
 
-When we create infra nodes, we apply them to existing worker nodes first. If you want to use the node as a purely infra node, you can remove the worker pool as follows:
-
-```
-oc label node ip-10-0-130-218.us-west-1.compute.internal node-role.kubernetes.io/worker-
-```
-
-This will not change the files on the node itself: the infra pool inherits from workers by default. This means that if you add a new MachineConfig to update workers, a purely infra node will still get updated. However this will mean that workloads scheduled for workers will no longer be scheduled on this node, as it no longer has the worker label.
-
 Next you need to create a MachineConfigPool that contains both the `worker` role and your custom one as MachineConfig selector as follows:
 
 ```console


briantward:
https://bugzilla.redhat.com/show_bug.cgi?id=1760675#c23

"Alex Crawford 2019-11-26 18:19:25 UTC

The design of the MCO is such that there is no intrinsic notion of an infra node. As a result, the infra nodes should carry two roles: infra and worker. The worker role is needed by the MCO in order to manage the lifecycle of the machines' configuration and the infra role is just needed for billing purposes. As it was mentioned above, the the linked documentation (https://access.redhat.com/documentation/en-us/openshift_container_platform/4.1/html/machine_management/creating-infrastructure-machinesets) has no mention of taints, which results in customer workloads potentially landing on the infra nodes. We'll need to get the docs updated to properly taint the infra nodes so that customer workloads don't get scheduled to them.
"

kikisdeliveryservice:
Those instructions were actually updated relatively recently to include manually removing the worker label, bc users requested it see: https://github.com/openshift/machine-config-operator/pull/1824 and the bugzilla comment you ref is stale. You don't _have_ to remove that worker label, but iirc @runcom adjusted it bc some users explicitly wanted the option for clarity while admin.


/hold

briantward:
@kikisdeliveryservice excellent thank you for that reference! I thought last time I checked (before that PR) that this particular reference was not in this doc.  At least I'm not crazy.  

On 4.3 and 4.4 there was some unreliable behavior here.  I saw node reboots end up with worker node relabeled in AWS clusters, and others saw failures to upgrade from 4.3 to 4.4.  Aware of anything that might have changed around reboots and upgrades?

I just tested on 4.5 and 4.6 reboots and upgrades though and they look clean.  Was going to sanity check myself with a 4.3 test.

kikisdeliveryservice:
I believe that the context of the change went along with exactly the issue you are mentioning and there was some work put in to make it better. This is definitely a more recent change (ie the worker label _was_ required before), I'd have to dig to see what exactly @runcom did but these directions should correct for more current ocp versions

briantward:
Thanks for the update. Closing this PR.

PR #2294, opened 2020-12-10
tighten the /etc/kubernetes/ca.crt

deads2k:
`/etc/kubernetes/ca.crt` appears to have certificates used to terminate kube-apiserver client connections, which doesn't seem necessary.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -129,7 +129,6 @@ func RenderBootstrap(
 	bundle = append(bundle, filesData[rootCAFile]...)
 	// Append the kube-ca if given.
 	if _, ok := filesData[kubeAPIServerServingCA]; ok {
-		bundle = append(bundle, filesData[kubeAPIServerServingCA]...)
 		spec.KubeAPIServerServingCAData = filesData[kubeAPIServerServingCA]
 	}
 	// Set the cloud-provider CA if given.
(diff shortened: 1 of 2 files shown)


deads2k:
/hold

let's see what happens  here.

deads2k:
Missing permissions to fetch Quotas and therefore will skip checking them: failed to load limits for servicequotas: failed to list default serviceqquotas

/retest

deads2k:
well, that installed nicely.

/hold cancel

rphillips:
I verified a cluster with this change. /etc/kubernetes/ca.crt contains just the openshift root-ca. The openshift root-ca has a 10 year life span.

yuqi-zhang:
I've included this in https://github.com/openshift/machine-config-operator/pull/2398 and updated the commit message (https://github.com/openshift/machine-config-operator/pull/2398/commits/05614046601633fd677e1799c7d4a35bd8431d32) while keeping you as the original author. If you are ok with it @deads2k can we close this in favour of 2398?

deads2k:
> I've included this in #2398 and updated the commit message ([0561404](https://github.com/openshift/machine-config-operator/commit/05614046601633fd677e1799c7d4a35bd8431d32)) while keeping you as the original author. If you are ok with it @deads2k can we close this in favour of 2398?

yes, thanks.

/hold

just keeping open  until #2398 so my todo list stays somewhat correct.

PR #2295, opened 2020-12-10
avoid draining if no action is required

deads2k:
If the only action is none (or perhaps some special value), then we don't need to drain.


deads2k:
Just a potential example.  No attachment.

deads2k:
https://github.com/openshift/machine-config-operator/pull/2398 looks like the modern version

/close

PR #2297, opened 2020-12-14
Bug 1907333: daemon: Move rollback removal to update loop

cgwalters:
See https://bugzilla.redhat.com/1907333

This needs to be a more proper part of the control loop so that
we retry, rather than having the MCD die.

Also by doing it here it will happen on firstboot when there's
no I/O contention with other processes.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -422,9 +422,14 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 	}
 
 	// Update OS
-	if err := dn.updateOS(newConfig, osImageContentDir); err != nil {
-		MCDPivotErr.WithLabelValues(dn.node.Name, newConfig.Spec.OSImageURL, err.Error()).SetToCurrentTime()
-		return err
+	if mcDiff.osUpdate {
+		if err := dn.updateOS(newConfig, osImageContentDir); err != nil {
+			MCDPivotErr.WithLabelValues(dn.node.Name, newConfig.Spec.OSImageURL, err.Error()).SetToCurrentTime()
+			return err
+		}
+		if err := dn.removeRollback(); err != nil {
+			return err
+		}
 	}
 
 	defer func() {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1029,10 +1029,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		return fmt.Errorf("error detecting previous SSH accesses: %v", err)
 	}
 
-	if err := dn.removeRollback(); err != nil {
-		return errors.Wrapf(err, "Failed to remove rollback")
-	}
-
 	// Bootstrapping state is when we have the node annotations file
 	if state.bootstrapping {
 		targetOSImageURL := state.currentConfig.Spec.OSImageURL


mandre:
Once again,
/bugzilla refresh

pkg/daemon/update.go
```diff
@@ -422,9 +422,14 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
...
 	// Update OS
-	if err := dn.updateOS(newConfig, osImageContentDir); err != nil {
-		MCDPivotErr.WithLabelValues(dn.node.Name, newConfig.Spec.OSImageURL, err.Error()).SetToCurrentTime()
-		return err
+	if mcDiff.osUpdate {
+		if err := dn.updateOS(newConfig, osImageContentDir); err != nil {
+			MCDPivotErr.WithLabelValues(dn.node.Name, newConfig.Spec.OSImageURL, err.Error()).SetToCurrentTime()
+			return err
+		}
+		if err := dn.removeRollback(); err != nil {
```
sinnykumari, line 430: This could be problematic where updated OS doesn't boot fine. Admin will no longer be able to manually rollback which I think can be the case in disconnected install.

I was thinking maybe we can keep the previous code and just ignore the error when deleting rollback failed? Keeping rollback deployment is anyway not a concern other than it provides additional free space.

cgwalters: Yeah, that is a concern.  That said, OpenShift doesn't really support rollbacks in general, and even specifically for the MCO doing it will trigger issues related to https://github.com/openshift/machine-config-operator/issues/1190 - basically the MCO will go degraded.

We don't really have a strong story for e.g. "The new kernel in 4.6.X broke booting on this type of hardware" other than "pause the update and wait for a fixed kernel to arrive and then update".

> I was thinking maybe we can keep the previous code and just ignore the error when deleting rollback failed?

The problem with that is that if we e.g. just log errors we have no real visibility into failures (though I guess we could add an alert) and:

>  Keeping rollback deployment is anyway not a concern other than it provides additional free space.

The original motivation for doing this is (very briefly) mentioned in the PR:
https://github.com/openshift/machine-config-operator/pull/2220

It's basically: compliance tooling that e.g. adjusts kernel parameters for security wants to avoid a system (accidentally) booting into a non-compliant configuration.  If the rollback deployment is there it's an easy down-arrow away for an admin.

yuqi-zhang: So to clarify, this will be removing the rollback (current) deployment when there is an osupdate, during the update cycle of the MCD. If something fails below (e.g. updating kargs), we also call a `removePendingDeployment`. Would that remove both deployments from rpm-ostree? i.e. if we somehow fail there, would we be able to recover at all?

We also run an explicit defer of `applyOSChanges` to roll back to the previous deployment, maybe we should modify that code just in case something was hitting that before.

cgwalters: > Would that remove both deployments from rpm-ostree? i.e. if we somehow fail there, would we be able to recover at all?

There's actually *three* deployments potentially; "pending", "booted", "rollback".  rpm-ostree will never let you remove the booted deployment.  There's no API to do it in rpm-ostree, and multiple layers of safeguards against it (e.g. [this code](https://github.com/ostreedev/ostree/blob/06a77bfd69bbdb6b69ec426dd56a33a14255c214/src/libostree/ostree-sysroot-cleanup.c#L261)), in addition to the entire codebase being designed to avoid mutating the running system.

Does that address the concern?

yuqi-zhang: Ah I see, sorry I got the terms confused.

In this case wouldn't the rollback deployment being removed here get pushed out anyways once we update and reboot? Assuming nothing went wrong the `booted` becomes `rollback`, the `pending` becomes `booted`, and the `rollback` being removed here would be gone anyways?

In a firstboot scenario, this code also runs but fresh nodes would only have a `booted` deployment, and updates to a `pending` deployment via the firstboot systemd unit. There wouldn't be a `rollback` to remove there, so if there are security concerns the firstboot would linger as a rollback and doesn't get removed until a further update.

cgwalters: Oh yes you're totally right :smile:   This won't have the intended effect indeed.

Blah...the architectural problem we have here is there's no "sync loop" in the MCD outside of trying to perform an update; the closest we have is a few ad-hoc goroutines like the ssh monitor.

cgwalters:
Closing in favor of https://github.com/openshift/machine-config-operator/pull/2302

PR #2298, opened 2020-12-14
[release-4.6] Bug 1899622: backport bond options fixes

trozet:
Includes fixes for bond options, dhcp client id, and iface parsing.


trozet:
@danwinship @celebdor

markmc:
(patch manager) looks like a fix for an issue that is affecting some real-world users

markmc:
@trozet @danwinship still needs approved label

yuqi-zhang (approved):

PR #2299, opened 2020-12-14
Bug 1907929: enable madvdontneed in system components

rphillips:
Golang 1.12 changed to use MADV_FREE. MADV_FREE is somewhat faster than 
MADV_DONTNEED; however, the kernel will not try and reclaim memory in golang processes until the system is memory constrained. Cloud servers can potentially page out executable caches, which can cause IOPS throttles. Since golang 1.16  will be changing the default to MADV_DONTNEED, we are going to enable the flag by default.

https://github.com/golang/go/issues/42330

### DIFF
--- a/templates/common/_base/units/crio.service.yaml
+++ b/templates/common/_base/units/crio.service.yaml
@@ -6,3 +6,7 @@ dropins:
       [Service]
       EnvironmentFile=/etc/mco/proxy.env
       {{end -}}
+  - name: 10-mco-default-madv.conf
+    contents: |
+      [Service]
+      Environment="GODEBUG=x509ignoreCN=0,madvdontneed=1"
--- a/templates/common/_base/units/kubelet.service.yaml
+++ b/templates/common/_base/units/kubelet.service.yaml
@@ -6,3 +6,7 @@ dropins:
       [Service]
       EnvironmentFile=/etc/mco/proxy.env
       {{end -}}
+  - name: 10-mco-default-madv.conf
+    contents: |
+      [Service]
+      Environment="GODEBUG=x509ignoreCN=0,madvdontneed=1"


rphillips:
We would also need to tweak the base image as well to set the GODEBUG: https://github.com/openshift/images/blob/515726d73cf1b35abbe0183860f5cde0aad5f387/base/Dockerfile.rhel#L23

cgwalters:
Is the thought that we should do this OpenShift-wide?  The basic tradeoff here is latency vs returning memory to the OS, right?  That seems to be what https://go-review.googlesource.com/c/go/+/135395/ is claiming.

One thing I'd note is at least for experimenting with this w/kubelet and crio, there's no need to patch the MCO; one could provide a `MachineConfig` object that adds a systemd-drop in for those two units and get most of the desired effect here I think.

rphillips:
@cgwalters correct. The thought is that instead of getting into a memory pressure scenario to have GO free the memory back to the kernel as it goes. Since, Golang 1.16 is going to default this option, it should have minimal impact to the system.

This PR is Part 2.
Part 1 of the PR for Openshift Images is here: https://github.com/openshift/images/pull/61

rphillips:
/close

closing for now...

rphillips:
/reopen

cc @smarterclayton

smarterclayton:
Is this EVERY component, or just openshift / rhcos system component services?

rphillips:
This is every component. Since Golang 1.16 is going to default the option it seems reasonable to default it everywhere.

smarterclayton:
Every component inside AND outside a container?  Or just all system golang services not running in a container?  If it's the latter, i think that's ok (we control all of them) BUT we would need to verify that teams in RHCOS that use golang are ready for it.  Probably can't do the former because we don't control all the possible images that run on the system and the golang code they were compiled with (customer could have a golang app that doesn't work with this that we don't want to take from their control).

In general we don't inject our logic into our customers code unless we have to to ensure they function.

rphillips:
This option does not get passed down to inside the container... so all system golang services not running in a container is correct.

Customer workloads do not see this option.

smarterclayton:
Ok, then I think this is acceptable for RHCOS and OCP and OKD to enforce.

rphillips:
This change is safer. It'll only target crio and kubelet. If other system components would like to enable it, then we can do another drop-in for that component.

kikisdeliveryservice:
> Is the thought that we should do this OpenShift-wide? The basic tradeoff here is latency vs returning memory to the OS, right? That seems to be what https://go-review.googlesource.com/c/go/+/135395/ is claiming.
> 
> One thing I'd note is at least for experimenting with this w/kubelet and crio, there's no need to patch the MCO; one could provide a `MachineConfig` object that adds a systemd-drop in for those two units and get most of the desired effect here I think.

new changes seem to fullfil @cgwalters ask and is limited to crio/kubelet so makes sense

/approve

PR #2302, opened 2020-12-14
Bug 1907333: daemon: Revert code to remove rollback

cgwalters:
See https://bugzilla.redhat.com/1907333

Doing this when the MCD lands on a node the first time means
we compete for I/O with lots of other containers, and on
slow systems that can cause a timeout.  Then a further
problem is that we were fatally erroring out in the MCD
rather than retrying.

But we don't have control loop in the MCD today outside of
trying to apply an update; let's defer this until we add one.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1029,10 +1029,6 @@ func (dn *Daemon) checkStateOnFirstRun() error {
 		return fmt.Errorf("error detecting previous SSH accesses: %v", err)
 	}
 
-	if err := dn.removeRollback(); err != nil {
-		return errors.Wrapf(err, "Failed to remove rollback")
-	}
-
 	// Bootstrapping state is when we have the node annotations file
 	if state.bootstrapping {
 		targetOSImageURL := state.currentConfig.Spec.OSImageURL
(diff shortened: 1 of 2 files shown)


yuqi-zhang:
Should probably remove https://github.com/openshift/machine-config-operator/blob/4a005cb1e426a4d844fad89bd88fef95e0ce0246/pkg/daemon/update.go#L700 as well, since there is no other reference to it.

Otherwise I'm :+1:

cgwalters:
Eh, I would have kept it since we might use it in the future, but eh I removed it.

That said I think what we really want actually is https://github.com/coreos/rpm-ostree/issues/2389

mandre:
/test e2e-openstack
Moment of truth...

mandre:
There is progress, at least now [it's deploying the cluster](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2302/pull-ci-openshift-machine-config-operator-master-e2e-openstack/1338747629693046784) :+1:

cgwalters:
> But we don't have control loop in the MCD today outside of trying to apply an update; let's defer this until we add one.

...time passes...we do now!  Added in https://github.com/openshift/machine-config-operator/pull/3070

So I think we could safely do a PR that re-reverts this (i.e. readds it) to fix https://bugzilla.redhat.com/show_bug.cgi?id=2104619

PR #2303, opened 2020-12-15
Bug 1844990:  Bump GoLang to 1.15 for improved TLS Security

darkmuggle:
This change is needed in order to fix CVE-2016-2183.

With [1] TLS 1.3 is on-by-default which improves the security posture of
the MCO. 

[1] https://github.com/golang/go/issues/30055


pkg/server/api.go
```diff
@@ -60,12 +60,23 @@ func NewAPIServer(a *APIHandler, p int, is bool, c, k string) *APIServer {
 
 // Serve launches the API Server.
 func (a *APIServer) Serve() {
```
kikisdeliveryservice, line 62: can these code changes be broken out into a separate commit from the go upgrade?

kikisdeliveryservice (commented):
Do we have to change here too @sinnykumari : https://github.com/openshift/machine-config-operator/blob/522f0fa36cc7b952c6e98e120c58c66e6d795544/machine-config-daemon.spec#L16

Admittedly, I'm not super knowledgeable about this go-bump... would probably perfer @runcom to take a look next week if there's time before the merge.

pkg/server/api.go
darkmuggle: I literally just pushed that commit :)

darkmuggle:
> Admittedly, I'm not super knowledgeable about this go-bump... would probably perfer @runcom to take a look next week if there's time before the merge.

I spent the better part of an afternoon evaluating whether bumping the GoLang version was the better path to resolving the CVE. I concluded that moving to Go1.15 was better due to:
- TLS 1.3 is on by default with downgradablilty to TLS 1.2, where on GoLang 1.13 it defaults to TLS 1.3. TLS 1.3 is a better security posture.
- The `cypto/tls` code exports mappings of Ciphers from their definition to the common name. There's an RFE for use a TLS network policy and by using a newer GoLang we map the common cipher names without custom mapping code (see https://docs.openshift.com/container-platform/4.5/networking/ingress-operator.html#nw-ingress-controller-tls-profiles_configuring-ingress)
- It brings us inline with the Kube Go which was bumped.

pkg/server/api.go
darkmuggle: I updated the `machines-config-daemon.spec` too -- it would be required.

cgwalters: That file is dead btw, the MCD isn't part of RHCOS anymore.

darkmuggle: I *thought* the `spec` file was dead, but wasn't 100% sure.

yuqi-zhang:
Should also do the equivalent of this: https://github.com/openshift/release/pull/8148 I believe

pkg/server/api.go
kikisdeliveryservice: >I literally just pushed that commit :)

great minds!

darkmuggle:
https://github.com/openshift/release/pull/14355 filed for the CI change.

pkg/server/api.go
sinnykumari: yeah as Colin said we no longer need and build machine-config-daemon rpm package. Even if someone wants to build m-c-d rpm package it will build fine with 1.13 because the current 1.15 requirement is added mainly to MCS.

darkmuggle:
I dropped the TLS config for now. Turns out there is no easy way to have HTTP/2 and provide a list of approved ciphers. However, since we are using TLS and HTTP/2 only approved ciphers are allowed so this will drop the SHA hashing and 3DES ciphers.

sinnykumari:
we will need a bug created to get this into 4.7.
LGTM.
/approve

yuqi-zhang (approved):
/lgtm

PR #2208, opened 2020-11-07
operator: allow kube-cloud-config cm with CA data but no cloud config

staebler:
**- What I did**
When installed on C2S regions on AWS, a CA bundle is needed for communication with the AWS API. This CA bundle will be placed in the kube-cloud-config ConfigMap in the openshift-config-managed namespace. When that ConfigMap is present, the machine-config-operator assumes that it will contain a cloud provider config in the "cloud.conf" key. However, for AWS, the cloud provider config is not required.

With these changes, the machine-config-operator will accept a ConfigMap that does not have the "cloud.conf" key. The MCO will still error when the "cloud.conf" key is absent for a platform where the cloud provider config is required.

https://issues.redhat.com/browse/CORS-1584

**- How to verify it**
Add a cloud-provider-config ConfigMap in the openshift-config namespace. Verify that a kube-cloud-config ConfigMap is created in the openshift-config-managed namespace. Restart the machine-config-operator pod. Verify in the logs of the new machine-config-operator pod that the RenderConfig is synced.

**- Description for the changelog**

Accept a kube-cloud-config ConfigMap without "cloud.conf" on platforms that do not require a cloud provider config.

PR #2305, opened 2020-12-16
Bug 1908159: operator: fix error syncing cloud config with CA data but no cloud config

staebler:
The machine-config-operator should accept a kube-cloud-config ConfigMap on platforms that do not require a cloud.conf when the ConfigMap has no cloud.conf data but does have ca-bundle.pem data. This was the intent of https://github.com/openshift/machine-config-operator/pull/2208, but there was a bug in that code.

### DIFF
--- a/pkg/operator/sync_test.go
+++ b/pkg/operator/sync_test.go
@@ -1,14 +1,156 @@
 package operator
 
 import (
+	"testing"
+
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/stretchr/testify/assert"
-	"testing"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes/fake"
+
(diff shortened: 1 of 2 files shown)


pkg/operator/sync_test.go
```diff
@@ -1,14 +1,152 @@
...
+
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
-func TestIsCloudConfigRequired(t *testing.T) {
-	testInfra := configv1.Infrastructure{}
-	testInfra.Status.Platform = "None"
-	required := isCloudConfigRequired(&testInfra)
-	assert.False(t, required)
+func TestSyncCloudConfig(t *testing.T) {
```
kikisdeliveryservice, line 16: Did we lose the platform = "none" as a case?

staebler: I am happy to add a test case for platform "None". Is there something special about that platform that you would like to me test?

staebler: Add a test case for None platform with no kube-cloud-config.

kikisdeliveryservice (approved):
Will let @sinnykumari do the final review

PR #2306, opened 2020-12-16
Bug 1901376: Fix upgrade issue with DNS controllerconfig field

cybertron:
The initial implementation of this had incorrect logic for handling
merges of configs. Because DNS is a new field, the existing one will
always be nil on upgrade. We need to remove the nil check for the
existing object so that it can be populated with the new object.


kikisdeliveryservice (approved):
Worked with @cybertron on this fix. As Baremetal didn't have an upgrade job it wasn't caught the first time.

This is correct now.

/lgtm
/approve

PR #2307, opened 2020-12-16
kubelet: Allow drop-in configuration of node labels and taints

abursavich:
**- What I did**

This allows users to configure node labels (e.g. node-role.kubernetes.io/infra, topology.kubernetes.io/region=foobar, topology.kubernetes.io/zone=foobar-a) and taints (node-role.kubernetes.io/infra=:NoSchedule) with systemd drop-ins before a node joins the cluster. This can prevent pods from scheduling on nodes that would otherwise be tainted after joining the cluster. This can also be used by (baremetal) provisioning systems to add role and topology labels to nodes.

**- How to verify it**

I confirmed that the kubelet's flag parsing methods allow empty taint and label values.

In the case of [taints](https://github.com/kubernetes/kubernetes/blob/v1.19.0/pkg/util/taints/taints.go#L103-L119), the flag value can be empty but comma-separated components may not be (e.g. `""` and `"node-role.kubernetes.io/infra=:NoSchedule"` are valid but `","` and `"node-role.kubernetes.io/infra=:NoSchedule,"` are not).

In the case of [labels](https://github.com/kubernetes/kubernetes/blob/v1.19.0/staging/src/k8s.io/component-base/cli/flag/configuration_map.go#L36-L49) both the flag value and comma-separated components may be empty (e.g. `""`, `","`, `"node-role.kubernetes.io/infra"`, and `"node-role.kubernetes.io/infra,"` are all valid).

Allow configuration of initial node labels and taints with systemd drop-in


rphillips:
@abursavich this looks reasonable. We will likely get to this PR at the beginning of the New Year. Happy Holidays.

/ok-to-test

abursavich:
The failures on that job don't seem relevant to this change, so I'm gonna assume that's a flake for now.

/retest

abursavich:
It looks like there are similar test failures on other recent PRs, including ones that have been merged. I'm gonna spare the resources and not rerun the job again.

abursavich:
Rebased to deal with merge conflicts. PTAL.

abursavich:
Again, the failing tests are unrelated to the change... I'll kick off another run, but I don't want to waste resources, so I probably won't keep doing it.

/retest

abursavich:
> @abursavich this looks reasonable. We will likely get to this PR at the beginning of the New Year. Happy Holidays.

@rphillips, please take a look :)

/honk

vrutkovs:
@rphillips mind having another look?

abursavich:
Please take a look! :)

abursavich:
:upside_down_face:

PR #2309, opened 2020-12-17
Bug 1895024: Restart ovs-configuration on-abnormal

juanluisvaladas:
We see that ovs-configuration may not start because some dependencies
failed. These dependencies are restarted on failure so we should
re-check later if it has started.

Should fix: #1895024
**- How to verify it**
Unfortunately I do not have a reliable reproducer and I'm not 100% certain that it will actually fix it.
However this should improve stability and I can't think of any scenario where this can cause problems.

Restart ovs-configuration on-abnormal failures.

**- What I did**
Force restarting ovs-configuration on-abnormal failures.

### DIFF
--- a/templates/common/_base/units/ovs-configuration.service.yaml
+++ b/templates/common/_base/units/ovs-configuration.service.yaml
@@ -17,6 +17,8 @@ contents: |
   ExecStart=/usr/local/bin/configure-ovs.sh {{.NetworkType}}
   StandardOutput=journal+console
   StandardError=journal+console
+  Restart=on-abnormal
+  RestartSec=90s
 
   [Install]
   WantedBy=network-online.target


juanluisvaladas:
/assign @trozet 
Tim, I can't think of any reason why this could potentially have problems, but I'd like you to have a look anyway.

kikisdeliveryservice:
Q: do we expect this ovn-step-registry to pass for this?

rbbratta:
This patch won't work.  

```
Dec 17 16:56:06 localhost systemd[1]: ovs-configuration.service: Service has Restart= setting other than no, which isn't allowed for Type=oneshot services. Refusing.
Dec 17 16:56:06 localhost systemd[1]: ovs-configuration.service: Cannot add dependency job, ignoring: Unit ovs-configuration.service has a bad unit file setting.
```

Restart for oneshot only works on newer systemds  https://github.com/systemd/systemd/pull/13754

rbbratta:
Not a systemd expert, but it looks like from the comments `# Need oneshot to delay kubelet` we need to delay kubelet.

It looks like we can re-run `/usr/local/bin/configure-ovs.sh` multiple times and it will detect existing config and not re-configure.

So maybe we could use `Type=exec` with `Restart=on-abnormal`.

`Type=exec` looks like it will wait until the process has execd before starting kubelet.

> The `exec` type is similar to simple, but the service manager will consider the unit started immediately after the main service binary has been executed. The service manager will delay starting of follow-up units until that point. (Or in other words: simple proceeds with further jobs right after `fork()` returns, while exec will not proceed before both `fork()` and `execve()` in the service process succeeded.)

```
[Service]
Type=exec
ExecStart=/usr/local/bin/configure-ovs.sh OpenShiftSDN
StandardOutput=journal+console
StandardError=journal+console
Restart=on-abnormal
```

PR #2310, opened 2020-12-17
Bug 1825417: Make the ctrcfg CR to MC mapping 1:1

umohnani8:
Fixes #1825417 (https://bugzilla.redhat.com/show_bug.cgi?id=1825417)

**- What I did**
Create a new MC object for each ctrcfg object created.
Before we were only creating 1 MC object for every ctrcfg
created, this would override the previouse ctrcfg CR options
causing a rollback to go all the way back to the default values
when the latest ctrcfg CR was deleted.
So if there was ctrcfg-1, ctrcfg-2 and we deleted ctrcfg-2, it would
skip over ctrcfg-1 and go back to the defaults.
Having a 1:1 mapping of ctrcfg to MC prevents this from happening.
So ctrcfg-1 will have mc-1 and ctrcfg-2 will have [mc-2](https://issues.redhat.com/browse/mc-2) - if ctrcfg-2
is deleted, the changes will roll back to mc-1, which is from ctrcfg-1
and not the initial defaults.

**- How to verify it**
Start a cluster and create multiple ctrcfg CRs, then delete them and see that the changes on the node rolls back to the previous ctrcfg CR and not directly to the defaults. Also, each ctrcfg CR will have its own MC and the suffix of the MC name will be stored as annotation in the ctrcfg CR.

Create a new MC for each ctrcfg object created.

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/clarketm/json"
@@ -446,8 +447,28 @@ func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.ContainerRuntimeConfig, err e
 	return err
 }
 
+// addAnnotation adds the annotions for a ctrcfg object with the given annotationKey and annotationVal
+func (ctrl *Controller) addAnnotation(cfg *mcfgv1.ContainerRuntimeConfig, annotationKey, annotationVal string) error {
+	annotationUpdateErr := retry.RetryOnConflict(updateBackoff, func() error {
+		newcfg, getErr := ctrl.mccrLister.Get(cfg.Name)
+		if getErr != nil {
+			return getErr
+		}
+		newcfg.SetAnnotations(map[string]string{
+			annotationKey: annotationVal,
+		})
+		_, updateErr := ctrl.client.MachineconfigurationV1().ContainerRuntimeConfigs().Update(context.TODO(), newcfg, metav1.UpdateOptions{})
+		return updateErr
+	})
+	if annotationUpdateErr != nil {
+		glog.Warningf("error updating the container runtime config with annotation key %q and value %q: %v", annotationKey, annotationVal, annotationUpdateErr)
+	}
+	return annotationUpdateErr
+}
+
 // syncContainerRuntimeConfig will sync the ContainerRuntimeconfig with the given key.
 // This function is not meant to be invoked concurrently with the same key.
+// nolint: gocyclo
 func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 	startTime := time.Now()
 	glog.V(4).Infof("Started syncing ContainerRuntimeconfig %q (%v)", key, startTime)
@@ -507,7 +528,7 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 	for _, pool := range mcpPools {
 		role := pool.Name
 		// Get MachineConfig
-		managedKey, err := getManagedKeyCtrCfg(pool, ctrl.client)
+		managedKey, err := getManagedKeyCtrCfg(pool, ctrl.client, cfg)
 		if err != nil {
 			return ctrl.syncStatusOnly(cfg, err)
 		}
@@ -552,6 +573,15 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 			mc, err = ctrlcommon.MachineConfigFromIgnConfig(role, managedKey, tempIgnCfg)
 			if err != nil {
 				return ctrl.syncStatusOnly(cfg, err, "could not create MachineConfig from new Ignition config: %v", err)
+			}
+			_, ok := cfg.GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
+			arr := strings.Split(managedKey, "-")
+			// If the MC name suffix annotation does not exist and the managed key value returned has a suffix, then add the MC name
+			// suffix annotation and suffix value to the ctrcfg object
+			if len(arr) > 4 && !ok {
+				if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
+					return ctrl.syncStatusOnly(cfg, err, "could not update annotation for containerRuntimeConfig")
+				}
 			}
 		}
 
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -2,10 +2,12 @@ package containerruntimeconfig
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/docker/reference"
@@ -19,6 +21,7 @@ import (
 	"github.com/openshift/runtime-utils/pkg/registries"
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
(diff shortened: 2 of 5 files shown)


umohnani8:
@yuqi-zhang @kikisdeliveryservice @rphillips PTAL
Once this is approved, I will pull in the changes for the kubeletconfig controller. Working on reworking the e2e tests for this as well, wanted to get the review process started while I get the tests updated :)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -553,6 +574,15 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 			if err != nil {
 				return ctrl.syncStatusOnly(cfg, err, "could not create MachineConfig from new Ignition config: %v", err)
 			}
+			_, ok := cfg.GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
+			arr := strings.Split(managedKey, "-")
+			// If the MC name suffix annotation does not exist and the maaged key value returned has a suffix, then add the MC name
```
kikisdeliveryservice, line 579: minor: maaged -> managed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -177,8 +180,54 @@ func getManagedKeyCtrCfgDeprecated(pool *mcfgv1.MachineConfigPool) string {
...
+		// Convert the suffix value to int so we can add 1 to it for the MC name for the latest ctrcfg object
+		suffixNum, err = strconv.Atoi(val)
+		if err != nil {
+			return "", fmt.Errorf("error converting %s to int: %v", val, err)
+		}
+	}
+	// The max suffix value that we can go till with this logic is 9 - this means that a user can create up to 10 different ctrcfg CRs.
+	// However, if there is a ctrcfg-1 mapping to mc-1 and ctrcfg-2 mapping to mc-2 and the user deletes ctrcfg-1, it will delete mc-1 but
+	// then if the user creates a ctrcfg-new it will map to mc-3. This is what we want as the latest ctrcfg created should be higher in priority
+	// so that those changes can be rolled out to the nodes. But users will have to be mindful of how many ctrcfg CRs they create. Don't think
```
kikisdeliveryservice, line 224: given this comment, instead of creating more why not just make them modifying the existing?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -177,8 +180,54 @@ func getManagedKeyCtrCfgDeprecated(pool *mcfgv1.MachineConfigPool) string {
...
+	// anyone should ever have the need to create 10 when they can simply update an existing ctrcfg unless it is to apply to another pool.
+	if suffixNum+1 > 9 {
```
kikisdeliveryservice, line 226: 9 feels arbitrary and surprising?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -553,6 +574,15 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+			// suffix annotation and suffix value to the ctrcfg object
+			if len(arr) > 4 && !ok {
```
kikisdeliveryservice, line 581: why is len 4? probably obvi but I dont understand

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -446,8 +447,28 @@ func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.ContainerRuntimeConfig, err e
 	return err
 }
 
+// addAnnotation adds the annotions for a ctrcfg object with the given annotationKey and annotationVal
+func (ctrl *Controller) addAnnotation(cfg *mcfgv1.ContainerRuntimeConfig, annotationKey, annotationVal string) error {
```
kikisdeliveryservice, line 451: what is an example of this key/value pair?

kikisdeliveryservice (commented):
Some initial questions..

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: The first MC name would be `99-generated-[pool]-containerruntime` and the following ones will have a suffix e.g `99-generated-[pool]-containerruntime-1` so we want to only add the suffix annotation if a suffix exists that is why we split at `-` and check if the length is greater than 4.

pkg/controller/container-runtime-config/helpers.go
umohnani8: Since the priority is alphanumeric, every time a new ctrcfg is created, the suffic name of the MC will be bumped up by 1. So here is an example of what they would look like:
```
ctrcfg-1 --> 99-generated-[pool]-containerruntime
ctrcfg-2 --> 99-generated-[pool]-containerruntime-1
ctrcfg-3 --> 99-generated-[pool]-containerruntime-2
etc
```

Now when we hit 9 and bump it up by 1, it will become 10, so based on the alphanumeric priority `-9` will be bigger than `-10` and that is why it doesn't work after 9. But, I don't see why anyone should ever have to make so many ctrcfg CRs.

pkg/controller/container-runtime-config/helpers.go
umohnani8: So you would want to create a new one if you are targeting a different pool, or if you want to modify some other options for a bit and then revert it back. We just want to give the users the option to make more than one ctrcfg CR.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
umohnani8: An example is the new suffix annotation key I added
```
MCNameSuffixAnnotationKey = "machineconfiguration.openshift.io/mc-name-suffix"
``` 
and its value will be the suffix from the MC name.
I made the function generic in the even we would want to add more annotations in the future.

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: Can we output to the commandline to let the user knows about this limit? ie if they reach 7 ctrcfgs we tell them they have 3 left and we will be deleting them at 10etc..

pkg/controller/container-runtime-config/helpers.go
kikisdeliveryservice: makes sense!

kikisdeliveryservice:
Overall concept of this seems sound (making it consistent with MCs) and it's correcting for what would be very suprising behavior to users:+1:

pkg/controller/container-runtime-config/container_runtime_config_controller.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: Hmm, I am not sure how to propagate a log to the command line. Anything that we print or log ends up in the MCC logs right now.

umohnani8:
Updated the e2e tests - creates 2 ctrcfg CRs and deletes one at a time to make sure the config rolls back to the previous ctrcfg and  not directly to the defaults.

pkg/controller/container-runtime-config/helpers.go
cgwalters: https://kubernetes.io/blog/2020/09/03/warnings/

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -177,8 +180,54 @@ func getManagedKeyCtrCfgDeprecated(pool *mcfgv1.MachineConfigPool) string {
...
+		return "", fmt.Errorf("the max MC name suffix value that can be used is 9, please delete some ctrcfg CRs to free up suffx values before a new ctrcfg can be created")
```
kikisdeliveryservice, line 227: Can we make the message more specific?
""the max MC name suffix value that can be used is 9, please delete some ctrcfg CRs to free up suffx values before a new ctrcfg can be created"

Can we say something like:
"Max number of supported ctrcfgs (9) has been reached. Please delete old or unused ctrcfgs before retrying"

Users don't know what mc suffixes are and don't really care, but they do know how many ctrcfgs they have... doesn't have to be that exact warning but something clearer would be great.

kikisdeliveryservice: Also to confirm, this _won't_ be output to the terminal? so if they have 9 and they try to create one, it'll just do nothing? that seems a little confusing.

kikisdeliveryservice (changes requested):

pkg/controller/container-runtime-config/helpers.go
umohnani8: The error should show up when a user does `oc describe ctrcfg [ctrcfgname]` but I will test and confirm that. And will make the message clearer

kikisdeliveryservice (approved):
Spoke to @umohnani8 and she's going to handle: https://github.com/openshift/machine-config-operator/pull/2310#discussion_r562086976 in a followup pr along with the kubeletcfg stuff shortly, so merging this as-is.

PR #2314, opened 2021-01-04
Bug 1891460: KubeletConfig: Allow only positive values for KubeReserved, SystemReserved, EvictionHard and EvictionSoft

harche:
Allow only positive values for KubeReserved, SystemReserved and EvictionHard

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1891460

**- What I did**
Validate if KubeReserved, SystemReserved and EvictionHard in KubeletConfig are positive or not. 

**- How to verify it**
https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c0

Allow only positive values for KubeReserved, SystemReserved, EvictionHard and EvictionSoft

### DIFF
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -149,18 +149,6 @@ func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
 		return fmt.Errorf("KubeletConfiguration: staticPodPath is not allowed to be set, but contains: %s", kcDecoded.StaticPodPath)
 	}
 
-	if kcDecoded.EvictionSoft != nil && len(kcDecoded.EvictionSoft) > 0 {
-		if kcDecoded.EvictionSoftGracePeriod == nil || len(kcDecoded.EvictionSoftGracePeriod) == 0 {
-			return fmt.Errorf("KubeletConfiguration: EvictionSoftGracePeriod must be set when evictionSoft is defined, evictionSoft: %v", kcDecoded.EvictionSoft)
-		}
-
-		for k := range kcDecoded.EvictionSoft {
-			if _, ok := kcDecoded.EvictionSoftGracePeriod[k]; !ok {
-				return fmt.Errorf("KubeletConfiguration: evictionSoft[%s] is defined but EvictionSoftGracePeriod[%s] is not set", k, k)
-			}
-		}
-	}
-
 	reservedResources := []v1.ResourceName{v1.ResourceCPU, v1.ResourceMemory, v1.ResourceEphemeralStorage}
 
 	if kcDecoded.KubeReserved != nil && len(kcDecoded.KubeReserved) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/controller/kubelet-config/helpers.go
```diff
@@ -170,8 +170,8 @@ func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
 				if err != nil {
 					return fmt.Errorf("KubeletConfiguration: invalid value specified for %s reservation in kubeReserved, %s", rr.String(), val)
 				}
-				if q.Sign() == -1 {
-					return fmt.Errorf("KubeletConfiguration: %s reservation value cannot be negative in kubeReserved", rr.String())
+				if q.Sign() < 0 || q.IsZero() {
```
sinnykumari, line 173: Do we want reserved value input > 0 ? 
Also see q.Sign() doc, return value of Sign() will be either -1, 0 or 1 https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L522

harche: Yes, we want the reserved value input > 0. 

I tried to mirror how the kubelet validates these parameters, 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/eviction/helpers.go#L221-L223

```golang
if quantity.Sign() < 0 || quantity.IsZero() {
	return nil, fmt.Errorf("eviction threshold %v must be positive: %s", signal, &quantity)
}
```

kikisdeliveryservice:
Looking at the underlying BZ I also see: "value greater than memory capacity" causes error... so this wouldn't close the bz fully.

see:https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c1

kikisdeliveryservice:
for now since this doesn't fully close the bz

/hold

harche:
> Looking at the underlying BZ I also see: "value greater than memory capacity" and "non-digital" causes error... so this wouldn't close the bz fully.
> 
> see:https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c1

We won't be supporting `Value greater than capacity` validation. I have closed a similar bug about that, https://bugzilla.redhat.com/show_bug.cgi?id=1881400. I have also highlighted that in the comment https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c6. 

Other than that, I have added more test cases for 'non-digital' values and also added fixes the validation for `KubeReserved`, `SystemReserved`, `EvicationHard` and `EvictionSoft`. 

/cc @rphillips

kikisdeliveryservice:
> > Looking at the underlying BZ I also see: "value greater than memory capacity" and "non-digital" causes error... so this wouldn't close the bz fully.
> > see:https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c1
> 
> We won't be supporting `Value greater than capacity` validation. I have closed a similar bug about that, https://bugzilla.redhat.com/show_bug.cgi?id=1881400. I have also highlighted that in the comment https://bugzilla.redhat.com/show_bug.cgi?id=1891460#c6.
> 
> Other than that, I have added more test cases for 'non-digital' values and also added fixes the validation for `KubeReserved`, `SystemReserved`, `EvicationHard` and `EvictionSoft`.
> 
> /cc @rphillips

Thanks for this @harche !!

/hold cancel

PR #2316, opened 2021-01-04
Bug 1865743: Add a prestart line to change ownership of openvswitch dir

abhat:
Temporary fix for making sure the ownership of openvswitch directory is set to `openvswitch` user till we fix https://github.com/coreos/rpm-ostree/issues/49.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/templates/common/_base/units/ovs-vswitchd.service.yaml
+++ b/templates/common/_base/units/ovs-vswitchd.service.yaml
@@ -4,3 +4,6 @@ dropins:
     contents: |
       [Service]
       Restart=always
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /var/lib/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /etc/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /run/openvswitch'


abhat:
/assign @cgwalters. I also thought about doing `DynamicUser=`, but I am not sure if using that I can actually specify the user to be `openvswitch`. It looked like setting that option, the system will create an unused uid/gid instead. 

Also, there are repercussions about using that option for persistent directories and how to set up StateDirectory/LogsDirectory etc. correctly without breaking existing assumptions about how openvswitch service works. Given that, this seemed to be a prudent choice.

abhat:
Ok so there is openvswitch.useropts file that seems to be setting the `OVS_USER_ID` and `OVS_USER_OPT` environment variables that the openvswitch.service file is using. The drop-in does change the permissions on the directories, but because the openvswitch process is running as a different group `hugetlbfs`, any new bridge created by ovs is done using that gid.

I don't know who owns that env file and how it gets populated during ignition (via some config). But it would seem to make sense to use the same env. variable in this PR potentially?

danielmellado:
The PR seems to be just applying the workaround mentioned in [this bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1865743), so overall LGTM

templates/common/_base/units/ovs-vswitchd.service.yaml
```diff
@@ -4,3 +4,6 @@ dropins:
     contents: |
       [Service]
       Restart=always
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /var/lib/openvswitch'
```
cgwalters, line 7: Are you sure about the leading `:` to only change the group?
There's quite a lot of syntatic sorcery going on here.
OK `$$` is to quote it from systemd's expansion, so the shell sees it.
The `##*:` is...I don't quite follow that, what is it doing?

abhat: I leveraged this sorcery from an existing ovs-vswitchd service file on a deployed cluster. I am assuming it works :).

PR #2317, opened 2021-01-04
Bug 1909943: check for scheduler support before setting

darkmuggle:
RHEL7 worker nodes do not support bfq. Rather than assuming the block
device supports bfq, this checks before applying the setting.

### DIFF
--- a/pkg/daemon/controlplane.go
+++ b/pkg/daemon/controlplane.go
@@ -35,8 +35,19 @@ func setRootDeviceSchedulerBFQ() error {
 		return err
 	}
 	schedulerContents := string(schedulerContentsBuf)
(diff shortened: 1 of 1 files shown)


darkmuggle:
Testing confirms that this works as expected -- bfq is used on RHCOS nodes, but not on RHEL 7 nodes.

sinnykumari:
LGTM.
/approve
/assign @cgwalters

cgwalters:
I think a cleaner fix is to fix this instead https://github.com/openshift/machine-config-operator/pull/2243#issuecomment-737541790

But no objections to this as is!

sinnykumari:
Let's backport this to 4.6 since we are using bfq there as well and it can impact rhel7 nodes in 4.6
/cherry-pick release-4.6

PR #2325, opened 2021-01-07
Bug 1913154: update.go: only set BFQ scheduler for masters

yuqi-zhang:
In https://github.com/openshift/machine-config-operator/pull/2243
we made it so that we switch to BFQ during updates, but unintentially
made it also apply to workers. This does not work on RHEL7 since it
breaks on both trying to set the scheduler as well as finding the root
device.

Change the behaviour back to master only, which also implies that only
CoreOS variants will do this.


darkmuggle (approved):
/approve
/lgtm

Based on prior PR comments, switching to `bfq` really only makes sense on the control plane.

yuqi-zhang:
Will test manually on 4.6 first

yuqi-zhang:
/hold
need to fix an error

sinnykumari:
okd-e2e-aws is known to be broken
/skip

PR #2326, opened 2021-01-07
Bug 1910738: [on-prem] fix nodeip-configuration.service template

patrickdillon:
**- What I did**
Updates the on-prem nodeip-configuration.service template so content is always written but enabled is toggled based on onPremPlatformAPIServerInternalIP. Prior to this change, if onPremPlatformAPIServerInternalIP is nil the enabled is true but content is empty. This change puts the template to be consistent with /pkg/templates/common/_base/units/nodeip-configuration.service.yaml.

**- How to verify it**
The product of the template can be verified in vSphere UPI CI (and another platform to check the opposite case).

Updates the on-prem nodeip-configuration.service template so content is always written but enabled is toggled based on onPremPlatformAPIServerInternalIP.

Discussion:
I believe this change will mitigate the issues in bz-1917038 but it is not clear to me why vSphere UPI CI does not encounter this issue. Comparing the [must-gather from a successful UPI CI run ](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ocp-4.7-e2e-vsphere-upi/1346946367452352512/artifacts/e2e-vsphere-upi/gather-must-gather/) vs the must gather from the BZ I see that the BZ cluster encounters the error in the machine-config-daemon and it looks like there is some sort of update being triggered:

```
2020-12-24T06:59:46.241456713Z I1224 06:59:46.241389    2389 update.go:588] Checking Reconcilable for config rendered-worker-c255d81fe61657c3062da9e2cfbaee99 to rendered-worker-36aba62b3fbb0fe978bc6a3521af0c3d
2020-12-24T06:59:46.268478900Z I1224 06:59:46.268447    2389 update.go:1844] Starting update from rendered-worker-c255d81fe61657c3062da9e2cfbaee99 to rendered-worker-36aba62b3fbb0fe978bc6a3521af0c3d: &{osUpdate:false kargs:false fips:false passwd:false files:true units:false kernelType:false extensions:false}
```
This seems to set off the writing of files which leads to the encountered error, but if anyone could shed some light on why this update happens in the BZ but not in CI that would be helpful.

cc @jcpowermac 
cc @bcrochet 

/test e2e-vsphere-upi

### DIFF
--- a/templates/common/on-prem/units/nodeip-configuration.service.yaml
+++ b/templates/common/on-prem/units/nodeip-configuration.service.yaml
@@ -1,7 +1,6 @@
 name: nodeip-configuration.service
-enabled: true
+enabled: {{if (onPremPlatformAPIServerInternalIP .)}}true{{else}}false{{end}}
 contents: |
-  {{ if (onPremPlatformAPIServerInternalIP .) -}}
   [Unit]
   Description=Writes IP address configuration so that kubelet and crio services select a valid node IP
   # This only applies to VIP managing environments where the kubelet and crio IP
@@ -45,4 +44,3 @@ contents: |
 
   [Install]
   WantedBy=multi-user.target
-  {{ end -}}


patrickdillon:
> @patrickdillon: An error was encountered searching for bug 1917038 on the Bugzilla server at https://bugzilla.redhat.com. No known errors were detected, please see the full error message for details.
> 
> Full error message.
> ` code 101: Bug #1917038 does not exist. `
> Please contact an administrator to resolve this issue, then request a bug refresh with `/bugzilla refresh`.

It doesn't https://bugzilla.redhat.com/show_bug.cgi?id=1910738?

/bugzilla refresh

kikisdeliveryservice:
Hey all, ci seems to be totally down right now, convo in 4-dev

patrickdillon:
`$ cat cluster-scoped-resources/machineconfiguration.openshift.io/machineconfigs/00-worker.yaml ` shows: 

```
 - contents: |
          [Unit]
          Description=Writes IP address configuration so that kubelet and crio services select a valid node IP
          # This only applies to VIP managing environments where the kubelet and crio IP
          # address picking logic is flawed and may end up selecting an address from a
          # different subnet or a deprecated address
          Wants=network-online.target
          After=network-online.target ignition-firstboot-complete.service
          Before=kubelet.service crio.service

          [Service]
          Type=oneshot
          # Would prefer to do Restart=on-failure instead of this bash retry loop, but
          # the version of systemd we have right now doesn't support it. It should be
          # available in systemd v244 and higher.
          ExecStart=/bin/bash -c " \
            until \
            /usr/bin/podman run --rm \
            --authfile /var/lib/kubelet/config.json \
            --net=host \
            --volume /etc/systemd/system:/etc/systemd/system:z \
            registry.apps.build01-us-west-2.vmc.ci.openshift.org/ci-op-xc6d35g2/stable@sha256:c95430e348f6bc6d6d9f7b6f9126be814332a5fb1c0d840d4b39ad7b07cca5cf \
            node-ip \
            set --retry-on-failure \
            <no value>; \
            do \
            sleep 5; \
            done"

          [Install]
          WantedBy=multi-user.target
        enabled: false
```
which looks good. Prior to this change, both CI and must-gather from the bz show:
```
      - contents: ""
        enabled: true
```

Jinyun Ma from QE has told me he is able to reproduce, so should be able to test.

@jcpowermac  does this look good to you?

@kikisdeliveryservice how do we do approval for this as it touches on-prem templates (albeit in a small way; do all platforms need to take a look or is this too small to merit attention? I will try to take a look at the CI tests to make sure everything looks good:
/test-e2e-openstack 
/test e2e-ovirt
/test e2e-metal-ipi

patrickdillon:
Actually `e2e-metal-ipi` already ran and passed, but it doesn't look like there's a must-gather for the test, so I can't check the manifest. Hopefully oVirt or OpenStack has must-gather as the manifest would be the same.

kikisdeliveryservice:
@patrickdillon let me ping openstack and baremetal on slack to get another :eye:

mandre (commented):
LGTM, let's just make sure the tests pass.

/test e2e-openstack

ashcrow:
Will this PR also fix https://bugzilla.redhat.com/show_bug.cgi?id=1909570?

patrickdillon:
> Will this PR also fix https://bugzilla.redhat.com/show_bug.cgi?id=1909570?

Yes. I marked 1909570 as a dupe of this BZ.

cybertron:
/lgtm

I'm not sure why this would just be coming up now and only on vsphere upi, but it does seem wrong to have an empty unit enabled. The downside of the fix is that it means the unit will be written on all platforms, but that seems like a lesser evil than having deployments break on other platforms. If it turns out to be a problem we can look for another solution.

kikisdeliveryservice:
Just waiting for that openstack test to finish

/approve
/hold

kikisdeliveryservice:
something seems wrong with the openstack job finishing but I see:

`2021/01/11 19:07:59 Container test in pod e2e-openstack completed successfully`

kikisdeliveryservice:
ill try one more time and if it stalls again ill just remove hold EOD

/test e2e-openstack

kikisdeliveryservice:
Finally!

/hold cancel

PR #2329, opened 2021-01-09
Bug 1911632: daemon/update: fix regression in realtime upgrades

darkmuggle:
During the extension refactor [1], the upgrade logic that installed the
realtime kernel RPM's was lost. 'rpm-ostree' does not have the ability
to do non-local replacements of installed overrides (the packages are
shipped as a bona-fida repo). The cleanest method for handling the
upgrade is to remove the override and the install the new ones; this
avoids having to calculate the rpm dependencies.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1155,12 +1155,11 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
 
 	defaultKernel := []string{"kernel", "kernel-core", "kernel-modules", "kernel-modules-extra"}
 	realtimeKernel := []string{"kernel-rt-core", "kernel-rt-modules", "kernel-rt-modules-extra", "kernel-rt-kvm"}
-	var args []string
 
 	dn.logSystem("Initiating switch from kernel %s to %s", canonicalizeKernelType(oldConfig.Spec.KernelType), canonicalizeKernelType(newConfig.Spec.KernelType))
 
 	if canonicalizeKernelType(oldConfig.Spec.KernelType) == ctrlcommon.KernelTypeRealtime && canonicalizeKernelType(newConfig.Spec.KernelType) == ctrlcommon.KernelTypeDefault {
-		args = []string{"override", "reset"}
+		args := []string{"override", "reset"}
 		args = append(args, defaultKernel...)
 		for _, pkg := range realtimeKernel {
 			args = append(args, "--uninstall", pkg)
@@ -1172,7 +1171,7 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
 
 	if canonicalizeKernelType(oldConfig.Spec.KernelType) == ctrlcommon.KernelTypeDefault && canonicalizeKernelType(newConfig.Spec.KernelType) == ctrlcommon.KernelTypeRealtime {
 		// Switch to RT kernel
-		args = []string{"override", "remove"}
+		args := []string{"override", "remove"}
 		args = append(args, defaultKernel...)
 		for _, pkg := range realtimeKernel {
 			args = append(args, "--install", pkg)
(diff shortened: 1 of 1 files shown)


darkmuggle:
Calling this WIP while I test.

pkg/daemon/update.go
```diff
@@ -1147,42 +1147,57 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+			return resetRealTimeOverride(fmt.Sprintf("Switching to kerneltype=%s", newConfig.Spec.KernelType))
+		}
+
+		// Upgrade of realtime kernel requires resetting and then adding back
+		// in the override. rpm-ostree only performs "local" replacements of overrides, which
+		// mean detecting the NVR of the package. In theory, we could just find the path to the
+		// rpm's in the extension repo, however it would bug prone due to dependencies.
 		if oldConfig.Spec.OSImageURL != newConfig.Spec.OSImageURL {
-			dn.logSystem("Updating rt-kernel packages on host: %+q", args)
-			_, err := runGetOut("rpm-ostree", args...)
```
darkmuggle, line 1181: This was the root-cause: `rpm-ostree` was called with an empty `[]string`.

sinnykumari: ah yeah passing empty args here is of no use and caused the bug. Simple way to fix it is just pass `update` as arg to rpm-ostree. Since kernel-rt packages are installed from extensions repo, package dependencies gets automatically handled.

pkg/daemon/update.go
```diff
@@ -1147,42 +1147,57 @@ func (dn *Daemon) switchKernel(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
 	defaultKernel := []string{"kernel", "kernel-core", "kernel-modules", "kernel-modules-extra"}
 	realtimeKernel := []string{"kernel-rt-core", "kernel-rt-modules", "kernel-rt-modules-extra", "kernel-rt-kvm"}
-	var args []string
 
 	dn.logSystem("Initiating switch from kernel %s to %s", canonicalizeKernelType(oldConfig.Spec.KernelType), canonicalizeKernelType(newConfig.Spec.KernelType))
 
-	if canonicalizeKernelType(oldConfig.Spec.KernelType) == ctrlcommon.KernelTypeRealtime && canonicalizeKernelType(newConfig.Spec.KernelType) == ctrlcommon.KernelTypeDefault {
-		args = []string{"override", "reset"}
+	// resetRealTimeOverride removes the realtime kernel override.
+	resetRealTimeOverride := func(msg string) error {
```
sinnykumari, line 1154: override reset is used only for base overrides (packages which is part of base OS). In our case we override regular kernel packages which is part of base OS. kernnel-rt packages are layered packages and are treated similar to any other layered packages. Layered packages should automatically updated from extensions repo when `rpm-ostree update` gets called.

sinnykumari:
Sigh, this regression occurred due to https://github.com/openshift/machine-config-operator/commit/adb9e707b0d0170d741cc40dec2eaa73aa201415#diff-349c0748c3f52201852d3027c29daf618b45300b949482728df6666a3f9ba245R963 . I blame copy paste

pkg/daemon/update.go
darkmuggle: > Sigh, this regression occurred due to [adb9e70#diff-349c0748c3f52201852d3027c29daf618b45300b949482728df6666a3f9ba245R963](https://github.com/openshift/machine-config-operator/commit/adb9e707b0d0170d741cc40dec2eaa73aa201415#diff-349c0748c3f52201852d3027c29daf618b45300b949482728df6666a3f9ba245R963) . I blame copy paste

Ah, you found the real culprit. `rpm-ostree`'s man page doesn't document `rpm-ostree update` (I went looking).

pkg/daemon/update.go
darkmuggle: Indeed, repushed :)

pkg/daemon/update.go
sinnykumari: yup, it is tricky

PR #2332, opened 2021-01-11
Bug 1913582: fix IsLikeTraditionalRhel7() to account for minor versions

kikisdeliveryservice:
Bug Summary: 
The VERSION_ID was 7.5 but our check IsLikeTraditionalRhel7() did not account for minor versions (only for "7" not for 7.x) and so returned false. This resulted in dn.loggerSupportsJournal being set to true which is the relevant value when selecting which logger to use. So instead of using storePendingStateLegacyLogger() (for cases where journald isn't avail) we used normal logger resulting in the --journald flag erroneously being used for a rhel7 node.

Also added a unit test for this func to keep it working going forward.  Pre-fix unit new test failed and returned false for 7.5, now it correctly returns true.

### DIFF
--- a/pkg/daemon/osrelease.go
+++ b/pkg/daemon/osrelease.go
@@ -38,6 +38,9 @@ func (os OperatingSystem) IsCoreOSVariant() bool {
 // yum based + kickstart/cloud-init (not Ignition).
 func (os OperatingSystem) IsLikeTraditionalRHEL7() bool {
 	// Today nothing else is going to show up with a version ID of 7
+	if len(os.VersionID) > 2 {
+		return strings.HasPrefix(os.VersionID, "7.")
+	}
 	return os.VersionID == "7"
 }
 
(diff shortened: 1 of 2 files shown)


ashcrow:
Thank you for working on this!

darkmuggle:
Change looks good so far.

darkmuggle (approved):
/lgtm
/approve
@kikisdeliveryservice thank you kindly for the quick fix

kikisdeliveryservice:
okd job is red so skipping

e2e-aws isn't looking very good today either..

/skip

ashcrow (approved):

kikisdeliveryservice:
No backport required, this VersionID check was introduced in 4.7 via https://github.com/openshift/machine-config-operator/pull/2242

kikisdeliveryservice:
no point of retesting rn

/hold

PR #2334, opened 2021-01-12
PoC of /etc/hosts.d functionality

cybertron:
As discussed in https://github.com/openshift/machine-config-operator/pull/2236 this is a PoC of how a /etc/hosts.d feature might work. Currently it is used to provide the api-int record for on-prem platforms.


cybertron:
/hold

It's a PoC and I'm not sure we're actually going to use this mechanism anymore. In designing the feature that needed it, we found some issues with using /etc/hosts for this. I wanted to push it because I'd already done most of the work on it and it might be useful for someone else. This logic should also be moved into MCD as discussed in the other PR about this: https://github.com/openshift/machine-config-operator/pull/2258

darkmuggle (commented):
Implicitly this creates a new path `/etc/hosts.d` that is used add hosts to `/etc/hosts` -- and I'm not sure that is the best path forward. `/etc/hosts.d` is not used by anything in RHCOS. I'm not wild using a block in `/etc/hosts`.

PR #2336, opened 2021-01-12
Bug 1917187: [release-4.5][vsphere] Remove prepend domain-name-servers option from…

cybertron:
… dhclient.conf

In this release, nameserver prepending is handled by a NetworkManager
dispatcher script. The existence of this option is causing render
errors on vsphere deployments, which causes some weird DHCP behavior.
All of the other on-prem platforms had already removed this option
in 4.5, so we should get rid of it here too.


cybertron:
This bug only exists in 4.5 because the file in question was removed completely in 4.6. What is the process to make bugzilla bot happy? Opening a bug against 4.6/4.7 doesn't make much sense, but maybe that's what needs to happen?

sinnykumari:
Steve may know the right way to proceed with having a valid BZ
/cc @ashcrow

cybertron:
If https://github.com/openshift/release/pull/16966 merges this should be unblocked.

kikisdeliveryservice:
@cybertron do you want this anymore?

cybertron:
Not sure. @jcpowermac is this a problem you've been seeing a lot?

sdodson:
/close
The linked BZ has been closed EOL

PR #2338, opened 2021-01-13
Bug 1913536: update.go: add broken symlink check + removal during unit enable

yuqi-zhang:
If unit enable fails, remove broken symlinks in multi-user.target.wants
and try again. This fixes a bug where enables would fail on cluster upgrades
with RHEL 7 nodes between 4.6 -> 4.7.

Context: before https://github.com/openshift/machine-config-operator/pull/2145,
the MCO hard coded a symlink from /etc/system/systemd/$UNIT to
/etc/systemd/system/multi-user.target.wants/$UNIT, which is not the case
for every unit and thus caused broken symlinks. On RHCOS/FCOS, the systemd
version is newer and is able to remove broken symlinks, but on RHEL 7 nodes,
it will not first attempt to remove broken symlinks and thus fails the
enable. As a workaround, this PR thus attempts to remove broken symlinks
when the first enable fails, and then try again. Successful FCOS/RHCOS upgrades
should not hit this, and failing ones would report full errors.

The error checking is perhaps a bit overkill but the original bug case should
only run through this logic once before it is fixed. Future errors are likely
actual errors and will be reported as such.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1432,6 +1432,9 @@ func (dn *Daemon) enableUnits(units []string) error {
 	args := append([]string{"enable"}, units...)
 	stdouterr, err := exec.Command("systemctl", args...).CombinedOutput()
 	if err != nil {
+		if !dn.os.IsLikeTraditionalRHEL7() {
+			return fmt.Errorf("error enabling units: %s", stdouterr)
+		}
 		// In RHEL7, the systemd version is too low, so it is unable to handle broken
 		// symlinks during enable. Do a best-effort removal of potentially broken
 		// hard coded symlinks and try again.
@@ -1461,7 +1464,7 @@ func (dn *Daemon) enableUnits(units []string) error {
 		}
 		stdouterr, err := exec.Command("systemctl", args...).CombinedOutput()
 		if err != nil {
-			return fmt.Errorf("error enabling unit: %s", stdouterr)
+			return fmt.Errorf("error enabling units: %s", stdouterr)
 		}
 	}
 	glog.Infof("Enabled systemd units: %v", units)
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1432,7 +1432,37 @@ func (dn *Daemon) enableUnits(units []string) error {
 	args := append([]string{"enable"}, units...)
 	stdouterr, err := exec.Command("systemctl", args...).CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("error enabling unit: %s", stdouterr)
+		// In RHEL7, the systemd version is too low, so it is unable to handle broken
+		// symlinks during enable. Do a best-effort removal of potentially broken
+		// hard coded symlinks and try again.
+		// See: https://bugzilla.redhat.com/show_bug.cgi?id=1913536
+		wantsPathSystemd := "/etc/systemd/system/multi-user.target.wants/"
+		for _, unit := range units {
+			unitLinkPath := filepath.Join(wantsPathSystemd, unit)
+			fi, fiErr := os.Lstat(unitLinkPath)
+			if fiErr != nil {
+				if !os.IsNotExist(fiErr) {
+					return fmt.Errorf("error trying to enable unit, fallback failed with %s (original error %s)",
+						fiErr, stdouterr)
+				}
+				continue
+			}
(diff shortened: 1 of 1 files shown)


pkg/daemon/update.go
```diff
@@ -1432,7 +1432,37 @@ func (dn *Daemon) enableUnits(units []string) error {
 	args := append([]string{"enable"}, units...)
 	stdouterr, err := exec.Command("systemctl", args...).CombinedOutput()
 	if err != nil {
```
darkmuggle, line 1434: Shouldn't we do a check for RHEL 7?

pkg/daemon/update.go
```diff
@@ -1432,7 +1432,37 @@ func (dn *Daemon) enableUnits(units []string) error {
...
-		return fmt.Errorf("error enabling unit: %s", stdouterr)
+		// In RHEL7, the systemd version is too low, so it is unable to handle broken
+		// symlinks during enable. Do a best-effort removal of potentially broken
+		// hard coded symlinks and try again.
+		// See: https://bugzilla.redhat.com/show_bug.cgi?id=1913536
+		wantsPathSystemd := "/etc/systemd/system/multi-user.target.wants/"
```
darkmuggle, line 1439: The enable code does not assume `multi-user.target.wants`; the unit could have a different target.

pkg/daemon/update.go
```diff
@@ -1432,7 +1432,37 @@ func (dn *Daemon) enableUnits(units []string) error {
 	args := append([]string{"enable"}, units...)
 	stdouterr, err := exec.Command("systemctl", args...).CombinedOutput()
```
darkmuggle, line 1433: The effect of this is that any failure to enable the units will trigger the check. Since you are changing on the on-disk state, you may have to do a `systemctl daemon-reload`. 

Under the axiom that RHEL 7 can't handle null-target symlinks, I wonder if:
- check for symlinks in `/etc/systed/system` (since you can't assume the unit is just for multi-user.target.wants)
- check if the target is dangling, remove it
- then systemctl-enable
- guard for RHEL 7 only

pkg/daemon/update.go
yuqi-zhang: that's another option, I could hard check for RHEL7 and remove broken symlinks

pkg/daemon/update.go
yuqi-zhang: Right, this attempts to fix ONLY `multi-user.target.wants`, because that's where we wrote broken symlinks back when we had hardcoding. This ONLY attempts to fix those, and ONLY in the list of enables, and other user error is not handled.

pkg/daemon/update.go
yuqi-zhang: > The effect of this is that any failure to enable the units will trigger the check. Since you are changing on the on-disk state, you may have to do a systemctl daemon-reload.

Right, and I propagate the old error just in case. Put another way I am just performing one extra "cleanup" when I fail the first time, so normal operation specifically doesn't hit this.

> check for symlinks in /etc/systed/system (since you can't assume the unit is just for multi-user.target.wants)

I don't want to do that because then we'd be potentially managing items **not defined in a machineconfig** which I do not want to happen

yuqi-zhang: Basically, if user error occurs, I still want to error as usual. I don't want this to actually fix incorrect unit definitions, for example, or if a user wrote a bad link with a `files` section config. I am only attempting to clean up bad links from our own code pre 4.7. I can try to make that clearer in the commit message if you think the approach makes sense

yuqi-zhang: Would you be ok if the logic was the same, but I only attempted this on RHEL7? Or if I parse the error specifically for "file exists" which is what the rhel 7 bug was? (somewhat hacky I feel)

yuqi-zhang:
Added a check for rhel

yuqi-zhang:
Rebased, all tests passed, tested locally with 4.6.6->4.6.11-> 4.7 nightly with this PR + rhel node, all errors were automatically fixed, upgrade was successful, and no dangling symlinks remain. This should be good to go (assuming we're ok with the above assumptions)

sinnykumari (approved):
Overall LGTM.
Will leave final approval on Ben.

darkmuggle:
/approve
/lgtm

After talking to Jerry offline, looks good to me.

PR #2339, opened 2021-01-13
Bug 1873593: fix error args paring causing unresolved string

QiWang19:
Fixed: 1873593 (https://bugzilla.redhat.com/show_bug.cgi?id=1873593)

**- What I did**
Fix wrapErrorWithCondition seleting the args when format the error message.
It should selsect the rest elements except the format string.
**- How to verify it**

### DIFF
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -208,7 +208,7 @@ func wrapErrorWithCondition(err error, args ...interface{}) mcfgv1.ContainerRunt
 	if len(args) > 0 {
 		format, ok := args[0].(string)
 		if ok {
-			condition.Message = fmt.Sprintf(format, args[:1]...)
+			condition.Message = fmt.Sprintf(format, args[1:]...)
 		}
 	}
 	return *condition
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -264,7 +264,7 @@ func wrapErrorWithCondition(err error, args ...interface{}) mcfgv1.KubeletConfig
 	if len(args) > 0 {
 		format, ok := args[0].(string)
 		if ok {
-			condition.Message = fmt.Sprintf(format, args[:1]...)
+			condition.Message = fmt.Sprintf(format, args[1:]...)
 		}
 	}
 	return *condition


kikisdeliveryservice:
ci is currently busted so not retesting. in the meantime:
/assign @umohnani8

mtrmac (commented):
I’m almost certainly being stupid, but I can’t see how this helps at all.

Looking at `wrapErrorCondition`, `condition.Message` is either `fmt.Sprintf("Error: %v", err)` if `err != nil && len(args) == 0`, or `fmt.Sprintf(args[0], args[1:])` in the other case (and if `args[0]` is a `string`.

I.e. if `args` is empty, `condition.Message` is just `err` without much context; if `args` is not empty, `condition.Message` is `args` and the text from `err` is not used at all — unless it is explicitly included in `args`. So, AFAICS, this PR should just result in `Message` fields that don’t contain `err` at all, a rather unhelpful outcome.

OTOH I can’t see how the `%v` in the bug report has not been interpreted as a format string, nor why the `could not add finalizers to ContainerRuntimeConfig` prefix has been added twice, so there definitely is something I’m missing.

So, for dumb people like me, how does this PR work?

kikisdeliveryservice:
same w as @mtrmac - could you also verify the new output as well?

/hold

QiWang19:
The logic in the wrapErrorWithCondition was not right. I have updated the fix. 
I can reproduce the bz error `%v` has not been interpreted as a format string. And confirmed after this patch the %v can be resolved.
```
$ oc describe ctrcfg/set-log-and-pid
// reproduce the bz bug
Status:
  Conditions:
    Last Transition Time:  2021-01-15T18:22:30Z
    Message:               could not add finalizers to ContainerRuntimeConfig: could not add finalizers to ContainerRuntimeConfig: %v
    Status:                False
  Observed Generation:     1

// result after applying this patch
Status:
  Conditions:
    Last Transition Time:  2021-01-15T20:40:49Z
    Message:               could not add finalizers to ContainerRuntimeConfig: ==== this is a test error
    Status:                False
```

mtrmac (commented):
Oh. Good catch, I have completely overlooked that.

Please update `pkg/controller/kubelet-config` as well.

mtrmac (commented):
/lgtm .

kikisdeliveryservice (approved):

kikisdeliveryservice:
@umohnani8 PTAL

umohnani8:
Nice catch, /lgtm

kikisdeliveryservice (approved):

kikisdeliveryservice:
for some reason the 2 lgtms above didn't get added to this pr, let's try again

/lgtm

PR #2341, opened 2021-01-14
handle interfaces with multiple ips

aojea:
The network manager script wasn´t discriminating by IP address, hence,
interfaces with multiple addresses broke the scripts.

This was partially fixed by https://github.com/openshift/machine-config-operator/pull/2312
~However, this still carries the problem that the parsing of the interface output return multiples fields, and those should be discriminated by the IP address received as parameter.~

ups, Sorry , I misread the script, I think that it only misses the `jq --argi, at least it failed to me without it

https://bugzilla.redhat.com/show_bug.cgi?id=1914250

### DIFF
--- a/templates/common/baremetal/files/NetworkManager-static-dhcpv6.yaml
+++ b/templates/common/baremetal/files/NetworkManager-static-dhcpv6.yaml
@@ -11,8 +11,8 @@ contents:
         exit 0
     fi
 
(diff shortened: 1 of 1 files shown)


aojea:
@cybertron please check if this will solve this bug too https://bugzilla.redhat.com/show_bug.cgi?id=1910165

aojea:
hmm, I think that the double quoting works too
my  bad sorry, I somehow  missed it

PR #2342, opened 2021-01-14
Bug 1882674: use environment file for proxy settings

darkmuggle:
Proxies that are url encoded represent sepcial characters with a '%'.
However, systemd uses '%' as control character in unit files and
dropins [1]. This moves the environment varialbes into a file and uses a
EnviromentFile stanza in the units which cannot safely contain '%' and
avoids adding complicated logic to the MCO for mangling the proxy
passwords (detecting if its URL encoded and special character handling).

Closes: 1882674

PR #2343, opened 2021-01-14
Bug 1882674: Escape systemd proxy strings

kwoodson:
Fixes: #1882674

Systemd does not allow single % characters followed by
a number of a character.  Systemd requires escaping '%'
by adding preceding percent to appear as '%%'. This is
an attepmt to escape the proxy string when applied in
/etc/systemd path.

**- What I did**
I added a simple template rendering function to double escape the `%` characters with `%%`.

**- How to verify it**
(Tested in Azure)
1. Create a cluster
2. Create a separate proxy host in a different resource group.
3. Peer the vnets
4. Setup squid proxy on the proxy vm with authentication and a user that has a `%` in the password.
5. Edit the cluster's proxy object (oc edit proxy cluster) and specify the proxy cluster's IP with auth.
```
spec:
  httpProxy: http://proxy:this%is%my%password@10.3.0.4:3128
  readinessEndpoints:
  - http://www.google.com
  trustedCA:
    name: user-ca-bundle
```
6. Verify that the unit files are **not** showing up in dmesg as failed and verify the password is escaped propertly (includes `%%`).

An attempt to properly escape % characters for systemd proxy files.

### DIFF
--- a/templates/common/_base/units/crio.service.yaml
+++ b/templates/common/_base/units/crio.service.yaml
@@ -6,12 +6,12 @@ dropins:
       {{if .Proxy -}}
       [Service]
       {{if .Proxy.HTTPProxy -}}
-      Environment=HTTP_PROXY={{.Proxy.HTTPProxy}}
+      Environment=HTTP_PROXY='{{.Proxy.HTTPProxy}}'
       {{end -}}
       {{if .Proxy.HTTPSProxy -}}
-      Environment=HTTPS_PROXY={{.Proxy.HTTPSProxy}}
+      Environment=HTTPS_PROXY='{{.Proxy.HTTPSProxy}}'
       {{end -}}
       {{if .Proxy.NoProxy -}}
-      Environment=NO_PROXY={{.Proxy.NoProxy}}
+      Environment=NO_PROXY='{{.Proxy.NoProxy}}'
       {{end -}}
       {{end -}}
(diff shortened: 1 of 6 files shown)


pkg/controller/template/render.go
```diff
@@ -523,3 +524,13 @@ func urlPort(u string) (interface{}, error) {
...
 }
+
+// escapeSystemdPercent is a template function that replaces the % character in a string to %%
+func escapeSystemdPercent(u string) (interface{}, error) {
+	percentReplacer := strings.NewReplacer("%", "%%")
+	if strings.Contains(u, "%") {
+		u = percentReplacer.Replace(u)
+	}
+
+	return u, nil
```
darkmuggle, line 535: Curious, why not?
```suggestion
	return strings.Replace(u, "%", "%%")
```
It seems much faster than allocating a new `func`, checking, then running the new `func.`

darkmuggle (commented):
This might not be safe since it mutates the entire URL. Consider a proxy string of:
`https://user@foobar:test.proxy.com/foo#meh` -- there are other special characters that could cause breakage, not just the %.

darkmuggle:
I chatted with @kwoodson on Slack, and the better path is to use a single `'`. The problem with mutating is that we would need to split the parts of the URL and we would need to handle all special characters. I think the safest path is to blindly pass through the value the user sends instead of mutating the string and prevents a future of dealing with most special characters (although we still need to deal with `'` too....)

kikisdeliveryservice:
since we now have https://github.com/openshift/machine-config-operator/pull/2342

/hold

darkmuggle:
@kwoodson I presume that its okay to close this PR?

PR #2344, opened 2021-01-15
Bug 1916711: vendor: update kubernetes and dependencies to 1.20.0

sinnykumari:
* Updated kube deps to 1.20.0
* Fixup as a result of kube update


templates/master/01-master-kubelet/_base/files/kubelet.yaml
```diff
@@ -32,6 +32,8 @@ contents:
       IPv6DualStack: true
 {{- end}}
       LegacyNodeRoleBehavior: false
+      # Will be removed in future openshift/api update https://github.com/openshift/api/commit/c8c8f6d0f4a8ac4ff4ad7d1a84b27e1aa7ebf9b4
+      RemoveSelfLink: false
```
sinnykumari, line 36: @deads2k with kube 1.20 bump, some of the kubelet-config unit tests were failing. Can you please let us know if the kubelet related changes in this PR are ok?

kikisdeliveryservice: @rphillips PTAL

kikisdeliveryservice:
codes changes are kubelet based so will defer approval until @sjenning @rphillips take a look/approves

/assign @sjenning @rphillips

templates/master/01-master-kubelet/_base/files/kubelet.yaml
rphillips: RemoveSelfLink is being removed in 1.21... I don't think we need this change.

rphillips: I retract that comment after more review. +1

PR #2347, opened 2021-01-18
Bug 1907636: [vsphere, openstack] Add missing proxy settings to resolv-prepender

cybertron:
Apparently in 4.5 the vsphere and openstack platforms had separate
templates for the resolv-prepender. This caused the backports for
the proxy settings to only apply to either the master or the worker
instead of both. We should have those settings everywhere, so this
change adds the appropriate env vars.

### DIFF
--- a/templates/master/00-master/openstack/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/master/00-master/openstack/files/NetworkManager-resolv-prepender.yaml
@@ -7,6 +7,19 @@ contents:
     set -eo pipefail
     IFACE=$1
     STATUS=$2
+
+    {{if .Proxy -}}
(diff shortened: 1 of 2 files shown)


mandre (commented):
/retest
LGTM, thanks

cybertron:
OpenStack passed and vsphere failed on docker.io rate limiting but it looks like the cluster deployed successfully, so I think we're in good shape.

patrickdillon:
LGTM 
cc @jcpowermac

jcpowermac:
lgtm as well

cc: @jstuever

kikisdeliveryservice:
ok since we have a quorum :smile: I'll just do the honors

/approve
/lgtm

PR #2348, opened 2021-01-18
Refactor kernel argument handling

darkmuggle:
The command to add or remove a kernel tunable is the same command with
different flags. This change simplifies and removes duplicate logic.

This also fixes a problem with using the `/proc/cmdline` for checking if
a kernel commandline in use. Running `rpm-ostree kargs` is safer since
is shows the kargs that will be used next boot instead of the current
boot (which could be ephemeral to boot).

Finally, due to difficulties with legacy systems, this introduces a
NodeUpdaterClient of `notCoreOSClient` to start putting differences
between CoreOS and RHEL 7. The first stub out is for the Kernel
Commandline handling.

### DIFF
--- a/pkg/daemon/kernelargs.go
+++ b/pkg/daemon/kernelargs.go
@@ -3,7 +3,6 @@ package daemon
 import (
 	"bufio"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
@@ -50,33 +49,30 @@ func isArgTunable(arg string) (bool, error) {
 	return false, nil
 }
 
-// isArgInUse checks to see if the argument is already in use by the system currently
-func isArgInUse(arg, cmdLinePath string) (bool, error) {
-	if cmdLinePath == "" {
-		cmdLinePath = CmdLineFile
-	}
-	content, err := ioutil.ReadFile(cmdLinePath)
+// isKernelArgInUse checks to see if the argument is already in use by the system currently
+func isKernelArgInUse(arg string) (bool, error) {
+	nu := NewNodeUpdaterClient()
+	checkable, err := nu.GetKernelArgs()
 	if err != nil {
 		return false, err
 	}
 
-	checkable := string(content)
-	if strings.Contains(checkable, arg) {
-		return true, nil
+	for _, v := range checkable {
+		if strings.HasPrefix(v, arg) {
+			return true, nil
+		}
 	}
 	return false, nil
 }
 
 // parseTuningFile parses the kernel argument tuning file
-func parseTuningFile(tuningFilePath, cmdLinePath string) ([]types.TuneArgument, []types.TuneArgument, error) {
+func parseTuningFile(tuningFilePath string) ([]types.TuneArgument, []types.TuneArgument, error) {
 	addArguments := []types.TuneArgument{}
 	deleteArguments := []types.TuneArgument{}
 	if tuningFilePath == "" {
 		tuningFilePath = KernelTuningFile
 	}
-	if cmdLinePath == "" {
-		cmdLinePath = CmdLineFile
-	}
+
 	// Read and parse the file
 	file, err := os.Open(tuningFilePath)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/daemon/not_cos.go
```diff
@@ -0,0 +1,60 @@
...
+var (
+	// notCoreOSClient is a NodeUpdaterClient.
+	_ NodeUpdaterClient = &notCoreOSClient{}
+
+	// ErrNotCoreosVariant is a generic error for un-supported tasks on legacy systems.
+	ErrNotCoreosVariant = errors.New("operating system is not a CoreOS Variant")
+)
+
+// GetBootedDeployment returns the booted demployment.
+func (noCos *notCoreOSClient) GetBootedDeployment() (*RpmOstreeDeployment, error) {
```
darkmuggle, line 33: Note: none of this is "safe" yet..I imagine that we should just return errors for all this. To get started, I thought the least disruptive/buggy approach is just to pass through a `RpmOstreeClient` for now.

pkg/daemon/kernelargs.go
```diff
@@ -93,63 +90,62 @@ func parseTuningFile(tuningFilePath, cmdLinePath string) ([]types.TuneArgument,
...
-			glog.V(2).Infof(`skipping malformed line in %s: "%s"`, tuningFilePath, line)
 		}
 	}
 	return addArguments, deleteArguments, nil
 }
 
 // UpdateTuningArgs executes additions and removals of kernel tuning arguments
-func UpdateTuningArgs(tuningFilePath, cmdLinePath string) (bool, error) {
-	if cmdLinePath == "" {
-		cmdLinePath = CmdLineFile
```
darkmuggle, line 149: Note: the use of `/proc/cmdline` is incorrect as it shows the kernel arguments for the current boot, not the kernel arguments for the next boot. `rpm-ostree` updates the BLSConfig and reads from it.

darkmuggle:
/unhold

Tested and confirmed to work. Test case:
- Create a cluster
- Added kargs via a machine-config
- Waited for rollout and then manually removed some kargs
- Delete the machine-config
- Waited and confirmed the kargs was removed

```
1915 update.go:1862] Running rpm-ostree [kargs --append=loglevel=7 --append=meh=bar --append=baz]
1915 rpm-ostree.go:279] Running captured: rpm-ostree kargs --append=loglevel=7 --append=meh=bar --append=baz
2175 update.go:1862] Running rpm-ostree [kargs --delete=loglevel=7 --delete=baz]
```

The effect of this PR is to confirm that any Kargs being "removed" truly is active.

pkg/daemon/kernelargs.go
```diff
@@ -50,33 +49,31 @@ func isArgTunable(arg string) (bool, error) {
...
+	checkable, err := nu.GetKernelArgs()
 	if err != nil {
 		return false, err
 	}
 
-	checkable := string(content)
-	if strings.Contains(checkable, arg) {
-		return true, nil
+	for _, v := range checkable {
+		if strings.Contains(v, arg) {
```
darkmuggle, line 62: /hold

I should fix this -- `strings.Contains` is wrong, it should be `strings.HasPrefix`. Otherwise `meh=baz` would match keys `meh` and `baz`.

sinnykumari:
Reason of bug doesn't seems to be related to kernelArgs, most likely it is related to https://bugzilla.redhat.com/show_bug.cgi?id=1916169#c7 .  Have created a minor fixup that came out from bug in https://github.com/openshift/machine-config-operator/pull/2349 .

Although this PR does some refactoring which can be a good discussion point. Possibly when master is on 4.8 ?

sinnykumari:
Also, wanted to highlight that applying tunable kargs through a file is a bit of legacy and it existed when we didn't have kargs support through MachineConfig. It is available only to keep backward compatibility and OKD needs it for tweaking FCOS default kargs . We don't encourage anybody to use it since kargs support added through MachineConfig.

darkmuggle:
> It is available only to keep backward compatibility and OKD needs it for tweaking FCOS default kargs .

Interesting...I was trying to figure out where it was actually used. If we're retaining if for backward compatibility, I would much rather that we work out an interface for dealing the matrix of OCP/RHCOS and OKD on FCOS (or maybe CentOS streams?)

I like the idea of targeting this for 4.8.

PR #2349, opened 2021-01-20
Bug 1918440: daemon: add check before updating kernelArgs

sinnykumari:
Today we are unnecessarily calling updateKernelArguments() even when there is no kargs to apply. This doesn't fixes bug 1916169 but discovered in https://bugzilla.redhat.com/show_bug.cgi?id=1916169#c8.
If needed, can file a new bug but for now getting it in as part of BZ#1916169

Update: Created a new bug to track it separately.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -992,12 +992,12 @@ func parseKernelArguments(kargs []string) []string {
 	return parsed
 }
 
-// generateKargsCommand performs a diff between the old/new MC kernelArguments,
+// generateKargs performs a diff between the old/new MC kernelArguments,
 // and generates the command line arguments suitable for `rpm-ostree kargs`.
 // Note what we really should be doing though is also looking at the *current*
 // kernel arguments in case there was drift.  But doing that requires us knowing
 // what the "base" arguments are. See https://github.com/ostreedev/ostree/issues/479
-func generateKargsCommand(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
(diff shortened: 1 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -451,8 +451,10 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 	}()
 
 	// Apply kargs
-	if err := dn.updateKernelArguments(oldConfig, newConfig); err != nil {
-		return err
+	if mcDiff.kargs {
```
darkmuggle, line 454: Doesn't `dn.updateKerenlArguments` do a check already? 
https://github.com/openshift/machine-config-operator/blob/82929b72f6901650c5849ba4afe7ccdbed48b17c/pkg/daemon/update.go#L1018-L1021

sinnykumari: not really, generateKargsCommand(0 function just deletes all kargs available in old Rendered MC and appened all from new rendered MC. variable name is a bit misleading, perhaps it is hanging around from our earlier implementation. We have simplified kargs logic over time due to various bugs we hit, see https://github.com/openshift/machine-config-operator/pull/2014/files . I will update variable name so that we don't get confused

darkmuggle: Ah, okay.... let me know when you're ready for another review.

sinnykumari: updated

darkmuggle (approved):
/lgtm
/approve

PR #2350, opened 2021-01-20
Bug 1915080: set max time to keepalived health checks

iamemilio:
We noticed a linearly growing number of tcp connections in haproxy in all standing clusters,
that are caused by healthcheck connections remaining open. adding a max time equal to the interval
at which healthchecks are run should prevent a growth in open tcp connections.

Fixes Bug 1915080

### DIFF
--- a/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
+++ b/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
@@ -17,7 +17,7 @@ contents:
     # api to take the VIP. This isn't preferred because it means all api
     # traffic will go through one node, but at least it keeps the api available.
     vrrp_script chk_ocp_lb {
-        script "/usr/bin/timeout 1.9 /etc/keepalived/chk_ocp_script.sh"
+        script "/etc/keepalived/chk_ocp_script.sh"
         interval 2
         weight 20
         rise 3
@@ -25,7 +25,7 @@ contents:
     }
 
     vrrp_script chk_ocp_both {
-        script "/usr/bin/timeout 1.9 /etc/keepalived/chk_ocp_script_both.sh"
+        script "/etc/keepalived/chk_ocp_script_both.sh"
         interval 2
         # Use a smaller weight for this check so it won't trigger the move from
         # bootstrap to master by itself.
@@ -37,7 +37,7 @@ contents:
     # TODO: Improve this check. The port is assumed to be alive.
     # Need to assess what is the ramification if the port is not there.
     vrrp_script chk_ingress {
-        script "/usr/bin/timeout 0.9 /usr/bin/curl -o /dev/null -Lfs http://localhost:1936/healthz/ready"
+        script "/usr/bin/curl -o /dev/null --max-time 0.9 -Lfs http://localhost:1936/healthz/ready"
         interval 1
         weight 50
     }
@@ -52,7 +52,7 @@ contents:
         advert_int 1
         {{`{{if .EnableUnicast}}`}}
         unicast_src_ip {{`{{.NonVirtualIP}}`}}
-        unicast_peer {            
+        unicast_peer {
             {{`{{range .LBConfig.Backends -}}
             {{if ne $nonVirtualIP .Address}}{{.Address}}{{end}}
             {{end}}`}}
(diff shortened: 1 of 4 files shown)


mandre (commented):
Can you try to also use the same `--max-time` flag to `chk_ingress` check, in `templates/master/00-master/on-prem/files/keepalived-keepalived.yaml`?

templates/master/00-master/on-prem/files/keepalived-script-both.yaml
```diff
@@ -3,4 +3,4 @@ path: "/etc/kubernetes/static-pod-resources/keepalived/scripts/chk_ocp_script_bo
 contents:
   inline: |
     #!/bin/bash
-    /usr/bin/curl -o /dev/null -kLfs https://localhost:{{`{{ .LBConfig.LbPort }}`}}/readyz && [ -e /var/run/keepalived/iptables-rule-exists ] || /usr/bin/curl -kLfs https://localhost:{{`{{ .LBConfig.ApiPort }}`}}/readyz
+    /usr/bin/curl -o /dev/null --max-time 1.9 -kLfs https://localhost:{{`{{ .LBConfig.LbPort }}`}}/readyz && [ -e /var/run/keepalived/iptables-rule-exists ] || /usr/bin/curl --max-time 1.9 -kLfs https://localhost:{{`{{ .LBConfig.ApiPort }}`}}/readyz
```
mandre, line 6: There are actually 2 checks, so we should have `--max-time` < 1 to ensure this fits in the 2 second interval at which the tests are run.

iamemilio: wait they are run concurrently (&&), i think that 1.9 seconds for each should be ok

iamemilio: jk, I always thought that && let you run each task as a thread, but aparently its a conditional chain of operations :zany_face:

iamemilio:
/retest

It looks like this did have a slight effect on things, but was likely not the only culprit. I think we should consider putting this through.

iamemilio:
/assign @yboaron @mandre 

It looks like it lowered the overall connections slightly..

latest netstat

iamemilio:
Compare to this graph, which is from the first PR, which should have actually changed nothing

netstat default

iamemilio:
also note that the timeouts [don't exist in 4.6](https://github.com/openshift/machine-config-operator/blob/release-4.6/templates/master/00-master/openstack/files/openstack-keepalived-script-both.yaml)!

cybertron:
/lgtm

It's a little concerning that we're timing out often enough for this to cause problems, but the change should be equivalent to what we had before and if it fixes a problem then let's do it.

iamemilio:
I am going to run a cluster over night to see if I get a better performance pattern than what Martin and Alex got. Ill post them below for reference.

Martin's:
image

Alex's:
bad tcp

templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
```diff
@@ -37,7 +37,7 @@ contents:
     # TODO: Improve this check. The port is assumed to be alive.
     # Need to assess what is the ramification if the port is not there.
     vrrp_script chk_ingress {
-        script "/usr/bin/timeout 0.9 /usr/bin/curl -o /dev/null -Lfs http://localhost:1936/healthz/ready"
+        script "/usr/bin/timeout 0.9 /usr/bin/curl -o /dev/null --max-time 1 -Lfs http://localhost:1936/healthz/ready"
```
mdbooth: Isn't this redundant, as timeout will kill the the curl before 1 second in any case. Any explicit cleanup curl does after 1 seconds would presumably not run.

mdbooth: Ah, you remove the timeout in the next commit anyway. If you update probably best to elide this commit.

iamemilio: NP haha. Yeah. I'm still working on proving this out, but my hypothesis right now is that syskilling the curl rather than using the built in timeout may not be closing the tcp connection. Yet to prove that though.

yboaron:
@iamemilio  I think  https://github.com/openshift/baremetal-runtimecfg/pull/117 should address the open TCP connections

iamemilio:
this does not actually solve the problem, closing

PR #2351, opened 2021-01-20
Bug 1918415: check dropins for nil-pointers

darkmuggle:
Check dropin content for nil pointer and content.


sinnykumari:
Thanks for fixing this!
/lgtm

PR #2355, opened 2021-01-21
Bug 1918677: templates: use valid binary in AuthorizedKeysCommand

vrutkovs:
Setting `AuthorizedKeysCommand` with invalid path breaks ssh on hosts:
```
Jan 21 10:55:36 ip-10-0-137-119 sshd[130647]: error: AuthorizedKeysCommand path is not absolute
Jan 21 10:55:36 ip-10-0-137-119 sshd[130647]: Connection closed by authenticating user core 10.0.188.12 port 41306 [preauth]
```
This PR would set it to `/bin/true`

Reverts #2087

### DIFF
--- a/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
+++ b/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
@@ -3,4 +3,4 @@ path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
 contents:
   inline: |
     # disable key lookup from ~/.ssh/authorized_keys.d/ on FCOS
-    AuthorizedKeysCommand none
+    AuthorizedKeysCommand /bin/true


sinnykumari:
was it always broken or any recent FCOS update caused it?

vrutkovs:
>was it always broken or any recent FCOS update caused it?

Always was broken, but we didn't get a chance to properly identify it :/

LorbusChris:
I don't think we want to remove that snippet entirely, since we still want to disable lookup from the authorized_keys.d dir.
Can the `none` command be replaced by a valid no-op command, e.g. `/usr/bin/false`?

/cc @bgilbert

vrutkovs:
Not sure why, but ssh'ing to masters from workers worked when I removed that setting (via `oc debug node/...`)

bgilbert:
Yeah, I'd try setting it to `/bin/true` instead.

vrutkovs:
okay, updated, PTAL

kikisdeliveryservice:
hi @vrutkovs !

do we expect that okd-e2e-aws test to pass, looking at past runs it seems kind of flaky right?

vrutkovs:
>do we expect that okd-e2e-aws test to pass, looking at past runs it seems kind of flaky right?

It should pass, yeah, must be flakes. This change doesn't affect e2e tests, just helps us debug failures.

/retest

bgilbert:
Content LGTM.  Commit message and PR title need an update.

kikisdeliveryservice:
> Content LGTM. Commit message and PR title need an update.

agree with ben :+1:  hold so can be fixed

/hold

templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
```diff
@@ -3,4 +3,4 @@ path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
 contents:
   inline: |
     # disable key lookup from ~/.ssh/authorized_keys.d/ on FCOS
-    AuthorizedKeysCommand none
+    AuthorizedKeysCommand /bin/true
```
LorbusChris, line 6: can you try explicitly setting it to nil/empty?

```suggestion
    AuthorizedKeysCommand
```

LorbusChris: this should re-instate the default

PR #2356, opened 2021-01-21
Bug 1919272: [release-4.6] NM resolve prepender: support appending a nameserver for systemd-resolved

vrutkovs:
FCOS 33 doesn't use NM to manage resolv.conf. Instead it uses systemd-resolved.
The script should not replace /etc/resolv.conf in this case, but instead create a dropin for systemd-resolved.

Cherry-pick of https://github.com/openshift/machine-config-operator/pull/2359 on release-4.6

PR #2359, opened 2021-01-22
Bug 1919271: [on-prem] NM resolve prepender: support appending a nameserver for systemd-resolved

vrutkovs:
FCOS 33 doesn't use NM to manage resolv.conf. Instead it uses systemd-resolved.

The script should not replace /etc/resolv.conf in this case, but instead create a dropin for systemd-resolved

4.6 cherry-pick: https://github.com/openshift/machine-config-operator/pull/2356


yuqi-zhang (approved):
Code looks fine, I'm not sure if the `Only leave the first 3 nameservers in /etc/resolv.conf` moving into the `$NAMESERVER_IP` condition affects anything. It'd be good for the metal team to have the final lgtm

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -52,14 +52,21 @@ contents:
             "{{ onPremPlatformIngressIP . }}")
         DOMAIN="{{.DNS.Spec.BaseDomain}}"
         if [[ -n "$NAMESERVER_IP" ]]; then
-            >&2 echo "NM resolv-prepender: Prepending 'nameserver $NAMESERVER_IP' to /etc/resolv.conf (other nameservers from /var/run/NetworkManager/resolv.conf)"
-            sed -e "/^search/d" \
-                -e "/Generated by/c# Generated by KNI resolv prepender NM dispatcher script\nsearch $DOMAIN\nnameserver $NAMESERVER_IP" \
-                /var/run/NetworkManager/resolv.conf > /etc/resolv.tmp
+            source /etc/os-release
+            if [[ "$NAME" == "Fedora" ]]; then
+                echo "[Resolve]" > /etc/systemd/resolved.conf.d/kni.conf
```
darkmuggle, line 57: https://www.freedesktop.org/software/systemd/man/resolved.conf.html

I *think* we should use something like `60-kni.conf` to make ordering easier.

vrutkovs: Good idea, updated

vrutkovs:
>I'm not sure if the Only leave the first 3 nameservers in /etc/resolv.conf moving into the $NAMESERVER_IP condition affects anything

Nameservers would be cleanup on RHCOS when new interface is up and NAMESERVER_IP is resolved (previously we'd clean those regardless of NAMESERVER_IP result). I don't think this would regress in any corner case really

kikisdeliveryservice:
holding for on-prem reviews

/hold

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -52,14 +52,21 @@ contents:
             "{{ onPremPlatformIngressIP . }}")
         DOMAIN="{{.DNS.Spec.BaseDomain}}"
         if [[ -n "$NAMESERVER_IP" ]]; then
-            >&2 echo "NM resolv-prepender: Prepending 'nameserver $NAMESERVER_IP' to /etc/resolv.conf (other nameservers from /var/run/NetworkManager/resolv.conf)"
-            sed -e "/^search/d" \
-                -e "/Generated by/c# Generated by KNI resolv prepender NM dispatcher script\nsearch $DOMAIN\nnameserver $NAMESERVER_IP" \
-                /var/run/NetworkManager/resolv.conf > /etc/resolv.tmp
+            source /etc/os-release
+            if [[ "$NAME" == "Fedora" ]]; then
```
mandre, line 56: What sets the $NAME variable? Don't you also need to check the version of Fedora?

Perhaps a better check would be to look for the use of systemd-resolved, so that there's no surprise when RHCOS switches to it.

vrutkovs: > What sets the $NAME variable?

Its being set after `source /etc/os-release`

> Don't you also need to check the version of Fedora?

We could add this later, but OKD 4.6 uses F33+, so systemd-resolved would certainly be used

> Perhaps a better check would be to look for the use of systemd-resolved

That was my initial approach, but its not clear if there is a nice way to do that - check that /etc/resolv.conf is a symlink? systemd-resolved service is running?

iamemilio:
/lgtm 

I think this is good to go

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
mandre: I think the most robust is `systemctl is-active --quiet systemd-resolved`

Fedosin (approved):
/lgtm

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
vrutkovs: For some reason that didn't work. I'll have another look and follow up on that later

Fedosin (approved):
/lgtm

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
mandre: Yeah, don't worry too much about it. I think we'll know where to look if suddenly DNS stops working.

mandre (commented):
/lgtm

kikisdeliveryservice:
I see a bunch of openstack approvals,

@bcrochet @celebdor @jcpowermac PTAL

/test e2e-vsphere

kikisdeliveryservice:
(pinged in slack channels)

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
celebdor: I think checking for the link should be fine.

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -52,14 +52,21 @@ contents:
...
+                echo "[Resolve]" > /etc/systemd/resolved.conf.d/60-kni.conf
+                echo "DNS=$NAMESERVER_IP" >> /etc/systemd/resolved.conf.d/60-kni.conf
+                echo "Domains=$DOMAIN" >> /etc/systemd/resolved.conf.d/60-kni.conf
```
celebdor, line 59: Wouldn't it be nicer to have a single heredoc or an echo with "\n" instead of three separate echo?

vrutkovs: The most cleanest way would be creating this file unconditionally - and it would be caught up in RHEL when it does the switch. WDYT?

vrutkovs: I don't have a preference how to write the file, seems echo with '\n' would the cleanest option. I don't mind rewriting it if you insist

cybertron:
/lgtm

Seems like everyone is in agreement on the functionality of this. If we want to modify how things get done we can always do that in a followup.

vrutkovs:
Skipping optional tests
/skip

PR #2360, opened 2021-01-23
daemon: unit-testable rpm-ostree wrapper and karg improvements

darkmuggle:
This is a first-step towards unit-testable rpm-ostree commands by
creating a mockable rpm-ostree wrapper. To demonstrate the idea, kernel
arguments were plumbed into the new facility.

The NodeUpdateClient is the top level interface used by the daemon for
doing node updates and configuration. However, not all `rpm-ostree`
commands were using the CoreOS implementation of `RpmOstreeClient.` To
facilate mockability, all `rpm-ostree` commands are now routed
through `RpmOstreeClient`. Mocking of `rpm-ostree` command is done
through the new struct field of `runRpmOstreeFunc`. Unit tests, can use
this faculty to mock output of `rpm-ostree`.

The first unit-test that uses this mocking feature is kernel arguments.
`rpm-ostree kargs --delete` will only delete kargs that it knows about,
and attempting to delete a run-time or missing karg results in an
error. By mocking `rpm-ostree` the logic is of adding, removing and
updating kernel arguments is provable.

Finally, this does introduce a behavior change. Previously if an admin
adjusted the kargs outside of the MCD and upgrade would fail due to the
kargs being missing. The MCD will now tolerate the missing kargs and
apply the expected state. Unit tests have been added deal with
expecting, albeit runtime missing kargs.

**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/daemon/not_cos.go
+++ b/pkg/daemon/not_cos.go
@@ -1,0 +1,64 @@
+package daemon
+
+import (
+	"io/ioutil"
+
+	"github.com/golang/glog"
+)
+
+// NotCoreOSClient is a wrapper around RpmOstreeClient that implements NodeUpdaterClient
+// safely on unsupported and legacy Operating systems.
+type notCoreOSClient struct{}
+
+// NewNodeUpdaterClientNotCoreOS returns a NodeUpdaterClient for traditional RHEL.
+func NewNodeUpdaterClientNotCoreOS() NodeUpdaterClient {
+	glog.Warning("Operating System is not a CoreOS Variant. Update functionality is disabled.")
+	return &notCoreOSClient{}
+}
+
+// notCoreOSClient is a NodeUpdateClient
+var _ NodeUpdaterClient = &notCoreOSClient{}
+
+// GetBootedDeployment returns the booted demployment.
+func (noCos *notCoreOSClient) GetBootedDeployment() (*RpmOstreeDeployment, error) {
+	return &RpmOstreeDeployment{}, nil
+}
+
+// Rebase calls rpm-ostree status
+func (noCos *notCoreOSClient) Rebase(a, b string) (bool, error) {
+	glog.Info("Rebase is not supported on this system.")
+	return false, nil
+}
+
+// GetStatus returns the rpm-ostree status
+func (noCos *notCoreOSClient) GetStatus() (string, error) {
+	return "", nil
+}
+
+// GetBootedOSImageURL returns the rpmOstree Booted OS Image
+func (noCos *notCoreOSClient) GetBootedOSImageURL() (string, string, error) {
+	return "", "", nil
+}
+
+// GetKernelArgs returns the real kernel arguments since we can't use rpmOstree
+func (noCos *notCoreOSClient) GetKernelArgs() ([]string, error) {
+	content, err := ioutil.ReadFile(CmdLineFile)
+	if err != nil {
+		return nil, err
+	}
+	return quoteSpaceSplit(string(content)), nil
+}
+
+// SetKernelArgs sets the kernel arguments
+func (noCos *notCoreOSClient) SetKernelArgs([]KernelArgument) (string, error) {
+	return "unsupported", nil
+}
+
+// RemovePendingDeployment is not supported on non-CoreOS machines.
+func (noCos *notCoreOSClient) RemovePendingDeployment() error {
+	return nil
+}
+
+func (noCos *notCoreOSClient) RunRpmOstree(noun string, args ...string) ([]byte, error) {
+	return nil, nil
+}
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -20,6 +20,9 @@ const (
 	numRetriesNetCommands = 5
 	// Pull secret.  Written by the machine-config-operator
 	kubeletAuthFile = "/var/lib/kubelet/config.json"
+
+	// realRpmOstreeCmd is the binary name for rpmOstree
+	realRpmOstreeCmd = "/usr/bin/rpm-ostree"
 )
 
 // rpmOstreeState houses zero or more RpmOstreeDeployments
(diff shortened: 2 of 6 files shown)


darkmuggle:
Throwing this up for early review. If the team likes the approach for unit-testable executables such as `rpm-ostree`, I can work out more of this.  Obviously, this won't help with the CI situation but can help make testing faster by testing the logic and not needing an actual cluster. In the case of `kargs` I was able to prove my prior PR was logically incorrect due to splitting strings. 

Anyway, if the team likes this approach I can continue with the work.

darkmuggle:
Okay, this is ready now that 4.8 is opened.

ashcrow (approved):
I really like having the functionality to be able to unit test! There is a lot in this PR however and I don't feel confident in my ability to be the only reviewer :-). I recommend having at least two other folks take a look as well. :+1:

pkg/daemon/not_cos.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"io/ioutil"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+)
+
+// NotCoreOSClient is a wrapper around RpmOstreeClient that implements NodeUpdaterClient
+// safely on unsupported and legacy Operating systems.
```
cgwalters, line 12: Well...that seems harsh :smile:   I usually say "yum-based" or "traditional".

cgwalters (approved):
Overall looks sane to me, but I only gave this a surface level review, didn't deep dive into the new logic.  Makes total sense to make some of this unit testable of course!

cgwalters:
Related to this we have https://github.com/coreos/rpm-ostree/issues/2389 and I hope to add Go rpm-ostree bindings at some point and take some of the trickier interaction logic there.  That might naturally lead to making things easier to mock.

That said I think by far the biggest MCO simplification would be if rpm-ostree implemented https://github.com/coreos/rpm-ostree/issues/2326 (this is also related to ye olde https://github.com/openshift/machine-config-operator/issues/1190 ) - basically the MCO would just translate the MachineConfig bits into e.g. `/etc/rpm-ostree.conf.d/mco.conf` and run `rpm-ostree rebuild` - no need to manually diff packages/kargs etc, let rpm-ostree take care of reconciling all that.

Anyways I think the rpm-ostree Go bindings would be extracted from this repo, and it makes total sense from that PoV to have things like "reconcile kernel args" be part of that.

pkg/daemon/rpm-ostree.go
```diff
@@ -245,17 +297,82 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
+type KernelArgument struct {
+	Operation int
+	Name      string
+}
+
+// runRpmOStree wraps the rpm-ostree command. Unless mocked, this
+// is the rpmOstreeCommander that is used by RpmOstreeClient.
+func runRpmOstree(noun string, args ...string) ([]byte, error) {
+	if len(args) > 2 {
+		return nil, fmt.Errorf("rpm-ostree command is malformed, not enough arguments")
```
yuqi-zhang, line 321: if anything shouldn't this be "too many arguments?"

pkg/daemon/rpm-ostree.go
```diff
@@ -245,17 +297,82 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
+	}
+	glog.Infof("Executing cmd: '%s %s'", realRpmOstreeCmd, args)
+	out, err := runGetOut(realRpmOstreeCmd, args...)
```
yuqi-zhang, line 324: I think because you don't pass the actual `noun` string here, any commands you pass to it will fail (since its not the full command), which in turn causes the very first invocation of `GetBootedDeployment` to fail since instead of
`rpm-ostree status --json` you give it `rpm-ostree --json`.

yuqi-zhang: I checked a daemon pod and sure enough it fails with

```F0221 21:54:09.287205   15334 start.go:150] Failed to initialize single run daemon: error reading osImageURL from rpm-ostree: error running /usr/bin/rpm-ostree --json```

I feel like since `noun` isn't referenced at all the golint should have caught it... weird

File: pkg/daemon/rpm-ostree.go [hunk 4]
```diff
@@ -245,17 +297,82 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
+	if err != nil {
+		glog.Errorf("'%s %v' failed to run: %v:\n %s", realRpmOstreeCmd, args, string(out), err)
+	}
+	return out, err
+}
+
+// SetKernelArgs sets kernel arguments on an RPM OStree systeR
```
yuqi-zhang, line 331: `RPM OStree systeR` -> `rpm-ostree system`? Or was this referring to something else

pkg/daemon/update.go
```diff
@@ -1766,11 +1745,6 @@ func (dn *Daemon) updateSSHKeys(newUsers []ign3types.PasswdUser) error {
 
 // updateOS updates the system OS to the one specified in newConfig
 func (dn *Daemon) updateOS(config *mcfgv1.MachineConfig, osImageContentDir string) error {
-	if !dn.os.IsCoreOSVariant() {
```
yuqi-zhang, line 1769: So this does introduce a difference in behaviour I think. Before we just silently accepted the os as "updated" always on RHEL nodes (not an error if you try to "update OS"), but by removing this and passing it off to the notCOS interface you error instead.

yuqi-zhang (changes requested):
I am generally in favour of splitting it up into interfaces, however with this PR I was unable to test the details due to an error (see below comments).

Also in general there are some points that may change behaviour (e.g. not allowing you to update OS now, whereas before it was a silent no-op, see below) that would block rhel nodes from updating. I think we should be very careful not to change that explicit behaviour.

pkg/daemon/rpm-ostree.go
darkmuggle: Oof, yeah, I had that worked out but not pushed. The latest push should resolve this pain.

darkmuggle:
@yuqi-zhang I dropped the errors for now to preserve the behavior. If this merges this week (no pressure -- it merges when it merges or not) I'll follow up with PR's that build on it.

pkg/daemon/rpm-ostree.go
```diff
@@ -245,17 +297,117 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
+// is the rpmOstreeCommander that is used by RpmOstreeClient.
+func runRpmOstree(noun string, args ...string) ([]byte, error) {
+	errNotEnoughArgs := errors.New("rpm-ostree command does not have enough arguments")
+	errTooManyArgs := errors.New("rpm-ostree command called with too many arguments")
+
+	// Perform some seriously basic command validation on the rpm-ostree
+	// commands. These checks are not meant to be exhaustive.
+	switch noun {
+	// Argless commands
+	case "cancel", "rollblack":
```
yuqi-zhang, line 327: rollblack -> rollback

pkg/daemon/rpm-ostree.go
```diff
@@ -245,17 +297,117 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
+	// Dumb checks for package operations.
+	case "override", "uninstall", "install":
+		if len(args) == 0 {
+			return nil, errNotEnoughArgs
+		}
+
+	// known error conditions
+	case "":
+		return nil, errNotEnoughArgs
+	default:
```
yuqi-zhang, line 351: "status" is still not supported in this switch case right?

yuqi-zhang: Yeah with this PR I get `Failed to initialize single run daemon: error reading osImageURL from rpm-ostree: unsupported command 'rpm-ostree status'`

darkmuggle:
/hold (again)

Putting this back into hold state pending validation on RHEL 7.

pkg/daemon/rpm-ostree.go
darkmuggle: I reworked this to include full command validation, but this will need some serious testing.

darkmuggle:
Withdrawling as stale.

PR #2361, opened 2021-01-25
Bug 1915235: add imagefs.inodesFree to resourceFields

rphillips:
**- What I did**
This is preventing upgrades in certain cases where the imagefs.inodesFree resource field is used.

**- How to verify it**
Upgrade from 4.6.12 to 4.7.nightly with the following custom resource:

```
spec:
  kubeletConfig:
    evictionHard:
      imagefs.available: 10%
      imagefs.inodesFree: 5%
      memory.available: 200Mi
      nodefs.available: 5%
      nodefs.inodesFree: 4%
    evictionPressureTransitionPeriod: 0s
    evictionSoft:
      imagefs.available: 15%
      imagefs.inodesFree: 10%
      memory.available: 500Mi
      nodefs.available: 10%
      nodefs.inodesFree: 5%
    evictionSoftGracePeriod:
      imagefs.available: 1m30s
      imagefs.inodesFree: 1m30s
      memory.available: 1m30s
      nodefs.available: 1m30s
      nodefs.inodesFree: 1m30s
    imageGCHighThresholdPercent: 80
    imageGCLowThresholdPercent: 75
    imageMinimumGCAge: 5m
    maxPods: 240
    podsPerCore: 80
  machineConfigPoolSelector:
    matchLabels:
      custom-kubelet: small-pods
```

### DIFF
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -183,7 +183,7 @@ func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
 		}
 	}
 
-	resourceFields := []string{"memory.available", "nodefs.available", "nodefs.inodesFree", "imagefs.available"}
+	resourceFields := []string{"memory.available", "nodefs.available", "nodefs.inodesFree", "imagefs.available", "imagefs.inodesFree", "pid.available"}
 
 	if kcDecoded.EvictionSoft != nil && len(kcDecoded.EvictionSoft) > 0 {
 		if kcDecoded.EvictionSoftGracePeriod == nil || len(kcDecoded.EvictionSoftGracePeriod) == 0 {


pkg/controller/kubelet-config/helpers.go
```diff
@@ -183,7 +183,7 @@ func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
 		}
 	}
 
-	resourceFields := []string{"memory.available", "nodefs.available", "nodefs.inodesFree", "imagefs.available"}
+	resourceFields := []string{"memory.available", "nodefs.available", "nodefs.inodesFree", "imagefs.available", "imagefs.inodesFree"}
```
yuqi-zhang, line 186: Not the most familiar with the flags but reading https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/ there's also a `pid.available`, maybe it's worth adding that as well for futureproof purposes? Or do you foresee that not being used

rphillips: Updated. Thanks

yuqi-zhang (approved):
This should fix the issue, it looks like this was introduced in https://github.com/openshift/machine-config-operator/commit/1eb4bd3fca7ad32bf74f436f494c431000e13716, so I guess we never checked the correctness of these flags before.

One last concern is how this error gets bubbled up (didn't find anything in MCC logs), might be good to investigate so the MCO can properly report this error, but that's another issue entirely

QiWang19 (commented):
LGTM

PR #2363, opened 2021-01-25
Bug 1920027: daemon: handle zero-length dropins/units

darkmuggle:
This fixes a problem were a previous dropin/unit is replaced with zero
length file by ensuring that drops/units have actual content.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1414,12 +1414,17 @@ func checkV3Units(units []ign3types.Unit) error {
 				content = *u.Dropins[j].Contents
 			}
 
-			// Backwards compatibility check: the new behavior is to noop when a drop-in for zero length.
-			// However, to maintain backwards compatibility, we allow existing zero length files to exist.
-			if err := checkFileContentsAndMode(path, []byte(content), defaultFilePermissions); err != nil {
-				if content == "" && os.IsNotExist(err) {
+			// As of 4.7 we now remove any empty defined dropins, check for that first
+			if _, err := os.Stat(path); content == "" && err != nil {
+				if os.IsNotExist(err) {
 					continue
 				}
+				return err
+			}
+
+			// To maintain backwards compatibility, we allow existing zero length files to exist.
+			// Thus we are also ok if the dropin exists but has no content
+			if err := checkFileContentsAndMode(path, []byte(content), defaultFilePermissions); err != nil {
 				return err
 			}
 		}
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1406,7 +1406,20 @@ func checkV3Units(units []ign3types.Unit) error {
 	for _, u := range units {
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
-			if err := checkFileContentsAndMode(path, []byte(*u.Dropins[j].Contents), defaultFilePermissions); err != nil {
+
+			var content string
+			if u.Dropins[j].Contents == nil {
+				content = ""
+			} else {
+				content = *u.Dropins[j].Contents
+			}
+
+			// Backwards compatibility check: the new behavior is to noop when a drop-in for zero length.
+			// However, to maintain backwards compatibility, we allow existing zero length files to exist.
+			if err := checkFileContentsAndMode(path, []byte(content), defaultFilePermissions); err != nil {
+				if content == "" && os.IsNotExist(err) {
+					continue
+				}
 				return err
 			}
 		}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1502,12 +1502,23 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 	for _, u := range units {
 		// write the dropin to disk
 		for i := range u.Dropins {
+			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
 			if u.Dropins[i].Contents == nil || *u.Dropins[i].Contents == "" {
 				glog.Infof("Dropin for %s has no content, skipping write", u.Dropins[i].Name)
+				if _, err := os.Stat(dpath); err != nil {
+					if os.IsNotExist(err) {
+						continue
+					}
+					return err
+				}
+				glog.Infof("Removing %q, updated file has zero length", dpath)
+				if err := os.Remove(dpath); err != nil {
+					return err
+				}
 				continue
 			}
+
 			glog.Infof("Writing systemd unit dropin %q", u.Dropins[i].Name)
-			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
 			if _, err := os.Stat("/usr" + dpath); err == nil &&
 				dn.os.IsCoreOSVariant() {
 				if err := createOrigFile("/usr"+dpath, dpath); err != nil {


pkg/daemon/daemon.go
```diff
@@ -1406,6 +1406,9 @@ func checkV3Units(units []ign3types.Unit) error {
 	for _, u := range units {
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
+			if u.Dropins[j].Contents == nil || *u.Dropins[j].Contents == "" {
```
yuqi-zhang, line 1409: Doesn't this technically allow the error in the bug to slip through in the future? i.e. if I deleted a dropin (or made it empty), we never verify if the file itself was properly deleted?

pkg/daemon/update.go
```diff
@@ -1469,12 +1469,22 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 	for _, u := range units {
 		// write the dropin to disk
 		for i := range u.Dropins {
+			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
+
 			if u.Dropins[i].Contents == nil || *u.Dropins[i].Contents == "" {
 				glog.Infof("Dropin for %s has no content, skipping write", u.Dropins[i].Name)
```
crawford, line 1475: Can you update this to mention that we are going to try to remove a file? That will give us some more context if the remove operation fails.

pkg/daemon/update.go
```diff
@@ -1469,12 +1469,22 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
-				continue
+				err := os.Remove(dpath)
```
yuqi-zhang, line 1476: While I think this should work, I'm leaning towards doing it one of two ways:

1. delete this in the deleteStaleData function
2. write an empty file to the location instead of deleting it

If we do 2 I think the benefit is potentially we don't affect verification. What do you think?

yuqi-zhang (commented):
Some questions/concerns

pkg/daemon/update.go
darkmuggle: Ah, indeed. `deleteStaleData` make sense here.

pkg/daemon/update.go
darkmuggle: Sure

pkg/daemon/daemon.go
darkmuggle: Moving this over the `deleteStaleData` should address that.

pkg/daemon/update.go
```diff
@@ -1391,6 +1391,17 @@ func (dn *Daemon) deleteStaleData(oldIgnConfig, newIgnConfig *ign3types.Config)
 				}
 				glog.Infof("Removed stale systemd dropin %q", path)
 			}
+
```
yuqi-zhang, line 1394: Ah, sorry, I think I confused both of us when I recommended doing it here. I think its better to `remove` in the writeUnits location and let this function do what it used to do. This function operates under the assumption of a diff, so its not the greatest place to remove "empty" units, also:

1. the logic here I think operators on `u` which is the old config (the one being updated from, not to), so if the unit was modified in the later one, this would wrongly remove it since it just bases it on the old unit definition
2. mucking with noOrig/Orig here is somewhat tricky here. We probably have to do it anyways in the other section. Example scenario: you had a default dropin shipped in the OS you want to disable, and you do so by writing an empty dropin over it with the same name. Then at some point you delete the empty dropin to re-enable the old dropin, we should correctly restore from the previous on-disk default (if you don't foresee this ever happening, we can ignore it maybe)
3. in the writeUnits function, its easier to keep backwards compatibility. An example of this with the first point would be, if we had a non-changing dropin definition that was empty in the first place, writeUnits can guarentee it doesn't write the file and remove any outstanding ones, if its defined, since it re-writes all unit definitions anyways

Let me know if you want to discuss in detail any of the points I raised. Sorry if its a bit confusing, I'm thinking through edge cases so I may be incorrect on these.

yuqi-zhang (changes requested):

pkg/daemon/update.go
darkmuggle: Let's roll with the first PR (re-pushed, btw). The edge cases are worth considering. I think backward compatibility is critical. I'll prioritize getting some unit tests here so that we can work through the edge cases.

pkg/daemon/update.go
```diff
@@ -1502,12 +1502,22 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 	for _, u := range units {
 		// write the dropin to disk
 		for i := range u.Dropins {
+			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
 			if u.Dropins[i].Contents == nil || *u.Dropins[i].Contents == "" {
 				glog.Infof("Dropin for %s has no content, skipping write", u.Dropins[i].Name)
+				if _, err := os.Stat(dpath); err != nil {
+					if _, ok := err.(*os.PathError); ok {
```
yuqi-zhang, line 1509: Minor point: correct me if I'm wrong but I think you are checking to see whether the file exists, and if it doesn't, just continue right?

For situations that err is not PathError (or I guess if we want to be specific, IsNotExist) should we just return the error here? Otherwise it will go to the remove below and may or may not cause other issues.

darkmuggle: Updated to `os.IsNotExist`.
 
I agree the logic is a little wonky...so changed. 

The check I wanted was:
- look to see if the file is missing, if so continue
- log a message about removing (@crawford asked for the message)
- then remove

pkg/daemon/daemon.go
```diff
@@ -1406,6 +1406,9 @@ func checkV3Units(units []ign3types.Unit) error {
...
+				continue
```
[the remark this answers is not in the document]
darkmuggle: Updated.

yuqi-zhang (commented):
Overall LGTM now, just one minor nit

pkg/daemon/daemon.go
```diff
@@ -1406,7 +1406,20 @@ func checkV3Units(units []ign3types.Unit) error {
...
+			var content string
+			if u.Dropins[j].Contents == nil {
+				content = ""
+			} else {
+				content = *u.Dropins[j].Contents
+			}
+
+			// Backwards compatibility check: the new behavior is to noop when a drop-in for zero length.
+			// However, to maintain backwards compatibility, we allow existing zero length files to exist.
+			if err := checkFileContentsAndMode(path, []byte(content), defaultFilePermissions); err != nil {
```
yuqi-zhang, line 1419: This doesn't return a IsNotExist since it Wraps the error as follows:

```
	fi, err := os.Lstat(filePath)
	if err != nil {
		return errors.Wrapf(err, "could not stat file %q", filePath)
	}
```
A separate if above this to check for empty file is probably needed

yuqi-zhang (changes requested):
Oh the e2e-gcp op failures is not a flake (not counting for the ones with not enough resources),
```
E0127 23:49:47.635348    2139 writer.go:135] Marking Degraded due to: unexpected on-disk state validating against rendered-node-log-level-50227eb514d3634e5728a485fcfee1a3: could not stat file "/etc/systemd/system/pivot.service.d/10-mco-default-env.conf": lstat /etc/systemd/system/pivot.service.d/10-mco-default-env.conf: no such file or directory
```

See below

yuqi-zhang:
The OP/upgrade tests passed, I think this should be good

sinnykumari:
/lgtm
e2e-aws-proxy has been always red, perhaps this ci test need some tweaking

PR #2365, opened 2021-01-25
Bug 1920027: use separate dropin files for kubelet

darkmuggle:
Test commit to see if rendering/merging dropins may be broken.

### DIFF
--- a/templates/common/_base/units/kubelet.service-proxy.yaml
+++ b/templates/common/_base/units/kubelet.service-proxy.yaml
@@ -1,0 +1,8 @@
+name: kubelet.service
+dropins:
+  - name: 10-mco-default-env.conf
+    contents: |
+      {{if .Proxy -}}
+      [Service]
+      EnvironmentFile=/etc/mco/proxy.env
+      {{end -}}
--- a/templates/common/_base/units/kubelet.service.yaml
+++ b/templates/common/_base/units/kubelet.service.yaml
@@ -1,11 +1,5 @@
 name: kubelet.service
 dropins:
-  - name: 10-mco-default-env.conf
-    contents: |
-      {{if .Proxy -}}
-      [Service]
-      EnvironmentFile=/etc/mco/proxy.env
-      {{end -}}
   - name: 10-mco-default-madv.conf
     contents: |
       [Service]


yuqi-zhang (commented):
Looking at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2365/pull-ci-openshift-machine-config-operator-master-e2e-aws/1353835083844292608/artifacts/e2e-aws/gather-extra/machineconfigs.json I think this worked! (search for kubelet.service, you can see that both of them are listed as dropins).

Although this doesn't answer the question of why having them in the same file doesn't work, that can be investigated another time. We can try running this against a proxy job just to be safe but I think I am +1 on this.

darkmuggle:
> Looking at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2365/pull-ci-openshift-machine-config-operator-master-e2e-aws/1353835083844292608/artifacts/e2e-aws/gather-extra/machineconfigs.json I think this worked! (search for kubelet.service, you can see that both of them are listed as dropins).
> 
> Although this doesn't answer the question of why having them in the same file doesn't work, that can be investigated another time. We can try running this against a proxy job just to be safe but I think I am +1 on this.

Right @yuqi-zhang, I think that we need to get to the root of the merging problem and _why_ this causing the unpleasant bug. That is something that I'll track down when I can incept more time.

yuqi-zhang (approved):
/lgtm

I don't see any risk with this change so we should be good

yuqi-zhang:
Hmm, in the recent proxy runs, I still see
```
Jan 29 14:51:56 ip-10-0-82-57 hyperkube[1433]: E0129 14:51:56.667021    1433 kubelet.go:2169] Container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: No CNI configuration file in /etc/kubernetes/cni/net.d/. Has your network provider started?
```

Maybe the error is something else?

wking:
Backported to 4.6 in #2652.

PR #2366, opened 2021-01-26
Bug 1840881: Make the kubelet config CR to MC mapping 1:1

umohnani8:
Fixes #1840881 (https://bugzilla.redhat.com/show_bug.cgi?id=1840881)

**- What I did**

Create a new MC object for each kubelet config object created.
Before we were only creating 1 MC object for every kubelet config
created, this would override the previouse kubelet config CR options
causing a rollback to go all the way back to the default values
when the latest ctrcfg CR was deleted.
So if there was kc-1, kc-2 and we deleted kc-2, it would
skip over kc-1 and go back to the defaults.
Having a 1:1 mapping of kubelet config to MC prevents this from happening.
So kc-1 will have mc-1 and kc-2 will have [mc-2](https://issues.redhat.com/browse/mc-2) - if kc-2
is deleted, the changes will roll back to mc-1, which is from kc-1
and not the initial defaults.
Add e2e tests for kubelet config controller as well.

**- How to verify it**

Start a cluster and create multiple kubelet config CRs, then delete them and see that the changes on the node rolls back to the previous kubelet config CR and not directly to the defaults. Also, each kubelet config CR will have its own MC and the suffix of the MC name will be stored as annotation in the kubelet config CR.

Create a new MC for each kubelet config object created.

### DIFF
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -180,6 +180,7 @@ func getManagedKeyCtrCfgDeprecated(pool *mcfgv1.MachineConfigPool) string {
 	return fmt.Sprintf("99-%s-%s-containerruntime", pool.Name, pool.ObjectMeta.UID)
 }
 
+// nolint: dupl
 func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, cfg *mcfgv1.ContainerRuntimeConfig) (string, error) {
 	// Get all the ctrcfg CRs
 	ctrcfgList, err := client.MachineconfigurationV1().ContainerRuntimeConfigs().List(context.TODO(), metav1.ListOptions{})
@@ -224,7 +225,7 @@ func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.In
 	// so that those changes can be rolled out to the nodes. But users will have to be mindful of how many ctrcfg CRs they create. Don't think
 	// anyone should ever have the need to create 10 when they can simply update an existing ctrcfg unless it is to apply to another pool.
 	if suffixNum+1 > 9 {
-		return "", fmt.Errorf("the max MC name suffix value that can be used is 9, please delete some ctrcfg CRs to free up suffx values before a new ctrcfg can be created")
+		return "", fmt.Errorf("max number of supported ctrcfgs (10) has been reached. Please delete old ctrcfgs before retrying")
 	}
 	// Return the default MC name with the suffixNum+1 value appended to it
 	return fmt.Sprintf("99-%s-generated-containerruntime-%s", pool.Name, strconv.Itoa(suffixNum+1)), nil
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -2,8 +2,10 @@ package kubeletconfig
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
 	osev1 "github.com/openshift/api/config/v1"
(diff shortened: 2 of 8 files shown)


umohnani8:
WIP because still running some tests, will update once done.

umohnani8:
@rphillips PTAL. Does feature gate need a similar fix? Or do we expect to only have one feature gate MC at a time

pkg/controller/kubelet-config/helpers.go
```diff
@@ -97,8 +99,55 @@ func findKubeletConfig(mc *mcfgv1.MachineConfig) (*ign3types.File, error) {
...
-	return ctrlcommon.GetManagedKey(pool, client, "99", "kubelet", getManagedKubeletConfigKeyDeprecated(pool))
+// nolint: dupl
+func getManagedKubeletConfigKey(pool *mcfgv1.MachineConfigPool, client mcfgclientset.Interface, cfg *mcfgv1.KubeletConfig) (string, error) {
+	// Get all the kubelet config CRs
+	kcList, err := client.MachineconfigurationV1().KubeletConfigs().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return "", fmt.Errorf("error listing kubelet configs: %v", err)
+	}
+	// If there is no kubelet config in the list, return the default MC name with no suffix
+	if kcList == nil || kcList.Items == nil {
```
rphillips, line 110: `kcList.Items == nil` should be `len(kcList.Items) == 0`

pkg/controller/kubelet-config/helpers.go
```diff
@@ -97,8 +99,55 @@ func findKubeletConfig(mc *mcfgv1.MachineConfig) (*ign3types.File, error) {
...
+		return ctrlcommon.GetManagedKey(pool, client, "99", "kubelet", getManagedKubeletConfigKeyDeprecated(pool))
+	}
+	for _, kc := range kcList.Items {
+		if kc.Name == cfg.Name {
```
rphillips, line 114: I suggest reversing the logic and keeping the if statement super small...

```
if kc.Name != cfg.Name {
  continue
}
... continue with the reset of the code...
```

pkg/controller/kubelet-config/helpers.go
```diff
@@ -97,8 +99,55 @@ func findKubeletConfig(mc *mcfgv1.MachineConfig) (*ign3types.File, error) {
...
+			}
+		}
+	}
+
+	// If we are here, this means that a new kubelet config was created, so we have to calculate the suffix value for its MC name
+	suffixNum := 0
+	// Get the suffix value of the second to last item in the kubelet config list. We use the second to last item because
+	// the most recent kubelet config being created has already been added to the list and is the kubelet config object that we are
+	// trying to figure out the suffix value for
+	val, ok := kcList.Items[len(kcList.Items)-2].GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
```
rphillips, line 133: Are we guaranteed to have 2 items in the slice? len(kcList.Items) could equal 0 - 2, thus an index of -2 or -1.

We need some sort of validation logic before this line.

rphillips:
I love the test cases!

pkg/controller/kubelet-config/helpers.go
umohnani8: Fixed

pkg/controller/kubelet-config/helpers.go
umohnani8: Fixed

pkg/controller/kubelet-config/helpers.go
umohnani8: I can add a check, but we are guaranteed to have at least 2 elements in the slice by the time we get to this point. When we create a CR for the first time, https://github.com/openshift/machine-config-operator/pull/2366/files#diff-4a2d0a4442e10d889a927322ad7311c67b22b259622067b52cbda7d2ece3ebefR119 will be true and it will return the original key. And as soon as you create a new CR it is appended to the `Items` list, so while we are figuring out what suffix to use for the CR, it already exists in the `Items` list and that is why the logic `len - 2` so we can get what the suffix was for the previous CR and calculate the suffix for the current one accordingly.

umohnani8:
@kikisdeliveryservice the error is propagated to the status of the CR and can be viewed with `oc describe...` :D
```
Spec:                                                                                                                                                         
  Kubelet Config:                                                                                                                                             
    Max Pods:  130                                                                                                                                            
  Machine Config Pool Selector:                                                                                                                               
    Match Labels:                                                                                                                                             
      Custom - Kubelet:  small-pods                                                                                                                           
Status:                                                                                                                                                       
  Conditions:                                                                                                                                                 
    Last Transition Time:  2021-01-27T16:08:53Z                                                                                                               
    Message:               could not get kubelet config key: Max number of supported kubelet config (10) has been reached. Please delete old kubelet c
onfigs before retrying
```

umohnani8:
This should be ready now, can I get a final review please @rphillips @kikisdeliveryservice

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -224,7 +225,7 @@ func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.In
 	// so that those changes can be rolled out to the nodes. But users will have to be mindful of how many ctrcfg CRs they create. Don't think
 	// anyone should ever have the need to create 10 when they can simply update an existing ctrcfg unless it is to apply to another pool.
 	if suffixNum+1 > 9 {
-		return "", fmt.Errorf("the max MC name suffix value that can be used is 9, please delete some ctrcfg CRs to free up suffx values before a new ctrcfg can be created")
+		return "", fmt.Errorf("Max number of supported ctrcfgs (10) has been reached. Please delete old ctrcfgs before retrying")
```
kikisdeliveryservice, line 228: Max -> max

no?

umohnani8: sure, fixed :)

kikisdeliveryservice (approved):
Thank you for adding the test coverage

:heart:

PR #2367, opened 2021-01-26
Bug 1920670: [Baremetal and friends] Properly handle SIGTERM - Keepalived container

yboaron:
When Keepalived container that holds VIP being SIGTERMed that should trigger a priority 0 advertisement
message to allow faster VIP migration.

This PR forwards the SIGTERM to child processes (the Keepalived processes) in Keepalived container
in order for SIGTERM to cause a VRRP priority 0 message to be sent.

### DIFF
--- a/templates/common/on-prem/files/keepalived.yaml
+++ b/templates/common/on-prem/files/keepalived.yaml
@@ -70,6 +70,13 @@ contents:
         - -c
         - |
           #/bin/bash
+          sigterm_handler()
+          {
+            if pid=$(pgrep -o keepalived); then
+              kill -s SIGTERM "$pid"
+            fi
+          }
+
           reload_keepalived()
           {
             if pid=$(pgrep -o keepalived); then
@@ -94,6 +101,9 @@ contents:
           declare -r keepalived_sock="/var/run/keepalived/keepalived.sock"
           export -f msg_handler
           export -f reload_keepalived
+          export -f sigterm_handler
+
+          trap sigterm_handler SIGTERM
           if [ -s "/etc/keepalived/keepalived.conf" ]; then
               /usr/sbin/keepalived -f /etc/keepalived/keepalived.conf --dont-fork --vrrp --log-detail --log-console &
           fi


yboaron:
/retitle bug 1920670: [Baremetal and friends] Properly handle SIGERM - Keepalived container

jcpowermac:
/test e2e-vsphere

cc: @jstuever @rvanderp3

cybertron:
/lgtm

The dangers of writing your own service handler. ;-)

dougsland:
looks sane to me. Just wondering if we have a CI job for covering such test. I don't see any change in the test files.
@kikisdeliveryservice @yboaron 
/lgtm

templates/common/on-prem/files/keepalived.yaml
```diff
@@ -70,6 +70,13 @@ contents:
         - -c
         - |
           #/bin/bash
+          sigterm_handler()
+          {
+            if pid=$(pgrep -o keepalived); then
```
darkmuggle, line 75: `killall /usr/sbin/keepalived` could simplify this a bit (has the same effect of `grep -o...`

kikisdeliveryservice:
@jcpowermac looks like the vsphere test is red since dec? (but not many runs)

/retest

kikisdeliveryservice:
> looks sane to me. Just wondering if we have a CI job for covering such test. I don't see any change in the test files.
> @kikisdeliveryservice @yboaron
> /lgtm

@dougsland onprem functionality isn't exercised in mainline mco ci tests, only in platform specific e2e (whereever that lives). so yall would have to review your ci tests to ensure adequate coverage. :)

yboaron:
/hold 
Need to verify that this change doesn't have negative influence on Keepalived restart process as a result of config update by the monitor container

mandre (commented):
/lgtm
Looks like it's only missing the approval.

kikisdeliveryservice:
@yboaron let's do this :smile: 

/approve

PR #2369, opened 2021-01-27
Bug 1904133: kubeletcfg: fix repeated status error msg

QiWang19:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1904133

**- What I did**
For kubeletconfig, do not append new status conditions if the error message stays the same
to avoid the repeated same conditions
This patch mirrors the handlerErr of the container runtime config controller.
**- How to verify it**

Verify this by applying the problematic YAML and waiting for several minutes to check the status
```
Status:
  Conditions:
    Last Transition Time:  2021-01-27T21:19:29Z
    Message:               Error: could not find any MachineConfigPool set for KubeletConfig
    Status:                False

// wait several minutes
Status:
  Conditions:
    Last Transition Time:  2021-01-27T21:26:16Z
    Message:               Error: could not find any MachineConfigPool set for KubeletConfig
    Status:                False
```

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -286,11 +286,6 @@ func (ctrl *Controller) handleErr(err error, key interface{}) {
 		return
 	}
 
-	if err == errCouldNotFindMCPSet {
-		ctrl.queue.AddAfter(key, 1*time.Minute)
-		return
-	}
-
 	if ctrl.queue.NumRequeues(key) < maxRetries {
 		glog.V(2).Infof("Error syncing kubeletconfig %v: %v", key, err)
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice (changes requested):
Just reading this you'll end up with the same err msg with updated timestamp so itll just stay 1 message? Makes sense

But can you update your commit message to indicate this is a kubelet config change?

QiWang19:
@kikisdeliveryservice PTAL. I changed the commit message to note it's a kubelet config change.

kikisdeliveryservice:
the rhel7 job is red no worries on retesting.

/skip

kikisdeliveryservice (approved):
Thanks for the changes @QiWang19 
for final approval:

/assign @rphillips

PR #2362, opened 2021-01-25
Bug 1916363: Fix how files from system-connections-merged get copied into system-connections

abhat:
In baremetal and vsphere platforms the merged directory serves as
overlayFS for supporting k8s nmstate. But because of the way it
gets mounted, if the corresponding files in that directory exist in
the lower directory, the files get their contents zeroed out.

This commit fixes the issue with zeroing out of nmconnection files.

Also, on other cloud platforms the merged directory doesn't exist.
The ovs-configuration script previously would try to copy files
from a non-existent path and fail. This commit also fixes that issue
by first checking if the merged directory exists before trying to
copy the nmconnection files.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

This is a follow-up fix for #2263

PR #2370, opened 2021-01-29
Use NetworkManager keyfile on all on-prem platforms

bcrochet:
Instead of making the entirety of the NetworkManager conf file for on-prem platforms
hinge on the existence of the API VIP, the keyfile setting should be set for on-prem
platforms at all times. This especially is present on the VSphere UPI platform.

This is related to https://github.com/openshift/machine-config-operator/pull/2362
that works around a problem where the overlayfs is mounted, but NetworkMangaer is
not pointed to the overlay.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

**- What I did**
Made the keyfile setting for NetworkManager present on all on-prem platforms.

**- How to verify it**
The 99-{platform}.conf file in /etc/NetworkManager/conf.d should always contain a [keyfile] section.

Made the keyfile setting for NetworkManager present on all on-prem platforms.

### DIFF
--- a/templates/common/on-prem/files/NetworkManager-onprem.conf.yaml
+++ b/templates/common/on-prem/files/NetworkManager-onprem.conf.yaml
@@ -9,3 +9,6 @@ contents:
     ipv6.dhcp-duid=ll
     ipv6.dhcp-iaid=mac
     {{ end -}}
+    [keyfile]
+    path=/etc/NetworkManager/system-connections-merged
+


cybertron:
/test e2e-vsphere-upi
/cc @jcpowermac @patrickdillon

This is primarily a change on the vsphere-upi platform so we should make sure it works there. Assuming it does, lgtm.

bcrochet:
> /test e2e-vsphere-upi
> /cc @jcpowermac @patrickdillon
> 
> This is primarily a change on the vsphere-upi platform so we should make sure it works there. Assuming it does, lgtm.

Thanks Ben. It closes a gap where the overlayfs is mounted, but NetworkManager isn't using it, causing confusion.

mandre:
Looks like e2e-vsphere-upi is green.
/lgtm

bcrochet:
No longer relevant.

PR #2374, opened 2021-01-29
[baremetal & friends] Move on-prem api-int record to dnsmasq

cybertron:
This is in preparation for moving the cluster-hosted network services
to a separate operator. With coredns no longer running as a static
pod, it will not be usable for providing the api-int record needed
for the node to register.

We decided to use dnsmasq instead of /etc/hosts because when a
deployer wants to use an external loadbalancer it will be necessary
to change the api-int record. If it's in /etc/hosts, that will require
restarting many/all of the pods to pick up the change. Using dnsmasq
allows us to just change the record in dnsmasq and SIGHUP it.

To allow dnsmasq and coredns to coexist on the node, coredns is moved
to port 5333 and dnsmasq has a server entry added to send queries
for the cluster domain to coredns.

Move api-int record for on-prem platforms to dnsmasq service.

### DIFF
--- a/templates/common/on-prem/units/on-prem-dnsmasq.yaml
+++ b/templates/common/on-prem/units/on-prem-dnsmasq.yaml
@@ -1,0 +1,28 @@
+name: on-prem-dnsmasq.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Run dnsmasq to provide local
+  Before=kubelet.service crio.service
+  [Service]
+  ExecStartPre=/bin/bash -c " \
+    until \
+    /usr/bin/podman run --rm \
+    --authfile /var/lib/kubelet/config.json \
+    --net=host \
+    --volume /etc/kubernetes:/etc/kubernetes:z \
+    --volume /etc/dnsmasq.d:/etc/dnsmasq.d:z \
+    {{ .Images.baremetalRuntimeCfgImage }} \
+    render \
+    /etc/kubernetes/kubeconfig \
+    --api-vip {{ onPremPlatformAPIServerInternalIP . }} \
+    --ingress-vip {{ onPremPlatformIngressIP . }} \
+    /etc/kubernetes/static-pod-resources/dnsmasq \
+    --out-dir /etc/dnsmasq.d; \
+    do \
+      sleep 5; \
+    done; \
+    restorecon -rF /etc/dnsmasq.d"
+  ExecStart=/usr/sbin/dnsmasq -k
+  [Install]
+  WantedBy=multi-user.target
--- a/templates/common/on-prem/files/coredns-corefile.yaml
+++ b/templates/common/on-prem/files/coredns-corefile.yaml
@@ -2,7 +2,7 @@ mode: 0644
 path: "/etc/kubernetes/static-pod-resources/coredns/Corefile.tmpl"
 contents:
   inline: |
-    . {
+    .:5333 {
         errors
         health :18080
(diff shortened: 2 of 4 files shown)


cybertron:
/hold
/cc @yboaron @celebdor 

The operator design is still in progress, but this is what we're currently planning. I've tested it with the operator and it works.

darkmuggle:
/hold

@crawford  @miabbott looks like others had my idea too -- moving from `/etc/hosts` to `dnsmasq`. 

From the perspective of the MCO, I think using `dnsmasq` is cleaner than manipulating `/etc/hosts`. I like this PR a lot better than the other PRs for manipulating hostnames. 

Given the discussions about `dnsmasq` and its supportability, accepting this PR would make `dnsmasq` an explicit dependency. 
Before we can accept this PR, we'll need ACK's from the RHCOS team that we're good to use dnsmasq.

templates/common/on-prem/units/on-prem-dnsmasq.yaml
```diff
@@ -0,0 +1,29 @@
+name: on-prem-dnsmasq.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Run dnsmasq to provide local
+  Before=kubelet.service crio.service
+  [Service]
+  ExecStart=/bin/bash -c " \
```
darkmuggle, line 8: Perhaps using this as an `ExecStartPre` and then using `ExecStart=/usr/bin/dnsmasq -k`?

cybertron: Makes sense. This is a holdover from the nodeip-configuration service I stole the pattern from, but in that case the podman call is the only thing being run.

cybertron:
> /hold
> 
> @crawford @miabbott looks like others had my idea too -- moving from `/etc/hosts` to `dnsmasq`.
> 
> From the perspective of the MCO, I think using `dnsmasq` is cleaner than manipulating `/etc/hosts`. I like this PR a lot better than the other PRs for manipulating hostnames.

One of our primary motivations for wanting this over /etc/hosts is that we need to be able to modify the records, potentially after initial deployment. With /etc/hosts I believe we'd have to restart all of the pods on the system to pick up changes. With dnsmasq, we just make the change, SIGHUP it (or the dbus equivalent), and all of the pods will use the new address immediately.

ashcrow:
> Before we can accept this PR, we'll need ACK's from the RHCOS team that we're good to use dnsmasq.

I've spoken with a few folks and the consensus is that using `dnsmasq` is acceptable.

yboaron:
Don't you need to update CoreDNS port also for 'friends' platforms files (e.g: https://github.com/openshift/machine-config-operator/blob/master/templates/common/vsphere/files/coredns-corefile.yaml) ?

cybertron:
> Don't you need to update CoreDNS port also for 'friends' platforms files (e.g: https://github.com/openshift/machine-config-operator/blob/master/templates/common/vsphere/files/coredns-corefile.yaml) ?

Hmm, that seems bad. It means both the on-prem and the vsphere template are going to be written to the same location. I guess it must happen to work out, but we should probably work with the vsphere team to converge those files.

cybertron:
Okay, I moved the config rendering to Pre and removed the Reload command because we aren't using it and I'm not sure it was working correctly anyway. I also have #2410 up to de-dupe the Corefiles because I think that's something we should do anyway. We either need to merge that or I'll need to change the port in those configs as well.

/hold

cybertron:
This is probably going to be superseded by https://github.com/openshift/machine-config-operator/pull/2450 but since that's a more significant change there's a possibility it will be nacked by the associated enhancement review.

PR #2380, opened 2021-02-01
Bug 1920807: [vsphere] set hostname with --static to provide consistent node name for CSR approval

rvanderp3:
Fixes: BZ1920807 - when creating a new machine, the node will get an unexpected hostname

**- What I did**
Added the --static flag to `hostnamectl` to prevent DHCP from overriding the hostname sourced from guestinfo.  This is only done once on vsphere nodes on the first boot.

**- How to verify it**
- Configure DHCP to assign a hostname as part of the lease
- Scale a new node
- The node should automatically join the cluster with a node name like ` <cluster>-<clusterid>-worker-<name>`

Resolves issue where the hostname provided by the DHCP overrides the guestinfo hostname thus preventing the machine-api-controller from reconciling new nodes which receive a hostname from DHCP.

### DIFF
--- a/templates/common/vsphere/files/vsphere-hostname.yaml
+++ b/templates/common/vsphere/files/vsphere-hostname.yaml
@@ -5,6 +5,9 @@ contents:
     #!/usr/bin/env bash
     set -e
 
+    # only run if the hostname is not set
+    test -f /etc/hostname && exit 0 || :
+
     if vm_name=$(/bin/vmtoolsd --cmd 'info-get guestinfo.hostname'); then
-        /usr/bin/hostnamectl --transient set-hostname ${vm_name}
+        /usr/bin/hostnamectl set-hostname --static ${vm_name}
     fi
--- a/templates/common/vsphere/units/vsphere-hostname.service.yaml
+++ b/templates/common/vsphere/units/vsphere-hostname.service.yaml
@@ -3,7 +3,14 @@ enabled: true
 contents: |
   [Unit]
   Description=vSphere hostname
+  Requires=vmtoolsd.service
   After=vmtoolsd.service
+
+  # Only run on first boot and only if the host does not have a name yet
+  ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
+  ConditionPathExists=!/etc/hostname
+  ConditionVirtualization=vmware
+  
   Before=kubelet.service
   Before=node-valid-hostname.service
   Before=NetworkManager.service


kikisdeliveryservice:
(need to make sure the actual platform affected by the PR has a ci run ^^^)

kikisdeliveryservice:
@rvanderp3 can you please update the commit to indicate this is a vsphere pr?

/hold

rvanderp3:
> @rvanderp3 can you please update the commit to indicate this is a vsphere pr?
> 
> /hold

ofcourse.  doing it now.

templates/common/vsphere/files/vsphere-hostname.yaml
```diff
@@ -6,5 +6,5 @@ contents:
     set -e
 
     if vm_name=$(/bin/vmtoolsd --cmd 'info-get guestinfo.hostname'); then
-        /usr/bin/hostnamectl --transient set-hostname ${vm_name}
+        /usr/bin/hostnamectl --transient --static set-hostname ${vm_name}
```
darkmuggle, line 9: The effect of this will be to set both the transient and the static hostname which will override and DHCP hostnames.

The transient hostname would change if the user removed the hostname from VSphere. In this case, both the transient hostname and the static hostname will be updated to whatever VSphere tells us. For other platforms, we've explicitly used `--transient` so all the forms of dynamic hostname discovery continues to work.

I would not recommend backporting this change.

rvanderp3: The issue we are seeing in 4.6.9 and later is if there is a hostname provided by DHCP, that hostname will be preferred over the one provided by the machine-api/vsphere configuration.  As a result, the machine-api will fail to reconcile the node as the hostname does not match what is expected and scaling machinesets fails.

rvanderp3: @darkmuggle Your explanation makes a lot of sense. I just verified that this issue can be addressed by changing the vsphere-hostname service to start after node-valid-hostname.service and NetworkManager.service.  The --static flag was removed as part of this test as well.  Would that be a preferred path to addressing this issue?

rvanderp3: I spoke a little too soon on that, it does address the issue, but there is a race condition depending on when NetworkManager sets the hostname which results in the DHCP hostname being applied some of the time.

darkmuggle: In general, we want to use DHCP over the VSphere hostname. The easier change would be to change https://github.com/openshift/machine-config-operator/blob/master/templates/common/vsphere/units/vsphere-hostname.service.yaml#L8 to `After`

jcpowermac: I think this should be tested where we can configure a DHCP server with a hostname for the guest. Since in this case I am making the assumption that the DHCP provided hostname is not the same as the guest name. I want to make sure that the cloud provider doesn't have a problem with the difference and that CSR approval still works correctly. I vaguely remember an issue with CSR approval and the hostname not matching - which ended up with using the extra config and this script that is being modified.

rvanderp3: I have an environment setup that can do this.  I'm working on getting additional testing on this.  

Adding /hold for the time being

rvanderp3: /hold

rvanderp3: @darkmuggle @jcpowermac I did some additional testing today with the proposed change and it seems to work as long as there is a hostname provided by DHCP.  If DHCP does not provide a hostname, a valid hostname does not get applied and as a result, boot of the node stops waiting for a valid hostname.  The node-valid-hostname service times out after 300 seconds and the hostname for the node remains `localhost`.  Which is not as desired as the current behavior where the node name matches the machine API given name unless DHCP provides a hostname.

darkmuggle: @rvanderp3 collaborated via Slack that gets around the whole hostname dance.

rvanderp3: Thanks!  I'm testing things out now.

rvanderp3: Unfortunately, the change did not seem the resolve the issue.   CSRs still had to be manually approved in order for an machineset created node to join the cluster.  The vSphere cloud provider is retrieving the node name from the from the hostname[vsphere.go#L314](https://github.com/openshift/kubernetes/blob/oc-4.7-kubernetes-1.20.1/staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go#L314).  The hostname-override does seem to update the hostname in the v1.Node resource, but does not have an impact on the node name reported by the kubelet.  The kube doc suggests the hostname-override flag may not be effective if there is a configured cloud provider and I think that may be what we are hitting here.  I'm going to move this PR to a WIP and regroup for the time being.

darkmuggle: You are indeed right @rvanderp3. And after a deep dive on the Cloud provider plugin, the kubelet and the golang `os` package -- sadly we will need to set the static name.

I pinged over a solution that I think should be an effective compromise.

rvanderp3: Just tested the fix on my cluster where I could reproduce the problem and things look good.  CSRs are auto-approving as expected on a scale up.

kikisdeliveryservice:
I'll defer to @darkmuggle and @jcpowermac for reviews

jcpowermac:
> I'll defer to @darkmuggle and @jcpowermac for reviews

@rvanderp3 and I talked yesterday about this PR.
I hope the 'on first boot' doesn't cause a problem. Good to see vmtoolds as required and the additional check of virtualization provider.

I would like to see the results of UPI though

darkmuggle:
> > I'll defer to @darkmuggle and @jcpowermac for reviews
> 
> @rvanderp3 and I talked yesterday about this PR.
> I hope the 'on first boot' doesn't cause a problem. Good to see vmtoolds as required and the additional check of virtualization provider.

We should be okay. The hostname needs to match the VM name for the certificate approval for the Kubelet and changing the hostname after the fact usually makes the kubelet unhappy.

darkmuggle (approved):
/lgtm

@rvanderp3  thank you to you and your team for the patience in working through this.

jcpowermac:
/lgtm

On slack with @rvanderp3 tested with UPI, no need to wait for CI.

PR #2381, opened 2021-02-01
daemon: block reboots during upgrades

darkmuggle:
This prevents any operator or user from issuing an unsafe `systemctl reboot` or "reboot" by masking the `reboot target` if the MCD is working on a node.  

Any reboot will emit a warning:

```
$ sudo systemctl reboot
Failed to reboot system via logind: Access denied
Failed to start reboot.target: Unit reboot.target is masked.
```

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -451,6 +451,16 @@ func (dn *Daemon) syncNode(key string) error {
 	// and then proceeds to check the state of the node, which includes
 	// finalizing an update and/or reconciling the current and desired machine configs.
 	if dn.booting {
+		// Prevent any reboots
+		if err := maskRebootTarget(); err != nil {
+			return err
+		}
+		defer func() {
+			if err := unmaskRebootTarget(); err != nil {
+				glog.Error(err)
+			}
+		}()
+
 		// Be sure only the MCD is running now, disable -firstboot.service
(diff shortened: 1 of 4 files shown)


darkmuggle:
This "fix" is actually short-sighted. To do reboots safely, there should be a locking mechanism (or leasing) and/or an API to block reboots. Regardless, this will prevent a reboot until its safe to do so.

cgwalters:
In https://github.com/kubernetes/kubernetes/pull/96129 the kubelet gained code to talk to systemd to inhibit reboots, which is the cleaner way of doing this.

yuqi-zhang:
Reading the kubelet code a bit, it seems to mostly be used to handle pod drains during shutdowns. Maybe not directly using it, but worth considering doing the same thing (i.e. using https://www.freedesktop.org/wiki/Software/systemd/inhibit/ to lock/unlock)

darkmuggle:
/hold 

I'll look into doing the Inhibit through dbus.

darkmuggle:
I concur with @yuqi-zhang's conclusion -- the Kubelet code is used to set a shutdown timer, but it doesn't _stop_ the shutdown. That said, after playing around with it, I can work out a DBus messager; I think its about a day or so worth of work.

darkmuggle:
Per an MCO team meeting, we're going to approach this problem differently.

PR #2382, opened 2021-02-01
add kubelet environment variable for extra node lables

yuvalk:
this can be used in conjunction with ignition that include something like:
```
$ cat /etc/systemd/system/kubelet.service.d/20-extralabels.conf
[Service]
Environment="FIRST_BOOT_ONLY_KUBELET_NODE_EXTRA_LABELS=node.kubernetes.io/role=rwn"
```
Which will allow to instantiate nodes directly into the desired MCP

Signed-off-by: Yuval Kashtan <yuvalkashtan@gmail.com>

**- What I did**

**- How to verify it**

### DIFF
--- a/docs/KubeletConfigDesign.md
+++ b/docs/KubeletConfigDesign.md
@@ -63,7 +63,7 @@ KubeletConfig
 
 ## VALIDATION
 
-It's importent to note that, since the fields of the kubelet configuration are directly fetched from upstream the validation 
+It's important to note that, since the fields of the kubelet configuration are directly fetched from upstream the validation 
 of those values is handled directly by the kubelet. Please refer to the upstream version of the relavent kubernetes for the 
 valid values of these fields. Invalid values of the kubelet configuration fields may render cluster nodes unusable.
 
--- a/install/0000_80_machine-config-operator_01_kubeletconfig.crd.yaml
+++ b/install/0000_80_machine-config-operator_01_kubeletconfig.crd.yaml
@@ -56,7 +56,7 @@ spec:
               description: The fields of the kubelet configuration are defined in 
                 kubernetes upstream. Please refer to the types defined in the 
                 version/commit used by OpenShift of the upstream kubernetes.
-                It's importent to note that, since the fields of the kubelet
+                It's important to note that, since the fields of the kubelet
                 configuration are directly fetched from upstream the validation 
                 of those values is handled directly by the kubelet. Please refer 
                 to the upstream version of the relavent kubernetes for the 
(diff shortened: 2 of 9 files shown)


kikisdeliveryservice:
So this would ensure that the node always has the correct label available and doesn't reboot back after getting the worker label here?

@yuvalk does this cover upi usecase? 

Can you walk me through the user flow using this change?

kikisdeliveryservice:
Adding a hold for further discussion/testing. Thanks for the PR

/hold

crawford:
This is what was suggested in https://bugzilla.redhat.com/show_bug.cgi?id=1850295, but that was closed. I assume this will be rejected for a similar reason, unfortunately.

kikisdeliveryservice:
update: seems that there might be a viable alternative path that they're going to explore targetting the ingress&friends scheduling issues directly. will leave this open for now.

slintes:
> This is what was suggested in https://bugzilla.redhat.com/show_bug.cgi?id=1850295, but that was closed. I assume this will be rejected for a similar reason, unfortunately.

If I understand correctly, the only argument in the end was preventing confusion about why you can set e.g. a custom role on registration, but not modify it? Not a strong argument IMHO, considering this would solve a big problem.

> update: seems that there might be a viable alternative path that they're going to explore targetting the ingress&friends scheduling issues directly. will leave this open for now.

To my best knowledge workload scheduling is only one part of the problem. Unneeded reboots and reprovisioning (from custom role which was installed using ignition, to worker role because of missing role label, and back to custom role after labels were set by "something") is the other part.

yuvalk:
pushed a small fix to rename the environment variable in a more friendly name that denote the fact that it can be used only once.

yuvalk:
> To my best knowledge workload scheduling is only one part of the problem. Unneeded reboots and reprovisioning (from custom role which was installed using ignition, to worker role because of missing role label, and back to custom role after labels were set by "something") is the other part.
Yep, we'd want to minimize reboots to shorten deployment time

also, I think this is very useful in more dynamic clusters, to prevent undesirable workload from reaching that node in the first place. consider the case of highly dynamic cluster. when the node is made schedulable on the 'general' worker pull it will start receiving workloads, just to evict it shortly afterwards..

darkmuggle:
In light of https://bugzilla.redhat.com/show_bug.cgi?id=1850295#c11, we would need the Node team to ack this PR. I'm not entirely sure that this is the _correct_ path.

kikisdeliveryservice:
> In light of https://bugzilla.redhat.com/show_bug.cgi?id=1850295#c11, we would need the Node team to ack this PR. I'm not entirely sure that this is the _correct_ path.

AFAIK, we are not going forward with this.

yuvalk:
/close
We have a path to achieve the same without changing MCO
https://github.com/yuvalk/ocp-node-labels

PR #2352, opened 2021-01-20
Bug 1918471: fix custom featuregates and add a unit test

rphillips:
**- What I did**
Fixed applying Custom FeatureGate to the MCO. Prior to this fix 'CustomNoUpgrade' featuregates would not be injected into ignition correctly. This fix adds a unit test as well.

```
apiVersion: config.openshift.io/v1
kind: FeatureGate
metadata:
  name: cluster
spec:
  featureSet: "CustomNoUpgrade"
  customNoUpgrade:
    enabled:
      - GracefulNodeShutdown
      - APIPriorityAndFairness
      - RotateKubeletServerCertificate
      - SupportPodPidsLimit
      - NodeDisruptionExclusion
      - ServiceNodeExclusion
      - SCTPSupport
    disabled:
      - LegacyNodeRoleBehavior
      - RemoveSelfLink
```

**- How to verify it**
Edit the cluster featuregate and paste in this spec. Also rely on the added unit test.

**- Description for the changelog**

PR #2386, opened 2021-02-02
[OCPCLOUD-1107] External cloud-provider support via FeatureGate in post-install

Danil-Grigorev:
**- What I did**

Implementing external cloud provider selection forced by feature gate. Described in the proposal: https://github.com/openshift/enhancements/pull/463/files#diff-3e0e2c48e70215076dfe36c13768a823ab7080d929d80292f37db2ef5a2121e8R201

This PR is main blocker on other work with integrating CCM in Openshift: https://issues.redhat.com/browse/OCPCLOUD-1107
 
**- How to verify it**

### DIFF
--- a/pkg/controller/common/constants.go
+++ b/pkg/controller/common/constants.go
@@ -18,4 +18,7 @@ const (
 
 	// MCNameSuffixAnnotationKey is used to keep track of the machine config name associated with a CR
 	MCNameSuffixAnnotationKey = "machineconfiguration.openshift.io/mc-name-suffix"
+
+	// ClusterFeatureInstanceName is a singleton name for featureGate configuration
+	ClusterFeatureInstanceName = "cluster"
 )
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -330,13 +330,13 @@ func (ctrl *Controller) handleFeatureErr(err error, key interface{}) {
 	ctrl.featureQueue.AddAfter(key, 1*time.Minute)
 }
 
-func (ctrl *Controller) generateOriginalKubeletConfig(role string) (*ign3types.File, error) {
+func (ctrl *Controller) generateOriginalKubeletConfig(role string, featureGate *configv1.FeatureGate) (*ign3types.File, error) {
 	cc, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
 	if err != nil {
 		return nil, fmt.Errorf("could not get ControllerConfig %v", err)
 	}
 	// Render the default templates
-	rc := &mtmpl.RenderConfig{ControllerConfigSpec: &cc.Spec}
+	rc := &mtmpl.RenderConfig{ControllerConfigSpec: &cc.Spec, FeatureGate: featureGate}
 	generatedConfigs, err := mtmpl.GenerateMachineConfigsForRole(rc, role, ctrl.templatesDir)
 	if err != nil {
 		return nil, fmt.Errorf("GenerateMachineConfigsforRole failed with error %s", err)
@@ -470,7 +470,7 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
 		return ctrl.syncStatusOnly(cfg, err)
 	}
 
-	features, err := ctrl.featLister.Get(clusterFeatureInstanceName)
+	features, err := ctrl.featLister.Get(ctrlcommon.ClusterFeatureInstanceName)
 	if macherrors.IsNotFound(err) {
 		features = createNewDefaultFeatureGate()
 	} else if err != nil {
@@ -478,7 +478,7 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
 		err := fmt.Errorf("could not fetch FeatureGates: %v", err)
 		return ctrl.syncStatusOnly(cfg, err)
 	}
-	featureGates, err := ctrl.generateFeatureMap(features)
+	featureGates, err := generateFeatureMap(features)
 	if err != nil {
 		err := fmt.Errorf("could not generate FeatureMap: %v", err)
 		glog.V(2).Infof("%v", err)
@@ -512,7 +512,7 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
 		userDefinedSystemReserved := make(map[string]string, 2)
 
 		// Generate the original KubeletConfig
-		originalKubeletIgn, err := ctrl.generateOriginalKubeletConfig(role)
+		originalKubeletIgn, err := ctrl.generateOriginalKubeletConfig(role, features)
 		if err != nil {
 			return ctrl.syncStatusOnly(cfg, err, "could not generate the original Kubelet config: %v", err)
 		}
--- a/pkg/controller/kubelet-config/kubelet_config_features.go
+++ b/pkg/controller/kubelet-config/kubelet_config_features.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/imdario/mergo"
 	osev1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/cloudprovider"
 	"github.com/vincent-petithory/dataurl"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -23,8 +24,11 @@ import (
 	"github.com/openshift/machine-config-operator/pkg/version"
 )
 
-const (
-	clusterFeatureInstanceName = "cluster"
+var (
+	// openshiftOnlyFeatureGates contains selection of featureGates which will be rejected by native kubelet
+	openshiftOnlyFeatureGates = []string{
+		cloudprovider.ExternalCloudProviderFeature,
+	}
 )
 
 func (ctrl *Controller) featureWorker() {
@@ -52,7 +56,7 @@ func (ctrl *Controller) syncFeatureHandler(key string) error {
 	}()
 
 	// Fetch the Feature
-	features, err := ctrl.featLister.Get(clusterFeatureInstanceName)
+	features, err := ctrl.featLister.Get(ctrlcommon.ClusterFeatureInstanceName)
 	if errors.IsNotFound(err) {
 		glog.V(2).Infof("FeatureSet %v is missing, using default", key)
 		features = &osev1.FeatureGate{
@@ -65,7 +69,7 @@ func (ctrl *Controller) syncFeatureHandler(key string) error {
 	} else if err != nil {
 		return err
 	}
-	featureGates, err := ctrl.generateFeatureMap(features)
+	featureGates, err := generateFeatureMap(features, openshiftOnlyFeatureGates...)
 	if err != nil {
 		return err
 	}
@@ -97,7 +101,7 @@ func (ctrl *Controller) syncFeatureHandler(key string) error {
 			}
 		}
 		// Generate the original KubeletConfig
-		originalKubeletIgn, err := ctrl.generateOriginalKubeletConfig(role)
+		originalKubeletIgn, err := ctrl.generateOriginalKubeletConfig(role, nil)
 		if err != nil {
 			return err
 		}
@@ -197,7 +201,8 @@ func (ctrl *Controller) deleteFeature(obj interface{}) {
 }
 
 //nolint:gocritic
-func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[string]bool, error) {
+// generateFeatureMap returns a map of enabled/disabled feature gate selection with exclusion list
+func generateFeatureMap(features *osev1.FeatureGate, exclusions ...string) (*map[string]bool, error) {
 	rv := make(map[string]bool)
 	set, ok := osev1.FeatureSets[features.Spec.FeatureSet]
 	if !ok {
@@ -218,6 +223,11 @@ func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[st
 		for _, featDisabled := range features.Spec.CustomNoUpgrade.Disabled {
 			rv[featDisabled] = false
 		}
+	}
+
+	// Remove features excluded due to being breaking for some reason
+	for _, excluded := range exclusions {
+		delete(rv, excluded)
 	}
 	return &rv, nil
 }
--- a/pkg/controller/template/render.go
+++ b/pkg/controller/template/render.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Masterminds/sprig"
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/cloudprovider"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/machine-config-operator/pkg/constants"
@@ -25,7 +26,8 @@ import (
 // RenderConfig is wrapper around ControllerConfigSpec.
 type RenderConfig struct {
 	*mcfgv1.ControllerConfigSpec
-	PullSecret string
+	PullSecret  string
+	FeatureGate *configv1.FeatureGate
 
 	// no need to set this, will be automatically configured
 	Constants map[string]string
@@ -332,6 +334,13 @@ func skipMissing(key string) (interface{}, error) {
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	if cfg.Infra.Status.PlatformStatus != nil {
+		external, err := cloudprovider.IsCloudProviderExternal(cfg.Infra.Status.PlatformStatus.Type, cfg.FeatureGate)
+		if err != nil {
+			glog.Error(err)
+		} else if external {
+			return "external", nil
+		}
+
 		switch cfg.Infra.Status.PlatformStatus.Type {
 		case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.OpenStackPlatformType, configv1.VSpherePlatformType:
 			return strings.ToLower(string(cfg.Infra.Status.PlatformStatus.Type)), nil
@@ -368,6 +377,13 @@ func cloudConfigFlag(cfg RenderConfig) interface{} {
 		cfg.Infra.Status.PlatformStatus = &configv1.PlatformStatus{
 			Type: "",
 		}
+	}
+
+	external, err := cloudprovider.IsCloudProviderExternal(cfg.Infra.Status.PlatformStatus.Type, cfg.FeatureGate)
+	if err != nil {
+		glog.Error(err)
+	} else if external {
+		return ""
 	}
 
 	flag := "--cloud-config=/etc/kubernetes/cloud.conf"
--- a/pkg/controller/template/template_controller.go
+++ b/pkg/controller/template/template_controller.go
@@ -5,10 +5,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"sort"
 	"time"
 
 	"github.com/golang/glog"
+	configv1 "github.com/openshift/api/config/v1"
+	osev1 "github.com/openshift/api/config/v1"
+	oseinformersv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	oselistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/machine-config-operator/lib/resourceapply"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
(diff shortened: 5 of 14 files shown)


pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -317,12 +317,24 @@ func (ctrl *Controller) handleFeatureErr(err error, key interface{}) {
 }
 
 func (ctrl *Controller) generateOriginalKubeletConfig(role string) (*ign3types.File, error) {
+	var features *map[string]bool
 	cc, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
 	if err != nil {
 		return nil, fmt.Errorf("could not get ControllerConfig %v", err)
 	}
+	if fg, err := ctrl.featLister.Get(clusterFeatureInstanceName); !macherrors.IsNotFound(err) {
```
kikisdeliveryservice, line 325: macherrors-> matcherrors

kikisdeliveryservice (commented):
feel like the team/arch also needs to review the enhancement 

/hold

kikisdeliveryservice:
Feels like this should leverage: https://github.com/openshift/machine-config-operator/pull/2352  via: https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L121

pkg/controller/template/render.go
```diff
@@ -332,6 +337,9 @@ func skipMissing(key string) (interface{}, error) {
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	if cfg.Infra.Status.PlatformStatus != nil {
+		if enabled, found := cfg.FeatureGates[featureGateExternalName]; found && enabled {
+			return "external", nil
+		}
```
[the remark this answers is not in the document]
Danil-Grigorev: Eventually, yes. In TechPreviewNoUpgrade it is on feature gate to set invariant configuration when found. I don't think anyone will risk setting such gate on production cluster for platform we don't have configs for. This just means less changes in development cycle, if we don't make it a switch.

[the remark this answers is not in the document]
JoelSpeed: I had thought we would want this to only affect things on platforms we have allowed it to, hence being provider specific, let's discuss on the call next week

kikisdeliveryservice (commented):
@Danil-Grigorev I've gone through the enhancements, but am having problems connecting the dots here.

Could you add some detail/specific description as to what this PR is trying to accomplish, whether this PR is like step 1 of X etc... if this is day2 post-install is this user initiated? etc? Will there be any followup prs or are there prereq prs for this? etc

Thanks! :)

Danil-Grigorev:
@kikisdeliveryservice Sure. The prerequisite PR will be an API change for FeatureGate in https://github.com/openshift/api/pull/738 The name of the featureGate, it's location and related changes in functionality may be corrected based on new data from https://github.com/kubernetes/enhancements/pull/2443. What it will result into - we will have a way to deploy a techPreview clusters with out-of-tree providers code, and attempt migration from in-tree in post install time, while identifying issues (CI use-case). It is user initiated.

Ideal scenario - follow up PRs will only remove selections from platform switch case, say: AWS is now by default `external` in 1.24. Realistically, we will still have to force some `kubelet` flags even after the upstream feature gate is implemented, as the goal is to identify issues in migration, which may break some storage, networking and cloud-credentials functionality in the cluster if done blindly.

Actual change, which we will probably need - implement our selection (or more realistically wait for some helper upstream implementation) in MCO install phase, where the bootstrap methods are called. This way a fresh cluster will get out-of-tree providers by default with related `DisableCloudProviders` feature gate . This is out-of-scope for this PR.

kikisdeliveryservice:
Thanks for the summary @Danil-Grigorev !

pkg/controller/kubelet-config/kubelet_config_features.go
```diff
@@ -197,7 +201,8 @@ func (ctrl *Controller) deleteFeature(obj interface{}) {
 }
 
 //nolint:gocritic
-func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[string]bool, error) {
+// generateFeatureMap returns a map of enabled/disabled feature gate selection with exclusion list
+func generateFeatureMap(features *osev1.FeatureGate, blacklist ...string) (*map[string]bool, error) {
```
JoelSpeed, line 205:
```suggestion
func generateFeatureMap(features *osev1.FeatureGate, exclusions ...string) (*map[string]bool, error) {
```

pkg/controller/kubelet-config/kubelet_config_features.go
```diff
@@ -219,5 +224,10 @@ func (ctrl *Controller) generateFeatureMap(features *osev1.FeatureGate) (*map[st
 			rv[featDisabled] = false
 		}
 	}
+
+	// Remove features excluded due to being breaking for some reason
+	for _, excluded := range blacklist {
```
JoelSpeed, line 229:
```suggestion
	for _, excluded := range exclusions {
```

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,54 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
 	}
 }
 
+func (ctrl *Controller) enqueueController() {
+	cfg, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get ControllerConfig on secret callback %#v", err))
+		return
+	}
+	glog.V(4).Infof("Re-syncing ControllerConfig %s due to secret change", cfg.Name)
```
JoelSpeed, line 191: This log line is no longer correct if this is being called from `updateFeature` for example, does this need to be more generic?
```suggestion
	glog.V(4).Infof("Re-syncing ControllerConfig %s due to dependency change", cfg.Name)
```

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,54 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
...
+			return
+		}
+		features, ok = tombstone.Obj.(*osev1.FeatureGate)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a FeatureGate %#v", obj))
+			return
+		}
+	}
+	glog.V(4).Infof("Deleted Feature %s and restored default controller config", features.Name)
+}
```
JoelSpeed, line 225: Do we not need to enqueue a reconcile if the feature is deleted?

pkg/controller/template/template_controller.go
```diff
@@ -418,7 +471,14 @@ func (ctrl *Controller) syncControllerConfig(key string) error {
 		}
 		pullSecretRaw = secret.Data[corev1.DockerConfigJsonKey]
 	}
-	mcs, err := getMachineConfigsForControllerConfig(ctrl.templatesDir, cfg, pullSecretRaw)
+
+	fg, err := ctrl.featLister.Get(ctrlcommon.ClusterFeatureInstanceName)
+	if err != nil && !errors.IsNotFound(err) {
+		glog.V(2).Infof("%v", err)
+		err := fmt.Errorf("could not fetch FeatureGate: %v", err)
```
JoelSpeed, line 478: Why not wrap the error before logging it? Gives more context right?

pkg/controller/template/render.go
```diff
@@ -332,6 +334,13 @@ func skipMissing(key string) (interface{}, error) {
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	if cfg.Infra.Status.PlatformStatus != nil {
+		external, err := cloudprovider.IsCloudProviderExternal(cfg.Infra.Status.Platform, cfg.FeatureGate) //nolint
```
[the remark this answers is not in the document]
Danil-Grigorev: Because of the deprecation warning. library-go still uses infra.status.platform, while is recommended to use infra.status.platformStatus.type now. This causes `verify` check to fail.

pkg/controller/template/render.go
```diff
@@ -332,6 +334,13 @@ func skipMissing(key string) (interface{}, error) {
...
+		if err != nil {
+			glog.Error(err)
+		} else if external {
+			return "external", nil
+		}
```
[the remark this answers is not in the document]
Danil-Grigorev: I'm only replicating usual error check with skipping the logic if is found. Will rather leave it as is.

pkg/controller/template/template_controller.go
Danil-Grigorev: No, it is usual for such static resources to be re-queued only on add/update. Removing it and reconciling is as if you rip off a part of cluster, and therefor destroying it's health.

JoelSpeed: This affects the state of the output of MCO though, so deleting it does infer a different desired state, is something going to automatically put it back if you delete it?  I'll leave this decision to the MCO owners but to me it seems natural to reconcile this on delete as well

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,54 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
 	}
 }
 
+func (ctrl *Controller) enqueueController() {
+	cfg, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get ControllerConfig on secret callback %#v", err))
```
JoelSpeed, line 188: Missed this one, also needs an update to be non-secret specific here too
```suggestion
		utilruntime.HandleError(fmt.Errorf("couldn't get ControllerConfig on dependency callback %#v", err))
```

pkg/controller/common/constants.go
```diff
@@ -18,4 +18,7 @@ const (
 
 	// MCNameSuffixAnnotationKey is used to keep track of the machine config name associated with a CR
 	MCNameSuffixAnnotationKey = "machineconfiguration.openshift.io/mc-name-suffix"
+
+	// ClusterFeatureInstanceName is a singletone name for featureGate configuration
```
kikisdeliveryservice, line 22: singleton?

pkg/controller/template/render.go
```diff
@@ -332,6 +334,13 @@ func skipMissing(key string) (interface{}, error) {
 
 func cloudProvider(cfg RenderConfig) (interface{}, error) {
 	if cfg.Infra.Status.PlatformStatus != nil {
+		external, err := cloudprovider.IsCloudProviderExternal(cfg.Infra.Status.Platform, cfg.FeatureGate) //nolint
+		if err != nil {
+			glog.Error(err)
```
kikisdeliveryservice, line 339: is there a reason you're not returning the error here?

pkg/controller/template/render.go
```diff
@@ -370,6 +379,13 @@ func cloudConfigFlag(cfg RenderConfig) interface{} {
 		}
 	}
 
+	external, err := cloudprovider.IsCloudProviderExternal(cfg.Infra.Status.Platform, cfg.FeatureGate) //nolint
+	if err != nil {
+		glog.Error(err)
```
kikisdeliveryservice, line 384: same as above - does only logging this error make sense?

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,55 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
 	}
 }
 
+func (ctrl *Controller) enqueueController() {
+	cfg, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get ControllerConfig on dependency callback %#v", err))
```
kikisdeliveryservice, line 188: the only thing is that changing this string makes the error much more generic

kikisdeliveryservice (commented):
made a pass through with some questions/comments :+1:

pkg/controller/template/render.go
Danil-Grigorev: Yes, the error check is made solely for featureGate validation, and is backed by the filtering behavior, which will return `false`. We don't want to break existing `cloudProvider` workflow, if the featureGate configuration is invalid. This is temporary solution: https://github.com/openshift/library-go/pull/993/files#diff-478af36e9fb994fc80d37b7d2f6ae207c67d8c43b94f98f6ae3e420808958ba9R41 Here is the place where the error comes from, it could be safely ignored and only printed.

pkg/controller/template/render.go
Danil-Grigorev: https://github.com/openshift/machine-config-operator/pull/2386#discussion_r587561388

pkg/controller/template/render.go
Danil-Grigorev: It will result in `external` false, which will not break and reconfigure cluster.

kikisdeliveryservice: thanks!

pkg/controller/template/template_controller.go
JoelSpeed: Could pass the type into this function as a string and format string it into the error? Bit of an odd pattern though maybe?

Danil-Grigorev:
@kikisdeliveryservice Latest enhancement update contains detailed information about flags which this PR is setting. Could you give it another review? https://github.com/openshift/enhancements/pull/463/commits/af7b324b8303bb43d80fc6296ca5f4bdeaecb9ef We can't test how CCM initialize node for provider without this patch. Everything is described in the proposal

Danil-Grigorev:
@rphillips @kikisdeliveryservice @JoelSpeed Could you guys give it a review and lablels please? We need it to introduce operator in 4.9, so once the branch opens this should be merged first thing.

templates/common/aws/units/afterburn-hostname.service.yaml
```diff
@@ -0,0 +1,17 @@
+name: afterburn-hostname.service
```
JoelSpeed, line 1: I thought this was being covered in a separate PR? We should keep the scope of this PR just to the feature gate changes and have a separate PR for the AWS hostname issue

JoelSpeed (commented):
LGTM, though I think we will want to address the afterburn issue in a separate PR

templates/common/aws/units/afterburn-hostname.service.yaml
Danil-Grigorev: Yes it was, although this implementation is the only one which worked. An issue is still there in the other PR and there is no way to fix it: https://github.com/openshift/machine-config-operator/pull/2401#issuecomment-789739611

The implementation does not belong here, was a temporary fix for testing. Removed the last commit.

JoelSpeed (commented):
There's a couple of files for group generation which I think are part of a code generation tool, that seem to have changed permissions, this seems unrelated to this PR, have you run `go mod tidy` and `go mod vendor`? I'm surprised the permissions would be changing here

pkg/controller/template/template_controller.go
```diff
@@ -102,27 +111,29 @@ func New(
...
 }
 
 func (ctrl *Controller) filterSecret(secret *corev1.Secret) {
 	if secret.Name == "pull-secret" {
-		cfg, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
-		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("couldn't get ControllerConfig on secret callback %#v", err))
-			return
-		}
-		glog.V(4).Infof("Re-syncing ControllerConfig %s due to secret change", cfg.Name)
```
kikisdeliveryservice, line 124: I keep coming back to removing this secret change logging... Can you leave the log line and maybe with a reword for it to make sense since there's another log line in the func called ?

kikisdeliveryservice: As an aside: @sinnykumari should this logging really be v4? on secret change maybe we want it to be more visible? is there an event for this somewhere?

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,55 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
...
+	}
+	glog.V(4).Infof("Re-syncing ControllerConfig %s due to dependency change", cfg.Name)
+	ctrl.enqueueControllerConfig(cfg)
+}
+
+func (ctrl *Controller) updateFeature(old, cur interface{}) {
+	oldFeature := old.(*osev1.FeatureGate)
+	newFeature := cur.(*osev1.FeatureGate)
+	if !reflect.DeepEqual(oldFeature.Spec, newFeature.Spec) {
+		glog.V(4).Infof("Update Feature %s", newFeature.Name)
```
kikisdeliveryservice, line 199:
```suggestion
		glog.V(4).Infof("Updating Feature: %s", newFeature.Name)
```

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,55 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
...
+		ctrl.enqueueController()
+	}
+}
+
+func (ctrl *Controller) addFeature(obj interface{}) {
+	features := obj.(*osev1.FeatureGate)
+	glog.V(4).Infof("Adding Feature %s", features.Name)
```
kikisdeliveryservice, line 206:
```suggestion
	glog.V(4).Infof("Adding Feature: %s", features.Name)
```

pkg/controller/template/template_controller.go
```diff
@@ -171,12 +182,55 @@ func (ctrl *Controller) deleteSecret(obj interface{}) {
...
+			utilruntime.HandleError(fmt.Errorf("Couldn't get object from tombstone %#v", obj))
+			return
+		}
+		features, ok = tombstone.Obj.(*osev1.FeatureGate)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a FeatureGate %#v", obj))
+			return
+		}
+	}
+	glog.V(4).Infof("Deleted Feature %s", features.Name)
```
kikisdeliveryservice, line 224:
```suggestion
	glog.V(4).Infof("Deleting Feature %s", features.Name)
```

It hasn't happened yet right?

kikisdeliveryservice (commented):
A few smaller comments but same q as joel on the permission change

pkg/controller/template/render.go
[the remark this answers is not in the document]
kikisdeliveryservice: double checking: is this still true?

kikisdeliveryservice:
`level=error msg=Cluster operator ingress Degraded is True with IngressDegraded: The "default" ingress controller reports Degraded=True: DegradedConditions: One or more other status conditions indicate a degraded state: PodsScheduled=False (PodsNotScheduled: Some pods are not scheduled: Pod "router-default-794c6d77d-msh47" cannot be scheduled: 0/4 nodes are available: 1 node(s) didn't match pod affinity/anti-affinity rules, 1 node(s) didn't match pod anti-affinity rules, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate. Make sure you have sufficient worker nodes.)`

/test e2e-agnostic-upgrade

kikisdeliveryservice:
Would really like to see a successful upgrade run on this but there seem to be a bunch of apimachinery failures.

pkg/controller/template/template_controller.go
sinnykumari: yes, there is no confidential information we are logging here. So, it makes sense to log it without verbosity.

pkg/controller/template/render.go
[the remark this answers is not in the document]
Danil-Grigorev: I guess not, we could use platformStatus.type, let me update this quickly. MCO does bootstrap part after cluster-config-operator, so it should work.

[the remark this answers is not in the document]
JoelSpeed: PlatformStatus.Type is default from OCP 4.2 so I'd expect it to be the one we use here anyway

pkg/controller/template/template_controller.go
Danil-Grigorev: Added the log line without verbosity

Danil-Grigorev:
@JoelSpeed @kikisdeliveryservice All comments addressed. I checked, permission change happens on revendor, and it seems expected. Please review.

JoelSpeed (commented):
/lgtm

Thanks @Danil-Grigorev

elmiko:
just wanted to note that i have been test driving this patch for the last few days and it seems to be working for me. i have not given a thorough testing of the negative case (eg when the gate is not applied), but i have done a basic smoke test on that.

elmiko:
@Danil-Grigorev , i talked with @kikisdeliveryservice in slack and it sounds like we could help here by adding a must-gather for a cluster that has had the feature gate enabled as a day 2 operation. so, basically:
1. create normal cluster with in-tree providers
2. enable feature gate and switch cluster to out-of-tree
3. once cluster is healthy, collect must-gather

i think we are already capturing this behavior in the [openshift-e2e-aws-ccm](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/e2e/aws/ccm/openshift-e2e-aws-ccm-workflow.yaml) step. but since we need this PR to make it work, [none of those jobs have been passing](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-cloud-controller-manager-operator-master-e2e-aws-ccm). @Danil-Grigorev does that make sense?

kikisdeliveryservice:
just to follow-on the above, would this ever be disabled day2 @Danil-Grigorev ?

Danil-Grigorev:
@kikisdeliveryservice No, applying the feature gate would move cluster to TechPreview state, where we officially do not support rollback. Yet the code does not prevent it from happening and works both ways.

elmiko:
quick update here, we have shared a must-gather with @kikisdeliveryservice

kikisdeliveryservice (approved):
Reviewed must gather and verified:

we started off with --cloud-provider=aws
and
successfully switched to --cloud-provider=external

Based on my understanding from: https://github.com/openshift/enhancements/pull/463/files#diff-3e0e2c48e70215076dfe36c13768a823ab7080d929d80292f37db2ef5a2121e8R270 also saw: ExternalCloudProvider: true alongside --cloud-provider=aws which helps force upgrade from intree to out of tree.

Since this has also been extensively tested by that team, let's merge this to unblock work and move forward.

Issue #2385, opened 2021-02-02
/var/home/core/.ssh/authorized_keys not created by MCO during install (OKD 4.6/4.7)


fortinj66:
**Description**

During installation MCO does not create /var/home/core/.ssh/authorized_keys which should have the ignition bootstrap key in it.  This prevents core from logging into the server if https://github.com/openshift/machine-config-operator/pull/2087 is enabled as reading authorized_keys.d is disabled.  if https://github.com/openshift/machine-config-operator/pull/2087 is disabled, login succeeds as authorized_keys.d is read and the ignition key is available.

**Steps to reproduce the issue:**
Install OKD 4.7 
**Describe the results you received:**
unable to ssh into server after installation.  If I manually create the authorized_keys file I can log in successfully

**Describe the results you expected:**
successful ssh into servers

**Additional information you deem important (e.g. issue happens only occasionally):**

```
[root@os-utils-d02 qa-4.7]# oc get mc
NAME                                               GENERATEDBYCONTROLLER                      IGNITIONVERSION   AGE
00-master                                          451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
00-worker                                          451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
01-master-container-runtime                        451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
01-master-kubelet                                  451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
01-worker-container-runtime                        451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
01-worker-kubelet                                  451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
99-master-generated-registries                     451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
99-master-okd-extensions                                                                      3.1.0             15h
99-master-ssh                                                                                 3.1.0             15h
99-okd-master-disable-mitigations                                                             3.1.0             15h
99-okd-worker-disable-mitigations                                                             3.1.0             15h
99-worker-generated-registries                     451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
99-worker-okd-extensions                                                                      3.1.0             15h
99-worker-ssh                                                                                 3.1.0             15h
rendered-master-24161f6ba4983107f6f1799baa94a8c3   451aa6521d737f849302a98fae2dff213a99eff4   3.2.0             15h
[...]

PR #2387, opened 2021-02-03
Revert disabling authorized_keys.d  Pull 2087  

fortinj66:
revert  https://github.com/openshift/machine-config-operator/pull/2087
            until https://github.com/openshift/machine-config-operator/issues/2385 is fixed

The authorized_keys file is not being created on cluster install and with authorized_keys.d disabled ssh into the
 cluster nodes is not possible without manual intervention.  Once that issue is fixed authorized_keys.d can be disabled.
    
 The entry has to be commented out so that the original value does not get overwritten.


kikisdeliveryservice (changes requested):
can you clean up the commit, there should only be one.

fortinj66:
Yep...  I'm not sure why I got two...  sigh

fortinj66:
extra commit included...  closing for https://github.com/openshift/machine-config-operator/pull/2388

fortinj66:
closing for https://github.com/openshift/machine-config-operator/pull/2388

PR #2388, opened 2021-02-03
revert Pull #2087 until issue #2385 is fixed

fortinj66:
revert  https://github.com/openshift/machine-config-operator/pull/2087
            until https://github.com/openshift/machine-config-operator/issues/2385 is fixed

The authorized_keys file is not being created on cluster install and with authorized_keys.d disabled ssh into the
 cluster nodes is not possible without manual intervention.  Once that issue is fixed authorized_keys.d can be disabled.
    
 The entry has to be commented out so that the original value does not get overwritten.

### DIFF
--- a/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
+++ b/templates/common/_base/files/sshd-disable-ssh-key-dir.yaml
@@ -3,4 +3,4 @@ path: "/etc/ssh/sshd_config.d/10-disable-ssh-key-dir.conf"
 contents:
   inline: |
     # disable key lookup from ~/.ssh/authorized_keys.d/ on FCOS
-    AuthorizedKeysCommand none
+    ###AuthorizedKeysCommand none


kikisdeliveryservice:
Next time please just push to your existing branch/PR instead of opening a new one. =)

/assign @vrutkovs

fortinj66:
@kikisdeliveryservice 
thanks!
I ended up having to rebase and it didn't want to do the right thing...  Sometimes Git and I have a love/hate relationship

kikisdeliveryservice:
> @kikisdeliveryservice
> thanks!
> I ended up having to rebase and it didn't want to do the right thing... Sometimes Git and I have a love/hate relationship

After I rebase I do a `git push -f` to push to existing branch/pr. :)

fortinj66:
this will be fixed with https://github.com/openshift/machine-config-operator/pull/2393

bgilbert:
This (re-)creates a security vulnerability in OKD.  Any SSH keys injected during node provisioning will not be managed by the MCO and will remain for the lifetime of the node, even if they are removed from the MachineConfig.

PR #2389, opened 2021-02-03
Bug 1885186: handle Passwd.Users removal more gracefully

kikisdeliveryservice:
Still WIP bc I'd like to do more testing but this seems like a better approach in line with MCO rules

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -828,6 +828,8 @@ func reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineConfigDif
 				if err := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1]); err != nil {
 					return nil, err
 				}
+			} else if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) == 0 {
+				return nil, errors.New("ignition passwd user section contains unsupported changes: user core may not be deleted")
 			}
 		}
 	}
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -814,22 +814,19 @@ func reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineConfigDif
 			return nil, errors.New("ignition Passwd Groups section contains changes")
 		}
 		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
-			// check if the prior config is empty and that this is the first time running.
-			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
-			if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) >= 1 {
-				// there is an update to Users, we must verify that it is ONLY making an acceptable
-				// change to the SSHAuthorizedKeys for the user "core"
-				for _, user := range newIgn.Passwd.Users {
-					if user.Name != coreUserName {
-						return nil, errors.New("ignition passwd user section contains unsupported changes: non-core user")
-					}
-				}
-				glog.Infof("user data to be verified before ssh update: %v", newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
-				if err := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1]); err != nil {
-					return nil, err
-				}
-			} else if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) == 0 {
+			if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) == 0 {
 				return nil, errors.New("ignition passwd user section contains unsupported changes: user core may not be deleted")
+			}
+			// there is an update to Users, we must verify that it is ONLY making an acceptable
+			// change to the SSHAuthorizedKeys for the user "core"
+			for _, user := range newIgn.Passwd.Users {
+				if user.Name != coreUserName {
+					return nil, errors.New("ignition passwd user section contains unsupported changes: non-core user")
+				}
+			}
+			glog.Infof("user data to be verified before ssh update: %v", newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
+			if err := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1]); err != nil {
+				return nil, err
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


sinnykumari:
change looks fine but this is adding another bug which I am seeing during initial cluster install. When I launched a cluster with this PR using clusterbot, nodes doesn't have any sshAuthorizedKeys added into /home/core/.ssh/authorized_keys although rendered config does have one entry in sshAuthorizedKeys.

kikisdeliveryservice:
> change looks fine but this is adding another bug which I am seeing during initial cluster install. When I launched a cluster with this PR using clusterbot, nodes doesn't have any sshAuthorizedKeys added into /home/core/.ssh/authorized_keys although rendered config does have one entry in sshAuthorizedKeys.

ahhh clusterconfig.. thanks @sinnykumari

pkg/daemon/update.go
```diff
@@ -814,9 +814,7 @@ func reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineConfigDif
 			return nil, errors.New("ignition Passwd Groups section contains changes")
 		}
 		if !reflect.DeepEqual(oldIgn.Passwd.Users, newIgn.Passwd.Users) {
-			// check if the prior config is empty and that this is the first time running.
-			// if so, the SSHKey from the cluster config and user "core" must be added to machine config.
-			if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) >= 1 {
+			if len(newIgn.Passwd.Users) >= 1 {
```
yuqi-zhang, line 817: I just realized that you could delete the entire "core" user and we'd silently be ok with it. Maybe for safety purposes its worth checking here? (Doesn't have to be part of this PR, just wanted to note)

kikisdeliveryservice: this pr is related to users doing that via deleting the 99-xxx-ssh

yuqi-zhang: Oh wait, this PR is saying that if you delete the entire `core` user, you should clear the sshkeys.
Do we support removing the whole user? Its not like we actually delete the user from the system, so if you had e.g. provided a password, its still accessible

kikisdeliveryservice: ok discussed with jerry, and have a new path forward so will repush with some updates!

kikisdeliveryservice:
Still WIP need to add one more thing.

pkg/daemon/update.go
```diff
@@ -814,21 +814,20 @@ func reconcilable(oldConfig, newConfig *mcfgv1.MachineConfig) (*machineConfigDif
...
-				// change to the SSHAuthorizedKeys for the user "core"
-				for _, user := range newIgn.Passwd.Users {
-					if user.Name != coreUserName {
-						return nil, errors.New("ignition passwd user section contains unsupported changes: non-core user")
-					}
-				}
-				glog.Infof("user data to be verified before ssh update: %v", newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1])
-				if err := verifyUserFields(newIgn.Passwd.Users[len(newIgn.Passwd.Users)-1]); err != nil {
-					return nil, err
+			if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) == 0 {
```
yuqi-zhang, line 817: Just curious, isn't the `len(oldIgn.Passwd.Users) > 0` check technically redundant now that the rest of the checks are pulled out of the if conditional?

This actually does introduce a behaviour change. Consider the following scenario: you want to give a node a ssh passwordHash via cluster installation manifests. This is fine because ignition would write that password, and so long as you didn't change it, we allowed it because it was a ignition written update and not a MC update.

Where this behaviour changes now is that we no longer allow it for the "initial update" from empty MC to initial config. This is because before, `verifyUserFields` (which doesn't allow anything except a core user and an sshkey) would fall under `if len(oldIgn.Passwd.Users) > 0 && len(newIgn.Passwd.Users) >= 1 {` , and the check gets skipped. Now it always gets verified and fails, so your nodes don't install properly (mcd-firstboot fails).

Now this would have eventually come back to bite you if you ever tried to update sshkeys, so we didn't have a great behaviour anyways. The underlying question really is, for ignition supported MC-not supported fields, if you don't change them (e.g. provisioning disks, which is officially supported), what should our behaviour be. Fortunately I don't think we have anywhere documented for other `passwd` sections, so maybe I'm being overly cautious here. If we want to error better that would be good too, because it may be confusing if this ever happened.

yuqi-zhang (commented):
Overall I think the approach is good, just a question about a detail

pkg/daemon/update.go
fortinj66: so right now, at least for OKD, ~core/.ssh/authorized_keys is not created during ignition...  doesn't work in either 4.6 nor 4.7.  There is a key created in ~core/.ssh/authorized_keys.d/

The validation for rendered_master_XXX and rendered_worker_XXX seems to only check for units and files, not Passwd...  99-master-ssh and 99-worker-ssh are in the rendered mc, but never applied

fortinj66:
> > change looks fine but this is adding another bug which I am seeing during initial cluster install. When I launched a cluster with this PR using clusterbot, nodes doesn't have any sshAuthorizedKeys added into /home/core/.ssh/authorized_keys although rendered config does have one entry in sshAuthorizedKeys.
> 
> ahhh clusterconfig.. thanks @sinnykumari

https://github.com/openshift/machine-config-operator/issues/2385

pkg/daemon/update.go
kikisdeliveryservice: @fortinj66 this PR isn't intended for OKD, I think you need to ping on your open issue since OKD has a different way it deals with keys which I believe they are working on

kikisdeliveryservice:
> 
> I added this as a bug a few days ago. #2385

This pr isn't intended to address existing OKD issues. That team is aware and working on it afaik.  Sinny was speaking about a change that I introduced in my PR accidentally in my WIP PR that has since been reverted.

pkg/daemon/update.go
fortinj66: OK, thanks...  but it would seem that @sinnykumari sees the same issue that I see with the authorized_keys file not created...

fortinj66:
> > I added this as a bug a few days ago. #2385
> 
> This pr isn't intended to address existing OKD issues. That team is aware and working on it afaik. Sinny was speaking about a change that I introduced in my PR accidentally in my WIP PR for RHCOS that has since been reverted.

Ok...  Thank you

pkg/daemon/update.go
kikisdeliveryservice: yes that was via a bug that i briefly introduced via an older version of this pr that no longer exists.

kikisdeliveryservice:
Prev run was all green, just cleaned up commits and repushed no code changes.

kikisdeliveryservice:
> Prev run was all green, just cleaned up commits and repushed no code changes.

Def jinxed myself on this one..

/retest

yuqi-zhang (approved):
/lgtm

This will now explicitly disallow deleting the core user (via deleting the sshkey config). We will look to improve on this later

PR #2377, opened 2021-02-01
Bug 1919271: NM resolve prepender: Update for systemd-resolved logic. systemd-resolved requires restart after dropin created

fortinj66:
**- What I did**
- remove OS check.  If node has systemd-resolved enabled this will work
- Check to see if systemd-resolved is enabled.  Otherwise skip
- Check if /etc/systemd/resolved.conf.d/60-kni.conf exists.  If it doesn't create it and restart systemd-resolved if active
- Add logging

**- How to verify it**

**- Description for the changelog**
NM resolve prepender:  Update for systemd-resolved logic

PR #2393, opened 2021-02-06
OKD: Add authorized keys if missing after ignition

fortinj66:
OKD Ignite does not create /home/core/.ssh/authorized_keys.  It creates /home/core/.ssh/authorized_keys.d/ignition instead.
This patch creates /home/core/.ssh/authorized_keys from /home/core/.ssh/authorized_keys.d/ignition if missing during
machine-config-daemon-firstboot.service.

This puts the system in sync with rendered machine configs.

I do not believe this will effect OCP.  My understanding is that ignite on OCP creates  /home/core/.ssh/authorized_keys so this will never be applied  (Modified so that it would handle no ssh keys being passed as cluster creation)

**- What I did**

Created writeMissingAuthorizedKeys() function which is called in updateFiles()

**- How to verify it**
/home/core/.ssh/authorized_keys file exists

Add authorized keys if missing after ignition

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1228,6 +1228,11 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error 
 	if err := dn.writeUnits(newIgnConfig.Systemd.Units); err != nil {
 		return err
 	}
+	// In OKD /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
+	// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+	if err := dn.writeMissingAuthorizedKeys(); err != nil {
+		return err
+	}
 	if err := dn.deleteStaleData(&oldIgnConfig, &newIgnConfig); err != nil {
 		return err
 	}
@@ -1491,6 +1496,39 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 		return fmt.Errorf("error running preset on unit: %s", stdouterr)
 	}
 	glog.Infof("Preset systemd unit %s", unit.Name)
+	return nil
+}
+
+// writeMissingAuthorizedKeys: In OKD /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
+// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+func (dn *Daemon) writeMissingAuthorizedKeys() error {
+	authKeyPath := filepath.Join(coreUserSSHPath, "authorized_keys")
+	authKeyFragPath := filepath.Join(coreUserSSHPath, "authorized_keys.d", "ignition")
+
+	// if authKeyFragPath doesn't exist return as the rest is unneeded
+	if _, err := os.Stat(authKeyFragPath); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+
+	if _, err := os.Stat(authKeyPath); err != nil {
+		if os.IsNotExist(err) {
+			glog.Infof("Creating missing %s", authKeyPath)
+			key, err := ioutil.ReadFile(authKeyFragPath)
+			if err != nil {
+				glog.Infof("Unable to read %s", authKeyFragPath)
+				return err
+			}
+			if err := dn.atomicallyWriteSSHKey(string(key)); err != nil {
+				glog.Infof("Unable to create %s", authKeyPath)
+				return err
+			}
+		} else {
+			glog.Infof("exists, keys already written")
+		}
+	}
 	return nil
 }
 


fortinj66:
If this is accepted, https://github.com/openshift/machine-config-operator/pull/2388 will not be needed

fortinj66:
Example log from my test cluster (OKD vSphere IPI):
```
Feb 06 19:34:14 localhost systemd[1]: Starting Machine Config Daemon Firstboot...
Feb 06 19:34:14 localhost machine-config-daemon[2382]: I0206 19:34:14.103968    2382 rpm-ostree.go:261] Running captured: rpm-ostree status --json
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.760772    2382 daemon.go:224] Booted osImageURL:  (33.20210117.3.1)
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.862158    2382 daemon.go:231] Installed Ignition binary version: 2.9.0
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.865477    2382 update.go:598] Checking Reconcilable for config mco-empty-mc to rendered-master-9c89b3d73b2747f940dd156d73c30d3f
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.866931    2382 update.go:1931] Starting update from mco-empty-mc to rendered-master-9c89b3d73b2747f940dd156d73c30d3f: &{osUpdate:true kargs:false fips:false passwd
:false files:false units:false kernelType:false extensions:true}
Feb 06 19:34:15 localhost root[2869]: machine-config-daemon[2382]: Starting update from mco-empty-mc to rendered-master-9c89b3d73b2747f940dd156d73c30d3f: &{osUpdate:true kargs:false fips:false passwd:false files:false units:false ker
nelType:false extensions:true}
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.871445    2382 update.go:1216] Updating files
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.871925    2382 update.go:1510] Creating missing /home/core/.ssh/authorized_keys
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.872032    2382 update.go:1762] Writing SSHKeys at "/home/core/.ssh/authorized_keys"
Feb 06 19:34:15 localhost machine-config-daemon[2382]: I0206 19:34:15.873484    2382 update.go:1294] Deleting stale data
```

fortinj66:
sorry for the churn...  git and I are not seeing eye to eye after I've made formatting issue...

vrutkovs:
Looks good to me. Needs to be squashed into a single commit with a short description (Ignition may create dropins in authorized_keys.d instead of a single authorized_keys file, so in order to properly manage this file MCO should convert it into a single authorized_keys entry)

pkg/daemon/update.go
```diff
@@ -1228,6 +1228,11 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	if err := dn.writeUnits(newIgnConfig.Systemd.Units); err != nil {
 		return err
 	}
+	// /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
```
kikisdeliveryservice, line 1231: Please add note that this is relevant to OKD ie "In OKD /var/home///..."

pkg/daemon/update.go
```diff
@@ -1494,6 +1499,31 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 	return nil
 }
 
+// writeMissingAuthorizedKeys /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
```
kikisdeliveryservice, line 1502: same comment as above

pkg/daemon/update.go
```diff
@@ -1494,6 +1499,31 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
...
+			if err != nil {
+				glog.Infof("Unable to read %s", authKeyFragPath)
+				return err
+			}
+			if err := dn.atomicallyWriteSSHKey(string(key)); err != nil {
+				glog.Infof("Unable to create %s", authKeyPath)
+				return err
+			}
+		} else {
+			glog.Infof("%s already exists", authKeyPath)
```
kikisdeliveryservice, line 1521: "exists, keys already written"

pkg/daemon/update.go
```diff
@@ -1494,6 +1499,31 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 	return nil
 }
 
+// writeMissingAuthorizedKeys /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
+// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+func (dn *Daemon) writeMissingAuthorizedKeys() error {
+	authKeyPath := filepath.Join(coreUserSSHPath, "authorized_keys")
+	authKeyFragPath := filepath.Join(coreUserSSHPath, "authorized_keys.d", "ignition")
+
+	if _, err := os.Stat(authKeyPath); err != nil {
```
kikisdeliveryservice, line 1508: Is this correct? It is valid for the authKeyPath in OCP to be missing if the user did not supply ssh keys. We do not want to create it in all cases only when it's OKD and there are keys to write.

kikisdeliveryservice (changes requested):
Some questions/comments

pkg/daemon/update.go
fortinj66: the authorized_keys.d/ignition file wouldn't exist in OCP correct? We could also add a check for the ignition file and then do the copy...

fortinj66:
> Looks good to me. Needs to be squashed into a single commit with a short description (Ignition may create dropins in authorized_keys.d instead of a single authorized_keys file, so in order to properly manage this file MCO should convert it into a single authorized_keys entry)

Once any additional changes needed are made and accepted I'll squash this down and update the commit comment...

pkg/daemon/update.go
fortinj66: That would also take care of the OKD case when no key is given at build time

pkg/daemon/update.go
```diff
@@ -1494,6 +1499,39 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 	return nil
 }
 
+// writeMissingAuthorizedKeys: In OKD /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
+// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+func (dn *Daemon) writeMissingAuthorizedKeys() error {
+	authKeyPath := filepath.Join(coreUserSSHPath, "authorized_keys")
+	authKeyFragPath := filepath.Join(coreUserSSHPath, "authorized_keys.d", "ignition")
+
+	// if authKeyFragPath doesn't exist return as the rest is unneeded
```
kikisdeliveryservice, line 1508: so this covers OCP & OKD if there are no sshkeys?

fortinj66: I believe so...  I can't test OCP though...

darkmuggle: The `core` user is created by Ignition by default unless the user opts out.

I worried a bit by two assumptions:
- not checking to see if this is a CoreOS variant (i.e. not RHEL)
- the assumption that the user exists and that user's path. The assumption of the user's path could be problematic. 

Perhaps something like:

```
	if !dn.os.IsCoreOSVariant() {
		return nil
	}

	user, err := user.Lookup("core")
	if err != nil {
		return nil
	}

	authKeyPath := filepath.Join(user.HomeDir, "authorized_keys")
	authKeyFragPath := filepath.Join(user.HomeDir, "authorized_keys.d", "ignition")

       <do your thing...>
```

pkg/daemon/update.go
```diff
@@ -1228,6 +1228,11 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
 	if err := dn.writeUnits(newIgnConfig.Systemd.Units); err != nil {
 		return err
 	}
+	// In OKD /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
```
sinnykumari, line 1231: I may have missed conversation around this(if already happened). What is the reason OKD by default writes to /var/home/core/.ssh/authorized_keys.d/ignition which differs from default OCP behavior?

pkg/daemon/update.go
```diff
@@ -1228,6 +1228,11 @@ func (dn *Daemon) updateFiles(oldConfig, newConfig *mcfgv1.MachineConfig) error
...
+	// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+	if err := dn.writeMissingAuthorizedKeys(); err != nil {
```
sinnykumari, line 1233: If this only impacts OKD, then I think it is better to add additional checks to run this only on FCOS nodes.

pkg/daemon/update.go
vrutkovs: IIUC the new Ignition does that, @LorbusChris or @bgilbert might know more

pkg/daemon/update.go
kikisdeliveryservice: Agree with @sinnykumari :100:

darkmuggle (changes requested):
~In general, I think its a bit clearer to check for the user and determine the path from asking Linux (in case the user has changed the home directory via Ignition) 

Also, why doesn't OKD create the `core` user via Ignition like RHCOS? https://github.com/openshift/okd-machine-os is the build location and ensuring that the CoreOS user exists via an Ignition Fragment seems like another suitable path. ~

pkg/daemon/update.go
fortinj66: @sinnykumari  ignition 2.x writes to there by default unless there is an override environment variable set... 

In ignition:  internal/exec/util/passwd.go in AuthorizeSSHKeys()

```                
                if distro.WriteAuthorizedKeysFragment() {
                        path, err = u.JoinPath(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition")
                        if err == nil {
                                err = writeAuthKeysFile(usr, path, []byte(ks))
                        }
                } else {
                        path, err = u.JoinPath(usr.HomeDir, ".ssh", "authorized_keys")
                        if err == nil {
                                err = writeAuthKeysFile(usr, path, []byte(ks))
                        }
                }
```

and from distro:
```
        // writeAuthorizedKeysFragment indicates whether to write SSH keys
        // specified in the Ignition config as a fragment to
        // ".ssh/authorized_keys.d/ignition" ("true"), or to
        // ".ssh/authorized_keys" ("false").
        writeAuthorizedKeysFragment = "true"
```
so its always true

but it looks like it can be over-written with an environment variable:

```func WriteAuthorizedKeysFragment() bool {
	return bakedStringToBool(fromEnv("WRITE_AUTHORIZED_KEYS_FRAGMENT", writeAuthorizedKeysFragment))
}

func fromEnv(nameSuffix, defaultValue string) string {
	value := os.Getenv("IGNITION_" + nameSuffix)
	if value != "" {
		return value
	}
	return defaultValue
}
```

I'm not sure how RHCOS/OCP is doing this but I'm guessing the environment variable is set in RHCOS...

pkg/daemon/update.go
fortinj66: This will only run on OKD nodes as OCP doesn't use /var/home/core/.ssh/authorized_keys.d/ 

that being said, is there a method to test OCP vs. OKD?

fortinj66:
> In general, I think its a bit clearer to check for the user and determine the path from asking Linux (in case the user has changed the home directory via Ignition)
> 
> Also, why doesn't OKD create the `core` user via Ignition like RHCOS? https://github.com/openshift/okd-machine-os is the build location and ensuring that the CoreOS user exists via an Ignition Fragment seems like another suitable path.

See the above ignition code...  It seems to default differently from RHCOS...

pkg/daemon/update.go
vrutkovs: CI should be creating test payloads for OCP and OKD.
See `build-log.txt` in `okd-e2e-aws` and `e2e-aws`, and it should have a line like `Create release image registry.build01.ci.openshift.org/ci-op-hs9i1yr4/release:latest` - this is the payload w/ test change.

However you might not have a pull secret to fetch it, so I'll help with testing those

pkg/daemon/update.go
bgilbert: The environment variable is for CI testing.  The variables in the distro package can be overridden at compile time, and the RHCOS package does so.

`writeAuthorizedKeysFragment = "true"` is the recommended default for an OS, allowing multiple tools to independently manage their own sets of SSH keys.  But on RHCOS there is only one set of keys, and it was decided that Ignition and the MCO would manage `~/.ssh/authorized_keys` directly.

bgilbert:
FCOS [does create the core user](https://github.com/coreos/fedora-coreos-config/blob/e68419cd04de2a68ba262ccd70188ae5f710750c/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/00-core.ign) from Ignition by default.

pkg/daemon/update.go
fortinj66: RHCOS is but FCOS is not doing this and is using the default from ignition and creating authorized_keys.d/ignition which is not what MCO is expecting and because of https://github.com/openshift/machine-config-operator/pull/2087, using authorized_keys.d/ignition is disabled so there is a need to create authorized_keys for logging in to the nodes...

darkmuggle:
I'm opposed to this change. 

After looking over the code and re-looking at the problem the issues is that FCOS use `.ssh/authorized_keys.d/ignition` while RCHOS uses `.ssh/authroized_keys`.  We should fix the `atomicallyWriteSSHKey` function to deal _safely_ with the differences.

fortinj66:
> FCOS [does create the core user](https://github.com/coreos/fedora-coreos-config/blob/e68419cd04de2a68ba262ccd70188ae5f710750c/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf/00-core.ign) from Ignition by default.

Yes, but the ssh key that is created is in /var/home/core/.ssh/authorized_keys.d not /var/home/core/.ssh/authorized_keys.  See the ignition code above for details.

pkg/daemon/update.go
bgilbert: #2087 was implemented specifically to address the difference in the paths of Ignition-written SSH keys between FCOS and RHCOS, ensuring that Ignition won't write keys on FCOS that the MCO then fails to manage.  The MCO still manages `~/.ssh/authorized_keys` on FCOS nodes.

fortinj66:
> I'm opposed to this change.
> 
> After looking over the code and re-looking at the problem the issues is that FCOS use `.ssh/authorized_keys.d/ignition` while RCHOS uses `.ssh/authroized_keys`. We should fix the `atomicallyWriteSSHKey` function to deal _safely_ with the differences.

How is rewriting `atomicallyWriteSSHKey` going to fix this...  Unless you are going to check for the existence of the key and create it the problem will still exist...

Or revert https://github.com/openshift/machine-config-operator/pull/2087 which is what revealed this issue to begin with and let MCO manage quthorized_keys.d

Note that OKD 4.7 is currently broken and you cannot login via ssh with the provided key.

pkg/daemon/update.go
fortinj66: Except that since the authorized_keys.d/ignition file is never moved/copied to authorized_keys, ssh into the server is disabled, until and if you can update 99_master_ssh and/or 99_worker_ssh.

And if there are issues with building the cluster, debugging becomes very difficult as you cannot ssh into the nodes...

bgilbert:
@fortinj66 Could you expand on what problem this is solving?  The current design on MCO/FCOS is that `authorized_keys` will not be written until the first time the MCO reconciles the MachineConfig.  That does mean that on FCOS, unlike on RHCOS, the machine cannot be SSHed to until the MCO is running.  But your change still has that property, does it not?

fortinj66:
> @fortinj66 Could you expand on what problem this is solving? The current design on MCO/FCOS is that `authorized_keys` will not be written until the first time the MCO reconciles the MachineConfig. That does mean that on FCOS, unlike on RHCOS, the machine cannot be SSHed to until the MCO is running. But your change still has that property, does it not?

Unfortunately, I don't have an OCP system to look at and thus can't look at RHCOS.  

However, on FCOS, the core ssh key is written to /var/home/core/.ssh/authorized_keys.d/ignition during ignition.  MCO does not reconcile authorized_keys in either the first boot or subsequent processing.  FILES and UNIT seem to get reconciled, but PASSWD (?) does not.  The data is in the rendered configs, but never written to disk.  If I update 99_master_ssh, after the cluster is created, the authorized_keys file is created as updated.  but for access/debugging that can be far too late.

I cloned and rebuilt MCO initially disabling https://github.com/openshift/machine-config-operator/pull/2087 so that I would be able to ssh into the nodes and access logs, etc.

So this is a solution...  

If there is a better solution, I'm all for it.  This is me, observing an issue, reporting it, and actually offering a potential solution.

kikisdeliveryservice:
Q: I feel like we've had this discussion before and the take-away was that OKD was going to fix this issue elsewhere. Was that shelved?

fortinj66:
> That does mean that on FCOS, unlike on RHCOS, the machine cannot be SSHed to until the MCO is running. But your change still has that property, does it not?

With this change I can log into the server machine-config-daemon-firstboot.service completes.

Here is the example log from my cluster with this change:

```
Feb 08 22:17:23 localhost systemd[1]: Starting Machine Config Daemon Firstboot...
Feb 08 22:17:24 localhost machine-config-daemon[2384]: I0208 22:17:24.041365    2384 rpm-ostree.go:261] Running captured: rpm-ostree status --json
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.103372    2384 daemon.go:224] Booted osImageURL:  (33.20210117.3.1)
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.210231    2384 daemon.go:231] Installed Ignition binary version: 2.9.0
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.212791    2384 update.go:598] Checking Reconcilable for config mco-empty-mc to rendered-master-b62a234c88f45eccb72d33c17c34a532
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.214076    2384 update.go:1939] Starting update from mco-empty-mc to rendered-master-b62a234c88f45eccb72d33c17c34a532: &{osUpdate:true kargs:false fips:false passw>
Feb 08 22:17:26 localhost root[2871]: machine-config-daemon[2384]: Starting update from mco-empty-mc to rendered-master-b62a234c88f45eccb72d33c17c34a532: &{osUpdate:true kargs:false fips:false passwd:false files:false units:false ke>
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.221307    2384 update.go:1216] Updating files
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.222066    2384 update.go:1518] Creating missing /home/core/.ssh/authorized_keys
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.222145    2384 update.go:1770] Writing SSHKeys at "/home/core/.ssh/authorized_keys"
Feb 08 22:17:26 localhost machine-config-daemon[2384]: I0208 22:17:26.223390    2384 update.go:1294] Deleting stale data
```

fortinj66:
Part of the issues I see is that RHCOS != FCOS.  FCOS behaves differently.  Look at what we had to do to make DNS work properly because of systemd-resolved (https://github.com/openshift/machine-config-operator/pull/2377)

I think this is one of those differences that has now reared its head.

Note that OKD 4.6 has the same behavior, but does not have https://github.com/openshift/machine-config-operator/pull/2087 applied.  I'm guessing that 4.5 has the same...

darkmuggle:
I commented in the bug. However, the must-gather indicates that the cluster is _very_ unhappy, from the start. Before we proceed any further, can we confirm the bug with a clean must-gather?

fortinj66:
> I commented in the bug. However, the must-gather indicates that the cluster is _very_ unhappy, from the start. Before we proceed any further, can we confirm the bug with a clean must-gather?

I can rebuild my cluster.  I'm guessing you want it from the latest 4.7 nightly?  I'll run the "vanilla" install without my changes...

kikisdeliveryservice:
I'd like to know what problem this is really solving bc I thought we already came to a final solution last year on handling authorized_keys in OKD via: https://github.com/openshift/machine-config-operator/pull/2087

fortinj66:
> I'd like to know what problem this is really solving bc I thought we already came to a final solution last year on handling authorized_keys in OKD via: #2087

#2087 breaks OKD ssh authentication because an authorized_keys file for the core user is never created and the authorized_keys.d/ignition file is now unavailable/disabled because of #2087 

@vrutkovs has reported this also...  https://github.com/openshift/machine-config-operator/pull/2283

I'll post the directory structure once the cluster is up and I can `oc debug node/xxx` into it...

fortinj66:
I'm also going to provision an OCP cluster...

kikisdeliveryservice:
> > I'd like to know what problem this is really solving bc I thought we already came to a final solution last year on handling authorized_keys in OKD via: #2087
> 
> #2087 breaks OKD ssh authentication because an authorized_keys file for the core user is never created and the authorized_keys.d/ignition file is now unavailable/disabled because of #2087
> 
> @vrutkovs has reported this also... #2283
> 
> I'll post the directory structure once the cluster is up and I can `oc debug node/xxx` into it...

This was a known tradeoff agreed up on the initial implementation. This PR doesn't feel like the correct fix to mitigating that tradeoff and as we discussed prior more work elsewhere is probably required.

vrutkovs:
I'm fine with leaving that change if FCOS adopts setting `WRITE_AUTHORIZED_KEYS_FRAGMENT` env var (just like RHCOS apparently does).

fortinj66:
### New OKD cluster.  ssh with key does not work...

[Newest must-gather](https://www.dropbox.com/s/orpmpwqi3xtx3j2/must-gather.tar.gz?dl=0)

**OKD Directory Structure for core after cluster install**

```
sh-5.0# pwd
/var/home/core
sh-5.0# find .
.
./.bash_logout
./.bash_profile
./.bashrc
./.ssh
./.ssh/authorized_keys.d
./.ssh/authorized_keys.d/ignition
```

**OCP Directory Structure for core after cluster install**
```
sh-4.4# pwd
/var/home/core
sh-4.4# find .
.
./.bash_logout
./.bash_profile
./.bashrc
./.ssh
./.ssh/authorized_keys
sh-4.4#
```

If you look at the logs for ignition, you can see where the files are created...

**OKD**
```
Feb 09 22:04:37 localhost ignition[676]: INFO     : Ignition 2.9.0
Feb 09 22:04:37 localhost ignition[676]: INFO     : Stage: files
Feb 09 22:04:37 localhost ignition[676]: INFO     : reading system config file "/usr/lib/ignition/base.d/00-core.ign"
Feb 09 22:04:37 localhost ignition[676]: DEBUG    : parsing config with SHA512: ff6a5153be363997e4d5d3ea8cc4048373a457c48c4a5b134a08a30aacd167c1e0f099f0bdf1e24c99ad180628cd02b767b863b5fe3a8fce3fe1886847eb8e2e
Feb 09 22:04:37 localhost ignition[676]: INFO     : reading system config file "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"
Feb 09 22:04:37 localhost ignition[676]: DEBUG    : parsing config with SHA512: a30a1921169d5a3b58ef9b25de60783be1add6ea8d05fd44a0746cb60dd1b8a8b34ab51eec5eb14eecc2df2ab6ba1cd3fd7351eed65793d22316ab262a857d95
Feb 09 22:04:37 localhost ignition[676]: INFO     : no config dir at "/usr/lib/ignition/base.platform.d/vmware"
Feb 09 22:04:37 localhost ignition[676]: INFO     : Adding "root-ca" to list of CAs
Feb 09 22:04:37 localhost ignition[676]: INFO     : files: ensureUsers: op(1): [started]  creating or modifying user "core"
Feb 09 22:04:37 localhost ignition[676]: DEBUG    : files: ensureUsers: op(1): executing: "useradd" "--root" "/sysroot" "--create-home" "--password" "*" "--comment" "CoreOS Admin" "--groups" "adm,sudo,systemd-journal,wheel" "core"
Feb 09 22:04:37 localhost ignition[676]: INFO     : files: ensureUsers: op(1): [finished] creating or modifying user "core"
Feb 09 22:04:37 localhost ignition[676]: INFO     : files: ensureUsers: op(2): [started]  adding ssh keys to user "core"
Feb 09 22:04:37 localhost ignition[676]: wrote ssh authorized keys file for user: core
```

**OCP**
```
Feb 09 22:33:57 localhost ignition[985]: INFO     : Ignition 2.6.0
Feb 09 22:33:57 localhost ignition[985]: INFO     : Stage: files
Feb 09 22:33:57 localhost ignition[985]: INFO     : reading system config file "/usr/lib/ignition/base.ign"
Feb 09 22:33:57 localhost ignition[985]: DEBUG    : parsing config with SHA512: ff6a5153be363997e4d5d3ea8cc4048373a457c48c4a5b134a08a30aacd167c1e0f099f0bdf1e24c99ad180628cd02b767b863b5fe3a8fce3fe1886847eb8e2e
Feb 09 22:33:57 localhost ignition[985]: INFO     : Adding "root-ca" to list of CAs
Feb 09 22:33:57 localhost ignition[985]: INFO     : files: ensureUsers: op(1): [started]  creating or modifying user "core"
Feb 09 22:33:57 localhost ignition[985]: DEBUG    : files: ensureUsers: op(1): executing: "useradd" "--root" "/sysroot" "--create-home" "--password" "*" "--comment" "CoreOS Admin" "--groups" "adm,sudo,systemd-journal,wheel" "core"
Feb 09 22:33:57 localhost ignition[985]: INFO     : files: ensureUsers: op(1): [finished] creating or modifying user "core"
Feb 09 22:33:57 localhost ignition[985]: INFO     : files: ensureUsers: op(2): [started]  adding ssh keys to user "core"
Feb 09 22:33:57 localhost ignition[985]: INFO     : files: ensureUsers: op(2): [finished] adding ssh keys to user "core"
Feb 09 22:33:57 localhost ignition[985]: wrote ssh authorized keys file for user: core
```

darkmuggle:
If the reason for this PR is a bootstrap, and OKD already uses a custom installer, it seems that the better place to handle the drift is in the custom installer, no? This allows the MCO and Ignition to maintain the aforementioned agreement while providing for the debuggability. 

https://github.com/darkmuggle/installer/commit/ca7adb5b5305209ab4f9d4a21fbdfa24070b0c1f (YMMV, POC commit)

fortinj66:
> If the reason for this PR is a bootstrap, and OKD already uses a custom installer, it seems that the better place to handle the drift is in the custom installer, no? This allows the MCO and Ignition to maintain the aforementioned agreement while providing for the debuggability.
> 
> [darkmuggle/installer@ca7adb5](https://github.com/darkmuggle/installer/commit/ca7adb5b5305209ab4f9d4a21fbdfa24070b0c1f) (YMMV, POC commit)

I don't really know anything about the installer but I'm OK with the concept...  I tested the symlink and it works, and MCO correctly removes the symlink and creates a regular file when I update 99_master_ssh as expected

darkmuggle:
> @vrutkovs will need to chime in on this...

IMO, a viable workaround that resolves the problem without changing the MCO and maintains the agreements between various stakeholders is the preferable path forward for the time being.  Since the FCOS branch for the installer is for OKD use-cases it seems like an appropriate place to handle this specific bootstrap problem. Without a clear argument for why using the installer to handle this nuanced case will require further review and stakeholder input (and thus delay merging).

vrutkovs:
>OKD already uses a custom installer

Not for long, we're about to get merged in master soon.

>a viable workaround that resolves the problem without changing the MCO

Why should we not change MCO? There two locations where Ignition may write an ssh key - and MCO only manages one of them. It sounds like MCO is the component which needs to be updated to support both? And this PR does exactly that?

fortinj66:
Also, with the addition of a couple of lines of code, we can remove the authorized_keys.d/ignition file and eliminate the need for https://github.com/openshift/machine-config-operator/pull/2087

Remember, this only runs at first boot and is a noop for RHCOS/OCP until and if RHCOS implements authorized_keys.d

darkmuggle:
> Why should we not change MCO? There are two locations where Ignition may write an ssh key - and MCO only manages one of them. It sounds like MCO is the component which needs to be updated to support both? And this PR does exactly that?

FCOS is using [1] which is tried _after_ `~/.ssh/authorized_keys`. The reason why the MCO should not manage both locations is that the FCOS implementation of `.ssh/authorized_keys.d` via [2] is unmanaged by the MCO -- the MCO doesn't create the files. Rather, Ignition wrote a file to a specific location. Assuming that the MCO has ownership over `~/.ssh/authorized_keys.d/ignition` is a bit pretentious and when RHCOS supports `~/.ssh/authorized_keys` the MCO will have to develop a more nuanced understanding of which configuration it owns instead of blindly assuming the ownership without a clear design. 

[1] ```
 AuthorizedKeysCommand
             Specifies a program to be used to look up the user's public keys.  The program must be owned by root, not writable by group or others and specified by an absolute path.  Arguments to AuthorizedKeysCommand accept the tokens described in the
             TOKENS section.  If no arguments are specified then the username of the target user is used.

             The program should produce on standard output zero or more lines of authorized_keys output (see AUTHORIZED_KEYS in sshd(8)).  AuthorizedKeysCommand is tried after the usual AuthorizedKeysFile files and will not be executed if a matching key
             is found there.  By default, no AuthorizedKeysCommand is run.
```
[2] https://github.com/coreos/ssh-key-dir

fortinj66:
> > Why should we not change MCO? There are two locations where Ignition may write an ssh key - and MCO only manages one of them. It sounds like MCO is the component which needs to be updated to support both? And this PR does exactly that?
> 
> FCOS is using [1] which is tried _after_ `~/.ssh/authorized_keys`. The reason why the MCO should not manage both locations is that the FCOS implementation of `.ssh/authorized_keys.d` via [2] is unmanaged by the MCO -- the MCO doesn't create the files. Rather, Ignition wrote a file to a specific location. Assuming that the MCO has ownership over `~/.ssh/authorized_keys.d/ignition` is a bit pretentious. When RHCOS supports `~/.ssh/authorized_keys` the MCO will have to develop a more nuanced understanding of which configuration it owns instead of blindly assuming the ownership without a clear and agreed upon design.
> 
> The CoreOS and MCO teams made a decision earlier due a security concern and resolution before the RHCOS support for `~/.ssh/authorized_keys.d` is implemented to solve a debug pain point without full stakeholder discussion is premature.
> 
> [1]
> 
> ```
> AuthorizedKeysCommand
>             Specifies a program to be used to look up the user's public keys.  The program must be owned by root, not writable by group or others and specified by an absolute path.  Arguments to AuthorizedKeysCommand accept the tokens described in the
>             TOKENS section.  If no arguments are specified then the username of the target user is used.
> 
>             The program should produce on standard output zero or more lines of authorized_keys output (see AUTHORIZED_KEYS in sshd(8)).  AuthorizedKeysCommand is tried after the usual AuthorizedKeysFile files and will not be executed if a matching key
>             is found there.  By default, no AuthorizedKeysCommand is run.
> ```
> 
> [2] https://github.com/coreos/ssh-key-dir

@darkmuggle I don't think we are suggesting that the MCO support authorized_keys.d...  If fact we are offering the opposite solution which is to create the supported method of using authorized_keys.  The issue revolves around the fact that the authorized_keys file is never created anywhere; not in ignition and not during first boot.  

This forces the creation of authorized_keys from the ignition created authorized_keys.d/ignition file if it exists and it can be supported by the MCO as normal. and if we also delete authorized_keys.d/ignition afterwards, the security issue also goes away...

sinnykumari:
> This forces the creation of authorized_keys from the ignition created authorized_keys.d/ignition file if it exists and it can be supported by the MCO as normal. and if we also delete authorized_keys.d/ignition afterwards, the security issue also goes away...

Source of truth for MCO to apply any config is through MachineConfig which we can validate at any point of time by looking at applied MC and rendered config.  For authorized ssh keys, MCO manages and keep track of only ~/.ssh/authorized_keys file. If MCO will start writing ~/.ssh/authorized_keys.d/ignition implicitly into ~/.ssh/authorized_keys , MCO will have no way to know from where these extra ssh keys are coming as they won't get reflected in applied MC or rendered MC. Also, since MCO doesn't manages them, it won't be able to delete these keys which user may want later on.

fortinj66:
I surrender...  I'm not an owner or member so I don't have any real responsibility for this codebase.  Although I think my solution is appropriate I will not continue to argue for its merits.  

The ssh issue still remains without a solution.  I hope someone is able to come up with one before OKD 4.7 is released.

Regards,
--John

bgilbert:
Let me take a step back here.  @fortinj66, the designed and implemented behavior is this:

1. On FCOS, Ignition writes `~/.ssh/authorized_keys.d/ignition` because we want to support allowing multiple tools to manage their own sets of SSH keys.  On RHCOS, there is only one set of tools that manages SSH keys (the combination of Ignition on first boot + MCO afterward) so we just use `~/.ssh/authorized_keys`, and there are no plans to switch to `~/.ssh/authorized_keys.d/ignition`.
2. Once the MCO starts managing the system, it syncs `~/.ssh/authorized_keys` from the MachineConfig.  At that point, SSH keys should work on both FCOS and RHCOS.  The MCO does not touch `~/.ssh/authorized_keys.d/ignition`.
3. On FCOS, the MCO also needs to disable `~/.ssh/authorized_keys.d/ignition`, or else old keys will continue to work after they've been removed from the MachineConfig.

https://github.com/openshift/machine-config-operator/pull/2393#issuecomment-776314529 implies that we need an installer change for SSH keys to work on the bootstrap node on FCOS, I assume because the MCO doesn't manage the bootstrap node.  If step 2 in the above list is not working, that sounds like an MCO bug, and I assume we should fix it rather than changing the architecture.  But I'm a CoreOS person, not an MCO person, and I can't speak for the MCO folks.

fortinj66:
> 2\. Once the MCO starts managing the system, it syncs `~/.ssh/authorized_keys` from the MachineConfig.  At that point, SSH keys should work on both FCOS and RHCOS.  The MCO does not touch `~/.ssh/authorized_keys.d/ignition`.

I do not believe this is actually the case.  There does not seem to be a "sync" any where for authorized_keys in OCP or OKD.  If there is, please reference the code as I cannot find it and I've looked...  You can update it after the cluster is up and that works fine, but it does not create a new authorized_keys file from 99_master_ssh or 99_worker_ssh.  

I can't speak to the installer.  @vrutkovs has already made a comment above with his thoughts...

fortinj66:
> Source of truth for MCO to apply any config is through MachineConfig which we can validate at any point of time by looking at applied MC and rendered config. For authorized ssh keys, MCO manages and keep track of only ~/.ssh/authorized_keys file. If MCO will start writing ~/.ssh/authorized_keys.d/ignition implicitly into ~/.ssh/authorized_keys , MCO will have no way to know from where these extra ssh keys are coming as they won't get reflected in applied MC or rendered MC. Also, since MCO doesn't manages them, it won't be able to delete these keys which user may want later on.

And I agree, after the cluster is up... but there is not a sync during cluster creation, otherwise the authorized_keys file would get created and it doesn't.  for RHCOS, it is created at ignition.  for FCOS, it is created in authorized_keys.d/ignition.  All this PR does is copy authorized_keys.d/ignition to authorized_keys so that ssh will work...  and since this key matches what is in the bootstrap 99_master_ssh and 99_worker_ssh, MCO is also in sync.

That is were the difference/disconnect is.  This PR does not meet your architectural guidance, so I have withdrawn it

LorbusChris:
I wonder if this can be fixed with a symlink (like @darkmuggle suggested above, see https://github.com/darkmuggle/installer/commit/ca7adb5b5305209ab4f9d4a21fbdfa24070b0c1f) - but in https://github.com/openshift/okd-machine-os instead of the installer - this would then account for both bootstrap and non-bootstrap nodes.

vrutkovs:
> this would then account for both bootstrap and non-bootstrap nodes.

The symlink would be applied only post-pivot, so the host can't be ssh'd to if it can't pivot

LorbusChris:
Before pivoting the config to disable `~/.ssh/authorized_keys.d` doesn't exist yet and everything should still work, since the MCO hasn't written the sshd config fragment at that point. Or am I missing something?

vrutkovs:
Experimenting with this in https://github.com/openshift/okd-machine-os/pull/85

PR #2394, opened 2021-02-08
[WIP] Bug 1853264: refactor remaining metrics to remove cardinality

kikisdeliveryservice:
@lilic PTAL

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -55,11 +55,11 @@ var (
 			Help: "state of kubelet health monitor",
 		})
 
-	// MCDRebootErr logs success/failure of reboot and errors
-	MCDRebootErr = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
+	// MCDRebootErr tallys failed reboot attempts
+	MCDRebootErr = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
 			Name: "mcd_reboot_err",
-		}, []string{"node", "message", "err"})
+		}, []string{})
 
 	// MCDUpdateState logs completed update or error
 	MCDUpdateState = prometheus.NewGaugeVec(
@@ -90,7 +90,6 @@ func registerMCDMetrics() error {
 
 	MCDDrainErr.Set(0)
 	KubeletHealthState.Set(0)
-	MCDRebootErr.WithLabelValues("", "", "").Set(0)
 	MCDUpdateState.WithLabelValues("", "").Set(0)
 
 	return nil
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -34,12 +34,12 @@ var (
 			Help: "logs failed drain",
 		})
 
-	// MCDPivotErr shows errors encountered during pivot
-	MCDPivotErr = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
+	// MCDPivotErr flags error encountered during pivot
+	MCDPivotErr = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
 			Name: "mcd_pivot_err",
-			Help: "errors encountered during pivot",
-		}, []string{"node", "pivot_target", "err"})
+			Help: "error encountered during pivot",
+		}, []string{"pivot_target"})
 
 	// MCDState is state of mcd for indicated node (ex: degraded)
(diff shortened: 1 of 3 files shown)


kikisdeliveryservice:
sigh

/retest

pkg/daemon/metrics.go
```diff
@@ -34,12 +34,12 @@ var (
 			Help: "errors from failed drain",
 		}, []string{"node", "err"})
 
-	// MCDPivotErr shows errors encountered during pivot
+	// MCDPivotErr flags error encountered during pivot
 	MCDPivotErr = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
```
jan--f, line 39: Looks like this is never decremented so maybe this can be
```suggestion
	MCDPivotErr = prometheus.NewCounterVec(
		prometheus.CounterOpts{
```

pkg/daemon/metrics.go
```diff
@@ -55,11 +55,11 @@ var (
 			Help: "state of kubelet health monitor",
 		})
 
-	// MCDRebootErr logs success/failure of reboot and errors
+	// MCDRebootErr logs failure of reboot and tallys count
 	MCDRebootErr = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
```
jan--f, line 60: Same as `MCDPivotErr` above
```suggestion
	MCDRebootErr = prometheus.NewCounterVec(
		prometheus.CounterOpts{
```

pkg/daemon/metrics.go
```diff
@@ -55,11 +55,11 @@ var (
...
 			Name: "mcd_reboot_err",
-		}, []string{"node", "message", "err"})
+		}, []string{"node", "err"})
```
jan--f, line 62: Is it feasible to the remove the `err` label altogether? In the normal error case the label `failed to run reboot` doesn't seem to add any information.
The other usage should never be reached anyway, so I'd argue to probably remove that.

jan--f (commented):
I certainly don't have insight into the whole system so consider this some comments from a new guy.

lilic (commented):
Sorry for the late review, missed the ping!

Thanks, its a good first step towards improving metrics! I left some comments if you need more practical or concrete suggestions just comment what you actually want to achieve with these metrics then we can come up with something.

Feel free to email or send me a msg on slack if something is unclear!

kikisdeliveryservice:
Thanks for the great feedback and detailed pointers @lilic and @jan--f !

pkg/daemon/metrics.go
```diff
@@ -34,12 +34,12 @@ var (
 			Help: "errors from failed drain",
 		}, []string{"node", "err"})
 
-	// MCDPivotErr shows errors encountered during pivot
+	// MCDPivotErr flags error encountered during pivot
 	MCDPivotErr = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "mcd_pivot_err",
```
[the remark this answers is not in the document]
kikisdeliveryservice: thanks for the links ill spend some time going thru them :heart:

pkg/daemon/update.go
```diff
@@ -1958,13 +1958,13 @@ func (dn *Daemon) reboot(rationale string) error {
...
-		MCDRebootErr.WithLabelValues(dn.node.Name, "failed to run reboot", err.Error()).SetToCurrentTime()
+		MCDRebootErr.WithLabelValues(dn.node.Name, "failed to run reboot").Inc()
 	}
 
 	// wait to be killed via SIGTERM from the kubelet shutting down
 	time.Sleep(defaultRebootTimeout)
 
 	// if everything went well, this should be unreachable.
-	MCDRebootErr.WithLabelValues(dn.node.Name, "reboot failed", "this error should be unreachable, something is seriously wrong").SetToCurrentTime()
+	MCDRebootErr.WithLabelValues(dn.node.Name, "reboot failed: this error should be unreachable, something is seriously wrong").Inc()
```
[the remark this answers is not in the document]
kikisdeliveryservice: good point, I think I can make this more concise

pkg/daemon/metrics.go
kikisdeliveryservice: Good point, I'm going to spend some time rethinking this. Thank you!

PR #2395, opened 2021-02-08
daemon: safer signal handling for shutdown

darkmuggle:
This armors the signal handling for the daemon blocking any shutdown until _after_ an update is complete.
   
The old functions `catchIgnoreSIGTERM` and `cancelSIGTERM` really didn't do much (they used the mutex and then set a bool) but there were no checks in the signal handling.

This also increases the time for the MCD to shutdown from 5min to a 1hr. Since the MCD will shut down immediately when safe to do so this shouldn't have negative effects except in the case of an already unhappy node.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -91,7 +91,7 @@ type Daemon struct {
 	kubeletHealthzEnabled  bool
 	kubeletHealthzEndpoint string
 
-	updateActive     bool
+	// updateActiveLock is used to signal that an update operation is in progress
 	updateActiveLock sync.Mutex
 
 	nodeWriter NodeWriter
@@ -351,15 +351,19 @@ func (dn *Daemon) worker() {
 }
 
 func (dn *Daemon) processNextWorkItem() bool {
-	key, quit := dn.queue.Get()
-	if quit {
+	select {
+	case <-dn.stopCh:
 		return false
+	default:
+		key, quit := dn.queue.Get()
+		if quit {
+			return false
+		}
+		defer dn.queue.Done(key)
+
+		err := dn.syncHandler(key.(string))
+		dn.handleErr(err, key)
 	}
-	defer dn.queue.Done(key)
-
-	err := dn.syncHandler(key.(string))
-	dn.handleErr(err, key)
-
 	return true
 }
 
@@ -580,29 +584,28 @@ func (dn *Daemon) RunFirstbootCompleteMachineconfig() error {
 	return dn.reboot(fmt.Sprintf("Completing firstboot provisioning to %s", mc.GetName()))
 }
 
-// InstallSignalHandler installs the handler for the signals the daemon should act on
-func (dn *Daemon) InstallSignalHandler(signaled chan struct{}) {
-	termChan := make(chan os.Signal, 2048)
-	signal.Notify(termChan, syscall.SIGTERM)
+// InterruptHandler ensures that shutdown operations are blocked until an
+// an update operation is finished.
+func (dn *Daemon) InterruptHandler(signaled chan int) {
+	glog.Info("Guarding against sigterm signal")
 
 	// Catch SIGTERM - if we're actively updating, we should avoid
 	// having the process be killed.
 	// https://github.com/openshift/machine-config-operator/issues/407
 	go func() {
-		for sig := range termChan {
-			//nolint:gocritic
-			switch sig {
-			case syscall.SIGTERM:
-				dn.updateActiveLock.Lock()
-				updateActive := dn.updateActive
-				dn.updateActiveLock.Unlock()
-				if updateActive {
-					glog.Info("Got SIGTERM, but actively updating")
-				} else {
-					close(signaled)
-					return
-				}
-			}
+		sig := make(chan os.Signal, 1)
+		signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
+
+		select {
+		case <-dn.stopCh:
+			glog.Info("sigterm handler function stopped")
+			dn.updateActiveLock.Lock()
+			return
+		case <-sig:
+			glog.Warningf("sigterm received waiting to claim update lock")
+			dn.updateActiveLock.Lock()
+			glog.Info("lock obtained, proceeding to shutdown")
+			signaled <- 15
 		}
 	}()
 }
@@ -617,8 +620,8 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	glog.Info("Starting MachineConfigDaemon")
 	defer glog.Info("Shutting down MachineConfigDaemon")
 
-	signaled := make(chan struct{})
-	dn.InstallSignalHandler(signaled)
+	signaled := make(chan int)
+	dn.InterruptHandler(signaled)
 
 	if dn.kubeletHealthzEnabled {
 		glog.Info("Enabling Kubelet Healthz Monitor")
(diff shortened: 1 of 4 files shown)


darkmuggle:
I dropped the Systemd Inhibit functionality to blow reboots. During the team discussion today, the prevailing view is that the MCO cannot block a reboot.

pkg/daemon/daemon.go
```diff
@@ -634,16 +637,16 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 
 	go wait.Until(dn.worker, time.Second, stopCh)
 
-	for {
```
darkmuggle, line 637: The `for` loop is superfluous since. Once we get the `sigterm` we _should_ finish the work and then get out of the way since a `sigkill` could/likely follow.

pkg/daemon/daemon.go
```diff
@@ -580,29 +584,28 @@ func (dn *Daemon) RunFirstbootCompleteMachineconfig() error {
 	return dn.reboot(fmt.Sprintf("Completing firstboot provisioning to %s", mc.GetName()))
 }
 
-// InstallSignalHandler installs the handler for the signals the daemon should act on
-func (dn *Daemon) InstallSignalHandler(signaled chan struct{}) {
-	termChan := make(chan os.Signal, 2048)
-	signal.Notify(termChan, syscall.SIGTERM)
+// InteruptHandler ensures that shutdown operations are blocked until an
+// an update operation is finished.
```
sinnykumari, line 588: nit: "an"  is written twice

sinnykumari:
This really looks good and handles well interruption during update process.
/approve

Colin, can you please also take a look to make sure we are not missing anything here?
cc @cgwalters

pkg/daemon/daemon.go
```diff
@@ -634,16 +637,16 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
-		case err := <-exitCh:
-			// This channel gets errors from auxiliary goroutines like loginmonitor and kubehealth
-			glog.Warningf("Got an error from auxiliary tools: %v", err)
-		}
+	select {
+	case <-stopCh:
+		return nil
+	case sig := <-signaled:
+		glog.Warningf("shutdown of machine-config-daemon triggered via signal %d", sig)
+		return fmt.Errorf("shutdown of the machine-config-daemon trigger via syscall signal %d", sig)
```
cgwalters, line 645: I think getting `SIGTERM` isn't an error, it's a normal condition.  Which relates to one of the original goals I had here in that in the "idle" aka "not applying updates" case, we don't install a `SIGTERM` handler at all - we just let the kernel unilaterally kill the process.

I more recently posted about this here: https://internals.rust-lang.org/t/should-rust-programs-unwind-on-sigint/13800/11

See also e.g. https://github.com/nhorman/rng-tools/pull/72

And related to all this, ideally of course we do https://github.com/openshift/machine-config-operator/issues/1190 - then we shouldn't need to handle SIGTERM at all in the MCO.  With that if we're killed (or the machine power cycled/kernel froze) in the middle of an update, we either have the old or new system.

pkg/daemon/daemon.go
```diff
@@ -580,29 +584,28 @@ func (dn *Daemon) RunFirstbootCompleteMachineconfig() error {
 	return dn.reboot(fmt.Sprintf("Completing firstboot provisioning to %s", mc.GetName()))
 }
 
-// InstallSignalHandler installs the handler for the signals the daemon should act on
-func (dn *Daemon) InstallSignalHandler(signaled chan struct{}) {
-	termChan := make(chan os.Signal, 2048)
-	signal.Notify(termChan, syscall.SIGTERM)
+// InteruptHandler ensures that shutdown operations are blocked until an
+// update operation is finished.
+func (dn *Daemon) InteruptHandler(signaled chan int) {
```
cgwalters, line 589:
```suggestion
func (dn *Daemon) InterruptHandler(signaled chan int) {
```
(two r's)

cgwalters (approved):
Would you say the core thing this is fixing here is that currently we were ignoring `SIGTERM` if caught during an update, but we didn't then exit *after* the update had finished?  And that was causing systemd to time out and go on the `SIGKILL` spree?

Overall I think this looks improved; one optional comment.

I find it hard to review all of this for correctness though - I'd reiterate that I think we really want to make this whole thing transactional w/ostree support.  I hope at some point in the next few months to land some infrastructure for that.

cgwalters:
Tangentially related to this...one thing that could help in the MCO design is if we had a "dual nature" as both a pod *and* a systemd unit.  For example, we could represent the "applying updates" phase via e.g. `systemctl start machine-config-operator-update`.

This would also more naturally handle cases where e.g. we're updating a node and while that happens a cluster upgrade is happening an a new daemonset is rolling out.  Now with correct `SIGTERM` handling the old pod won't die until done, but you could imagine that instead e.g. we allow the pod to be killed, but what's actually doing the updates is the systemd unit, and the new pod discovers that state instead.  IOW think of the pod as like a "control plane proxy".  A systemd unit would more obviously natively interact with native systemd features (including e.g. things like ordering `Before/After=` and possibly using the systemd inhibit functionality, etc.)

darkmuggle:
> Tangentially related to this...one thing that could help in the MCO design is if we had a "dual nature" as both a pod _and_ a systemd unit. For example, we could represent the "applying updates" phase via e.g. `systemctl start machine-config-operator-update`.
> 

I had a similar thought, although your idea is a bit more refined. My idea was to have something that would ensure that if a user did a shutdown/reboot in the middle that RHCOS would wait until the update was done. However, the idea was NAK'd during our recent team discussion. Regardless, I do think that RHCOS and MCO could work a little better on coordinating when an update is safe to apply or block an update from starting if the machine is shutting down (such as in the case when a user has SSH'd).

pkg/daemon/daemon.go
darkmuggle: Fixed up, thanks!

pkg/daemon/daemon.go
darkmuggle: I did go back on and forth on whether or to report `SIGTERM` as an error or not for the reasons your blog highlights and ultimately went with returning the error to signal the reason for the death of the process (and thus make it obvious in the logs). In retrospect, I think just logging and moving on is the better path. 

FWIW I was arguing for making the MCD block reboots, however, @crawford argued that fully transactional updates would negate the need and is the better decision. My view is that until we get the transactional update mechanism, a short-lived inhibitor is _better_ than nothing and could prevent some support cases and bug reports. 

> And related to all this, ideally of course we do #1190 - then we shouldn't need to handle SIGTERM at all in the MCO. 

I would disagree -- we should at, the very least, through away the pending transaction (or have `rpm-ostree` do it).

darkmuggle:
> Would you say the core thing this is fixing here is that currently we were ignoring SIGTERM if caught during an update, but we didn't then exit after the update had finished? And that was causing systemd to time out and go on the SIGKILL spree?

@cgwalters this was fix four issues that were observed were:
- when https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L1965 was called, it would in fact _wait_ for 24 hours to be killed.
- This is because all `sigterms` were simply swallowed. https://github.com/openshift/machine-config-operator/pull/2395/files#diff-349c0748c3f52201852d3027c29daf618b45300b949482728df6666a3f9ba245L1917-L1933 shows that the prior logic literally did nothing. The `updateActive` bool was used to log that an update was active but didn't change the daemon's execution. 
- This fixes a race condition between the MCD receiving a `sigterm` and the MCD starting work. We have a report of the NTO applying an MCC while the SRIOV Operator triggers a reboot. Effectively since the `sigterm` was being trapped, the daemon might log if an update is active, but otherwise would continue normal operations until its work queue was closed _or_ it was `sigkill`'d.
- The `stopCh` channel did not stop work. https://github.com/openshift/machine-config-operator/blob/61a13776c9db16420380aae2da50a806f26965bb/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go#L140 is the code that runs the worker func. The interesting problem is that `wait.Until` will run the function until stop is signaled, but it won't stop the func. Since https://github.com/openshift/machine-config-operator/blob/61a13776c9db16420380aae2da50a806f26965bb/pkg/daemon/daemon.go#L348-L351 runs as an infinite loop work can start after its supposed to stop. 

So the core problem this seeks to solve is:
- wait to die until work finishes
- don't start work after sigterm or signaled to stop
- respond as soon as possible to either `stopCh` or `sigterm`

sinnykumari:
Thanks Colin for reviewing.
There are some good discussion points in the PR which will be be good to revisit and make update and reboot better.

@Ben Latest push broke some tests which needs to get fixed first.

darkmuggle:
Arg, I forgot to do the `make verify; make test-unit` dance after PR review update. That's been fixed. @ksinny should be good to go now.

PR #2398, opened 2021-02-09
Add kubelet CA to no-reboot action list/Do not drain for non-reboot actions

yuqi-zhang:
Incorporates https://github.com/openshift/machine-config-operator/pull/2294 and https://github.com/openshift/machine-config-operator/pull/2290, such that a kubelet cert signer rotation causes a "none" action.

In 4.7 we introduced non-reboot updates for select cases, but still drained nodes for safety. This is still a major workload disruption, and non-reboot updates do not require draining for correctness. This PR will also introduce the ability to not drain at all for non-reboot updates, and in case of error, return the error directly instead of rebooting in an attempt to fix it.

This will cause updates to effectively to not cordon nodes at all, and on AWS a change to e.g. kubelet ca would talk ~20 seconds to apply to a 3-node pool.

One last thing to consider is what behaviour we want for changes to `/etc/containers/registries.conf`, since not draining may cause a mismatch of image reference between existing and new containers.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -132,8 +132,7 @@ func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string
 			if dn.recorder != nil {
 				dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedServiceReload", fmt.Sprintf("Reloading %s service failed. Error: %v", serviceName, err))
 			}
-			dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)
-			dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+			return fmt.Errorf("Could not apply update: reloading %s configuration failed. Error: %v", serviceName, err)
 		}
 
 		if dn.recorder != nil {
@@ -147,14 +146,12 @@ func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string
 	// Get current state of node, in case of an error reboot
 	state, err := dn.getStateAndConfigs(configName)
 	if err != nil {
-		glog.Errorf("Error processing state and configs, node will reboot instead. Error: %v", err)
-		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+		return fmt.Errorf("Could not apply update: error processing state and configs. Error: %v", err)
 	}
 
 	var inDesiredConfig bool
 	if inDesiredConfig, err = dn.updateConfigAndState(state); err != nil {
-		glog.Errorf("Setting node's state to Done failed, node will reboot instead. Error: %v", err)
-		return dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+		return fmt.Errorf("Could not apply update: setting node's state to Done failed. Error: %v", err)
 	}
 	if inDesiredConfig {
 		return nil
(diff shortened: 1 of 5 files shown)


yuqi-zhang:
https://github.com/openshift/machine-config-operator/pull/2294 and https://github.com/openshift/machine-config-operator/pull/2290 can still be merged separately, this exists as a consolidated PR if we want to merge altogether. The commits are still separate for cherry-picking, if needed.

/hold pending investigation of the last point above

yuqi-zhang:
After some discussion and consideration, it is likely better for us to still drain on registries.conf changes (crio reload), to surface errors quicker if the file is bad. See commit message for details.

/hold cancel

pkg/daemon/update.go
```diff
@@ -132,8 +132,7 @@ func (dn *Daemon) performPostConfigChangeAction(postConfigChangeActions []string
 			if dn.recorder != nil {
 				dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedServiceReload", fmt.Sprintf("Reloading %s service failed. Error: %v", serviceName, err))
 			}
-			dn.logSystem("Reloading %s configuration failed, node will reboot instead. Error: %v", serviceName, err)
-			dn.reboot(fmt.Sprintf("Node will reboot into config %s", configName))
+			return fmt.Errorf("Could not apply update: reloading %s configuration failed. Error: %v", serviceName, err)
```
sinnykumari, line 135: I am not convinced why we shouldn't reboot to try to auto fix the problem. These errors may not be user initiated, user will be anyway clueless and will end up creating bug report or rebooting node explicitly to fix. Since MCO has already drained the node, attempting to auto-recover by reboot makes sense and if problem still remains we will at least have logs that rebooting doesn't help and we can investigate further.

yuqi-zhang: So I have two considerations here although I do not feel super strongly either way:
1. In terms of failure conditions of the reload, it could be due to
 1a) a bad config provided by the user causing the reload to fail, in which case rebooting does not help anyways
 1b) a transient issue with e.g. systemd, but since our logic keeps on retrying through a new sync, it will resolve itself eventually if its just a transient issue
 1c) an issue that isn't transient but resolvable by a reboot. This would be covered by a reboot but I'm not sure what an example would be
2. Since we are saying this change is non-reboot, a customer may be confused if it rebooted (due to an error) and not necessarily have insight to what the error was. Since we make calls to systemctl elsewhere in the MCD anyways, I thought it would be consistent to display the same failure mode (returns error, logs it, but will retry upon a later sync)

sinnykumari: thanks Jerry for providing the reasoning, I am convinced now :slightly_smiling_face:

sinnykumari (commented):
This looks good to me, will let others to review as well.

Since https://github.com/openshift/machine-config-operator/pull/2290 and https://github.com/openshift/machine-config-operator/pull/2294 are open, we may want to bring everyone on this PR or get those PRs merged first

yuqi-zhang:
I've updated the commit messages (keeping original authors) so the two commits are easier understood.

https://github.com/openshift/machine-config-operator/pull/2398/commits/05614046601633fd677e1799c7d4a35bd8431d32
https://github.com/openshift/machine-config-operator/pull/2398/commits/3aabfec2ece0d7bd3487399e615e27b8efbfe808

If the original authors are ok, we can go ahead with this as the main PR. Will ask on the separate PRs.

rphillips:
/approve
Thanks!

yuqi-zhang:
/retest

did some manual testing, the behaviour is as expected

darkmuggle (approved):

PR #2400, opened 2021-02-09
Dockerfile: update reference to builder to new CI location

yuqi-zhang:
Update reference, so that doing a `make image` pulls from the updated location to match CI.

Interestingly `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8` (the one referenced in Dockerfile.rhel7) doesn't seem to work, so using 4.7 one for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
 WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
 # FIXME once we can depend on a new enough host that supports globs for COPY,
 # just use that.  For now we work around this by copying a tarball.
 RUN make install DESTDIR=./instroot && tar -C instroot -cf instroot.tar .
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.ci.openshift.org/ocp/4.8:base
 COPY --from=builder /go/src/github.com/openshift/machine-config-operator/instroot.tar /tmp/instroot.tar
 RUN cd / && tar xf /tmp/instroot.tar && rm -f /tmp/instroot.tar
 COPY install /manifests


kikisdeliveryservice:
> Update reference, so that doing a `make image` pulls from the updated location to match CI.
> 
> Interestingly `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8` (the one referenced in Dockerfile.rhel7) doesn't seem to work, so using 4.7 one for now.

do we need to wait for that to be created or something?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
```
kikisdeliveryservice, line 1: i thought this was just removing the `svc` ??

kikisdeliveryservice: like: `registry.ci.openshift.org/openshift/release:golang-1.15`

yuqi-zhang: This just matches https://github.com/openshift/machine-config-operator/blob/master/Dockerfile.rhel7#L2

The previous edition of it rather: https://github.com/openshift/machine-config-operator/pull/2337/files

Maybe we shouldn't reference builder? I'm not sure. I just know the updated location does work. `registry.ci.openshift.org/openshift/release:golang-1.15` does not seem to exist.

kikisdeliveryservice:
Wait.. is it not avail bc we're waiting for RHCOS?..

I'm gonna go ask around

/hold

sinnykumari:
This is confusing, how did `images` ci test passed then on https://github.com/openshift/machine-config-operator/pull/2396? Does that mean ART and ci uses
Dockerfile.rhel7 file and local dev build (make image) uses Dockerfile ?

yuqi-zhang:
Updated to 4.8 location, I had an expired secret :facepalm: 

> Does that mean ART and ci uses Dockerfile.rhel7 file and local dev build (make image) uses Dockerfile ?

I believe so. Dockerfile.rhel7 starting of https://github.com/openshift/machine-config-operator/commit/1cd3056d6c0d9229b6f71230f606777a1ee1a66b was supposed to be a copy of Dockerfile (why have both then...?) but the ART process seem to only update the other one, making that process kinda pointless.

Maybe we should just drop Dockerfile and point the Makefile to Dockerfile.rhel7?

sinnykumari:
> Maybe we should just drop Dockerfile and point the Makefile to Dockerfile.rhel7?

yeah, or better way would be talk to ART team and see if they can use and sync Dockerfile directlyand we get away with Dockerfile.rhel7

Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
 WORKDIR /go/src/github.com/openshift/machine-config-operator
 COPY . .
 # FIXME once we can depend on a new enough host that supports globs for COPY,
 # just use that.  For now we work around this by copying a tarball.
 RUN make install DESTDIR=./instroot && tar -C instroot -cf instroot.tar .
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.ci.openshift.org/ocp/4.8:base
```
yuqi-zhang, line 8: Sorry, forgot to update this line, we should remove all references to `svc`. I tested this also works with correct creds to `registry.ci`

kikisdeliveryservice: good catch!

PR #2401, opened 2021-02-10
[OCPCLOUD-1106] Add afterburn task to update AWS hostname to match instance metadata

Danil-Grigorev:
**- What I did**

During kubelet migration from `--cloud-provider=aws` to `--cloud-provider=external` kubelet determines `Node` name by the hostname of the machine where it is running. AWS hostname is mismatched, usually it is a reversed DNS mapping of machine ip, like  `ip-10-0-195-176`. This results in node name mismatch in kubelet, and API server admission plugin rejects changes for the object.

This fixes the issue.

**- How to verify it**

- Ensure that new nodes provisioned by kubelet will report hostname with aws prefix, typically reversed IP mapping + `<region>.compute.internal` even when the cloud-provider is set to `external`.

### DIFF
--- a/templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml
+++ b/templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml
@@ -1,0 +1,10 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # The following configuration allows 90-long-hostname.sh
+    # to manage setting transient hostname instead of NetworkManager itself.
+    # See: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+    #      https://bugzilla.redhat.com/show_bug.cgi?id=1872885
+    [main]
+    hostname-mode=none
--- a/templates/common/aws/units/aws-hostname.service.yaml
+++ b/templates/common/aws/units/aws-hostname.service.yaml
@@ -1,0 +1,20 @@
+name: aws-hostname.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Set AWS Transient Hostname
+  # Block services relying on networking being up.
+  Before=network-online.target
+  # Wait for NetworkManager to report it's online
+  After=NetworkManager-wait-online.service
+  # Run before hostname checks
+  Before=node-valid-hostname.service
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStart=/usr/local/bin/mco-hostname --aws
+
+  [Install]
+  WantedBy=multi-user.target
+  WantedBy=network-online.target
--- a/templates/common/_base/files/usr-local-bin-mco-hostname.yaml
+++ b/templates/common/_base/files/usr-local-bin-mco-hostname.yaml
@@ -22,6 +22,16 @@ contents:
         exit 0
     }
 
+    set_aws_hostname() {
+        /usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname
+
+        local host_name=$(cat /run/afterburn.hostname)
+
+        echo "setting transient hostname to ${host_name}"
+        /bin/hostnamectl --transient set-hostname "${host_name}"
+        exit 0
+    }
+
     set_gcp_hostname() {
         /usr/bin/afterburn --provider gcp --hostname=/run/afterburn.hostname
 
@@ -59,6 +69,7 @@ contents:
     case "${arg}" in
         --wait) wait_localhost;;
         --gcp) set_gcp_hostname;;
+        --aws) set_aws_hostname;;
         *) echo "Unhandled arg $arg"; exit 1
     esac
 


yuqi-zhang:
Hi Danil,

What is the context of this PR? I don't believe any of the cloud platforms today use afterburn directly to acquire their hostname (other than GCP to work around GCP hostname truncating issues). Is there a bug you are trying to resolve?

darkmuggle:
Is this related to the cloud provider plugin requiring that the hostname and the node name match for admission to the cluster? I know this is a problem for VMware and if we have to do this for AWS....I fear we'll be doing it for the other clouds too.

Danil-Grigorev:
What happened is that I tried an automated reconfiguration for kubelet with `--cloud-provider=external` flag for AWS and got this issue with admission lock on the `Node` resource in api server. The hostname was different, so the node was not applied by kubelet, which resulted in the thing never running. I'm still trying to verify if this will fix it for AWS, but yeah. That's what @darkmuggle mentioned is closely related. Same attempt for GCP was working, and the `external` configuration provisioned Node without issues.

templates/common/aws/units/afterburn-hostname.service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  Description=Afterburn Hostname for AWS
+  After=NetworkManager-wait-online.service
+  Before=node-valid-hostname.service
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStartPre=/usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname
+  # Manually set current hostname at /proc/sys/kernel/hostname to correct value for kubelet.service
+  ExecStart=/bin/bash -c "cp /run/afterburn.hostname /proc/sys/kernel/hostname; cp /run/afterburn.hostname /etc/hostname"
```
Danil-Grigorev, line 14: `/proc/sys/kernel/hostname` is the default and only place where golang `os.Hostname` used in kubelet will gain the value. `node-valid-hostname.service` `hostnamectl` call fails to update current hostname in running machine, so there is a manual copy.

```
Could not set property: Connection timed out
```

Danil-Grigorev:
@darkmuggle @yuqi-zhang PTAL.

rphillips:
We need to fix this in the external cloud providers. There are too many platforms this can effect.

Danil-Grigorev:
@rphillips I tested it in GCP, Azure, migration works fine on those platforms. vSphere has a fix, same for OpenStack, so it is possible this is the only place left to be changed. We can't really fix it for `external` cloud providers - it is common for all platform-less kubernetes configurations. The issue happens to be here: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/node/node.go#L56

There is no way for standard `os` go library to default to `FQDN` hostname, and I'm not sure changing this core functionality will be easily accepted.

Danil-Grigorev:
> Is this related to the cloud provider plugin requiring that the hostname and the node name match for admission to the cluster? I know this is a problem for VMware and if we have to do this for AWS....I fear we'll be doing it for the other clouds too.

@darkmuggle I tested it in other clouds, does not seem to be the case. Only AWS is affected. So there is a fix.

rphillips:
Upstream issue: https://github.com/kubernetes/kubernetes/issues/70897

rphillips:
We should use the following script/function to set the hostname, which uses hostnamectl:

https://github.com/openshift/machine-config-operator/blob/522f0fa36cc7b952c6e98e120c58c66e6d795544/templates/common/_base/files/usr-local-sbin-set-valid-hostname.yaml#L38

rphillips:
The script gets installed at `/usr/local/sbin/set-valid-hostname.sh`

templates/common/aws/units/afterburn-hostname.service.yaml
Danil-Grigorev: @rphillips Here is an issue with hostnamectl I observed, re: https://github.com/openshift/machine-config-operator/pull/2401#issuecomment-789739611

Danil-Grigorev:
@rphillips Your feedback should be addressed now, could you take a look? The script is now using `set-valid-hostname.sh`
/retest

Danil-Grigorev:
@darkmuggle @yuqi-zhang PTAL. AWS provider is the only one affected and as @rphillips pointed out it is coreOS issue- https://github.com/kubernetes/kubernetes/issues/70897. They propose to use afterburner for managing such tasks, which is this PR exactly doing.

yuqi-zhang:
Hi, sorry for the late comment, I'm not the most comfortable with hostnames so I defer to @darkmuggle if he has the time to take a look. Just a quick question, was comparing this to: https://github.com/openshift/machine-config-operator/blob/bb2630cee2ee42a39638e968da9601c726467494/templates/common/gcp/units/gcp-hostname.service.yaml#L7

That service appears to not run on firstboot. Should this do the same?

darkmuggle:
The change looks fine, it's consistent with other platforms. 
/approve
/lgtm
(other than the failing CI)

sinnykumari:
@rphillips Since you have been also reviewing this PR, can you please take a look at the final PR. thanks!

sinnykumari (approved):
After looking closely, I see that all the concerns has been addressed.

sinnykumari:
As per conversation on slack with Danil, https://github.com/openshift/machine-config-operator/pull/2401#discussion_r574448305 is still unaddressed.
Waiting on test result that hostnamectl is working as expected during boot on aws

/approve cancel

Danil-Grigorev:
@sinnykumari This is not an issue of this PR. It is an issue of `hostnamectl` on AWS coreos side. There is a BZ somewhere in rhel/coreos space describing the exact same issue. This PR has all comments adressed, so I don't think removing approval is good path to take here. We can put hold on it and untold when we are confident.

sinnykumari:
Any MCO team can re-approve the PR once it has been tested. We are regularly monitoring the open PRs and in case we miss it, please ping us again, we will be happy to get it merged slack :slightly_smiling_face:

elmiko:
i have been testing out this pr today and i think it is working as intended. here's what i tried:

1. build an mco with this pr + #2386 
2. created a custom release payload based on `registry.ci.openshift.org/ocp/release:4.9.0-0.ci-2021-06-19-105001`
3. added the external feature gate manifest to my install, and ran `create cluster`

what i observed is that i see this entry in the machine objects:
```yaml
status:
  addresses:
  - address: 10.0.152.55
    type: InternalIP
  - address: ip-10-0-152-55.us-east-2.compute.internal
    type: InternalDNS
  - address: ip-10-0-152-55.us-east-2.compute.internal
    type: Hostname
```
if i understand correctly, this hostname is being gathered from the host o/s which should be demonstrating this pr.

elmiko:
i ssh'd into the machine in my previous message to confirm the findings, and i'm slightly confused.

i see this on the shell:
```bash
[core@ip-10-0-152-55 ~]$ hostname
ip-10-0-152-55
```

and it looks like perhaps the `afterburn-hostname.service` did not run to completion?
```
[core@ip-10-0-152-55 ~]$ systemctl status afterburn-hostname
● afterburn-hostname.service - Afterburn Hostname for AWS
   Loaded: loaded (/etc/systemd/system/afterburn-hostname.service; enabled; vendor preset: enabled)
   Active: active (exited) since Mon 2021-06-21 18:46:31 UTC; 2h 17min ago
  Process: 1197 ExecStart=/bin/bash -c source /usr/local/sbin/set-valid-hostname.sh; set_valid_hostname `cat /run/afterburn.hostname` (code=exited, status=>
  Process: 1170 ExecStartPre=/usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname (code=exited, status=0/SUCCESS)
 Main PID: 1197 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 49006)
   Memory: 0B
      CPU: 0
   CGroup: /system.slice/afterburn-hostname.service

Jun 21 18:46:06 ip-10-0-152-55 systemd[1]: Starting Afterburn Hostname for AWS...
Jun 21 18:46:06 ip-10-0-152-55 afterburn[1170]: Jun 21 18:46:06.256 INFO Putting http://169.254.169.254/latest/api/token: Attempt #1
Jun 21 18:46:06 ip-10-0-152-55 afterburn[1170]: Jun 21 18:46:06.258 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/hostname: Attempt #1
Jun 21 18:46:06 ip-10-0-152-55 afterburn[1170]: Jun 21 18:46:06.261 INFO Fetch successful
Jun 21 18:46:31 ip-10-0-152-55 bash[1197]: Could not set property: Connection timed out
Jun 21 18:46:31 ip-10-0-152-55 systemd[1]: Started Afterburn Hostname for AWS.
```

elmiko:
i was able to manually run the commands from the afterburn-hostname, and those did work:

```bash
[root@ip-10-0-152-55 sbin]# source /usr/local/sbin/set-valid-hostname.sh
[root@ip-10-0-152-55 sbin]# set_valid_hostname `cat /run/afterburn.hostname`
exit
[core@ip-10-0-152-55 sbin]$ echo $?
0
[core@ip-10-0-152-55 sbin]$ hostname
ip-10-0-152-55.us-east-2.compute.internal
```

not sure why this is failing during boot though. perhaps there is another dependency we need to wait on, or add more retries?

templates/common/aws/units/afterburn-hostname.service.yaml
```diff
@@ -0,0 +1,18 @@
...
+  After=NetworkManager-wait-online.service
+  Before=node-valid-hostname.service kubelet.service
+
+  [Service]
+  Restart=on-failure
+  RestartSec=15
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStartPre=/usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname
+  ExecStart=/bin/bash -c "source /usr/local/sbin/set-valid-hostname.sh; set_valid_hostname `cat /run/afterburn.hostname`"
```
cgwalters, line 15: This will need reworking to be based on https://github.com/openshift/machine-config-operator/pull/2618

cgwalters (changes requested):
Also, this needs to do the same thing as we're doing on GCP here: https://github.com/openshift/machine-config-operator/blob/master/templates/common/gcp/files/etc-networkmanager-conf.d-hostname.yaml

I guess just copy that file  into `templates/common/aws`, longer term it'd be good to dedup them.

Otherwise the hostname change may get reverted when NM renews the DHCP lease.

templates/common/aws/units/afterburn-hostname.service.yaml
```diff
@@ -0,0 +1,18 @@
+name: afterburn-hostname.service
```
cgwalters, line 1: Also please name this `aws-hostname.service` or so similarly to `gcp-hostname.service`.

cgwalters:
https://github.com/coreos/afterburn/issues/509#issuecomment-866295230

elmiko:
i created a small pr based off of @cgwalters suggestions and it seems to be working well for me.

```
diff --git a/templates/common/_base/files/usr-local-bin-mco-hostname.yaml b/templates/common/_base/files/usr-local-bin-mco-hostname.yaml
index 0ad56d8c..d3a398bf 100644
--- a/templates/common/_base/files/usr-local-bin-mco-hostname.yaml
+++ b/templates/common/_base/files/usr-local-bin-mco-hostname.yaml
@@ -22,6 +22,16 @@ contents:
         exit 0
     }
 
+    set_aws_hostname() {
+        /usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname
+
+        local host_name=$(cat /run/afterburn.hostname)
+
+        echo "setting transient hostname to ${host_name}"
+        /bin/hostnamectl --transient set-hostname "${host_name}"
+        exit 0
+    }
+
     set_gcp_hostname() {
         /usr/bin/afterburn --provider gcp --hostname=/run/afterburn.hostname
 
@@ -58,6 +68,7 @@ contents:
     arg=${1}; shift;
     case "${arg}" in
         --wait) wait_localhost;;
+        --aws) set_aws_hostname;;
         --gcp) set_gcp_hostname;;
         *) echo "Unhandled arg $arg"; exit 1
     esac
diff --git a/templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml b/templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml
new file mode 100644
index 00000000..de974ae2
--- /dev/null
+++ b/templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml
@@ -0,0 +1,10 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # The following configuration allows 90-long-hostname.sh
+    # to manage setting transient hostname instead of NetworkManager itself.
+    # See: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+    #      https://bugzilla.redhat.com/show_bug.cgi?id=1872885
+    [main]
+    hostname-mode=none
diff --git a/templates/common/aws/units/afterburn-hostname.service.yaml b/templates/common/aws/units/afterburn-hostname.service.yaml
deleted file mode 100644
index 6126d872..00000000
--- a/templates/common/aws/units/afterburn-hostname.service.yaml
+++ /dev/null
@@ -1,18 +0,0 @@
-name: afterburn-hostname.service
-enabled: true
-contents: |
-  [Unit]
-  Description=Afterburn Hostname for AWS
-  After=NetworkManager-wait-online.service
-  Before=node-valid-hostname.service kubelet.service
-
-  [Service]
-  Restart=on-failure
-  RestartSec=15
-  Type=oneshot
-  RemainAfterExit=yes
-  ExecStartPre=/usr/bin/afterburn --provider aws --hostname=/run/afterburn.hostname
-  ExecStart=/bin/bash -c "source /usr/local/sbin/set-valid-hostname.sh; set_valid_hostname `cat /run/afterburn.hostname`"
-
-  [Install]
-  WantedBy=network-online.target
diff --git a/templates/common/aws/units/aws-hostname.service.yaml b/templates/common/aws/units/aws-hostname.service.yaml
new file mode 100644
index 00000000..00b80017
--- /dev/null
+++ b/templates/common/aws/units/aws-hostname.service.yaml
@@ -0,0 +1,20 @@
+name: aws-hostname.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Set AWS Transient Hostname
+  # Block services relying on networking being up.
+  Before=network-online.target
+  # Wait for NetworkManager to report it's online
+  After=NetworkManager-wait-online.service
+  # Run before hostname checks
+  Before=node-valid-hostname.service
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStart=/usr/local/bin/mco-hostname --aws
+
+  [Install]
+  WantedBy=multi-user.target
```

when i install a cluster with this patch in place, i see machines joining and nodes created with the proper names. here is an example of the CSRs after bootstrap:

```
NAME                                       AGE   SIGNERNAME                                    REQUESTOR                                                                         CONDITION
csr-5zbw8                                  50m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-196-71.us-east-2.compute.internal                             Approved,Issued
csr-fgt7s                                  50m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
csr-gddbc                                  57m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
csr-hb24h                                  57m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-171-34.us-east-2.compute.internal                             Approved,Issued
csr-hhnh8                                  57m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-213-252.us-east-2.compute.internal                            Approved,Issued
csr-jz5g7                                  51m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-161-224.us-east-2.compute.internal                            Approved,Issued
csr-kzd2k                                  51m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
csr-nzc9m                                  51m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-156-210.us-east-2.compute.internal                            Approved,Issued
csr-t6wf4                                  57m   kubernetes.io/kubelet-serving                 system:node:ip-10-0-155-68.us-east-2.compute.internal                             Approved,Issued
csr-ttm2t                                  57m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
csr-wtvsn                                  57m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
csr-xvs8h                                  51m   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper         Approved,Issued
system:openshift:openshift-authenticator   56m   kubernetes.io/kube-apiserver-client           system:serviceaccount:openshift-authentication-operator:authentication-operator   Approved,Issued
```

templates/common/aws/files/etc-networkmanager-conf.d-hostname.yaml
```diff
@@ -0,0 +1,12 @@
+mode: 0644
+path: "/etc/NetworkManager/conf.d/hostname.conf"
+contents:
+  inline: |
+    # On GCP the hostname is likely to be longer than 63 chars.
+    # /etc/networkmanager/dispatcher.d/90-long-hostname.sh will set
+    # the hostname. The following configuration allows 90-long-hostname.sh
+    # to manage setting transient hostname instead of NetworkManager itself.
+    # See: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+    #      https://bugzilla.redhat.com/show_bug.cgi?id=1872885
```
elmiko, line 10: i don't think we need the gcp specific text
```suggestion
    # The following configuration allows 90-long-hostname.sh
    # to manage setting transient hostname instead of NetworkManager itself.
    # See: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
    #      https://bugzilla.redhat.com/show_bug.cgi?id=1872885
```

elmiko (commented):
just a minor nit, but otherwise this looks good and i know it works ;)

cgwalters (approved):
LGTM.

(At some point I am curious around the larger backstory on why in AWS DHCP doesn't give us the hostname we need)

templates/common/aws/units/aws-hostname.service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  Before=node-valid-hostname.service
+
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStart=/usr/local/bin/mco-hostname --aws
+
+  [Install]
+  WantedBy=multi-user.target
+  WantedBy=network-online.target
```
darkmuggle, line 20: Missing a new line. 

But otherwise, LGTM

darkmuggle (approved):

kikisdeliveryservice (approved):
and adding lgtm for Ben & Colin's reviews

/lgtm

PR #2402, opened 2021-02-10
Bug 1927366: Fixes ovs-configuration with cloned connections

trozet:
When using nm clone for a connection where we cannot find the current
keyfile, the file is assumed to be in the regular NM keyfile path.
However on platforms using the "merged" keyfile path, this fails,
because the file only exists in the merged directory.

Additionally, after we copy the cloned file to edit for static IP
addressing, NM will fail to load it because of selinux. Therefore we
need to restorecon after we copy it.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -43,8 +43,11 @@ contents:
           ifconfig "$intf" allmulti
         fi
       }
-
-      NM_CONN_PATH="/etc/NetworkManager/system-connections"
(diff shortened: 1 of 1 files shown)


trozet:
@bcrochet @celebdor

osherdp:
@trozet hey man
Any idea when it will be merged?

romfreiman:
@yuqi-zhang @sinnykumari @cgwalters PTAL? It blocks us for 4.8

sinnykumari (approved):
LGTM

PR #2403, opened 2021-02-10
Bug 1921937: kubelet.service: start after hostnamed

haircommander:
There currently exists a race between the Kubelet and hostnamed.
If Kubelet starts a container that bind mounts /etc/hostname before hostnamed finishes creating /etc/hostname,
CRI-O needs to create the not yet created /etc/hostname as a directory (conforming with age-old docker behavior).

This causes hostnamed to perpetually fail, which causes all sorts of trouble

Instead of allowing this race, we can start the kubelet after systemd-hostnamed is finished, so containers
always mount the newly created /etc/hostname

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1921937

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -4,7 +4,7 @@ contents: |
   [Unit]
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
-  After=network-online.target crio.service
+  After=network-online.target crio.service systemd-hostnamed.service
 
   [Service]
   Type=notify
--- a/templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml
@@ -4,7 +4,7 @@ contents: |
   [Unit]
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
-  After=network-online.target crio.service
+  After=network-online.target crio.service systemd-hostnamed.service
 
   [Service]
   Type=notify


haircommander:
@rphillips @mrunalp @umohnani8 PTAL

haircommander:
Note: we *could* have CRI-O specify "After..." instead of Kubelet, but CRI-O's service file comes from the CRI-O rpm, not MCO, so we cannot test it in the same way we can the kubelet.service.

It ultimately doesn't matter. CRI-O won't create the directory until Kubelet tells it to create a container that requests the directory. If kubelet starts after hostnamed.service, we should prevent this race (and test in CI if it likes this change at the same time!)

templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
```diff
@@ -4,7 +4,7 @@ contents: |
   [Unit]
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
-  After=network-online.target crio.service
+  After=network-online.target crio.service systemd-hostnamed.service
```
darkmuggle, line 7: Does this fix the bug? Unless the hostname is statically set, then you won't see an `/etc/hostname`. A lot of the platforms use a transient hostname. 

AFAIK, mounting in `/proc/kernel/hostname` would be better since it would have the current hostname.

haircommander: hmmm I hoped that systemd-hostnamed finishing would mean *something* would be in /etc/hostname.

That's a good idea! I will look into making that switch

haircommander: I cannot find what container actually attempts to mount /etc/hostname. The only thing cri-o can really do is switch references to it to /proc/sys/kernel/hostname, which doesn't seem that safe...

PR #2407, opened 2021-02-12
WIP: Debugging MCD "Got SIGTERM, but actively updating" when it shouldn't be updating

wking:
Reading the code hasn't gotten me there, so here's some shot-in-the-dark debug logging for https://bugzilla.redhat.com/show_bug.cgi?id=1927041#c5

### DIFF
--- a/cmd/machine-config-daemon/start.go
+++ b/cmd/machine-config-daemon/start.go
@@ -4,10 +4,12 @@ import (
 	"bufio"
 	"flag"
 	"io"
+	"math/rand"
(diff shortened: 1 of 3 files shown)


darkmuggle:
I found that the daemon's sigterm handling was not working as expected. https://github.com/openshift/machine-config-operator/pull/2395 is the result that I _hope_ makes this particular issue go away.

wking:
Yeah, #2395 touched all the code I was concerned about.

/close

PR #2410, opened 2021-02-12
Remove platform-specific Corefile templates

cybertron:
As far as I can tell, these are all providing the exact same records,
just using variations on the specific plugin. For the sake of simplicity,
let's converge on one Corefile template.

A couple of notes:
1) I used the on-prem one because it handles ipv4 and ipv6 records
   in an RFC-compliant way, which none of the other plugins do.
2) I left the kubevirt db file. I'm not sure why it's there since
   there doesn't seem to be a corresponding Corefile for the platform,
   and since I don't understand it I think it's safest to leave it
   alone.

**- What I did**

**- How to verify it**


cybertron:
/retest

metal-ipi failed to deploy the masters before this ever would come into play (and shouldn't be affected by this change anyway), openstack and vsphere both appear to have successfully deployed the nodes and failed on other things, which means coredns should be working. Unfortunately my must-gather patch hasn't merged yet so we don't have logs from the infra containers there yet.

Gal-Zaidman:
@eslutsky can you please review this change ?

mandre (commented):
LGTM for OpenStack.

OpenStack was able to deploy the cluster which indicates DNS is working as expected. The job failed on an unrelated monitoring alert.

kikisdeliveryservice:
FYI: we now have owners files for template dirs so y'all will need to actually approve =)
see: https://github.com/openshift/machine-config-operator/pull/2410#issuecomment-778385934

mandre:
> FYI: we now have owners files for template dirs so y'all will need to actually approve =)
> see: [#2410 (comment)](https://github.com/openshift/machine-config-operator/pull/2410#issuecomment-778385934)

Great! in that case,

/lgtm
/approve

cybertron:
/retest

C'mon openstack, you can do it! :-)

mandre:
FYI I'm moving openstack job to a more stable platform with https://github.com/openshift/release/pull/16936

cybertron:
/retest

Looks like the openstack patch merged, so let's see if we have better luck now.

mandre:
> /retest
> 
> Looks like the openstack patch merged, so let's see if we have better luck now.

Sadly we experienced our first issue with Vexxhost right after we moved the openstack job there. It is fixed now.

/test e2e-openstack

cybertron:
/test e2e-openstack

Failed to bootstrap. Since bootstrapping is no longer dependent on DNS, I doubt it's related.

cybertron:
This went in as part of #2465 .

PR #2411, opened 2021-02-15
[OKD] Provide rendering support for OKD and SSH keys path

darkmuggle:
In order to make OKD specific changes safe, this change does two things:
- provides `daemon/okd.go` where specific run-time changes to the MCD can be done (such as setting paths)
- provides a template boolean function of `okd` to do _render_ tests
- does not presume the homedir of the `core` user
- adds unit tests

The goal is to provide a way for OKD to render templates that _will_ not affect OCP renderings. This means that instead of logic tests run on the node to determine the code path, the rendered configuration will only have a single path.


darkmuggle:
To clarify: this is a POC to demonstrate an idea. The MCO team will need to discuss if this path is better than maintaining an OKD branch.

LorbusChris (approved):
I am fine with this approach if the MCO team feels like it's beneficial and makes OKD-specific things easier to maintain on branches shared with OCP.

darkmuggle:
I'm withdrawing this PR for the time being. There's been little interest in this from either OKD or OCP and without further technical directions, this is PR will rot.

PR #2414, opened 2021-02-17
Bug 1902963: templates: add After=ostree-finalize-staged.service to kubelet.service

jlebon:
**- What I did**

We've seen occasional issues during upgrades caused by what seems to be
the kubelet still being active during OSTree finalization. If the
kubelet changes things in `/etc`, it'll confuse OSTree which is trying
to do the `/etc` merge.

We want to be sure that the kubelet is done modifying everything it
needs to in `/etc` and that it exited before we finalize the deployment.

Add a `After=ostree-finalize-staged.service` for this. The way this
works is that that service runs in its `ExecStop`, and shutdown ordering
is the reverse of startup. So this will cause the kubelet to exit before
`ostree-finalize-staged.service` is stopped.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1902963

**- How to verify it**

It's tricky to verify because it's a flake. So verification would be getting this in and not seeing that bug pop up anymore. (Could still at least verify though from the journal logs that the kubelet exited before `ostree-finalize-staged.service`.)

During upgrades, we now wait for the kubelet to exit before finalizing the staged deployment to ensure no conflicts occur.

### DIFF
--- a/templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml
@@ -5,6 +5,7 @@ contents: |
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
   After=network-online.target crio.service
+  After=ostree-finalize-staged.service
 
   [Service]
   Type=notify
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -5,6 +5,7 @@ contents: |
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
   After=network-online.target crio.service
+  After=ostree-finalize-staged.service
 
   [Service]
   Type=notify
--- a/templates/master/01-master-kubelet/on-prem/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/on-prem/units/kubelet.service.yaml
@@ -5,6 +5,7 @@ contents: |
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
   After=network-online.target crio.service
+  After=ostree-finalize-staged.service
 
   [Service]
   Type=notify
--- a/templates/worker/01-worker-kubelet/on-prem/units/kubelet.service.yaml
+++ b/templates/worker/01-worker-kubelet/on-prem/units/kubelet.service.yaml
@@ -5,6 +5,7 @@ contents: |
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
   After=network-online.target crio.service
+  After=ostree-finalize-staged.service
 
   [Service]
   Type=notify


jlebon:
(Not tested.)

cgwalters:
This is probably related to the "shutdown timeout" issue the kubelet folks are hitting sometimes where systemd times out waiting for containers and goes on a SIGTERM spree.  

Hmm you know, I bet we could move the `/etc` merge into the initramfs on next boot or so.  That would close off all races.

cgwalters:
Worth noting of course that https://github.com/openshift/machine-config-operator/issues/1190 would also fix this.

But the more I think about this the more this feels like a pure ostree-level bug - if systemd is in the "SIGTERM all the things" phase, we shouldn't try to do the ostree finalization at all.

darkmuggle:
I presume we'll want to backport this to 4.6?

jlebon:
> But the more I think about this the more this feels like a pure ostree-level bug - if systemd is in the "SIGTERM all the things" phase, we shouldn't try to do the ostree finalization at all.

Hmm, but even so the root issue is still that there's no ordering between those services, right? So even in the graceful case, we need some strong ordering.

> Hmm you know, I bet we could move the `/etc` merge into the initramfs on next boot or so. That would close off all races.

Doesn't that imply writing to the bootloader at shutdown still? Also not sure about adding more fallible code into the post-reboot path. It's nice that right now if it fails, we just go back into the same deployment.

Hmm, another easier approach is to have `ostree-finalize-staged.service` remount `/etc` read-only before starting the `/etc` merge. (That doesn't really fix any race, but at least it makes the operation more reliable.)

jlebon:
> I presume we'll want to backport this to 4.6?

So in one RHBZ, this was hit in 4.5 going to 4.6, in another one, it was hit going from 4.6 to 4.7. So ideally, we'd backport all the way to 4.5? (It doesn't seem like a super common issue, though at least it doesn't require new bootimages so!)

Mostly first looking to confirm this does work though. Will we be able to see the journal logs from the `e2e-agnostic-upgrade` test once it completes?

templates/worker/01-worker-kubelet/_base/units/kubelet.service.yaml
```diff
@@ -5,6 +5,7 @@ contents: |
   Description=Kubernetes Kubelet
   Wants=rpc-statd.service network-online.target crio.service
   After=network-online.target crio.service
+  After=ostree-finalize-staged.service
```
miabbott, line 8: Is it worth putting a condition in the template so that this is only used for RHCOS nodes?

sinnykumari: good point, this will work on OSTree based nodes but break RHEL nodes

jlebon: That's a good idea!

Though hmm... AFAICT there doesn't seem to be a variable already exposed for this. So I think this would require wiring it through the render code first. Meh... maybe not worth the trouble in that case? (It's totally harmless on non-RHCOS nodes.)

jlebon: > good point, this will work on OSTree based nodes but break RHEL nodes

systemd just ignores `Before=`/`After=` constraints if the targetunit is missing.

sinnykumari: cool, RHEL nodes are safe then :laughing:

jlebon:
OK cool, looks like this works! Looking at the journal logs from one of the nodes in ci/prow/e2e-agnostic-upgrade, I see:

```
   8916 Feb 22 16:25:27.917657 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 root[81740]: machine-config-daemon[70704]: Rebooting node
   8917 Feb 22 16:25:27.921197 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 root[81741]: machine-config-daemon[70704]: initiating reboot: Node will reboot into config rendered-worker-251b5c4768cfdff11d117502b97d1508
   8918 Feb 22 16:25:27.934014 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Started machine-config-daemon: Node will reboot into config rendered-worker-251b5c4768cfdff11d117502b97d1508.
   8919 Feb 22 16:25:27.948589 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopping Kubernetes Kubelet...
   8920 Feb 22 16:25:27.948842 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 hyperkube[1537]: I0222 16:25:27.948627    1537 dynamic_cafile_content.go:182] Shutting down client-ca-bundle::/etc/kubernetes/kubelet-ca.crt
   8921 Feb 22 16:25:27.970705 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: kubelet.service: Succeeded.
   8922 Feb 22 16:25:27.971695 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopped Kubernetes Kubelet.
   8923 Feb 22 16:25:27.971891 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: kubelet.service: Consumed 8min 7.090s CPU time
   8924 Feb 22 16:25:27.986455 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd-logind[1199]: System is rebooting.
   8925 Feb 22 16:25:27.994245 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: machine-config-daemon-reboot.service: Succeeded.
   8926 Feb 22 16:25:27.994948 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopped machine-config-daemon: Node will reboot into config rendered-worker-251b5c4768cfdff11d117502b97d1508.
   8927 Feb 22 16:25:27.995596 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: machine-config-daemon-reboot.service: Consumed 19ms CPU time
...
   9073 Feb 22 16:25:28.681767 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopping OSTree Finalize Staged Deployment...
   9074 Feb 22 16:25:28.688061 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopped target Paths.
   9075 Feb 22 16:25:28.694123 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: console-login-helper-messages-issuegen.path: Succeeded.
   9076 Feb 22 16:25:28.696019 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopped Monitor console-login-helper-messages runtime issue snippets directory for changes.
   9077 Feb 22 16:25:28.791278 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 ostree[81828]: Finalizing staged deployment
   9078 Feb 22 16:25:31.146325 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 ostree[81828]: Copying /etc changes: 14 modified, 0 removed, 168 added
   9079 Feb 22 16:25:31.146846 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 ostree[81828]: Copying /etc changes: 14 modified, 0 removed, 168 added
   9080 Feb 22 16:25:31.222965 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 kernel: EXT4-fs (sda3): re-mounted. Opts: (null)
   9081 Feb 22 16:25:32.033315 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 ostree[81828]: Bootloader updated; bootconfig swap: yes; deployment count change: 0
   9082 Feb 22 16:25:32.033832 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 ostree[81828]: Bootloader updated; bootconfig swap: yes; deployment count change: 0
   9083 Feb 22 16:25:33.187729 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: ostree-finalize-staged.service: Succeeded.
   9084 Feb 22 16:25:33.188550 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: Stopped OSTree Finalize Staged Deployment.
   9085 Feb 22 16:25:33.196203 ci-op-i1ivpc34-57c36-7nfmp-worker-b-k7qs2 systemd[1]: ostree-finalize-staged.service: Consumed 2.464s CPU time
```

kikisdeliveryservice:
@jlebon both of the remaining (but unrequired) tests are super flaky/red so I wouldn't' expect them to pass on any pr at this point ;)

/assign @sinnykumari @darkmuggle

jlebon:
Anyone want to `/lgtm` this?

kikisdeliveryservice (approved):
@sinnykumari PTAL

sinnykumari:
ah, missed to get to this PR earlier.

/lgtm

sinnykumari:
@jlebon shall we cherrypick this or we want to wait for a while to see how it goes in master?

jlebon:
> @jlebon shall we cherrypick this or we want to wait for a while to see how it goes in master?

I think this is fairly safe to backport, but maybe to start with let's just backport it to 4.7?

PR #2416, opened 2021-02-18
controller: remove etcd quorum-guard check on master pool

sinnykumari:
etcd-quorum-guard has been part of cluster-etcd-operator
for a while and it already guards against node drain to
prevent from quorum loss.

With having support for different types of cluster like
SingleReplica and HighlyAvailable, it allows MCO to not
think about what quorum guard behaviour should be
depending upon cluster type.

Also use GetScaledValueFromIntOrPercent() instead of
deprecated function GetValueFromIntOrPercent() and
update unit tests accordingly.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -944,21 +944,12 @@ func maxUnavailable(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) (int, 
 	if pool.Spec.MaxUnavailable != nil {
 		intOrPercent = *pool.Spec.MaxUnavailable
 	}
-	maxunavail, err := intstrutil.GetValueFromIntOrPercent(&intOrPercent, len(nodes), false)
+	maxunavail, err := intstrutil.GetScaledValueFromIntOrPercent(&intOrPercent, len(nodes), false)
 	if err != nil {
 		return 0, err
 	}
 	if maxunavail == 0 {
 		maxunavail = 1
-	}
-	if pool.Name == masterPoolName {
-		// calculate the fault tolerance dynamically for the master pool
-		// to avoid risking losing etcd quorum.
-		tolerance := len(nodes) - ((len(nodes) / 2) + 1)
-		if maxunavail > tolerance {
-			glog.Warningf("Refusing to honor master pool maxUnavailable %d to prevent losing etcd quorum, using %d instead", maxunavail, tolerance)
-			return tolerance, nil
-		}
 	}
 	return maxunavail, nil
 }
--- a/pkg/controller/node/node_controller_test.go
+++ b/pkg/controller/node/node_controller_test.go
@@ -454,30 +454,15 @@ func TestMaxUnavailable(t *testing.T) {
 			expected:   0,
 			err:        true,
 		}, {
-			poolName:   "master",
-			maxUnavail: intStrPtr(intstr.FromInt(1)),
-			nodes:      newNodeSet(5),
-			expected:   1,
-			err:        false,
+			maxUnavail: intStrPtr(intstr.FromString("60")),
+			nodes:      newNodeSet(4),
+			expected:   0,
+			err:        true,
 		}, {
-			poolName:   "master",
-			maxUnavail: intStrPtr(intstr.FromInt(2)),
-			nodes:      newNodeSet(3),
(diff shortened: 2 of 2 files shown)


hexfusion:
minor upgrade from 4.7

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1362402572626825216

yuqi-zhang (approved):
I think this PR is safe for non-HA. The only situation in a regular non-HA operation that we would hit this code would be if the user somehow changed maxUnavailable on a master pool, which generally shouldn't happen.

Although that said, there is a small difference in behaviour: with this change and a maxUnavailable of 2 in a 3-master cluster, I think two nodes can start updating simultaneously. One node will simply fail to drain and block until the other is finished, but you will see 2 masters listed as "Schedulingdisabled". This is probably still fine from etcd quorum perspective, but it might be worth testing out what would happen if such an update was triggered

sinnykumari:
> I think this PR is safe for non-HA. The only situation in a regular non-HA operation that we would hit this code would be if the user somehow changed maxUnavailable on a master pool, which generally shouldn't happen.
> 
> Although that said, there is a small difference in behaviour: with this change and a maxUnavailable of 2 in a 3-master cluster, I think two nodes can start updating simultaneously. One node will simply fail to drain and block until the other is finished, but you will see 2 masters listed as "Schedulingdisabled". This is probably still fine from etcd quorum perspective, but it might be worth testing out what would happen if such an update was triggered

This situation came to my mind but I think that is a user fault if they set maxUnavailable knowing that they have only 3 master nodes.

yuqi-zhang:
> This situation came to my mind but I think that is a user fault if they set maxUnavailable knowing that they have only 3 master nodes.

Agreed. We are removing a check that purely existed to prevent users from making that mistake, which makes me curious how the error would manifest if a user were to do it (if it would error at all).

hexfusion:
> minor upgrade from 4.7
> 
> https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1362402572626825216

upgrade passed

hexfusion:
> Agreed. We are removing a check that purely existed to prevent users from making that mistake, which makes me curious how the error would manifest if a user were to do it (if it would error at all).

Protecting user from mistake is a big reason to keep it. Silent upgrade errors and finger-pointing don't fair well at SEV1.

hexfusion:
In general, I feel quorum-guard should be the source of truth for "can we perform actions that may result in quorum loss". But it sounds like this also checks for user misconfiguration. If 2 nodes were Schedulingdisabled static pod operators could not roll revisions on the majority of nodes. I really don't want to do expose new failure cases.

sinnykumari:
yeah, it can cause problem in certain condition. One example that came to my mind: user cluster has 3 control plane nodes (say node1, node2, node3) and has set maxUnavailable to 2. Now, user wants to either upgrade cluster or apply a MachineConfig. With this change, MCO will initiate update on 2 nodes in parallel (say node1 and node2). MCO marks node1 and node2 as unscedulable. node1 starts drain which will start progressing as etcd-quorum-gurad is satisfied. node2 won't be able to get drained as it will be prevented by etcd-quorum-guard. Now, we have only node3 left where drained pod can be rescheduled. If node3 doesn't have enough resource to scheduled drained pods, drain from node1 may not finish and it can get stuck here.

If we think we should try to prevent such possibilities then we should keep this check in MCO for HA cluster.

yuqi-zhang:
So I did test this on a base 3-master setup on AWS, and it was able to complete successfully. The master nodes went unscheduleable two at a time, but the quorum guard prevent one drain from progression, so it was basically the same as upgrading 1 node at a time.

Here's a (maybe too hacky) alternative: set a min maxUnavailable=1. (Take the minimal of len(nodes) - ((len(nodes) / 2) + 1) or 1). This will guarentee that on master pools, the maxUnavailable is still calculated but cannot go below one, making it work both for 3-master or 1-master cases.

Funnily enough this does allow 2-master configs to upgrade (there shouldn't be any), but is a middle ground

hexfusion:
Based on additional research from ccx team on exposure to outlier failures outlined above we feel this change is safe to move forward.

/lgtm
/retest

PR #2420, opened 2021-02-18
Bug 1844990: server/api: Disable unsafe ciphers for CVE-2016-2183

darkmuggle:
With [1], SHA and 3DES Ciphers are considered unsafe. 

http/2 requires these inseucre Ciphers which means that fixing the CVE requires disableing http/2

[1] https://access.redhat.com/security/cve/cve-2016-2183

This change will not conflict with the TLS profile work coming later.

### DIFF
--- a/pkg/server/api.go
+++ b/pkg/server/api.go
@@ -60,14 +60,7 @@ func NewAPIServer(a *APIHandler, p int, is bool, c, k string) *APIServer {
 
 // Serve launches the API Server.
 func (a *APIServer) Serve() {
-	mcs := &http.Server{
-		Addr:    fmt.Sprintf(":%v", a.port),
-		Handler: a.handler,
-		// We don't want to allow 1.1 as that's old.  This was flagged in a security audit.
-		TLSConfig: &tls.Config{
-			MinVersion: tls.VersionTLS12,
-		},
-	}
+	mcs := getHTTPServerCfg(fmt.Sprintf(":%v", a.port), a.handler)
 
 	glog.Infof("Launching server on %s", mcs.Addr)
 	if a.insecure {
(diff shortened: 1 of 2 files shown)


yuqi-zhang (approved):
I tested this with the steps in the BZ and I see:
`SWEET32 (CVE-2016-2183, CVE-2016-6329)    not vulnerable (OK)`

so I think the functionality is good. Code also generally LGTM, but I am uncertain about some of the specifics. e.g. does disabling HTTP/2 affect anything?

darkmuggle:
> so I think the functionality is good. Code also generally LGTM, but I am uncertain about some of the specifics. e.g. does disabling HTTP/2 affect anything?

Disabling `http/2` is not really a big deal - the features like multiplexing, prioritization controls, compression, etc -- we're not using. More importantly, `http/2` by definition is insecure for the SWEET32 ciphers; there is simply no use `http/2` with TLS 1.2 (required with Azure). The only way to use `http/2` is to force TLS 1.3 and that can't happen. If the MCS was doing things like streaming media or trying to handle prefetching or HTTP pipelining, then it would be a different story.

pkg/server/api.go
```diff
@@ -340,3 +333,62 @@ func (h *defaultHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				first = append(first, c.ID)
+			}
+			if v == tls.VersionTLS12 && allowable(c) {
+				inFirst := false
+				for _, id := range first {
+					if c.ID == id {
+						break
+					}
+				}
+				if !inFirst {
```
sinnykumari, line 386: `inFirst` value will always be False as we are not setting it True when it was not found in `first`.

pkg/server/api.go
```diff
@@ -340,3 +333,62 @@ func (h *defaultHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// cipherOrder returns an ordered list of Ciphers that are considered secure
+// Deprecated ciphers are not returned.
+func cipherOrder() []uint16 {
+	var first []uint16
+	var second []uint16
+
+	allowable := func(c *tls.CipherSuite) bool {
+		// Disallow block ciphers using straight SHA1
+		// See: https://tools.ietf.org/html/rfc7540#appendix-A
+		if strings.HasSuffix(c.Name, "CBC_SHA") {
```
sinnykumari, line 364: Since we are matching with Suffix here, it won't match ciphers listed in blacklist that ends with `CBC_SHA256`, `CBC_SHA384` . Are these safe?

pkg/server/api.go
```diff
@@ -340,3 +333,62 @@ func (h *defaultHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	return
 }
+
+// getHTTPServerCfg returns the basic HTTP Server
+func getHTTPServerCfg(addr string, handler http.Handler) *http.Server {
+	return &http.Server{
+		Addr:    addr,
+		Handler: handler,
+		// CVE-2016-2183: disable http/2, which by definition, requires insecure ciphers
+		// Per https://golang.org/src/net/http/doc.go this is the runtime method of disabling HTTP/2
```
sinnykumari, line 343: Really like the useful comments you have added throughout the PR :heart:

pkg/server/api.go
darkmuggle: SHA256, 384, and 512 are considered secure....for now.

pkg/server/api.go
darkmuggle: Good catch -- fixed up.

pkg/server/api.go
```diff
@@ -340,3 +333,63 @@ func (h *defaultHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	for _, c := range tls.CipherSuites() {
+		for _, v := range c.SupportedVersions {
+			if v == tls.VersionTLS13 {
+				first = append(first, c.ID)
+			}
+			if v == tls.VersionTLS12 && allowable(c) {
+				inFirst := false
+				for _, id := range first {
+					if c.ID == id {
+						first = true
```
yuqi-zhang, line 383: I think you meant `inFirst` here

darkmuggle: My local toolchain must be really wonky -- `make verify` and `make test` -- all worked.

sinnykumari:
/lgtm

@darkmuggle As you have done good digging, can you create a followup doc PR into https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigServer.md with details and reason of MCO sticking to http and TLS 1.2 version

PR #2421, opened 2021-02-19
WIP: Add conntrack to gcp route script

michaelgugino:

### DIFF
--- a/templates/master/00-master/gcp/files/opt-libexec-openshift-gcp-routes-sh.yaml
+++ b/templates/master/00-master/gcp/files/opt-libexec-openshift-gcp-routes-sh.yaml
@@ -93,6 +93,11 @@ contents:
                 ensure_rule nat "${CHAIN_NAME}-local" --dst "${vip}" -j REDIRECT
             fi
         done
+        sleep 5
+        for vip in "${!vips[@]}"; do
+            echo "Removing stale conntrack connections for ${vip}"
+            conntrack -D -r "${vip}" || echo "unable to run conntrack"
+        done
     }
 
     clear_rules() {


templates/master/00-master/gcp/files/opt-libexec-openshift-gcp-routes-sh.yaml
```diff
@@ -93,6 +93,11 @@ contents:
                 ensure_rule nat "${CHAIN_NAME}-local" --dst "${vip}" -j REDIRECT
             fi
         done
+        sleep 5
+        for vip in "${!vips[@]}"; do
+            echo "Removing stale conntrack connections for ${vip}"
+            conntrack -D -r "${vip}" || echo "unable to run conntrack"
```
cgwalters, line 99: It looks like `/usr/sbin/conntrack` is actually part the same package `conntrack-tools` which includes a whole daemon `conntrackd.service` that...I don't think we would use in OpenShift.

One option is to change the `Dockerfile` here to do `yum -y install conntrack-tools` and teach this script to exec back into the MCD container to run this.

This all relates to the larger goal of having more of these scripts migrate to Go code that's part of the MCD instead of injected onto the host.  We have `systemd-run` as a way for the pod to schedule code on the host, we could invent e.g. `/run/bin/machine-config-daemon exec conntrack -D -r "${vip"}` which would use a local socket to exec code in the MCD pod.

cgwalters: Another alternative is to `yum -y install conntrack-tools` in here, and then change the pod startup to copy the binary out to the host.

Another alternative is to make it a hidden extension we only install where needed.

Or finally, we could just ship it in CoreOS but then it quickly becomes an "API" that we have to maintain ~forever and makes it hard to keep track of who's using it and what version they need etc.

cgwalters: (Or yet another option is to have the MCD use the netlink API to do that directly, but I wouldn't willingly inflict programming netlink API on anyone I like :wink: )

cgwalters: Or finally, we could say this whole problem domain should move to the apiserver operator/team and they have a daemonset that runs on the control plane :smile:

cgwalters: For now, a PR to add to the host: https://github.com/openshift/os/pull/502

michaelgugino: I'm a fan of 'do whatever is easiest' in this case.  It appears that today, installing conntrack-tools is easiest.  If this is seen as a negative, we should make a better option the 'easiest' option.

PR #2422, opened 2021-02-20
Bug 1925698: apiserver-watcher handle the host iptables instead of using systemd scripts

aojea:
apiserver-watcher migration to go
    
    Openshift use different mechanism to provide HA to the apiserver.
    Typically, on cloud providers, that functionaliy is done by the
    load balancer, however, each cloud provide LB has its nuances that
    we need to workaround using iptables rules.
    
    Until now, we were using a mix of a pod checking the apiserver status
    and systemd unit files with bash scripts. This setup is complicated
    to maintain and to operate, and can lead to race conditions.
    
    We can use the pod that checks the apiserver status to install the
    corresponding iptables rules.
    

aojea:
/hold 
this is not working

aojea:
/hold cancel
/assign @squeed 
This is ready for review, I've verified there are not more stale conntrack entries

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,149 @@
...
+// interval_in_seconds = 5
+// number_of_probes    = 2
+func (az *azureHandler) successThreshold() int {
+	return 1
+}
+
+func (az *azureHandler) failureThreshold() int {
+	// LB = 6 seconds, plus 10 seconds for propagation
+	return 8
+}
```
aojea, line 106: I took the existing values but it will be good if somebody can review them, I think that azure and gcp lb using different parameters but the previous code was using the same for both

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	// bz1930457 delete stale conntrack entries to the apiserver
+	ip := net.ParseIP(az.vip)
+	if ip.IsGlobalUnicast() {
+		var family netlink.InetFamily
+		family = netlink.FAMILY_V4
+		if ip.To4() == nil {
+			family = netlink.FAMILY_V6
+		}
+		filter := &netlink.ConntrackFilter{}
+		filter.AddIP(netlink.ConntrackOrigDstIP, ip)
```
michaelgugino, line 85: I'm pretty sure this should be 'ConntrackReplySrcIP'

michaelgugino (changes requested):

cmd/apiserver-watcher/azure_handler.go
aojea: > sh-4.4# cat /proc/net/nf_conntrack| grep SYN | grep 6443
> ipv4     2 tcp      6 118 SYN_SENT src=35.191.10.141 dst=10.0.0.2 sport=32768 dport=6443 [UNREPLIED] src=10.0.0.2 dst=35.191.10.141 sport=6443 dport=32768 mark=0 secctx=system_u:object_r:unlabeled_t:s0 zone=0 use=2
> ipv4     2 tcp      6 115 SYN_SENT src=35.191.10.143 dst=10.0.0.2 sport=32768 dport=6443 [UNREPLIED] src=10.0.0.2 dst=35.191.10.143 sport=6443 dport=32768 mark=0 secctx=system_u:object_r:unlabeled_t:s0 zone=0 use=2
> ipv4     2 tcp      6 119 SYN_SENT src=35.191.10.135 dst=10.0.0.2 sport=32768 dport=6443 [UNREPLIED] src=10.0.0.2 dst=35.191.10.135 sport=6443 dport=32768 mark=0 secctx=system_u:object_r:unlabeled_t:s0 zone=0 use=2

this runs before the redirect are working, so it flush everything targeting the VIPs, but maybe `ConntrackReplySrcIP` is safer indeed

aojea: changed, the name is different in the netlink library
```	
ConntrackNatSrcIP         // -src-nat ip    Source NAT ip
```

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	// FIXME: assume IPv4 only because GCP doesn't support IPv6
+	for _, vip := range gcp.vips {
+		// sanity check to not flush unexpected IPs
+		// provided by the metadata service
+		ip := net.ParseIP(vip)
+		if ip == nil || !ip.IsGlobalUnicast() {
+			continue
+		}
+		filter := &netlink.ConntrackFilter{}
+		filter.AddIP(netlink.ConntrackNatSrcIP, ip)
```
michaelgugino, line 90: This is deprecated: https://github.com/vishvananda/netlink/blob/master/conntrack_linux.go#L337

We definitely want ConntrackReplySrcIP here.

Also, we need to ensure the conntrack steps are applied after iptables rules sync, not before, so we don't race against ourselves.

michaelgugino (changes requested):

cmd/apiserver-watcher/gcp_handler.go
aojea: > This is deprecated: https://github.com/vishvananda/netlink/blob/master/conntrack_linux.go#L337
> We definitely want ConntrackReplySrcIP here.

We are vendoring the netlink stable release v1.0.0, and the code is backwards compatible, ConntrackReplySrcIP is an alias, we can bump the library if we want to have the new name and bring the modifications I've added to the conntrack filtering https://github.com/vishvananda/netlink/commit/bca67dfc8220b44ef582c9da4e9172bf1c9ec973 , but do we really need this or is  a nice to have?

> Also, we need to ensure the conntrack steps are applied after iptables rules sync, not before, so we don't race against ourselves.

yeah, it is how is implemented, check the onSuccess() methods:
syncOnce() do a first sync, if errors, it return,s otherwise it flush the conntrack entries (iptables rules are there already) and then it syncs rules in a loop until the stop signal is sent onFailure()

michaelgugino: As long as we call the conntrack bits whenever we add the rules, we are good.

We should use whatever version of netlink other parts of OpenShift are using for the time being, I guess.  1.1.0 is the latest tag I see there, but even that branch is over a year old now.  I'll leave that decision up to you.

michaelgugino (approved):
LGTM

aojea:
:thinking: 
/test e2e-azure
/retest

cmd/apiserver-watcher/iptables
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec chroot /rootfs /usr/sbin/iptables "$@"
```
yuqi-zhang, line 3: Question: what is this binary being used for? Does go-iptables make direct calls to the binary?

aojea: right, but the binary use some shared libraries, so if the call is not chrooted it fails

cmd/apiserver-watcher/azure_handler.go
squeed: /cc @sttts

cmd/apiserver-watcher/iptables
squeed: FYI, we also need to know if iptables-nft or iptables-legacy is in use. And we might need to `modprobe`

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,166 @@
...
+const azureMasqChainName = "azure-vips"
+
+// deleteRules deletes the iptables rules added for the VIPs
+func (az *azureHandler) deleteRules() error {
+	glog.Infof("Deleting iptables VIP rules for: %v", az.vip)
+	return az.ip4tables.ClearChain("nat", azureMasqChainName)
+}
+
+// syncRulesOnce syncs ip masquerade rules
+// TODO ipv6
```
squeed, line 118: tsk tsk :-)
Please open a bug for this.

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return &gcpHandler{
+		ip4tables: ipt4,
+		stopCh:    make(chan struct{}),
+	}, nil
+}
+
+// onFailure: either stop the routes service, or write downfile
+func (gcp *gcpHandler) onFailure() error {
```
squeed, line 62: We need to be cleverer here - this is a bit unlike Azure.

redirection is required for *all* GCP LB VIPs, not just the apiserver. If, for whatever reason, there is another load-balancer pointing to this node, it will also need redirection configured - but we shouldn't remove that redirection if the apiserver happens to be down.

So, we need to track which vip, specifically, applies to the apiserver, and only manage that one.

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	return 1
+}
+
+func (gcp *gcpHandler) failureThreshold() int {
+	// LB = 6 seconds, plus 10 seconds for propagation
+	return 8
+}
+
+// deleteRules deletes the iptables rules added for the VIPs
+func (gcp *gcpHandler) deleteRules() error {
```
squeed, line 117: Ideally, we always handle traffic coming from the load balancer - even if the apiserver is maybe going down, it would be good not to drop it on the floor.

The challenge will be telling the difference between traffic from the LB, and traffic from the pods. Perhaps we need a separate rule for traffic from the default interface?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,209 @@
...
+	// bz1930457 delete stale conntrack entries to the apiserver
+	// FIXME: assume IPv4 only because GCP doesn't support IPv6
+	for _, vip := range gcp.vips {
+		// sanity check to not flush unexpected IPs
+		// provided by the metadata service
+		ip := net.ParseIP(vip)
+		if ip == nil || !ip.IsGlobalUnicast() {
+			continue
+		}
+		filter := &netlink.ConntrackFilter{}
```
squeed, line 89: any states we can add here, just in case we over-match?

cmd/apiserver-watcher/iptables
aojea: This is the same as in openshift/sdn https://github.com/openshift/sdn/blob/master/images/iptables-scripts/ip6tables
I think that OS symlinks /usr/sbin iptables, and since this pods mounts the whole rootfs under /rootfs doing a chroot will work with whatever the OS has configured.
Modprobe is done by the iptables userspace tools IIRC

cmd/apiserver-watcher/gcp_handler.go
aojea: it keeps the VIPs that are active and remove the stale ones, is just calling syncRuleOnce()  again.
I don't know the logic in GCP between Loadbalancer, forwarded-ips, instances and where to obtain the relations

smarterclayton:
We still on track for this?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,214 @@
...
+		return err
+	}
+	var s []string
+	err = json.Unmarshal([]byte(resp), &s)
+	if err != nil {
+		return err
+	}
+	newIPs := sets.NewString(s...)
+	staleVips := oldIPs.Difference(newIPs).List()
+	gcp.vips = newIPs.List()
```
squeed, line 146: This isn't quite right.

We need to do two things:
1. for ALL vips, regardless of up or down, ensure there is a rule in POSTROUTING (that is, for external traffic).
2. for **only** the down apiserver vip, delete the rule in OUTPUT, but just for that vip.

This is so any connections from the LB are handled, but connections from the local kubelet go to out to the LB.

aojea: how can you know which is the apiserver VIP, the metadata returns a list of IPs that are being forwarded to the instance
```
[]string{`curl`, `-H`, `Metadata-Flavor: Google`, `http://169.254.169.254/computeMetadata/v1/instance/network-interfaces/0/forwarded-ips/?recursive=true`, `[10.0.0.2,34.75.177.157,34.75.192.108]` }
```
does it makes sense keeping redirect rules to IPs that are not there?

squeed: you know the apiserver internal vip because you know the apiserver hostname; the apiserver-watcher looks it up for you.

All other LB vips need to be left alone.

squeed: you know the apiserver internal vip because you know the apiserver hostname; the apiserver-watcher looks it up for you.

All other LB vips need to be left alone.

aojea: talked with Casey offline, he explained me the correct behavior, is the vip we passed as parameter the one that this code has to handle.

cmd/apiserver-watcher/gcp_handler.go
aojea: this happens once, only on success and if the nated ip is the vip (that means it wasn't redirected)
It looks safe to me

cmd/apiserver-watcher/run.go
```diff
@@ -67,10 +62,28 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
...
+		}
+		if utilnet.IsIPv6String(vip) != runOpts.ipv6 {
+			return fmt.Errorf("vip %s IP family mismatch, ipv6 option %v", vip, runOpts.ipv6)
+		}
+	} else {
+		addrsList, err := net.LookupHost(uri.Hostname())
+		if err != nil {
+			return fmt.Errorf("failed to lookup host %s: %v", uri.Hostname(), err)
+		}
+		addrs := filterAddrList(addrsList, runOpts.ipv6)
```
aojea, line 80: This solves the problem with dualstack urls, before it checked it only returned 1 address for all families, breaking dualstack.
This keeps the same restriction of 1 address, but for only one IP family

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,168 @@
...
+const azureMasqChainName = "azure-vips"
+
+// deleteRules deletes the iptables rules added for the VIPs
+func (az *azureHandler) deleteRules() error {
+	glog.Infof("Deleting iptables VIP rules for: %v", az.vip)
+	return az.ipt.ClearChain("nat", azureMasqChainName)
+}
+
+// syncRulesOnce syncs iptables rules
+// FIXME add IPv6 support
```
dcbw, line 119: @aojea is this FIXME still relevant?

dcbw:
/lgtm

@kikisdeliveryservice any chance you can take a look at this? Thanks!

cmd/apiserver-watcher/azure_handler.go
aojea: nope , I have to remove it

aojea:
I've splitted the dual-stack fix so we can cherry-pick it

cmd/apiserver-watcher/azure_handler.go
aojea: fixed

aojea:
quota issues in aws
```
level=error msg=Error: Error creating IAM instance profile ci-op-gl9qdyyb-5e954-76xmr-bootstrap-profile: LimitExceeded: Cannot exceed quota for InstanceProfilesPerAccount: 1000
```

dcbw:
ovn-step-registry is failing because of `: [sig-arch] Managed cluster should only include cluster daemonsets that have maxUnavailable update of 10 or 33 percent [Suite:openshift/conformance/parallel]` which shoudl be fixed by https://github.com/openshift/origin/pull/25962

/retest

aojea:
@yuqi-zhang is this something you can help us to get reviewed?

kikisdeliveryservice:
@aojea : Is this exercised by some e2e test? or are tests generally passing evidence that it is working correctly without regressions?

aojea:
> @aojea : Is this exercised by some e2e test? or are tests generally passing evidence that it is working correctly without regressions?

if the apiserver-watcher doesn't work azure and gcp will fail during the installation process because the pods will not be able to reach the apiserver.

aojea:
> or are tests generally passing evidence that it is working correctly without regressions?

the answer is that is just not the test, it will fail at installation time

dcbw:
@kikisdeliveryservice see Antonio's response; he didn't @ you. Thanks!

cmd/apiserver-watcher/run.go
```diff
@@ -267,3 +272,14 @@ func (sl *healthTracker) OnComplete(state *health.State) {
 		}
 	}
 }
+
+// filterAddrList filter a list of IP addresses by IP family
```
kikisdeliveryservice, line 242: filterAddrList filters a list of IP addresses and return list of addresses in IP family enabled by flag runOpts.ipv6

or some such thing that I'll be able to read and not forget the purpose of this whole thing ;)

cmd/apiserver-watcher/README.md
```diff
@@ -39,7 +39,7 @@ it's down, or else the node (i.e. kubelet) loses access to the apiserver VIP
 and becomes unmanagable.
```
kikisdeliveryservice, line 41: thanks for updating the docs :heart:

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,167 @@
...
+//
+// Example rules using vip 10.0.0.1:
+//
+// -A PREROUTING -m comment --comment "azure LB vip overriding for pods" -j azure-vips
+// -A OUTPUT -m comment --comment "azure LB vip overriding for local clients" -j azure-vips
+// -A azure-vips --dst 10.0.0.1 -j REDIRECT
+
+type azureHandler struct {
+	vip    string
+	ipt    *iptables.IPTables
```
kikisdeliveryservice, line 34: woud love this to be iptable or iptbl

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,167 @@
...
+	if isIPv6 {
+		proto = iptables.ProtocolIPv6
+	}
+
+	ipt, err := iptables.New(
+		iptables.IPFamily(proto),
+		iptables.Timeout(iptablesTimeout),
+	)
+	if err != nil {
+		return nil, err
```
kikisdeliveryservice, line 51: do we want a message here about new azure handler....blah?

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return &azureHandler{
+		vip:       vip,
+		ip4tables: ipt4,
+		ip6tables: ipt6,
+		stopCh:    make(chan struct{}),
+	}, nil
+}
+
+// onFailure: either stop the routes service, or write downfile
```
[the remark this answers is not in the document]
kikisdeliveryservice: is this downfile comment correct given ^^^

cmd/apiserver-watcher/azure_handler.go
kikisdeliveryservice: no idea on this did anyone double check?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,215 @@
...
+
+}
+
+// https://github.com/openshift/installer/blob/master/data/data/gcp/network/lb-private.tf
+// healthy_threshold   = 3
+// unhealthy_threshold = 3
+// check_interval_sec  = 2
+// timeout_sec         = 2
+
+func (gcp *gcpHandler) successThreshold() int {
```
kikisdeliveryservice, line 120: are the azure/gcp success/failure thresholds always the same or is this a coincidence?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	return 1
+}
+
+func (gcp *gcpHandler) failureThreshold() int {
+	// LB = 6 seconds, plus 10 seconds for propagation
+	return 8
+}
+
+// syncRulesOnce syncs ip masquerade rules
+// TODO ipv6
```
kikisdeliveryservice, line 130: is there a reason this is a TODO and not done?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,215 @@
...
+func (gcp *gcpHandler) syncRulesOnce() error {
+	// sync VIPs from the metadata
```
kikisdeliveryservice, line 132: q: in practice are these errors clear enough for troubleshooting (that aren't really wrapped) - i dont know, you tell me =)

cmd/apiserver-watcher/run.go
kikisdeliveryservice: could we add this as a comment here bc i think that would be helpful

kikisdeliveryservice (commented):
since this is transitioning scripts to go and doing some dual stack razzle dazzle had a few clarifying questions/comments as a non-apiserver sme

cmd/apiserver-watcher/gcp_handler.go
aojea: gcp doesn't support IPv6 yet, I can remove the comment or explain the reason, whatever you think is better

cmd/apiserver-watcher/azure_handler.go
aojea: `ipt` is the way it is implemented in kube-proxy, I was keeping the same naming

cmd/apiserver-watcher/gcp_handler.go
kikisdeliveryservice: explain is cool with me :)

cmd/apiserver-watcher/azure_handler.go
aojea: will add the message to the caller in run.go

cmd/apiserver-watcher/azure_handler.go
aojea: this doesn't change the timers, so it will keep working as before

cmd/apiserver-watcher/gcp_handler.go
aojea: this is how it was before, hence my comment here 
https://github.com/openshift/machine-config-operator/pull/2422/commits/4a6ddb3218cca0f53a406fddb14062960ff00400#r584729261

cmd/apiserver-watcher/gcp_handler.go
aojea: wrapping is better

aojea: the iptables errors are meaningful though, but this from the instance metdata I don't really now

cmd/apiserver-watcher/README.md
aojea: :smiley:

cmd/apiserver-watcher/azure_handler.go
kikisdeliveryservice: cool fair point

aojea: naming is hard :upside_down_face:

aojea:
hmm
is this failure legit?
```
=== RUN   TestReconcileAfterBadMC
panic: test timed out after 1h15m0s
```

yuqi-zhang:
/retest

>=== RUN   TestReconcileAfterBadMC
panic: test timed out after 1h15m0s

This is the test hitting the overall timeout. It seems that we've been hitting it more frequently recently

aojea:
/retest
This time it failed later, I don't know the tests to judge if this change is impacting it

```
--- PASS: TestNoReboot (40.76s)
=== RUN   TestPoolDegradedOnFailToRender
--- PASS: TestPoolDegradedOnFailToRender (14.14s)
=== RUN   TestReconcileAfterBadMC
    utils_test.go:67: Pool worker has rendered config add-a-file-0f35350f-3a5f-4c88-8870-caf8de4fb850 with rendered-worker-c365550ff4c5dadde9a0105fb31d1599 (waited 4.036454685s)
    utils_test.go:89: Pool worker has completed rendered-worker-9a4870d75192b6490b868f7fa05a1b74 (waited 20.019772933s)
--- PASS: TestReconcileAfterBadMC (36.25s)
=== RUN   TestDontDeleteRPMFiles
    utils_test.go:67: Pool infra has rendered config modify-host-file-4153658b-c30c-4fa6-aa95-4075961b9390 with rendered-infra-a2b524477b6cbdc0e30e6c3f730e0476 (waited 6.038291025s)
    utils_test.go:89: Pool infra has completed rendered-infra-a2b524477b6cbdc0e30e6c3f730e0476 (waited 2m52.01998607s)
    utils_test.go:89: Pool infra has completed rendered-infra-9a4870d75192b6490b868f7fa05a1b74 (waited 1m50.019919309s)
    utils_test.go:89: Pool infra has completed rendered-infra-9a4870d75192b6490b868f7fa05a1b74 (waited 2.019176059s)
--- PASS: TestDontDeleteRPMFiles (297.04s)
=== RUN   TestIgn3Cfg
    mcd_test.go:707: Created 99-ign3cfg-infra-f204a08f-76c0-42fd-813e-43b7f8f51279
    utils_test.go:67: Pool infra has rendered config 99-ign3cfg-infra-f204a08f-76c0-42fd-813e-43b7f8f51279 with rendered-infra-a1774c96e7b3cdb0099e26905ba698c3 (waited 6.054712351s)
panic: test timed out after 1h15m0s
```

aojea:
/retest
upgrade job failed with
```

4/89 Tests Failed. | expand_less
-- | --
: [sig-node] pods should never transition back to pending expand_more0s | : [sig-node] pods should never transition back to pending expand_more | 0s
: [sig-node] pods should never transition back to pending expand_more | 0s
: [sig-network] pods should successfully create sandboxes by writing network status expand_more | : [sig-network] pods should successfully create sandboxes by writing network status expand_more
: [sig-network] pods should successfully create sandboxes by writing network status expand_more
```

kikisdeliveryservice:
> /retest
> upgrade job failed with
> 
> ```
> 
> 4/89 Tests Failed. | expand_less
> -- | --
> : [sig-node] pods should never transition back to pending expand_more0s | : [sig-node] pods should never transition back to pending expand_more | 0s
> : [sig-node] pods should never transition back to pending expand_more | 0s
> : [sig-network] pods should successfully create sandboxes by writing network status expand_more | : [sig-network] pods should successfully create sandboxes by writing network status expand_more
> : [sig-network] pods should successfully create sandboxes by writing network status expand_more
> ```

this is: https://bugzilla.redhat.com/show_bug.cgi?id=1933760

hoping to get a fix soon

aojea:
I can confirm this fixes the dual-stack problem too

aojea:
ok, now we have other failing test that I saw in other PRs
```
kubelet terminates kube-apiserver gracefully 
```

The ci/prow/e2e-gcp-op failed because couldn't take a lease

/retest

smarterclayton:
Good to see us go green on zero downtime upgrade on GCP.

What remaining issues are there?  I will grant override exceptions on the upgrade flake as necessary.

dcbw:
@smarterclayton e2e-gcp-op failure looks like a setup/install issue, but I can't tell why:

```
error: some steps failed:
  * could not run steps: step e2e-gcp-op failed: "e2e-gcp-op" pre steps failed: "e2e-gcp-op" pod "e2e-gcp-op-ipi-install-install" failed: the pod ci-op-hvb5kv5c/e2e-gcp-op-ipi-install-install was deleted without completing after 31m20s (failed containers: )
Link to step on registry info site: https://steps.ci.openshift.org/reference/ipi-install-install
Link to job on registry info site: https://steps.ci.openshift.org/job?org=openshift&repo=machine-config-operator&branch=master&test=e2e-gcp-op
time="2021-03-18T09:18:27Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:step_failed:utilizing_lease:executing_test:executing_multi_stage_test'"
```

but the install-install artifacts say success:

```
level=info msg=Creating infrastructure resources...
level=info msg=Waiting up to 20m0s for the Kubernetes API at https://api.ci-op-hvb5kv5c-1354f.origin-ci-int-gce.dev.openshift.com:6443...
level=info msg=API v1.20.0-1085+e1bc274ae64069-dirty up
level=info msg=Waiting up to 30m0s for bootstrapping to complete...
level=info msg=Destroying the bootstrap resources...
level=info msg=Waiting up to 40m0s for the cluster at https://api.ci-op-hvb5kv5c-1354f.origin-ci-int-gce.dev.openshift.com:6443 to initialize...
level=info msg=Waiting up to 10m0s for the openshift-console route to be created...
level=info msg=Install complete!
level=info msg=To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/tmp/installer/auth/kubeconfig'
level=info msg=Access the OpenShift web-console here: https://console-openshift-console.apps.ci-op-hvb5kv5c-1354f.origin-ci-int-gce.dev.openshift.com
level=info msg=Time elapsed: 31m4s
Setup phase finished, prepare env for next steps
Copying log bundle...
Removing REDACTED info from log...
Copying required artifacts to shared dir
```

so no clue what's happening there. Same issue with aws-workers-rhel7

smarterclayton:
The first e2e-gcp failure was I believe related to infrastructure issues in CI (not caused by this PR)

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,167 @@
...
+
+	ipt, err := iptables.New(
+		iptables.IPFamily(proto),
+		iptables.Timeout(iptablesTimeout),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &azureHandler{
```
dcbw, line 54: @aojea do we need to remove stale VIP rules on startup? Looks like the old script might do that:

```
remove_stale() {
        ## find extra iptables rules
        for ipt_vip in $(iptables -w -t nat -S "${CHAIN_NAME}" | awk '$4{print $4}' | awk -F/ '{print $1}'); do
            if [[ ! -v v4vips[${ipt_vip}] ]] || [[ "${v4vips[${ipt_vip}]}" = down ]]; then
                echo removing stale vip "${ipt_vip}" for local clients
                iptables -w -t nat -D "${CHAIN_NAME}" --dst "${ipt_vip}" -j REDIRECT
            fi
        done

```

aojea: my observation is that when the pod starts it always goes to onfailure state, and whenever you calls that it flush the rules there
```
// deleteRules deletes the iptables rules added for the VIPs
func (az *azureHandler) deleteRules() error {
	glog.Infof("Deleting iptables VIP rules for: %v", az.vip)
	return az.ipt.ClearChain("nat", azureMasqChainName)
}
```

dcbw: Fair enough.

smarterclayton:
/retest

upgrade was machine pool not completing before upgrade complete (unrelated)

aojea:
/retest
```
: [sig-network] pods should successfully create sandboxes by writing network status expand_less
```

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	// wait for iptables sync loop to finish
+	time.Sleep(iptablesTimeout * time.Second)
+
+	// indicate the host script that handles GCP LB vips
+	// to not forward traffic to the apiserver VIP
+	glog.Infof("apiserver down: create downfile: %s.down", gcp.vip)
+	if err := writeVipStateFile(gcp.vip, "down"); err != nil {
+		return fmt.Errorf("error writing state file for vip UP: %v", err)
+	}
```
aojea, line 95: @squeed on failures we now create the vip.down file to indicate the host process that the vip is down

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// onSuccess: either start routes service, or remove down file
+func (gcp *gcpHandler) onSuccess() error {
+	// indicate the host script that handles GCP LB vips that
+	// it should not delete the apiserver VIP
+	glog.Infof("apiserver down: deleting downfile: %s.down", gcp.vip)
+	if err := removeVipStateFile(gcp.vip, "down"); err != nil {
+		return fmt.Errorf("error writing state file for vip UP: %v", err)
+	}
```
aojea, line 108: @squeed on success we just remove the vip.down file, no need to use the vip.up since we install the vip iptables rules if it doesn't already exist

aojea:
Hmm the gcp job fails now

aojea:
/hold
I will keep the iptables logic in the scripts, and move all the conntrack to the apiserver-watcher

aojea:
new PR for the conntrack issue here https://github.com/openshift/machine-config-operator/pull/2477

aojea:
/retest
```
: [sig-network] pods should successfully create sandboxes by writing network status expand_less
```

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,167 @@
...
+// https://github.com/openshift/installer/blob/master/data/data/azure/vnet/internal-lb.tf
+// interval_in_seconds = 5
+// number_of_probes    = 2
+func (az *azureHandler) successThreshold() int {
+	return 1
+}
+
+func (az *azureHandler) failureThreshold() int {
+	// LB = 6 seconds, plus 10 seconds for propagation
+	return 8
```
sttts, line 106: why 8 ?

cmd/apiserver-watcher/azure_handler.go
```diff
@@ -0,0 +1,167 @@
...
+	go az.syncRulesUntil()
+	return nil
+
+}
+
+// https://github.com/openshift/installer/blob/master/data/data/azure/vnet/internal-lb.tf
+// interval_in_seconds = 5
+// number_of_probes    = 2
+func (az *azureHandler) successThreshold() int {
+	return 1
```
sttts, line 101: why 1 ?

cmd/apiserver-watcher/gcp_handler.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := gcp.ipt.AppendUnique("nat", gcpLocalMasqChainName, "--dst", gcp.vip, "-j", "REDIRECT"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// syncRulesUntil syncs the iptables rules until it receives the stop signal
+func (gcp *gcpHandler) syncRulesUntil() {
+	ticker := time.NewTicker(30 * time.Second)
```
sttts, line 188: why do you think 30s is correct? Don't we sync much faster in the bash script (we have to!) ?

sttts:
/hold

Please ask us for final review when ready. This is crucial for apiserver and workloads teams.

Note: we have nothing against moving this to Golang, but we have to make sure that the semantics match jour expectations.

cmd/apiserver-watcher/azure_handler.go
aojea: those were the previous values, https://github.com/openshift/machine-config-operator/pull/2422/commits/bcba00705b625c0fb230b0a5cb8edeb880b0107d#r584729261
I kept them

cmd/apiserver-watcher/azure_handler.go
aojea: carried over the previous values

cmd/apiserver-watcher/gcp_handler.go
aojea: the bash script has to sync faster because is not able to react to apiserver changes, it depends on this pod, to writes the api status file, and the insert the rules.
This scripts reacts intimidately to the apiserver status so the rules are programmed immediately., this cycle is just to be sure that nobody delete our rules

aojea:
> /hold
> 
> Please ask us for final review when ready. This is crucial for apiserver and workloads teams.
> 
> Note: we have nothing against moving this to Golang, but we have to make sure that the semantics match jour expectations.

absolutely

aojea:
/close
if we are going to do it there are much better and elegant options

PR #2423, opened 2021-02-22
Bug 1931507: ovsdb-server: chown when restarting ovsdb-server

vrutkovs:
**- What I did**
Added a dropin which sets permissions for openvswitch files in `ovsdb-server` service.

In OVN `ovsdb-server` service starts before `ovs-vswitchd`, so chown
needs to run there. Unlike a dropin in `ovsdb-server` setting a group
is insufficient - user needs to own the OVS DB lock file too.

**- How to verify it**
Upgrade 4.6 OVN to 4.7. All nodes boot correctly after MCD updates the node.

Permissions for openvswitch files are being updated for OVN installs on upgrade.

### DIFF
--- a/templates/common/_base/units/ovsdb-server.service.yaml
+++ b/templates/common/_base/units/ovsdb-server.service.yaml
@@ -5,3 +5,6 @@ dropins:
     contents: |
       [Service]
       Restart=always
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /var/lib/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /etc/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /run/openvswitch'


templates/common/_base/units/ovsdb-server.service.yaml
```diff
@@ -5,3 +5,6 @@ dropins:
     contents: |
       [Service]
       Restart=always
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /var/lib/openvswitch'
```
abhat, line 8: Why not use `:$${OVS_USER_ID##}`?

vrutkovs: It was insufficient in my case - not entirely sure why. This script should ensure that a different user owns these folders anyway (at least its the case in current OKD 4.6 -> 4.7 upgrade)

abhat: It's unclear to me what insufficient means. What happens if you do `:$${OVS_USER_ID##*:}`?

abhat: OVS_USER_ID is coming from openvswitch.useropts which in itself is getting set from either default.conf or /etc/config/openvswitch.

vrutkovs: >It's unclear to me what insufficient means. What happens if you do :$${OVS_USER_ID##*:}?

`ovsdb-server` service fails on boot. I can't ssh and check more details.

Changing this to `chown -R ${OVS_USER_ID}` makes the machine boot

LorbusChris:
@vrutkovs this is one of the missing pieces to get OKD back to using mainline. Should we re-open?

PR #2425, opened 2021-02-22
Bug 1928473: Apply system-connections-merged to all platforms

bcrochet:
In order to reduce complexity, and to promote consistency between platforms,
and thereby reducing an opportunity for future issues, this patch introduces
the OverlayFS mounted /etc/NetworkManger/system-connections-merged to all
platforms. It is meant to be a transparent change, and should not change the
functionality of NetworkManager on any platform. The purpose is to make any
NetworkManager configuration that is not performed by Machine Config Operator
be ephemeral.

**- What I did**

**- How to verify it**


darkmuggle:
/lgtm 

This was discussed at a high level in a meeting with the Networking team.

PR #2431, opened 2021-02-24
Bug 1927731: Revert "pkg/daemon: Add IgnitionVersion to Daemon"

bgilbert:
https://github.com/openshift/machine-config-operator/pull/1729 introduced code to obtain the version of the Ignition binary shipped in the OS root filesystem.  That PR contemplated that the version would be used more widely in the MCD, but that hasn't occurred, so its only current use is to log the Ignition version at MCD startup.  On balance, I think that code path introduces more risk than value, so this PR backs it out.  My reasoning is:

1. By the time this code is invoked, Ignition will never run again, so its version is useful for forensics and not much else.  In particular, the version of the Ignition binary in machine-os-content probably doesn't match the one in the bootimage.
2. MCD is invoking the Ignition binary via a private (non-contractual) path in the root filesystem, but Ignition is designed and tested only for use in the initramfs.  In practice, `ignition --version` _should_ be harmless since it exits very early in Ignition startup, but see below.
3. If the MCD fails to run the binary, say because the path to Ignition has changed, MCD considers that a fatal error.
4. And indeed, we now have a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1927731) where an `ignition --version` segfault is apparently blocking an upgrade from 4.5 to 4.6, on exactly one customer node.  From the [stack trace](https://bugzilla.redhat.com/show_bug.cgi?id=1927731#c7), the problem could be in Ignition itself, in the initialization code of a vendored library, or in the Go runtime.  As a practical matter, the crash is unlikely to be root-caused.

cc @LorbusChris @sinnykumari @ashcrow from the original PR.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -57,9 +57,6 @@ type Daemon struct {
 	// os the operating system the MCD is running on
 	os OperatingSystem
 
-	// IgnitionVersion is the version of the installed Ignition binary on the system
-	IgnitionVersion string
-
 	// mock is set if we're running as non-root, probably under unit tests
 	mock bool
 
@@ -128,8 +125,6 @@ type Daemon struct {
 }
 
 const (
-	// pathIgnition is the path where the ignition binary resides
-	pathIgnition = "/usr/lib/dracut/modules.d/30ignition/ignition"
 	// pathSystemd is the path systemd modifiable units, services, etc.. reside
 	pathSystemd = "/etc/systemd/system"
 	// pathDevNull is the systems path to and endless blackhole
@@ -202,7 +197,6 @@ func New(
 	var (
 		osImageURL string
 		osVersion  string
-		ignVersion string
 		err        error
 	)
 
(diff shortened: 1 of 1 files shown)


darkmuggle:
@bgilbert and I have spoken at length on this.  

However, I support the removal of this check based solely on the reason 3 -- the check is a fatal error for purely informational purposes at this point. The MCD team spoke quite a bit about simplifying and making the code safe; this PR accomplishes that goal.

kikisdeliveryservice:
Would love for ci to finish running but on an initial pass, it seems like this isn't providing much value but causing problems (via  reason 3 above).

LorbusChris (approved):

kikisdeliveryservice:
Tests are a mess today... will check back in later.

/retest

sinnykumari (approved):
Agree with reasoning provided here. In future if we need to know and utilize ignition version, we can revisit and discuss with ignition team to do it in safer way.

kikisdeliveryservice (approved):
seems like we have agreement this is good so

/lgtm

kikisdeliveryservice:
That umm didnt work..

/lgtm

kikisdeliveryservice:
@bgilbert do we want to/how far should we backport this?

bgilbert:
It looks like the code was introduced in 4.6.  Since [RHBZ 1927731](https://bugzilla.redhat.com/show_bug.cgi?id=1927731) reports a failure in a 4.6 upgrade, I'd be inclined to backport to both 4.6 and 4.7.

bgilbert:
4.6 cherry-pick in https://github.com/openshift/machine-config-operator/pull/2439.

PR #2433, opened 2021-02-24
Bug 1932105: status: do not update status until master pool completes

yuqi-zhang:
There appears to be situations where our master pool is still progressing to the new version's rendered config, but the version reports completed.

Looking at our versioning/progress reporting it doesn't seem to be explicitly tied to pool status, maybe that's by design? Adding in explicit dependency on required pools will at least prevent version from updating before pool completion.

Still digging in to why this (mostly) works today and why in some scenarios it does not, marking WIP

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -34,6 +34,22 @@ func (optr *Operator) syncVersion() error {
 		return nil
 	}
 
+	pools, err := optr.mcpLister.List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	for _, pool := range pools {
+		_, hasRequiredPoolLabel := pool.Labels[requiredForUpgradeMachineConfigPoolLabelKey]
+		if hasRequiredPoolLabel {
+			updated := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated)
+			if !updated {
+				// Don't update version until master pool has settled.
+				return nil
+			}
+		}
+	}
+
 	if !optr.vStore.Equal(co.Status.Versions) {
 		mcoObjectRef := &corev1.ObjectReference{
 			Kind:      co.Kind,


pkg/operator/status.go
```diff
@@ -34,6 +34,22 @@ func (optr *Operator) syncVersion() error {
 		return nil
```
wking, line 34: This^^ condition, where we refuse to bump the version when `Progressing == True && Degraded == True`, is [ancient][1] (born with `syncVersion`), but doesn't make sense to me.  The required-pool guard you're adding here makes a lot more sense to me.  Feelings about dropping this guard now that you're adding the new pool guard?

What other operands does the MCO have?  There are the MCDs that get rolled out before pools, and the MCSs that... are also rolled out before the pools?  Is it worth checking those explicitly here?  Or do we lean on them blocking pool update attempts, and then only need to check the pools explicitly here?

[1]: https://github.com/openshift/machine-config-operator/commit/6faac492e7de523a2e54ae23d7bc0c0294d5946a#diff-8a6a064b4db19eb3b29e5280447ac3ba9496e847391489d25c1282767c3354c5R30

wking: Looking at `Progressing`, the only touches under `pkg/` seem to be [`syncProgressingStatus`][1].  The only significant conditional there is on `optr.vStore.Equal(co.Status.Versions)`, so... it's going `Progressing=True` when the current requested operator version diverges from the ClusterOperator's `status.versions`, and then, because of this^^ guard, stays `Progressing` until `Degraded!=True`, after which the version levels?  I'm going to go look at `Degraded`...

[1]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/status.go#L115-L157

wking: Ok, [`syncDegradedStatus`][1] seems to hold the current meat.  Stepping through my current understanding of [`syncAll`][2]:

1. CVO bumps the MCO to a new version, including the environment variable with the new target version.
2. `syncProgressingStatus` notices the new requested version diverges from the ClusterOperator `status.versions`, and sets `Progressing=True`.  Great.
3. `syncAll` iterates through sync functions sequentially.  If there's an error, it bails out of that loop, and...
4. ... passes the error through to `syncDegradedStatus`.  That's overly aggressive, because there could be a number of errors with syncing that do not mean [you're missing SLOs or degrading QoS][3].
5. On error, `syncDegradedStatus` sets `Degraded=True`.
6. [`syncAvailableStatus`][4] sets `Available=False`.  Also overly aggressive, because being Degraded=True doesn't mean you're immediately completely dead.
7. Then we get down to `syncVersion`, and currently bail out because of this^^ guard, because `Progressing == True && Degraded==True`.

That cycle repeats until we get through all the sync functions without errors, after which `syncDegradedStatus` notices the lack of error and sets `Degraded=False`.  That is enough to get through this^^ guard, and `syncVersion` bumps `status.Versions`.  That is enough to get `syncProgressingStatus` to set `Progressing=False`.

[1]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/status.go#L191-L239
[2]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/sync.go#L60-L109
[3]: https://github.com/openshift/api/blob/70b287c0a0a51be5cdc88d40858de59e31222f5f/config/v1/types_cluster_operator.go#L153-L169
[4]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/status.go#L85-L113

wking: Ok, [`syncDegradedStatus`][1] seems to hold the current meat.  Stepping through my current understanding of [`syncAll`][2]:

1. CVO bumps the MCO to a new version, including the environment variable with the new target version.
2. `syncProgressingStatus` notices the new requested version diverges from the ClusterOperator `status.versions`, and sets `Progressing=True`.  Great.
3. `syncAll` iterates through sync functions sequentially.  If there's an error, it bails out of that loop, and...
4. ... passes the error through to `syncDegradedStatus`.  That's overly aggressive, because there could be a number of errors with syncing that do not mean [you're missing SLOs or degrading QoS][3].
5. On error, `syncDegradedStatus` sets `Degraded=True`.
6. [`syncAvailableStatus`][4] sets `Available=False`.  Also overly aggressive, because being Degraded=True doesn't mean you're immediately completely dead.
7. Then we get down to `syncVersion`, and currently bail out because of this^^ guard, because `Progressing == True && Degraded==True`.

That cycle repeats until we get through all the sync functions without errors, after which `syncDegradedStatus` notices the lack of error and sets `Degraded=False`.  That is enough to get through this^^ guard, and `syncVersion` bumps `status.Versions`.  That is enough to get `syncProgressingStatus` to set `Progressing=False`.

[1]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/status.go#L191-L239
[2]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/sync.go#L60-L109
[3]: https://github.com/openshift/api/blob/70b287c0a0a51be5cdc88d40858de59e31222f5f/config/v1/types_cluster_operator.go#L153-L169
[4]: https://github.com/openshift/machine-config-operator/blame/f53cb9aa4a9c81ec4d0dfe6cb734957d4523676a/pkg/operator/status.go#L85-L113

yuqi-zhang:
Lifting WIP, after digging at logs I'm reasonably confident this should work as a temporary measure to patch the problem if we so wish. However, our syncRequiredMachineConfigPools function seem to be poorly constructed, as well as syncVersion and syncProgressing also can be improved upon. If we want a more robust solution I can try to fix it more next week.

kikisdeliveryservice:
> Lifting WIP, after digging at logs I'm reasonably confident this should work as a temporary measure to patch the problem if we so wish. However, our syncRequiredMachineConfigPools function seem to be poorly constructed, as well as syncVersion and syncProgressing also can be improved upon. If we want a more robust solution I can try to fix it more next week.

Would rather just take this now as I'm working in the sync functions for my epic.

kikisdeliveryservice:
I don't quite think this is the right fix, working on a PR...

/hold

wking:
I think #2440 should cover this, and I want the 4.8/master bug ON_QA so we can confirm it works in nightlies.  Closing this to allow the bug to move forward.

/close

PR #2435, opened 2021-02-25
[release-4.6] Bug 1933075: backport systemd dropin/units zero-length handling

darkmuggle:
Dependency backports for Bug 1933077.

This fixes a problem were a previous dropin/unit is replaced with zero
length file by ensuring that drops/units have actual content.

(cherry picked from commit c6df6d8)

The previous commit checks for IsNotExist as a return of
checkFileContentsAndMode, which since we wrap the error, will
never be the case, thus always erroring there. Instead check
for the error in a separate case.

Signed-off-by: Yu Qi Zhang <jerzhang@redhat.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1381,7 +1381,20 @@ func checkV3Units(units []ign3types.Unit) error {
 	for _, u := range units {
 		for j := range u.Dropins {
 			path := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[j].Name)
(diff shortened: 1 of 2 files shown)


darkmuggle:
Cherry-picks did not cleanly apply, this is a hand-crafted artisanal commit:
https://github.com/openshift/machine-config-operator/pull/2363#issuecomment-785991550

kikisdeliveryservice:
Does this need to merge first? Or second?

kikisdeliveryservice (approved):
failing test is very red

/skip

darkmuggle:
> Does this need to merge first? Or second?

@kikisdeliveryservice this should merge first

derekwaynecarr:
(Patch Manager) met criteria.

PR #2440, opened 2021-02-26
Bug 1932105: operator/sync.go restore err when required pools not leveled

kikisdeliveryservice:
In the course of working on wait for all pools (#2231) an error when a required pool isn't yet finished updating was inadvertently removed. Restoring it here.

Replaces  #2433

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -617,6 +617,8 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 					isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated) {
 					continue
 				}
+				lastErr = fmt.Errorf("error required pool %s is not ready, retrying. Status: (total: %d, ready %d, updated: %d, unavailable: %d, degraded: %d)", pool.Name, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount, pool.Status.DegradedMachineCount)
+				return false, nil
 			}
 		}
 		syncstatuserr := optr.syncUpgradeableStatus()


wking (approved):

wking:
[Kubelet / API-server failure][1] unrelated.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2440/pull-ci-openshift-machine-config-operator-master-e2e-aws-serial/1365231667953799168

pkg/operator/sync.go
```diff
@@ -617,6 +617,8 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 					isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdated) {
 					continue
 				}
+				lastErr = fmt.Errorf("error required pool %s is not ready, retrying. Status: (total: %d, ready %d, updated: %d, unavailable: %d, degraded: %d)", pool.Name, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount, pool.Status.DegradedMachineCount)
```
yuqi-zhang, line 620: Thanks for the fix! Just want to make sure I understand this correctly, the `pool.Status.ObservedGeneration` only gets updated to `pool.Generation` when the pool completes its update?

wking: The pool controller should bump `status.observedGeneration` whenever it sees a new `metadata.generation`.  The block above is accepting the pool if the `Updated=True` condition is set and the observed generation is new enough for us to know that that `Updated` is the controller's view vs. the current `spec`.  The restored error this thread is a comment on is what happens when the pool is not acceptable, because the pool controller hasn't noticed a `metadata.generation` bump, the pool isn't `Updated=True` yet, or both.

PR #2442, opened 2021-03-01
Bug 1917282: Explicitly set keyfile as the default plugin of NetworkManager for RHEL7

pliurh:
1. Configure keyfile as the default plugin of NetworkManager explicitly for all nodes. It is the default setting in RHCOS.
2. Fix configure-ovs.sh script for RHEL7 workers with old version NetworkManager.

### DIFF
--- a/templates/common/_base/files/NetworkManager-keyfiles.yaml
+++ b/templates/common/_base/files/NetworkManager-keyfiles.yaml
@@ -2,6 +2,8 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-keyfiles.conf"
 contents:
   inline: |
+    [main]
+    plugins=keyfile,ifcfg-rh
     [keyfile]
     path=/etc/NetworkManager/systemConnectionsMerged
 
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -16,9 +16,10 @@ contents:
       dst_path="/etc/NetworkManager/system-connections"
       if [ -d $src_path ]; then
         echo "$src_path exists"
-        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
+        # In RHEL7 files in /{etc,run}/NetworkManager/system-connections end without the suffix '.nmconnection', whereas in RHCOS they end with the suffix.
+        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
         for file in ${fileList[*]}; do
-          if [ ! -f $dst_path/$file ]; then
+          if [ ! -f $dst_path/$file ] && [ -f $src_path/$file ]; then
             cp $src_path/$file $dst_path/$file
           else
             echo "Skipping $file since it exists in $dst_path"
@@ -179,9 +180,11 @@ contents:
       # bring down any old iface
       nmcli device disconnect $iface
 
+      # use ${extra_phys_args[@]+"${extra_phys_args[@]}"} instead of ${extra_phys_args[@]} to be compatible with bash 4.2 in RHEL7.9
+      extra_args=${extra_phys_args[@]+"${extra_phys_args[@]}"}
       if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
         nmcli c add type ${iface_type} conn.interface ${iface} master ovs-port-phys0 con-name ovs-if-phys0 \
-          connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} "${extra_phys_args[@]}"
+          connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} ${extra_args}
       fi
 
       # Update connections with master property set to use the new device name


mtnbikenc:
workers-rhel7 failing, needs https://github.com/openshift/machine-config-operator/pull/2443

kikisdeliveryservice:
> workers-rhel7 failing, needs #2443

adding hold until the other pr merges

/hold

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -11,9 +11,10 @@ contents:
       dst_path="/etc/NetworkManager/system-connections"
       if [ -d $src_path ]; then
         echo "$src_path exists"
-        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
+        # NetworkManager 1.15.2 or later now requires that files in /{etc,run}/NetworkManager/system-connections end in the suffix '.nmconnection'
+        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
         for file in ${fileList[*]}; do
-          if [ ! -f $dst_path/$file ]; then
+          if [ ! -f $dst_path/$file ] && [ -f $src_path/$file ]; then
```
trozet, line 17: @pliurh is this really necessary? if the file we wrote is not there, then this should be a failure

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -11,9 +11,10 @@ contents:
       dst_path="/etc/NetworkManager/system-connections"
       if [ -d $src_path ]; then
         echo "$src_path exists"
-        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
+        # NetworkManager 1.15.2 or later now requires that files in /{etc,run}/NetworkManager/system-connections end in the suffix '.nmconnection'
```
trozet, line 14: I'm not sure this even works. We had NM bugs like:
https://bugzilla.redhat.com/show_bug.cgi?id=1852106

which require minimum NM package versions in rhel 7.9z. See:
https://bugzilla.redhat.com/show_bug.cgi?id=1882667#c13

templates/common/_base/files/configure-ovs-network.yaml
pliurh: It's because `fileList` contains keyfiles with and without the extension '.nmconnection', e.g. 'br-ex' and 'br-ex.nmconnection'. So to make it work for both RHEL and RHCOS, I add the source file check here.

templates/common/_base/files/configure-ovs-network.yaml
pliurh: Here's the fix in NM https://github.com/NetworkManager/NetworkManager/commit/d37ad15f12bafd91cf724cda50aea7093e04bf7a. It has been merged in NM 1.18.8.

However, in my test with RHEL7.9 and NM 1.18.8-2.el7_9, the NM still generates the keyfile names without the extension 'nmconnection'.

```
[root@rhel-worker ~]# ls /etc/NetworkManager/system-connections-merged/
br-ex  ovs-if-br-ex  ovs-if-phys0  ovs-port-br-ex  ovs-port-phys0
```

trozet:
@msherif1234 PTAL

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -180,8 +181,9 @@ contents:
       nmcli device disconnect $iface
 
       if ! nmcli connection show ovs-if-phys0 &> /dev/null; then
+        # use ${extra_phys_args[@]+"${extra_phys_args[@]}"} instead of ${extra_phys_args[@]} to be compatible with bash 4.2 in RHEL7.9
```
abhat, line 184: can you please comment on why bash4.2 on rhel7.9 needs the complicated syntax to parse `extra_phys_args` array?

pliurh: prior bash 4.4, using `${extra_phys_args[@]}` when `extra_phys_args` is empty and the nounset option is enabled, it throws an unbound variable error. 

https://git.savannah.gnu.org/cgit/bash.git/tree/CHANGES?id=3ba697465bc74fab513a26dea700cc82e9f4724e#n878

templates/common/_base/files/NetworkManager-keyfiles.yaml
```diff
@@ -2,6 +2,8 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-keyfiles.conf"
 contents:
   inline: |
+    [main]
+    plugins=keyfile,ifcfg-rh
```
msherif1234, line 6: why are we doing this while Bug 1928473: Apply system-connections-merged to all platforms already fixed the above ?

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: I understood the file extension differences between the two env but I am not clear on why we need src file check still 
not sure in when we don't have the src file ?

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: interesting never seen this pattern b4 to help with readability can u store the new form of extra_phys_args[@] in new var and use it below

templates/common/_base/files/NetworkManager-keyfiles.yaml
pliurh: No, I don't think it fixed this issue. On RHCOS, the default plugin is set explicitly with `plugins=keyfile,ifcfg-rh` in 10-disable-default-plugins.conf. So I think we need to do the same thing for RHEL7.

templates/common/_base/files/configure-ovs-network.yaml
pliurh: In RHEL we don't have src files with the extension, in RHCOS we don't the src files without the extension.

templates/common/_base/files/configure-ovs-network.yaml
pliurh: fixed.

kikisdeliveryservice:
@mtnbikenc does this work for rhel7?

kikisdeliveryservice:
@abhat Has this been tested on RHEL7 or is the successful CI run sufficient?

abhat:
> @abhat Has this been tested on RHEL7 or is the successful CI run sufficient?

@pliurh has most probably manually tested this on RHEL7. The CI job also captures bringing up OVN on RHEL7 using configure-ovs-network script.

pliurh:
Yes, I did test it manually. I don't think we have any CI job for RHEL+OVN setup. As far as I can tell, the job e2e-aws-workers-rhel7 is openshift-sdn based. That is also the reason why we didn't find this issue earlier.

abhat:
@kikisdeliveryservice is this good to go?

knobunc:
@yuqi-zhang Does this look okay?

PR #2446, opened 2021-03-01
Bug 1933805: Add node label to service monitor

smarterclayton:
In order to associate the metrics of each target, we want to clearly
identify which node the MCD is scheduled on to. This will allow us
to more accurately track target health as it relates to nodes in
order to improve the TargetDown alert during upgrades. In the future,
other alerts should be based on node name, not instance name, which
should simplify behavior.

If the node label is already set, do nothing.

The linked bug reports TargetDown firing for MCD because at least one
instance is down the entire time of the upgrade due to overlapping
control plane / worker updates. We are changing the alert to exclude
instances that are deliberately being upgraded (today by using the
unschedulable flag, in the future potentially by a more targeted MCO
driven metric indicating the phase of node upgrade). We will apply
the fix to all daemonsets in the platform, this is being tested here
first.

Metrics reported by the machine config daemon now include a node label for the node the metric originated from, if the metric did not already include a node label.

### DIFF
--- a/install/0000_90_machine-config-operator_00_servicemonitor.yaml
+++ b/install/0000_90_machine-config-operator_00_servicemonitor.yaml
@@ -15,6 +15,15 @@ spec:
     port: metrics
     scheme: https
     path: /metrics
+    relabelings:
+    - action: replace
+      regex: ;(.*)
+      replacement: $1
+      separator: ";"
+      sourceLabels:
+      - node
+      - __meta_kubernetes_pod_node_name
+      targetLabel: node
     tlsConfig:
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

while we discuss this with monitoring team

simonpasquier (commented):
/lgtm

smarterclayton:
Ok, monitoring team confirmed this is ok.  I will be using this to suppress TargetDown firing on daemonset alerts.

kikisdeliveryservice (approved):
wfm!

PR #2448, opened 2021-03-01
Bug 1930636: Fix overlaySize & pidslimit ctrcfg

QiWang19:
Close: Bug 1930636 https://bugzilla.redhat.com/show_bug.cgi?id=1930636

**- What I did**
Add validation: If the overlaySize < 0, pidslimit < 0, return the failure status.

Fix a corner case of overlaySize and pidslimit containerruntimeconfig


pkg/apis/machineconfiguration.openshift.io/v1/types.go
```diff
@@ -432,7 +432,7 @@ type ContainerRuntimeConfigSpec struct {
 // ContainerRuntimeConfiguration defines the tuneables of the container runtime
 type ContainerRuntimeConfiguration struct {
 	// pidsLimit specifies the maximum number of processes allowed in a container
-	PidsLimit int64 `json:"pidsLimit,omitempty"`
+	PidsLimit *int64 `json:"pidsLimit,omitempty"`
```
mtrmac, line 435: Shouldn’t the CRD be updated to indicate the value is optional? (I don’t understand what are source files and what is generated, and I don’t know all that much about CRDs, this just looks curious to me.)

If `0` is a special case, shouldn’t it be documented? (BTW CRI-O does _not_ document 0 as a special case (but then again, runtime-spec/specs-go does document it), so it is curious that it should be supported here if it is undocumented in CRI-O.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -313,9 +317,9 @@ func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []generatedConfig
 			glog.V(2).Infoln(cfg, err, "error updating user changes for log-level to crio.conf.d: %v", err)
 		}
 	}
-	if ctrcfg.PidsLimit != 0 {
+	if *ctrcfg.PidsLimit >= 0 {
```
mtrmac, line 320: Looking at the caller above, it seems possible that the pointer is `nil` at this point.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -431,14 +435,18 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 	}
 
 	ctrcfg := cfg.Spec.ContainerRuntimeConfig
-	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < minPidsLimit {
-		return fmt.Errorf("invalid PidsLimit %q, cannot be less than 20", ctrcfg.PidsLimit)
+	if *ctrcfg.PidsLimit > 0 && *ctrcfg.PidsLimit < minPidsLimit {
```
mtrmac, line 438: What ensures the pointer is not `nil`?

mtrmac (commented):
(I don’t really understand the problem space, some comments based on reading just the local parts affected by the PR.)

pkg/apis/machineconfiguration.openshift.io/v1/types.go
kikisdeliveryservice: It's been a long day, so pardon the simple question but why the change to *int64?

QiWang19: > It's been a long day, so pardon the simple question but why the change to *int64?

Changed to *int64 to indicate whether pidsLimit is set in the containerruntimeconfig definition. Using int64 the decode of containerruntimeconfig definatoin can't differentiate if the configuration is set to 0 or there's no specific pidslimit configuration.

QiWang19: > 
> If `0` is a special case, shouldn’t it be documented? (BTW CRI-O does _not_ document 0 as a special case (but then again, runtime-spec/specs-go does document it), so it is curious that it should be supported here if it is undocumented in CRI-O.)

I have' t found the documentation in CRIO but have checked with the crio team, the value will go down to runc, 0 is valid means no limit will apply.

pkg/controller/container-runtime-config/helpers.go
QiWang19: fixed

pkg/controller/container-runtime-config/helpers.go
QiWang19: fixed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -431,14 +435,18 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 	}
 
 	ctrcfg := cfg.Spec.ContainerRuntimeConfig
-	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < minPidsLimit {
-		return fmt.Errorf("invalid PidsLimit %q, cannot be less than 20", ctrcfg.PidsLimit)
+	if ctrcfg.PidsLimit != nil && *ctrcfg.PidsLimit > 0 && *ctrcfg.PidsLimit < minPidsLimit {
```
umohnani8, line 438: This should be `>= 0` as well right?

QiWang19: Not really. if it's 0, it is valid, do not need to return the error below.

kikisdeliveryservice (commented):
@QiWang19 this is failing unit tests, so will need to fix those
`FAIL	github.com/openshift/machine-config-operator/pkg/controller/container-runtime-config `

QiWang19:
> @QiWang19 this is failing unit tests, so will need to fix those
> `FAIL github.com/openshift/machine-config-operator/pkg/controller/container-runtime-config `

Fixed. Thanks!

kikisdeliveryservice:
```
level=fatal msg=failed to fetch Worker Machines: failed to generate asset "Worker Machines": failed to fetch availability zones: failed to list zones: googleapi: Error 403: Quota exceeded for quota group 'ListGroup' and limit 'List requests per 100 seconds' of service 'compute.googleapis.com' for consumer 'project_number:...'., rateLimitExceeded
```

will report

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -313,9 +317,9 @@ func createCRIODropinFiles(cfg *mcfgv1.ContainerRuntimeConfig) []generatedConfig
 			glog.V(2).Infoln(cfg, err, "error updating user changes for log-level to crio.conf.d: %v", err)
 		}
 	}
-	if ctrcfg.PidsLimit != 0 {
+	if ctrcfg.PidsLimit != nil && *ctrcfg.PidsLimit >= 0 {
```
mtrmac, line 320: Why are negative values silently ignored here? (Or should this rely on `validateUserContainerRuntimeConfig` and not do redundant checks? The previous version didn’t check for `minPidsLimit`.)

pkg/controller/container-runtime-config/helpers.go
mtrmac, line 438: Why are negative values silently ignored here?

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -863,14 +863,18 @@ func TestRegistriesValidation(t *testing.T) {
 // TestContainerRuntimeConfigOptions tests the validity of allowed and not allowed values
 // for the options in containerruntime config
 func TestContainerRuntimeConfigOptions(t *testing.T) {
```
mtrmac, line 865: Maybe add more test cases for the new code.

pkg/apis/machineconfiguration.openshift.io/v1/types.go
mtrmac: - I’m still curious about the CRD
- So, ideally _both_ CRI-O should document `0` (not part of this PR, sure), and this repo should document `0` (a new feature introduced by this PR, so should be documented by this PR), shouldn’t they?

pkg/controller/container-runtime-config/helpers.go
umohnani8: Yup, this relies on `validateUserContainerRuntimeConfig`. It won't get to this point if the validation fails.

pkg/controller/container-runtime-config/helpers.go
umohnani8: This should fail on negative values. @haircommander iirc you were testing out something by setting this to 0 and/or -1, should -1 be a valid value?

haircommander: confusingly, negative values result in the "max" number of allowed, but zero (on my machine) returns `37725`

I'm putting together a cri-o PR to test this

haircommander: https://github.com/cri-o/cri-o/pull/4614

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -431,14 +435,18 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 	}
 
 	ctrcfg := cfg.Spec.ContainerRuntimeConfig
-	if ctrcfg.PidsLimit > 0 && ctrcfg.PidsLimit < minPidsLimit {
-		return fmt.Errorf("invalid PidsLimit %q, cannot be less than 20", ctrcfg.PidsLimit)
+	if ctrcfg.PidsLimit != nil && *ctrcfg.PidsLimit > 0 && *ctrcfg.PidsLimit < minPidsLimit {
+		return fmt.Errorf("invalid PidsLimit %q, cannot be less than 20", *ctrcfg.PidsLimit)
```
haircommander, line 439: this error message doesn't match what we're verifying anymore

QiWang19: minPidsLimit is 20. I am not sure why this value is checked. Should it be dropped?

haircommander: apparently I can't read. No that's fine :)

pkg/controller/container-runtime-config/helpers.go
umohnani8: @QiWang19 actually this can just check for `nil` as we are using a pointer now. It was checking for `0` initially as that is the unset value of an int. Leave the validation check to `validateUserContainerRuntimeConfig`.

pkg/controller/container-runtime-config/helpers.go
umohnani8: The minimum is 20 because anything less than 20 will most definitely not be enough pids for an application, so we set this limit to stop the user from hitting this issue. However, setting it to 0 sets it to the max value of pids allowed.

pkg/controller/container-runtime-config/helpers.go
QiWang19: already fixed

pkg/controller/container-runtime-config/helpers.go
mtrmac: > confusingly, negative values result in the "max" number of allowed, but zero (on my machine) returns 37725

Are these features, documented implementation details, or undocumented implementation details?

The bug originally started just with “The MCO should validate configurations and reject invalid ones”; it’s not clear to me that it requests explicitly implementing all the possible obscure values. AFAICT (but I haven’t looked into it deeply), the original reporter might have been content with just rejecting `pidsLimit: 0` as invalid, rather than explicitly supporting that (or negative values) as special cases.

(OTOH, sure, if CRI-O implements `pidsLimit: 0`, it’s plausible that the MCO should allow configuring that. But then again, OpenShift very intentionally does not expose every single feature of the underlying components. I have no idea whether the 0 and negative cases qualify as “useful OpenShift features” or “obscure implementation details OpenShift should hide”.)

pkg/controller/container-runtime-config/helpers.go
mtrmac: So should this say “if positive, cannot be less than 20”, to be pedantic?

(This is not the purpose of this PR, but given the demonstrated confusion, it would be nice to clean that up to say `"%d", minPidsLimit` instead of hard-coding the string.)

pkg/controller/container-runtime-config/helpers.go
haircommander: > Are these features, documented implementation details, or undocumented implementation details?
AFACT, undocumented implementation details. As such, I am leaning towards not allowing them here. 

One day, we may get a request to allow `max` pids_limit, and then we can decide whether it's worth adding the logic. Until then, I don't think it's worth the extra logic to allow 0 or -1

QiWang19:
@mtrmac @haircommander @umohnani8 PTAL

pkg/controller/container-runtime-config/helpers.go
umohnani8: @QiWang19 let's not change the existing behavior so we don't accidentally break someones use case. Let's leave 0 as an allowed value as it already was and we can fix the error logged for an invalid value. 
You can now control the pids limit via the kubelet, so we should open up a follow up PR to deprecate it here and document how to use kubelet instead to set it.

QiWang19: @umohnani8  Could you elaborate how to leave 0 as allowed value? Current behavior is ignoring the 0 and negative values, no error will be logged and, no new configuration will be rolled out, pidslimit will be the default value 1024. So just remove the changes of the pidslimit in this PR is enough?

umohnani8: So throw an error if `pidsLimit < 0 || (pidsLimt > 0 && pidsLimit < minPidsLimit)` This way we cover negative values and any value less than 20 that is not 0. We want 0 to be a valid value as it was.

haircommander:
/approve

LGTM

umohnani8:
/lgtm
@kikisdeliveryservice can you please approve

kikisdeliveryservice (approved):

PR #2449, opened 2021-03-02
Bug 1934113: Improve error handling for os updates

yuqi-zhang:
Add a nil check, so that if a firstboot osupdate fails, the error
is properly captured instead of panicking, since the node object
is nil.

Also just report the error directly if the command fails, since
runGetOut doesn't return stdout when it errors.


sinnykumari (approved):
/lgtm
/retest

PR #2450, opened 2021-03-02
Remove coredns from on-prem platforms

cybertron:
The only thing we have to use coredns for at this point is node
name resolution, and it turns out we don't actually need that anymore.
The internal records it provides can be just as easily provided by
dnsmasq running on the host and then we don't have to deploy and
manage an extra pod with its attendant init and monitor containers.
It also means we can remove the mdns-publisher pod.

**- What I did**

**- How to verify it**


templates/common/on-prem/units/on-prem-dnsmasq.yaml
```diff
@@ -0,0 +1,10 @@
+name: on-prem-dnsmasq.service
```
cybertron, line 1: I'm not sure we actually need a separate service for this now that it's being run as vanilla dnsmasq. I'm also not sure if there's a simple way to enable an existing OS service using MCO. I'm open to suggestions.

cybertron (commented):
/hold

This should probably wait on the outcome of https://github.com/openshift/enhancements/pull/654 as it's making a lot of the same architectural changes as that enhancement.

templates/common/on-prem/units/on-prem-dnsmasq.yaml
celebdor: I would prefet to have a separate service or a systemctl override of the default one.

PR #2452, opened 2021-03-05
Bug 1931615: OVS Config: fixes detecting bond NM files with static IP

trozet:
With static IP addressing we try to copy the original NM keyfile so that
we can modify it and retain the original IP settings. In the case where
a bond was configured with a static IP address we grep for the UUID in
the keyfiles to determine which keyfile belongs to the original
connection. However, bond slaves will also contain the UUID in their
keyfile, and the grep was accidentally returning multiple files which
causes the script to fail.

(cherry picked from commit e73bd012f14fe5025ccae35059c02d4835ecf4e8)

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -179,8 +179,8 @@ contents:
         if nmcli --fields ipv4.method,ipv6.method conn show $old_conn | grep manual; then
           echo "Static IP addressing detected on default gateway connection: ${old_conn}"
           # find and copy the old connection to get the address settings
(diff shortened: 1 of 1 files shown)


trozet:
@danwinship PTAL

yuqi-zhang (approved):
Backport of https://github.com/openshift/machine-config-operator/pull/2424, out of curiosity, why not use the bot to cherry pick?

trozet:
> Backport of #2424, out of curiosity, why not use the bot to cherry pick?

Just habit. I'll try to remember to use it in the future.

yuqi-zhang:
/hold
until cherry-pick approval. Should the bot be retesting optional tests?

PR #2376, opened 2021-02-01
Add new extension for qemu-kiwi

jensfr:
**- What I did**
This adds support for a new RHCOS extension called
"qemu-kiwi". It will install qemu-kiwi and its
dependencies, 8 RPMs with a total size of 20MB when installed.

It will allow users to run lightweight virtual machines using QEMU.

**- How to verify it**
We tested it by building MCO and building a machine-os-content image with the new RPMs included.

```oc image extract -a ~/.openshift/pull-secret  --path /:/extracted-oscontent quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ecfac365596289b8474e6a0f0bffea996e04b0bdd1379e8fb4776b8e3ef2af1b```

Then we build the image with this Dockerfile:

```
FROM quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ecfac365596289b8474e6a0f0bffea996e04b0bdd1379e8fb4776b8e3ef2af1b
RUN dnf install -y createrepo
RUN dnf install -y --downloadonly --downloaddir=/extensions/qemu-img qemu-img
RUN createrepo /extensions
ENTRYPOINT ["bin/bash"]
```

Create a new release image:

```
oc adm release new -n origin  --from-release=quay.io/openshift-release-dev/ocp-release-nightly@sha256:d1a968a86cf04a0e621c8ed265d932f365c5a3bc7977a8c0b1c6679d79f3b864  --to-image=docker.io/bpradipt/ocp-release:v4.7.1 machine-config-operator=quay.io/bpradipt/machine-config-operator:latest machine-os-content=quay.io/bpradipt/extension-test@sha256:83aae59f986b376813916807e3d19716e9df8f06b56661fce327c9e27c3e5896
```

Do an update:

```
oc adm upgrade --force --to-image  docker.io/bpradipt/ocp-release:v4.7.1 --allow-explicit-upgrade
```
Verify osImage URL points to overridden machine-os-content:
```
oc -n openshift-machine-config-operator edit configmap/machine-config-osimageurl


apiVersion: v1
data:
  osImageURL: quay.io/bpradipt/extension-test@sha256:83aae59f986b376813916807e3d19716e9df8f06b56661fce327c9e27c3e5896
  releaseVersion: 4.7.0-0.nightly-2021-01-19-083458
kind: ConfigMap
metadata:
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
[...]

PR #2456, opened 2021-03-09
Add new extension for sandboxed-containers

fidencio:
**- What I did**
This adds support for a new RHCOS extension called
"sandboxed-containres". It will install kata-containers  and its
dependencies, 9 RPMs with a total size of ~55MB when downloaded, ~150MB
when installed.

It will allow users to run kernel isolated containers via the Sandboxed
Containers operator.

**- How to verify it**
TBD

 Added support for a new RHCOS extension to install `sandboxed-containers`.

**IMPORTANT:**
Let me add a big note here that I'm experimenting with the fact of having a kata-containers extension, superseding https://github.com/openshift/machine-config-operator/pull/2376, and it comes from a suggestion made by @cgwalters while reviewing https://github.com/openshift/enhancements/pull/677#discussion_r588732079 (also, refer to: https://coreos.slack.com/archives/CK1AE4ZCK/p1614974860056400).

While modifying https://github.com/openshift/enhancements/pull/677 according to the reviewers suggestions, I'm experimenting, I'm trying to have all the pieces together, so we can make the proposal solid and based on work that's either done or close to be done, rather than basing it on what we think that should be done.

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1048,8 +1048,9 @@ func getSupportedExtensions() map[string][]string {
 	// These are RHCOS supported extensions.
 	// Each extension keeps a list of packages required to get enabled on host.
 	return map[string][]string{
-		"usbguard":     {"usbguard"},
-		"kernel-devel": {"kernel-devel", "kernel-headers"},
+		"usbguard":             {"usbguard"},
+		"kernel-devel":         {"kernel-devel", "kernel-headers"},
+		"sandboxed-containers": {"kata-containers"},
 	}
 }
 
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -259,7 +259,7 @@ func TestExtensions(t *testing.T) {
 			Config: runtime.RawExtension{
 				Raw: helpers.MarshalOrDie(ctrlcommon.NewIgnConfig()),
 			},
-			Extensions: []string{"usbguard", "kernel-devel"},
+			Extensions: []string{"usbguard", "kernel-devel", "sandboxed-containers"},
 		},
 	}
 
@@ -283,11 +283,12 @@ func TestExtensions(t *testing.T) {
 	var expectedPackages []string
 	if isOKD {
 		// OKD does not support grouped extensions yet, so installing kernel-devel will not also pull in kernel-headers
+		// "sandboxed-containers" extension is not available on OKD
 		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel")
 		expectedPackages = []string{"usbguard", "kernel-devel"}
 	} else {
-		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel", "kernel-headers")
-		expectedPackages = []string{"usbguard", "kernel-devel", "kernel-headers"}
+		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-q", "usbguard", "kernel-devel", "kernel-headers", "kata-containers")
+		expectedPackages = []string{"usbguard", "kernel-devel", "kernel-headers", "kata-containers"}
 	}
 	for _, v := range expectedPackages {
 		if !strings.Contains(installedPackages, v) {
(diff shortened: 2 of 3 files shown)


fidencio:
NOTE: As it's right now, this is a test to see whether we can easiler supersede https://github.com/openshift/machine-config-operator/pull/2376, if needed.

pkg/daemon/update.go
```diff
@@ -1102,8 +1102,9 @@ func getSupportedExtensions() map[string][]string {
 	// These are RHCOS supported extensions.
 	// Each extension keeps a list of packages required to get enabled on host.
 	return map[string][]string{
-		"usbguard":     {"usbguard"},
-		"kernel-devel": {"kernel-devel", "kernel-headers"},
+		"usbguard":             {"usbguard"},
+		"kernel-devel":         {"kernel-devel", "kernel-headers"},
+		"sandboxed-containers": {"ipxe-roms-qemu", "kata-containers", "libpmem", "pixman", "qemu-kiwi", "qemu-kvm-common", "seabios-bin", "seavgabios-bin", "sgabios-bin"},
```
sinnykumari, line 1107: You can specify here top level packages that you need to enable the functionality. You don't need to specify those package which will for sure gets pulled in as dependency, rpm-ostree takes care of resolving and downloading deps.

For example, on my regular Fedora system I want to install qemu. So, I will just do `dnf install qemu` and qemu packagers make sure that dnf pulls in deps they need for qemu to work.

test/e2e/mcd_test.go
```diff
@@ -317,9 +317,9 @@ func TestExtensions(t *testing.T) {
 
 	if isOKD {
 		// OKD does not support grouped extensions yet, so installing kernel-devel will not also pull in kernel-headers
-		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "usbguard", "kernel-devel")
+		installedPackages = execCmdOnNode(t, cs, infraNode, "chroot", "/rootfs", "rpm", "-qa", "usbguard", "kernel-devel", "ipxe-roms-qemu", "kata-containers", "libpmem", "pixman", "qemu-kiwi", "qemu-kvm-common", "seabios-bin", "seavgabios-bin", "sgabios-bin")
```
sinnykumari, line 320: don't check for packages here as we are not enabling the extension on OKD

pkg/daemon/update.go
fidencio: Oh, that's lovely to know. So, we just need to ensure we have the correct repos enabled on RHCOS side.

Thanks a lot for the review, @sinnykumari!

test/e2e/mcd_test.go
fidencio: Ack!

pkg/daemon/update.go
fidencio: Ah, @sinnykumari, down below in this PR, we do the check about the expected and installed packages.
Would there make sense to have all the deps listed? Or are we good to go just with`kata-containers`?

fidencio:
I've updated the PR according to @sinnykumari's review. (thanks!)

pkg/daemon/update.go
sinnykumari: If `kata-containers` package ensures that it will install deps, then yes just verifying for kata-containers should be enough.

sinnykumari: > Oh, that's lovely to know. So, we just need to ensure we have the correct repos enabled on RHCOS side.

yes, as per https://github.com/openshift/enhancements/pull/677/files#diff-3baabdf22a2dfc2542b74496b417654de8a95693f9d1bb09605c066afb790ed1R471 I think enabling required repos and specifying just `kata-containers` for sandboxed-containers` extension in https://github.com/openshift/os/blob/master/extensions.yaml should be enough.

fidencio:
@sinnykumari, as this one supersedes https://github.com/openshift/machine-config-operator/pull/2376, can we have the 4.8 label added to this one?

@jensfr, can we have https://github.com/openshift/machine-config-operator/pull/2376 closed?

jensfr:
> @jensfr, can we have #2376 closed?

done

sinnykumari:
> @sinnykumari, as this one supersedes #2376, can we have the 4.8 label added to this one?

added 4.8 label. Once you have machine-os-content updated with correct set of packages, let us know in the PR.

fidencio:
This PR has been updated, and depends on https://github.com/openshift/os/pull/525 (which depends on a downstream PR).

fidencio:
@sinnykumari, I guess this one is ready to get another round of review and, hopefully, get merged at some point soon.
/cc @yuqi-zhang, who also reviewed https://github.com/openshift/machine-config-operator/pull/2376.

sinnykumari (approved):
looks clean, thanks for the PR!
/lgtm

JAORMX (approved):

PR #2457, opened 2021-03-09
Skip drain on Single Node deployment

sinnykumari:
With the introduction of different ControlPlaneTopology types
in the OpenShift cluster, the behaviour of the cluster may differ
based on cluster type. For example: cluster with Single
controlPlane node, it doesn't make sense to perform workload drain.

ControllerConfig now understands ControlPlaneTopology:TopologyMode
set in the cluster. Node controller later on reads the value from
controllerConfig and sets value into node annotation
`machineconfiguration.openshift.io/controlPlaneTopology`.

While performing a configuration update, machine-config-daemon
will read the annotation and based on controlPlaneTopology
type, it will decide drain action.
MCD skips drain if controlPlaneTopology is SingleReplica.

Part of - https://github.com/openshift/enhancements/pull/560/

This PR also:
- refactors drain logic
- adds related unit tests

### DIFF
--- a/manifests/controllerconfig.crd.yaml
+++ b/manifests/controllerconfig.crd.yaml
@@ -175,6 +175,14 @@ spec:
                         like the web console to tell users where to find the Kubernetes
                         API.
                       type: string
+                    controlPlaneTopology:
+                      description: controlPlaneTopology expresses the expectations for
+                        operands that normally run on control nodes. The default is
+                        HighlyAvailable, which represents the behavior operators have
+                        in a normal cluster. The SingleReplica mode will be used in
+                        single-node deployments and the operators should not configure
+                        the operand for highly-available operation.
+                      type: string
                     etcdDiscoveryDomain:
                       description: 'etcdDiscoveryDomain is the domain used to fetch
                         the SRV records for discovering etcd servers and clients.
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -80,9 +80,11 @@ type Controller struct {
 	syncHandler              func(mcp string) error
 	enqueueMachineConfigPool func(*mcfgv1.MachineConfigPool)
 
+	ccLister   mcfglistersv1.ControllerConfigLister
 	mcpLister  mcfglistersv1.MachineConfigPoolLister
 	nodeLister corelisterv1.NodeLister
 
+	ccListerSynced   cache.InformerSynced
 	mcpListerSynced  cache.InformerSynced
 	nodeListerSynced cache.InformerSynced
 
@@ -94,6 +96,7 @@ type Controller struct {
 
 // New returns a new node controller.
 func New(
+	ccInformer mcfginformersv1.ControllerConfigInformer,
 	mcpInformer mcfginformersv1.MachineConfigPoolInformer,
 	nodeInformer coreinformersv1.NodeInformer,
 	schedulerInformer cligoinformersv1.SchedulerInformer,
@@ -129,8 +132,10 @@ func New(
 	ctrl.syncHandler = ctrl.syncMachineConfigPool
 	ctrl.enqueueMachineConfigPool = ctrl.enqueueDefault
 
+	ctrl.ccLister = ccInformer.Lister()
 	ctrl.mcpLister = mcpInformer.Lister()
 	ctrl.nodeLister = nodeInformer.Lister()
+	ctrl.ccListerSynced = ccInformer.Informer().HasSynced
 	ctrl.mcpListerSynced = mcpInformer.Informer().HasSynced
 	ctrl.nodeListerSynced = nodeInformer.Informer().HasSynced
 
@@ -145,7 +150,7 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer ctrl.queue.ShutDown()
 
-	if !cache.WaitForCacheSync(stopCh, ctrl.mcpListerSynced, ctrl.nodeListerSynced, ctrl.schedulerListerSynced) {
+	if !cache.WaitForCacheSync(stopCh, ctrl.ccListerSynced, ctrl.mcpListerSynced, ctrl.nodeListerSynced, ctrl.schedulerListerSynced) {
 		return
 	}
 
@@ -758,6 +763,10 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		return err
 	}
 
+	if err := ctrl.setClusterConfigAnnotation(nodes); err != nil {
+		return goerrs.Wrapf(err, "error setting clusterConfig Annotation for node in pool %q, error: %v", pool.Name, err)
+	}
+
 	candidates, capacity := getAllCandidateMachines(pool, nodes, maxunavail)
 	if len(candidates) > 0 {
 		ctrl.logPool(pool, "%d candidate nodes for update, capacity: %d", len(candidates), capacity)
@@ -800,6 +809,30 @@ func (ctrl *Controller) getNodesForPool(pool *mcfgv1.MachineConfigPool) ([]*core
 	return nodes, nil
 }
 
+// setClusterConfigAnnotation reads cluster configs set into controllerConfig
+// and add/updates required annotation to node such as ControlPlaneTopology
+// from infrastrcture object.
+func (ctrl *Controller) setClusterConfigAnnotation(nodes []*corev1.Node) error {
+	cc, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodes {
+		glog.Infof("Initiating controlPlaneTopology annotation %s\n", string(cc.Spec.Infra.Status.ControlPlaneTopology))
+		if node.Annotations[daemonconsts.ClusterControlPlaneTopologyAnnotationKey] != string(cc.Spec.Infra.Status.ControlPlaneTopology) {
+			_, err := internal.UpdateNodeRetry(ctrl.kubeClient.CoreV1().Nodes(), ctrl.nodeLister, node.Name, func(node *corev1.Node) {
+				node.Annotations[daemonconsts.ClusterControlPlaneTopologyAnnotationKey] = string(cc.Spec.Infra.Status.ControlPlaneTopology)
+				glog.Infof("Updated controlPlaneTopology annotations\n")
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func (ctrl *Controller) setDesiredMachineConfigAnnotation(nodeName, currentConfig string) error {
 	return clientretry.RetryOnConflict(nodeUpdateBackoff, func() error {
 		oldNode, err := ctrl.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
@@ -815,6 +848,7 @@ func (ctrl *Controller) setDesiredMachineConfigAnnotation(nodeName, currentConfi
 		if newNode.Annotations == nil {
 			newNode.Annotations = map[string]string{}
 		}
+
 		if newNode.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey] == currentConfig {
 			return nil
 		}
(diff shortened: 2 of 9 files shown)


pkg/daemon/daemon.go
```diff
@@ -1283,7 +1284,7 @@ func (dn *Daemon) prepUpdateFromCluster() (*mcfgv1.MachineConfig, *mcfgv1.Machin
 // "transient state" file, which signifies that all of those prior steps have
 // been completed.
 func (dn *Daemon) completeUpdate(node *corev1.Node, desiredConfigName string) error {
-	if err := drain.RunCordonOrUncordon(dn.drainer, node, false); err != nil {
+	if err := dn.cordonOrUncordonNode(false); err != nil {
```
yuqi-zhang, line 1287: Since you use the `dn.Node` reference instead, the linter is complaining the `node` variable isn't used anymore here

sinnykumari: yeah, fixed it

sinnykumari:
unit test seems to be failing due to unrelated test
/retest

yuqi-zhang (commented):
Overall looks good! Just some minor questions/nits.

Also, let's say a user has a default 3-3 cluster. If they set that controllerconfig spec, would they be able to skip drains on their updates? Or does something overwrite that?

One last question: I wonder if `ensureControllerConfigSpec` function needs to be updated or not

pkg/controller/node/node_controller.go
```diff
@@ -800,6 +809,30 @@ func (ctrl *Controller) getNodesForPool(pool *mcfgv1.MachineConfigPool) ([]*core
...
+// from infrastrcture object.
+func (ctrl *Controller) setClusterConfigAnnotation(nodes []*corev1.Node) error {
+	cc, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodes {
+		glog.Infof("Initiating controlPlaneTopology annotation %s\n", string(cc.Spec.Infra.Status.ControlPlaneTopology))
+		if node.Annotations[daemonconsts.ClusterControlPlaneTopologyAnnotationKey] != string(cc.Spec.Infra.Status.ControlPlaneTopology) {
```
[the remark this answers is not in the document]
sinnykumari: It will always get set and would have default value HighlyAvailable, see https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml#L190

pkg/daemon/drain.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (dn *Daemon) performDrain() error {
+	// Skip drain process when we're not cluster driven
+	if dn.kubeClient == nil {
+		return nil
+	}
+
+	if !dn.drainRequired() {
+		if isSingleNodeTopology(dn.getControlPlaneTopology()) {
```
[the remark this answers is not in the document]
sinnykumari: right, was refactoring additional logic into drainRequired() but it didn't go well. missed to restructure code later on. Fixed.

pkg/daemon/daemon.go
```diff
@@ -1617,3 +1618,23 @@ func (dn *Daemon) senseAndLoadOnceFrom(onceFrom string) (interface{}, onceFromOr
...
 	return nil, onceFromUnknownConfig, fmt.Errorf("unable to decipher onceFrom config type: %v", err)
 }
+
+func isSingleNodeTopology(topology configv1.TopologyMode) bool {
+	return topology == configv1.SingleReplicaTopologyMode
+}
+
+// getControlPlaneTopology reads from node annotation and returns
+// controlPlaneTopology value set in the cluster. If annotation value
+// is unrecognized, we considere it as highly available cluster.
```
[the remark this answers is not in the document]
sinnykumari: fixed

manifests/controllerconfig.crd.yaml
```diff
@@ -175,6 +175,14 @@ spec:
                         like the web console to tell users where to find the Kubernetes
                         API.
                       type: string
+                    controlPlaneTopology:
```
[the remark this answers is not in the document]
sinnykumari: syncMachineConfigController() calls [resourceapply.ApplyControllerConfig](https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L405) that calls [resourcemerge.EnsureControllerConfig](https://github.com/openshift/machine-config-operator/blob/82868e63176fee2bc806c1deb308ed1fc8965d84/lib/resourcemerge/machineconfig.go#L17) which calls ensureControllerConfigSpec() and [here]([https://github.com/openshift/machine-config-operator/blob/82868e63176fee2bc806c1deb308ed1fc8965d84/lib/resourcemerge/machineconfig.go#L81]) we check if Infra object has been modified and update controllerconfig if changed  . Infra object is the one where `controlPlaneTopology` value exist as well along with other data https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L86 . MCO was already reading and updating Infra content, so adding controlPlaneToplogy field in crd populates its value as well.

sinnykumari:
> Overall looks good! Just some minor questions/nits.
> 
> Also, let's say a user has a default 3-3 cluster. If they set that controllerconfig spec, would they be able to skip drains on their updates? Or does something overwrite that?

No, MCO operator pod keeps resyncing value of controllerConfigspec and we read data from infrastructure cluster object, so in next sync it will revert the value to whatever is set cluster wide.

> One last question: I wonder if `ensureControllerConfigSpec` function needs to be updated or not

not needed, as we already do that at https://github.com/openshift/machine-config-operator/blob/82868e63176fee2bc806c1deb308ed1fc8965d84/lib/resourcemerge/machineconfig.go#L81 .

sinnykumari:
log from a SNO cluster created using clusterbot and applied a MachineConfig:
```
I0310 10:38:15.014517    8164 update.go:543] Checking Reconcilable for config rendered-master-dd9d479a34df41e1ba15683557aac98f to rendered-master-032640efd4eaabc7a384aa49328dc323
I0310 10:38:15.115007    8164 update.go:1851] Starting update from rendered-master-dd9d479a34df41e1ba15683557aac98f to rendered-master-032640efd4eaabc7a384aa49328dc323: &{osUpdate:false kargs:false fips:false passwd:false files:false units:false kernelType:false extensions:true}
I0310 10:38:15.149990    8164 update.go:1851] Node has been successfully cordoned
I0310 10:38:15.158087    8164 update.go:1851] Drain not required, skipping
I0310 10:38:15.163018    8164 update.go:1166] Updating files

```

manifests/controllerconfig.crd.yaml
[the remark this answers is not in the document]
yuqi-zhang: Ahh, ok thanks for the explanation! Right it gets synced as part of the infra field. Not the clearest of code paths :facepalm:

sinnykumari:
gcp-op failed due to unrelated reason, retesting
/test e2e-gcp-op

pkg/daemon/daemon.go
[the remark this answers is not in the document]
crawford: Super minor nit: the fix ended up in a later commit (1261271d).

sinnykumari:
Talked to SNO team, aws-single-node test is good from MCO side. Failing sub-tests should pass with https://github.com/openshift/origin/pull/25936 .

pkg/controller/node/node_controller_test.go
```diff
@@ -904,6 +942,46 @@ func TestShouldDoNothing(t *testing.T) {
 	f.run(getKey(mcp, t))
 }
 
+func TestControlPlaneTopology(t *testing.T) {
```
yuqi-zhang, line 945: Sorry, I'm a bit confused by what this is trying to test. Is it checking to see if the controllerconfig topology propagates to node annotations? Where is that being checked?

pkg/daemon/drain.go
```diff
@@ -0,0 +1,113 @@
...
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", err.Error())
+		return errors.Wrap(err, "failed to drain node")
+	}
+
+	return nil
+}
+
+func (dn *Daemon) performDrain() error {
+	// Skip drain process when we're not cluster driven
+	if dn.kubeClient == nil {
```
yuqi-zhang, line 80: I know this is from our old code, but just curious, do you know if we would ever hit this condition?

pkg/daemon/drain.go
```diff
@@ -0,0 +1,113 @@
...
+	dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Cordon", "Cordoned node to apply update")
+
+	if !dn.drainRequired() {
+		dn.logSystem("Drain not required, skipping")
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Drain", "Drain not required, skipping")
+		return nil
+	}
+
+	// We are here, that means we need to cordon and drain node
+	MCDDrainErr.WithLabelValues(dn.node.Name, "").Set(0)
```
yuqi-zhang, line 97: Hmm, why do we set this here again? Is it to clear a previously failing drain's error if that one hit the global timeout?

pkg/daemon/drain.go
```diff
@@ -0,0 +1,113 @@
...
+	// Skip drain process when we're not cluster driven
+	if dn.kubeClient == nil {
+		return nil
+	}
+
+	if err := dn.cordonOrUncordonNode(true); err != nil {
+		return err
+	}
+	dn.logSystem("Node has been successfully cordoned")
+	dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Cordon", "Cordoned node to apply update")
```
yuqi-zhang, line 88: Very minor: this event seems a bit unnecessary, since either the "Drain not required" or "Draining node" event would fire immediately after this, which implies we cordoned the node right? I'm ok with keeping it too but was wondering if this would be useful to check for anything

yuqi-zhang (approved):
Code is looking good! Just a few more minor questions/nits. Approving :tada: 

Will try to do some manual testing as well as give others a chance to provide feedback before LGTM

pkg/daemon/drain.go
sinnykumari: yeah, we do hit it when update() gets called during firstboot as firstboot is not cluster driven.

My initial goal was to move this logic into drainRequired() but it was complicating some logic because during firstboot dn.node is null and it causes to crash things.

pkg/daemon/drain.go
sinnykumari: maybe, I am not sure. I kept logic from earlier implementation. @kikisdeliveryservice would know better.

pkg/daemon/drain.go
sinnykumari: That's right. I don't have strong opinion here. I added thinking drain and cordon are two different actions. Does adding extra event impact our logging system?

pkg/controller/node/node_controller_test.go
sinnykumari: Here, we are currently checking that with a valid controlPlaneToplogy i.e. `SingleReplica` ,  setClusterConfigAnnotation() gets called and works as expected. Since, in this unit test we have created controllerConfig with `SingleReplica` and we have set node annotation `machineconfiguration.openshift.io/controlPlaneTopology` to `SingleReplica` , expected result is no node change action such as `patch`.

Later on, I am thinking of extending the test to also perform updating node annotation update action. Adding this is a bit tricky.

pkg/daemon/drain.go
yuqi-zhang: Not really, probably fine as is :+1:

pkg/controller/node/node_controller_test.go
yuqi-zhang: Hmm, sorry for the dumb question, but I'm still not sure how we are testing that. If an action change happens as a bug, which line of code would return the error? I see we do the same for the above `TestShouldDoNothing` but I don't see any expectations for "nothing"

yuqi-zhang:
Tried to do some manual testing but failed, will try again next week and LGTM if nobody else has any concerns

pkg/controller/node/node_controller_test.go
sinnykumari: when in test run() calls runController(), node-controller syncHandler(pool) gets called at https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller_test.go#L114 . As you know when this runs, various actions can be generated based on what all operation on node has been performed. In test we filter out list and watch actionhttps://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller_test.go#L121as these actions doesn't really update anything. For other actions like patch, additional action will be check in later line during checkAction()

yuqi-zhang:
I think this looks good! Let's get this in, and iterate from there if there are any issues.
/lgtm

sinnykumari:
skipping optional test from retest
/skip

PR #2458, opened 2021-03-10
Bug 1825417: Make getting the suffix of an MC more robust

umohnani8:
Fixes #1825417 (https://bugzilla.redhat.com/show_bug.cgi?id=1825417)

**- What I did**
Instead of relying on the order in which the ctrcfg
objects are added to the list, let's go through the
list and find the maximum suffix value used. We can
then add 1 to the max suffix value to be used for the
MC when creating the new ctrcfg object.

**- How to verify it**
Create multiple ctrcfg CRs and a new MC with a different suffix will be created for each one.

### DIFF
--- a/pkg/controller/container-runtime-config/helpers.go
+++ b/pkg/controller/container-runtime-config/helpers.go
@@ -188,7 +188,7 @@ func getManagedKeyCtrCfg(pool *mcfgv1.MachineConfigPool, client mcfgclientset.In
 		return "", fmt.Errorf("error listing container runtime configs: %v", err)
 	}
 	// If there is no ctrcfg in the list, return the default MC name with no suffix
-	if ctrcfgList == nil || ctrcfgList.Items == nil {
+	if ctrcfgList == nil || len(ctrcfgList.Items) == 0 {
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
(Warning: based only on the diff, without fully understanding the operators.)

- Presumably `getManagedKubeletConfigKey` needs to be updated as well.
- This is Nth case of such duplication causing risk of not fixing the same bug everywhere. Wouldn’t it be well worth it to deduplicate the logic, by now? 
- It would be nice to have tests — perhaps split the `client.` API accesses from the underlying logic, and write unit tests for the logic.

umohnani8:
There are 2 separate bugs so opened separate PRs, the kubelet config fix is in https://github.com/openshift/machine-config-operator/pull/2459

yuqi-zhang (approved):
The code itself seems fine to me. Just curious, I recall trying to reproduce this bug with you, what exactly was the reproduction condition?

umohnani8:
@yuqi-zhang I wasn't able to always reproduce the issue, but when I did I noticed that the new ctrcfg was being added to the beginning of the items list instead of being appended to it. That was causing issues when were relying on the order to be newest at the end of the items list. So fixed it up to just look through the list and find the max suffix instead of checking the most recent ctrcfg that was created.

PR #2459, opened 2021-03-10
Bug 1840881: Make getting the suffix of a kubeletconfig MC more robust

umohnani8:
Fixes #1840881 (https://bugzilla.redhat.com/show_bug.cgi?id=1840881)

**- What I did**
Instead of relying on the order in which the kubeletconfig
objects are added to the list, let's go through the
list and find the maximum suffix value used. We can
then add 1 to the max suffix value to be used for the
MC when creating the new kubeletconfig object.

**- How to verify it**
Create multiple kubelet config CRs and each one will have a different MC for it.


umohnani8:
The test failures look unrelated, can we skip them? @kikisdeliveryservice

kikisdeliveryservice (approved):
Looks good, like the new check.

PR #2460, opened 2021-03-11
Bug 1948022: Add on-prem namespaces to relatedObjects

cybertron:
Previously the kni namespace was explicitly listed in must-gather,
but when I went to add the other on-prem ones it was pointed out
that this should probably be done with relatedObjects. This adds
the namespaces here so we don't need anything in must-gather itself.

Added on-prem service namespaces to relatedObjects so they will be included in must-gather.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -72,6 +72,11 @@ func (optr *Operator) syncRelatedObjects() error {
 		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigs"},
 		// gathered because the machineconfigs created container bootstrap credentials and node configuration that gets reflected via the API and is needed for debugging
 		{Group: "", Resource: "nodes"},
+		// Gathered for the on-prem services running in static pods.
+		{Resource: "namespaces", Name: "openshift-kni-infra"},
+		{Resource: "namespaces", Name: "openshift-openstack-infra"},
+		{Resource: "namespaces", Name: "openshift-ovirt-infra"},
+		{Resource: "namespaces", Name: "openshift-vsphere-infra"},
 	}
 
 	if !equality.Semantic.DeepEqual(coCopy.Status.RelatedObjects, co.Status.RelatedObjects) {


cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

As requested on the must-gather review doing the same.

mandre (commented):
I'm a bit unfamiliar with this part of the code, but LGTM on the principle.

So, once this merge, you could update https://github.com/openshift/must-gather/pull/209/ to remove the `openshift-kni-infra` namespace from the named resource collection?

cybertron:
> So, once this merge, you could update [openshift/must-gather#209](https://github.com/openshift/must-gather/pull/209) to remove the `openshift-kni-infra` namespace from the named resource collection?

Yes, we would no longer need anything specific to the on-prem platforms in must-gather. The only reason I initially added the kni namespace there was that I didn't know about this. :-)

cybertron:
/retest

Note that although the vsphere job failed, it did capture logs this time so it confirms that this is working.

pkg/operator/status.go
```diff
@@ -72,6 +72,10 @@ func (optr *Operator) syncRelatedObjects() error {
 		{Group: "machineconfiguration.openshift.io", Resource: "machineconfigs"},
 		// gathered because the machineconfigs created container bootstrap credentials and node configuration that gets reflected via the API and is needed for debugging
 		{Group: "", Resource: "nodes"},
+		{Resource: "namespaces", Name: "openshift-kni-infra"},
```
kikisdeliveryservice, line 75: could you add a comment above the new additions just to memorialize what this is for?

cybertron: Done

cybertron:
Both of the remaining failed jobs look pretty red in general. They don't seem to be related to this patch (not sure how they could be).

cybertron:
@Gal-Zaidman @mandre Can I get another lgtm on this? I lost it when I added the comment. Thanks.

mandre (commented):
/lgtm

kikisdeliveryservice (approved):
Double-checked and this seems to be working correctly.

mandre:
/retest
Why isn't prow restarting e2e-agnostic-upgrade?

kikisdeliveryservice:
@cybertron Do you still want this backported?

cybertron:
Definitely. It's a real hassle debugging on-prem network issues without it. Let's see if I remember the cherry-pick syntax...

/cherry-pick release-4.7

PR #2461, opened 2021-03-15
apiserver-watcher support dual-stack

aojea:
The apiserver watchers validates that the url passed as argument
only resolves to one IP address, but this is a  problem on dual
stack environments, that must be legit to resolve to one address
per IP family.

The apiserver-watcher must check that the restriction only applies
per IP family instead.


aojea:
I don't know the procedures here, just opened it if is not ok to merge as part of https://github.com/openshift/machine-config-operator/pull/2422

PR #2463, opened 2021-03-15
Bug 1939270: pkg/operator/status: Use 'DegradedPool' reason for Upgradeable=False

wking:
c6e2fa8674 (#2231) added this check, but [`Reason` is supposed to be a machine-readable slug, while the human-oriented string goes in `Message`][1].  Possibly worth a backport to 4.7?

[1]: https://github.com/openshift/api/blob/4b79815405ec40f1d72c3a74bae0ae7da543e435/config/v1/types_cluster_operator.go#L128-L136


kikisdeliveryservice:
good catch! let me attach a bz to this to backport.

/hold

kikisdeliveryservice (approved):
thanks @wking for catching this

/lgtm

PR #2465, opened 2021-03-17
Bug 1946506: Add hosts template for monitor-based resolution

cybertron:
Now that node resolution is not required for bootstrapping, we no
longer need to use mdns to build the list of nodes. We can just wait
until the api is up and retrieve it from there.

This change removes the mdns plugin from the coredns config and
replaces it with a hosts plugin template which will be populated
by the coredns-monitor.

Removed mdns node resolution in favor of a kubernetes api-based implementation.


cybertron:
This is dependent on https://github.com/openshift/baremetal-runtimecfg/pull/129. I suspect it may not pass ci until that merges.

cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

metal-ipi is currently broken, but the runtimecfg patch merged so this should be good to go whenever it starts working again.

Gal-Zaidman:
/lgtm

oVirt seems to be ok with the change :)

kikisdeliveryservice:
Ping:  @mandre  @jcpowermac @patrickdillon

jcpowermac:
Reviewing the e2e-vsphere logs the test failures are unrelated.

Upgrade is failing at MCO which might indicate a problem - I will look more in depth tomorrow.

kikisdeliveryservice:
> Reviewing the e2e-vsphere logs the test failures are unrelated.
> 
> Upgrade is failing at MCO which might indicate a problem - I will look more in depth tomorrow.

MCD having trouble with image registry pod: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2465/pull-ci-openshift-machine-config-operator-master-e2e-vsphere-upgrade/1377258688363892736/artifacts/e2e-vsphere-upgrade/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-daemon-nzw6c_machine-config-daemon.log

It's covered by: https://bugzilla.redhat.com/show_bug.cgi?id=1944762 

It's hitting some jobs incl vsphere

templates/common/on-prem/files/coredns-corefile.yaml
```diff
@@ -5,7 +5,6 @@ contents:
     . {
         errors
         health :18080
-        mdns {{ .DNS.Spec.BaseDomain }} 0 {{`{{.Cluster.Name}}`}} {{`{{.NonVirtualIP}}`}}
```
mandre, line 8: Seems like we no longer need the rule for mdns in the security groups and it should be removed from the installer. Are we tracking this somewhere?

templates/common/on-prem/files/coredns.yaml
```diff
@@ -19,7 +19,7 @@ contents:
           path: "/etc/kubernetes/static-pod-resources/coredns"
       - name: kubeconfig
         hostPath:
-          path: "/etc/kubernetes/kubeconfig"
```
mandre, line 22: Are the two changes of updating the path of the kubeconfig file and removing the mdns plugin related? If not they should probably go in separate PRs or at least separate commits.

templates/common/on-prem/files/coredns-corefile.yaml
cybertron: Ah, no I hadn't considered that because baremetal doesn't have security groups. What would be the best way to make sure that doesn't get lost?

templates/common/on-prem/files/coredns.yaml
cybertron: Yes, in order for the coredns-monitor to read the node list and populate this new template we need the kubelet kubeconfig. The old one doesn't have permission to read nodes.

templates/common/on-prem/files/coredns-corefile.yaml
mandre: Normally, I would say in the same BZ that tracks the removal of mdns, or just a patch in the installer, but with only 3 days to feature freeze I think it's unrealistic to expect both patch to merge before FF. I'll create the BZ.

mandre (commented):
/retest
/lgtm

mandre:
/retitle Bug 1946506: Add hosts template for monitor-based resolution

templates/common/on-prem/files/coredns-corefile.yaml
mandre: We're now tracking mdns removal in https://bugzilla.redhat.com/show_bug.cgi?id=1946506. Aside from allowing to merge past feature freeze, this also ensures QE looks at and verifies the code change.

mandre: Do we still need mdns-publisher?

cybertron: As it stands right now, we do. My thought was that after this and https://github.com/openshift/machine-config-operator/pull/2410 merged we could remove mdns-publisher completely, but right now some platforms are still going to be using mdns.

Given how close we are to feature freeze now, maybe I need to pull 2410 into this PR and do everything at once.

staebler:
Coming from https://github.com/openshift/installer/pull/4821, I was expecting to see code in this PR that removed the `--mdns-publisher-image` flag.

cybertron:
I pulled in the de-dupe commit from #2410 and removed mdns-publisher as well. This should be all of the coredns changes we need to get in for 4.8. We'll need to clean up the mdns-publisher image bits @staebler  mentioned, but that will have to happen in a separate PR.

cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

Once more unto the breach!

cybertron:
/test e2e-metal-ipi

Looks like the cluster deployed fine but we have some flaky tests on metal. I will fire off a local deployment too just in case I somehow broke just the metal job when I consolidated everything.

mandre:
> /test e2e-openstack
> /test e2e-ovirt
> /test e2e-vsphere
> 
> Once more unto the breach!

Strange, this didn't trigger the jobs.

/retest
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

mandre:
The bootstrap [still uses mdns](https://github.com/openshift/machine-config-operator/blob/5f55012609c9fa7b6548b98e12dc5848bee7edfa/manifests/on-prem/coredns-corefile.tmpl#L4), is it expected?

cybertron:
> The bootstrap [still uses mdns](https://github.com/openshift/machine-config-operator/blob/5f55012609c9fa7b6548b98e12dc5848bee7edfa/manifests/on-prem/coredns-corefile.tmpl#L4), is it expected?

Nope, that was an oversight on my part. I'll fix it.

cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

Maybe Prow is not a fan of Shakespeare? ;-)

cybertron:
/retest

This patch seems to be working fine in the metal job. Coredns looks happy, so we must be failing on other issues.

cybertron:
/retest

CI hates this, but AFAICT it's working as intended. The failures aren't related from what I can see (although I have no logs for the non-metal platforms so I could be wrong).

cybertron:
/retest

Hopefully jobs will be more stable now, and we should get logs if they do fail.

cybertron:
I think the non-metal failures may be legitimate here. It looks like I need to add the fallthrough option to the hosts plugin or it eats requests. I didn't run into this on metal because we don't access external addresses from the cluster, but platforms that need to reach the cloud provider can't get out.

jcpowermac:
@cybertron for vsphere I was seeing DNS resolution failing for vCenter. That is also a problem with WMCO currently

cc: @aravindhp

cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

Argh, forgot to re-add the optional jobs again. >:-(

cybertron:
> @cybertron for vsphere I was seeing DNS resolution failing for vCenter. That is also a problem with WMCO currently
> 
> cc: @aravindhp

Okay thanks, I'll keep that in mind, but all of the platforms were seeing similar errors with this patch. I guess we'll see.

cybertron:
/test e2e-metal-ipi

403 downloading go. o.O

cybertron:
/test e2e-aws

I think the vsphere-upgrade and okd jobs are known flaky, but it would be nice to get an e2e-aws pass.

cybertron:
/test e2e-aws

I'm seeing the same errors in other PRs which eventually merged, so I don't think it's related to this one.

cybertron:
Okay, this is now passing all of the expected jobs. The only change since the previous lgtms was adding fallthrough to fix external DNS resolution.

cybertron:
@mandre @Gal-Zaidman @jcpowermac Can someone lgtm this again? It should be good to go now, and this fixes a pretty big scaling issue with on-prem. Thanks!

JAORMX (approved):
/retest

PR #2467, opened 2021-03-18
Bug 1883916: [4.6z backport] Fixes setting route metric for ovs-config

trozet:
ovs-configuration: use NM default ethernet route metric
    
    Instead of the default OVS interface route metric, which is much lower,
    and causes problems when secondary interfaces are added to the machine.


trozet:
@yuqi-zhang PTAL

yuqi-zhang (approved):

PR #2470, opened 2021-03-19
Bug 1940939: Do "systemctl daemon-reload" after running "runtimecfg node-ip"

danwinship:
I have not managed to reproduce this bug, but the circumstantial evidence is strong.

we write out 20-nodenet.conf at 11:08:14:

```
Mar 17 11:08:14 ...-infra-0-qx58v bash[1716]: time="2021-03-17T11:08:14Z" level=info msg="Writing Kubelet service override with content [Service]\nEnvironment=\"KUBELET_NODE_IP=192.168.48.21\" \"KUBELET_NODE_IPS=192.168.48.21\"\n"
```

then start kubelet 2 seconds later:
```
Mar 17 11:08:16 ...-infra-0-qx58v systemd[1]: Starting Kubernetes Kubelet...
```

but kubelet doesn't see `${KUBELET_NODE_IP}`:
```
    root        2113 11.7  1.0 3212924 168416 ?      Ssl  Mar17 184:29 kubelet --config=/etc/kubernetes/kubelet.conf --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig --kubeconfig=/var/lib/kubelet/kubeconfig --container-runtime=remote --container-runtime-endpoint=/var/run/crio/crio.sock --runtime-cgroups=/system.slice/crio.service --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_id=rhcos --node-ip= --address= ...
```

There are some instances of

    Mar 17 08:09:01 ...-infra-0-qx58v sh[209732]: Warning: The unit file, source configuration file or drop-ins of kubelet.service changed on disk. Run 'systemctl daemon-reload' to reload units.

in the logs, though none at the time kubelet is started. (Maybe it only gets logged if something that actually runs "systemctl _something_ kubelet" happens to have its output get logged?)

/hold
still testing
cc @celebdor

### DIFF
--- a/templates/common/_base/units/nodeip-configuration.service.yaml
+++ b/templates/common/_base/units/nodeip-configuration.service.yaml
@@ -25,6 +25,7 @@ contents: |
     do \
     sleep 5; \
     done"
+  ExecStart=/bin/systemctl daemon-reload
 
   {{if .Proxy -}}
   EnvironmentFile=/etc/mco/proxy.env
--- a/templates/common/on-prem/units/nodeip-configuration.service.yaml
+++ b/templates/common/on-prem/units/nodeip-configuration.service.yaml
@@ -29,6 +29,7 @@ contents: |
     do \
     sleep 5; \
     done"
+  ExecStart=/bin/systemctl daemon-reload
 
   {{if .Proxy -}}
   EnvironmentFile=/etc/mco/proxy.env


danwinship:
Toni points out that files read by `EnvironmentFile` don't have the "needs daemon-reload" constraint like drop-in files do, so an alternative approach would be to write the file somewhere else and then read it in via `EnvironmentFile` rather than writing it to `kubelet.service`'s drop-in dir and having it be read automatically. (Assuming we actually care about the "write it somewhere else" part, that would require a runtimecfg update too, although I have one of those coming anyway for other node-ip bugs...)

danwinship:
/hold cancel

ok, there doesn't seem to be any way to get systemd to tell you _why_ a unit needs a daemon-reload, but I've confirmed that on a running bare-metal system:
```bash
# systemctl show --property=NeedDaemonReload kubelet
NeedDaemonReload=yes
```
whereas on an Azure system, for comparison:
```bash
# systemctl show --property=NeedDaemonReload kubelet
NeedDaemonReload=no
```

So we're doing _something_ wrong on bare metal, and it seems likely that it's this.

templates/common/_base/units/nodeip-configuration.service.yaml
```diff
@@ -25,6 +25,7 @@ contents: |
     do \
     sleep 5; \
     done"
+  ExecStart=/bin/systemctl daemon-reload
```
cybertron, line 28: TIL that oneshots can have multiple ExecStarts. :-)

cybertron (approved):
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere
/test e2e-vsphere-upi

Looks fine to me, just testing all the affected on-prem platforms.

mandre (commented):
/lgtm

danwinship:
/assign @sinnykumari
for approval because the bot said so

sinnykumari:
test from affected platform needs to be green
/test e2e-vsphere

sinnykumari:
vsphere test is failing on the PR
@jcpowermac please review the PR from vsphere side

danwinship:
@sinnykumari the e2e-vsphere failures must be flakes; the test seems flaky in general, and neither of the files modified by this PR would be used in that environment. (The one in `templates/common/_base` would be installed, but it's `enabled: {{if eq .Infra.Status.PlatformStatus.Type "None"}}true{{else}}false{{end}}`, and e2e-vsphere uses the vsphere platform.)

jcpowermac:
> @sinnykumari the e2e-vsphere failures must be flakes; the test seems flaky in general, 

Depends on what is going on the vSphere CI cluster at the time.

> and neither of the files modified by this PR would be used in that environment. (The one in `templates/common/_base` would be installed, but it's `enabled: {{if eq .Infra.Status.PlatformStatus.Type "None"}}true{{else}}false{{end}}`, and e2e-vsphere uses the vsphere platform.)

But the other template we do use for vSphere IPI.

I checked the last CI job failure and I think it was unreleated to this change as the cluster provisioned.
/lgtm

sinnykumari (approved):
@jcpowermac thank you for reviewing and confirming.

@danwinship I understand that PR may not be cause of test failure. Since, MCO team doesn't have expertise in platform specific changes and impact, before approval we try to get it reviewed first by respective team :slightly_smiling_face:

PR #2478, opened 2021-03-22
Bug 1939740: Use new --prefer-ipv6 flag to "runtimecfg node-ip" as appropriate

danwinship:
**- What I did**
Make MCO templates use new `runtimecfg node-ip --prefer-ipv6` flag when on single-stack IPv6, so that we pick the correct node IP when bringing up a single-stack IPv6 cluster on a node that happens to also have IPv4 IPs.

**- How to verify it**
Our infra doesn't currently create any such clusters anywhere but we can confirm that the option gets passed (and accepted) in single-stack IPv6 clusters, and not elsewhere. (Of course, I don't think MCO has a single-stack IPv6 presubmit job, so...)

(Why does the template ask for this? What changelog?)

/hold
for testing

### DIFF
--- a/templates/common/_base/units/nodeip-configuration.service.yaml
+++ b/templates/common/_base/units/nodeip-configuration.service.yaml
@@ -21,7 +21,11 @@ contents: |
     --volume /etc/systemd/system:/etc/systemd/system:z \
     {{ .Images.baremetalRuntimeCfgImage }} \
     node-ip \
-    set --retry-on-failure; \
+    set \
+    {{if eq .IPFamilies "IPv6" -}}
+    --prefer-ipv6 \
+    {{end -}}
+    --retry-on-failure; \
     do \
     sleep 5; \
     done"


danwinship:
>  Our infra doesn't currently create any such clusters anywhere but we can confirm that the option gets passed (and accepted) in single-stack IPv6 clusters, and not elsewhere.

except that the patch only affects UPI, not IPI, and dev-scripts only does IPI, so in fact, I can't test this at all. The actual baremetal-runtimecfg patch (https://github.com/openshift/baremetal-runtimecfg/pull/130) was tested to make sure it works when you invoke it manually but I don't know how to bring up a cluster that will test this PR

/cc @celebdor

danwinship:
OK, at least managed to confirm that it is getting used on single-stack IPv6 UPI:

```
● nodeip-configuration.service - Writes IP address configuration so that kubelet and crio services select a valid node IP
   Loaded: loaded (/etc/systemd/system/nodeip-configuration.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Tue 2021-03-23 20:01:10 UTC; 3min 53s ago
  Process: 1469 ExecStart=/bin/bash -c     until    /usr/bin/podman run --rm    --authfile /var/lib/kubelet/config.json    --net=host    --volume /etc/systemd/system:/etc/systemd/system:z registry.build01.ci.openshift.org/ci-ln-2sbrfgk/stable@sha256:d38f22111f380edb40f4d6a38822da074196f74447026008cc89ca514c868e3d    node-ip    set    --prefer-ipv6    --retry-on-failure;    do
   sleep 5;    done (code=exited, status=0/SUCCESS)
      CPU: 9.727s
```

(though on this host there was only an IPv6 IP to choose from anyway... but this at least confirms that (a) we're passing `--prefer-ipv6` when we should, and (b) it doesn't break anything. The unit tests on the runtimecfg side (and manual testing on dual-stack hosts) confirm it would have done the right thing if there had been an IPv4 IP.)

danwinship:
/retest
e2e-agnostic-upgrade seems super flaky for everyone...

danwinship:
/assign @kikisdeliveryservice
for approval

danwinship:
poke. can I get an approve? @kikisdeliveryservice @sinnykumari

sinnykumari (approved):

PR #2487, opened 2021-03-24
operator: upgradeable false when pools are still updating

kikisdeliveryservice:
Sets machine-config clusteroperator Upgradeable: False when pools are updating (covering future upgrades), does not block current upgrade from reporting finished on all completing.

To verify roll out a mcp then once pool is updating, verify Upgradeable: False:
```oc get co/machine-config -oyaml```

when pool finished, should toggle back to Upgradeable: True

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -262,13 +262,26 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		Status: configv1.ConditionTrue,
 		Reason: asExpectedReason,
 	}
+	var updating, degraded bool
 	for _, pool := range pools {
-		degraded := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolDegraded)
+		// collect updating status but continue to check each pool to see if any pool is degraded
+		if isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdating) {
+			updating = true
+		}
+		degraded = isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolDegraded)
+		// degraded should get top billing in the clusteroperator status, if we find this, set it and update
 		if degraded {
 			coStatus.Status = configv1.ConditionFalse
 			coStatus.Reason = "DegradedPool"
-			coStatus.Message = "One or more machine config pool is degraded, please see `oc get mcp` for further details and resolve before upgrading"
+			coStatus.Message = "One or more machine config pools are degraded, please see `oc get mcp` for further details and resolve before upgrading"
+			break
 		}
+	}
+	// updating and degraded can occur together, in that case defer to the degraded Reason that is already set above
+	if updating && !degraded {
+		coStatus.Status = configv1.ConditionFalse
+		coStatus.Reason = "PoolUpdating"
+		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
 	return optr.updateStatus(co, coStatus)
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -624,10 +624,6 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 				return false, nil
 			}
 		}
-		syncstatuserr := optr.syncUpgradeableStatus()
-		if syncstatuserr != nil {
-			glog.Errorf("Error syncingUpgradeableStatus: %q", syncstatuserr)
-		}
 		glog.Info("required machine-config pools synchronized")
 		return true, nil
 	}); err != nil {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/operator/status.go [hunk 1]
```diff
@@ -262,14 +262,27 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		Status: configv1.ConditionTrue,
 		Reason: asExpectedReason,
 	}
+	var updating, degraded bool
 	for _, pool := range pools {
-		degraded := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolDegraded)
+		// collect updating status but continue to check each pool to see if any pool is degraded
+		if isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdating) {
```
yuqi-zhang, line 268: Minor: maybe this can be a one-liner like degraded?

(updating = isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdating)

pkg/operator/sync.go
```diff
@@ -605,6 +605,15 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 				return false, nil
 			}
 
+			updating := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdating)
+			if updating {
+				glog.Infof("pool %s is still updating, Status: (pool updating: %v total: %d, ready %d, updated: %d, unavailable: %d)", pool.Name, updating, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount)
+				syncerr := optr.syncUpgradeableStatus()
```
yuqi-zhang, line 611: As a general question, why sync the upgradeable status here in a per-pool loop? The `syncUpgradeableStatus()` already does so per pool.

We also call this right after this loop (once master pool completes) and then again at https://github.com/openshift/machine-config-operator/blob/889af725b2dc5e7503c5b0cb9b0b710fd62da967/pkg/operator/sync.go#L91. Do you think it'd be worth it to consolidate some of those calls?

yuqi-zhang (commented):
Overall looks good, some minor questions below

Also in description it says `when pool finished, should toggle back to Upgradeable: False`, should that be `Upgradeable: True` when pools finish?

pkg/operator/status.go
sinnykumari: I think current code is fine as we want to flag `updating` to `true` when looping over all the pools without overriding previous value.

pkg/operator/sync.go
sinnykumari: I too think that syncUpgradeableStatus() shouldn't be needed here as we will anyway sync after we iterate over all pools.

sinnykumari (commented):
Overall LGTM, few inline commemts

pkg/operator/sync.go
```diff
@@ -605,6 +605,15 @@ func (optr *Operator) syncRequiredMachineConfigPools(_ *renderConfig) error {
 				return false, nil
 			}
 
+			updating := isPoolStatusConditionTrue(pool, mcfgv1.MachineConfigPoolUpdating)
+			if updating {
+				glog.Infof("pool %s is still updating, Status: (pool updating: %v total: %d, ready %d, updated: %d, unavailable: %d)", pool.Name, updating, pool.Status.MachineCount, pool.Status.ReadyMachineCount, pool.Status.UpdatedMachineCount, pool.Status.UnavailableMachineCount)
```
sinnykumari, line 610: can we not print status here? This floods machine-config-controller logs whenever a pool is progressing.

kikisdeliveryservice: Agree will remove

pkg/operator/sync.go
kikisdeliveryservice: Agree will remove

pkg/operator/status.go
kikisdeliveryservice: Yep exactly what Sinny said, we're tracking the existence of a true while finishing going thru the pools - we don't want to reset it to false (and lose that value) in the same iteration of the loop.

sinnykumari (approved):
This looks fine to me, will let Jerry do final review.

Noting this explicitly here that with this PR, MCO will block next upgrade until nodes in all the pools have finished updating. We may need to adjust the behavior in future when EUS workflow is finalized.

kikisdeliveryservice:
This was updated, can I get a final LGTM? Given the timing & ci state it could take a while to merge..

yuqi-zhang:
/lgtm

Just want to make sure of something, until master pool is ready, the syncAll function doesn't return since it's looping on https://github.com/openshift/machine-config-operator/blob/616a208c55e3df7651efc0c104d2604c7792396f/pkg/operator/sync.go#L614 right?

If so the first upgradeable status will happen here https://github.com/openshift/machine-config-operator/blob/889af725b2dc5e7503c5b0cb9b0b710fd62da967/pkg/operator/sync.go#L91, which is fine because the MCO hasn't reported the previous update from completing thereby blocking further upgrades. It just means that upgradeable=true during the master pool update, and then flips to upgradeable=false once master pool completes if the worker pool hasn't.

Is that assessment correct?

yuqi-zhang:
/lgtm

Eventually it might be nice to rework the sync a bit, such that we don't have to syncUpgradeable in the middle of the pool sync, but for now I think this will work as well

PR #2488, opened 2021-03-24
Bug 1929160: Add NetworkManager reload to resolv-prepender

cybertron:
If resolv.conf gets populated after NetworkManager has attempted to
do a reverse lookup on the node IP to get the hostname, we can end
up with a hostname of localhost because the lookup has no DNS servers
to use. To avoid this possibility, reload NetworkManager after we
populate resolv.conf, which should force it to redo the lookup.

Reload NetworkManager after populating resolv.conf to ensure we correctly populate the node hostname.

### DIFF
--- a/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
@@ -72,6 +72,10 @@ contents:
                 # Only leave the first 3 nameservers in /etc/resolv.conf
                 sed -i ':a $!{N; ba}; s/\(^\|\n\)nameserver/\n# nameserver/4g' /etc/resolv.tmp
                 mv -f /etc/resolv.tmp /etc/resolv.conf
+                # Workaround for bz 1929160. Reload NetworkManager to force it to
+                # re-run the lookup of the hostname now that we know we have DNS
+                # servers configured correctly in resolv.conf.
+                nmcli general reload dns-rc
             fi
         fi
         ;;


cybertron:
/hold

I'm hoping we can get this tested in an environment where the race condition has reproduced pretty consistently.

cybertron:
/retest
/hold cancel

We had a successful test in the failing environment, so I think we're good to merge this.

mandre (commented):
/lgtm

JAORMX (approved):
/retest

cybertron:
/test e2e-vsphere

The vsphere job looks pretty red lately. Might need to check in with that team.

cybertron:
/test e2e-vsphere

Couldn't acquire lease.

yuvalk:
cluster created with this PR still exhibit the issue described in the BZ.
ie one of the masters still shows with 'localhost'
https://bugzilla.redhat.com/show_bug.cgi?id=1929160#c21

cybertron:
/hold

It seems this alone wasn't enough to fix the problem. We're waiting on a NetworkManager change and will revisit this.

kikisdeliveryservice:
@cybertron do you still want this open?

cybertron:
Yes, I think we eventually need a patch like this. However, we need https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/832/diffs?commit_id=43fc98498968ef64edb270c5488c04ad94bdd232 to show up in RHCOS before it will work as intended. I'm not seeing it yet in my local deployments so I don't think this can proceed.

/hold

cybertron:
/test e2e-ovirt
/test e2e-vsphere

I've updated this to use the new nmcli reload command that forces NM to redo the DNS lookup. Hopefully this will fix the problem once and for all.

I'm going to leave the hold because it would be nice to test this in an environment where the problem reproduced.

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
```diff
@@ -72,6 +72,10 @@ contents:
                 # Only leave the first 3 nameservers in /etc/resolv.conf
                 sed -i ':a $!{N; ba}; s/\(^\|\n\)nameserver/\n# nameserver/4g' /etc/resolv.tmp
                 mv -f /etc/resolv.tmp /etc/resolv.conf
+                # Workaround for bz 1929160. Reload NetworkManager to force it to
+                # re-run the lookup of the hostname now that we know we have DNS
+                # servers configured correctly in resolv.conf.
+                nmcli general reload dns-rc
```
mandre, line 78: Strangely, `nmcli` version 1.30.6-1.fc34 doesn't show the reload command as a valid command:
```
nmcli general -h     
Usage: nmcli general { COMMAND | help }

COMMAND := { status | hostname | permissions | logging }

  status

  hostname [<hostname>]

  permissions

  logging [level <log level>] [domains <log domains>]
```

However, it's rightfully documented in the man page:
```
nmcli general {status | hostname | permissions | logging | reload} [ARGUMENTS...]
```

The man page also says that the `reload dns-rc` command "Update DNS configuration, which usually involves writing /etc/resolv.conf anew". Isn't it a problem for us? Won't it overwrite our changes to `/etc/resolv.conf`?

cybertron: Yeah, I actually delayed updating this patch because it wasn't showing up in the cli, but at some point I realized it was there anyway. :-)

The reason I used dns-rc is because https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/832 is the patch in response to this bug and it mentions using SIGUSR1 to notify NM that the DNS config had changed. In the docs, dns-rc corresponds to SIGUSR1. In my local tests it seemed to work fine, but I'm waiting on the bug reporter to try it in an environment where the bug reproduces. If it doesn't work then I may just do a general reload without parameters to reload everything.

cybertron:
/hold cancel

I'm still not aware of anyone having tested this in an environment where the problem reproduces, but based on the investigation of the problem so far there's a reasonable expectation that this will fix it and there's no point holding out a probably-good patch while we wait for pre-verification. I'm going to propose that we merge this and if it ends up not fixing the problem we can revisit.

yuvalk:
> /hold cancel
> 
> I'm still not aware of anyone having tested this in an environment where the problem reproduces, but based on the investigation of the problem so far there's a reasonable expectation that this will fix it and there's no point holding out a probably-good patch while we wait for pre-verification. I'm going to propose that we merge this and if it ends up not fixing the problem we can revisit.

I own a lab where issue can be reproduced, but currently we are running with this WA - 
https://github.com/openshift-kni/cnf-features-deploy/tree/master/feature-configs/deploy/localhost-workaround
I didnt have time to first try without it, make sure is reproducible then test with this PR ...

LorbusChris:
We get reports with hostname set to `localhost` rather frequently for OKD. Latest example: https://github.com/openshift/okd/discussions/1039

cybertron:
Yes, my reasoning from https://github.com/openshift/machine-config-operator/pull/2488#issuecomment-983828260 still holds. I'd like to get it in and verify the bug after. It will be easier once it's in real builds and doesn't require pulling down a PR.

templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
mandre: For reference, here is a excerpt of what `NetworkManager` man page says about `SIGUSR1`:
```
When NetworkManager is not managing DNS, the signal forces a restart of operations that depend on the DNS
configuration (like the resolution of the system hostname via reverse DNS, or the resolution of WireGuard peers);
therefore, it can be used to tell NetworkManager that the content of resolv.conf was changed externally. In the future,
further actions may be added.
```

So it sounds like something we should do, since we set `rc-manager=unmanaged` to tell NM not to write to `/etc/resolv.conf`.

mandre (commented):
/lgtm

PR #2493, opened 2021-03-26
Bug 1940585: conditionally set nodeip-service enabled for vSphere

patrickdillon:
The nodeip service should not be enabled for vSphere UPI. Prior to this change, enabled was always true but the content was empty for vSphere UPI. In 4.7, this has already been remedied and the service is correctly set to enabled: false. The discrepency between 4.6 and 4.7 breaks upgrades, so this commit fixes the issue in 4.6.

@yuqi-zhang @jcpowermac @rvanderp3

### DIFF
--- a/templates/common/vsphere/units/nodeip-configuration.service.yaml
+++ b/templates/common/vsphere/units/nodeip-configuration.service.yaml
@@ -1,11 +1,26 @@
 name: nodeip-configuration.service
-enabled: true
+enabled: {{if .Infra -}}
+         {{if .Infra.Status -}}
+         {{if .Infra.Status.PlatformStatus -}}
+         {{if .Infra.Status.PlatformStatus.VSphere -}}
+         {{if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
+         true
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end }}
 contents: |
-  {{ if .Infra -}}
-  {{ if .Infra.Status -}}
-  {{ if .Infra.Status.PlatformStatus -}}
-  {{ if .Infra.Status.PlatformStatus.VSphere -}}
-  {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
   [Unit]
   Description=Writes IP address configuration so that kubelet and crio services select a valid node IP
   # This only applies to VIP managing environments where the kubelet and crio IP
@@ -18,6 +33,11 @@ contents: |
   [Service]
   # Need oneshot to delay kubelet
   Type=oneshot
+  {{ if .Infra -}}
+  {{ if .Infra.Status -}}
+  {{ if .Infra.Status.PlatformStatus -}}
+  {{ if .Infra.Status.PlatformStatus.VSphere -}}
+  {{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}
   # Would prefer to do Restart=on-failure instead of this bash retry loop, but
   # the version of systemd we have right now doesn't support it. It should be
   # available in systemd v244 and higher.
@@ -34,6 +54,11 @@ contents: |
     do \
     sleep 5; \
     done"
+  {{ end -}}
+  {{ end -}}
+  {{ end -}}
+  {{ end -}}
+  {{ end -}}
   ExecStart=/bin/systemctl daemon-reload
 
   {{if .Proxy -}}
@@ -50,8 +75,3 @@ contents: |
 
   [Install]
   WantedBy=multi-user.target
-  {{ end -}}
-  {{ end -}}
-  {{ end -}}
-  {{ end -}}
-  {{ end -}}


jcpowermac:
Since this is 4.6 there will be docker rate limiting.
Even with UPI test failures this should be ok. Lets try IPI again.

/test e2e-vsphere

templates/common/vsphere/units/nodeip-configuration.service.yaml
```diff
@@ -1,5 +1,5 @@
 name: nodeip-configuration.service
-enabled: true
+enabled: {{ if .Infra -}}{{ if .Infra.Status -}}{{ if .Infra.Status.PlatformStatus -}}{{ if .Infra.Status.PlatformStatus.VSphere -}}{{ if .Infra.Status.PlatformStatus.VSphere.APIServerInternalIP -}}true{{else}}false{{end}}{{end}}{{end}}{{end}}{{end}}
```
yuqi-zhang, line 2: If I am reading this correctly, this is basically
```
if a
  if b
    if c
      true
    else
      false
    end
  end
end
```

Shouldn't it be `false` for each condition that it doesn't meet?

```
if a
  if b
    if c
      true
    else
      false
    end
  else
    false
  end
else
  false
end
```

such that if any conditions aren't met it still gets written (not empty).

yuqi-zhang (commented):
Three comments:

1. Maybe we should align with https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/units/nodeip-configuration.service.yaml and have the unit contents always be written (remove the if conditionals there) and just have the if on the enabled field?

2. See below comment

3. I think we should also relax the check in enableUnits function such that if the contents are empty we just no-op. Do you think that would work? If so I can try to patch that into 4.8/4.7 so we don't have to block too many edges

templates/common/vsphere/units/nodeip-configuration.service.yaml
patrickdillon: @yuqi-zhang  Thanks for the thorough review. I agree with all of this and have pushed the changes. I will need to check CI to make sure the whitespace formatting is good on the new conditional, but I think all is good as far as I know if you want to take another look. 

> I think we should also relax the check in enableUnits function such that if the contents are empty we just no-op (in addition to this PR). Do you think that would work? If so I can try to patch that into 4.8/4.7 so we don't have to block too many edges

This sounds reasonable to me. When you say it will no-op, you mean in the sense of allowing upgrades to progress even if enabled: true contents: <empty>? That sounds like another graceful way of handling this situation we know is problematic.

yuqi-zhang: I think the new logic looks correct. Since this is 4.6 only I think I'm ok with the fix as is assuming formatting is correct.

> When you say it will no-op, you mean in the sense of allowing upgrades to progress even if enabled: true contents: ? That sounds like another graceful way of handling this situation we know is problematic.

Yes, I will look to add a check that on empty units (with enable defined) we just output a warning in the MCD logs. We can do it in conjunction with this PR.

yuqi-zhang:
/test e2e-vsphere-upi
/test e2e-vsphere

to check template rendering

yuqi-zhang:
Opened https://github.com/openshift/machine-config-operator/pull/2498 as a complementary solution

yuqi-zhang:
Oh the upi-vsphere job actually reports an error:

```
error running MCC[BOOTSTRAP]: failed to create MachineConfig for role master: failed to execute template: template: /etc/mcc/templates/common/vsphere/units/nodeip-configuration.service.yaml:48:12: executing "/etc/mcc/templates/common/vsphere/units/nodeip-configuration.service.yaml" at <.Infra.Status.PlatformStatus.VSphere.APIServerInternalIP>: nil pointer evaluating *v1.VSpherePlatformStatus.APIServerInternalIP
```

I'm not an expert on formatting, are the `-`'s causing a problem?

patrickdillon:
> Oh the upi-vsphere job actually reports an error:
> 
> ```
> error running MCC[BOOTSTRAP]: failed to create MachineConfig for role master: failed to execute template: template: /etc/mcc/templates/common/vsphere/units/nodeip-configuration.service.yaml:48:12: executing "/etc/mcc/templates/common/vsphere/units/nodeip-configuration.service.yaml" at <.Infra.Status.PlatformStatus.VSphere.APIServerInternalIP>: nil pointer evaluating *v1.VSpherePlatformStatus.APIServerInternalIP
> ```
> 
> I'm not an expert on formatting, are the `-`'s causing a problem?

The formatting should just be removing whitespace as far as I understand. I have pushed a new commit with whitespace formatting that is more conventional because I now see that the leading `-` isn't necessary when preceded by a trailing `-`. 

I'll let CI see how it feels about that while I try to figure out what's going on. The whole purpose of these `if`s is to catch nils...

patrickdillon:
Doh!

@yuqi-zhang There was an `APIServerInternalIP` in the contents here:
https://github.com/openshift/machine-config-operator/pull/2493/files#diff-198ed38eadea16913d0c0b3ff43d17fd847a1c459ea4df3f627e693f2e651c13L33

So when I removed the conditional we hit the nil pointer. I removed my changes to the contents, tidied up the formatting and I think we should be good to go.

patrickdillon:
Machine configs look good to me:
```yaml
      - contents: |
          [Unit]
          Description=Writes IP address configuration so that kubelet and crio services select a valid node IP
          # This only applies to VIP managing environments where the kubelet and crio IP
          # address picking logic is flawed and may end up selecting an address from a
          # different subnet or a deprecated address
          Wants=network-online.target
          After=network-online.target ignition-firstboot-complete.service
          Before=kubelet.service crio.service

          [Service]
          Type=oneshot
          ExecStart=/bin/systemctl daemon-reload

          [Install]
          WantedBy=multi-user.target
        enabled: false
```
looks like e2e-vsphere-upi had a successful install but hit the docker rate limiting @jcpowermac was referring to.
/test e2e-vsphere-upi

patrickdillon:
Relevant tests are green.

yuqi-zhang (approved):
/lgtm

PR #2496, opened 2021-03-29
Bug 1899057: configure-ovs-network: fix spurious OVS warnings

trozet:
Fixes the following error:
ERR|interface br-ex: ignoring mac in Interface record

While benign, this error would fill up the OVS logs. Old versions of
Network Manager required the cloned-mac to be set on both the bridge and
the bridge interface. However, in newer versions of NM this has been
fixed and it only needs to be set on the interface. See:

https://bugzilla.redhat.com/show_bug.cgi?id=1899745#c8


trozet:
@danwinship PTAL

kikisdeliveryservice (approved):
approve deferring to @danwinship for final review

/assign @danwinship

PR #2498, opened 2021-03-30
WIP: Bug 1940585: units: do not fail enable/disable for empty systemd units

yuqi-zhang:
Pre-4.7, we allowed specifying enable/disable for a systemd unit
definition that had empty content. This would write a dangling
symlink if the service didn't otherwise exist on the system, but
we did not hard error.

In 4.7 we now hard error on units that cannot be enabled, such that
the following definition:

enabled: true

with empty contents would fail. This was shipped in 4.6 as a template
that would manifest on UPI vsphere, and could cause upgrade failures
if you start a 4.7 upgrade before worker pools completed a 4.6 upgrade
(we do not have wait for all pools in 4.6).

To make it more consistent with previous behaviour, relax the error
such that it becomes a warning instead, only for the above situation
where an empty content unit is trying to be enabled. Units that otherwise
has content defined still hard error, in case your unit is misconfigured.

This can be done in conjunction with https://github.com/openshift/machine-config-operator/pull/2493 which fixes the template. Marking WIP since the implementation is not very clean

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1445,6 +1445,7 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 }
 
 // writeUnits writes the systemd units to disk
+//nolint:gocyclo
 func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 	var enabledUnits []string
 	var disabledUnits []string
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
cc @patrickdillon

patrickdillon:
Seems reasonable to my untrained eye.
/test e2e-vsphere-upgrade

PR #1920, opened 2020-07-10
[release-4.5] Bug 1855543: KubeletConfig: Validate kubeReserved and systemReserved for KubeletConfig

harche:
This is an automated cherry-pick of https://github.com/openshift/machine-config-operator/pull/1885

PR #2499, opened 2021-03-30
WIP: Bug 1930636: Fix validation for LogSizeMax and OverlaySize in ContainerRuntimeConfig

harche:

**- What I did**

**- How to verify it**


pkg/controller/container-runtime-config/helpers.go
```diff
@@ -442,12 +443,26 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 		return fmt.Errorf("invalid PidsLimit %v", *ctrcfg.PidsLimit)
 	}
 
+	q, err := resource.ParseQuantity(ctrcfg.LogSizeMax.String())
```
mtrmac, line 446: `ctrcfg.LogSizeMax` must already be a valid object for the `.String()` call to be possible. What is this supposed to do? (E.g. note that `UnmarshalJSON` already calls `ParseQuantity`).

The bug references a
```
unable to decode an event from the watch stream: unable to decode watch event: v1.ContainerRuntimeConfig.Spec: v1.ContainerRuntimeConfigSpec.MachineConfigPoolSelector: ContainerRuntimeConfig: v1.ContainerRuntimeConfiguration.OverlaySize: unmarshalerDecoder: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$', error found in #10 byte of ...|\":\"9asadG\"
```

Apparently at that point validation already happens. If we can’t catch that error (can’t we?), why is `.String()` valid to call here?

Also compare https://github.com/openshift/machine-config-operator/blob/95a704487887311129e4d9831312e9f3393184d7/pkg/controller/container-runtime-config/container_runtime_config_controller.go#L556 .

---

Also, the field is defined as `omitempty`. What happens if the value is missing? `resource.ParseQuantity` rejects both `""` and `null`, unlike `resource.Quantity.UnmarshalJSON`.

---

This may well be just my ignorance of something obvious everyone should already know; if not, please describe the mechanism in code comments down, and make sure it is consistent (along with handling of missing values) throughout the codebase.

(In both instances.)

harche: Thanks @mtrmac for your comments. I have marked this PR `WIP` to allow me to play with the changes using the cluster bot. So this work is far from over. 

Handling the negative value check for `OverlaySize` is pretty simple, but I also wanted to explore if I could capture `non-digital` values like `9asadG`. The main inspiration was to try out something that I had used in KubeletConfig once upon a time, https://github.com/openshift/machine-config-operator/pull/1920/files#diff-4a2d0a4442e10d889a927322ad7311c67b22b259622067b52cbda7d2ece3ebefR128-R140

The idea was to use `resource.ParseQuantity` to run the input through similar validation. But you are right, which I also verified using cluster bot, that `OverlaySize` is already a valid object by the time `validateUserContainerRuntimeConfig` in invoked. 

I am still exploring how can we capture, 
```
unable to decode an event from the watch stream: unable to decode watch event: v1.ContainerRuntimeConfig.Spec: v1.ContainerRuntimeConfigSpec.MachineConfigPoolSelector: ContainerRuntimeConfig: v1.ContainerRuntimeConfiguration.OverlaySize: unmarshalerDecoder: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$', error found in #10 byte of ...|\":\"9asadG\"
``` 

Please let me know if you know how can we handle that.

harche: Maybe simply checking for `OverlaySize.Value() > 0` and if it's of type `int` or not should be good enough.

mtrmac: > I have marked this PR WIP to allow me to play with the changes using the cluster bot. So this work is far from over.

Ah, my mistake, I didn’t notice.

I’m afraid I don’t have any suggestions, I’m not familiar with the intricacies of the RawExtension mechanism and informers.

harche:
Closing this in favor of https://github.com/openshift/machine-config-operator/pull/2503

PR #2501, opened 2021-03-30
WIP - Try Single node with more system reserved memory

omertuc:
Just trying to run CI with a small diff, please ignore

PR #2466, opened 2021-03-18
OCPNODE-367: Auto node sizing for Openshift

harche:
Signed-off-by: Harshal Patil <harpatil@redhat.com>

**- What I did**
Introduce auto node sizing for Openshift - https://issues.redhat.com/browse/OCPNODE-367

**- How to verify it**
To enable auto node sizing on worker node use following yaml

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: KubeletConfig
metadata:
  name: dynamic-node 
spec:
  autoSizingReserved: true 
  machineConfigPoolSelector:
    matchLabels:
      pools.operator.machineconfiguration.openshift.io/worker: ""
```
Once it's rolled out, log into the worker node,

```bash
# ps -ef | grep kubelet
root        1613       1 11 06:49 ?        00:00:05 kubelet --config=/etc/kubernetes/kubelet.conf --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig --kubeconfig=/var/lib/kubelet/kubeconfig --container-runtime=remote --container-runtime-endpoint=/var/run/crio/crio.sock --runtime-cgroups=/system.slice/crio.service --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_id=rhcos --node-ip= --minimum-container-ttl-duration=6m0s --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec --cloud-provider=azure --cloud-config=/etc/kubernetes/cloud.conf --pod-infra-container-image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7b8e2e2857d8ac3499c9eb4e449cc3296409f1da21aa21d0140134d611e65b84 --system-reserved=cpu=0.07,memory=2.5Gi --v=2
```
You should be able to see kubelet's `--system-reserved` flag is set for optimal values of `cpu` and `memory`

**- Description for the changelog**
Enable auto node sizing post installation

PR #2504, opened 2021-03-31
Bug 1945017: Increase system reserved memory from 1Gi to 1.8Gi to support single node clusters

omertuc:
When running E2E tests on single node clusters, the 1Gi reserved for
system memory is insufficient.

During this PR: https://github.com/openshift/machine-config-operator/pull/2501 -

I had 3 e2e test runs on AWS single node, the peak recorded system memory
usage during those tests was 1.40, 1.31 and 1.19 GiB respectively. In
this PR I also saw a run that peaked at 1.56 GiB:

image

image

The SystemMemoryExceedsReservation alerts demands that the actual usage
would be less than 90% of the amount reserved, so that means the
corresponding thresholds that should be set are at least 1.44, 1.46, 1.32 and
1.74 GiB.

Or in short, the reserved memory should be increased to 1.8GiB to
support single node (with some hopefully sufficient padding).

Possible future improvements -

1) Different threshold depending on whether the cluster is a single node
cluster or not

2) Find a way to lower single node system memory usage

### DIFF
--- a/templates/common/_base/files/kubelet-auto-sizing.yaml
+++ b/templates/common/_base/files/kubelet-auto-sizing.yaml
@@ -85,7 +85,7 @@ contents:
     }
     function static_node_sizing {
         rm -f ${NODE_SIZES_ENV}
-        echo "SYSTEM_RESERVED_MEMORY=1Gi" >> ${NODE_SIZES_ENV}
+        echo "SYSTEM_RESERVED_MEMORY=1.8Gi" >> ${NODE_SIZES_ENV}
         echo "SYSTEM_RESERVED_CPU=500m" >> ${NODE_SIZES_ENV}
     }
 


kikisdeliveryservice:
@rphillips any issues with this?

@omertuc please update the BZ as bot requested

omertuc:
> @omertuc please update the BZ as bot requested

It has to has to be triaged before I can do that, see comment in bz

rphillips:
For SNO, we will likely want the Auto Node Sizing feature enabled: https://github.com/openshift/machine-config-operator/pull/2466

rphillips:
The PR will somehow need to detect if SNO is enabled to set the reserve to 1.8GB

mrunalp:
/hold
The preferred approach would be to bump up the system reserved just for the single node CI jobs.

omertuc:
Closed in favor of https://github.com/openshift/release/pull/17403

PR #2508, opened 2021-03-31
Bug 1945431: alerts: SystemMemoryExceedsReservation triggers too quickly

smarterclayton:
As a warning, the alert should only fire over sustained usage
because this is a symptom and may peak briefly during startup.
Set the timeout to 15m and slightly improve the alert message.

The `SystemMemoryExceedsReservation` alert requires a longer period of extended usage before firing, and the description is slightly clarified.


kikisdeliveryservice (approved):
This is fine with me

since he added this alert:
/assign @sjenning

PR #2509, opened 2021-04-01
daemon: drop `systemctl stop kubelet` during node reboot

sinnykumari:
* With the upstream implementation of kubelet graceful shutdown feature, we don't explicitly stop the kubelet
 so that kubelet can gracefully shutdown pods when `GracefulNodeShutdown` feature gate is enabled. kubelet uses systemd inhibitor locks to delay node shutdown to terminate pods.
https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
* On HA cluster MCO today performs explicit pod drain on node, so this feature is not much useful there. But quite useful on Non -HA clusters (like Single Node OpenShift) , where we skip pod drain. Such clusters can take advantage of `GracefulNodeShutdown` feature to terminate critical pods.


cgwalters (approved):
Makes sense to me!

mrunalp:
We can enable this after we rebase to 4.8 and enable the settings for node graceful shutdown.

sinnykumari:
ok, I am putting hold on the PR until then.
@mrunalp please let us know when we have enabled graceful node shutdown feature in 4.8

/hold

kikisdeliveryservice:
is nobody going to ask how @sinnykumari got the font in the title to be cool?

rphillips:
+1 after the rebase lands

JAORMX (approved):

kikisdeliveryservice:
@mrunalp @rphillips can we merge this now? do we want to backport?

sinnykumari:
Not sure but I think node team is testing it in https://github.com/openshift/machine-config-operator/pull/2524 where tests are currently failing

sinnykumari:
still valid until https://github.com/openshift/machine-config-operator/pull/2524 gets merged

cgwalters:
Just to xref I think we want this *and* https://github.com/openshift/machine-config-operator/pull/3136

yuqi-zhang:
Just to double check, is graceful shutdown now in OCP 4.12?

sinnykumari:
Not yet, but I think plan is to have this feature available in 4.12. So, we will need this PR to adapt to this feature

cgwalters:
> Just to xref I think we want this and https://github.com/openshift/machine-config-operator/pull/3136

I was wrong because I was confused by the difference between delay and blocking locks.  #3136 would be needed for us to honor *blocking* locks, which I think we also do want, but isn't a blocker (heh) for this.

/approve

I think we can just ship this without even having graceful shutdown enabled right now.

sinnykumari:
Fixes [MCO-345](https://issues.redhat.com/browse/MCO-345)

sinnykumari:
Removing hold since this patch can go in independently without waiting from graceful shutdown enabled.
/hold cancel

sinnykumari:
@yuqi-zhang PTAL and LGTM if this looks alright to get merged

yuqi-zhang (approved):
/lgtm

As I understand, this should be safe in our model anyways?

sinnykumari:
Looks like ci issue.
/retest

PR #2516, opened 2021-04-07
Bug 1899187: run afterburn-hostname only when Network Manager is up

Fedosin:
To correct setting of hostname, afterburn requires Network Manager to be up. To prevent potential race conditions this commit ensures that we start afterburn only when Network Manager is online and blocks services relying on Networking being up.

### DIFF
--- a/templates/common/openstack/units/afterburn-hostname.service.yaml
+++ b/templates/common/openstack/units/afterburn-hostname.service.yaml
@@ -3,6 +3,11 @@ enabled: true
 contents: |
   [Unit]
   Description=Afterburn Hostname
+  # Block services relying on Networking being up.
+  Before=network-online.target
+  # Wait for NetworkManager to report its online
+  After=NetworkManager-wait-online.service
+  # Run before hostname checks
   Before=node-valid-hostname.service
 
   [Service]


mandre:
Please add a description to the PR and a body to the commit message.

templates/common/openstack/units/afterburn-hostname.service.yaml
```diff
@@ -3,6 +3,9 @@ enabled: true
 contents: |
   [Unit]
   Description=Afterburn Hostname
```
mandre, line 5: I think we'll also want to block services relying on Networking being up, same as [what GCP did](https://github.com/openshift/machine-config-operator/blob/0e66c5d6124440d298c6014a5372fe3f0f136c8e/templates/common/gcp/units/gcp-hostname.service.yaml#L8-L9).

Fedosin: It's already done: https://github.com/openshift/machine-config-operator/blob/master/templates/common/openstack/units/afterburn-hostname.service.yaml#L13

mandre (commented):
/lgtm
/approve

PR #2519, opened 2021-04-09
Bug 1943320: configure-ovs doesn't handle bond interfaces correctly for OVNKubernetes

msherif1234:
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>

update all existing connection with master set to use the new devices name.

**- What I did**
Walk through all existing connection and modify any connection with master to use the new connection device's name.

**- How to verify it**
follow repro steps doc in 1943320

update connection UUID for bond interfaces for OVNKubernetes

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -173,6 +173,15 @@ contents:
           connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} ${extra_phys_args}
       fi
 
+      # Update connections with master property set to use the new device name
+      new_device=$(nmcli --get-values connection.interface-name conn show ovs-if-phys0)
+      for conn_uuid in $(nmcli -g UUID connection show) ; do
+        if [ "$(nmcli -g connection.master connection show uuid "$conn_uuid")" != "$old_conn" ]; then
+          continue
+        fi
+        nmcli conn mod uuid ${conn_uuid} connection.master ${new_device}
+      done
+
       nmcli conn up ovs-if-phys0
 
       if ! nmcli connection show ovs-if-br-ex &> /dev/null; then


templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -173,6 +173,13 @@ contents:
           connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} ${extra_phys_args}
       fi
 
+      # Move bond/vlan member interfaces to the new ovs-if-phys0 connection
+      if [[ ${iface_type} == bond || ${iface_type} == vlan ]]; then
+        new_conn=$(nmcli --get-values connection.uuid conn show ovs-if-phys0)
+        sed -i s/master=${old_conn}/master=${new_conn}/ $NM_CONN_PATH/*.nmconnection
```
trozet, line 179: let's use nmcli con modi connection.master=${device_name}

Advantages are:
1. we don't have to modify a file and assume a path.
2. we don't need to revert the change later because using the device ID should work for both connections

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -275,6 +282,11 @@ contents:
       echo "ERROR: Failed to activate ovs-if-br-ex NM connection"
       # if we made it here networking isnt coming up, revert for debugging
       set +e
+      # Revert back bond and vlan interfaces old connection
```
trozet, line 285: i dont think we need to revert, i think we are just going to get ourselves in trouble if the script bails out mid run and doesnt revert the change. IMO we should just set device name

trozet (changes requested):

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: as discussed offline changed to use device name instead of UID and kept sed for simplicity

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: removed the revert logic

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -173,6 +173,13 @@ contents:
           connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} ${extra_phys_args}
       fi
 
+      # Update all connections that has master to the new device
+      conns=$(nmcli --fields UUID conn show | sed 1d)
```
thom311, line 177: `$(nmcli -g UUID connection)` ?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -173,6 +173,13 @@ contents:
...
+      while IFS= read -r conn; do
+        new_device=$(nmcli --get-values connection.interface-name conn show ${conn})
+        nmcli conn mod ${conn} connection.master=${new_device}
```
thom311, line 180: this doesn't seem right. It sets the master property of a profile to its own interface-name.

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: Cool Thanks!

templates/common/_base/files/configure-ovs-network.yaml
msherif1234: hmm I though the above will give the same effect of doing 
sed -i s/master=${old_conn}/master=${new_device}/ $NM_CONN_PATH/*.nmconnection

thom311: ah, I think it only would have the same effect,if the previous `master=` as set to `${old_conn}`.

So you would need something like
```
   nmcli -g connection.master connection show uuid "$conn" | grep -q -F "$old_conn" || continue
```

msherif1234: nmcli -g connection.master connection show 
Error: invalid field 'connection.master'; allowed fields: NAME,UUID,TYPE,TIMESTAMP,TIMESTAMP-REAL,AUTOCONNECT,AUTOCONNECT-PRIORITY,READONLY,DBUS-PATH,ACTIVE,DEVICE,STATE,ACTIVE-PATH,SLAVE,FILENAME.

thom311: i said `nmcli -g connection.master connection show uuid "$conn"` :)

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -173,6 +173,15 @@ contents:
           connection.autoconnect-priority 100 802-3-ethernet.mtu ${iface_mtu} ${extra_phys_args}
       fi
 
+      # Update connections with master property set to use the new device name
+      for conn_uuid in $(nmcli -g UUID connection show) ; do
+        if [ "$(nmcli -g connection.master connection show uuid "$conn_uuid")" != "$old_conn" ]; then
+          continue
+        fi
+        new_device=$(nmcli --get-values connection.interface-name conn show ${conn_uuid})
```
marbindrakon, line 181: Shouldn't this be getting the device name from $old_conn rather than $conn_uuid?

marbindrakon: or maybe just use $iface

trozet: It needs to be from ovs-if-phys0 I think and this line should be moved outside of the for loop so we only do it one time

msherif1234: $iface it seems to be set only for bond and vlan I will use ovs-if-phys0 that return the proper device name
```
sh-4.4# new_device=$(nmcli --get-values connection.interface-name conn show ovs-if-phys0)
sh-4.4# echo $new_device
ens4
```

trozet (approved):
/lgtm

marbindrakon (approved):
Looks good to me now

kikisdeliveryservice (approved):

PR #2521, opened 2021-04-09
Bug 1948431: Bump openshift/api

gnufied:
This will allow us to get latest feature gates about CSI migration

Note: this is needed for CSI migration feature to work in OCP 4.8.
Goal of this PR is to apply --feature-gate=CSIMigrationAWS=true,CSIMigrationOpenStack=true to kubelet when TechPreviewNoUpgrade featureSet is enabled, which was introduced in openshift/api#880


jsafrane:
Note: this is needed for CSI migration feature to work in OCP 4.8.
Goal of this PR is to apply `--feature-gate=CSIMigrationAWS=true,CSIMigrationOpenStack=true` to kubelet when `TechPreviewNoUpgrade` featureSet is enabled, which was introduced in https://github.com/openshift/api/pull/880

kikisdeliveryservice:
Looks like this breaks the kubelet unit test TestFeatureGateDrift. You'll need to update =)

jsafrane:
@kikisdeliveryservice added BZ and fixed unit tests. PTAL.

kikisdeliveryservice:
@rphillips can you PTAL since this touches some kubelet templates?

kikisdeliveryservice (approved):

sinnykumari (approved):
/retest

PR #2523, opened 2021-04-12
Test for keepalived vip cleanup

cybertron:
Just combining #2511 and #2460 to see why the former is failing. This is just a shortcut while 2460 is in ci purgatory.


cybertron:
This didn't work anyway because without the api vip you can't get must-gather.

PR #2524, opened 2021-04-12
WIP [do not merge] adds graceful shutdown vars for testing

wgahnagl:
**- What I did**

**- How to verify it**


sinnykumari:
You may need to apply https://github.com/openshift/machine-config-operator/pull/2509 as well for testing this feature.

templates/master/01-master-kubelet/_base/files/kubelet.yaml
```diff
@@ -19,6 +19,8 @@ contents:
     kubeAPIQPS: 50
     kubeAPIBurst: 100
     rotateCertificates: true
+    shutdownGracePeriod: 100
```
rphillips, line 22: these times will need to be much larger... maybe 600 for 10 minutes.

rphillips:
Adding a breadcrumb to this PR.... We ran into issues with the Graceful Shutdown feature, and it will need to be fixed upstream.

PR #2527, opened 2021-04-14
Bug 1949387: Fix the typo in reserved calculation in auto sizing script

harche:

Fixes : https://bugzilla.redhat.com/show_bug.cgi?id=1949387

**- What I did**
Fixed the typo in auto node sizing script

**- How to verify it**
Enable auto node sizing where `getconf _NPROCESSORS_ONLN` is 2, and the auto node sizing should not fail. 

Fix the typo reserved cpu calculation in auto sizing script

### DIFF
--- a/templates/common/_base/files/kubelet-auto-sizing.yaml
+++ b/templates/common/_base/files/kubelet-auto-sizing.yaml
@@ -37,7 +37,7 @@ contents:
             recommended_systemreserved_memory=$(echo $recommended_systemreserved_memory 6.72 | awk '{print $1 + $2}')
             total_memory=$((total_memory-112))
         fi
-        if (($total_memory >= 128)); then # 2% of any memory above 128GB
+        if (($total_memory >= 0)); then # 2% of any memory above 128GB
             recommended_systemreserved_memory=$(echo $recommended_systemreserved_memory $(echo $total_memory 0.02 | awk '{print $1 * $2}') | awk '{print $1 + $2}')
         fi
         echo "SYSTEM_RESERVED_MEMORY=${recommended_systemreserved_memory}Gi">> ${NODE_SIZES_ENV}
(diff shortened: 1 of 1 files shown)


templates/common/_base/files/kubelet-auto-sizing.yaml
```diff
@@ -37,7 +37,7 @@ contents:
             recommended_systemreserved_memory=$(echo $recommended_systemreserved_memory 6.72 | awk '{print $1 + $2}')
             total_memory=$((total_memory-112))
         fi
-        if (($total_memory >= 128)); then # 2% of any memory above 128GB
+        if (($total_memory >= 0)); then # 2% of any memory above 128GB
```
harche, line 40: this was a bug, if the memory was above 128 without this fix it will yield incorrect result.

templates/common/_base/files/kubelet-auto-sizing.yaml
```diff
@@ -46,15 +46,16 @@ contents:
         total_cpu=$(getconf _NPROCESSORS_ONLN)
         recommended_systemreserved_cpu=0
         if (($total_cpu <= 1)); then # 6% of the first core
-            recommended_systemreserved_cpu=$(echo $total_cpu 0.60 | awk '{print $1 * $2}')
+            recommended_systemreserved_cpu=$(echo $total_cpu 0.06 | awk '{print $1 * $2}')
```
harche, line 49: A typo, we need to get 6% of the CPU. Because of the typo, it was calculating 60%. However, since we never have system in CI or in production with just 1 CPU. This bug never surfaced.

templates/common/_base/files/kubelet-auto-sizing.yaml
```diff
@@ -46,15 +46,16 @@ contents:
...
         else
             recommended_systemreserved_cpu=0.06
             total_cpu=$((total_cpu-1))
         fi
         if (($total_cpu <= 1)); then # 1% of the next core (up to 2 cores)
-            recommended_systemreserved_cpu=$(echo $recommended_systemreserved_cpu $(echo $total_cpu 0.10 | awk '{print $1 * $2}') | awk '{print $1 + $2}')
-            total_cpu=0 else
+            recommended_systemreserved_cpu=$(echo $recommended_systemreserved_cpu $(echo $total_cpu 0.01 | awk '{print $1 * $2}') | awk '{print $1 + $2}')
+            total_cpu=0
+        else
```
harche, line 58: `else` in this line was misplaced.

templates/common/_base/files/kubelet-auto-sizing.yaml
```diff
@@ -65,7 +66,7 @@ contents:
             recommended_systemreserved_cpu=$(echo $recommended_systemreserved_cpu 0.01 | awk '{print $1 + $2}')
             total_cpu=$((total_cpu-2))
         fi
-        if (($total_cpu >= 4)); then # 0.25% of any cores above 4 cores
+        if (($total_cpu >= 0)); then # 0.25% of any cores above 4 cores
```
harche, line 69: This was a bug too. Above 4 cores, without this fix we would start to slightly deviate away from the desired output.

harche:
```
Core       Reserved Value
1              0.06
2              0.07
3              0.075
4              0.08
5              0.0825

after 5, keep adding 0.0025 for per cpu core
```

templates/common/_base/files/kubelet-auto-sizing.yaml
harche:
```
memory    reserved 
128           9.32Gi
129           9.34Gi
130           9.36Gi
```
after 128, add 2% of the remaining memory to the reserved.

harche:
/hold cancel

Verified with various inputs.

templates/common/_base/files/kubelet-auto-sizing.yaml
harche: Also, there was a typo on line 56 where we wanted to get 1% of CPU but we were getting 10%.

sinnykumari (approved):
Not blocking the PR but writing unit test would be more appropriate way to make sure we are getting expecting result in each cases.

harche:
> Not blocking the PR but writing unit test would be more appropriate way to make sure we are getting expecting result in each cases.

This script is embedded in rhcos ignition file. There is no way to execute this script unless deployed on the node by the installer.

PR #2528, opened 2021-04-14
Bug 1949664: [on-prem] Disable liveness probe until keepalived.conf exists

cybertron:
Keepalived doesn't necessarily start immediately when its container
is created. It waits until keepalived.conf exists. In some case,
it takes longer for the config file to be created than for the
liveness probe to time out, which triggers unnecessary restarts of
the container. Once the monitor populates keepalived.conf it will
send the reload command, which starts keepalived.

This change just adds a check for the keepalived.conf file to the
liveness probe. This way, the liveness probe will always succeed
until keepalived.conf is created.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/on-prem/files/keepalived.yaml
+++ b/templates/common/on-prem/files/keepalived.yaml
@@ -125,7 +125,7 @@ contents:
             - /bin/bash
             - -c
             - |
-              echo "State = FAULT" > /tmp/keepalived.data && kill -s SIGUSR1 "$(pgrep -o keepalived)" && for i in $(seq 5); do grep -q "State = FAULT" /tmp/keepalived.data && sleep 1 || exit 0; done && exit 1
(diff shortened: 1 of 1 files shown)


cybertron:
/bugzilla refresh
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

I'm not sure this is actually hurting anything, but it will clean up some unnecessary errors in the logs.

mandre (commented):
/lgtm

kikisdeliveryservice:
ping @jcpowermac @patrickdillon

JAORMX (approved):
/retest

yboaron:
@kikisdeliveryservice , Do you think we can merge this one? 

It should stabilize CI gates (since it should  prevent keepalived containers failure and restarts  during deployment time )

PR #2531, opened 2021-04-15
Miscellaneous fixes

crawford:


Makefile
```diff
@@ -20,6 +20,8 @@ export GOPROXY=https://proxy.golang.org
 GOTAGS = "containers_image_openpgp exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_ostree_stub"
 
 .PHONY: clean test test-unit test-e2e verify update install-tools
+.DEFAULT_GOAL = binaries
```
sinnykumari, line 23: I like this!

sinnykumari (approved):
This looks good to me.

hack/build-image
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
sinnykumari: Since we are making it shell script in later commit. We can drop this commit.

PR #2535, opened 2021-04-17
Bug 1950133: Add cleanupStatus to keep three kubelet config conditions

QiWang19:

Close: Bug 1950133 https://bugzilla.redhat.com/show_bug.cgi?id=1950133

**- What I did**
Add cleanUpStatus method to keep a limited number of conditions of different
timestamps. The list keeps the three latest conditions. Updates only the timestamp
if the new status condition stays the same as the last condition.

**- How to verify it**

fix the bug MCO creates empty conditions on the kubeletconfig object

### DIFF
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -346,16 +346,13 @@ func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.KubeletConfig, err error, arg
 		if getErr != nil {
 			return getErr
 		}
-		// To avoid a long list of same statuses, only append a status if it is the first status
+		// Keeps a list of three status to avoid a long list of same statuses,
+		// only append a status if it is the first status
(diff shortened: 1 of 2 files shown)


QiWang19:
@umohnani8 @rphillips PTAL

umohnani8:
LGTM
/retest

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -346,16 +346,13 @@ func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.KubeletConfig, err error, arg
 		if getErr != nil {
 			return getErr
 		}
-		// To avoid a long list of same statuses, only append a status if it is the first status
+		// Keeps a list of three status to avoid a long list of same statuses,
```
kikisdeliveryservice, line 349: thanks for the clear comments here =)

kikisdeliveryservice (approved):
Thanks @QiWang19 Appreciate the tests as well !

Issue #1365, opened 2020-01-10
Capture and report systemd unit failures by default to the MCS


cgwalters:
We've debated this a few times - who watches host failures?  I think we should do something like have a small systemd unit `machine-config-daemon-host-monitor.service` which watches systemd and if any unit fails, does a `POST` to the MCS.

Yes this implies a MCS write endpoint, not just read.  We could start by just dumping the data into the pod logs.

This is also related to https://github.com/coreos/ignition/issues/585 which we'd also want a MCS endpoint for.

cgwalters:
Also related to this, something in the platform really must start watching for kernel oopses.

Also xref https://github.com/kubernetes/node-problem-detector

cgwalters:
Now that https://github.com/openshift/machine-config-operator/pull/1766/ landed - we can make this be part of the MCD itself quite easily.

cgwalters:
/lifecycle frozen
We really need this.  See e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1906057 where the real problem is lost in the black hole somewhere in Ignition or firstboot.

I'd been hoping we could reuse https://github.com/openshift/enhancements/pull/443 to authenticate the MCS endpoint for failure report submission.

PR #2539, opened 2021-04-20
Bug 1928581: validate the proxy by trying oc image info

QiWang19:
Try use skopeo inspect image using http proxy config for proxy validation. If the skopeo command fails, do not render the proxy.

**- What I did**

Close Bug 1928581 https://bugzilla.redhat.com/show_bug.cgi?id=1928581

**- How to verify it**

Add HTTP proxy validation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,10 @@ COPY --from=builder /go/src/github.com/openshift/machine-config-operator/instroo
 RUN cd / && tar xf /tmp/instroot.tar && rm -f /tmp/instroot.tar
 COPY install /manifests
 
+RUN dnf -y update && dnf -y reinstall shadow-utils && \
+dnf -y install skopeo && dnf -y install openshift-clients && \
+rm -rf /var/cache /var/log/dnf*
+
 RUN if [[ "${TAGS}" == "fcos" ]] || [[ "${TAGS}" == "scos" ]]; then \
     # comment out non-base/extensions image-references entirely for fcos/scos
     sed -i '/- name: rhel-coreos-8-/,+3 s/^/#/' /manifests/image-references && \
@@ -21,11 +25,6 @@ RUN if [[ "${TAGS}" == "fcos" ]] || [[ "${TAGS}" == "scos" ]]; then \
     if [[ "${TAGS}" == "fcos" ]]; then sed -i 's/rhel-coreos-8/fedora-coreos/g' /manifests/*; \
     elif [[ "${TAGS}" == "scos" ]]; then sed -i 's/rhel-coreos-8/centos-stream-coreos-9/g' /manifests/*; fi && \
     if ! rpm -q util-linux; then yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*; fi
-
-RUN dnf -y update && dnf -y reinstall shadow-utils && \
-dnf -y install skopeo && \
-rm -rf /var/cache /var/log/dnf*
-
 COPY templates /etc/mcc/templates
 ENTRYPOINT ["/usr/bin/machine-config-operator"]
 LABEL io.openshift.release.operator true
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,11 @@ RUN if [[ "${TAGS}" == "fcos" ]] || [[ "${TAGS}" == "scos" ]]; then \
     if [[ "${TAGS}" == "fcos" ]]; then sed -i 's/rhel-coreos-8/fedora-coreos/g' /manifests/*; \
     elif [[ "${TAGS}" == "scos" ]]; then sed -i 's/rhel-coreos-8/centos-stream-coreos-9/g' /manifests/*; fi && \
     if ! rpm -q util-linux; then yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*; fi
+
+RUN dnf -y update && dnf -y reinstall shadow-utils && \
+dnf -y install skopeo && \
+rm -rf /var/cache /var/log/dnf*
+
 COPY templates /etc/mcc/templates
 ENTRYPOINT ["/usr/bin/machine-config-operator"]
 LABEL io.openshift.release.operator true
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -1,11 +1,19 @@
 package v1
 
 import (
+	"context"
 	"fmt"
+	"os/exec"
+	"strings"
+	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// proxyTestImage will be skopeo inspect through http proxy if set for proxy validation
+const proxyTestImage = "quay.io/podman/stable"
 
 // NewMachineConfigPoolCondition creates a new MachineConfigPool condition.
 func NewMachineConfigPoolCondition(condType MachineConfigPoolConditionType, status corev1.ConditionStatus, reason, message string) *MachineConfigPoolCondition {
@@ -195,4 +203,47 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
 		return nil
 	}
 	return fmt.Errorf("ControllerConfig has not completed: completed(%v) running(%v) failing(%v)", completed, running, failing)
+}
+
+// ProxyValidation inspects the proxyTestImage to validate if the http proxy is valid.
+func ProxyValidation(proxy *configv1.ProxyStatus) error {
+	if proxy == nil {
+		return nil
+	}
+	return skopeoInspectUseProxy(proxyTestImage, proxy.HTTPProxy, proxy.HTTPSProxy)
+}
+
+// skopeoInspectUseProxy uses skopeo to inspect the proxyTestImage, through the http(s) proxy if exists.
+// returns error if the inspection fails.
+func skopeoInspectUseProxy(imageURL, httpProxy, httpsProxy string) error {
+	// proxy error from docker registry starts with proxyErr
+	proxyErr := "proxyconnect"
+	skopeo, err := exec.LookPath("skopeo")
+	if err != nil {
+		return err
+	}
+	imageSpec := fmt.Sprintf("docker://%s", imageURL)
+	var envs []string
+	if httpProxy != "" {
+		envs = append(envs, fmt.Sprintf("HTTP_PROXY=%s", httpProxy))
+	}
+	if httpsProxy != "" {
+		envs = append(envs, fmt.Sprintf("HTTPS_PROXY=%s", httpsProxy))
+	}
+
+	args := []string{"inspect", imageSpec}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, skopeo, args...)
+	cmd.Env = envs
+	rawOut, err := cmd.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(rawOut), proxyErr) {
+			return fmt.Errorf("invalid http proxy: %w: error running %s %s: %s", err, skopeo, strings.Join(args, " "), string(rawOut))
+		}
+		return fmt.Errorf("%w: error running %s %s: %s", err, skopeo, strings.Join(args, " "), string(rawOut))
+	}
+	return nil
 }
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
@@ -1,19 +1,11 @@
 package v1
 
 import (
-	"context"
 	"fmt"
-	"os/exec"
-	"strings"
-	"time"
 
-	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
-// proxyTestImage will be skopeo inspect through http proxy if set for proxy validation
-const proxyTestImage = "quay.io/podman/stable"
 
 // NewMachineConfigPoolCondition creates a new MachineConfigPool condition.
 func NewMachineConfigPoolCondition(condType MachineConfigPoolConditionType, status corev1.ConditionStatus, reason, message string) *MachineConfigPoolCondition {
@@ -203,47 +195,4 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
 		return nil
 	}
 	return fmt.Errorf("ControllerConfig has not completed: completed(%v) running(%v) failing(%v)", completed, running, failing)
-}
-
-// ProxyValidation inspects the proxyTestImage to validate if the http proxy is valid.
-func ProxyValidation(proxy *configv1.ProxyStatus) error {
-	if proxy == nil {
-		return nil
-	}
-	return skopeoInspectUseProxy(proxyTestImage, proxy.HTTPProxy, proxy.HTTPSProxy)
-}
-
-// skopeoInspectUseProxy uses skopeo to inspect the proxyTestImage, through the http(s) proxy if exists.
-// returns error if the inspection fails.
-func skopeoInspectUseProxy(imageURL, httpProxy, httpsProxy string) error {
-	// proxy error from docker registry starts with proxyErr
-	proxyErr := "proxyconnect"
-	skopeo, err := exec.LookPath("skopeo")
-	if err != nil {
-		return err
-	}
-	imageSpec := fmt.Sprintf("docker://%s", imageURL)
-	var envs []string
-	if httpProxy != "" {
-		envs = append(envs, fmt.Sprintf("HTTP_PROXY=%s", httpProxy))
-	}
-	if httpsProxy != "" {
-		envs = append(envs, fmt.Sprintf("HTTPS_PROXY=%s", httpsProxy))
-	}
-
-	args := []string{"inspect", imageSpec}
-
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	defer cancel()
-
-	cmd := exec.CommandContext(ctx, skopeo, args...)
-	cmd.Env = envs
-	rawOut, err := cmd.CombinedOutput()
-	if err != nil {
-		if strings.Contains(string(rawOut), proxyErr) {
-			return fmt.Errorf("invalid http proxy: %w: error running %s %s: %s", err, skopeo, strings.Join(args, " "), string(rawOut))
-		}
-		return fmt.Errorf("%w: error running %s %s: %s", err, skopeo, strings.Join(args, " "), string(rawOut))
-	}
-	return nil
 }
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
@@ -3,9 +3,11 @@ package v1
 import (
 	"errors"
 	"fmt"
+	"os/exec"
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -178,4 +180,15 @@ func TestIsControllerConfigCompleted(t *testing.T) {
 			}
 		})
 	}
+}
+
+// TestSkopeoInspectUseProxy tests the proxyTestImage can be inspected by skopeo successfully without a proxy
+// This does not test some proxy. If skopeoInspectUseProxy failed, it would caused by the proxy not the function itself.
+func TestInspectUseProxy(t *testing.T) {
+	_, err := exec.LookPath("skopeo")
+	if err != nil {
+		t.Skip("skip the test since skopeo not available")
+	}
+	err = skopeoInspectUseProxy(proxyTestImage, "", "")
+	require.NoError(t, err)
 }
--- a/pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
+++ b/pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
@@ -3,11 +3,9 @@ package v1
 import (
 	"errors"
 	"fmt"
-	"os/exec"
 	"reflect"
 	"testing"
 
-	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -180,15 +178,4 @@ func TestIsControllerConfigCompleted(t *testing.T) {
 			}
 		})
 	}
-}
-
-// TestSkopeoInspectUseProxy tests the proxyTestImage can be inspected by skopeo successfully without a proxy
-// This does not test some proxy. If skopeoInspectUseProxy failed, it would caused by the proxy not the function itself.
-func TestInspectUseProxy(t *testing.T) {
-	_, err := exec.LookPath("skopeo")
-	if err != nil {
-		t.Skip("skip the test since skopeo not available")
-	}
-	err = skopeoInspectUseProxy(proxyTestImage, "", "")
-	require.NoError(t, err)
 }
--- a/pkg/controller/bootstrap/bootstrap.go
+++ b/pkg/controller/bootstrap/bootstrap.go
@@ -182,8 +182,11 @@ func (b *Bootstrap) Run(destDir string) error {
 		}
 		configs = append(configs, kconfigs...)
 	}
-	if err := mcfgv1.ProxyValidation(cconfig.Spec.Proxy); err != nil {
-		return err
+
+	if releaseVersion, ok := cconfig.Annotations[ctrlcommon.ReleaseImageVersionAnnotationKey]; ok {
+		if err := ctrlcommon.ProxyValidation(cconfig.Spec.Proxy, releaseVersion, icspRules); err != nil {
+			return err
+		}
 	}
 
 	fpools, gconfigs, err := render.RunBootstrap(pools, configs, cconfig)
--- a/pkg/controller/bootstrap/bootstrap.go
+++ b/pkg/controller/bootstrap/bootstrap.go
@@ -182,6 +182,9 @@ func (b *Bootstrap) Run(destDir string) error {
 		}
 		configs = append(configs, kconfigs...)
 	}
+	if err := mcfgv1.ProxyValidation(cconfig.Spec.Proxy); err != nil {
+		return err
+	}
 
 	fpools, gconfigs, err := render.RunBootstrap(pools, configs, cconfig)
 	if err != nil {
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -14,11 +14,16 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"os/exec"
 	"reflect"
 	"sort"
 	"strings"
+	"time"
+
+	runtimeutils "github.com/openshift/runtime-utils/pkg/registries"
 
 	"github.com/clarketm/json"
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	fcctbase "github.com/coreos/fcct/base/v0_1"
 	"github.com/coreos/ign-converter/translate/v23tov30"
 	"github.com/coreos/ign-converter/translate/v32tov22"
@@ -39,6 +44,8 @@ import (
 	validate3 "github.com/coreos/ignition/v2/config/validate"
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
+	apicfgv1 "github.com/openshift/api/config/v1"
+	apioperatorsv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"github.com/vincent-petithory/dataurl"
 	kerr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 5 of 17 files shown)


pkg/controller/bootstrap/bootstrap.go
```diff
@@ -133,6 +135,21 @@ func (b *Bootstrap) Run(destDir string) error {
...
 
+	if cconfig.Spec.Proxy != nil && cconfig.Spec.Proxy.HTTPProxy != "" {
+		if err := os.Setenv("http_proxy", "cconfig.Spec.Proxy.HTTPProxy"); err == nil {
+			if err := podmanCreateUseProxy("k8s.gcr.io/pause:3.1"); err != nil {
+				return err
+			}
+		}
+	}
+	if cconfig.Spec.Proxy != nil && cconfig.Spec.Proxy.HTTPSProxy != "" {
+		if err := os.Setenv("https_proxy", "cconfig.Spec.Proxy.HTTPProxy"); err == nil {
```
rphillips, line 146: I think you will want to refactor this so the http or https proxy get set prior to running the podmanCreateUseProxy... The environment variables should be HTTP_PROXY and HTTPS_PROXY.

Make sure to remember to remove the quotes around cconffig.Spec>proxy.HTTPProxy

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -234,3 +251,14 @@ func parseManifests(filename string, r io.Reader) ([]manifest, error) {
 		manifests = append(manifests, m)
 	}
 }
+
+func podmanCreateUseProxy(imageURL string) error {
+	command := "podman"
+	args := []string{"create", "--http-proxy", imageURL}
```
rphillips, line 257: should be a podman pull...

add the environment variables from above to the cmd context here...

cmd.Env = append("HTTP_PROXY=something", "HTTPS_PROXY=something_else")

pkg/controller/bootstrap/bootstrap.go
rphillips: These environment variables get pushed into the cmd.Env mentioned below

sgreene570:
What about proxy config changes made as a day 2 operator? Does the podman "test pull" code in this PR verify that image pulls work when modifying the cluster proxy config after installation?

sinnykumari:
Considering that proxy config is a global config in the cluster, shouldn't this be fixed at source i.e checking for proxy config when it gets applied/updated to the cluster? Consumers like MCO comes into picture later on.

QiWang19:
> Considering that proxy config is a global config in the cluster, shouldn't this be fixed at source i.e checking for proxy config when it gets applied/updated to the cluster? Consumers like MCO comes into picture later on.

@sinnykumari  From the Bugzilla discussion, when applying to the cluster the proxy will be validated by CNO. [comment 17](https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c17).  Is it the not a MCO change.
or could you give me a pointer where would be a proper place to validate it, looks like? https://github.com/openshift/machine-config-operator/blob/01f5f79da605e854d320ba2feeac093f6673cbe9/lib/resourceapply/machineconfig.go#L14

sinnykumari:
> @sinnykumari From the Bugzilla discussion, when applying to the cluster the proxy will be validated by CNO. [comment 17](https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c17). It is the not a MCO change.

ah ok, I may have been confused then because this PR is making changes to MCO's MCC bootstrap mode. If CNO is going to validate the proxy (which makes sense to me), shouldn't this validation code be in the CNO repo?

QiWang19:
If the bootstrap mode has an invalid proxy, the CNO pod will fail to launch since the iCNO images cannot be pulled down.

QiWang19:
> > @sinnykumari From the Bugzilla discussion, when applying to the cluster the proxy will be validated by CNO. [comment 17](https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c17). It is the not a MCO change.
> 
> ah ok, I may have been confused then because this PR is making changes to MCO's MCC bootstrap mode. If CNO is going to validate the proxy (which makes sense to me), shouldn't this validation code be in the CNO repo?

@sinnykumari PTAL. Add validation for places after the installation.

QiWang19:
The error `executable file not found in $PATH` does not appear in the ci, the skopeo has been installed in MCO image.

QiWang19:
which username:password should I use here for registry.redhat.io? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2539/pull-ci-openshift-machine-config-operator-master-e2e-aws-workers-rhel7/1466809754041978880/artifacts/e2e-aws-workers-rhel7/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-operator-84b4cf549b-d4ldg_machine-config-operator.log
```
error running /usr/bin/skopeo inspect docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458: time="2021-12-03T17:06:54Z" level=fatal msg="Error parsing image name \"docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458\": unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication"
```

kikisdeliveryservice:
> which username:password should I use here for registry.redhat.io? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2539/pull-ci-openshift-machine-config-operator-master-e2e-aws-workers-rhel7/1466809754041978880/artifacts/e2e-aws-workers-rhel7/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-operator-84b4cf549b-d4ldg_machine-config-operator.log
> 
> ```
> error running /usr/bin/skopeo inspect docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458: time="2021-12-03T17:06:54Z" level=fatal msg="Error parsing image name \"docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458\": unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication"
> ```

@QiWang19 did you get this question answered?

QiWang19:
> > which username:password should I use here for registry.redhat.io? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2539/pull-ci-openshift-machine-config-operator-master-e2e-aws-workers-rhel7/1466809754041978880/artifacts/e2e-aws-workers-rhel7/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-operator-84b4cf549b-d4ldg_machine-config-operator.log
> > ```
> > error running /usr/bin/skopeo inspect docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458: time="2021-12-03T17:06:54Z" level=fatal msg="Error parsing image name \"docker://registry.redhat.io/openshift4/ose-openshift-proxy-pull-test-rhel8@sha256:438b3fb47bcf03e9f712a337fc992c690083193164cc0df54cea9098b91bf458\": unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication"
> > ```
> 
> @QiWang19 did you get this question answered?

The rest platform does not have the credentials for registry.redhat.io. We changed to use a public image from quay.io that does not require credentials.

QiWang19:
@rphillips PTAL. Do we want to get this in? Installed skopeo to MCO and inspect image from quay to avoid credential issue.

Dockerfile
```diff
@@ -10,6 +10,11 @@ COPY --from=builder /go/src/github.com/openshift/machine-config-operator/instroo
 RUN cd / && tar xf /tmp/instroot.tar && rm -f /tmp/instroot.tar
 COPY install /manifests
 RUN if ! rpm -q util-linux; then yum install -y util-linux && yum clean all && rm -rf /var/cache/yum/*; fi
+
+RUN dnf -y update; dnf -y reinstall shadow-utils; \
```
rphillips, line 14: semicolons should be && so there are not hidden failures.

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -196,3 +202,40 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
 	}
 	return fmt.Errorf("ControllerConfig has not completed: completed(%v) running(%v) failing(%v)", completed, running, failing)
 }
+
+// ProxyInspectTest inspects the proxyTestImage to test if the http proxy is valid.
+func ProxyInspectTest(proxy *configv1.ProxyStatus) error {
+	if proxy == nil {
+		return nil
+	}
+	if err := skopeoInspectUseProxy(proxyTestImage, proxy.HTTPProxy, proxy.HTTPSProxy); err != nil {
```
rphillips, line 211: this if statement is the same as just returning

```
return skopeoInspectUseProxy(proxyTestImage, proxy.HTTPProxy, proxy.HTTPSProxy)
```

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -196,3 +202,40 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
...
+	}
+	if httpsProxy != "" {
+		envs = append(envs, fmt.Sprintf("HTTPS_PROXY=%s", httpsProxy))
+	}
+
+	args := []string{"inspect", imageSpec}
+	cmd := exec.Command(skopeo, args...)
+	cmd.Env = envs
+	rawOut, err := cmd.CombinedOutput()
+	if err != nil && strings.Contains(string(rawOut), proxyErr) {
```
rphillips, line 237: this if statement should be split up because err might be thrown for a different reason.

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -196,3 +202,40 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
...
+	var envs []string
+	if httpProxy != "" {
+		envs = append(envs, fmt.Sprintf("HTTP_PROXY=%s", httpProxy))
+	}
+	if httpsProxy != "" {
+		envs = append(envs, fmt.Sprintf("HTTPS_PROXY=%s", httpsProxy))
+	}
+
+	args := []string{"inspect", imageSpec}
+	cmd := exec.Command(skopeo, args...)
```
rphillips, line 234: we need to add a context here so it times out the skopeo command as well.

QiWang19:
@kikisdeliveryservice @yuqi-zhang Could you add approved label?

pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
```diff
@@ -179,3 +181,14 @@ func TestIsControllerConfigCompleted(t *testing.T) {
 		})
 	}
 }
+
+// TestSkopeoInspectUseProxy tests the proxyTestImage can be inspected by skopeo successfully
+// This does not test the proxy
```
kikisdeliveryservice, line 186: I'm confused by the line "this test does not test the proxy" what is the expect flow in this test?

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -196,3 +204,44 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
 	}
 	return fmt.Errorf("ControllerConfig has not completed: completed(%v) running(%v) failing(%v)", completed, running, failing)
 }
+
+// ProxyInspectTest inspects the proxyTestImage to test if the http proxy is valid.
+func ProxyInspectTest(proxy *configv1.ProxyStatus) error {
```
kikisdeliveryservice, line 209: Generally things with Test in the name are tests. Can you rename to show what this check is doing?

pkg/apis/machineconfiguration.openshift.io/v1/helpers.go
```diff
@@ -196,3 +204,44 @@ func IsControllerConfigCompleted(ccName string, ccGetter func(string) (*Controll
...
+	if proxy == nil {
+		return nil
+	}
+	return skopeoInspectUseProxy(proxyTestImage, proxy.HTTPProxy, proxy.HTTPSProxy)
+}
+
```
kikisdeliveryservice, line 215: add func description

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -182,6 +182,9 @@ func (b *Bootstrap) Run(destDir string) error {
 		}
 		configs = append(configs, kconfigs...)
 	}
+	if err := mcfgv1.ProxyInspectTest(cconfig.Spec.Proxy); err != nil {
```
kikisdeliveryservice, line 185: Below you check if proxy != nil but here you don't?

pkg/apis/machineconfiguration.openshift.io/v1/helpers_test.go
QiWang19: The test confirmed the image inspect function is working correctly without proxy.  If skopeoInspectUseProxy failed, it would caused by the proxy not the function itself.

kikisdeliveryservice: thanks qi, can you update the test description

pkg/controller/bootstrap/bootstrap.go
QiWang19: proxy != nil this was checked inside the function body ProxyInspectTest

QiWang19:
@kikisdeliveryservice I cleaned up the reviews. PTAL.

yuqi-zhang (changes requested):
Sorry I didn't see this earlier. Just re-read the bugzilla and this is an interesting issue. To help think through this, I have a few questions:

1. I see you are installing skopeo (the binary) into the MCO image, and then invoking skopeo via the command line to check for a string (`proxyconnect`) to see if the proxy has connection errors (or other errors). Should we perhaps be vendoring skopeo directly instead?
2. The idea here is that if a proxy is detected, instead of rendering it into the renderconfig (thus the MCC), we first check if the proxy can be connected to. If not, we don't render the proxy at all and fail at the MCO level. Do you know what the error would manifest as?
3. I'm tentative with the approach generally, and I think that as a main concern, I'm not sure if our definition of "is a proxy correct" as "can we pull quay.io/podman/stable" is the correct assumption. e.g. could there be environments that set up a proxy for disconnected, that don't have access to quay, but have a proxy for internal usage? What if the user blocks quay as part of their registries configuration?

I think this may not be a great idea for all environments since we could be rejected existing user setups that work already, I guess as my main concern

QiWang19:
we need an executable binary to inspect the image, it can not be done by vendor. The test machine of CI running mco image does not have skopeo or podman installed and the ci test can failed. 

If the proxy does not allow access to quay, shouldn't we consider the proxy invalid? From the [discussion](https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c8), the cluster must be able to pull images from quay, otherwise, it would cause the CNO Bug.
quay is the registry used by payload, it is not allowed to block it.

yuqi-zhang:
> we need an executable binary to inspect the image, it can not be done by vendor

Ah ok, I thought skopeo may have had a library of some sort we could use to do this

> If the proxy does not allow access to quay, shouldn't we consider the proxy invalid? From the [discussion](https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c8), the cluster must be able to pull images from quay, otherwise, it would cause the CNO Bug.
quay is the registry used by payload, it is not allowed to block it.

I'm not sure on this either, so my concern may be invalid. I was thinking of disconnected environments, etc.
https://docs.openshift.com/container-platform/4.10/installing/disconnected_install/installing-mirroring-installation-images.html

But maybe you can't use a proxy in a disconnected environment? Would there be non-disconnected environments where the user doesn't actually want to pull from quay but their own registry? Is that allowed?

If we do have somewhere that explicitly says, if you don't have access to quay, your cluster is invalid, then I am fine with this. I just want to make sure that if there is a user out there with proxy+no quay access, we don't break them

yuqi-zhang:
> @yuqi-zhang What if we added a boolean on the controller config (like: enableProxyValidation, defaulted: false). We enable the check when the boolean is true, and document the risks and image registry that is used to validate the proxy.

I think that would be a safer approach, so I would be +1 (not necessarily on the controllerconfig since I think the user is not supposed to modify that directly, maybe as part of... proxy object? MCO pod options?).

Again, I am not necessarily against the current approach either, its just that I would like to make sure we check for all the cases so we don't break existing user setups that work but don't necessarily have access to pull from quay, if that exists

kikisdeliveryservice:
> I think that would be a safer approach, so I would be +1 (not necessarily on the controllerconfig since I think the user is not supposed to modify that directly, maybe as part of... proxy object? MCO pod options?).

+1 not on controllerconfig as that's not something that users should be directly modifying.

palonsoro:
@yuqi-zhang The way to do the check correctly. even for the (extremely exotic) that somebody uses a proxy in a disconnected environment would be to try to access one of the release images after applying image content source policies (which say "when you go to `quay.io`, actually go to `myregistry.internal`).

Although it may sound a bit complex, I'd suggest the following check if possible:
- Get the release image (from clusterversion if possible).
- From release image, get the CNO image (in a similar way to `oc adm release info $blahblah --image-for=$blahblah`)
- Then, apply any image content source policy that may apply to it.
- Then do the proxy check with the final result.

I know this process looks a bit of an overkill, but I see it covering some interesting points:
- It takes into account a disconnected installations when quay.io pullspecs are used with ICSPs that redirect to a mirror, with the proxy.
- It takes into account a disconnected installation when the pullspecs do resemble your mirror directly.
- CNO image is the best test because that's the image we need always to pull. Reason is as simple as: If we can pull CNO image, it does not matter that much that we cannot pull other images, because we can fix the proxy settings. If we cannot pull CNO image, we won't be able to fix the proxy settings.

yuqi-zhang:
> Although it may sound a bit complex, I'd suggest the following check if possible:
> - Get the release image (from clusterversion if possible).
> - From release image, get the CNO image (in a similar way to oc adm release info $blahblah --image-for=$blahblah)
> - Then, apply any image content source policy that may apply to it.
> - Then do the proxy check with the final result.

This makes more sense to me, although this does sound... quite complex, and I'm hesitant on whether this degree of network checking should belong more to the MCO. It feels to me that this level of network configuration (proxy) correctness should be done at a higher level, and not in the MCO before we render the proxy into each node's config.

But I think this would once again go back to the original BZ and get it passed around. I don't have a very good picture in mind on what the final flow should look like at the moment. If there are any discussions for this, I'd be happy to chat. At the end of the day, I think it definitely would be good to have some more safeguards for the user. I am just not sure the best way to achieve that

palonsoro:
IMHO, this should have been done in CNO while reading the proxy spec and before writing the status. But I have the feeling that I am missing something.

cgwalters:
A key safeguard we have by default is that only one node is updated at a time, so you will have at most one node down.

I'm (weakly) against adding lots of special case checks for this.  There are a *ton* of ways to break things, and I think we will get much more out of solving the generic issues:

- Ensuring we [capture and report runtime errors](https://github.com/openshift/machine-config-operator/issues/1365)
- Helping users understand that in MachineAPI managed scenarios (and for workers mainly), to recover they can just delete the machine object and let the node be reprovisioned
- Ensuring it's easy to test changes outside of a running cluster node (covered by [layering](https://github.com/openshift/enhancements/blob/master/enhancements/ocp-coreos-layering.md))
- Try unifying with greenboot and similar approaches that do automatic rollback

palonsoro:
@cgwalters I am not sure to understand your point with this comment.

The key thing with this issue is that the cluster can become blocked forever and you will become completely unable to fix the proxy configuration cluster-wide. This is not something that can be fixed or workarounded by killing a single node.

I would agree that MCO may be a bit of a surprising place to put the fix. When I started the bug, I was thinking more of the CNO validating the settings before asking the MCO to roll them out. But now that we are here, we do need this to be fully checked somewhere.

palonsoro:
@cgwalters I am not sure to understand your point with this comment.

The key thing with this issue is that the cluster can become blocked forever and you will become completely unable to fix the proxy configuration cluster-wide. This is not something that can be fixed or workarounded by killing a single node.

I would agree that MCO may be a bit of a surprising place to put the fix. When I started the bug, I was thinking more of the CNO validating the settings before asking the MCO to roll them out. But now that we are here, we do need this to be fully checked somewhere.

QiWang19:
@yuqi-zhang @palonsoro  Could you review? The new commit can install openshift-client and exec oc command to fetch CNO image pull spec.

palonsoro:
@QiWang19 it looks good to me. Thanks!

Dockerfile
```diff
@@ -12,6 +12,10 @@ COPY --from=builder /go/src/github.com/openshift/machine-config-operator/instroo
 RUN cd / && tar xf /tmp/instroot.tar && rm -f /tmp/instroot.tar
 COPY install /manifests
 
+RUN dnf -y update && dnf -y reinstall shadow-utils && \
+dnf -y install skopeo && dnf -y install openshift-clients && \
```
cgwalters, line 16: It's a bit unfortunate to add a whole new copy of skopeo and `oc` into the image, because we have them right there on the host too...
(And for that matter, it's actually useful to validate the proxy configuration *from the host network namespace* since that's where most image pulling will be happening)
Tricky to deal with without making the MCC privileged enough for host mounts though.  But OTOH, the MCC really is privileged in a cluster sense anyways, so making it a privileged container (at least enough for host mounts) isn't really adding any new attack surface.

yuqi-zhang:
Sorry for the delay. I think I would need to check in with the team on this as a more general discussion on how to proceed.

Dockerfile
palonsoro: @cgwalters I agree with the host network part, you made a great point here, that may be something worth considering.

However, regarding including the binaries, that's a usual burden that we are already paying a lot of times in a lot of components, nothing that should be surprising to us if we have OCP4 consisting of a number of clusteroperators which deploy a number of operands, almost all of them running inside containers. 

Making the MCC deployment require access to the host and require the host to always have these binaries available, even when not crazy in practice, goes much against the spirit of having every component in a container so it is, well, self-contained.

A possible way of improvement here would be to make MCO image derive from the `tools` image shipped in the release instead of `base`, because the `tools` image includes the correct version of `oc` already. Other images that require `oc` already do it and benefit from image layer de-duplication in what regards the oc client.

cgwalters: > A possible way of improvement here would be to make MCO image derive from the tools image shipped in the release instead of base, because the tools image includes the correct version of oc already.

Yeah, that'd help, but doesn't get us out of also shipping skopeo, which today also vendors large parts of the container runtime again.

Hmm, do we actually need to use skopeo vs just forking `oc`?

palonsoro: Yes, we do. `oc` is only used to find out which image must be checked (CNO image). Once found, the check is done with `skopeo` (`oc` doesn't provide a way to do it).

cgwalters: Wouldn't just running e.g. `oc image info` be sufficient?

palonsoro: mmm, as long as executing this successfully guarantees a successful pull, it would.

cgwalters: `oc` today vendors the docker Go library for interacting with registries, whereas skopeo uses the github.com/containers/image bits.  But ultimately...I can't imagine a case where one worked but not the other.

Today `oc`'s fetching is kind of load-bearing because it's where we have e.g. `oc image mirror` etc that people use for disconnected.

I can't imagine a case where `oc` succeeds but `skopeo` (and/or podman/cri-o) would fail with respect to the proxy.

QiWang19: I agree, we can just run `oc image info`

pkg/operator/sync.go
```diff
@@ -301,6 +303,25 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
...
+		return err
+	}
+	if proxy != nil && clusterversionCfg != nil {
+		icspRules, err := optr.icspLister.List(labels.Everything())
+		if err != nil && errors.IsNotFound(err) {
+			icspRules = []*apioperatorsv1alpha1.ImageContentSourcePolicy{}
+		} else if err != nil {
+			return err
+		}
+		if err := ctrlcommon.ProxyValidation(&proxy.Status, clusterversionCfg.Status.Desired.Version, icspRules); err != nil {
```
jkyros, line 321: I feel like you could talk me into putting this in a separate place for "sanity checks" before config rollout/eviction/drain, but I have concerns about the proposed location in `SyncRenderedconfig()` -- if the proxyconfig fails this test for whatever reason, we don't get a `RenderConfig`, which prevents the rest of the sync functions from running, and that is problematic for general cluster stability during normal operations (among other things, it affects certificate rotations). 

Specifically, in a case where the proxy was "valid" when it was configured, but is down/unreachable/etc at the time of the check, the MCO would degrade, wouldn't it? 

I don't know that I have a spot picked out where it *should* go, because the MCO has typically not accepted preflight checks of this nature, but I'm kind of sympathetic here   :smile:

QiWang19: @yuqi-zhang could you help locate where MCO deploys the proxy settings to the Nodes so that the proxyvalidatoin can be done there?

yuqi-zhang: Right, so the exact details are a bit up to debate. The way it's set up in this PR right now is blocking at the `operator` level which is potentially dangerous for reasons John has listed, and I agree that we should probably think about moving this to a consolidated "checking" location.

So, if we want to achieve the point of "don't roll out the proxy to nodes unless it *most likely* works", then it likely will have to happen at the controller level, between:
1. https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/render/render_controller.go#L546, where the rendered MC is generated for a pool
2. https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L846, where the config is rolled out to the MCP

So then this would be something like `validateIncomingRenderedConfig` before it gets rolled out to the pool, somehow, where right now we just validate the proxy, but could be extended as a node-specific pre-flight check of some sort. We could even have a flag that enables/disables this, if we don't want to change default behaviour.

The other side of this issue is, as we move towards layering, what if I built a new format OS image with a proxy built into it somehow? This validation path would not catch that if done directly in the image.

One last extension thought on validation that's a bit more encompassing: have a (flag enabled?) option to create an extra canary node on incoming updates to see if that node would break, before upgrading the rest of the nodes. That's a bit too far though.

In summary, I think this is a pretty complex topic. Right now I think maybe the safest option is at the controller level, but how exactly that would be done is a bit up in the air

cgwalters: > The other side of this issue is, as we move towards layering, what if I built a new format OS image with a proxy built into it somehow? This validation path would not catch that if done directly in the image.

Ultimately I think what we want is https://github.com/ostreedev/ostree/issues/2725 - basically, we try booting the new configuration, and roll back if kubelet isn't able to start.

QiWang19: @yuqi-zhang Please review, need help with the operator code auto-generation regarding to the `func (f *fixture) newController()` (pkg/controller/render/render_controller_test.go)

pkg/controller/bootstrap/bootstrap.go
```diff
@@ -183,6 +183,12 @@ func (b *Bootstrap) Run(destDir string) error {
 		configs = append(configs, kconfigs...)
 	}
 
+	if releaseVersion, ok := cconfig.Annotations[ctrlcommon.ReleaseImageVersionAnnotationKey]; ok {
+		if err := ctrlcommon.ProxyValidation(cconfig.Spec.Proxy, releaseVersion, icspRules); err != nil {
```
yuqi-zhang, line 187: Hmm, just for my own curiosity, this will check via the bootstrap network on the bootstrap node right?

Is there a possibility that the bootstrap network is different? Does it even use the proxy you provide to the cluster?

pkg/controller/common/helpers.go
```diff
@@ -1001,3 +1008,144 @@ func GetDefaultBaseImageContainer(cconfigspec *mcfgv1.ControllerConfigSpec) stri
...
+	}
+
+	return ocImageInfoUseProxy(oc, cnoImg, proxy.HTTPProxy, proxy.HTTPSProxy, icspfile)
+}
+
+func cnoImageSpec(oc, releaseVersion, icspfile string) (string, error) {
+	const (
+		tagName           = "cluster-network-operator"
+		imageInfo         = "adm release info %s --image-for %s"
+		imageInfoWithICSP = "adm release info %s --image-for %s --icsp-file %s"
```
yuqi-zhang, line 1045: Could you explain how the ICSP affects the imagespec here?

pkg/operator/sync.go
```diff
@@ -301,6 +303,25 @@ func (optr *Operator) syncRenderConfig(_ *renderConfig) error {
 		}
 	}
 	spec.AdditionalTrustBundle = trustBundle
+	clusterversionCfg, err := optr.configClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
```
yuqi-zhang, line 306: We are looking to remove these changes right?

pkg/controller/render/render_controller.go
```diff
@@ -482,6 +491,18 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
...
 
+	icspRules, err := ctrl.icspLister.List(labels.Everything())
+	if err != nil && errors.IsNotFound(err) {
+		icspRules = []*apioperatorsv1alpha1.ImageContentSourcePolicy{}
+	} else if err != nil {
+		return err
+	}
+	if releaseVersion, ok := cc.Annotations[ctrlcommon.ReleaseImageVersionAnnotationKey]; ok {
+		if err := ctrlcommon.ProxyValidation(cc.Spec.Proxy, releaseVersion, icspRules); err != nil {
+			return err
```
yuqi-zhang, line 502: so if we err out here, I think we just don't generate the rendered config right? I feel like maybe we should still generate the rendered config, the have the node controller do the validation and fail there, so we can reference which rendered MC is failing

yuqi-zhang: Put another way, we still would have an issue where the rendered config doesn't get generated if we do it here, I think?

pkg/controller/common/helpers.go
```diff
@@ -1001,3 +1008,144 @@ func GetDefaultBaseImageContainer(cconfigspec *mcfgv1.ControllerConfigSpec) stri
...
+	defer cancel()
+
+	cmdExec := exec.CommandContext(ctx, oc, args...)
+	cmdExec.Env = envs
+	rawOut, err := cmdExec.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(rawOut), proxyErr) {
+			return fmt.Errorf("invalid http proxy: %w: error running %s %s: %s", err, oc, strings.Join(args, " "), string(rawOut))
+		}
+		return fmt.Errorf("%w: error running %s %s: %s", err, oc, strings.Join(args, " "), string(rawOut))
```
yuqi-zhang, line 1148: Should we have some kind of retry here, for transient failures?

I guess we always retry via re-syncing technically, maybe it would be worth adding a requeue somewhere...

The issue I'm thinking about is, let's say the network is unstable, and we happen to fail the one validation, but the proxy is otherwise valid, what is the user experience there?

pkg/operator/sync.go
yuqi-zhang: > need help with the operator code auto-generation regarding to the func (f *fixture) newController() (pkg/controller/render/render_controller_test.go)

Sorry, I don't quite follow, what is the issue with the code autogen?

pkg/controller/common/helpers.go
QiWang19: the oc command can accept the icsp file as an alternative source to retrieve the release image. https://github.com/openshift/oc/blob/3cdf3c29f0c109c94eb67124548a6b21fc5f6a22/pkg/cli/admin/release/info.go#L136. 
If the icsp has been configured on the cluster I think we should pass them to the oc to get the release image and get the cno pull spec.

QiWang19:
> Sorry, I don't quite follow, what is the issue with the code autogen?

the function signature change made for getting icsp https://github.com/openshift/machine-config-operator/pull/2539/files#diff-d38c494535eacf2f0876136ce2b6a6329c78e91d238f7cb2b8f75379427747c0R80
So in the test we need to match the arguments  in the call,  the arguments for this function are auto-generated by informer-gen,  https://github.com/QiWang19/machine-config-operator/blob/a49df6a2bcf2803f77aff5c2247d549fbdc62fff/pkg/controller/render/render_controller_test.go#L69
I have run `make update` but it did not generate.

pkg/controller/common/helpers.go
QiWang19: yes, I agree. we can retry to deal with the risks.

pkg/controller/render/render_controller.go
QiWang19: yes, that error was out before the render config was generated and synced.
To let the node controller do the check, we can drop the validation from render_conller, and in the node controller we add validation before this line https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L846,  something like:
```go
cconfigs, err := ctrl.ccLister.List(labels.Everything())
for _, cc := range cconfigs {
	retry(validation)
}
```
what do you think?

yuqi-zhang: I think somewhere in the sync MCP function could work.

Although, hmm, this does mean that every time we perform an update of any sort, for every node that gets synced, we re-check the proxy, and even for scenarios that don't have any changes to the proxy, we re-validate, which seems like... a lot of unnecessary work.

So in that case, maybe having it as we do now is better, but only validate on a change to the proxy between old and new?

What do you think @jkyros ? I'm leaning towards reducing the # of times we validate if there isn't a change in the proxy, just not sure where the best place to do so would be. From a logic perspective, I think maybe render controller is easier, but comes with the downside of not generating a new rendered MC.

I think in my view, the best place would be, after we generate the rendered MC, before we roll out to a MCP, we do a **one time check** if the current->desired MC contains a proxy change, before allowing the node controller to roll out. Such that if there is an error, the user would see a new rendered MC, but the MCP not start an upgrade due to it being degraded on checking for proxy (with a certain amount of retries). But I don't know how well that plugs into what we have now without it either 1. looking clunkly or 2. adding a whole new interface of some sort to do so

yuqi-zhang:
> I have run make update but it did not generate.

Hmm, it's been a long time since we last updated that test.

What happens if you just try to manually update the test function with the additional necessary items?

i.e. a `f.operatorClient = fakeoperatorclient.NewSimpleClientset(f.operatorObjects...)` -> `oi := operatorinformer.NewSharedInformerFactory(f.operatorClient, noResyncPeriodFunc())` -> `oi.Operator().V1alpha1().ImageContentSourcePolicies(),` like we do for e.g. containerruntimeconfigcontroller?

yuqi-zhang:
I've spent some time thinking through this general problem (validation of the configuration on the nodes), and I'd like to bring this up more for general discussion.

First, I'd like to go back to the MCO mission statement:
```
The MCO keeps the underlying CoreOS system up-to-date and applies configs. The team chooses:
Simplicity over sophistication
Being OPEN and Transparent over being opinionated
Verbosity and clarity over imposed safety
Empowering debugging over preventing bugs
```

The MCO was never designed to be, and I believe should still not be, the place where we provide `imposed safety`. The MCO does not check whether your configuration is correct today (other than syntax); instead it is simply the bridge between your configuration and the nodes. If we wanted to ensure configuration safety, there is simply too large of a matrix to ensure every configuration you provide is "safe", and the validation complexity will only increase as we move towards CoreOS Layering and providing image based updates. Simply put, this is a tradeoff we have made.

Side note: there is a sort of mitigation in place for "breaking updates": we roll out changes one node at a time (generally), and any singular node should always be replaceable. 

Back to the point of this PR, proxy has always been a contentious issue. Fundamentally, the proxy object is not owned by the MCO. If any validation were to happen, that the root object owner should be making validations to the object changes before it is provided to the cluster for consumption. If a user provides a broken proxy, shouldn't the change be rejected in the first place? Instead of having it get all the way to the MCO generating a new config before saying: actually your proxy changes aren't valid because the MCC container can't pull the CNO image. The CNO could have done that before we even got here, reducing the complexity of transit. The MCO would then react to it https://github.com/openshift/machine-config-operator/blob/69933a3dccd6f78a5b9fe09d5b3830c9c8e39a03/pkg/operator/render.go#L175 and not lay down the "bad proxy"

Side note again: more broadly speaking, `validation at source` doesn't cover some other cases we've ran into in the past, such as important secrets/certs. etc. being deleted. Some of these objects are created during install time and never "managed", and the MCO simply consumes them.

I would also like to revisit the bug for a moment: up until this point https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c12 we were still discussing how to properly validate at a CNO level, but right after we did some component switching for which there is no context for in the bug. Trevor makes some good points in https://bugzilla.redhat.com/show_bug.cgi?id=1928581#c17 and then we flipped it back to node. Did we ever get a chance to discuss this at a higher level?

Now, to also look at the flip slide, "MCO does not do validation" is not a view that cannot be changed. Openshift is constantly growing and adapting, such that if there is sufficient need to tackle a problem, I think we should consider it. As I see it, there are a few alternatives floating in mind:
1. the `MCO does validation as we see fit` approach: this would be basically this PR: we add validations for issues that are seen a lot and are annoying to deal with, and we scatter it across the code. This obviously does not scale well, but can solve some immediate issues
2. the `MCO creates a new validation schema/API/controller` approach: we would spend time designing and crafting a whole new method (controller?) that allows us to create extendable methods to validate configuration, with potential options to allow users to specify extra validation schema. This would probably take more design and I am not sure what's the best way to do so today
3. the `root owner takes responsibility` approach, where the validation happens at a higher level before it reaches the MCO, and the MCO continues to be a consumer. This would likely also work better with layering, since the validation would happen pre-build
4. the `validate coreos layered images` approach, where we create a new image validation schema specifically for the new layered image update workflow
5. the `create OCP config validator` flow, where a new operator is generally responsible for watching important configuration changes 

And lastly, I feel bad for this writeup, since many people have put a lot of work into this PR, but after all the back and forth I am still leaning towards "this isn't something we should do in the MCO". I am happy to discuss this further in any context, and I am willing to change my mind.

cgwalters:
Excellent writeup, I agree with most of it.  My view on this is what we really want is automated rollbacks.  Basically in this scenario:

- node boots into new config
- we fail to contact the proxy (I don't think kubelet fails in this scenario, but we can't fetch OS updates anymore?  Presumably other pod workloads on the node fail)
- This is detected by a health check
- Roll back to previous config
- Error from previous state is saved and reported

A question here is whether we then later try to reconcile again later.  I think it'd make sense to do so, with a backoff to only trying the change again e.g. once a day at most or so?

sinnykumari:
Thank you Jerry for adding all the context and reasoning, great explanation! 100% agree with it and will echo again that validation should be done at the source not at the consumer level. This scales better, less error prone as provider have better knowledge of what is correct.
On the similar note, we backed off recently proposal which involved MCO updating infra object https://github.com/openshift/enhancements/pull/1102#issuecomment-1201019554

palonsoro:
I agree that it should have been CNO and not MCO who does this test. Honestly, I don't understand how the bug ended in MCO in the first place. But if this is to be returned to CNO, we need some higher level co-ordination to make this possible.

BTW, the PR may not be the best discussion place for all of this but the bugzilla.

rphillips:
I am in agreement with the latest discussion. Let's close this PR and document a procedure to test the settings. The proxy settings should be tested in a staging environment.

palonsoro:
> I am in agreement with the latest discussion. Let's close this PR and document a procedure to test the settings. The proxy settings should be tested in a staging environment.

I agree with closing this in what regards MCO, because this should (and never should have been) checked by MCO.

However, just relying on customer validation on stage environment is not a correct approach, because mistakes will always happen. The main point of this bug was not to protect from the error itself, but from the fact that there is no sane way to recover from it once it happened. 

If this issue raises again, we shall open bug to the CNO, which where some proper solution should be placed.

yuqi-zhang:
Thanks everyone for the work and comments! will try to continue tracking this in Jira so we don't lose the context

PR #2540, opened 2021-04-20
Bug 1934176: ensure SSH key uniqueness

cheesesashimi:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1934176

**- What I did**
I added some code to `pkg/controller/common/helpers.go` to ensure that when Ignition 2 configs are converted to Ignition 3, all `PasswdUser` objects contain per-user-unique `AuthorizedSSHKey`. Not ensuring the SSH keys are unique results in a degraded state for the MCO.

**- How to verify it**

Ensures SSH keys are unique for translation from Ignition 2 to Ignition 3.

### DIFF
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -479,7 +479,8 @@ func removeIgnDuplicateFilesUnitsUsers(ignConfig ign2types.Config) (ign2types.Co
 				outUser.SSHAuthorizedKeys = append(outUser.SSHAuthorizedKeys, users[i].SSHAuthorizedKeys[j])
 			}
 		}
-		ignConfig.Passwd.Users = []ign2types.PasswdUser{outUser}
+		// Ensure SSH key uniqueness
+		ignConfig.Passwd.Users = []ign2types.PasswdUser{dedupePasswdUserSSHKeys(outUser)}
 	}
 
 	// outFiles and outUnits should now have all duplication removed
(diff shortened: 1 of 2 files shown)


pkg/controller/common/helpers.go
```diff
@@ -236,13 +236,53 @@ func ConvertRawExtIgnitionToV2(inRawExtIgn *runtime.RawExtension) (runtime.RawEx
...
+
 // convertIgnition2to3 takes an ignition spec v2.2 config and returns a v3.2 config
 func convertIgnition2to3(ign2config ign2types.Config) (ign3types.Config, error) {
 	// only support writing to root file system
 	fsMap := map[string]string{
 		"root": "/",
 	}
 
+	// Ensure SSH keys are unique for each PasswdUser
+	ign2config.Passwd.Users = dedupeSSHKeys(ign2config.Passwd.Users)
```
yuqi-zhang, line 284: I think we can remove this, since basically every time this function gets called, the `removeIgnDuplicateFilesUnitsUsers` will also have been called.

pkg/controller/common/helpers.go
```diff
@@ -236,13 +236,53 @@ func ConvertRawExtIgnitionToV2(inRawExtIgn *runtime.RawExtension) (runtime.RawEx
...
+
+	// Overwrite the keys with the deduped list.
+	passwdUser.SSHAuthorizedKeys = dedupedSSHKeys
+
+	return passwdUser
+}
+
+// Ensures SSH keys are unique for all Ign 2 PasswdUsers in a slice
+// See: https://bugzilla.redhat.com/show_bug.cgi?id=1934176
+func dedupeSSHKeys(passwdUsers []ign2types.PasswdUser) []ign2types.PasswdUser {
```
yuqi-zhang, line 267: Given that the below function isn't really needed, I think we can also remove this function, and just have one `dedupePasswdUserSSHKeys`

pkg/controller/common/helpers_test.go
```diff
@@ -99,7 +114,12 @@ func TestParseAndConvert(t *testing.T) {
 
 	// Make a Ign2 comp config
 	testIgn2Config := ign2types.Config{}
-	tempUser2 := ign2types.PasswdUser{Name: "core", SSHAuthorizedKeys: []ign2types.SSHAuthorizedKey{"5678", "abc"}}
+	tempUser2SSHKeys := []ign2types.SSHAuthorizedKey{
```
yuqi-zhang, line 117: Minor: I think just having the above testing in `TestConvertIgnition2to3` should be enough. Unless you feel that this covers more use cases

yuqi-zhang (commented):
Overall looks good! Some minor comments

pkg/controller/common/helpers_test.go
cheesesashimi: Fine with me!

pkg/controller/common/helpers.go
cheesesashimi: I noticed in [ConvertRawExtIgnitionToV3](https://github.com/openshift/machine-config-operator/blob/870f8c4de5a3927ef71e0cfd0e81bd5b75b424fd/pkg/controller/common/helpers.go#L152), `convertIgnition2to3` gets called without `removeIgnDuplicateFilesUnitsUsers` being called first.

However, based upon my testing, adding the deduplication call to `removeIgnDuplicateFilesUnitsUsers` is sufficient.

pkg/controller/common/helpers.go
cheesesashimi: Fine with me!

cheesesashimi:
@yuqi-zhang I think this is ready for another look!

yuqi-zhang (approved):
/lgtm
/bugzilla refresh

PR #2542, opened 2021-04-22
Bug 1952368: daemon: return nil for unsupported operation on an OS

sinnykumari:
MCO doesn't support rpm-ostree related operation on RHEL
nodes like KernelType change, extensions. Instead of returning
error, log message and return nil so that nodes don't go
degraded and block updates for the rest of the
nodes in the pool.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1952368

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -972,12 +972,14 @@ func generateKargs(oldConfig, newConfig *mcfgv1.MachineConfig) []string {
 
 // updateKernelArguments adjusts the kernel args
 func (dn *Daemon) updateKernelArguments(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	if !dn.os.IsCoreOSVariant() {
+		glog.Info("updating kargs on non-CoreOS nodes is not supported")
+		return nil
+	}
+
 	kargs := generateKargs(oldConfig, newConfig)
 	if len(kargs) == 0 {
 		return nil
-	}
-	if !dn.os.IsCoreOSVariant() {
-		return fmt.Errorf("updating kargs on non-CoreOS nodes is not supported: %v", kargs)
 	}
 
 	args := append([]string{"kargs"}, kargs...)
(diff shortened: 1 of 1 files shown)


yuqi-zhang (commented):
I'm surprised we haven't hit this before. I guess nobody tried to adjust e.g. kargs for clusters with RHEL nodes?

Looking at the bug, it seems that the cluster thinks there's a RT kernel upgrade when the user applied a ICSP change. Is that the `canonicalizeKernelType` logic getting something wrong?

I'm a bit ambivalent on whether we should error on not. Taking RT kernel for example, I don't think you can do it at all for RHEL nodes right? So if you try to apply to a cluster with RHEL nodes, maybe we should error and tell the user that because you have RHEL nodes, you can't do this?

On the other hand we technically say the MCO doesn't handle OS updates in RHEL so maybe it makes sense we allow everything and let the user figure it out. In that case, we can maybe even considering removing the individual checks and just add a check to the top of `applyOSChanges` and just return nil for anything in that general criteria.

WDYT @sinnykumari (apologies for the text dump)

sinnykumari:
> I'm surprised we haven't hit this before. I guess nobody tried to adjust e.g. kargs for clusters with RHEL nodes?

yeah, it would be definitely a pure luck.

> Looking at the bug, it seems that the cluster thinks there's a RT kernel upgrade when the user applied a ICSP change. Is that the `canonicalizeKernelType` logic getting something wrong?

canonicalizeKernelType logic is fine. In the bug, what is happening is since we don't do early return, it reaches to the point https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L1111 where we return error as we don't support kernel switch on RHEL nodes.  The reason we let code run through because in case of OSImageURL has changed, we would like to update realtime kernel which could be the case during upgrade.
 
> I'm a bit ambivalent on whether we should error on not. Taking RT kernel for example, I don't think you can do it at all for RHEL nodes right? So if you try to apply to a cluster with RHEL nodes, maybe we should error and tell the user that because you have RHEL nodes, you can't do this?

We don't even support OS Update on RHEL node and today we simply `return nil` with info message. This behavior is documented in [OpenShift doc](https://docs.openshift.com/container-platform/4.7/updating/updating-cluster-rhel-compute.html#update-service-overview_updating-cluster-rhel-compute). I think rather than degrading, it would be better to document explicitly in our doc. It would be simply a no-op on OS that doesn't support something.

> On the other hand we technically say the MCO doesn't handle OS updates in RHEL so maybe it makes sense we allow everything and let the user figure it out. In that case, we can maybe even considering removing the individual checks and just add a check to the top of `applyOSChanges` and just return nil for anything in that general criteria.

I thought about it as well, but I would rather leave individual functionality decide what they support. For example: we don't support today realtime kernel switch on OKD. 

> WDYT @sinnykumari (apologies for the text dump)

Not at all, thanks for giving deep thought :)

yuqi-zhang (commented):
> We don't even support OS Update on RHEL node and today we simply return nil with info message. This behavior is documented in OpenShift doc. I think rather than degrading, it would be better to document explicitly in our doc. It would be simply a no-op on OS that doesn't support something.

Right, I see, that makes sense

> I thought about it as well, but I would rather leave individual functionality decide what they support. For example: we don't support today realtime kernel switch on OKD.

Right, and now I think about it another benefit is we will explicitly log which config change we didn't support

Thanks for the thorough thought! This mostly lgtm, one minor comment (weird didn't save the first time around, separately commented):

pkg/daemon/update.go
```diff
@@ -1076,15 +1078,17 @@ func validateExtensions(exts []string) error {
 }
 
 func (dn *Daemon) applyExtensions(oldConfig, newConfig *mcfgv1.MachineConfig) error {
+	// Right now, we support extensions only on CoreOS nodes
+	if !dn.os.IsCoreOSVariant() {
+		glog.Info("applying extensions on non-CoreOS nodes is not supported on ")
```
yuqi-zhang, line 1083: Did you mean to log the OS variant? `"not supported on " cuts off at the end there

sinnykumari: ah no. I was rewording sentence to make it consistent with others and it was left as a trail. Thanks for catching this. Fixed.

PR #2543, opened 2021-04-22
Bug 1950974: change system-connections-merged directory to systemConnectionsMerged

Prashanth684:
Looks like there is a bug in 8.4 systemd which causes the system-connections-merged mount unit to not be enabled. On further investigation looks like systemd does not like the `\x2d` character when enabling presets. While this issue needs to be investigated by the systemd folks, changing the name here so that UPI baremetal deploys can succeed.


cybertron:
/test e2e-metal-ipi

too many requests bug.

cybertron (approved):
/lgtm

The two remaining ci failures are "normal", so they don't need to block this.

Prashanth684:
@kikisdeliveryservice good for an approval ? thanks!

yuqi-zhang (approved):

PR #2549, opened 2021-04-27
DO NOT MERGE: Debug evictions with grace period zero

smarterclayton:
This is just a test PR to see why we might have a grace period
zero (to confirm the theory)


smarterclayton:
Terminated pods are removed from apiserver, so the evictions with grace zero were all completed pods. Closing.

PR #2551, opened 2021-04-28
Bug 1943539: templates: order service files that use podman around crio-wipe

haircommander:
or else they may race against each other, causing issues with crio-wipe running on a forced shutdown

both machine-config-daemon-pull.service and nodeip-configuration.service must run after crio wipe runs, to protect against a corrupted image storage from a forced reboot

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
+++ b/templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
@@ -8,8 +8,7 @@ contents: |
   # Removal of this file signals firstboot completion
   ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
   After=machine-config-daemon-pull.service
-  Before=crio.service crio-wipe.service
-  Before=kubelet.service
+  Before=crio.service kubelet.service
 
   [Service]
   Type=oneshot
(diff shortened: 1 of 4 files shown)


cgwalters (approved):
Eeek.

EDIT: To clarify the "Eeek" response was about the *problem* - the patch itself looks totally fine and makes sense!

haircommander:
as a note to ease the mind: this is only recently relevant. crio-wipe is in the process of adding support for detecting a non-graceful shutdown (which caused corrupted images in the past) in 4.8. Before this, we didn't interfere with podman containers, so this wasn't relevant

haircommander:
the fact that no e2e test has passed cannot be a coincidence.
/hold
looking into it

haircommander:
there was a cyclical dependency that should be fixed now

QiWang19 (commented):
LGTM

templates/common/_base/units/machine-config-daemon-firstboot.service.yaml
```diff
@@ -8,8 +8,7 @@ contents: |
   # Removal of this file signals firstboot completion
   ConditionPathExists=/etc/ignition-machine-config-encapsulated.json
   After=machine-config-daemon-pull.service
-  Before=crio.service crio-wipe.service
```
yuqi-zhang, line 11: out of curiosity why are you removing this dependency? the firstboot service does have scenarios today where it would make a call to podman

haircommander: crio-wipe should run before firstboot, which is ensured by the After= field in mcd-pull. Should I add an After=crio-wipe.service here too?

yuqi-zhang: Ah right I forgot that explicit dependency. No that's fine as is

PR #2552, opened 2021-04-29
add kubelet skew check for MCO upgradeable

deads2k:
Implementation of https://github.com/openshift/enhancements/pull/762 with locality at the point of action.

Produces a condition like
`KubeletSkewTooFar node/first-node must be updated or removed before the cluster can upgrade, current version v1.18`

/hold  just a POC demonstrating ease and benefits of locality.

cc @crawford @rphillips

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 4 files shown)


nstielau:
I'm likely biased but some recent reports of difficulty, but explicit expectations seems better than implicit expectations.  I can't even think of a place where we'd need a valid override (but there very well could be one).

pkg/operator/status.go
```diff
@@ -289,9 +293,83 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if ran, err := optr.isKubeletSkewTooFar(); ran && err != nil {
+		if coStatus.Status != configv1.ConditionFalse {
+			coStatus.Status = configv1.ConditionFalse
```
nstielau, line 299: I'd still love a `coStatus.DocumentationLink` type field here ;)

QiWang19:
> Implementation of [openshift/enhancements#762](https://github.com/openshift/enhancements/pull/762) with locality at the point of action.
> 
> Produces a condition like
> `KubeletSkewTooFar node/first-node must be updated or removed before the cluster can upgrade, current version v1.18`
> 
> /hold just a POC demonstrating ease and benefits of locality.
> 

@deads2k what do you mean `just a POC`? Is this PR supposed to be merged?

kikisdeliveryservice:
this is being worked on via another pr closing as it's unneeded.

PR #2554, opened 2021-04-30
DO NOT MERGE: Should openshift-gcp-routes.service be tied as network?

smarterclayton:
Testing theory that openshift-gcp-routes needs to be shutdown after
all pods are stopped (which includes apiserver), currently it is
shutdown at the same time.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
+++ b/templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
@@ -7,6 +7,7 @@ contents: |
   ConditionKernelCommandLine=|ignition.platform.id=gcp
   Wants=network-online.target
   After=network-online.target
+  PartOf=network.target
 
   [Service]
   Type=simple


templates/master/00-master/gcp/units/openshift-gcp-routes.service.yaml
```diff
@@ -7,6 +7,7 @@ contents: |
   ConditionKernelCommandLine=|ignition.platform.id=gcp
   Wants=network-online.target
   After=network-online.target
+  PartOf=network.target
```
aojea, line 10: I'm far of a systemd expert, but should this After=network.target?
> Services using the network should hence simply place an After=network.target dependency in their unit files, and avoid any Wants=network.target or even Requires=network.target.
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

aojea:
So, the relation is
apiserver watcher pod checks the apiserver and drops a file consumed by the gcp-route hack script
https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/templates/master/00-master/azure/files/etc-kubernetes-manifests-apiserver-watcher.yaml

apiserver watcher doesn't handle sigterm :/ 
https://github.com/openshift/machine-config-operator/blob/8bd0dd547059e9a5cbb43d2f58b40e6bece6c72f/cmd/apiserver-watcher/run.go#L118

If the apiserver watcher pods detects a failure in the apiserver, current threshold for failure is 8 
https://github.com/openshift/machine-config-operator/blob/8bd0dd547059e9a5cbb43d2f58b40e6bece6c72f/cmd/apiserver-watcher/run.go#L99
and the checks are every 2 seconds
https://github.com/openshift/machine-config-operator/blob/8bd0dd547059e9a5cbb43d2f58b40e6bece6c72f/cmd/apiserver-watcher/run.go#L108
it signals the script to remove the entries

smarterclayton:
Merged in #2561

PR #2557, opened 2021-04-30
[WIP][TEST] apiserver/GCP loadbalancer tweaks

aojea:
- the gcp-routes script that handles the GCP LB wait for the network to be down before exiting, so it still can keep the LB connections alive
- increase the apiserver-watcher pod grace period to 60s, so it can follow apiserver lifecyle <<< I made up the number, TODO use the same value as the apiserver pod???
- add iptables rules to not delete the established connections >> TODO check with Casey
- apiserver-watcher implement graceful timeout, wait for apiserver to be unhealthy before exiting on SIGTERM
- apiserver-watcher be more conservative, current timeouts are pretty aggressive, and can create flapping
- it seems apiserver keep serving for ~100s since it starts to report http.status with errors, so don't redirect local traffic externally until that
- implement a short-circuit in case that the apiserver doesn't have the tcp port open and fail inmidiatly


aojea:
something is doing this change that the apiserver are overloaded and doesn't answer anything

> the server has received too many requests and has asked us to try again late

PR #2558, opened 2021-04-30
Bug 1952358: vSphere platform none vmxnet3v4 fix

jcpowermac:
This adds `templates/common/none` directory
with the fix `platform: vsphere` for vmxnet3 v4.

### DIFF
--- a/templates/common/none/OWNERS
+++ b/templates/common/none/OWNERS
@@ -1,0 +1,7 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file just uses aliases defined in OWNERS_ALIASES.
+
+approvers:
+  - vsphere-approvers
+reviewers:
+  - vsphere-reviewers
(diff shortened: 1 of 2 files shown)


jcpowermac:
I am unsure who the owners should be. Probably not vsphere

yuqi-zhang:
Two quick questions:

1. Aside from metal-on-vsphere, do we know if there are other use cases for `none`?
2. As an alternative, we could ship this as a base template, and gate on vsphere platform (I guess the script already does, because `vmxnet3` should only be used on vsphere right?). It *should* not affect other platforms and doesn't open the potential can of worms of a "none" platform. Would there be other downsides to this?

jcpowermac:
> Two quick questions:
> 
> 1. Aside from metal-on-vsphere, do we know if there are other use cases for `none`?

Any cloud platform that we currently don't support directly and baremetal upi. There could be more and I am forgetting

> 2. As an alternative, we could ship this as a base template, and gate on vsphere platform (I guess the script already does, because `vmxnet3` should only be used on vsphere right?). It _should_ not affect other platforms and doesn't open the potential can of worms of a "none" platform. Would there be other downsides to this?

I think this is up to the MCO team. I just tried out adding the none template directory and it worked. I figured that we are limiting our potential for an issue. If mco feels more comfortable moving it to base I have no problem moving it.

PR #2559, opened 2021-04-30
Bug 1952358: vsphere: platform none, vmxnet3v4 fix move to base

jcpowermac:
Alternative to #2558


yuqi-zhang:
It'd be also good if we can at least manually verify this for `none`, but I am more in favour of this approach that 2558

jcpowermac:
> It'd be also good if we can at least manually verify this for `none`, but I am more in favour of this approach that 2558

Sounds like a good idea, let me do that.

jcpowermac:
As requested tested a `platform: none` install (vSphere)
After install upgraded a single guest - `control-plane-0` to hardware version 17.

`install-config.yaml` snipit
```
apiVersion: v1
baseDomain: vmc.devcluster.openshift.com
metadata:
  name: jcupi 
platform:
  none: {}
pullSecret: |
```


```
_  vsphere-upi oc debug node/control-plane-0
Starting pod/control-plane-0-debug ...
Pod IP: 192.168.19.4
sh-4.4# chroot /host
```
Disabled udp offload as expected
```
sh-4.4# ethtool -k ens192 | grep -v fixed
Features for ens192:
rx-checksumming: on
tx-checksumming: on
        tx-checksum-ip-generic: on
scatter-gather: on
        tx-scatter-gather: on
tcp-segmentation-offload: on
        tx-tcp-segmentation: on
        tx-tcp-mangleid-segmentation: off
        tx-tcp6-segmentation: on
generic-segmentation-offload: on
generic-receive-offload: on
large-receive-offload: off
rx-vlan-offload: on
tx-vlan-offload: on
receive-hashing: on
highdma: on
tx-udp_tnl-segmentation: off
tx-udp_tnl-csum-segmentation: off
rx-gro-list: off
tx-nocache-copy: off
```

Dispatch file, as expected
```
sh-4.4# cat /etc/NetworkManager/dispatcher.d/99-vsphere-disable-tx-udp-tnl 
#!/bin/bash
# Workaround:
# https://bugzilla.redhat.com/show_bug.cgi?id=1941714
# https://bugzilla.redhat.com/show_bug.cgi?id=1935539

driver=$(nmcli -t -m tabular -f general.driver dev show "${DEVICE_IFACE}")

if [[ "$2" == "up" && "${driver}" == "vmxnet3" ]]; then
  logger -s "99-vsphere-disable-tx-udp-tnl triggered by ${2} on device ${DEVICE_IFACE}."
  ethtool -K ${DEVICE_IFACE} tx-udp_tnl-segmentation off
  ethtool -K ${DEVICE_IFACE} tx-udp_tnl-csum-segmentation off
fi
```
No cloud-config, as expected
```
sh-4.4# cat /etc/kubernetes/cloud.conf 
sh-4.4# 
```

jcpowermac:
vSphere job failure caused by single tests, I would think unrelated to this change.

sinnykumari:
Find this implementation better than https://github.com/openshift/machine-config-operator/pull/2558 . Can we also add description in commit message to include reasoning of this template move?

jcpowermac:
> Find this implementation better than #2558 . Can we also add description in commit message to include reasoning of this template move?

@sinnykumari updated.

ashcrow:
```
level=error msg=Error: could not contact Ironic API: timeout reached
```
Failure could be infra issue.

jcpowermac:
vsphere test failures look unrelated.

jcpowermac:
Are we waiting for those last jobs to pass? From dealing with other PRs in MCO those rarely are green.

Still going to take quite a bit of time for this to merge and then backported to 4.8 and 4.7.

ashcrow:
My $0.02: If we can explain why these are failing and have confidence this works then I don't believe we need to wait.

yuqi-zhang (approved):
I think the failing CI are unrelated and not required, so let's go ahead and move forwards with this given 

1. Joseph was able to confirm this works on `none`
2. This does not seem to affect any other platform
3. We will be looking to revert this as soon as the underlying issues are resolved
/lgtm

PR #2560, opened 2021-05-01
Bug 1944605: Use recommended retry intervals

QiWang19:
Fix bug 1944605. The nodes get stuck after applying the kubeletconfig,
the log has timeout error requesting data from the node.
use recommended retry intervals for possible timeout errors. I am not sure
if this will fix the bug since there's no way to reproduce.

**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -61,12 +61,6 @@ var (
 	controllerKind = mcfgv1.SchemeGroupVersion.WithKind("ContainerRuntimeConfig")
 )
 
-var updateBackoff = wait.Backoff{
-	Steps:    5,
-	Duration: 100 * time.Millisecond,
-	Jitter:   1.0,
-}
-
 // Controller defines the container runtime config controller.
 type Controller struct {
 	templatesDir string
@@ -417,7 +411,7 @@ func generateOriginalContainerRuntimeConfigs(templateDir string, cc *mcfgv1.Cont
 }
 
 func (ctrl *Controller) syncStatusOnly(cfg *mcfgv1.ContainerRuntimeConfig, err error, args ...interface{}) error {
-	statusUpdateErr := retry.RetryOnConflict(updateBackoff, func() error {
+	statusUpdateErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		newcfg, getErr := ctrl.mccrLister.Get(cfg.Name)
 		if getErr != nil {
 			return getErr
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
How/what is this intended to help?

If I’m reading the code correctly:
- With `updateBackoff`, there are 5 attempts at about 0, 100ms, 200ms, 300ms, 400ms.
- With `retry.DefaultRetry`, there are 5 attempts at about 0, 10ms, 20ms, 30ms, 40ms.

And all of those retries only happen on _conflict_ errors; there is no difference for any other kinds of errors.

How is using the same amount of retries with less wait in between less likely to fail?

---

Either way, the failing queue item should get requeued and per `DefaultControllerRateLimiter`, retried over time for many seconds later. So there should be plenty of opportunities to either resolve a conflict, or avoid a transient failure, whether the immediate retries are in 10s or 100s of ms.

I realize I’m very ignorant of the finer points of k8s reliability; this it may well make a great deal of difference for efficiency/speed of state convergence, I wouldn’t know — but I can’t see how it could make a difference for “stuck” nodes. Why _does_ it make a difference?

mtrmac:
(I’m not saying that the timeouts must stay as they are, the code does not document any specific reason for using non-default values; it’s just non-obvious to me that it should make a difference either way. The platform is supposed to be resilient against much more severe hardware and network failures, including the MCC being abruptly terminated.)

kikisdeliveryservice:
Same questions as @mtrmac above. 

I don't see how this would fix the issue. Looking at the must gather on 3.17 :
  - lastTransitionTime: "2021-03-17T11:53:59Z"
    message: |-
      DaemonSet "openshift-multus/network-metrics-daemon" is not available (awaiting 1 nodes)
      DaemonSet "openshift-ovn-kubernetes/ovs-node" is not available (awaiting 1 nodes)
      DaemonSet "openshift-ovn-kubernetes/ovnkube-node" is not available (awaiting 1 nodes)

but the new rendered config is generated and rolling out after this ovn/network problem already exists (see the time stamp):
  - lastTransitionTime: "2021-03-17T12:56:12Z"
    message: All nodes are updating to rendered-worker-d7212e4cc75170aa2391bb094627438e
    reason: ""
    status: "True"
    type: Updating

I don't think the kubelet config caused this issue, rather the issue predates that and is preventing it from being rolled out as that node isn't reachable.

I'd direct the BZ to ovn to take a look
/hold

kikisdeliveryservice:
Underlying bug was closed and duped to https://bugzilla.redhat.com/show_bug.cgi?id=1961899 by OVN team so this PR is no longer needed. Closing.

PR #2565, opened 2021-05-04
Bug 1947066: Allow wildcard in noProxy field

cheesesashimi:
If a wildcard (*) is used in the `noProxy` field, it is not properly quoted before being rendered into YAML, causing `resourceread.ReadDaemonSetV1OrDie()` to fail because the rendered YAML was `noProxy: *` instead of `noProxy: "*"`.

The reason for the read failure is because an unquoted asterisk has special meaning in YAML, typically used to repeat other sections (which we're not doing here).

Fixes: #1947066

**- What I did**
I added quotes to the `manifests/machineconfigdaemon/daemonset.yaml` file around the optional `noProxy` field.

**- How to verify it**
I've updated the unit test suite `pkg/operator/render_test.go` to read the rendered templates using the appropriate `resourceread` function.

Allow wildcard (*) in noProxy field

### DIFF
--- a/pkg/operator/render_test.go
+++ b/pkg/operator/render_test.go
@@ -6,7 +6,9 @@ import (
 	"testing"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/machine-config-operator/lib/resourceread"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestClusterDNSIP(t *testing.T) {
@@ -81,7 +83,7 @@ func TestRenderAsset(t *testing.T) {
 	tests := []struct {
 		Path         string
 		RenderConfig *renderConfig
-		FindExpected string
+		FindExpected []string
 		Error        bool
 	}{{
 		// Simple test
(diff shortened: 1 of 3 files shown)


yuqi-zhang:
The quotations look fine from a MCO CI perspective. Let me know if you are able to install a cluster with a wildcard noProxy. Thanks!

cheesesashimi:
I was able to configure a preexisting cluster with a wildcard `noProxy` using this PR. I also tested it by creating an `install-config.yaml` file with a `noProxy: "*"` field. The newly installed cluster behaved as anticipated and I did not see any related error messages.

pkg/operator/render_test.go
```diff
@@ -108,12 +110,46 @@ func TestRenderAsset(t *testing.T) {
...
+				Proxy: &configv1.ProxyStatus{
+					HTTPSProxy: "https://i.am.a.proxy.server",
+					NoProxy:    "*", // See: https://bugzilla.redhat.com/show_bug.cgi?id=1947066
+				},
+			},
+		},
+		FindExpected: []string{
+			"image: mco-operator-image",
+			"image: oauth-proxy-image",
+			"- name: HTTPS_PROXY\n            value: https://i.am.a.proxy.server",
```
yuqi-zhang, line 132: Just curious, is the extra spaces in between how the output is formatted?

yuqi-zhang (approved):
Overall lgtm! Test failures seem unrelated but just to make sure
/retest

sinnykumari:
This looks great!

/lgtm

Issue #2529, opened 2021-04-14
oVirt/RHV disabled nodeip-configuration.service


mriedmann:
**Description**

Our RHV IPI installed clusters (4.6.21, 4.6.24) are showing the `nodeip-configuration.service` as disabled (`enabled: false`) in `00-master` and `00-worker` machine config resources. We are currently investigating a major cluster stability problem since our upgrade from 4.6.18 to 4.6.22 in one cluster. This lead us to another issue (https://bugzilla.redhat.com/show_bug.cgi?id=1940939) that discribes a problem with this service, but is explicitly mentioning that the service is enabled (but not working properly). The --node-ip parameter of kubelet is empty, resulting in a flip-flop behavior on nodes binding 2 ips (like the ingress or api IPs). 

As far as I understand the code at hand the service should be enabled, according to following source-code parts, because ingress and API IPs are set in the referenced resource (controllerconfig/machine-config-controller): 
* https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/templates/common/on-prem/units/nodeip-configuration.service.yaml#L2
* https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/pkg/controller/template/render.go#L442
* https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/pkg/operator/render.go#L236

A current workaround is to start the nodeip-configuration service manually, calling systemctl daemon-reload and restarting kubelet. After that the --node-ip argument is showing the right ip and the cluster starts to stabalize. 

**Steps to reproduce the issue:**
1. Install OCP 4.6.21 against RHV
2. Inspect machine-config 00-master
3. Notice that the `nodeip-configuration.service` is disabled (enabled: false)

**Describe the results you received:**

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  name: 00-master
...
          enabled: false
          name: nodeip-configuration.service
...
```

**Describe the results you expected:**

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  name: 00-master
...
          enabled: true
          name: nodeip-configuration.service
...
```

**Additional information you deem important (e.g. issue happens only occasionally):**

[...]

PR #2567, opened 2021-05-05
Bug 1948533: Add nodeip-configuration.service for oVirt

Gal-Zaidman:
**- What I did**

This change adds the nodeip-configuration.service to the oVirt platform.
On 4.7 due to the unification of the unit files of each on prem platform to the "on-prem" directory oVirt got the "nodeip-configuration.service" file but on 4.6 release we are missing it which can cause hard to debug problems to customers.

This change also adds fixes for BZ#1940939 https://github.com/openshift/machine-config-operator/pull/2470

** I was not able to reproduce the issue on BZ#1948533 but this change seems relevant and according to the comments on BZ#1948533 will fix the issues

**- How to verify it**
Install a 4.6 ovirt cluster

### DIFF
--- a/templates/common/ovirt/units/nodeip-configuration.service.yaml
+++ b/templates/common/ovirt/units/nodeip-configuration.service.yaml
@@ -1,0 +1,78 @@
+name: nodeip-configuration.service
+enabled: {{if .Infra -}}
+         {{if .Infra.Status -}}
+         {{if .Infra.Status.PlatformStatus -}}
+         {{if .Infra.Status.PlatformStatus.Ovirt -}}
+         {{if .Infra.Status.PlatformStatus.Ovirt.APIServerInternalIP -}}
+         true
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end -}}
+         {{- else -}}
+         false
+         {{- end }}
+contents: |
+  [Unit]
+  Description=Writes IP address configuration so that kubelet and crio services select a valid node IP
+  # This only applies to VIP managing environments where the kubelet and crio IP
+  # address picking logic is flawed and may end up selecting an address from a
+  # different subnet or a deprecated address
+  Wants=network-online.target
+  After=network-online.target ignition-firstboot-complete.service
+  Before=kubelet.service crio.service
+
+  [Service]
+  # Need oneshot to delay kubelet
+  Type=oneshot
+  {{ if .Infra -}}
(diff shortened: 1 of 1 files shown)


Gal-Zaidman:
/hold

For tests

Gal-Zaidman:
/unhold

oVirt job passed

templates/common/ovirt/units/nodeip-configuration.service.yaml
```diff
@@ -0,0 +1,48 @@
+name: nodeip-configuration.service
+enabled: true
```
patrickdillon, line 2: Do you want to disable this service in the case of a UPI install? That is what we do for vSphere, but not sure if applicable for your case: https://github.com/openshift/machine-config-operator/blob/release-4.6/templates/common/vsphere/units/nodeip-configuration.service.yaml

patrickdillon:
Fixes #2529

yuqi-zhang (approved):
Generally LGTM pending Patrick's comment. Will also let him do the final lgtm when ready

templates/common/ovirt/units/nodeip-configuration.service.yaml
Gal-Zaidman: Hi thanks for the review,
I added the enabled logic but removed the ifs from the "contents" since I think that if the service is disabled it doesn't matter

templates/common/ovirt/units/nodeip-configuration.service.yaml
```diff
@@ -0,0 +1,68 @@
...
+  ExecStart=/bin/bash -c " \
+    until \
+    /usr/bin/podman run --rm \
+    --authfile /var/lib/kubelet/config.json \
+    --volume /etc/systemd/system:/etc/systemd/system:z \
+    --net=host \
+    {{ .Images.baremetalRuntimeCfgImage }} \
+    node-ip \
+    set --retry-on-failure \
+    {{.Infra.Status.PlatformStatus.Ovirt.APIServerInternalIP }}; \
```
patrickdillon, line 48: Does this attribute exist in UPI? We had the conditionals to protect against an exception in UPi where this value does not exist.

Gal-Zaidman: Nope, it wouldn't exist, but I thought that if the service is disabled it doesn't matter.
That is not the case?

patrickdillon: The template is written regardless, so without the condition the mco will throw an exception trying to access the value while writing the template.

Gal-Zaidman: thanks for the explanation, fixed

PR #2568, opened 2021-05-05
Bug 1957374: install: Automatically fill in the pod for MCDDrainError

wking:
Instead of asking folks to look up the pod using the node name, use the pod label that the Prometheus scraper adds automatically when scraping metrics off the daemon pod.  While I'm touching this, I'm pivoting to the automatic label for the namespace as well, even though that is very unlikely to change.  The container being scraped is `oauth-proxy`, so I'm leaving the target container hard-coded to `machine-config-daemon`, since that's where the interesting logs are.


kikisdeliveryservice:
will make a bz thanks @wking 

/approve
/lgtm

PR #2570, opened 2021-05-06
Bug 1955517: Add cleanUpDuplicatedMC

QiWang19:
Fix BZ 1955517: upgrade when there are more than one configs,
these generated MC will be duplicated by upgraded MC with
number suffixed name (func getManagedKubeletConfigKey()) and fails the upgrade.
Remove the mc if its generated and degraded.

**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -616,8 +616,36 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 		}
 		glog.Infof("Applied ContainerRuntimeConfig %v on MachineConfigPool %v", key, pool.Name)
 	}
+	if err := ctrl.cleanUpDuplicatedMC(); err != nil {
+		return err
+	}
 
 	return ctrl.syncStatusOnly(cfg, nil)
+}
+
+// cleanUpDuplicatedMC removes the MC of uncorrected version if format of its name contains 'generated-xxx'.
+// BZ 1955517: upgrade when there are more than one configs, these generated MC will be duplicated
+// by upgraded MC with number suffixed name (func getManagedKeyCtrCfg()) and fails the upgrade.
+func (ctrl *Controller) cleanUpDuplicatedMC() error {
+	generatedCtrCfg := "generated-containerruntime"
+	// Get all machine configs
+	mcList, err := ctrl.client.MachineconfigurationV1().MachineConfigs().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("error listing containerruntime machine configs: %v", err)
+	}
+	for _, mc := range mcList.Items {
+		if !strings.Contains(mc.Name, generatedCtrCfg) {
+			continue
+		}
+		// delete the containerruntime mc if its degraded
+		if mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey] != version.Hash {
+			if err := ctrl.client.MachineconfigurationV1().MachineConfigs().Delete(context.TODO(), mc.Name, metav1.DeleteOptions{}); err != nil {
+				return fmt.Errorf("error deleting degraded containerruntime machine config %s: %v", mc.Name, err)
+			}
+
+		}
+	}
+	return nil
 }
 
 // mergeConfigChanges retrieves the original/default config data from the templates, decodes it and merges in the changes given by the Custom Resource.
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
@@ -37,6 +37,7 @@ import (
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	"github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/fake"
 	informers "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/machine-config-operator/pkg/version"
 	"github.com/openshift/machine-config-operator/test/helpers"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/controller/kubelet-config/helpers.go
```diff
@@ -159,7 +159,7 @@ func getManagedKubeletConfigKey(pool *mcfgv1.MachineConfigPool, client mcfgclien
 		}
 		val, ok := kc.GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
 		// If we find a matching kubelet config and it is the only one in the list, then return the default MC name with no suffix
-		if !ok && len(kcList.Items) < 2 {
+		if !ok {
```
umohnani8, line 162: We should add a condition for val so `if !ok || val == ""`. We will need to do the same for the container-runtime-config controller as well.

QiWang19:
@umohnani8 PTAL. I tested the upgrade scenario as the Bug 1955517 mentioned. It can successfully upgrade from 4.6 to 4.7.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -616,10 +616,39 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
 		glog.Infof("Applied ContainerRuntimeConfig %v on MachineConfigPool %v", key, pool.Name)
 	}
+	if err := ctrl.cleanUpDegradedMC(); err != nil {
+		return err
+	}
 
 	return ctrl.syncStatusOnly(cfg, nil)
 }
 
+// cleanUpDegradedMc removes the MC of uncorrected version if format of its name is non-suffixed generate-xxx.
```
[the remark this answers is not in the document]
QiWang19: fixed

QiWang19:
github.com/pkg/errors

QiWang19:
@kikisdeliveryservice @umohnani8 PTAL

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -616,10 +616,39 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+	mcList, err := ctrl.client.MachineconfigurationV1().MachineConfigs().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("error listing containerruntime machine configs: %v", err)
+	}
+	for _, mc := range mcList.Items {
+		mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
+		if !strings.HasSuffix(mc.Name, ctrcfgMCSuffix) {
+			continue
+		}
+		// delete the mc if its non-suffixed and degraded
```
haircommander, line 641: I feel like I'm missing some information, but here are a couple of notes: 
- above we're skipping if it doesn't have the suffix, but the comment here says we're deleting if it doesn't have the suffix suffixed. 
- also, the comment says it will delete if degraded, but I'm not seeing us check if it is degraded.

rphillips:
A unit test would be great to make sure the cleanup function works.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
QiWang19: I should make the commet more clear, we will delete the MC without number suffix, those MC usually has ctrcfgMCSuffix.
The version.Hash is check the degradation.

umohnani8: @QiWang19 I don't think we should just restrict this to non-suffixed MCs. We should delete any MC that is still using the old ignition version regardless of if it has a suffix or not.
I would suggest renaming `ctrcfgMCSuffix` to something like `generatedCtrcfg`  and check if the MC has that in its name.

QiWang19: fixed

QiWang19:
Added unit test for `cleanUpDuplicatedMC`.

QiWang19:
@kikisdeliveryservice @umohnani8 @rphillips PTAL.

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -633,10 +633,38 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
...
+	}
 
 	return ctrl.syncStatusOnly(cfg, nil)
 }
 
+// cleanUpDuplicatedMC removes the MC of uncorrected version if format of its name contains 'generated-xxx'.
+// BZ 1955517: upgrade when there are more than one configs, these generated MC will be duplicated
+// by upgraded MC with number suffixed name (func getManagedKeyCtrCfg()) and fails the upgrade.
+func (ctrl *Controller) cleanUpDuplicatedMC() error {
+	generatedKubeletCfg := "generated-kubelet"
```
rphillips, line 647: does this cleanup a duplicated kubelet config? The function should be called cleanupDuplicatedKC

QiWang19: Only the machine config is duplicated, it only cleans up the machine config. The result of `oc get kubeletconfigs.machineconfiguration.openshift.io` doesn't change.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -616,10 +616,39 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
...
+	if err != nil {
+		return fmt.Errorf("error listing containerruntime machine configs: %v", err)
+	}
+	for _, mc := range mcList.Items {
+		mcCtrlVersion := mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey]
+		if !strings.Contains(mc.Name, generatedCtrCfg) {
+			continue
+		}
+		// delete the containerruntime mc if its degraded
+		if mcCtrlVersion != version.Hash {
```
umohnani8, line 642: nit: can do `if mc.Annotations[ctrlcommon.GeneratedByControllerVersionAnnotationKey] != version.Hash` instead

pkg/controller/kubelet-config/kubelet_config_controller.go
umohnani8: You can create a common function for this in the common/helpers.go file as this is cleaning up MCs and is not specific to ctrcfg or kubeletconfig objects.

QiWang19: > You can create a common function for this in the common/helpers.go file as this is cleaning up MCs and is not specific to ctrcfg or kubeletconfig objects.

Didn't add it as a common, the cleanup functions are specific to different Controller.

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -633,10 +633,37 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
...
+	if err := ctrl.cleanUpDuplicatedMC(); err != nil {
+		return err
+	}
 
 	return ctrl.syncStatusOnly(cfg, nil)
 }
 
+// cleanUpDuplicatedMC removes the MC of uncorrected version if format of its name contains 'generated-xxx'.
+// BZ 1955517: upgrade when there are more than one configs, these generated MC will be duplicated
+// by upgraded MC with number suffixed name (func getManagedKeyCtrCfg()) and fails the upgrade.
```
umohnani8, line 645: nit: getManagedKeyCtrCfg() should be getManagedKubeletConfigKey()

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -956,3 +957,58 @@ func getKey(config *mcfgv1.ContainerRuntimeConfig, t *testing.T) string {
 	}
 	return key
 }
+
+// TestCleanUpDuplicatedMC test the function removes the MC from the MC list
+// if the MC is of old version.
+func TestCleanUpDuplicatedMC(t *testing.T) {
```
umohnani8, line 963: @QiWang19 can you please add similar tests for kubeletconfig

umohnani8:
LGTM after adding tests for kubeletconfig as well.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
QiWang19: test added.

QiWang19:
@kikisdeliveryservice @umohnani8 @rphillips PTAL.

QiWang19:
Need approval label.

kikisdeliveryservice:
@rphillips PTAL

Just holding my approval until Ryan gets a chance to approve. :)

rphillips:
looks ready!

/lgtm

kikisdeliveryservice (approved):
let's get this in :)

PR #2571, opened 2021-05-06
Bug 1957972: Remove remnants of mdns functionality

cybertron:
mDNS was removed in #2465, but a couple of the related templates were
missed. This just cleans those up too.

Note that there are also references to mdns in the MCO code, but I'm going
to leave the removal of those until 4.9 so we don't risk breaking anything on
upgrade.


mandre (commented):
I see all the related CI jobs have passed.
/lgtm

On a side note, we should also merge https://github.com/openshift/installer/pull/4821

mandre:
Per the bot suggestion, assigning Kirsten for approval.
/assign @kikisdeliveryservice

kikisdeliveryservice (approved):
Makes sense to me.

PR #2574, opened 2021-05-11
Bug 1948137: crio: enable internal_wipe option

haircommander:
to make sure we're calling CNI del on a node reboot

Signed-off-by: Peter Hunt <pehunt@redhat.com>

PR #2575, opened 2021-05-11
Bug 1921937: crio: reject /etc/hostname mount if absent

haircommander:
otherwise, we create /etc/hostname as a directory, which causes issues on reboot

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -25,6 +25,9 @@ contents:
         "/run/containers/oci/hooks.d",
     ]
     manage_ns_lifecycle = true
+    absent_mount_sources_to_reject = [
+        "/etc/hostname",
+    ]
 
     [crio.image]
     global_auth_file = "/var/lib/kubelet/config.json"
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -25,6 +25,9 @@ contents:
         "/run/containers/oci/hooks.d",
     ]
     manage_ns_lifecycle = true
+    absent_mount_sources_to_reject = [
+        "/etc/hostname",
+    ]
 
     [crio.image]
     global_auth_file = "/var/lib/kubelet/config.json"


mtrmac (commented):
(Not really a NAK, a weakly held opinion: I understand this was discussed elsewhere already.)

Why is special-casing `/etc/hostname` the right fix, and why is unilaterally refusing to mount this one `hostPath` into containers the right thing to do? Couldn’t a container be set up to intentionally and correctly (whatever it means) handle the situation of `/etc/hostname` missing on the node, e.g. to create and populate the file if it is empty?

In particular, https://kubernetes.io/docs/concepts/storage/volumes/#hostpath seems to already give the Pod spec enough power to deal with this situation however the pod author wants. Does that not work, or is it not available?

Just reading that (without testing), using a `type: File` in Pods should have the same effect as this PR. If the author meant to do that but didn’t, that seems to me more appropriate for a linter than for RHCOS to forcefully refuse.

---

If the concern is that node’s `/etc/hostname` is owned by the node/RHCOS and containers shouldn’t be modifying it in any way, `/etc/hostname` seems just a very specific special case of the larger concern — shouldn’t we completely prohibit `hostPath` mounts of `/etc`, or the like? I guess we don’t want to do that because we want the wide flexibility for unpredictable third-parties, but then it’s not obvious to me how `/etc/hostname` is special.

haircommander:
> (Not really a NAK, a weakly held opinion: I understand this was discussed elsewhere already.)
> 
> Why is special-casing `/etc/hostname` the right fix, and why is unilaterally refusing to mount this one `hostPath` into containers the right thing to do? Couldn’t a container be set up to intentionally and correctly (whatever it means) handle the situation of `/etc/hostname` missing on the node, e.g. to create and populate the file if it is empty?
> 
> In particular, https://kubernetes.io/docs/concepts/storage/volumes/#hostpath seems to already give the Pod spec enough power to deal with this situation however the pod author wants. Does that not work, or is it not available?
> 
> Just reading that (without testing), using a `type: File` in Pods should have the same effect as this PR. If the author meant to do that but didn’t, that seems to me more appropriate for a linter than for RHCOS to forcefully refuse.
> 
> If the concern is that node’s `/etc/hostname` is owned by the node/RHCOS and containers shouldn’t be modifying it in any way, `/etc/hostname` seems just a very specific special case of the larger concern — shouldn’t we completely prohibit `hostPath` mounts of `/etc`, or the like? I guess we don’t want to do that because we want the wide flexibility for unpredictable third-parties, but then it’s not obvious to me how `/etc/hostname` is special.

the problem is that when the source of a mount doesn't exist, and the user doesn't specify "file" or "fileOrCreate" as the hostpath type, crio needs to create a directory at that path *before* it's mounted. this is to maintain compatibility with docker. If the container asks to mount it, it doesn't have a chance to create it--either it's created by crio or already exists.

`/etc/hostname` is also a special case because it's is sometimes mounted, *and* it often doesn't exist. Folks in the RHCOS team also seem to be of the mindset that there should be very few entities thinking about `/etc/hostname` (because of headaches with DHCP in the past). As such, we can't even special case by creating a file instead of a directory if it's `/etc/hostname`. Rejecting the request seems to be the best solution (to help teach users not to do this, while also protecting against nodes not coming up after a reboot). 

We considered special-casing a hard-coded path directly in cri-o, but we instead decided to make it configurable, which allows us to add other files we may want in the future (thus preventing the need to do this for all of /etc, until someone runs into this with a new path :smiley: )

haircommander:
/hold

for https://github.com/openshift/machine-config-operator/pull/2574

cgwalters:
> The overall bug is that “a buggy privileged container with hostPath mounts can break the node”, which… is kind of indistinguishable from the desirable feature of allowing privileged containers to do things to the node; 

I think what's different here is trying to do a *specific* mount.  Most privileged containers that want to inspect the whole host mount the whole host `/` to `/host` or so.  Or another pattern is to cherry pick just e.g. `/etc` -> `/host/etc` or so.  Those are generally safe patterns, particularly when mounted read-only.

I guess one thing we we should ship on the host (or the MCO) is a small service that just does `rmdir /etc/hostname || true` since it's never useful to have it as a directory.  (OTOH, I think anyone affected by this also won't be able to get that fix via upgrades, so we just need docs/customer portal page about this)

haircommander:
> > The overall bug is that “a buggy privileged container with hostPath mounts can break the node”, which… is kind of indistinguishable from the desirable feature of allowing privileged containers to do things to the node;
> 
> I think what's different here is trying to do a _specific_ mount. Most privileged containers that want to inspect the whole host mount the whole host `/` to `/host` or so. Or another pattern is to cherry pick just e.g. `/etc` -> `/host/etc` or so. Those are generally safe patterns, particularly when mounted read-only.
> 
> I guess one thing we we should ship on the host (or the MCO) is a small service that just does `rmdir /etc/hostname || true` since it's never useful to have it as a directory. (OTOH, I think anyone affected by this also won't be able to get that fix via upgrades, so we just need docs/customer portal page about this)

for clarity, WDYT about the approach in this PR? If we have both a systemd unit and docs discouraging this, do you think this PR is needed?

templates/master/01-master-container-runtime/_base/files/crio.yaml
```diff
@@ -2,6 +2,9 @@ mode: 0644
 path: "/etc/crio/crio.conf.d/00-default"
 contents:
   inline: |
+    [crio]
+    internal_wipe = true
```
cgwalters, line 6: Can you motivate the relationship of this change to the bug a bit more?  Is it that in order to have the policy of rejecting containers that mount `/etc/hostname` apply after an upgrade we need to clear the containers after reboot?

(What's the relationship of this and the `crio-wipe.service`?  Are both needed?)

cgwalters:
We have strong evidence this is a real world problem, it's also an insidious one because it acts as a "time bomb" - it only breaks the node on the *next boot*.  I think this plus above factors motivate a band-aid fix, even if it's not the best/most-ideal one.

On the more general problem though...the more I think about this the more it seems best addressed generically - ideally push this bug/discussion into upstream Kube too. Instead of having the container runtime reject it, we could create and ship an admission webhook that prevents the pod from even being created or so?

cgwalters (approved):

cgwalters:
I'm good to ship this but we should keep this discussion open.

templates/master/01-master-container-runtime/_base/files/crio.yaml
haircommander: oh shoot this needs a rebase, this is unrelated but was built off of it to prevent conflicts

mrunalp (approved):

haircommander:
/hold cancel
as it's been rebased now

PR #2576, opened 2021-05-11
Hard-code 127.0.0.1 to unicast_peer list in keepalived

cybertron:
In the latest version of keepalived we have (2.1.5), the behavior
of unicast_peers changed so that if the list is empty it reverts to
multicast and ignores messages from other unicast nodes. This is a
problem during bootstrapping because it can cause a master to
incorrectly take the VIP and ignore attempts by the bootstrap node
to take it back.

Note that we filter the node's own IP out of the peer list because if
we don't it triggers warnings in the keepalived logs. 127.0.0.1 does
not trigger the same warnings, so it should be safe to use.

**- What I did**

**- How to verify it**


cybertron:
I think my concerns with https://github.com/openshift/baremetal-runtimecfg/pull/137 were unfounded, so we don't need this.

PR #2577, opened 2021-05-12
Bug 1959599: append the internalRegistryHostname to allowedRegistries list

QiWang19:
Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1959599

Append the internalRegistryHostname to allowedRegistries list.
If the internalRegistryHostname configured, add the host to allowdRegtries by default
to avoid the explicitly configure the same host under allowdRegtries. Otherwise, the
internal host will be rejected and cause errors.

**- What I did**

**- How to verify it**

Append the internalRegistryHostname to allowedRegistries list

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -678,6 +678,11 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		}
 	}
 
+	// append the internalRegistryHostname to allowedRegistries list
+	allowedRegs := imgcfg.Spec.RegistrySources.AllowedRegistries
+	if len(imgcfg.Status.InternalRegistryHostname) > 0 {
+		allowedRegs = append(allowedRegs, imgcfg.Status.InternalRegistryHostname)
+	}
 	// Get ControllerConfig
 	controllerConfig, err := ctrl.ccLister.Get(ctrlcommon.ControllerConfigName)
 	if err != nil {
@@ -712,7 +717,7 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		}
 		if err := retry.RetryOnConflict(updateBackoff, func() error {
 			registriesIgn, err := registriesConfigIgnition(ctrl.templatesDir, controllerConfig, role,
-				imgcfg.Spec.RegistrySources.InsecureRegistries, blockedRegs, imgcfg.Spec.RegistrySources.AllowedRegistries,
+				imgcfg.Spec.RegistrySources.InsecureRegistries, blockedRegs, allowedRegs,
 				imgcfg.Spec.RegistrySources.ContainerRuntimeSearchRegistries, icspRules)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -678,6 +678,11 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 		}
 	}
 
+	// append the internalRegistryHostname to allowedRegistries list
```
mtrmac, line 681: (Absolutely non-blocking I don’t think this comment adds any value right now; the code says just that already. A comment saying _why_ we add it could be useful.)

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -678,6 +678,11 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
+	allowedRegs := imgcfg.Spec.RegistrySources.AllowedRegistries
+	if len(imgcfg.Status.InternalRegistryHostname) > 0 {
```
mtrmac, line 683: Shouldn’t this happen only if `len(allowedRegs) > 0`? Otherwise this turns an “allow all” configuration into “allow only the internal registry”, with no way to achieve “allow all”.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -678,6 +678,11 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
+		allowedRegs = append(allowedRegs, imgcfg.Status.InternalRegistryHostname)
+	}
```
mtrmac, line 685: Please organize this code to only have one copy of the logic; at least a helper function like `getValidBlockedRegs`, or ideally in a way that is impossible to bypass by mistake (_maybe_ pass all of `imgcfg` into `registriesConfigIgnition` and move the log somewhere inside that call subtree).

mtrmac (commented):
The bug references a discussion about changing the end-user documentation instead. Why hard-code this instead? It’s not obviously the right thing to do.

To the extent this is justified by (guessing) ”we must allow pulling from OpenShift product repositories for OpenShift to work at all”:
- That’s not really what the reported bug is about: the cluster seems to work well enough, the concern is with some default templates. I’m not saying those are OK to just break but they are less important that the core product images.
- … the cluster seems to work well enough because the core cluster images are delivered though `quay.io/openshift-release-dev/`, possibly remapped through ICSP, and `allowedRegistries` is currently (undocumented?) interpreted _before_ ICSP applies; so the critical repository to auto-add would be something in `quay.io`, not `InternalRegistryHostname` _even if_ the images were hosted on the internal registry (which I don’t think ever happens anyway, or the cluster could not bootstrap). It makes little sense to me to auto-add the secondary template images repository and not auto-add the core repository the product needs to work at all.
- I can also _imagine_ that some sites would not want application developers to rely on the template images, because they only want to build applications on top of their own base images. It’s not obvious to me that auto-adding that repo is the right thing to do — admittedly that’s an unsubstantiated guess.
- https://github.com/openshift/api/blob/d0a89da74761b4b02914a4b1609c216d57132865/config/v1/types_image.go#L110 explicitly says
  > All other registries are denied.
  
  I’m not strongly opposed to automatically allowing repos necessary core product features, but if we do that, all relevant documentation must be updated in tandem.
- Compare also the discussion about allowing namespace/repo scopes in allowed/blocked registries (Cc: @umohnani8 ); if we decide to auto-add the core OCP repos under `quay.io/openshift-release-dev`, it would make much more sense to only allow that namespace and not _all_ of `quay.io`. If so, it would probably be better to implement the namespace allowing features first, and only then auto-add the smaller repo, rather than auto-add all of `quay.io` now, and possibly break clusters on upgrades if we auto-add a smaller scope in the future.

---

`Status.InternalRegistryHostname` is documented to be set by the registry operator, i.e. presumably it is not set during bootstrap already. If that’s the case (is it?), doesn’t this cause an extra node reboot during bootstrap?

PR #2578, opened 2021-05-13
Clean up keepalived log messages

cybertron:
The new version of keepalived recommends setting a config value of
max_auto_priority. Since we are not having performance issues with
keepalived to my knowledge, I'm just setting it to -1 to suppress
the message.

In addition, the worker config was missing a couple of settings
related to running check scripts. Those were already present in the
master config so I've added them for workers too.


mandre (commented):
/lgtm

Consistency FTW :+1:

patrickdillon:
LGTM

sinnykumari (approved):

PR #2580, opened 2021-05-19
Bug 1966602: Don't explicitly set the IPv6DualStack feature gate

danwinship:
It's on by default now, and setting it explicitly may result in an initial kubelet config that doesn't match the post-bootstrap one.

I think we're not hitting this in CI because we're still explicitly setting the feature gate there...

### DIFF
--- a/templates/master/01-master-kubelet/_base/files/kubelet.yaml
+++ b/templates/master/01-master-kubelet/_base/files/kubelet.yaml
@@ -26,9 +26,6 @@ contents:
       ephemeral-storage: 1Gi
     featureGates:
       APIPriorityAndFairness: true
-{{- if eq .IPFamilies "DualStack"}}
-      IPv6DualStack: true
-{{- end}}
       LegacyNodeRoleBehavior: false
       NodeDisruptionExclusion: true
       RotateKubeletServerCertificate: true
--- a/templates/worker/01-worker-kubelet/_base/files/kubelet.yaml
+++ b/templates/worker/01-worker-kubelet/_base/files/kubelet.yaml
@@ -26,9 +26,6 @@ contents:
       ephemeral-storage: 1Gi
     featureGates:
       APIPriorityAndFairness: true
-{{- if eq .IPFamilies "DualStack"}}
-      IPv6DualStack: true
-{{- end}}
       LegacyNodeRoleBehavior: false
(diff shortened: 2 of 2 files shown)


danwinship:
dev-scripts fix: https://github.com/openshift-metal3/dev-scripts/pull/1242
Whichever one merges first will break dual-stack CI. There is no good way to avoid this.

dcbw:
Looks like we're now failing metal dual-stack tests in ovnkube due to this:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ovn-kubernetes/551/pull-ci-openshift-ovn-kubernetes-master-e2e-metal-ipi-ovn-dualstack/1396842961500639232

```
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: May 24 16:22:41.626: Some cluster operators are not ready: kube-apiserver (Upgradeable=False FeatureGates_RestrictedFeatureGates_IPv6DualStackNoUpgrade: FeatureGatesUpgradeable: "IPv6DualStackNoUpgrade" does not allow updates)
```

kikisdeliveryservice:
@danwinship is there a BZ for this? or will it wait?

danwinship:
/assign @kikisdeliveryservice 
oops, belatedly created the bz. This is needed for 4.8

danwinship:
Bot got fooled by a comment in the dev-scripts PR. This is still needed

kikisdeliveryservice:
PTAL needs kubelet review

/assign  @rphillips @sjenning @mrunalp

danwinship:
> PTAL needs kubelet review
> 
> /assign @rphillips @sjenning @mrunalp

It doesn't really need _much_ kubelet review; all the patch does is remove a feature-gate which is now enabled by default anyway, reverting https://github.com/openshift/machine-config-operator/pull/2277/commits/a0c44de31e93f323b8a8abb015695f89a128a3db.

(Without this patch, the bootstrap-time kubelet config won't match the runtime kubelet config now that we are no longer explicitly adding the feature gate to the runtime kubelet config, which will then mean that the node will fail to come up because it computes the wrong hash for its MachineConfig or whatever.)

mrunalp (approved):

PR #2581, opened 2021-05-19
Bug 1962288: [ovirt] Stop using db file for static internal records

cybertron:
The db file does not allow fallthrough for queries in the cluster
domain that cannot be answered by our coredns. To get around this,
we can switch to different plugins that do have a fallthrough
option. This makes the ovirt configuration the same as vsphere.


kikisdeliveryservice:
they can fully approve and merge this

/assign  @Gal-Zaidman @rgolangh

Gal-Zaidman:
/lgtm

I see that this is exactly the same file as vsphere and the tests are passing

PR #2583, opened 2021-05-20
Bug 1958812: daemon: Change runGetOut to not intermix stdout/stderr

cgwalters:
Using `CombinedOutput()` here is wrong since e.g. for invoking
things like `rpm-ostree status --json` we don't want to try to
ever parse stderr text.

Instead, propagate the stderr text into the error, so it
will be reported up into the daemon logs where we can see the
actual error text.

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -253,13 +253,28 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 	return
 }
 
+func truncate(input string, length int) string {
+	asRunes := []rune(input)
+
+	if length > len(asRunes) {
+		length = len(asRunes)
+	}
+
+	return string(asRunes[:length])
+}
+
 // runGetOut executes a command, logging it, and return the stdout output.
 func runGetOut(command string, args ...string) ([]byte, error) {
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	rawOut, err := cmd.CombinedOutput()
+	rawOut, err := cmd.Output()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error running %s %s: %s", command, strings.Join(args, " "), string(rawOut))
+		errtext := ""
+		if e, ok := err.(*exec.ExitError); ok {
+			// Trim to max of 256 characters
+			errtext = fmt.Sprintf("\n%s", truncate(string(e.Stderr), 256))
+		}
+		return nil, fmt.Errorf("error running %s %s: %s%s", command, strings.Join(args, " "), err, errtext)
 	}
 	return rawOut, nil
 }
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -647,3 +647,25 @@ func checkIrreconcilableResults(t *testing.T, key string, reconcilableError erro
 		t.Errorf("Different %s values should not be reconcilable.", key)
 	}
 }
+
+func TestRunGetOut(t *testing.T) {
+	o, err := runGetOut("true")
+	assert.Nil(t, err)
+	assert.Equal(t, len(o), 0)
+
+	o, err = runGetOut("false")
+	assert.NotNil(t, err)
+
+	o, err = runGetOut("echo", "hello")
+	assert.Nil(t, err)
+	assert.Equal(t, string(o), "hello\n")
+
+	// base64 encode "oops" so we can't match on the command arguments
+	o, err = runGetOut("/bin/sh", "-c", "echo hello; echo b29wcwo= | base64 -d 1>&2; exit 1")
+	assert.Error(t, err)
+	errtext := err.Error()
+	assert.Contains(t, errtext, "exit status 1\noops\n")
+
+	o, err = runGetOut("/usr/bin/test-failure-to-exec-this-should-not-exist", "arg")
+	assert.Error(t, err)
+}


pkg/daemon/rpm-ostree.go
```diff
@@ -253,13 +253,28 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
 // runGetOut executes a command, logging it, and return the stdout output.
 func runGetOut(command string, args ...string) ([]byte, error) {
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	rawOut, err := cmd.CombinedOutput()
+	rawOut, err := cmd.Output()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error running %s %s: %s", command, strings.Join(args, " "), string(rawOut))
+		errtext := ""
+		if e, ok := err.(*exec.ExitError); ok {
```
yuqi-zhang, line 273: On the off chance that this assertion doesn't hold, should we put an else condition to error more gracefully?

yuqi-zhang (commented):
Generally makes sense. To make sure I understand this correctly, in the bug we were basically erroneously parsing the stdout/stderr from runGetOut, such that it was actually returning an error, but it returned it as stdout instead?

Also 2 quick questions:

pkg/daemon/rpm-ostree.go
```diff
@@ -253,13 +253,28 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 	return
 }
 
+func truncate(input string, length int) string {
+	asRunes := []rune(input)
```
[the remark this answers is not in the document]
cgwalters: https://stackoverflow.com/a/56129336/12240197
This helps ensure we don't incorrectly cut off between UTF-8 codepoints.

I commented on internal chat, this is a perfect example of the superficial simplicity of Go materializing as debt paid later.  Go just treats `[]byte` and `string` as [interchangable types](https://blog.golang.org/strings), which is convenient...but also makes it easy to do the wrong thing.  Rust in contrast for example doesn't allow indexing `&str`.  Though I just learned `String::truncate()` exists but is [also a trap](https://stackoverflow.com/questions/38461429/how-can-i-truncate-a-string-to-have-at-most-n-characters).

pkg/daemon/rpm-ostree.go
cgwalters: In that case we'll omit the standard error text, which seems fine?  What would be in the `else`?

I think it's possible to get a different error if we fail to `exec()` instead of the child process itself failing, so it can happen.  But I think there'd just be an extra `: ` in the error message.

But I cleaned this up and added a test for it too.

kikisdeliveryservice (approved):
propagating the stderr text into the error will be super helpful!

pkg/daemon/rpm-ostree.go
```diff
@@ -253,13 +253,28 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
...
 	glog.Infof("Running captured: %s %s", command, strings.Join(args, " "))
 	cmd := exec.Command(command, args...)
-	rawOut, err := cmd.CombinedOutput()
+	rawOut, err := cmd.Output()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error running %s %s: %s", command, strings.Join(args, " "), string(rawOut))
+		errtext := ""
+		if e, ok := err.(*exec.ExitError); ok {
+			// Trim to max of 256 characters
+			errtext = fmt.Sprintf("\n%s", truncate(string(e.Stderr), 256))
```
sinnykumari, line 275: can we keep all error message. This will help to keep complete error messages captured when something fails.
Like, longer error message from one the bug we have seen:
```
E1113 00:46:05.603931    3636 writer.go:135] Marking Degraded due to: error running rpm-ostree override remove kernel kernel-core kernel-modules kernel-modules-extra --install kernel-rt-core --install kernel-rt-modules --install kernel-rt-modules-extra --install kernel-rt-kvm: Checking out tree 859d427...done
Enabled rpm-md repositories: coreos-extensions
rpm-md repo 'coreos-extensions' (cached); generated: 2020-11-11T18:27:37Z
Importing rpm-md...done
Resolving dependencies...done
error: Could not depsolve transaction; 4 problems detected:
 Problem 1: conflicting requests
  - nothing provides linux-firmware >= 20200619-99.git3890db36 needed by kernel-rt-core-4.18.0-240.1.1.rt7.55.el8_3.x86_64
 Problem 2: package kernel-rt-modules-extra-4.18.0-240.1.1.rt7.55.el8_3.x86_64 requires kernel-rt-uname-r = 4.18.0-240.1.1.rt7.55.el8_3.x86_64, but none of the providers can be installed
  - conflicting requests
  - nothing provides linux-firmware >= 20200619-99.git3890db36 needed by kernel-rt-core-4.18.0-240.1.1.rt7.55.el8_3.x86_64
 Problem 3: package kernel-rt-modules-4.18.0-240.1.1.rt7.55.el8_3.x86_64 requires kernel-rt-uname-r = 4.18.0-240.1.1.rt7.55.el8_3.x86_64, but none of the providers can be installed
  - conflicting requests
  - nothing provides linux-firmware >= 20200619-99.git3890db36 needed by kernel-rt-core-4.18.0-240.1.1.rt7.55.el8_3.x86_64
 Problem 4: package kernel-rt-kvm-4.18.0-240.1.1.rt7.55.el8_3.x86_64 requires kernel-rt = 4.18.0-240.1.1.rt7.55.el8_3, but none of the providers can be installed
  - conflicting requests
  - nothing provides linux-firmware >= 20200619-99.git3890db36 needed by kernel-rt-core-4.18.0-240.1.1.rt7.55.el8_3.x86_64
: exit status 1
```

sinnykumari:
Looking at the bug, it is surprising to see the error message that it failed to unmarshal output. Ideally, if `rpm-ostree status --json` command fails then we would get `err! = nil` should have handled it. But looks like, in this case something unrelated in between got printed on stderr that got captured by `cmd.CombinedOutput()` andn hence Unmarshal failed ?

pkg/daemon/rpm-ostree.go
kikisdeliveryservice: @sinnykumari (apologies for the potentially simple question) but wouldn't that error be contained/returned in rawOut (below)

cgwalters: Hmm.  Right, so at 256 characters that would be truncated to something like:
```
error: Could not depsolve transaction; 4 problems detected:
 Problem 1: conflicting requests
  - nothing provides linux-firmware >= 20200619-99.git3890db36 needed by kernel-rt-core-4.18.0-240.1.1.rt7.55.el8_3.x86_64
```

Which isn't bad?

Actually though...why are we using `runGetOut()` for that?  Take this one for example:
https://github.com/openshift/machine-config-operator/blob/c415ce6aed25604bc1d2478951db16759dac31f6/pkg/daemon/update.go#L987
There we're gathering the stdout (and stderr) to a pipe, but just throwing it away.

I think perhaps the best is to pass through stdout (so regular information messages go to the MCD pod log) and capture stderr - write the whole stderr to our stderr, and also capture e.g. the last chunk of it here (which actually is similar to what this https://golang.org/pkg/os/exec/#ExitError is already doing, though it doesn't define "long" but from a look at current Go sources it seems to be 64k, which is actually wayyy too long to include in something like status text).

cgwalters:
> Looking at the bug, it is surprising to see the error message that it failed to unmarshal output. Ideally, if rpm-ostree status --json command fails then we would get err! = nil should have handled it. 

That's the problem - stderr is going into the same bucket as stdout right now, which we're just throwing away because the child process exited with an error.

> But looks like, in this case something unrelated in between got printed on stderr/stdout that got captured by cmd.CombinedOutput() and hence Unmarshal failed ?

Actually wait, you must be right - I had thought the command was exiting with an error, but that seems to not be the case because otherwise we'd have propagated that error.

But either way, we shouldn't be intermixing stdout and stderr when we're actually going to parse stdout.

pkg/daemon/rpm-ostree.go
sinnykumari: I think it was just laziness that we have been reusing runGetOut() in both cases (just executing command and for parsing output). Thanks for creating https://github.com/openshift/machine-config-operator/pull/2587

yuqi-zhang:
> Actually wait, you must be right - I had thought the command was exiting with an error, but that seems to not be the case because otherwise we'd have propagated that error.
> But either way, we shouldn't be intermixing stdout and stderr when we're actually going to parse stdout.

Should we then also add a line to https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/pkg/daemon/rpm-ostree.go#L89 to return `output` and not just `err`, so we can see what we failed to Unmarshal?

cgwalters:
Let's start this ball rolling, this is an easy low risk backport.
/cherrypick release-4.7

cgwalters:
OK right, it looks like we already have the changes on 4.8, just need to retarget the BZ.

PR #2585, opened 2021-05-21
Bug 1955929: operator/sync.go confirm renderedconfig osimageurl matches cvo

kikisdeliveryservice:
Add check to operator to verify that rendered config osImageURL matches the url in CVO/rendered config/cc. sync.go uses isMachineConfigPoolConfigurationValid() which checks the pool sources for GeneratedByVersionAnnotationKey - but osimageurl is not a source, so let's explicitly add a check since they also need to match to verify the rendered config is the most recently rendered. This prevents the operator from reporting upgraded before the controller rerenders a new config when there is no new MCO commit and only an osImageURL change. Also add unit test.

Also add event when a new rendered config is created to further debug this bug (and also help in debugging other issues when previously logs are lost post-reboots). Finally removed some excessive logging from operator that was added to debug.

This PR doesn't close this bug, but would like to get this merged to get some data points.


kikisdeliveryservice:
ERRO[2021-05-21T03:31:06Z] level=error msg=	status code: 400, request id: d146308d-6b2f-4168-8011-c808d0f6feba 
ERRO[2021-05-21T03:31:06Z] level=error                                  
ERRO[2021-05-21T03:31:06Z] level=error msg=  on ../tmp/openshift-install-031473860/vpc/master-elb.tf line 45, in resource "aws_lb_target_group" "api_internal": 
ERRO[2021-05-21T03:31:06Z] level=error msg=  45: resource "aws_lb_target_group" "api_internal" { 

/test e2e-aws
/test e2e-vsphere-upgrade

kikisdeliveryservice:
LB issues 

/test e2e-aws-serial

kikisdeliveryservice:
shouldnt have refactored ci has api-machinery failures :/

/retest

kikisdeliveryservice:
Remove 1 more logging line that wasn't useful and was piling up in mco logs.

pkg/controller/render/render_controller.go
```diff
@@ -496,6 +496,7 @@ func (ctrl *Controller) syncGeneratedMachineConfig(pool *mcfgv1.MachineConfigPoo
 	if apierrors.IsNotFound(err) {
 		_, err = ctrl.client.MachineconfigurationV1().MachineConfigs().Create(context.TODO(), generated, metav1.CreateOptions{})
 		glog.V(2).Infof("Generated machineconfig %s from %d configs: %s", generated.Name, len(source), source)
+		ctrl.eventRecorder.Eventf(pool, corev1.EventTypeNormal, "RenderedConfigGenerated", "%s successfully generated", generated.Name)
```
cgwalters, line 499: Nit: Not a new issue, but we'll log this even if an error occurs which could be confusing, maybe hoist up the `if err != nil` check to be before this?

kikisdeliveryservice:
oh boy.. what did i do now.. :/

cgwalters (approved):
Based on my understanding so far, I think this is will close the race and it's pretty self-contained and easy to validate. And so it'll also be easy to backport which is good.

Longer term I think we could clean this up though - something like having the operator also explicitly wait for the render controller to synchronize.  Another alternative would be to inject our version as an annotation into the rendered MC.

I think it'd be good add a new mco e2e that explicitly exercises the OS update path, but as far as testing this, if we see the new error message in logs then that means the race would have been hit which should be sufficient.

pkg/controller/render/render_controller.go
kikisdeliveryservice: hmm good call!

kikisdeliveryservice: wait i think i need to keep the old one _and_ add the new one in the if?

cgwalters: Yeah we need both, we want to return an error if create fails, or if there was some other error.

kikisdeliveryservice: done!

kikisdeliveryservice:
Tweaked a commit message, no substantive changes. This is ready for final lgtm.

@cgwalters could you do the honors?

cgwalters (approved):
/lgtm

One thing we need to keep in mind is that the current `e2e-aws` doesn't do upgrades, and `e2e-agnostic-upgrade` doesn't do OS updates.  So the place to watch for this after landing will be around things like https://prow.ci.openshift.org/?job=promote-release-openshift-machine-os-content-e2e-aws-4.8

kikisdeliveryservice:
> /lgtm
> 
> One thing we need to keep in mind is that the current `e2e-aws` doesn't do upgrades, and `e2e-agnostic-upgrade` doesn't do OS updates. So the place to watch for this after landing will be around things like https://prow.ci.openshift.org/?job=promote-release-openshift-machine-os-content-e2e-aws-4.8

perfect! I'll check that out and some of the periodic jobs. thanks!! :+1:

kikisdeliveryservice:
e2e-aws-disruptive and e2e-aws-serial both failing on unrelated:
```[sig-instrumentation][Late] Alerts shouldn't report any alerts in firing or pending state apart from Watchdog and AlertmanagerReceiversNotConfigured and have no gaps in Watchdog firing [Suite:openshift/conformance/parallel]```

/retest

kikisdeliveryservice:
(updated the attached bz to leave the original bz open -as there is another non-mco bug blocking overall verification - and use the new bz tailored to the osimageurl race so we can backport it quickly)

PR #2586, opened 2021-05-21
Bug 1916890: Stop setting nopreempt on bootstrap keepalived.conf

cybertron:
The nopreempt option has the opposite effect to what we want on the
bootstrap. In our current LB architecture, we want the VIP to be on
the bootstrap any time the bootstrap api is running. nopreempt
breaks this by not allowing the VIP to move back to the bootstrap
if there happens to be a blip in the bootstrap apiserver that causes
keepalived to stop temporarily while one of the masters is running.

The following sequence of events is the problematic one:
1) A master keepalived instance comes up
2) There is a temporary outage of the bootstrap apiserver.
3) Bootstrap keepalived-monitor notices the outage and stops the
   bootstrap keepalived.
4) The master keepalived takes the VIP
5) The bootstrap apiserver comes back up.
6) Bootstrap keepalived-monitor notices that the api is back and
   restarts keepalived
7) nopreempt stops bootstrap keepalived from taking the API VIP
   back, which wedges the deployment because the VIP remains
   pointing at a node without an api available.

Note that the bootstrap priority is always higher than the master
priority, so any time the bootstrap is active it should hold the VIP.

Dropping nopreempt allows the VIP to move back to the bootstrap node
in this scenario which allows the deployment to continue.

**- What I did**

**- How to verify it**

Fixed a bug with VIP handling during bootstrap that could prevent a deployment from completing.

### DIFF
--- a/manifests/on-prem/keepalived.conf.tmpl
+++ b/manifests/on-prem/keepalived.conf.tmpl
@@ -10,7 +10,6 @@
     virtual_router_id {{.Cluster.APIVirtualRouterID }}
     priority 70
     advert_int 1
-    nopreempt
     {{ if .EnableUnicast }}
     unicast_src_ip {{.NonVirtualIP}}
     unicast_peer {
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -1797,7 +1797,6 @@ var _manifestsOnPremKeepalivedConfTmpl = []byte(`# Configuration template for Ke
     virtual_router_id {{.Cluster.APIVirtualRouterID }}
     priority 70
     advert_int 1
-    nopreempt
     {{ if .EnableUnicast }}
     unicast_src_ip {{.NonVirtualIP}}
     unicast_peer {


cybertron:
/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere

Not cc'ing anyone yet because I want to make sure this doesn't somehow break something.

staebler:
This looks good to me. Nice catch.

cybertron:
/retest

Looks like bootstrapping was fine on the failed jobs.

cybertron:
/retest

metal-ipi should be working again. Let's see how the other jobs do.

mandre (commented):
Wow, ok. We've also observed this behavior in OpenStack deployments and could not pin the root cause. Thanks for the bug fix, it looks good to me. I bet we'll be able to close a few bugs as dups too.

/lgtm
/retest

cybertron:
/cc @Gal-Zaidman @jcpowermac @patrickdillon
/test e2e-vsphere

The vsphere failures have been something to do with DNS that seems unrelated to what this patch does on the bootstrap node.

jcpowermac:
/lgtm

We fixed the issue with DNS this morning

mandre:
Per the bot's suggestion,
/assign @sinnykumari

We'd really like to see this one merge :wink:

sinnykumari (approved):
LGTM

sinnykumari:
seems like bot is busy, trying again
/approve

PR #2587, opened 2021-05-21
Bug 1958812: daemon: Rework rpm-ostree synchronous invocation logging

cgwalters:
This is the dual/inverse of https://github.com/openshift/machine-config-operator/pull/2583
Replace all the parts where we were using `runGetOut` but not actually processing
the output with a function that:

- Writes rpm-ostree stdout to our stdout
- Captures stderr for propagation into the error we return

Instead of gathering the data from both streams into a pipe but then
discarding it.

(Now the daemon also logs into the journal, but there's not much
 harm from capturing both client and daemon output)

### DIFF
--- a/pkg/daemon/rpm-ostree.go
+++ b/pkg/daemon/rpm-ostree.go
@@ -77,6 +77,13 @@ func NewNodeUpdaterClient() NodeUpdaterClient {
 	return &RpmOstreeClient{}
 }
 
+// Synchronously invoke rpm-ostree, writing its stdout to our stdout,
+// and gathering stderr into a buffer which will be returned in err
+// in case of error.
+func runRpmOstree(args ...string) error {
+	return runCmdSync("rpm-ostree", args...)
+}
+
 // GetBootedDeployment returns the current deployment found
 func (r *RpmOstreeClient) GetBootedDeployment() (*RpmOstreeDeployment, error) {
 	var rosState rpmOstreeState
@@ -245,7 +252,7 @@ func (r *RpmOstreeClient) Rebase(imgURL, osImageContentDir string) (changed bool
 	args := []string{"rebase", "--experimental", fmt.Sprintf("%s:%s", repo, ostreeCsum),
 		"--custom-origin-url", customURL, "--custom-origin-description", "Managed by machine-config-operator"}
 
-	if _, err = runGetOut("rpm-ostree", args...); err != nil {
+	if err = runRpmOstree(args...); err != nil {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/daemon/rpm-ostree.go
```diff
@@ -77,6 +78,22 @@ func NewNodeUpdaterClient() NodeUpdaterClient {
 	return &RpmOstreeClient{}
 }
 
+// Synchronously invoke rpm-ostree, writing its stdout to our stdout,
+// and gathering stderr into a buffer which will be returned in err
+// in case of error.
+func runRpmOstree(args ...string) error {
```
sinnykumari, line 84: Perhaps we can rename it to something more generic like `runCmd(command string, args ...string)` or something similar. We are using runGetOut() function today for running non rpm-ostree command as well like https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L105 .

cgwalters: I looked at this and updated the commit message around this too.  In the systemctl case...it doesn't really print to stdout, but the extra logging we do makes sense I think for this one.  So updated!

cgwalters: (I kept the `runRpmOstree()` thin wrapper because we call it often and we may want to do something else there in the future)

pkg/daemon/update.go
```diff
@@ -1857,6 +1848,22 @@ PENDING=%d`, pendingStateMessageID, pending.GetName(), dn.bootID, isPending))
...
+// Synchronously invoke a command, writing its stdout to our stdout,
+// and gathering stderr into a buffer which will be returned in err
+// in case of error.
+func runCmdSync(cmdName string, args ...string) error {
+	glog.Infof("Running: %s %s", strings.Join(args, " "))
+	cmd := exec.Command(cmdName, args...)
+	var stderr bytes.Buffer
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Wait(); err != nil {
```
sinnykumari, line 1860: did we mean to call here cmd.Run() instead?

sinnykumari:
@cgwalters when you get time, please update the PR so that this goes in as well with other related PRs that got merged :slightly_smiling_face:

cgwalters:
(Sorry, missed notifications from this)

- :surfer: Rebased
- :white_check_mark:  Comments addressed
- :heavy_check_mark:  Self-referential checklist created

kikisdeliveryservice:
gcp-op :
``` ERRO[2021-06-09T18:33:17Z] error: Failed to acquire resource, current capacity: 0 free, 70 leased 
INFO[2021-06-09T18:33:17Z] Ran for 2h5m9s                               
ERRO[2021-06-09T18:33:17Z] Some steps failed:                           
ERRO[2021-06-09T18:33:17Z]   * could not run steps: step e2e-gcp-op failed: failed to acquire lease for "gcp-quota-slice": resources not found 
```
:-1:

PR #2588, opened 2021-05-21
Bug 1958812: daemon: Display more non-JSON text from rpm-ostree status --json

cgwalters:
If we get an error deserializing, let's log a bit of the actual output
so we can better figure out what's going on there.

xref https://bugzilla.redhat.com/show_bug.cgi?id=1958812


cgwalters:
Rebased :surfer: and merged the truncate functions.

kikisdeliveryservice:
@sinnykumari can you PTAL since you were reviewing the other PR?

PR #2594, opened 2021-06-02
[WIP] Bug 1952694: bump drain backoff

kikisdeliveryservice:
[WIP] As per discussion today, drain backoff time is too short and causing issues. I don't believe we ever touched these times before, nor had a good reason why they were set the way they are set.

Not convinced exponential backoff is the right thing here, experimenting a bit and will pick up tomorrow.

### DIFF
--- a/pkg/daemon/drain.go
+++ b/pkg/daemon/drain.go
@@ -46,9 +46,10 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
 
 func (dn *Daemon) drain() error {
 	backoff := wait.Backoff{
-		Steps:    5,
+		Steps:    8,
 		Duration: 10 * time.Second,
 		Factor:   2,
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
Note: generally neutral to these timings, if you'd like something else, wfm

kikisdeliveryservice:
Thinking about this more this should be lower threshold since this is for 1 mcd...

yuqi-zhang:
One thing to consider is, near the end we are waiting multiple hours before retrying, so in case of e.g. CI workloads (5hr), we are maybe waiting two hours for the MCD to retry the operation and report success after the last pod has drained.

Maybe we should cap the retry timeout at 15 mins or something like that? Reading https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Backoff I think it will just stop retrying if I understood that correctly. Maybe with a outer retry loop somehow?

PR #2596, opened 2021-06-03
Bug 1943315: daemon: make node drain logic smarter for ICSP changes

sinnykumari:
Today, we perform drain for any changes made into container registry config file /etc/containers/registries.conf.
    
- To avoid users workload disruption, we can safely skip drain for some of the changes made into the config file. Such as:
    1. A new mirror is added to an existing registry that has `mirror-by-digest-only=true`
    2. A new registry has been added that has `mirror-by-digest-only=true`
    
- If any other unsafe changes are detected in registry conf, MCD will default to perform node drain.

- Also added unit tests.

### DIFF
--- a/pkg/daemon/drain_test.go
+++ b/pkg/daemon/drain_test.go
@@ -1,0 +1,359 @@
+package daemon
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/vincent-petithory/dataurl"
+)
+
+func TestIsDrainRequired(t *testing.T) {
+	machineConfigs := map[string]*mcfgv1.MachineConfig{
+		"mc1": helpers.NewMachineConfig("01-test", nil, "dummy://", []ign3types.File{{
+			Node: ign3types.Node{
+				Path: "/etc/containers/registries.conf",
+			},
+			FileEmbedded1: ign3types.FileEmbedded1{
+				Contents: ign3types.Resource{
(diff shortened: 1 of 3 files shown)


sinnykumari:
/cc @mrunalp @umohnani8 
can you please also take a look to make sure we are not adding any false positive or negative.

kikisdeliveryservice:
Wow Sinny. :heart: 

Is this really a bug though?  Feels like it should've gone through other channels...

sinnykumari:
> Wow Sinny. heart
> 
> Is this really a bug though? Feels like it should've gone through other channels...

Yeah, it is not a bug but was good to have in 4.8 if possible for  :)

pkg/daemon/drain_test.go
```diff
@@ -0,0 +1,343 @@
...
+		},
+		{
+			// skip drain: only additional unqualified-search-registries has been added
+			actions:        []string{postConfigChangeActionReloadCrio},
+			oldConfig:      machineConfigs["mc3"],
+			newConfig:      machineConfigs["mc4"],
+			expectedAction: false,
+		},
+		{
+			// perform drain: prefix value of one or more registry has changed
```
yuqi-zhang, line 312: Very minor, comparing 7 to 1, I don't think `prefix` value was changed, instead a registries.mirror was added with mirror-by-digest-only true (this is fine) but the other registry's mirror-by-digest-only changed from true to false, which is an unsafe change covered by the last test in the list. Did you mean to use other MCs here?

yuqi-zhang (approved):
Overall looks great, logic was very easy to follow, and amazing test coverage! Just one minor question below, otherwise will LGTM after giving others a chance to take a look. Thanks for working on this!

pkg/daemon/drain_test.go
sinnykumari: good catch Jerry! fixed it by adding a new MC that has only prefix change.

umohnani8:
LGTM. Thanks for working on this!

PR #2597, opened 2021-06-03
Bug 1968019: Bump drain timeout to 1h

kikisdeliveryservice:
As per retro and subsequent follow-up conversation, make a short term bump to the pool drain degradation to 1hour since the existing times are too short for many users and cause premature erroring/degrading.

In addition:
Refactor drain() since exponential backoff is not suitable for these longer durations. 
Refactor the metric and reduce cardinality, and remove 15min delay on alert.

### DIFF
--- a/pkg/daemon/drain.go
+++ b/pkg/daemon/drain.go
@@ -73,6 +73,7 @@ func (dn *Daemon) drain() error {
 		done <- true
 		failMsg := fmt.Sprintf("failed to drain node : %s after 1 hour", dn.node.Name)
 		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)
+		MCDDrainErr.Set(1)
 		return errors.New(failMsg)
 	case <-drainer():
 		return nil
@@ -98,7 +99,6 @@ func (dn *Daemon) performDrain() error {
 	}
 
 	// We are here, that means we need to cordon and drain node
-	MCDDrainErr.WithLabelValues(dn.node.Name, "").Set(0)
 	dn.logSystem("Update prepared; beginning drain")
 	startTime := time.Now()
 
@@ -111,7 +111,7 @@ func (dn *Daemon) performDrain() error {
 	dn.logSystem("drain complete")
 	t := time.Since(startTime).Seconds()
 	glog.Infof("Successful drain took %v seconds", t)
-	MCDDrainErr.WithLabelValues(dn.node.Name, "").Set(0)
+	MCDDrainErr.Set(0)
 
 	return nil
 }
--- a/pkg/daemon/drain.go
+++ b/pkg/daemon/drain.go
@@ -45,34 +45,38 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
 }
 
 func (dn *Daemon) drain() error {
-	backoff := wait.Backoff{
-		Steps:    5,
-		Duration: 10 * time.Second,
-		Factor:   2,
-	}
-	var lastErr error
-	if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
-		err := drain.RunNodeDrain(dn.drainer, dn.node.Name)
-		if err != nil {
-			lastErr = err
-			glog.Infof("Draining failed with: %v, retrying", err)
-			return false, nil
-		}
-		return true, nil
+	done := make(chan bool, 1)
 
-	}); err != nil {
-		if err == wait.ErrWaitTimeout {
-			failMsg := fmt.Sprintf("%d tries: %v", backoff.Steps, lastErr)
-			MCDDrainErr.WithLabelValues(dn.node.Name, "WaitTimeout").Set(float64(backoff.Steps))
-			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)
-			return errors.Wrapf(lastErr, "failed to drain node (%d tries): %v", backoff.Steps, err)
-		}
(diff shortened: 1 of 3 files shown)


pkg/daemon/drain.go
```diff
@@ -45,34 +45,30 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
...
-			failMsg := fmt.Sprintf("%d tries: %v", backoff.Steps, lastErr)
-			MCDDrainErr.WithLabelValues(dn.node.Name, "WaitTimeout").Set(float64(backoff.Steps))
-			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)
-			return errors.Wrapf(lastErr, "failed to drain node (%d tries): %v", backoff.Steps, err)
-		}
-		MCDDrainErr.WithLabelValues(dn.node.Name, "UnknownError").Set(float64(backoff.Steps))
-		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", err.Error())
-		return errors.Wrap(err, "failed to drain node")
+	select {
+	case <-time.After(24 * time.Hour):
```
dhellmann, line 64: I'm still relatively new to go, but it looks like if this error case is triggered nothing is going to stop `drainer()`. Is that right, or is there something subtle that I'm not seeing?

kikisdeliveryservice: that was uhh.. not right :)

dhellmann: This looks better. :-) I would probably move line 76 up to happen first in this block, but the order may not actually matter.

kikisdeliveryservice: sure, gave it a whirl :smile:

pkg/daemon/drain.go
```diff
@@ -45,34 +45,39 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
...
-			failMsg := fmt.Sprintf("%d tries: %v", backoff.Steps, lastErr)
-			MCDDrainErr.WithLabelValues(dn.node.Name, "WaitTimeout").Set(float64(backoff.Steps))
-			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", failMsg)
-			return errors.Wrapf(lastErr, "failed to drain node (%d tries): %v", backoff.Steps, err)
-		}
-		MCDDrainErr.WithLabelValues(dn.node.Name, "UnknownError").Set(float64(backoff.Steps))
-		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "FailedToDrain", err.Error())
-		return errors.Wrap(err, "failed to drain node")
+	drainer := func() chan error {
+		ret := make(chan error)
```
dhellmann, line 51: Does something need to close this channel if there's a timeout? Or will the garbage collector do that?

pkg/daemon/drain.go
```diff
@@ -45,34 +45,39 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
...
+		go func() {
+			for {
+				select {
+				case <-done:
+					return
+				default:
+					if err := drain.RunNodeDrain(dn.drainer, dn.node.Name); err != nil {
+						glog.Infof("Draining failed with: %v, retrying", err)
```
dhellmann, line 59: Should this log message include the delay? As written, it implies that the retry might start immediately.

dhellmann (commented):
The logic looks good. Another go noob question about the gc and one nit about a log message.

pkg/daemon/drain.go
kikisdeliveryservice: No it's fine as is and consistent with how it was done before.

pkg/daemon/drain.go
darkmuggle: The channel will be GC'd when `drain()` exists since the scope of `done` is contained entirely in `drain()`

kikisdeliveryservice:
I think I might attach this to a different bug, removing the WIP but adding hold to sort that out. Otherwise, this is tested and ready to go.

/hold

pkg/daemon/drain.go
```diff
@@ -45,34 +45,39 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
...
+					return
+				}
+			}
+		}()
+		return ret
 	}
 
-	return nil
+	select {
+	case <-time.After(1 * time.Hour):
```
darkmuggle, line 72: This is a cleaner way, IMO, than an exponential back-off due to length of time involved. The usefulness of an exponential back-off diminishes when the time scales move much beyond 5min. Further, you'll have the ability to tweak the timeouts easily.

darkmuggle (approved):
IMO, this is easier to ready than an exponential back-off. 

FWIW, /lgtm

pkg/daemon/drain.go
dhellmann: Thanks for explaining!

kikisdeliveryservice: thanks for your review @dhellmann !!

kikisdeliveryservice:
Since this wouldn't close Bugzilla 1952694 fully created a new bug so we can backport this as an intermediate fix and to get signal.

cc: @sdodson

pkg/daemon/drain.go
kikisdeliveryservice: agree was much easier when testing to tweak and much clearer/more predictable in logs etc..

thanks for your help!!

kikisdeliveryservice:
seem to be hitting https://bugzilla.redhat.com/show_bug.cgi?id=1968079 in gcp-op

kikisdeliveryservice:
This has been ready since last week & has been reviewed - can someone please LGTM

PR #2598, opened 2021-06-04
Bug 1946584: Check suffix annotation is a number

QiWang19:
Fix: Bug 1946584 https://bugzilla.redhat.com/show_bug.cgi?id=1946584
When checking whether managedKey has MC name suffix annotation, add validation to confirm it is a number type as we added.

**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/controller/container-runtime-config/container_runtime_config_controller.go
+++ b/pkg/controller/container-runtime-config/container_runtime_config_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -579,8 +580,11 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 			// If the MC name suffix annotation does not exist and the managed key value returned has a suffix, then add the MC name
 			// suffix annotation and suffix value to the ctrcfg object
 			if len(arr) > 4 && !ok {
-				if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
-					return ctrl.syncStatusOnly(cfg, err, "could not update annotation for containerRuntimeConfig")
+				_, err := strconv.Atoi(arr[len(arr)-1])
+				if err == nil {
+					if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
+						return ctrl.syncStatusOnly(cfg, err, "could not update annotation for containerRuntimeConfig")
+					}
 				}
 			}
 		}
--- a/pkg/controller/kubelet-config/kubelet_config_controller.go
+++ b/pkg/controller/kubelet-config/kubelet_config_controller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


kikisdeliveryservice:
node team ptal:

/assign @umohnani8 @rphillips

umohnani8:
LGTM
@QiWang19 can you do the same for containerruntimeconfig controller as well.

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -593,12 +594,15 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
 			}
 			mc.ObjectMeta.UID = uuid.NewUUID()
 			_, ok := cfg.GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
-			arr := strings.Split(managedKey, "-")
 			// If the MC name suffix annotation does not exist and the managed key value returned has a suffix, then add the MC name
 			// suffix annotation and suffix value to the kubelet config object
-			if len(arr) > 4 && !ok {
```
QiWang19, line 599: @umohnani8 PTAL. Is the check for len(arr) > 4 necessary? If the custom pool name has multiple `-`, it will also satisfy this case.

umohnani8: the check was there because the first MC won't have a number suffix so there would be nothing to add to the annotation. But your patch makes sense, we can get rid of this length check

pkg/controller/kubelet-config/kubelet_config_controller.go
```diff
@@ -593,12 +594,15 @@ func (ctrl *Controller) syncKubeletConfig(key string) error {
...
-				if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
-					return ctrl.syncStatusOnly(cfg, err, "could not update annotation for kubeletConfig")
+			if strings.Contains(managedKey, "-generated-kubelet-") && !ok {
```
umohnani8, line 599: We shouldn't need this check as the managed key will always have `generated-kubelet` in it as this controller is for the kubeletconfig stuff only.

pkg/controller/kubelet-config/kubelet_config_controller.go
QiWang19: Thanks for the explanation. I keep the check there to protect the `arr[len(arr)-1])` from the panic.

QiWang19:
@umohnani8 PTAL. Added the same check for containerruntimeconfig controller.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -579,8 +580,11 @@ func (ctrl *Controller) syncContainerRuntimeConfig(key string) error {
 			// If the MC name suffix annotation does not exist and the managed key value returned has a suffix, then add the MC name
 			// suffix annotation and suffix value to the ctrcfg object
 			if len(arr) > 4 && !ok {
-				if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
-					return ctrl.syncStatusOnly(cfg, err, "could not update annotation for containerRuntimeConfig")
+				_, err := strconv.Atoi(arr[len(arr)-1])
+				if err == nil {
+					if err := ctrl.addAnnotation(cfg, ctrlcommon.MCNameSuffixAnnotationKey, arr[len(arr)-1]); err != nil {
+						return ctrl.syncStatusOnly(cfg, err, "could not update annotation for containerRuntimeConfig")
+					}
```
kikisdeliveryservice, line 587: do you need to handle err != nil?

(i dont know, just asking)

kikisdeliveryservice (commented):
does this need updated unit tests or is it covered already?

pkg/controller/container-runtime-config/container_runtime_config_controller.go
QiWang19: No. If the err != nil, the suffix annotation is not a number type, we have to do noting about it since it is not an MC name suffix.

kikisdeliveryservice: ah yes, read too fast. thanks :)

QiWang19:
@umohnani8 PTAL

QiWang19:
@sinnykumari Could you add approved label?

kikisdeliveryservice:
@QiWang19 can you answer plz: https://github.com/openshift/machine-config-operator/pull/2598#pullrequestreview-676557575

QiWang19:
> need updated unit tests

I don't see this worthwhile updating the unit test. It almost just tests whether the go pkg strconv can report an error converting non-int string to int. We don't have to worry about this part.

QiWang19:
@kikisdeliveryservice Can this PR get in?

QiWang19:
Needs approval label.

yuqi-zhang (approved):

PR #2601, opened 2021-06-04
Bug 1969681: maxUnavailable of ds/machine-config-daemon isn't changed even after Bug 1899535 was fixed 

jkyros:
The configuration change fix from Bug #1899535 ( set maxUnavaiable to "10%" instead of "1") worked only against newly installed clusters because the EnsureDaemonSet function in resourcemerge doesn't merge the UpdateStrategy part of the DaemonSet structure. 

This PR makes EnsureDaemonSet additionally merge the UpdateStrategy part of the structure so it also works for cluster upgrades. 

- Fixes #1959258

**- What I did**
- Added code to merge the UpdateStrategy portion of the DaemonSet as part of a resource merge if it has been updated
- Added unit test just to make sure the UpdateStrategy structure gets merged as part of EnsureDaemonSet (if UpdateStrategy has been updated)

**- How to verify it**
- Verified that this works by upgrading 4.5 -> 4.6 -> 4.7 -> this commit (4.8) and verifying that maxUnavailable changed from "1" to "10%", and it did. 

Merge UpdateStrategy as part of DaemonSet resource merge.

### DIFF
--- a/lib/resourcemerge/apps.go
+++ b/lib/resourcemerge/apps.go
@@ -36,5 +36,10 @@ func EnsureDaemonSet(modified *bool, existing *appsv1.DaemonSet, required appsv1
 		existing.Spec.Selector = required.Spec.Selector
 	}
 
+	if !equality.Semantic.DeepEqual(existing.Spec.UpdateStrategy, required.Spec.UpdateStrategy) {
+		*modified = true
+		existing.Spec.UpdateStrategy = required.Spec.UpdateStrategy
+	}
+
 	ensurePodTemplateSpec(modified, &existing.Spec.Template, required.Spec.Template)
 }
--- a/lib/resourcemerge/apps_test.go
+++ b/lib/resourcemerge/apps_test.go
@@ -1,0 +1,147 @@
+package resourcemerge
+
+import (
+	"fmt"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func TestMergeDaemonSetUpdateStrategy(t *testing.T) {
+	// The merge functions set "modified" to true if pointer fields
+	// are nil. Since this test doesn't "zero" all possible pointer
+	// fields, modified will be true, so this test expects this
+	// behavior
+	daemonset_update_strategy_tests := []struct {
+		existing appsv1.DaemonSet
+		input    appsv1.DaemonSet
+
+		expectedModified bool
+		expected         appsv1.DaemonSet
+	}{
+
+		{
+			existing: appsv1.DaemonSet{
+				Spec: appsv1.DaemonSetSpec{
+					UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
+						Type: "RollingUpdate",
+						RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+							MaxUnavailable: &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+							MaxSurge:       &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+						},
+					},
+				},
+			},
+			input: appsv1.DaemonSet{
+				Spec: appsv1.DaemonSetSpec{
+					UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
+						Type: "RollingUpdate",
+						RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+							MaxUnavailable: &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+							MaxSurge:       &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+						},
+					},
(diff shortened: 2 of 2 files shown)


lib/resourcemerge/appt_test.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+		},
+	}
+
+	for idx, test := range daemonset_update_strategy_tests {
+		t.Run(fmt.Sprintf("test#%d", idx), func(t *testing.T) {
+			modified := BoolPtr(false)
+			EnsureDaemonSet(modified, &test.existing, test.input)
+
+			if *modified != test.expectedModified {
```
yuqi-zhang, line 104: So because of the above comment, this will always be `true` right?

I think this is fine as is, but of curiosity how much of the daemonset object would we have to mock for this to work as expected? It looks like ObjectMeta needs labels and annotations, and the Selector needs to be populated. Would that be enough?

yuqi-zhang (approved):
Generally lgtm! One question:

(also regarding bugzilla labels, still looking into the correct way to apply them)

lib/resourcemerge/appt_test.go
```diff
@@ -0,0 +1,113 @@
+package resourcemerge
```
sinnykumari, line 1: typo in the file name, should be apps_test.go

lib/resourcemerge/appt_test.go
```diff
@@ -0,0 +1,113 @@
...
+					UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
+						Type: "OldThing",
+						RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+							MaxUnavailable: &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+							MaxSurge:       &intstr.IntOrString{Type: 0, IntVal: 1, StrVal: ""},
+						},
+					},
+				},
+			},
+			//this "true" is expected true because of the nil pointers elsewhere in the DaemonSet struct
```
sinnykumari, line 47: optional: for consistency and better readability with rest of the repo, it will be nice to add a space in the beginning of comment i.e. after "//".

lib/resourcemerge/appt_test.go
jkyros: Correct. 

ObjectMeta and Selector would be enough if it EnsureDaemonSet didn't also call ensurePodTemplateSpec(), but because it does we have to:

1.  populate ObjectMeta on the template (because it calls EnsureObjectMeta also)
2.  populate the container lists because ensurePodSpec() doesn't ever initialize the existingCurr pointer for either set of containers, so if Spec->Template->PodSpec->InitContainers or Spec->Template->PodSpec->Containers is an empty list, it ends up treating it just like a nil and always returning modified
3. Same as #2, but for Spec->Template->PodSpec->Volumes
4. populate other PodSpec fields because they are are also checked (NodeSelector, SecurityContext,Affinity,Priority, etc) 

I have the test code to mock all of it stashed, so if we really want to mock the whole thing I will, I just was worried that the next poor soul was going to have to hunt through all of that again like I did if the structure changed and they didn't know why the test started failing.

lib/resourcemerge/appt_test.go
jkyros: Whoops, nice catch, I've fixed it, thanks!

lib/resourcemerge/appt_test.go
jkyros: Comment spacing has been fixed, I will remember that for the future :)

lib/resourcemerge/appt_test.go
```diff
@@ -0,0 +1,113 @@
...
+			//it always takes new struct and sets modified to "true" in event of nil
+			expectedModified: true,
+			expected: appsv1.DaemonSet{
+				Spec: appsv1.DaemonSetSpec{
+					UpdateStrategy: appsv1.DaemonSetUpdateStrategy{
+						Type: "OldThing",
```
sinnykumari: Not sure if it is required but I would suggest it is better to use realistic value for Type which can be either "RollingUpdate" or "RollingUpdate". See https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L500

jkyros: I picked something "synthetic" because I was just testing whether the field got updated or not -- in reality my understanding is the RollingUpdate field is only populated if the Type is "RollingUpdate". I didn't want it to be confusing having "realistic" values in a circumstance where they wouldn't make sense (e.g. having the RollingUpdate field populated when set to OnDelete, etc). 

I've added another test to the list that goes between OnDelete and Rolling update and replaced the synthetic values with realistic ones (so now I can both test that the field changes, and have it look real).  If you think it would be better from a best practice standpoint use the consts out of types.go instead of the explicit strings, I can do that too.

sinnykumari: sounds great, thank you for adding a test with realistic data as well. I don't see that added test though, perhaps it is still in your local repo :)

jkyros: Yes, it's in there now, sorry. I figured while I was in there I would would squash also.

PR #2603, opened 2021-06-08
Bug 1817075: operator/controller pods: faster leader elections

yuqi-zhang:
As a temporary solution in lieu of graceful shutdowns, have the
operator and controller pods use faster defaults for lease/renew/
retry, which shouldn't otherwise notably affect operation other
than the new pod acquiring a new lease faster during upgrades,
saving some waiting time.

Example in CVO: https://github.com/openshift/cluster-version-operator/blob/706a0062cade643767e95da2e6142823e8399986/pkg/start/start.go#L51

### DIFF
--- a/cmd/common/helpers.go
+++ b/cmd/common/helpers.go
@@ -16,11 +16,11 @@ import (
 
 const (
 	// LeaseDuration is the default duration for the leader election lease.
-	LeaseDuration = 90 * time.Second
+	LeaseDuration = 30 * time.Second
 	// RenewDeadline is the default duration for the leader renewal.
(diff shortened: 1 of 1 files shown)


deads2k:
This isn't the direction we want to go.  In order to handle short-term network (or other) disruption, all clients (including those using leases) should be able to make it at least 60 seconds without the kube-apiserver available.  To have rapid hand-off in normal cases, clients using leases need to make use of graceful release.

We found this general case (not this one specifically) during single-node, when a 60 second disruption led to mass failures.

Is the alternative to this PR, providing graceful lease release already open so we can assess the size?

PR #2610, opened 2021-06-11
Use openshift-on-prem-infra namespace for all on-prem platforms

cybertron:
While this alone won't solve the problem of unnecessary namespaces
being deployed in clusters that don't need them, at least there will
only be one.

This is a mitigation of https://bugzilla.redhat.com/show_bug.cgi?id=1920199

**- What I did**

**- How to verify it**


pkg/controller/template/render.go
```diff
@@ -381,20 +381,7 @@ func cloudConfigFlag(cfg RenderConfig) interface{} {
 
 func onPremPlatformShortName(cfg RenderConfig) interface{} {
```
cybertron, line 382: This probably needs to be removed completely, but before I go through and touch all the templates I figured I would make the simple change first.

PR #2613, opened 2021-06-14
Ignition 3.3.0 & Ignition Kargs

arithx:
**- What I did**

1. Bump the vendored version of ignition & ign-converter
2. Switch internal usage from v3_2 Ignition structs to v3_3
3. Add support for the MCS to generate Ignition snippets modifying kargs based on the underlying MachineConfig

**- How to verify it**

1. Boot cluster w/ bootimage containing Ignition v3.2.0
2. Boot cluster w/ bootimage containing Ignition v3.3.0
3. Boot cluster w/ bootimage containing Ignition v3.3.0, w/ MachineConfig kargs specified before first boot of a machine. Validate that Ignition set the karg rather than the MCD.

 - Add Ignition v3.3.0 support
 - Utilize Ignition to set kernel arguments during the first boot


pkg/controller/common/helpers.go
```diff
@@ -337,39 +420,46 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+				return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3_0, rptV3_0)
 			}
-			return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.0 failed with error: %v\nReport: %v", errV3_0, rptV3_0)
+			return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3_1, rptV3_1)
 		}
-		return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.1 failed with error: %v\nReport: %v", errV3_1, rptV3_1)
+		return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.2 failed with error: %v\nReport: %v", errV3_2, rptV3_2)
 	}
-	return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.2 failed with error: %v\nReport: %v", errV3_2, rptV3_2)
+	return ign3types.Config{}, errors.Errorf("parsing Ignition config spec v3.3 failed with error: %v\nReport: %v", errV3_3, rptV3_3)
```
mtrmac, line 462: This, and to a smaller extent `ConvertRawExtIgnitionToV3`, is getting unreasonable.

Is there any chance to make this a table-driven loop instead?
```go
for _, attempt := range []struct{attempt func([]byte) (interface{}, error)} {
    func([]byte) (interface{}, error) {
        …
    },
    …
} {
   res, err := attempt(rawIgn)
   if err != nil { return err}
   if res != nil { return res, nil}
}
```
or maybe with the attempt functions out of line?

I don’t expect that we can get rid of all the duplication without generics, but at least expressing the top-level structure of this function and getting the error wrappers closer to the error locations should be possible.

Or maybe this just needs inverting the `if errV3_3.Error() == ign3error.ErrUnknownVersion.Error()` checks, so that the function aborts early, and all of this turns into a long repetitive straight-line code, with no nesting and no `func` intermediates.

kikisdeliveryservice:
@sinnykumari PTAL

pkg/controller/common/helpers.go
```diff
@@ -22,9 +24,12 @@ import (
...
 	ign3_1types "github.com/coreos/ignition/v2/config/v3_1/types"
-	ign3 "github.com/coreos/ignition/v2/config/v3_2"
-	translate3 "github.com/coreos/ignition/v2/config/v3_2/translate"
-	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
+	ign3_2 "github.com/coreos/ignition/v2/config/v3_2"
+	translate3_2 "github.com/coreos/ignition/v2/config/v3_2/translate"
+	ign3_2types "github.com/coreos/ignition/v2/config/v3_2/types"
+	ign3 "github.com/coreos/ignition/v2/config/v3_3_experimental"
+	translate3 "github.com/coreos/ignition/v2/config/v3_3_experimental/translate"
+	ign3types "github.com/coreos/ignition/v2/config/v3_3_experimental/types"
```
yuqi-zhang, line 32: I think instead of moving this to spec 3.3, we should add 3.3 as a separate case for now so the default is still 3.2.

Alternatively, we can move individual items (e.g. ParseAndConvertConfig) to spew out 3.2, such that rendered configs are 3.2

pkg/controller/common/helpers.go
bgilbert: Can we use the new API from https://github.com/coreos/ignition/pull/1193 here?

darkmuggle:
Thank you @arithx for your good work. I've picked up this PR in https://github.com/openshift/machine-config-operator/pull/2675

PR #2615, opened 2021-06-14
[release-4.6] Bug 1927555: [vsphere] set hostname with --static to provide consistent node name for CSR approval

rvanderp3:
Fixes: BZ1927555 - when creating a new machine, the node will get an unexpected hostname

Note: This pull-request is a replacement for PR 2404.  A problem was found in release-4.7 which required a change to PR 2404.

- What I did
Added the --static flag to hostnamectl to prevent DHCP from overriding the hostname sourced from guestinfo. 

- How to verify it

Configure DHCP to assign a hostname as part of the lease
Scale a new node
The node should automatically join the cluster with a node name like <cluster>-<clusterid>-worker-<name>
- Description for the changelog
Resolves issue where the hostname provided by the DHCP overrides the guestinfo hostname thus preventing the machine-api-controller from reconciling new nodes which receive a hostname from DHCP.

### DIFF
--- a/templates/common/vsphere/files/vsphere-hostname.yaml
+++ b/templates/common/vsphere/files/vsphere-hostname.yaml
@@ -4,7 +4,8 @@ contents:
   inline: |
     #!/usr/bin/env bash
     set -e
-
+    # only run if the hostname is not set
+    test -f /etc/hostname && exit 0 || :
     if vm_name=$(/bin/vmtoolsd --cmd 'info-get guestinfo.hostname'); then
-        /usr/bin/hostnamectl --transient set-hostname ${vm_name}
+        /usr/bin/hostnamectl set-hostname --static ${vm_name}
     fi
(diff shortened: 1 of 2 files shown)


rvanderp3:
cc @jcpowermac

jcpowermac:
/lgtm

e2e-vsphere is hitting docker rate limiting but is installing correctly

kikisdeliveryservice:
@rvanderp3 since this only affect vsphere templates, someone from the vsphere owners should be able to approve: https://github.com/openshift/machine-config-operator/blob/b2fec386856a114de16fcf711bcdbd4e08b3031f/OWNERS_ALIASES#L36

kikisdeliveryservice:
doh i guess the owners file is by version too :frowning:

kikisdeliveryservice (approved):

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.70

approved

PR #2618, opened 2021-06-14
Bug 1956836: templates: Rework node-valid-hostname to fix SELinux denial

cgwalters:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1956836

Unfortunately right now, executing `bash` skips a domain transition
(see https://github.com/fedora-selinux/selinux-policy/issues/778)
so the way we're sourcing the script means we stay in `initrc_t`
and end up triggering a SELinux policy denial.

(BTW this denial turns out to just delay the successful exit
 of the script, which will then end up just delaying kubelet start.
 it's otherwise harmless, but we also don't want SELinux policy denials
 in our product by default)

Fix this in two ways:

- First, just move the thing to `/usr/local/bin` to avoid issues
  with labeling of `/usr/local/sbin` that were fixed in https://github.com/openshift/os/pull/551
- Second, rework it to be executed directly

While we're here:

- Clean the confusing+outdated comment about being a NM dispatcher
- Drop the `logger` bit which was only necessary as a NM dispatcher;
  since we're *always* running under systemd, this makes `journalctl -u node-valid-hostname`
  actually show the script output.
- Make it crystal clear that the "truncate hostname" is only run in GCP.
- Fix various typos
- Use the more precise term "non-localhost" in various places instead of the more ambiguous
  terms "real"/"valid"


cgwalters:
/hold
while we review the CI results form this PR on AWS and GCP (two platforms which will exercise the relevant codepaths) - we should see success/failure in the node journals.

cgwalters:
OK so the new code still works:
GCP logs look good:
`Jun 14 22:16:42.102005 ci-op-k5y0btj8-1354f-8rg5c-master-0.c.openshift-gce-devel-ci.int mco-hostname[1280]: ci-op-k5y0btj8-1354f-8rg5c-master-0.c.openshift-gce-devel-ci.internal is longer than 63 characters, using truncated hostname`

AWS logs look good:
`Jun 14 23:07:26.927262 ip-10-0-128-41 mco-hostname[1575]: waiting for non-localhost hostname to be assigned
Jun 14 23:07:26.928323 ip-10-0-128-41 mco-hostname[1575]: node identified as ip-10-0-128-41`

But unfortunately because each `iptables` invocation ends up in the audit log, and we do a *lot* of those, any AVC denials very quickly get rotated out and lost.  This is really useful data we're just throwing away :cry: 

I'm trying to dig at how to filter that stuff out of the audit logs.

darkmuggle (approved):
/lgtm

The fix-ups are nice -- especially dropping things that are not longer needed after the hack of dispatcher scripts were dropped.

cgwalters:
/hold cancel
Verified that there's no more SELinux denial in a fresh GCP cluster launched from this PR.

cgwalters:
@sinnykumari this got a lgtm, mind adding an `/approve` of the general idea?  :pray:

cgwalters:
xref https://github.com/linux-audit/audit-userspace/issues/205

sinnykumari (approved):

PR #2626, opened 2021-06-18
Bug 1971715: configure-ovs: fix nondeterministic master in slave profiles

jcaamano:
**- What I did**
configure-ovs sets the master device name as connection.master on slave
connection profiles when the default gateway being replaced is a master
device.

Those slave profiles may activate before the master profile and if so,
NM may chose non-deterministically any master profile that might provide
a connection for the master device name, which may be the old master
connection profile instead of the replacement connection profile. The
auto-connect priority does not seem to be considered as this might not
be regarded by NM as an auto-connect of the master profile in strict
sense.

Setting the master connection profile uuid as connection.master on slave
connection profiles ensures that the correct master profile is activated
when the slave auto-connects. This does require a bit more effort when
reverting the configuration.

Additionally, if there is any modification to the slave connection
profiles, those need to be persisted as well from
systemConnectionsMerged to system-connections

**- How to verify it**
Run `./configure-ovs  OVNKubernetes` on a device with a bond interface as default gateway, using the following NM configuration:
```
$ cat /etc/NetworkManager/conf.d/99-keyfiles.conf
[keyfile]
path=/etc/NetworkManager/systemConnectionsMerged
```
where `/etc/NetworkManager/systemConnectionsMerged` is an overlay fs as:
```
overlay on /etc/NetworkManager/systemConnectionsMerged type overlay (rw,relatime,seclabel,lowerdir=/etc/NetworkManager/system-connections,upperdir=/run/nm-system-connections,workdir=/run/nm-system-connections-work)
```
Verify that the applied network configuration persists accross reboots.

Improved configuration of OVS with bond interfaces for OVNKuberentes 

fixes: #2519
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1971715

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@redhat.com>

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -4,28 +4,59 @@ contents:
   inline: |
     #!/bin/bash
     set -eux
-    
+
     # This file is not needed anymore in 4.7+, but when rolling back to 4.6
     # the ovs pod needs it to know ovs is running on the host.
     touch /var/run/ovs-config-executed
+
+    if [ -d "/etc/NetworkManager/systemConnectionsMerged" ]; then
+      NM_CONN_PATH="/etc/NetworkManager/systemConnectionsMerged"
+    else
+      NM_CONN_PATH="/etc/NetworkManager/system-connections"
+    fi
+
+    # In RHEL7 files in /{etc,run}/NetworkManager/system-connections end without the suffix '.nmconnection', whereas in RHCOS they end with the suffix.
+    managed_nm_conn_files=($(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection))
 
     # Workaround to ensure OVS is installed due to bug in systemd Requires:
     # https://bugzilla.redhat.com/show_bug.cgi?id=1888017
     copy_nm_conn_files() {
       src_path="/etc/NetworkManager/systemConnectionsMerged"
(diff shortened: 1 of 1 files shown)


trozet (approved):
seems OK to me, but want @msherif1234 to take a look

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -27,6 +27,18 @@ contents:
         done
       fi
     }
+
+    # Used to replace an old master connection uuid with a new one on all connections
+    replace_connection_master() {
+      local old="$1"
+      local new="$2"
+      for conn_uuid in $(nmcli -g UUID connection show) ; do
+        if [ "$(nmcli -g connection.master connection show uuid "$conn_uuid")" != "$old" ]; then
```
trozet, line 36: @jcaamano what if the old connection has a connection.master of device name instead of UUID, should we check that too?

jcaamano: Done, we look for both things.

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -177,25 +204,38 @@ contents:
...
+      trap "handle_exit_error" EXIT
+
+      # Update connections with master property set to use the new connection
+      replace_connection_master $old_conn $new_conn
+      replace_connection_master $iface $new_conn
+
+      # bring down old iface
+      nmcli device disconnect $iface
+
+      # bring up new connectionx
```
trozet, line 238: nit typo on the connectionx

trozet (approved):
/lgtm

sinnykumari (approved):
LGTM. OVS team has already reviewed the functionality

PR #2627, opened 2021-06-18
Bug 1973525: [on-prem] Drop runlevel from infra namespaces

cybertron:
This appears to have been copy-pasta from the MCO namespace, but we
don't need it for the on-prem services.

**- What I did**

**- How to verify it**

Dropped the runlevel specifier for the on-prem namespaces. It was not necessary and may have caused confusion about the services running in the namespace.

### DIFF
--- a/install/0000_80_machine-config-operator_00_namespace.yaml
+++ b/install/0000_80_machine-config-operator_00_namespace.yaml
@@ -24,7 +24,6 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     name: openshift-openstack-infra
-    openshift.io/run-level: "1"
 ---
 apiVersion: v1
 kind: Namespace
@@ -37,7 +36,6 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     name: openshift-kni-infra
-    openshift.io/run-level: "1"
 ---
 apiVersion: v1
 kind: Namespace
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice (approved):
Adding approve and hold remove hold once you get the lgtms you need =)

/hold

mandre (commented):
/test e2e-openstack
LGTM provided e2e-openstack comes back green.

yuvalk:
my PR is probably a dup
https://github.com/openshift/machine-config-operator/pull/2655
but
I think you've missed the run-level 1 for openshift-machine-config-operator

cybertron:
> my PR is probably a dup
> #2655
> but
> I think you've missed the run-level 1 for openshift-machine-config-operator

Possibly. This change came out of a conversation about the infra namespaces, which clearly did not need the runtime attribute. I'm not certain whether that's the case for MCO itself.

kikisdeliveryservice:
> > my PR is probably a dup
> > #2655
> > but
> > I think you've missed the run-level 1 for openshift-machine-config-operator
> 
> Possibly. This change came out of a conversation about the infra namespaces, which clearly did not need the runtime attribute. I'm not certain whether that's the case for MCO itself.

Just to followup, I closed the other PR in favor for this one that only touches on-prem things.

cybertron:
/hold cancel

All affected platforms are passing and we have signoff from 3 of them. Let's get this in.

PR #2630, opened 2021-06-21
Add support for bind-address flag

enxebre:
**- What I did**

**- How to verify it**

Currently the MCS bind itself to 0.0.0.0, this keeps current behaviour by default while letting a consumer to specify a different binding address e.g 127.0.0.1.


sinnykumari:
@cgwalters PTAL

PR #2641, opened 2021-06-25
Bug 1976232: Explicitly set keyfile as the default plugin of NetworkManager for RHEL7

trozet:
1. Configure keyfile as the default plugin of NetworkManager explicitly
for all node. It is the default setting in RHCOS.
2. Fix configure-ovs.sh script for RHEL7 workers with old version
NetworkManager

(cherry picked from commit 5af273d4c986bc7018882afdfeab6d3479469bb6)

### DIFF
--- a/templates/common/_base/files/NetworkManager-keyfiles.yaml
+++ b/templates/common/_base/files/NetworkManager-keyfiles.yaml
@@ -2,6 +2,8 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-keyfiles.conf"
 contents:
   inline: |
+    [main]
+    plugins=keyfile,ifcfg-rh
     [keyfile]
     path=/etc/NetworkManager/system-connections-merged
 
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -16,9 +16,10 @@ contents:
       dst_path="/etc/NetworkManager/system-connections"
       if [ -d $src_path ]; then
         echo "$src_path exists"
-        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
+        # In RHEL7 files in /{etc,run}/NetworkManager/system-connections end without the suffix '.nmconnection', whereas in RHCOS they end with the suffix.
+        fileList=$(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection)
         for file in ${fileList[*]}; do
-          if [ ! -f $dst_path/$file ]; then
+          if [ ! -f $dst_path/$file ] && [ -f $src_path/$file ]; then
             cp $src_path/$file $dst_path/$file
           else
             echo "Skipping $file since it exists in $dst_path"


trozet:
@pliurh  @msherif1234 PTAL

templates/common/_base/files/NetworkManager-keyfiles.yaml
```diff
@@ -2,6 +2,8 @@ mode: 0644
 path: "/etc/NetworkManager/conf.d/99-keyfiles.conf"
 contents:
   inline: |
+    [main]
+    plugins=keyfile,ifcfg-rh
```
cgwalters, line 6: Hmm.  So this is something like the networking layer implicitly grew a dependency on the fact that RHCOS switched to use keyfile as the default, and that broke rhel7 workers?  Note FCOS *only* supports keyfiles...quickly testing it out, looks like this will result in

`Jun 25 15:38:31 cosa-devsh NetworkManager[1358]: <warn>  [1624635511.9838] settings: could not load plugin 'ifcfg-rh' from file '/usr/lib64/NetworkManager/1.30.4-1.fc34/libnm-settings-plugin-ifcfg-rh.so': No such file or directory`
but is otherwise harmless.

It's a bit unfortunate to override the RHCOS config just to fix RHEL7.  Although, perhaps traditional RHEL8 workers would also need this?  In that case, perhaps the correct thing really is to put this config in https://github.com/openshift/openshift-ansible and not in the MCO.

cgwalters: Also, ideally even traditional RHEL9 switches to use keyfiles by default.  I think there was some discussion about changing that in Fedora baseline too, not sure if that went anywhere.

sdodson: That's an RHCOS build time defaulting rather than something encapsulated in machine config?

pliurh: In RHCOS, there is NM config file `/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf` which set keyfile as the default plugin. This file is not injected by MCO. I cannot find such a file on RHEL7 or RHEL8 either.

So it is not to override the RHCOS config, but to align the RHEL NM config with RHCOS.

cgwalters: > That's (keyfile) an RHCOS build time defaulting rather than something encapsulated in machine config?

https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/lib/NetworkManager/conf.d/10-disable-default-plugins.conf

sdodson: Thanks, that was a curiosity for future consideration of what goes into the OS at build time versus via MachineConfig which would be applied to RHEL workers. I think we should move forward with this PR since it aligns with what's already in 4.8.

pliurh: For future improvement, we can either: 
1. remove the `10-disable-default-plugins.conf` for RHCOS and keep the MCO change here, 
2. or keep RHCOS unchanged, make this change to openshift-ansible. 

I prefer the first approach, so that we can have all the NM configurations in one place. WDYT?

sinnykumari:
Putting hold from retesting until it gets either Group Lead or cherry-picked approval
/hold

sinnykumari:
sorry, wrong PR :)
/hold cancel

sdodson:
@sinnykumari @kikisdeliveryservice Can either of you approve this?

sinnykumari (approved):

PR #2642, opened 2021-06-25
 daemon: Dump systemctl status rpm-ostreed on load failure

cgwalters:
daemon: Rename `os` variable to `hostos`

Prep for a future patch; `os` conflicts with the standard Go package.

---

daemon: Dump systemctl status rpm-ostreed on load failure

This is yet another patch which should help debugging
https://bugzilla.redhat.com/show_bug.cgi?id=1958812
like situations in the future.

I also have https://github.com/coreos/rpm-ostree/pull/2932
cooking but it's going to take a while to cycle down.

---

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -201,9 +201,9 @@ func New(
 		err        error
 	)
 
-	os := OperatingSystem{}
+	hostos := OperatingSystem{}
 	if !mock {
-		os, err = GetHostRunningOS()
+		hostos, err = GetHostRunningOS()
 		if err != nil {
 			HostOS.WithLabelValues("unsupported", "").Set(1)
 			return nil, errors.Wrapf(err, "checking operating system")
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -201,19 +201,25 @@ func New(
...
 	}
 
 	// Only pull the osImageURL from OSTree when we are on RHCOS or FCOS
-	if os.IsCoreOSVariant() {
+	if hostos.IsCoreOSVariant() {
 		osImageURL, osVersion, err = nodeUpdaterClient.GetBootedOSImageURL()
 		if err != nil {
+			// If this fails for some reason, let's dump the unit status
+			// into our logs to aid future debugging.
+			cmd := exec.Command("systemctl", "status", "rpm-ostreed")
```
sinnykumari, line 219: Does printing status will capture enough error logs or should we print here rpm-ostreed journal log from current boot?

cgwalters: Status will capture the last few lines of logs at the point in time this happens, which I think will be enough.  

https://github.com/openshift/must-gather/pull/244 will help for the general case.

sinnykumari (approved):
/lgtm

cgwalters:
just to xref, https://github.com/coreos/rpm-ostree/pull/2945 is driving this down into rpm-ostree, so eventually we'll want to remove this code.  But right now, shipping *this* PR involved literally nothing more than me submitting it and having it merge, but changing rpm-ostree in RHEL will take much longer and involve a *lot* more exciting paperwork.

PR #2643, opened 2021-06-28
Bug 1971715: configure-ovs: fix bond ifcfg backed configuration

jcaamano:
Pull request #2626 did not account for a possible scenario where
networking configuration would be backed with ifcfg files instead of
nmconnection files. Attempting to persist slave nmconnection changes in
case of an overlay setup of NM configuration directory would fail.

Additionally opted for a strategy of cloning slave connection profiles
instead of modifying them directly to avoid any interference with MCO
existing configuration.

Also fixed shellcheck reported issues in 4.8 backport pull request #2639.

fixes: #2626
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1971715

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@redhat.com>

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -9,36 +9,81 @@ contents:
     # the ovs pod needs it to know ovs is running on the host.
     touch /var/run/ovs-config-executed
 
-    if [ -d "/etc/NetworkManager/systemConnectionsMerged" ]; then
-      NM_CONN_PATH="/etc/NetworkManager/systemConnectionsMerged"
+    NM_CONN_OVERLAY="/etc/NetworkManager/systemConnectionsMerged"
+    NM_CONN_UNDERLAY="/etc/NetworkManager/system-connections"
+    if [ -d "$NM_CONN_OVERLAY" ]; then
+      NM_CONN_PATH="$NM_CONN_OVERLAY"
     else
-      NM_CONN_PATH="/etc/NetworkManager/system-connections"
+      NM_CONN_PATH="$NM_CONN_UNDERLAY"
     fi
 
     # In RHEL7 files in /{etc,run}/NetworkManager/system-connections end without the suffix '.nmconnection', whereas in RHCOS they end with the suffix.
-    managed_nm_conn_files=($(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection))
+    MANAGED_NM_CONN_FILES=($(echo {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0} {br-ex,ovs-if-br-ex,ovs-port-br-ex,ovs-if-phys0,ovs-port-phys0}.nmconnection))
+    MANAGED_NM_CONN_SUFFIX="-slave-ovs-clone"
 
     # Workaround to ensure OVS is installed due to bug in systemd Requires:
     # https://bugzilla.redhat.com/show_bug.cgi?id=1888017
     copy_nm_conn_files() {
-      src_path="/etc/NetworkManager/systemConnectionsMerged"
-      dst_path="/etc/NetworkManager/system-connections"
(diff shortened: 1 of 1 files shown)


templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -41,6 +45,25 @@ contents:
...
+    # Used to add NM connection files to manage by uuid if any
+    add_managed_nm_conn_file() {
+      local uuid="$1"
+
+      echo "Finding nmconnection file for $uuid"
+      conn_path="$(egrep -l --include=*.nmconnection uuid=$uuid ${NM_CONN_PATH}/* || true)"
+      # there is no corresponding nmconnection file to handle, assume this is backed by an ifcfg file
+      if [ -z "$conn_path" ]; then
+        echo "No nmconnection file found for $uuid, assuming it's backed by an ifcfg file"
+        continue
```
trozet, line 57: should this be a return?

templates/common/_base/files/configure-ovs-network.yaml
```diff
@@ -369,7 +391,7 @@ contents:
       ovs-vsctl --timeout=30 --if-exists del-br br-int -- --if-exists del-br br-local -- --if-exists del-br br-ex
 
       if [[ -n "$iface" ]]; then
-        nmcli device connect $iface
+        nmcli c up ifname $iface
```
trozet, line 394: whats the reason for this change? will this work if the device was disconnected?

jcaamano: For some reason, when config is backed up by ifcfg files and we remove connection profiles, the corresponding device disappears and then we cannot connect by device. So at the end is more reliable to connect by profile and fortunately we can do this in combination with the device name as well.

rbbratta:
This seems to work for the `ifcfg` case with reboots.  I need to do more testing with NetworkManager connection files.

trozet (approved):
/lgtm

ci failures have nothing to do with this

rbbratta:
Will testing I noticed that this doesn't work when the vmxnet3 is bonded and we reboot.

```shell
      # if the interface is of type vmxnet3 add multicast capability for that driver
      # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
      function configure_driver_options {
        intf=$1
        driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
        echo "Driver name is" $driver
        if [ "$driver" = "vmxnet3" ]; then
          ifconfig "$intf" allmulti
        fi
      }.
```

The cluster seems fine without it, so maybe we don't need it anymore.

Filed a medium BZ to track https://bugzilla.redhat.com/show_bug.cgi?id=1982052

sinnykumari:
holding approval until  concern in https://github.com/openshift/machine-config-operator/pull/2643#issuecomment-879599807 has been addressed.

trozet:
> While testing I noticed that this doesn't work when the vmxnet3 is bonded and we reboot.
> 
> ```shell
>       # if the interface is of type vmxnet3 add multicast capability for that driver
>       # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
>       function configure_driver_options {
>         intf=$1
>         driver=$(cat "/sys/class/net/${intf}/device/uevent" | grep DRIVER | awk -F "=" '{print $2}')
>         echo "Driver name is" $driver
>         if [ "$driver" = "vmxnet3" ]; then
>           ifconfig "$intf" allmulti
>         fi
>       }.
> ```
> 
> The cluster seems fine without it, so maybe we don't need it anymore.
> 
> Filed a medium BZ to track https://bugzilla.redhat.com/show_bug.cgi?id=1982052

I think we should keep it, but not configure it for a bond type. That can be a separate PR.

jcaamano:
@sinnykumari @trozet Since it is a superfluous issue, we could use @rbbratta bz to fix it. On the other side, this fixes the original issue and it is important to get it out on 4.9.

sinnykumari:
> @sinnykumari @trozet Since it is a superfluous issue, we could use @rbbratta bz to fix it. On the other side, this fixes the original issue and it is important to get it out on 4.9.

+1
/approve

rbbratta:
Verified with `ifcfg-bond0`, karg `bond=` (and `.nmconnection`), and `ifcfg` `teamd` configs on vSphere UPI.

PR #2378, opened 2021-02-01
Bug 1920027: templates: split crio dropins into separate files

yuqi-zhang:
Somewhere in the template rendering stack the dropin merge logic
is broken, causing only the last dropin defined in a file to get
rendered. Split the dropin templates into separate files for now
to fix CI failures, and revisit later the template rendering logic
to see where this breaks.

Fixes https://github.com/openshift/machine-config-operator/commit/0c7541ec30748c774d1c9e0f27afde2ae6c1e0fb#diff-934bed5b78da27a91177feabe213288a794cc4d1851be77d92c3ff5426f1d88c and https://github.com/openshift/machine-config-operator/commit/fe7fea755afa7f6f1f4706cb4b924a68a0a6a59d#diff-934bed5b78da27a91177feabe213288a794cc4d1851be77d92c3ff5426f1d88c

PR #2652, opened 2021-07-01
Bug 1978041: templates: split unit dropins into separate files

yuqi-zhang:
The template rendering has broken merge logic, which causes us
to lose some templates. This was originally fixed for 4.7
(see https://github.com/openshift/machine-config-operator/pull/2378/)
but previous cherry-pick failures + a recent backport of another
bug broke this for 4.6.

Split the crio/kubelet dropins such that we don't break the proxy.

### DIFF
--- a/templates/common/_base/units/crio.service-proxy.yaml
+++ b/templates/common/_base/units/crio.service-proxy.yaml
@@ -1,0 +1,17 @@
+name: crio.service
+dropins:
+  - name: 10-mco-default-env.conf
+    contents: |
+      [Unit]
+      {{if .Proxy -}}
+      [Service]
+      {{if .Proxy.HTTPProxy -}}
+      Environment=HTTP_PROXY={{.Proxy.HTTPProxy}}
+      {{end -}}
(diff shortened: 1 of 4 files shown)


yuqi-zhang:
Essentially a manual cherry pick for https://github.com/openshift/machine-config-operator/pull/2378#issuecomment-785991657

Fixes https://github.com/openshift/machine-config-operator/pull/2632

wking (approved):
/lgtm

sinnykumari:
wondering of we also need https://github.com/openshift/machine-config-operator/pull/2436

cheesesashimi:
I'm wondering if we need https://github.com/openshift/machine-config-operator/pull/2565 as well.

sinnykumari:
> I'm wondering if we need #2565 as well.

I think this would only impact if someone has set noProxy config in their proxy setting with wildcard like `*`

sinnykumari:
Created PR https://github.com/openshift/machine-config-operator/pull/2653 on top of https://github.com/openshift/machine-config-operator/pull/2436 to see if that helps.

PR #2653, opened 2021-07-01
templates: split unit dropins into separate files

sinnykumari:
The template rendering has broken merge logic, which causes us
to lose some templates. This was originally fixed for 4.7
(see #2378)
but previous cherry-pick failures + a recent backport of another
bug broke this for 4.6.

Split the crio/kubelet dropins such that we don't break the proxy.


sinnykumari:
Cherrypicked https://github.com/openshift/machine-config-operator/pull/2436 as well for easier testing

sinnykumari:
Not needed since original PR https://github.com/openshift/machine-config-operator/pull/2652 seems to be working with proxy config.

PR #2654, opened 2021-07-01
GRPA-3788: Machine-config-daemon: support RHEL 8 OS versions

jkyros:
This PR: 

- Gives the MCD the ability to tell if a node is a RHEL8 node 
- Adds a simple unit test to make sure the version is being recognized properly 
- Modifies the e2e tests such that RHEL7/8 unsupported features no longer result in e2e test failures

JIRA ticket link: https://issues.redhat.com/browse/GRPA-3788

The places where we were explicitly checking for RHEL7 were places where RHEL7 was "broken" compared to everything else, and that was due to older versions of software (systemd, logger). RHEL8 is newer (and closer to RHCOS than 7 was), it doesn't suffer from any of the current RHEL7 issues, and as a result we don't yet have a need to explicitly check for it anywhere right now.

There are still places in the code where we check to see if the OS is a CoreOS variant so we can determine whether the OS supports a feature or not, but those checks will still mark RHEL8 as "non-CoreOS", just like they did RHEL7. so our existing decision-making there does not need to change for RHEL8. 

The e2e tests were failing against RHEL7/8 nodes because they were checking for the results of functionality that was explicitly not supported by non-CoreOS nodes (KernelArguments, Kernel, Extensions) and  I have modified the tests to take those unsupported items into account so the RHEL nodes can now be e2e tested.

### DIFF
--- a/pkg/daemon/osrelease.go
+++ b/pkg/daemon/osrelease.go
@@ -44,6 +44,15 @@ func (os OperatingSystem) IsLikeTraditionalRHEL7() bool {
 	return os.VersionID == "7"
 }
 
+// IsLikeTraditionalRHEL8 is true if the OS is traditional RHEL8 or CentOS8:
+func (os OperatingSystem) IsLikeTraditionalRHEL8() bool {
+
+	if len(os.VersionID) > 2 {
+		return strings.HasPrefix(os.VersionID, "8.")
+	}
+	return os.VersionID == "8"
+}
+
 // ToPrometheusLabel returns a value we historically fed to Prometheus
 func (os OperatingSystem) ToPrometheusLabel() string {
 	// We historically upper cased this
--- a/pkg/daemon/osrelease_test.go
+++ b/pkg/daemon/osrelease_test.go
@@ -16,3 +16,15 @@ func TestIsLikeTraditionalRHEL7(t *testing.T) {
 	testOS.VersionID = "6.8"
 	assert.False(t, testOS.IsLikeTraditionalRHEL7())
 }
+
+func TestIsLikeTraditionalRHEL8(t *testing.T) {
+	var testOS OperatingSystem
+	testOS.VersionID = "8"
+	assert.True(t, testOS.IsLikeTraditionalRHEL8())
+	testOS.VersionID = "8.5"
+	assert.True(t, testOS.IsLikeTraditionalRHEL8())
+	testOS.VersionID = "7"
+	assert.False(t, testOS.IsLikeTraditionalRHEL8())
+	testOS.VersionID = "6.8"
+	assert.False(t, testOS.IsLikeTraditionalRHEL8())
+}
--- a/test/e2e/mcd_test.go
+++ b/test/e2e/mcd_test.go
@@ -119,14 +119,20 @@ func TestKernelArguments(t *testing.T) {
 	infraNode := helpers.GetSingleNodeByRole(t, cs, "infra")
 	assert.Equal(t, infraNode.Annotations[constants.CurrentMachineConfigAnnotationKey], renderedConfig)
 	assert.Equal(t, infraNode.Annotations[constants.MachineConfigDaemonStateAnnotationKey], constants.MachineConfigDaemonStateDone)
-	kargs := helpers.ExecCmdOnNode(t, cs, infraNode, "cat", "/rootfs/proc/cmdline")
-	expectedKernelArgs := []string{"nosmt", "foo=bar", "foo=baz", "baz=test", "bar=hello world"}
-	for _, v := range expectedKernelArgs {
-		if !strings.Contains(kargs, v) {
-			t.Fatalf("Missing %q in kargs: %q", v, kargs)
+
+	// Non-CoreOS nodes don't support kernel/args today, we need to make sure they survive this test
+	if helpers.IsCoreOSNode(&infraNode) {
+		kargs := helpers.ExecCmdOnNode(t, cs, infraNode, "cat", "/rootfs/proc/cmdline")
+		expectedKernelArgs := []string{"nosmt", "foo=bar", "foo=baz", "baz=test", "bar=hello world"}
+		for _, v := range expectedKernelArgs {
+			if !strings.Contains(kargs, v) {
+				t.Fatalf("Missing %q in kargs: %q", v, kargs)
+			}
 		}
+		t.Logf("Node %s has expected kargs", infraNode.Name)
+	} else {
+		t.Logf("Non-CoreOS node %s does not support kernel and argument changes (expected) ", infraNode.Name)
 	}
-	t.Logf("Node %s has expected kargs", infraNode.Name)
 
 	// cleanup - delete karg mc and rollback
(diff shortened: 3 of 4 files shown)


test/e2e/mcd_test.go
```diff
@@ -119,14 +119,20 @@ func TestKernelArguments(t *testing.T) {
...
+	if helpers.IsCoreOSNode(&infraNode) {
+		kargs := helpers.ExecCmdOnNode(t, cs, infraNode, "cat", "/rootfs/proc/cmdline")
+		expectedKernelArgs := []string{"nosmt", "foo=bar", "foo=baz", "baz=test", "bar=hello world"}
+		for _, v := range expectedKernelArgs {
+			if !strings.Contains(kargs, v) {
+				t.Fatalf("Missing %q in kargs: %q", v, kargs)
+			}
 		}
+		t.Logf("Node %s has expected kargs", infraNode.Name)
+	} else {
```
kikisdeliveryservice, line 133: this test change doesn't quite makes sense bc we are explicitly testing the above Fatalf error for a specific state so we don't need to be checking if it's a coreosnode or have the else as this test should _only_ pass if the fatal is not hit.

kikisdeliveryservice (changes requested):
can we drop the e2e changes, it doesn't quite make sense to me. if you want to test rhel, i think you should look into modifying the e2e-rhel job

pkg/daemon/osrelease.go
```diff
@@ -44,6 +44,15 @@ func (os OperatingSystem) IsLikeTraditionalRHEL7() bool {
 	return os.VersionID == "7"
 }
 
+// IsLikeTraditionalRHEL8 is true if the OS is traditional RHEL8 or CentOS8:
+func (os OperatingSystem) IsLikeTraditionalRHEL8() bool {
```
kikisdeliveryservice, line 48: Also this new func doesn't seem to be called anywhere not in a test? bc to support rhel8 i'd expect an update to update.go/daemon.go as well?

test/e2e/mcd_test.go
jkyros: Right, but while the test failure should be fatal for a CoreOS node, for a RHEL node it shouldn't. If you run our e2e tests against a cluster with RHEL nodes in it (like with make test-e2e), the e2e tests fail even though the MCO and the RHEL nodes are behaving properly -- RHEL nodes can receive kernel config during normal operation, but they're supposed to do nothing with it by design.
	
I was looking at this as "test failure" == "something is broken", and nothing is broken when the changes do not take effect on RHEL nodes.

 Is there some sort of general governing principle as to what goes in our repo's e2e tests and what needs to go in release/step-registry?

kikisdeliveryservice: why not just skip the test entirely if it can't be run on RHELX nodes? I'm not fully understanding the utility of having it roll out things that shouldn't work instead of just not running the test at all.

pkg/daemon/osrelease.go
```diff
@@ -44,6 +44,15 @@ func (os OperatingSystem) IsLikeTraditionalRHEL7() bool {
 	return os.VersionID == "7"
 }
 
+// IsLikeTraditionalRHEL8 is true if the OS is traditional RHEL8 or CentOS8:
```
kikisdeliveryservice, line 47: Is this func only used in the below unit test?

test/e2e/mcd_test.go
kikisdeliveryservice: if it's in a mixed node cluster (i have no idea how the suite running on rhel nodes is set up or if this exists), it selects a node to be a pool to roll a change out to, why not just check whether or not that node is RHELX there instead of rolling out the machineconfig that isn't supported?

pkg/daemon/osrelease_test.go
```diff
@@ -16,3 +16,15 @@ func TestIsLikeTraditionalRHEL7(t *testing.T) {
 	testOS.VersionID = "6.8"
 	assert.False(t, testOS.IsLikeTraditionalRHEL7())
 }
+
+func TestIsLikeTraditionalRHEL8(t *testing.T) {
```
kikisdeliveryservice, line 20: since the func is only used here, i dont think we need this?

test/e2e/mcd_test.go
jkyros: I can do it there, too, but I wasn't looking at the rest of the test being useless. I was looking at it as testing what happens in a mixed node cluster to make sure that you can: 

1.  "push out a config with unsupported features"
2. "do nothing with the parts you can't understand, like you're supposed to" 
3.  "make sure you can roll back from it" 

because you could receive unsupported config in a mixed-node cluster, and  you'd want to be sure it behaved properly? 

I actually tested it by pulling all of my CoreOS nodes out of my cluster and leaving only the RHEL ones so I wasn't at the mercy of the random chooser.

pkg/daemon/osrelease.go
jkyros: It's only used there because we don't know of any ways RHEL8 is broken yet (all of the RHEL7 decisions were because something was "wrong" with RHEL7 relative to CoreOS, , but at some point if we have to do something RHEL8-specific (bug, feature, whatever) we'll have to add it. 

I can take it out and hold it if we don't do "speculative code" that we don't 100% have a use for right now.

test/e2e/mcd_test.go
kikisdeliveryservice: > "do nothing with the parts you can't understand, like you're supposed to"

the problem is the there is no check in the else seeing if the node/mcd did what it was supposed to do and whether or not the node hit an error state, it's just logging success.

pkg/daemon/osrelease.go
kikisdeliveryservice: Yes, please remove. :+1:

test/e2e/mcd_test.go
jkyros: True, I only went in half way (mostly because degradation under those conditions was more the thing I was worried about, not necessarily the doing nothing).

If I check the before and after to verify that it did nothing across kernel/kernelargs/extensions, is that acceptable?

kikisdeliveryservice: sure, adding a check to verifying what you consider to be success (for the non-coreos node case) in the else would wfm

pkg/daemon/osrelease.go
sinnykumari: agree with Kirsten, it doesn't make sense to add a function in advance if there is no usecase today. It can be added as part of fix if required in future.

test/e2e/mcd_test.go
sinnykumari: I think modifying current e2e gcp-op test doesn't makes sense today. Tests are written and added as we go on adding additional coverage. I don't know if it is possible today to create  a OCP cluster in ci with mixed nodes(RHCOS + RHEL) in a pool. Before improving these tests it would be worthwhile checking it.

jkyros:
Closing this  -- thank you everyone for your time and review -- it looks like there is nothing we need to implement here at this time :)

sinnykumari:
Thank you John for putting all the effort. In future when there is a need, we can definitely reuse some part of this PR :)

PR #2658, opened 2021-07-02
Bug 1986453: Check for API server and node versions skew

QiWang19:
ref: https://issues.redhat.com/browse/OCPNODE-595
replace https://github.com/openshift/machine-config-operator/pull/2552

**- What I did**

Check for API server and node versions skew.
Update with the message the Kube API version is skew too far, but do not force
`Upgradeable=False` according to enhancement https://github.com/openshift/enhancements/pull/762/files

**- How to verify it**

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
@@ -16,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	v1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 )
 
@@ -243,6 +246,13 @@ func (optr *Operator) syncDegradedStatus(ierr syncError) (err error) {
 	return optr.updateStatus(co, coStatus)
 }
 
+const (
+	skewUnchecked   = "KubeletSkewUnchecked"
+	skewSupported   = "KubeletSkewSupported"
+	skewUnsupported = "KubeletSkewUnsupported"
+	skewPresent     = "KubeletSkewPresent"
+)
+
 // syncUpgradeableStatus applies the new condition to the mco's ClusterOperator object.
 func (optr *Operator) syncUpgradeableStatus() error {
 	co, err := optr.fetchClusterOperator()
@@ -289,7 +299,166 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		skewStatus, status, err := optr.isKubeletSkewSupported(pools)
+		if err != nil {
+			glog.Errorf("Error checking version skew: %v, kubelet skew status: %v, status reason: %v, status message: %v", err, skewStatus, status.Reason, status.Message)
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			return optr.updateStatus(co, coStatus)
+		}
+		switch skewStatus {
+		case skewUnchecked:
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			return optr.updateStatus(co, coStatus)
+		case skewUnsupported:
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			mcoObjectRef := &corev1.ObjectReference{
+				Kind:      co.Kind,
+				Name:      co.Name,
+				Namespace: co.Namespace,
+				UID:       co.GetUID(),
+			}
+			glog.Infof("kubelet skew status: %v, status reason: %v", skewStatus, status.Reason)
+			optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, coStatus.Reason, coStatus.Message)
+			return optr.updateStatus(co, coStatus)
+		case skewPresent:
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			glog.Infof("kubelet skew status: %v, status reason: %v", skewStatus, status.Reason)
+			return optr.updateStatus(co, coStatus)
+		}
+	}
 	return optr.updateStatus(co, coStatus)
+}
+
+// isKubeletSkewSupported checks the version skew of kube-apiserver and node kubelet version.
+// Returns the skew status. version skew > 2 is not supported.
+func (optr *Operator) isKubeletSkewSupported(pools []*v1.MachineConfigPool) (skewStatus string, coStatus configv1.ClusterOperatorStatusCondition, err error) {
+	coStatus = configv1.ClusterOperatorStatusCondition{}
+	kubeAPIServerStatus, err := optr.configClient.ConfigV1().ClusterOperators().Get(context.TODO(), "kube-apiserver", metav1.GetOptions{})
+	if err != nil {
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err)
+		return skewUnchecked, coStatus, err
+	}
+	// looks like
+	//   - name: kube-apiserver
+	//     version: 1.21.0-rc.0
+	kubeAPIServerVersion := ""
+	for _, version := range kubeAPIServerStatus.Status.Versions {
+		if version.Name != "kube-apiserver" {
+			continue
+		}
+		kubeAPIServerVersion = version.Version
+		break
+	}
+	if kubeAPIServerVersion == "" {
+		err = fmt.Errorf("kube-apiserver does not yet have a version")
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err.Error())
+		return skewUnchecked, coStatus, err
+	}
+	kubeAPIServerMinorVersion, err := getMinorKubeletVersion(kubeAPIServerVersion)
+	if err != nil {
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err)
+		return skewUnchecked, coStatus, err
+	}
+	var (
+		lastError      error
+		kubeletVersion string
+	)
+	nodes, err := optr.GetAllManagedNodes(pools)
+	if err != nil {
+		err = fmt.Errorf("getting all managed nodes failed: %v", err)
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when getting all the managed nodes: %v", err.Error())
+	}
+	for _, node := range nodes {
+		// looks like kubeletVersion: v1.21.0-rc.0+6143dea
+		kubeletVersion = node.Status.NodeInfo.KubeletVersion
+		if kubeletVersion == "" {
+			continue
+		}
+		nodeMinorVersion, err := getMinorKubeletVersion(kubeletVersion)
+		if err != nil {
+			lastError = err
+			continue
+		}
+		if nodeMinorVersion+2 < kubeAPIServerMinorVersion {
+			coStatus.Reason = skewUnsupported
+			coStatus.Message = fmt.Sprintf("One or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet version of at least %v.", getMinimalSkewSupportNodeVersion(kubeAPIServerVersion))
+			return skewUnsupported, coStatus, nil
+		}
+		if nodeMinorVersion+2 == kubeAPIServerMinorVersion {
+			coStatus.Reason = skewPresent
+			coStatus.Message = fmt.Sprintf("Current kubelet version %v will not be supported by newer kube-apiserver. Please upgrade the kubelet first if plan to upgrade the kube-apiserver", kubeletVersion)
+			return skewPresent, coStatus, nil
+		}
+	}
+	if kubeletVersion == "" {
+		err = fmt.Errorf("kubelet does not yet have a version")
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err.Error())
+		return skewUnchecked, coStatus, err
+	}
+	if lastError != nil {
+		coStatus.Reason = skewUnchecked
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err)
+		return skewUnchecked, coStatus, lastError
+	}
+	return skewSupported, coStatus, nil
+}
+
+// GetAllManagedNodes returns the nodes managed by MCO
+func (optr *Operator) GetAllManagedNodes(pools []*v1.MachineConfigPool) ([]*corev1.Node, error) {
+	nodes := []*corev1.Node{}
+	for _, pool := range pools {
+		selector, err := metav1.LabelSelectorAsSelector(pool.Spec.NodeSelector)
+		if err != nil {
+			return nil, fmt.Errorf("label selector for pool %v failed %v", pool.Name, err)
+		}
+		poolNodes, err := optr.nodeLister.List(selector)
+		if err != nil {
+			return nil, fmt.Errorf("could not list nodes for pool %v with error %v", pool.Name, err)
+		}
+		nodes = append(nodes, poolNodes...)
+	}
+	return nodes, nil
+}
+
+// getMinorKubeletVersion parses the minor version number of kubelet
+func getMinorKubeletVersion(version string) (int, error) {
+	tokens := strings.Split(version, ".")
+	if len(tokens) < 2 {
+		return 0, fmt.Errorf("incorrect version syntax: %q", version)
+	}
+	minorVersion, err := strconv.ParseInt(tokens[1], 10, 32)
+	if err != nil {
+		return 0, err
+	}
+	return int(minorVersion), nil
+}
+
+// getMinimalSkewSupportNodeVersion returns the minimal supported node kubelet version.
+func getMinimalSkewSupportNodeVersion(version string) string {
+	// drop the pre-release and commit hash
+	if strings.Contains(version, "-") {
+		version = version[:strings.Index(version, "-")]
+	}
+	if strings.Contains(version, "+") {
+		version = version[:strings.Index(version, "+")]
+	}
+	tokens := strings.Split(version, ".")
+	if minorVersion, err := strconv.ParseInt(tokens[1], 10, 32); err == nil {
+		tokens[1] = strconv.Itoa(int(minorVersion) - 2)
+		return strings.Join(tokens, ".")
+	}
+	return version
 }
 
 func (optr *Operator) fetchClusterOperator() (*configv1.ClusterOperator, error) {
--- a/pkg/operator/status_test.go
+++ b/pkg/operator/status_test.go
@@ -7,6 +7,10 @@ import (
 	"reflect"
 	"testing"
 
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
+	clientgotesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -20,6 +24,7 @@ import (
 	cov1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
+	"github.com/openshift/machine-config-operator/test/helpers"
 )
 
 func TestIsMachineConfigPoolConfigurationValid(t *testing.T) {
@@ -152,12 +157,23 @@ func TestIsMachineConfigPoolConfigurationValid(t *testing.T) {
 	}
 }
 
-type mockMCPLister struct{}
+type mockMCPLister struct {
+	pools []*mcfgv1.MachineConfigPool
+}
 
 func (mcpl *mockMCPLister) List(selector labels.Selector) (ret []*mcfgv1.MachineConfigPool, err error) {
-	return nil, nil
+	return mcpl.pools, nil
 }
 func (mcpl *mockMCPLister) Get(name string) (ret *mcfgv1.MachineConfigPool, err error) {
+	if mcpl.pools == nil {
+		return nil, nil
+	}
+	for _, pool := range mcpl.pools {
+		if pool.Name == name {
+			return pool, nil
+		}
+
+	}
 	return nil, nil
 }
 
@@ -550,7 +566,24 @@ func TestOperatorSyncStatus(t *testing.T) {
 			eventRecorder: &record.FakeRecorder{},
 		}
 		optr.vStore = newVersionStore()
-		optr.mcpLister = &mockMCPLister{}
+		optr.mcpLister = &mockMCPLister{
+			pools: []*mcfgv1.MachineConfigPool{
+				helpers.NewMachineConfigPool("master", nil, helpers.MasterSelector, "v0"),
+				helpers.NewMachineConfigPool("workers", nil, helpers.WorkerSelector, "v0"),
+			},
+		}
+
+		nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+		optr.nodeLister = corelisterv1.NewNodeLister(nodeIndexer)
+		nodeIndexer.Add(&corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{Name: "first-node", Labels: map[string]string{"node-role/worker": ""}},
+			Status: corev1.NodeStatus{
+				NodeInfo: corev1.NodeSystemInfo{
+					KubeletVersion: "v1.21",
+				},
+			},
+		})
+
 		coName := fmt.Sprintf("test-%s", uuid.NewUUID())
 		co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: coName}}
 		cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse})
@@ -559,6 +592,14 @@ func TestOperatorSyncStatus(t *testing.T) {
 		cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorUpgradeable, Status: configv1.ConditionUnknown})
 		co.Status.Versions = append(co.Status.Versions, configv1.OperandVersion{Name: "operator", Version: "test-version"})
 		optr.name = coName
+		kasOperator := &configv1.ClusterOperator{
+			ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
+			Status: configv1.ClusterOperatorStatus{
+				Versions: []configv1.OperandVersion{
+					{Name: "kube-apiserver", Version: "1.21"},
+				},
+			},
+		}
 
 		for j, sync := range testCase.syncs {
 			optr.inClusterBringup = sync.inClusterBringUp
@@ -567,7 +608,7 @@ func TestOperatorSyncStatus(t *testing.T) {
 			} else {
 				optr.vStore.Set("operator", "test-version")
 			}
-			optr.configClient = fakeconfigclientset.NewSimpleClientset(co)
+			optr.configClient = fakeconfigclientset.NewSimpleClientset(co, kasOperator)
 			err := optr.syncAll(sync.syncFuncs)
 			if sync.expectOperatorFail {
 				assert.NotNil(t, err, "test case %d, sync call %d, expected an error", idx, j)
@@ -597,12 +638,35 @@ func TestInClusterBringUpStayOnErr(t *testing.T) {
 	}
 	optr.vStore = newVersionStore()
 	optr.vStore.Set("operator", "test-version")
-	optr.mcpLister = &mockMCPLister{}
+	optr.mcpLister = &mockMCPLister{
+		pools: []*mcfgv1.MachineConfigPool{
+			helpers.NewMachineConfigPool("master", nil, helpers.MasterSelector, "v0"),
+			helpers.NewMachineConfigPool("workers", nil, helpers.WorkerSelector, "v0"),
+		},
+	}
+	nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	optr.nodeLister = corelisterv1.NewNodeLister(nodeIndexer)
+	nodeIndexer.Add(&corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "first-node", Labels: map[string]string{"node-role/worker": ""}},
+		Status: corev1.NodeStatus{
+			NodeInfo: corev1.NodeSystemInfo{
+				KubeletVersion: "v1.21",
+			},
+		},
+	})
 	co := &configv1.ClusterOperator{}
+	kasOperator := &configv1.ClusterOperator{
+		ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
+		Status: configv1.ClusterOperatorStatus{
+			Versions: []configv1.OperandVersion{
+				{Name: "kube-apiserver", Version: "1.21"},
+			},
+		},
+	}
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse})
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse})
 	cov1helpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorDegraded, Status: configv1.ConditionFalse})
-	optr.configClient = fakeconfigclientset.NewSimpleClientset(co)
+	optr.configClient = fakeconfigclientset.NewSimpleClientset(co, kasOperator)
 	optr.inClusterBringup = true
 
 	fn1 := func(config *renderConfig) error {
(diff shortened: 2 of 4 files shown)


pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		} else if err != nil {
+			coStatus.Reason = coStatus.Reason + "::" + "KubeletSkewTooFar"
+			coStatus.Message = coStatus.Message + "\n" + err.Error()
+		}
+	}
+
 	return optr.updateStatus(co, coStatus)
 }
 
+func (optr *Operator) isKubeletSkewTooFar() (bool, error) {
```
kikisdeliveryservice, line 310:
```suggestion
func (optr *Operator) isKubeletSkewSupported() (bool, error) {
```

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if ran, err := optr.isKubeletSkewTooFar(); ran && err != nil {
+		if coStatus.Status != configv1.ConditionFalse {
+			coStatus.Reason = "KubeletSkewTooFar"
```
kikisdeliveryservice, line 299:
```suggestion
			coStatus.Reason = "KubeletSkewUnsupported"
```

pkg/operator/status.go
kikisdeliveryservice: also please add func description above declaration

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+	}
+	if len(kubeAPIServerVersion) == 0 {
+		return false, fmt.Errorf("kube-apiserver does not yet have a version")
+	}
+	kubeAPIServerMinorVersion, err := getMinorVersion(kubeAPIServerVersion)
+	if err != nil {
+		return true, err
+	}
+
+	skewedNodeErrors := []error{}
```
kikisdeliveryservice, line 342: do we really want to have a list of errors and print a list at the CO level?

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if ran, err := optr.isKubeletSkewTooFar(); ran && err != nil {
```
kikisdeliveryservice, line 297: why ran when you aren't checking if it ran but if the skew is supported?

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		}
+		nodeMinorVersion, err := getMinorVersion(kubeletVersion)
+		if err != nil {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v has malformed kubelet version %q: %v", node.Name, kubeletVersion, err))
+		}
+		if nodeMinorVersion+2 <= kubeAPIServerMinorVersion {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v must be updated or removed before the cluster can upgrade, current version %v", node.Name, kubeletVersion))
+		}
+	}
+	return true, errors.NewAggregate(skewedNodeErrors)
```
kikisdeliveryservice, line 357: I'm kind of confused about what the bool return values in this func signify? is it t/f skew supported? or t/r function ran?  I think it should be the former, but it's not clear reading this.

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		kubeletVersion := node.Status.NodeInfo.KubeletVersion
+		if len(kubeletVersion) == 0 {
+			continue
+		}
+		nodeMinorVersion, err := getMinorVersion(kubeletVersion)
+		if err != nil {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v has malformed kubelet version %q: %v", node.Name, kubeletVersion, err))
+		}
+		if nodeMinorVersion+2 <= kubeAPIServerMinorVersion {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v must be updated or removed before the cluster can upgrade, current version %v", node.Name, kubeletVersion))
```
kikisdeliveryservice, line 354: I think this message should be more explicit, ie tell them how many versions they are off/what the skew is

kikisdeliveryservice (commented):
some initial questions/comments

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if ran, err := optr.isKubeletSkewTooFar(); ran && err != nil {
+		if coStatus.Status != configv1.ConditionFalse {
+			coStatus.Reason = "KubeletSkewTooFar"
+			coStatus.Message = err.Error()
```
kikisdeliveryservice, line 300: Feel like this really should be a const Message

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		} else if err != nil {
+			coStatus.Reason = coStatus.Reason + "::" + "KubeletSkewTooFar"
```
kikisdeliveryservice, line 302: what are we doing Reason::KubeletSkewTooFar?instead of just KubeletSkewTooFar?

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			coStatus.Message = coStatus.Message + "\n" + err.Error()
```
kikisdeliveryservice, line 303: similar Q: coStats.Message+error?

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
```
kikisdeliveryservice, line 296: I'm not convinced we should be reporting on a node level. In a large cluster this could  result  in a big list of nodes when 1) the MCO doesn't think via individual nodes but rather pools and 2) a long list is unmangeable as a ui

I think this should probably be reported at a pool level.

kikisdeliveryservice (commented):
a few more comments

/hold

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+	skewedNodeErrors := []error{}
+	for _, node := range nodes {
+		// looks like kubeletVersion: v1.21.0-rc.0+6143dea
+		kubeletVersion := node.Status.NodeInfo.KubeletVersion
+		if len(kubeletVersion) == 0 {
+			continue
+		}
+		nodeMinorVersion, err := getMinorVersion(kubeletVersion)
+		if err != nil {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v has malformed kubelet version %q: %v", node.Name, kubeletVersion, err))
```
kikisdeliveryservice, line 351: what does this mean to a user? what are they supposed to do to fix?

pkg/operator/status.go
kikisdeliveryservice: a message like: Worker pool contains nodes with an unsupported kubelet skew. Please see `oc get nodes` for details and ensure all nodes have a kubelet Version of at least $version.

Would be more concise but also more descriptive.

QiWang19: @kikisdeliveryservice  What is reporting at the pool level? The version skew should be checked for each node, follow the same implementation as https://github.com/openshift/machine-config-operator/blob/db56bd2856b4892f3b72e6056904a29676d362d4/pkg/operator/status.go#L289, the message is reported at pool level.  Could you elaborate?

pkg/operator/status.go
QiWang19: This only provides the user with more information that there's an error when checking the version. Might be helpful debugging if they run into other issues but no way to fix it at this point. Should we `glog.Error(err)` this error or silently ignore it?

pkg/operator/status.go
kikisdeliveryservice: I have a suggestion above: https://github.com/openshift/machine-config-operator/pull/2658/files#r663253475

pkg/operator/status.go
kikisdeliveryservice: Would this error manifest as another error that already exists? IE how does a user get into this situation and would it already be reflected in the state of the pool (would the pool be degraded, etc..)

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if ran, err := optr.isKubeletSkewTooFar(); ran && err != nil {
+		if coStatus.Status != configv1.ConditionFalse {
+			coStatus.Reason = "KubeletSkewTooFar"
+			coStatus.Message = err.Error()
+		} else if err != nil {
```
kikisdeliveryservice, line 301: aren't we already guaranteed err!=nil from `ran && err != nil` above?

pkg/operator/status.go
```diff
@@ -289,9 +293,82 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v has malformed kubelet version %q: %v", node.Name, kubeletVersion, err))
+		}
+		if nodeMinorVersion+2 <= kubeAPIServerMinorVersion {
+			skewedNodeErrors = append(skewedNodeErrors, fmt.Errorf("node/%v must be updated or removed before the cluster can upgrade, current version %v", node.Name, kubeletVersion))
+		}
+	}
+	return true, errors.NewAggregate(skewedNodeErrors)
+}
+
+func getMinorVersion(version string) (int, error) {
```
kikisdeliveryservice, line 360:
```suggestion
func getMinorKubeletVersion(version string) (int, error) {
```

kikisdeliveryservice:
@QiWang19 I know that this PR is based on someone else's previous PR, but I think we can refine it. What most of my comments are asking for is: using that work as a baseline, how can we make this into an understandable & actionable status for a user. Also keep in mind a pool can be large, so the info needs to be easy to consume if we have, say, 50 nodes that have unsupported skew.  

Users can get into this situation bc a pool is paused and so it did not upgrade to match the apiserver and is still at an older version. They will likely have to let those pools upgrade (to at a minimum a supported skew) before they can initiate another clusterwide upgrade (which would cause the kubeapiserver to get even further away).  So we need to think about telling them the state in a meaningful way, but also give them some hint about what they need to do to remedy it.

Happy to discuss further if you have any questions. =)

pkg/operator/status.go
QiWang19: > Would this error manifest as another error that already exists? IE how does a user get into this situation and would it already be reflected in the state of the pool (would the pool be degraded, etc..),would it fail the kubelet health check, etc..

I agree with the above situation. I tend to only report the message and reason if the version has been successfully checked and compared to verify there's a version skew.

QiWang19:
@kikisdeliveryservice Thanks for the explanation. I have cleaned up some reviews. PTAL.

QiWang19:
@kikisdeliveryservice Thanks for the explanation. I have cleaned up some reviews. Could you PTAL?

pkg/operator/status.go
```diff
@@ -289,9 +291,79 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if supported, v := optr.isKubeletSkewSupported(); !supported {
+		coStatus.Reason = "KubeletSkewUnsupported"
+		coStatus.Message = fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", v)
```
kikisdeliveryservice, line 297: `upgrade all nodes so that they have a kubelet Version of at least %v.", v`

If v is the apiserver version, the nodes don't have to be updated to that version but to that version-2 at a minimum no?

pkg/operator/status.go
```diff
@@ -289,9 +291,79 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if we didn't run
+	if supported, v := optr.isKubeletSkewSupported(); !supported {
+		coStatus.Reason = "KubeletSkewUnsupported"
```
kikisdeliveryservice, line 296: As is this would be overwriting a degraded and updating status, I think we probably only want to do this if we are available but not updating/degraded? Doesn't seem like this should preempt those more impt messages? wdyt

kikisdeliveryservice (commented):
a few more comments

pkg/operator/status.go
QiWang19: > `upgrade all nodes so that they have a kubelet Version of at least %v.", v`
> 
> If v is the apiserver version, the nodes don't have to be updated to that version but to that version-2 at a minimum no?

The version skew does not support version - 2 version number. It should be the version with n-1 (n is the minor version of the version). If we want to report the version with a specific version number, will need another parsing function to do the calculation. https://github.com/openshift/machine-config-operator/pull/2658/files#diff-8a6a064b4db19eb3b29e5280447ac3ba9496e847391489d25c1282767c3354c5R371
Do you think it worth adding this func?

pkg/operator/status.go
```diff
@@ -289,9 +291,99 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating and degraded
+	if !updating && !degraded {
+		if supported, v := optr.isKubeletSkewSupported(); !supported {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
+		}
```
kikisdeliveryservice, line 299: can we also add an event here?

see an example here: https://github.com/openshift/machine-config-operator/blob/4f555d05550bca9a67f5354afa2721cea5e36b02/pkg/operator/status.go#L44

pkg/operator/status.go
```diff
@@ -289,9 +291,99 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		kubeletVersion := node.Status.NodeInfo.KubeletVersion
+		if kubeletVersion == "" {
+			continue
+		}
+		nodeMinorVersion, err := getMinorKubeletVersion(kubeletVersion)
+		if err != nil {
+			glog.Error(err)
+			continue
+		}
+		if nodeMinorVersion+2 <= kubeAPIServerMinorVersion {
```
kikisdeliveryservice, line 348: note: we're checking on this

pkg/operator/status.go
kikisdeliveryservice: I think your update covers it by printing minimalsupported version. :+1:

kikisdeliveryservice:
Just noting that when this is done, we need to get the commits updated so Qi is a co-author.

pkg/operator/status.go
```diff
@@ -289,9 +291,99 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating and degraded
+	if !updating && !degraded {
+		if supported, v := optr.isKubeletSkewSupported(); !supported {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
```
kikisdeliveryservice, line 298: There's been a bunch of great changes on this PR, so I'm giving this message a little thought.

pkg/operator/status.go
QiWang19: @kikisdeliveryservice  I checked this with @rphillips . We can keep it as is. Considering the 2nd example here https://docs.google.com/document/d/1yXzWPBWmKLuKBkvOKMk0FusyBpUqjyT3hGXYL8tUafU/edit#. we get into this situation when we do upgrade kube-apiserver is at 1.21 and 1.20 , then the kubelet can only support 1.20 and 1.19.

pkg/operator/status.go
QiWang19: @kikisdeliveryservice I added the event. PTAL.

pkg/operator/status.go
kikisdeliveryservice: Wouldn't <= mean that we only support 1? if apiserver is at 1.20 and node is at 1.18 and I add 2 it's = apiserver and not supported. Is this correct?

deads2k: Up to n-2 is support.

1. 1.21 kas supports 1.21, 1.20, 1.19 kubelet
2. 1.20 kas supports 1.20, 1.19, 1.18 kubelet

The allowable skew is the intersection, in this case 1.20 and 1.19.

QiWang19: > Up to n-2 is support.
> 
>     1. 1.21 kas supports 1.21, 1.20, 1.19 kubelet
> 
>     2. 1.20 kas supports 1.20, 1.19, 1.18 kubelet
> 
> 
> The allowable skew is the intersection, in this case 1.20 and 1.19.

@deads2k So the skew calculation should take into account the upgrade, like both 1.21 and 1.20 in the above example?

kikisdeliveryservice: right but my question is that <= correct?

and isn't apiserver has various versions only applicable when we're mid-update? post update, they would be the same.

kikisdeliveryservice:
Would like an answer on my question about:

nodeMinorVersion+2 <= kubeAPIServerMinorVersion

means that if i have a 1.19 and apiserver is 1.21 we are saying this isn't supported and they must upgrade even tho the skew is a supported 2 versions.  shouldn't this be:

nodeMinorVersion+2 < kubeAPIServerMinorVersion

rphillips:
@kikisdeliveryservice According to this [doc](https://docs.google.com/presentation/d/1PJvN6-hLm9Vr8KuecMmiyk9J5vszGDcTAsppKLUet70/edit#slide=id.gbab5de315b_0_112) the version skew will always be the current version and the next version. The n+2 version should be flagged as an error.

kikisdeliveryservice:
>the version skew will always be the current version and the next version. The n+2 version should be flagged as an error.

@rphillips I'm really confused on this bc I thought the whole point was to allow an n+2 version skew, which is also what upstream is?

kikisdeliveryservice:
but also the file that you are linking is allowing n+2

kikisdeliveryservice:
Wait I think my issue is the phrasing bc n+2 is supported, but we're _advising_ you at that point to upgrade if you are not intending to stay there.. I need more coffee.

rphillips:
| Openshift Version | Kubelet Version |
| ------------------------| ----------------------|
| 4.6 | 1.0 |
| 4.7 | 1.1 |
| 4.8 | *1.2* |
| 4.9 | 1.3 |
| 4.10 | *1.4* |

Control Plane upgrades to 4.9.x  & Worker nodes remain on 4.8.x, then the control plane upgrades to 4.10. Kirsten is correct... It shouldn't be <=.

rphillips:
```diff
diff --git a/pkg/operator/status.go b/pkg/operator/status.go
index 62f0ea81..3bf5b06f 100644
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -346,7 +346,7 @@ func (optr *Operator) isKubeletSkewSupported() (skewSupported bool, kubeAPIServe
                        glog.Error(err)
                        continue
                }
-               if nodeMinorVersion+2 <= kubeAPIServerMinorVersion {
+               if nodeMinorVersion+2 < kubeAPIServerMinorVersion {
                        skewSupported = false
                        break
                }
diff --git a/pkg/operator/status_test.go b/pkg/operator/status_test.go
index 83cd8b92..15209418 100644
--- a/pkg/operator/status_test.go
+++ b/pkg/operator/status_test.go
@@ -637,7 +637,7 @@ func TestKubeletSkewSupported(t *testing.T) {
                ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
                Status: configv1.ClusterOperatorStatus{
                        Versions: []configv1.OperandVersion{
-                               {Name: "kube-apiserver", Version: "1.20"},
+                               {Name: "kube-apiserver", Version: "1.21"},
                        },
                },
        }
@@ -705,7 +705,7 @@ func TestKubeletSkewSupported(t *testing.T) {
        if upgradeable.Status != configv1.ConditionTrue {
                t.Fatal(upgradeable)
        }
-       if upgradeable.Message != "machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least 1.19." {
+       if upgradeable.Message != "machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least 1.20." {
                t.Fatal(upgradeable)
        }
        if upgradeable.Reason != "KubeletSkewUnsupported" {
```

QiWang19:
> ```diff
> -       if upgradeable.Message != "machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least 1.19." {
> +       if upgradeable.Message != "machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least 1.20." {
> ```

@rphillips should this at least version be 1.19? The test case kube-apiserver 1.21, node version 1.18. If the policy allows at most 2 versions skew. The minimum version they need to upgrade to is 1.19.

rphillips:
@QiWang19 Agreed... Can you fix that up?

QiWang19:
> @QiWang19 Agreed... Can you fix that up?

Fixed.

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		if supported, v := optr.isKubeletSkewSupported(); !supported {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
```
kikisdeliveryservice, line 298:
```suggestion
			coStatus.Message = fmt.Sprintf("one or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
```

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		return 0, fmt.Errorf("incorrect version syntax: %q", version)
+	}
+	minorVersion, err := strconv.ParseInt(tokens[1], 10, 32)
+	if err != nil {
+		return 0, err
+	}
+	return int(minorVersion), nil
+}
+
+// getMinimalSkewSupportNodeVersion returns the minimal node kubelet version should be
```
kikisdeliveryservice, line 370:
```suggestion
// getMinimalSkewSupportNodeVersion returns the minimal supported node kubelet version.
```

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+// upgrade to for skew support.
```
kikisdeliveryservice, line 371:
```suggestion
```

pkg/operator/status.go
sinnykumari: Shouldn't this event be of type "EventTypeWarning" ?

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		if supported, v := optr.isKubeletSkewSupported(); !supported {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
+			optr.eventRecorder.Eventf(co, corev1.EventTypeNormal, "KubeletSkewUnsupported", fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew with kube-apiserver version %v.", v))
```
sinnykumari, line 299: Shouldn't this event be of type "EventTypeWarning"?

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		}
+	}
+
 	return optr.updateStatus(co, coStatus)
 }
 
+// isKubeletSkewSupported checks the version skew of kube-apiserver and node kubelet version.
+// Returns false is the version skew is not supported. version skew > 2 is not supported.
```
sinnykumari, line 307: nit: rewording "Returns false if the version skew is not supported"

pkg/operator/status.go
```diff
@@ -289,9 +291,100 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+func (optr *Operator) isKubeletSkewSupported() (skewSupported bool, kubeAPIServerVersion string) {
+	skewSupported = true
+	nodes, err := optr.nodeLister.List(labels.Everything())
+	if err != nil {
+		return skewSupported, ""
```
sinnykumari, line 312: I am confused, in case of error shouldn't we returning "skewSupported=false"?

sinnykumari: Possibly also return error and set error message when an error has been encountered.

pkg/operator/status_test.go
```diff
@@ -621,3 +631,84 @@ func TestInClusterBringUpStayOnErr(t *testing.T) {
 
 	assert.False(t, optr.inClusterBringup)
 }
+
+func TestKubeletSkewSupported(t *testing.T) {
```
sinnykumari, line 635: Right now, we have test for happy path.To make it more robust, it feels like we should add more coverage in this unit tests by adding more test cases like kubelet version is something like  v1.21.0-rc.0+6143dea, when kubelet skew is supported, etc.

pkg/operator/status.go
QiWang19: See comment https://github.com/openshift/machine-config-operator/pull/2658#discussion_r663273670. The former discussion has question about this kind of error should cause failure in other processes and the message returned from here would not be very useful for customers. So ignore the error and the skew hasn't been checked at this point, shouldn't return false.

pkg/operator/status.go
kikisdeliveryservice: agree good catch

pkg/operator/status.go
```diff
@@ -289,9 +291,105 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		if supported, v := optr.isKubeletSkewSupported(); !supported {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("one or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
```
wking, line 298: I'm just noticing that [the enhancement][1] doesn't currently have much detail here.  There's some follow-up work still in flight [here][2] with:

> The Machine Config Operator should report a message in its reporting that admins should perform an action.

But based on my fuzzy recollection of some in-person meeting at some point, the MCO was going to warn folks who had outdated kubelets that they were behind, with hints about how to update.  Or maybe they were going to warn folks who were up against the boundary (so with the Kube-API-server's `Upgradeable=False` going off), with hints about how to update.  Staying silent until the nodes are _past_ the boundary seems like it's waiting too long.

[1]: https://github.com/openshift/enhancements/blame/3716ed25593447d9e624cb4fbabd90fcec66c58f/enhancements/update/eus-upgrades-mvp.md
[2]: https://github.com/openshift/enhancements/pull/800/files#diff-91c1b36fc770bc12ef9147ff77360aa00fb1f79d3df986a471d75d1c4a33891cR292-R293

QiWang19:
@kikisdeliveryservice PTAL. I saw the KubeletSkewUnsupported on the cluster, apiserver version is 1.21.1, kubelet version is 1.18.0. I will clean up new reviews and get must-gather.
```
$ os describe ClusterOperator
    Last Transition Time:  2021-07-20T22:28:30Z
    Message:               machine config pool contains nodes with an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least 1.19.1.
    Reason:                KubeletSkewUnsupported
    Status:                True
    Type:                  Upgradeable
```

pkg/operator/status.go
kikisdeliveryservice: This goes hand in hand with kube-apiserver, but might be worth having 2 cases:

1) KubeletSkewUnsupported for people past the boundary (which I think can happen since this won't be backported to infinity?) 
2) KubeletSkewPresent (or some better name than what I can think of right now) where we are informationally saying: that we recommend upgrading your pool (maybe they are just 1 skew), let them know then and maybe they will just upgrade and save us all the trouble? 

Basically if there is skew you always say something, but you'll probably want to say different things

pkg/operator/status.go
kikisdeliveryservice: I think what @sinnykumari is pointing out is that the issue is that if there is an error we're saying the skew IS supported, when we don't know?

kikisdeliveryservice: We use this func here:
```
if supported, v := optr.isKubeletSkewSupported(); !supported 
```
but if the kubeAPIServerVersion="" is intended to indicate that we didn't actually check the skew, you'd have to look at the value of v there.

QiWang19: > but if the kubeAPIServerVersion="" is intended to indicate that we didn't actually check the skew, you'd have to look at the value of v there.

Here it failed to get the node kublet version. Could you explain your idea?

QiWang19:
In the must-gather, the event shows in `registry-build01-ci-openshift-org-ci-ln-m0wnn6k-stable-sha256-955070604a96dd31ae3e857876595d97fc13eea864445a89abfa2f634d1929fa/namespaces/openshift-machine-config-operator/pods/machine-config-operator-57fcc46c4d-27xl7/machine-config-operator/machine-config-operator/logs/current.log`
[must-gather.tar.gz](https://github.com/openshift/machine-config-operator/files/6856568/must-gather.tar.gz)

pkg/operator/status.go
QiWang19: @kikisdeliveryservice Follow your suggestion, what if change to the following check and warn the users there's going to be an unsupported version skew?
```go
minorVSkew, kasV, kubeletV := optr.isKubeletSkewSupported()
if minorVSkew > 2 {
	coStatus.Reason = "KubeletSkewUnsupported"
	coStatus.Message = fmt.Sprintf("one or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(v))
} else if minorVSkew == 2 {
	coStatus.Reason = "KubeletSkewPresent"
	coStatus.Message = fmt.Sprintf("current kubelet version %v will not be supported by newer kube-apiserver. Please upgrade the the kubelet first if plan to upgrade the kube-apiserver")
}
```

pkg/operator/status.go
kikisdeliveryservice: Thinking about sinny's comment we can either:
1) log in the operator logs the error and the node
and also maybe:
2) log in the status the failure to check the skew and to see the operator logs for details

@sinnykumari wdyt?

pkg/operator/status.go
```diff
@@ -289,9 +291,105 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+	}
+
 	return optr.updateStatus(co, coStatus)
 }
 
+// isKubeletSkewSupported checks the version skew of kube-apiserver and node kubelet version.
+// Returns false if the version skew is not supported. version skew > 2 is not supported.
+func (optr *Operator) isKubeletSkewSupported() (skewSupported bool, kubeAPIServerVersion string) {
+	skewSupported = true
+	nodes, err := optr.nodeLister.List(labels.Everything())
```
kikisdeliveryservice, line 316: One question here: will this pull ALL the nodes in the cluster? bc after thinking about this we only want it to pull the nodes managed by the MCO (ie nodes that are in machine config pools)?

pkg/operator/status.go
sinnykumari: Definitely we should do 1, also doing 2 makes sense i.e. reporting error message to MCO coStatus with error message and details on where to look for further debugging.

pkg/operator/status_test.go
QiWang19: @sinnykumari Updated the test for both path tests and added more version parsing tests.

pkg/operator/status.go
QiWang19: > Definitely we should do 1, also doing 2 makes sense i.e. reporting error message to MCO coStatus with error message and details on where to look for further debugging.

@sinnykumari  Just to be clear, is this correct: 
- log in the operator logs is using glog.Error()
- log in the status is the coStatus.Message

pkg/operator/status.go
rphillips: Great question. I think we want all the nodes in the cluster, since a version skew that goes out of bounds might leave a kubelet not able to communicate to the API server.

kikisdeliveryservice: The problem with this is that those nodes wouldn't be managed by the MCO and by reporting it in our status we seem to be implying that we, in fact, do.

QiWang19:
@kikisdeliveryservice @sinnykumari PTAL. Added a commit https://github.com/openshift/machine-config-operator/pull/2658/commits/07d50206435b7ab99683d9878af255e20b1aed91 for the above reviews about the coStatus.Message and log errors.

kikisdeliveryservice:
> @kikisdeliveryservice @sinnykumari PTAL. Added a commit [07d5020](https://github.com/openshift/machine-config-operator/commit/07d50206435b7ab99683d9878af255e20b1aed91) for the above reviews about the coStatus.Message and log errors.

This is failing verify and unity tests. Could you please fix before we review?

pkg/operator/status.go
kikisdeliveryservice: The more I consider this, the less it makes sense to me as it's not within the operator's mandate to consider and/or touch nodes that are not members of MCPs.

QiWang19: If we don't check the non MCO nodes, not sure if other components will do the check?
What will be the result of that wrong implication? I saw the MCO manage the nodes using labels, if they use the MCO to manage the node, will they get notified of missing labels? So it will not be a big concern here.

QiWang19:
> > @kikisdeliveryservice @sinnykumari PTAL. Added a commit [07d5020](https://github.com/openshift/machine-config-operator/commit/07d50206435b7ab99683d9878af255e20b1aed91) for the above reviews about the coStatus.Message and log errors.
> 
> This is failing verify and unit tests. Could you please fix before we review?

Fixed. PTAL

pkg/operator/status.go
kikisdeliveryservice: Back to my original question: does this check pull in non-MCO managed nodes? If it does, the concern is that the MCO only manages nodes that are members of a pool, we do not manage, upgrade, etc.. nodes that are not in pools. Verifying/managing  these non-pool nodes is very much out of scope for the operator. Reporting their status as part of our operator status creates an untrue implication that the MCO manages them/watches them generally.

kikisdeliveryservice: Users are generally responsible for non-MCO managed nodes in a cluster, I don't think we want that relationship to change.

pkg/operator/status.go
kikisdeliveryservice: Skew is also present if the diff is 1 and I think Trevor is pointing out that we should say something whenever there is a skew and not wait.

pkg/operator/status.go
QiWang19: Fixed.

pkg/operator/status.go
```diff
@@ -289,9 +301,165 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		kubeletVersion string
+		nodes          []*corev1.Node
+	)
+	masterNodes, err := optr.nodeLister.List(labels.SelectorFromSet(labels.Set{ctrlcommon.MasterLabel: ""}))
+	if err != nil {
+		coStatus.Reason = "KubeletSkewUnchecked"
+		coStatus.Message = fmt.Sprintf("An error occurred when checking kubelet version skew: %v", err)
+		return skewUnchecked, coStatus, err
+	}
+	workerNodes, err := optr.nodeLister.List(labels.SelectorFromSet(labels.Set{node.WorkerLabel: ""}))
```
kikisdeliveryservice, line 390: Clusters can have custom pools as well.

pkg/operator/status.go
```diff
@@ -289,9 +301,165 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			lastError = err
+			continue
+		}
+		if nodeMinorVersion+2 < kubeAPIServerMinorVersion {
+			coStatus.Reason = "KubeletSkewUnsupported"
+			coStatus.Message = fmt.Sprintf("One or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet Version of at least %v.", getMinimalSkewSupportNodeVersion(kubeAPIServerVersion))
+			return skewUnsupported, coStatus, nil
+		}
+		if nodeMinorVersion+2 == kubeAPIServerMinorVersion {
+			coStatus.Reason = "KubeletSkewPresent"
```
kikisdeliveryservice, line 414: why is this not the value of the const?

pkg/operator/status.go
```diff
@@ -243,6 +246,15 @@ func (optr *Operator) syncDegradedStatus(ierr syncError) (err error) {
...
 }
 
+type versionSkew int
+
+const (
+	skewUnchecked versionSkew = iota
+	skewSupported
+	skewUnsupported
+	skewPresent
+)
```
kikisdeliveryservice, line 256: Can you explain what's happening here with these const? Why not assign them the value of the Reason strings?

kikisdeliveryservice (changes requested):
verify is still failing + some below comments since some of this seems confusing.

@rphillips there have been a bunch of changes, can you review/verify and let us know when this is ready?

QiWang19:
The new commit is a tentative commit based on the comments about report error if the version skew has been checked  https://github.com/openshift/machine-config-operator/pull/2658#discussion_r673590213 and about the KubeletSkewPresent https://github.com/openshift/machine-config-operator/pull/2658#discussion_r673567119

pkg/operator/status.go
rphillips: There will be a skew regardless on upgrades. Is the intent to report to the user on an upgrade there is a skew?

pkg/operator/status.go
rphillips: They can have custom pools, but they are always Masters or Workers, right?

kikisdeliveryservice: custom pools don't have to have a worker label: https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md#creating-a-custom-pool

QiWang19: Does those nodes have label llike `labels.SelectorFromSet(labels.Set{fmt.Sprintf("node-role.kubernetes.io/%s", pool.Name): ""})`?
If they do we can list the pools and filter the nodes by the label.

pkg/operator/status.go
rphillips: The enhancement is for >=2 skew... If we want to create a further enhancement for 4.10 for skew < 2 we can do that. It's out of scope of this PR.

pkg/operator/status.go
QiWang19: @sinnykumari PTAL. Added the error to coStatus and glog.Info() to log the current status.

rphillips:
We've add a few fixes to this PR from feedback here. A log line for the operator was added [here](https://github.com/openshift/machine-config-operator/pull/2658/files#diff-8a6a064b4db19eb3b29e5280447ac3ba9496e847391489d25c1282767c3354c5R310). Additionally, we added a new function `GetAllManagedNodes` to gather all the managed nodes in the system from all the pools. This will cover custom pools and the default pools we create.

pkg/operator/status.go
QiWang19: @kikisdeliveryservice  As https://github.com/openshift/machine-config-operator/pull/2658#issuecomment-885218200 commeted, the new function will get the mco managed node.

pkg/operator/status.go
rphillips: This should now be resolved in the latest commit.

rphillips:
@kikisdeliveryservice I believe this PR is ready... Takes into account all discussions on this thread.

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			continue
+		}
+		if nodeMinorVersion+2 < kubeAPIServerMinorVersion {
+			coStatus.Reason = skewUnsupported
+			coStatus.Message = fmt.Sprintf("One or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet version of at least %v.", getMinimalSkewSupportNodeVersion(kubeAPIServerVersion))
+			return skewUnsupported, coStatus, nil
+		}
+		if nodeMinorVersion+2 == kubeAPIServerMinorVersion {
+			coStatus.Reason = skewPresent
+			coStatus.Message = fmt.Sprintf("Current kubelet version %v will not be supported by newer kube-apiserver. Please upgrade the the kubelet first if plan to upgrade the kube-apiserver", kubeletVersion)
```
sinnykumari, line 404:
```suggestion
			coStatus.Message = fmt.Sprintf("Current kubelet version %v will not be supported by newer kube-apiserver. Please upgrade the kubelet first if plan to upgrade the kube-apiserver", kubeletVersion)
```

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			return skewPresent, coStatus, nil
+		}
```
sinnykumari, line 406: Looks like for `nodeMinorVersion == kubeAPIServerMinorVersion`, this function will return skewSupported which couldn't be confusing sine there is no skew at all. Shouldn't we return `noSkew` for this case?

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		skewStatus, status, err := optr.isKubeletSkewSupported(pools)
+		if err != nil {
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			return optr.updateStatus(co, coStatus)
```
sinnykumari, line 308: Can we log error message here as well so that operator pod log captures it?

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		}
+		glog.Infof("kubelet skew status: %v, status reason: %v, status message: %v", skewStatus, status.Reason, status.Message)
```
sinnykumari, line 310: I think we should not log message when there is no version skew, otherwise operator log will be flooded with the message while user doesn't have to take any action. Related to https://github.com/openshift/machine-config-operator/pull/2658/files#r675479794 . We may have to do same for `skewSupported` as well but that can be done later on as well.

sinnykumari:
Thanks for making all the changes! Few minor suggestions, other than that LGTM

pkg/operator/status.go
QiWang19: renamed `skewSupported` to `noSkew`.

pkg/operator/status.go
QiWang19: Added glof.Fatalf

QiWang19:
@sinnykumari PTAL.

pkg/operator/status.go
QiWang19: Are you suggesting `nodeMinorVersion == kubeAPIServerMinorVersion` shouldbe `noSkew` and `kubeAPIServerMinorVersion - nodeMinorVersion == 1` shoule be `skewSupported`?

For `coStatus`, `kubeAPIServerMinorVersion - nodeMinorVersion < 2` is out of the scoop of this PR https://github.com/openshift/machine-config-operator/pull/2658#discussion_r675082603, I don't see it is necessary for `isKubeletSkewSupported` to descrimintate it at this moment.

rphillips: @sinnykumari I think you misread the line perhaps? there is a +2 which would be skewPresent.

sinnykumari: What I meant was to add a new field `noSkew` to handle condition `nodeMinorVersion == kubeAPIServerMinorVersion`. If you replace it, that would mean for `nodeMinorVersion+1 == kubeAPIServerMinorVersion` as well it will return noSkew, although there is skewPresent. I think @rphillips  can tell better what is best here. Because in the end how these status is being consumed and message we want to give customer matters.

QiWang19: I revert the rename change, keep using `skewSupported` for `kubeAPIServerMinorVersion - nodeMinorVersion < 2`  until a consensus is reached.

sinnykumari:
I will let Kirsten do final review.

QiWang19:
@kikisdeliveryservice PTAL.

pkg/operator/status.go
rphillips: we don't want Fatalf, that will cause the controller to exit.

QiWang19: fixed.

pkg/operator/status.go
kikisdeliveryservice: @sinnykumari is this now ok?

kikisdeliveryservice (commented):
A few comments.

Can we also see a must gather since there've been so many changes to this PR?

Thanks!

pkg/operator/status.go
sinnykumari: yeah, based on https://github.com/openshift/machine-config-operator/pull/2658#discussion_r675082603 it seems handling n<2 is not part of current enhancement. So, it is upto node team how they get handled in 4.9.

QiWang19:
> A few comments.
> 
> Can we also see a must gather since there've been so many changes to this PR?
> 
> Thanks!

[must-gather.tar.gz](https://github.com/openshift/machine-config-operator/files/6885819/must-gather.tar.gz)
must-gather attached.  apiserver version is 1.21.1, kubelet version is 1.18.0.
```
// namespaces/openshift-machine-config-operator/pods/machine-config-operator-bbb5555d6-z5rvz/machine-config-operator/machine-config-operator/logs/current.log
2021-07-27T03:55:19.559426141Z I0727 03:55:19.559244       1 start.go:43] Version: 4.9.0-0.ci.test-2021-07-27-033800-ci-ln-04clxvb-latest (Raw: machine-config-daemon-4.6.0-202006240615.p0-975-g9f036fe9-dirty, Hash: 9f036fe9b49b6de2eacc991c37678587e14c0386)
2021-07-27T03:55:19.562886709Z I0727 03:55:19.562779       1 leaderelection.go:243] attempting to acquire leader lease openshift-machine-config-operator/machine-config...
2021-07-27T03:57:15.259786604Z I0727 03:57:15.259685       1 leaderelection.go:253] successfully acquired lease openshift-machine-config-operator/machine-config
2021-07-27T03:57:16.311510738Z I0727 03:57:16.311439       1 operator.go:262] Starting MachineConfigOperator
2021-07-27T03:57:22.603593629Z I0727 03:57:22.603512       1 status.go:325] kubelet skew status: KubeletSkewUnsupported, status reason: KubeletSkewUnsupported
2021-07-27T03:57:22.603811222Z I0727 03:57:22.603689       1 event.go:282] Event(v1.ObjectReference{Kind:"", Namespace:"", Name:"machine-config", UID:"4753add8-791d-4b39-8ab8-3ecfbc169be2", APIVersion:"", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'KubeletSkewUnsupported' One or more nodes have an unsupported kubelet version skew. Please see `oc get nodes` for details and upgrade all nodes so that they have a kubelet version of at least 1.19.1.
```

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		case skewUnsupported:
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			mcoObjectRef := &corev1.ObjectReference{
+				Kind:      co.Kind,
+				Name:      co.Name,
+				Namespace: co.Namespace,
+				UID:       co.GetUID(),
+			}
+			optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, coStatus.Reason, fmt.Sprintf("machine config pool contains nodes with an unsupported kubelet version skew."))
```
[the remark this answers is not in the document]
QiWang19: Fixed

pkg/operator/status.go
```diff
@@ -289,9 +299,172 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+		case skewPresent:
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			mcoObjectRef := &corev1.ObjectReference{
+				Kind:      co.Kind,
+				Name:      co.Name,
+				Namespace: co.Namespace,
+				UID:       co.GetUID(),
+			}
+			optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, coStatus.Reason, fmt.Sprintf("machine config pool contains nodes with an N-1 kubelet version skew."))
```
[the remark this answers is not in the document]
QiWang19: Fixed

pkg/operator/status.go
```diff
@@ -289,9 +299,175 @@ func (optr *Operator) syncUpgradeableStatus() error {
 		coStatus.Message = "One or more machine config pools are updating, please see `oc get mcp` for further details"
 	}
 
+	// don't overwrite status if updating or degraded
+	if !updating && !degraded {
+		skewStatus, status, err := optr.isKubeletSkewSupported(pools)
+		if err != nil {
+			glog.Errorf("Error checking version skew: %v, kubelet skew status: %v, status reason: %v, status message: %v", err, skewStatus, status.Reason, status.Message)
```
QiWang19, line 306: Wheather this `glog.Errorf` casued the CI failures? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2658/pull-ci-openshift-machine-config-operator-master-okd-e2e-aws/1419855633997369344/artifacts/e2e-aws/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-operator-598979445f-sd8qf_machine-config-operator.log

rphillips:
I tested this PR locally and it does report the skew, and go back to no skew as expected.

pkg/operator/status.go
```diff
@@ -289,9 +299,175 @@ func (optr *Operator) syncUpgradeableStatus() error {
...
+			coStatus.Reason = status.Reason
+			coStatus.Message = status.Message
+			mcoObjectRef := &corev1.ObjectReference{
+				Kind:      co.Kind,
+				Name:      co.Name,
+				Namespace: co.Namespace,
+				UID:       co.GetUID(),
+			}
+			glog.Infof("kubelet skew status: %v, status reason: %v", skewStatus, status.Reason)
+			optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, coStatus.Reason, coStatus.Message)
```
kikisdeliveryservice, line 338: just talked to ryan, and @QiWang19 can you remove this event? Since this is a supported state that might last fora  while, we don't think that filling the MCO's logs with it every 5 mins for a long time is going to be helpful and will create v v large log files.

kikisdeliveryservice (commented):
one last thing otherwise looks good

pkg/operator/status.go
QiWang19: Removed.

kikisdeliveryservice: thanks Qi!

kikisdeliveryservice (approved):
Thanks for all of the work on this. I think we've gotten it to a good state.

PR #2657, opened 2021-07-02
daemon: add log and event for uncordoning node

sinnykumari:
For better debugging in future for bugs like
https://bugzilla.redhat.com/show_bug.cgi?id=1975907,
let's explicitly log and add event when a node has
been successfully uncordoned by MCO.

PR #2659, opened 2021-07-02
pkg/daemon/drain: Clarify "cordon/uncordon" messages

wking:
So we're symmetric for both cordoning and uncording.

Including the node name in the message is not strictly necessary, because it is already in the event's referenced object.  But the message is pretty short otherwise, so repeating the node name inline seemed like it might be a good use of space.

### DIFF
--- a/pkg/daemon/drain.go
+++ b/pkg/daemon/drain.go
@@ -28,6 +28,11 @@ func (dn *Daemon) drainRequired() bool {
 }
 
 func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
+	verb := "cordon"
+	if !desired {
+		verb = "uncordon"
+	}
+
 	backoff := wait.Backoff{
 		Steps:    5,
 		Duration: 10 * time.Second,
@@ -38,16 +43,17 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
 		err := drain.RunCordonOrUncordon(dn.drainer, dn.node, desired)
 		if err != nil {
 			lastErr = err
-			glog.Infof("cordon/uncordon failed with: %v, retrying", err)
+			glog.Infof("%s failed with: %v, retrying", verb, err)
 			return false, nil
 		}
 		return true, nil
 	}); err != nil {
 		if err == wait.ErrWaitTimeout {
-			return errors.Wrapf(lastErr, "failed to cordon/uncordon node (%d tries): %v", backoff.Steps, err)
+			return errors.Wrapf(lastErr, "failed to %s node (%d tries): %v", verb, backoff.Steps, err)
 		}
-		return errors.Wrap(err, "failed to cordon/uncordon node")
+		return errors.Wrapf(err, "failed to %s node", verb)
 	}
+
 	return nil
 }
 


kikisdeliveryservice:
we don't need this, sinny already has an approved PR that is waiting to merge:
https://github.com/openshift/machine-config-operator/pull/2657

/hold

pkg/daemon/drain.go
```diff
@@ -38,16 +45,19 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
 		err := drain.RunCordonOrUncordon(dn.drainer, dn.node, desired)
 		if err != nil {
 			lastErr = err
-			glog.Infof("cordon/uncordon failed with: %v, retrying", err)
+			glog.Infof("%s failed with: %v, retrying", verb, err)
```
wking, line 48: @kikisdeliveryservice , any interest in this portion of the PR, or should we just close this?

sinnykumari: yep, it makes sense to keep this part.

wking: Rerolled to keep this, but leave eventing alone, in 5e2ca0f6 -> f302d7e4.  The new commit message tries to motivate logging but not including events within `cordonOrUncordonNode`, but let me know if my attempt at that motivation doesn't make sense.

pkg/daemon/drain.go
```diff
@@ -38,16 +43,18 @@ func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
...
 	}); err != nil {
 		if err == wait.ErrWaitTimeout {
-			return errors.Wrapf(lastErr, "failed to cordon/uncordon node (%d tries): %v", backoff.Steps, err)
+			return errors.Wrapf(lastErr, "failed to %s node (%d tries): %v", verb, backoff.Steps, err)
 		}
-		return errors.Wrap(err, "failed to cordon/uncordon node")
+		return errors.Wrapf(err, "failed to %s node", verb)
 	}
+
+	dn.logSystem("Node has been successfully %s", verb)
```
sinnykumari, line 57: For consistency, I would say let caller decide both logging and sending event.

wking: ok, rebased onto master and dropped this on-success log change with f302d7e4 -> 342c0135.

pkg/daemon/drain.go
```diff
@@ -28,6 +28,11 @@ func (dn *Daemon) drainRequired() bool {
 }
 
 func (dn *Daemon) cordonOrUncordonNode(desired bool) error {
+	verb := "cordoned"
```
kikisdeliveryservice, line 31:
```suggestion
	verb := "cordon"
```

pkg/daemon/drain.go
```diff
@@ -28,6 +28,11 @@ func (dn *Daemon) drainRequired() bool {
...
+	if !desired {
+		verb = "uncordoned"
```
kikisdeliveryservice, line 33:
```suggestion
		verb = "uncordon"
```

kikisdeliveryservice (commented):
I like this, just 1 change to match sentence.

pkg/daemon/drain.go
wking: Done with 342c0135 -> da38bbbd.

pkg/daemon/drain.go
wking: Done with 342c0135 -> da38bbbd.

kikisdeliveryservice (approved):
Thanks Trevor!

kikisdeliveryservice:
are commands dead or am impatient?

kikisdeliveryservice:
HAHAHA. Impatient it is.

PR #2666, opened 2021-07-07
OCPNODE-600: Add ContainersOOMKilled alert

harche:

**- What I did**
Add an alert that uses `container_runtime_crio_containers_oom_total` metric 

**- How to verify it**
If more than 5 containers on the host get OOMKilled then this alert should fire. 

ContainersOOMKilled alert to notify users if more than 5 containers get OOMKilled on the host.


install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -57,3 +57,13 @@ spec:
...
+    - name: containers-oom-killed
+      rules:
+        - alert: ContainersOOMKilled
+          expr: |
+            sum by (node) (container_runtime_crio_containers_oom_total) > 5
+          for: 15m
+          labels:
+            severity: warning
+          annotations:
+            message: "{{ $value | humanize }} containers have been killed on {{ $labels.node }} because they ran out of memory (OOM). Please make sure to allocate adequate enough memory for the pods or try to reduce the total number of pods."
```
saschagrunert, line 69: Yes, let's add an information that overcommitting the node is not recommended:

```suggestion
            message: "{{ $value | humanize }} containers have been killed on {{ $labels.node }} because they ran out of memory (OOM). Please make sure to limit the memory for the containers to not overcommit the node or try to reduce the total number of pods per node."
```

saschagrunert (approved):
LGTM

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -57,3 +57,23 @@ spec:
...
+          labels:
+            severity: warning
+          annotations:
+            message: "{{ $value | humanize }} containers have been killed on {{ $labels.node }} because they ran out of memory (OOM). Please make sure to limit the memory for the containers to not overcommit the node or try to reduce the total number of pods per node."
+    - name: containers-oom-killed-critical
+      rules:
+        - alert: ContainersOOMKilledCritical
+          expr: |
+            sum by (node) (container_runtime_crio_containers_oom_total) > 5
+          for: 15m
```
kikisdeliveryservice, line 75: Is there a reason to wait 15 min here to alert? We don't seem to be counting the number of OOM containers which would make sense to wait for but the number of killed containers which won't change even if recovered. So trying to figure out what the delay accomplishes? Or am I misunderstanding the alert?

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -57,3 +57,23 @@ spec:
...
+          labels:
+            severity: critical
+          annotations:
+            message: "{{ $value | humanize }} containers have been killed on {{ $labels.node }} because they ran out of memory (OOM). Please make sure to limit the memory for the containers to not overcommit the node or try to reduce the total number of pods per node."
```
kikisdeliveryservice, line 79: Also Q: the message is exactly the same for both, is there something more a user needs to know in the critical state? will functionality be impaired? etc..

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
harche: This alert is an a replacement for the existing alert `MultipleContainersOOMKilled` which also waits for 15min. The expression for `MultipleContainersOOMKilled` looks like, 

```
sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[12m])) and max by(namespace, container, pod) (kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}) == 1) > 5
```

@saschagrunert @haircommander How long crio metric `container_runtime_crio_containers_oom_total` will keep reporting OOMKilled that count after the container is OOMKilled? @kikisdeliveryservice has a point, that eventually `container_runtime_crio_containers_oom_total` should reduce the OOMKilled count after a certain timeout otherwise the recovered containers will keep getting counted in and the alert will keep on firing even after system comes out of memory pressure.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
harche: If just one of the container gets OOMKilled we want to raise a warning to the user so that they can look into that particular pod but if multiple containers are getting killed then we want to raise a critical alert because it may impair the system. I thought the `severity` of the alert was good enough to differentiate, but let me know if you have a better way to express that message.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
haircommander: I believe it is a global count for the time cri-o has been running. Is there persisted state for this alert? as in, can we check the amount that's been increased, or subtract the amount that we previously alerted on?

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
kikisdeliveryservice: What I'm getting at more is that you're using the same message for both and only changing the severity. I'd think you'd want to tell the user something different in the case of a critical situation?

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
rphillips: I think we want to use `rate` here to smooth out when the node comes out of memory pressure.

simonpasquier: +1 to what @rphillips said, you definitely want to apply `rate()` (or `increase()`) to the counter metric otherwise your alert might continue to fire forever.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -57,3 +57,23 @@ spec:
...
+    - name: containers-oom-killed-critical
+      rules:
+        - alert: ContainersOOMKilledCritical
+          expr: |
+            sum by (node) (container_runtime_crio_containers_oom_total) > 5
+          for: 15m
+          labels:
+            severity: critical
```
simonpasquier, line 77: Critical alerts should be carefully evaluated. The [Alerting consistency](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md) proposal isn't perfect but it provides some [guidelines](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#critical-alerts) to determine whether an alert deserves being critical.
cc @RiRa12621

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -57,3 +57,23 @@ spec:
...
+          for: 15m
+          labels:
+            severity: warning
+          annotations:
+            message: "{{ $value | humanize }} containers have been killed on {{ $labels.node }} because they ran out of memory (OOM). Please make sure to limit the memory for the containers to not overcommit the node or try to reduce the total number of pods per node."
+    - name: containers-oom-killed-critical
+      rules:
+        - alert: ContainersOOMKilledCritical
+          expr: |
+            sum by (node) (container_runtime_crio_containers_oom_total) > 5
```
ravitri, line 74: As I understand it, the criteria here is to alert on the total OOM killed pods at a node level crossing threshold within 15 minutes. I have following thoughts/queries:
1. Correct me if am wrong but I think the count of this metric will also increase if an insufficient memory limit is set for any app. In a scenario where the node has enough free memory but end user/customer app's (any app not in openshift-* or kube-* namespace) memory limit is insufficient, this OOM killed count will still increase and the rate can increase with more apps. For SRE, this alert would not be actionable at all if it's purely customer app configuration issue. 
2. Regarding OOM killed containers in general, if we as such scope it for the node then do you think it could be better to refer to the free/available node memory as the eventual symptom to alert upon? Else can scope the OOM killed alert for core openshift-*, kube-* and default namespaces?
3. If anyone gets paged for this alert, the message indicates that "some containers are OOM killed on a node". This could be any container/pod across all the namespaces. I think with reference to above two points, we could make the "actionable" message to be more informative.

harche: @ravitri Thanks for your input. I am happy to take any suggestions to make the alert message more actionable.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
harche: @rphillips @simonpasquier Thanks for your input, I have updated the PR with rate()

PR #2674, opened 2021-07-12
    apiserver-watcher: use a lockfile.

squeed:
We eventually want to rename this pod (since it's in the wrong namespace), but before we can do that, we have to handle multiple instances of the process. This is because the kubelet often handles static pod updates by creating the new one before deleting the old one.
    
So, write a lockfile. Fortunately, we're hostPID, so we can actually detect stale lockfiles. Whew.

/cc @aojea

### DIFF
--- a/cmd/apiserver-watcher/run.go
+++ b/cmd/apiserver-watcher/run.go
@@ -11,12 +11,14 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
 	"sync"
 	"time"
 
 	health "github.com/InVisionApp/go-health"
 	"github.com/InVisionApp/go-health/checkers"
 	"github.com/golang/glog"
+	"github.com/nightlyone/lockfile"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/machine-config-operator/pkg/version"
@@ -40,6 +42,10 @@ var (
 // indicates the route to the VIP should be withdrawn.
 const downFileDir = "/run/cloud-routes"
 
+// try to get the lock for 5 minutes
+const lockRetries = 30
+const lockRetryInterval = 10 * time.Second
+
 func init() {
 	rootCmd.AddCommand(runCmd)
 	runCmd.PersistentFlags().StringVar(&runOpts.rootMount, "root-mount", "/rootfs", "where the nodes root filesystem is mounted for writing down files or chrooting.")
@@ -48,6 +54,8 @@ func init() {
 
 type handler struct {
 	vips []string
+
+	lockfile lockfile.Lockfile
 }
 
 func runRunCmd(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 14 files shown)


cmd/apiserver-watcher/run.go
```diff
@@ -122,6 +134,9 @@ func runRunCmd(cmd *cobra.Command, args []string) error {
 			if err := handler.onFailure(); err != nil {
 				glog.Infof("Failed to mark service down on signal: %s", err)
 			}
+			if err := handler.lockfile.Unlock(); err != nil {
```
aojea, line 137: I couldn't put much though on this sorry, but I have the feeling this can go wrong in several places, just brainstorming here:

1. we get the lock on start and fail if we can't
2. we ONLY unlock on signal received

I don't know if there are some edge case we can exit without unlocking so we never will be able to start.

In case we always unlock on exit, a parallel pod restart will release the lock, despite the original pod is still active, since it never checks again if he has the lock ...

squeed: Since we're in HostPID namespace, we'll do the right thing and remove stale locks if the holding process is dead. I think we should be safe in this regard.

squeed:
hey @aojea I've rebased this for 4.10. Want to take a look? There's nobody else who understands this better than you and me.

aojea:
> This is because the kubelet often handles static pod updates by creating the new one before deleting the old one.

@mfojtik you've implemented something similar for the apiserver-operator and I remember the topic of kubelet updating static pod was discussed, but I can't remember the conclusion , I know that someone proposed that if the kubelet keeps 2 instance of the same static pod that should be considered a bug

squeed:
Right, the issue is that we want to change the namespace of the watcher. So those two pods are independent. It's not a kubelet bug, just a rename :/

cmd/apiserver-watcher/run.go
aojea: ahh, smart
> // Please note, that existing lockfiles containing pids of dead processes
> // and lockfiles containing no pid at all are simply deleted.

cmd/apiserver-watcher/run.go
```diff
@@ -149,9 +164,39 @@ func newHandler(uri *url.URL) (*handler, error) {
 		vips: addrs,
 	}
 	glog.Infof("Using VIPs %v", h.vips)
+
+	if err := os.MkdirAll(filepath.Dir(lockpath), 0755); err != nil {
+		return nil, fmt.Errorf("could not create run directory %s: %w", filepath.Dir(lockpath), err)
+	}
```
aojea, line 170: how was this created before?

aojea:
/retest
lgtm
one question about why we didnt have to create the folder before

aojea:
/lgtm
/hold
holding just for this question https://github.com/openshift/machine-config-operator/pull/2674/files#r704800469 , the tests are passing , so feel free to unhold if needed

cmd/apiserver-watcher/run.go
squeed: It was being created by the iptables scripts.

squeed:
@aojea @s-urbaniak reminded me we need to get this in. So, I've rebased and answered your question. Would you mind re-lgtm-ing?

aojea:
/lgtm
/approve
/hold cancel
Sorry for missing it :disappointed: 

/retest required

kikisdeliveryservice:
There is an alternative PR https://github.com/openshift/machine-config-operator/pull/3106

Adding a hold to both and you can let me know which you want to go with.

/hold
/assign @deads2k @squeed @aojea

aojea:
the alternative looks much simpler :smile:

PR #2675, opened 2021-07-12
Ignition 3.3.0 & Ignition Kargs

darkmuggle:
Pick-up PR @arithx  who has moved on
Original PR: https://github.com/openshift/machine-config-operator/pull/2613

---
**What I did**

- Bump the vendored version of ignition & ign-converter
- ~Switch internal usage from v3_2 Ignition structs to v3_3~
- Add support for the MCS to generate Ignition snippets modifying kargs based on the underlying MachineConfig

**How to verify it**

- Boot cluster w/ bootimage containing Ignition v3.2.0
- Boot cluster w/ bootimage containing Ignition v3.3.0
- Boot cluster w/ bootimage containing Ignition v3.3.0, w/ MachineConfig kargs specified before first boot of a machine. Validate that Ignition set the karg rather than the MCD.

**Description for the changelog**

Add Ignition v3.3.0 support
Utilize Ignition to set kernel arguments during the first boot


bgilbert:
> Switch internal usage from v3_2 Ignition structs to v3_3

I believe that line is from an older version of the PR, since we're no longer planning to do it for 4.9.

darkmuggle:
Let's /hold pending the usual test, git-smash, re-test (and I'm coming up to speed on the deep magic in this PR)

pkg/server/bootstrap_server.go
```diff
@@ -87,36 +88,66 @@ func (bsc *bootstrapServer) GetConfig(cr poolRequest) (*runtime.RawExtension, er
...
-			return nil, err
+			return nil, nil, err
 		}
 	}
 
+	return mc, &ignConf, nil
+}
+
+// GetConfig fetches the machine conf(type - Ignition) based on the pool request
+func (bsc *bootstrapServer) GetConfig(cr poolRequest) (*runtime.RawExtension, error) {
```
kikisdeliveryservice, line 118: GetConfig -> GetIgnConfig

for those of us skimming

pkg/server/cluster_server.go
```diff
@@ -80,27 +82,53 @@ func (cs *clusterServer) GetConfig(cr poolRequest) (*runtime.RawExtension, error
...
 		}
 	}
 
-	rawConf, err := json.Marshal(ignConf)
+	return mc, &ignConf, nil
+}
+
+// GetConfig fetches the machine config(type - Ignition) from the cluster,
+// based on the pool request.
+func (cs *clusterServer) GetConfig(cr poolRequest) (*runtime.RawExtension, error) {
```
kikisdeliveryservice, line 104: samesies as above

pkg/controller/common/helpers.go
```diff
@@ -278,6 +280,28 @@ func convertIgnition32to31(ign3config ign3types.Config) (ign3_1types.Config, err
 	return converted31, nil
 }
 
+// AddIgnitionKernelArguments adds the given kernel arguments to the Ignition configs
+// kernelArguments.shouldExist field
+func AddIgnitionKernelArguments(inRawExtIgn *runtime.RawExtension, kargs []string) (runtime.RawExtension, error) {
```
kikisdeliveryservice: maybe explicitly say firstboot somewhere here? so it's clear when this is used

pkg/controller/common/helpers.go
```diff
@@ -361,39 +361,34 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
...
+	// ParseCompatibleVersion will parse any config <= N to version N
+	ignCfgV3, rptV3, errV3 := ign3_3.ParseCompatibleVersion(rawIgn)
+	if errV3 == nil && !rptV3.IsFatal() {
+		// Regardless of the input version it has now been translated to a 3.3 config, downtranslate to 3.2 to match
+		// what is used internally
+		ignCfgV3_2, errV3_2 := v33tov32.Translate(ignCfgV3)
+		if errV3_2 != nil {
+			// This case should only be hit if fields that only exist in v3_3 are being used which are not
+			// currently supported by the MCO
+			return ign3types.Config{}, errors.Errorf("translating Ignition config failed with error %v", errV3_2)
```
kikisdeliveryservice, line 373: wondering if it's worth being explicit in the error message like in the comment?

kikisdeliveryservice (commented):
note last commit has typo in username.

Maybe worth updating the e2e to use v3 in some case:  https://github.com/openshift/machine-config-operator/blob/a5421965cf4e440b94bd1a4774072c517291fda0/test/e2e/mcd_test.go#L683

pkg/controller/common/helpers.go
```diff
@@ -278,6 +280,28 @@ func convertIgnition32to31(ign3config ign3types.Config) (ign3_1types.Config, err
...
+		return outRawExt, errors.Errorf("parsing Ignition config failed with error: %v\nReport: %v", errV3, rptV3)
+	}
+
+	for _, karg := range kargs {
+		ignCfg.KernelArguments.ShouldExist = append(ignCfg.KernelArguments.ShouldExist, ign3_3types.KernelArgument(karg))
+	}
+
+	outIgn, err := json.Marshal(ignCfg)
+	if err != nil {
+		return outRawExt, errors.Errorf("failed to marshal converted config: %v", err)
```
kikisdeliveryservice, line 298: this message is from here: https://github.com/openshift/machine-config-operator/blob/8bd0dd547059e9a5cbb43d2f58b40e6bece6c72f/pkg/controller/common/helpers.go#L177 
but i dont think converting is applicable in this case?

pkg/controller/common/helpers.go
```diff
@@ -337,39 +361,34 @@ func ValidateMachineConfig(cfg mcfgv1.MachineConfigSpec) error {
 // IgnParseWrapper parses rawIgn for both V2 and V3 ignition configs and returns
 // a V2 or V3 Config or an error. This wrapper is necessary since V2 and V3 use different parsers.
 func IgnParseWrapper(rawIgn []byte) (interface{}, error) {
-	ignCfgV3_2, rptV3_2, errV3_2 := ign3.Parse(rawIgn)
-	if errV3_2 == nil && !rptV3_2.IsFatal() {
+	// ParseCompatibleVersion will parse any config <= N to version N
+	ignCfgV3, rptV3, errV3 := ign3_3.ParseCompatibleVersion(rawIgn)
```
kikisdeliveryservice, line 365: this is a sweet change

pkg/server/api.go
```diff
@@ -134,10 +134,27 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	var serveConf *runtime.RawExtension
-	if reqConfigVer.Equal(*semver.New("3.2.0")) {
+	if reqConfigVer.Equal(*semver.New("3.3.0")) {
+		// If the config version is greater or equal to 3.3.0 we can add the kargs from the
+		// MachineConfig into the returned config
+		kargs, err := sh.server.GetKernelArguments(cr)
+		if err != nil {
+			w.Header().Set("Content-Length", "0")
+			w.WriteHeader(http.StatusInternalServerError)
+			glog.Errorf("couldn't get kernel arguments for req: %v, error: %v", cr, err)
```
kikisdeliveryservice, line 146: is there a reason we dont return here?

kikisdeliveryservice:
@darkmuggle sorry I think I left my comments prematurely - noticing there's a hold :grimacing:

pkg/server/bootstrap_server.go
darkmuggle: I agree that this is confusing, however, this comes from https://github.com/arithx/machine-config-operator/commit/7a18cfb59f5bb52587b6543101f26e26496ecc3a. 

I can happily change it -- and I wanted to -- but this PR is doing some heavy lifting already.

pkg/controller/common/helpers.go
darkmuggle: How so?

pkg/server/api.go
darkmuggle: Indeed it should. I'm not actually sure I like the code flow of fetching the config from https://github.com/openshift/machine-config-operator/pull/2675/files/513a02f4f53a5ac38bd5973f8d97da8ea510b58e#diff-9e2561386a5b88c1f09d35f98d4d4fa1cf4a3fb15a41183e21d5a42d2c24d939R125 and then re-feteching it in `GetKernelArguments`. I wanted to stay as true to Stephen's code and so I was approaching it from minizing changes. That said, since there is a chance for an error, it should be handled.

pkg/controller/common/helpers.go
kikisdeliveryservice: when i read it yesterday it didnt seem like this was a converted config? (correct me if im wrong)

pkg/server/bootstrap_server.go
kikisdeliveryservice: if possible would love this to be updated.

pkg/controller/common/helpers.go
darkmuggle: > @darkmuggle sorry I think I left my comments prematurely - noticing there's a hold 😬

pfft :) happy to have early feedback

pkg/server/api.go
darkmuggle: Fixed up

darkmuggle:
MCO team, don't mind me while I git-smash my way through my first E2E test.

test/e2e/mcs_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
+)
+
+// TestIgn33Kargs ensures that Kernel arguments are available in the Ingition
+// payloads when the client requests a 3.3.0 Spec.
+func TestIgn33Kargs(t *testing.T) {
```
kikisdeliveryservice, line 23: Is there anyway to combine this with :
https://github.com/openshift/machine-config-operator/blob/9d1da2759b62e4b7292f9825e7cfc8483167b427/test/e2e/mcd_test.go#L683

I don't think we need 2 ign tests.

But also for this test it's slow bc you are applying to whole worker pool. Rather, use a custom pool with 1 node for a faster rollout like the link above.

pkg/server/api.go
```diff
@@ -134,31 +134,47 @@ func (sh *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		// If the config version is greater or equal to 3.3.0 we can add the kargs from the
+		// MachineConfig into the returned config
+		kargs, err := sh.server.GetKernelArguments(cr)
+		if err != nil {
+			w.Header().Set("Content-Length", "0")
+			w.WriteHeader(http.StatusInternalServerError)
+			glog.Errorf("couldn't get kernel arguments for v3.3.0 req: %v, error: %v", cr, err)
+			return
+		}
+		if err := ctrlcommon.AddV33IgnitionKernelArguments(conf, kargs); err != nil {
```
yuqi-zhang, line 149: So I could be wrong on this, but the bootstrap server and the regular in-cluster server both can make use of this Addition of Ignition Kargs, right?

What's preventing the commonly-seen bootstrap issue of drifting between configs? i.e. the initial bootstrap served config will have the karg snippet if its using ignition spec 3.3.0, but the actual in-cluster rendered master MC does not. Won't that cause a `rendered-master-xxx` not found error? Or is this transparent to that process/I am misunderstanding something.

yuqi-zhang (commented):
Overall lgtm, one question below:

Thanks for picking up this work!

travier:
I don't know the status of this work but AFAIK this should probably happen at some point.
/remove-lifecycle rotten

bgilbert:
It should happen, but it's not being actively worked on from the CoreOS side.

PR #2676, opened 2021-07-14
Bug 1982052: Handle team/bond interfaces with a more clear message

abhat:
When teamed/bonded interfaces are presented to the ovs-configure
script, the uevent file doesn't exist under the device dir because
there is no actual net-device corresponding to the bond. This
commit adds a more clear error message for handling team/bonded
interfaces.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -74,10 +74,14 @@ contents:
       # REMOVEME: Once BZ:1854355 is fixed, this needs to get removed.
       function configure_driver_options {
         intf=$1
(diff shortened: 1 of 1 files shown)


trozet (commented):
@abhat one question, do we need to set the allmulti on the slaves if they are using vmxnet3 driver or no?

abhat:
From what @rbbratta mentioned it's working without setting the `allmulti` option on the bonded interfaces. So I feel like just sipping the bond/team interface is sufficient.

kikisdeliveryservice (approved):

PR #2677, opened 2021-07-15
daemon: Explicitly start rpm-ostreed, restart if we detect active txn

cgwalters:
This is a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1982389
which is already fixed in rpm-ostree in
https://github.com/coreos/rpm-ostree/pull/2995
because it will take a fair while until we can ship the fixed
rpm-ostreed in RHEL and then OpenShift stable versions.
(Yes, this is a sad recurring pattern)

The updater client gains an explicit `Initialize` method, where
we also explicitly `systemctl start rpm-ostreed` which
then effectively rolls in the change from
https://github.com/coreos/rpm-ostree/pull/2945
too.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -212,14 +212,12 @@ func New(
 
 	// Only pull the osImageURL from OSTree when we are on RHCOS or FCOS
 	if hostos.IsCoreOSVariant() {
+		err := nodeUpdaterClient.Initialize()
+		if err != nil {
+			return nil, fmt.Errorf("error initializing rpm-ostree: %v", err)
+		}
 		osImageURL, osVersion, err = nodeUpdaterClient.GetBootedOSImageURL()
 		if err != nil {
-			// If this fails for some reason, let's dump the unit status
-			// into our logs to aid future debugging.
-			cmd := exec.Command("systemctl", "status", "rpm-ostreed")
-			cmd.Stdout = os.Stdout
-			cmd.Stderr = os.Stderr
-			_ = cmd.Run()
 			return nil, fmt.Errorf("error reading osImageURL from rpm-ostree: %v", err)
 		}
 		glog.Infof("Booted osImageURL: %s (%s)", osImageURL, osVersion)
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -212,7 +212,10 @@ func New(
 
 	// Only pull the osImageURL from OSTree when we are on RHCOS or FCOS
 	if hostos.IsCoreOSVariant() {
-		osImageURL, osVersion, err = nodeUpdaterClient.GetBootedOSImageURL()
+		err := nodeUpdaterClient.Initialize()
+		if err != nil {
+			return nil, fmt.Errorf("error initializing rpm-ostree: %v", err)
+		}
 		if err != nil {
```
kikisdeliveryservice, line 219: how do we ever hit this second `if err != nil` if we already returned `if err != nil` in line 216 above? Feels like we either don't need 219-227 or we're missing some other call right before that check?

pkg/daemon/rpm-ostree.go
```diff
@@ -96,7 +97,47 @@ func (r *RpmOstreeClient) GetBootedDeployment() (*RpmOstreeDeployment, error) {
 		return nil, errors.Wrapf(err, "failed to parse `rpm-ostree status --json` output (%s)", truncate(string(output), 30))
 	}
 
-	for _, deployment := range rosState.Deployments {
+	return &rosState, nil
+}
+
+func (r *RpmOstreeClient) Initialize() error {
+	// This replicates https://github.com/coreos/rpm-ostree/pull/2945
```
kikisdeliveryservice, line 104: should this comment be for the whole func?

kikisdeliveryservice (commented):
Two questions :)

pkg/daemon/daemon.go
sinnykumari: +1, looks like it can either be moved after GetBootedOSImageURL() call or removed if we no longer need since we just restarted rpm-ostreed service

cgwalters: Good catch!  Fixed.

pkg/daemon/rpm-ostree.go
cgwalters: Mmm...I think it's right as is because we're just doing the `systemctl start` bit, but not the "dump status on error" bit which is instead in the daemon code.  But it'd probably make sense to live here too....done!

sinnykumari (approved):

kikisdeliveryservice:
Thanks Colin!

/lgtm

PR #2678, opened 2021-07-19
Add CRI-O metrics collectors configuration

saschagrunert:
**- What I did**
We now can configure metrics by referencing the collector since
https://github.com/cri-o/cri-o/pull/5061 has been merged.

This patch adds the default CRI-O metrics collectors and reflects the
previously overwritten state from the Cluster Monitoring Operator (CMO).

**- How to verify it**
Verify that the configured metrics are being served.

Added CRI-O `metrics_collectors` configuration.

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -45,3 +45,19 @@ contents:
     [crio.metrics]
     enable_metrics = true
     metrics_port = 9537
+    metrics_collectors = [
+      "operations",
+      "operations_latency_microseconds_total",
+      "operations_latency_microseconds",
+      "operations_errors",
+      "image_pulls_layer_size",
+      "containers_oom_total",
+      "containers_oom",
+      # Drop metrics with excessive label cardinality.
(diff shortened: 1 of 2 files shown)


saschagrunert:
@QiWang19 @mrunalp @rphillips @haircommander PTAL

QiWang19:
Do we also need add the same to `templates/worker/01-worker-container-runtime/_base/files/crio.yaml`?

saschagrunert:
> Do we also need add the same to `templates/worker/01-worker-container-runtime/_base/files/crio.yaml`?

Yes, sure. Thank you for the hint. Updated the PR as suggested.

QiWang19 (commented):
LGTM

saschagrunert:
PTAL @yuqi-zhang @kikisdeliveryservice

kikisdeliveryservice:
As MCO doesn't own these templates, would like one of the SMEs to look:

/assign @umohnani8 @mtrmac @haircommander

kikisdeliveryservice (approved):
This seemed generally fine and since an SME weighted in, let's merge.

PR #2679, opened 2021-07-19
Bug 1982766: [on-prem] Make ingress VIP more tolerant to failures

mandre:
We've noticed that the ingress VIP could be flapping in some
environments where the keepalived check fails due to timeout. We should
make the ingress check more resilient to errors and start using `fall`
and `raise` to mitigate flapping as we did for the API checks.

### DIFF
--- a/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
+++ b/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
@@ -41,12 +41,16 @@ contents:
         script "/usr/bin/timeout 0.9 /usr/bin/curl -o /dev/null -Lfs http://localhost:1936/healthz/ready"
         interval 1
         weight 20
+        rise 3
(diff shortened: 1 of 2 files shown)


mandre:
/cc @cybertron @yboaron 
It seems like a sensible thing to add `fall` and `raise` directives to the ingress checks like we do on the API check and I was wondering if this was something that has previously been discussed and discarded or just an overlook. Reviews welcome :)

cybertron:
/lgtm

Agreed, this seems quite reasonable.

/test e2e-openstack
/test e2e-ovirt
/test e2e-vsphere
/cc @Gal-Zaidman @jcpowermac @patrickdillon

mandre:
As the bot suggests.
/assign sinnykumari

All relevant jobs are passing, PTAL.

sinnykumari (approved):
All relevant tests are passing.

PR #2681, opened 2021-07-19
Bug 2020003: Node upgrade stuck due to not writing through dangling symlink '/etc/machine-config-daemon/orig/etc/issue.mcdorig'

jkyros:
This PR fixes a condition where: 

1.  The mcd backs up a relative/dangling symlink while backing up original files during a configuration rollout 
2.  The next time configuration gets rolled out, the mcd checks to see if it has already backed up the original files 
3.  os.Stat returns an error against the previously backed-up symlink because the symlink's target does not exist
4.  The mcd interprets that error as the file not existing and tries to copy over it, resulting in  "cp: not writing through dangling symlink"

This PR adds a symlink check before the file existence check, to make sure we don't try to back up over the dangling symlink. 
 
This also adds a unit test for original file backups that tests the symlink case as well as the "normal" case. I broke the backup file paths out into package variables so I could override them for the test. 

**This PR does NOT address the problem of a user undesirably using machine config to overwrite symlinks with files, it only prevents the MCO from degrading in the event that such a configuration were to occur.** 

Fixes:  https://bugzilla.redhat.com/show_bug.cgi?id=2020003

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -56,6 +56,11 @@ const (
 	postConfigChangeActionReloadCrio = "reload crio"
 )
 
+var (
+	origParentDirPath   = filepath.Join("/etc", "machine-config-daemon", "orig")
+	noOrigParentDirPath = filepath.Join("/etc", "machine-config-daemon", "noorig")
+)
+
 func writeFileAtomicallyWithDefaults(fpath string, b []byte) error {
 	return writeFileAtomically(fpath, b, defaultDirectoryPermissions, defaultFilePermissions, -1, -1)
 }
@@ -1603,11 +1608,11 @@ func (dn *Daemon) writeFiles(files []ign3types.File) error {
 }
 
 func origParentDir() string {
-	return filepath.Join("/etc", "machine-config-daemon", "orig")
+	return origParentDirPath
 }
 
 func noOrigParentDir() string {
-	return filepath.Join("/etc", "machine-config-daemon", "noorig")
+	return noOrigParentDirPath
 }
 
 func origFileName(fpath string) string {
@@ -1638,6 +1643,14 @@ func createOrigFile(fromPath, fpath string) error {
 			return errors.Wrapf(err, "creating no orig parent dir: %v", err)
 		}
 		return writeFileAtomicallyWithDefaults(noOrigFileStampName(fpath), nil)
+	}
+
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1970959
+	// orig file might exist, but be a relative/dangling symlink
+	if symlinkTarget, err := os.Readlink(origFileName(fpath)); err == nil {
+		if symlinkTarget != "" {
+			return nil
+		}
 	}
 	if _, err := os.Stat(origFileName(fpath)); err == nil {
 		// the orig file is already there and we avoid creating a new one to preserve the real default
--- a/pkg/daemon/update_test.go
+++ b/pkg/daemon/update_test.go
@@ -2,7 +2,10 @@ package daemon
 
 import (
 	"fmt"
+	"io/ioutil"
 	"math/rand"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/daemon/update.go
```diff
@@ -1605,11 +1610,11 @@ func (dn *Daemon) writeFiles(files []ign3types.File) error {
 }
 
 func origParentDir() string {
-	return filepath.Join("/etc", "machine-config-daemon", "orig")
+	return origParentDirPath
```
kikisdeliveryservice, line 1613: I think you did this for testing (?), but is there a reason you can't just mock the return value of this func? instead of having the var and func both existing at the same time?

kikisdeliveryservice (commented):
A question

pkg/daemon/update.go
jkyros: Yes, I am testing to make sure createOrigFile() is working properly, and I need a way to get it to write its files to a temporary directory during the test, but the file locations are essentially hard-coded because of those function calls.

Maybe I shouldn't worry about the test (or maybe it should be an e2e test because it does file i/o?). I don't know where you want to draw the line. 

Anyway, my options as I saw them were: 
	

1. make origParentDir/noOrigParentDir function pointers and then mock/override their functions for the test:
`var origParentDir = func() string {
	return filepath.Join("/etc", "machine-config-daemon", "orig")
}`
2. make package variables for the strings and have createOrigFile use those strings directly instead of the function calls ( I figured that would be "too much change" for no benefit )
3. have origParentDir/noOrigParentDir/createOrigFile take an path/prefix argument so I could redirect where they put their files ( again, I felt like that was too much change for no benefit) 
4. make package variables for the path strings and just have the existing functions return them 

I went with option 4, but I can do option 1 if that's preferable?

pkg/daemon/update_test.go
```diff
@@ -687,3 +690,74 @@ func TestRunGetOut(t *testing.T) {
...
+	err = os.Remove(fileOverSymlink)
+	assert.Nil(t, err)
+	err = ioutil.WriteFile(fileOverSymlink, []byte("replacement contents"), 0755)
+	assert.Nil(t, err)
+
+	// Try to back it up again make sure it knows it's already backed up
+	err = createOrigFile(relativeSymlink, relativeSymlink)
+	assert.Nil(t, err)
+
+	// Finally, make sure we can restore the relative symlink if we rollback
```
yuqi-zhang, line 759: Just to make sure, even if it is a dangling symlink, we still restore it now correctly right?

I see we do a `cp -a --reflink=auto` so it seems that it should restore the (still dangling) symlink

yuqi-zhang (approved):
Sorry for the late review, I think this generally LGTM, thanks for the tests!

I think the current variable paths method is fine, but will defer to others for a review as well.

One question below:

pkg/daemon/update_test.go
jkyros: Yes, we still restore it correctly. I was worried about that too! :) 

Like in the case of /etc/issue, the symlink wasn't "wrong", it was only dangling because of where we put it, if we restored it, it wouldn't be dangling anymore, it would just be relative (and correct).

pkg/daemon/update.go
```diff
@@ -58,6 +58,11 @@ const (
 	postConfigChangeActionReloadCrio = "reload crio"
 )
 
+var (
```
sinnykumari, line 61: These variable are not supposed to be updated, so this could be made const.

sinnykumari:
I am not sure if this is the right way to fix the issue. I looked at the issue and it seems, user is trying to update content of /etc/issue by applying a MachineConfig. On a 4.8 cluster RHCOS node, I see:
$ ls -l /etc/issue lrwxrwxrwx. 1 root root 16 Jul 28 14:38 /etc/issue -> ../usr/lib/issue

pkg/daemon/update.go
jkyros: - I would have made them const if they weren't paths being assembled with filepath.Join (which I assume was done as good practice for cross-platform separators, but I don't know how likely it is that this will need to be cross platform)
- If I make them constants, I can't assign to them in the test to redirect the files for testing

jkyros:
I see what you're saying. writeFileAtomically's behavior is to replace the symlink with a file, not to write the file contents to the symlink's target

https://github.com/openshift/machine-config-operator/blob/4ad850b8649af61992796dcf923d986861bd82ab/pkg/daemon/update.go#L67

 because under the hood it's a "rename" operation not a direct write. https://github.com/openshift/machine-config-operator/blob/4ad850b8649af61992796dcf923d986861bd82ab/vendor/github.com/google/renameio/tempfile.go#L106

I was assuming this was "desired behavior" :smile: 

Also...the compliance operator seems to encourage modifying /etc/issue on RHCOS as part of its audits -- at least one of the users that hit this was using the compliance operator -- so while the current behavior is maybe non-intuitive/incorrect,  if we do change this to act on the target file vs the symlink itself,  there will be some consequence.

kikisdeliveryservice:
while this is being sorted out
/hold

yuqi-zhang:
I think there's 2 issues here:

1. handling of dangling symlinks in the MCD in general
2. writing to locations we shouldn't be writing to

Which are somewhat tangential. We don't really do a great job of guarding against 2, so it's on a per-error basis. I agree with Sinny in that in the particular case of the BZ the user should have probably just written to `/etc/issue.d`. I am ok with the PR as a fix to 1, for other potential dangling symlinks that would normally be in writable locations.

As for the compliance operator, if we overwrite `/etc/issue` directly, technically its ok because the location of `/usr/lib/issue` would just not get used at all, and the symlink just becomes a file. However if we want to move compliance operator to using `issue.d`, maybe we can move the BZ and get their thoughts on it as well.

Would that be an ok path? (merge this for 1, move BZ for 2)

sinnykumari:
@jkyros I agree that MCO is not really doing good job in updating symlink files.

@yuqi-zhang sure, we can go ahead with this PR for in general improvement and move this bug to compliance team to see if they can update instructions to write to /etc/issue.d/ instead of in /etc/issue file. Suggesting to directly write directly into /etc/issue by making symlink to a file shouldn't be ideally done since that file is coming a spart of OS update and may cause issue(RHCOS team would have better idea)

In addition to above two points, I think we should also create a story for MCO to discuss about right way of handling symlink files for both cases editable and read-only.

Based on above discussion, this PR would need update in messaging and perhaps creating a new bz if we want to backport to older releases.

yuqi-zhang (approved):
/lgtm

updated commit message looks good, I think we are good to merge this, thanks!

kikisdeliveryservice:
since jerry is the same person who gave the initial approval waiting for sinny to re-review

/assign @sinnykumari

sinnykumari:
It took me a while to recollect context of this bug. PR is LGTM but before we merge want to know what we decided upon https://github.com/openshift/machine-config-operator/pull/2681#issuecomment-889180191 i.e.
1. Creating a new bug that handles dangling symlink and use that bug to backport this fix if needed
2. Move original bug bz#1970959 to compliance team to fix the recommendation with using /etc/issue.d/ to write new content instead of writing directly to /etc/issue
3. Creating a jira card to track better handling of writing to symlink using MC (may already exist)

sinnykumari:
Thanks John for making all the changes. LGTM
/hold cancel

PR #2684, opened 2021-07-19
WIP crio: add a tighter seccomp profile and use it by default

haircommander:
[for fun and profit](https://www.openshift.com/blog/seccomp-for-fun-and-profit)

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio-seccomp-json.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio-seccomp-json.yaml
@@ -1,0 +1,109 @@
+mode: 0644
+path: "/etc/crio/crio-seccomp.json"
+contents:
+  inline: |
+    {
+        "defaultAction": "SCMP_ACT_ALLOW",
+        "archMap": [
+            {
+                "architecture": "SCMP_ARCH_X86_64",
+                "subArchitectures": [
+                    "SCMP_ARCH_X86",
+                    "SCMP_ARCH_X32"
+                ]
+            },
+            {
+                "architecture": "SCMP_ARCH_AARCH64",
+                "subArchitectures": [
+                    "SCMP_ARCH_ARM"
+                ]
+            },
+            {
+                "architecture": "SCMP_ARCH_MIPS64",
+                "subArchitectures": [
+                    "SCMP_ARCH_MIPS",
+                    "SCMP_ARCH_MIPS64N32"
+                ]
+            },
+            {
+                "architecture": "SCMP_ARCH_MIPS64N32",
+                "subArchitectures": [
+                    "SCMP_ARCH_MIPS",
+                    "SCMP_ARCH_MIPS64"
+                ]
+            },
+            {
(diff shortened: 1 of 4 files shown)


cgwalters:
This PR will be useful as a reference point for e2e testing with that profile enabled by default.  But I don't think we can ever merge this as is - we have to make it opt-in.

Long term I personally think the best thing would be:

- podman vendors https://github.com/containerd/containerd/blob/main/contrib/seccomp/seccomp_default.go
- cri-o somehow dynamically loads the seccomp policy via forking `podman` (i.e. we only have one default seccomp policy which lives in Fedora/RHEL and applies uniformly to `podman run` and OpenShift)
- Document how to deploy and enforce custom hardened policies in OpenShift, with particular carve-outs for things like privileged pods and use cases like `buildah-in-OpenShift` that require `clone(CLONE_NEWUSER)` etc.

haircommander:
Another option is we could take the approach we did for capabilities: add an mc that adds these files as well as one that does not have these files. Then, when a new user installs, they'd get the tighter default, but if they upgrade they won't inherit it.

haircommander:
> This PR will be useful as a reference point for e2e testing with that profile enabled by default. But I don't think we can ever merge this as is - we have to make it opt-in.
> 
> Long term I personally think the best thing would be:
> 
>     * podman vendors https://github.com/containerd/containerd/blob/main/contrib/seccomp/seccomp_default.go

I'm not sure how viable this will be for podman/cri-o teams as we use different spec generation libraries as well. Further, I think there are syscalls that podman-in-podman or podman-in-crio would need. I am not sure about what to do about the PinP case, but for the PinC case, we could investigate the applying the buildah-in-Openshift profile as well.

haircommander:
cc @saschagrunert

cgwalters:
> Another option is we could take the approach we did for capabilities: add an mc that adds these files as well as one that does not have these files. Then, when a new user installs, they'd get the tighter default, but if they upgrade they won't inherit it.

I think as an OpenShift admin I'd expect this to be tied to `securitycontextconstraint`.  That already supports a `seccompProfiles` array.

templates/master/01-master-container-runtime/_base/files/crio-seccomp-json.yaml
```diff
@@ -0,0 +1,241 @@
+mode: 0644
+path: "/etc/crio/crio-seccomp.json"
```
saschagrunert, line 2: What's the reason for adding it here as file in favor of having the profile in-memory: https://github.com/cri-o/cri-o/blob/af2035b294e34cfe64863922d985468d27d9de79/internal/config/seccomp/seccomp.go#L30 ?

haircommander: all encompassed in this PR to test :)

saschagrunert: Got it, thank you for the clarification.

PR #2686, opened 2021-07-20
[MCC][MCD]: Introduce in progress taint

ravisantoshgudimetla:
**- What I did**
All the upgrade candidate nodes in a mcp would be applied
`UpdateInProgress: PreferNoSchedule` taint.
The taint will be removed MCD once the upgrade is complete.
Since kubernetes/kubernetes#104251 landed,
the nodes not having PreferNoSchedule taint will have higher score.
Before the upgrade starts, MCC will taint all the nodes in the
cluster that are supposed to be upgraded. Once the upgrade is
complete since MCD will remove the taint, none of the nodes will
have `UpdateInProgress: PreferNoSchedule` taint. This ensures
the score of the nodes will be equal again.

Why is this needed?
This reduces the pod churn when the cluster upgrade is in progress.
When the non-upgraded nodes in the cluster have `UpdateInProgress:
PreferNoSchedule` taint, they would get lesser score and the pods
would prefer to land onto untainted(upgraded) nodes there by
reducing the chances of landing onto an unupgraded node which
can cause one more reschedule
**- How to verify it**

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"github.com/openshift/machine-config-operator/internal"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/pkg/constants"
 	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
 	daemonconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
@@ -64,12 +65,6 @@ const (
 	// masterPoolName is the control plane MachineConfigPool name
 	masterPoolName = "master"
 )
-
-var nodeUpdateBackoff = wait.Backoff{
-	Steps:    5,
-	Duration: 100 * time.Millisecond,
-	Jitter:   1.0,
-}
 
 // Controller defines the node controller.
 type Controller struct {
@@ -488,6 +483,10 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 			ctrl.logPoolNode(pool, curNode, "changed labels")
 			changed = true
 		}
+		if !reflect.DeepEqual(oldNode.Spec.Taints, curNode.Spec.Taints) {
+			ctrl.logPoolNode(pool, curNode, "changed taints")
+			changed = true
+		}
 	}
 
 	if !changed {
@@ -766,7 +765,18 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 	if err := ctrl.setClusterConfigAnnotation(nodes); err != nil {
 		return goerrs.Wrapf(err, "error setting clusterConfig Annotation for node in pool %q, error: %v", pool.Name, err)
 	}
-
+	// Taint all the nodes in the node pool, irrespective of their upgrade status.
+	ctx := context.TODO()
+	for _, node := range nodes {
+		// All the nodes that need to be upgraded should have `NodeUpdateInProgressTaint` so that they're less likely
+		// to be chosen during the scheduling cycle.
+		targetConfig := pool.Spec.Configuration.Name
+		if node.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey] != targetConfig {
+			if err := ctrl.setUpdateInProgressTaint(ctx, node.Name); err != nil {
+				return goerrs.Wrapf(err, "failed applying %s taint for node %s", constants.NodeUpdateInProgressTaint.Key, node.Name)
+			}
+		}
+	}
 	candidates, capacity := getAllCandidateMachines(pool, nodes, maxunavail)
 	if len(candidates) > 0 {
 		ctrl.logPool(pool, "%d candidate nodes for update, capacity: %d", len(candidates), capacity)
@@ -834,7 +844,7 @@ func (ctrl *Controller) setClusterConfigAnnotation(nodes []*corev1.Node) error {
 }
 
 func (ctrl *Controller) setDesiredMachineConfigAnnotation(nodeName, currentConfig string) error {
-	return clientretry.RetryOnConflict(nodeUpdateBackoff, func() error {
+	return clientretry.RetryOnConflict(constants.NodeUpdateBackoff, func() error {
 		oldNode, err := ctrl.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
 		if err != nil {
 			return err
@@ -871,7 +881,6 @@ func (ctrl *Controller) setDesiredMachineConfigAnnotation(nodeName, currentConfi
 // capacity.  It is the reponsibility of the caller to choose a subset of the nodes given the capacity.
 func getAllCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*corev1.Node, maxUnavailable int) ([]*corev1.Node, uint) {
 	targetConfig := pool.Spec.Configuration.Name
-
 	unavail := getUnavailableMachines(nodesInPool)
 	// If we're at capacity, there's nothing to do.
 	if len(unavail) >= maxUnavailable {
@@ -888,10 +897,8 @@ func getAllCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*core
 			}
 			continue
 		}
-
 		nodes = append(nodes, node)
 	}
-
 	// Nodes which are failing to target this config also count against
 	// availability - it might be a transient issue, and if the issue
 	// clears we don't want multiple to update at once.
@@ -899,7 +906,6 @@ func getAllCandidateMachines(pool *mcfgv1.MachineConfigPool, nodesInPool []*core
 		return nil, 0
 	}
 	capacity -= failingThisConfig
-
 	return nodes, uint(capacity)
 }
 
@@ -971,6 +977,47 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool, 
 		ctrl.eventRecorder.Eventf(pool, corev1.EventTypeNormal, "SetDesiredConfig", "Set target for %d nodes to config %s", targetConfig)
 	}
 	return nil
+}
+
+// setUpdateInProgressTaint applies in progress taint to all the nodes that are to be updated.
+// The taint on the individual node is removed by MCD once the update of the node is complete.
+// This is to ensure that the updated nodes are being preferred to non-updated nodes there by
+// reducing the number of reschedules.
+func (ctrl *Controller) setUpdateInProgressTaint(ctx context.Context, nodeName string) error {
+	return clientretry.RetryOnConflict(constants.NodeUpdateBackoff, func() error {
+		oldNode, err := ctrl.kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		oldData, err := json.Marshal(oldNode)
+		if err != nil {
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
+			newNode.Spec.Taints = []corev1.Taint{}
+		}
+
+		for _, taint := range newNode.Spec.Taints {
+			if taint.MatchTaint(constants.NodeUpdateInProgressTaint) {
+				return nil
+			}
+		}
+
+		newNode.Spec.Taints = append(newNode.Spec.Taints, *constants.NodeUpdateInProgressTaint)
+		newData, err := json.Marshal(newNode)
+		if err != nil {
+			return err
+		}
+
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, corev1.Node{})
+		if err != nil {
+			return fmt.Errorf("failed to create patch for node %q: %v", nodeName, err)
+		}
+		_, err = ctrl.kubeClient.CoreV1().Nodes().Patch(ctx, nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
+		return err
+	})
 }
 
 func maxUnavailable(pool *mcfgv1.MachineConfigPool, nodes []*corev1.Node) (int, error) {
--- a/pkg/controller/node/node_controller_test.go
+++ b/pkg/controller/node/node_controller_test.go
@@ -3,6 +3,7 @@ package node
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/machine-config-operator/pkg/constants"
 	"reflect"
 	"testing"
 	"time"
@@ -789,9 +790,22 @@ func TestShouldMakeProgress(t *testing.T) {
 	mcpWorker := helpers.NewMachineConfigPool("worker", nil, helpers.WorkerSelector, "v1")
 	mcp.Spec.MaxUnavailable = intStrPtr(intstr.FromInt(1))
 
+	// Node2 is at desired config, so need to do a get on the node2 to check for the taint status
+	node2 := newNodeWithLabel("node-2", "v1", "v1", map[string]string{"node-role/worker": "", "node-role/infra": ""})
+	// Update node-2 to have the needed taint, this should still have no effect
+	node2.Spec.Taints = []corev1.Taint{*constants.NodeUpdateInProgressTaint}
+	node3 := newNodeWithLabel("node-3", "v0", "v0", map[string]string{"node-role/worker": "", "node-role/infra": ""})
+	// Update node-3 to have the needed taint, this should result in taint being applied, however we won't see patch
+	// to the annotations as maxUnavailable is set to 1.
+	node3.Spec.Taints = []corev1.Taint{*constants.NodeUpdateInProgressTaint}
 	nodes := []*corev1.Node{
+		// Since node-0 is at desired config, it shouldn't be tainted.
 		newNodeWithLabel("node-0", "v1", "v1", map[string]string{"node-role/worker": "", "node-role/infra": ""}),
 		newNodeWithLabel("node-1", "v0", "v0", map[string]string{"node-role/worker": "", "node-role/infra": ""}),
+		// This node is at desiredConfig and has taint, no patch calls
+		node2,
+		// This node is not at desiredConfig and has taint, get call
+		node3,
 	}
 
 	f.ccLister = append(f.ccLister, cc)
@@ -801,10 +815,10 @@ func TestShouldMakeProgress(t *testing.T) {
 	for idx := range nodes {
 		f.kubeobjects = append(f.kubeobjects, nodes[idx])
 	}
-
+	// First patch to apply taint
 	f.expectGetNodeAction(nodes[1])
 	expNode := nodes[1].DeepCopy()
-	expNode.Annotations[daemonconsts.DesiredMachineConfigAnnotationKey] = "v1"
+	expNode.Spec.Taints = append(expNode.Spec.Taints, *constants.NodeUpdateInProgressTaint)
 	oldData, err := json.Marshal(nodes[1])
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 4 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -973,6 +988,47 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 	return nil
 }
 
+// setUpdateInPrgressTaint applies in progress taint to all the nodes that are to be updated.
```
wking, line 991: nit: `setUpdateInPrgressTaint` -> `setUpdateInProgressTaint`

pkg/controller/node/node_controller.go
```diff
@@ -963,6 +974,10 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 		if err := ctrl.setDesiredMachineConfigAnnotation(node.Name, targetConfig); err != nil {
 			return goerrs.Wrapf(err, "setting desired config for node %s", node.Name)
 		}
+		ctrl.logPool(pool, "Apply UpdateInProgress taint %s target to %s", node.Name, targetConfig)
+		if err := ctrl.setUpdateInProgressTaint(node.Name); err != nil {
+			return goerrs.Wrapf(err, "failed applying updateinprogress taint for node %s", node.Name)
```
wking, line 979: possibly pull the key out of `NodeUpdateTaint.Key` here and a few lines above, so you don't have to worry about matching spelling/casing/typos.

pkg/controller/node/node_controller.go
```diff
@@ -963,6 +974,10 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 		if err := ctrl.setDesiredMachineConfigAnnotation(node.Name, targetConfig); err != nil {
 			return goerrs.Wrapf(err, "setting desired config for node %s", node.Name)
 		}
+		ctrl.logPool(pool, "Apply UpdateInProgress taint %s target to %s", node.Name, targetConfig)
```
kikisdeliveryservice, line 977:
```suggestion
		ctrl.logPool(pool, "Applying UpdateInProgress taint to %s",  node.Name)
```

pkg/controller/node/node_controller.go
```diff
@@ -488,6 +495,10 @@ func (ctrl *Controller) updateNode(old, cur interface{}) {
 			ctrl.logPoolNode(pool, curNode, "changed labels")
 			changed = true
 		}
+		if !reflect.DeepEqual(oldNode.Spec.Taints, curNode.Spec.Taints) {
```
kikisdeliveryservice, line 498: prob need a nil check here?

kikisdeliveryservice (commented):
quick pass, generally makes sense. will come back tomorrow to finish looking.

pkg/controller/node/node_controller.go
```diff
@@ -973,6 +988,47 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
...
+		if err != nil {
+			return err
+		}
+		oldData, err := json.Marshal(oldNode)
+		if err != nil {
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
```
kikisdeliveryservice, line 1007: should this be nil or len(newNode.Spec.Taints) == 0 

(goes back to my q about whether this will always be nil or if it can be empty)

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1294,67 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+		if err != nil {
+			return err
+		}
+		oldData, err := json.Marshal(oldNode)
+		if err != nil {
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
```
kikisdeliveryservice, line 1332: len(newNode.Spec.Taints) == 0 ? same q as above

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1294,67 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+	}
+
 	dn.logSystem("Update completed for config %s and node has been successfully uncordoned", desiredConfigName)
 	dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "Uncordon", fmt.Sprintf("Update completed for config %s and node has been uncordoned", desiredConfigName))
 
 	return nil
 }
 
+func (dn *Daemon) removeUpdateInProgressTaint() error {
+	// TODO: Move these 2 vars to common location to be shared with
```
kikisdeliveryservice, line 1308: nice <3

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1294,67 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+		oldData, err := json.Marshal(oldNode)
+		if err != nil {
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
+			newNode.Spec.Taints = []corev1.Taint{}
+		}
+
```
kikisdeliveryservice, line 1335: I think I'm tired but can we get a comment here on this loop thru append are doing bc future me will also be tired and easily confused ;)

kikisdeliveryservice:
Just for ref:

```
// Like TaintEffectNoSchedule, but the scheduler tries not to schedule
	// new pods onto the node, rather than prohibiting new pods from scheduling
	// onto the node entirely. Enforced by the scheduler.
TaintEffectPreferNoSchedule TaintEffect = "PreferNoSchedule"
``` 

https://github.com/kubernetes/kubernetes/blob/f0b7ad3ee06c5168fef5fa4f01fe445ece595f89/pkg/apis/core/types.go#L2684

pkg/controller/node/node_controller.go
```diff
@@ -65,11 +65,18 @@ const (
...
-	Duration: 100 * time.Millisecond,
-	Jitter:   1.0,
-}
+var (
+	nodeUpdateBackoff = wait.Backoff{
+		Steps:    5,
+		Duration: 100 * time.Millisecond,
+		Jitter:   1.0,
+	}
+	// NodeUpdateTaint is a taint applied by MCC when the update of node starts.
```
kikisdeliveryservice, line 74: nit: when you consolidate these vars elsewhere, maybe add the "why" of this to the comment?

sinnykumari:
Once this is ready, can we please add better description of why we are adding this taint in MCO. Also, will be nice to have details in commit message.

Also since feature freeze is on Friday, from the slack conversation I believe we leaned towards implementing this in 4.10.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Sure. Do you have a location where I can share code across MCO and MCC?

pkg/controller/node/node_controller.go
ravisantoshgudimetla: Why would one be non-nil empty and other be nil but I am if it is confusing, I can have `len()` checks as well.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: It depends on declaration. In this case, we have a slice but no backing array so nil is ok but there is nothing wrong in checking len as it works for both nil and empty slices. For example,

var a []string is a nil slice as it doesn't have underlying array
c := []string{} is an empty slice. 

Usually all the API objects returned will be nil slices as it encoding in JSON will be easier.

pkg/daemon/daemon.go
ravisantoshgudimetla: I just want to remove a taint from the node. I can add a comment. Technically, I can convert this to set so that deletion will be easier but I don't think the current implementation is going to be that expensive considering the number of taints we have.

ravisantoshgudimetla (commented):
Thank you for the reviews @sinnykumari @kikisdeliveryservice @wking 

> Once this is ready, can we please add better description of why we are adding this taint in MCO. Also, will be nice to have details in commit message.

Of course, I opened this to see if I am in the right direction or not.

>Also since feature freeze is on Friday, from the slack conversation I believe we leaned towards implementing this in 4.10.

I am fine with 4.10 implementation but I am wondering if this is too much of invasive change?

pkg/daemon/daemon.go
kikisdeliveryservice: I think it's fine to leave implementation as-is now, just add comment :+1: 

we can refactor if necessary later

pkg/controller/node/node_controller.go
kikisdeliveryservice: i think we could maybe use constants.go?

https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/pkg/constants/constants.go

pkg/daemon/daemon.go
```diff
@@ -8,6 +8,10 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
kikisdeliveryservice, line 11: failing verify here: 

`pkg/daemon/daemon.go:11: File is not `goimports`-ed (goimports)`

pkg/controller/node/node_controller.go
soltysh: I doubt spec will ever be nil at this point in time, as for `Taints` `DeepEqual` handles nil case, for you.

pkg/controller/node/node_controller.go
```diff
@@ -65,11 +65,18 @@ const (
...
+	NodeUpdateTaint = &corev1.Taint{
```
soltysh, line 75: Nit: `NodeUpdateInProgressTaint`?

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1294,67 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+		}
+
+		taintIndex := 0
+		for i, taint := range newNode.Spec.Taints {
+			if taint.MatchTaint(NodeUpdateTaint) {
+				taintIndex = i
+			}
+		}
+
+		newNode.Spec.Taints = append(newNode.Spec.Taints[:taintIndex], newNode.Spec.Taints[taintIndex+1:]...)
```
soltysh, line 1343: Alternatively, we can just iterate over taints leaving out the one we want to remove and then assign that. The complexity is similar, a bit more expensive memory wise, but easier to read, imo.

sinnykumari:
> I am fine with 4.10 implementation but I am wondering if this is too much of invasive change?

I understand this change is not much invasive. Main hesitation is because of unplanned changes that get added in repo without having full context and time crunch makes things worse. With limited team capacity and already other planned work and unplanned bugs, it becomes a bit difficult. Having some time buffer allows to think and review code better for future maintenance. 

Also whatever code we write, we all do our best to test and review and think that this shouldn't break existing fetaures but we all know what happens in reality ;)

kikisdeliveryservice:
> > I am fine with 4.10 implementation but I am wondering if this is too much of invasive change?
> 
> I understand this change is not much invasive. Main hesitation is because of unplanned changes that get added in repo without having full context and time crunch makes things worse. With limited team capacity and already other planned work and unplanned bugs, it becomes a bit difficult. Having some time buffer allows to think and review code better for future maintenance.
> 
> Also whatever code we write, we all do our best to test and review and think that this shouldn't break existing fetaures but we all know what happens in reality ;)

Agree with Sinny. next release is the better time frame. FTR I don't think this is a PR quality issue and more of a review/testing/discussion issue and I don't see a reason we can't work together to get this in next time (on the earlier side) pending those convos and testing/review/ :smile:

pkg/controller/node/node_controller.go
ravisantoshgudimetla: That is correct. Technically, the code above is right. I believe @kikisdeliveryservice's concern is more on the lines of readability.

ravisantoshgudimetla:
https://github.com/kubernetes/kubernetes/pull/104251 merged upstream, once https://github.com/openshift/kubernetes/pull/1076 lands, we should be able to use beta3 version which has increased weights

kikisdeliveryservice:
As per call, we'd also like an OTA approver on this PR along with a test verifying functionality before merging.

/assign @wking

pkg/daemon/daemon.go
kikisdeliveryservice: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2686/pull-ci-openshift-machine-config-operator-master-verify/1467926127505838080

kikisdeliveryservice:
@ravisantoshgudimetla can you ping me for review once you've addressed the outstanding comments?

pkg/controller/node/node_controller.go
```diff
@@ -65,11 +66,13 @@ const (
 	masterPoolName = "master"
 )
 
-var nodeUpdateBackoff = wait.Backoff{
-	Steps:    5,
-	Duration: 100 * time.Millisecond,
-	Jitter:   1.0,
-}
+var (
+	nodeUpdateBackoff = wait.Backoff{
```
wking: you have a public `NodeUpdateBackoff` in `pkg/constants` now.  Can you drop this one?

pkg/controller/node/node_controller.go
```diff
@@ -963,6 +970,10 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 		if err := ctrl.setDesiredMachineConfigAnnotation(node.Name, targetConfig); err != nil {
 			return goerrs.Wrapf(err, "setting desired config for node %s", node.Name)
 		}
+		ctrl.logPool(pool, "Applying %s taint to %s", constants.NodeUpdateInProgressTaint.Key, node.Name)
+		if err := ctrl.setUpdateInProgressTaint(node.Name); err != nil {
+			return goerrs.Wrapf(err, "failed applying %s taint for node %s", constants.NodeUpdateInProgressTaint.Key, node.Name)
```
wking, line 967: nit: the wrapper and log are both making some assumptions about what `setUpdateInProgressTaint` is up to.  Probably pretty safe assumptions, so feel free to ignore, but it would be slightly more robust to push those down into `setUpdateInProgressTaint` itself.

pkg/controller/node/node_controller.go
wking: fixed as of  652de25.

pkg/controller/node/node_controller.go
wking: and it looks like you do drop it in 652de252.  Maybe reshuffle your hunks between commits or squash down to one to avoid touching something in one commit and then removing it in the second?

pkg/controller/node/node_controller.go
```diff
@@ -973,6 +984,47 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 	return nil
 }
 
+// setUpdateInPrgressTaint applies in progress taint to all the nodes that are to be updated.
+// The taint on the individual node is removed by MCD once the update of the node is complete.
+// This is to ensure that the updated nodes are being preferred to non-updated nodes there by
+// reducing the number of reschedules.
+func (ctrl *Controller) setUpdateInProgressTaint(nodeName string) error {
+	return clientretry.RetryOnConflict(constants.NodeUpdateBackoff, func() error {
+		oldNode, err := ctrl.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
wking, line 985: nit: can we add a context argument to `setUpdateInProgressTaint` to push this `context.TODO` up a level?  No need to plumb `Context` through legacy call stacks, but new functions that need a context should take a context argument to avoid growing even more tech debt ;).

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1295,55 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
+			newNode.Spec.Taints = []corev1.Taint{}
+		}
+
+		// Get the index of NodeUpdateInProgressTaint.
+		taintIndex := 0
```
wking, line 1325: nit: you can avoid a no-op patch by initializing to `-1` and then `return nil` if you're still `-1` after iterating through looking for a match.

wking (approved):
I've gone through this again and left a few local-scoped nits.  But nothing update-specific; everything in here is good on that front.  So feel free to address as many or few of my nits as you like for me, and once the MCO folks are happy, I have no problems with this landing however it ends up.

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1295,55 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+		for i, taint := range newNode.Spec.Taints {
+			if taint.MatchTaint(commonconstants.NodeUpdateInProgressTaint) {
+				taintIndex = i
+			}
+		}
+		// Remove the NodeUpdateInProgressTaint.
+		newNode.Spec.Taints = append(newNode.Spec.Taints[:taintIndex], newNode.Spec.Taints[taintIndex+1:]...)
```
yuqi-zhang, line 1332: If `newNode.Spec.Taints = []corev1.Taint{}` still is the case here, wouldn't this panic?

wking: [this][1] would be one way to avoid that panic.

[1]: https://github.com/openshift/machine-config-operator/pull/2686#discussion_r764383642

yuqi-zhang: Ah right. Alternatively if the taint is not found, we could probably return early above.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: I thought, we usually error or log but not both in fn or method. IIUC, you want log to be present in `setUpdateInProgressTaint` method?

pkg/daemon/daemon.go
```diff
@@ -1290,12 +1295,59 @@ func (dn *Daemon) completeUpdate(desiredConfigName string) error {
...
+			return err
+		}
+
+		newNode := oldNode.DeepCopy()
+		if newNode.Spec.Taints == nil {
+			newNode.Spec.Taints = []corev1.Taint{}
+		}
+
+		// Get the index of NodeUpdateInProgressTaint.
+		taintIndex := -1
```
kikisdeliveryservice, line 1326: instead of doing it this way maybe doing it the way soltyish suggests below is more readable/less confusing? this block feels a little hard to understand on a run thru

kikisdeliveryservice: loop thru array, appending everything that isn't a taint, if len(clean_array) != len(original array) assign it to newNode.Spec.Taints else return nil

pkg/controller/node/node_controller.go
kikisdeliveryservice: what ravi did here seems consistent with the func call above

kikisdeliveryservice: tho @wking is there a specific change to the error message above you think would be better?

pkg/controller/node/node_controller.go
```diff
@@ -963,6 +963,10 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
 		if err := ctrl.setDesiredMachineConfigAnnotation(node.Name, targetConfig); err != nil {
 			return goerrs.Wrapf(err, "setting desired config for node %s", node.Name)
 		}
+		ctrl.logPool(pool, "Applying %s taint to %s", constants.NodeUpdateInProgressTaint.Key, node.Name)
```
sinnykumari, line 966: s/Applying/applying/

pkg/controller/node/node_controller.go
```diff
@@ -963,6 +963,10 @@ func (ctrl *Controller) updateCandidateMachines(pool *mcfgv1.MachineConfigPool,
...
+		if err := ctrl.setUpdateInProgressTaint(ctx, node.Name); err != nil {
```
sinnykumari, line 967: Is the goal is to apply UpdateInProgress taint only to nodes that are currently being updated?
I may have misunderstood but I thought we want to apply UpdateInProgress taint to all the nodes in the pool that has not been updated yet. And scheduler will make use of this taint to ensure that we try to rescheduled drained pods to node that doesn't have that taint(implies node has already been updated).

sinnykumari (commented):
Thanks Ravi for updating the PR.
Added few inline comments. Additionally, an you update description of the PR and commit message with details on what this change is about. This will help anyone who is freshly looking at the PR.

pkg/controller/node/node_controller.go
ravisantoshgudimetla: I thought:

All the nodes in the pool that have not been updated yet == All the nodes that doesn't have desired configuration? Is it not the case?

pkg/controller/node/node_controller.go
ravisantoshgudimetla: I thought it is ok to have capital letters for log whereas it is recommended to not have error string starting with capitals.

pkg/controller/node/node_controller.go
sinnykumari: Desired Configuration are not set on all node together when a new update starts. Number of nodes where we set desiredConfig is based on maxUnavailable value set. So, by default it will be set only on one node that has been picked up for update. See https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/node/node_controller.go#L770, we calculate number of candidates that can be updated before calling updateCandidateMachines()

sinnykumari:
Thanks for updating the commit message and description.

sinnykumari (approved):
Overall LGTM. I don't see any major review comment left from others that can change behavior of this feature. Let's get this merged and if any minor concerns needs to get fixed, it can be fixed as follow-up PR.

Also, with Trevor's approval we should be good to go from upgrade team side.

/lgtm

pkg/controller/node/node_controller_test.go
```diff
@@ -3,6 +3,7 @@ package node
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/machine-config-operator/pkg/constants"
```
ravisantoshgudimetla, line 6: oops ordering needs to be sorted out.

PR #2688, opened 2021-07-22
Support `ssh-key-dir` on OKD/FCOS

LorbusChris:
**- What I did**
In older versions of OKD, the SSH keys were written to
`/home/core/.ssh/authorized_keys`. Newer versions of OKD will
expect the keys at `/home/core/.ssh/authorized_keys.d/ignition`.
In the case of a newer version of OKD, the keys file is removed
from the legacy path in the updateSSHKeys function.
It will then be recreated at the new fragment path by the
atomicallyWriteSSHKey function that is called right after.

**- How to verify it**
CI

Support ssh-key-dir on OKD/FCOS

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1695,7 +1695,12 @@ func (dn *Daemon) atomicallyWriteSSHKey(keys string) error {
 		return err
 	}
 
-	authKeyPath := filepath.Join(coreUserSSHPath, "authorized_keys")
+	var authKeyPath string
+	if dn.os.IsFCOS() {
+		authKeyPath = filepath.Join(coreUserSSHPath, "authorized_keys.d", "ignition")
+	} else {
+		authKeyPath = filepath.Join(coreUserSSHPath, "authorized_keys")
+	}
 
 	// Keys should only be written to "/home/core/.ssh"
 	// Once Users are supported fully this should be writing to PasswdUser.HomeDir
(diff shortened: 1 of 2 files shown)


fortinj66:
@LorbusChris, https://github.com/openshift/machine-config-operator/pull/2087 will need to be reverted for this to work correct?

LorbusChris:
@fortinj66 yes. The plan is to manage the `ignition` key fragment in the future.

travier:
The order in which we will merge things here will matter unfortunately as this involves Ignition package changes too. One option we have is to make sure we coordinate this to happen on a specific day.

travier (approved):
Code LGTM but I'm not really familiar with the MCO code base.

LorbusChris:
Rebased, dropped one commit from this PR that was already implemented in https://github.com/openshift/machine-config-operator/pull/2813.

Maybe it makes sense to also drop the first commit from this PR (`daemon: Add IsRHCOS9() function`) until that is actually testable. Otherwise should be good to go. Please take a look @bgilbert.

pkg/daemon/osrelease.go
```diff
@@ -22,7 +25,16 @@ func (os OperatingSystem) IsRHCOS() bool {
 	return os.ID == "rhcos"
 }
 
-// IsFCOS is true if the OS is RHEL CoreOS
+// IsRHCOS9 is true if the OS is RHEL CoreOS 9
+func (os OperatingSystem) IsRHCOS9() bool {
```
sinnykumari, line 29: Do we know if we have infra to test this today in RHCOS9? Maybe we can track this in a card and implement along with test once we have OCP based on RHCOS 9?

LorbusChris: Not sure, but I agree that it'd be better to hold off on the RHCOS9 part until that's testable. For OKD it'd be good to get this in already, though.

LorbusChris:
I've removed the RHCOS9 parts from this PR

pkg/daemon/update.go
```diff
@@ -1726,6 +1731,57 @@ func (dn *Daemon) updateSSHKeys(newUsers []ign3types.PasswdUser) error {
 		}
 	}
 	if !dn.mock {
+		// On RHCOS8 and in older versions of OKD, the keys are written to `/home/core/.ssh/authorized_keys`.
```
sinnykumari, line 1734: Until we add support for RHCOS 9, let's keep the comment specific to FCOS/OKD

pkg/daemon/update.go
```diff
@@ -1726,6 +1731,57 @@ func (dn *Daemon) updateSSHKeys(newUsers []ign3types.PasswdUser) error {
...
+		// Remove the keys file from the old path here. It will be recreated at the correct path by the
+		// atomicallyWriteSSHKey function called right after.
+		authKeyPath := filepath.Join(coreUserSSHPath, "authorized_keys")
+		authKeyFragmentDirPath := filepath.Join(coreUserSSHPath, "authorized_keys.d")
+
+		if dn.os.IsFCOS() {
+			// Check if the authorized_keys file exists. If it does, remove it.
+			_, err := os.Stat(authKeyPath)
+			if err == nil {
+				err := os.RemoveAll(authKeyPath)
```
sinnykumari, line 1745: Since this is a file, I think just calling for Remove() should do.

Also, no strong opinion but I was wondering should we delete here entire file authorized_keys or delete ssh keys inside it that are managed by MCO? It could be possible that user have other keys added that without leveraging MCO.

pkg/daemon/update.go
```diff
@@ -1726,6 +1731,57 @@ func (dn *Daemon) updateSSHKeys(newUsers []ign3types.PasswdUser) error {
...
+						if err != nil {
+							return fmt.Errorf("failed to remove path '%s': %v", keyPath, err)
+						}
+					}
+				}
+			} else if !os.IsNotExist(err) {
+				// This shouldn't ever happen
+				return fmt.Errorf("unexpectedly failed to get info for path '%s': %v", authKeyFragmentDirPath, err)
+			}
+		} else {
```
sinnykumari, line 1770: Since we are not handling RHCOS 9, we can skip else section.

pkg/daemon/update.go
LorbusChris: I was thinking about this, too, but I deemed it safer to use RemoveAll() just in case somebody had unexpectedly added a dir with that name here. Happy to change this though.

Regarding additional keys, is that supported today though? i.e. wouldn't those keys be overwritten by the MCD via `atomicallyWriteSSHKey` on update anyway?

pkg/daemon/update.go
sinnykumari: umm, or this could be needed like you said during downgrade but we haven't started thinking about this move yet. I haven't tested myself but do you know if it can cause undesired behavior if both /home/core/.ssh/authorized_keys.d/ignition and /home/core/.ssh/authorized_keys are present on a node?

pkg/daemon/update.go
sinnykumari: good point, MCO will overwrite the content during any node update.

sinnykumari (approved):
Thanks Christain for making the changes.
/lgtm

sinnykumari:
Just noticed https://github.com/openshift/machine-config-operator/pull/2688#issuecomment-932160648 , @LorbusChris do we need any ignition update here? Putting hold until this question is addressed.
/hold

LorbusChris:
> Just noticed [#2688 (comment)](https://github.com/openshift/machine-config-operator/pull/2688#issuecomment-932160648) , @LorbusChris do we need any ignition update here? Putting hold until this question is addressed. /hold

No, this will only be needed for RHCOS9

PR #2689, opened 2021-07-22
Add parsing for registries.conf wildcard entries

umohnani8:

**- What I did**
The registries.conf file now supports wildcard entries.
Add validation for possible entries for insecure, blocked,
and allowed and vendor in updated runtime-utils for handling
the wildcard entries correctly.
Add some more test cases for wildcard entries.

**- How to verify it**
Add wildcard entries such as "*.foo.com" to the cluster wide Image CR for insecure, blocked, or allowed registries. When matching certificates, if we are pulling from "example.foo.com", it will match the settings for "*.foo.com" to it. So users don't need to have a long list of entries in their registries.conf file if they all match to a common dns name.

Add support for registries.conf wildcard entries.


umohnani8:
@kikisdeliveryservice PTAL

kikisdeliveryservice:
@umohnani8 is this... the minimum set of changes for the re-vendoring? Because I'm kind of concerned at updating& adding 1200 files.

umohnani8:
@kikisdeliveryservice yup, I just had to update two repos and go mod pulled in everything else on its own based on that.

sinnykumari:
We are now also parsing /etc/containers/registry.conf to determine whether a change made is safe to skip node drain or not https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/drain.go#L166 , especially, for icsp related changes. Do you think this PR may require some update in our logic?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -8,11 +8,12 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/docker/reference"
-	"github.com/containers/image/pkg/sysregistriesv2"
 	signature "github.com/containers/image/signature"
```
mtrmac, line 15: Now that runtime-utils uses v5, please update this to image/v5/ as well, incl. `helpers_test.go`; that should make the v1 go away and decrease the code size.

mtrmac:
> @umohnani8 is this... the minimum set of changes for the re-vendoring? Because I'm kind of concerned at updating& adding 1200 files.

A lot of that is a problematic dependency chain that brings the full containers/storage driver set, via https://github.com/containers/image/issues/1146 . It’s not fundamentally necessary, but I’m also not sure we can fix it today. I can take a stab at it if it is important.

mtrmac:
> We are now also parsing /etc/containers/registries.conf to determine whether a change made is safe to skip node drain or not https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/drain.go#L166 , especially, for icsp related changes. Do you think this PR may require some update in our logic?

Is this supposed to work only on MCO-generated data, or in general, including user-provided ignition files?

If the latter, oh boy, that looks rather brittle, possibly breaking any time a new option is added to the config file.
I would strongly recommend at least defaulting to drain on any completely unrecognized config file entries, like https://github.com/containers/image/issues/1284 . And for config file entries known to `sysregistriesv2` but not to MCO (right now, `CredentialHelpers`+the short-name configuration, probably rework this to work via a single top-level deep equality comparison, only editing out known-acceptable differences.

---

Either way, the code that uses `reg.Location` for a lookup key, and compares entries for unexpected changes, must prefer `reg.Prefix` if it set; runtime-utils now generates configuration like that.

sinnykumari:
> Is this supposed to work only on MCO-generated data, or in general, including user-provided ignition files?

This doesn't come into picture for initial boot. MCO is parsing and determining the change performed on cluster as a day2 operation if registries.conf content changed between old and newly generated rendered-config.

> I would strongly recommend at least defaulting to drain on any completely unrecognized config file entries, like 
[containers/image#1284](https://github.com/containers/image/issues/1284) . And for config file entries known to `sysregistriesv2` but not to MCO (right now, `CredentialHelpers`+the short-name configuration, probably rework this to work via a single top-level deep equality comparison, only editing out known-acceptable differences.

+1
I agree that current parsing is quite delicate for getting new field being introduced. Tried best to handle all available structs and types I got after parsing the toml. Perhaps, need to revisit the implementation.

mtrmac:
> Either way, the code that uses `reg.Location` for a lookup key, and compares entries for unexpected changes, must prefer `reg.Prefix` if it set; runtime-utils now generates configuration like that.

To be very explicit, I think this is a blocker for merging.

pkg/controller/container-runtime-config/helpers.go

umohnani8:
@mtrmac I accounted for the Prefix if it has a value https://github.com/openshift/machine-config-operator/pull/2689/commits/3e408f1eca5b0b38b59c8fe07f29a4a343816d25#diff-58db213cf6f6259a621f8d3421b5e8af55ed71c9f52c15b255f66f0b1e78a946R226, PTAL

pkg/daemon/drain.go
```diff
@@ -223,12 +223,20 @@ func isSafeContainerRegistryConfChanges(oldConfig, newConfig *mcfgv1.MachineConf
...
 	}
 
 	newRegHashMap := make(map[string]sysregistriesv2.Registry)
 	for _, reg := range tomlConfNewReg.Registries {
-		newRegHashMap[reg.Location] = reg
+		scope := reg.Location
+		if reg.Prefix != "" {
+			scope = reg.Prefix
+		}
+		newRegHashMap[scope] = reg
```
mtrmac, line 239: A bit below, I think the `reflect.DeepEqual` should also verify `Location` is unchanged (that used to be implied by using `Location` as the map key, but isn’t anymore).

mtrmac (commented):
ACK, one more location.

mtrmac:
> > Is this supposed to work only on MCO-generated data, or in general, including user-provided ignition files?
> 
> This doesn't come into picture for initial boot. MCO is parsing and determining the change performed on cluster as a day2 operation if registries.conf content changed between old and newly generated rendered-config.

Yes, but users can add new MachineConfig resources that contain ignition content, overriding what the MCO runtime-config mechanism creates, can’t they? (I don’t know, that might be unsupported, but IIRC some people are sharing such recipes.)

pkg/daemon/drain.go
umohnani8: fixed

mtrmac (commented):
/lgtm

Thanks!

umohnani8:
@sinnykumari @kikisdeliveryservice can I please get an approve here :)

sinnykumari:
> > > Is this supposed to work only on MCO-generated data, or in general, including user-provided ignition files?
> > 
> > This doesn't come into picture for initial boot. MCO is parsing and determining the change performed on cluster as a day2 operation if registries.conf content changed between old and newly generated rendered-config.
> 
> Yes, but users can add new MachineConfig resources that contain ignition content, overriding what the MCO runtime-config mechanism creates, can’t they? (I don’t know, that might be unsupported, but IIRC some people are sharing such recipes.)

Possible but are they supported? I don't have full picture. I thought in OCP this would  ideally be managed only by MCO by applying ctrcfg. @umohnani8  can you confirm this?

If this is true, I would suggest to hold this PR for 4.10 where we can work on improving drain behaviour accordingly.

umohnani8:
Users can override the registries.conf setting by applying a MC that adds a drop-in file at `/etc/containers/registries.conf.d`. We don't recommend this method and always ask users to use the ICSP or Image CR for any registries.conf changes.
We don't support the MC way, but I believe that method has been shared in the past when the user absolutely had to use it. I don't think we need to hold this PR as it is not changing any of the fields already accepted by both ICSP and Image CRs, we are just allowing for wildcard entries to be recognized.

sinnykumari:
Thanks @umohnani8 for confirming! I am fine with this to go in as we don't see any known issues.
gcp-op test has failed, looks like it needs some fixing.

```
 INFO[2021-07-23T16:30:50Z] go test -failfast -timeout 90m -v${WHAT:+ -run="$WHAT"} ./test/e2e/
# pkg-config --cflags  -- devmapper
Package devmapper was not found in the pkg-config search path.
Perhaps you should add the directory containing `devmapper.pc'
to the PKG_CONFIG_PATH environment variable
No package 'devmapper' found
pkg-config: exit status 1 
```

umohnani8:
@sinnykumari the test should be fixed now. Can I please get an approve and lgtm here

sinnykumari:
Yeah, looks like test is fixed now. Since, later changes made was only in Makefile to fix gcp-op test and  @mtrmac has already added lgtm for container runtime specific changes. Should be good to get merged.

/lgtm
/approve

mtrmac:
> @umohnani8 is this... the minimum set of changes for the re-vendoring? Because I'm kind of concerned at updating& adding 1200 files.

https://github.com/containers/storage/pull/972 + https://github.com/containers/image/pull/1312 will reduce the effect, with that this PR would count 749 modified files (and a lot of that is dependency upgrades, not just net-new additions). We should go further than that to exclude the compression implementations entirely, but that part of https://github.com/containers/image/issues/1146 (and its impact on the stable API) looks much more difficult.

mtrmac:
> > @umohnani8 is this... the minimum set of changes for the re-vendoring? Because I'm kind of concerned at updating& adding 1200 files.
> 
> [containers/storage#972](https://github.com/containers/storage/pull/972) + [containers/image#1312](https://github.com/containers/image/pull/1312) will reduce the effect

#2695 .

PR #2690, opened 2021-07-23
Bug 1970129: Set ovs syslog level to info

npinaeva:
This bug was previously fixed in a different way: https://github.com/openshift/cluster-network-operator/pull/1142
And then reverted: https://github.com/openshift/cluster-network-operator/pull/1163

**- What I did**
Set ovs syslog level to info (default is err), to collect info level logs with must-gather since it uses systemd journal.
ovs-vswitchd service runs `ovs-ctl` to start and reload service (https://github.com/openvswitch/ovs/blob/master/rhel/usr_lib_systemd_system_ovs-vswitchd.service.in#L23).
`ovs-ctl` doesn't let you set log level on start (https://github.com/openvswitch/ovs/blob/master/utilities/ovs-ctl.in#L214)
So, I added `ExecStartPost` to the service which sets right log level on start and restart.
There is also `reload` operation defined for the service (https://github.com/openvswitch/ovs/blob/master/rhel/usr_lib_systemd_system_ovs-vswitchd.service.in#L28), but `ExecStartPost` is not executed on reload, so I added `ExecReload` command which is executed after the main service `ExecReload`, therefore ovs has right log settings after reload also.

**- How to verify it**
Check logs and ovs log level on a cluster, then create new MachineConfig object that will update all your nodes, then check logs and log level again.

Check logs: 
- run `oc adm must-gather -- gather_network_logs` and see log file from `host_service_logs/<node_name>/ovs-vswitchs_service.log`, find (or check there are None at the beginning) INFO level logs there
- get access to node binaries:  run `oc debug --image fedora:31 node/<node name>` then `chroot /host` 
- check log levels on host: after previous step run `ovs-appctl vlog/list`, check syslog column is set to ERR at the beginning and at the end is set to INFO

Create new MachineConfig object (from https://github.com/openshift/machine-config-operator#applying-configuration-changes-to-the-cluster):
Create MachineConfig (two copies for worker and master, change `machineconfiguration.openshift.io/role` accordingly):
```
kind: MachineConfig
metadata:
  labels:
  name: 50-change-ovs-loglevel
spec:
  config:
    ignition:
      version: 3.2.0
    systemd:
      units:
      - dropins:
        - contents: |
            [Service]
            Restart=always
            ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /var/lib/openvswitch'
            ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /etc/openvswitch'
            ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /run/openvswitch'
            ExecStartPost=-/usr/bin/ovs-appctl vlog/set syslog:info
            ExecReload=-/usr/bin/ovs-appctl vlog/set syslog:info
          name: 10-ovs-vswitchd-restart.conf
        name: ovs-vswitchd.service
```
Then run `oc create -f <machine config file name>` for both files to create MachineConfig that corresponds presented changes.
Then use `oc describe machineconfigpool/worker(master)` to watch the progress of the rollout. You should see that new config successfully generated, then the nodes will get updated, e.g.
```
  Normal  RenderedConfigGenerated  62m   machineconfigcontroller-rendercontroller  rendered-worker-eac6cd3e114c5c3d1be4db3005827d08 successfully generated
  Normal  SetDesiredConfig         62m   machineconfigcontroller-nodecontroller    Targeted node ci-ln-7xlpli2-f76d1-gwkvq-worker-a-r4864 to config rendered-worker-eac6cd3e114c5c3d1be4db3005827d08
```
After all nodes are updated, check the logs once again the same way as it the beginning (it should now collect more log messages including WARN and INFO levels), also check `ovs-appctl vlog/list`

Now you can access host binaries as described in Check logs section above and check that log levels are stable after service restart/reload. Run 
- `systemctl restart ovs-vswitchd` and then 'ovs-appctl vlog/list'
- `systemctl reload ovs-vswitchd` and then 'ovs-appctl vlog/list'

Set ovs syslog level to info

### DIFF
--- a/templates/common/_base/units/ovs-vswitchd.service.yaml
+++ b/templates/common/_base/units/ovs-vswitchd.service.yaml
@@ -7,3 +7,5 @@ dropins:
       ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /var/lib/openvswitch'
       ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /etc/openvswitch'
       ExecStartPre=-/bin/sh -c '/usr/bin/chown -R :$${OVS_USER_ID##*:} /run/openvswitch'
+      ExecStartPost=-/usr/bin/ovs-appctl vlog/set syslog:info
+      ExecReload=-/usr/bin/ovs-appctl vlog/set syslog:info


abhat:
this will work, but there is $OPTIONS variable in the /etc/sysconfig/openvswitch which is probably where we should specify the logging options.

npinaeva:
> this will work, but there is $OPTIONS variable in the /etc/sysconfig/openvswitch which is probably where we should specify the logging options.

Thank you for suggestion, I tried to figure out the best way to set OPTIONS, but machine-config doesn't let you edit files and fully overwriting `/etc/sysconfig/openvswitch` file may be dangerous (since it stores OVS_USER_ID), so my best suggestion is to use sed on /etc/sysconfig/openvswitch during ExecStartPre.

Please let me know is you see a better way to specify OPTIONS

rbbratta:
The current `EnvironmentFiles` order seems to be:

```shell
# systemctl show  ovs-vswitchd.service | grep Environ
Environment=XDG_RUNTIME_DIR=/var/run/openvswitch
EnvironmentFiles=/etc/openvswitch/default.conf (ignore_errors=no)
EnvironmentFiles=/etc/sysconfig/openvswitch (ignore_errors=yes)
EnvironmentFiles=/run/openvswitch.useropts (ignore_errors=yes)
```

```shell
# systemctl cat ovs-vswitchd.service | grep Environ
Environment=XDG_RUNTIME_DIR=/var/run/openvswitch
EnvironmentFile=/etc/openvswitch/default.conf
EnvironmentFile=-/etc/sysconfig/openvswitch
EnvironmentFile=-/run/openvswitch.useropts
```

The last file has precedence.

It looks like we already do an `ExecStartPre` trick to set `OVS_USER_ID` for security in `ovsdb-server.service`

https://github.com/openvswitch/ovs/commit/27e25e18c1f4cdd789d5670ab9e01dcf02a86b6f#diff-e2b87a549c9d95310910486a3f416086306ced43c7eede2c67b6294ea4ac4d63

Will this be the last `$OPTIONS` override we do for OVS on OCP?  If so then OCP should probably take ownership of the last `EnvironmentFile` by adding our own.

Or it we want RHCOS to own OVS then maybe we can convince the `openvswitch` RHCOS RPM owners to change the default log level?

dcbw:
Would a systemd "dropin" unit file work better here?

`man systemd.unit`
See also /lib/systemd/system/ovn-northd.service for a possible example.

```
       Along with a unit file foo.service, a "drop-in" directory foo.service.d/ may exist. All files with the suffix ".conf" from this directory will be merged in the alphanumeric order and parsed after
       the main unit file itself has been parsed. This is useful to alter or add configuration settings for a unit, without having to modify unit files. Each drop-in file must contain appropriate section
       headers. For instantiated units, this logic will first look for the instance ".d/" subdirectory (e.g.  "foo@bar.service.d/") and read its ".conf" files, followed by the template ".d/" subdirectory
       (e.g.  "foo@.service.d/") and the ".conf" files there. Moreover for unit names containing dashes ("-"), the set of directories generated by repeatedly truncating the unit name after all dashes is
       searched too. Specifically, for a unit name foo-bar-baz.service not only the regular drop-in directory foo-bar-baz.service.d/ is searched but also both foo-bar-.service.d/ and foo-.service.d/. This
       is useful for defining common drop-ins for a set of related units, whose names begin with a common prefix. This scheme is particularly useful for mount, automount and slice units, whose systematic
       naming structure is built around dashes as component separators. Note that equally named drop-in files further down the prefix hierarchy override those further up, i.e.
       foo-bar-.service.d/10-override.conf overrides foo-.service.d/10-override.conf.

       In cases of unit aliases (described above), dropins for the aliased name and all aliases are loaded. In the example of default.target aliasing graphical.target, default.target.d/,
       default.target.wants/, default.target.requires/, graphical.target.d/, graphical.target.wants/, graphical.target.requires/ would all be read. For templates, dropins for the template, any template
       aliases, the template instance, and all alias instances are read. When just a specific template instance is aliased, then the dropins for the target template, the target template instance, and the
       alias template instance are read.

       In addition to /etc/systemd/system, the drop-in ".d/" directories for system services can be placed in /usr/lib/systemd/system or /run/systemd/system directories. Drop-in files in /etc/ take
       precedence over those in /run/ which in turn take precedence over those in /usr/lib/. Drop-in files under any of these directories take precedence over unit files wherever located. Multiple drop-in
       files with different names are applied in lexicographic order, regardless of which of the directories they reside in.
```

npinaeva:
@dcbw since this PR do change dropin file, and `EnvironmentFile` options override settings made with `Environment` how exactly do you suppose to use a dropin?

dcbw:
@npinaeva more about Ross' comment and Aniket's. Typically modifying the RPM-laid-down files isn't something we want to do, and systemd provides us a mechanism to use dropins. Whether that mechanism works the way we want, I don't know :) But if it does, it would be cleaner than modifying /etc/sysconfig/openvswitch directly.

squeed:
Annoyingly, it doesn't seem like there's a way to append to an existing environment variable via systemd units.

So, the easiest way to do this is to add an ExecStartPost:

```
ExecStartPost=-/usr/bin/ovs-appctl -t ovs-vswitchd vlog/set syslog:info
```

npinaeva:
That's funny, so you think I should revert it back? https://github.com/openshift/machine-config-operator/commit/c58663a44bd7ccc568716dac4fc7c7b85705ea99

squeed:
Hah. That is funny.

Because there isn’t a specific logging environment variable (as far as I can tell), I personally prefer the `appctl` to the `sed`. Idempotent and doesn’t touch /etc.

I fear we’re drifting off in to bike shedding, though. If it works, it works.

npinaeva:
Yeah, totally agree,
Btw my sed is idempotent! (sets --ovs-vswitchd-options only if there is no such substring in OPTIONS)

rbbratta:
is there a requirement for dev to enable `dbg` logging on boot?  If so maybe document a procedure to enable.

npinaeva:
> is there a requirement for dev to enable `dbg` logging on boot? If so maybe document a procedure to enable.

@abhat @dcbw @squeed does anyone have an answer to that? Please

trozet:
> > is there a requirement for dev to enable `dbg` logging on boot? If so maybe document a procedure to enable.
> 
> @abhat @dcbw @squeed does anyone have an answer to that? Please

we will need to create a knowledge base article to do something similar to kubelet: https://docs.openshift.com/container-platform/4.6/rest_api/editing-kubelet-log-level-verbosity.html

So that a user has a way to set on each node manually or via mco.

trozet (approved):
/lgtm

npinaeva:
I changed it back to `ovs-appctl` based on our discussion.
I will also work on adding doc section explaining how to change ovs log level manually

kikisdeliveryservice (approved):

PR #2699, opened 2021-07-31
Bug 1975626: Fix KubeletConfig nil error

QiWang19:
Add nil check for KubeletConfig to fix the nil pointer panic.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1975626

**- What I did**

**- How to verify it**

### DIFF
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -92,7 +92,7 @@ func getManagedKubeletConfigKeyDeprecated(pool *mcfgv1.MachineConfigPool) string
 
 // validates a KubeletConfig and returns an error if invalid
 func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
-	if cfg.Spec.KubeletConfig.Raw == nil {
+	if cfg.Spec.KubeletConfig == nil || cfg.Spec.KubeletConfig.Raw == nil {
 		return nil
 	}
 	kcDecoded, err := decodeKubeletConfig(cfg.Spec.KubeletConfig.Raw)


QiWang19:
@kikisdeliveryservice @sinnykumari PTAL.

sinnykumari:
Please get any kubelet related PR reviewed by node team first since node team owns it and they have better context.
/assign @rphillips 

Also, looks like this bug (BZ#1975626) depends on https://bugzilla.redhat.com/show_bug.cgi?id=1921248 which doesn't seem relevant to this PR. When was this fix introduced in 4.7 and 4.8?

QiWang19:
> Also, looks like this bug (BZ#1975626) depends on https://bugzilla.redhat.com/show_bug.cgi?id=1921248 which doesn't seem relevant to this PR. When was this fix introduced in 4.7 and 4.8?

The change is part of https://github.com/openshift/machine-config-operator/commit/1030b2c2fef5043c7611e8694c6c3081b7b6ea5f and introduced from 4.7.

sinnykumari:
> The change is part of [1030b2c](https://github.com/openshift/machine-config-operator/commit/1030b2c2fef5043c7611e8694c6c3081b7b6ea5f) and introduced from 4.7.

ok, since this change was already part of 4.7 and there is no corresponding 4.7 bz, you can check with your team lead and get direct valid bz label on this PR. Ideally https://bugzilla.redhat.com/show_bug.cgi?id=1921248 should be removed from depends on as it is not related.

QiWang19:
Needs approved label.

pkg/controller/kubelet-config/helpers.go
```diff
@@ -92,7 +92,7 @@ func getManagedKubeletConfigKeyDeprecated(pool *mcfgv1.MachineConfigPool) string
 
 // validates a KubeletConfig and returns an error if invalid
 func validateUserKubeletConfig(cfg *mcfgv1.KubeletConfig) error {
-	if cfg.Spec.KubeletConfig.Raw == nil {
+	if cfg.Spec.KubeletConfig == nil || cfg.Spec.KubeletConfig.Raw == nil {
```
yuqi-zhang, line 95: Curious, do we consider Spec.KubeletConfig mandatory at all? I assume no since e.g. a user should be able to create a kubeletconfig to just modify the logLevel via https://github.com/openshift/machine-config-operator/blob/master/install/0000_80_machine-config-operator_01_kubeletconfig.crd.yaml#L144 ?

QiWang19: @yuqi-zhang  No, it's not mandatory so the [Line96](https://github.com/openshift/machine-config-operator/pull/2699/files#diff-4a2d0a4442e10d889a927322ad7311c67b22b259622067b52cbda7d2ece3ebefR96) does not rise an error. But if `Spec.KubeletConfig` is nil the `cfg.Spec.KubeletConfig.Raw` will lead to nil pointer panic.

yuqi-zhang (approved):

ecordell:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

picked

PR #2703, opened 2021-08-04
Bug 1986757: Set timeoutSeconds for keepalived liveness probe

cybertron:
timeoutSeconds defaults to 1 second, but our liveness probe attempts
to wait for 5 seconds. This seems to be causing frequent liveness
probe timeouts in some environments which triggers unnecessary
restarts.

This patch sets timeoutSeconds to 5 to match the expectations of
the probe.

Fixed an issue with the keepalived liveness probe that could cause spurious timeouts.


mandre (commented):
/lgtm

cybertron:
All of the affected platforms are now passing. This is a pretty simple change that isn't platform-specific so I think it could probably go in (although it would be even better if someone from the remaining two platforms could chime in).

kikisdeliveryservice (approved):
Adding an approval and also a hold to be removed by @cybertron once he feels it's ready and has necessary approvals.

/hold

yboaron:
@cybertron  I think we can remove the hold and merge this PR

cybertron:
/hold cancel

Agreed, thanks.

cybertron:
The failing jobs all appear to be permafails and none of them even deploy the changed template. I think we just need to skip them.

PR #2716, opened 2021-08-16
Bug 1980844: bump SystemMemoryExceedsReservation alert threshold to 95%

harche:

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1980844

**- What I did**

**- How to verify it**

SystemMemoryExceedsReserved alert should fire at 95% utilization instead of existing 90%.

### DIFF
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -51,7 +51,7 @@ spec:
       rules:
         - alert: SystemMemoryExceedsReservation
           expr: |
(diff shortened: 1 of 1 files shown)


sinnykumari:
For future reference, can we add reason of the bump in commit message?
/assign @rphillips

harche:
> For future reference, can we add reason of the bump in commit message?
> /assign @rphillips

Thanks, I updated the commit message.

sinnykumari (approved):

wking:
I've filed #2722 to catch the `message` up with this change.

harche:
> I've filed #2722 to catch the `message` up with this change.

Thanks @wking

PR #2718, opened 2021-08-16
Bug 1994277: delete the memory manager state file before the kubelet start

cynepco3hahue:
Once you change the memory manager policy it requires removing the memory manager state file, otherwise, the kubelet will fail to restart. To avoid kubelet start failures during the change of the memory manager policy we can remove the state as `PreStart` condition under the systemd unit file.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

**- What I did**

**- How to verify it**

Delete the memory manager state file before the kubelet start

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -12,6 +12,7 @@ contents: |
   Type=notify
   ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
   ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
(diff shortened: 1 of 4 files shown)


kikisdeliveryservice:
@cynepco3hahue can you give us some background for this change?

cynepco3hahue:
> @cynepco3hahue can you give us some background for this change?

Added more information under the description, please let me know if it is good enough. In general, the memory manager state file validation is the same as for the CPU manager, so we need to do similar steps to be sure that it will not prevent the kubelet restart.

rphillips:
Is there a BZ?

cynepco3hahue:
@rphillips I attached the BZ.

wabouhamad (commented):
/lgtm

PR #2721, opened 2021-08-18
Bug 1955300: tighten operator availability conditions

kikisdeliveryservice:
The MCO available status should only be set to False when we believe that the operator/operands have problems/aren't working correctly.  An Operator can be degraded (such as in the case of a degraded master pool) but that does not mean that the operator itself is Unavailable and not functioning properly.

We are setting Available = False when we have Degraded = True across the board and by far this is being set when we have a problem with syncRequiredPools. Pool issues do not mean that the operator itself has a problem (if it did it wouldve failed the other syncs first). Instead, let syncAvailable() look at the results of the syncs itself and only set Available = False when the syncing of operands, etc.. fails (which occurs much less often). When those other syncs fail, we have good reason to believe something is fundamentally wrong with the MCO thus setting Available = False

Also fix Available message version. We are using the incoming version as Available at before the version is fully rolled out to the MCO. Instead we should be looking at the clusteroperator version and using that instead.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -100,15 +100,13 @@ func (optr *Operator) syncAvailableStatus() error {
 		return nil
 	}
 
-	optrVersion, _ := optr.vStore.Get("operator")
 	degraded := cov1helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorDegraded)
-	message := fmt.Sprintf("Cluster has deployed %s", optrVersion)
+	message := fmt.Sprintf("Cluster has deployed %s", co.Status.Versions)
 
 	available := configv1.ConditionTrue
 
 	if degraded {
 		available = configv1.ConditionFalse
-		message = fmt.Sprintf("Cluster not available for %s", optrVersion)
 		mcoObjectRef := &corev1.ObjectReference{
 			Kind:      co.Kind,
 			Name:      co.Name,
@@ -117,6 +115,7 @@ func (optr *Operator) syncAvailableStatus() error {
 		}
 
 		optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, "OperatorNotAvailable", message)
+		message = fmt.Sprintf("Cluster not available for %s", co.Status.Versions)
 	}
 
 	coStatus := configv1.ClusterOperatorStatusCondition{
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -91,7 +91,7 @@ func (optr *Operator) syncRelatedObjects() error {
 }
 
 // syncAvailableStatus applies the new condition to the mco's ClusterOperator object.
-func (optr *Operator) syncAvailableStatus() error {
+func (optr *Operator) syncAvailableStatus(ierr syncError) error {
 	co, err := optr.fetchClusterOperator()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
Tests passed but unfortunately must gather failed

`Step e2e-gcp-op-test succeeded after 1h4m50s.`
` Step e2e-gcp-op-gather-extra failed after 10s.`

/retest

pkg/operator/status.go
```diff
@@ -117,6 +116,7 @@ func (optr *Operator) syncAvailableStatus() error {
 		}
 
 		optr.eventRecorder.Eventf(mcoObjectRef, corev1.EventTypeWarning, "OperatorNotAvailable", message)
+		message = fmt.Sprintf("Cluster not available for %s", co.Status.Versions)
```
yuqi-zhang, line 119: Am I correct in understanding that, theoretically, this failure can happen in one of two cases:

1. the MCO is not performing an upgrade, but e.g. the user touched the daemonset definition and now it is having trouble rendering MCD definitions. We should report failing on current version
2. we are performing an upgrade, and the new incoming MCO is failing to sync on something (this is far more likely). 

In the second case, should we maybe distinguish that fact here somehow? i.e. technically the new version is being deployed, and we can no longer roll back to the old version anymore. But the new version is failing right now, not the old one.

yuqi-zhang (commented):
Overall lgtm, curious, does this put us on par with the Openshift operator availability definition? Or is there more than just the requiredPools available=false scenario?

One more question below:

pkg/operator/status.go
kikisdeliveryservice: This is just moving the same message we'd always log (if unavailable)  just moved down (from line 111) and logged less (not the reqd pools case). Mostly I'd like to merge this to to get some signal from CI and potentially iterate.

kikisdeliveryservice:
Would like to get this in to get some ci runs to further iterate. Can I please get a LGTM as it's a high priority BZ.

sinnykumari:
will try to review it by tomorrow unless Jerry gets to it before that.

pkg/operator/status.go
sinnykumari: Took a quick look at recommend way by CVO on how it should report different state of operator. Overall this looks fine, we will probbaly need to fix messaging according to doc https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md#conditions . What I understood from doc, if there is an error, then operator should report following:
```
If an error blocks reaching 4.0.1, the conditions might be:

    Degraded is true with a detailed message Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.
    Available is true with message Cluster has deployed 4.0.0
    Progressing is true with message Unable to apply 4.0.1: a required object is missing
```

This means if we are reporting available it would report existing version from which upgrade started.

Does `co.Status.Versions` target updated version or the existing version from which upgrade started?

sinnykumari:
We can do any followup work on messaging in subsequent PR about https://github.com/openshift/machine-config-operator/pull/2721#discussion_r710360036 . Considering priority, let's get it merged as this would help Kirsten to monitor further logs in CI and periodic runs.
/lgtm

pkg/operator/status.go
kikisdeliveryservice: Yeah I actually fixed that in the other commit: https://github.com/openshift/machine-config-operator/pull/2721/commits/28c305d9550946702f145ba061d6cec8fcf9502f

co.status.version is the actual version the MCO is on, not the incoming one

kikisdeliveryservice: I noticed we were doing it wrong before, but will also verify things once I get some periodic runs in. :smile_cat:

PR #2723, opened 2021-08-19
Bug 1995785: crio: complete crio default config

haircommander:
Instead of relying on /etc/crio/crio.conf (shipped in the cri-o rpm) for fields, we should import all relevant fields into MCO.
This will allow us to eventaully use MCO to remove /etc/crio/crio.conf (and have the rpm stop shipping it).

We do this because there exists a feature where rpm-ostree pins a file if it was ever managed by an RPM, and doesn't update it.
however, we want to be able to change crio configuration post-installation, and having a pinned crio config could interfere with that.

Finally, even though conmon path ("conmon" option) is set to the default, we include it here, to work around the aforementioned bug manifesting
in CRI-O failing to come up.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -4,12 +4,18 @@ contents:
   inline: |
     [crio]
     internal_wipe = true
+    storage_driver = "overlay"
+    storage_option = [
+        "overlay.override_kernel_check=1",
+    ]
 
     [crio.api]
     stream_address = ""
     stream_port = "10010"
 
     [crio.runtime]
+    selinux = true
+    conmon = ""
     conmon_cgroup = "pod"
     default_env = [
         "NSS_SDB_USE_CACHE=no",
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -4,12 +4,18 @@ contents:
   inline: |
     [crio]
     internal_wipe = true
+    storage_driver = "overlay"
+    storage_option = [
+        "overlay.override_kernel_check=1",
+    ]
 
     [crio.api]
     stream_address = ""
     stream_port = "10010"
 
     [crio.runtime]
+    selinux = true
+    conmon = ""
     conmon_cgroup = "pod"
     default_env = [
         "NSS_SDB_USE_CACHE=no",


haircommander:
relevant line is
```
./bin/%{service_name} \
      --selinux \ 
      --cgroup-manager "systemd" \
      --storage-driver "overlay" \
      --cni-plugin-dir "%{_libexecdir}/cni" \
      --storage-opt "overlay.override_kernel_check=1" \
      --metrics-port 9537 \
      --enable-metrics \
      config > %{service_name}.conf


```
this enables all not currently enabled items in that command

cgwalters:
For completeness this is related to https://bugzilla.redhat.com/show_bug.cgi?id=1993385#c3

mrunalp:
This looks fine but the commit message could use some more detail.

cgwalters:
I built a 4.9 image for this as `registry.ci.openshift.org/coreos/openshift49-bz1993385-followup`

cgwalters:
> We do this because there exists a feature where rpm-ostree pins a file if it was ever managed by an RPM, and doesn't update it.

Not worth a respin, but that's not quite correct.  It's more that all files in `/etc` are treated as `%config(noreplace)` by ostree.  And in this case, from ostree's PoV the file was "manually" modified (by the MCO), then the MCO stopped modifying it.  ostree doesn't know that though, so the file became "unmanaged state".

Or to rephrase, the problem is that the file was edited compared to the RPM default, not that it was "managed by an RPM".

haircommander:
updated the commit, ptal @cgwalters

mtrmac:
> Instead of allowing nodes to not come up, we populate the MCO template with all relevant fields in the cri-o rpm (including "conmon" option, even though it's being set to the default),

Do we need to think about how these “all relevant fields” will be kept up-to-date? I.e. is there a risk of CRI-O upstream, and/or the RPM maintainer, adjusting the default config file, and MCO not being updated?

Alternatively, would it be possible to somehow convince ostree to return `/etc/crio/crio.conf` to fully-managed state, coming from the RPM, and only have MCO ship its intentional overrides in `/etc/crio/crio.conf.d`?

Or are the relevant maintainers actually paying more attention to the MCO than to the upstream and/or RPM configuration, so this is not really a concern at all?

cgwalters:
> Alternatively, would it be possible to somehow convince ostree to return /etc/crio/crio.conf to fully-managed state, coming from the RPM, and only have MCO ship its intentional overrides in /etc/crio/crio.conf.d?

Absolutely, that's easy: we can change the MCO to `cp {/usr,}/etc/crio/crio.conf`.  However, this only works on ostree systems and unfortunately we need to consider traditional `yum` systems too at the moment, which don't have `/usr/etc`.  (IMO, this is one of the cooler features of ostree) 

This path though *is* aligned with the yum/rpm case of the semantics of `%config(noreplace)` - on a traditional yum system, reverting the file to the default will also return it to being managed by RPM.    There just isn't a convenient way to do it without redownloading the RPM.

There's code in the MCO which tries to simulate this with `.mcdorig` files, but I think that's buggy.

I think this is the second or third bug we've had of this form.  I think the most robust fix is: crio ships defaults in `/usr/lib/crio.conf` or just in the binary itself, and there is no `/etc/crio/crio.conf` at all by default.  The MCO or admins can write drop-ins in `/etc/crio.conf.d`, and that's it.

Also, related to this: when the MCO writes its drop in, it should only write values that actually changed, and not write the whole config file.

wking:
[upgrade failed][1] with:

```
[sig-network] pods should successfully create sandboxes by not timing out
```

That's [not unique to this PR][2], and sounds orthogonal.  I think we should:

```
```

to get this landed and unblock the backport chain for this `FastFix` bug.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2723/pull-ci-openshift-machine-config-operator-master-e2e-agnostic-upgrade/1428428458068807680
[2]: https://search.ci.openshift.org/chart?maxAge=24h&type=junit&search=pods%20should%20successfully%20create%20sandboxes%20by%20not%20timing%20out

kikisdeliveryservice:
Given the urgency and fact that the e2e test is required by basically red. Let's merge to unblock 

PR #2742, opened 2021-08-31
Bug 1970021: Revert ephemeral NM configuration change

yboaron:
This PR :
1- Deletes system-connections-merged directory from all on-prem platforms
2- Sets NetworkManager to look for its keyfiles in default dir (/etc/NetworkManager/system-connections/)

 With this change,  networking config change made outside of a MachineConfig (for example by k8s-nmstate) will result in it being persistent.

This reverts commits: 
                   eef0b76fb55658d602218a459e6a4027b678a784
                    a2815aa4993954ec6f30785c8c0a1d0a79109a4b
                    6666e7f9302083153031dadff7385887905eb2e4


cybertron (changes requested):
There are references to this in the configure-ovs script too[0]. I think that will break if we don't fix it in this change too.

https://github.com/openshift/machine-config-operator/blob/d6e945e6647bfd00a66b78810ae3ef6948451fba/templates/common/_base/files/configure-ovs-network.yaml#L11

sinnykumari:
> There are references to this in the configure-ovs script too[0]. I think that will break if we don't fix it in this change too.
> 
> https://github.com/openshift/machine-config-operator/blob/d6e945e6647bfd00a66b78810ae3ef6948451fba/templates/common/_base/files/configure-ovs-network.yaml#L11

Since, MCO team doesn't have much context about these templates.
/cc @trozet

yboaron:
> There are references to this in the configure-ovs script too[0]. I think that will break if we don't fix it in this change too.
> 
> https://github.com/openshift/machine-config-operator/blob/d6e945e6647bfd00a66b78810ae3ef6948451fba/templates/common/_base/files/configure-ovs-network.yaml#L11

Yep, noticed that, thought to have it in a follow-up PR.
IIUC,   ovs-configure should work even if systemConnectionsMerged directory doesn't exist, see [1] .

@cybertron @trozet  WDYT ? 

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml#L13-#L17

cybertron:
/retitle Bug 1970021: Revert ephemeral NM configuration change

cybertron:
> > There are references to this in the configure-ovs script too[0]. I think that will break if we don't fix it in this change too.
> > https://github.com/openshift/machine-config-operator/blob/d6e945e6647bfd00a66b78810ae3ef6948451fba/templates/common/_base/files/configure-ovs-network.yaml#L11
> 
> Yep, noticed that, thought to have it in a follow-up PR.
> IIUC, ovs-configure should work even if systemConnectionsMerged directory doesn't exist, see [1] .
> 
> @cybertron @trozet WDYT ?
> 
> [1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml#L13-#L17

Ah, okay that should be fine then. We should do a followup, but it won't hurt to leave the functionality in the script. Might make backports for the script a little easier too.

/lgtm

cybertron:
Hmm, if the bot never accepts my retitle can someone add that bug reference? Thanks.

kikisdeliveryservice:
@cybertron @trozet is this ready to merge?

cybertron:
This is failing the metal-ipi ipv6 ci jobs due to a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=1998643

However, I have successfully tested it locally (the bug causes bootstrap to timeout, but the cluster still deploys). configure-ovs is working fine with this change.

While this is a non-trivial change in behavior, it's actually moving us back to a more normal network configuration so if anything it will probably avoid more issues than it causes.

I'm not sure what, if any, of the other platform jobs exercise OVNK. I would hope at least one of them is, and if so then we should have sufficient ci coverage even with the metal-ipi jobs.

kikisdeliveryservice (approved):
Adding an approval and a hold. Please remove the hold once you are satisfied with ci/approvals from affected parties

/hold

mandre (commented):
/lgtm
/test e2e-vsphere

yboaron:
I can see that other PRs also fail to run vsphere gate, @jcpowermac @patrickdillon do you think we can merge this PR?

cybertron:
I discussed this with the vsphere team the other day and they said if the vsphere-upgrade job passes then that covers the regular vsphere job too. Since vsphere-upgrade passed we shouldn't need to block on the failing one. Once agnostic upgrade is green I think we're good.

/cc @phoracek @qinqon

kikisdeliveryservice:
> I discussed this with the vsphere team the other day and they said if the vsphere-upgrade job passes then that covers the regular vsphere job too. Since vsphere-upgrade passed we shouldn't need to block on the failing one. Once agnostic upgrade is green I think we're good.
> 
> /cc @phoracek @qinqon

This is great to know!

/test e2e-agnostic-upgrade

kikisdeliveryservice:
This is passing reqd tests, remove the hold when you're ready @cybertron

yboaron:
All required gates are green and we have a green light from all representatives of relevant platforms,  I think we can merge it  
/hold cancel

PR #2752, opened 2021-09-02
Bug 1993922: fixes 1 to 1 kubelet config mapping

rphillips:
**- What I did**
The bug reported a problem with the 1-to-1 mapping to MCPs and Kubelet Configs. I figured out the `getManageKubeletConfigKey` function did not include a LabelSelector to filter out the KubeletConfig for the Pool, thus including all the KubeletConfigs in the system for the limit validation.

**- How to verify it**
The patch adds a test to validate multiple pools and their 1:1 relationship with KubeletConfigs.

```
Fix KubeletConfig 1-to-1 relationship to Pools.
```

### DIFF
--- a/pkg/controller/kubelet-config/helpers.go
+++ b/pkg/controller/kubelet-config/helpers.go
@@ -12,6 +12,7 @@ import (
 	"github.com/vincent-petithory/dataurl"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice:
PTAL

/assign @umohnani8 @QiWang19

kikisdeliveryservice:
@rphillips might need to double check whether this change affects e2e bc it's failing in the kc section.

QiWang19:
Launched a cluster of this pr, failed to create the [kubeletconfig ](https://github.com/openshift/machine-config-operator/blob/master/docs/KubeletConfigDesign.md#example---setting-the-kubelet-log-level)

```
Status:
  Conditions:
    Last Transition Time:  2021-09-02T20:56:07Z
    Message:               could not get kubelet config key: error listing kubelet configs: unable to parse requirement: <nil>: Invalid value: "&LabelSelector{MatchLabels:map[string]string{machineconfiguration.openshift.io/role:": prefix part a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'); name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')

```

rphillips:
@QiWang19 thanks... I had to add a filter. Looks like we do this in most of the other areas of the code.

QiWang19:
> @QiWang19 thanks... I had to add a filter. Looks like we do this in most of the other areas of the code.

Yes, good to know

QiWang19 (commented):
/lgtm

pkg/controller/kubelet-config/helpers.go
```diff
@@ -145,22 +146,36 @@ func findKubeletConfig(mc *mcfgv1.MachineConfig) (*ign3types.File, error) {
...
+		kcList = append(kcList, kc)
+	}
+
+	for _, kc := range kcList {
 		if kc.Name != cfg.Name {
 			continue
 		}
 		val, ok := kc.GetAnnotations()[ctrlcommon.MCNameSuffixAnnotationKey]
 		// If we find a matching kubelet config and it is the only one in the list, then return the default MC name with no suffix
 		// add check len(kcList.Items) < 2, mc name should not suffixed if cfg is the first kubelet config to be updated/created
```
kikisdeliveryservice, line 177:
```suggestion
		// add check len(kcList) < 2, mc name should not suffixed if cfg is the first kubelet config to be updated/created
```

kikisdeliveryservice (commented):
just one small thing bc the comment is now not accurate

kikisdeliveryservice (approved):
Adding approval and hold. If you prefer to fix in a followup feel free to remove.

/hold

kikisdeliveryservice:
removing hold and re-adding lgtm since this was a comment tweak

/hold cancel
/lgtm

PR #2755, opened 2021-09-05
[OCPNODE-553] Support allowMirrorByTags From config/v1 ImageContentPolicy

QiWang19:

**- What I did**
Support allowMirrorByTags From config/v1 ImageContentPolicy
REF: https://issues.redhat.com/browse/OCPNODE-553

Depends on the  openshift/runtime-utils PR: https://github.com/openshift/runtime-utils/pull/12

**- How to verify it**
Apply the ImageContentPolicy to a cluster, debug into the node, `etc/containers/registries.conf` does not set `mirror-by-digest-only=true`.
```yml
kind: ImageContentPolicy
metadata:
  name: ubi8repo
spec:
  repositoryDigestMirrors:
  - mirrors:
    - example.io/example/ubi-minimal 
    source: registry.access.redhat.com/ubi8/ubi-minimal
    allowMirrorByTags: true
```


QiWang19:
Close PR, the allow mirror by tag feature will be implemented by https://github.com/openshift/machine-config-operator/pull/3037

PR #2764, opened 2021-09-14
[OCPNODE-709] add control-plane label

rphillips:
Add control plane label to master nodes.

[OCPNODE-709](https://issues.redhat.com/browse/OCPNODE-709)

### DIFF
--- a/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml
@@ -29,7 +29,7 @@ contents: |
         --container-runtime=remote \
         --container-runtime-endpoint=/var/run/crio/crio.sock \
         --runtime-cgroups=/system.slice/crio.service \
-        --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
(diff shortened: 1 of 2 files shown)


cgwalters:
> Add control plane label to master and worker nodes.

I think you mean "Add control plane label where we have master label"?

rphillips:
/hold

this PR needs the kubelet PR https://github.com/openshift/kubernetes/pull/957

rphillips:
/hold

rhcos is not building for 4.10 yet.

rphillips:
/hold cancel

4.10 cluster installed correctly.

cgwalters:
(I don't follow upstream kube much, I should but...) a question I didn't see answered here in the PR or the ticket: is this happening in upstream kube at all?  Or are we leading by example?  The latter is fine by me but if there's not an effort there I would hope we start one.

rphillips:
We are not necessarily leading the change here. Upstream has already implemented this in kubeadm [[PR](https://github.com/kubernetes/kubernetes/pull/95382)]. It is my understanding that upstream controllers will eventually migrate to the control-plane label.

yuqi-zhang:
Just a quick question: I remember there being a canonical list of accepted role labels that the kubelet can bootstrap with in OCP, and that was (at least in part) preventing booting into custom pools.

Do you know if that list still exists and whether the above still holds?

PR #2739, opened 2021-08-27
manifests: Use embed instead of go-bindata to embed assets

LorbusChris:
**- What I did**

**- How to verify it**

**- Description for the changelog**

PR #2765, opened 2021-09-14
Use go "embed" module to store manifests

Fedosin:
This PR starts using "embed" go module to store manifests in the binary. It is easier and more natural than doing so with go-bindata.


cgwalters:
Dup of https://github.com/openshift/machine-config-operator/pull/2739 ?

Fedosin:
Oh, sorry, didn't notice that patch. As I see it's closed now, so we can continue here.

sinnykumari:
LGTM
@LorbusChris Since you also invested time in this with PR https://github.com/openshift/machine-config-operator/pull/2739 , can you take a look in case I missed something.

LorbusChris (approved):
LGTM

sinnykumari (approved):

PR #2770, opened 2021-09-16
Bug 1990625: configure-ovs: Persist addr-gen-mode for ipv6 connections

cybertron:
Currently br-ex defaults to the mode stable-privacy because that's
the default for new nmcli connections. However, the host interface
is actually configured for eui64, which causes inconsistency from
initial boot to bridge configuration.

To fix this we can just persist the addr-gen-mode like we do with
the DHCP parameters.

**- What I did**

**- How to verify it**

Persist addr-gen-mode for ipv6 connection when creating br-ex.

### DIFF
--- a/templates/common/_base/files/configure-ovs-network.yaml
+++ b/templates/common/_base/files/configure-ovs-network.yaml
@@ -166,6 +166,11 @@ contents:
         extra_brex_args+="ipv6.dhcp-duid ${dhcp6_client_id} "
       fi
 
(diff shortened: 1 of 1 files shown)


cybertron:
Had a missing quote in the first version of this. The current one should work now.

/test e2e-metal-ipi-ovn-ipv6

cybertron:
/retest-required

The metal-ipi OVN job passed, which is the main one we need to worry about. Just need a pass from the required jobs and we should be good to go.

bfournie (approved):

trozet (approved):
/lgtm

kikisdeliveryservice (approved):

PR #2778, opened 2021-09-28
Merge OKD-specific changes into master

LorbusChris:
I'm opening this draft PR to open a forum for MCO and OKD maintainers to discuss how the OKD-specific changes to the MCO code base can be merged back into master. This would allow for dropping the fork and lighten the OKD maintenance burden.

I'd like to ask @fortinj66 and @vrutkovs to expand on the reasoning behind those individual changes, and provide links to associated issue/bug reports (if any). I am open to opening more fine-grained PRs against master for each issue addressed with these changes.

My hopes are this will make it easier for @openshift/openshift-team-mco to review the changes.

### DIFF
--- a/templates/common/_base/units/machine-config-daemon-pull.service.yaml
+++ b/templates/common/_base/units/machine-config-daemon-pull.service.yaml
@@ -18,8 +18,8 @@ contents: |
   # See https://github.com/coreos/fedora-coreos-tracker/issues/354
   ExecStart=/bin/sh -c '/bin/mkdir -p /run/bin && chcon --reference=/usr/bin /run/bin'
   ExecStart=/bin/sh -c "while ! /usr/bin/podman pull --authfile=/var/lib/kubelet/config.json --quiet '{{ .Images.machineConfigOperator }}'; do sleep 1; done"
-  ExecStart=/bin/sh -c "/usr/bin/podman run --rm --quiet --net=host --entrypoint=cat '{{ .Images.machineConfigOperator }}' /usr/bin/machine-config-daemon > /run/bin/machine-config-daemon.tmp"
-  ExecStart=/bin/sh -c '/usr/bin/chmod a+x /run/bin/machine-config-daemon.tmp && mv /run/bin/machine-config-daemon.tmp /run/bin/machine-config-daemon'
+  ExecStart=/usr/bin/podman run --rm --quiet --net=host -v /run/bin:/host/run/bin:z --entrypoint=cp '{{ .Images.machineConfigOperator }}' /usr/bin/machine-config-daemon /host/run/bin
+  ExecStart=/bin/chcon system_u:object_r:bin_t:s0 /run/bin/machine-config-daemon
   {{if .Proxy -}}
   EnvironmentFile=/etc/mco/proxy.env
   {{end -}}


kikisdeliveryservice:
I know this is a draft, but just noting for future discussion, wondering why this is reverting a (relatively) new crio PR: https://github.com/openshift/machine-config-operator/pull/2805

I think for this type of change would prefer to do it as a separate PR, but that can wait for now while this is still in draft.

vrutkovs:
>I know this is a draft, but just noting for future discussion, wondering why this is reverting a (relatively) new crio PR: #2805

temporary fix until we get 1.23, [slack thread](https://coreos.slack.com/archives/CBBJY9GSX/p1636708657087400)

pkg/daemon/update.go
```diff
@@ -1078,9 +1078,9 @@ func (dn *Daemon) applyExtensions(oldConfig, newConfig *mcfgv1.MachineConfig) er
 	}
 
 	// Validate extensions allowlist on RHCOS nodes
-	if err := validateExtensions(newConfig.Spec.Extensions); err != nil && dn.os.IsRHCOS() {
-		return err
-	}
+	// if err := validateExtensions(newConfig.Spec.Extensions); err != nil && dn.os.IsRHCOS() {
```
vrutkovs, line 1081: This probably can be dropped - afaik we don't use extensions since 4.8, so it won't affect upgrades.
I'll give that a try later on

pkg/daemon/update.go
```diff
@@ -1164,6 +1164,11 @@ func (dn *Daemon) updateFiles(oldIgnConfig, newIgnConfig ign3types.Config) error
 	if err := dn.writeUnits(newIgnConfig.Systemd.Units); err != nil {
 		return err
 	}
+	// In OKD /var/home/core/.ssh/authorized_keys is not created during ignition.  Create
+	// from /var/home/core/.ssh/authorized_keys.d/ignition if missing
+	if err := dn.writeMissingAuthorizedKeys(); err != nil {
+		return err
+	}
```
LorbusChris, line 1171: @vrutkovs this commit can be dropped in the next rebase now that https://github.com/openshift/machine-config-operator/pull/2688 has merged

pkg/daemon/update.go
LorbusChris: I believe this would be a no-op on FCOS anyway, due to the `err != nil && dn.os.IsRHCOS()`, no?

pkg/daemon/kernelargs.go
```diff
@@ -44,10 +38,8 @@ func isArgTunable(arg string) (bool, error) {
 
 	if os.IsRHCOS() {
 		return tuneableRHCOSArgsAllowlist[arg], nil
-	} else if os.IsFCOS() {
-		return tuneableFCOSArgsAllowlist[arg], nil
 	}
-	return false, nil
+	return true, nil
```
LorbusChris, line 42:
```
	} else if os.IsFCOS() {
		return true, nil
	}
	return false, nil
```
in order to still consider other systems this may run on, like RHEL.

cmd/machine-config-daemon/pivot.go
```diff
@@ -79,6 +79,17 @@ func run(_ *cobra.Command, args []string) (retErr error) {
 		}
 	}
 
+	// Check to see if we need to tune kernel arguments
+	tuningChanged, err := daemon.UpdateTuningArgs(daemon.KernelTuningFile, daemon.CmdLineFile)
```
LorbusChris, line 83: since we're not using the kernel-args file anymore in 4.9+, I think this commit can be dropped too

templates/common/_base/units/ovsdb-server.service.yaml
```diff
@@ -5,3 +5,6 @@ dropins:
     contents: |
       [Service]
       Restart=always
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /var/lib/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /etc/openvswitch'
+      ExecStartPre=-/bin/sh -c '/usr/bin/chown -R ${OVS_USER_ID} /run/openvswitch'
```
LorbusChris, line 10: We have to check if upgrades from 4.9 to 4.10 are still affected by this. If not, this commit can be dropped.

cmd/machine-config-daemon/pivot.go
vrutkovs: We use kernel-args for bootstrap

LorbusChris: I think now that https://github.com/openshift/installer/pull/5439 has merged, the installer doesn't use it anymore either. I think that was the last place?

PR #2782, opened 2021-09-28
Bug 2002121: Periodically send gratuitous ARPs

cybertron:
Under some circumstances, the network infrastructure might miss the
GARPs sent when a VIP fails over. This causes the VIP to become
inaccessible because the network doesn't know where it is. Worse, in
some cases the network might not realize it doesn't have the correct
information because it has previously cached the old VIP location.

To get around this, configure keepalived to periodically send GARPs.
This will ensure that even if the failover is missed, the network
will eventually catch up.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
+++ b/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml
@@ -6,6 +6,7 @@ contents:
         enable_script_security
         script_user root
         max_auto_priority -1
+        vrrp_garp_master_refresh 60
     }
 
     # These are separate checks to provide the following behavior:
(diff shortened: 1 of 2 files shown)


mandre (commented):
/lgtm

kikisdeliveryservice (approved):
Adding hold until all reviews are in and hopefully vsphere test passes.

Remove when ready @cybertron

flavio-fernandes (commented):
/lgtm

cybertron:
vsphere-upgrade passed and the vsphere failure doesn't seem to be related.

@jcpowermac @rvanderp3 are you okay with merging this?

cybertron:
/hold cancel

vsphere-upgrade passed and all of the other platforms have signed off too.

PR #2413, opened 2021-02-16
Bug 1927042: [baremetal & friends] Don't write empty static pod manifests

cybertron:
Kubelet complains about empty files in the manifests directory because
it can't parse them. While this doesn't cause any harm, it has caused
bugs to be mis-routed because people see the error messages while
debugging and think something is wrong.

Instead of conditionalizing the content of the files, we can
conditionalize the path and just write them to a disabled location
so they won't be deployed and they also won't be processed by kubelet.
Note that I tried just using /dev/null to avoid writing the files
entirely, but I was not able to get that to work.

**- What I did**

**- How to verify it**

**- Description for the changelog**

PR #2783, opened 2021-09-28
[release-4.7] Bug 2008692: Don't write empty static pod manifests

cybertron:
Kubelet complains about empty files in the manifests directory because
it can't parse them. While this doesn't cause any harm, it has caused
bugs to be mis-routed because people see the error messages while
debugging and think something is wrong.

Instead of conditionalizing the content of the files, we can
conditionalize the path and just write them to a disabled location
so they won't be deployed and they also won't be processed by kubelet.
Note that I tried just using /dev/null to avoid writing the files
entirely, but I was not able to get that to work.

**- What I did**

**- How to verify it**

### DIFF
--- a/templates/common/on-prem/files/coredns.yaml
+++ b/templates/common/on-prem/files/coredns.yaml
@@ -1,8 +1,7 @@
 mode: 0644
-path: "/etc/kubernetes/manifests/coredns.yaml"
+path: {{ if (onPremPlatformAPIServerInternalIP .) -}} "/etc/kubernetes/manifests/coredns.yaml" {{ else }} "/etc/kubernetes/disabled-manifests/coredns.yaml" {{ end }}
 contents:
   inline: |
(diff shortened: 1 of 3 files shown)


mandre (commented):
/lgtm

cybertron:
/test e2e-vsphere-upi

This is the test that is most affected by this change.

kikisdeliveryservice:
Original fix PR (since this couldn't be auto-cherrypicked): https://github.com/openshift/machine-config-operator/pull/2413

kikisdeliveryservice:
@cybertron is this good now? vsphere upi test passed.

cybertron:
@kikisdeliveryservice Thanks for linking the original PR. I didn't think to do that when I pushed this.

With the upi job passing we should be good to merge this.

kikisdeliveryservice (approved):

PR #2786, opened 2021-09-30
Vendor library go changes for enable gcp and vsphere ccms

lobziik:
Updated library-go dependency with changes from openshift/library-go#1221. Needed for proceeding with GCP and vSphere out of tree providers related work.


pkg/controller/template/render_test.go
```diff
@@ -48,11 +45,26 @@ func TestCloudProvider(t *testing.T) {
...
+		featureGate: newFeatures("cluster", "CustomNoUpgrade", []string{cloudprovider.ExternalCloudProviderFeature}, nil),
+		res:         "external",
+	}, {
+		platform:    configv1.OpenStackPlatformType,
+		featureGate: newFeatures("cluster", "CustomNoUpgrade", []string{cloudprovider.ExternalCloudProviderFeature}, nil),
+		res:         "external",
+	}, {
+		platform:    configv1.OpenStackPlatformType,
+		featureGate: newFeatures("cluster", "CustomNoUpgrade", []string{cloudprovider.ExternalCloudProviderFeature}, nil),
+		res:         "external",
```
kikisdeliveryservice, line 60: Is there a reason why these two OpenStack cases seem to be identical? Looks like an accidental dupe?

kikisdeliveryservice (commented):
test q

pkg/controller/template/render_test.go
lobziik: Yup, accidental. Fixed.

go.mod
```diff
@@ -35,9 +35,9 @@ require (
 	github.com/imdario/mergo v0.3.12
 	github.com/mattn/go-isatty v0.0.12 // indirect
 	github.com/opencontainers/go-digest v1.0.0
-	github.com/openshift/api v0.0.0-20210921141042-49b82c6d1ca9
+	github.com/openshift/api v0.0.0-20210924154557-a4f696157341
```
sinnykumari, line 38: Also need to pin this version by updating openshift/api version in replace section https://github.com/openshift/machine-config-operator/blob/master/go.mod#L38

pkg/controller/template/render_test.go
```diff
@@ -31,9 +31,6 @@ func TestCloudProvider(t *testing.T) {
 		featureGate *configv1.FeatureGate
 		res         string
 	}{{
-		platform: configv1.AWSPlatformType,
```
sinnykumari, line 34: I don't fully understand. Are these no longer in use i.e. cloudProvider ID will be only set using featuregate?

lobziik: Just moved this case to line 61.

go.mod
lobziik: Done

sinnykumari (approved):
LGTM. Will wait for Kirsten in case there are some final questions/concerns.

kikisdeliveryservice (approved):
seems good to go now =)

/lgtm

PR #2790, opened 2021-10-05
Bug 2013726: [on-prem] Update NetworkManager-resolv-prepender.yaml

Rupesh-git-eng:
Fixies # Don't add white namespace while generating /etc/resolv.conf on nodes if variables are empty. 
Empty NetworkManager variables cause generating whitespace in /etc/resolv.conf between search and search domains.
This patch checks if the variables are empty. If empty, don't add it to Variable $DOMAINS 

i.e 
 
$ cat <sosreport>/etc/resolv.conf 
# Generated by KNI resolv prepender NM dispatcher script
search   abc.com ## **three whitespace added here between search and abc.com** 
nameserver xx.xx.xx.xx
nameserver xx.xx.xx.xx
nameserver xx.xx.xx.xx

**- What I did**
Validate if variables are set to empty. 

**- How to verify it**
Don't pass search domain via dhcp service. 

Validate the NetworkManager variables before using them in the dispatcher script. 

Validate the NetworkManager variables before using them in the dispatcher script.

### DIFF
--- a/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
+++ b/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml
@@ -50,7 +50,18 @@ contents:
             show \
             "{{ onPremPlatformAPIServerInternalIP . }}" \
             "{{ onPremPlatformIngressIP . }}")
-        DOMAINS="${IP4_DOMAINS} ${IP6_DOMAINS} {{.DNS.Spec.BaseDomain}}"
+            
+        # Ignore white space while creating /etc/resolv.conf between search and search domain names if the NM variables are empty.  
+        
+        if   [[ -z "$IP4_DOMAINS" && -z "$IP6_DOMAINS" ]] ; then 
+             DOMAINS="{{.DNS.Spec.BaseDomain}}" 
+        elif [[ -n "$IP4_DOMAINS" && -z "$IP6_DOMAINS" ]] ; then
+             DOMAINS="${IP4_DOMAINS} {{.DNS.Spec.BaseDomain}}" 
+        elif [[ -z "$IP4_DOMAINS" && -n "$IP6_DOMAINS" ]] ; then
+             DOMAINS="${IP6_DOMAINS} {{.DNS.Spec.BaseDomain}}" ; else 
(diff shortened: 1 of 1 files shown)


mandre:
It's not clear to me what we're trying to achieve here. Is it a cosmetic change?

Rupesh-git-eng:
> @Rupesh-git-eng just double checking: is this related to a bug?

Yes.

Rupesh-git-eng:
[https://bugzilla.redhat.com/show_bug.cgi?id=2013726](https://bugzilla.redhat.com/show_bug.cgi?id=2013726)

mandre:
> > @sinnykumari: [Bugzilla bug 2013726](https://bugzilla.redhat.com/show_bug.cgi?id=2013726) is in a bug group that is not in the allowed groups for this repo. Allowed groups for this repo are:
> > ```
> > * qe_staff
> > 
> > * redhat
> > ```
> 
> Not sure, why this is complaining.

It complains because the bug is private. This looks like a dup of https://github.com/openshift/machine-config-operator/pull/2823/, @cybertron do you confirm?

cybertron:
It's not exactly a dupe, but #2823 will also fix this. I would prefer to just merge that because it also fixes another bug.

I'm also not entirely sure about all of the changes in this patch. My understanding when we added IP4 and IP6_DOMAINS was that they were supposed to include DHCP-provided nameservers (see https://github.com/openshift/machine-config-operator/commit/bbbb72292b1f03c5f86676c9b1cefffbec387f80#diff-255f8a4599166f31961853ea8626f969ca4231c55aacbc20a5bb3ceb640f911d). One side benefit of #2823 is that we just use what NetworkManager provides instead of having to come up with a complete list ourselves. That should result in less surprising behavior and eliminate questions like this.

kikisdeliveryservice:
adding a hold until the above is resolved. 

/hold

cybertron:
Note that #2823 has merged so the bug being addressed by this should be fixed. Even if we do need an additional patch, this one won't work because the other PR changed how the domain list is built.

kikisdeliveryservice:
> Note that #2823 has merged so the bug being addressed by this should be fixed. Even if we do need an additional patch, this one won't work because the other PR changed how the domain list is built.

Thanks for the update and details, @cybertron ! Given the above, I'll close this PR.

PR #2795, opened 2021-10-07
Proactively detect config drift

cheesesashimi:
This enables the MCD to check if on-disk configuration has drifted from the currently applied MachineConfig. This work is being tracked in https://issues.redhat.com/browse/MCO-69

**- What I did**
I added a goroutine to the MCD to run a Config Drift Monitor. Under the hood, the Config Drift Monitor uses fsnotify to wire up handlers for all of the directories referenced in a given MachineConfig. If a write event is detected for any file (or files relating to a Systemd unit) defined in a MachineConfig, the Config Drift Monitor will run the `validateOnDiskState` function. If the on-disk configuration has drifted from what is specified in the MachineConfig, the node will be marked Degraded. Additionally, I added a `validateOnDiskState` check to the `syncNode` routine so that if the config has drifted prior to an update, it will prevent the cluster from getting into an inconsistent state.

Care was taken to ensure that the Config Drift Monitor only runs whenever no updates are occurring. When the MCD is booting, updates may still be pending. To avoid spurious config drift errors (since technically, the config will "drift" from one config to another), the Config Drift Monitor will only run when the MCD has finished booting. Additionally, prior to applying an update, the Config Drift Monitor is shut down. Again, this is done because the config will "drift" from one config to another while the update is being applied. If no reboot is required, the Config Drift Monitor will be started once the update is complete.

**- How to verify it**
1. Within a cluster, create a new MachineConfig, e.g.:
```yaml
---
kind: MachineConfig
metadata:
  labels:
  name: etc-file
spec:
  config:
    ignition:
      version: 3.2.0
    storage:
      files:
      - contents:
          source: data:,hello%20world%0A
        mode: 420
        path: /etc/etc-file
```

2. Verify that the Config Drift Monitor has started by reviewing the MCD logs:
```console
I1203 15:28:31.366342    1883 daemon.go:647] Starting MachineConfigDaemon
I1203 15:28:31.366426    1883 daemon.go:654] Enabling Kubelet Healthz Monitor
I1203 15:28:38.461070    1883 daemon.go:1343] Updating Node ip-10-0-132-121.ec2.internal
I1203 15:28:38.461091    1883 daemon.go:1343] Updating Node ip-10-0-132-86.ec2.internal
I1203 15:28:38.461111    1883 daemon.go:1343] Updating Node ip-10-0-150-37.ec2.internal
I1203 15:28:38.461116    1883 daemon.go:1343] Updating Node ip-10-0-151-158.ec2.internal
I1203 15:28:38.461121    1883 daemon.go:1343] Updating Node ip-10-0-164-151.ec2.internal
I1203 15:28:38.461125    1883 daemon.go:1343] Updating Node ip-10-0-170-92.ec2.internal
I1203 15:28:39.462097    1883 daemon.go:402] Started syncing node "ip-10-0-132-121.ec2.internal" (2021-12-03 15:28:39.462086333 +0000 UTC m=+9.130821879)
I1203 15:28:39.462153    1883 daemon.go:404] Finished syncing node "ip-10-0-132-121.ec2.internal" (63.876µs)
I1203 15:28:39.462162    1883 daemon.go:402] Started syncing node "ip-10-0-132-86.ec2.internal" (2021-12-03 15:28:39.462160974 +0000 UTC m=+9.130896512)
I1203 15:28:39.462215    1883 daemon.go:394] Node ip-10-0-132-86.ec2.internal is not labeled node-role.kubernetes.io/master
I1203 15:28:39.468002    1883 daemon.go:910] Current config: rendered-infra-e40714a84270cd41458ddf52dd8c465a
I1203 15:28:39.468060    1883 daemon.go:911] Desired config: rendered-infra-fcc7352d9598d0d4441fdd00aac05739
I1203 15:28:39.474917    1883 update.go:1994] Disk currentConfig rendered-infra-fcc7352d9598d0d4441fdd00aac05739 overrides node's currentConfig annotation rendered-infra-e40714a84270cd41458ddf52dd8c465a
I1203 15:28:39.477776    1883 daemon.go:1193] Validating against pending config rendered-infra-fcc7352d9598d0d4441fdd00aac05739
I1203 15:28:39.495045    1883 daemon.go:1211] Validated on-disk state
I1203 15:28:39.495129    1883 event.go:282] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"ip-10-0-132-86.ec2.internal", UID:"b9afb880-9b75-45ef-b454-a89eb37b6d6f", APIVersion:"", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'NodeDone' Setting node ip-10-0-132-86.ec2.internal, currentConfig rendered-infra-fcc7352d9598d0d4441fdd00aac05739 to Done
I1203 15:28:39.508451    1883 daemon.go:1343] Updating Node ip-10-0-132-86.ec2.internal
I1203 15:28:39.509827    1883 daemon.go:1262] Completing pending config rendered-infra-fcc7352d9598d0d4441fdd00aac05739
I1203 15:28:39.509839    1883 drain.go:44] Initiating uncordon on node (currently schedulable: false)
I1203 15:28:39.523698    1883 drain.go:62] RunCordonOrUncordon() succeeded but node is still not in uncordon state, retrying
I1203 15:28:39.524490    1883 daemon.go:1343] Updating Node ip-10-0-132-86.ec2.internal
I1203 15:28:39.558790    1883 daemon.go:1343] Updating Node ip-10-0-132-86.ec2.internal
I1203 15:28:49.523812    1883 drain.go:44] Initiating uncordon on node (currently schedulable: true)
I1203 15:28:49.523838    1883 drain.go:66] uncordon succeeded on node (currently schedulable: true)
I1203 15:28:49.523846    1883 update.go:1994] Update completed for config rendered-infra-fcc7352d9598d0d4441fdd00aac05739 and node has been successfully uncordoned
I1203 15:28:49.526114    1883 daemon.go:1278] In desired config rendered-infra-fcc7352d9598d0d4441fdd00aac05739
I1203 15:28:49.536226    1883 config_drift_monitor.go:240] Config Drift Monitor started
```

3. Once the new MachineConfig has been applied to all pool nodes, modify the contents on-disk of this file:
```console
$ oc debug node/<node-name> -- printf "not-the-data" > /host/etc/etc-file
```

4. Check the MCD logs for the node this targeted. You should see something that resembles:
```console
E1203 15:40:28.247846    1883 on_disk_validation.go:221] content mismatch for file "/etc/etc-file" (-want +got):
  bytes.Join({
- 	"hello world\n",
+ 	"not-the-data",
  }, "")
E1203 15:40:28.247952    1883 daemon.go:733] content mismatch for file "/etc/etc-file"
E1203 15:40:28.247984    1883 writer.go:135] Marking Degraded due to: content mismatch for file "/etc/etc-file"
I1203 15:40:28.248939    1883 event.go:282] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"ip-10-0-132-86.ec2.internal", UID:"b9afb880-9b75-45ef-b454-a89eb37b6d6f", APIVersion:"", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'ConfigDriftDetected' content mismatch for file "/etc/etc-file"
I1203 15:40:28.272864    1883 daemon.go:1343] Updating Node ip-10-0-132-86.ec2.internal
I1203 15:40:29.273087    1883 daemon.go:402] Started syncing node "ip-10-0-132-86.ec2.internal" (2021-12-03 15:40:29.27307442 +0000 UTC m=+718.941809966)
E1203 15:40:29.286249    1883 on_disk_validation.go:221] content mismatch for file "/etc/etc-file" (-want +got):
  bytes.Join({
- 	"hello world\n",
+ 	"not-the-data",
  }, "")
E1203 15:40:29.286289    1883 daemon.go:506] Preflight config drift check failed: content mismatch for file "/etc/etc-file"
I1203 15:40:29.286296    1883 daemon.go:404] Finished syncing node "ip-10-0-132-86.ec2.internal" (13.220356ms)
```

To recover, either change the file contents back or `oc debug node/<node-name> -- touch /host/run/machine-config-daemon-force`. The latter will force the MCD to apply the update and will cause a reboot which may or may not be desirable.
 
Proactively detect on-disk config drift

### DIFF
--- a/pkg/daemon/config_drift_monitor.go
+++ b/pkg/daemon/config_drift_monitor.go
@@ -1,0 +1,370 @@
+package daemon
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+
+	ign2types "github.com/coreos/ignition/config/v2_2/types"
+	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
+	"github.com/fsnotify/fsnotify"
+	"github.com/golang/glog"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	ctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Outermost error type for config drift errors
+type configDriftErr error
+
+// Error type for file config drifts
+type fileConfigDriftErr error
+
+// Error type for systemd unit config drifts
+type unitConfigDriftErr error
+
+type ConfigDriftMonitor interface {
+	Start(ConfigDriftMonitorOpts) error
+	Done() <-chan struct{}
+	IsRunning() bool
+	Stop()
+}
+
+type ConfigDriftMonitorOpts struct {
+	// Called whenever a config drift is detected.
+	OnDrift func(error)
+	// The currently applied MachineConfig.
+	MachineConfig *mcfgv1.MachineConfig
+	// The Systemd dropin path location.
+	// Defaults to /etc/systemd/system
+	SystemdPath string
+	// Channel to report unknown errors
+	ErrChan chan<- error
+}
+
+// Holds the Config Drift Watcher and ensures we only have a single instance
+// running at a given time.
+type configDriftMonitor struct {
+	cdw    *configDriftWatcher
+	mu     sync.Mutex
+	stopCh chan struct{}
+}
+
+// Implements the Config Drift Watcher
+type configDriftWatcher struct {
+	ConfigDriftMonitorOpts
+	watcher   *fsnotify.Watcher
+	filePaths sets.String
+	wg        sync.WaitGroup
+	stopCh    chan struct{}
+}
+
+// Holds a single Config Drift Watcher and starts / stops it as necessary while
+// ensuring that only a single Config Drift Monitor is running at any given
+// time (within its scope).
+func NewConfigDriftMonitor() ConfigDriftMonitor {
+	return &configDriftMonitor{
+		mu:     sync.Mutex{},
+		stopCh: make(chan struct{}),
+	}
+}
+
+// Determines if a Config Drift Watcher is running; useful for avoiding certain
+// logic such as emitting a startup event.
+func (c *configDriftMonitor) IsRunning() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.cdw != nil
+}
+
+// Starts the Config Drift Monitor if not already started. It will create a new
+// Config Drift Monitor instance, if necessary.
+func (c *configDriftMonitor) Start(opts ConfigDriftMonitorOpts) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.cdw != nil {
+		return nil
+	}
+
+	var err error
+
+	// We don't have a Config Drift Watcher running, create one.
+	c.cdw, err = newConfigDriftWatcher(opts)
+	if err != nil {
+		return err
+	}
+
+	c.cdw.start()
+
+	return nil
+}
+
+// Provides a channel for listening if the currently running Config Drift
+// Monitor has been stopped. Can be used similar to <-ctx.Done().
+func (c *configDriftMonitor) Done() <-chan struct{} {
+	out := make(chan struct{})
+
+	go func() {
+		select {
+		case <-c.stopCh:
+			out <- struct{}{}
+			close(out)
+			return
+		}
+	}()
+
+	return out
+}
+
+// Stops the currently running Config Drift Monitor and deletes its instance so
+// that it cannot be reused. Also sends a signal to the Done() listener.
+func (c *configDriftMonitor) Stop() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.cdw != nil {
+		c.cdw.stop()
+		c.cdw = nil
+		c.stopCh <- struct{}{}
+	}
+}
+
+// Creates a new config drift watcher for a given machineconfig and registers
+// a callback for when a drift occurs.
+func newConfigDriftWatcher(opts ConfigDriftMonitorOpts) (*configDriftWatcher, error) {
+	if opts.OnDrift == nil {
+		return nil, fmt.Errorf("no ondrift function attached")
+	}
+
+	if opts.ErrChan == nil {
+		return nil, fmt.Errorf("no error channel provided")
+	}
+
+	if opts.MachineConfig == nil {
+		return nil, fmt.Errorf("no machine config provided")
+	}
+
+	if opts.SystemdPath == "" {
+		opts.SystemdPath = pathSystemd
+	}
+
+	c := &configDriftWatcher{
+		ConfigDriftMonitorOpts: opts,
+		stopCh:                 make(chan struct{}),
+	}
+
+	if err := c.initialize(); err != nil {
+		return nil, fmt.Errorf("could not initialize config drift monitor: %w", err)
+	}
+
+	return c, nil
+}
+
+// Performs the initial setup and initialization of the Config Drift Monitor,
+// such as identifying files from the MachineConfig, wiring up the watchers,
+// etc.
+func (c *configDriftWatcher) initialize() error {
+	var err error
+
+	// Initialize the fsnotify watcher
+	c.watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("could not start watcher: %w", err)
+	}
+
+	glog.V(4).Infof("Initializing Config Drift Monitor")
+	glog.V(4).Infof("Current MachineConfig: %s", c.MachineConfig.Name)
+
+	// Even though we wire up fsnotify to use the parent directories of each
+	// file, we keep track of the file paths individually so that we can ignore
+	// files in the same directory which are not referenced by the MachineConfig.
+	// This is useful when, for example, someone places a file in /etc using a
+	// MachineConfig (e.g., /etc/foo), but an unknown (to the MachineConfig) file
+	// in /etc (e.g., /etc/bar) is written to.
+	c.filePaths, err = getFilePathsFromMachineConfig(c.MachineConfig, c.SystemdPath)
+	if err != nil {
+		return fmt.Errorf("could not get file paths from machine config: %w", err)
+	}
+
+	// fsnotify (presently) uses inotify instead of fanotify on Linux.
+	// See: https://github.com/fsnotify/fsnotify/issues/114
+	//
+	// We must be judicious about the number of files we have wired up. However,
+	// inotify cannot recurse into directories. To work around these limitaions,
+	// we get the dirnames for all of the files in the Ignition config and dedupe
+	// them and attach watchers to those dirs.
+	dirPaths := getDirPathsFromFilePaths(c.filePaths)
+
+	glog.V(4).Infof("Will watch %d directories containing %d files:", len(dirPaths), len(c.filePaths))
+
+	// Wire up fsnotify to watch our config dirs
+	for _, path := range dirPaths {
+		glog.V(4).Infof("Watching dir: \"%s\"", path)
+		if err := c.watcher.Add(path); err != nil {
+			return fmt.Errorf("could not add fsnotify watcher to dir \"%s\": %w", path, err)
+		}
+	}
+
+	return nil
+}
+
+// Starts the Config Drift Watcher
+func (c *configDriftWatcher) start() {
+	c.wg = sync.WaitGroup{}
+	c.wg.Add(1)
+
+	go func() {
+		defer c.wg.Done()
+		for {
+			select {
+			case event := <-c.watcher.Events:
+				// Our watcher is reporting an event that we should look at.
+				if err := c.handleFileEvent(event); err != nil {
+					// Send unknown file event errors to the error channel.
+					c.ErrChan <- err
+				}
+			case err := <-c.watcher.Errors:
+				// Send fsnotify errors directly to the error channel.
+				c.ErrChan <- fmt.Errorf("fsnotify error: %w", err)
+			case <-c.stopCh:
+				// We received a stop signal, shutdown our watcher.
+				c.watcher.Close()
+				return
+			}
+		}
+	}()
+
+	glog.Info("Config Drift Monitor started")
+}
+
+// Stops the watcher.
+// Note: Once a Config Drift Watcher has been stopped, it cannot be started
+// again. A new instance must be created.
+func (c *configDriftWatcher) stop() {
+	c.stopCh <- struct{}{}
+	c.wg.Wait()
+	glog.Info("Config Drift Monitor has shut down")
+}
+
+// Handles the filesystem event for any of the files we're watching and
+// filters any config drift errors to the provided callback.
+func (c *configDriftWatcher) handleFileEvent(event fsnotify.Event) error {
+	err := c.checkMachineConfigForEvent(event)
+
+	if err == nil {
+		return nil
+	}
+
+	var cdErr configDriftErr
+	if errors.As(err, &cdErr) {
+		c.OnDrift(cdErr)
+		// Don't bubble this error up further since it's handled by OnDrift.
+		return nil
+	}
+
+	return fmt.Errorf("unknown config drift error: %w", err)
+}
+
+// Validates on disk state for potential config drift.
+func (c *configDriftWatcher) checkMachineConfigForEvent(event fsnotify.Event) error {
+	// Ignore events for files not found in the MachineConfig.
+	if !c.filePaths.Has(event.Name) {
+		return nil
+	}
+
+	if err := validateOnDiskState(c.MachineConfig, c.SystemdPath); err != nil {
+		return configDriftErr(err)
+	}
+
+	return nil
+}
+
+// Finds the paths for all files in a given MachineConfig.
+func getFilePathsFromMachineConfig(mc *mcfgv1.MachineConfig, systemdPath string) (sets.String, error) {
+	ignConfig, err := ctrlcommon.IgnParseWrapper(mc.Spec.Config.Raw)
+	if err != nil {
+		return sets.String{}, fmt.Errorf("could not get dirs from ignition config: %w", err)
+	}
+
+	switch typedConfig := ignConfig.(type) {
+	case ign3types.Config:
+		return getFilePathsFromIgn3Config(ignConfig.(ign3types.Config), systemdPath), nil
+	case ign2types.Config:
+		return getFilePathsFromIgn2Config(ignConfig.(ign2types.Config), systemdPath), nil
+	default:
+		return sets.String{}, fmt.Errorf("unexpected type for ignition config: %v", typedConfig)
+	}
+}
+
+// Extracts all unique directories from a given Ignition 3 config.
+func getFilePathsFromIgn3Config(ignConfig ign3types.Config, systemdPath string) sets.String {
+	files := sets.NewString()
+
+	// Get all the file paths from the ignition config
+	for _, ignFile := range ignConfig.Storage.Files {
+		if _, err := os.Stat(ignFile.Path); err == nil && !os.IsNotExist(err) {
+			files.Insert(ignFile.Path)
+		}
+	}
+
+	// Get all the file paths for systemd dropins from the ignition config
+	for _, unit := range ignConfig.Systemd.Units {
+		unitPath := getIgn3SystemdUnitPath(systemdPath, unit)
+		if _, err := os.Stat(unitPath); err == nil && !os.IsNotExist(err) {
+			files.Insert(unitPath)
+		}
+
+		for _, dropin := range unit.Dropins {
+			dropinPath := getIgn3SystemdDropinPath(systemdPath, unit, dropin)
+			if _, err := os.Stat(dropinPath); err == nil && !os.IsNotExist(err) {
+				files.Insert(dropinPath)
+			}
+		}
+	}
+
+	return files
+}
+
+// Extracts all unique directories from a given Ignition 2 config.
+func getFilePathsFromIgn2Config(ignConfig ign2types.Config, systemdPath string) sets.String {
+	files := sets.NewString()
+
+	// Get all the file paths from the ignition config
+	for _, ignFile := range ignConfig.Storage.Files {
+		if _, err := os.Stat(ignFile.Path); err == nil && !os.IsNotExist(err) {
+			files.Insert(ignFile.Path)
+		}
+	}
+
+	// Get all the file paths for systemd dropins from the ignition config
+	for _, unit := range ignConfig.Systemd.Units {
+		unitPath := getIgn2SystemdUnitPath(systemdPath, unit)
+		if _, err := os.Stat(unitPath); err == nil && !os.IsNotExist(err) {
+			files.Insert(unitPath)
+		}
+
+		for _, dropin := range unit.Dropins {
+			dropinPath := getIgn2SystemdDropinPath(systemdPath, unit, dropin)
+			if _, err := os.Stat(dropinPath); err == nil && !os.IsNotExist(err) {
+				files.Insert(dropinPath)
+			}
+		}
+	}
+
+	return files
+}
+
+// Gets the directories for all the MachineConfig file paths while
+// deduplicating them.
+func getDirPathsFromFilePaths(filePaths sets.String) []string {
+	dirPaths := sets.NewString()
+
+	for filePath := range filePaths {
+		dirPaths.Insert(filepath.Dir(filePath))
+	}
+
+	return dirPaths.List()
+}
--- a/pkg/daemon/config_drift_monitor_test.go
+++ b/pkg/daemon/config_drift_monitor_test.go
@@ -1,0 +1,430 @@
+package daemon
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/uuid"
+)
+
+func TestConfigDriftMonitor(t *testing.T) {
+	// Our errors. Their contents don't matter because we're not basing our
+	// assertions off of their contents. Instead, we're more concerned about
+	// their types.
+	fileErr := configDriftErr(fileConfigDriftErr(fmt.Errorf("file error")))
+	unitErr := configDriftErr(unitConfigDriftErr(fmt.Errorf("unit error")))
+
+	// Filesystem Mutators
+	// These are closures to avoid namespace collisions and pollution since
+	// they're not useful outside of this test.
+	changeFileContent := func(path string) error {
+		return ioutil.WriteFile(path, []byte("notthecontents"), defaultFilePermissions)
+	}
+
+	touchFile := func(path string) error {
+		now := time.Now()
+		return os.Chtimes(path, now.Local(), now.Local())
+	}
+
+	renameFile := func(path string) error {
+		return os.Rename(path, path+".old")
+	}
+
+	overwriteFile := func(path string) error {
+		tmpPath := filepath.Join(filepath.Dir(path), "overwritten")
+		if err := writeFileAtomicallyWithDefaults(tmpPath, []byte("notthecontents")); err != nil {
+			return err
+		}
+
+		return os.Rename(tmpPath, path)
+	}
+
+	chmodFile := func(path string) error {
+		return os.Chmod(path, 0755)
+	}
+
+	// The general idea for this test is as follows:
+	// 1. We create a temporary directory.
+	// 2. For each test case, we create an Ignition config as usual.
+	// 3. Apply the temp dir to paths in the testcase Ignition config.
+	// 4. Write the files to the temporary directory.
+	// 5. Start the Config Drift Monitor and supply a mock ondrift function.
+	// 6. Mutate the files on disk with our mutate func.
+	// 7. Wait a few milliseconds for the watcher to catch up.
+	// 8. Assert that we have config drift.
+	// 9. Shut down the monitor and move on.
+	//
+	// Note: We try to run the testcases in parallel for speed with each one
+	// mutating its own temp directory.
+	testCases := []configDriftMonitorTestCase{
+		// Ignition File
+		// These target the file called /etc/a-config-file defined by the test
+		// fixture.
+		{
+			name:        "ign file content drift",
+			expectedErr: fileErr,
+			mutateFile:  changeFileContent,
+		},
+		{
+			name:       "ign file touch",
+			mutateFile: touchFile,
+		},
+		{
+			name:        "ign file rename",
+			expectedErr: fileErr,
+			mutateFile:  renameFile,
+		},
+		{
+			name:        "ign file delete",
+			expectedErr: fileErr,
+			mutateFile:  os.Remove,
+		},
+		{
+			name:        "ign file overwrite",
+			expectedErr: fileErr,
+			mutateFile:  overwriteFile,
+		},
+		{
+			name:        "ign file chmod",
+			expectedErr: fileErr,
+			mutateFile:  chmodFile,
+		},
+		// Systemd Unit tests
+		// These target the systemd unit files in the test fixture:
+		// /etc/systemd/system/unittest.service
+		{
+			name:        "ign unit content drift",
+			expectedErr: unitErr,
+			mutateUnit:  changeFileContent,
+		},
+		{
+			name:       "ign unit touch",
+			mutateUnit: touchFile,
+		},
+		{
+			name:        "ign unit rename",
+			expectedErr: unitErr,
+			mutateUnit:  renameFile,
+		},
+		{
+			name:        "ign unit delete",
+			expectedErr: unitErr,
+			mutateUnit:  os.Remove,
+		},
+		{
+			name:        "ign unit overwrite",
+			expectedErr: unitErr,
+			mutateUnit:  overwriteFile,
+		},
+		{
+			name:        "ign unit chmod",
+			expectedErr: unitErr,
+			mutateUnit:  chmodFile,
+		},
+		// Systemd Dropin tests
+		// These target the dropin files belonging to the unittest systemd unit in
+		// the test fixture: /etc/systemd/system/unittest.service.d/10-unittest-service.conf
+		{
+			name:         "ign dropin content drift",
+			expectedErr:  unitErr,
+			mutateDropin: changeFileContent,
+		},
+		{
+			name:         "ign dropin touch",
+			mutateDropin: touchFile,
+		},
+		{
+			name:         "ign dropin rename",
+			expectedErr:  unitErr,
+			mutateDropin: renameFile,
+		},
+		{
+			name:         "ign dropin delete",
+			expectedErr:  unitErr,
+			mutateDropin: os.Remove,
+		},
+		{
+			name:         "ign dropin overwrite",
+			expectedErr:  unitErr,
+			mutateDropin: overwriteFile,
+		},
+		{
+			name:         "ign dropin chmod",
+			expectedErr:  unitErr,
+			mutateDropin: chmodFile,
+		},
+	}
+
+	for _, testCase := range testCases {
+		testCase := testCase
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			tmpDir := t.TempDir()
+			defer os.RemoveAll(tmpDir)
+
+			testCase.tmpDir = tmpDir
+			testCase.systemdPath = filepath.Join(tmpDir, pathSystemd)
+
+			testCase.run(t)
+		})
+	}
+}
+
+// Holds a testcase and its associated helper funcs
+type configDriftMonitorTestCase struct {
+	// Name of the test case
+	name string
+	// The expected error, if any
+	expectedErr error
+	// The tmpdir for the test case (assigned at runtime)
+	tmpDir string
+	// The systemdroot for the test case (assigned at runtime)
+	systemdPath string
+	// Only one of these may be used per testcase:
+	// The mutation to apply to the Ignition file
+	mutateFile func(string) error
+	// The mutation to apply to the systemd unit file
+	mutateUnit func(string) error
+	// The mutation to apply to the systemd dropin file
+	mutateDropin func(string) error
+}
+
+// Runs the test case
+func (tc configDriftMonitorTestCase) run(t *testing.T) {
+	// Get our test fixtures
+	ignConfig, mc := tc.getFixtures(t)
+
+	// Create our unexpected error channel
+	errChan := make(chan error, 5)
+
+	// Create a new instance of the Config Drift Monitor
+	cdm := NewConfigDriftMonitor()
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	// Create a goroutine to listen on the Done() method.
+	doneCalled := false
+	go func() {
+		defer wg.Done()
+		select {
+		case <-cdm.Done():
+			doneCalled = true
+			return
+		}
+	}()
+
+	// Create a goroutine to listen on the supplied error channel.
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case err, ok := <-errChan:
+				// Closing the channel causes a nil value to be sent. So we
+				// disambiguate a nil error object being sent from the channel closing.
+				if err != nil {
+					t.Errorf("unexpected error via error channel: %s", err)
+				}
+
+				// Only break out of the loop if the error channel is closed.
+				if !ok {
+					return
+				}
+			}
+		}
+	}()
+
+	onDriftCalled := false
+
+	// To listen on when
+	onDriftChan := make(chan struct{})
+
+	// Configure the config drift monitor
+	opts := ConfigDriftMonitorOpts{
+		ErrChan:       errChan,
+		SystemdPath:   tc.systemdPath,
+		MachineConfig: mc,
+		OnDrift: func(err error) {
+			go func() {
+				onDriftChan <- struct{}{}
+			}()
+			onDriftCalled = true
+			tc.onDriftFunc(t, err)
+		},
+	}
+
+	// Start the config drift monitor
+	require.Nil(t, cdm.Start(opts))
+
+	// Ensure that it's running
+	assert.True(t, cdm.IsRunning())
+
+	// Mutate the filesystem
+	require.Nil(t, tc.mutate(ignConfig))
+
+	timeout := 100 * time.Millisecond
+	start := time.Now()
+
+	// Give the watcher time to fire or time out if it doesn't
+	select {
+	case <-onDriftChan:
+		t.Logf("Took %v to fire", time.Since(start))
+	case <-time.After(timeout):
+		if tc.expectedErr != nil {
+			t.Errorf("expected onDrift to be called, but timed out after: %v", timeout)
+		}
+	}
+
+	// Stop the config drift monitor
+	cdm.Stop()
+
+	// Ensure that it's no longer running
+	assert.False(t, cdm.IsRunning())
+
+	// Closing our errChan causes the monitoring goroutine to stop.
+	close(errChan)
+
+	// Wait for our listener goroutines to shut down
+	wg.Wait()
+
+	// Ensure that our done channel was called
+	assert.True(t, doneCalled, "expected to receive a done event")
+
+	// If we expect an error, make sure onDrift was called. Otherwise, make sure
+	// it wasn't called.
+
+	if tc.expectedErr == nil {
+		assert.False(t, onDriftCalled, "expected onDrift not to be called")
+	} else {
+		assert.True(t, onDriftCalled, "expected onDrift to be called")
+	}
+}
+
+// Creates the Ignition Config test fixture
+func (tc configDriftMonitorTestCase) getIgnConfig() ign3types.Config {
+	return ign3types.Config{
+		Ignition: ign3types.Ignition{
+			Version: ign3types.MaxVersion.String(),
+		},
+		Storage: ign3types.Storage{
+			Files: []ign3types.File{
+				helpers.CreateIgn3File("/etc/a-config-file", "data:,theafilecontents", int(defaultFilePermissions)),
+			},
+		},
+		Systemd: ign3types.Systemd{
+			Units: []ign3types.Unit{
+				{
+					Name:     "unittest.service",
+					Contents: helpers.StrToPtr("unittest-unit-contents"),
+					Dropins: []ign3types.Dropin{
+						{
+							Contents: helpers.StrToPtr("unittest-service-contents"),
+							Name:     "10-unittest-service.conf",
+						},
+						{
+							Name: "20-unittest-service.conf",
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// Determines which file to mutate and runs the appropriate mutator.
+func (tc configDriftMonitorTestCase) mutate(ignConfig ign3types.Config) error {
+	if tc.mutateFile != nil {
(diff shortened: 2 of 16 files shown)


pkg/daemon/daemon.go
```diff
@@ -1093,14 +1167,12 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
 
-	if _, err := os.Stat(constants.MachineConfigDaemonForceFile); err != nil {
-		if err := dn.validateOnDiskState(expectedConfig); err != nil {
-			return fmt.Errorf("unexpected on-disk state validating against %s: %v", expectedConfig.GetName(), err)
-		}
-		glog.Info("Validated on-disk state")
-	} else {
-		glog.Infof("Skipping on-disk validation; %s present", constants.MachineConfigDaemonForceFile)
+	if skipped, err := dn.maybeValidateOnDiskState(expectedConfig); skipped {
+		// We skipped validation, so immediately trigger the update
```
yuqi-zhang, line 1171: Ah so this answers my question from the demo, I see you adapted the original behaviour of always triggering an update here. Reading the code it appears that only on checkStateOnFirstRun do we always force an upgrade if a forcefile exists, but going through the logic, we appear to "apply" a change even if current==desired, and then... write the desired configuration mostly as it stands without checking for diff (e.g. re-write all the files, comparing the OSURL of the desired NOT with the current but with the BOOTED url). Then... reboot because the postConfigChangeAction honours the forcefile still.

So then we actually CAN apply a no-diff update (I didn't think this was the case, but it appears to be) and barring some subfields (e.g. kargs since that works on a diff check system) this mostly works out (at least I think so based on my understanding of the logic).

Hmm, regardless, we seem to be somewhat inconsistent here. Happy to discuss this further, I could be completely mistaken as well on some of my assumptions.

pkg/daemon/daemon.go
```diff
@@ -1120,6 +1192,54 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
+	if err == nil {
+		return true
+	}
+
+	// We know the file doesn't exist
+	if goErrors.Is(err, os.ErrNotExist) {
+		return false
+	}
+
+	// We couldn't stat the file for other reasons, assume it doesn't exist
```
yuqi-zhang, line 1208: Hmm, maybe we should at least log something here. I guess we technically never checked for other errors, so this is fine as is as well

pkg/daemon/daemon.go
```diff
@@ -1120,6 +1192,54 @@ func (dn *Daemon) checkStateOnFirstRun() error {
...
+		return true, nil
+	}
+
+	// Don't skip validation
+	err := dn.runValidateOnDiskState(expectedConfig)
+
+	return false, err
+}
+
+func (dn *Daemon) runValidateOnDiskState(expectedConfig *mcfgv1.MachineConfig) error {
```
yuqi-zhang, line 1226: I see you have many layering of this function, namely:
```
validateCurrentOnDiskStateInSyncNode
validateCurrentOnDiskState
maybeValidateOnDiskState
runValidateOnDiskState
validateOnDiskState
```
Do you think it would be possible to simplify that down a bit? Maybe to merge some of the functions and do some ifs inside? It feels a bit confusing to follow so many different layers of validation.

pkg/daemon/daemon.go
```diff
@@ -689,6 +708,61 @@ func (dn *Daemon) applySSHAccessedAnnotation() error {
...
+					dn.updateErrorState(err)
+				} else {
+					exitCh <- err
+				}
+			}
+		}
+	}
+}
+
+// Not sure if this is actually needed or if I can use getCurrentConfigOnDisk instead?
```
yuqi-zhang, line 734: Technically possible to have difference. My interpretation is that the on-disk version should always be correct, although in practice it doesn't seem to be that rigorous.

pkg/daemon/daemon.go
```diff
@@ -460,6 +465,18 @@ func (dn *Daemon) syncNode(key string) error {
 		return nil
 	}
 
+	// This allows skip behavior based upon the presence of
+	// /run/machine-config-daemon-force, though I don't know if that is desirable
+	// here.
+	if err := dn.validateCurrentOnDiskStateInSyncNode(); err != nil {
```
yuqi-zhang, line 471: 2 thoughts:

1. this technically double validates if dn.booting, since checkStateOnFirstRun() also validates
2. this technically differs from checkStateOnFirstRun() since that forces an upgrade (even if current==desired) which this does not

Random thought: maybe we should consolidate the checkStateOnFirstRun loop? Perhaps we want to be a bit more fancy during sync checks than just this.

yuqi-zhang (commented):
Overall I think the direction looks good, although I have a few concerns, namely on what we do in the non-booting sync loops and how we handle "no-change forcefile updates". More details in comments below, and please do let me know if I am misunderstanding any of this, I am not sure if I have the details correct

pkg/daemon/daemon.go
cheesesashimi: We should definitely consolidate / simplify the `checkStateOnFirstRun` loop, though I don't think this PR is the best place to do that.

pkg/daemon/daemon.go
```diff
@@ -1325,44 +1385,83 @@ func (dn *Daemon) triggerUpdateWithMachineConfig(currentConfig, desiredConfig *m
 	return dn.update(currentConfig, desiredConfig)
 }
 
+// maybeValidateOnDiskState will conditionally run validateOnDiskState
+// depending on the presence of the forcefile and provides an indication that
+// validation was skipped. If the forcefile is present, this will *not* run
+// validateOnDiskState.
+func (dn *Daemon) maybeValidateOnDiskState(expectedConfig *mcfgv1.MachineConfig) (bool, error) {
```
kikisdeliveryservice, line 1392: Can this be consolidated into validateOnDiskState or elsewhere? This func seems confusing - the bool is false if you do validateOnDiskState? Seems counterintuitive but also really not clear what's happening would prefer to just have this if check condensed as Jerry requested elsewhere.

pkg/daemon/daemon.go
```diff
@@ -1325,44 +1385,83 @@ func (dn *Daemon) triggerUpdateWithMachineConfig(currentConfig, desiredConfig *m
...
 		}
-		return nil
-	default:
-		return errors.Errorf("unexpected type for ignition config: %v", typedConfig)
 	}
+
+	start := time.Now()
+
+	if err := validateFunc(); err != nil {
+		wErr := fmt.Errorf("on-disk state mismatch: %v", err)
```
kikisdeliveryservice, line 1454: what does wErr signify?

pkg/daemon/daemon.go
```diff
@@ -1325,44 +1385,83 @@ func (dn *Daemon) triggerUpdateWithMachineConfig(currentConfig, desiredConfig *m
...
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "OnDiskStateValidationFailed", wErr.Error())
+		return wErr
+	}
+
+	end := time.Since(start)
+
+	glog.Infof("Validated on-disk state (took %s)", end)
+	dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "OnDiskStateValidationSuccess", "On-disk configuration state validated to equal %s (took %s).", currentConfig.GetName(), end)
```
kikisdeliveryservice, line 1462: For general eventing: How often is this check running, in a long lived cluster I don't think we want to fill event log with success if the check is happening often.

On failure yes, but on success? Maybe not.

pkg/daemon/daemon.go
```diff
@@ -1641,3 +1740,21 @@ func (dn *Daemon) getControlPlaneTopology() configv1.TopologyMode {
 		return configv1.HighlyAvailableTopologyMode
 	}
 }
+
+// forceFileExists determines if /run/machine-config-daemon-force is present.
```
kikisdeliveryservice, line 1744: This is only called by your maybe.. function, let's consolidate this as well.

pkg/daemon/daemon.go
```diff
@@ -123,6 +124,9 @@ type Daemon struct {
 	loggerSupportsJournal bool
 
 	drainer *drain.Helper
+
+	// Ensures we only have a single validateOnDiskState running at a time.
+	diskValidatorLock sync.Mutex
```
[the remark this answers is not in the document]
cheesesashimi: `dn.booting` only covers the bootstrap case. My thought behind the mutex was to ensure that the periodic check doesn't run in the middle of an update.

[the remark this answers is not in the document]
sinnykumari: I think this runs whenever MCD pod restarts or recreated, otherwise validateOnDiskState() would never get called today when we perform update/upgrade and we would have never got any bugs filed around it. @yuqi-zhang what do you think?

cheesesashimi:
Here's where I am:
- I reverted `validateOnDiskState` back to it's original scope of just validating the on-disk state with the addition of the `diskValidatorLock`. It will not emit any events on success or failure. That responsibility will fall on the caller.
- I'm decoupling the forcefile detection logic from `validateOnDiskState` to make what's happening more explicit. That means no more `maybeValidateOnDiskState`.
- I will keep the `forceFileExists` function because I think it's a bit cleaner and easier to read though I'm open to arguments against.
- To be more explicit, I'm creating a function called `checkForConfigDrift` which runs `validateOnDiskState` against the current on-disk config and will only emit an event when it finds an error. However, it will log whenever it runs.
- I've also added a `runPeriodicConfigDriftCheck` which executes the aforementioned function but emits events upon error with the specific context of being run periodically as opposed to on-demand.

To respond to @sinnykumari's comments:
- Using `dn.booting` would only prevent the periodic check from running when the MCD is still booting. It won't prevent the periodic check from running while an update is in progress. If the MCD is performing an update and the periodic check runs, it could cause unexpected behavior.
- I'm not sure if the additional logic found in https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/daemon.go#L1071-L1104 is applicable for the context of checking for drift relative to the current on-disk configuration in both the periodic case as well as prior to applying an update.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
yuqi-zhang: > I think this runs whenever MCD pod restarts
So this part we are not sure of. I think we were able to test a scenario where a machineconfig triggered update does not reset dn.booting flag, so the rebooted run did not re-check the state. Also now that we have non-rebooting updates, it becomes a bit harder to just validate on `booting`. Zack and I were thinking of de-coupling validation and booting in general, and maybe even move more items out of the check (see https://issues.redhat.com/browse/MCO-86)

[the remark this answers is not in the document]
cheesesashimi: We only call `validateOnDiskState()` whenever the MCD restarts (inside `checkStateOnFirstRun`). After that, we [set `dn.booting` to false and it never gets run again until the MCD restarts](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/daemon.go#L445-L461). I remember discussing this with @yuqi-zhang and we were both surprised that it only runs there.

test/helpers/mcd_config_drift.go
```diff
@@ -0,0 +1,299 @@
...
+	mcdPod, err := MCDForNode(cs, &targetNode)
+	require.Nil(t, err)
+
+	// Check that we have our kernel args pre-recovery
+	assertHasKernelArgs(t, cs, targetNode, mcadd.Spec.KernelArguments)
+
+	// Get our kernel args for future verification
+	kargs := getKernelArgs(t, cs, targetNode)
+
+	t.Logf("Target MCD Pod: %s. To stream pod logs, copy / paste the below:", mcdPod.Name)
```
kikisdeliveryservice, line 85:
```suggestion
	t.Logf("Target MCD Pod: %s", mcdPod.Name)
```

test/helpers/mcd_config_drift.go
```diff
@@ -0,0 +1,299 @@
...
+	// Verify that the expected error message is in the console logs.
+	t.Logf("Verifying pod logs")
+
+	logs, err := cs.Pods(mcdPod.Namespace).GetLogs(mcdPod.Name, &corev1.PodLogOptions{
+		Container: "machine-config-daemon",
+	}).DoRaw(context.TODO())
+	require.Nil(t, err)
+
+	if !strings.Contains(string(logs), logEntry) {
+		t.Fatalf("expected a content mismatch!")
```
kikisdeliveryservice, line 187:
```suggestion
		t.Fatalf("expected a content mismatch error")
```

pkg/daemon/daemon.go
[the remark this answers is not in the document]
kikisdeliveryservice: This seems correct bc it's only set to true here: https://github.com/openshift/machine-config-operator/blob/7c1ac8b51423448397bda3f349f56f6e94261b64/pkg/daemon/daemon.go#L251

pkg/daemon/daemon.go
```diff
@@ -247,15 +254,17 @@ func New(
...
+		booting:                  true,
+		os:                       hostos,
+		NodeUpdaterClient:        nodeUpdaterClient,
+		bootedOSImageURL:         osImageURL,
+		bootID:                   bootID,
+		exitCh:                   exitCh,
+		currentConfigPath:        currentConfigPath,
+		loggerSupportsJournal:    loggerSupportsJournal,
+		diskValidatorLock:        sync.Mutex{},
+		configDriftCheckInterval: 10 * time.Second,
```
kikisdeliveryservice, line 267: (just flagging this to remember to change remove once updated near merge time)

test/helpers/mcd_config_drift.go
```diff
@@ -0,0 +1,299 @@
...
+		return m.Status.DegradedMachineCount == 0 &&
+			allMCPConditionsTrue(trueConditions, m) &&
+			allMCPConditionsFalse(falseConditions, m)
+	})
+}
+
+func assertNodeReachesState(t *testing.T, cs *framework.ClientSet, target corev1.Node, stateFunc func(corev1.Node) bool) {
+	t.Helper()
+
+	maxWait := 20 * time.Minute
```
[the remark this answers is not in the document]
cheesesashimi: I'm not sure why I did that... I'll set it to 5 minutes.

test/helpers/mcd_config_drift.go
cheesesashimi: Beyond my current development needs, we don't. I'll take this out.

cheesesashimi:
Update:
- I adopted pretty much all of the changes @kikisdeliveryservice requested.
- I also pulled this PR out of draft so all the tests can run.
- I realized that I introduced an import cycle, so I fixed that. However, I'm not completely happy with my fix, so I'm completely open to feedback and suggestions.

TODO:
- Get the import cycle fix into a state I and others are happy with.
- Address lint / test failures.
- Squash commits.

test/sharedtests/mcd_config_drift.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func assertNodeReachesState(t *testing.T, cs *framework.ClientSet, target corev1.Node, stateFunc func(corev1.Node) bool) {
+	t.Helper()
+
+	maxWait := 5 * time.Minute
+
+	startTime := time.Now()
+
+	var lastErr error = nil
```
kikisdeliveryservice, line 240:
```suggestion
	var lastErr error
```

test/sharedtests/mcd_config_drift.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func assertPoolReachesState(t *testing.T, cs *framework.ClientSet, target mcfgv1.MachineConfigPool, stateFunc func(mcfgv1.MachineConfigPool) bool) {
+	t.Helper()
+
+	maxWait := 5 * time.Minute
+
+	startTime := time.Now()
+
+	var lastErr error = nil
```
kikisdeliveryservice, line 277:
```suggestion
	var lastErr error
```

pkg/daemon/daemon.go
```diff
@@ -474,6 +495,23 @@ func (dn *Daemon) syncNode(key string) error {
...
+	// the forcefile: /run/machine-config-daemon-force.
+	if forceFileExists() {
+		glog.Infof("Skipping preflight config drift check; %s present", constants.MachineConfigDaemonForceFile)
+		return nil
+	}
+
+	err := dn.checkForConfigDrift()
+
+	if err != nil {
+		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeWarning, "PreflightConfigDriftCheckFailed", err.Error())
```
sinnykumari, line 509: An info message would be good so that we can see that in daemon log as well.

pkg/daemon/daemon.go
```diff
@@ -466,6 +477,16 @@ func (dn *Daemon) syncNode(key string) error {
...
+		if err := dn.runPreflightConfigDriftCheck(); err != nil {
+			return err
+		}
+
+		// Make sure we don't attempt to validate on disk contents while the update
+		// is running
+		dn.diskValidatorLock.Lock()
+		defer dn.diskValidatorLock.Unlock()
+
 		if err := dn.triggerUpdateWithMachineConfig(current, desired); err != nil {
```
sinnykumari, line 490: This function gets called at multiple places. wondering if it will make more sense to do mutex lock inside triggerUpdateWithMachineConfig() function definition.

sinnykumari (commented):
overall logic looks LGTM. Few minor comments.
Would like to test this out locally once.

pkg/daemon/daemon.go
```diff
@@ -689,6 +729,55 @@ func (dn *Daemon) applySSHAccessedAnnotation() error {
...
+			}
+		}
+	}
+}
+
+func (dn *Daemon) runPeriodicConfigDriftCheck() error {
+	glog.Info("Running periodic config drift check")
+	err := dn.checkForConfigDrift()
+	if err != nil {
+		// TODO: Use better error handling here
```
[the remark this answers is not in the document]
cheesesashimi: I'll tidy this up with a custom error type instead of doing string matching.

pkg/daemon/daemon.go
```diff
@@ -146,13 +150,19 @@ const (
 	// the MCC/MCO will time out.  We don't want to spam our logs with the same
 	// error.
 	updateDelay = 5 * time.Second
+
 	// maxUpdateBackoff is the maximum time to react to a change as we back off
 	// in the face of errors.
 	maxUpdateBackoff = 60 * time.Second
+
+	// How frequently to check for config drift
+	configDriftCheckInterval = 60 * time.Minute
```
kikisdeliveryservice, line 159: @sinnykumari are you ok with this timing? I thought you wanted longer?

sinnykumari: not really, this needs discussion within team so that we all agree on a particular interval. Let's do this in one of the meeting.

pkg/daemon/daemon.go
cheesesashimi: I tested this and discovered that it can dead-lock under certain conditions. If we acquire the lock in `dn.triggerUpdateWithMachineConfig` and defer the unlock, we won't unlock until the call to `dn.update` returns.

When we call `dn.update`, we eventually get to `dn.performPostConfigChangeAction`. However, under certain conditions, `dn.performPostConfigChangeAction` will call [`dn.triggerUpdateWithMachineConfig` again](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L161).

However, since we haven't returned from the `dn.update` call in `dn.triggerUpdateWithMachineConfig`, the lock never gets released and it hangs on the second call waiting for the lock to become available.

cheesesashimi: In the other places where `dn.triggerUpdateWithMachineConfig` is called, I don't think locking / unlocking matters as much because those are the booting or bootstrapping phases. Even though the periodic config drift checker is started by that time, that just starts the goroutine; it doesn't actually run a scan until after the check interval has been satisfied. I could add some guards to skip the scan if we're bootstrapping or booting but I don't think that's necessary.

cheesesashimi: I added the guards in `dn.runPeriodicConfigDriftCheck` to check if we're booting or if we're bootstrapping. If either of those cases is met, we'll skip the config drift check.

cheesesashimi:
I added some minor improvements and guards to the test helpers:
- Instead of shelling out to `oc` for adding / removing labels from nodes, I make the changes via an API call instead since we already have the API available.
- I also added some checks around ensuring that we have an `oc` command in our `$PATH` to shell out to (since I know that was an issue for @mkenigs when he first tried to run some of the e2e tests).
- I added some checking around Kubeconfigs. In short, if one does `NewClientSet("/path/to/kubeconfig")`, then one tries to call `ExecCmdOnNode` with that clientset, it won't know which Kubeconfig should be used since `ExecCmdOnNode` depends on having `$KUBECONFIG` set. Now, we get the value from the clientset and explicitly inject it into the environment where `oc` runs.

I'm happy to move those changes to a separate PR to be merged after this one, if desired.

cgwalters:
One option is to use e.g. [fanotify](https://man7.org/linux/man-pages/man7/fanotify.7.html) to recursively monitor `/etc`.  I think the cost would be pretty low for that.

Another option is to increase efficiency a bit here by at least doing timestamp checking on the directories.

Yet another option is to edge trigger off "likely drift sources" - e.g. when someone logs into the node (via ssh) start monitoring more frequently.

cheesesashimi:
I pushed up a first pass which uses fsnotify (which provides a wrapper around fanotify, et. al.). Bonus: We were already indirectly dependent on it, so it was already vendored. 

I have not yet wired it into the MCD as I want to get feedback on that portion of the code first.

cheesesashimi:
I think I have this into a sensible implementation with fsnotify. In general, here's the approach I used as well as some overall notes:
- Rather than grafting a bunch of methods onto the MCD Daemon struct, I created another struct to hold the Config Drift Monitor. The Config Drift Monitor has its own unit-test suite. A caveat I ran into is that they don't run well in OS X, so I have to use a container. Those who are running Linux should be able to run them with no issues.
- The Config Drift Monitor is intended to be run in a separate goroutine and I've strategically placed the start function in a few places where it will only be started once the MCD is out of bootstrap and not booting. I've made the start and stop functions in the MCD idempotent so there is no harm in calling either more than once.
- To perform a config update, we have to shut down the Config Drift Monitor as writing the config to disk will be perceived as a config "drift". Once the update is complete, the Config Drift Monitor can be started again using the current on-disk config (assuming a reboot does not occur).
- The Config Drift Monitor reads the MachineConfig and extracts the paths of the files as well as the Systemd Units / Dropins. Next, I get their parent directory names and wire up fsnotify to their parents. This is because under the hood, fsnotify is currently using inotify instead of fanotify and has a finite limit of paths it can watch. Furthermore, a limitation of inotify is that it cannot recurse into subdirectories. So if you wire it up to monitor `/etc/`, for example, it won't report events for files under `/etc/system/`. However, files placed in `/etc` (e.g., `/etc/passwd`) would cause events to be fired. However, I ensure that we only check for config drift when files declared by the MachineConfig are mutated. If it's not declared in the MachineConfig, it's ignored. 
- When we detect a potential drift event, `validateOnDiskState` gets called and either confirms or denies that a config drift actually happened. From simple observation, this mechanism is very quick to degrade the node whenever it detects a config drift. It's worth mentioning that I refactored `dn.validateOnDiskState` very slightly to allow easy reuse of the code as well as make testing a bit easier.
- I created custom error types for config drift events. For now, they're very simple wrappers around the preexisting errors. Additional context will be added to those errors in the future as that is outside the scope of this PR.

pkg/daemon/configdriftmonitor.go
```diff
@@ -0,0 +1,376 @@
+package daemon
+
```
kikisdeliveryservice, line 2: not urgent at all but can you snake case this file name?

cheesesashimi:
This is now in a state I'm happy with and is ready for a final review (modulo passing CI).

kikisdeliveryservice:
any idea why this is failing?
/test e2e-gcp-op-single-node

cheesesashimi:
I have a couple of ideas though no solid leads:
- [This failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1465801584679063552) is addressed by my most recent commit. The short version is that I foolishly introduced an unnecessary change which required the control plane to still be available while the node rebooted. That's fine in an HA cluster, but not in SNO; hence this particular failure.
- [This one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466106627542224896) looks like something went awry during installation, though I don't think it's MCO / MCD-related.
- [This one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264) is puzzling. It looks like a timeout was hit, but judging from both the [e2e test output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/e2e-gcp-op-single-node/test/build-log.txt) as well as the [ci-operator output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/ci-operator.log) Prow aborted the run for some reason. Looking at the e2e test logs shows that is indeed the case, but it did so after the `TestKernelArguments` and `TestKernelType` tests completed; 28 minutes after the e2e test suite began running. The [event log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/e2e-gcp-op-single-node/gather-must-gather/artifacts/event-filter.html) has some interesting events around a transient `RenderConfigFailed` event, but that lasted for several minutes during the OpenShift installation phase, so I'm pretty sure it's unrelated. Also, I don't see any other MCO / MCD degradation events.

pkg/daemon/config_drift_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+		c.cdw = nil
+		c.stopCh <- struct{}{}
+	}
+}
+
+// Creates a new config drift watcher for a given machineconfig and registers
+// a callback for when a drift occurs.
+func newConfigDriftWatcher(opts ConfigDriftMonitorOpts) (*configDriftWatcher, error) {
+	if opts.OnDrift == nil {
+		return nil, fmt.Errorf("no ondrift function attached")
```
kikisdeliveryservice, line 139: partially feel like this phrasing "no ondrift function" isn't very clear

pkg/daemon/config_drift_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+	for filePath := range filePaths {
+		dirPaths.Insert(filepath.Dir(filePath))
+	}
+
+	return dirPaths.List()
+}
+
+// Computes the absolute path for a given systemd unit and dropin, given an
+// optional root path which defaults to /etc/systemd.
+func getIgn2SystemdPath(systemdPath string, unit ign2types.Unit, dropin ign2types.SystemdDropin) string {
```
kikisdeliveryservice, line 364: are these funcs only called but the getfilepaths? if so I think we could just combine them into those funcs since the code is trivial.

pkg/daemon/daemon.go
```diff
@@ -474,6 +488,32 @@ func (dn *Daemon) syncNode(key string) error {
 	return nil
 }
 
+func (dn *Daemon) runPreflightConfigDriftCheck() error {
```
kikisdeliveryservice, line 491: add doc string to this func

pkg/daemon/daemon.go
```diff
@@ -689,6 +729,70 @@ func (dn *Daemon) applySSHAccessedAnnotation() error {
 	return nil
 }
 
+func (dn *Daemon) onConfigDrift(err error) {
```
kikisdeliveryservice, line 732: same add doc since if feels mysterious until i search/scroll down

test/sharedtests/mcd_config_drift.go
```diff
@@ -0,0 +1,396 @@
+package sharedtests
```
kikisdeliveryservice, line 1: Request: retitle dir to use "e2e" like e2e-shared-tests also use convention of _test.go in the file name.  Current naming is unclear. 

Might also be nice to add a README to your dir to specific what you mean by shared tests- these are shared between e2e and single-node and that's the purpose of the dir (correct me if im wrong on that ;) )

pkg/daemon/config_drift_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+func (c *configDriftWatcher) initialize() error {
+	var err error
+
+	// Initialize the fsnotify watcher
+	c.watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("could not start watcher: %w", err)
+	}
+
+	glog.V(4).Infof("Initializing Config Drift Monitor")
```
kikisdeliveryservice, line 178: are there any glog.Info level config drift monitor running messages? i see starting, i see initializing, i see stopping but (and my search might be bad here) is there anything that just explicitly says yep it's running now:"

kikisdeliveryservice (commented):
A couple comments/questions but also I think your PR description is now stale?

kikisdeliveryservice:
> I have a couple of ideas though no solid leads:
> 
>     * [This failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1465801584679063552) is addressed by my most recent commit. The short version is that I foolishly introduced an unnecessary change which required the control plane to still be available while the node rebooted. That's fine in an HA cluster, but not in SNO; hence this particular failure.
> 
>     * [This one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466106627542224896) looks like something went awry during installation, though I don't think it's MCO / MCD-related.
> 
>     * [This one](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264) is puzzling. It looks like a timeout was hit, but judging from both the [e2e test output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/e2e-gcp-op-single-node/test/build-log.txt) as well as the [ci-operator output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/ci-operator.log) Prow aborted the run for some reason. Looking at the e2e test logs shows that is indeed the case, but it did so after the `TestKernelArguments` and `TestKernelType` tests completed; 28 minutes after the e2e test suite began running. The [event log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op-single-node/1466128583431819264/artifacts/e2e-gcp-op-single-node/gather-must-gather/artifacts/event-filter.html) has some interesting events around a transient `RenderConfigFailed` event, but that lasted for several minutes during the OpenShift installation phase, so I'm pretty sure it's unrelated. Also, I don't see any other MCO / MCD degradation events.

Thanks for the details! That jobs seems pretty red generally, if it's failling tomorrow we can ask around about the status it's not required and in current state isn't a blocker.

pkg/daemon/config_drift_monitor.go
cheesesashimi: No, I use these in my tests. I was also planning to add them to `checkV2Units` and `checkV3Units`, respectively.

My intent is to avoid duplicating the `filepath.Join` calls in multiple places especially since we manipulate one of the values passed into `filepath.Join` (ign3types.Systemd.Name + ".d").

pkg/daemon/config_drift_monitor.go
cheesesashimi: No, but it would be a good idea to have them!

test/sharedtests/mcd_config_drift.go
cheesesashimi: You're correct! I was struggling with a good name for these, so that helps.

pkg/daemon/config_drift_monitor.go
cheesesashimi: I agree, I'll cogitate on this a bit.

test/sharedtests/mcd_config_drift.go
cheesesashimi: I can't use the `_test.go` convention in the filename, otherwise I get:

```console
go build github.com/openshift/machine-config-operator/test/e2e-shared-tests: no non-test Go files in /Users/zzlotnik/go/src/github.com/openshift/machine-config-operator/test/e2e-shared-tests
```

kikisdeliveryservice: ahhhhhh gotcha

pkg/daemon/config_drift_monitor.go
kikisdeliveryservice: i think i missed that, will relook :+1:

cheesesashimi:
@kikisdeliveryservice Your suggestions around organizing the shared test folder gave me an idea to provide easier hooks into creating and running a shared test in both a HA and single-node contexts. I put those changes in https://github.com/openshift/machine-config-operator/pull/2795/commits/b2f42c8b6eba57a2ae419befb8fa8bce33c50753. If that's too much for this PR, I'm fine with dropping that particular commit.

kikisdeliveryservice:
> @kikisdeliveryservice Your suggestions around organizing the shared test folder gave me an idea to provide easier hooks into creating and running a shared test in both a HA and single-node contexts. I put those changes in [b2f42c8](https://github.com/openshift/machine-config-operator/commit/b2f42c8b6eba57a2ae419befb8fa8bce33c50753). If that's too much for this PR, I'm fine with dropping that particular commit.

I'm not mad at the refactor :smile: , in fact it was more of what i expected for shared tests. Felt a little bulky/potentially drifty holding more code separately before :+1:

cheesesashimi:
While refactoring the Config Drift Monitor unit test suite in a similar vein to what I did to the shared e2e tests, I realized there was an uncovered edge case in the form of systemd unit files. So I fixed that, added some additional tests there, and added similar tests to the e2e suite as well.

cheesesashimi:
@sinnykumari @kikisdeliveryservice I think this is ready for final review.

test/e2e-shared-tests/README.md
```diff
@@ -0,0 +1,16 @@
+# e2e Shared Tests
+
+To reduce duplication across multiple e2e test suites, tests which can be
+reused across all MCO test suites should live within this package.
+
+## To Add A New Shared Test
```
kikisdeliveryservice, line 6: love  this!

test/e2e-shared-tests/helpers.go
```diff
@@ -0,0 +1,266 @@
...
+		return isDegraded && hasReason
+	})
+
+	mcdPod, err := helpers.MCDForNode(cs, &node)
+	require.Nil(t, err)
+
+	assertLogsContain(t, cs, mcdPod, logEntry)
+
+	// Assert that the MachineConfigPool eventually reaches a degraded state and has the config mismatch as the reason.
+	t.Log("Verifying MachineConfigPool becomes degraded")
```
kikisdeliveryservice, line 98: Request: when looking thru the logs we don't explicitly say config mismatch detected or anything referencing config,  [example logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-config-operator/2795/pull-ci-openshift-machine-config-operator-master-e2e-gcp-op/1466917418067562496/artifacts/e2e-gcp-op/test/build-log.txt). Can we add something around here which references config drift or config mismatch?

```suggestion
	t.Log("Verifying MachineConfigPool becomes degraded due to config mismatch")
```

kikisdeliveryservice:
As a followup to this PR, let's update our documentation and add something about recovery from your PR description, we'll also want some update to official OCP docs too. Both would include something about:

`To recover, either change the file contents back or oc debug node/<node-name> -- touch /host/run/machine-config-daemon-force. The latter will force the MCD to apply the update and will cause a reboot which may or may not be desirable.`

After that is done, we can update the error message to point to the documentation for recovery steps.

wdyt: @sinnykumari @cheesesashimi ?

test/e2e-shared-tests/helpers.go
cheesesashimi: For sure. Great idea!

cheesesashimi:
@kikisdeliveryservice I agree. I was planning on doing config drift error message improvements after this, so I can tackle adding a link to the docs as part of that.

kikisdeliveryservice:
@cheesesashimi I don't think you need to retest, as there are some ci issues at the moment in aws, seeing it specifically in the e2e-aws job rn

kikisdeliveryservice (approved):
I don't have any more comments and I think all of my previous ones were address, so leaving my approval. Great job, Zach!

Leaving to rest of team for their final reviews.

/assign @sinnykumari

cheesesashimi:
@sinnykumari made the observation that if multiple files drift at the same time, the Config Drift Monitor will only see the first file. This is because `validateOnDiskState` halts on the first error it finds. In other words, if a MachineConfig defines two files (`/etc/foo` and `/etc/bar`) and both files are mutated, `validateOnDiskState` will only report one of the two files in the console log and in the degradation event. While this should not block this PR (as this PR leverages the preexisting behavior of `validateOnDiskState`), it is something to be considered when config drift error messages are revamped.

In the future, maybe `validateOnDiskState` should validate the complete MachineConfig and present all errors found instead of halting on the first error it finds. As an implementation detail, we could use https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L35 as a method of error aggregation.

pkg/daemon/config_drift_monitor_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+func TestConfigDriftMonitor(t *testing.T) {
+	// Our errors. Their contents don't matter because we're not basing our
+	// assertions off of their contents. Instead, we're more concerned about
+	// their types.
+	fileErr := configDriftErr(fileConfigDriftErr(fmt.Errorf("file error")))
+	unitErr := configDriftErr(unitConfigDriftErr(fmt.Errorf("unit error")))
+
+	// Filesystem Mutators
+	// These don't have to be closures. I just wanted to avoid potential
```
sinnykumari, line 30: Can you re-frame the comment to not include "I" in sentence.

pkg/daemon/config_drift_monitor_test.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	overwriteFile := func(path string) error {
+		tmpPath := filepath.Join(filepath.Dir(path), "overwritten")
+		if err := writeFileAtomicallyWithDefaults(tmpPath, []byte("notthecontents")); err != nil {
+			return err
+		}
+
+		return os.Rename(tmpPath, path)
+
```
sinnykumari, line 52: nit: newline not needed here

pkg/daemon/config_drift_monitor_test.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	chmodFile := func(path string) error {
```
sinnykumari, line 55: Is this function for file or directory? Directory usually have permission 755 but a file would have 644

pkg/daemon/config_drift_monitor_test.go
```diff
@@ -0,0 +1,414 @@
...
+	// If the testcase asks for a unitConfigDriftErr, be sure we got one.
+	var uErr unitConfigDriftErr
+	if errors.As(tc.expectedErr, &uErr) {
+		assert.ErrorAs(t, err, &uErr)
+	}
+}
+
+func (tc configDriftMonitorTestCase) writeFileForTest(t *testing.T, path string, contents *string) {
+	t.Helper()
+	// I know this isn't how writeFiles actually does this, but I did it this way
```
sinnykumari, line 407: Same here. Try to reword comment without referring to "I" . This applies to other comments in the PR as well in case I missed some of them.

pkg/daemon/daemon.go
```diff
@@ -1334,34 +1446,8 @@ func (dn *Daemon) validateOnDiskState(currentConfig *mcfgv1.MachineConfig) error
...
-			return err
-		}
-		if err := checkV2Units(ignconfigi.(ign2types.Config).Systemd.Units); err != nil {
-			return err
-		}
-		return nil
-	default:
-		return errors.Errorf("unexpected type for ignition config: %v", typedConfig)
-	}
+	return validateOnDiskState(currentConfig, pathSystemd)
```
sinnykumari, line 1450: Instead of passing const value pathSystemd to all the places. Maybe move this const variable to https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/constants/constants.go

pkg/daemon/on_disk_validation.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		if err := checkFileContentsAndMode(path, []byte(unit.Contents), defaultFilePermissions); err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+// checkFiles validates the contents of  all the files in the
```
sinnykumari, line 142: instead of `checkFiles` , it can be `check3Files` to match comment with function.

pkg/daemon/on_disk_validation.go
```diff
@@ -0,0 +1,250 @@
...
+// target config.
+
```
sinnykumari, line 144: Newline can be skipped.

pkg/daemon/on_disk_validation.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+		}
+		if err := checkFileContentsAndMode(f.Path, contents.Data, mode); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// checkFiles validates the contents of all the files in the
```
sinnykumari, line 171: same, s/checkFiles/checkV2Files

sinnykumari (commented):
Few minor comments. Other than that looks great!

pkg/daemon/daemon.go
cheesesashimi: Yes. For easy cleanup, the Config Drift Monitor test suite creates a tempdir to write the Ignition files, systemd unit and dropin files to. To detect the systemd unit and dropin files, I modified `validateOnDiskState` (and some of its helpers) to accept a `systemdPath` argument instead of having `pathSystemd` hard-coded. That said, Config Drift Monitor, as well as `validateOnDiskState's` helpers will default to `pathSystemd` if an empty string is passed in.

pkg/daemon/config_drift_monitor_test.go
cheesesashimi: It's for a file. The test uses `defaultFilePermission` when it creates the file. This function just modifies the file permissions to something that is not 644 to cause the Config Drift Monitor to report that a config drift has occurred.

pkg/daemon/config_drift_monitor_test.go
cheesesashimi: Yes. In the future, I'll put "I" comments in the PR as a note to reviewers.

cheesesashimi:
The cause of the unit test failure is a timing issue caused by the `time.Sleep` statement in the Config Drift Monitor test suite.  The filesystem event took longer than the hard-coded value for the event to fire. I'll refactor the unit test to use a more event-driven approach.

pkg/daemon/config_drift_monitor_test.go
sinnykumari: aha, interesting. Makes sense

sinnykumari (approved):
This looks great and good to go in.

One small change before merging, can we add short summary of this PR in "check for config drift " commit of what it is doing.

sinnykumari (commented):
Thanks for keeping all the patience with review comments :) Nice work! Let's get this merged :tada: 

/lgtm

PR #2797, opened 2021-10-11
Bug 2012780: Avoid dynamically allocated port range for haproxy

cybertron:
The haproxy stats and health endpoints were configured with high
port numbers (50000 and 50936) that conflict with the range used for
dynamic allocation of ports:

$ cat /proc/sys/net/ipv4/ip_local_port_range
32768   60999

This can occasionally cause haproxy to fail if the kernel happens to
pick one of those ports for a client connection.

To avoid this, we should use ports outside the range shown above.
This patch changes the health endpoint to 30936, while the stats
port is controlled by runtimecfg and will be fixed in a separate
commit there.

**- What I did**

**- How to verify it**

Moved the haproxy health check port to avoid conflicts with dynamically allocated client ports.

### DIFF
--- a/templates/master/00-master/on-prem/files/haproxy-haproxy.yaml
+++ b/templates/master/00-master/on-prem/files/haproxy-haproxy.yaml
@@ -20,7 +20,7 @@ contents:
       bind :::{{`{{ .LBConfig.LbPort }}`}} v4v6
       default_backend masters
     listen health_check_http_url
-      bind :::50936 v4v6
+      bind :::30936 v4v6
       mode http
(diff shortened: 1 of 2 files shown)


mandre (commented):
Looks good except for commit message that mentions port 50000.

sinnykumari (approved):
LGTM.
/hold
Feel free to remove hold once all the relevant team has LGTM'ed

cybertron:
/retest-required
/hold cancel

CI is passing on all platforms and we have lgtm from 3/4. Should be good to go.

PR #2800, opened 2021-10-12
Bug 1981549: lib/resourcemerge: handle container env var deletions

yuqi-zhang:
The logic in the resourcemerge functions only iterate through required
variables, meaning any removed variable is not handled.

As a fix to bug 1981549, this adds removal check for env vars, which
ensures that e.g. a removed proxy will correctly propagate to the
daemonset definition, which needs the proxy injected as an environment
variable to allow MCD to pull os image updates.

This of course is also a problem for everything else being synced via
these lib functions, but for now I only added the fix to EnvVar for
proxy, as it is the most likely issue for users to hit. If we want to
fix all other variables, we should probably also consider reworking
the resource* libraries in general, since they are outdated and error
prone.

To test, you can pause the pool, add a cluster proxy and then remove
it, checking the MCD daemonset after both steps to see the proxy being
added/removed. Interestingly, the adding of the proxy is almost instant,
whereas the removal can take up to 10 minutes due to the MCO seemingly
not resyncing (no action from proxy informer?). I am investigating that
separately.

Also add some initial tests for EnvVar handling specifically, to make sure
we don't regress proxy. Other unit tests should be added as we clean up
the code.

### DIFF
--- a/lib/resourcemerge/core_test.go
+++ b/lib/resourcemerge/core_test.go
@@ -1,0 +1,133 @@
+package resourcemerge
+
+import (
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+)
+
+// Test added for now to test ensureContainer (mostly so we don't break Env Vars - proxy)
+func TestEnsureContainer(t *testing.T) {
+	tests := []struct {
+		existing corev1.Container
+		input    corev1.Container
+
+		expectedModified bool
+		expected         corev1.Container
+	}{
+		// Change in proxy
+		{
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
CVO, as an example, has their own fork as well: https://github.com/openshift/cluster-version-operator/tree/master/lib. There is a card https://issues.redhat.com/browse/GRPA-3832 but I am not sure what came out of that discussion

sinnykumari (approved):
Thanks Jerry for fixing this!
Tested locally, proxy changes get applied as expected to MCD pods.
/lgtm

palonsoro:
(I removed my cherry-pick comment because I thought it had a typo but it didn't)

PR #2802, opened 2021-10-16
Send alert when MCO can't safely apply updated Kubelet CA on nodes in paused pool

jkyros:
**- What I did**

1. Added metrics reporting to machine-config-controller because it did not previously have that capability by adding:  
- In manifests:
   - Cluster Roles
   - Cluster Role Bindings
   - ServiceMonitor for metrics
   - Service for metrics
   - oauth-proxy container for machine-config-controller deployment
   - mcc-proxy-tls secret for machine-config-controller
  
- In controller:
   - metrics handler function in machine-config-controller common
   - machine config lister in node_controller and node_controller_test
- References:
   - For the handler I cribbed off of: 557303ffd70ba6ccfa834f86ff7d4acd1f97dd17
   - And then to add oauth: 3ab692f423cb34d70d7cb4dc808206849362ba05	

2. Added an alert that fires when the kubelect-ca certificate is pending in a paused pool: 
   - added a GaugeVec (MCCImportantConfigPaused) for important pending config 
   - added a map in node_controller to store the table of "important config names" : "important config files"
   - added functions to check for pending files, and release the alerts once the pools have been unpaused
 
3. I don't have tests yet

**- How to verify it**

1. To fire the alert:

   - build a cluster
   - oc edit mcp worker 
   - change spec.paused: false to spec.paused: true 
   - Trigger a certificate rotation:  ```oc patch secret -p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}' kube-apiserver-to-kubelet-signer -n openshift-kube-apiserver-operator```
   - Observe prometheus metrics: ```TOKEN=`oc sa get-token prometheus-k8s -n openshift-monitoring`  
POD=`oc get pods -n openshift-machine-config-operator | grep controller | awk '{print $1}'`  
oc rsh $POD curl -k -H "Authorization: Bearer $TOKEN" https://localhost:9001/metrics  ```
   - look for mcc_important_config_paused
   - (or watch the web UI, they show up there too)
   
 2. To stop the alert: 
    - oc edit mcp worker 
    - change spec.paused: true to spec.paused: false 
    - once again observe prometheus metrics -- the alert will stop firing/set to 0 

Send alert when MCO can't safely apply updated Kubelet CA on nodes in paused pool

### DIFF
--- a/cmd/machine-config-controller/start.go
+++ b/cmd/machine-config-controller/start.go
@@ -29,10 +29,10 @@ var (
 	}
 
 	startOpts struct {
-		kubeconfig string
-		templates  string
-
-		resourceLockNamespace string
+		kubeconfig               string
+		templates                string
+		promMetricsListenAddress string
+		resourceLockNamespace    string
 	}
 )
 
@@ -40,6 +40,7 @@ func init() {
 	rootCmd.AddCommand(startCmd)
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
 	startCmd.PersistentFlags().StringVar(&startOpts.resourceLockNamespace, "resourcelock-namespace", metav1.NamespaceSystem, "Path to the template files used for creating MachineConfig objects")
+	startCmd.PersistentFlags().StringVar(&startOpts.promMetricsListenAddress, "metrics-listen-address", "127.0.0.1:8797", "Listen address for prometheus metrics listener")
 }
 
 func runStartCmd(cmd *cobra.Command, args []string) {
@@ -55,6 +56,9 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	}
 	run := func(ctx context.Context) {
 		ctrlctx := ctrlcommon.CreateControllerContext(cb, ctx.Done(), componentName)
+
+		// Start the metrics handler
+		go ctrlcommon.StartMetricsListener(startOpts.promMetricsListenAddress, ctrlctx.Stop)
 
 		controllers := createControllers(ctrlctx)
 
@@ -140,6 +144,7 @@ func createControllers(ctx *ctrlcommon.ControllerContext) []ctrlcommon.Controlle
 		// The node controller consumes data written by the above
 		node.New(
 			ctx.InformerFactory.Machineconfiguration().V1().ControllerConfigs(),
+			ctx.InformerFactory.Machineconfiguration().V1().MachineConfigs(),
 			ctx.InformerFactory.Machineconfiguration().V1().MachineConfigPools(),
 			ctx.KubeInformerFactory.Core().V1().Nodes(),
 			ctx.ConfigInformerFactory.Config().V1().Schedulers(),
--- a/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
+++ b/install/0000_90_machine-config-operator_01_prometheus-rules.yaml
@@ -1,6 +1,42 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
+  name: machine-config-controller
+  namespace: openshift-machine-config-operator
+  labels:
+    k8s-app: machine-config-controller
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: mcc-paused-pool-kubelet-ca
+      rules:
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             max by (namespace,pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
+          for: 60m
+          labels:
+            severity: warning
+          annotations:
+            summary: "Paused machine configuration pool '{{$labels.pool}}' is blocking a necessary certificate rotation and must be unpaused before the current kube-apiserver-to-kubelet-signer certificate expires on {{ $value | humanizeTimestamp }}."
+            description: "Machine config pools have a 'pause' feature, which allows config to be rendered, but prevents it from being rolled out to the nodes. This alert indicates that a certificate rotation has taken place, and the new kubelet-ca certificate bundle has been rendered into a machine config, but because the pool '{{$labels.pool}}' is paused, the config cannot be rolled out to the nodes in that pool. You will notice almost immediately that for nodes in pool '{{$labels.pool}}', pod logs will not be visible in the console and interactive commands (oc log, oc exec, oc debug, oc attach) will not work. You must unpause machine config pool '{{$labels.pool}}' to let the certificates through before the kube-apiserver-to-kubelet-signer certificate expires on {{ $value | humanizeTimestamp }} or this pool's nodes will cease to function properly."
+            runbook_url: https://github.com/openshift/blob/master/alerts/machine-config-operator/MachineConfigControllerPausedPoolKubeletCA.md
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             max by (namespace,pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m]) - time()) < (86400 * 14) AND max by (namespace,pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
+          for: 60m
+          labels:
+            severity: critical
+          annotations:
+            summary: "Paused machine configuration pool '{{$labels.pool}}' is blocking a necessary certificate rotation and must be unpaused before the current kube-apiserver-to-kubelet-signer certificate expires in {{ $value | humanizeDuration }}."
+            description: "Machine config pools have a 'pause' feature, which allows config to be rendered, but prevents it from being rolled out to the nodes. This alert indicates that a certificate rotation has taken place, and the new kubelet-ca certificate bundle has been rendered into a machine config, but because the pool '{{$labels.pool}}' is paused, the config cannot be rolled out to the nodes in that pool. You will notice almost immediately that for nodes in pool '{{$labels.pool}}', pod logs will not be visible in the console and interactive commands (oc log, oc exec, oc debug, oc attach) will not work. You must unpause machine config pool '{{$labels.pool}}' to let the certificates through before the kube-apiserver-to-kubelet-signer certificate expires. You have approximately {{ $value | humanizeDuration }} remaining before this happens and nodes in '{{$labels.pool}}' cease to function properly." 
+            runbook_url: https://github.com/openshift/blob/master/alerts/machine-config-operator/MachineConfigControllerPausedPoolKubeletCA.md
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
   name: machine-config-daemon
   namespace: openshift-machine-config-operator
   labels:
--- a/manifests/machineconfigcontroller/deployment.yaml
+++ b/manifests/machineconfigcontroller/deployment.yaml
@@ -27,10 +27,42 @@ spec:
             cpu: 20m
             memory: 50Mi
         terminationMessagePolicy: FallbackToLogsOnError
+      - name: oauth-proxy
+        image: {{.Images.OauthProxy}}
+        ports:
+        - containerPort: 9001
+          name: metrics
+          protocol: TCP
+        args:
+        - --https-address=:9001
+        - --provider=openshift
+        - --openshift-service-account=machine-config-controller
+        - --upstream=http://127.0.0.1:8797
+        - --tls-cert=/etc/tls/private/tls.crt
+        - --tls-key=/etc/tls/private/tls.key
+        - --cookie-secret-file=/etc/tls/cookie-secret/cookie-secret
+        - '--openshift-sar={"resource": "namespaces", "verb": "get"}'
+        - '--openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get"}}'
+        resources:
+          requests:
+            cpu: 20m
+            memory: 50Mi
+        volumeMounts:
+        - mountPath: /etc/tls/private
+          name: proxy-tls
+        - mountPath: /etc/tls/cookie-secret
+          name: cookie-secret
       serviceAccountName: machine-config-controller
       nodeSelector:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
+      volumes:
+        - name: proxy-tls
+          secret:
+            secretName: mcc-proxy-tls
+        - name: cookie-secret
+          secret:
+            secretName: cookie-secret
       restartPolicy: Always
       tolerations:
       - key: node-role.kubernetes.io/master
--- a/pkg/controller/common/helpers.go
+++ b/pkg/controller/common/helpers.go
@@ -2,12 +2,15 @@ package common
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"net/url"
 	"os"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/clarketm/json"
 	fcctbase "github.com/coreos/fcct/base/v0_1"
@@ -739,4 +742,51 @@ func GetIgnitionFileDataByPath(config *ign3types.Config, path string) ([]byte, e
 		}
 	}
 	return nil, nil
+}
+
+// GetNewestCertificatesFromPEMBundle breaks a pem-encoded bundle out into its component certificates
+func GetCertificatesFromPEMBundle(pemBytes []byte) ([]*x509.Certificate, error) {
+	var certs []*x509.Certificate
+	// There can be multiple certificates in the file
+	for {
+		// Decode a block to parse
+		block, rest := pem.Decode(pemBytes)
+		// Once we get no more blocks, we've read all the certs
+		if block == nil {
+			break
+		}
+		// Right now we just care about certificates, not keys
+		if block.Type == "CERTIFICATE" {
+			cert, err := x509.ParseCertificate(block.Bytes)
+			if err != nil {
+				// This isn't fatal, *this* cert could just be junk, next one could be okay
+				glog.Warningf("Failed to parse certificate: %v", err.Error())
+			} else {
+				certs = append(certs, cert)
+			}
+		}
+		// Keep reading from where we left off
+		pemBytes = rest
+	}
+	return certs, nil
+}
+
+// GetLongestValidCertificate returns the latest-expiring certificate from a given list of certificates
+// whose Subject.CommonName also matches any of the given common-name prefixes
+func GetLongestValidCertificate(certificateList []*x509.Certificate, subjectPrefixes []string) *x509.Certificate {
+	// Sort is smallest-to-largest, so we're putting the cert with the latest expiry date at the top
+	sort.Slice(certificateList, func(i, j int) bool {
+		return certificateList[i].NotAfter.After(certificateList[j].NotAfter)
+	})
+	// For each certificate in our list
+	for _, certificate := range certificateList {
+		// Check it against our prefixes
+		for _, prefix := range subjectPrefixes {
+			// If it matches, this is the latest-expiring one since it's closest to the "top"
+			if strings.HasPrefix(certificate.Subject.CommonName, prefix) {
+				return certificate
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/controller/common/metrics.go
+++ b/pkg/controller/common/metrics.go
@@ -1,0 +1,68 @@
+package common
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const (
+	// DefaultBindAddress is the port for the metrics listener
+	DefaultBindAddress = ":8797"
+)
+
+var (
+	// MachineConfigControllerPausedPoolKubeletCA logs when a certificate rotation is being held up by pause
+	MachineConfigControllerPausedPoolKubeletCA = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "machine_config_controller_paused_pool_kubelet_ca",
+			Help: "Set to the unix timestamp in utc of the current certificate expiry date if a certificate rotation is pending in specified paused pool",
+		}, []string{"pool"})
+
+	metricsList = []prometheus.Collector{
+		MachineConfigControllerPausedPoolKubeletCA,
+	}
+)
+
+func RegisterMCCMetrics() error {
+	for _, metric := range metricsList {
+		err := prometheus.Register(metric)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// StartMetricsListener is metrics listener via http on localhost
+func StartMetricsListener(addr string, stopCh <-chan struct{}) {
+	if addr == "" {
+		addr = DefaultBindAddress
+	}
+
+	glog.Info("Registering Prometheus metrics")
+	if err := RegisterMCCMetrics(); err != nil {
+		glog.Errorf("unable to register metrics: %v", err)
+		// No sense in continuing starting the listener if this fails
+		return
+	}
+
+	glog.Infof("Starting metrics listener on %s", addr)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	s := http.Server{Addr: addr, Handler: mux}
+
+	go func() {
+		if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			glog.Errorf("metrics listener exited with error: %v", err)
+		}
+	}()
+	<-stopCh
+	if err := s.Shutdown(context.Background()); err != http.ErrServerClosed {
+		glog.Errorf("error stopping metrics listener: %v", err)
+	}
+}
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -2,11 +2,13 @@ package node
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
 
+	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
 	cligoinformersv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
@@ -24,6 +26,7 @@ import (
 	goerrs "github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -64,7 +67,14 @@ const (
 
 	// masterPoolName is the control plane MachineConfigPool name
 	masterPoolName = "master"
+
+	// kubeletCAFilePath is the expected file path for the kubelet ca
+	kubeletCAFilePath = "/etc/kubernetes/kubelet-ca.crt"
 )
+
+// kubeAPIToKubeletSignerNamePrefixes is the list of subject common names that are regarded as a kube-apiserver-to-kubelet-signer ca certificate
+// Based on naming convention from https://github.com/openshift/library-go/blob/ed9bc958bd8a2fff079d52976806e4e0a8a7c315/pkg/operator/certrotation/signer.go#L132
+var kubeAPIToKubeletSignerNamePrefixes = []string{"openshift-kube-apiserver-operator_kube-apiserver-to-kubelet-signer@", "kube-apiserver-to-kubelet-signer"}
 
 // Controller defines the node controller.
 type Controller struct {
@@ -76,10 +86,12 @@ type Controller struct {
 	enqueueMachineConfigPool func(*mcfgv1.MachineConfigPool)
 
 	ccLister   mcfglistersv1.ControllerConfigLister
+	mcLister   mcfglistersv1.MachineConfigLister
 	mcpLister  mcfglistersv1.MachineConfigPoolLister
 	nodeLister corelisterv1.NodeLister
 
 	ccListerSynced   cache.InformerSynced
+	mcListerSynced   cache.InformerSynced
 	mcpListerSynced  cache.InformerSynced
 	nodeListerSynced cache.InformerSynced
 
@@ -92,6 +104,7 @@ type Controller struct {
 // New returns a new node controller.
 func New(
 	ccInformer mcfginformersv1.ControllerConfigInformer,
+	mcInformer mcfginformersv1.MachineConfigInformer,
 	mcpInformer mcfginformersv1.MachineConfigPoolInformer,
 	nodeInformer coreinformersv1.NodeInformer,
 	schedulerInformer cligoinformersv1.SchedulerInformer,
@@ -128,9 +141,11 @@ func New(
 	ctrl.enqueueMachineConfigPool = ctrl.enqueueDefault
 
 	ctrl.ccLister = ccInformer.Lister()
+	ctrl.mcLister = mcInformer.Lister()
 	ctrl.mcpLister = mcpInformer.Lister()
 	ctrl.nodeLister = nodeInformer.Lister()
 	ctrl.ccListerSynced = ccInformer.Informer().HasSynced
+	ctrl.mcListerSynced = mcInformer.Informer().HasSynced
 	ctrl.mcpListerSynced = mcpInformer.Informer().HasSynced
 	ctrl.nodeListerSynced = nodeInformer.Informer().HasSynced
 
@@ -145,7 +160,7 @@ func (ctrl *Controller) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer ctrl.queue.ShutDown()
 
-	if !cache.WaitForCacheSync(stopCh, ctrl.ccListerSynced, ctrl.mcpListerSynced, ctrl.nodeListerSynced, ctrl.schedulerListerSynced) {
+	if !cache.WaitForCacheSync(stopCh, ctrl.ccListerSynced, ctrl.mcListerSynced, ctrl.mcpListerSynced, ctrl.nodeListerSynced, ctrl.schedulerListerSynced) {
 		return
 	}
 
@@ -743,8 +758,16 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		if mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
 			glog.Infof("Pool %s is paused and will not update.", pool.Name)
 		}
+
+		// Only check for pending files if we're out of sync
+		if pool.Spec.Configuration.Name != pool.Status.Configuration.Name {
+			ctrl.setPendingFileMetrics(pool)
+		}
 		return ctrl.syncStatusOnly(pool)
 	}
+
+	// We aren't paused anymore, so reset the metrics
+	ctrl.resetPendingFileMetrics(pool)
 
 	nodes, err := ctrl.getNodesForPool(pool)
 	if err != nil {
@@ -1041,4 +1064,113 @@ func getErrorString(err error) string {
 		return err.Error()
 	}
 	return ""
+}
+
+// setPendingFileMetrics checks to see if there are any important files in the
+// machineconfig that the pool should be moving to, and sets metrics if there are
+func (ctrl *Controller) setPendingFileMetrics(pool *mcfgv1.MachineConfigPool) {
+	// Retrieve and parse the pool's machine config
+	currentConfig, pendingConfig, err := ctrl.parseConvertMachineConfigFilesForPool(pool)
+	if err != nil {
+		glog.Warningf("Error converting pool configs for %s pool: %v", pool.Name, err)
+		return
+	}
+
+	// Figure out what files differ between pool.Spec and pool.Status
+	fileDiff := ctrlcommon.CalculateConfigFileDiffs(currentConfig, pendingConfig)
+
+	// Go through our files until we hit the kubelet CA bundle
+	for _, path := range fileDiff {
+		// We only care about the kubelet CA bundle
+		if path != kubeletCAFilePath {
+			continue
+		}
+
+		// If it's there, get the *newest* (in case there have been multiple rotations) kube-apiserver-to-kubelet signer certifiate out of the bundle
+		newestSignerCertificate, err := ctrl.getNewestAPIToKubeletSignerCertificate(currentConfig)
+		if err != nil {
+			glog.Warningf("Error retrieving kubelet-ca certificates from pool %s: %v", pool.Name, err)
+		} else {
+			// Set the metric value to the UTC expiry date of that cert so we can count down to it
+			glog.V(2).Infof("Kubelet CA is stuck in paused pool %s. Setting metric to expiry date of %s (%s)", pool.Name, newestSignerCertificate.Subject.CommonName, newestSignerCertificate.NotAfter.UTC())
+			ctrlcommon.MachineConfigControllerPausedPoolKubeletCA.WithLabelValues(pool.Name).Set(float64(newestSignerCertificate.NotAfter.UTC().Unix()))
+		}
+		break
+	}
+}
+
+// resetPendingFileMetrics turns off any "paused file" metrics that were firing for the pool
+func (ctrl *Controller) resetPendingFileMetrics(pool *mcfgv1.MachineConfigPool) {
+	// Set the metric for this pool back to zero
+	ctrlcommon.MachineConfigControllerPausedPoolKubeletCA.WithLabelValues(pool.Name).Set(0)
+}
+
+// parseConvertMachineConfigFilesForPool retrieves the current and pending configurations for
+// a pool, parses and converts them, and returns them as ignition v3 Config objects. The controller needs
+// to retrieve and examine the actual configurations so it can diff the file lists and figure out which new
+// files are "stuck" behind a paused pool.
+func (ctrl *Controller) parseConvertMachineConfigFilesForPool(pool *mcfgv1.MachineConfigPool) (current, pending *ign3types.Config, err error) {
+	// The config we're in right now
+	currentName := pool.Status.Configuration.Name
+	// The config we would be going to
+	pendingName := pool.Spec.Configuration.Name
+
+	// Get the machine config objects
+	currentConfig, err := ctrl.mcLister.Get(currentName)
+	if apierrors.IsNotFound(err) {
+		glog.V(2).Infof("MachineConfig %v has been deleted", currentName)
+		return nil, nil, err
+	}
+
+	pendingConfig, err := ctrl.mcLister.Get(pendingName)
+	if apierrors.IsNotFound(err) {
+		glog.V(2).Infof("MachineConfigPool %v has been deleted", pendingName)
+		return nil, nil, err
+	}
+
+	// Make sure we can coax the objects into ignitionv3
+	currentIgnConfig, err := ctrlcommon.ParseAndConvertConfig(currentConfig.Spec.Config.Raw)
+	if err != nil {
+		return nil, nil, err
+	}
+	pendingIgnConfig, err := ctrlcommon.ParseAndConvertConfig(pendingConfig.Spec.Config.Raw)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &currentIgnConfig, &pendingIgnConfig, nil
+}
+
+// getNewestAPIToKubeletSignerCertificate returns the newest kube-apiserver-to-kubelet-signer
+// certificate present in the kubelet-ca.crt bundle. We extract the certificate so we can use its
+// expiry date in our metrics/alerting. It's a very important certificate and its expiry will cause
+// nodes using it to cease communicating with the cluster.
+func (ctrl *Controller) getNewestAPIToKubeletSignerCertificate(statusIgnConfig *ign3types.Config) (*x509.Certificate, error) {
+	// Retrieve the file data from ignition
+	kubeletBundle, err := ctrlcommon.GetIgnitionFileDataByPath(statusIgnConfig, kubeletCAFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse that bundle into its component certificates
+	containedCertificates, err := ctrlcommon.GetCertificatesFromPEMBundle(kubeletBundle)
+	if err != nil {
+		return nil, err
+	}
+
+	// We have other problems if this is empty, but it's possible
+	if len(containedCertificates) == 0 {
+		return nil, fmt.Errorf("No certificates found in bundle")
+	}
+
+	// The *original* signer has a different name, the rotated ones have longer names
+	// The suffix changes with the timstamp on rotation, which is why I'm using prefix here not exact match
+	newestCertificate := ctrlcommon.GetLongestValidCertificate(containedCertificates, kubeAPIToKubeletSignerNamePrefixes)
+
+	// Shouldn't come back with nothing, but just in case we do
+	if newestCertificate == nil {
+		return nil, fmt.Errorf("No matching kube-apiserver-to-kubelet-signer certificates found in bundle")
+	}
+
+	return newestCertificate, nil
 }
--- a/test/e2e/mcc_test.go
+++ b/test/e2e/mcc_test.go
@@ -1,0 +1,111 @@
+package e2e_test
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/openshift/machine-config-operator/test/framework"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test case to make sure the MCC alerts properly when pools are
+// paused and a certificate rotation happens
+func TestMCCPausedKubeletCAAlert(t *testing.T) {
+	var testPool = "master"
+
+	cs := framework.NewClientSet("")
+
+	// Get the machine config pool
+	mcp, err := cs.MachineConfigPools().Get(context.TODO(), testPool, metav1.GetOptions{})
+	require.Nil(t, err)
+
+	// Update our copy of the pool
+	newMcp := mcp.DeepCopy()
+	newMcp.Spec.Paused = true
+
+	t.Logf("Pausing pool")
+
+	// Update the pool to be paused
+	_, err = cs.MachineConfigPools().Update(context.TODO(), newMcp, metav1.UpdateOptions{})
+	require.Nil(t, err)
+	t.Logf("Paused")
+
+	// Rotate the certificates
+	t.Logf("Patching certificate")
+	err = helpers.ForceKubeApiserverCertificateRotation(cs)
+	require.Nil(t, err)
+	t.Logf("Patched")
+
+	// Verify the pool is paused and the config is pending but not rolling out
+	t.Logf("Waiting for rendered config to get stuck behind pause")
+	err = helpers.WaitForPausedConfig(t, cs, testPool)
+	require.Nil(t, err)
+	t.Logf("Certificate stuck behind paused (as expected)")
+
+	// Retrieve the token from the prometheus secret, we need it to auth to the oauth proxy
+	t.Logf("Getting monitoring token")
+	token, err := helpers.GetMonitoringToken(t, cs)
+	require.Nil(t, err)
+
+	// Get the service details so we can ge the IP and port to check for the metrics
+	t.Logf("Getting metrics listener service details")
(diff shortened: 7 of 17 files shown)


install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,30 @@
...
+  groups:
+    - name: mcc-important-config-paosed
+      rules:
+        - alert: MCCImportantConfigPaused
+          expr: |
+             mcc_important_config_paused > 0
+          labels:
+            severity: warning
+          annotations:
+            message: "Important config {{$labels.desc}} ({{ $labels.file }}) is blocked by pause in {{ $labels.pool }}"
```
wking, line 22: I'm pretty sure you want `summary` and `description`, but not `message`.  For example, see openshift/cluster-version-operator#547, which was successfully verified by QE, and [these monitoring docs][1], which call for `summary` and `description`, but do not mention the older `message`.

[1]: https://github.com/monitoring-mixins/docs/pull/10/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R183

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,30 @@
...
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: mcc-important-config-paosed
+      rules:
+        - alert: MCCImportantConfigPaused
+          expr: |
+             mcc_important_config_paused > 0
+          labels:
+            severity: warning
```
wking, line 20: Firing `warning` alerts immediately seems like it's wound a bit too tightly.  We can go a while before the lack of certificate rotation becomes a problem (significant fractions of a year?).  So setting a large-ish `for` here seems useful to avoid desensitizing folks by alerting every time they briefly pause.  Of course, you could have someone leaving pools paused for the bulk of that time, and then very briefly unpause, have Prom happen to scrape, and then pause again, before the pool had time to rotate out many certs.  Which is one reason "are we paused now, and have we been paused for a bit?" only a rough proxy for "are we at risk of certs expiring?".  And that makes picking an appropriate `for` difficult.  But if we stick with the paused-pool angle on this, I'd expect at least `for: 60m` on this alert, based on [our consistency guidelines][1].

[1]: https://github.com/openshift/enhancements/blob/15aebc7549095cfc7eebc26d6460defc8e3927f6/enhancements/monitoring/alerting-consistency.md#warning-alerts

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,30 @@
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: mcc-important-config-paosed
+      rules:
+        - alert: MCCImportantConfigPaused
+          expr: |
+             mcc_important_config_paused > 0
```
wking, line 18: you will probably want some aggregation here to avoid twitching off and on if/when the machine-config controller container rolls.  Because that will give you a small gap when `mcc_important_config_paused`, and the new container may come up with different injected scrape labels, and both of those may shuffle the time series matching this expression, leading to new alerts (after cooking through [`for`][1]).  I addressed the scrape-labels portion for a CVO-side alert in openshift/cluster-version-operator#643, but missed the metrics-downtime portion.  Something like:

    sum by (description, path, pool) (max_over_time(mcc_important_config_paused[5m])) > 0

should address both issues, but while the `max_over_time(...[5m])` smearing protects you from alert-churning leader blips, it also makes the alert less responsive to the metric actually getting cleared.  So something like "`mcc_important_config_paused` looks bad, or some signal that the machine-config controller is currently not exporting metrics" would be better.

[1]: https://github.com/openshift/machine-config-operator/pull/2802#discussion_r731511094

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+var (
+	// DefaultBindAddress is the port for the metrics listener
+	DefaultBindAddress = ":8797"
+
+	// MCCImportantConfigPaused logs when important config is being held up by pause
+	MCCImportantConfigPaused = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mcc_important_config_paused",
```
wking, line 19: I think about the machine-config controller often enough to recognize the `mcc` acronym.  But many folks browsing available metrics on their cluster will not.  Can we write this out `machine_config_controller_content_paused` or some such?

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+			Help: "important config type is waiting in specified paused pool",
+			// "platform config kubelet ca certificate is stuck in paused pool worker"
+		}, []string{"desc", "file", "pool"})
```
wking, line 22: Is there a benefit to the `desc` abbreviation over a full `description`?

You can also avoid landing the description in the metric itself by using path-to-description lookup logic in the alert's `description`.  I'm not familiar enough with Prom or the alert stack to know if that's worth the trouble.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,30 @@
...
+          labels:
+            severity: warning
+          annotations:
+            message: "Important config {{$labels.desc}} ({{ $labels.file }}) is blocked by pause in {{ $labels.pool }}"
+            summary: "Important configuration is being prevented from syncing by paused pool"
```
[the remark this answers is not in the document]
sinnykumari: ok, I see that you are keeping the alert name generic for future. Not needed then.

pkg/controller/node/node_controller.go
```diff
@@ -744,7 +757,21 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
 		if mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) {
 			glog.Infof("Pool %s is paused and will not update.", pool.Name)
 		}
+
+		// See if there are any important files stuck behind pause and alert if there are
+		err := ctrl.alertIfImportantFilesPending(pool)
+		// I don't think we want errors here to be fatal if everything else is fine, just log them
```
kikisdeliveryservice, line 763:
```suggestion

```

pkg/controller/node/node_controller.go
```diff
@@ -744,7 +757,21 @@ func (ctrl *Controller) syncMachineConfigPool(key string) error {
...
+		if err != nil {
+			glog.Errorf("Unable to check pool %s for pending important configuration: %v", pool.Name, err)
+		}
+
 		return ctrl.syncStatusOnly(pool)
+
+	} else {
+		// Pool is no longer paused, so we need to turn off the paused pool file alerts
+		// TODO: I don't like the overhead of having do this every time, but the alternative is keeping a
```
kikisdeliveryservice, line 772:
```suggestion

```

remove TODO, create a jira card instead.

pkg/controller/node/node_controller.go
```diff
@@ -995,3 +1022,92 @@ func getErrorString(err error) string {
...
+			glog.Infof("File diff: detected change to %v", newFile.Path)
+			diffFileSet = append(diffFileSet, path)
+		}
+	}
+
+	// Go through our list of files and alert if they match our important list
+	for _, path := range diffFileSet {
+		// If this is an important file, trigger its alert
+		for alertFilePath, alertFileDesc := range alertOnPausedFiles {
+			if path == alertFilePath {
```
sinnykumari, line 1097: Instead of sending alert for each important file, it will good to send a single alert with important change list we care about.

pkg/controller/node/node_controller.go
```diff
@@ -995,3 +1022,92 @@ func getErrorString(err error) string {
...
+	if err != nil {
+		return err
+	}
+	specIgnConfig, err := ctrlcommon.ParseAndConvertConfig(specConfig.Spec.Config.Raw)
+	if err != nil {
+		return err
+	}
+
+	// Go through the files and see what is new or different
+	oldFileSet := make(map[string]ign3types.File)
```
sinnykumari, line 1059: Perhaps we can move calculating file diff list logic into another function so that it can be reused here as well as in [calculatePostConfigChangeActionFromFileDiffs() function](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L424)

jkyros:
Ok I think I've got this updated with the suggested feedback and ready for review again:

 - Used existing context/stop channel for metrics handler
 - Adjusted alert name/variables to be clearer, fixed fields, added future runbook link
 - Smoothed out the alerting period 
 - Broke the file diff function out into helpers and added tests 
 - Focused on just kubelet-ca.crt 
   - I know previous feedback here was leaning 'one alert for all config', but subsequent discussions seemed to lean back the other way to 'just the certificate for now' when we talked about keeping state for the alerts somewhere. 
   - If, as you're going through this, you think we want to go back the other way, I really just want to do what we think is right.  

  
   

- The e2e test is of less concern right now than everything else, but I've included it.
  -  It works, but I really would love a more elegant way than "rsh into a pod" to check that metrics endpoint. 

- I'm still trying to find a more concrete answer as to how long after certificates are paused the 'bad stuff' happens so we can set the critical time threshold appropriately
  -  ( we supposedly have 73 days, but according to what we saw here, we seem to only have 2 weeks: https://bugzilla.redhat.com/show_bug.cgi?id=1970329#c17 )

test/helpers/utils.go
```diff
@@ -120,6 +122,73 @@ func WaitForPoolComplete(t *testing.T, cs *framework.ClientSet, pool, target str
 	return nil
 }
 
+// WaitForPoolComplete polls a pool until it has completed any update
+func WaitForPoolCompleteAny(t *testing.T, cs *framework.ClientSet, pool string) error {
```
cheesesashimi, line 126: Something to be mindful about is if you run this test in an SNO context, this function will prematurely error because the control plane may go offline.

To work around that, you'd need to do something like: https://github.com/openshift/machine-config-operator/blob/master/test/e2e-single-node/sno_mcd_test.go#L355-L374 or: https://github.com/openshift/machine-config-operator/pull/2795/files#diff-b240aacd848b27ec33331f84d7e3fc0a4bc20045f45f2cd536727396a756e97bR195-R223.

The short version is:
1. Ignore the error you get from the `cs.MachineConfigPools().Get()`.
2. Run the clock out.
3. If you're still getting an error from `cs.MachineConfigPools().Get()`, then fail.

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// StartMetricsListener is metrics listener via http on localhost
+func StartMetricsListener(addr string, stopCh <-chan struct{}) {
```
cheesesashimi, line 40: This looks very similar to [`daemon.StartMetricsListener`](https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/metrics.go#L100-L109). I wonder if there's an opportunity for consolidating the boilerplate of starting the listener, e.g.:

```golang
func StartMetricsListener(addr string, stopCh <-chan struct{}, register func() error) {
  // Boilerplate
  if err := register(); err != nil {
    // Handle error
  }
  // More boilerplate
}
```

Then the controller and daemon can call this and pass in their registration funcs.

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	if addr == "" {
+		addr = DefaultBindAddress
+	}
+
+	glog.Info("Registering Prometheus metrics")
+	if err := RegisterMCCMetrics(); err != nil {
+		glog.Errorf("unable to register metrics: %v", err)
```
cheesesashimi, line 47: This logs that an error occurred during metrics registration and then continues starting the metrics listener. Does it make sense to continue trying to start the metrics listener if registration fails? Or would it make more sense to return early?

test/e2e/mcc_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/openshift/machine-config-operator/test/framework"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test case for https://github.com/openshift/machine-config-operator/issues/358
+func TestMCCPausedKubeletCAAlert(t *testing.T) {
+
+	var testPool = "master"
```
cheesesashimi, line 18: Does this have to target the `master` MachineConfigPool? I ask because a common pattern in the e2e test suite is that we create an ephemeral `infra` pool for the purpose of the test, assign a single worker node to that pool, then apply the MachineConfig. The reason is because we can apply the MachineConfig to a single node (and reset when we're done) much faster than rolling it out to the entire MachineConfigPool.

Here's an example: https://github.com/openshift/machine-config-operator/pull/2795/files#diff-32df39b541567d57b84246d0d48a44792350b5af97d060cd0e44c33949c07370R55-R96

pkg/controller/common/helpers_test.go
```diff
@@ -380,3 +381,35 @@ func TestRemoveIgnDuplicateFilesAndUnits(t *testing.T) {
 
 	assert.Equal(t, expectedIgn2Config, convertedIgn2Config)
 }
+
+func TestDiffMachineConfigFilesForPool(t *testing.T) {
```
cheesesashimi, line 385: I really like that you repackaged this and wrote a test for it! 👍

pkg/controller/common/helpers.go
```diff
@@ -622,3 +622,44 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
 
 	return passwdUser
 }
+
+// CalculateConfigFileDiffs compares the files present in two ignition configurations and returns the list of files
+// that are different between them
+func CalculateConfigFileDiffs(oldIgnConfig, newIgnConfig ign3types.Config) []string {
```
cheesesashimi, line 628: I really like that you repackaged this and wrote a test for it! 👍

pkg/controller/node/node_controller.go
```diff
@@ -995,3 +1011,74 @@ func getErrorString(err error) string {
...
+}
+
+// diffMachineConfigFilesForPool retrieves the machine config specified by the pool spec and status,
+// compares them, and returns the list of file paths that are different
+func (ctrl *Controller) diffMachineConfigFilesForPool(pool *mcfgv1.MachineConfigPool) ([]string, error) {
+
+	// The config we would be going to
+	specName := pool.Spec.Configuration.Name
+	// The config we're in right now
+	statusName := pool.Status.Configuration.Name
```
cheesesashimi, line 1051: A convention I've seen elsewhere in the MCO is to refer to these as `currentConfig` (the config we're currently in) and `pendingConfig` (the config we're moving to).

test/e2e/mcc_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	t.Logf("Pausing pool")
+
+	// Update the pool to be paused
+	_, err = cs.MachineConfigPools().Update(context.TODO(), newMcp, metav1.UpdateOptions{})
+	require.Nil(t, err)
+	t.Logf("Paused")
+
+	t.Logf("Patching certificate")
+	// Take note that the slash had to be encoded as ~1 because it's a reference: https://www.rfc-editor.org/rfc/rfc6901#section-3
```
cheesesashimi, line 38: This comment makes more sense in `helpers.ForceKubeApiserverCertificateRotation()` as opposed to here.

test/e2e/mcc_test.go
jkyros: So I did originally have it set up the "infra"  way, but:

1. I flipped it to master when I was testing against single node (because if a node is member of multiple pools, master seems to supersede it). 
2. There is no escaping the certificate rotation -- all unpaused pools will get it -- I don't have a way to confine a certificate rotation to certain pools (well, aside from pause, which is what I am testing) 

It probably does save a little bit of time having just one node be paused, otherwise we'd have to wait once we unpause. 

I'll put it back to infra  (and maybe copy this test as it sits into the single node tests if we care?)

test/helpers/utils.go
jkyros: Excellent. Thanks. I'll rearrange the test after your helpers merge. I did run the test against single node, so I know it works (aside from handling the control plane conditions you're describing).

test/e2e/mcc_test.go
jkyros: Left that one behind when I broke that out into a function :(

pkg/controller/node/node_controller.go
jkyros: I've updated the language to favor the current/pending convention for clarity - my head was full of pool specs and statuses from hanging out in the node controller.

jkyros:
I had to rebase/push to grab the new bootstrap test changes, I didn't just force push for fun. 

Also, now instead of just alerting boolean yes/no when config is pending, the metric now gets set to the expiry date of the most recent kube-api-to-kubelet signer certificate from the pool's current config.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+            runbook_url: https://github.com/openshift/blob/master/alerts/machine-config-operator/MachineConfigControllerPausedPoolKubeletCA.md
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             sum by (pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m]) - time()) < (86400 * 365) AND sum by (pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
```
jkyros, line 28: **This is set to 365 for testing so it always alerts**. 

For real I was thinking 14 or 21 days but I would love suggestions. I want to give enough time that an outage can be scheduled and completed. 

The thought here was that: 
-  max_over_time ignores zeroes, so we would take 5 minutes to shut off once the condition is fixed
-  last_over_time ignores missing data like max_over_time (so restarts, etc) but pays attention when we get a zero 

So the joint condition lets us avoid flapping while turning off the alert immediately once the pool syncs and sets it back to zero.

sinnykumari:
Please pull this PR out of draft when this is ready for review.

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
 
 	return passwdUser
 }
+
+// CalculateConfigFileDiffs compares the files present in two ignition configurations and returns the list of files
+// that are different between them
+func CalculateConfigFileDiffs(oldIgnConfig, newIgnConfig *ign3types.Config) []string {
+
```
sinnykumari, line 668: nit: extra newline

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+		newFileSet[f.Path] = f
+	}
+	diffFileSet := []string{}
+
+	// First check if any files were removed
+	for path := range oldFileSet {
+		_, ok := newFileSet[path]
+		if !ok {
+			// debug: remove
+			glog.Infof("File diff: %v was deleted", path)
```
sinnykumari, line 685: Do we need this debug log here?

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+			diffFileSet = append(diffFileSet, path)
+		}
+	}
+
+	// Now check if any files were added/changed
+	for path, newFile := range newFileSet {
+		oldFile, ok := oldFileSet[path]
+		if !ok {
+			// debug: remove
+			glog.Infof("File diff: %v was added", path)
```
sinnykumari, line 695: same question as above

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+	return diffFileSet
+}
+
+// GetNewestCertificatesFromPEMBundle breaks a pem-encoded bundle out into its component certificates
+func GetCertificatesFromPEMBundle(pemBytes []byte) ([]*x509.Certificate, error) {
+	var certs []*x509.Certificate
+
+	// There can be multiple certificates in the file
+	for {
+
```
sinnykumari, line 712: nit: newline not needed here

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+		block, rest := pem.Decode(pemBytes)
+
+		// Once we get no more blocks, we've read all the certs
+		if block == nil {
+			break
+		}
+
+		// Right now we just care about certificates, not keys
+		if block.Type == "CERTIFICATE" {
+
```
sinnykumari, line 723: nit: extra newline

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+			}
+		}
+	}
+	return nil, nil
+}
+
+// GetLongestValidCertificate returns the latest-expiring certificate from a given list of certificates
+// whose Subject.CommonName also matches any of the given common-name prefixes
+func GetLongestValidCertificate(certificateList []*x509.Certificate, subjectPrefixes []string) *x509.Certificate {
+
```
sinnykumari, line 762: nit: extra newline

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+	// Sort is smallest-to-largest, so we're putting the cert with the latest expiry date at the top
+	sort.Slice(certificateList, func(i, j int) bool {
+		return certificateList[i].NotAfter.After(certificateList[j].NotAfter)
+	})
+	// For each certificate in our list
+	for _, certificate := range certificateList {
+
```
sinnykumari, line 769: nit: extra newline

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+		// Check it against our prefixes
+		for _, prefix := range subjectPrefixes {
+
```
sinnykumari, line 772: nit: extra newline

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
+		// Keep reading from where we left off
+		pemBytes = rest
+	}
+
+	return certs, nil
+}
+
+// GetIgnitionFileDataByPath retrieves the file data for a specified path from a given ignition config
+func GetIgnitonFileDataByPath(config *ign3types.Config, path string) ([]byte, error) {
+
```
sinnykumari, line 742: nit: extra newline

pkg/daemon/update.go
```diff
@@ -437,39 +437,7 @@ func calculatePostConfigChangeActionFromFileDiffs(oldIgnConfig, newIgnConfig ign
...
-			// debug: remove
-			glog.Infof("File diff: %v was added", path)
-			diffFileSet = append(diffFileSet, path)
-		} else if !reflect.DeepEqual(oldFile, newFile) {
-			// debug: remove
-			glog.Infof("File diff: detected change to %v", newFile.Path)
-			diffFileSet = append(diffFileSet, path)
-		}
-	}
+	diffFileSet := ctrlcommon.CalculateConfigFileDiffs(&oldIgnConfig, &newIgnConfig)
```
sinnykumari, line 440: @mkenigs Looks like John's PR is already refactoring fileDiff list in a separate function (related to https://github.com/openshift/machine-config-operator/pull/2851)

sinnykumari: @john Logging about changed file will make more sense here.

pkg/controller/common/helpers.go
sinnykumari: suggested alternative at  https://github.com/openshift/machine-config-operator/pull/2802/files#r766787404

pkg/daemon/update.go
mkenigs: What's the best way to handle that? Should I wait for my PR until this one lands?

pkg/controller/common/helpers.go
```diff
@@ -658,3 +661,120 @@ func dedupePasswdUserSSHKeys(passwdUser ign2types.PasswdUser) ign2types.PasswdUs
...
 	return passwdUser
 }
+
+// CalculateConfigFileDiffs compares the files present in two ignition configurations and returns the list of files
+// that are different between them
+func CalculateConfigFileDiffs(oldIgnConfig, newIgnConfig *ign3types.Config) []string {
+
+	// Go through the files and see what is new or different
+	oldFileSet := make(map[string]ign3types.File)
+	for _, f := range oldIgnConfig.Storage.Files {
```
mkenigs, line 671: Not sure if it'd be worth the effort or an improvement, but when I was looking at this I was wondering about using 2 loops instead of 4: one loop to construct the `oldFileSet`, and then one loop to simultaneously add every new file and remove every unchanged file from that set

pkg/controller/common/helpers.go
jkyros: It was in the original code that I refactored, and it never got taken out. I've seen us look at it  a couple times when debugging. 

I was too chicken to take it out, but yeah it's been in there long enough that we should probably either take the debug comment off it and accept that we're keeping it or remove both the log statement and the comment . :smiley:

pkg/controller/common/helpers.go
jkyros: I am...so sorry about my love of extra newlines. I will clean all these up.

pkg/controller/common/helpers.go
jkyros: Also, we lose the added/changed/deleted distinction once we return the list, it's just an array of strings representing "stuff that's different than it was before", but we don't keep "how".

test/e2e/mcc_test.go
sinnykumari: I think for this particular test we don't really need to create infra pool because cert rotation would take place for all pools in the cluster. And for test purpose it is fine to pause only master pool if it helps with SNO case too.

test/helpers/utils.go
sinnykumari: Agree with Zack that for SNO, cluster can be itself unreachable because everything is running on the same node and when node reboots, it will be unreachable for a while. However, for this particular test case I won't be worried much because applying kube cert shouldn't cause drain or reboot. Ideally, cluster availability shouldn't be impacted and if it does happen, something is wrong.

sinnykumari:
Overall logic looks fine to me. Added few inline comments. There are extra newline at lot of places(added inline comment at some of them), will be good to get them fixed. Newline are usually not required at following places:
-  Start of a function
-  Start of if, else or other construct like for loop
-  Just before closing curly braces.

Also, this PR adds some of the functionalities that will worthwhile to get reviewed by corresponding teams. Node or apiserver team for cert part, monitoring team for alerts/metrics.

pkg/daemon/update.go
sinnykumari: @mkenigs Ah no, It's fine to continue with your code. Noone is blocked on anyone, tagged you to just for FYI. Whichever PR get merged later may need to fix merge conflict.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
sinnykumari: No strong opinion on exact timing. But yeah, giving 2 weeks of buffer sounds good.

test/helpers/helpers.go
```diff
@@ -150,6 +151,18 @@ func NewMachineConfigPool(name string, mcSelector, nodeSelector *metav1.LabelSel
...
+func NewMachineConfigFile(path, data string, mode int) *ign3types.File {
+
+	tempFile := ign3types.File{
+		Node: ign3types.Node{Path: path},
+		FileEmbedded1: ign3types.FileEmbedded1{Contents: ign3types.Resource{Source: StrToPtr(dataurl.EncodeBytes([]byte(data)))},
+			Mode: &mode}}
+
+	return &tempFile
+
+}
```
mkenigs, line 164: I recently added a similar function which I think you could use: https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update_test.go#L585

Unless do you need mode for some reason?

pkg/controller/common/helpers_test.go
```diff
@@ -380,3 +381,35 @@ func TestRemoveIgnDuplicateFilesAndUnits(t *testing.T) {
...
+
+	var testIgn3ConfigOld ign3types.Config
+	var testIgn3ConfigNew ign3types.Config
+
+	oldTempFile := helpers.NewMachineConfigFile("/etc/kubernetes/kubelet-ca.crt", "oldcertificates", 420)
+	newTempFile := helpers.NewMachineConfigFile("/etc/kubernetes/kubelet-ca.crt", "newcertificates", 420)
+
+	// Make an "old" config with the existing file in it
+	testIgn3ConfigOld.Ignition.Version = "3.2.0"
+	testIgn3ConfigOld.Storage.Files = append(testIgn3ConfigOld.Storage.Files, *oldTempFile)
```
mkenigs, line 395: We have a helper for creating MachineConfigs (https://github.com/openshift/machine-config-operator/blob/master/test/helpers/helpers.go#L37). Maybe I should add one with fewer parameters though :thinking:

test/helpers/helpers.go
jkyros: I don't need mode, and it is practically the same function, but I'd have to go fish it out of daemon to use it. I can take the mode argument off of the one I put in helpers?

pkg/controller/common/helpers_test.go
mkenigs, line 385: Would `TestCalculateConfigFileDiffs` be more conventional?

jkyros: Indeed it would!

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var (
+	// DefaultBindAddress is the port for the metrics listener
+	DefaultBindAddress = ":8797"
```
mkenigs, line 14: `const` instead of `var`?

jkyros: yes, we don't write to it, but in my defense I stole that verbatim from the daemon. :smiley: 

This is probably in line with Zack's comments about the boilerplate: https://github.com/openshift/machine-config-operator/pull/2802/#discussion_r765177360 whatever we fix here, we should fix across all our metrics listeners so there is "one true metrics listener" setup function.

pkg/controller/common/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// MachineConfigControllerPausedPoolKubeletCA logs when a certificate rotation is being held up by pause
+	MachineConfigControllerPausedPoolKubeletCA = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "machine_config_controller_paused_pool_kubelet_ca",
+			Help: "a certificate rotation is waiting in specified paused pool",
```
mkenigs, line 20: I do not fully understand the context of this PR, but I don't know which pool is "specified"

jkyros: Yeah, it makes more sense when the metrics actually come out of prometheus as a gauge. In the prometheus output it looks like: 
```
#HELP machine_config_controller_paused_pool_kubelet_ca a certificate rotation is waiting in specified paused pool
#TYPE machine_config_controller_paused_pool_kubelet_ca gauge
machine_config_controller_paused_pool_kubelet_ca{pool="master"} 0
machine_config_controller_paused_pool_kubelet_ca{pool="worker"} 1.670651677e+09
```

jkyros: I also don't think anyone ever actually looks at the HELP but I can't be sure, so I fill it in.

jkyros:
- Cleaned up all of the newlines :) 
- Made naming changes based on team suggestions 
- Fixed some typos and updated alert language to more clearly describe impact 
- Rebased + resolved conflicts

cmd/machine-config-controller/start.go
```diff
@@ -40,6 +40,7 @@ func init() {
 	rootCmd.AddCommand(startCmd)
 	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
 	startCmd.PersistentFlags().StringVar(&startOpts.resourceLockNamespace, "resourcelock-namespace", metav1.NamespaceSystem, "Path to the template files used for creating MachineConfig objects")
+	startCmd.PersistentFlags().StringVar(&startOpts.promMetricsURL, "metrics-url", "127.0.0.1:8797", "URL for prometheus metrics listener")
```
simonpasquier, line 43: strictly speaking, it's an address, not a URL so `metrics-listen-addres` would be more correct.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: mcc-paused-pool-kubelet-ca
+      rules:
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             sum by (pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
```
simonpasquier, line 18: as per the [alerting guidelines](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md), you should probably preserve the `namespace` label.

```suggestion
             sum by (namespace, pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
```

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+            runbook_url: https://github.com/openshift/blob/master/alerts/machine-config-operator/MachineConfigControllerPausedPoolKubeletCA.md
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             sum by (pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m]) - time()) < (86400 * 365) AND sum by (pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
+          for: 60m
```
simonpasquier, line 29: if the alert expression already delays the firing, it might not be necessary to add a `for` clause.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+          labels:
+            severity: critical
```
simonpasquier, line 31: ping @openshift/sre-alert-sme since it's adding a critical alert.

pkg/controller/common/helpers.go
sinnykumari: right, we will lose why exactly file changes. Let's keep the debug log as it is for now and we can move it in future if it start causing annoyance instead of being helpful.

pkg/controller/node/node_controller.go
```diff
@@ -1042,3 +1064,108 @@ func getErrorString(err error) string {
...
+	}
+
+	// We have other problems if this is empty, but it's possible
+	if len(containedCertificates) == 0 {
+		return nil, fmt.Errorf("No certificates found in bundle")
+	}
+
+	// The *original* signer has a different name, the rotated ones have longer names
+	// The suffix changes with the timstamp on rotation, which is why I'm using prefix here not exact match
+	newestCertificate := ctrlcommon.GetLongestValidCertificate(containedCertificates, kubeAPIToKubeletSignerNamePrefixes)
```
sinnykumari, line 1163: One question: Does looking for longest valid cert mean that we will not be sending critical alert for cert rotation that occurs at 365 days to delete old expired cert?

test/helpers/helpers.go
```diff
@@ -150,6 +151,16 @@ func NewMachineConfigPool(name string, mcSelector, nodeSelector *metav1.LabelSel
...
 }
 
+// NewIgnFile returns an ignition3 file object created from arguments passed to it
+func NewIgnFile(path, data string, mode int) *ign3types.File {
+	tempFile := ign3types.File{
+		Node: ign3types.Node{Path: path},
+		FileEmbedded1: ign3types.FileEmbedded1{Contents: ign3types.Resource{Source: StrToPtr(dataurl.EncodeBytes([]byte(data)))},
+			Mode: &mode}}
+
+	return &tempFile
```
mkenigs, line 161: Could you just `return &ign3types.File{}`?

pkg/controller/node/node_controller.go
jkyros: Correct, it would not go critical under that circumstance because the certificates would have too many days remaining. 

 - Do we consider the rotation for removal of that certificate critical? ( And would someone be able to quantify the bad thing that happens if that can't take place?) 
 
I can figure out if we're in that condition by doing something like "If the pending machine configs removed certificates from this bundle, and we still have plenty of days remaining, then this is the the 'cleanup' rotation".

sinnykumari: nice, this is expected behavior. Was just making sure that we are not going critical during cert removal change.

test/helpers/helpers.go
mkenigs: :see_no_evil: perfect one function in utils and one in helpers :joy:

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+          annotations:
+            summary: "Paused machine configuration pool '{{$labels.pool}}' is blocking a necessary certificate rotation and must be unpaused before the current kube-api-to-kubelet-signer expires in {{ $value | humanizeDuration }}."
+            description: "Machine config pools have a 'pause' feature, which allows config to be rendered, but prevents it from being rolled out to the nodes. This alert indicates that a certificate rotation has taken place, and the new kubelet-ca certificate bundle has been rendered into a machine config, but because the pool '{{$labels.pool}}' is paused, the config cannot be rolled out to the nodes in that pool. You will notice almost immediately that for nodes in pool '{{$labels.pool}}', pod logs will not be visible in the console and interactive commands (oc log, oc exec, oc debug, oc attach) will not work. You must unpause machine config pool '{{$labels.pool}}' to let the certificates through before the kube-api-to-kubelet-signer certificate expires. You have approximately {{ $value | humanizeDuration }} remaining before this happens and nodes in '{{$labels.pool}}' cease to function properly."
```
dofinn, line 34: This is an excellent description!

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
dofinn: should also preserve `namespace` label here as per @simonpasquier previous comment ?  Else LGTM

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
simonpasquier: I'm also not sure that `sum()` is the correct aggregation to use. `max()` is probably a better choice.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
sinnykumari: @dofinn From your SRE experience, any thoughts on how many days in advance alert should be moved to critical before cluster gets unresponsive?

pkg/controller/node/node_controller.go
```diff
@@ -64,8 +67,14 @@ const (
 
 	// masterPoolName is the control plane MachineConfigPool name
 	masterPoolName = "master"
+
+	// kubeletCAFilePath is the expected file path for the kubelet ca
+	kubeletCAFilePath = "/etc/kubernetes/kubelet-ca.crt"
 )
 
+// kubeAPIToKubeletSignerNamePrefixes is the list of subject common names that are regarded as a kube-apiserver-to-kubelet-signer ca certificate
+var kubeAPIToKubeletSignerNamePrefixes = []string{"openshift-kube-apiserver-operator_kube-apiserver-to-kubelet-signer@", "kube-apiserver-to-kubelet-signer"}
```
sinnykumari, line 76: Do we know that these are exclusive prefix or there can be some other prefix too?

jkyros: It's a combination of the namespace,  the secret name, and the timestamp, and the generation happens via the libary-go functions: 
 https://github.com/openshift/library-go/blob/ed9bc958bd8a2fff079d52976806e4e0a8a7c315/pkg/operator/certrotation/signer.go#L132
 
 Which the apiserver-operator has vendored here: 
 https://github.com/openshift/cluster-kube-apiserver-operator/blob/b98a8911dfb04c5c8ccc66dc3bb8c9feebabe4f8/vendor/github.com/openshift/library-go/pkg/operator/certrotation/signer.go#L133
 
 So as long as we don't change the names of the secrets or the namespaces, we should only ever have:
- the original name (non-prefixed/suffixed, not named by library-go) CN=kube-apiserver-to-kubelet-signer
-  the rotated names (prefixed/suffixed, named by library-go) CN=openshift-kube-apiserver-operator_kube-apiserver-to-kubelet-signer@timestamp

sinnykumari: great, thanks for detail!

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
dofinn: @sinnykumari I have not encountered this situation yet, one consideration in managed OpenShift space would be hibernation and potential implications of paused pools with respect to this alert. https://issues.redhat.com/browse/SDE-1173

pkg/controller/node/node_controller.go
```diff
@@ -995,3 +1011,74 @@ func getErrorString(err error) string {
 	}
 	return ""
 }
+
+// alertIfImportantFilesPending checks to see if there are any important files in the
+// machineconfig that the pool should be moving to, and alerts if there are
+func (ctrl *Controller) alertIfImportantFilesPending(pool *mcfgv1.MachineConfigPool) {
```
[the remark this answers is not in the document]
jkyros: Agreed. I've changed it to setPendingFileMetrics/resetPendingFileMetrics

test/e2e/mcc_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"testing"
+
+	"github.com/openshift/machine-config-operator/test/framework"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test case for https://github.com/openshift/machine-config-operator/issues/358
+func TestMCCPausedKubeletCAAlert(t *testing.T) {
```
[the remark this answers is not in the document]
jkyros: This is really cool. Thank you. I don't have it working cleanly yet, so it might not make it in for this PR but I appreciate it.

cmd/machine-config-controller/start.go
jkyros: Updated to reference listen address. This was copied more or less verbatim from the machine-config-daemon, and while I'd like to change it there too, I don't know if anyone actually uses that parameter over there (we don't), so I'm going to lave that one for now until I'm sure it's not going to break anyone.

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
jkyros: Thank you, I've preserved the namespace and switched from sum() to max()

pkg/controller/common/metrics.go
jkyros: I've refactored StartMetricsListener to take a function argument as suggested, and adjusted daemon to call it from controller/common.

pkg/controller/common/metrics.go
jkyros: I adjusted it to return if it fails to register, but I didn't have it return the error all the way up, I assumed we didn't want the failure of metrics registration to be fatal/impact anything else aside from the metrics handler. 

If I'm wrong and we *do* want it to be fatal for some reason, let me know :)

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
```diff
@@ -1,5 +1,41 @@
...
+            runbook_url: https://github.com/openshift/blob/master/alerts/machine-config-operator/MachineConfigControllerPausedPoolKubeletCA.md
+        - alert: MachineConfigControllerPausedPoolKubeletCA
+          expr: |
+             max by (namespace,pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m]) - time()) < (86400 * 365) AND max by (namespace,pool) (last_over_time(machine_config_controller_paused_pool_kubelet_ca[5m])) > 0
```
jkyros, line 28: @dofinn we were thinking we would trigger critical with 14 days remaining in the certificate life (if still paused) -- we wanted to make sure there was enough time to schedule an outage to unpause the pool. Do you have any thoughts on the timing here?  Is that enough time? Too much? Just right? 

(It's set to < 365 days right now for testing so it always alerts, but I was going to change it to 14 before merge unless stronger opinions manifest)

install/0000_90_machine-config-operator_01_prometheus-rules.yaml
sinnykumari: That's interesting usecase and will be interesting to tackle the cert rotation problem when cluster is in hibernation for long time

pkg/controller/common/metrics.go
cheesesashimi: In terms of error-handling, I think this is fine for now; we can revisit bubbling the error further later. A further improvement would be adding an additional error line (or tweaking the existing one) to say something like, "could not start metrics listener", just so it's clear that the metrics listener is not running.

simonpasquier: In practice, a failure during metric registration is a bug in your code (you registered the same metric twice for instance) and should be treated as a hard failure. Which is why `prometheus.MustRegister()` exists so you don't have to worry about the potential error.

simonpasquier (commented):
`RegisterMCCMetrics()` and  `RegisterMCCMetrics()` both use the global Prometheus registry but I suppose that they're called together in the same binary?
For clarity, it might be better that the caller creates the registry and pass it to the StartMetricsListener() and Register*Metrics() functions.

manifests/machineconfigcontroller/deployment.yaml
```diff
@@ -27,10 +27,42 @@ spec:
             cpu: 20m
             memory: 50Mi
         terminationMessagePolicy: FallbackToLogsOnError
+      - name: oauth-proxy
```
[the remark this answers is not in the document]
jkyros: Because I didn't know about it :) 

It sounds like we probably should use it -- I set this up to mirror what the machine-config-daemon does, since we already had precedent, and it uses oauth-proxy. I probably won't change the machine-config-controller to operate this way today since I don't fully understand all of the ramifications, but I have added a JIRA ticket to investigate this as part of our monitoring improvement epic https://issues.redhat.com/browse/MCO-128

jkyros:
Rebased and squashed. 
- I rolled back the metrics handler "unification" changes, that body of work will be addressed here: https://issues.redhat.com/browse/MCO-134
- Thanks to Simon for the kube-rbac-proxy info, that body of work will be addressed here: https://issues.redhat.com/browse/MCO-128
- This now uses the helpers that were factored out into #2870
- Alert is no longer set to testing value and will go critical with 14 days remaining on the certificate

sinnykumari:
https://github.com/openshift/machine-config-operator/pull/2802/commits/c9eeef1507a8e98a5e0c171aab5bfccfe4a0f593 commit is shouldn't be part of this PR since https://github.com/openshift/machine-config-operator/pull/2870 has already been merged. Instead just remove the PR and rebase your branch with MCO master

jkyros:
> [c9eeef1](https://github.com/openshift/machine-config-operator/commit/c9eeef1507a8e98a5e0c171aab5bfccfe4a0f593) commit shouldn't be part of this PR since #2870 has already been merged. Instead just remove the PR and rebase your branch with MCO master

This is rebased on master. #2870 added the helpers, but it looks like it didn't update the daemon to use them.  Master still had https://github.com/openshift/machine-config-operator/blob/4f93ba390d3c10b86cc8b530c6faa0d180b15127/pkg/daemon/update.go#L431

mkenigs:
I removed the duplicate code in https://github.com/openshift/machine-config-operator/pull/2851 which is approved but not yet merged

jkyros:
Okay, cool,  that makes sense. I've pulled that commit out.

sinnykumari (approved):

sinnykumari:
Thanks John, nice work :tada: 
I will let someone take a final look and add lgtm

yuqi-zhang:
/lgtm

Did a sanity check and I think this can merge. Good descriptions and clearly documented followups. Great work everyone!

sinnykumari:
Trying upgrade on some other platform
/test e2e-gcp-upgrade

sinnykumari:
With Feature Freeze in place, if we want to get this PR merged it will require a valid bug.

aravindhp:
> With Feature Freeze in place, if we want to get this PR merged it will require a valid bug.

@sinnykumari please don't merge this feature as a bug

sinnykumari:
> > With Feature Freeze in place, if we want to get this PR merged it will require a valid bug.
> 
> @sinnykumari please don't merge this feature as a fix for a bug

ack.

Putting hold so that we don't merge this feature work in 4.10 unless anything changes.
/hold

pkg/controller/node/node_controller.go
```diff
@@ -1042,3 +1065,108 @@ func getErrorString(err error) string {
...
+	}
+}
+
+// resetPendingFileMetrics turns off any "paused file" metrics that were firing for the pool
+func (ctrl *Controller) resetPendingFileMetrics(pool *mcfgv1.MachineConfigPool) {
+	// Set the metric for this pool back to zero
+	ctrlcommon.MachineConfigControllerPausedPoolKubeletCA.WithLabelValues(pool.Name).Set(0)
+}
+
+// parseConvertMachineConfigFilesForPool retrieves the current and pending configurations for a pool, parses and converts them, and
```
kikisdeliveryservice, line 1108: can we add some mention about metrics in docstring, bc it wasn't clear to me how this was going to be used given the name and current docstring.

pkg/controller/node/node_controller.go
```diff
@@ -1042,3 +1065,108 @@ func getErrorString(err error) string {
...
+	}
+	pendingIgnConfig, err := ctrlcommon.ParseAndConvertConfig(pendingConfig.Spec.Config.Raw)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &currentIgnConfig, &pendingIgnConfig, nil
+}
+
+// getNewestAPIToKubeletSignerCertificate returns the newest kube-apiserver-to-kubelet-signer certificate present in the
```
kikisdeliveryservice, line 1142: same comment as above, a bit of the why in this doctstring might be helpful for future us.

kikisdeliveryservice (commented):
As someone who didnt' read this until now, looks good but 2 small requests so things are clear to unfamiliar readers on a skim.

jkyros:
Added some more detailed comments to those functions as requested and re-pushed.

test/e2e/mcc_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/openshift/machine-config-operator/test/framework"
+	"github.com/openshift/machine-config-operator/test/helpers"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
```
kikisdeliveryservice, line 14: Why add an e2e test for an alert if there's a unit test? Shouldn't the unit test be sufficient?

jkyros: The unit test is good for us checking the metric, the e2e test was to make sure that monitoring would be able to retrieve it (secrets, auth, etc were all set up right). 

Long term we talked about having it talk to the thanos query endpoint to make sure monitoring actually got it (like Simon was saying here: https://github.com/openshift/machine-config-operator/pull/2802#discussion_r755919982)

pkg/controller/common/helpers.go
```diff
@@ -740,3 +743,50 @@ func GetIgnitionFileDataByPath(config *ign3types.Config, path string) ([]byte, e
 	}
 	return nil, nil
 }
+
+// GetNewestCertificatesFromPEMBundle breaks a pem-encoded bundle out into its component certificates
+func GetCertificatesFromPEMBundle(pemBytes []byte) ([]*x509.Certificate, error) {
+	var certs []*x509.Certificate
+	// There can be multiple certificates in the file
+	for {
```
kikisdeliveryservice, line 751: wait sorry - do we really want an infinite loop here? could we make this a little more explicit?

kikisdeliveryservice: Future John will come back to this later to add clarifying comments which is fine with me.

PR #2811, opened 2021-10-27
Bug 2017756: Remove crio settings that overwrite /etc/containers/storage.conf

palonsoro:
This intends to fix https://bugzilla.redhat.com/show_bug.cgi?id=2017756

The way it does is:
- Removing crio options from `/etc/crio/crio.conf.d/00-default` that may overwrite changes in `/etc/containers/storage.conf` introduced by `ContainerRuntimeConfig` custom resources.
- Wiping default `/etc/crio/crio.conf`, as it also includes the offending settings and any other interesting default has already been moved to MCO-managed configuration `/etc/crio/crio.conf.d/00-default`.

### DIFF
--- a/templates/master/01-master-container-runtime/_base/files/crio.yaml
+++ b/templates/master/01-master-container-runtime/_base/files/crio.yaml
@@ -4,10 +4,6 @@ contents:
   inline: |
     [crio]
     internal_wipe = true
-    storage_driver = "overlay"
-    storage_option = [
-        "overlay.override_kernel_check=1",
-    ]
     version_file_persist = "/var/lib/crio/version"
 
     [crio.api]
--- a/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
+++ b/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
@@ -4,10 +4,6 @@ contents:
   inline: |
     [crio]
     internal_wipe = true
-    storage_driver = "overlay"
-    storage_option = [
-        "overlay.override_kernel_check=1",
-    ]
     version_file_persist = "/var/lib/crio/version"
 
     [crio.api]


yuqi-zhang:
> The one risk this presents is there may be customers who have customized crio.conf and not used a drop-in. Since we've been recommending a drop-in since 4.5, I think this represents a small percentage of customers.

Theoretically, so long as they are providing it via a machineconfig that alphanumerically overrides the base "wipe" conf we are shipping here, they should see no change to that file.

That said, I'm not sure how I feel about us shipping "wipe" files in general. If we think a file should not be there, I am still leaning towards removing this from the default OS in the first place somehow, and not have the MCO attempt "removing" local files like this.

haircommander:
> That said, I'm not sure how I feel about us shipping "wipe" files in general. If we think a file should not be there, I am still leaning towards removing this from the default OS in the first place somehow, and not have the MCO attempt "removing" local files like this.

yeah I see what you mean, I wasn't sure if this approach was totally idiomatic but seemed like a good solution. One way we can "remove" (at least for new installs) is drop the config file from the rpm, which I am attempting now. If it doesn't break anything :crossed_fingers: then 4.10 installs and beyond will not have it at least

templates/master/01-master-container-runtime/_base/files/crio-conf-wipe.yaml
```diff
@@ -0,0 +1,4 @@
+mode: 0644
+path: "/etc/crio/crio.conf"
+contents:
+  inline: ''
```
haircommander, line 4: per @yuqi-zhang 's comments, I think we should drop this

palonsoro: Only as long as we ensure that we are shipping no such file within RPM. Can we get confirmation of that somehow?

haircommander: I had attempted to drop the file from the rpm but I broke the rpm in the process, I think I've fixed. once it's trickled into the images I'll build a cluster to prove it's been done :)

haircommander:
I think we should pivot a bit here. I don't think we should overwrite /etc/crio/crio.conf, nor do I think we should not ship it in the RPM. The reasoning for the latter is because without cri-o being configured OOTB to update it's cni plugin dir, cri-o starts but doesn't accept container creation requests and basically endlessly stalls. Even though no one is using RHCOS+CRI-O without MCO, it feels too weird to have that behavior come from the vanilla RPM. 

Instead, I have dropped the storage related fields in the crio.conf from the RPM, and I think we should do the same here. then, we will ineherit it from /etc/containers/storage.conf, which correctly configures everything.

WDYT @palonsoro

palonsoro:
If you have removed the storage related overrides from crio.conf on RPM, and that removal can be backported to crio versions used in older OCPs (as necessary), I think I can just drop the crio.conf modification completely and only leave the rest of the changes. Please confirm if that makes sense for you and I'll just go ahead and do it.

palonsoro:
@haircommander ^^^

haircommander:
yup sounds good to me, thanks @palonsoro

palonsoro:
Done.

haircommander:
/lgtm

thanks for sticking with this and waiting while we figured out the solution here :)

yuqi-zhang (approved):
New changes lgtm

palonsoro:
Thanks to you for all your assistance

PR #2814, opened 2021-10-28
Alibaba Cloud Provider: use --provider-id kubelet parameter

kwoodson:
**- What I did**
Alibaba Cloud provider requires the `--provider-id` kubelet parameter instead of `--hostname-override`. 

I added custom kubelet.service files for `alibabacloud`.

**- How to verify it**
Install a cluster on Alibaba and verify that the control plane nodes have the proper settings in `/cat /etc/systemd/system/kubelet.service`.

AlibabaCloud provider-id kubelet flag

### DIFF
--- a/templates/master/01-master-kubelet/alibabacloud/units/kubelet.service.yaml
+++ b/templates/master/01-master-kubelet/alibabacloud/units/kubelet.service.yaml
@@ -1,0 +1,52 @@
+name: kubelet.service
+enabled: true
+contents: |
+  [Unit]
+  Description=Kubernetes Kubelet
+  Wants=rpc-statd.service network-online.target
+  Requires=crio.service kubelet-auto-node-size.service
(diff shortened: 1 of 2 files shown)


yuqi-zhang (approved):
Generally lgtm, would prefer someone with more alibaba cloud knowledge to take a look as well.

Interestingly enough we split the master/worker kubelet.conf manifests but they seem to be identical. I forget if there was some reason behind it

kwoodson:
@kwoodson Thanks! Unfortunately, I'm the Alibaba guy right now. I can get another on my team or another person to TAL if that helps. 

cc @elmiko @rvanderp3 Either of you mind reviewing?

The kubelet for control plane and for the worker are only different when it comes to node labels and taints.

elmiko (commented):
given what i know of alibaba, this looks correct to me
/lgtm

PR #2817, opened 2021-11-01
Pass parsed Ignition config to updateFiles

mkenigs:
Instead of passing machine configs that must be parsed, just pass the
already parsed Ignition configs

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -325,12 +325,8 @@ func removePendingDeployment() error {
 	return runRpmOstree("cleanup", "-p")
 }
 
-func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr error) {
+func (dn *Daemon) applyOSChanges(mcDiff machineConfigDiff, oldConfig, newConfig *mcfgv1.MachineConfig) (retErr error) {
 	// Extract image and add coreos-extensions repo if we have either OS update or package layering to perform
-	mcDiff, err := newMachineConfigDiff(oldConfig, newConfig)
-	if err != nil {
-		return err
-	}
 
 	if dn.recorder != nil {
 		dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "OSUpdateStarted", mcDiff.osChangesString())
@@ -356,6 +352,7 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 		if dn.recorder != nil {
 			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "InClusterUpgrade", fmt.Sprintf("Updating from oscontainer %s", newConfig.Spec.OSImageURL))
 		}
+		var err error
 		if osImageContentDir, err = ExtractOSImage(newConfig.Spec.OSImageURL); err != nil {
 			return err
 		}
@@ -481,7 +478,7 @@ func calculatePostConfigChangeActionFromFileDiffs(oldIgnConfig, newIgnConfig ign
 	return
 }
 
-func calculatePostConfigChangeAction(oldConfig, newConfig *mcfgv1.MachineConfig) ([]string, error) {
+func calculatePostConfigChangeAction(diff *machineConfigDiff, oldIgnConfig, newIgnConfig ign3types.Config) ([]string, error) {
 	// If a machine-config-daemon-force file is present, it means the user wants to
 	// move to desired state without additional validation. We will reboot the node in
 	// this case regardless of what MachineConfig diff is.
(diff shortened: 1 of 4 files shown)


yuqi-zhang (approved):
Generally lgtm!

Side note, we are not very consistent on passing by reference/value in the daemon code... although I suppose it's not too critical for our case, we technically shouldn't be modifying the objects anyways.

pkg/daemon/update.go
```diff
@@ -579,29 +579,29 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
+		return fmt.Errorf("parsing old Ignition config failed with error: %v", err)
+	}
+	newIgnConfig, err := ctrlcommon.ParseAndConvertConfig(newConfig.Spec.Config.Raw)
+	if err != nil {
+		return fmt.Errorf("parsing new Ignition config failed with error: %v", err)
+	}
+
 	// update files on disk that need updating
-	if err := dn.updateFiles(oldConfig, newConfig); err != nil {
+	if err := dn.updateFiles(&oldIgnConfig, &newIgnConfig); err != nil {
```
sinnykumari, line 592: Since we are not going to modify params, let's pass them by value.

mkenigs: `updateFiles` calls `deleteStaleData` which takes references, but I could just pass references to the copy.

In general, when is it good practice to pass by value? Is it just a tradeoff between safety/no garbage collection vs size of a copy?

mkenigs:
Just saw looks like we should pass the parsed files to reconcilable as well, so I'll change that too

mkenigs:
Pushed a few more commits. If they look good, do we usually squash all commits for a PR?

sinnykumari:
> Pushed a few more commits. If they look good, do we usually squash all commits for a PR?

Yes, squashing makes sense in this case because all changes belongs to one logical change. If you are tackling multiple small features or fixes in same PR, it can have multiple commits.

mkenigs:
Squashed. If you've already taken a look at a commit, is it good for me to leave it split until you see it, or just force push it?

pkg/daemon/update.go
```diff
@@ -356,7 +352,8 @@ func (dn *Daemon) applyOSChanges(oldConfig, newConfig *mcfgv1.MachineConfig) (re
 		if dn.recorder != nil {
 			dn.recorder.Eventf(getNodeRef(dn.node), corev1.EventTypeNormal, "InClusterUpgrade", fmt.Sprintf("Updating from oscontainer %s", newConfig.Spec.OSImageURL))
 		}
-		if osImageContentDir, err = ExtractOSImage(newConfig.Spec.OSImageURL); err != nil {
+		osImageContentDir, err := ExtractOSImage(newConfig.Spec.OSImageURL)
```
sinnykumari, line 355: This is not idiomatic in go. You can leave the way it was earlier. Also, osImageContentDir var has already been created earlier in the function, so you are limiting the scope of this var by recreating a local scoped variable.

sinnykumari:
Note that the unit tests also need to be fixed.

@cheesesashimi @jkyros a round of review from you would be great as well :)

pkg/daemon/update.go
```diff
@@ -541,6 +525,15 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
 	oldConfigName := oldConfig.GetName()
 	newConfigName := newConfig.GetName()
 
+	oldIgnConfig, err := ctrlcommon.ParseAndConvertConfig(oldConfig.Spec.Config.Raw)
+	if err != nil {
+		return fmt.Errorf("parsing old Ignition config failed with error: %v", err)
```
cheesesashimi, line 530: Instead of `%v`, use `%w`. This will allow the error from `ctrlcommon.ParseAndConvertConfig()` to be cleanly wrapped inside the error this returns. This will be useful at a later date when we revamp how we handle errors.

pkg/daemon/update.go
mkenigs: Redefined `err` instead

pkg/daemon/update.go
mkenigs: Looks like I still have to use `%v` with `t.Errorf` though?

mkenigs:
Fixed unit tests

pkg/daemon/update.go
```diff
@@ -580,28 +573,19 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
 	// update files on disk that need updating
-	if err := dn.updateFiles(oldConfig, newConfig); err != nil {
+	if err := dn.updateFiles(oldIgnConfig, newIgnConfig); err != nil {
 		return err
 	}
 
 	defer func() {
 		if retErr != nil {
-			if err := dn.updateFiles(newConfig, oldConfig); err != nil {
+			if err := dn.updateFiles(oldIgnConfig, newIgnConfig); err != nil {
```
sinnykumari, line 582: any reason to swap oldConfig and newConfig param? defer function is supposed to get called when an error occurred in between and we would try to revert back the changes we applied.

pkg/daemon/update.go
```diff
@@ -627,13 +611,13 @@ func (dn *Daemon) update(oldConfig, newConfig *mcfgv1.MachineConfig) (retErr err
...
 
-	if err := dn.applyOSChanges(oldConfig, newConfig); err != nil {
+	if err := dn.applyOSChanges(*diff, oldConfig, newConfig); err != nil {
 		return err
 	}
 
 	defer func() {
 		if retErr != nil {
-			if err := dn.applyOSChanges(newConfig, oldConfig); err != nil {
+			if err := dn.applyOSChanges(*diff, oldConfig, newConfig); err != nil {
```
sinnykumari, line 620: same question https://github.com/openshift/machine-config-operator/pull/2817/files#r742665160

PR #2818, opened 2021-11-01
Bug 1966445: Cleanup systemd unit files for empty units

mkenigs:
Empty systemd units should ensure no files exist for that unit

This fixes edge cases such as when a previous symlink to /dev/null is
never removed, which inappropriately continues to mask the unit

Closes: #1966445

### DIFF
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1474,11 +1474,11 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 			return err
 		}
 
+		// write (or cleanup) path in /etc/systemd/system
 		fpath := filepath.Join(pathSystemd, u.Name)
-
-		// check if the unit is masked. if it is, we write a symlink to
-		// /dev/null and continue
 		if u.Mask != nil && *u.Mask {
+			// if the unit is masked, symlink fpath to /dev/null and continue
+
 			glog.V(2).Info("Systemd unit masked")
 			if err := os.RemoveAll(fpath); err != nil {
 				return fmt.Errorf("failed to remove unit %q: %v", u.Name, err)
@@ -1501,12 +1501,21 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 					return err
 				}
 			}
-			// write the unit to disk
 			if err := writeFileAtomicallyWithDefaults(fpath, []byte(*u.Contents)); err != nil {
 				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
 			}
 
 			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		} else if u.Mask != nil && !*u.Mask {
+			// if mask is explicitly set to false, make sure to remove a previous mask
+			// see https://bugzilla.redhat.com/show_bug.cgi?id=1966445
+			// Note that this does not catch all cleanup cases; for example, if the previous machine config specified
+			// Contents, and the current one does not, the previous content will not get cleaned up. For now we're ignoring some
+			// of those edge cases rather than introducing more complexity.
+			glog.V(2).Infof("Ensuring systemd unit %q has no mask at %q", u.Name, fpath)
+			if err := os.RemoveAll(fpath); err != nil {
+				return fmt.Errorf("failed to cleanup %s: %v", fpath, err)
+			}
 		}
 
 		// if the unit doesn't note if it should be enabled or disabled then
--- a/pkg/daemon/update.go
+++ b/pkg/daemon/update.go
@@ -1430,41 +1430,48 @@ func (dn *Daemon) presetUnit(unit ign3types.Unit) error {
 	return nil
 }
 
+// write dropins to disk
+func (dn *Daemon) writeDropins(u ign3types.Unit) error {
+	for i := range u.Dropins {
+		dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
+		if u.Dropins[i].Contents == nil || *u.Dropins[i].Contents == "" {
+			glog.Infof("Dropin for %s has no content, skipping write", u.Dropins[i].Name)
+			if _, err := os.Stat(dpath); err != nil {
+				if os.IsNotExist(err) {
+					continue
+				}
+				return err
+			}
+			glog.Infof("Removing %q, updated file has zero length", dpath)
+			if err := os.Remove(dpath); err != nil {
+				return err
+			}
+			continue
+		}
+
+		glog.Infof("Writing systemd unit dropin %q", u.Dropins[i].Name)
+		if _, err := os.Stat("/usr" + dpath); err == nil &&
+			dn.os.IsCoreOSVariant() {
+			if err := createOrigFile("/usr"+dpath, dpath); err != nil {
+				return err
+			}
+		}
+		if err := writeFileAtomicallyWithDefaults(dpath, []byte(*u.Dropins[i].Contents)); err != nil {
+			return fmt.Errorf("failed to write systemd unit dropin %q: %v", u.Dropins[i].Name, err)
+		}
+
+		glog.V(2).Infof("Wrote systemd unit dropin at %s", dpath)
+	}
+	return nil
+}
+
 // writeUnits writes the systemd units to disk
 func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 	var enabledUnits []string
 	var disabledUnits []string
 	for _, u := range units {
-		// write the dropin to disk
-		for i := range u.Dropins {
-			dpath := filepath.Join(pathSystemd, u.Name+".d", u.Dropins[i].Name)
-			if u.Dropins[i].Contents == nil || *u.Dropins[i].Contents == "" {
-				glog.Infof("Dropin for %s has no content, skipping write", u.Dropins[i].Name)
-				if _, err := os.Stat(dpath); err != nil {
-					if os.IsNotExist(err) {
-						continue
-					}
-					return err
-				}
-				glog.Infof("Removing %q, updated file has zero length", dpath)
-				if err := os.Remove(dpath); err != nil {
-					return err
-				}
-				continue
-			}
-
-			glog.Infof("Writing systemd unit dropin %q", u.Dropins[i].Name)
-			if _, err := os.Stat("/usr" + dpath); err == nil &&
-				dn.os.IsCoreOSVariant() {
-				if err := createOrigFile("/usr"+dpath, dpath); err != nil {
-					return err
-				}
-			}
-			if err := writeFileAtomicallyWithDefaults(dpath, []byte(*u.Dropins[i].Contents)); err != nil {
-				return fmt.Errorf("failed to write systemd unit dropin %q: %v", u.Dropins[i].Name, err)
-			}
-
-			glog.V(2).Infof("Wrote systemd unit dropin at %s", dpath)
+		if err := dn.writeDropins(u); err != nil {
+			return err
 		}
 
 		fpath := filepath.Join(pathSystemd, u.Name)


pkg/daemon/update.go
```diff
@@ -1503,22 +1503,27 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
 			glog.V(2).Infof("Created symlink unit %q to %s", u.Name, pathDevNull)
 
 			continue
-		}
-
-		if u.Contents != nil && *u.Contents != "" {
+		} else if u.Contents != nil && *u.Contents != "" {
```
mkenigs, line 1506: I thought putting the else here makes this more readable, but not necessary because of the continue

pkg/daemon/update.go
```diff
@@ -1503,22 +1503,27 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
+			// write the unit to fpath
 			if err := writeFileAtomicallyWithDefaults(fpath, []byte(*u.Contents)); err != nil {
 				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
 			}
 
 			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		} else {
+			// otherwise fpath should be empty
+			glog.V(2).Info("Cleaning up systemd unit")
+			if err := os.RemoveAll(fpath); err != nil {
```
mkenigs, line 1523: Should this go in `deleteStaleData`? Wasn't sure, but I think cleaning up the symlinks from a mask needs to happen before preset

pkg/daemon/update.go
```diff
@@ -1503,22 +1503,27 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
 				}
 			}
-			// write the unit to disk
+			// write the unit to fpath
 			if err := writeFileAtomicallyWithDefaults(fpath, []byte(*u.Contents)); err != nil {
 				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
 			}
 
 			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		} else {
```
mkenigs, line 1520: Are there cases we need to leave something there? Or create an orig file or something?

pkg/daemon/update.go
yuqi-zhang: No strong feelings here but it does make the `else` condition a bit awkward below, I think

pkg/daemon/update.go
yuqi-zhang: So, let's say we got here, we just know that mask is nil and contents is nil, if I'm reading this correctly.

A user could have created dropins for an existing service, for example. It doesn't mean we should clean up the unit. If anything I don't think this should be gated on the else

pkg/daemon/update.go
yuqi-zhang: Yes I think attempted removal (in this case, diffing off a mask of a service) should go into deleteStateData I think.

Units are a bit more complex in this way. Just because mask and content are nil does not mean other subfields can be extrapolated. We should instead be checking (for this specific bug) to see whether a mask was previous defined and now gone.

yuqi-zhang (changes requested):
Added some comments based on my understanding but I could be missing some context, happy to discuss further if you have questions!

pkg/daemon/update.go
mkenigs: I should probably reword the comment, but as far as I can tell dropins go into `u.Name+".d"` rather than `u.Name`, which is what I'm removing here. Is there a case where mask is nil and contents is nil, but there should still be something in the file `u.Name`?

yuqi-zhang: An example: enabling or disabling a service that already exists on the system. To do so you can simply pass a machineconfig that contains that service, with no mask and no content, and it should enable. Example spec:

```
spec:
  config:
    ignition:
      version: 3.1.0
    systemd:
      units:
        - name: test.service
          enabled: true
```

The main issue with our assumptions here is that all the systemd services are fully defined via machineconfigs, which is not the case. There are many that are shipped by default on an RHCOS system, for example

mkenigs: Ok I was assuming those were presets, but I guess that's not true

mkenigs: How can we unmask something then? Masking overwrites everything at `fpath`, and I don't see a backup. If it's not a preset, how would we restore `fpath` when unmasking?

sinnykumari:
I may be missing some context about handling units in MCD. But I am wondering what will happen in the scenario that associated bug describes.
Suppose, I created a service with mask: true , MCD will see that mask is true so it will create a symlink for the service pointing to /dev/null. Now, I either edited the applied MC (umask: false) or deleted the MC. From the code flow it looks to me that it will write unit content and then will call systemctl enable <service>. Don't we first need to unlink/remove existing symlink that service was already pointing to?

pkg/daemon/update.go
yuqi-zhang: After discussing with Matthew, we think this actually should work. Although the current logic where if we have a unit with no mask and no content (e.g. enable/dropin) still feels minorly... incorrect and inefficient, although I can't think of a scenario that would cause issues.

One option if we want to make absolutely sure is instead of this as an else condition, just have it be an if condition that handles only mask=false, so we aren't "removing" something that isn't there on e.g. an enable operation, saving that extra RemoveAll call when the user did not define mask=false. For masks that are deleted straight up the deleteStaleData will handle it as usual. WDYT @mkenigs

yuqi-zhang (commented):
> I may be missing some context about handling units in MCD. But I am wondering what will happen in the scenario that associated bug describes.
Suppose, I created a service with mask: true , MCD will see that mask is true so it will create a symlink for the service pointing to /dev/null. Now, I either edited the applied MC (umask: false) or deleted the MC. From the code flow it looks to me that it will write unit content and then will call systemctl enable . Don't we first need to unlink/remove existing symlink that service was already pointing to?

So the codepath for deleting a mask and setting a mask to false will work out differently. Deleting a mask will simply have us rewrite over the mask'ed unit, if it exists, or if that was the last MC definition, our deleteStaleData so happens to think it just should remove the whole file, which works out (even though it didnt remove a unit, just a symlink to /dev/null in this case)

Setting a mask to false is what isn't handled today. There are a few associated problems with that as well. Namely, we technically "wipe" the service from /etc/systemd completely - no orig files or anything, under the assumption that we should be defining that file somewhere else in the MC and we can recover with that if we unmask. Fortunately this seems to be true, where the defaults shipped in rhcos go into /usr/lib/systemd/system (e.g. chronyd) which allows us to not actually wipe the file off the system. Such that either way we are able to restore it if we set mask=false.

As to overwriting the symlink with the unit definition again, I think it works (Matthew would know better) without having to remove the symlink first.

(A thought: technically the mask=false is a very weird directive for the MCO. Assuming we just supply masked=false first, it should just be a noop since there isn't anything to unmask. In the MCO it's almost like we should support masked=true -> masked or empty mask -> not masked)

pkg/daemon/update.go
mkenigs: A case that misses is `mask: true` -> `enabled: true`. Since both mask and content are nil, `fpath` won't get touched, and the unit would still be masked

mkenigs: Another bug (I think):
`content: someContent` -> `enabled: true`. Since content is nil for the second config, `fpath` doesn't get removed, and `someContent` still applies rather than whatever ships with RHCOS

mkenigs: I think we either need to accept buggy/weird behavior or use both the old units and the new units to decide what action to take

mkenigs:
@yuqi-zhang I think this is in line with what you said now

pkg/daemon/update.go
```diff
@@ -1494,12 +1501,20 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
-			// write the unit to disk
 			if err := writeFileAtomicallyWithDefaults(fpath, []byte(*u.Contents)); err != nil {
 				return fmt.Errorf("failed to write systemd unit %q: %v", u.Name, err)
 			}
 
 			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		} else if u.Mask != nil && !*u.Mask {
+			// if mask is explicitly set to false, make sure to remove a previous mask
+			// see https://bugzilla.redhat.com/show_bug.cgi?id=1966445
+			// note that this does not catch all cleanup cases; for example, if the previous machine config specified Contents, and the current one does not, the previous content will not get cleaned up. For now we're ignoring some of those edge cases rather than introducing more complexity.
```
yuqi-zhang, line 1512: This comment line is a bit long, could you wrap it such that it would be easier to read in an editor? (For example, see below for longer comments)

pkg/daemon/update.go
```diff
@@ -1494,12 +1501,20 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
+			glog.V(2).Info("Ensuring systemd unit is not masked")
```
yuqi-zhang, line 1513: Might be easier for debugging if we added unit path as well here in the log line. (Side note, I believe V(2) is the default, so I think we can just make it glog.infof)

yuqi-zhang (commented):
Overall looks good! Some nits below:

pkg/daemon/update.go
mkenigs: Yes! Was actually wondering, do we have a standard for line length? Could you add that to new hire docs? Or probably even better, could we enable `lll` in `golangci-lint`?

pkg/daemon/update.go
mkenigs: Everything else in the method uses `V(2)`, do you want me to make it inconsistent?

mkenigs: Actually I guess it's a mix

yuqi-zhang: Yeah I think we were never that good on the consistency front. Not a major issue either way

pkg/daemon/update.go
yuqi-zhang: > Was actually wondering, do we have a standard for line length? Could you add that to new hire docs?

Good question. I am not sure. Will see if we have any docs on this. I always just went with whatever the standard was for the rest of the file.

> Or probably even better, could we enable lll in golangci-lint?

That might be a good option (although we'd probably have to fix a bunch of comments by hand elsewhere)

mkenigs: Just wrapped manually for now

pkg/daemon/update.go
```diff
@@ -1494,12 +1501,22 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
 			}
 
 			glog.V(2).Infof("Successfully wrote systemd unit %q: ", u.Name)
+		} else if u.Mask != nil && !*u.Mask {
+			// if mask is explicitly set to false, make sure to remove a previous mask
+			// see https://bugzilla.redhat.com/show_bug.cgi?id=1966445
+			// Note that this does not catch all cleanup cases; for example, if the previous machine config specified
+			// Contents, and the current one does not, the previous content will not get cleaned up. For now we're ignoring some
+			// of those edge cases rather than introducing more complexity.
+			glog.V(2).Info("Ensuring systemd unit %q has no mask at %q", u.Name, fpath)
```
yuqi-zhang, line 1515: The verify test is failing on this with:
```
printf: (github.com/golang/glog.Verbose).Info call has possible formatting directive %q (govet)
```

So Info -> Infof

pkg/daemon/update.go
```diff
@@ -1494,12 +1501,22 @@ func (dn *Daemon) writeUnits(units []ign3types.Unit) error {
...
+			if err := os.RemoveAll(fpath); err != nil {
+				return fmt.Errorf("failed to cleanup %s: %v", fpath, err)
+			}
+			glog.V(2).Infof("Removed %q", fpath)
```
yuqi-zhang, line 1519: I think we can drop this comment. It doesn't always remove the file since if the file doesn't currently exist, RemoveAll also returns no error. (we also cover this with the `Ensuring systemd unit %q has no mask at %q` above)

yuqi-zhang (commented):
Mostly lgtm, 2 comments below:

pkg/daemon/update.go
mkenigs: Removed

pkg/daemon/update.go
mkenigs: Fixed!

PR #2819, opened 2021-11-02
Allow blocking payload reg if it has mirrors configured

umohnani8:

This relates to https://issues.redhat.com/browse/RFE-2166

**- What I did**
Some disconnected use cases require adding the payload reg
to the list of blocked registries for certain compliance reasons.
So allow adding the payload registry to the list of blocked registries
if we find a mirror configuration set for it.

**- How to verify it**
Bring up a cluster and check what registry the payload is being pulled from. Then test the following cases:
- Add the payload registry to the blocked registry list --> the payload reg should not be added under blocked registries in /etc/containers/registries.conf
- Create a mirror rule using ICSP for the payload reg and then add the payload reg to the blocked registry list --> the payload reg should be added under the blocked registries in /etc/containers/registries.conf and there should be a warning in the logs about adding the payload registry to the list of blocked registries.

Allow blocking the payload registry if there are mirror rules set for it.


pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -849,12 +867,12 @@ func TestRegistriesValidation(t *testing.T) {
...
 		imgcfg := newImageConfig(test.name, test.config)
 		cvcfg := newClusterVersionConfig("version", "blah.io/myuser/myimage:test")
-		blocked, err := getValidBlockedRegistries(cvcfg.Status.Desired.Image, &imgcfg.Spec)
+		blocked, err := getValidBlockedRegistries(cvcfg.Status.Desired.Image, &imgcfg.Spec, test.icspRules)
 		if err != nil {
 			t.Errorf("%s: failed", test.name)
 		}
 		for _, reg := range blocked {
-			if reg == "blah.io" {
+			if reg == "blah.io" && len(test.icspRules) == 0 {
```
mtrmac, line 875: This is a non-obvious dependency between the test contents and mechanism; a new `expectedBlocked` member of `successTetss` would be cleaner.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -496,15 +496,38 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
 
+// payloadRegHasMirrorConfig returns true if the payload registry has mirror rules configured for it
+func payloadRegHasMirrorConfig(payloadReg string, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy) bool {
+	if len(icspRules) == 0 || icspRules == nil {
+		return false
+	}
+	// Go through the mirror rules configured and check if a source matches the payload registry
+	for _, rule := range icspRules {
+		for _, mirror := range rule.Spec.RepositoryDigestMirrors {
+			if strings.Contains(mirror.Source, payloadReg) {
```
mtrmac, line 523: (`Contains` is very imprecise. I get that this is a best-effort code to prevent breaking clusters, it just offends my aesthetic sensibilities :) .  https://github.com/openshift/runtime-utils/blob/8b8348d80d1d1e7b6cf06fb009d5965e0b55baa2/pkg/registries/registries.go#L13 can probably be used for an exact check.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -496,15 +496,38 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
 	for i, reg := range imgSpec.RegistrySources.BlockedRegistries {
```
mtrmac, line 496: Above we compute `payloadReg`. Do we know that all of the release is in the `releaseImage` repo? If that’s the case, we could probably make a much more accurate check, especially WRT interpreting ICSP correctly.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -496,15 +496,38 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
 		// if there is a match, return all the blocked registries except the one that matched and return an error as well
 		if reg == payloadReg {
-			blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
-			return blockedRegs, fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", payloadReg)
+			// If the payload registry doesn't have mirror rules configured for it, then don't add it to the blocked registries list
```
mtrmac, line 499: Mirrors can also be blocked; so this should probably verify that the mirrors that are defined are not blocked either.

Then the ugly questions:

And if there are multiple mirrors, and only a subset is blocked, I guess we allow that, as long as there is one unblocked mirror, arguing that a blocked mirror is the same as a blocked primary, and not _alone_ a reason to reject the whole configuration.

Alternatively, I could see an argument that any definition of a mirror pointing at a blocked repo is by definition invalid, and should be rejected. OTOH this controller can not “reject” configuration, it can only “ignore” it, and I’m not at all sure to what extent ignoring admin’s security policy to block a registry is acceptable.

mtrmac (commented):
I know, WIP — just a few things to possibly consider. (Or maybe this has already been discussed and designed.)

pkg/controller/container-runtime-config/helpers.go
umohnani8: Yes, all the release is in the releaseImage repo - how could we make a more accurate check here?

mtrmac: Conceptually:
- Build the effective registries.conf from all the ICSP, processing overrides/merging etc. = `EditRegistriesConfig`; ideally by only calling that once in this package, and doing this registries validation on the outcome, instead of calling `EditRegistriesConfig` twice.
- Look up the effective configuration for 1) `releaseRepo` = repo of `releaseImage` 2) _any_ configuration for `releaseRepo:$anyTag` or `releaseRepo@anyDigest`. That’s basically `pkg/registries.scopeIsNestedInsideScope(entryInQuestion.{Prefix,Location}, releaseRepo`.
  - Or, alternatively, `c/image/pkg/sysregistriesv2.FindRegistry(releaseRepo)` can be used to find one matching entry, but not possible `:anyTag`/`@anyDigest` subscopes.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: We are calling `EditRegistriesConfig` only once. We are doing this validation on the initial input of blocked registries before calling EditRegistriesConfig on it.

pkg/controller/container-runtime-config/helpers.go
umohnani8: Added a check to go through the mirrors configured to check if a mirror has been configured for the payload reg, if so then verify that at least one of the mirrors configured is not blocked.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
mtrmac: `expectedBlocked` is never set now?!

I was thinking
```go
expectedBlocked []string
…
assert.Equal(t, test.expectedBlocked, blocked)
```
and the `"blah.io"` constant is moved from the test code into the data.

pkg/controller/container-runtime-config/helpers.go
mtrmac: Somehow GitHub is not showing the code this comments on — but if this was was a “`reg` is blocking `payloadReg`”, I now see just a `==` compare; shouldn’t that be `ScopeIsNestedInsideScope(payloadReg, reg)`?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
+			// TODO maybe? Add an opt-in configuration to ICSP and check if that is set as well
+			if !payloadRegHasMirrorConfig(payloadReg, icspRules, imgSpec.RegistrySources.BlockedRegistries) {
+				blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
```
mtrmac, line 519: Assuming the `reg == blockedReg` above becomes a `ScopeIsNestedInsideScope`, it will be possible for more than one `blockedReg` value to be invalid, and in this case each needs to be individually checked and possibly rejected; we can’t terminate on a first violation.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
 	for i, reg := range imgSpec.RegistrySources.BlockedRegistries {
 		// if there is a match, return all the blocked registries except the one that matched and return an error as well
 		if reg == payloadReg {
-			blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
-			return blockedRegs, fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", payloadReg)
+			// If the payload registry doesn't have mirror rules configured for it, then don't add it to the blocked registries list
+			// TODO maybe? Add an opt-in configuration to ICSP and check if that is set as well
```
mtrmac, line 517: I have quite likely forgotten; from this comment, I don’t understand what the “opt-in configuration to ICSP” is supposed to opt into.

pkg/controller/container-runtime-config/helpers.go
mtrmac: … actually, I was overcomplicating things. In `EditRegistries` a single `blockedScopes` entry can create multiple `sysregistriesv2.Registry` entries, but they always cover the _full_ subtree of that blocked scope; there is no way to “unblock” a sub-scope.

So, just a `ScopeIsNestedInsideScope(releaseRepo, blockedReg)` should determine if that `blockedReg` value affects `releaseRepo`. That would allow blocking `quay.io/foo` without limitation, because it doesn’t affect `releaseRepo`.

pkg/controller/container-runtime-config/helpers.go
mtrmac: … actually, `payloadReg` is the whole registry. But see the other comment, we ideally want to match any `reg` that affects `releaseRepo`.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
+// payloadRegHasMirrorConfig returns true if the payload registry has mirror rules configured for it
+func payloadRegHasMirrorConfig(payloadReg string, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy, blockedRegs []string) bool {
+	if len(icspRules) == 0 || icspRules == nil {
+		return false
+	}
+	var payloadMirrors []string
+	// Go through the mirror rules configured and check if a source matches the payload registry
+	for _, rule := range icspRules {
+		for _, mirror := range rule.Spec.RepositoryDigestMirrors {
+			if runtimeutils.ScopeIsNestedInsideScope(mirror.Source, payloadReg) {
```
mtrmac, line 541: This will match `mirror.Source == "quay.io/foo"` with `payloadReg == "quay.io"`.

So we probably do _need_ the more accurate `payloadRepo`, and then this would become `ScopeIsNestedInsideScope(payloadRepo, mirror.Source)`.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
+				payloadMirrors = mirror.Mirrors
```
mtrmac, line 542: Shouldn’t this append `mirror.Mirrors`, and continue collecting values from other rules?

Or, possibly, just do the mirror check here instead of using another loop.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
+			}
+		}
+	}
+	// Go through the list of mirrors and check that at least one of the payload mirrors is not blocked
+	for _, mirror := range payloadMirrors {
+		if !common.InSlice(mirror, blockedRegs) {
```
mtrmac, line 548: Shouldn’t this also be a subscope check?

Actually, to be extra correct, it should use `sysregistriesv2.Registry.PullSourcesFromReference` to remap `releaseRepo`, and then check whether that result is blocked — consider a mirror (`quay.io` → `mirror.example.com`), with a `blockedRegs` entry for `mirror.example.com/openshfit-release-dev` (not an exact match for the host name, not an exact match for the repo, a parent repo of the release repo mirror). 

(That would only be possible when integrated in the first loop, where `Source` is available.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
 		}
 		// Was not a match to the registry being used by the payload, so add to valid blocked registries
 		blockedRegs = append(blockedRegs, reg)
 	}
 	return blockedRegs, nil
 }
 
+// payloadRegHasMirrorConfig returns true if the payload registry has mirror rules configured for it
+func payloadRegHasMirrorConfig(payloadReg string, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy, blockedRegs []string) bool {
+	if len(icspRules) == 0 || icspRules == nil {
```
mtrmac, line 534: This seems to be handled correctly by the general code below, and doesn’t need to be special-cased. (Also `icspRules == nil` implies `len(icspRules) == 0`.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -511,15 +513,45 @@ func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec)
...
-			blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
-			return blockedRegs, fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", payloadReg)
+			// If the payload registry doesn't have mirror rules configured for it, then don't add it to the blocked registries list
+			// TODO maybe? Add an opt-in configuration to ICSP and check if that is set as well
+			if !payloadRegHasMirrorConfig(payloadReg, icspRules, imgSpec.RegistrySources.BlockedRegistries) {
+				blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
+				return blockedRegs, fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", payloadReg)
+			}
+			// Log a warning that we are adding the payload registries to the blocked registries list as there are mirror rules for it
+			glog.Warningf("%q is the payload registry, but will add it to the list of blocked registries as there are mirror rules configured for it", reg)
```
mtrmac, line 523: Is this a warning-level event? Everything is working as intended. (OTOH I haven’t taken the time to look up any guidelines about log levels.)

mtrmac (commented):
I can’t shake a feeling that I’m overcomplicating things, looking at all the “registry” variable names everywhere.

OTOH `EditRegistriesConfig` does explicitly document the parameter as `blockedScopes`, and that it supports both wildcards and sub-registry scopes. So I’m taking that as correct.

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: This was if we wanted to add a new field to either the image or ICSP CRD that is an explicit opt-in for blocking the payload registry - so kind of a two level protection, but not sure if that is really needed.

pkg/controller/container-runtime-config/helpers.go

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: So if I understand correctly, calling `sysregistriesv2.Registry.PullSourcesFromReference` on the `releaseRepo` grabs all the possible places that image can be pulled from? so essentially it grabs all the mirrors for that image? 
And you are suggesting we get all the possible pull sources and then check if any of those pull sources are in the list of blocked registries?

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: This warning would show up in the machine-config-controller logs. I would like to log a warning when we block the payload registry so that in the event where the cluster goes down because something is wrong with the mirrors configured etc, the logs can show that the payload registry was blocked and maybe that is why things are breaking.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+	)
 
 	// Get the registry being used by the payload from the releaseImage
 	ref, err := reference.ParseNamed(releaseImage)
 	if err != nil {
 		return nil, errParsingReference
 	}
-	payloadReg := reference.Domain(ref)
-	for i, reg := range imgSpec.RegistrySources.BlockedRegistries {
+	payloadRepo := reference.TrimNamed(ref).String()
```
mtrmac, line 514: (I think this can be just `ref.Name()`. This works fine.)

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+	for _, reg := range imgSpec.RegistrySources.BlockedRegistries {
 		// if there is a match, return all the blocked registries except the one that matched and return an error as well
```
mtrmac, line 516:
```suggestion
		// if there is a match, return all the blocked registries except those that matched and return an error as well
```

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
-			blockedRegs = append(blockedRegs, imgSpec.RegistrySources.BlockedRegistries[i+1:]...)
-			return blockedRegs, fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", payloadReg)
+		if runtimeutils.ScopeIsNestedInsideScope(payloadRepo, reg) {
+			// If the payload registry doesn't have mirror rules configured for it, then don't add it to the blocked registries list
+			if !payloadRegHasMirrorConfig(payloadRepo, icspRules, imgSpec.RegistrySources.BlockedRegistries) {
+				blockErr = append(blockErr, reg)
+				continue
+			}
+			// Log a warning that we are adding the payload registries to the blocked registries list as there are mirror rules for it
+			glog.Warningf("%q is the payload registry, but will add it to the list of blocked registries as there are mirror rules configured for it", reg)
```
mtrmac, line 524:
```suggestion
			glog.Warningf("%q matches the payload repository, but will add it to the list of blocked registries as there are mirror rules configured for it", reg)
```

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+
 		}
 		// Was not a match to the registry being used by the payload, so add to valid blocked registries
 		blockedRegs = append(blockedRegs, reg)
 	}
-	return blockedRegs, nil
+	var retErr error
+	if len(blockErr) > 0 {
+		retErr = fmt.Errorf("error adding %q to blocked registries, cannot block the registry being used by the payload", blockErr)
```
mtrmac, line 532:
```suggestion
		retErr = fmt.Errorf("error adding %q to blocked registries, cannot block the repository being used by the payload", blockErr)
```

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+		for _, mirror := range rule.Spec.RepositoryDigestMirrors {
+			if runtimeutils.ScopeIsNestedInsideScope(payloadRepo, mirror.Source) {
+				payloadMirrors = append(payloadMirrors, mirror.Mirrors...)
+			}
+		}
+	}
+	// Go through the list of mirrors and check that at least one of the payload mirrors is not blocked
+	for _, mirror := range payloadMirrors {
+		for _, blockReg := range blockedRegs {
+			if !runtimeutils.ScopeIsNestedInsideScope(mirror, blockReg) {
```
mtrmac, line 551: This terminates on the first blocking entry that doesn’t match a mirror — but there might be another blocking entry that matches that mirror. Shouldn’t this succeed only if we find a mirror that his blocked by none of `blockedRegs`?

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,64 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 // getValidBlockedRegistries gets the blocked registries in the image spec and validates that the user is not adding
 // the registry being used by the payload to the list of blocked registries.
 // If the user is, we drop that registry and continue with syncing the registries.conf with the other registry options
-func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec) ([]string, error) {
+func getValidBlockedRegistries(releaseImage string, imgSpec *apicfgv1.ImageSpec, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy) ([]string, error) {
```
mtrmac, line 499: Do we need a similar validation for `allowedRegistries` to ensure that the `payloadRepo` is allowed?

Alternatively, we could hard-code that to be allowed in `policy.json`, as was discussed for `blockedRegistries = ["quay.io"]`; then we wouldn’t quite need to _validate_ `allowedRegistries`.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
mtrmac: (Still outstanding.)

pkg/controller/container-runtime-config/helpers.go
mtrmac: That’s a good point, users using `blockedRegistries: foo` without an ICSP with `neverContactSource` would probably (?) end up with a policy.json entry blocking `foo` — but as long as the cluster is up, they can recover, so I don’t think we _have_ to reject that situation.

Here, in MCO, we can’t really “reject” a configuration, so I think it makes sense to be a bit conservative with validation, so that such semantics problems don’t prevent changing _other_ parts of the registry configuration, and getting to a working cluster.

pkg/controller/container-runtime-config/helpers.go
mtrmac: > So if I understand correctly, calling `sysregistriesv2.Registry.PullSourcesFromReference` on the `releaseRepo` grabs all the possible places that image can be pulled from? so essentially it grabs all the mirrors for that image?

Yes — the complete appropriate repos.

> And you are suggesting we get all the possible pull sources and then check if any of those pull sources are in the list of blocked registries?

I think we might need to, see that example above. A `ScopeIsNestedInsideScope(mirror, blockedReg)` would not find it.

pkg/controller/container-runtime-config/helpers.go
mtrmac: That’s a good point.

mtrmac (commented):
Looks very good.

---

Recording some points discussed separately, so that we have something in public GitHub (thanks for thinking about this!):

It’s not clear that it’s in scope of _this_ PR, but eventually MCO will need to compute _different_ lists of registries to block for `registries.conf` and for `policy.json`.
  - Eventually that would be driven by ICSP’s `neverContactSource`
  - We have discussed (for now, until `neverContactSource` arrives; and also for extra resiliency) hard-coding that the `payloadRepo` is allowed in `policy.json` (even if some parent namespace is in `blockedRegistries` — and perhaps there is nothing matching in `allowedRegistries`).

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -406,3 +407,100 @@ func TestValidateRegistriesConfScopes(t *testing.T) {
...
+			},
+			expectedRegistriesBlocked: []string{"block.io", "payload-reg.io", "block-2.io"},
+			expectedPolicyBlocked:     []string{"block.io", "block-2.io"},
+			expectedErr:               false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotRegistries, gotPolicy, err := getValidBlockedRegistries(tt.releaseImg, tt.imgSpec, tt.icspRules)
+			if err != nil && !tt.expectedErr {
```
mtrmac, line 498: This should also fail on `tt.expectedErr && err == nil`.

mtrmac (commented):
> I don't think adding the payload repo to be always allowed in policy.json would work here as we are only allowed to set either `allowedRegistries` or `blockedRegistries` and setting that to always allowed when the user blocks it means that they won't be able to add any other registries to the blocked list.

That’s an OpenShift semantic restriction, not, AFAICT, a property of `policy.json`.

`policy.json` can be set to have
```js
"": { "type": "insecureAcceptAnything" }, // blockedRegistries user input: default allow
"user-blocked.com": { "type: reject" },
"payload-reg.io/payload-repo":  { "type": "insecureAcceptAnything" }
```
as well as
```js
"": { "type": "reject" }, // allowedRegistries user input: default allow
"user-allowed": { "type: insecureAcceptAnything" },
"payload-reg.io/payload-repo":  { "type": "insecureAcceptAnything" }
```
(as long as there are no duplicate keys)

Looking at the tests, it seems what we might actually have to do that

mtrmac:
> Looking at the tests, it seems what we might actually have to do that

That’s https://github.com/openshift/machine-config-operator/pull/2819#discussion_r838767606 .

mtrmac:
(Note to self: https://github.com/openshift/machine-config-operator/pull/2819#discussion_r819976018 is outstanding from an earlier review.)

pkg/controller/container-runtime-config/helpers.go
umohnani8: @mtrmac are you saying we should always add the payloadRepo to the allowed list even when the user isn't trying to block it?

mtrmac: The first suggestion is essentially a new feature: we now validate `blockedRegistries` to make sure it does not block `payloadRepo`. But if the user sets up very restrictive `allowedRegistries` that does not allow `payloadRepo`, which is, in effect, the same kind of failure, the MCO does not complain.

I no longer think we strictly need that new feature:


The “alternatively” train of thought ultimately leads to https://github.com/openshift/machine-config-operator/pull/2819#discussion_r838767606 : the concept of validating `blockedRegistries` and _completely dropping_ entries that block `payloadRepo` is not quite what we want — we want to specially allow `payloadRepo` _but not more_. If we do that by ~hard-coding that {in the `policy.json` generation code, by submitting appropriate configuration to the `policy.json` generation code}, it seems possible (but I didn’t actually write the code), that the implementation could with very little extra effort handle the `allowedRegistries` configuration as well, making it unnecessary to specifically _validate_ `allowedRegistries`.

cgwalters:
I only gave this a superficial skim, but
/approve
of the general idea.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,69 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+		// if there is a match, return all the blocked registries except those that matched and return an error as well
+		if runtimeutils.ScopeIsNestedInsideScope(payloadRepo, reg) {
+			// If the payload registry doesn't have mirror rules configured for it, then don't add it to the blocked registries list
+			if !payloadRegHasMirrorConfig(payloadRepo, icspRules, imgSpec.RegistrySources.BlockedRegistries) {
+				blockErr = append(blockErr, reg)
+				continue
+			}
+			// Log a warning that we are adding the payload registries to the blocked registries list as there are mirror rules for it
+			glog.Warningf("%q matches the payload repository, but will add it to the list of blocked registries as there are mirror rules configured for it", reg)
+			registriesBlocked = append(registriesBlocked, reg)
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -367,10 +367,10 @@ func verifyRegistriesConfigAndPolicyJSONContents(t *testing.T, mc *mcfgv1.Machin
 	// This is not testing updateRegistriesConfig, which has its own tests; this verifies the created object contains the expected
 	// configuration file.
 	// First get the valid blocked registries to ensure we don't block the registry where the release image is from
-	blockedRegistries, _ := getValidBlockedRegistries(releaseImageReg, &imgcfg.Spec)
+	regisriesBlocked, policyBlocked, _ := getValidBlockedRegistries(releaseImageReg, &imgcfg.Spec, icsps)
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -495,29 +496,69 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
 // getValidBlockedRegistries gets the blocked registries in the image spec and validates that the user is not adding
 // the registry being used by the payload to the list of blocked registries.
 // If the user is, we drop that registry and continue with syncing the registries.conf with the other registry options
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers_test.go
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -830,30 +831,86 @@ func TestRegistriesValidation(t *testing.T) {
...
+				{
+					Spec: apioperatorsv1alpha1.ImageContentSourcePolicySpec{
+						RepositoryDigestMirrors: []apioperatorsv1alpha1.RepositoryDigestMirrors{
+							{Source: "blah.io/payload", Mirrors: []string{"mirror-1.io/payload", "mirror-2.io/payload"}},
+						},
+					},
+				},
+			},
+			expectedRegistriesBlocked: []string{"blah.io", "docker.io"},
+			expectedPolicyBlocked:     []string{"docker.io"},
```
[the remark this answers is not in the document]
umohnani8: I have added a check where if the payload repo is `blah.io/payload` and if the block entry is `blah.io/other-repo`, then we add `blah.io/other-repo` to the blocked list for policy.json as well. The payload repo is being added to the allowed list now. PTAL

pkg/controller/container-runtime-config/helpers.go
umohnani8: Taking this into account I have updated the code to allow the payload repo if we detect there is an attempt to block it. PTAL at the update code

umohnani8:
@mtrmac have made updates based on the reviews, PTAL

pkg/controller/container-runtime-config/helpers.go
umohnani8: Fixed. Using PullSourcesFromReference now, PTAL

umohnani8:
@mtrmac Updated the code to use `PullSourcesFromReference` as suggested above. PTAL

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -830,30 +832,88 @@ func TestRegistriesValidation(t *testing.T) {
...
+			expectedAllowed:           []string{"blah.io/payload/myimage"},
 		},
 	}
 
 	// Failure Tests
 	for _, test := range failureTests {
 		imgcfg := newImageConfig(test.name, test.config)
-		cvcfg := newClusterVersionConfig("version", "blah.io/myuser/myimage:test")
-		blocked, err := getValidBlockedRegistries(cvcfg.Status.Desired.Image, &imgcfg.Spec)
+		cvcfg := newClusterVersionConfig("version", "blah.io/payload/myimage:test")
```
mtrmac, line 911: These tests (really, all callers of `newClusterVersionConfig`) should now use digested references to be representative; otherwise `PullSourcesFromReference`’s tagged/digested configuration would interpret the configuration in ways that don’t match production.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -830,30 +832,88 @@ func TestRegistriesValidation(t *testing.T) {
...
+			},
+			icspRules: []*apioperatorsv1alpha1.ImageContentSourcePolicy{
+				{
+					Spec: apioperatorsv1alpha1.ImageContentSourcePolicySpec{
+						RepositoryDigestMirrors: []apioperatorsv1alpha1.RepositoryDigestMirrors{
+							{Source: "blah.io", Mirrors: []string{"mirror-1.io", "mirror-2.io"}},
+						},
+					},
+				},
+			},
```
mtrmac, line 901: Oh… this is a problematic case.

`registries.conf` only uses the `sysregistriesv2.Registry` entry with the _closest matching_ prefix. So, `BlockedRegistries: blah.io/payload` creates a `Registry` entry which “shadows” the `blah.io` entry for ICSP, effectively completely disabling the mirror configuration for the `blah.io` namespace.

OTOH OpenShift’s `BlockedRegistries` and ICSP are not defined to interact in any way. That’s a bug in runtime-utils: it can propagate blocked/insecure to nested sub-scopes, but it can’t propagate mirrors to nested sub-scopes.

For the purpose of _this_ PR, I think changing the three scopes in the mirror to add `/payload` (resulting in one `Registry` entry that carries both the blocked flat and the mirrors) would work. Separately, runtime-utils will need fixing…

pkg/controller/container-runtime-config/helpers.go
mtrmac: This works correctly for the validation of `blockedRegistries`.

It doesn’t enforce payload repo presence in `allowedRegistries`, but that would really be a new feature… let’s not increase the scope of this PR again.

mtrmac (commented):
I apologize for the terribly late response.

Two of the review comments are two-part, linked with (A)…(A) (B)…(B) markers.

Note primarily the (B) part and the “problematic case” in tests — they are linked and fixing (B) seems to mean mostly removing code.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -701,29 +701,30 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
+	// Find all ImageContentSourcePolicy objects
+	icspRules, err := ctrl.icspLister.List(labels.Everything())
+	if err != nil && errors.IsNotFound(err) {
+		icspRules = []*apioperatorsv1alpha1.ImageContentSourcePolicy{}
+	} else if err != nil {
+		return err
+	}
+
+	var registriesBlocked, policyBlocked, allowedRegs []string
 	if clusterVersionCfg != nil {
```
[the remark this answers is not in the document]
mtrmac: The move of `allowedRegs = append(…)` would be the easiest way to do the `allowedRegs` validation mentioned in https://github.com/openshift/machine-config-operator/pull/2819#discussion_r869696456 , so maybe let’s do it at that point.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
mtrmac: … Actually… if `blah.io/$repo` is blocked in registries.conf, it it _also_ blocked in `policy.json`, so the set of mirrors configured for blah.io/$repo` currently doesn’t matter at all. The only current exception is payloadRepo itself, which we treat specially. So this bug is not nearly that severe. (That will change with ITMS/IDMS and their `neverContactSource` option.)

mtrmac: … correcting myself one more time, even with `neverContactSource` this is not relevant, because `neverContactSource` would set up the mirror and the blocked bit on the _same_ `Registry` object. If there is a mirror configuration for $superScope, then any $subScope is either configuring a different mirror set (= no need to inherit the mirrors), or it is just adding a blocked bit (= no point in inheriting the mirrors).

So… this really only matters for, basically `BlockedRegistries: quay.io`, doesn’t it? But in that case we do need to propagate the mirrors… Let me take a stab at that.

mtrmac: https://github.com/openshift/runtime-utils/pull/16 should help.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,141 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+				}
+			}
+		}
+		// Grab the mirrors for each pull source
+		payloadMirrors = append(payloadMirrors, getMirrors(s.Reference.Name(), icspRules)...)
+	}
+
+	var isUsable bool
+	// Go through the list of mirrors and check that at least one of the payload mirrors is not blocked
+	for _, mirror := range payloadMirrors {
```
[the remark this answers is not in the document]
mtrmac: … this could be _even more_ c/image-like as
```go
reg2 := sysregistriesv2.FindRegistry(ourConfig, mirror);
isUsable :=  reg2 == nil || !reg2.Blocked
```
but AFAICS the two should be equivalent.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -701,29 +701,30 @@ func (ctrl *Controller) syncImageConfig(key string) error {
...
-		return fmt.Errorf("could not get ControllerConfig: %w", err)
-	}
-
-	// Find all ImageContentSourcePolicy objects
-	icspRules, err := ctrl.icspLister.List(labels.Everything())
-	if err != nil && errors.IsNotFound(err) {
-		icspRules = []*apioperatorsv1alpha1.ImageContentSourcePolicy{}
-	} else if err != nil {
-		return err
+		return fmt.Errorf("could not get ControllerConfig %v", err)
```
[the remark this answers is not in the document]
umohnani8: nope, reverted.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -745,8 +746,8 @@ func (ctrl *Controller) syncImageConfig(key string) error {
 			return err
 		}
 		if err := retry.RetryOnConflict(updateBackoff, func() error {
-			registriesIgn, err := registriesConfigIgnition(ctrl.templatesDir, controllerConfig, role,
-				imgcfg.Spec.RegistrySources.InsecureRegistries, blockedRegs, imgcfg.Spec.RegistrySources.AllowedRegistries,
+			registriesIgn, err := registriesConfigIgnition(ctrl.templatesDir, controllerConfig, role, clusterVersionCfg.Status.Desired.Image,
```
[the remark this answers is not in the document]
umohnani8: Yes, moved things around to the above nil check

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
umohnani8: There was a reason to put this check in place back then, I am unable to recall why. Leaving this check in place and added a new releaseImage variable that is set when clusterVersionCfg is not nil. So there should be no failures in the event it is nil.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
```diff
@@ -818,31 +819,31 @@ func registriesConfigIgnition(templateDir string, controllerConfig *mcfgv1.Contr
 	// Generate the original registries config
 	_, originalRegistriesIgn, originalPolicyIgn, err := generateOriginalContainerRuntimeConfigs(templateDir, controllerConfig, role)
 	if err != nil {
-		return nil, fmt.Errorf("could not generate origin ContainerRuntime Configs: %w", err)
+		return nil, fmt.Errorf("could not generate original ContainerRuntime Configs: %v", err)
```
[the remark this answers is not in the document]
umohnani8: nope, reverted

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
```diff
@@ -862,17 +922,16 @@ func TestRegistriesValidation(t *testing.T) {
 
 	// Successful Tests
 	for _, test := range successTests {
+		fmt.Println(test.name)
```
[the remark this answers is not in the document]
umohnani8: was left over from debugging, fixed.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,141 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+		policyBlocked = append(policyBlocked, reg)
+	}
+	if len(blockErr) > 0 {
+		retErr = fmt.Errorf("error adding %q to blocked registries, cannot block the repository being used by the payload", blockErr)
+	}
+	return registriesBlocked, policyBlocked, allowed, retErr
+}
+
+// payloadRegHasMirrorConfig returns true if the payload registry has mirror rules configured for it
+func payloadRegHasMirrorConfig(payloadRepo reference.Named, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy, imgSpec *apicfgv1.ImageSpec) (bool, error) {
```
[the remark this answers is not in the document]

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,141 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+		return false, err
+	}
+	sources, err := r.PullSourcesFromReference(payloadRepo)
+	if err != nil {
+		return false, err
+	}
+
+	// Go through the pull sources and grab their mirrors
+	// Also check if any of them are in the blocked list
+	for _, s := range sources {
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,141 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+		for _, blockReg := range imgSpec.RegistrySources.BlockedRegistries {
+			if s.Reference.Name() == blockReg {
+				payloadMirrors = getMirrors(s.Reference.Name(), icspRules)
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,141 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+				if len(payloadMirrors) == 0 {
+					return false, nil
+				}
+			}
+		}
+		// Grab the mirrors for each pull source
+		payloadMirrors = append(payloadMirrors, getMirrors(s.Reference.Name(), icspRules)...)
+	}
+
+	var isUsable bool
```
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -271,12 +275,68 @@ func TestUpdatePolicyJSON(t *testing.T) {
...
+			allowed:       []string{"allow.io"},
+			errorExpected: true,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := updatePolicyJSON(templateBytes, tt.blocked, tt.allowed)
+			got, err := updatePolicyJSON(templateBytes, tt.blocked, tt.allowed, "release-reg.io/image/release")
 			if err != nil {
+				if tt.errorExpected {
```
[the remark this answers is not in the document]
umohnani8: No, that is if we expect a test to fail, so err will not be nil and we are just checking if that was expected here

pkg/controller/container-runtime-config/helpers_test.go
```diff
@@ -406,3 +466,144 @@ func TestValidateRegistriesConfScopes(t *testing.T) {
...
+			imgSpec: &apicfgv1.ImageSpec{
+				RegistrySources: apicfgv1.RegistrySources{
+					BlockedRegistries: []string{"block.io", "block-2.io"},
+				},
+			},
+			icspRules: []*apioperatorsv1alpha1.ImageContentSourcePolicy{
+				{
+					Spec: apioperatorsv1alpha1.ImageContentSourcePolicySpec{
+						RepositoryDigestMirrors: []apioperatorsv1alpha1.RepositoryDigestMirrors{
+							{Source: "src.io/payload", Mirrors: []string{"mirror-1.io/payload", "mirror-2.io/payload"}},
```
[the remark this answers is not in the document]
umohnani8: Will fix in a follow up PR

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
umohnani8: Yeah, I started hitting this when I fixed up the code based on comments. I am still a little confused why this is happening, basically `PullSourcesFromReference` is unable to detect that `mirror-1.io/payload/image` is a mirror of the release image. 
So what exactly in runtime-utils needs fixing for this?

I have added `/payload` to fix the test for now.

umohnani8:
@mtrmac addressed all the comments. Have a little confusion on the last review, PTAL.

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
mtrmac: - I.e. `releaseImage` can now be `""`? Worrying about maintenance, I’d like the affected functions (`registriesConfigIgnition` and `updatePolicyJSON` to document that special case.)
- I’m not sure this actually works right now — `updatePolicyJSON`, if it needs `releaseImage`, calls `getPayloadRepo`, and that AFAICS fails on an empty string.

pkg/controller/container-runtime-config/helpers.go
```diff
@@ -478,29 +489,124 @@ func validateUserContainerRuntimeConfig(cfg *mcfgv1.ContainerRuntimeConfig) erro
...
+	tmpFile, err := os.CreateTemp("", "regtemp")
+	if err != nil {
+		return nil, err
+	}
+	_, err = tmpFile.Write(newData.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	defer tmpFile.Close()
```
mtrmac, line 598: This should happen just after `os.CreateTemp` succeeds, so that it is called if `Write` fails.

pkg/controller/container-runtime-config/helpers_test.go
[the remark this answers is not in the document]
mtrmac: Yeah, but if we expect the test to fail, and it didn’t, that’s at least unexpected, and potentially bad.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
mtrmac: The test caused generation of `registries.conf` roughly like this:
```toml
[[registry]]
location = "blah.io/payload"
blocked = true

[[registry]]
location = "blah.io"
mirrors = …
```

The way c/image works, only the most specific match matters, i.e. the `blah.io/payload` entry shadows the `blah.io` entry. So, given a config like this, `payloadRepoHasUnblockedMirror("blah.io/payload")` _correctly_ mirrors c/image behavior and finds no mirrors that apply to `blah.io`.

Does that help clarify, or is there something else unexpected going on that I’m missing?

I think https://github.com/openshift/runtime-utils/pull/16 fixes the underlying cause (of not generating `registries.conf` that correctly interprets the provided BlockedRegistries + ICSP), at least per its unit tests — but I didn’t actually go back to this test to verify that it now behaves as expected.

---

This problem manifests for configurations where blocking is nested inside mirrors, e.g. (`quay.io` has mirrors configured, `quay.io/openshift-release-dev` is blocked); those are technically legal, but I’d expect them to be much rarer than nesting mirrors inside blocking, e.g. (all of `quay.io` is blocked, `quay.io/openshift-release-dev` has mirrors).

So… I think this is a bug that we definitely need to fix, but I’m not at this point sure that it needs to block _this_ PR. (Also noting that I have had to re-evaluate my analysis two times, so it’s fairly likely I’m wrong again. I’d definitely appreciate a careful independent look.)

pkg/controller/container-runtime-config/helpers.go
umohnani8: fixed

pkg/controller/container-runtime-config/helpers_test.go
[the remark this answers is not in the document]
umohnani8: fixed

pkg/controller/container-runtime-config/container_runtime_config_controller.go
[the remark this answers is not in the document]
mtrmac: Sure, qualifying the documentation that `""` can happen with saying that it is very unlikely and broken would be useful.

WRT `updatePolicyJSON` completely failing in that case, I was worried that that might prevent error recovery … but, on second thought, in that situation the etcd database and the API server must be mostly accessible, and the fix for the missing `clusterVersionCfg` would essentially be database-level, so MCC failing wouldn’t prevent fixing that — so ultimately I’m fine with not adding a special case and just letting it fail.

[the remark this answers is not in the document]
umohnani8: Added a comment accordingly.

pkg/controller/container-runtime-config/container_runtime_config_controller_test.go
mtrmac: > but I didn’t actually go back to this test to verify that it now behaves as expected.

Confirming that the runtime-utils PR allows this test, as it was originally, to pass.

umohnani8: Awesome! I will pull it in here and fix up the test case once it is merged.

